PR #828, opened 2023-05-28
default transformer dropout to 0

JudeDavis1:
The reason for 0 dropout (at the pretraining/causal lm stage is that on limited hardware, you want the model to learn the dataset patterns quickly first, then use dropout at fine-tuning. Because dropout can slow down accuracy at the lm stage. So optional dropout may allow features to be quickly learned first.


geohot:
Don't change behavior, but I'm okay with that being a parameter.

JudeDavis1:
Ok, can change default to 0.1.

JudeDavis1:
@geohot done

PR #830, opened 2023-05-28
Tensor.where method

python273:


test/test_ops.py
```diff
@@ -66,13 +66,13 @@ def test_where(self):
     helper_test_op(
       [(100,)],
       lambda x: torch.where(x > 0.5, 4, 2),
-      lambda x: Tensor.where(x > 0.5, 4, 2), forward_only=True)
+      lambda x: (x > 0.5).where(4, 2), forward_only=True)
```
geohot, line 69: The key of removing the staticmethod is both version should work.

python273: yea, both versions work. Do you want an explicit test for `Tensor.where(x>0.5, 4, 2)`?

python273: linter doesn't like non-`self` first arg, if that's what you mean

geohot: Eh, it's all good.

PR #831, opened 2023-05-28
Support tensor type casting while loading weights

kunwar31:
Solves float16 inference, part of https://github.com/geohot/tinygrad/issues/829 
Changes made:
1. Allow casting weights to a specific dtype in `fake_torch_load_zipped`, one layer at a time.  

> 2. Setting default_type for a `Tensor` also sets dtype in `arange`, `rand`, `randn`, `dropout`.
> 3. HALF env var controls default_type for a Tensor and allows `fake_torch_load_zipped` to cast weights to half, one layer at a time.  

Edit: 2, 3 are removed as suggested @geohot

### DIFF
--- a/extra/utils.py
+++ b/extra/utils.py
@@ -3,7 +3,7 @@ import numpy as np
 from tqdm import tqdm
 import tempfile, platform
 from collections import defaultdict
-from tinygrad.helpers import prod, getenv, DEBUG, dtypes, DType
+from tinygrad.helpers import prod, getenv, DEBUG
 from tinygrad.ops import GlobalCounters
 from tinygrad.tensor import Tensor
 from tinygrad.lazy import LazyNumpyArray, Device
--- a/extra/utils.py
+++ b/extra/utils.py
@@ -3,7 +3,7 @@ import numpy as np
 from tqdm import tqdm
 import tempfile, platform
 from collections import defaultdict
-from tinygrad.helpers import prod, getenv, DEBUG, HALF, dtypes
+from tinygrad.helpers import prod, getenv, DEBUG, dtypes, DType
 from tinygrad.ops import GlobalCounters
 from tinygrad.tensor import Tensor
 from tinygrad.lazy import LazyNumpyArray, Device
@@ -128,14 +128,14 @@ def create_casting_function(prev_func, dtype):
     return lna.astype(dtype)
   return casting_function
 
-def post_process(t: Tensor):
+def post_process(t: Tensor, load_dtype: DType):
   if t is not None:
-    if HALF:
-      t.lazydata.op.arg.fxn = create_casting_function(t.lazydata.op.arg.fxn, dtype=np.float16)
-      t.lazydata = t.lazydata.cast(dtypes.float16)
+    if load_dtype != t.dtype:
+      t.lazydata.op.arg.fxn = create_casting_function(t.lazydata.op.arg.fxn, dtype=load_dtype.np)
+      t.lazydata = t.lazydata.cast(load_dtype)
     t.realize()
 
-def fake_torch_load_zipped(fb0, load_weights=True, multithreaded=True):
+def fake_torch_load_zipped(fb0, load_weights=True, multithreaded=True, load_dtype=dtypes.float32):
   if Device.DEFAULT in ["TORCH", "GPU", "CUDA"]: multithreaded = False  # multithreaded doesn't work with CUDA or TORCH. for GPU it's a wash with _mmap
 
   import zipfile
@@ -148,7 +148,7 @@ def fake_torch_load_zipped(fb0, load_weights=True, multithreaded=True):
         with myzip.open(f'{base_name}/data/{k}') as myfile:
           for v in vv:
             load_single_weight(v[2], myfile, v[3], v[4], v[0], v[5], mmap_allowed=True)
-            post_process(v[2])
+            post_process(v[2], load_dtype=load_dtype)
       if multithreaded:
         import concurrent.futures
         # 2 seems fastest
--- a/extra/utils.py
+++ b/extra/utils.py
@@ -79,8 +79,8 @@ def my_unpickle(fb0):
   return MyPickle(fb0).load(), key_prelookup
 
 def post_process(t: Tensor, load_dtype):
-    if load_dtype is not None and load_dtype != t.dtype: t.lazydata = t.lazydata.cast(load_dtype)
-    if not t.lazydata.realized: t.realize()
+  if load_dtype is not None and load_dtype != t.dtype: t.lazydata = t.lazydata.cast(load_dtype)
+  if not t.lazydata.realized: t.realize()
 
 def load_single_weight(t:Tensor, myfile, shape, strides, dtype, storage_offset, mmap_allowed=False):
   bytes_size = np.dtype(dtype).itemsize
--- a/extra/utils.py
+++ b/extra/utils.py
@@ -79,7 +79,6 @@ def my_unpickle(fb0):
   return MyPickle(fb0).load(), key_prelookup
 
 def post_process(t: Tensor, load_dtype: DType):
-  if t is not None:
     if load_dtype != t.dtype: t.lazydata = t.lazydata.cast(load_dtype)
     if not t.lazydata.realized: t.realize()
 
--- a/extra/utils.py
+++ b/extra/utils.py
@@ -78,7 +78,15 @@ def my_unpickle(fb0):
 
   return MyPickle(fb0).load(), key_prelookup
 
-def load_single_weight(t:Tensor, myfile, shape, strides, dtype, storage_offset, mmap_allowed=False):
+def post_process(t: Tensor, load_dtype: DType):
+  if t is not None:
+    if load_dtype != t.dtype:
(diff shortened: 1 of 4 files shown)


python273:
imho seems way too hardcoded for halfs

In weights loading code, instead of casts/args scattered all over, would be nicer to find a single place that would call a passed function for postprocessing, like after each `load_single_weight` call or something

kunwar31:
@python273 Good point. Added the change, and moved tensor.realize() from load_single_weight to post_process, along with the typecasts to HALF

tinygrad/helpers.py
```diff
@@ -27,7 +27,12 @@ def __call__(self, x): self.value = x
   def __bool__(self): return self.value != 0
   def __ge__(self, x): return self.value >= x
 
-DEBUG, IMAGE = DebugSingleton(), getenv("IMAGE", 0)
+class HalfSingleton:
```
geohot, line 30: Definitely not.

kunwar31: the class, env var or both?

geohot: Both. If you want this sort of stuff, it has to follow normal broadcasting rules.

tinygrad/tensor.py
```diff
@@ -512,7 +514,7 @@ def batchnorm(self, weight:Optional[Tensor], bias:Optional[Tensor], mean:Tensor,
   def dropout(self, p=0.5) -> Tensor:
     if not Tensor.training: return self
     # TODO: why is this going through numpy?
-    _mask: np.ndarray = np.asarray(Tensor._rng.binomial(1, 1.0-p, size=self.shape), dtype=np.float32)
+    _mask: np.ndarray = np.asarray(Tensor._rng.binomial(1, 1.0-p, size=self.shape), dtype=Tensor.default_type.np)
```
geohot, line 517: This should be refactored to not use numpy directly.

tinygrad/tensor.py
```diff
@@ -144,12 +144,14 @@ def arange(stop, start=0, step=1, **kwargs): return Tensor(np.arange(start=start
   def manual_seed(seed=None): Tensor._rng = np.random.default_rng(seed=seed)
 
   @staticmethod
-  def rand(*shape, **kwargs) -> Tensor: return Tensor(LazyNumpyArray(lambda lna: Tensor._rng.random(size=lna.shape, dtype=lna.dtype), shape, np.float32), **kwargs)
+  def rand(*shape, **kwargs) -> Tensor: return Tensor(LazyNumpyArray(lambda lna: Tensor._rng.random(size=lna.shape, dtype=lna.dtype), shape, Tensor.default_type.np), **kwargs)
```
geohot, line 147: This should come from the kwargs.

tinygrad/tensor.py
kunwar31: @geohot done

kunwar31: @geohot though now it looks like this could have been a separate PR

python273:
needs a test `test/extra/test_utils.py`

also `load_single_weight` is called in multiple places, and it won't realize in `fake_torch_load` it seems

kunwar31:
@python273 Added tests for loading in specific dtype.  since load_single_weight is called multiple times, I replaced the t.realise() calls with post_process calls so it should not cause any issues.

Also, I realized that I don't need to manually cast the np array, as lazydata.cast will handle that.

extra/utils.py
```diff
@@ -136,7 +140,7 @@ def fake_torch_load_zipped(fb0, load_weights=True, multithreaded=True):
       def load_weight(k, vv):
         with myzip.open(f'{base_name}/data/{k}') as myfile:
           for v in vv:
-            load_single_weight(v[2], myfile, v[3], v[4], v[0], v[5], mmap_allowed=True)
+            load_single_weight(v[2], myfile, v[3], v[4], v[0], v[5], mmap_allowed=True, load_dtype=load_dtype)
```
python273, line 143: I think would be cleaner to just have 2 calls to `post_process` (after each `load_single_weight`). Here and in 
https://github.com/geohot/tinygrad/blob/0fc4cf72a2261fff7ef432ed9eb9bd003459dcd9/extra/utils.py#L184

extra/utils.py
```diff
@@ -78,7 +78,12 @@ def persistent_load(self, pid):
 
   return MyPickle(fb0).load(), key_prelookup
 
-def load_single_weight(t:Tensor, myfile, shape, strides, dtype, storage_offset, mmap_allowed=False):
+def post_process(t: Tensor, load_dtype: DType):
+    if load_dtype != t.dtype: t.lazydata = t.lazydata.cast(load_dtype)
+    if not t.lazydata.realized: t.realize()
+
+def load_single_weight(t:Tensor, myfile, shape, strides, dtype, storage_offset, mmap_allowed=False, load_dtype=None):
+  if load_dtype is None: load_dtype = dtypes.from_np(dtype)
```
python273, line 86: this check can be in `post_process`

test/extra/test_utils.py
```diff
@@ -37,24 +37,28 @@ def __init__(self):
...
-      for name, a in model.state_dict().items():
-        b = model2[name]
-        a, b = a.numpy(), b.numpy()
-        assert a.shape == b.shape
-        assert a.dtype == b.dtype
-        assert np.array_equal(a, b)
+      for loaded_is_float16, load_dtype in [(False, dtypes.float32), (True, dtypes.float16), (isfloat16, None)]:
+        if isfloat16 == loaded_is_float16:
+          print(f"Saving with float16={isfloat16}")
+          print(f"Loading with dtype={load_dtype}")
```
python273, line 43: as I understand it never checks actual casts like f32 -> f16, but only matching types.

maybe make a separate test function specifically for casts? (torch model can be simpler)

tinygrad/tensor.py
```diff
@@ -511,9 +511,7 @@ def batchnorm(self, weight:Optional[Tensor], bias:Optional[Tensor], mean:Tensor,
 
   def dropout(self, p=0.5) -> Tensor:
     if not Tensor.training: return self
-    # TODO: why is this going through numpy?
-    _mask: np.ndarray = np.asarray(Tensor._rng.binomial(1, 1.0-p, size=self.shape), dtype=np.float32)
-    return self * Tensor(_mask, requires_grad=False, device=self.device) * (1/(1.0 - p))
+    return self * Tensor(Tensor._rng.binomial(1, 1.0-p, size=self.shape), requires_grad=False, device=self.device, dtype=self.dtype) * (1/(1.0 - p))
```
python273, line 514: separate PR would likely get it merged faster

kunwar31:
@python273 Made the reviewed changes, will create a separate PR for dropout change

test/extra/test_utils.py
```diff
@@ -36,25 +36,33 @@ def __init__(self):
...
-        assert a.dtype == b.dtype
-        assert np.array_equal(a, b)
+    for isfloat16_aftersave in [True, False]:
+      for isfloat16_beforesave in [True, False]:
+        for loaded_is_float16, load_dtype in [(False, dtypes.float32), (True, dtypes.float16), (isfloat16_beforesave, None)]:
+          torch_finally_isfloat16 = isfloat16_aftersave or isfloat16_beforesave
+          if torch_finally_isfloat16 == loaded_is_float16:
+            print(f"isfloat16_aftersave={isfloat16_aftersave}")
+            print(f"isfloat16_beforesave={isfloat16_beforesave}")
+            print(f"load_dtype={load_dtype}")
```
python273, line 46: so complicated and has redundant cases when you half model before and after

can just hardcode cases, would be more readable
```python
    cases = [
      # model is half, loaded dtype
      [1, dtypes.float16],
      [1, None],
      [0, dtypes.float32],
      [0, dtypes.float16],
      [0, None],
    ]
    
    # instead of
    # if isf16_after: model = model.half()
    if load_dtype == dtypes.float16: model = model.half()
```

extra/utils.py
```diff
@@ -78,6 +78,11 @@ def persistent_load(self, pid):
 
   return MyPickle(fb0).load(), key_prelookup
 
+def post_process(t: Tensor, stored_dtype: DType, load_dtype: DType):
+    if load_dtype is None: load_dtype = dtypes.from_np(stored_dtype)
+    if load_dtype != t.dtype: t.lazydata = t.lazydata.cast(load_dtype)
```
python273, line 83: you don't need `stored_dtype`, I think can be just
```python
if load_dtype is not None and load_dtype != t.dtype:
```

python273: and load_dtype is Optional (check mypy)

python273:
ok, nice and clean now 😅 

@geohot

extra/utils.py
```diff
@@ -78,6 +78,10 @@ def persistent_load(self, pid):
 
   return MyPickle(fb0).load(), key_prelookup
 
+def post_process(t: Tensor, load_dtype):
+    if load_dtype is not None and load_dtype != t.dtype: t.lazydata = t.lazydata.cast(load_dtype)
+    if not t.lazydata.realized: t.realize()
```
wozeparrot, line 83: Should be 2 spaces here.

geohot:
All of extra/utils is deprecated now and replaced with disktensors.

PR #832, opened 2023-05-28
ONNX tests: exclude all unsupported filetype tests

SnakeOnex:
There are ONXX tests for regular ops with a unsupported data types, such as test_max_uint8_cpu (int16, unint8, uint16, uint64, float64, ...). They don't get excluded by the current regex expressions.
I changed the exclusion regex, to include all tests containing these datatypes in their name, although I could write out all the tests in the format of test_\<OP\>_\<TYPE\>*, for clarity as well.
PASSED count is same pre & after the change (doesn't exclude any passing tests), but SKIPPED goes 1862 -> 1885.

Is a small change, but the first thing I noticed for getting to 0 FAILED. If some of these data types should work, I could look into making the tests PASS too.

### DIFF
--- a/test/external/external_test_onnx_backend.py
+++ b/test/external/external_test_onnx_backend.py
@@ -53,10 +53,15 @@ backend_test.exclude('test_momentum_*')
 backend_test.exclude('test_eyelike_*')
 
 # we only support float32
-backend_test.exclude('test_add_uint8_*')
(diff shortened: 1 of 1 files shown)


geohot:
Yea, I think getting to 0 failed is a good goal if things are obviously unsupported.

PR #833, opened 2023-05-28
Add max_pool2d dilation (also for ONNX)

marcellofuschi:


geohot:
Cool, 2 more onnx tests passing.

PR #835, opened 2023-05-28
conv1d & conv3d onnx tests

SnakeOnex:
Implemented the ONNX Conv1D as Conv2D, all the 8 Conv1D tests do pass. I detect, that it is a Conv1D, by the fact that the tensors are shape of length 3.

### DIFF
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -90,12 +90,8 @@ def MaxPool(X, kernel_shape, auto_pad="NOTSET", ceil_mode=0, dilations=1, pads=N
 
 def Conv(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_shape=None, pads=None, strides=1):
   if len(X.shape) == 3 and len(W.shape) == 3: # detect conv_1d conv
-    X = X.unsqueeze(3)
-    W = W.unsqueeze(3)
-    if isinstance(dilations, tuple): dilations = dilations[0]
-    if isinstance(strides, tuple): strides = strides[0]
-    out = X.conv2d(W, B, stride=strides, groups=group, dilation=dilations, padding=(0, 0, pads[0], pads[1]) if pads is not None else 0)
-    return out.reshape(out.shape[:-1])
+    return X.conv2d(W, B, stride=strides, groups=group, dilation=dilations, padding=(pads[0], pads[1]) if pads is not None else 0)
+
   return X.conv2d(W, B, stride=strides, groups=group, dilation=dilations, padding=(pads[1], pads[3], pads[0], pads[2]) if pads is not None else 0)
 
 def ConvTranspose(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_shape=None, pads=None, strides=1):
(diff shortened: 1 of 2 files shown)


extra/onnx_ops.py
```diff
@@ -89,6 +89,13 @@ def MaxPool(X, kernel_shape, auto_pad="NOTSET", ceil_mode=0, dilations=1, pads=N
   return _padding(X, pads, auto_pad, constant_value=-np.inf, axes=tuple(range(len(X.shape)))[-2:]).max_pool2d(kernel_shape, stride=strides, dilation=dilations)
 
 def Conv(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_shape=None, pads=None, strides=1):
+  if len(X.shape) == 3 and len(W.shape) == 3: # detect conv_1d conv
```
geohot, line 92: Why doesn't this just work? I get the padding is wrong, but I expect conv2d to just work as a 1d conv the same way it just works as a 3d conv.

geohot: Can you add 1d conv tests to test_ops without onnx?

SnakeOnex: > Why doesn't this just work? I get the padding is wrong, but I expect conv2d to just work as a 1d conv the same way it just works as a 3d conv.

Well, I just assumed you need a (BxCxWxH) tensor for a conv2D and it's simple to turn a B x C x W -> B x C x W x 1. Does the conv2d actually work for conv3d tho? Wouldn't you have reshape your B x C x D x H x W tensor in some complicated way.

As for making it just work, without adding the extra dimension and then taking it away, I will look into it more. But from the first glance, it looks as the way conv2d is implemented, the output shape breaks when there is padding involved and contains negative dimensions.

> Can you add 1d conv tests to test_ops without onnx?

I will look into it.

geohot: conv2d just works for conv2d, yea. See the test_ops.

SnakeOnex:
Okay, so I looked more into, and you were right, it should just work, just like the conv3d. The problem is with the padding. I think a design decision has to be made about the conv2d function. It is called conv2d, but if we want it to be used as conv1d and conv3d, the way padding argument works needs to be changed. 

Specifically, as it is right now, if you get padding (1,1), it is turned into (1,1,1,1), since it is written only with 2D conv in mind. So if you want it to be implemented only as one conv function, I think you would have to enforce a padding length of len(HW.shape) * 2, aka specify it for each dimension fully, to avoid ambiguity. In the last commit, which fails some tests due to this change, I added this enforcing assert and padding generation code. 
I also suspect this would help with fixing the ONNX conv3d tests, which fail on padding too.

I could look into the other tests that are failing and fix them, but since this is a relatively large design change, if you could just give me your input on it first.

geohot:
I'm fine with three types of padding:
1. A single number. It pads both sides with that.
2. A len(HW.shape) tuple. It pads both sides with each number
3. A len(HW.shape)*2 tuple. It pads each side.

The 1d-2d-3d nature of the conv should be decided by the len of the shape of the weight. 3=1d, 4=2d, 5=3d.

geohot:
Sweet, 9 new passing tests!

SnakeOnex:
Yes, it actually made the Conv3D ONNX tests work aswell, I was still planning to implement the Conv1D tests, but I guess I will create a new pull request when I get to them.

PR #836, opened 2023-05-28
Fix cuda

crthilakraj:
Fixed to work on CUDA.

Can be tested using the following commands
```
CUDA=1 pytest test/test_ops.py
```
```
CUDA=1 DEBUG=3 OPTLOCAL=1 python3 -c "from tinygrad.tensor import Tensor;
N = 1024; a, b = Tensor.randn(N, N), Tensor.randn(N, N);
c = (a.reshape(N, 1, N) * b.permute(1,0).reshape(1, N, N)).sum(axis=2);
print((c.numpy() - (a.numpy() @ b.numpy())).mean())"

```

PS: let me know if I should move the header file out of `disassemblers/nvidia/`


geohot:
So instead of this (huge file!), I would rather that the CUDA backend just doesn't emit float4 ALU ops.

crthilakraj:
I don't know any options other than having a self-hosted runner for testing CUDA, since GitHub doesn't support GPU.

geohot:
Don't disable float4 entirely, only the ALU. See my RDNA assembly PR.

PR #837, opened 2023-05-28
int8/uint8 support

wozeparrot:
This was done in the style of the float16 stuff. I tested that this works on GPU and CPU and TORCH.

### TODO:

- [x] Check https://github.com/geohot/tinygrad/blob/master/datasets/preprocess_imagenet.py. need to download imagenet first
- [x] Maybe do cast tests between half and int8/uint8 types

LLVM and CLANG don't work for half for me, so I'm opening this to test with CI.

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -86,5 +86,83 @@ class TestDtype(unittest.TestCase):
     assert c.dtype == dtypes.float32
     np.testing.assert_allclose(c.numpy(), [[1,2],[3,4]])
 
+  def test_int8_to_np(self):
+    a = Tensor([1,2,3,4], dtype=dtypes.int8)
+    print(a)
+    na = a.numpy()
+    print(na, na.dtype, a.lazydata.realized)
+    assert na.dtype == np.int8
+    np.testing.assert_allclose(na, [1,2,3,4])
+
+  def test_int8_add(self):
+    a = Tensor([1,2,3,4], dtype=dtypes.int8)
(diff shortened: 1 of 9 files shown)


geohot:
Hehe I knew this one was too easy. Can you make the preprocess_imagenet dataset work as a test of this?

test/test_dtype.py
```diff
@@ -86,5 +86,168 @@ def test_half_matmul_upcast(self):
     assert c.dtype == dtypes.float32
     np.testing.assert_allclose(c.numpy(), [[1,2],[3,4]])
 
+  def test_int8_to_np(self):
```
geohot, line 89: Is there some function here to prevent copy and paste?

wozeparrot:
yup, this was way easier than I thought it would be :)

Should I make preprocess_imagenet an external test or a normal one fine?

geohot:
Oh just testing locally is fine.

wozeparrot:
huh it seems like preprocess_imagenet just works, unless I'm testing it wrong.

geohot:
Yea...it works, but it has to have uint8 dtype or the file is 4x too big.

wozeparrot:
> Yea...it works, but it has to have uint8 dtype or the file is 4x too big.

I think it does? my file is 1.8gb, not sure what it's supposed to be.

geohot:
xxd it

geohot:
Are there any tests of negative numbers? Are uint8 and int8 actually different?

geohot:
Easy $200! Will send.

PR #839, opened 2023-05-29
Rewrite Adam/W as functions of LAMB

wozeparrot:
Follow up to https://github.com/geohot/tinygrad/pull/821

I'm not sure if we actually want this, it does save quite a few lines but it kind of loses the actual adam implementation, and it "adds" a test for the `adam` flag.

LAMB is essentially just LARS applied to Adam so if we just set the trust ratio to `1.0` its just Adam.

I'm not quite sure whats going on with METAL tho.

### DIFF
--- a/tinygrad/nn/optim.py
+++ b/tinygrad/nn/optim.py
@@ -37,27 +37,9 @@ class SGD(Optimizer):
       t.assign(t.detach() - g * self.lr)
     self.realize(self.b)
 
-class AdamW(Optimizer):
-  def __init__(self, params: List[Tensor], lr=0.001, b1=0.9, b2=0.999, eps=1e-8, wd=0.01):
-    super().__init__(params)
-    # NOTE: self.t is a tensor so Adam can be jitted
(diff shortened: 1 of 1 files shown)


tinygrad/nn/optim.py
```diff
@@ -76,9 +58,12 @@ def step(self) -> None:
       m_hat = self.m[i] / (1.0 - self.b1**self.t)
```
geohot, line 58: Is this why better than the way the Adam does it?

geohot:
I think this is worth it, maybe add a comment explaining this fact. The LAMB part is very isolated.

tinygrad/nn/optim.py
wozeparrot: If I do this the way the Adam implementation does it, it breaks for LAMB

geohot: Meh, this way I find more readable anyway, and it's a wash in the JIT.

PR #842, opened 2023-05-29
use tensor dtype for zeros_like()

ubihot:
- [x] need to add tests

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -156,33 +156,11 @@ class TestTinygrad(unittest.TestCase):
         np.testing.assert_allclose(a.numpy(), b.numpy())
 
   def test_zeros_like_has_same_dtype(self):
-    a = Tensor([1, 2, 3])
-    b = Tensor.zeros_like(a)
-    self.assertTrue(a.dtype == b.dtype)
-
-    a = Tensor([1, 2, 3], dtype=dtypes.float16)
-    b = Tensor.zeros_like(a)
-    self.assertTrue(a.dtype == b.dtype)
-
-    a = Tensor([1, 2, 3], dtype=dtypes.float32)
-    b = Tensor.zeros_like(a)
-    self.assertTrue(a.dtype == b.dtype)
-
-    a = Tensor([1, 2, 3], dtype=dtypes.int8)
-    b = Tensor.zeros_like(a)
-    self.assertTrue(a.dtype == b.dtype)
-
-    a = Tensor([1, 2, 3], dtype=dtypes.int32)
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -128,7 +128,7 @@ def zeros(*shape, **kwargs): return Tensor([0], **kwargs).reshape([1]*len(shape)
   def ones(*shape, **kwargs): return Tensor([1], **kwargs).reshape([1]*len(shape)).expand(shape).contiguous()
 
   @staticmethod
-  def zeros_like(tensor, **kwargs): return Tensor.zeros(*tensor.shape, **kwargs)
+  def zeros_like(tensor, **kwargs): return Tensor.zeros(*tensor.shape, dtype = tensor.dtype, **kwargs)
```
python273, line 131: remove spaces `_=_` -> `=`

test/test_tensor.py
```diff
@@ -154,5 +155,34 @@ def test_random_fns_are_deterministic_with_seed(self):
...
         np.testing.assert_allclose(a.numpy(), b.numpy())
 
+  def test_zeros_like_has_same_dtype(self):
+    a = Tensor([1, 2, 3])
+    b = Tensor.zeros_like(a)
+    self.assertTrue(a.dtype == b.dtype)
+
+    a = Tensor([1, 2, 3], dtype=dtypes.float16)
+    b = Tensor.zeros_like(a)
+    self.assertTrue(a.dtype == b.dtype)
```
python273, line 165: I'd say just 2 cases is enough (default and say int8), you can add `assert` to check shapes are the same

test/test_tensor.py
```diff
@@ -154,5 +155,11 @@ def test_random_fns_are_deterministic_with_seed(self):
         b = random_fn(10,10).realize()
         np.testing.assert_allclose(a.numpy(), b.numpy())
 
+  def test_zeros_like_has_same_dtype(self):
+    for dtype in dtypes.__annotations__:
+      a = Tensor([1, 2, 3], dtype=dtypes.float16)
+      b = Tensor.zeros_like(a)
+      self.assertTrue(a.dtype == b.dtype)
```
python273, line 162: remove loop, don't really need to test every dtype (and currently you have a bug, `dtype` var is unused)

just have 2 cases and add shape assert

test/test_tensor.py
ubihot: @python273 updated the test, let me know if there's a better way of doing this

geohot:
Thanks for the fix.

PR #844, opened 2023-05-29
stack based context manager for DEBUG and IMAGE

leopf:
Supports

```python

a = Tensor.ones(1)
b = Tensor.ones(1)

# DEBUG=getenv("DEBUG")

with Context(DEBUG=4):
    (a+b).numpy() # DEBUG=4

# DEBUG=getenv("DEBUG") again
```

This should be compatible with the current way of handling these vars. `DEBUG(x)` still works.

Could be useful for Device.DEFAULT as well.

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -20,14 +20,26 @@ def mnum(i) -> str: return str(i) if i >= 0 else f"m{-i}"
 @functools.lru_cache(maxsize=None)
 def getenv(key, default=0): return type(default)(os.getenv(key, default))
 
-# NOTE: hack to allow DEBUG(x) to set the debug value at runtime
-class DebugSingleton:
-  def __init__(self): self.value = getenv("DEBUG", 0)
-  def __call__(self, x): self.value = x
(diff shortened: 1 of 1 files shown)


fikisipi:
> Could be useful for Device.DEFAULT as well.

Oof, hard agree! I hardcode that one all the time when I'm working on a single runtime and env variables become a mess.

geohot:
I like it! Do `context_defaults` and `context_stack` have to be global? Also the names are repeated twice.

leopf:
This should do it! Globals removed, names only used once.

The only constraint is that all context vars have to be registered before a context is entered (len(ctx_stack) > 1).

leopf:
Sorry for the commit spam!
I wanted the ContextVar to be useful enough to replace getenv.

You can now do:
```python
with Context(TEST=1):
    TEST = ContextVar("TEST", 0)
    assert TEST.value == 1 # will be true
```

geohot:
Yea, it should work for things like training too.

leopf:
Sorry for that!

geohot:
Install the pre-commit hooks on your machine to prevent this.

PR #849, opened 2023-05-29
Update readme

hiteshhedwig:
Hi, @geohot 

I am trying to make readme a bit better. I've done some changes. Although, I still have some more changes to do (like creating table of contents, some formatting and stuffs). But i wanted to have input from you (& hopefully from other contributors as well) before making more changes in my side. That's why i'm creating this PR to understand your feedback and possibly have some of my query resolved about refactoring the readme.

Questions :
1. The changes i've done. What do you think about it ?
2. What direction you'd expect structure of the readme to go into.
3. Generally i've seen there is different readme for contribution, examples. But i am afraid it'll truncate alot reading length of Main readme and make it too short. How much should be the expected length of the readme for now. (the current one seems fine for me, considering  tinygrad will grow then we can refactor it once more)
4. Is refactoring readme even needed right now? (at the current stage)
5. Should i go ahead and make the relevant changes or close the PR ?

Let me know your thoughts  
Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,17 +8,27 @@
 
 [![tinygrad discord](https://discordapp.com/api/guilds/1068976834382925865/widget.png?style=banner2)](https://discord.gg/ZjZadyC7PK)
 
-For something in between a [pytorch](https://github.com/pytorch/pytorch) and a [karpathy/micrograd](https://github.com/karpathy/micrograd)
 
-This may not be the best deep learning framework, but it is a deep learning framework.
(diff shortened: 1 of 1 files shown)


geohot:
So I'm not going to merge a major change to the README from a first time contributor. If you'd like to do README-like stuff, write a very clean tutorial and add it to docs.

hiteshhedwig:
Understandable. Ok !

PR #852, opened 2023-05-29
add Tensor.full and Tensor.full_like and reuse them

ubihot:
Added `Tensor.full` and `Tensor.full_like` and reused them in `zeros, ones, zeros_like, ones_like`

TODO
- need to improve the tests so we don't repeat code

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -131,6 +131,9 @@ class Tensor:
   def zeros_like(tensor, dtype:Optional[DType]=None, **kwargs): return Tensor.zeros(*tensor.shape, dtype=tensor.dtype if dtype is None else dtype, **kwargs)
 
   @staticmethod
+  def ones_like(tensor, dtype:Optional[DType]=None, **kwargs): return Tensor.ones(*tensor.shape, dtype=tensor.dtype if dtype is None else dtype, **kwargs)
+
+  @staticmethod
   def empty(*shape, device=Device.DEFAULT, dtype:Optional[DType]=None, **kwargs):
     # NOTE: we do the reshape to fix interpreted buffers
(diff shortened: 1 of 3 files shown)


wozeparrot:
I'm wondering if it would be better to have something like `torch.full_like`.

Maybe switch `Tensor.ones` to be `Tensor.full` as well then add Tensor.ones back calling `Tensor.full`

ubihot:
yeah makes sense, gonna work on it

tinygrad/tensor.py
```diff
@@ -122,13 +122,25 @@ def to(self, device:str):
...
-  def ones(*shape, **kwargs): return Tensor([1], **kwargs).reshape([1]*len(shape)).expand(shape).contiguous()
+  def zeros(*shape, **kwargs): return Tensor.full(shape, 0, **kwargs)
 
   @staticmethod
-  def zeros_like(tensor, dtype:Optional[DType]=None, **kwargs): return Tensor.zeros(*tensor.shape, dtype=tensor.dtype if dtype is None else dtype, **kwargs)
+  def ones(*shape, **kwargs): return Tensor.full(shape, 1, **kwargs)
+
+  @staticmethod
+  def full_like(tensor, fill_value, dtype:Optional[DType]=None, **kwargs):
+    return Tensor.full(tensor.shape, fill_value, dtype=tensor.dtype if dtype is None else dtype, **kwargs)
```
wozeparrot, line 137: Don't think you actually need the dtype to be specified here because it should be part of kwargs.

wozeparrot:
make sure to add tests for `full` and `full_like` as well

tinygrad/tensor.py
ubihot: I've tried to remove it and other stuff but it then complains that dtype is passed twice in kwargs

wozeparrot: Ah I see, yea should be fine then, this matches what torch does.

geohot:
LGTM!

PR #855, opened 2023-05-29
Make triton backend work again

christopherm99:
Global loops with 4 or more dims, local, and cast aren't implemented.

4 or more dims should be easy to fix. Local should use block pointers(?) or tl.arange(). I think cast is ok not implemented? Since its only used for float4?


geohot:
Can you get triton (and CUDA) in CI? That's what's really needed to make them first class citizens.

christopherm99:
Sure, should I just add runners to `test.yml`? Or write additional tests?

geohot:
The issue is how to do it :)

christopherm99:
Hmm, it looks like the simplest method is to use a self-hosted runner with a GPU. Unless there's some consistent method for emulating CUDA on CPU.

geohot:
Yea, I might have to do that. I got $5k in AWS credits to burn, wonder what the cheapest machine is and how easy it is to set up?

christopherm99:
I looked a bit into this, and it looks like there are a couple projects to emulate CUDA on CPU, but most of them seem abandoned. Apparently `nvcc` even used to support outputting CPU code but that feature was dropped a long time ago. Everything else is third-party, at various levels of unmaintained:

- [gpgpu-sim](https://github.com/gpgpu-sim/gpgpu-sim_distribution/tree/dev)
- [CU2CL](https://github.com/vtsynergy/CU2CL)
- [gpuocelot](https://github.com/gtcasl/gpuocelot)
- [MCUDA](http://impact.crhc.illinois.edu/mcuda.aspx)

I'm sure there are many more, but this is most of what I found.

geohot:
Yea none of these look great. I'm standing up a self hosted machine with a 3080 Ti now.

geohot:
Getting a bunch of failures with `TRITON=1 pytest`

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_triton.py           113               0
lines added in the tinygrad folder: 113
```

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_triton.py           113               0
lines added in the tinygrad folder: 113
```

geohot:
I fixed four or more dims upstream.

geohot:
Added $200 bounty for fixing this (all tests pass) + testing this + CUDA in CI.

geohot:
We are close to CUDA CI and can revisit this.

christopherm99:
Sorry about the long hiatus on this, got pretty busy and working on this fell to the wayside.

`test/test_ops.py` is now fully passing, but there are still a fair amount of tests that need some work. In particular, all types except float32 are broken (I'm not sure how to typecast in triton). 

Also, I need to figure out how to make sure that reduction/local loops are actually getting parallelized (this may require block tensors) and implement emulated testing using gpuocelot.

geohot:
Ahh, cool. Locking the bounty to you. Check out my work in #1085 on the actually getting parallelized.

christopherm99:
Sounds good. I'll start by getting the CI working and fixing the bugs, and leave the parallelization to @ptillet. With regards to the `test_dtype.py` errors, is there a reason why computations are done all in float32? The two failed tests I'm still getting are matmul in fp64 and fp16, since the reduction is computed in fp32, and I haven't figured out how to cast in triton yet. Should I switch to doing computation in the respective precision or cast to fp32?

geohot:
All computation should be float32.

christopherm99:
Ok sounds good, I'll use cast in the triton code. 

As far as I know the ops tests are real, perhaps triton is automatically figuring out how to optimize reasonably well.

christopherm99:
Ok almost all the tests are passing locally now. The only tests not are a couple of the model tests in `test_efficientnet.py` and `test_onnx.py`, which seems to be a problem with the CUDA backend overall.

tinygrad/runtime/ops_cuda.py
```diff
@@ -55,13 +55,14 @@ def __init__(self, name:str, prg:str, binary=False):
...
     # TODO: name is wrong, so we get it from the ptx using hacks
-    self.prg = cuda.module_from_buffer(prg.encode('utf-8')).get_function(prg.split(".visible .entry ")[1].split("(")[0])
+    self.prg, self.shared = cuda.module_from_buffer(prg.encode('utf-8')).get_function(prg.split(".visible .entry ")[1].split("(")[0]), shared
 
   def __call__(self, global_size, local_size, *args, wait=False):
     if wait:
       start, end = cuda.Event(), cuda.Event()
       start.record()
-    self.prg(*[x._buf for x in args], block=tuple(local_size), grid=tuple(global_size))
+    if getenv("CUDACPU"): self.prg(*[x._buf for x in args], block=tuple(local_size), grid=tuple(global_size))
```
geohot, line 64: Can you fix this for CUDACPU?

tinygrad/codegen/triton.py
```diff
@@ -0,0 +1,96 @@
...
+      if uop == UOps.LOOP:
+        for i,var in enumerate(args[0]):
+          if isinstance(var, NumNode): continue # python doesnt have block scope
+          else:
+            if args[1] == "global":
+              global_size.append(var.max+1)
+              kk(f"{var.expr} = {gid[i]} # {var.max+1}")
+            elif args[1] == "local":
+              local_size.append(var.max+1)
+              kk(f"{var.expr} = tl.arange({var.min}, {var.max+1})")
```
geohot, line 50: This doesn't work, it only doesn't complain because there's no optimizer. Can you put back `hand_coded_optimizations` to match cstyle?

geohot:
Those test failures are fine, see comments above.

test/test_ops.py
```diff
@@ -910,7 +910,7 @@ def test_maxpool2d_bigger_stride(self):
           lambda x: torch.nn.functional.max_pool2d(x, kernel_size=(2,2), stride=stride),
           lambda x: Tensor.max_pool2d(x, kernel_size=(2,2), stride=stride))
 
-  @unittest.skipIf(Device.DEFAULT == "CUDA", "CUDA fails on this")
+  @unittest.skipIf(Device.DEFAULT == "CUDA" or Device.DEFAULT == "TRITON", "CUDA fails on this")
```
geohot, line 913: Unneeded

tinygrad/runtime/ops_cuda.py
```diff
@@ -11,12 +11,13 @@
 if getenv("CUDACPU", 0) == 1:
   import ctypes, ctypes.util
   lib = ctypes.CDLL(ctypes.util.find_library("gpuocelot"))
+  print(lib.ptx_run.argtypes)
```
geohot, line 14: Leftover?

tinygrad/runtime/ops_cuda.py
christopherm99: Fixed.

tinygrad/runtime/ops_cuda.py
christopherm99: Yes, deleted.

test/test_ops.py
christopherm99: Removed.

tinygrad/codegen/triton.py
christopherm99: Working on this, implementing local loops has caused some issues with block pointers in triton, and implementing `ReduceOps.SUM` has proved challenging, specifically finding the dimension to reduce along from just the output of the linearizer. Working on changing the linearizer to specify the axis in question.

geohot:
If tests pass, I'll merge and the bounty is yours.

geohot:
So the Triton test wasn't actually running Triton, it was running CPU since there was no CUDA=1.

Fixed, now sadly test is failing.

wozeparrot:
Missing `pycuda` from dependencies

christopherm99:
Hmm, so the error in CI is that the version of torch installed doesn't have CUDA support. Not sure if it's possible to force pip to install the CUDA-enabled version in CI (or whether that would even work if it is possible). I also was looking into using `@triton.jit(interpret=True)`, but it looks like this requires torch with CUDA too. Triton can also produce HIP code, which may be easier to emulate? Otherwise I may need to create a shim for torch that still allows triton to generate PTX to test with gpuocelot.

ptillet:
@christopherm99 do the tests pass locally? were you able to generate triton code conducive to tensor cores and peak perf? if yes i'm willing to see what I can do. Triton in interpreter mode shouldn't require torch-cuda.

christopherm99:
Almost all the tests pass locally (still need to fix up a few bugs). The generated triton code is definitely suboptimal however, as far as I know tensor cores are not being used. It looks like triton relies on torch-cuda to detect the local gpu capabilities and various other minor things (if I can successfully hardcode shim torch-cuda out of triton, I'll have a list of the dependencies).

geohot:
How is this coming?

geohot:
Going to unlock bounty in 24 hours.

karan-amuse:
Am starting work on this. If I finish this, pls send bounty money to @christopherm99 as he did most of the work and this is very low hanging fruit at this point.

karan0handa:
Am starting work on this. If I finish this, pls send bounty money to @christopherm99 as he did most of the work and this is very low hanging fruit at this point.

geohot:
Upped bounty to $300 for this, we need it brought across the finish line!

wozeparrot:
Closing in favour of #1547

PR #856, opened 2023-05-29
Fix binary operation tests that have float as input

eichenroth:
Before: `=== 329 failed, 497 passed, 1808 skipped, 2 warnings in 18.43s ===`

Now: `=== 315 failed, 511 passed, 1808 skipped, 2 warnings in 19.33s ===`

Might have the flaw that we now just assume that the type of the input is `float32`. How can we be sure?


geohot:
I don't actually know how to be sure. There's probably something deep in the ONNX spec about this. This info might be thrown away in my onnx parser, that's the first place I'd check.

eichenroth:
Will have a look at the ONNX parser - however, I would argue, that merging this would still be okay, since 15 tests get fixed.

PR #857, opened 2023-05-29
Fix ONNX dropout and unify the implementation

marcellofuschi:
I had to change `Tensor._rng` from outside because the ONNX tests use `np.random.seed()` for the expected dropout mask, and using `default_rng(seed)` won't give the same results as when the same seed is used in `np.random.seed(seed)` (because the latter [uses](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) a [RandomState](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState)).

### DIFF
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -97,8 +97,13 @@ def ConvTranspose(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_
   return X.conv_transpose2d(W, B, stride=strides, groups=group, dilation=dilations, padding=(pads[1], pads[3], pads[0], pads[2]) if pads is not None else 0)
 
 def Dropout(data, ratio=0.5, training_mode=False, seed=None):
+  class _FakeRNG:
+    def __init__(self, seed): self.seed = seed
+    def random(self, size=None, dtype=None):
+      ret = np.random.RandomState(self.seed).random(size)
+      return ret.astype(dtype) if dtype else ret
   Tensor.training = training_mode
-  if seed is not None: Tensor._rng = np.random.RandomState(seed)  # Need to use RandomState because of how test data gets generated.
+  if seed is not None: Tensor._rng = _FakeRNG(seed)  # Need to use RandomState because of how test data gets generated.
   ratio = ratio.lazydata.realize().toCPU()[0] if isinstance(ratio, Tensor) else ratio
   return data.dropout(p=ratio, return_mask=True)
 
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -96,11 +96,13 @@ def Conv(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_shape=Non
 def ConvTranspose(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_shape=None, pads=None, strides=1):
   return X.conv_transpose2d(W, B, stride=strides, groups=group, dilation=dilations, padding=(pads[1], pads[3], pads[0], pads[2]) if pads is not None else 0)
 
+# Reimplemented here because you need legacy RNG for passing ONNX tests.
 def Dropout(data, ratio=0.5, training_mode=False, seed=None):
-  Tensor.training = training_mode
-  if seed is not None: Tensor.manual_seed(seed=seed, legacy=True)  # Need to use legacy RNG because of how test data gets generated.
+  if not training_mode: return data, Tensor.ones(*data.shape, dtype=dtypes.bool)  # if mask is requested as output it will contain all True's.
+  rng = np.random.RandomState(seed)
   ratio = ratio.lazydata.realize().toCPU()[0] if isinstance(ratio, Tensor) else ratio
-  return data.dropout(p=ratio, return_mask=True)
+  mask = Tensor((rng.random(data.shape) >= ratio), requires_grad=False, device=data.device)
+  return data * mask * (1/(1.0 - ratio)), mask
 
 def Shape(data, end=None, start=0): return list(data.shape)[start:end]
 def Size(data): return prod(data.shape)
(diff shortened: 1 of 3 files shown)


geohot:
Hmm, interesting test failure.

tinygrad/tensor.py
```diff
@@ -521,11 +521,15 @@ def batchnorm(self, weight:Optional[Tensor], bias:Optional[Tensor], mean:Tensor,
...
-  def dropout(self, p=0.5) -> Tensor:
-    if not Tensor.training: return self
+  def dropout(self, p=0.5, return_mask=False) -> Union[Tensor, Tuple[Tensor, Tensor]]:
+    if not Tensor.training:
+      # if mask is requested as output it will contain all True's.
+      return (self, Tensor.ones(*self.shape, requires_grad=False, dtype=dtypes.bool)) if return_mask else self
     # TODO: why is this going through numpy?
-    _mask: np.ndarray = np.asarray(Tensor._rng.binomial(1, 1.0-p, size=self.shape), dtype=np.float32)
-    return self * Tensor(_mask, requires_grad=False, device=self.device) * (1/(1.0 - p))
+    _mask: np.ndarray = Tensor._rng.choice(a=[False, True], size=self.shape, p=[p, 1-p])
```
geohot, line 529: Hmm, is there a way to write this using the existing rng functions and not `Tensor._rng`

dc-dc-dc: might be related to int ops not being supported yet, the llvm ir codegen is using float functions and bool is using int

dc-dc-dc:
> Interesting test failure. bool related?

might be related to int ops not being supported yet, the llvm ir codegen is using float functions and bool is using int

tinygrad/tensor.py
marcellofuschi: Added a commit for this.

New onnx_ops code looks a bit ugly though because I had to write a wrapper for `RandomState`. This is because `RandomState.random()` doesn't accept a `dtype` param while `Generator.random()` does (and `Tensor.rand` uses it).

Plus, it slightly bothers me to replace `Tensor._rng` with an object that's not the type-hinted `np.random.Generator`.

marcellofuschi:
Cool, test failures disappeared ✌️

extra/onnx_ops.py
```diff
@@ -96,14 +96,16 @@ def Conv(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_shape=Non
...
-  _mask : np.ndarray = np.asarray(Tensor._rng.binomial(1, 1.0-ratio, size=data.shape), dtype=data.dtype)
-  mask = Tensor(_mask, requires_grad=False, device=data.device)
-  return data * mask * (1/(1.0 - ratio)), mask
+  class _FakeRNG:
+    def __init__(self, seed): self.seed = seed
+    def random(self, size=None, dtype=None):
+      ret = np.random.RandomState(self.seed).random(size)
+      return ret.astype(dtype) if dtype else ret
+  Tensor.training = training_mode
+  if seed is not None: Tensor._rng = _FakeRNG(seed)  # Need to use RandomState because of how test data gets generated.
```
geohot, line 106: We need something better than this. This affects all future rng in a weird way.

marcellofuschi: True that! I hadn't considered it would stick to `Tensor`. Will fix in a bit.

marcellofuschi: Added a param to `Tensor.manual_seed` to support [legacy random generation](https://numpy.org/doc/stable/reference/random/legacy.html).

tinygrad/tensor.py
```diff
@@ -150,16 +150,20 @@ def where(self:Tensor, input_:Union[Tensor, float], other:Union[Tensor, float]):
   # ***** (numpy) rng helper functions *****
   # TODO: move randomness generation out of numpy
 
-  _rng: ClassVar[np.random.Generator] = np.random.default_rng()
+  _rng: ClassVar[Union[np.random.Generator, np.random.RandomState]] = np.random.default_rng()
```
geohot, line 153: Why do we need two different ones? Anything wrong with just changing it?

marcellofuschi: The methods in those classes may have different signatures. See for example [RandomState.standard_normal](https://numpy.org/doc/stable/reference/random/generated/numpy.random.RandomState.standard_normal.html#numpy.random.RandomState.standard_normal) vs [Generator.standard_normal](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.standard_normal.html#numpy.random.Generator.standard_normal)

If you just assign a `RandomState` to `Tensor._rng` someone could call methods on it thinking they can use `Generator`'s signatures when they actually can't.

marcellofuschi: If instead by "just changing it" you meant keeping only the legacy one, I think it'd be quite weird to force the worse implementation because of ONNX tests suite requirements.

geohot: Why is it worse?

marcellofuschi: See here: https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState

> Container for the _slow_ Mersenne Twister pseudo-random number generator. Consider using a different BitGenerator with the Generator container instead.

Also here: https://numpy.org/doc/stable/reference/random/legacy.html#legacy-random-generation

> The [RandomState](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState) provides access to legacy generators. [...] This class should only be used if it is essential to have randoms that are identical to what would have been produced by previous versions of NumPy.

geohot:
Hmm, so I def don't want to support them both. I think the best way to do this is to just code dropout explicitly in the onnx function.

marcellofuschi:
> Hmm, so I def don't want to support them both. I think the best way to do this is to just code dropout explicitly in the onnx function.

OK, done!

kunwar31:
@marcellofuschi Hey, I think we trying to do something similar in tensor.dropout https://github.com/geohot/tinygrad/pull/864

PR #864, opened 2023-05-30
Dropout without numpy

kunwar31:
Fix for https://github.com/geohot/tinygrad/blob/master/tinygrad/tensor.py#L543

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -515,9 +515,7 @@ class Tensor:
 
   def dropout(self, p=0.5) -> Tensor:
     if not Tensor.training: return self
-    # TODO: why is this going through numpy?
-    _mask: np.ndarray = np.asarray(Tensor._rng.binomial(1, 1.0-p, size=self.shape), dtype=np.float32)
-    return self * Tensor(_mask, requires_grad=False, device=self.device) * (1/(1.0 - p))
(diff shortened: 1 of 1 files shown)


tinygrad/tensor.py
```diff
@@ -515,9 +515,7 @@ def batchnorm(self, weight:Optional[Tensor], bias:Optional[Tensor], mean:Tensor,
 
   def dropout(self, p=0.5) -> Tensor:
     if not Tensor.training: return self
-    # TODO: why is this going through numpy?
-    _mask: np.ndarray = np.asarray(Tensor._rng.binomial(1, 1.0-p, size=self.shape), dtype=np.float32)
-    return self * Tensor(_mask, requires_grad=False, device=self.device) * (1/(1.0 - p))
+    return self * Tensor(Tensor._rng.binomial(1, 1.0 - p, size=self.shape), requires_grad=False, device=self.device, dtype=self.dtype) * (1 / (1.0 - p))
```
geohot, line 518: This is still numpy

kunwar31: @geohot I see, so is this about Tensor._rng not being based on numpy. should I replace it with https://docs.python.org/3/library/random.html?

geohot: No, it's about reducing calls to Tensor._rng and funneling them into the tinygrad randomness methods we already have.

kunwar31: by tinygrad randomness methods, do you mean Tensor.rand, Tensor.randn, etc? don't they also depend on Tensor._rng

kunwar31: @geohot changed it to use Tensor.rand

PR #865, opened 2023-05-30
Audio transforms: STFT, iSTFT and (log-) Mel Spectrogram

radenmuaz:
tinygrad port of 
https://github.com/KinWaiCheuk/nnAudio/blob/master/Installation/nnAudio/features/stft.py

inverse STFT uses col2im ported from
https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py#L864

STFT is differentiable.
iSTFT is maybe differentiable... my col2im implementation uses mask-multiply trick to gather input elements from col2im indices.


geohot:
So this doesn't belong in core tinygrad? Why do I want this?

radenmuaz:
For MLPerf RNNT
`This is not typical preprocessing, since it takes place as part of the model's measured runtime, not before the model runs.`
https://github.com/mlcommons/inference/tree/master/speech_recognition/rnnt

And for speech models in general, ASR, TTS, audio generative models.

We can move this to `models/` if it doesn't belong to `tinygrad/`

geohot:
Ahh, I see. I think mlperf helpers or something is the best place. It's not core tinygrad functionality and it's not a model.

geohot:
Until it's used in an end to end way to accomplish a task I'm not going to merge this. Do you want to try for the RNN-T training bounty?

radenmuaz:
Really wanted to try but don't have much free time. I'll see what I can do.

geohot:
Going to close until it's used in a task.

PR #867, opened 2023-05-30
context var for Device.DEFAULT

leopf:
Allows changing the default device with a context var.
```python
from tinygrad.lazy import Device, DEFAULT_DEVICE

DEFAULT_DEVICE("cpu") # internally uses name "DEVICE"

a = Tensor.ones(1) # on cpu

with Context(device="gpu"):
    b = Tensor.ones(1) # on gpu
    print((a.to("gpu") + b).numpy())
```

The current implementation breaks `Device.DEFAULT = x` (only used by TG in abstractions.py). 

I also changed the context to key.upper() allowing for lowercase "device" and added support for using `Context` as a function decorator. This should also be helpful for the "training" context var.

I don't really like that "DEFAULT_DEVICE" is named differently than its internal context var "DEVICE", but I wanted to avoid having both `Device` and `DEVICE` in lazy.py.

This also means, that there are now 2 ways of selecting a device with envs.

### DIFF
--- a/docs/abstractions.py
+++ b/docs/abstractions.py
@@ -22,8 +22,8 @@ from abc import ABC
 # let's trace an addition down through the layers of abstraction.
 
 # we will be using the clang backend
-from tinygrad.lazy import Device
-Device.DEFAULT = "CLANG"
+from tinygrad.lazy import Device, DEFAULT_DEVICE
+DEFAULT_DEVICE("CLANG")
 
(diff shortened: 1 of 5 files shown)


geohot:
I'm not sure device should be in context, but no_grad and training definitely should. Let's start with those ones.

leopf:
Ok, I will work on that! I am closing this PR and open a new one once I have something.

PR #868, opened 2023-05-30
arange without numpy, simple cumsum helper method

dee7ine:
Removed the usage of numpy in arange method, added simple cumsum helper method.
test_tensor.py passed, added test to compare result with numpy function

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -115,6 +115,13 @@ class TestTinygrad(unittest.TestCase):
     expected = Tensor(np.arange(stop=1337))
     assert type(arange) is Tensor
     np.testing.assert_allclose(arange.numpy(), expected.numpy())
+    
+  def test_cumsum(self):
+    arange = np.arange(stop=100)
+    expected = Tensor(np.cumsum(arange))
+    cumsum = Tensor.cumsum(arange)
+    assert type(cumsum) is Tensor
+    np.testing.assert_allclose(cumsum.numpy(), expected.numpy())
 
   #@unittest.skipUnless(Device.DEFAULT == Device.CPU, "float64 not supported on GPU")
   @unittest.skip("float64 support broken")
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -110,19 +110,6 @@ class TestTinygrad(unittest.TestCase):
     expected = n * (1 - rate)
     np.testing.assert_allclose(non_zeros, expected, rtol=2e-3)
     
-  def test_arange(self):
-    arange = Tensor.arange(stop=1337)
-    expected = Tensor(np.arange(stop=1337))
-    assert type(arange) is Tensor
-    np.testing.assert_allclose(arange.numpy(), expected.numpy())
-    
-  def test_cumsum(self):
-    arange = np.arange(stop=100)
-    expected = Tensor(np.cumsum(arange))
-    cumsum = Tensor.cumsum(arange)
-    assert type(cumsum) is Tensor
-    np.testing.assert_allclose(cumsum.numpy(), expected.numpy())
-
   #@unittest.skipUnless(Device.DEFAULT == Device.CPU, "float64 not supported on GPU")
   @unittest.skip("float64 support broken")
   def test_jacobian(self):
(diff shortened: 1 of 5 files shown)


tinygrad/tensor.py
```diff
@@ -133,9 +134,13 @@ def eye(dim, **kwargs): return Tensor([1], **kwargs).slice(((0,dim+1),)).reshape
 
   # TODO: below line, remove use of numpy here and make lazy
   # TODO: requires cumsum to remove numpy
+  
   @staticmethod
-  def arange(stop, start=0, step=1, **kwargs): return Tensor(np.arange(start=start, stop=stop, step=step, dtype=np.float32), **kwargs)
-
+  def cumsum(x: Iterable): return Tensor(itertools.accumulate(x))
```
wozeparrot, line 139: There should be a test for `cumsum` as well.

dee7ine: Added test for cumsum, all tests passed.

dee7ine: https://github.com/geohot/tinygrad/blob/6738ad3f9e461a00fea08723234f0c07e18b2bac/tinygrad/tensor.py#L137-L138

test/test_tensor.py
```diff
@@ -109,6 +109,19 @@ def test_dropout(self):
     non_zeros = np.count_nonzero(w.cpu().numpy())
     expected = n * (1 - rate)
     np.testing.assert_allclose(non_zeros, expected, rtol=2e-3)
+    
+  def test_arange(self):
+    arange = Tensor.arange(stop=1337)
+    expected = Tensor(np.arange(stop=1337))
+    assert type(arange) is Tensor
+    np.testing.assert_allclose(arange.numpy(), expected.numpy())
```
wozeparrot, line 117: I think we already have an arange test in `test/test_ops.py`

Maybe move the `cumsum` test there as well?

dee7ine: Sure, moved the cumsum test to test_ops.
All tests passed.

tinygrad/tensor.py
```diff
@@ -133,9 +133,13 @@ def eye(dim, **kwargs): return Tensor([1], **kwargs).slice(((0,dim+1),)).reshape
...
   # TODO: below line, remove use of numpy here and make lazy
   # TODO: requires cumsum to remove numpy
+  
   @staticmethod
-  def arange(stop, start=0, step=1, **kwargs): return Tensor(np.arange(start=start, stop=stop, step=step, dtype=np.float32), **kwargs)
-
+  def cumsum(x: Iterable, **kwargs): return Tensor([i for i in itertools.accumulate(x)], **kwargs)
+  
+  @staticmethod
+  def arange(stop, start=0, step=1, **kwargs): return Tensor([x for x in range(start, stop, step)], **kwargs)
```
geohot, line 141: This is not what I intended when I said remove numpy.

tinygrad/tensor.py
```diff
@@ -133,9 +133,13 @@ def eye(dim, **kwargs): return Tensor([1], **kwargs).slice(((0,dim+1),)).reshape
 
   # TODO: below line, remove use of numpy here and make lazy
   # TODO: requires cumsum to remove numpy
+  
   @staticmethod
-  def arange(stop, start=0, step=1, **kwargs): return Tensor(np.arange(start=start, stop=stop, step=step, dtype=np.float32), **kwargs)
-
+  def cumsum(x: Iterable, **kwargs): return Tensor([i for i in itertools.accumulate(x)], **kwargs)
```
geohot, line 138: No, why would this be on the Tensor class?!?

dee7ine: I assumed it would be better to put it in helper.py module as a helper function. 
https://github.com/dee7ine/tinygrad/blob/a44c7e5504a25052f86c7e21bb0d5175c1e3ece5/tinygrad/helpers.py#L16-L19

One thing I couldn't find in the discussion is why numpy should be removed. Is it to remove any dependencies?

dee7ine: I assumed it could be better to move it to helpers.py module. 
https://github.com/geohot/tinygrad/blob/a44c7e5504a25052f86c7e21bb0d5175c1e3ece5/tinygrad/helpers.py#L17-L19

PR #872, opened 2023-05-30
Linearizer Refactor

save-buffer:
Copy-pasting from my Discord post

The call to codegen will be broken down into the following steps:
process (which I've renamed to canonicalize ), which moves reductions to the end, creates the shapetrackers etc.
optimize which just iterates through and calls a list of functions which define the optimization passes. I think this is both more flexible and simpler to understand than the current hand_coded_optmizations thing. This will also help organize the code better by moving each optimization into its own function
linearize which walks the AST and generates a list of uops
uops_to_program which runs through the list of uops and generates source code (whether it be textual metal shaders, or LLVM IR)

For now I've just been writing a LinearizerV2, once it's ready we could delete old Linearizer and rename this one

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -1,5 +1,5 @@
 from typing import List, Tuple, Any, Optional, cast, DefaultDict, NamedTuple, TypeVar, Dict
-import itertools, math
+import itertools, math, abc
 from collections import defaultdict
 from enum import Enum, auto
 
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py            71               1
lines added in the tinygrad folder: 70
```

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py            71               1
lines added in the tinygrad folder: 70
```

PR #874, opened 2023-05-30
mlperf bert training

wozeparrot:
Wooo, Bert training time. :rocket: 

Going to try to do a run where I overfit on the eval set just to make sure everything is working.

Kinda annoying that its a different dataset again and you have to load from a pretrained tensorflow model.


geohot:
Hehe, you get one locked bounty. You want this one or RNN-T locked?

wozeparrot:
I'll take this locked, RNN-T is a bit too annoying on the loss side :)

wozeparrot:
this is probably ready to scale. pretty much everything is implemented and its end to end jitted. might write a naive multigpu training just to go faster.

wozeparrot:
Already this should be ready to go, need one of the dataset archives from here: https://drive.google.com/drive/folders/1cywmDnAsrP5-2vsr8GDc6QUc7VWe-M3v extracted into `datasets/wikipedia/raw/`. You should end up with a directory `datasets/wikipedia/raw/results4/`.
Then pull `eval.txt` out of `datasets/wikipedia/raw/results4/` into `datasets/wikipedia/raw/`.

Then create directories `datasets/wikipedia/eval/` and `datasets/wikipedia/train/`.

Then run `python datasets/wikipedia.py pre-eval` to preprocess the eval set.
Then run `python datasets/wikipedia.py pre-train <part number>` for every part (there are 500) to preprocess the training set.

Then for the initial weights you need the tf1 checkpoint from here: https://drive.google.com/drive/folders/1NJaS5Kj7vNkE4Xl-6uMwRPHGTxSMFiithttps://drive.google.com/drive/folders/1NJaS5Kj7vNkE4Xl-6uMwRPHGTxSMFiit

It should stop once it hits the target accuracy, and it saves a checkpoint every 10k steps.

geohot:
Sweet! So for the bounty, we have to hit the accuracy in 12 hours on some machine.

Do you have an estimate of where the step time is and what we have to get the step time to to hit that? Are we 2x off, 10x off, or 50x off?

2x should be just some minor fixes
10x is doable with tensor cores + kernel search OR multigpu + minor fixes
50x is multigpu + tensor cores + kernel search AKA full speed in torch

I should be able to get you a tinybox for ssh in a couple weeks. I also added a $1,000 bounty for "clean MultiGPU training with ring allreduce" that we can tack on to this if you'd like. I think it'll only be 100-200 lines of code, might need to refactor `.to` a bit to specialize and make sure the copies are actually async.

wozeparrot:
I'm not sure what the exact training time will be, its taking me about 1 hour 30 for the eval step, but that's for a BS of 1, the BS can def be bumped up on a 7900xtx. Lets say about 10x off, I think the big thing will be multigpu.

Is it fine if I lock the multigpu bounty and just do it on a tinybox? cuz my system doesn't havve multiple gpus. tho it should probably be doable with like the torch backend on cpu and a gpu, i don't see why not.

PR #875, opened 2023-05-31
Support `output_padding` in transposed conv

kposborne2:
Works by adjusting the padding in the call to `conv2d`
Added some tests for asymmetric args/bias while debugging this, we also now pass some more of the onnx tests

### DIFF
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -99,8 +99,8 @@ def Conv(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_shape=Non
   padding = [p for ps in zip(pads[:len(pads)//2][::-1], pads[len(pads)//2:][::-1]) for p in ps] if pads is not None else 0 # reorder padding
   return X.conv2d(W, B, stride=strides, groups=group, dilation=dilations, padding=padding)
 
-def ConvTranspose(X, W, B=None, auto_pad="NOTSET", dilations=1, group=1, kernel_shape=None, pads=None, strides=1):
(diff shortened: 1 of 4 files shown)


geohot:
I'm not sure this is right. Wouldn't output padding be on the output?

kposborne2:
> I'm not sure this is right in all cases. Wouldn't output padding be on the output?

It is not quite the same as adding padding to the output. The idea is that if you supply one set of arguments to a convolution to go from shape A -> B, you should be able to supply the same set of arguments in a transposed convolution to go from shape B -> shape A. But, when stride > 1, it is not fully determined from shape B + the convolution arguments what is the shape A, so the extra argument `output_padding` is added to the transposed convolution.

Here is a good blog post about it how the different parameters impact the transposed convolution behavior: https://numbersmithy.com/understanding-transposed-convolutions-in-pytorch/

I copied this demonstrative example, it has a convolution input shape of 7, with stride=2, dilation=2, and padding=1, resulting in an output shape of 3. But, an input shape of 8 with those parameters would also result in an output shape of 3. Distinguishing that case a transposed convolution with stride=2, dilation=2, and padding=1 requires adding the `output_padding` argument of 1. But, that final output value is not 0, it still follows the normal logic. 
Translating that to the implementation in tinygrad, including the correct values is equivalent to padding the convolution call to on the low end by 3: `(k-1)*d-p`, and on the high end by 4:  `(k-1)*d-p+op`

image

geohot:
Cool, I trust you've thought about it :)

Issue #861, opened 2023-05-30
I want sugar


xl0:
Are we ok with adding a bunch of numpy/pytorch convenience methods, like `.numel()`, `.ndim()`, `.isinf()`, etc, etc?

How do we choose which ones are desirable and which ones are not? And most important, what should be the name when the two disagree?

How about, starting with this table:

| Funciton     | PyTorch| NumPy | TinyGrad | Want? |
| ----------- | ----------- |----------- |----------- |----------- |
| Number of elements | `.numel()`, | `.size()` | `.numel()` | :+1: |
| `len()`|  | `len()` | `len()` | :-1:  |
| Numer of bytes | `.numel() * .element_size()` | `.nbytes()` | `.nbytes()` | :heavy_check_mark:  |

(I'll add more as I run into things that are missing, and other people can add more in comments)
And let @geohot :+1: :-1: them as a form of bloat control. Then we can convert the :+1:s into :heavy_check_mark:s .

sounds like a plan?

geohot:
Yea, I'm fine with most PyTorch matching one liners in tensor.py

geohot:
Should be fixed with #869 and #876

PR #869, opened 2023-05-30
Add tensor.numel

nima10khodaveisi:

PR #876, opened 2023-05-31
Add Tensor methods .ndim , .numel, .element_size, .nbytes and also__len__

xl0:


xl0:
Fixes #861

@nima10khodaveisi Sorry did not see your #869, but also `.numel` should be a `@peoperty` .

geohot:
Tests failing. (btw, install the pre-commit hooks to run the linter on commit)

xl0:
@geohot , my bad, I pushed a branch that was not rebased. Should be good now.

geohot:
I've never seen anyone use `is_floating_point`

xl0:
- I'm using `is_floating_point` because you can't calculate mean/std on a non-float tensor.  If the criteria for adding stuff is anyone is using it, then yes, I'm using it. If you hate having it, I'll drop it.
- My bad re `__len__`, I guess I never use `len()`. I'll drop it.
- I'll drop `.nbytes` because I agree the semantics is confusing - should it return the real bytes, or just `numel * sizeof`. Non-obvious methods are not ideal. It's easy enough to multiply 2 numbers explicitly if one needs to.

tinygrad/tensor.py
```diff
@@ -552,6 +549,16 @@ def cast(self, dtype:DType) -> Tensor: return mlops.Cast.apply(self, dtype=dtype
...
   def half(self) -> Tensor: return self.cast(dtypes.float16)
 
+
+  # ***** Convenience stuff *****
+  @property
+  def ndim(self) -> int: return len(self.shape)
+  def numel(self) -> int: return math.prod(self.shape)
+  def element_size(self) -> int: return self.dtype.itemsize
+  def nbytes(self) -> int: return self.numel() * self.dtype.itemsize
+  def is_floating_point(self) -> bool: return dtypes.is_float(self.dtype)
```
geohot, line 559: If you really want to keep this, at least remove the 8 lines of (untested) logic from dtypes. `self.dtype in [dtypes.float16, dtypes.float32]`

xl0: Do you mean `is_int()`, `is_insgned()`? They are used in codegen.

How do you feel about 
```
class Dtype(...):
    [...]
    is_int: bool
    is_float: bool
    is_signed: bool
```

And then just `tensot.dtype.is_float`, etc in the code? This can turn 6 lines into 3.

xl0:
@nima10khodaveisi , don't think there was a duplication, except for my earlier commits. Do you see them?

PR #881, opened 2023-05-31
Add unbiased std and corresponding tests

tomatillos:


tomatillos:
Subtracting one from the divisor didn't work because a/(ab-1) != 1/(b-1). Instead have done 1/ ( (ab/a) -1 ). Would easily extend/swap to `correction` which replaces `unbiased` in Torch 2.0

geohot:
Oh correction is nice, https://pytorch.org/docs/stable/generated/torch.std.html

I'd just copy that, it's okay if the behavior changes tinygrad is pre 1.0. Good attention to detail.

tomatillos:
Changed unbiased option to correction. Have set default to correction=1, in line with torch (but not with numpy!)

PR #873, opened 2023-05-30
Training ResNet_50v1.5 on ImageNet

shauray8:
* Using SGD to train ResNet
* Implement cross-entropy Loss
* Adding Transforms
Should reach the desired accuracy within 50 epochs in theory ( Implementation details - https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/Classification/ConvNets/resnet50v1.5 )
* SGD doesn't have weight decay???

PR #883, opened 2023-05-31
Add weight decay to SGD

wozeparrot:
This is needed for #873 and maybe one other mlperf model

This matches the pytorch implementation.

There is a strange bug? I'm trying to fix with two specific high lr tests, `test_multistep_sgd_high_lr_wd` and `test_multistep_sgd_high_lr_nesterov_momentum_wd` which seem to have corrupted/zeroed weights on all backends. If I drop the lr for these two tests to 9 it does work but that doesn't really fix the problem.

### DIFF
--- a/test/test_optim.py
+++ b/test/test_optim.py
@@ -17,6 +17,7 @@ class TinyNet():
 
   def forward(self):
     out = self.x.matmul(self.W).relu()
+    # print(out.detach().numpy())
     out = out.log_softmax(1)
     out = out.mul(self.m).add(self.m).sum()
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/optim.py                       3               3
```

wozeparrot:
I think I may have just picked the worst possible hyperparams that perfectly cancel out weights, doesn't explain why it won't match torch tho.

geohot:
Meh, LGTM

PR #885, opened 2023-05-31
Small Onnx Parser Improvements

eichenroth:
Coming back to the big unknown of the filetypes in #856.

Constants should return tensors, they do now, therefore no type conversion is necessary for the inputs of `Add`, `Sub`, `Mul` & `Pow`.

Also added types here and there.


geohot:
How does this affect number of onnx tests passed?

eichenroth:
Fixes one more test. Tests are down to 294 passed on this branch now.

However, this implements the correct type conversion, that is mentioned in PR #856

PR #877, opened 2023-05-31
refactor context manager

ductm104:
Using [PEP 448](https://peps.python.org/pep-0448/) to merge dicts.
```python
>>> {**{'x': 1}, **{'x': 2}}
{'x': 2}
```

PR #886, opened 2023-06-01
Tensor context vars no_grad and training.

leopf:
Context vars for training and no_grad. Allows for:
```python

@Context(training=True)
def train_cifar():
    ...

@Context(no_grad=True)
def run_llama():
    ...    

# or

with Context(training=True):
    ...
    # train some model
```

Tests pass for me. I was careful, but hope I did not break any examples.

I also saw #877. I will add some tests for the context vars.

### DIFF
--- a/examples/benchmark_train_efficientnet.py
+++ b/examples/benchmark_train_efficientnet.py
@@ -4,9 +4,9 @@ import time
 from tqdm import trange
 from models.efficientnet import EfficientNet
 from tinygrad.nn import optim
-from tinygrad.tensor import Tensor
+from tinygrad.tensor import Tensor, NO_GRAD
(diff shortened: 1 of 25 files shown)


leopf:
Added some tests for training and no_grad.

Testing for training behavior in dropout and batchnorm2d. Also testing if no_grad actually results in there being no grad.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        8               7
tinygrad/nn/__init__.py                    2               2
tinygrad/tensor.py                         6               5
total                                     16              14
```

geohot:
Absolutely massive change I can't read.

The functionality should be introduced in a separate PR from the one that uses it.

PR #887, opened 2023-06-01
document environment variables

smurfd:
Gathered the env vars with a short description.
some might be wrongly described. but atleast its a good start.


geohot:
Looks good! But I'd put it in its own file, not abstractions. Even better, want to write a README.md for this folder and include it there, so someone will see it when they click docs?

smurfd:
yes that sounds good, see this commit

geohot:
LGTM

PR #777, opened 2023-05-15
Zero dim Tensor support

Joqsan:
This PR adds the following features:

- Zero-dim support:
    
    ```python
    >>> from tinygrad.tensor import Tensor
    >>> b = Tensor(10)
    >>> b.shape
    ()
    >>> ones = Tensor.ones(2,2,2,2)
    >>> ones.sum().shape
    ()
    >>> (ones.sum() * b).numpy()
    array(160., dtype=float32)
    >>> (ones.sum() * b.reshape(1)).numpy()
    array([160.], dtype=float32)
    ```
    
- varargs support for `Tensor.zeros()` and `.ones()`: not only `Tensor.zeros(45, 65)` works, but also `Tensor.zeros([45, 65])`.
    - Why: Make them consistent with their torch counterparts: e.g., `Tensor.zeros([])` should return a 0dim tensor.
- A refactored `Tensor.__getitem__()`.
    - ~5.8x faster slicing:
        
        ```python
        d = torch.arange(64**3).reshape(64,64,64)
        %timeit old_getitem(d, s[1:32, 32:64, 1:64], print_results=False)
        %timeit new_getitem(d, s[1:32, 32:64, 1:64], print_results=False)
        23.6 µs ± 975 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
        4.06 µs ± 46.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
        
        %timeit old_getitem(d, s[1:32, 32:64, None, 1:64, None, None], print_results=False)
        %timeit new_getitem(d, s[1:32, 32:64, None, 1:64, None, None], print_results=False)
        28.5 µs ± 638 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
        4.86 µs ± 91.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
        ```
        
    - Remove the use of `.slice()`; it just relies on `.shrink()` and `.reshape()`.
    - Fix weird behaviors when slicing out of range:
        
        ```python
        >>> w = np.arange(8**3).reshape(8,8,8).astype(np.float32)
        >>> tg = Tensor(w)
        >>> pt = torch.from_numpy(w)
        >>> tg[-4:2].shape  # tensor of shape (-2, 8, 8)
        >>> pt[-4:2].shape  # tensor([], size=(0, 8, 8))
        
        ```
        
    - More torch-like exception messages if invalid slices passed.
[...]

PR #890, opened 2023-06-01
ConstantOfShape ONNX test fixed.

stevenandersonz:
Before
=== 296 failed, 530 passed, 1808 skipped, 3 warnings in 21.67s ===
After
=== 294 failed, 532 passed, 1808 skipped, 3 warnings in 19.38s ===

### DIFF
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -205,9 +205,9 @@ def Or(x:Tensor, y:Tensor): return Where((x==y), x, Tensor.ones(*x.shape)).cast(
 def Xor(x:Tensor, y:Tensor): return Where((x==y), Tensor.zeros(*x.shape), Tensor.ones(*x.shape)).cast(dtypes.bool)
 def Not(x:Tensor): return Where((x==1), Tensor.zeros(*x.shape), Tensor.ones(*x.shape)).cast(dtypes.bool)
 
-def ConstantOfShape(input, value=0.0):
+def ConstantOfShape(input, value=Tensor([0.0])):
   shape = [int(x) for x in safe_numpy(input)]
-  return Tensor.ones(*shape, dtype=value.dtype) * value 
+  return Tensor.ones(*shape, dtype=value.dtype) * (value if input.shape !=(0,) else 1)
 
 # this is obviously wrong, but since we don't have types, it's better than nothing
 def Cast(input, to):
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -207,7 +207,7 @@ def Not(x:Tensor): return Where((x==1), Tensor.zeros(*x.shape), Tensor.ones(*x.s
 
 def ConstantOfShape(input, value=0.0):
   shape = [int(x) for x in safe_numpy(input)]
-  return Tensor.ones(*shape) * value
+  return Tensor.ones(*shape, dtype=value.dtype) * value if value > 0 and shape[0] > 0 else Tensor.zeros(*shape, dtype=value.dtype)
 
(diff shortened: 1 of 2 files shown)


extra/onnx_ops.py
```diff
@@ -207,7 +207,7 @@ def Not(x:Tensor): return Where((x==1), Tensor.zeros(*x.shape), Tensor.ones(*x.s
 
 def ConstantOfShape(input, value=0.0):
   shape = [int(x) for x in safe_numpy(input)]
-  return Tensor.ones(*shape) * value
+  return Tensor.ones(*shape, dtype=value.dtype) * value if value > 0 and shape[0] > 0 else Tensor.zeros(*shape, dtype=value.dtype)
```
geohot, line 210: Do you need this if statement?

stevenandersonz: Without it test_constantofshape_int_shape_zero_cpu fails with AssertionError: zeros not allowed in shape (0,).
Tinygrad as far I understand doesn't support empty tensor operations with zero as torch does and one of the test cases test for a tensor of shape(0,) .

```python
torch.ones(0) * 0.0 # realizes into tensor([])
Tensor.ones(0) * 0.0 # AssertionError: zeros not allowed in shape (0,)
```

Maybe I'm missing something?

wozeparrot: Seems related to #777

stevenandersonz: I Fetched Joqsan zero_dim_support branch and tried my changes there, but it still doesn't allow the behavior described above.

I could remove the if statement in my code and once the behavior is allowed in Tinygrad that test will pass automatically.

extra/onnx_ops.py
```diff
@@ -207,7 +207,7 @@ def Not(x:Tensor): return Where((x==1), Tensor.zeros(*x.shape), Tensor.ones(*x.s
 
 def ConstantOfShape(input, value=0.0):
   shape = [int(x) for x in safe_numpy(input)]
-  return Tensor.ones(*shape) * value
+  return Tensor.ones(*shape, dtype=value.dtype) * value
```
geohot, line 210: Err, what if value is 0.0?

stevenandersonz: @geohot  yeah I mentioned in a comment above:

The if was necessary when the value was 0.0 otherwise throws AssertionError: zeros not allowed in shape (0,) however it would be enough to just check the shape `if input.shape != (0,) 

I was wondering also if Tinygrad will support the behavior below then the if wouldn't be needed at all

```python
torch.ones(0) * 0.0 # realizes into tensor([])
Tensor.ones(0) * 0.0 # AssertionError: zeros not allowed in shape (0,)
```
Which one of the two should I follow?

geohot: 0.0.dtype

extra/onnx_ops.py
```diff
@@ -205,9 +205,9 @@ def Or(x:Tensor, y:Tensor): return Where((x==y), x, Tensor.ones(*x.shape)).cast(
 def Xor(x:Tensor, y:Tensor): return Where((x==y), Tensor.zeros(*x.shape), Tensor.ones(*x.shape)).cast(dtypes.bool)
 def Not(x:Tensor): return Where((x==1), Tensor.zeros(*x.shape), Tensor.ones(*x.shape)).cast(dtypes.bool)
 
-def ConstantOfShape(input, value=0.0):
+def ConstantOfShape(input, value=Tensor([0.0])):
```
geohot, line 208: When does this tensor get created?

PR #892, opened 2023-06-01
Bert pretraining

kolakows:
I've seen that the bounty is already locked, but still wanted to post my take on the Bert's pretraining code. 

If it's not worth including in the repo, feel free to close the PR.

Details:

For the Bert model, I've modified [models/transformer.py](https://github.com/geohot/tinygrad/blob/ae83e9844c7ed07f04087af7c865df97540be745/models/transformer.py#LL52C10-L52C10) to use prenorm, new_gelu activation and turned on gradient for the embedding layer, the rest is the same.

Dataset prep is a modified version of [nanoGPTs dataset preps](https://github.com/karpathy/nanoGPT/blob/master/data/openwebtext/prepare.py).

During training batches are sampled from the pretokenized train.bin blob and masked. I've ommited the NSP (Next Sentence Prediction) loss as it was shown in later papers to not improve models' quality. 

I've run the code on A10 gpu, achieving ~6s/step with bert_base architecture (12 layers, 12 heads, 786 embed_dim, ffn = 4*embed_dim), context 128 and BS=32. Original Bert was trained for 1,000,000 steps with BS=256.

### DIFF
--- a/datasets/wikibooks.py
+++ b/datasets/wikibooks.py
@@ -1,0 +1,70 @@
+# based on nanoGPT's prepare of openwebtext
+# https://github.com/karpathy/nanoGPT/blob/master/data/openwebtext/prepare.py
+
+import os
+import re
+from tqdm import tqdm
+import numpy as np
+from transformers import BertTokenizer
+from datasets import load_dataset, concatenate_datasets  # huggingface datasets
+
+# number of workers in .map() call
+# good number to use is ~order number of cpu cores // 2
+num_proc = 13
+
+# prepare bookcorpus
+bookcorpus = load_dataset("bookcorpus", split='train', num_proc=num_proc)
+
+# prepare wiki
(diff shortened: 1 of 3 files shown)


datasets/wikibooks.py
```diff
@@ -0,0 +1,70 @@
...
+
+# prepare bookcorpus
+bookcorpus = load_dataset("bookcorpus", split='train', num_proc=num_proc)
+
+# prepare wiki
+wiki = load_dataset("wikipedia", "20220301.en", split='train',  num_proc=num_proc)
+wiki = wiki.remove_columns([col for col in wiki.column_names if col != "text"])
+
+def normalize_whitespaces(example):
+    text = example['text']
```
jla524, line 23: Keep the indents at 2 spaces

kolakows: Good point, corrected

PR #895, opened 2023-06-01
Add GA to calculate sz-diff

fedemengo:
This GA is based on cloc, as of now it targets only `tinygrad` dir (as `sz.sh` used to do, just noticed is missing) for ready-to-review PRs

- checks out base ref (`tinygrad:master`)
- checks out PR ref
- `cloc --include-ext="$EXT=py" --diff "base/$TARGET=tinygrad" "pr/$TARGET=tinygrad"`
- count total, and comment with markdown table

here an example https://github.com/fedemengo/code-metrics-for-pr/pull/9

it _might_ be necessary to add some permissions to the GA in the repo settings


geohot:
Use the sz.py script in the repo.

fedemengo:
- checks out base ref (`tinygrad:master`)
- checks out PR ref
- run `sz.py` for both ref and calculate diff
- comment with an extended version of the output of `sz.py`

here an example https://github.com/fedemengo/code-metrics-for-pr/pull/4

fedemengo:
Looks like tinyb0t is already doing something similar. Although it doesn't seem to refresh the stats on new commits.

I could integrate [radon](https://radon.readthedocs.io/en/latest/intro.html) to keep track of some complexity metrics (Cyclomatic Complexity, Maintainability Index, Halstead Metrics) in case there is interest.

.github/workflows/sz-diff.py
```diff
@@ -0,0 +1,30 @@
...
+from tabulate import tabulate
+import itertools
+
+if __name__ == '__main__':
+  base, pr = gen_stats(sys.argv[1]), gen_stats(sys.argv[2])
+  base_files, pr_files = {x[0]: x for x in base}, {x[0]: x for x in pr}
+  base_loc, pr_loc = sum(x[1] for x in base), sum(x[1] for x in pr)
+  def first_dir(x): return x.rsplit("/", 1)[0]
+
+  modified = [[x[0], x[1], x[2], x[1]-base_files[x[0]][1], "M"] for x in pr if x[0] in base_files]
```
geohot, line 14: There has to be better ways to write this

jla524:
> Looks like tinyb0t is already doing something similar. Although it doesn't seem to refresh the stats on new commits.
> 
> I could integrate [radon](https://radon.readthedocs.io/en/latest/intro.html) to keep track of some complexity metrics (Cyclomatic Complexity, Maintainability Index, Halstead Metrics) in case there is interest.

tinyb0t will update the comments when new commits are added, but it doesn't show the total line count

fedemengo:
@jla524 is tinyb0t using `sz.py`?

jla524:
> @jla524 is tinyb0t using `sz.py`?

No, it doesn't checkout the code. It only calls the GitHub API to see how many lines were added and removed.

fedemengo:
I think it's still valuable to use the logic from `sz.py`, in particular the github api counts line counts differently than what we are interested it (whitespaces)

geohot:
Copy pasted sz python in two places, complexity I don't think is needed. I'd merge a *very simple* size commenter on PRs to replace tinyb0t with something internal, but this is complex looking

PR #898, opened 2023-06-01
Slightly improved readability of reshape method condition check 

dee7ine:
Replaced all() function call with simple check if 0 is anywhere in an iterable.
`assert 0 not in new_shape, f"zeros not allowed in shape {new_shape}"`


tinyb0t:
Changes made in `tinygrad/`:
```
```

Issue #882, opened 2023-05-31
Standard tinygrad weight format?


wozeparrot:
Inspired by #878 and kinda needed for mlperf training? not sure if mlperf specifies the checkpoint format.

We might want to look into having a standardized way of storing model weights and the code to read and write them.

I'm not sure what the best way to go about this is, for example in https://github.com/wozeparrot/tinyrwkv I'm just pickling the state_dict converted to ndarrays. The thing is the ML community seems to be moving away from pickle due to its possible security concerns during weight loading.

Maybe we adopt [safetensors](https://github.com/huggingface/safetensors)? But safetensors is pretty big and has a rust dependency which seems like way too much stuff to bring in.

Thoughts?

geohot:
Disktensors! I think we just convert the state_dict to a bunch of disktensors in a folder, then for distribution, you can tar.xz that folder.

wozeparrot:
Right! I totally forgot that you implemented those :facepalm:

Need a good way to load them tho `Tensor.empty(*shape, device='disk:/tmp/file')` doesn't seem very scalable to the entire state_dict.

Do we need metadata with the format?

geohot:
Yea, and I don't have a great way to do that. An option I was playing with was putting the dtype and shape in the file extension, but I don't love that.

Maybe metadata is just json concatted to the end of the file that describes the tensor(s) in it.

wozeparrot:
> then for distribution, you can tar.xz that folder.

I don't know if this would actually be good, because it means that we can no longer mmap the tensors directly and have to have a step of extracting them first.

I do quite like safetensors tho, It would be nice if we could find a way to make it work with disktensors or at least incorporate parts of it.

geohot:
Oh safe tensors sound good. Re: working with disktensors. Add a "@<start>-<end>" to the filename to get only that region?

wozeparrot:
Doesn't look like there are compliance tests, closest I could find are the tests in the reference rust library: https://github.com/huggingface/safetensors/blob/main/safetensors/src/tensor.rs

geohot:
Ugh, cast isn't exactly what I want...

```
t = Tensor.empty(1, dtype=dtypes.int64, device=f"disk:{fn}").numpy()[0]
j = Tensor.empty(8+int(t), dtype=dtypes.uint8, device=f"disk:{fn}")[8:]
print(json.loads(j.numpy().tobytes()))
```

geohot:
[...]

PR #900, opened 2023-06-01
Add a quick start guide

wozeparrot:
This adds a pretty simple quick start guide/tutorial that goes over the high level tinygrad API.

~~I do think it would be good to mention the JIT in this, probably just at the end as a note.~~ Added

I was aiming for this to be enough so that someone with zero knowledge of tinygrad should be able to learn enough to implement some larger models.
We do need more detailed docs tho.

Edit: ~~need to add saving and loading models as well but that depends on #882~~ Added


geohot:
Great work! Maybe this is the thing that comes up when you click the docs folder and it links to the other two?

wozeparrot:
I think it would be better if we had a sort of landing page for the docs, the current one is not great.

I'll throw something together.

wozeparrot:
hmm, maybe I should just replace the repo readme as well, cuz what I have right now for the docs readme could also work for the repo readme with some tweaks and adding some missing stuff like the ops breakdown.

The current repo readme seems ok tho.

wozeparrot:
I think this is in a state where we can get some comments on it.

You can see the rendered version here: https://github.com/wozeparrot/tinygrad/tree/tutorial/

This got a lot larger than I was intending for it to be but oh well we got nicer docs now ;)

geohot:
I liked the copy and paste one-liner for lazy_matmul.

wozeparrot:
> I like the copy and paste one-liner for lazy_matmul. I use it at a test all the time, and don't want to edit a file.

The main reason for splitting it up is that the one liner is a bit jarring for newcomers, as the readme will be the first thing that they see so it should be as well layed out as possible.

> I think the first thing in the README should be llama and stable diffusion (have to make sure the weight autodownloaders are good) Meh, maybe just surfacing the showcase readme is good and mention LLaMA and Stable Diffusion.

I do like the split out showcase as it makes the docs directory structure cleaner, and it lets me mention the showcase in more places. Yea I think if I just put a link to the showcase at the top and mention llama and stable diffusion it should be fine.

geohot:
I need the one liner back

wozeparrot:
It might also be good to mention mlperf somewhere here. Not sure tho

geohot:
Big improvement, thanks!

docs/showcase.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Generative Adversarial Networks
+
+Take a look at [mnist_gan.py](/examples/mnist_gan.py).
+
+![mnist gan by tinygrad](/docs/showcase/mnist_by_tinygrad.jpg)
+
+### Stable Diffusion
+
+You will need to download the [weights](https://huggingface.co/CompVis/stable-diffusion-v-1-4-original/resolve/main/sd-v1-4.ckpt) of Stable Diffusion and put it into the [weights/](/weights) directory.
```
python273, line 44: it's auto downloading them

docs/showcase.md
```diff
@@ -0,0 +1,61 @@
...
+python3 examples/stable_diffusion.py
+```
+
+![a horse sized cat eating a bagel](/docs/showcase/stable_diffusion_by_tinygrad.jpg)
+
+*"a horse sized cat eating a bagel"*
+
+### LLaMA
+
+You will need to download and put the weights into the [weights/LLaMA](/weightsLLaMA) directory, which may need to be created.
```
python273, line 56: typo in the link, though not sure if link is needed

PR #901, opened 2023-06-02
nn init matches torch

geohot:


tinygrad/nn/__init__.py
```diff
@@ -48,16 +50,18 @@ class ConvTranspose2d:
...
 
   def __call__(self, x):
     return x.conv_transpose2d(self.weight, self.bias, padding=self.padding, output_padding=self.output_padding, stride=self.stride, dilation=self.dilation, groups=self.groups)
 
 class Linear:
-  def __init__(self, in_features, out_features, bias=True, initialization: str='kaiming_uniform'):
-    self.weight = getattr(Tensor, initialization)(out_features, in_features)
-    self.bias = Tensor.zeros(out_features) if bias else None
+  def __init__(self, in_features, out_features, bias=True):
+    self.weight = Tensor.kaiming_uniform(out_features, in_features, a=math.sqrt(5))
```
wozeparrot, line 62: I think you can just replace this with `Tensor.uniform` at least according to https://github.com/pytorch/pytorch/blob/683753fb0fc842afa9c43bc2a4576053af529429/torch/nn/modules/linear.py#L104-L111

geohot: Meh, I'll leave it to match torch until there's tests. We should really add distribution tests for common torch layers, I did them manually in the CIFAR training.

PR #905, opened 2023-06-02
Add .linspace()

xl0:
More from #861


jla524:
Looks good! Do you want to add some tests?

xl0:
@jla524 , do we have a test-it-all policy? I can add it, but also it just calls np.
I think the test should be added when we move it away from NumPy because then the test writers will have a better feel of the corner cases.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         2               0
```

geohot:
Calling numpy is bad. Really the numpy calls should be moved to a special section and should eventually all be removed.

Also, I have never used `linspace` If you insist on it, it must be an "hlop"

xl0:
I will look into how to rewrite `linspace` (and `arange` ?) without numpy as an exercise.

PR #909, opened 2023-06-02
Get passing Metal tests in CI

jla524:
Changes made:
- Add metal tests in GitHub workflow
- Fix metal segfaults
- Skip tests that fail on GitHub runners (pass on m1)

TODO:
- Check if skipped tests can be fixed


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        2               2
tinygrad/lazy.py                           2               2
total                                      6               6
```

test/models/test_mnist.py
```diff
@@ -94,13 +95,15 @@ def test_conv(self):
     train(model, X_train, Y_train, optimizer, steps=100)
     assert evaluate(model, X_test, Y_test) > 0.94   # torch gets 0.9415 sometimes
 
+  @unittest.skipIf(getenv("CI", "") and Device.DEFAULT == "METAL", "broken on some CI runners")
```
geohot, line 98: Hmm, why is this broken? This is a very important test

tinygrad/runtime/ops_cpu.py
```diff
@@ -27,7 +27,7 @@ def mulacc(a, b, new_shape):
   return mulacc
 
 numpy_fxn_for_op: Dict[Op, Callable] = {**base_fxn_for_op, **{
-  UnaryOps.NOOP: lambda x: np.require(x, requirements='C'), UnaryOps.EXP: np.exp, UnaryOps.LOG: np.log, UnaryOps.CAST: lambda x,y: x.astype(y.np), UnaryOps.SIN: np.sin,
+  UnaryOps.NOOP: np.ascontiguousarray, UnaryOps.EXP: np.exp, UnaryOps.LOG: np.log, UnaryOps.CAST: lambda x,y: x.astype(y.np), UnaryOps.SIN: np.sin,
```
geohot, line 30: Is this different?

geohot:
The copies introduce a big time cost, what's the absolute minimum case you need them for? Also, why are the tests being skipped for METAL? I'm only okay with this if there's some good first principles explanation.

test/models/test_mnist.py
jla524: It really depends on the CI runners. Sometimes it passes, and sometimes it fails. So unfortunately I have to skip this.

jla524:
> The copies introduce a big time cost, what's the absolute minimum case you need them for? This PR would introduce them for all backends.

I understand. It's really to avoid segfaults when using the metal backend. I had to switch to GPU when I worked on UNet3D inference, because METAL would randomly segfault.

geohot:
Are the segfaults really random? My guess would be use after frees, probably out of scoping something that needs to be kept around. This is Apple, I doubt they are race conditions or anything terrible.

jla524:
> Also, why are the tests being skipped for METAL? I'm only okay with this if there's some good first principles explanation.

Those tests passed on my m1 and intel Mac (except `test_vit`, we can look into it), but on the CI runners they fail about 60% of the time. I figured this out when I pushed an empty commit, and the test went from failing to passing.

jla524:
> Are the segfaults really random? My guess would be use after frees, probably out of scoping something that needs to be kept around (do they still happen if you disable the GC?). This is Apple, I doubt they are race conditions or anything terrible.

Yeah they were quite random. Usually it happens at the start the run, but occasionally it's near the end of the run (after a few hours). I haven't tried disabling the GC, will check and see if it makes a difference.

tinygrad/runtime/ops_cpu.py
jla524: It should be the same. This was from reverting the "no copy" commit. I'll change it back.

jla524:
This is really confusing… tests are passing on my m1 (including the skipped ones), yet the ones in CI are failing left and right

geohot:
`test_conv_with_bn` usually fails if things are broken with assignment.

geohot:
See #931 I think that's (at least one of) the segfault issues.

jla524:
> See #931 I think that's (at least one of) the segfault issues.

I added a class in helpers (+7 lines) to stored the buffer and I think it fixed the segfault.

Not sure what to do about CI runners though. I can't reproduce those results locally.

.github/workflows/test.yml
```diff
@@ -60,6 +60,22 @@ jobs:
...
+     - name: Checkout Code
+       uses: actions/checkout@v3
+     - name: Set up Python 3.10
+       uses: actions/setup-python@v4
+       with:
+         python-version: '3.10'
+     - name: Install Dependencies
+       run: pip install -e '.[metal,testing]'
+     - name: Run Pytest
+       run: python -m pytest -s -v -n=auto test/
```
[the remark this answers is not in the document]
jla524: Since the metal dependencies are installed, it should use the metal backend by default
https://github.com/geohot/tinygrad/blob/e8a23d43311ad751bb23919657e76a0e7382358d/tinygrad/lazy.py#L320-L325

tinygrad/helpers.py
```diff
@@ -82,6 +82,14 @@ def from_np(x) -> DType: return asdict(dtypes())[np.dtype(x).name]
   int64: Final[DType] = DType(2, 8, "int64", np.int64)
   uint8: Final[DType] = DType(0, 1, "uchar", np.uint8)
 
+class NPArray(np.ndarray):
```
kposborne2, line 85: I see you implemented the wrapper class to permit adding a property, however, the `MetalBuffer` is what we need to keep alive, I tried this change to your branch on `ops_metal.py`, and it passed my [segfaulting test](https://github.com/geohot/tinygrad/issues/931) :) 
```
diff --git a/tinygrad/runtime/ops_metal.py b/tinygrad/runtime/ops_metal.py
index ccb0af04..d105ec1e 100644
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -1,9 +1,10 @@
 # pip3 install pyobjc-framework-Metal pyobjc-framework-Cocoa pyobjc-framework-libdispatch
 import os, subprocess, pathlib
 import Metal, Cocoa, libdispatch # type: ignore
+import numpy as np
 from typing import List, Any
 from tinygrad.codegen.cstyle import CStyleCodegen, CStyleLanguage
-from tinygrad.helpers import prod, getenv, DEBUG, DType
+from tinygrad.helpers import prod, getenv, DEBUG, DType, NPArray
 from tinygrad.ops import Compiled
 from tinygrad.runtime.lib import RawBufferMapped
 
@@ -25,9 +26,9 @@ class RawMetalBuffer(RawBufferMapped):
   def __del__(self):
     self._buf.release()
     super().__del__()
-  def _buffer(self):
+  def toCPU(self) -> np.ndarray:
     METAL.synchronize()
-    return self._buf.contents().as_buffer(self._buf.length())
+    return NPArray(np.frombuffer(self._buf.contents().as_buffer(self._buf.length()), dtype=self.dtype.np),buffer=self)
 
 def unwrap(x):
   ret, err = x
```

jla524: Thanks @kposborne2! This fixes most of the test fails on my intel Mac. I'm not sure why tests are still failing in CI though...

jla524:
Yeah.. tests are failing randomly and I’m not able to reproduce it locally. I suspect this is a problem with the CI runners. I’ll put this on hold for now.

tinygrad/runtime/ops_metal.py
```diff
@@ -67,6 +72,10 @@ def __call__(self, global_size, local_size, *bufs, wait=False):
 
     assert prod(local_size) <= self.pipeline_state.maxTotalThreadsPerThreadgroup(), f"local size {local_size} bigger than {self.pipeline_state.maxTotalThreadsPerThreadgroup()} with exec width {self.pipeline_state.threadExecutionWidth()} memory length {self.pipeline_state.staticThreadgroupMemoryLength()}"
     command_buffer = METAL.mtl_queue.commandBuffer()
+    command_encoder = command_buffer.computeCommandEncoder()
+    if i, a in enumerate(bufs): command_encoder.setBuffer_offset_atIndex_(a._buf, 0, i)
```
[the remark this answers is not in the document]
jla524: Thanks @Maroc-OS . This passes on my m1, but then again, fails in the CI runners.

geohot:
Then Jacky will get $200 once this is merged :)

I think there's a few issues, and I suggest making small PRs to address them. The use-after-free on numpy is top priority, I don't like the new class, there has to be a way to register the backing buffer with numpy.

tinygrad/helpers.py
geohot: There has to be a better way to do this. There's no way to register a backing buffer with numpy? I really don't like the subclass.

geohot:
Oh yea...OOMs are segfaults in CI, maybe it's that? We probably should add an assert for this.

jla524:
> Oh yea...OOMs are segfaults in CI, maybe it's that? We probably should add an assert for this.
> 
> If it's fast enough, we don't have to run the tests in parallel. That can introduce OOM nondeterminism.

I don't get it ._. the Mac runners should have more RAM than the linux runners

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

jla524:
We might have a problem with OpenCL on Intel iGPUs. I'll try to reproduce the error on my Intel Mac.
https://community.amd.com/t5/opencl/question-about-opencl-local-work-size-local-size-256-1-doesn-t/td-p/430396

jla524:
I have to skip `test_speed_v_torch.py` due to the lack of VRAM in the Mac CI runner. Although we can probably add an option for `TINY=1` that reduces the size of Tensors and allow us to run the test.

Most of the tests in `test_ops.py` in are fine, except `test_mul_const` and `test_div_const` which will fail for backward. I can't reproduce this either, but I'll take a look at the code for this to try and understand the issue.

test/test_ops.py
```diff
@@ -166,13 +166,15 @@ def test_neg(self):
   def test_mul(self):
     helper_test_op([(64,64), (64,64)], lambda x,y: x*y, Tensor.mul)
     helper_test_op([(), ()], lambda x,y: x*y, Tensor.mul)
+  @unittest.skipIf(getenv("CI", "") != "" and Device.DEFAULT == "METAL", "broken in METAL CI")
```
geohot, line 169: These need to be fixed, I don't see why they should be broken. I'm fine with skipping ViT

tinygrad/runtime/ops_metal.py
```diff
@@ -64,7 +64,6 @@ def __call__(self, global_size, local_size, *bufs, wait=False):
     if local_size is None: local_size = [32]
     local_size += [1] * (3-len(local_size))
 
-    assert prod(local_size) <= self.pipeline_state.maxTotalThreadsPerThreadgroup(), f"local size {local_size} bigger than {self.pipeline_state.maxTotalThreadsPerThreadgroup()} with exec width {self.pipeline_state.threadExecutionWidth()} memory length {self.pipeline_state.staticThreadgroupMemoryLength()}"
```
geohot, line 67: Why was this removed?

.github/workflows/test.yml
```diff
@@ -114,6 +114,33 @@ jobs:
...
+        python-version: '3.10'
+    - name: Install Dependencies
+      run: pip install -e '.[metal,testing]'
+    # Run separately to avoid OOM
+    - name: Run Pytest (extra)
+      run: METAL=1 python -m pytest -s -v -n=auto test/extra
+    - name: Run Pytest (unit)
+      run: METAL=1 python -m pytest -s -v -n=auto test/unit
+    - name: Run Pytest (mnist models only)
+      run: METAL=1 python -m pytest -s -v test/models --ignore=test/models/test_mnist.py
```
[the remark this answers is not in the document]
jla524: updated

test/test_ops.py
jla524: looks like they are passing now, just have to separate them out

tinygrad/runtime/ops_metal.py
jla524: it was for testing, I've reverted this change

test/test_ops.py
jla524: I discussed with @cloud11665 since he wrote the tests. He believes the test fail is caused by the architectural difference between the iGPU (in CI runners) vs the dGPUs I'm testing on.

jla524:
Only `test_conv_with_bn` is failing now. It's likely memory related.

test/test_dtype.py
```diff
@@ -52,6 +52,7 @@ def test_half_mul_upcast_float(self): _test_mul_upcast(Tensor([1,2,3,4], dtype=d
   def test_half_matmul_upcast_float(self): _test_matmul_upcast(Tensor([[1,2],[3,4]], dtype=dtypes.float16), Tensor.eye(2, dtype=dtypes.float32), dtypes.float32, [[1,2],[3,4]])
   def test_int8_matmul_upcast_half(self): _test_matmul_upcast(Tensor([[1,2],[3,4]], dtype=dtypes.int8), Tensor.eye(2, dtype=dtypes.float16), dtypes.float16, [[1,2],[3,4]])
 
+@unittest.skipIf(Device.DEFAULT == "METAL", "metal doesn't support float64")
```
geohot, line 55: You can skip this for any OSX

tinygrad/runtime/ops_metal.py
geohot: No it's not. I wrote that assert.

.github/workflows/test.yml
```diff
@@ -114,9 +119,45 @@ jobs:
...
+      run: pip install -e '.[metal,testing]'
+    # Run separately to avoid OOM
+    - name: Run Pytest (extra)
+      run: METAL=1 python -m pytest -s -v -n=auto test/extra
+    - name: Run Pytest (unit)
+      run: METAL=1 python -m pytest -s -v -n=auto test/unit
+    - name: Run Pytest (base excluding ops and test_speed_v_torch)
+      run: METAL=1 python -m pytest -s -v test --ignore=test/external --ignore=test/extra --ignore=test/unit --ignore=test/models --ignore=test/test_ops.py --ignore=test/test_speed_v_torch.py
+
+    - name: Run Pytest (ops excluding test_mul_const and test_div_const)
```
geohot, line 143: Why do those ones have to be tested separately? This likely points to issues with the method cache.

jla524: they are combined now, looks much cleaner

jla524:
Closing this. Bye metal 🥲

PR #911, opened 2023-06-03
Allow Tensor(tuple)

xl0:


tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
Is there a real use for this?

xl0:
I was using it in my code that was ported to tinygrad, and it works in torch, so don't see a reason it should not work in tinygrad.

geohot:
kk, want to rebase?

xl0:
Idk man, every time I add a small thing that I use in torch but does not work in tinygrad you give me the same reaction. I get it, it's PR noise, and I don't expect a medal for any of them, but what do you want me to do? I can't accumulate them because the code base moves fast.

geohot:
So we shouldn't just support things because torch does. The downside I see to tuples is it's easier to get confused with a shape. I'll merge it if you want, but I like a justification for the feature, and if it's not worth that, it's probably not worth having.

xl0:
Rebased.


I used tuple in my code because Python still does not have `const`, so when I define a list that should be constant, I tend to make it a tuple instead, just as a good habit in case I try to mutate it down the line by mistake.

Another use case would be
```
def asdf( value = (1,2,3,4,5) ):
    t = Tensor(value)
    ...
```
Using a list as a default value is a bad idea.



But in reality, the reason to allow tuples is that both NumPy and PyTorch (and probably TF too) allow it. It just saves a bunch of time and frustration for people porting code.

Look, I agree that if someone is just going through the API and generates PRs for the low-hanging differences, it's not a good use of your (and their) time. But what do you want us to do when we are porting code and run into those differences? What should be the workflow?

I see 3 option:
1. Do nothing - unless it's really bad, just work around differences.
2. Submit a PR and expect to engage in a discussion over stupid shit like `.linspace` vs `.arange` or the use of tuples.
3. Submit a PR and expect it to get merged if it works like in Torch and NumPy, even if George does not use the API in the same way.

Can do 1. or 3. I feel that 2. is not a good use of our time.

PR #912, opened 2023-06-03
Fixed test suite for Clip

stevenandersonz:
Tests fixed:
- test_clip_default_inbounds_cpu 
- test_clip_default_max_cpu 

Before:
=== 294 failed, 532 passed, 1808 skipped, 2 warnings in 32.39s ====
After:
=== 292 failed, 534 passed, 1808 skipped, 2 warnings in 33.74s ====

### DIFF
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -153,8 +153,9 @@ def Softmax_13(input, axis=-1): return input.softmax(axis)
 Softmax = {1: Softmax_1, 13: Softmax_13}   # Softmax default axis changed
 def LogSoftmax(input, axis=-1): return input.log_softmax(axis)
 def Clip(input, min:Tensor=None, max:Tensor=None):
-  max_, min_ = Tensor([3.4e38]) if max is None else max, Tensor([-3.4e38]) if min is None else min
-  return  input if min is None and max is None else (-input if min is None else input).clip(min_,max_)
+  if min is None: min = Tensor([-3.4e38])
+  if max is None: max = Tensor([3.4e38])
+  return input.clip(min, max)
 
 def Sin(x): return x.sin()
 def Cos(x): return x.cos()
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -152,9 +152,9 @@ def Softmax_1(input, axis=1): return input.softmax(axis)
 def Softmax_13(input, axis=-1): return input.softmax(axis)
 Softmax = {1: Softmax_1, 13: Softmax_13}   # Softmax default axis changed
 def LogSoftmax(input, axis=-1): return input.log_softmax(axis)
-def Clip(input, min:Tensor=None, max:Tensor=None):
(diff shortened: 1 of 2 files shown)


geohot:
Very confusing code. What's the real issue?

stevenandersonz:
You're right, I'll do better.  I was trying to go around Tensor.clip() for cases like min == [-3.4e38], max == [0]  

```python
h = np.array([-0.67246044, -0.35955316,   0.17742614]).astype(np.float32)
h.clip(-3.4e38, 0) #                                -> [-0.67246044  -0.35955316   0.        ]
Tensor(h).clip(-3.4e38, 0).numpy() #  -> [0.          0.         -0.17742614]
```

I'll fix the clip function inside tensor.py and write a test for it.

tinyb0t:
Changes made in `tinygrad/`:
```
```

extra/onnx_ops.py
```diff
@@ -152,8 +152,10 @@ def Softmax_1(input, axis=1): return input.softmax(axis)
 def Softmax_13(input, axis=-1): return input.softmax(axis)
 Softmax = {1: Softmax_1, 13: Softmax_13}   # Softmax default axis changed
 def LogSoftmax(input, axis=-1): return input.log_softmax(axis)
-def Clip(input, min=-3.4e38, max=3.4e38): return input.clip(min, max)
-
+def Clip(input, min:Tensor=None, max:Tensor=None):
```
geohot, line 155: This is unneeded

stevenandersonz: Do you mean the whole function or just the typings?

extra/onnx_ops.py
```diff
@@ -152,8 +152,10 @@ def Softmax_1(input, axis=1): return input.softmax(axis)
 def Softmax_13(input, axis=-1): return input.softmax(axis)
 Softmax = {1: Softmax_1, 13: Softmax_13}   # Softmax default axis changed
 def LogSoftmax(input, axis=-1): return input.log_softmax(axis)
-def Clip(input, min=-3.4e38, max=3.4e38): return input.clip(min, max)
-
+def Clip(input, min=None, max=None):
+  if min is None: min = -3.4e38
```
geohot, line 156: Ugh, really?

stevenandersonz: I know, the problem is that Onnx Model calls the Clip function with an explicit None for min.
```python
# onnx.py - line 88 
# inp -> [<Tensor buffer<60, dtypes.float> on TORCH with grad None>, None, <Tensor buffer<1, dtypes.float> on TORCH with grad None>]
real_fxn(*inp, **opt)
```
That's why it would never default min to -3.4e38 and the test fails without the if.

stevenandersonz: Wait what I'm I doing I can create an elif for this function and do what's needed there.

geohot: Ahh, I see. Meh, onnx is dumb, will merge.

PR #915, opened 2023-06-03
JIT support input size changing

amatraturi:
Added support for input size changing (as per https://github.com/geohot/tinygrad/issues/692)
Added Test for size changing jit (passes for this PR)
Let me know if this is proper.

### DIFF
--- a/test/test_jit.py
+++ b/test/test_jit.py
@@ -75,11 +75,9 @@ class TestJit(unittest.TestCase):
     def add(a, b): return (a+b).realize()
     for i in range(20):
       if i < 10:
-        a = Tensor.randn(10, 10)
-        b = Tensor.randn(10, 10)
+        a,b = Tensor.randn(10, 10),Tensor.randn(10, 10)
       else:
-        a = Tensor.randn(20, 20)
-        b = Tensor.randn(20, 20)
+        a,b = Tensor.randn(20, 20),Tensor.randn(20, 20)
       d = add(a,b)
       np.testing.assert_equal(d.numpy(), a.numpy()+b.numpy())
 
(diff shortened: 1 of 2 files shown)


geohot:
Code. No tests.

Does this add new behavior? Why isn't the new behavior tested?

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/jit.py                           24              14
```

amatraturi:
> If you add code, you have to add tests.
> 
> Does this add new behavior? The new behavior needs to be tested.

Yes, understood. Added the test for input size changing, passes with this PR.

Existing usage (same size) would behave same as previously.

tinygrad/jit.py
```diff
@@ -19,33 +19,41 @@ def __get__(self, obj, objtype): return functools.partial(self.__call__, obj)
 
   def __call__(self, *args, **kwargs) -> Any:
     if Device.DEFAULT not in ["GPU", "CLANG", "METAL", "CUDA"]: return self.fxn(*args, **kwargs)  # only jit on the GPU codegen
+
     # NOTE: this cast is needed since although we know realize will create a ".realized" DeviceBuffer, the type checker doesn't
     input_rawbuffers: Dict[Union[int, str], RawBuffer] = {cast(Union[int, str], k):cast(RawBuffer, v.realize().lazydata.realized) for k,v in itertools.chain(enumerate(args), kwargs.items()) if isinstance(v, Tensor)}
+    specialised_key = f'{hash("".join([str(input_rawbuffers[i].size) for i,k in input_rawbuffers.items()]))}'
```
geohot, line 25: What if the shapetrackers are different? Like permuted? Is just checking the size enough?

test/test_jit.py
```diff
@@ -70,5 +70,16 @@ def __call__(self, b:Tensor) -> Tensor:
       c = fun(b)
       np.testing.assert_equal(c.numpy(), fun.a.numpy()+b.numpy())
 
+  def test_shape_change_jit(self):
```
geohot, line 73: I will merge just this test with an expected failure for now if you want to submit it.

PR #920, opened 2023-06-04
Eliminate LoadOps.FROMCPU

kposborne2:
- new `LazyBuffer.fromCPU` method to create from `ndarray` + changing the `__init__` method to have optional `LazyOp`/`RawBuffer`
- Add `int32` support for `RawMallocBuffer`

### DIFF
--- a/tinygrad/lazy.py
+++ b/tinygrad/lazy.py
@@ -7,6 +7,7 @@ from tinygrad.helpers import prod, getenv, DType, dtypes, flatten, ImageDType, D
 from tinygrad.shape.shapetracker import ShapeTracker, get_contraction
 from tinygrad.ops import Compiled, Interpreted, UnaryOps, BinaryOps, ReduceOps, MovementOps, LoadOps, OpType, LazyOp, get_lazyops, get_buffers, map_buffers
 from tinygrad.runtime.lib import RawConst, RawBuffer, RawBufferMapped
+from tinygrad.runtime.ops_cpu import RawNumpyBuffer
 from tinygrad.runtime.ops_disk import RawDiskBuffer
 
 # lazy can recurse a lot
@@ -105,6 +106,7 @@ class LazyBuffer:
     if GRAPH >= 3: log_op(self, self.op, phantom=True)
 
   def __repr__(self): return f"<LB {self.shape} {self.dtype} op:{self.op.op if self.realized is None else self.realized} st:{self.st}>"
+  def _device_extra_args(self) -> Dict[str, str]: return {"device": self.device.split(":")[1]} if ":" in self.device else {}
 
   def realize(self:LazyBuffer) -> LazyBuffer:
     if self.realized is None:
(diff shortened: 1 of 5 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                          20              17
tinygrad/runtime/lib.py                    7               1
total                                     29              21
```

geohot:
This is +8 lines. It should be -10 lines.

kposborne2:
Were you thinking of a different overall approach? Or is there other code that can be also removed now that this capability is available? I didn't really see one that could actually reduce the line count by 10 - if I remove the bugfix for empty tensors, inline the LazyBuffer construction method to tensor `__init__` (requires `ShapeTracker` import), and remove the assertion, it is still only -2 on the net, didn't really see a way to get to -10

tinygrad/tensor.py
```diff
@@ -44,8 +45,7 @@ def __init__(self, data:Union[int, float, list, LazyBuffer, np.ndarray], device=
       assert dtype is None or dtype == data.dtype, "dtype doesn't match, and casting isn't supported"
       lazydata = data if data.device == device else LazyBuffer.loadop(LoadOps.FROM, data.shape, data.dtype, device, src=data)
     elif isinstance(data, np.ndarray):
-      # TODO: create CPUBuffer directly
-      lazydata = LazyBuffer.loadop(LoadOps.FROMCPU, data.shape, dtypes.from_np(data.dtype), device, data)
+      lazydata = LazyBuffer(device, ShapeTracker(tuple(data.shape)), LoadOps, Device[device].buffer.fromCPU(data, **Device.extra_args(device)), dtypes.from_np(data.dtype))
```
geohot, line 48: Create a CPUBuffer then you can copy it later.

tinygrad/lazy.py
```diff
@@ -87,32 +87,29 @@ def create_lazybuffer(device:str, shape:Union[ShapeTracker, Tuple[int, ...]], op
 
 class LazyBuffer:
   __deletable__ = ('op',)
-  def __init__(self, device:str, st:ShapeTracker, optype:OpType, op:LazyOp, dtype:DType):
+  def __init__(self, device:str, st:ShapeTracker, optype:OpType, src:Union[LazyOp, RawBuffer], dtype:DType):
```
geohot, line 90: I'd make a different constructor for this.

tinygrad/tensor.py
kposborne2: I see, so it should still defer the load if the device is not the cpu, that can simplify it

tinygrad/lazy.py
kposborne2: WDYT of a `fromCPU` method that takes the `np.ndarray`? The current `RawBuffer` does not remember what was its device

geohot:
Nice. What about that TODO?

kposborne2:
> Nice. What about that TODO?

In the original implementation, I added a new EmptyBuffer class to handle buffers with 0 size, I reverted that to reduce the lines added. It is contained in this diff here: https://github.com/geohot/tinygrad/pull/920/commits/bbdcdc8713f60725012d9f1602e18de4bdd3a4ed, does that solution make sense to you? Note that this type of tensor is already broken in master if you try to call `.realize()` on it.

geohot:
Hmm, I don't think a buffer should ever be 0 size, just 0 dim.

kposborne2:
I see, since the loading for non-cpu devices is deferred again, it won't hit the issue, so I will restore that test

geohot:
Nice work!

geohot:
Hmm, this broke docs/abstractions.py (which isn't tested in CI)

Want to fix and add to tests?

kposborne2:
> Hmm, this broke docs/abstractions.py (which isn't tested in CI)
> 
> Want to fix and add to tests?

Yes, I can do that

geohot:
Done!

narangkay:
Note that without a bump to the pip release, running (now updated) abstractions.py in colab gives me

```
LazyOp(op=<LoadOps.FROMCPU: 1>, src=(), arg=<tinygrad.helpers.LazyNumpyArray object at 0x7fd0e8112140>)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
[<ipython-input-7-7442bbed48f4>](https://localhost:8080/#) in <cell line: 3>()
      1 print(lazyop.src[0].op)
      2 assert lazyop.src[0].op.op == LoadOps.FROMCPU
----> 3 assert lazyop.src[0].op.src[0].realized.toCPU()[0] == 2, "the arg of the FROM LazyOP is a LazyBuffer holding [2.]"
      4 assert result.lazydata.realized is None, "the LazyBuffer is not realized yet"

IndexError: tuple index out of range
``` 

Might be worth adding some sort of version check in abstractions.py

PR #921, opened 2023-06-04
Refactor

Uzaaft:
List of refactors:
- Refactor comprehensions to generators
- Simplify and cleanup different if else tests
- Cleanup nested if-else logic
- Attempt to use more of walruss operator
- f-strings


Uzaaft:
Swapping `tuple()` with `()` microbenchmark:
image

tinyb0t:
Changes made in `tinygrad/`:
```
```

Uzaaft:
Benchmark with f-string vs Concatinations:
Concatinations:
image

f-string:
image

Pretty similar

tinygrad/codegen/linearizer.py
```diff
@@ -202,14 +201,8 @@ def global_store(self, i, idxs:List[Variable], store:List[Token], ssa) -> None:
...
     # float4 grouping (optional)
-    should_upcast = self.supports_float4 and (self.bufs[i].dtype not in (dtypes.float16, dtypes.int8, dtypes.uint8)) and len(self.float4_axis(i)) == 1
-    if should_upcast:
-      store_offset_new = {}
-      for k,out_tokens in self._group_float4(i, store_offset).items():
-        if all_same([x.name for x in out_tokens]) and tuple(range(4)) == tuple(x.offset for x in out_tokens):
-          store_offset_new[k] = Token(out_tokens[0].name, LocalTypes.float4)
-        else:
-          store_offset_new[k] = self.uop(UOps.CAST, ssa("alu", LocalTypes.float4), out_tokens)
+    if (should_upcast := self.supports_float4 and (self.bufs[i].dtype not in (dtypes.float16, dtypes.int8, dtypes.uint8)) and len(self.float4_axis(i)) == 1):
```
Uzaaft, line 204: In theory, we shouldn't need the `should_upcast` variable  now.

tinygrad/codegen/linearizer.py
```diff
@@ -171,8 +171,7 @@ def global_load(self, i, idxs:List[Variable], const=None) -> List[Token]:
     load_offset: Dict[Tuple[int, ...], Any] = {uidxs:(LocalTypes.float,uidxs)+self.sts[i].expr_idxs(idxs+[Variable.num(x) for x in uidxs[::-1]]) for uidxs in self.shape_offsets(i)}
 
     # float4 grouping (optional)
-    should_upcast = self.supports_float4 and len(self.float4_axis(i)) == 1
-    if should_upcast:
+    if self.supports_float4 and len(self.float4_axis(i)) == 1:
```
Uzaaft, line 174: In theory, we shouldn't need the `should_upcast` variable  now.

tinygrad/codegen/linearizer.py
```diff
@@ -171,8 +171,7 @@ def global_load(self, i, idxs:List[Variable], const=None) -> List[Token]:
     load_offset: Dict[Tuple[int, ...], Any] = {uidxs:(LocalTypes.float,uidxs)+self.sts[i].expr_idxs(idxs+[Variable.num(x) for x in uidxs[::-1]]) for uidxs in self.shape_offsets(i)}
 
     # float4 grouping (optional)
-    should_upcast = self.supports_float4 and len(self.float4_axis(i)) == 1
-    if should_upcast:
+    if should_upcast := self.supports_float4 and len(self.float4_axis(i)) == 1:
```
Uzaaft, line 174: In theory, we shouldn't need the `should_upcast` variable  now.

tinygrad/codegen/linearizer.py
Uzaaft: Removing the variable due to the linter.

tinygrad/codegen/linearizer.py
Uzaaft: Removing the variable due to the linter.

geohot:
PRs should never mix whitespace and functional code changes. Do you trust the tests enough that *nothing* at all broke?

roelofvandijk:
Using generators instead of list comprehensions might be slower in some instances.

Uzaaft:
> Using generators instead of list comprehensions might be slower in some instances.

That's true.

But from geohots streams, I've understood that he wants things to be lazy(as much as possible), and I believe generators should then be used instead.

roelofvandijk:
I have no idea what the priority is - @rayanht is going the other way (generators -> list comprehensions) [here](https://github.com/geohot/tinygrad/pull/812/files#diff-7703e7b63c9844f019ae3609629fd9c9bab290d11ae6608ed2d09edd49bab2dbR250). The performance gain might be important in very hot functions.

rayanht:
generators are way slower in places where we use them because the iterators are quite small so the generator overhead dominates

Uzaaft:
> generators are way slower in almost all places where we currently use them because the iterators are quite small so the generator overhead dominates

Today I learnt this. I'll remove the commits where comprehensions are converted to generators.
Perhaps this would be worth revisiting in the future. Seems like there is a regression in generators going from python 3.10 to 3.11

geohot:
I don't think this is much easier to read, and I don't trust the tests enough to merge this.

PRs that look like changes all over the codebase without new functionality or a demonstrable 20%+ speedup won't be merged.

PR #922, opened 2023-06-04
Add cumsum with n-dim inputs

tomatillos:
Extended the existing implementation to work with n-dimensional inputs with a specified axis. Also added relevant cumsum tests.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -81,6 +81,7 @@ class TestOps(unittest.TestCase):
   def test_eye(self):
     helper_test_op([], lambda: torch.eye(10), lambda: Tensor.eye(10), forward_only=True)
 
+  @unittest.skipIF(IMAGE>0, "no arange on images")
   def test_arange(self):
     helper_test_op([], lambda: torch.arange(10), lambda: Tensor.arange(10), forward_only=True)
   def test_where(self):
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

tomatillos:
Oh, have used that conv2d works with more than 2 image dimensions, but nn/image.py::image_conv2d only allows two...

test/test_tensor.py
```diff
@@ -190,5 +190,14 @@ def test_element_size(self):
       dtype = f.default
       assert dtype.itemsize == Tensor.randn(3, dtype=dtype).element_size(), f"Tensor.element_size() not matching Tensor.dtype.itemsize for {dtype}"
 
+  def test_cumsum(self):
```
geohot, line 193: Should be in test_ops. It's okay if multidimensional doesn't work with images, but you have to skip them appropriately

test/test_ops.py
```diff
@@ -81,6 +81,7 @@ def test_ones_like(self):
   def test_eye(self):
     helper_test_op([], lambda: torch.eye(10), lambda: Tensor.eye(10), forward_only=True)
 
+  @unittest.skipIf(IMAGE>0, "no arange on images")
```
geohot, line 84: Why? I want single dim cumsum to work on images.

tomatillos: Fair enough. Cumsum works on images, it's arange that fails because arange asks for cumsum of a 1-d tensor. My implementation of cumsum relies on conv2d taking arbitrary many dimensions, however image_conv2d can only take 2d kernels.

Can see obvious options as:

Option 1: rewrite cumsum to only call conv2d with a true 2d kernel. This should take +1 or +2 lines of code, requiring some permutes before and after the 1d cumsum implementation.

Option 2 (hackier): Make arange call cumsum on a 2d tensor with shape (1,n) and flatten after. No extra lines of code, and arange now works on images.

geohot: Option 1 sounds good, I'm okay with a few lines.

tomatillos: This seems to work nicely now.

geohot:
Thanks!

Issue #834, opened 2023-05-28
examples/stable_diffusion exception on first run


rweichler:
race condition maybe?

```
tinygrad$ PYTHONPATH="." CLANG=1 python3 examples/stable_diffusion.py
https://huggingface.co/CompVis/stable-diffusion-v-1-4-original/resolve/main/sd-v1-4.ckpt: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4.27G/4.27G [35:44<00:00, 1.99MB/s]
loading 1144 ram used:  4.26 GB: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▊| 1144/1145 [00:03<00:00, 367.86it/s]
Traceback (most recent call last):
  File "examples/stable_diffusion.py", line 620, in <module>
    dat = fake_torch_load_zipped(open(FILENAME, "rb"))
  File "/Users/reed/code/gh/geohot/tinygrad/extra/utils.py", line 146, in fake_torch_load_zipped
    if future.exception() is not None: raise future.exception()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/reed/code/gh/geohot/tinygrad/extra/utils.py", line 139, in load_weight
    load_single_weight(v[2], myfile, v[3], v[4], v[0], v[5], mmap_allowed=True)
  File "/Users/reed/code/gh/geohot/tinygrad/extra/utils.py", line 114, in load_single_weight
    myfile.readinto(t.lazydata.realized._buffer())
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/zipfile.py", line 940, in read
    data = self._read1(n)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/zipfile.py", line 1030, in _read1
    self._update_crc(data)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/zipfile.py", line 958, in _update_crc
    raise BadZipFile("Bad CRC-32 for file %r" % self.name)
zipfile.BadZipFile: Bad CRC-32 for file 'archive/data/1143'
```

[...]

PR #923, opened 2023-06-04
quick doc fixups

wozeparrot:
Followup to #900 with some additional fixes.

Gonna start documenting API after this.

Re stable diffusion: we should probably make sure that #834 isn't happening anymore.

### DIFF
--- a/docs/showcase.md
+++ b/docs/showcase.md
@@ -41,8 +41,6 @@ Take a look at [mnist_gan.py](/examples/mnist_gan.py).
 
 ### Stable Diffusion
 
-You will need to download the [weights](https://huggingface.co/CompVis/stable-diffusion-v-1-4-original/resolve/main/sd-v1-4.ckpt) of Stable Diffusion and put it into the [weights/](/weights) directory.
-
 ```sh
 python3 examples/stable_diffusion.py
 ```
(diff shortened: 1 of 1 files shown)


docs/showcase.md
```diff
@@ -53,7 +51,7 @@ python3 examples/stable_diffusion.py
 
 ### LLaMA
 
-You will need to download and put the weights into the [weights/LLaMA](/weightsLLaMA) directory, which may need to be created.
+You will need to download and put the weights into the `weights/LLaMA` directory, which may need to be created.
```
python273, line 54: maybe link to llama-dl? 🤔 

https://github.com/shawwn/llama-dl/blob/56f50b96072f42fb2520b1ad5a1d6ef30351f23c/llama.sh

wozeparrot: Are we legally allowed to do so? Just cuz of the license on LLaMA and that the original download links are behind the signup wall.

geohot: We should probably switch to MPT or something by default. I don't think we should link to it.

wozeparrot: There does exist OpenLLaMA: https://github.com/openlm-research/open_llama

geohot: Is it as good as LLaMA? We really need a LM test suite. Unlocked that bounty.

* Get LLaMA 7B running with uint8 weights (leave float16 support too)
* Evaluate both the float16 one and uint8 one with ARC+HellaSwag+MMLU+TruthfulQA and confirm they match https://huggingface.co/spaces/HuggingFaceH4/open_llm_leaderboard

wozeparrot: They claim that it has performance comparable to llama and better in some places.

I'll take a crack at getting the eval harness working. Do you want to depend on the harness or try to implement everything here as much as possible?

Issue #925, opened 2023-06-04
`setup.py` package dependency issues


the-praxs:
1. `README` should contain a list of the extra dependencies.

2. `python3 -m pip install -e '.[testing]'` produces an error: This works instead: `python3 -m pip install -e .[testing]`.

3. `llvm` optional dependency requires additional packages to `llvmlite` for proper installation.

4. `cuda` optional dependency installation fails on Windows because it is unable to create a wheel file for `pycuda`. Additionally `Microsoft C++ Build Tools` are required as pre-requisite to start the process of building the `pycuda` wheel. A workaround to this is to download and install the pycuda wheel from [here](https://www.lfd.uci.edu/~gohlke/pythonlibs/#pycuda).

5. `triton` optional dependency is broken. No suitable version is found to satisfy the dependencies.

6. `metal` optional dependency should be mentioned in the `README` to be specific to Mac only.

PR #926, opened 2023-06-04
fix package dependencies

the-praxs:
Fixing #925

TODO:
- [x] `llvm` fixed
- [ ] `README` add info
- [ ] Fix command for optional dependency
- [ ] Fix `cuda`
- [ ] Fix `triton`
- [ ] Add warning for `metal` on windows

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(name='tinygrad',
       install_requires=['numpy', 'requests', 'pillow', 'tqdm', 'networkx', 'pyopencl'],
       python_requires='>=3.8',
       extras_require={
-        'llvm': ["llvmlite"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -22,7 +22,7 @@
       install_requires=['numpy', 'requests', 'pillow', 'tqdm', 'networkx', 'pyopencl'],
       python_requires='>=3.8',
       extras_require={
-        'llvm': ["llvmlite"],
+        'llvm': ["llvmlite", "pytz", "requests-oauthlib", "Click", "scikit-learn", "scipy"],
```
python273, line 25: ???

the-praxs: `llvmlite` install requires these packages too

geohot:
Huh? It should auto install them. This PR doesn't make sense.

PR #930, opened 2023-06-05
Update broken `docs/abstractions.py` for changed ops, and add to CI

kposborne2:
After https://github.com/geohot/tinygrad/pull/920, the docs were broken, by still using the `FROMCPU` LoadOp, updated and added to ci


wozeparrot:
Can you update the readme and https://github.com/geohot/tinygrad/blob/master/docs/adding_new_accelerators.md as well?

PR #935, opened 2023-06-05
fix inf/nan codegen

cloud11665:
We were appending f to NAN and INFINITY constatns, and now we don't!
Constants taken from https://www.gnu.org/software/libc/manual/html_node/Infinity-and-NaN.html

### DIFF
--- a/tinygrad/codegen/cstyle.py
+++ b/tinygrad/codegen/cstyle.py
@@ -120,8 +120,7 @@ def uops_to_cstyle(uops:List[UOp], bufs:List[Union[LocalBuffer,LazyBuffer]], lan
       # TODO: merge with CONST?
       if bufs[args.i] is not None and isinstance(bufs[args.i].realized, RawConst):
         assert newvar.ltype == LocalTypes.float, "const can't be float4"
-        # nan? inf?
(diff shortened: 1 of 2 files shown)


tinygrad/codegen/cstyle.py
```diff
@@ -120,8 +120,7 @@ def kk(s): kernel.append("  "*depth+s)
       # TODO: merge with CONST?
       if bufs[args.i] is not None and isinstance(bufs[args.i].realized, RawConst):
         assert newvar.ltype == LocalTypes.float, "const can't be float4"
-        # nan? inf?
-        val = f"{bufs[args.i].realized._buf}" + ("f" if not dtypes.is_int(bufs[args.i].dtype) else "")
+        val = str({float("inf"):"INFINITY", float("nan"):"NAN"}.get(bufs[args.i].realized._buf, bufs[args.i].realized._buf)) + ("f" if not dtypes.is_int(bufs[args.i].dtype) and not bufs[args.i].realized._buf in [float("inf"), float("nan")] else "")
```
python273, line 123: can we stop with golfed 250 char one-liners? It's going to be more readable if you use proper `if` and I'm sure it will be more fast without all of these intermediate dicts and lists

python273:
Also would be good to add a test for operations on `inf`s

geohot:
Thanks!

PR #940, opened 2023-06-06
Added a new implementation of Tensor.pow()

willrandship:
…()), so this has an expanded domain relative to standard pow(), and should never output NaN.


willrandship:
Also, since this does equal precision for all operations, it's a bit lower precision on the output than an IEEE compliant implementation.

tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
Tests failing.

willrandship:
Yeah...misunderstanding of how it was expecting to be used. I'll rewrite it in mlops instead of as a Tensor method. (Tensor method will just be the broadcasted mlop as before.) That should fix the scalar input issue.
While I'm at it I'll see if I can get the nan behavior working properly.
I'll submit another PR when that's working.

Issue #927, opened 2023-06-04
Exception running the stable_diffusion example


dvtkrlbs:
I am getting 
```py
Traceback (most recent call last):
  File "C:\DEV\tinygrad\examples\stable_diffusion.py", line 615, in <module>
    load_state_dict(model, torch_load(FILENAME)['state_dict'], strict=False)
  File "c:\dev\tinygrad\tinygrad\state.py", line 107, in torch_load
    _, _, _, rwd, _, ids, base_offset = pkl.load(), pkl.load(), pkl.load(), f.tell(), pkl.load(), pkl.load(), f.tell()
_pickle.UnpicklingError: persistent IDs in protocol 0 must be ASCII strings
```

I am using Python 3.10.11 on WIndows 11 with Cuda Tookit version 12.1.1.

python273:
which weights you use?

dvtkrlbs:
The hash matches same issue also seems to be happening with the whisper example same error.

dvtkrlbs:
here are output of pip freeze
```
appdirs==1.4.4
audioread==3.0.0
certifi==2023.5.7
cffi==1.15.1
charset-normalizer==3.1.0
colorama==0.4.6
decorator==5.1.1
filelock==3.12.0
idna==3.4
Jinja2==3.1.2
joblib==1.2.0
lazy_loader==0.2
librosa==0.10.0.post2
llvmlite==0.40.0
MarkupSafe==2.1.3
mpmath==1.3.0
msgpack==1.0.5
networkx==3.1
numba==0.57.0
numpy==1.24.3
packaging==23.1
Pillow==9.5.0
platformdirs==3.5.1
pooch==1.6.0
pycparser==2.21
pyopencl==2023.1
pytools==2022.1.14
requests==2.31.0
scikit-learn==1.2.2
scipy==1.10.1
soundfile==0.12.1
soxr==0.3.5
sympy==1.12
threadpoolctl==3.1.0
-e git+https://github.com/geohot/tinygrad@091bd65a68677cbed4c7eccaa50d34c5dacc1214#egg=tinygrad
torch==2.0.1
torchaudio==2.0.2
tqdm==4.65.0
typing_extensions==4.6.3
urllib3==2.0.2
```

python273:
can you add `print(t[0:32].numpy())` here:
https://github.com/geohot/tinygrad/blob/079ea217a38bc90ce8e64137cf7d04d664ad9db6/tinygrad/state.py#L94-L95
[...]

PR #944, opened 2023-06-06
[bug fix] limit split to 1 due to windows path containing :

dc-dc-dc:
Disk tensor bug on windows, when parsing the device string the split does not limit to 1. Since window path includes a ":" this results in the disk tensor only getting the drive letter as the path. Limiting the split to 1 passes the full path to file to disk tensor.

Closes (#927)

### DIFF
--- a/examples/stable_diffusion.py
+++ b/examples/stable_diffusion.py
@@ -1,6 +1,7 @@
 # https://arxiv.org/pdf/2112.10752.pdf
 # https://github.com/ekagra-ranjan/huggingface-blog/blob/main/stable_diffusion.md
-
+import os
+import tempfile
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


python273:
also does old torch format has magic header? Maybe we should add an explicit check and raise an error if not found

dc-dc-dc:
hmm not sure, but can add a check if the tensor buffer is still empty raise an error as nothing was loaded

Something like
`
if tuple(t[0:2].numpy()) == (0,0): raise Exception(f'data failed to load at {fn}')
`

dc-dc-dc:
it would open the letter drive as 'C:\whatever' would get passed to disk tensor as 'C' 

It wouldn't crash because disk_tensor opens in append mode, so it would create the file if not there

PR #946, opened 2023-06-06
made seed None by default -> numpy picks a random seed

SnakeOnex:
made seed None by default, stable diffusion now generates a different image each time


geohot:
Hmm, I want the loadop to be deterministic. Just pick a random seed, don't use None.

PR #947, opened 2023-06-06
Onnx: added ArrayFeatureExtractor, GatherElements

geohotstan:
========== 282 failed, 541 passed, 1811 skipped, 2 warnings in 20.25s ==========
========== 265 failed, 558 passed, 1811 skipped, 2 warnings in 19.62s ==========

for ArrayFeatureExtractor, I modified the original Gather implementation
for GatherElements, I tried to keep everything in Tensor but that was a little too hard. I could give it another go before merge if needed since I had an idea 
for Where change, Where returns int64 Tensor which is used as input later on and causes `AssertionError: dtype mismatch on realize got dtypes.float expected dtypes.int64`
for Gather, there are Gather ops in tests that don't have axis. kinda hacky.


geohot:
This just uses numpy. If it's converting the tensor to numpy what's the point?

geohotstan:
You're right. 
I'm trying instead to write a _gather() that mimics torch.gather()'s behavior that uses `arange`, a `compare`, and a `multiply` like you mentioned in issues 858. I'll then express `ArrayFeatureExtractor`, `GatherElements`, and `Gather` using _gather() by reshaping the tensors a little.
I'll close this pr for now, sorry. I'll reopen one when I figured it out.

PR #948, opened 2023-06-06
Updated test_mnist.py

athy125:
Dataset is preprocessed only once using the preprocess_mnist() function and saved in a more efficient format using NumPy's np.savez() function. The preprocessed dataset is then loaded using the load_mnist() function, which retrieves the data from the saved file.

Additionally, the setUpClass() method is used as a class method to perform the dataset loading and preprocessing only once, regardless of the number of test methods present in the test case.

### DIFF
--- a/test/models/test_mnist.py
+++ b/test/models/test_mnist.py
@@ -5,108 +5,103 @@ from tinygrad.tensor import Tensor, Device
 from tinygrad.nn import optim, BatchNorm2d
 from extra.training import train, evaluate
 from datasets import fetch_mnist
+import multiprocessing
 
-# load the mnist dataset
-X_train, Y_train, X_test, Y_test = fetch_mnist()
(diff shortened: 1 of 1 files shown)


python273:
need to indent with 2 spaces

geohot:
Why does this read like ChatGPT?

test/models/test_mnist.py
```diff
@@ -5,108 +5,103 @@
...
-    self.l2 = Tensor.scaled_uniform(128, 10)
-
-  def parameters(self):
-    return optim.get_parameters(self)
-
-  def forward(self, x):
-    return x.dot(self.l1).relu().dot(self.l2).log_softmax()
-
-# create a model with a conv layer
-class TinyConvNet:
```
jla524, line 25: Tests will fail (e.g. test_conv) with this model removed

PR #950, opened 2023-06-06
Whisper: there is a better way to do that!

cloud11665:
In response to a quote from latest stream (https://youtu.be/Mr0rWJhv9jU?t=13758)


geohot:
hehe nice

PR #728, opened 2023-03-25
Use the Tensor Cores on M1+ ($1000 bounty)

geohot:
Claim the bounty by implementing this and having tinygrad generate GEMM kernels for M1 that are faster than torch/Accelerate.framework.

Clean code only, must be merged to claim bounty.

PR #955, opened 2023-06-08
Fixed imports of coremltools for version 6.3 and 7.0.

mxmln11:
Hi George,
here's a tiny fix ;)
I am just starting with the examples and test...
Best, Max

### DIFF
--- a/accel/ane/1_build/coreml_ane.py
+++ b/accel/ane/1_build/coreml_ane.py
@@ -1,7 +1,13 @@
 #!/usr/bin/env python3
 import numpy as np
 import coremltools as ct
-from coremltools.models.neural_network import datatypes, NeuralNetworkBuilder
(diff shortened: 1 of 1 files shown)


jla524:
Hi Max, thanks for sending in the fix. Unfortunately, ANE is no longer supported. However, we have bounties for supporting tensor cores (see #728).

geohot:
Yea, the issue with the ANE is that it is now super locked down, so we've discontinued support until someone finds a clean way to use it.

philipturner:
> Yea, the issue with the ANE is that it is now super locked down, so we've discontinued support until someone finds a clean way to use it.

Bruh just use `MPSGraph.optimizationLevel1`

https://github.com/dougallj/applegpu/issues/44

Screenshot 2023-06-13 at 10 48 20 PM

PR #958, opened 2023-06-09
Test nllloss

stevenandersonz:
Before
=== 227 failed, 597 passed, 1810 skipped, 2 warnings in 34.87s ===
After
=== 209 failed, 615 passed, 1810 skipped, 2 warnings in 27.66s ===

The implementation is slower than torch.NLLLoss but I had to do the indexing using a mask. Also, I only use NumPy to gather the weights but if that's a problem I could probably find another way.


geohot:
If it turns things to numpy, it's not valid tinygrad.

PR #959, opened 2023-06-09
pow implementation - now with working scalars and reverse

willrandship:
It's a bit ugly and big, but the tests should be much better than the last PR. the 3 conditionals are for:

scalar ^ tensor (as tensor.pow(scalar, reverse=True) )
reversed tensor^tensor (should be rare, probably should have just done return x.pow(self) in retrospect)
regular tensor^whatever (using only the tensor-first functions, so it doesn't matter if the other side is a scalar)

### DIFF
--- a/tinygrad/mlops.py
+++ b/tinygrad/mlops.py
@@ -50,22 +50,6 @@ class Exp(Function):
   def backward(self, grad_output:LazyBuffer) -> LazyBuffer:
     return self.ret.binary_op(BinaryOps.MUL, grad_output)
 
-class Log2(Function):
-  def forward(self, x:LazyBuffer) -> LazyBuffer:
-    self.x = x
-    return x.unary_op(UnaryOps.LOG2)
-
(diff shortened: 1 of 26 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         7               1
```

tinygrad/mlops.py
```diff
@@ -50,6 +50,22 @@ def forward(self, x:LazyBuffer) -> LazyBuffer:
   def backward(self, grad_output:LazyBuffer) -> LazyBuffer:
     return self.ret.binary_op(BinaryOps.MUL, grad_output)
 
+class Log2(Function):
```
geohot, line 53: Definitely not

willrandship:
How about this? just pow + the iterated sqrt, rsqrt.

There are also 7 removable lines in mlops.py with these changes (Pow class), but I've been leaving those.

geohot:
Going to close, this is unreadable and not explained or tested.

PR #960, opened 2023-06-09
clean pad argument

ilyesarf:
This pull request introduces improvements to the codebase, specifically targeting the `pad` and `pad2d` functions. Prior to these changes, the pad function strictly accepted arguments in the form of ((1,1), (1,1)). However, with the updates, the functions have become more flexible. Now, it can handle arguments like (1,1,1,1) or (1,1, (2,3)). 

This was inspired from pytorch's `pad` function where they accept different types of padding, <a href="https://github.com/ilyesarf/tinygrad/blob/56e06fe746638b9a2aa8034c19dc26995cb1e39b/tinygrad/tensor.py#L212"> `flatten` </a> them, <a href="https://github.com/ilyesarf/tinygrad/blob/56e06fe746638b9a2aa8034c19dc26995cb1e39b/tinygrad/tensor.py#L213"> turn them into `Tuple[Tuple[int, int], ...]` format  </a> and pass them to the function.

Also, there's a bool argument to turn `clean_padding` off in the <a href="https://github.com/ilyesarf/tinygrad/blob/56e06fe746638b9a2aa8034c19dc26995cb1e39b/tinygrad/tensor.py#L207"> `pad` </a> function.

Finally, some more test cases are added to <a href="https://github.com/ilyesarf/tinygrad/blob/7040eb201d3c5097a5455980044f5d3ca7c06572/test/test_ops.py#L304"> `test_pad2d` </a> and all of them are passing.


test/test_ops.py
```diff
@@ -301,7 +301,9 @@ def test_slice_one_multi(self):
     helper_test_op([(10,10)], lambda x: x[1], lambda x: x[1])
 
   def test_pad2d(self):
-    helper_test_op([(3,3,3,3)], lambda x: torch.nn.functional.pad(x, (1,2,3,4)), lambda x: x.pad2d(padding=(1,2,3,4)))
+    paddings = ((1,1,2,2), (1,2,(3,4)), (0,(1,2,1),[3,3]))
```
geohot, line 304: How is `(0,(1,2,1),[3,3])` valid

geohot:
I agree pad needs to be cleaned up, but this needs to be thought through a lot better.

`helper_test_op([(1,1,2,2,3,3)], lambda x: torch.nn.functional.pad(x, Tensor([0]).flatten_pad(padding)), lambda x: x.pad2d(padding=padding))`

is not a test! It's calling the same function on both sides of the test.

PR #962, opened 2023-06-09
Cuda on cpu

cloud11665:
tested singular ptx kernels. Setting up the env for this is a pain


cloud11665:
image
I have compiled a fully static library for executing ptx on the cpu. No need to mess with building gpuocelot

geohot:
I mean...you can't commit a binary blob. Can it be built in CI with the tests?

cloud11665:
There is https://github.com/actions/cache which could be useful, as building gpuocelot requires building llvm, and we wouldn't want to do that every commit.

geohot:
Yea, caching is fine, but CI has to be able to build it. LLVM isn't too slow to build if you only build one backend.

wozeparrot:
> There is [actions/cache](https://github.com/actions/cache) which could be useful, as building gpuocelot requires building llvm, and we wouldn't want to do that every commit.

I've tried building gpuocelot before and it didn't require the entirety of llvm to be built unless you are doing a debug build.

cloud11665:
kk, a simple python wrapper is done:
```py
kernel = r"""
.version 7.5
.target sm_35
.address_size 64

        // .globl       _Z4E_16Pf

.visible .entry _Z4E_16Pf(
        .param .u64 _Z4E_16Pf_param_0
)
{
        .reg .b32       %r<6>;
        .reg .b64       %rd<5>;


        ld.param.u64    %rd1, [_Z4E_16Pf_param_0];
        cvta.to.global.u64      %rd2, %rd1;
        mov.u32         %r1, %ntid.x;
        mov.u32         %r2, %ctaid.x;
        mov.u32         %r3, %tid.x;
        mad.lo.s32      %r4, %r1, %r2, %r3;
        mul.wide.s32    %rd3, %r4, 4;
        add.s64         %rd4, %rd2, %rd3;
        mov.u32         %r5, 1065353216;
        st.global.u32   [%rd4], %r5;
        ret;

}
"""

import numpy as np

buff = np.zeros(16, dtype=np.float32)
buffp = ffi.cast("void*", buff.ctypes.data)

lib.run_ptx(kernel.encode(), 1, [buffp], 16, 1, 1, 1, 1, 1)

print(buff) # [1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
```

cloud11665:
I'm experimenting in docker, and the longest step is downloading the cuda toolkit.
Also, I will not be forking gpuocelot, and just maintain a patch in the tinygrad repo

geohot:
Patch is fine.

cloud11665:
> Patch is fine. How big is CUDA? If it's huge (>200MB), maybe we figure out how to cache it.

All apt packages (cuda included) are ~8000MiB. It only takes a minute to fetch and install as Microsoft has them all cached on azure

cloud11665:
but we do all of that only for the libcudacpu.so file, so caching that would save 3-4 mins

cloud11665:
Tested locally and it worked, but there are 2 more things I want to add:
- a non_stable mode for newer nvcc version (compiling for sm_50 and hoping it works :laughing: )
- a docker for a fully static compilation of libcudacpu.so (basically only adding static libboost since that's the nony "nonstandard" lib)
```
$ ldd extra/cudacpu/libtinygrad.so 
extra/cudacpu/libtinygrad.so: /lib64/libtinfo.so.6: no version information available (required by extra/cudacpu/libtinygrad.so)
        linux-vdso.so.1 (0x00007ffec333f000)
        libGLEW.so.2.2 => /lib64/libGLEW.so.2.2 (0x00007fe90f851000)
        libz.so.1 => /lib64/libz.so.1 (0x00007fe90f837000)
        libboost_thread.so.1.74.0 => /lib64/libboost_thread.so.1.74.0 (0x00007fe90f818000)
        libGL.so.1 => /lib64/libGL.so.1 (0x00007fe90cf79000)
        libtinfo.so.6 => /lib64/libtinfo.so.6 (0x00007fe90f7e5000)
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fe90cc00000)
        libm.so.6 => /lib64/libm.so.6 (0x00007fe90ce99000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fe90ce79000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fe90ca23000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fe90f92f000)
        libX11.so.6 => /lib64/libX11.so.6 (0x00007fe90c8dc000)
        libGLX.so.0 => /lib64/libGLX.so.0 (0x00007fe90ce47000)
        libXext.so.6 => /lib64/libXext.so.6 (0x00007fe90c8c7000)
        libGLdispatch.so.0 => /lib64/libGLdispatch.so.0 (0x00007fe90c80e000)
        libxcb.so.1 => /lib64/libxcb.so.1 (0x00007fe90c7e3000)
        libXau.so.6 => /lib64/libXau.so.6 (0x00007fe90f7db000)
```

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_cuda.py              32              17
lines added in the tinygrad folder: 15
```

cloud11665:
only tests failing in test_ops are:
test/test_ops.py::TestOps::test_slice_negative_strides
test/test_ops.py::TestOps::test_strided_conv_transpose2d

cloud11665:
tests:
FAILED test/test_ops.py::TestOps::test_broadcast_full
FAILED test/test_ops.py::TestOps::test_output_padded_conv_transpose2d
FAILED test/test_ops.py::TestOps::test_pow
FAILED test/test_ops.py::TestOps::test_slice_negative_strides
FAILED test/test_ops.py::TestOps::test_sqrt
FAILED test/test_ops.py::TestOps::test_slice_stride_gt_one
FAILED test/test_ops.py::TestOps::test_sin
FAILED test/test_ops.py::TestOps::test_std
FAILED test/test_ops.py::TestOps::test_strided_conv1d_simple
FAILED test/test_ops.py::TestOps::test_strided_conv2d_simple
FAILED test/test_ops.py::TestOps::test_strided_conv_transpose2d
FAILED test/test_ops.py::TestOps::test_cos
FAILED test/test_ops.py::TestOps::test_pow_const
FAILED test/test_ops.py::TestOps::test_broadcast_partial
FAILED test/test_ops.py::TestOps::test_std_keepdim
FAILED test/test_ops.py::TestOps::test_tan
FAILED test/test_ops.py::TestOps::test_std_axis
FAILED test/test_ops.py::TestOps::test_cumsum
are aborting when transpiled to llvm with the following message:
`LLVM Verifier failed for kernel: _Z6E_5525PfPKf : "PHINode should have one entry for each predecessor of its parent basic block!`

cloud11665:
I'm currently fuzzing strided access, as that's the only bug left

geohot:
So it's currently taking 18 minutes, really needs to be ~5. How is this so much slower than OpenCL Test?

cloud11665:
my best guess is that it's due to emulation and opencl tests are jit compiled

cloud11665:
yea, this is way too slow...

cloud11665:
I think I'll have to die on that hill (gpuocelot)... Other alternatives do not exist (not using self-hosted runners*)
- [CU2CL](https://github.com/vtsynergy/CU2CL) - couldn't get it to compiler
- [ZLUDA](https://github.com/vosen/ZLUDA) - only a libcuda replacement only, we also rely on libcudart
- AMD hipify - cuda -> hip and then hip -> spirv -> mlir-spirv-cpu-runner, couldn't get hipify to work after finally building it after like an hour of errors
- [COX](https://arxiv.org/abs/2112.10034) - only a theoretical paper by people from the same university that made gpuocelot. Interestingly, it mentions [DPC++ compat](https://www.intel.cn/content/www/cn/zh/developer/tools/oneapi/dpc-compatibility-tool.html) - a tool from intel for auto migrating cuda code to SYCL (funnily as with everything intel, finding it via google was impossible)

So. I think there are only 2 sensible options. Either we stick with ocelot emulation (as the ptx->llvmir backend is way too unstable), optimize it a bit, cache the compilations, reduce test sizes, and run forward pass only - sub 15 is doable. O we scrap everything and dive into the intel tool.

cloud11665:
Interesting that it fails...

python273:
seems like some workflows got stuck?

cloud11665:
> seems like some workflows got stuck?
> 
> https://github.com/geohot/tinygrad/actions?query=is%3Ain_progress

yea, I'll have to investigate, thought the earliest one will die in approx. 15 min (6h timeout)

cloud11665:
funnily enough, you can't run tests with a merge conflict. weird

cloud11665:
installing pycuda and compilation now take up most of the time :)

geohot:
Let's aim for as many tests as we can get in under 10 minutes. Why is installing deps 3 minutes here and only 1 minute on others?

Core tinygrad (like `tinygrad/runtime/ops_cuda.py`) can't import from extra. And instead of the test skips I'd just skip the whole file. Things like `test_mnist` and `test_onnx` don't need to run.

cloud11665:
Installing deps takes so long because of pycuda which is not installed in any other test.
As for moving that into core tinygrad, should it live in ops_cuda or some separate file ?

extra/cudacpu/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+void ptx_kernel_destroy(void* kernel);
+void ptx_call(void* kernel, int n_args, void* args[],
+             int blck_x, int blck_y, int blck_z,
+             int grid_x, int grid_y, int grid_z);
+""")
+lib = ffi.dlopen("./extra/cudacpu/libcudacpu.so")
+
+DEBUGCUDACPU = getenv("DEBUGCUDACPU", 0)
+
+class PTXKernel:
```
geohot, line 18: Wrong spacing.

extra/cudacpu/__init__.py
```diff
@@ -0,0 +1,82 @@
+from typing import List, Tuple, Any
+from enum import Enum
+from tinygrad.helpers import getenv
+import cffi
+
+ffi = cffi.FFI()
```
geohot, line 6: Can you use types instead of cffi?

extra/cudacpu/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __call__(self, args:Tuple[Any,...], block:Tuple[int, ...], grid:Tuple[int, ...]):
+        # print(self.kernel, args)
+        lib.ptx_call(self.kernel, len(args), [ffi.cast("void*", ffi.from_buffer(x._buffer())) for x in args], *block, *grid)
+        # print(self.kernel, args)
+
+    def __del__(self):
+        lib.ptx_kernel_destroy(self.kernel)
+
+# def create_kernel(source):
```
geohot, line 30: All this stuff needs to be cleaned up, maybe put in a test.

extra/cudacpu/__init__.py
cloud11665: done.

extra/cudacpu/__init__.py
cloud11665: done.

cloud11665:
!!!
image
sub 8 minutes !

geohot:
I'll set a clear limit. This functionality is worth +50 lines total to m, with max 10 in core tinygrad. If it's more than that, I don't think it's worth it.

cloud11665:
Does the CI config file count, as that alone is about 50 ?

cloud11665:
cudacpu.cpp (15) + patch (23) + tests (3) = 41
ops_cuda.py (+28-14)

test/test_ops.py
```diff
@@ -451,6 +451,7 @@ def test_slice_stride_gt_one(self):
     helper_test_op([(7,5,10)], lambda x: x[1:5:2, 3, ::4], lambda x: x[1:5:2, 3, ::4])
     helper_test_op([(7,5,10)], lambda x: x[1:5:2, None, None, 3, None, ::4], lambda x: x[1:5:2, None, None, 3, None, ::4])
 
+  @unittest.skipIf(getenv("CUDACPU",0) == 1, "Emulator fails at the last assert")
```
geohot, line 454: Why?

test/test_ops.py
```diff
@@ -654,6 +655,7 @@ def test_dilated_conv_transpose2d(self):
         lambda x,w: torch.nn.functional.conv_transpose2d(x,w,dilation=dilation).relu(),
         lambda x,w: Tensor.conv_transpose2d(x,w,dilation=dilation).relu(), atol=1e-4, grad_rtol=1e-5)
 
+  @unittest.skipIf(getenv("CUDACPU",0) == 1, "Emulator fails on the first stride")
```
geohot, line 658: Why?

extra/cudacpu/gpuocelot.patch
```diff
@@ -0,0 +1,23 @@
+diff --git a/ocelot/CMakeLists.txt b/ocelot/CMakeLists.txt
+index fef7ee9b..534a4d7c 100644
+--- a/ocelot/CMakeLists.txt
++++ b/ocelot/CMakeLists.txt
+@@ -21,6 +21,8 @@ find_package(FLEX 2.5 REQUIRED)
+ find_package(BISON 2.5 REQUIRED)
+ find_package(GLEW REQUIRED)
+ find_package(ZLIB REQUIRED)
++set(TINFO_INCLUDE_DIRS "/usr/include")
```
geohot, line 9: Is there a system way to fix this?

extra/cudacpu/cudacpu.cpp
```diff
@@ -0,0 +1,15 @@
+#include <ocelot/ir/Module.h>
```
geohot, line 1: Do you really need this file?

.github/workflows/test.yml
```diff
@@ -187,3 +187,67 @@ jobs:
...
+        git apply gpuocelot.patch
+    - if: ${{ steps.cache-build.outputs.cache-hit != 'true' }}
+      name: Compile gpuocelot and libcudacpu
+      run: |
+        cd ${{ github.workspace }}/extra/cudacpu/gpuocelot/ocelot
+        mkdir build
+        cd build
+        cmake .. -Wno-dev -G Ninja
+        ninja
+    - name: install libcudacpu.so
```
geohot, line 239: Just LD_PRELOAD it

geohot:
Still 132. Look, there's nicer ways to do a lot of this stuff. I don't want to keep going back and forth.

Ask, is this beautiful? If it's not, make it more beautiful

cloud11665:
I'll try again tomorrow, it's getting late here (2am). I'm not giving up on this.

tinygrad/runtime/ops_cuda.py
```diff
@@ -1,19 +1,32 @@
...
+    def __del__(self): lib.ptx_kernel_destroy(self.kernel)
+else:
+  import pycuda.autoprimaryctx # type: ignore # pylint: disable=unused-import # noqa: F401
+  import pycuda.driver as cuda # type: ignore
+  class RawCUDABuffer(RawBufferCopyInOut):
+    def __init__(self, size, dtype): super().__init__(size, dtype, cuda.mem_alloc(size * dtype.itemsize))
+    def _copyin(self, x:np.ndarray, stream:Optional[cuda.Stream]=None): cuda.memcpy_htod_async(self._buf, x, stream)
+    def _copyout(self, x:np.ndarray): cuda.memcpy_dtoh(x, self._buf)
 
 class CUDAProgram:
```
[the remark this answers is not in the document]
cloud11665: due to calls to cuda.Event() and cuda.Context.synchronize, I don't think there is a clean way to do this.

cloud11665:
We're not patching ocelot anymore, using a simpler interface, compiling our "frontend" g++ and linking it with libgpuocelot directly. The interface used is the same as with the tests inside ocelot, so it's the intended way. I have experimented with preloading, but pycuda is looking directly for libcuda.so and libcudart.so. As for the cleanness, I think not patching anything and a single cpp file for an interface passes the bar.

cloud11665:
oh, sorry. tests were passing locally due to having other version of ocelot installed on my system. This kernel crashes the emulator
```
Failed to parse file '':
32      cvta.to.global.u64      %rd6, %rd2;
33      mov.u32         %r1, %ntid.y;
34      mov.u32         %r2, %ctaid.y;
35      mov.u32         %r3, %tid.y;
36      mad.lo.s32      %r4, %r1, %r2, %r3;
37      mov.u32         %r5, %ntid.x;
38      mov.u32         %r6, %ctaid.x;
39      mov.u32         %r7, %tid.x;
40      mad.lo.s32      %r8, %r5, %r6, %r7;
41      mad.lo.s32      %r9, %r4, 5, %r8;
42      mul.wide.s32    %rd7, %r9, 4;
43      add.s64         %rd8, %rd6, %rd7;
44      ld.global.f32   %f1, [%rd8];
45      mul.wide.s32    %rd9, %r8, 4;
46      add.s64         %rd10, %rd5, %rd9;
47      ld.global.f32   %f2, [%rd10];
48      max.f32         %f3, %f1, %f2;
49      setp.eq.f32     %p1, %f3, %f1;
50      selp.f32        %f4, 0f3F800000, 0f00000000, %p1;
51      add.s64         %rd11, %rd4, %rd7;
52 
```
 (41, 5): Parsed invalid instruction mad.lo.s32 %r9, %r4, 5, %r8 : type of operand A b32 does not equal type of operand B s32

I'll upstream a change to ocelot

dmikushin:
> pycuda is looking directly for libcuda.so and libcudart.so

@cloud11665 Every modern CUDA app also has this problem, because libcudart_static internally dlopens libcuda and dlsym. ~~If you came from OpenGL, you can't stop repeating OpenGL everywhere.~~

Therefore, preloading won't help, you are right.

What you (and GPUOcelot) need is this: https://github.com/dmikushin/easyprof/blob/master/src/loader.cu#L13 and below.

I'm half-way integrating the same into GPUOcelot, but please feel free to get it landed it faster.

cloud11665:
I think you linked to a private repository, as I cannot view it.

dmikushin:
> I think you linked to a private repository, as I cannot view it.

Sorry, made it public now.

cloud11665:
not counting the .github, there's only like 30 lines changed. Is this clean and tiny enough ?

test/test_ops.py
cloud11665: it doesn't fail now. I don't know why.

test/test_ops.py
cloud11665: it doesn't fail now. I don't know why.

test/test_ops.py
geohot: Remove skip?

tinygrad/runtime/ops_cuda.py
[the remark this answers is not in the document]
geohot: There is

extra/cudacpu.cpp
```diff
@@ -0,0 +1,17 @@
+#include <ocelot/ir/Module.h>
+#include <ocelot/executive/EmulatedKernel.h>
+extern "C" void ptx_run(const char* source, int n_args, void* args[], int blck_x, int blck_y, int blck_z, int grid_x, int grid_y, int grid_z) {
+	std::stringstream ss; ss << source;
```
geohot, line 4: Bad spacing.

geohot:
The size is mostly okay, but there's still bad code quality in a few places.

Example: What if I want to use this outside CI?

.github/workflows/test.yml
```diff
@@ -187,3 +187,56 @@ jobs:
...
+    #     cache-name: cache-gpuocelot-build
+    #   with:
+    #     path: ${{ github.workspace }}/extra/cudacpu/gpuocelot/ocelot/build
+    #     key: ubuntu22.04-gpuocelot-${{ hashFiles('${{ github.workspace }}/extra/cudacpu/cudacpu.cpp', '${{ github.workspace }}/extra/cudacpu/gpuocelot.patch') }}
+    #     restore-keys: |
+    #       ubuntu22.04-gpuocelot-
+    - name: Clone gpuocelot repo
+      uses: actions/checkout@v3
+      with:
+        repository: gpuocelot/gpuocelot
```
geohot, line 218: Pin a commit?

.github/workflows/test.yml
```diff
@@ -187,3 +187,56 @@ jobs:
...
+        path: ${{ github.workspace }}/gpuocelot
+        submodules: true
+    - name: Compile gpuocelot and libcudacpu
+      run: |
+        cd ${{ github.workspace }}/gpuocelot/ocelot
+        mkdir build
+        cd build
+        cmake .. -Wno-dev -G Ninja
+        ninja
+        sudo cp ./libgpuocelot.so /usr/lib
```
geohot, line 228: Does gpuocelot have a make install?

tinygrad/runtime/ops_cuda.py
```diff
@@ -49,15 +57,13 @@ class CUDACodegen(CStyleCodegen):
     kernel_prefix = "__global__", smem_prefix = "__shared__ ", barrier = "__syncthreads();", float4 = "make_float4",
     gid = [f'blockDim.{chr(120+i)}*blockIdx.{chr(120+i)}+threadIdx.{chr(120+i)}' for i in range(3)],
     lid = [f'threadIdx.{chr(120+i)}' for i in range(3)],
+    header = "typedef unsigned char uchar; typedef long long int64;",
```
geohot, line 60: I'd just fix this in dtypes.

cloud11665:
Re: Example: What if I want to use this outside CI?

I think I'll just upstream this function so that it's in libocelot.so and we don't have any cpp files in tinygrad. And if somebody wants to get it working, they'll just have to normally install ocelot.

extra/cudacpu.cpp
cloud11665: fully removed :)

.github/workflows/test.yml
cloud11665: now it has !

tinygrad/runtime/ops_cuda.py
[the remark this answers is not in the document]
cloud11665: I don't know how to do it cleanly though.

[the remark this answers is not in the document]
cloud11665: if you really want, I can stub pycuda in C code directly in ocelot, but I don't think that's clean.

.github/workflows/test.yml
cloud11665: done.

tinygrad/runtime/ops_cuda.py
cloud11665: done.

cloud11665:
oh, the commit order is messed up because the clock on my laptop desynced as it discharged to 0 overnight :(

tinygrad/runtime/ops_cuda.py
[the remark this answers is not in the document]
geohot: Make a class called "cuda" in "if EMULATING"

cloud11665:
I've also added a config in ocelot to not build tests, so we're building only 2/3 the objects. We will still cache it though.

tinygrad/runtime/ops_cuda.py
```diff
@@ -1,18 +1,41 @@
...
+    module_from_buffer = lambda src: cuda.module(src)
+    class Event:
+      def __init__(self): pass
+      def record(self): pass
+      def time_till(self, other): return 0.0
+      def synchronize(self): pass
+    class Context:
+      synchronize = lambda:0
+    CompileError = Exception
+  def _cuda_compile(prg, **kwargs): return cuda_compile(prg, **{**kwargs, 'arch': 'sm_35'})
```
geohot, line 29: There should be a better way to do this. Can you get the arch from cuda?

tinygrad/runtime/ops_cuda.py
```diff
@@ -1,18 +1,41 @@
...
+  class cuda:
+    class module:
+      def __init__(self, src): self.src = src
+      def get_function(self, _): return self
+      def __call__(self, *args, block, grid): lib.ptx_run(self.src, len(args), (ctypes.c_void_p * len(args))(*[ctypes.cast(x, ctypes.c_void_p) for x in args]), *block, *grid)
+    module_from_buffer = lambda src: cuda.module(src)
+    class Event:
+      def __init__(self): pass
+      def record(self): pass
+      def time_till(self, other): return 0.0
```
geohot, line 24: Make this actually work!

tinygrad/runtime/ops_cuda.py
```diff
@@ -1,18 +1,42 @@
...
+  lib.ptx_run.argtypes = [ctypes.c_char_p, ctypes.c_int, ctypes.POINTER(ctypes.c_void_p), ctypes.c_int, ctypes.c_int, ctypes.c_int, ctypes.c_int, ctypes.c_int, ctypes.c_int]
+  class cuda:
+    class module:
+      def __init__(self, src): self.src = src
+      def get_function(self, _): return self
+      def __call__(self, *args, block, grid): lib.ptx_run(self.src, len(args), (ctypes.c_void_p * len(args))(*[ctypes.cast(x, ctypes.c_void_p) for x in args]), *block, *grid)
+    module_from_buffer = lambda src: cuda.module(src) # pylint: disable=unnecessary-lambda # noqa: E731
+    class Event:
+      def __init__(self): pass
+      def record(self): self.start = time.time()
```
geohot, line 23: Don't use `time.time` for timing. Read tinygrad and see how it's done elsewhere, and also understand why and what the issue is

tinygrad/codegen/assembly_ptx.py
```diff
@@ -4,7 +4,7 @@
 from tinygrad.codegen.linearizer import UOps
 from tinygrad.helpers import dtypes
 
-dtype_to_nvtype = {dtypes.float32: "f32", dtypes.float16: "u16", dtypes.int64: "s64", dtypes.int32: "s32", dtypes.bool: "pred", dtypes.uint64: "u64", dtypes.uint32: "u32"}
+dtype_to_nvtype = {dtypes.float32: "f32", dtypes.float16: "u16", dtypes.int64: "s64", dtypes.int32: "s32", dtypes.bool: "pred", getattr(dtypes, "unsigned long"): "u64", getattr(dtypes, "unsigned int"): "u32"}
```
geohot, line 7: Did you test this?

cloud11665:
ehh, it's this again. I'll just do it in kernel_prefix.

cloud11665:
It was previously in the half_prekernel, but since that was made conditional to only be added if code had fp16 dtypes, then I've moved it into kernel_prefix.

cloud11665:
the desired behavior of
image
Is already implemented because of this piece of code in pycuda.compiler:
```py
    if arch is None:
        from pycuda.driver import Error

        try:
            from pycuda.driver import Context

            arch = "sm_%d%d" % Context.get_device().compute_capability()
        except Error:
            pass
```
No need for mock the python wrapper of [CUContext](https://github.com/inducer/pycuda/blob/6b0fe7ab43f80932e12e9375f239f53ca7ae27b3/src/cpp/cuda.hpp#L451).
arch is `None` by default, and it's overridden only in emulation mode.
image
which matches this table for generation and arch
image

cloud11665:
I have mocked the driver context.
```patch
commit 3304bc777e4dcbe6011842861e958bb21d587cf2
Author: cloud11665 <cloud11665@gmail.com>
Date:   Tue Jun 20 16:43:16 2023 +0200

    mock pycuda.driver.Context

diff --git a/tinygrad/runtime/ops_cuda.py b/tinygrad/runtime/ops_cuda.py
index d985818..807787b 100644
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -26,9 +26,12 @@ if getenv("CUDACPU", 0) == 1:
     class Context:
       synchronize = lambda:0 # noqa: E731
     CompileError = Exception
-  __cuda_compile = cuda_compile
-  def _cuda_compile(prg, **kwargs): return __cuda_compile(prg, **{**kwargs, 'arch': 'sm_35', 'options': kwargs.get('options', []) + ['-Wno-deprecated-gpu-targets']}) # noqa: E731
-  cuda_compile = _cuda_compile # noqa: F811
+  class context:
+    class device:
+      compute_capability = lambda: (3,5) # pylint: disable=unnecessary-lambda # noqa: E731
+    get_device = lambda: context.device # pylint: disable=unnecessary-lambda # noqa: E731
+  import pycuda.driver # type: ignore
+  pycuda.driver.Context = context
   RawCUDABuffer = RawMallocBuffer
 else:
   import pycuda.autoprimaryctx # type: ignore # pylint: disable=unused-import # noqa: F401
@@ -46,7 +49,7 @@ class CUDAProgram:
           f.write(cuda_compile(prg, target="cubin", no_extern_c=True))
         sass = subprocess.check_output(['nvdisasm', '/tmp/cubin']).decode('utf-8')
         print(sass)
-      if not binary: prg = cuda_compile(prg, target="ptx", no_extern_c=True).decode('utf-8')
+      if not binary: prg = cuda_compile(prg, target="ptx", no_extern_c=True, options=['-Wno-deprecated-gpu-targets']).decode('utf-8')
     except cuda.CompileError as e:
       if DEBUG >= 3: print("FAILED TO BUILD", prg)
       raise e

```

PR #963, opened 2023-06-10
multidevice training

wozeparrot:
Opening for visibility.

This currently works for the CPU, LLVM, and CLANG backends.

But as we are mainly targeting multigpu stuff this needs more thought. The main problem right now is that pyopencl/opencl in general really does not like multiprocessing. I could work around that by creating opencl contexts in each process but that *may* break future p2p work. This current iteration hangs on the enqueue_copy for GPU.

I'll try an iteration with threading instead of multiprocessing but I'm not sure how bad the GIL overhead will be.


geohot:
Can you launch the new processes before you init the CL contexts?

wozeparrot:
> Can you launch the new processes before you init the CL contexts?

We might be able to, not sure how this might affect p2p if the devices are in different contexts, or even if p2p would go through opencl at all, it would still be possible to share tensors by sending them to host shared memory first. Tho trying to launch the processes before won't make for the best user experience tho since importing Tensor sets up the CL contexts. We could delay context creation to be later tho hmm.

I do want to try and get this to work with multiprocessing as much as possible because it is way more flexible and easier to scale to multiple boxes later on.

geohot:
Hmm, so if they don't share a CL context there's no good way to share them? CUDA has good stuff for this, maybe OpenCL doesn't.

wozeparrot:
From what I've read, OpenCL doesn't even provide guaranteed support for doing GPU-to-GPU transfers without a host copy. It seems like you can hint it to probably get it to skip tthe host copy, but its not guaranteed. AMD does have a specific [p2p transfer OpenCL extension](https://github.com/jlgreathouse/test_cl_amd_copy_buffer_p2p) that does work across contexts (in fact it has to work across contexts) but I'm not sure what the current state of it is considering this was one of the only pieces of information I could find.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                           3               1
tinygrad/runtime/lib.py                    9               0
tinygrad/runtime/ops_gpu.py               18              15
total                                     30              16
```

wozeparrot:
Probably still going to have to use multithreading as opposed to multiprocessing, as I can't share the buffers/contexts/queues across different processes as their addresses are process/opencl context specific, I could do really cursed stuff and reach into the other processes memory to make it work but that is way too cursed. There are some open APIs that have some level of process compatibility like vulkan which allows some of its objects to be shared across processes but opencl doesn't.

I'm not sure if HIP allows you to access the same GPU memory across processes, but I know that CUDA can.

tinygrad/runtime/ops_gpu.py
```diff
@@ -18,65 +18,72 @@
...
+    self.events_in_flight = []
+  def post_init(self, device=None):
     platforms: List[List[cl.Device]] = [y for y in ([x.get_devices(device_type=cl.device_type.GPU) for x in cl.get_platforms()] + [x.get_devices(device_type=cl.device_type.CPU) for x in cl.get_platforms()]) if len(y)]
-    devices: List[cl.Device] = [x for x in platforms[getenv('CL_PLATFORM', 0)] if x.name not in getenv('CL_EXCLUDE', '').split(',')]
-    if DEBUG >= 1: print(f"using devices: {[d.hashable_model_and_version_identifier for d in devices]}")
-    self.cl_ctx: cl.Context = cl.Context(devices=devices)
-    self.cl_queue: List[cl.CommandQueue] = [cl.CommandQueue(self.cl_ctx, device=device, properties=cl.command_queue_properties.PROFILING_ENABLE) for device in self.cl_ctx.devices]
-
+    self.cl_platform = cl.get_platforms()[getenv('CL_PLATFORM', 0)]
+    self.cl_ctxs: List[cl.Context] = [cl.Context(devices=[x]) for x in platforms[getenv('CL_PLATFORM', 0)] if x.name not in getenv('CL_EXCLUDE', "").split(",")] if device is None else [cl.Context(devices=[platforms[getenv('CL_PLATFORM', 0)][device]])]
```
geohot, line 25: What's the advantage of a different context for each device?

geohot:
To set a target for this bounty.

I want to run `hlb_cifar10.py` across 6 7900XTX and get the ~same result + at least 5x the performance of one.

tinygrad/runtime/ops_gpu.py
wozeparrot: P2P on opencl requires that the devices be in different contexts, but I found that getting multiple contexts on the same device from multiple processes causes the queues to hang.

Antonio12389 (approved):
Si por favor gracias

PR #968, opened 2023-06-10
cuda: add uchar, int64 typedefs

cloud11665:
`test/test_dtype.py` -> tests failed went down to 3 from 20.
only ones left are f16, but I haven't looked into them yet.

### DIFF
--- a/tinygrad/codegen/cstyle.py
+++ b/tinygrad/codegen/cstyle.py
@@ -24,6 +24,7 @@ class CStyleLanguage(NamedTuple):
   float4: Optional[str] = None
   half_prekernel: Optional[str] = None
   uses_vload: bool = False
+  header: Optional[str] = None
 
 def to_image_idx(base_shape:Tuple[int, ...], idxy:Node, valid:Node, validhacks=False) -> Tuple[Node, Node]:
   idy = (idxy//(4*base_shape[1]))
@@ -171,6 +172,8 @@ def uops_to_cstyle(uops:List[UOp], bufs:List[Union[LocalBuffer,LazyBuffer]], lan
     [', '.join([f'{t} {bufnames[i]}' for i,t in buftypes] + lang.extra_args)] +
     [") {\n"] + list(prekernel) + ['\n'.join(kernel), "\n}"])
 
+  if lang.header:
+    prg =''.join([f"{lang.header}", "\n", prg])
   if lang.half_prekernel:
     prg =''.join([f"{lang.half_prekernel}", "\n", prg])
   return prg, global_size, local_size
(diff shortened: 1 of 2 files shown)


cloud11665:
simply casting true and false branch of a ternary statement got rid of the issue.

cloud11665:
Oops, was also casting to float4, and given it's implementation quirks it broke the opencl tests.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/cstyle.py                 9               2
tinygrad/runtime/ops_cuda.py               5               1
total                                     14               3
```

cloud11665:
I'm getting 1100mspt on a 3090 with `JIT=1 OPT=4 OPTLOCAL=2` with cuda and 180mspt with opencl.
It's not looking too good for cuda atm, I'll have to investigate it further, but I think this is ready to push.

geohot:
Tons of lines and a file open, I don't think you need them all.

cloud11665:
should the half4 stuff be inlined into CUDAProgram then ? Imo it's not worth it as that'd make the output much more noisy. On the other hand, we could solve both of these issues by inlining and having a token like `//KERNEL START` that goes after evevrything in header.

As for the code size, I can compress the ternary expr ifs

cloud11665:
is this tiny enough ?

tinygrad/runtime/ops_cuda.py
```diff
@@ -47,8 +47,21 @@ def __call__(self, global_size, local_size, *args, wait=False):
...
-    lid = [f'threadIdx.{chr(120+i)}' for i in range(3)])
+    lid = [f'threadIdx.{chr(120+i)}' for i in range(3)],
+    half_prekernel = """
+      #include <cuda_fp16.h>
+      struct __align__(8) half4 {
+        half2 x, y;
+        __device__ explicit operator float4() const;
+      };
+      __device__ __forceinline__ half4::operator float4() const {return make_float4(__half2float(x.x), __half2float(x.y), __half2float(y.x), __half2float(y.y)); }
+      __device__ __forceinline__ half4 operator+(half4 a, half4 b) { return half4{a.x+b.x, a.y+b.y}; }
```
geohot, line 59: Are these operators used anywhere?

cloud11665: they are not, llama tests pass without them. Actually, we can also define the cast operator in the struct directly. my bad.

tinygrad/codegen/cstyle.py
```diff
@@ -141,8 +141,10 @@ def kk(s): kernel.append("  "*depth+s)
       # NOTE: if min and max are both 0, it should be a CONST in the Linearizer
       if args.valid.min == 1: kk(f"{newvar.render(True)} = {val};")
       else:
-        zero = f"{lang.float4}(0.0f, 0.0f, 0.0f, 0.0f);" if newvar.ltype == LocalTypes.float4 else "0.0f"
-        kk(f"{newvar.render(True)} = ({args.valid.render(render_cl)}) ? ({val}) : {zero};")
+        if   newvar.ltype == LocalTypes.float4: expr = f"({val}) : {lang.float4}(0.0f, 0.0f, 0.0f, 0.0f);"
```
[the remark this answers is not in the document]
cloud11665: yes!

PR #972, opened 2023-06-11
One hot

stevenandersonz:
Before
=== 225 failed, 615 passed, 1794 skipped, 3 warnings in 16.35s ===

After
=== 221 failed, 619 passed, 1794 skipped, 2 warnings in 16.10s ===

I'm not sure if using `int(depth.lazydata.realize().realized._buf)` is allowed. depth is a Tensor scalar but Tensor.arange doesn't support tensors.

### DIFF
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -252,3 +252,32 @@ def NegativeLogLikelihoodLoss(input, target, weight=None, ignore_index=None, red
   if reduction == "mean": return loss.mean() if weight is None else loss.sum() / weight.sum()
   elif reduction == "sum": return loss.sum()
   return loss.reshape(t_shape) if len(i_shape) != 3 else loss
+
+def OneHot(indices, depth, values, axis=-1):
+  print(f"indices: {indices.numpy()}")
+  print(f"depth: {depth.numpy()}")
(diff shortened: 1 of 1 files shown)


extra/onnx_ops.py
```diff
@@ -251,4 +251,12 @@ def NegativeLogLikelihoodLoss(input, target, weight=None, ignore_index=None, red
   loss = (-mask * input).sum(axis=1) * (1 if weight is None else weight)  
   if reduction == "mean": return loss.mean() if weight is None else loss.sum() / weight.sum()
   elif reduction == "sum": return loss.sum()
-  return loss.reshape(t_shape) if len(i_shape) != 3 else loss
\ No newline at end of file
+  return loss.reshape(t_shape) if len(i_shape) != 3 else loss
+
+def OneHot(indices, depth, values, axis=-1):
+  indices, rank = (indices.cast(dtypes.float32) < 0).where(indices+depth, indices), len(indices.shape)
+  C = int(depth.lazydata.realize().realized._buf)
```
dc-dc-dc, line 258: `safe_numpy(depth).item()` will extract the single scalar value

stevenandersonz: Sweet, thank u.

PR #973, opened 2023-06-12
Add BCELoss and BCELossWithLogits to training.py

MahirAbbas:
initial implementation. two implementations of BCELoss, one which adds epsilon small (same way in pytorch) and one with clamping.
also BCELossWithLogits. no clamping/EPS as sigmoid should remove issue of 0 in 99% scenarios. can add.


MahirAbbas:
pytorch does use EPS, and not clamping. probably more performant

geohot:
Tests failing. Also, this introduces new functionality. Whenever new functionality is introduced it needs to be tested.

MahirAbbas:
put test file in /test (don't know if correct place). BCEWithLogitsLoss outputs different values  than in pytorch. definitions seem to be the same. created two models in /test/test_BCELoss. one in pytorch and one in tinygrad. manually set weights. loss is different from pytorch

MahirAbbas:
need to add more test. initial test works now.

geohot:
Tests not in a framework, EPS isn't a param, not used anywhere. For now, won't merge.

PR #975, opened 2023-06-12
Update graph.py and Simplifies assignment logic for G

metacritical:
Simplifies assignment logic for G

### DIFF
--- a/tinygrad/graph.py
+++ b/tinygrad/graph.py
@@ -1,8 +1,5 @@
 import os, atexit, itertools
-try:
-  import networkx as nx  # type: ignore
-except ImportError:
-  nx = None # graph won't work
+import networkx as nx
 from collections import defaultdict
 from typing import Dict, List, Optional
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/graph.py                          1               1
```

tinygrad/graph.py
```diff
@@ -14,7 +14,7 @@
 
 # **** debugging and graphing ****
 
-G = nx.DiGraph() if nx is not None else None
+G = nx.DiGraph()
```
jla524, line 17: if networkx is not installed, this will fail with `AttributeError: 'NoneType' object has no attribute 'DiGraph'`

jla524:
@geohot can we get rid of this try except logic, since networkx is in `setup.py`?
```
try:
  import networkx as nx  # type: ignore
except ImportError:
  nx = None  # graph won't work
```

tinygrad/graph.py
metacritical: If setup doesnt work we have a different problem also this is handled above.

PR #976, opened 2023-06-13
F64 support

### LINKED ISSUE #891: float64 support
Support for float64 over all plattforms.
This would also fix up to 20 ONNX tests.

Is there a plan to add this? Otherwise there is no way to fully support ONNX, since some tests are hard relying on that.

dc-dc-dc:
Adds float64 support to the different runtimes, also included an excluded field on dtypes for unsupported platforms

From my testing float64 is not supported on metal and OpenCL on M2 Mac. OpenCL float64 is fine on my windows machine although this is an optional feature and ideally should check the OpenCL extensions for it.

Also brings the onnx stat line to:
190 failed, 650 passed, 1794 skipped, 2 warnings in 14.73s

Tested on:
- CPU
- GPU
- CUDA
- LLVM
- CLANG

Closes #891

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -53,6 +53,7 @@ class DType(NamedTuple):
   itemsize: int
   name: str
   np: type  # TODO: someday this will be removed with the "remove numpy" project
+  excluded: Tuple[str] = ()
   def __repr__(self): return f"dtypes.{self.name}"
 
 # dependent typing?
@@ -69,7 +70,7 @@ class dtypes:
   @staticmethod # static methds on top, or bool in the type info will refer to dtypes.bool
   def is_int(x: DType)-> bool: return x in (dtypes.int8, dtypes.uint8, dtypes.int32, dtypes.int64)
   @staticmethod
-  def is_float(x: DType) -> bool: return x in (dtypes.float16, dtypes.float32)
+  def is_float(x: DType) -> bool: return x in (dtypes.float16, dtypes.float32, dtypes.float64)
   @staticmethod
   def is_unsigned(x: DType) -> bool: return x in (dtypes.uint8)
   @staticmethod
@@ -77,6 +78,7 @@ class dtypes:
   bool: Final[DType] = DType(0, 1, "bool", bool)
   float16: Final[DType] = DType(0, 2, "half", np.float16)
   float32: Final[DType] = DType(4, 4, "float", np.float32)
+  float64: Final[DType] = DType(5, 8, "double", np.float64, ("METAL"))
   int8: Final[DType] = DType(0, 1, "char", np.int8)
   int32: Final[DType] = DType(1, 4, "int", np.int32)
   int64: Final[DType] = DType(2, 8, "int64", np.int64)
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -80,7 +80,7 @@ class dtypes:
   bool: Final[DType] = DType(0, 1, "bool", bool)
   float16: Final[DType] = DType(0, 2, "half", np.float16)
   float32: Final[DType] = DType(4, 4, "float", np.float32)
-  float64: Final[DType] = DType(5, 8, "double", np.float64, ("METAL", "GPU" if OSX else ""))
+  float64: Final[DType] = DType(5, 8, "double", np.float64, ("METAL", "GPU") if OSX else ())
   int8: Final[DType] = DType(0, 1, "char", np.int8)
   int32: Final[DType] = DType(1, 4, "int", np.int32)
   int64: Final[DType] = DType(2, 8, "int64", np.int64)
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
+import platform
 from dataclasses import dataclass, asdict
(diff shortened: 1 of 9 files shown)


dc-dc-dc:
Will add the following in the kernel prefix
```
#ifdef cl_khr_fp64
#pragma OPENCL EXTENSION cl_khr_fp64 : enable
#elif defined(cl_amd_fp64)
#pragma OPENCL EXTENSION cl_amd_fp64 : enable
#endif
```

tinygrad/tensor.py
```diff
@@ -36,6 +36,7 @@ class Tensor:
 
   def __init__(self, data:Union[int, float, list, tuple, LazyBuffer, np.ndarray], device=Device.DEFAULT, dtype:Optional[DType]=None, requires_grad:Optional[bool]=None):
     assert dtype is None or isinstance(dtype, DType), f"invalid dtype {dtype}"
+    assert dtype is None or device not in dtype.excluded, f"dtype: {dtype} not supported on device: {device}"
```
geohot, line 39: You want this after the device `canonicalize` and you only want to check before the colon.

"gpu:0", "GPU:0" can both be passed into device

tinygrad/runtime/ops_gpu.py
```diff
@@ -87,7 +86,7 @@ def __call__(self, global_size, local_size, *bufs, wait=False) -> Optional[float
 
 class CLCodegen(CStyleCodegen):
   lang = CStyleLanguage(
-    kernel_prefix = "#define int64 long\n__kernel", buffer_prefix = "__global ", smem_prefix = "__local ",
+    kernel_prefix = "#ifdef cl_khr_fp64\n#pragma OPENCL EXTENSION cl_khr_fp64 : enable\n#elif defined(cl_amd_fp64)\n#pragma OPENCL EXTENSION cl_amd_fp64 : enable\n#endif\n#define int64 long\n__kernel", buffer_prefix = "__global ", smem_prefix = "__local ",
```
geohot, line 89: Want to do something similar to half_prekernel so this isn't in all the kernels?

tinygrad/helpers.py
```diff
@@ -69,14 +72,15 @@ class dtypes:
...
-  def is_float(x: DType) -> bool: return x in (dtypes.float16, dtypes.float32)
+  def is_float(x: DType) -> bool: return x in (dtypes.float16, dtypes.float32, dtypes.float64)
   @staticmethod
   def is_unsigned(x: DType) -> bool: return x in (dtypes.uint8)
   @staticmethod
   def from_np(x) -> DType: return asdict(dtypes())[np.dtype(x).name]
   bool: Final[DType] = DType(0, 1, "bool", bool)
   float16: Final[DType] = DType(0, 2, "half", np.float16)
   float32: Final[DType] = DType(4, 4, "float", np.float32)
+  float64: Final[DType] = DType(5, 8, "double", np.float64, ("METAL", "GPU" if OSX else ""))
```
geohot, line 83: I don't think you want "" here if not OSX, I think you just don't want that element in the tuple

dc-dc-dc: yeah will change to `("METAL", "GPU") if OSX else ()`

tinygrad/runtime/ops_gpu.py
dc-dc-dc: misread this, ignore the last commit

tinygrad/tensor.py
```diff
@@ -37,6 +37,7 @@ class Tensor:
   def __init__(self, data:Union[int, float, list, tuple, LazyBuffer, np.ndarray], device=Device.DEFAULT, dtype:Optional[DType]=None, requires_grad:Optional[bool]=None):
     assert dtype is None or isinstance(dtype, DType), f"invalid dtype {dtype}"
     device = Device.canonicalize(device)
+    assert dtype is None or device.split(":", 1)[0] not in dtype.excluded, f"dtype: {dtype} not supported on device: {device}"
```
geohot, line 40: These asserts should probably be at a much lower level, like when you create the rawbuffer. Then you only need one assert, and you don't need the excluded on dtypes either.

geohot:
Thanks! Now those onnx tests should pass

rayanht:
Should probably add a pytest decorator to skip the float64 tests on Mac since it's not supported

jdlib:
As @dc-dc-dc mentioned initially:

> OpenCL float64 is fine on my windows machine although this is an optional feature and ideally should check the OpenCL extensions for it.

I ran on Windows with a Intel Iris XE card and it fails the TestFloat64Dtype tests
```
error: use of type 'double' requires cl_khr_fp64 extension to be enabled
```
The generated cl file starts as expected, e.g. with
```
#ifdef cl_khr_fp64
#pragma OPENCL EXTENSION cl_khr_fp64 : enable
#elif defined(cl_amd_fp64)
#pragma OPENCL EXTENSION cl_amd_fp64 : enable
#endif
__kernel void E_4(__global double* data0, const __global double* data1, const __global double* data2) {
```
but the extensions of that card only contain `cl_khr_fp16` but not `cl_khr_fp64`.
Also since the error message is somewhat misleading, would it be of advantage to add a #else branch to the generated code?
```
...
#else
#error "fp64 not supported by device"
#endif
````

PR #979, opened 2023-06-14
FIX: Can't pickle local object

Maroc-OS:
_early_exec_process is a local function that is defined whiting the scope of another function, should be global.

also :
* replaced while 1 by while True
* moved imports to top.

this is the error example when testing GPU Backend.
https://github.com/geohot/tinygrad/issues/980


geohot:
Will merge if tests pass

Issue #981, opened 2023-06-14
CLANG on windows errors when running clang command


dc-dc-dc:
When trying to run the CLANG runtime on windows I get an error when building the lib.

The issues stems from [here](https://github.com/geohot/tinygrad/blob/2d567ef688473734cd21f31ae9344d06d01d6ec4/tinygrad/runtime/ops_clang.py#LL13C11-L13C11)  as this is building a shared library for the unix system with os specific flags.

From some quick testing the two flags in question are
- fPIC (I think for windows this is already enabled by default if its a shared library)
- lm (attempting to link the math lib but on windows its named something else?)

Also saw that ctypes has some flags for windows which may play a factor here as well

geohot:
Should be fixed with #986!

PR #986, opened 2023-06-15
clang rt for msvc

bl-ke:
addressing #981 
- split clang cflags, dynlib extension, and function export decorator by platform
- removed clang-rt from msvc linker args based on comment, added msvc dllexport decorator for kernel function
- retained darwin cflags from the linux config as i am not setup to verify correctness (assuming clang rt worked as expected on darwin)

tested on: win10, arch

note: these cflags assume the clang in PATH is set to target the host machine by default, some additional configuration would allow cross compilation


bl-ke:
Trimmed the fat. Should I look into adding Windows CI as well?

geohot:
Merged, but yea, that'd be cool to prevent regressions.

PR #812, opened 2023-05-26
Remove usage of `isinstance` in the codebase + various optimizations

rayanht:
# Summary

Remove the overwhelming majority of `isinstance` calls in favor of `x.__class__` comparisons. Closes #810 

# Test Plan

## Correctness

Local test suite is passing. Wait for CI

## Performance

`test_net_speed.py` benchmark:

### Before

```
test/test_net_speed.py::TestConvSpeed::test_mnist torch forward pass:  19.182 ms
torch backward pass: 19.889 ms
         362545 function calls (351966 primitive calls) in 234.580 seconds

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        5   12.607    2.521   12.832    2.566 ops_gpu.py:44(_copyin)
     1142   12.266    0.011   42.102    0.037 symbolic.py:59(sum)
    55734    8.959    0.000    8.959    0.000 {built-in method builtins.isinstance} <<<<<<<<<<<<<
    11876    7.671    0.001   18.786    0.002 symbolic.py:30(__mul__)
8036/4978    6.373    0.001   11.315    0.002 symbolic.py:15(render)
    10086    5.861    0.001    9.720    0.001 symbolic.py:128(__init__)
       16    5.621    0.351    5.621    0.351 {built-in method io.open}
      791    5.115    0.006   34.018    0.043 shapetracker.py:185(reshape)
     1845    4.884    0.003    7.792    0.004 shapetracker.py:33(__init__)
   910/20    4.742    0.005  139.127    6.956 lazy.py:107(realize)

forward pass:  16.117 ms, 0.84x off baseline 19.182 ms
backward pass: 31.474 ms, 1.58x off baseline 19.889 ms
```

### After

```
test/test_net_speed.py::TestConvSpeed::test_mnist torch forward pass:  24.423 ms
torch backward pass: 22.493 ms
         362565 function calls (351986 primitive calls) in 230.995 seconds

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1142   11.326    0.010   39.657    0.035 symbolic.py:59(sum)
       16   10.550    0.659   10.550    0.659 {built-in method io.open}
    11876    7.674    0.001   19.093    0.002 symbolic.py:30(__mul__)
    10086    5.659    0.001    9.878    0.001 symbolic.py:125(__init__)
      791    5.268    0.007   33.978    0.043 shapetracker.py:185(reshape)
        5    4.744    0.949    4.989    0.998 ops_gpu.py:44(_copyin)
[...]

PR #987, opened 2023-06-15
Optimizations in lazy.py

rayanht:
Before:

```
lazy.py: 0.9059 seconds
tensor.py: 0.4541 seconds
shapetracker.py: 0.3458 seconds
ops.py: 0.2825 seconds
symbolic.py: 0.1162 seconds
helpers.py: 0.0586 seconds
```

After:

```
lazy.py: 0.5744 seconds
tensor.py: 0.4551 seconds
shapetracker.py: 0.2559 seconds
helpers.py: 0.2406 seconds
ops.py: 0.2400 seconds
symbolic.py: 0.1097 seconds
```

Deltas:

`lazy.py         -0.33s`
`shapetracker.py -0.10s`
`ops.py          -0.05s`
`helpers.py      +0.19s`

Net perf gain = `-0.29s`


tinygrad/graph.py
```diff
@@ -4,9 +4,10 @@
 except ImportError:
   nx = None # graph won't work
 from collections import defaultdict
-from typing import Dict, List, Optional
-from tinygrad.ops import UnaryOps, BinaryOps, ReduceOps, MovementOps, LoadOps, FusedOps, Op, OpType, LazyOp, get_buffers, get_lazyops
-from tinygrad.lazy import LazyBuffer
+from typing import TYPE_CHECKING, Dict, List, Optional
+from tinygrad.ops import UnaryOps, BinaryOps, ReduceOps, MovementOps, LoadOps, FusedOps, Op, OpType, LazyOp
+if TYPE_CHECKING:
```
geohot, line 9: Is this needed? Does it make it faster?

tinygrad/helpers.py
```diff
@@ -99,3 +109,36 @@ class GlobalCounters:
   cache: ClassVar[Optional[List[Tuple[Callable, Any]]]] = None
   @staticmethod
   def reset(): GlobalCounters.global_ops, GlobalCounters.global_mem, GlobalCounters.time_sum_s, GlobalCounters.kernel_count, GlobalCounters.cache = 0,0,0.0,0,None
+
+# Stripped down version of a WeakSet
```
geohot, line 113: Wow, are these classes really faster? If so, even though it adds lines, I'm supportive of this. It surfaces complexity.

Do we really need both a Set and a Dictionary, or is just a Dictionary enough?

tinygrad/graph.py
rayanht: The reason for this is that we used to import the graph utilities in the `__init__` of LazyBuffer

We need this to break the circular dependency and be able to import the graph utils only once at the top-level of `lazy.py`

tinygrad/helpers.py
```diff
@@ -1,18 +1,23 @@
...
 # NOTE: helpers is not allowed to import from anything else in tinygrad
 OSX = platform.system() == "Darwin"
 
 def dedup(x): return list(dict.fromkeys(x))   # retains list order
-def prod(x:Union[List[int], Tuple[int, ...]]) -> int: return math.prod(x)
-def argfix(*x): return tuple() if len(x) == 0 else tuple(x[0]) if isinstance(x[0], (tuple, list)) else tuple(x)
+def argfix(*x): 
+  if x[0].__class__ in {tuple, list}:
+    try: return tuple(x[0])
+    except IndexError: return tuple()
```
geohot, line 17: No way an exception is faster than a len check

tinygrad/graph.py
```diff
@@ -49,11 +50,11 @@ def str_dtype(dtyp):
   ret = str(dtyp)[7:]
   return "" if ret == 'float' else f"\n{ret}"
 
-def log_op(ret: LazyBuffer, ast: LazyOp, show_graph: Optional[bool] = None, phantom=False):
+def log_op(ret: 'LazyBuffer', ast: LazyOp, show_graph: Optional[bool] = None, phantom=False):
```
geohot, line 53: Yea if this is for circular imports, there should be a cleaner way.

tinygrad/helpers.py
rayanht: This is a statistical optimization, `try` blocks that don't throw an exception are virtually free and in the overwhelming majority of calls to argfix we are in the exception-free path

tinygrad/lazy.py
```diff
@@ -198,89 +164,120 @@ def cast(self:LazyBuffer, arg:DType) -> LazyBuffer: return elementwise_op(UnaryO
...
+        return self.op.replace_with_movement_ops([(MovementOps.PERMUTE, arg)])
+    ret = create_lazybuffer(self.device, ShapeTracker(self.st).permute(arg), MovementOps, LazyOp(MovementOps.PERMUTE, (self,), arg, (self,)), self.dtype)
+    return self.prune_movement_nops(ret)
+  
+  def shrink_op(self:LazyBuffer, arg:Tuple[int, ...]) -> LazyBuffer:
+    if all([b - a == s for s, (a, b) in zip(self.shape, arg)]): return self
+    if not self.realized and self.op.op == MovementOps.SHRINK:
+      self.op.src[0].children.discard(self)
+      return self.op.src[0].shrink_op(tuple([(b1+b2, b1+e2) for (b1,e1),(b2,e2) in zip(self.op.arg, arg)]))
+    if SHUFFLE_MOVEMENT_OPS and self.optype == BinaryOps and not self.realized and len(self.children) == 0:
```
geohot, line 244: Hmm, a lot of copy pasted code in these. I'm never willing to trade off maintainability/readability for a couple percent of speed.

rayanht: hmm here I can extract the `SHUFFLE_MOVEMENT_OPS` optimization into a helper (similar to what i did with `REMOVE_MOVEMENT_NOPS` and the `prune_movement_nops` helper)

tinygrad/graph.py
geohot: Hmm, I think there has to be cleaner way. The graph utils should sit at the top of the stack, I'm even fine with dynamically importing them only if GRAPH env var is set.

tinygrad/lazy.py
geohot: Yea, overall this change is +144 lines. I'm okay with lines, but not if it's copy/paste harder to read.

What would make this code the most readable? I'm okay with losing a few percent of perf.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py            17              17
tinygrad/graph.py                          7               6
tinygrad/helpers.py                       52               9
tinygrad/lazy.py                         206             153
tinygrad/mlops.py                         37              19
tinygrad/ops.py                           47              23
tinygrad/runtime/lib.py                    4               0
total                                    370             227
lines added in the tinygrad folder: 143
```

tinygrad/codegen/linearizer.py
```diff
@@ -308,13 +308,13 @@ def ssa(name, ltype=LocalTypes.float) -> Token:
...
         # there's no AST here (and there's no shape for the reduce LazyOp)
-        self.ast_parse(LazyOp(self.reduceop.op, ("LOCAL_BUFFER",)), [acc[off] for off in self.acc_offsets(-1)], loaded_buffers, ssa, do_reduce=True)
+        self.ast_parse(LazyOp(self.reduceop.op, ("LOCAL_BUFFER",), None, (loaded_buffers["LOCAL_BUFFER"],)), [acc[off] for off in self.acc_offsets(-1)], loaded_buffers, ssa, do_reduce=True)
 
         # end the late reduce loop
         self.uop(UOps.ENDLOOP, None, [], (end_local_idxs, "late_reduce"))
 
     # load latebufs
-    loaded_buffers.update({b:self.global_load(i, global_idxs[:removed]+fake_reduce_idxs) for i,b in enumerate(self.bufs) if b not in self.earlybufs and i != 0 and not isinstance(b, LocalBuffer)})
+    loaded_buffers.update({b:self.global_load(i, global_idxs[:removed]+fake_reduce_idxs) for i,b in enumerate(self.bufs) if b not in self.earlybufs and i != 0 and not b.__class__ == LocalBuffer})
```
roelofvandijk, line 317: `b.__class__ is LocalBuffer`

tinygrad/codegen/linearizer.py
```diff
@@ -442,14 +442,14 @@ def simplify_merge_adjacent(self):
...
-    for i,x in enumerate(rets): self.sts[i].reshape(tuple(y[0] for y in x))
+    for i,x in enumerate(rets): self.sts[i].reshape(tuple([y[0] for y in x]))
 
   # ******************** GPU simplifiers ********************
 
   def required_optimizations(self, early_only=False):
     for buf_index,buf in enumerate(self.bufs):
       unit_stride_axes_mul_4 = [i for i in self.sts[buf_index].unit_stride_axes() if self.sts[buf_index].shape[i]%4 == 0]
-      if (not early_only or buf in self.earlybufs) and isinstance(self.bufs[buf_index].dtype, ImageDType):
+      if (not early_only or buf in self.earlybufs) and self.bufs[buf_index].dtype.__class__ is ImageDType:
```
roelofvandijk, line 452: `self.bufs[buf_index].dtype is ImageDType` should suffice.

tinygrad/helpers.py
```diff
@@ -1,18 +1,23 @@
...
+  return tuple(x)
 def argsort(x): return type(x)(sorted(range(len(x)), key=x.__getitem__)) # https://stackoverflow.com/questions/3382352/equivalent-of-numpy-argsort-in-basic-python
-def all_same(items): return all(x == items[0] for x in items) if len(items) > 0 else True
+def all_same(items): return all([x == items[0] for x in items]) if len(items) > 0 else True
```
roelofvandijk, line 20: `def all_same(items): return all([x == items[0] for x in items]) if len(items) > 1 else True`

tinygrad/helpers.py
```diff
@@ -89,6 +97,8 @@ def from_np(x) -> DType: return asdict(dtypes())[np.dtype(x).name]
   uint32: Final[DType] = DType(1, 4, "uint", np.uint32)
   uint64: Final[DType] = DType(2, 8, "uint64", np.uint64)
 
+# HACK: staticmethods are not callable in 3.8 so we have to compare the class
+DTYPES_DICT = {k: v for k, v in dtypes.__dict__.items() if not k.startswith('__') and not callable(v) and not v.__class__ == staticmethod}
```
roelofvandijk, line 101: You could also remove the `dataclass` decorator, instantiate `dtypes()` once at the end of this file and cache the `from_np` function. This will also cache the repeated `np.dtype(x).name` calls.

rayanht:
Cooled down the copy paste, we lost a tiny bit of perf but the lazy ops are much more concise and readable now (apart from permute which has some optimizations that don't apply to other ops). We lost a tiny bit of perf but nothing crazy

| Filename        	| Before 	| After 	| Delta        	|
|-----------------	|--------	|-------	|--------------	|
| lazy.py         	| 0.90s  	| 0.59s 	| -0.31s       	|
| tensor.py       	| 0.45s  	| 0.44s 	| not stat-sig 	|
| ops.py          	| 0.28s  	| 0.23s 	| -0.05s       	|
| shapetracker.py 	| 0.34s  	| 0.22s 	| -0.12s       	|
| helpers.py      	| 0.05s  	| 0.22s 	| +0.17s       	|
| symbolic.py     	| 0.11s  	| 0.14s 	| +0.03s       	|
| **TOTAL**       	| 2.13s  	| 1.84s 	| -0.29s       	|

tinygrad/graph.py
rayanht: We can't do that because the `GRAPH` env var itself is set in `graph.py`. Maybe we could move all env vars to a separate file?

tinygrad/helpers.py
rayanht: Yeah this is faster because the stdlib weak data structures are threadsafe-ish and iteration safe, both features we don't need.

I think we do need both because the dict is more expensive than the set so we only use the dict where it is absolutely necessary (for the addressable cache in `lazy.py`)

tinygrad/lazy.py
```diff
@@ -25,23 +28,23 @@
 def _ast_reduceops(self:LazyBuffer) -> LazyOp:
   # TODO: this can also corealize a binary op after the reduce, not just before
   src = self.op.src[0]
-  if MERGE_ELEMENTWISE_INTO_REDUCE and src.realized is None and src.optype == BinaryOps and len(src.children) <= 1:
+  if MERGE_ELEMENTWISE_INTO_REDUCE and not src.realized and src.optype == BinaryOps and len(src.children) <= 1:
```
roelofvandijk, line 31: Optype is BinaryOps

tinygrad/lazy.py
```diff
@@ -51,125 +54,91 @@ def _ast_binaryops(self:LazyBuffer) -> LazyOp:
...
+  @property
+  def key(self):
+    try: return (self.dtype.key, self.realized.key, self.st.key)
+    except AttributeError: return (self.dtype.key, self.op.op, self.st.key)
+
   def _device_extra_args(self) -> Dict[str, str]: return {"device": self.device.split(":", 1)[1]} if ":" in self.device else {}
 
   def realize(self:LazyBuffer) -> LazyBuffer:
-    if self.realized is None:
+    if not self.realized:
```
roelofvandijk, line 115: Can `exec_ast` fail to return anything and still return `None`?

tinygrad/codegen/linearizer.py
```diff
@@ -94,12 +94,12 @@ def __init__(self, ast:LazyOp, output_buffer:LazyBuffer):
...
 
     # get the output buffers
-    self.bufs = [output_buffer] + dedup(get_buffers(ast))
+    self.bufs = [output_buffer] + dedup(ast.buffers)
 
     # key for lookup in cache (can change, str might not be right)
     # bufs are needed because kernels like f(x) = x + x and f(x, y) = x + y have the same str(ast), but are different kernels.
     # mapping the buffers to integers is required because a-b != b-a (and how would you tell a and b apart?)
-    self.key = f"ASTKernelKey ast={str(map_buffers({x:i for i,x in enumerate(self.bufs)}, ast))} bufs={self.bufs}"
+    self.key = (ast.map_buffers({x:i for i,x in enumerate(self.bufs)}).key, tuple(map(attrgetter("key"), self.bufs)))
```
geohot, line 102: [x["key"] for x in self.bufs] is slower?

tinygrad/ops.py
```diff
@@ -18,19 +20,41 @@ class LoadOps(Enum): EMPTY = auto(); RAND = auto(); CONST = auto(); FROM = auto(
...
-def get_buffers(op:LazyOp) -> List[Any]: return functools.reduce(operator.add, [get_buffers(x) if isinstance(x, LazyOp) else [x] for x in op.src], [])
-def get_lazyops(op:LazyOp) -> List[LazyOp]: return functools.reduce(operator.add, [get_lazyops(x) for x in op.src if isinstance(x, LazyOp)], [op])
-def map_buffers(real_srcs:Dict[Any, Any], x:Any) -> LazyOp:
-  if len(real_srcs) and x in real_srcs: return map_buffers(real_srcs, real_srcs[x]) if isinstance(real_srcs[x], LazyOp) else real_srcs[x]
-  return LazyOp(x.op, tuple((map_buffers(real_srcs, y) if isinstance(y, LazyOp) else real_srcs[y]) for y in x.src), x.arg)
+  __slots__ = "op", "src", "arg", "buffers", "__weakref__"
+  op: Op
+  src: Tuple[Union[LazyOp, LazyBuffer], ...]
+  arg: Union[LazyOp, LazyBuffer]
+  buffers: Tuple[LazyBuffer]
```
geohot, line 29: I don't like that this is storing state in two places. What if this doesn't match src?

tinygrad/codegen/linearizer.py
rayanht: Oh yeah this should just be `x.key for x in self.bufs`, there was a reason why i did that when testing but i can't remember it now

tinygrad/codegen/linearizer.py
geohot: I don't think that's right.`ImageDType` is a class, `dtype` is an instance of that class.

tinygrad/graph.py
geohot: oh I'm fine with `GRAPH` env var moving to helpers.

geohot:
Readability of movement ops looks good.

My main concern now is the addition of `buffers` to LazyOp. This puts state in two places, both in `src` and `buffers`. Can this be a cached method instead?

tinygrad/helpers.py
geohot: I think these classes should probably have some unit tests.

Also, I propose moving them to `lazy.py`, that's the only place that should use this weakref stuff.

tinygrad/ops.py
rayanht: That can't happen by construction

We never delete buffers from the AST and ops are always replaced, never modified in-place so the final computational graph will always be consistent

geohot: Sure, LazyOps are immutable. But my concern is an invalid LazyOp will be constructed where the state doesn't match. There's a few places where they are created, and it's so easy to write a bug. Can't the class just cache the buffers on first access? Or even on construction.

rayanht: On construction would work, we can keep the same `__init__` and just remove the ability to pass in buffers in the constructor, would that alleviate your concerns?

The option to pass in the buffers to the constructor was added because for pretty much all internal operations in `lazy.py`, we know in advance what the buffers are since we're just shuffling around nodes in the AST and so we don't need the additional recursion in the constructor

geohot: Yea, that works. Then it can't have a bug somewhere with a mismatch.

rayanht:
Ok about to push the changes you requested above once I make mypy happy

re: mypy, it's super upset because I tried to put actual types on the LazyOp object (it used to have `Any` for src and args) since we have no code completion otherwise. For example here the IDE has no idea that self.op.src[0] is a LazyBuffer and so it doesn't find `expand_op`. 

Now the issue with having better types is that we need to help mypy in narrowing the types but i can't think of a clean way to do that, we could:

1. Have casts and asserts everywhere, don't think this is a reasonable fix
2. Have an abstract class that defines the methods and have `LazyOp` and `LazyBuffer` both inherit it
3. Content with no code completion

Screenshot 2023-06-19 at 00 17 12

geohot:
I think the abstract class approach is good, yea, I was never happy about that Any.

rayanht:
Yeah it would raise a `NotImplementedError` if called on a LazyOp (which shouldn't happen by construction but that can potentially double up as an extra measure to catch breaking changes?)

geohot:
Ahh, then you might not need an abstract class, perhaps add the stub methods to LazyOp and use a Union?

rayanht:
Hmm yeah okay cool I'm happy with just stubbing

rayanht:
Actually this only solves half of the battle, the other half is attribute accesses. We can get around that with having `@property` stubs but then it gets a bit ridiculous and we end up with something like this in the `LazyOp` class:

```
def realize(self): raise NotImplementedError
def reduce_op(self, _): raise NotImplementedError
def reshape_op(self, _): raise NotImplementedError
def pad_op(self, _): raise NotImplementedError
def expand_op(self, _): raise NotImplementedError
def permute_op(self, _): raise NotImplementedError
def shrink_op(self, _): raise NotImplementedError
def stride_op(self, _): raise NotImplementedError
@property
def children(self): raise NotImplementedError
@property
def realized(self): raise NotImplementedError
@property
def shape(self): raise NotImplementedError
@property
def st(self): raise NotImplementedError
@property
def src(self): raise NotImplementedError
[a few more of these...]
 ```

I'm not quite sure what's preferable between

1. Disabling type checking in `lazy.py` in favor of having code completion
2. Reverting to `Any` and content with no completion

My gut feeling says 1. because `lazy.py` is entirely internal APIs so we don't break anything for end-users by playing fast and loose with types and i feel like the QoL of having completion outweighs having 100% sound typing

In the long term we probably don't actually need separate classes for `LazyOp` and `LazyBuffer`, i think it can all live under `LazyBuffer` and that would solve all the typing shenanigans

**Comment by geohot** [comment]
I actually don't mind 25 lines of stubs. It's a good way to start thinking about merging the classes.

File: tinygrad/lazy.py [hunk 14]
```diff
@@ -51,125 +54,91 @@ def _ast_binaryops(self:LazyBuffer) -> LazyOp:
...
-  if isinstance(y, LazyBuffer):
-    for op, arg in ops: y = y.movement_op(op, arg)
-    return y
-  assert y.op in BinaryOps or y.op in UnaryOps
-  return elementwise_op(y.op, *[replace_with_movement_ops(z, ops) for z in y.src], arg=y.arg)   # type: ignore
+def get_movementroot(root:LazyBuffer, allow_contiguous=False) -> LazyBuffer: return get_movementroot(root.op.src[0], allow_contiguous) if not root.realized and (root.optype == MovementOps or (root.op.op == LoadOps.CONTIGUOUS and allow_contiguous and root.op.src[0].st.contiguous)) else root
+def get_movementroot_contiguous(x:LazyBuffer) -> LazyBuffer: return get_movementroot_contiguous(x.op.src[0]) if not x.realized and x.op.op == LoadOps.CONTIGUOUS else (get_movementroot(x, True) if x.optype == MovementOps and x.st.contiguous else x)
 
-lazycache: WeakValueDictionary[Tuple[str, DType, OpType, LazyOp], LazyBuffer] = WeakValueDictionary()
+lazycache: LightWeakValueDictionary[Tuple[str, DType, OpType, LazyOp], LazyBuffer] = LightWeakValueDictionary()
```
roelofvandijk, line 67: Have you tried removing this wop cache completely? Makes it faster for me.

tinygrad/lazy.py
```diff
@@ -51,125 +54,91 @@ def _ast_binaryops(self:LazyBuffer) -> LazyOp:
...
 def create_lazybuffer(device:str, shape:Union[ShapeTracker, Tuple[int, ...]], optype:OpType, op:LazyOp, dtype:DType):
```
roelofvandijk, line 68: I am not certain this cache actually helps the performance.

rayanht: Yeah i've experimented with it and it's way faster without the lazycache but then we end up with duplicated buffers

The cost of compiling the same buffer multiple times is way lower than this caching mechanism but we would have to change a lot of tests relating to garbage collection...

rayanht:
Update on this, working on merging LazyOp and LazyBuffer into one, should give us a net saving in nlines and probably some additional perf as well

geohot:
Cool, I'm curious to read it. My distinction was two fold.

LazyBuffer has a ShapeTracker. LazyBuffer can be realized. LazyOp doesn't and can't.

geohot:
Maybe for now we should just merge this with the stubs? So it doesn't go stale.

rayanht:
Updated with stubs and addressed the thing about graph imports

geohot:
Sweet, going to merge!

master:
```
codegen         mean runtime:  149.24ms, runs:   158.54,  146.31,  154.19,  150.24,  168.12,  143.27,  142.01,  142.79,  143.06,  143.87
methodcache     mean runtime:  141.55ms, runs:   158.67,  136.71,  136.79,  136.77,  136.66,  136.63,  165.18,  135.95,  135.75,  136.37
profile         mean runtime:  577.50ms, runs:   612.50,  595.11,  596.83,  554.52,  554.58,  562.47,  563.12,  568.43,  596.47,  571.01
```

this branch:
```
codegen         mean runtime:  103.21ms, runs:   186.60,   90.75,   90.47,   90.88,   90.44,  115.21,   91.01,   92.16,   91.96,   92.64
methodcache     mean runtime:   91.17ms, runs:    81.05,  111.85,   88.09,   88.08,   86.29,   85.49,   85.24,  113.39,   85.95,   86.26
profile         mean runtime:  362.23ms, runs:   401.66,  356.05,  346.72,  363.68,  376.81,  361.58,  352.67,  354.32,  356.23,  352.55
```

geohot:
Is there more from #812?

I'm seeing 109.27 ms for the whole thing on my M1 Max

rayanht:
Yeah a couple more minor things, PR coming

tinygrad/helpers.py
roelofvandijk: Correct - however, almost all lists that go in here are <10 items (mostly 1 or 2) long.
In practice, all([]) and any([]) is faster in this codebase based on profiling the calls.

PR #988, opened 2023-06-15
Refactor metal backend for memory efficiency

Maroc-OS:
I hope this one is more clear than the previews PR.

changes made are minimal

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -79,7 +79,7 @@ class MetalProgram:
 
 class MetalCodegen(CStyleCodegen):
   lang = CStyleLanguage(
-    kernel_prefix = "#include <metal_stdlib>;\n#define int64 long\nusing namespace metal;\nkernel", buffer_prefix = "device ", smem_prefix = "threadgroup ",
+    kernel_prefix = "#include <metal_stdlib>\n#define int64 long\nusing namespace metal;\nkernel", buffer_prefix = "device ", smem_prefix = "threadgroup ",
     barrier = "threadgroup_barrier(mem_flags::mem_threadgroup);", float4 = "float4",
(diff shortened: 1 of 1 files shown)


geohot:
There's still mixed whitespace/typing and functional changes.

tinygrad/runtime/ops_metal.py
```diff
@@ -23,10 +22,14 @@ def synchronize(self):
 class RawMetalBuffer(RawBufferMapped):
   def __init__(self, size:int, dtype:DType):
     assert dtype != dtypes.float64, "metal doesn't support float64"
-    super().__init__(size, dtype, METAL.device.newBufferWithLength_options_(size*dtype.itemsize, Metal.MTLResourceStorageModeShared))
+    super().__init__(size, dtype, METAL.device.newBufferWithBytesNoCopy_length_options_deallocator_(None, size*dtype.itemsize, Metal.MTLResourceStorageModeShared, None))
```
geohot, line 25: What if the backing object is freed?

tinygrad/runtime/ops_metal.py
```diff
@@ -23,10 +22,14 @@ def synchronize(self):
...
   def __del__(self):
-    self._buf.release()
+    self.release()
     super().__del__()
+  def release(self):
```
geohot, line 29: Why does this have to be a function?

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_metal.py             10               8
```

tinygrad/runtime/ops_metal.py
geohot: Wait this makes no sense. The pointer is None. Isn't this just the same thing?

geohot:
I'm going to close this. I still don't understand what it gains, there's no Metal CI tests yet, and I don't like passing None to the no copy, it just seems like the wrong constructor.

PR #996, opened 2023-06-17
fix: vals may be None here

Maroc-OS:


geohot:
Please read the else clause

PR #1001, opened 2023-06-18
Fixed issues of dot product with 1D tensors

terafo:
Currently something as simple as `Tensor.arange(2)@Tensor.arange(2)` crashes because dot product tries to access `shape[-2]`. Now 1D@1D and 2D@1D works as expected, in all instances.


tinyb0t:
Changes made in `tinygrad/`:
```
```

terafo:
Found some holes in my fix that weren't in original code, dot product still needs fixing.

terafo:
Actually fixed that, a bit messy, but works with everything I threw on it

tinygrad/tensor.py
```diff
@@ -467,12 +467,13 @@ def conv2d(self, weight:Tensor, bias:Optional[Tensor]=None, groups=1, stride=1,
     ret = (x * weight.reshape(1, groups, rcout, *[1 for _ in range(len(oyx))], cin, *HW)).sum([-1-i for i in range(1+len(oyx))], keepdim=True).reshape(bs, cout, *oyx)
     return ret if bias is None else ret.add(bias.reshape(1, -1, *[1 for _ in range(len(HW))]))
 
-  def dot(self, w:Tensor) -> Tensor:
```
geohot, line 470: Why did the letter change?

tinygrad/tensor.py
```diff
@@ -467,12 +467,13 @@ def conv2d(self, weight:Tensor, bias:Optional[Tensor]=None, groups=1, stride=1,
...
-    if (n1:=len(self.shape))*(n2:=len(w.shape)) == 0: raise RuntimeError(f"both arguments to matmul need to be at least 1D, but they are {n1}D and {n2}D")
+  def dot(self, n:Tensor) -> Tensor:
+    if (n1:=len(self.shape))*(n2:=len(n.shape)) == 0: raise RuntimeError(f"both arguments to matmul need to be at least 1D, but they are {n1}D and {n2}D")
     x = self.reshape(*self.shape[0:-1], 1, self.shape[-1])
-    w = w.reshape(*w.shape[0:-2], 1, w.shape[-2], w.shape[-1]).transpose(-1, -2)
+    w = n.reshape((*n.shape[0:-2], 1, n.shape[-2], n.shape[-1]) if len(n.shape)>1 else (x.shape[-1],1)).transpose(-1, -2)
```
geohot, line 473: Please pay attention to spacing.

tinygrad/tensor.py
terafo: Only because I need original shape later, just realized that it would be better to just move that condition into separate variable and compute at the beginning

geohot:
Was fixed in #1045

PR #1006, opened 2023-06-19
Minor improvements + cleanup to `ops_gpu.py`

Ameobea:
## Changes

 * Add some previously undocumented environment variables from `ops_gpu.py` to `env_vars.md`
 * Update debug print for OpenCL to print the devices that will be used post-filtering with `CL_EXCLUDE`
   * Print more detailed device info rather than raw Python objects with addresses etc.
 * Remove a couple unused or superfluous variables and assignments
 * Use `fromimport` shorthand to shave off a couple precious LOC
 * Couple small whitespace changes to clean things up

## Summary

The main motivation for this was to ease setup for users like me that have to filter out an iGPU from their OpenCL device list.  Trying to launch kernels on the iGPU causes the Python script to hang at best and my computer to blackscreen at worst.

I felt that printing out the _actual_ devices that are used post-filtering made more sense, as without that it seems that the filtering isn't working.

I don't know if some of these changes would be considered "code golf".  I tried to match the terseness and density of the existing code as best I could.

### DIFF
--- a/tinygrad/runtime/ops_gpu.py
+++ b/tinygrad/runtime/ops_gpu.py
@@ -9,26 +9,24 @@ from tinygrad.runtime.lib import RawBufferCopyInOut
 from tinygrad.codegen.cstyle import CStyleCodegen, CStyleLanguage
 
 OSX_TIMING_RATIO = (125/3) if OSX else 1.0   # see test/external_osx_profiling.py to determine this ratio. it's in like GPU clocks or something
-FLOAT16 = getenv("FLOAT16", 0)
 
 # TODO: if you fork and exit the child process after creating anything with cl on AMD, it hangs on e.wait()
 ROCM_LLVM_PATH = pathlib.Path("/opt/rocm/llvm/bin")
 #ROCM_LLVM_PATH = pathlib.Path(__file__).parent.parent.parent.parent / "extra/rocm/build/llvm-project/bin"
 if DEBUG >= 5:
-  from extra.helpers import enable_early_exec
-  early_exec = enable_early_exec()
+  early_exec = fromimport("extra.helpers", "enable_early_exec")()
 
 class _CL:
   def __init__(self):
-    platforms: List[List[cl.Device]] = [y for y in ([x.get_devices(device_type=cl.device_type.GPU) for x in cl.get_platforms()] + [x.get_devices(device_type=cl.device_type.CPU) for x in cl.get_platforms()]) if len(y)]
-    if DEBUG >= 1: print(f"using {platforms[getenv('CL_PLATFORM', 0)]}")
-    self.cl_ctx: cl.Context = cl.Context(devices=[x for x in platforms[getenv('CL_PLATFORM', 0)] if x.name not in getenv('CL_EXCLUDE', "").split(",")])
+    platform: List[cl.Device] = [y for y in ([x.get_devices(device_type=cl.device_type.GPU | cl.device_type.CPU) for x in cl.get_platforms()]) if len(y)][getenv('CL_PLATFORM', 0)]
+    if DEBUG >= 1: print(f"using devices: {[d.hashable_model_and_version_identifier for d in platform if d.name not in getenv('CL_EXCLUDE', '').split(',')]}")
+    self.cl_ctx: cl.Context = cl.Context(devices=[x for x in platform if x.name not in getenv('CL_EXCLUDE', '').split(',')])
     self.cl_queue: List[cl.CommandQueue] = [cl.CommandQueue(self.cl_ctx, device=device, properties=cl.command_queue_properties.PROFILING_ENABLE) for device in self.cl_ctx.devices]
+
   def synchronize(self):
     for q in self.cl_queue: q.finish()
 CL = _CL()
 
-# TODO: merge CLImage in here
 class CLBuffer(RawBufferCopyInOut):
   def __init__(self, size, dtype, device='0'):
     assert not OSX or dtype != dtypes.float64, "OpenCL on Mac doesn't support float64"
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_gpu.py               10              13
```

tinygrad/runtime/ops_gpu.py
```diff
@@ -9,26 +9,24 @@
...
-  from extra.helpers import enable_early_exec
-  early_exec = enable_early_exec()
+  early_exec = fromimport("extra.helpers", "enable_early_exec")()
 
 class _CL:
   def __init__(self):
-    platforms: List[List[cl.Device]] = [y for y in ([x.get_devices(device_type=cl.device_type.GPU) for x in cl.get_platforms()] + [x.get_devices(device_type=cl.device_type.CPU) for x in cl.get_platforms()]) if len(y)]
-    if DEBUG >= 1: print(f"using {platforms[getenv('CL_PLATFORM', 0)]}")
-    self.cl_ctx: cl.Context = cl.Context(devices=[x for x in platforms[getenv('CL_PLATFORM', 0)] if x.name not in getenv('CL_EXCLUDE', "").split(",")])
+    platform: List[cl.Device] = [y for y in ([x.get_devices(device_type=cl.device_type.GPU | cl.device_type.CPU) for x in cl.get_platforms()]) if len(y)][getenv('CL_PLATFORM', 0)]
```
wozeparrot, line 21: This changes the behaviour of this line. Originally they are split to put CPU devices on a lower priority than GPU devices.

tinygrad/runtime/ops_gpu.py
```diff
@@ -92,7 +91,5 @@ class CLCodegen(CStyleCodegen):
     half_prekernel = "#pragma OPENCL EXTENSION cl_khr_fp16 : enable",
     barrier = "barrier(CLK_LOCAL_MEM_FENCE);", float4 = "(float4)",
     gid = [f'get_global_id({i})' for i in range(3)], lid = [f'get_local_id({i})' for i in range(3)], uses_vload=True)
-  supports_float4_alu = True
```
geohot, line 95: Why was this removed?

Ameobea: The parent class defaults these to `True` currently, so behavior is unchanged: https://github.com/geohot/tinygrad/blob/master/tinygrad/codegen/cstyle.py#L182-L183

I figured it was worth the tradeoff to leave these off to save a couple LOC, but can def. add them back in if you'd prefer the explicitness.

tinygrad/runtime/ops_gpu.py
Ameobea: Ahh, makes sense.  I'll revert to the original logic here.

tinygrad/runtime/ops_gpu.py
geohot: Nah, didn't realize they were true. Removing is fine

tinygrad/runtime/ops_gpu.py
```diff
@@ -9,26 +9,24 @@
...
-  from extra.helpers import enable_early_exec
-  early_exec = enable_early_exec()
+  early_exec = fromimport("extra.helpers", "enable_early_exec")()
 
 class _CL:
   def __init__(self):
     platforms: List[List[cl.Device]] = [y for y in ([x.get_devices(device_type=cl.device_type.GPU) for x in cl.get_platforms()] + [x.get_devices(device_type=cl.device_type.CPU) for x in cl.get_platforms()]) if len(y)]
-    if DEBUG >= 1: print(f"using {platforms[getenv('CL_PLATFORM', 0)]}")
-    self.cl_ctx: cl.Context = cl.Context(devices=[x for x in platforms[getenv('CL_PLATFORM', 0)] if x.name not in getenv('CL_EXCLUDE', "").split(",")])
+    if DEBUG >= 1: print(f"using devices: {[d.hashable_model_and_version_identifier for d in platforms[getenv('CL_PLATFORM', 0)] if d.name not in getenv('CL_EXCLUDE', '').split(',')]}")
```
geohot, line 22: This line is duplicated logic from 23

tinygrad/runtime/ops_gpu.py
geohot: Btw, this project doesn't care about LOC anymore, it values readability.

tinygrad/runtime/ops_gpu.py
Ameobea: I could pull out the filtered list of devices for the current platform to a variable, but would require adding another LOC.

I could also move the log to after the context creation, but that would mean it wouldn't log the devices used to create it in the case that context creation raises an exception.

Sorry for all the hassle on this tiny PR; I'm quite new to this project and I'm not sure how far the low-LOC preference goes.

geohot: LOC doesn't matter!!

What's the most readable?

Ameobea: In that case, I'd definitely go for pulling it out to a variable:

```py
    platforms: List[List[cl.Device]] = [y for y in ([x.get_devices(device_type=cl.device_type.GPU) for x in cl.get_platforms()] + [x.get_devices(device_type=cl.device_type.CPU) for x in cl.get_platforms()]) if len(y)]
    devices: List[cl.Device] = [x for x in platforms[getenv('CL_PLATFORM', 0)] if x.name not in getenv('CL_EXCLUDE', '').split(',')]
    if DEBUG >= 1: print(f"using devices: {[d.hashable_model_and_version_identifier for d in devices]}")
    self.cl_ctx: cl.Context = cl.Context(devices=devices)
```

I'll push that up.

tinygrad/runtime/ops_gpu.py
Ameobea: I looked into adding a test.  Found one test that seemed like it was applicable at `test/external/external_test_image.py` but it's not currently working due to depending on removed `CLImage` class.

I'll look fixing this a bit more, but would probably be suited for a separate PR at this point.

geohot:
Cool, will merge if tests pass.

Btw, if you want a fun project to learn the internals of tinygrad, fix https://github.com/geohot/tinygrad/pull/1005/files#r1233484292

Ameobea:
Thanks for your attention here @geohot and help getting me up to speed.

Very cool project; it's the first time I feel like I actually have any kind of view whatsoever on how high-level ML code gets transformed into something that actually runs on the GPU.

And first time I'm getting anywhere near the perf I should be getting on my AMD GPU

PR #1011, opened 2023-06-19
Windows & Ubuntu CLANG CI support

dc-dc-dc:
Main changes were converting uses of '/tmp' to using a helper function that creates os specific paths with temp directory. Only had to skip one test(test_conv_transpose2d) because it was running out of memory in the runner.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,8 +112,10 @@ jobs:
         python-version: 3.8
     - name: Install Dependencies
       run: pip install -e '.[testing]' --extra-index-url https://download.pytorch.org/whl/cpu
+    - name: Set env
+      run: echo -e "CLANG=1\nENABLE_METHOD_CACHE=1" >> $GITHUB_ENV
     - name: Run Pytest
-      run: ENABLE_METHOD_CACHE=1 CLANG=1 python -m pytest -s -v -n=auto test/
+      run: python -m pytest -s -v -n=auto test/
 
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -95,6 +95,27 @@ jobs:
...
+    - name: Checkout Code
+      uses: actions/checkout@v3
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.8
+    - name: Install Dependencies
+      run: pip install -e '.[testing]' --extra-index-url https://download.pytorch.org/whl/cpu
+    - name: Set env
+      run: printf "CI=1\nCLANG=1\nENABLE_METHOD_CACHE=1" >> $GITHUB_ENV
```
geohot, line 116: ENABLE_METHOD_CACHE is the default, you don't need to explicitly turn it on (should be removed from LLVM).

I think `CI=1` is automatic too in CI

geohot:
Nice! Will merge as is, the extra flags don't really matter.

.github/workflows/test.yml
dc-dc-dc: Just checked github docs and `CI=true` by default so can remove it as well as ENABLE_METHOD_CACHE

geohot: Yea, I'd merge a small PR cleaning it up, but no big deal.

PR #1015, opened 2023-06-21
Quickstart

foreign-sub:
* fix quickstart.md
* test quickstart in ci

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -6,7 +6,8 @@ It is intended to be a very quick overview of the high level API that tinygrad p
 This guide is also structured as a tutorial which at the end of it you will have a working model that can classify handwritten digits.
 
 We need some imports to get started:
-```py
+
+```python
 import numpy as np
 import time
 ```
@@ -17,19 +18,22 @@ Tensors are the base data structure in tinygrad. They can be thought of as a mul
 All high level operations in tinygrad operate on these tensors.
 
 The tensor class can be imported like so:
-```py
+
+```python
 from tinygrad.tensor import Tensor
 ```
 
 Tensors can be created from an existing data structure like a python list or numpy ndarray:
-```py
+
+```python
 t1 = Tensor([1, 2, 3, 4, 5])
 na = np.array([1, 2, 3, 4, 5])
 t2 = Tensor(na)
 ```
 
 Tensors can also be created using one of the many factory methods:
-```py
+
+```python
 full = Tensor.full(shape=(2, 3), fill_value=5) # create a tensor of shape (2, 3) filled with 5
 zeros = Tensor.zeros(2, 3) # create a tensor of shape (2, 3) filled with 0
 ones = Tensor.ones(2, 3) # create a tensor of shape (2, 3) filled with 1
@@ -45,24 +49,28 @@ rand = Tensor.rand(2, 3) # create a tensor of shape (2, 3) filled with random va
 randn = Tensor.randn(2, 3) # create a tensor of shape (2, 3) filled with random values from a normal distribution
 uniform = Tensor.uniform(2, 3, low=0, high=10) # create a tensor of shape (2, 3) filled with random values from a uniform distribution between 0 and 10
 ```
+
 There are even more of these factory methods, you can find them in the [tensor.py](/tinygrad/tensor.py) file.
 
 All the tensors creation methods can take a `dtype` argument to specify the data type of the tensor.
-```py
+
+```python
 from tinygrad.helpers import dtypes
 
 t3 = Tensor([1, 2, 3, 4, 5], dtype=dtypes.int32)
 ```
 
 Tensors allow you to perform operations on them like so:
-```py
+
+```python
 t4 = Tensor([1, 2, 3, 4, 5])
 t5 = (t4 + 1) * 2
 t6 = (t5 * t4).relu().log_softmax()
 ```
 
 All of these operations are lazy and are only executed when you realize the tensor using `.realize()` or `.numpy()`.
-```py
+
+```python
 print(t6.numpy())
 # [-56. -48. -36. -20.   0.]
 ```
@@ -77,7 +85,8 @@ These operations are commonly grouped into the `__call__` method of a class whic
 These classes do not need to inherit from any base class, in fact if they don't need any trainable parameters they don't even need to be a class!
 
 An example of this would be the `nn.Linear` class which represents a linear layer in a neural network.
-```py
+
+```python
 # from tinygrad.nn import Linear
 class Linear:
   def __init__(self, in_features, out_features, bias=True, initialization: str='kaiming_uniform'):
@@ -87,12 +96,14 @@ class Linear:
   def __call__(self, x):
     return x.linear(self.weight.transpose(), self.bias)
 ```
+
 There are more neural network modules already implemented in [nn](/tinygrad/nn/__init__.py), and you can also implement your own.
 
 We will be implementing a simple neural network that can classify handwritten digits from the MNIST dataset.
 Our classifier will be a simple 2 layer neural network with a Leaky ReLU activation function.
 It will use a hidden layer size of 128 and an output layer size of 10 (one for each digit) with no bias on either Linear layer.
-```py
+
+```python
 from tinygrad.nn import Linear
 
 class TinyNet:
@@ -108,6 +119,7 @@ class TinyNet:
 
 net = TinyNet()
 ```
+
 We can see that the forward pass of our neural network is just the sequence of operations performed on the input tensor `x`.
 We can also see that functional operations like `leakyrelu` and `log_softmax` are not defined as classes and instead are just methods we can just call.
 Finally, we just initialize an instance of our neural network, and we are ready to start training it.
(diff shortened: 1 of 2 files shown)


docs/quickstart.md
```diff
@@ -101,17 +119,33 @@ class TinyNet:
...
+    x = self.l1(x).leakyrelu()
+    return self.l2(x).log_softmax()
 
 net = TinyNet()
+
 ```
+
 We can see that the forward pass of our neural network is just the sequence of operations performed on the input tensor `x`.
-We can also see that functional operations like `leakyrelu` and `log_softmax` are not defined as classes and instead are just methods we can just call.
+We can also see that functional operations like `leakyrelu` and `log_softmax` are not defined as classes and instead are just methods we can just call like that or with `x = x.leakyrelu()`.
```
wozeparrot, line 130: This line becomes a little confusing with this change.

docs/quickstart.md
```diff
@@ -135,46 +173,76 @@ def cross_entropy(out, Y):
...
 Namely, operations that are load/store like indexing a tensor with another tensor or assigning a value to a tensor at a certain index.
 Load/store ops are not supported in tinygrad because they add complexity when trying to port to different backends and 90% of the models out there don't use/need them.
 
 For our optimizer we will be using the traditional stochastic gradient descent optimizer with a learning rate of 3e-4.
+
 ```py
 from tinygrad.nn.optim import SGD
 
-opt = SGD([net.l1.weight, net.l2.weight], lr=3e-4)
+#opt = SGD([net.l1.weight, net.l2.weight], lr=3e-4)
```
wozeparrot, line 188: Why was this commented out?

docs/quickstart.md
```diff
@@ -135,46 +173,76 @@ def cross_entropy(out, Y):
...
+Now we have everything we need to start training our neural network.
+We will be training for a target loss of 0.03 with a batch size of 64.
+
+```py
+target_loss = 0.03
+tstep, losst, bs = 0, 1., 64
 
+from tqdm import tqdm
+
+tbar = tqdm(total=0, initial=0, ncols=0, bar_format='Training: {desc}')
```
wozeparrot, line 238: Probably shouldn't use tqdm to keep this guide from depending on more dependencies then just tinygrad.

docs/quickstart.md
```diff
@@ -135,46 +173,76 @@ def cross_entropy(out, Y):
...
+def run_net(X,Y,bs):
   # random sample a batch
-  samp = np.random.randint(0, X_train.shape[0], size=(64))
-  batch = Tensor(X_train[samp], requires_grad=False)
+  samp = np.random.randint(0, X.shape[0], size=(bs))
+  batch = Tensor(X[samp], requires_grad=False)
   # get the corresponding labels
-  labels = Y_train[samp]
-
+  labels = Y[samp]
```
wozeparrot, line 222: this shouldn't be in the run_net function as they are not directly related to running the network, they are part of the data loading process.

docs/quickstart.md
```diff
@@ -135,46 +173,76 @@ def cross_entropy(out, Y):
...
+```py
+from tinygrad.nn.optim import get_parameters
+opt = SGD(get_parameters(net), lr=3e-4)
+
+```
 
 Now that we have our network, loss function, and optimizer defined all we are missing is the data to train on!
 There are a couple of dataset loaders in tinygrad located in [/datasets](/datasets).
-We will be using the MNIST dataset loader.
+We will be using the MNIST dataset loader and assign it.
```
wozeparrot, line 205: What does "assign it" mean in this context?

docs/quickstart.md
foreign-sub: I did not want to remove it but at the same time it explained what was sent to SGD in the call below so it ended up with a comment

docs/quickstart.md
foreign-sub: just assign the data to the training and test vars

docs/quickstart.md
wozeparrot: I think keeping the SGD here and removing the one below would be better because this one is more explicit.

docs/quickstart.md
wozeparrot: Maybe be more explicit about what variables its assigning to, like `fetch_mnist` returns a tuple with 4 elements, would be better to specify exactly what is being returned and assigned.

docs/quickstart.md
foreign-sub: i added them to show how much code was shared between running and training in this case otherwise my run_net function was kinda empty and pointless

docs/quickstart.md
foreign-sub: i used tqdm because it is already a requirement from state.py load_state_dict

docs/quickstart.md
```diff
@@ -135,46 +173,76 @@ def cross_entropy(out, Y):
   y = y.reshape(list(Y.shape)+[num_classes])
   y = Tensor(y)
   return out.mul(y).mean()
+
```
wozeparrot, line 176: Depending on when/if #1018 is merged this will have to change to match.

docs/quickstart.md
wozeparrot: Ah OK yea this is fine then.

docs/quickstart.md
foreign-sub: understood

docs/quickstart.md
foreign-sub: yes, i'll rephrase it

docs/quickstart.md
foreign-sub: done

docs/quickstart.md
foreign-sub: yes

docs/quickstart.md
```diff
@@ -4,11 +4,14 @@ This guide assumes no prior knowledge of pytorch or any other deep learning fram
 It is intended to be a very quick overview of the high level API that tinygrad provides.
 
 This guide is also structured as a tutorial which at the end of it you will have a working model that can classify handwritten digits.
+The code for this tutorial is also available in [quickstart.py](/examples/quickstart.py).
```
geohot, line 7: Why is it in two places? They should be merged.

foreign-sub: merging

docs/quickstart.py
```diff
@@ -79,40 +177,65 @@ def cross_entropy(out, Y):
...
 opt = SGD([net.l1.weight, net.l2.weight], lr=3e-4)
 
+# We can see that we are passing in the parameters of our neural network to the optimizer.
+# This is due to the fact that the optimizer needs to know which parameters to update.
+# The parameters are just listed out explicitly here for clarity.
+# There is a simpler way to do this just by using `get_parameters(net)` from `tinygrad.nn.optim`
+# which will return a list of all the parameters in the neural network like below.
+
 #from tinygrad.nn.optim import get_parameters
 #opt = SGD(get_parameters(net), lr=3e-4)
```
wozeparrot: You can keep this uncommented

docs/quickstart.py
```diff
@@ -79,40 +177,65 @@ def cross_entropy(out, Y):
...
-tbar = tqdm(total=0, initial=0, ncols=0, bar_format='Training: {desc}')
-st = time.perf_counter()
-
-while losst > target_loss:
-  labels, out = run_net(X_train, Y_train, bs)
   # compute loss
-  #loss = cross_entropy(out, labels)
-  loss = sparse_categorical_crossentropy(out, labels)
+  loss = cross_entropy(out, labels)
+  #loss = sparse_categorical_crossentropy(out, labels)
```
wozeparrot, line 238: I would say just remove this line and add a comment above saying that `cross_entropy` is the same as `sparse_categorical_crossentropy` from `extra.training`

docs/quickstart.py
foreign-sub: ok

docs/quickstart.py
```diff
@@ -126,11 +249,11 @@ def run_net(X,Y,bs):
...
   pred = np.argmax(out.numpy(), axis=-1)
   acc = (pred == labels).mean()
-  losst = loss.numpy()
+  step_loss = loss.numpy()
 
   tstep += 1
   if tstep % 100 == 0 or tstep == 1:
-    tbar.desc=(f"Loss: {losst:5.4f} | Accuracy: {acc:5.4f} | Step: {tstep}")
+    tbar.desc=(f"Loss: {step_loss:5.4f} | Accuracy: {acc:5.4f} | Step: {tstep}")
     tbar.update()
```
wozeparrot: can we not use `set_description` here?

wozeparrot:
I'm not actually sure I like this...

This doesn't read as nicely as a formatted markdown doc would, this seems kinda cluttered with broken links as stuff, harder for new people to navigate.

Maybe we just ditch the code file for the quickstart but having it tested in CI to make sure it doesn't break would be nice. Thoughts @geohot?

docs/quickstart.py
foreign-sub: added comment

docs/quickstart.py
foreign-sub: yes we can

geohot:
Honestly, I'm pretty indifferent to this whole thing. It's easy to get carried away in reviews on a PR that just shouldn't happen.

When I see PRs like this that are +500 lines, my instinct is not to merge. Code costs money to maintain, probably on the order of $100 in core tinygrad and $10 elsewhere.

foreign-sub:
yeah it looks big because i have not removed quickstart.md yet, but without it the balance would be more like +20 lines

docs/quickstart.md
foreign-sub: removed the run_net function

foreign-sub:
reworked it to be only markdown updates and markdown code test passing in ci

geohot:
Ahh, now this has value. What's the minimal change to have the guide pass CI?

foreign-sub:
the markdown fixes

foreign-sub:
repushed with minimal changes

geohot:
Sweet, merged!

PR #1018, opened 2023-06-21
add onehot and cross_entropy to Tensor

ReinBentdal:
Added `onehot(num_classes)` and `categorical_cross_entropy(Y:Tensor)` to `Tensor` class. `categorical_cross_entropy` assumes Y is onehotted, yields same result as `sparse_categorical_crossentropy`. Both implementations are one-liners, using Tensor manipulations.

Example
```python
out = model.forward(X)
Y = Y.onehot(num_classes)
loss = out.categorical_cross_entropy(Y)
```

Value proposition:
- Firstly, I'm new to tinygrad and i might not yet know why you wouldn't want implementations of these functions in `Tensor`. I might be a dumbass
- Implementing both functions without realizing value of tensor, keeping it lazy.
- To me it feels more aligned with the design of tinygrad, extending the concept of the computation graph on tensors
- Core functionality easily accessible without importing extra files.

Although its nice to be implementable in one-liners, it might be a good idea to add some asserts for easier error messages (check if Y in cross entropy is onehotted).

Tests:
Member function `test_onehot` added to test_tensor.py `TestTinygrad` class. 
Not added test of `categorical_cross_entropy` since yet but same behavior as `nll_loss`. Simple test which tests equality of `categorical_cross_entropy` with `sparse_categorical_cross_entropy` and torch `nll_loss`:
```python
import numpy as np
from extra.training import sparse_categorical_crossentropy

softmax = [[0.1, 0.05, 0.2, 0.6, 0.05], [0.86, 0.05, 0.04, 0.04, 0.01]]
Y = [3, 0]

# tinygrad, Tensor
x = Tensor(softmax)
y = Tensor(Y).onehot(5)
loss = x.categorical_cross_entropy(y)
print(loss.numpy())

# tinygrad, extra
x = Tensor(softmax)
y = np.array(Y)
loss = sparse_categorical_crossentropy(x, y)
print(loss.numpy())

import torch
from torch.nn.functional import nll_loss

# pytorch
x = torch.tensor(softmax)
y = torch.tensor(Y)
loss = nll_loss(x, y)
print(loss.item())
```
> -0.7300001
> -0.73
> -0.7300000190734863

Compare:
New implementations 
```python
def onehot(self, num_classes) -> Tensor: return self.reshape(list(self.shape)+[1]).repeat([1]*len(self.shape)+[num_classes]).eq(Tensor.arange(num_classes))

def categorical_cross_entropy(self, Y:Tensor) -> Tensor: return -self.shape[-1]*self.mul(Y).mean()
```
vs what Is implemented in source code (examples.mlperf.metrics.py & extra.training.py):
```python
def one_hot(arr, num_classes=3):
  res = np.eye(num_classes)[np.array(arr).reshape(-1)]
  arr = res.reshape(list(arr.shape) + [num_classes])
  arr = arr.transpose((0, 4, 1, 2, 3)).astype(np.float32)
  return arr

def sparse_categorical_crossentropy(out, Y):
  num_classes = out.shape[-1]
  YY = Y.flatten().astype(np.int32)
  y = np.zeros((YY.shape[0], num_classes), np.float32)
  # correct loss for NLL, torch NLL loss returns one per row
  y[range(y.shape[0]),YY] = -1.0*num_classes
  y = y.reshape(list(Y.shape)+[num_classes])
  y = Tensor(y)
  return out.mul(y).mean()
```


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         4               0
lines added in the tinygrad folder: 4
```

geohot:
Do these match torch or numpy APIs? Are they fully tested?

ReinBentdal:
> Do these match torch or numpy APIs? Are they fully tested?

## API
Context, from `CONTRIBUTING.md`:
> If there is a feature in PyTorch and numpy that you have actually seen people use, we probably want it. All new features must include good robust tests, and in general matching the PyTorch API is good.
### onehot
The proposal:
```python
class Tensor:
  def onehot(self, num_classes) -> Tensor
```
From torch we have
```python
torch.nn.functional.one_hot(tensor, num_classes=- 1) → LongTensor
```
If `num_classes == -1` then the dimension will be inferred from the data in the tensor. I think its not a good idea to infer, might lead to unexpected errors at a later stage. And always easy just to provide this number.
pytorch one_hot returns LongTensor. Not sure whats ideal for tinygrad.

For `onehot` vs `one_hot`, think `onehot` aligns a tiny bit better with tinygrad naming convention (dropout, batchnorm, hardtanh, ...).

### cross entropy
The proposal:
```python
class Tensor:
  def negative_log_likelihood(self, target:Tensor, ignore_index=-100, weight:Tensor = None, reduction:Literal['none', 'mean', 'sum'] = 'mean') -> Tensor: 
  def cross_entropy(self, target:Tensor, ignore_index=-100, weight:Tensor = None, reduction:Literal['none', 'mean', 'sum'] = 'mean') -> Tensor:
```
From torch we have (not considering deprecated stuff)
```python
torch.nn.functional.nll_loss(input, target, weight=None, ignore_index=- 100, reduction='mean')
torch.nn.functional.cross_entropy(input, target, weight=None, ignore_index=- 100, reduction='mean', label_smoothing=0.0)
```
They are functionally the same except i didn't implement `label_smoothing`, didn't invest time figuring out what it is. 
Note, it differs from `sparse_categorical_crossentropy`, implemented in tinygrad `extra.training`, where the target is a numpy array, not Tensor.
The only difference between NLL and cross_entropy is that cross_entropy operates on logits while NLL operates on log probs.

Now a simple train could look like this:
```python
out = model.forward(X)
loss = out.cross_entropy(Y)
optim.zero_grad()
loss.backward()
optim.stem()
```

## Tests
Tests for both implemented in `test_tensor.py`. Implementations tested head to head with pytorch.

## Discussion
These are mainly convenience functions. The implementations could be simplified if we removed some of the function parameters which pytorch includes. If anyone needs a specific variant of these functions, they may just make it from scratch themself easily.
I initially wanted `negative_log_likelihood` to be even simpler, having onehot targets as parameter. onehot parameters is not common in other libraries because its possible to do computations more efficient without. Currently the functions implemented use onehot internally anyways since it makes it easier to only use tensor operations. I think from a conceptual point, it would be nice to have  `negative_log_likelihood` simple and easily understandable (onehot parameter targes instead of class indices) and have `cross_entropy` as this more convenient thing most people will use.

Drawback of implementing loss functions inside `Tensor` is that it makes it a bit less elegant to implement trainer functions.

ReinBentdal:
Not sure why i'm getting
> FAILED test/test_tensor.py::TestTinygrad::test_cross_entropy - AssertionError: dtype mismatch on realize got dtypes.long expected dtypes.int

from unit test. Will have to look into it later, unless someone else figures it out

tinygrad/tensor.py
```diff
@@ -617,6 +617,30 @@ def dropout(self, p=0.5) -> Tensor:
     mask = (Tensor.rand(*self.shape, requires_grad=False) >= p).cast(dtypes.bool)
     return self * mask * (1/(1.0 - p))
 
+  # note: the way onehot is implemented, if value is out of range, all 0s are returned
+  def onehot(self, num_classes) -> Tensor: 
+    assert dtypes.is_int(self.dtype), f"onehot only supported for ints, got {self.dtype}"
+    return self.reshape(list(self.shape)+[1]).repeat([1]*len(self.shape)+[num_classes]).eq(Tensor.arange(num_classes, dtype=dtypes.int32)) # might not be ideal type, could be bool or int64 (convenience when type is passed further)
```
ReinBentdal, line 623: safe to assume `Tensor.arange(num_classes, dtype=dtypes.int32)` will never overflow? torch one_hot returns long

tinygrad/tensor.py
```diff
@@ -617,6 +617,30 @@ def dropout(self, p=0.5) -> Tensor:
...
+  
+  def negative_log_likelihood(self, target:Tensor, weight:Optional[Tensor] = None, ignore_index=-100, reduction:Literal['none', 'mean', 'sum'] = 'mean') -> Tensor: 
+    assert target.shape == self.shape[:-1], f"Y dimensions {target.shape} must match all except last self dimension {self.shape}"
+    assert weight is None or (len(weight.shape) == 1 and weight.shape[0] == self.shape[-1]), f"weight dimensions {weight.shape} must match last self dimension {self.shape[-1]}"
+    # label_valid = (Y >= 0).mul(Y < self.shape[-1]).add(Y == ignore_index).minimum(1).min().cast(dtypes.bool)
+    # assert label_valid == True, "Cross entropy label out of range or not 'ignore_index' value" # cant check without realizing tensor :(  
+    num_classes = self.shape[-1]
+    y = target.onehot(num_classes)
+    if weight is not None:
+      W = Tensor.eye(num_classes, dtype=weight.dtype) * weight
```
ReinBentdal, line 633: probably not a big deal in this example, but what would generally be a good choice of dtype in these cases?

test/test_tensor.py
```diff
@@ -105,6 +105,39 @@ def test_dropout(self):
...
+    np.testing.assert_allclose(y.cpu().numpy(), y_torch.cpu().numpy())
+
+  def test_cross_entropy(self):
+    np.random.seed(1337)
+    n, classes = 100, 10
+    x = np.random.random_sample((n, classes)).astype(np.float32)
+    label_mask = np.random.randint(2, size=n)
+    targets = (label := np.random.randint(0, classes, n, dtype=np.int64), np.where(label_mask == 0, -100, label))
+    weights = np.random.choice((0, 1, 2, 3, 4), size=classes, p=(0.9, 0.05, 0.025, 0.015, 0.01)).astype(np.float32), None # torch requires float
+    reductiuons = ['mean', 'sum', 'none']
```
shelwinsunga, line 123: `reductiuons` is a typo?

ReinBentdal: Yep

ReinBentdal:
Have done a review and restricted the types of the functions such that it should always result in valid results. And operations does not change input type. Starting to think the unit test error might be caused by some other code:
> FAILED test/test_tensor.py::TestTinygrad::test_cross_entropy - AssertionError: dtype mismatch on realize got dtypes.double expected dtypes.float

jla524:
Since the PyTorch functions are stored in `torch.nn.functional`, I suggest to move our functions to `tinygrad/nn/functional.py` to match.

ReinBentdal:
```python
Z = X.cross_entropy(Y)
```
Hmm yes, maybe it's a good idea to keep `Tensor` "pure", and not be like for example numpy (where the namespace has everything). From a conceptual point I like the idea that cross_entropy operates on X with Y fusing in (and not a function combining them). Keeping a main computation flow through X to Z. It's then more obvious as well that Z should inherit the type of X (which I think probably always should be the case for operations on X unless otherwise explicitly specified. Thinking into the future, if a chip only supporting float16 and int16 hit the marked, it would be nice if there where no hidden types in functions not specified by the user forcing the use of for example int32).
I think there is a case for both. I would also be happy to keep stuff like this in Tensor, for the reasons above as well as having everything in one place like Numpy is conventient. But then it's less obvious when to keep functions in Tensor and when to move them out and Tensor might become less pure and more bloated. As well as deviating a bit from torch, which I don't see a too big problem with since the functions them self function the same. I would personally be happy to discover I didn't have to import cross entropy from functional.
Maybe not desired but could keep all core Tensor members in a base class and more high level functions in Tensor inherited from base, in the future to make the split more obvious.

wozeparrot:
I think `onehot` is fine to put in tensor, because that's how our gathers are currently implemented and its useful for other cases as well.

Variony (approved):

tinygrad/tensor.py
```diff
@@ -617,6 +617,38 @@ def dropout(self, p=0.5) -> Tensor:
     mask = (Tensor.rand(*self.shape, requires_grad=False) >= p).cast(dtypes.bool)
     return self * mask * (1/(1.0 - p))
 
+  # note: the way onehot is implemented, if value is out of range, all 0s are returned
+  def onehot(self, num_classes) -> Tensor: 
+    assert dtypes.is_int(self.dtype), f"onehot only supported for ints, got {self.dtype}"
+    return self.reshape(list(self.shape)+[1]).repeat([1]*len(self.shape)+[num_classes]).eq(Tensor.arange(num_classes, dtype=self.dtype)) # could cast to smaller type, currently just same as self
+  
+  def negative_log_likelihood(self, target:Tensor, weight:Optional[Tensor] = None, ignore_index=-100, reduction:Literal['none', 'mean', 'sum'] = 'mean') -> Tensor:
```
geohot, line 625: This is tons of lines.

ReinBentdal: yes agree, the extra asserts where to see if something unexpected was going on with the types, ref the failed test. But seems to maybe be HW thing.
Pushed more compact impl.

extra: though if the idea of making the type checks HW-agnostic by instead checking dtype priorities and assure that way the behavior is as expected. For example a future compact accelerator only having float16 should still work.

ReinBentdal: the shape asserts is not needed but makes the error message more understandable, don't have to dig into the code

geohot:
Tests failing, going to close.

PR #1021, opened 2023-06-22
fixed division by zero for fast kernels

EliTheCoder:
When a kernel takes a very short amount of time, calculating GFLOPS does a division by zero when DEBUG is greater than or equal to 2. This is fixed by checking when the estimated time is 0 and if it is, setting it to a very small number


tinyb0t:
Changes made in `tinygrad/`:
```
```

PR #1022, opened 2023-06-22
feat: fix shell alias on readme

Lantianyou:


geohot:
I don't understand, it's the same thing twice

Lantianyou:
If I use the original command in readme on windows, which is `py3 -m pip install -e .`. I was prompted error:
 ``` 
 cannot find py3 as cmdlet, function or shell scripts
 ```
 
I think you may set alias py3=python3 on windows. But for people who have not set these alias, they should use `python3` instead of `py3`.

wozeparrot:
I may have confused `py3` for `py` when I wrote this, haven't used windows in a while. I had the alternate command for windows cuz on windows `python3` might not be in path depending on the install method, which if a user just clicks through the python installer the default is to not put `python3` in the path and only put the python launcher (https://docs.python.org/3/using/windows.html#launcher)

Lantianyou:
I am using windows. I found `py` can be found but not `py3`. 

According to Python official docs, python windows launcher should be called `py` and "The launcher is compatible with all available versions of Python, so it does not matter which version is installed." https://docs.python.org/3.11/using/windows.html#launcher

However, from Pytorch's installation guide, the way to verify whether python is installed correctly is to type `python` on command line. https://pytorch.org/get-started/locally/. Even it is installed on windows and user should add `python` to PATH variable.

image

Also, I did a global GitHub search for py3 on markdown files in sourcegraph, I found very few use cases.

https://sourcegraph.com/search?q=context%3Aglobal+lang%3AMarkdown+%22py3+%22+&patternType=standard&sm=1&groupBy=repo


So, I believe `py` or `python3` is appropriate here. If a windows user have only installed Python3 and , `python` would also be correct.

wozeparrot:
Yea it was a mistype on my end for `py3` its supposed to be `py`.

If you look at the install guide for python on windows (https://docs.python.org/3/using/windows.html#installation-steps), the add to path option is not checked by default, which means that `python3` may not exist for all users depending on how they installed python.

Lantianyou:
Yes. Fully agree. `python3` may not exit for all users, but most popular python package kind of assume people put it on PATH variable. And Python official doc specifially instructed people mannually put it in PATH variable. https://docs.python.org/3/using/windows.html#configuring-python

IMHO, using `python3` is appropriate here. For folks who have not put `python` in PATH, they can easily google or ask ChatGPT  to figure out the solution.

wozeparrot:
Yea, I agree with that, I think just having `python3` should be fine.

PR #1025, opened 2023-06-22
8-bit quantization for llama

terafo:
Weights are quantized with just running `quantize.py`. To inference with quantize weights `--quantized` flag should be used when running LLaMa.
What works:
Absmax quantization of LLaMa weights and their inference
TODO:
Integrate LM eval


geohot:
This is a huge change.

I'm fine with `datasets` -> `tinydatasets` as a clean separate PR first.

geohot:
What's the status on this? Can quantizing be done at load time not in separate file? What do the benchmarks say the perf difference between fp16 and int8 is?

geohot:
Closing this, no progress in over a week.

PR #1026, opened 2023-06-22
[tinygrad] basic `nixos` dev shell on `cudakit` 11.6

b7r6:
**Summary**

I'm unsure if this is a useful thing to merge: my NVIDIA box is running NixOS and it seems at least possible that others might want to use `tinygrad` on that platform. This currently only supports `x86_64-linux` and is only tested with `cudakit` 11.6, though that would be easy to change. As a possible ancillary benefit, if this finds any adoption it would ease the path into `nixpkgs`.

**Test Plan**

I've run all the unit tests under `CPU`, `GPU`, and `CUDA` and there don't seem to be any regressions against a vanilla `virtualenv` installation.

Usage is: `nix develop` and start hacking.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,127 @@
+{
+  description =
+    "tinygrad: For something between PyTorch and karpathy/micrograd. Maintained by tiny corp.";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+
+    flake-utils = {
+      url = "github:numtide/flake-utils";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = { self, nixpkgs, flake-utils }@I:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        P = import nixpkgs {
+          inherit system;
+          config.allowUnfree = true;
+        };
+        B = builtins;
+        L = P.lib;
+
+        pythonVersion = "310";
+        pythonPackageStr = "python${pythonVersion}";
+        pythonNixpkgs = P."${pythonPackageStr}Packages";
+
+        tinygrad = pythonNixpkgs.buildPythonPackage {
+          pname = "tinygrad";
+          version = "v0.6.0";
+          src = ./.;
+
+          propagatedBuildInputs = with pythonNixpkgs; [
+            networkx
+            numpy
+            pillow
+            pyopencl
+            pyyaml
+            requests
+            tqdm
+          ];
+
+          # TODO(b7r6): enable when all tests pass...
+          doCheck = false;
+          pythonImportsCheck = [ "tinygrad" ];
+        };
+
+        commonBuildInputs = with P; [ git nixfmt ];
+        commonPythonPackages = pypkgs:
+          with pypkgs; [
+            # test and development
+            flake8
+            mypy
+            onnx
+            onnxruntime
+            opencv4
+            pylint
+            pytest
+            pytest-xdist
+            pytorch-bin
+            safetensors
+
+            # for LLaMA
+            sentencepiece
+
+            # package
+            tinygrad
+          ];
+
+        shellHook = ''
+          export __NIX_PS1__="tinygrad";
+          source "${P.bash-completion}/etc/profile.d/bash_completion.sh";
+          source "${P.git}/share/bash-completion/completions/git";
+          SCRIPT_DIR="${B.toString ./.}"
+          export PYTHONPATH="$SCRIPT_DIR:$PYTHONPATH"
+        '';
+
+      in {
+        packages = rec {
+          inherit tinygrad;
+          default = tinygrad;
+        };
+
+        devShells = rec {
+          cpu = P.mkShell rec {
+            name = tinygrad.pname;
+            python = P."${pythonPackageStr}".withPackages
+              (pypkgs: (commonPythonPackages pypkgs));
+            buildInputs = commonBuildInputs ++ [ python ];
+            inherit shellHook;
+          };
+
+          cuda = P.mkShell rec {
+            name = tinygrad.pname;
+            cudaVersion = "11_6";
+            cudaPackagesStr = "cudaPackages_${cudaVersion}";
+            python = P."${pythonPackageStr}".withPackages (pypkgs:
+              ((commonPythonPackages pypkgs) ++ (with pypkgs; [ pycuda ])));
+            buildInputs = commonBuildInputs ++ [
+              python
+              P."${cudaPackagesStr}".cudatoolkit
+              P."${cudaPackagesStr}".cuda_nvcc
+            ];
+            inherit shellHook;
+          };
+
+          llvm = P.mkShell rec {
+            name = tinygrad.pname;
+            python = P."${pythonPackageStr}".withPackages (pypkgs:
+              ((commonPythonPackages pypkgs) ++ (with pypkgs; [ llvmlite ])));
+            buildInputs = commonBuildInputs
+              ++ [ python P.llvmPackages_13.clang ];
+            inherit shellHook;
+          };
+
+          triton = P.mkShell rec {
+            name = tinygrad.pname;
+            python = P."${pythonPackageStr}".withPackages
+              (pypkgs: (commonPythonPackages pypkgs));
+            buildInputs = commonBuildInputs ++ [ python P.triton ];
+            inherit shellHook;
+          };
+
+          default = cpu;
+        };
+      });
+}
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ datasets/open-images-v6-mlperf
 datasets/kits/
 datasets/audio*
 venv
+result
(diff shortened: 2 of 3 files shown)


wozeparrot:
I do use NixOS, and I would appreciate this. But, keeping with the spirit of tinygrad, this is a bit large for what it does, it shouldn't need flake-utils-plus, and the tinygrad derivation shouldn't need that many buildInputs.

b7r6:
@wozeparrot Sounds good, I'll strip out the `FLP` usage and some of the "convenience" dependencies (`matplotlib`, etc.).

Is that roughly what you had in mind or something more?

wozeparrot:
Yea, thats about it, the convenience dependencies shouldn't be in the derivation, infact pycuda shouldn't be in the derivation either, because tinygrad doesn't search for it during install, pycuda can just be in the devshell.

Maybe split the devshell as well to one with cuda and one without so that people without cuda don't need to pull in the huge cuda dependency.

b7r6:
Hi @wozeparrot,

I've updated the PR in an attempt to address your feedback. The main changes are:

 - remove `FLP` dependency
 - strip all "extras" out of the `tinygrad` derivation proper, it's almost the same as what you linked above
 - done some `PYTHONPATH` manipulation in the `shellHook` to give semantics like `pip install -e .`
 - added separate dev shells for: `CPU`, `CUDA`, `LLVM`, and `TRITON` (`GPU` seems to work on all of them)
 - cut some baggage like `metadata` license and stuff

Let me know if this is what you had in mind? Sorry this was a little slow to turn around, I'm multitasking quite a bit today.

flake.nix
```diff
@@ -0,0 +1,123 @@
...
+        B = builtins;
+        L = P.lib;
+
+        pythonVersion = "310";
+        pythonPackageStr = "python${pythonVersion}";
+        pythonNixpkgs = P."${pythonPackageStr}Packages";
+
+        tinygrad = pythonNixpkgs.buildPythonPackage {
+          pname = "tinygrad";
+          version = "v.0.6.0";
```
wozeparrot, line 30: Is the `.` here between `v` and `0` intentional?

flake.nix
```diff
@@ -0,0 +1,123 @@
...
+            safetensors
+            tinygrad
+          ];
+
+        shellHook = ''
+          export __NIX_PS1__="tinygrad";
+          source "${P.bash-completion}/etc/profile.d/bash_completion.sh";
+          source "${P.git}/share/bash-completion/completions/git";
+          SCRIPT_DIR="${B.toString ./.}"
+          export PYTHONPATH="$SCRIPT_DIR/src:$PYTHONPATH"
```
wozeparrot, line 71: We should avoid having extra things in shellHook and can just leave it up to the user considering stuff like direnv. The PYTHONPATH should be fine to leave tho. I think the PYTHONPATH is wrong tho? shouldn't it just be `builtins.toString ./.`?

flake.nix
```diff
@@ -0,0 +1,123 @@
...
+            onnx
+            onnxruntime
+            opencv4
+            pylint
+            pytest
+            pytest-xdist
+            python-lsp-server
+            pytorch-bin
+            pyyaml
+            safetensors
```
wozeparrot, line 62: We shouldn't need safetensors here cuz tinygrad supports loading safetensors natively.

flake.nix
```diff
@@ -0,0 +1,123 @@
...
+            mypy
+            onnx
+            onnxruntime
+            opencv4
+            pylint
+            pytest
+            pytest-xdist
+            python-lsp-server
+            pytorch-bin
+            pyyaml
```
wozeparrot, line 61: duplicated pyyaml import? should be propagated by tinygrad

flake.nix
```diff
@@ -0,0 +1,123 @@
...
+          with pypkgs; [
+            flake8
+            mypy
+            onnx
+            onnxruntime
+            opencv4
+            pylint
+            pytest
+            pytest-xdist
+            python-lsp-server
```
wozeparrot, line 59: We should avoid pulling in lsp stuff in the devshell cuz peoples dev environments/editors may be different and may use different lsp implementations.

flake.nix
b7r6: Nope just a typo: good catch, I'l fix it.

flake.nix
b7r6: I'll do what you want on this, I personally think that it's brittle to completely override environment variables that Nix is known to mess around with in subtle, implementation specific ways. It's generally lower risk to throw an extra entry on a `PATH` style envvar than to blow it away, but I think your proposal currently works.

LMK what you want done here.

flake.nix
b7r6: I found that some of the more important examples don't work under any runtime without some imports like e.g. `safetensors` and `sentencepiece`. There's no objectively right answer here, but I'm biased towards people can `nix develop` and immediately run the marquee models that `tinygrad` supports, e.g. `LLaMA`.

Again, I will defer to your judgement here.

flake.nix
b7r6: #willfix.

flake.nix
b7r6: Also a little subtle. I have `python-lsp-server` in my home manager, but my home manager python is `311` which can't work for `tinygrad` at the moment because `torch` deps on `tensorboard` which breaks on `311`.

I'll defer to your judgement here, but `python-lsp-server` seems in 2023 to be a comparably core dependency to e.g. `pytest` for people who still use like emacs and vim.

LMK how you want this resolved.

geohot:
I'm not sure this belongs in tinygrad. tinygrad is supported in Ubuntu, Mac M1, and sort of Windows. I don't understand the advantage of this, can Nix not pip install?

whazor:
Related, I also used Nix but for installing all the ROCm/LLVM stuff. I got Tinygrad working on AMD in minutes thanks to Nix.

flake.nix
wozeparrot: `tensorboard` should be fixed on 3.11 with https://github.com/NixOS/nixpkgs/pull/227957. I myself do use neovim as my editor but I use pyright as a language server, so I'm not sure how `python-lsp-server` is a dependency for `pytest`. Is there a reason I'm not getting?

flake.nix
wozeparrot: Oh yea I see there is a test that depends on safetensors, ok yea safetensors is fine.

flake.nix
wozeparrot: for the `PYTHONPATH` I meant like the `/src` at the end, shouldn't it just be the repo root?

wozeparrot:
> I'm not sure this belongs in tinygrad. tinygrad is supported in Ubuntu, Mac M1, and sort of Windows. I don't understand the advantage of this, can Nix not pip install?

NixOS doesn't support pip install because that would modify world state which should be immutable/conflicts with non-fhs stuff. This also helps with people who run `nix` on other operating systems like MacOS where it's sometimes used to replace homebrew.

b7r6:
@wozeparrot I think this addresses most of the things you raised. I pulled both `sentenancepiece` and `python-lsp-server` because it sounds like there isn't a consensus that those things should be in by default.

My personal opinion is that you really want `sentencepiece` because you want someone to be able to hit `nix develop .#triton` (or whatever) and immediately hit:

Screenshot 2023-06-26 at 3 00 54 PM

It's scary impressive that a few thousand lines of Python can run like LLaMA and Stable Diffusion U-Net stuff, and I'd want to make that just really easy to run, but we can modify it later.

Most importantly @geohot sounds skeptical that Nix support is currently in scope for `tinygrad`, so if he's not on board with this we should probably junk this PR and maybe revisit it later.

wozeparrot:
No I think `sentencepiece` should be fine to keep, you missed the duplicated `pyyaml`.

One thing we should look into is upstreaming tinygrad to nixpkgs.

b7r6:
`pyyaml` duplicate dep removed, `sentencepiece` back in. :) WDYT?

flake.nix
```diff
@@ -0,0 +1,127 @@
...
+            # test and development
+            flake8
+            mypy
+            onnx
+            onnxruntime
+            opencv4
+            pylint
+            pytest
+            pytest-xdist
+            pytorch-bin
```
wozeparrot, line 60: Does just `torch` not work here? or is this for CUDA support?

b7r6:
A good flake is generally 80% of the way to being in `nixpkgs`. One important thing to remember about PyTorch on Nix is that there's just no good way to get the right flags for your hardware because Hydra can't host `allowUnfree = true` stuff, so you either build the world or you take `pytorch-bin` which isn't tuned for your system.

b7r6:
^ This is one of many great selling points for `tinygrad`.

flake.nix
b7r6: Thoughts below v.

wozeparrot:
I've done some thought on this, while I do think it would be great to have flake support in the tinygrad repo, I don't think that this is the right implementation for it. It still feels way too complex for what it does and I feel it should be able to be a lot simpler.

PR #1027, opened 2023-06-22
Fix cuda dtype tests

b7r6:
**Summary**

The `CUDA` backend is failing a number of tests (at least on my hardware). This PR addresses two issues, albeit in a somewhat crude way:

 - The `half_prekernel` prelude in `ops_cuda.py` is sometimes non prepended in places where e.g. the `uchar` typedef is wanted which breaks kernel compilation.
 - The implicit cast in the generated `CUDA` code for the `dtypes.int8` -> `dtypes.uint8` seems to be triggering what looks like UB when applied to negative numbers. I've separately isolated that in a standalone `pycuda` mock and typing the out parameter as `char*` instead of `uchar*` seems to fix the issue.
 
 **Test Plan**
 
 I've run `test_dtype.py` with these changes and get all green on the `CUDA` backend with no obvious regressions elsewhere, though I suppose CI will be the judge of that...


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_cuda.py               5               1
total                                      6               2
lines added in the tinygrad folder: 4
```

wozeparrot:
This pulls in the flake.nix commit from #1026, can you drop that commit?

b7r6:
@wozeparrot Good catch, I believe it should be dropped now.

tinygrad/runtime/ops_cuda.py
```diff
@@ -45,13 +45,17 @@ class CUDACodegen(CStyleCodegen):
...
+    # `dtypes.int8` -> `dtypes.uint8` cast with negative inputs. for now stripping
+    # the `unsigned` off of `uchar` seems the lowest impact fix...
     half_prekernel = """
       #include <cuda_fp16.h>
       struct __align__(8) half4 {
         half2 x, y;
         __device__ __forceinline__ explicit operator float4() const {return make_float4(__half2float(x.x), __half2float(x.y), __half2float(y.x), __half2float(y.y)); }
       };
-      typedef unsigned char uchar;
+      typedef char uchar;
```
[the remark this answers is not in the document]
b7r6: The original test that demonstrated the issue is `test_int8_to_uint8_negative`, which I suppose is suggestive but not conclusive. I had wired up a standalone mock (https://gist.github.com/b7r6/6cb5aadfd06fdfc366e791f32a7cdb58), but you raise an excellent point that this belongs in the suite, so I've updated the PR to include `test_uint8_to_uint8_gt_signed`.

Does that seem adequate or is more coverage indicated? The last time I worked on this stuff was in the Caffe2 / early PyTorch days and we used `hypothesis` to sweep these kinds of primitives, but that seems a bit heavy for the `tinygrad` philosophy?

cloud11665:
This was mostly fixed in https://github.com/geohot/tinygrad/commit/2407690d821cfbb1747d5bf8088a4af3e5ac0769. As for the cast itself, we'd have to check if the instruction has the `.sat` modifier (`cvt.sat.s8.u8`)

b7r6:
@cloud11665 I've rebased this on your (much better) fix and it still seems to fix the one remaining failed test on my `CUDA` 11.6 gear. I'm a brand-new contributor so I don't know if this is "expedient hack" or "go back and do it properly" and could use any guidance on that. Happy to abandon and take a different approach. WDYT?

Sidebar: your linked commit indicates that you understand the `CUDA` runtime *way* better than I do, my NVIDIA is *really* rusty. I had somewhat arbitrarily selected the `CUDA` runtime as a place to start contributing, mostly because my GPU machine has a 3090 in it. Would you mind if I dropped you the occasional question in Discord about how best to add value there?

cloud11665:
@b7r6 I think this is a great learning experience, and I'd be happy to answer any questions about cuda / tinygrad on the discord!

tinygrad/runtime/ops_cuda.py
```diff
@@ -69,7 +69,10 @@ def __call__(self, global_size, local_size, *args, wait=False):
 
 class CUDACodegen(CStyleCodegen):
   lang = CStyleLanguage(
-    kernel_prefix = "typedef unsigned char uchar;\ntypedef unsigned int uint;\ntypedef unsigned long ulong;\n__global__", smem_prefix = "__shared__ ", barrier = "__syncthreads();", float4 = "make_float4",
+    # TODO(b7r6): `typedef char uchar` is a rather ugly hack to deal with (IIUC) saturation semantics
+    # on CUDA, roughly writing negative `float`s into `unsigned char*` seems to yield zero,
+    # and a more elegant fix probably requires a bit of refactoring...
+    kernel_prefix = "typedef char uchar;\ntypedef unsigned int uint;\ntypedef unsigned long ulong;\n__global__", smem_prefix = "__shared__ ", barrier = "__syncthreads();", float4 = "make_float4",
```
geohot, line 75: I don't like that it's called `uchar` at all, it should be `unsigned char`

And I"m definitely not merging something that typedefs `unsigned char` as `char`

b7r6: @geohot Absolutely, this diff started iffy and got worse post rebase on https://github.com/geohot/tinygrad/commit/2407690d821cfbb1747d5bf8088a4af3e5ac0769.

I'll clean this up, what's your preference for how to handle this in an organized and idiomatic way?

geohot: Read the TODO, I think there's an obvious answer.

You can't write a float into an unsigned char, you have to CAST it, and you have to make CAST behavior as expected,

geohot:
Closing, the right solution is proper casting I think

PR #1033, opened 2023-06-23
Update sz.py

vaibhavb02:
1. The TOKEN_WHITELIST constant was missing the token.COMMENT type.
2. The count_tokens() function was not properly closing the tokenize.open() file handle.
3. The tabulate() function was not being called with the floatfmt argument.


geohot:
This is more lines and I don't see the benefit.

PR #1035, opened 2023-06-23
Docker CI Test Fix

bdkiran:
Added a fix for the docker CI test. Updated the dockerfile by explicitly downloading python3.8 as that seems like preferred python version of tinygrad. 

Locally Tested:
```
burakkiran@Buraks-MacBook-Pro tinygrad % docker run --rm tinygrad /usr/bin/env python3 -c "from tinygrad.tensor import Tensor; print(Tensor.eye(3).numpy())"
[[1. 0. 0.]
 [0. 1. 0.]
 [0. 0. 1.]]
burakkiran@Buraks-MacBook-Pro tinygrad % git status
```

Tested and ran in github actions:
https://github.com/bdkiran/tinygrad/actions/runs/5360270277/jobs/9724929952


geohot:
So the problem is that this only works on master. If you can make it work on branches, I'll merge

huge3006:
I'm impressed with how frequently you update this repo. Keep up the good work!

bdkiran:
Okay I updated the Dockerfile to work with the localfiles instead. It will only copy in src code to build tinygrad and nothing else. The dockefile may seem a little verbose but I think that is a good thing for some people to be able to follow along.

bdkiran:
Again, test runs successfully locally:
```
burakkiran@Buraks-MacBook-Pro tinygrad % docker run --rm tinygrad /usr/bin/env python3 -c "from tinygrad.tensor import Tensor; print(Tensor.eye(3).numpy())"

[[1. 0. 0.]
 [0. 1. 0.]
 [0. 0. 1.]]
 
 ```

And successfully in github actions:
https://github.com/bdkiran/tinygrad/actions/runs/5370725918/jobs/9743014601

geohot:
Merged the other docker change, but thanks for reopening this conversation!

bdkiran:
Understood. Just want to make a note here. Using relative paths in dockerfiles is not recommended and can lead to inconsistent behavior. In case this flares up in the future refer to the dockerfile in this commit.

PR #1037, opened 2023-06-24
[WIP] Winograd Convolution

chaosagent:
WIP. Code/progress is not interesting at the moment. Posting just for reference.

Approach:
- _pool 4x4 with stride=2*stride
- permute kernel HW to the first 2 dims. 
- winograd by indexing into the first 2 dims
- permute kernel HW back to the bottom
- hope that the movement ops get pushed upwards (in the generated code) so that we have a sane winograd kernel

Old status:
- passes a simple test case
- horrifyingly slow
- does not actually emit the winograd convolution
  - m2 and m3 and used in sums for r1 and r2, but they get merged with the add in the Tensor.stack at the end, losing their LazyBuffer and preventing its reuse in the two expressions.
  - perhaps we need expression dedup in the AST processor, or some way to reference the same op as src for two separate ops.
  - adding something to m2.lazydata.children and m3.lazydata.children makes it emit winograd, but it uses a lot of kernels and is all around bad.

A quick summary of current progress: I have an approach that will get this to work. It may not be the cleanest solution, but I think it is rather clean. A design discussion may be forthcoming, there are alternatives that can be considered. There are four parts, all orthogonal to each other:

- [ ] dedup kernel arguments with the same underlying RawBuffer (#1141)
- [ ] dedup repeated loads and arithmetic
- [ ] upcast masked dimensions
- [ ] implement Winograd using the permute slice permute pattern as described above.

The current implementation has all of these working to various extents; but some of them are implemented with extremely hacky code (especially upcast masked dimensions).

There are also some misc tasks:
- [ ] optimize precomputation of g
- [ ] maximize kernel fusion
- [ ] integrate with tensor cores
- [ ] definitely more


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                          17               0
tinygrad/tensor.py                        65              16
total                                     82              16
lines added in the tinygrad folder: 66
```

geohot:
Good luck! I think for the bounty we have to be matching PyTorch in speed.

You can use the tensor_cores branch to get a fast GEMM.

chaosagent:
An update, for reference. The implementation is still just experimental. If anyone has any thoughts, please let me know!

the original approach of indexing into the 4x4 HW pools seems fine. We want our kernel to have the Winograd multiplications unrolled, and we can do this by upcasting the 2x2output dimensions.

We detect the 2x2 output dimensions by checking for small dimensions with masks, which indicates a pad, stack, etc. This is justified since operations on masked out portions of a small dimension are nops, which can be unrolled and optimized if we upcast that dimension.

A rough sketch of what this looks like in practice: Tensor.stack([a, b]) is implemented as [a, 0] + [0, b]. the pad gets pushed all the way down the AST, so if a and b are functions, we have a([x, 0]) + b([0, x]). for even indices (in this simple example), we just want to calculate a, but the inputs needed for b are simply masked out, and b is calculated as normal (and the result will always equal 0 for even indices). This is a waste; luckily after unrolling the stack dimension, the shape tracker can tell that for even indices, the mask for the b inputs is just False.

Further, small sums are upcasted into a loop within the kernel, allowing us to use tensor cores for large cin in the future.

This takes care of generating the winograd convolution as a localized kernel. However, we still need to deal with the problem of not recalculating identical subexpressions. Right now I am attempting to dedup these calculations in the linearize step. I think this is sound in theory (just a dedup compiler pass). Some refactoring / design needs to be done to get input buffers with the same underlying rawbuffer (but different shapetracker) to be loaded as the same variable. Also, we might want to get rid of expressions of the form `0 ? some_val : 0.0f` completely. This is more or less standard compiler pass work; how much of this belongs in tinygrad? Perhaps if we get the input dedup to work nvcc will take care of everything for us. 

I don't know if that is the right design. The alternative is to do this at the AST level, making it into an abstract syntax dag. This seems more conceptually complicated.

I disabled the float4 stuff temporarily, since I did not want to deal with it. Float4 behaves like another local dimension. A little thought might (probably not) be needed here to get this to interact well with the mask simplification described above.

geohot:
Is there still anything here?

chaosagent:
Nope!

PR #1039, opened 2023-06-24
Docker fix

jla524:
credits to [bdkiran](https://github.com/bdkiran)

changes made:
- enable docker test in CI
- run full set of tests in docker
  - skip CLANG in multi-device test, I think we are missing dependencies
- copy current folder into docker container
- install python dependencies

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -224,7 +224,7 @@ jobs:
     - name: Build Docker
       run: docker build -t tinygrad -f test/Dockerfile .
     - name: Test Docker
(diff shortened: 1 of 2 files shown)


geohot:
So now I'm reconsidering the purpose of this test.

Originally it was designed to confirm that a minimal install of tinygrad works. I don't think it should be running the main test suite, that's already covered.

jla524:
> So now I'm reconsidering the purpose of this test.
> 
> Originally it was designed to confirm that a minimal install of tinygrad works. I don't think it should be running the main test suite, that's already covered by the rest of CI.
> 
> I'm not sure it really makes sense. Can you think of anything this would catch?

Hmm yeah, everything should behave the same in docker, so it doesn't make sense to run the full test.

I've reverted that change.

geohot:
Will merge, the test is very fast.

PR #1040, opened 2023-06-24
feat: :sparkles: Sub-Pixel Convolutional

aabbdev:
Learns a set of upscaling filters to transform low-resolution feature maps into high-resolution outputs by rearranging sub-pixels spatially

### DIFF
--- a/test/test_nn.py
+++ b/test/test_nn.py
@@ -5,11 +5,10 @@ from extra.utils import WINDOWS
 from tinygrad.helpers import getenv
 from tinygrad.jit import TinyJit
 from tinygrad.tensor import Tensor, Device
-from tinygrad.nn import BatchNorm2d, Conv2d, ConvTranspose2d, Linear, GroupNorm, LayerNorm, LayerNorm2d, Embedding, InstanceNorm
+from tinygrad.nn import BatchNorm2d, Conv2d, ConvTranspose2d, Linear, GroupNorm, LayerNorm, LayerNorm2d, Embedding, InstanceNorm, PixelShuffle, PixelUnshuffle
 import torch
 
 class TestNN(unittest.TestCase):
-
   def test_batchnorm2d(self, training=False):
     szs = [4, 8, 16, 32]
     for sz in szs:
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/image.py                      18               0
tinygrad/tensor.py                         3               1
total                                     21               1
```

jla524:
Can you add some tests?

geohot:
What models is this used in? I've never seen it before.

Also, like @jla524 said, all added features need tests. And I'm not sure this belongs in main tinygrad, it's +28 lines.

aabbdev:
Hello
I didn't have time this morning lol
This layer is used in the majority of SOTA super-resolution models (30% according to statistics). It was introduced by ESPCN.
I haven't finished my PR, I'll add more tests, and I wanted to add a real model example.

docs:
https://paperswithcode.com/method/pixelshuffle
https://medium.com/@hirotoschwert/introduction-to-deep-super-resolution-c052d84ce8cf

wozeparrot:
At the very least, I don't think this belongs in `nn`.

aabbdev:
as you wish, that's how it is in pytorch. I thought I heard you say you wanted to keep a similar api

aabbdev:
I missed an rebase, I made a mess I will cleanup

aabbdev:
I added tests

jla524:
> as you wish, that's how it is in pytorch. I thought I heard you say you wanted to keep a similar api
> 
> https://pytorch.org/docs/stable/generated/torch.nn.PixelShuffle.html

We want to keep the interface similar. At the same time, we really want to keep tinygrad minimal.

If these layers are only used in one example, then I suggest to move the code to the same file as the example. So everything goes in `examples/espcn.py` or `models/espcn.py`.

aabbdev:
I totally understand your point of view,
you don't want to implement things that are rarely used or "fashionable" but PixelShuffle is an essential building block for super-resolution models.
From my point of view, your logic of minimalism should be on the interface -> backend side only,
PixelShuffle reuses existing functions it doesn't need to modify backends, or add a new ops.

user engineers' point of view :
Ease of model creation

integrator engineers' viewpoint :
Make it easier to implement their accelerator/platform inside the library to run the models.

In this logic, I'm only improving the user experience without touching the integration.

Besides, letting users implement PixelShuffle themselves, however simple, can lead to bugs. For a long time, PixelShuffle didn't work properly in Pytorch, until it caused problems with the ONNX runtime.

Capture d’écran 2023-06-25 à 11 18 40
Capture d’écran 2023-06-25 à 11 18 14

geohot:
Closing, citing @jla524's last post. If operators are only used in one or two models, they belong in the model itself, not core tinygrad.

PR #1045, opened 2023-06-25
Reduce tensor dot line count and fixed 1d tensor dot

francescocastelli:
Reduced line counts for tensor.dot by 1 while also supporting 1d tensors as input (previously broken). 
Now should work in all the following cases: 

- 1d @ 1d
- 1d @ Nd
- Nd @ 1d
- Nd @ Nd

Also added the corresponding tests. 

Eventually we could reduce duplicated code like: 
```
  def dot(self, w:Tensor) -> Tensor:
    if (n1:=len(self.shape))*(n2:=len(w.shape)) == 0: raise RuntimeError(f"both arguments to matmul need to be at least 1D, but they are {n1}D and {n2}D")
    x = self.reshape(*self.shape[0:-1], *(m1:=[1]*min(n1-1, n2-1, 1)), self.shape[-1])
    w = w.reshape(*w.shape[0:-2], *m1, *w.shape[-(m2:=min(n2, 2)):]).transpose(-1, -m2)
    return (x*w).sum(-1)
```

Let me know!

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -279,12 +279,13 @@ class TestOps(unittest.TestCase):
       return x*torch.tanh(torch.nn.functional.softplus(x))
     helper_test_op([(45,65)], _mish_pytorch, Tensor.mish, atol=1e-4)
     helper_test_op([()], _mish_pytorch, Tensor.mish, atol=1e-4)
-  def test_dot(self):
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         3               4
```

geohot:
Now this is a winning PR. More functionality and less lines!

PR #1047, opened 2023-06-26
Dedup params in `Optimizer`

Ameobea:
Passing the same tensor multiple times in the set of learnable params passed to optimizers can result in models completely failing to learn, but no errors are produced.  De-duping the passed tensors avoids the problem.

I ran into this when porting some of my TensorFlow code to work with Tinygrad.  Had a maddening few hours before I finally tracked down the issue.

I also noticed that the `Optimizer.realize()`'s `extra` arg doesn't seem to ever be set when calling it from within the tinygrad project.  Makes sense if it's meant to be a public API, or maybe I missed something, but if not it might be possible to remove that and get rid of some code.

### DIFF
--- a/tinygrad/nn/optim.py
+++ b/tinygrad/nn/optim.py
@@ -1,5 +1,6 @@
 # sorted in order of increasing complexity
-from typing import List, Set
+from typing import List
+from tinygrad.helpers import dedup
 from tinygrad.tensor import Tensor
 
 class Optimizer:
@@ -8,9 +9,8 @@ class Optimizer:
     for x in params:
       if x.requires_grad is None: x.requires_grad = True
 
-    deduped_params: Set[Tensor] = set(params)
-    self.params: List[Tensor] = [x for x in deduped_params if x.requires_grad]
-    self.buffers: List[Tensor] = [x for x in deduped_params if not x.requires_grad]   # buffers are still realized
+    self.params: List[Tensor] = dedup([x for x in params if x.requires_grad])
+    self.buffers: List[Tensor] = dedup([x for x in params if not x.requires_grad])   # buffers are still realized
 
   def zero_grad(self):
     for param in self.params: param.grad = None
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/optim.py                       2               1
```

wozeparrot:
`extra` is set [here](https://github.com/Ameobea/tinygrad/blob/6017dfc850968e69ded3f849d537fec867841121/tinygrad/nn/optim.py#L69) and [here](https://github.com/Ameobea/tinygrad/blob/6017dfc850968e69ded3f849d537fec867841121/tinygrad/nn/optim.py#L39)

Ameobea:
Got it - indeed missed those, thx!

tinygrad/nn/optim.py
```diff
@@ -8,8 +8,9 @@ def __init__(self, params: List[Tensor]):
     for x in params:
       if x.requires_grad is None: x.requires_grad = True
 
-    self.params: List[Tensor] = [x for x in params if x.requires_grad]
-    self.buffers: List[Tensor] = [x for x in params if not x.requires_grad]   # buffers are still realized
+    deduped_params: Set[Tensor] = set(params)
```
geohot, line 11: Use `from tinygrad.helpers import dedup` instead. It preserves order and is `O(n)`

geohot:
Also, a test never hurts, preferably a small complete integration test that exposes how this problem can occur in practice.

Ameobea:
alright, I'll see if I can get a test for this together

Ameobea:
Alright, I added a very basic test that just verifies that the optimizer behaves the same with or without duped params.

I tried to find an example where training completely fails like I was seeing, but wasn't able to get it.  The main thing it seems to be doing is just acting as if the learning rate is higher for the duped params.

The case where I ran into issues was an RNN with regularizers and a custom activation function, though, so it's possible things become worse in some more complex situations.

geohot:
LGTM!

PR #1048, opened 2023-06-26
HIP CI support and some unit test fixes for HIP GPU/CPU backend

nanamiwang:
1. HIP CI support by adding HIP CPU backend, see test.yml modifications.
How to run HIP CPU backend locally:
- Install HIP-CPU by following these instructions: [https://github.com/ROCm-Developer-Tools/HIP-CPU/blob/master/docs/gcc_linux.md](https://github.com/ROCm-Developer-Tools/HIP-CPU/blob/master/docs/gcc_linux.md)
- Run tinygrad code using HIP-CPU:
```
HIP=1 HIP_CPU=1 python3 test/test_dtype.py
```
2. Since HIP-CPU is a header only library, [hip_cpu_wrapper.py](https://github.com/geohot/tinygrad/compare/master...nanamiwang:tinygrad:hip_ci?expand=1#diff-39d50bf559ac666b3e21dbf5fb7123756bfbdc3187713bb1f05befb03a482d8a) was added for wrapping HIP-CPU APIs into a shared library.
3. For HIP CPU backend, kernel and kernel launcher code have to be compiled together into one shared library, so add 2 properties (need_kernel_launcher&build_kernel_launcher) to CStyleLanguage.
HIP CPU kernels will look like this:
```
extern "C" __global__
 void E_2925(float* data0, const float* data1, const float* data2) {
{ int gidx0 = blockDim.x*blockIdx.x+threadIdx.x;  /* 2925 */
  float val1_0 = data1[gidx0];
  float val2_0 = data2[gidx0];
  float alu0 = (val1_0+val2_0);
  data0[gidx0] = alu0;
} /* global */
}
extern "C" void launch_and_wait_E_2925(
  std::uint32_t grid_dim_x,
  std::uint32_t grid_dim_y,
  std::uint32_t grid_dim_z,
  std::uint32_t block_dim_x,
  std::uint32_t block_dim_y,
  std::uint32_t block_dim_z,
  std::uint32_t shared_mem_bytes,
  hipStream_t stream,
float* data0, const float* data1, const float* data2
) {
  hipLaunchKernelGGL(E_2925, dim3(grid_dim_x, grid_dim_y, grid_dim_z), dim3(block_dim_x, block_dim_y, block_dim_z), shared_mem_bytes, stream,
data0, data1, data2);
  hipStreamSynchronize(stream);
}
```
4. Add half4, float4 pow/log/max supports to HIP GPU/CPU backend, fix some test_ops.py/test_dtype.py test failures.

#### Remained issues
HIP CPU backend is slow, so HIP CI test will take more than 5 hours and get cancelled on github actions.  Even run the tests locally, it is still very slow and memory consumption is very high. [CI log](https://github.com/nanamiwang/tinygrad/actions/runs/5373605398/jobs/9748138423)

### DIFF
--- a/tinygrad/codegen/cstyle.py
+++ b/tinygrad/codegen/cstyle.py
@@ -24,6 +24,8 @@ class CStyleLanguage(NamedTuple):
   float4: Optional[str] = None
   half_prekernel: Optional[str] = None
   double_prekernel: Optional[str] = None
+  need_kernel_launcher: bool = False
+  build_kernel_launcher: Optional[Callable] = None
   uses_vload: bool = False
 
 def to_image_idx(base_shape:Tuple[int, ...], idxy:Node, valid:Node, validhacks=False) -> Tuple[Node, Node]:
@@ -174,6 +176,8 @@ def uops_to_cstyle(uops:List[UOp], bufs:List[Union[LocalBuffer,LazyBuffer]], lan
 
   if lang.half_prekernel and any(x.dtype == dtypes.float16 for x in bufs): prg = ''.join([f"{lang.half_prekernel}", "\n", prg])
   if lang.double_prekernel and any(x.dtype == dtypes.float64 for x in bufs): prg = ''.join([f"{lang.double_prekernel}", "\n", prg])
+  if lang.need_kernel_launcher and lang.build_kernel_launcher:
+    prg = ''.join([prg, '\n'] + lang.build_kernel_launcher(bufnames, buftypes))
   return prg, global_size, local_size
 
 class CStyleCodegen(Linearizer):
--- a/tinygrad/runtime/ops_hip.py
+++ b/tinygrad/runtime/ops_hip.py
@@ -121,6 +121,16 @@ class HIPCodegen(CStyleCodegen):
   lang = CStyleLanguage(
     kernel_prefix = r"""
 #define INFINITY (__builtin_inff())
+__device__ float4 pow(float x, float4 y) {
+  return float4(pow(x, y.x), pow(x, y.y), pow(x, y.z), pow(x, y.w));
+}
+
+__device__ float4 pow(float4 x, float4 y) {
+  return float4(pow(x.x, y.x), pow(x.y, y.y), pow(x.z, y.z), pow(x.w, y.w));
+}
+__device__ float4 log2(float4 x) {
+  return float4(log2(x.x), log2(x.y), log2(x.z), log2(x.w));
+}
 extern "C" __global__""",
     smem_prefix = "__shared__ ", barrier = "__syncthreads();", float4 = "make_float4",
     half_prekernel = r"""
--- a/tinygrad/runtime/ops_hip.py
+++ b/tinygrad/runtime/ops_hip.py
@@ -131,15 +131,19 @@ class HIPCodegenCPU(CStyleCodegen):
 typedef unsigned char uchar;
 using half_float::half;
 MAKE_VECTOR_TYPE(half, half)
+
 inline float max(float x, float y) {
   return fmax(x, y);
 }
+
 inline float4 max(float4 x, float4 y) {
   return float4(fmax(x.x, y.x), fmax(x.y, y.y), fmax(x.z, y.z), fmax(x.w, y.w));
 }
-inline float4 pow(float4 x, float y) {
-  return float4(pow(x.x, y), pow(x.y, y), pow(x.z, y), pow(x.w, y));
+
+inline float4 pow(float x, float4 y) {
+  return float4(pow(x, y.x), pow(x, y.y), pow(x, y.z), pow(x, y.w));
 }
+
 inline float4 log2(float4 x) {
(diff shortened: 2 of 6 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/cstyle.py                 4               0
tinygrad/runtime/ops_hip.py              148               7
total                                    152               7
lines added in the tinygrad folder: 145
```

geohot:
So this is enough code I think it should be not in `ops_hip.py`

nanamiwang:
> Bounty locked.
> 
> So this is enough code I think it should be not in `ops_hip.py`
> 
> It should almost be a separate package. Like `pip install pyhip` or `pyrocm`

What about moving HIP CPU stuffs to ops_hip_cpu.py as a new backend?
For the new pip package, I think pyrocm is a better name, since it is for wrapping rocm c++ libraries (HIP GPU, HIP CPU, and might also [rocWMMA](https://github.com/ROCmSoftwarePlatform/rocWMMA) in the future)?

geohot:
Ideally it's refactored so it shares as many codepaths in tinygrad as possible. Currently, it's a different Buffer, Codegen, and Runtime!

nanamiwang:
> Ideally it's refactored so it shares as many codepaths in tinygrad as possible. Currently, it's a different Buffer, Codegen, and Runtime!

I did some refactorings so that HIP CPU and HIP GPU will share the same Buffer class and share some kernel prefixes. Codegen and Runtime can't be shared though.
Moving HIP wrappers  to seperate pip package is easy, should we move the code in ops_hip.py to seperate pip package as much as possible?

wozeparrot:
Just wanted an update on this. How's it coming along?

nanamiwang:
> Just wanted an update on this. How's it coming along?

@geohot @wozeparrot three questions:
1) Unit tests for HIP CPU backend are very slow currently, do we need to speed up it?
2) Currently Codegen & Program is still different for CPU and GPU backend, is it ok?
2) I am planning to create a pip package named pyrocm, move extra/hip_wrapper.py and extra/hip_cpu_wrapper.py to this package, also move some code in ops_hip.py (like build_kernel_launcher) to this package, keep ops_hip.py as short as possible, is it ok?

tinygrad/codegen/cstyle.py
```diff
@@ -24,6 +24,8 @@ class CStyleLanguage(NamedTuple):
   float4: Optional[str] = None
   half_prekernel: Optional[str] = None
   double_prekernel: Optional[str] = None
+  need_kernel_launcher: bool = False
+  build_kernel_launcher: Optional[Callable] = None
```
wozeparrot, line 28: This shouldn't be part of cstyle, it should be part of HIPProgramCPU

tinygrad/runtime/ops_hip.py
```diff
@@ -55,11 +62,120 @@ class PackageStruct(ctypes.Structure):
...
+""" + common_kernel_prefix,
+    smem_prefix = "__shared__ ", barrier = "__syncthreads();", float4 = "make_float4", uses_vload=True,
+    half_prekernel = r"""
+using half_float::half;
+MAKE_VECTOR_TYPE(half, half)
+""" + common_half_prekernel,
+    need_kernel_launcher = True,
+    build_kernel_launcher = build_kernel_launcher,
     gid = [f'blockDim.{chr(120+i)}*blockIdx.{chr(120+i)}+threadIdx.{chr(120+i)}' for i in range(3)],
     lid = [f'threadIdx.{chr(120+i)}' for i in range(3)])
```
wozeparrot, line 179: Really, the codegen should be the same between HIP and HIP_CPU

nanamiwang: They can't, since you could see that the required header files, macros and typedefs are different for CPU backend and GPU backend.

tinygrad/codegen/cstyle.py
nanamiwang: Let me explained more for this and for build_kernel_launcher:
For HIP GPU backend, we just call hiprtcCompileProgram to build the kernel  and call hipModuleLaunchKernel, it works well.
HIP CPU backend has no HIP RTC support, so we need to find another approaches.
I tried building the kernel into a dynamic library, export it as function and call this function using hipModuleLaunchKernel, it doesn't work, hipModuleLaunchKernel can't access the function since they are from different HIP CPU contexts.
build_kernel_launcher is for fixing this issue, it will append calls to hipLaunchKernelGGL/hipStreamSynchronize after the kernel, so that all the code are compiled together by HIPProgramCPU into one library, and a special function named launch_and_wait_KERNEL_NAME_PLACEHOLDER is exported for launching the kernel.
Can we fix this without modifying cstyle.py?

nanamiwang:
You can close this PR now. Will create a new one in case I can make a better solution for HIP CPU CI.

dc-dc-dc:
Tracking #1700

PR #1067, opened 2023-06-27
Onnx Simplifications

eichenroth:
Small improvements to the ONNX executor.

- typing
- understandable code
- less lines

No additional tests got fixed

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -5,7 +5,7 @@ import numpy as np
 from tinygrad.tensor import Tensor
 from tinygrad.helpers import prod, getenv, DEBUG, dtypes
 from typing import Dict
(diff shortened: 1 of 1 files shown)


geohot:
Is all this changed code well tested in CI?

eichenroth:
I see your point, but I think we have an exception here.

The only file that is touched is `onnx.py` - which is still WIP and very buggy in itself. I cannot prove that I did not introduce any bugs, however the number of failed onnx tests did not go up - therefore I think I did not make it worse.

Would you merge, if I fix some onnx tests with this PR?

geohot:
Even better, test `onnx.py`

eichenroth:
😄 that sounds reasonable

PR #1073, opened 2023-06-28
Over 90% on CIFAR with examples/hlb_cifar10.py 

kunwar31:
Haven't made any real changes yet, just wanted to have a correct baseline first

Changes made:
- Eval needs to be done on entire 10k eval set
- Fetch batches serially after shuffling (to avoid overlaps b/w batches) 

Main tricks:
- [x] LR Scheduling 
- [x] image augmentations (flip lr, mixup)
- [x] whitening (not needed to cross 90%)
- [x] Label smoothing (didnt help)
- [x] Most importantly, hyperparameter tuning


geohot:
Bounty locked, good luck!

To be eligible, It can't use more compute or time than what's in master, and needs to need line count to a relative minimum.

kunwar31:
With OneCycle LR Scheduling and weight decay, jump from 82.79 --> 87.94!

g1y5x3:
There's a [extra/lr_scheduler.py](https://github.com/geohot/tinygrad/blob/master/extra/lr_scheduler.py) class already implemented. Instead of writing a whole new optim.SGD class, I would just add the OneCycle LR there.

kunwar31:
@geohot I've just crossed 90%, though some tweaking of hyperparams could probably lead to higher acc

extra/lr_scheduler.py
```diff
@@ -61,3 +62,20 @@ def __init__(self, optimizer: Optimizer, T_max: int, eta_min=0):
 
   def get_lr(self) -> float:
     return self.eta_min + 0.5 * (self.eta_max - self.eta_min) * (1 + math.cos((self.epoch_counter/self.T_max) * math.pi))
+
+class OneCycleLR(LR_Scheduler):
+  def __init__(self, optimizer: Optimizer, max_lr: float, initial_div_factor: float, final_div_factor: float, total_steps: int, pct_start: float):
+    self.initial_lr = Tensor([max_lr / initial_div_factor]) # TODO: why is Tensor of lr over 2x faster?
```
geohot, line 68: Because it's not constant folding and recompiling

geohot:
Can this be done with a smaller diff? And is the whole thing not being jitted anymore?

geohot:
Yea, the whole training step should be in the JIT

kunwar31:
@geohot so its not that eval is slower, the previous eval was just on a single batch and I changed it to run on entire 10k data (which will be 20 batches on eval_bs=500). I tried adding JIT for eval step and it breaks it, need to figure out why

Regarding the train step not being fully JITted, the reason for that is the lr scheduler needs to update LR, and JITting that part does not change the LR correctly (idk why yet, need to understand how TinyJIT works prolly). But it still does take the same amount of time for a training step as the grad calculation is still JITTed

kunwar31:
`eval 9008/10000 90.08%,    0.41 val_loss STEP=2000`
BTW @geohot noticed that it says STEP=2000, if you are running with BS=256 and STEPS=2000, its not needed, 
default hparameters set in train_cifar perform better

kunwar31:
Also, I noticed that removing JIT completely doesnt have a drastic performance hit, on my 3060 I get ~1.4 TFLOPS with JIT (full JIT and current hacked JIT give the same perf) and ~1.2 TFLOPS without JIT

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/optim.py                       1               1
```

kunwar31:
@geohot should be ready now! added mixup to get a small boost in accuracy, because it didnt need many lines
Tried label smoothing as well but did not help 
Defaults to TTA=0
with TTA=1: 90.64%
without TTA=0: 90.33%

tinygrad/nn/optim.py
```diff
@@ -45,7 +45,7 @@ def Adam(params: List[Tensor], lr=0.001, b1=0.9, b2=0.999, eps=1e-8): return LAM
 class LAMB(Optimizer):
   def __init__(self, params: List[Tensor], lr=0.001, b1=0.9, b2=0.999, eps=1e-6, wd=0.0, adam=False):
     super().__init__(params)
-    self.lr, self.b1, self.b2, self.eps, self.wd, self.adam, self.t = lr, b1, b2, eps, wd, adam, Tensor([0], requires_grad=False).realize()
+    self.lr, self.b1, self.b2, self.eps, self.wd, self.adam, self.t = Tensor([lr]), b1, b2, eps, wd, adam, Tensor([0], requires_grad=False).realize()
```
wozeparrot, line 48: Should `requires_grad=False` here and above in SGD

extra/lr_scheduler.py
```diff
@@ -57,7 +58,25 @@ def __init__(self, optimizer: Optimizer, T_max: int, eta_min=0):
     super().__init__(optimizer)
     self.T_max = T_max
     self.eta_min = eta_min
-    self.eta_max = optimizer.lr
+    self.eta_max = optimizer.lr.numpy()[0]
 
   def get_lr(self) -> float:
```
wozeparrot, line 63: wrong return type here

examples/hlb_cifar10.py
```diff
@@ -24,12 +27,12 @@ def __init__(self, channels_in, channels_out, short, se=True):
...
   def __call__(self, x):
     x = self.conv[0](x).max_pool2d(2)
-    x = self.norm[0](x).relu()
+    x = self.norm[0](x).clip(1e-15, 1e15).gelu()
     if self.short: return x
     residual = x
-    mult = self.se2(self.se1(residual.mean((2,3))).relu()).sigmoid().reshape(x.shape[0], x.shape[1], 1, 1) if self.se else 1.0
-    x = self.norm[1](self.conv[1](x)).relu()
-    x = self.norm[2](self.conv[2](x) * mult).relu()
+    mult = self.se2(self.se1(residual.mean((2,3))).clip(1e-15, 1e15).gelu()).sigmoid().reshape(x.shape[0], x.shape[1], 1, 1) if self.se else 1.0
```
geohot, line 33: What's up with these weird clips?

tinygrad/nn/optim.py
```diff
@@ -24,7 +24,7 @@ def realize(self, extra=None):
 class SGD(Optimizer):
   def __init__(self, params: List[Tensor], lr=0.001, momentum=0, weight_decay=0.0, nesterov=False):
     super().__init__(params)
-    self.lr, self.momentum, self.wd, self.nesterov = lr, momentum, weight_decay, nesterov
+    self.lr, self.momentum, self.wd, self.nesterov = Tensor([lr], requires_grad=False), momentum, weight_decay, nesterov
```
geohot, line 27: Does `Tensor(lr, required_grad=False)` work? Like a 0-D tensor instead of a 1-D?

geohot:
Screenshot 2023-07-02 at 12 41 31 PM

The first 8 kernels on the train step are not jitted. Check with `DEBUG=2`

geohot:
Got `eval 8996/10000 89.96%,    0.43 val_loss STEP=1000` with `GPU=1 CL_PLATFORM=1 PYTHONPATH="." python3 examples/hlb_cifar10.py`

examples/hlb_cifar10.py
kunwar31: gelu causes NaNs if there are no clips, probably gradidnt clipping would be a better idea

geohot: Then fix it in gelu, not here

geohot:
I don't think gradient clipping is the right fix. Is that how torch is doing it?

kunwar31:
> I don't think gradient clipping is the right fix. Is that how torch is doing it?

Agree, after gradient clipping there are no NaNs but the training diverges. looking into gelu

kunwar31:
> Screenshot 2023-07-02 at 12 41 31 PM
> 
> The first 8 kernels on the train step are not jitted. Check with `DEBUG=2`

that was the mixup and flipLR, moved it to jit. all purple now

geohot:
Don't think we want grad clip. Also, carefully read all diffs in the GitHub interface, it would catch things like the divide by 1. In here, there's extra weird parens.

kunwar31:
yes I'll be cleaning it up once I figure out 90+ 
I'm just pushing code to pull in a cloud VM, def. not ideal

kunwar31:
without TTA:
```
eval 9062/10000 90.62%,    0.41 val_loss STEP=1000
``` 
with TTA:
```
eval 9109/10000 91.09%,    0.40 val_loss STEP=1000
```

kunwar31:
https://docs.google.com/spreadsheets/d/1mDzpl2yO0j4cabUgd8DINoMqbj4C8X5MAUhkWB-09i8/edit#gid=0
Just a log of hparams I tried
(for gelu I used older code which had gradient clipping)

geohot:
Congrats! You sent a payment request last time, want to do that again?

PR #1080, opened 2023-06-30
Improvement on CIFAR-10 hlb_cifar10_torch.py 89%

g1y5x3:
Since both torch and tinygrad implemented the same model and pipeline, I started with comparing to [https://github.com/tysam-code/hlb-CIFAR10/blob/main/main.py#L274](url) as suggested. 

Two major differences:
1. Whitening layer where the weight is frozen for better data distribution
2. LR scheduler (OneCycleLR) for more efficient training

By just adding the OneCycleLR, I was able to get as high as 89% without spending too much time tuning all the parameters inside OneCycleLR, I do believe once the whitening is applied, reaching 90% should be easily reachable since the space is much cleaner.

### DIFF
--- a/examples/hlb_cifar10_torch.py
+++ b/examples/hlb_cifar10_torch.py
@@ -77,6 +77,7 @@ def fetch_batch(X_train, Y_train, BS):
 
 def train_cifar():
   BS = getenv("BS", 512)
+  STEPS = getenv("STEPS", 10)
   if getenv("FAKEDATA"):
     N = 2048
     X_train = np.random.default_rng().standard_normal(size=(N, 3, 32, 32), dtype=np.float32)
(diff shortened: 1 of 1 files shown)


kunwar31:
@g1y5x3 currently eval is only done on a single batch of eval data, do you get the same score on entire 10k samples? (see https://github.com/geohot/tinygrad/pull/1073)

g1y5x3:
> 

Not yet but I think that's probably why I am seeing this variation of results in between different runs. And I don't want to make too much changes to the code base yet. But I am looking into the pytorch `Dataloader` class and try to make a minimal one.

g1y5x3:
As this was used just to compare with the tinygrad implementation which I am glad the OneCycleLR trick helped with #1073. Once the that got merged, there's probably no reason to keep this pytorch implementation anyway.

PR #1081, opened 2023-06-30
dtypes: add int16, priority change, refactor

ReinBentdal:
Did the following changes:
- removed `float` and `half` dtypes alias for `float32` and `float16` respectively
- added `int16` and `uint16` dtypes (short, ushort)
- changed dtype priorities such that no dtype priority is the same (thus order of ops wont effect resulting dtype)
- renamed `is_int` to `is_signed_int`. renamed `is_unsigned` to `is_unsigned_int`. Added `is_int` which encapsulates both signed and unsigned.
- removed `_half4` type since not in use
- added some tests for int16, moved tests for int64 from int8 class to its own int64 test class

### DIFF
--- a/tinygrad/codegen/cstyle.py
+++ b/tinygrad/codegen/cstyle.py
@@ -119,11 +119,11 @@ def uops_to_cstyle(uops:List[UOp], bufs:List[Union[LocalBuffer,LazyBuffer]], lan
     elif uop == UOps.LOAD and newvar is not None:
       # TODO: merge with CONST?
       if bufs[args.i] is not None and isinstance(bufs[args.i].realized, RawConst):
-        assert newvar.dtype == dtypes.float, "const can't be float4"
+        assert newvar.dtype == dtypes.float32, "const can't be float4"
         x = bufs[args.i].realized._buf
         if math.isnan(x): val = "NAN"
         elif math.isinf(x): val = ("-" if x < 0 else "") + "INFINITY"
-        else: val = f"{x}" +  ("f" if not dtypes.is_int(bufs[args.i].dtype) else "")
+        else: val = f"{x}" +  ("f" if not dtypes.is_signed_int(bufs[args.i].dtype) else "")
       elif isinstance(bufs[args.i].dtype, ImageDType):
         assert newvar.dtype == dtypes._float4, f"image must be float4 {newvar}"
         prekernel.add("const sampler_t smp = CLK_NORMALIZED_COORDS_FALSE | CLK_ADDRESS_CLAMP | CLK_FILTER_NEAREST;\n")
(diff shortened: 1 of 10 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/assembly.py               2               2
tinygrad/codegen/assembly_rdna.py          2               2
tinygrad/codegen/cstyle.py                 4               4
tinygrad/codegen/linearizer.py             3               3
tinygrad/codegen/llvmir.py                 4               4
tinygrad/helpers.py                       19              18
total                                     34              33
```

ReinBentdal:
to add support for a new type, theres some not-obvious places where it has to be added support for. I would like to propose also moving these mapping to inside Dtype. Something like:
```python
int16: Final[DType] = DType(3, 2, "short", np.int16, c=ctypes.c_int16, torch=torch.int16, llvm=ir.IntType(16))
uint16: Final[DType] = DType(3, 2, "ushort", np.uint16, c=ctypes.c_uint16, torch=None, llvm=None)
```
Also move from `helpers.py` to `dtype.py` or something

ReinBentdal:
> Tests failing

I'll look into it.
@geohot What do you think about my suggestion to move C, torch and LLVM type map to DType class?

tinygrad/codegen/llvmir.py
```diff
@@ -101,7 +101,7 @@ def uops_to_llvm_ir(uops:List[UOp], bufs:List[LazyBuffer]) -> str:
       element = lvars[vin[0]]
       if func_dtypes[0] != ir.FloatType():
         if dtypes.is_int(bufs[args.i].dtype):
```
ReinBentdal, line 103: This was the reason I changed syntax from `is_int` to `is_signed_int`. This used to be `is_int`, which only checked for signed type, which I believe to be a bug. Since I changed `is_int` to encapsulate both signed and unsigned type, this should now be correct. No idea about the implications

tinygrad/helpers.py
```diff
@@ -79,31 +79,32 @@ def __init__(self, priority, itemsize, name, np, shape):
   def __repr__(self): return f"dtypes.{self.name}({self.shape})"
 
 class dtypes:
-  @staticmethod # static methds on top, or bool in the type info will refer to dtypes.bool
-  def is_int(x: DType)-> bool: return x in (dtypes.int8, dtypes.uint8, dtypes.int32, dtypes.int64)
-  @staticmethod
-  def is_float(x: DType) -> bool: return x in (dtypes.float16, dtypes.float32, dtypes.float64, dtypes._half4, dtypes._float4)
-  @staticmethod
-  def is_unsigned(x: DType) -> bool: return x in (dtypes.uint8, dtypes.uint32, dtypes.uint64)
```
ReinBentdal, line 87: `is_int` used to be wrong? Checking for signed types + `uint8`.  Used in `llvmir.py` to check if type is integer, which I also believe to be a bug (since `is_int` does not include all int types). Hence why I changed from `is_int`, `is_unsigned_int` and `is_signed_int`

tinygrad/codegen/cstyle.py
```diff
@@ -119,11 +119,11 @@ def kk(s): kernel.append("  "*depth+s)
     elif uop == UOps.LOAD and newvar is not None:
       # TODO: merge with CONST?
       if bufs[args.i] is not None and isinstance(bufs[args.i].realized, RawConst):
-        assert newvar.dtype == dtypes.float, "const can't be float4"
+        assert newvar.dtype == dtypes.float32, "const can't be float4"
         x = bufs[args.i].realized._buf
         if math.isnan(x): val = "NAN"
         elif math.isinf(x): val = ("-" if x < 0 else "") + "INFINITY"
-        else: val = f"{x}" +  ("f" if not dtypes.is_int(bufs[args.i].dtype) else "")
+        else: val = f"{x}" +  ("f" if not dtypes.is_signed_int(bufs[args.i].dtype) else "")
```
ReinBentdal, line 126: this might also have been wrongly checked. Assume it should have checked for all int types and not just signed?

tinygrad/helpers.py
```diff
@@ -79,31 +79,32 @@ def __init__(self, priority, itemsize, name, np, shape):
...
+  def is_signed_int(x)-> bool: return x in (dtypes.int8, dtypes.int16, dtypes.int32, dtypes.int64)
+  @staticmethod
+  def is_unsigned_int(x) -> bool: return x in (dtypes.uint8, dtypes.uint16, dtypes.uint32, dtypes.uint64)
+  @staticmethod
+  def is_int(x) -> bool: return dtypes.is_signed_int(x) or dtypes.is_unsigned_int(x)
+  @staticmethod
+  def is_float(x) -> bool: return x in (dtypes.float16, dtypes.float32, dtypes.float64, dtypes._float4)
   bool: Final[DType] = DType(0, 1, "bool", bool)
-  float16: Final[DType] = DType(0, 2, "half", np.float16)
-  half = float16
```
geohot, line 94: Why were these aliases removed?

ReinBentdal: I removed float and half because I think its a bit confusing with two types which are identical. And I think float16 and float32 are more readable

ReinBentdal: I do changes which I in the moment makes sense to me. Maybe for future PR I should discuss these thinks first

geohot:
Can this be split into separate PRs. Functionality and refactors usually shouldn't be mixed, and I like the float and half aliases

ReinBentdal:
Sure! I'll do it soon

PR #1089, opened 2023-07-01
word error update and small cleanup

xhulianoThe1:


xhulianoThe1:
Is the current implementation not incorrect?

wozeparrot:
How so?

xhulianoThe1:
1). Reference is your x, Hypothesis/Prediction is your y
2). The for loop is unnecessary and it causes your denominator to be wrong.  
Ex: 
ref = "The dog is under the table"
hypothesis = "The dog is the fable"

Total words here is 16 with the current implementation should be 6 so WER should be .33 but the output is .5

wozeparrot:
Code is a simplified version from here: https://github.com/mlcommons/inference/blob/890af6b844cdc3d2e3053f51a40df20041b3852e/speech_recognition/rnnt/pytorch/metrics.py#L40

xhulianoThe1:
Ah okay, I think how it's being used in model_eval threw me off. Seems redundant.

xhulianoThe1:
Ah I'm dumb the input is not what I thought it would be. Anyway, if I have time, I think its usage in model eval can be simplified.  Thanks.

PR #1090, opened 2023-07-01
skip TestFloat64Dtype tests

danielbarter:
float64 not often supported on integrated GPUs, so probably good to skip related tests as discussed here: https://github.com/geohot/tinygrad/issues/1087


geohot:
Would merge ripping out float64 entirely.

PR #1093, opened 2023-07-02
Added torch.meshgrid functionality in tensor.py

AditMeh:
Added https://pytorch.org/docs/stable/generated/torch.meshgrid.html. Currently only supports 1-D `Tensors` as inputs (no scalars yet) and doesn't have an indexing argument.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -81,7 +81,12 @@ class TestOps(unittest.TestCase):
     helper_test_op([], lambda: torch.ones_like(b), lambda: Tensor.ones_like(a), forward_only=True)
   def test_eye(self):
     helper_test_op([], lambda: torch.eye(10), lambda: Tensor.eye(10), forward_only=True)
-
+  def test_meshgrid(self):
+    a = torch.meshgrid((torch.tensor([1,2,3]), torch.tensor([4,5,6,7])))
+    b =  Tensor.meshgrid(Tensor([1,2,3]), Tensor([4,5,6,7]))  
+    for (t1, t2) in zip(a, b):
+      helper_test_op([], lambda: t1, lambda: t2, forward_only=True)
+    
   def test_arange(self):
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                        11               0
```

dee7ine:
Maybe merge both the asserts into one. Even if it saves just one line I think it would look better. 
`assert(type(tensor) == Tensor and len(tensor.shape) == 1) # doesn't support scalars; currently only supports 1-D tensors`

geohot:
I've never heard of this method. In general, things only move to tensor.py after they are used in three models.

AditMeh:
A common use of this method is in Neural radiance field implementations ([vanilla NeRF](https://github.com/yenchenlin/nerf-pytorch/blob/63a5a630c9abd62b0f21c08703d0ac2ea7d4b9dd/run_nerf_helpers.py#L154), [voxel NeRF](https://github.com/sunset1995/DirectVoxGO/blob/341e1fc4e96efff146d42cd6f31b8199a3e536f7/lib/dvgo.py#L120))

If I included a minimal tinygrad implementation of a radiance field inside of `examples/` along with this PR, would it be sufficient for moving this to `tensor.py`?

SequoiaCoding:
> A common use of this method is in Neural radiance field implementations ([vanilla NeRF](https://github.com/yenchenlin/nerf-pytorch/blob/63a5a630c9abd62b0f21c08703d0ac2ea7d4b9dd/run_nerf_helpers.py#L154), [voxel NeRF](https://github.com/sunset1995/DirectVoxGO/blob/341e1fc4e96efff146d42cd6f31b8199a3e536f7/lib/dvgo.py#L120))
> 
> If I included a minimal tinygrad implementation of a radiance field inside of `examples/` along with this PR, would it be sufficient for moving this to `tensor.py`?

Have you started work on this? I was planning to do this with a friend but am not looking to duplicate effort!

PR #1098, opened 2023-07-02
Restoring old tests WIP DO NOT MERGE

posix4e:
I'm going through the repo and mostly just abusing the pull request feature to run tests and see how stuff works. Thanks for the great project? So what's the deal with these old broken tests? Do we wanna remove it? Is it important to test tensors this way. In my humble opinion we can just remove them.


geohot:
Much better to fix than to remove

PR #1101, opened 2023-07-02
Remove float64

rezaarezvan:
Removes all `float64` instances along with all associated tests with them.

### DIFF
--- a/test/external/external_test_onnx_backend.py
+++ b/test/external/external_test_onnx_backend.py
@@ -60,6 +60,7 @@ backend_test.exclude('uint32')
 backend_test.exclude('uint64')
 backend_test.exclude('int8')
 backend_test.exclude('int16')
+backend_test.exclude('float64')
 backend_test.exclude('string')
 
 backend_test.exclude('test_pow_types_int*')
--- a/test/external/external_test_onnx_backend.py
+++ b/test/external/external_test_onnx_backend.py
@@ -60,7 +60,6 @@ backend_test.exclude('uint32')
 backend_test.exclude('uint64')
 backend_test.exclude('int8')
 backend_test.exclude('int16')
-backend_test.exclude('float64')
 backend_test.exclude('string')
 
 backend_test.exclude('test_pow_types_int*')
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -105,6 +105,39 @@ class TestTinygrad(unittest.TestCase):
     expected = n * (1 - rate)
     np.testing.assert_allclose(non_zeros, expected, rtol=2e-3)
 
+  @unittest.skip("TODO: fix")
+  def test_jacobian(self):
+    W = np.random.RandomState(1337).random((10, 5))
+    x = np.random.RandomState(7331).random((1, 10)) - 0.5
+
+    torch_x = torch.tensor(x, requires_grad=True)
+    torch_W = torch.tensor(W, requires_grad=True)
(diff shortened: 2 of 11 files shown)


test/external/external_test_onnx_backend.py
```diff
@@ -60,7 +60,6 @@ def supports_device(cls, device: str) -> bool:
 backend_test.exclude('uint64')
 backend_test.exclude('int8')
 backend_test.exclude('int16')
-backend_test.exclude('float64')
```
geohot, line 63: Doesn't this exclude float64?

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/assembly_rdna.py          1               1
tinygrad/codegen/cstyle.py                 0               1
tinygrad/codegen/llvmir.py                 1               5
tinygrad/helpers.py                        1               2
tinygrad/runtime/lib.py                    1               1
tinygrad/runtime/ops_gpu.py                1               2
tinygrad/runtime/ops_metal.py              1               2
total                                      7              15
lines added in the tinygrad folder: -8
```

test/test_tensor.py
```diff
@@ -105,41 +105,6 @@ def test_dropout(self):
     expected = n * (1 - rate)
     np.testing.assert_allclose(non_zeros, expected, rtol=2e-3)
 
-  #@unittest.skipUnless(Device.DEFAULT == Device.CPU, "float64 not supported on GPU")
-  @unittest.skip("float64 support broken")
```
posix4e, line 109: If we remove this shouldn't we remove extra/gradcheck.py

test/external/external_test_onnx_backend.py
rezaarezvan: Oh yeah - my bad, I'll revert that

test/test_tensor.py
rezaarezvan: > If we remove this shouldn't we remove extra/gradcheck.py

Why?

geohot: Can you fix this test instead of removing it?

rezaarezvan: > Can you fix this test instead of removing it?

Sure, I'll see what I can do

geohot: would also merge with this test skipped and not removed

rezaarezvan: > would also merge with this test skipped and not removed

I think it's cleaner if we merge now. I'll open a new PR that fixes those tests instead

geohot:
Thanks!

Deferf:
Now when trying to import a random array from Numpy it returns `KeyError: 'float64'`. 

```
import numpy as np

X = np.random.random((3,3))
Tensor(X)
```

Are we expected to change the dtype? Or is it better to add a casting operation to fp32?

geohotstan:
Also for `numpy_fxn_for_op[BinaryOps.MAX]` in `ops_cpu.py`: 
```
In [33]: np.promote_types(np.float32, np.int32)
Out[33]: dtype('float64')

To reproduce:
In [1]: from tinygrad.lazy import Device                                                                                                                                                                                                          
In [2]: from tinygrad.tensor import Tensor                                                                                                                                                        
In [3]: from tinygrad.helpers import dtypes                                                                                                                                                                                                       
In [4]: Device.DEFAULT = "CPU"                                                                                                                   
In [5]: (Tensor.arange(3,dtype=dtypes.int32) == Tensor([1.0,1.0,2.412], dtype=dtypes.float32)).realize()
KeyError: 'float64'
```
torch on the other hand handles promote_types as promoting to float32 in these cases.

PR #1104, opened 2023-07-03
Remove POW llop and add SQRT llop

EliTheCoder:
Using exp log trick for pow


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/mlops.py                         11               2
tinygrad/runtime/ops_cpu.py                2               2
total                                     19               9
```

EliTheCoder:
A lot of the backends still need to be updated

tinygrad/mlops.py
```diff
@@ -133,7 +142,7 @@ def backward(self, grad_output:LazyBuffer) -> Tuple[Optional[LazyBuffer], Option
 class Pow(Function):
```
geohot, line 142: This should be removed and moved to hlops.

Pedroo1001:
ol h

tinygrad/tensor.py
```diff
@@ -579,7 +580,7 @@ def __matmul__(self, x) -> Tensor: return self.matmul(x)
   def __radd__(self, x) -> Tensor: return self.add(x, True)
   def __rsub__(self, x) -> Tensor: return self.sub(x, True)
   def __rmul__(self, x) -> Tensor: return self.mul(x, True)
-  def __rpow__(self, x) -> Tensor: return self.pow(x, True)
+  def __rpow__(self, x) -> Tensor: return Tensor(x, requires_grad=False).pow(self)
```
geohot, line 583: This shouldn't be needed, why doesn't pow match?

tinygrad/tensor.py
```diff
@@ -553,13 +553,14 @@ def _broadcasted(self, fxn:Type[Function], other:Union[Tensor, float], reverse:b
...
   def mul(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Mul, x, reverse) if x.__class__ is Tensor or x != 1.0 else self
-  def pow(self, x:Union[Tensor, float], reverse=False) -> Tensor:
-    if x.__class__ is not Tensor and not reverse:
+  def div(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Div, x, reverse) if x.__class__ is Tensor or reverse or not x else self.mul(1/x)
+  def pow(self, x:Union[Tensor, float]) -> Tensor:
+    if x.__class__ is not Tensor:
       # simple pow identities
       if x == 2.0: return self*self
+      if x == 1.0: return self
       if x == -1.0: return 1/self
```
geohot, line 562: Add sqrt and rsqrt here

tinygrad/tensor.py
```diff
@@ -553,13 +553,14 @@ def _broadcasted(self, fxn:Type[Function], other:Union[Tensor, float], reverse:b
   def add(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Add, x, reverse) if x.__class__ is Tensor or x else self
   def sub(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Sub, x, reverse) if x.__class__ is Tensor or x or reverse else self
   def mul(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Mul, x, reverse) if x.__class__ is Tensor or x != 1.0 else self
-  def pow(self, x:Union[Tensor, float], reverse=False) -> Tensor:
-    if x.__class__ is not Tensor and not reverse:
+  def div(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Div, x, reverse) if x.__class__ is Tensor or reverse or not x else self.mul(1/x)
```
geohot, line 556: Why did this have to change? Was there a bug?

tinygrad/tensor.py
EliTheCoder: This only gets called when you do float**Tensor. Instead of having a "reverse" argument, I just convert the float to a Tensor and then run pow. I don't know if there is a way to get the log of a float without converting it to a Tensor.

tinygrad/tensor.py
EliTheCoder: div didn't change it just got moved around so that add, sub, mul, and div are right next to each other.

tinygrad/tensor.py
EliTheCoder: What do you mean by negative being separate from the number?

geohot: eh, it's a minor thing

tinygrad/tensor.py
geohot: umm, `math.log`?

EliTheCoder: I wanted to do it on the GPU but I guess it doesn't matter

test/test_optim.py
```diff
@@ -71,10 +71,10 @@ def test_adamw(self): self._test_adamw(1, {'lr': 0.001}, 1e-5, 0)
   def test_adamw_high_lr(self): self._test_adamw(1, {'lr': 10}, 1e-5, 1e-5)
 
   def test_multistep_adam(self): self._test_adam(10, {'lr': 0.001}, 1e-5, 0)
-  def test_multistep_adam_high_lr(self): self._test_adam(10, {'lr': 10}, 1e-5, 3e-4)
+  def test_multistep_adam_high_lr(self): self._test_adam(10, {'lr': 10}, 2e-4, 5e-4)
```
geohot, line 74: Why are these worse? Does Adam use pow?

tinygrad/tensor.py
```diff
@@ -553,13 +553,16 @@ def _broadcasted(self, fxn:Type[Function], other:Union[Tensor, float], reverse:b
   def add(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Add, x, reverse) if x.__class__ is Tensor or x else self
   def sub(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Sub, x, reverse) if x.__class__ is Tensor or x or reverse else self
   def mul(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Mul, x, reverse) if x.__class__ is Tensor or x != 1.0 else self
+  def div(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Div, x, reverse) if x.__class__ is Tensor or reverse or not x else self.mul(1/x)
   def pow(self, x:Union[Tensor, float], reverse=False) -> Tensor:
+    a = self
```
geohot, line 558: Is this `a` needed?

tinygrad/tensor.py
```diff
@@ -553,13 +553,16 @@ def _broadcasted(self, fxn:Type[Function], other:Union[Tensor, float], reverse:b
...
       # simple pow identities
-      if x == 2.0: return self*self
-      if x == -1.0: return 1/self
-    return self._broadcasted(mlops.Pow, x, reverse) if x.__class__ is Tensor or x != 1.0 or reverse else self
-  def div(self, x:Union[Tensor, float], reverse=False) -> Tensor: return self._broadcasted(mlops.Div, x, reverse) if x.__class__ is Tensor or reverse or not x else self.mul(1/x)
+      if x < 0: return (1.0/self).pow(-x)
+      if x == 2.0: return a*a
+      if x == 1.0: return a
+      if x == 0.5: return a.sqrt()
+    return self.log().mul(x).exp() if not reverse or isinstance(x, Tensor) else self.mul(log(x)).exp()
```
geohot, line 565: Is this handling reverse correctly?

test/test_optim.py
wozeparrot: There was a bit of discussion about it in the discord a while back. Its why the old pr for this was pretty complex https://github.com/geohot/tinygrad/pull/959

geohot: Wait but the onnx pow isn't full two tensor pow. It's just like 0.9**<tensor>. This isn't accurate?

tinygrad/tensor.py
EliTheCoder: Yes. It passes test_pow and test_pow_const. It would fail if you tried to run Tensor.pow(Tensor, True), though. That is extremely abnormal usage of pow but do you want it to work?

test/test_optim.py
wozeparrot: Yea but it still goes through the approximation, and with higher lrs, the weights will be pretty big, and at large numbers it does start to diverge: https://godbolt.org/z/GfKPxcKTb

geohot: Hmm, what is torch adam doing? pow is so slow.

tinygrad/tensor.py
geohot: err, is there no way rpow can ever be called like that? it used to work, didn't it?

test/test_optim.py
wozeparrot: Yea it seems that they are: https://github.com/pytorch/pytorch/blob/main/torch/optim/adam.py#L366-L367

geohot: Hmm, this is making me reconsider pow removal. I thought it wasn't really used, but if it's in Adam...

I still don't understand why the approximation is so bad. It has to be the inaccuracy in `exp2` itself right? Aka with a perfect `exp2` this would be perfect?

wozeparrot: Yea its probably the inaccuracy in exp2, but yea, with a perfect `exp2` it should be perfect, like if you just do `math.exp2(math.log2(0.9) * 50)` (needs 3.11 for exp2) it's way more accurate cuz its doubles instead of floats.

tinygrad/tensor.py
EliTheCoder: rpow only gets called when you do `float**Tensor` so it will never get called for `Tensor**Tensor`
it did used to work without having to add anything extra since it could just pass the `reverse` argument to the `self._broadcasted` but it would take about two more lines to make it work here

geohot:
Discussed in voice chat. It's only the high LR Adam...and exp2 is a lot faster than GPU pow.

Fundamentally when we build hardware, I'd love to not need pow. So let's plan early :)

Thanks!

PR #1109, opened 2023-07-03
Small fix to abstractions.py so it runs errorless on Windows

tricky-labyrinth:
Here's the error message we were getting on Windows when we tried to run abstractions.py: error message

We add the dll export in to fix this.


bl-ke:
lgtm

PR #1111, opened 2023-07-03
Use generators instead of lists in `any`s and `all`s

AnselmC:
Small optimization for best/average case for `any` and `all` uses.  Be lazy when we can.
```shell
In [2]: %timeit any([x for x in range(int(1e6))])
18.9 ms ± 284 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [3]: %timeit any(x for x in range(int(1e6)))
296 ns ± 3.39 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit all([False for _ in range(int(1e6))])
16.2 ms ± 354 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit all(False for _ in range(int(1e6)))
279 ns ± 0.332 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -314,7 +314,7 @@ class Tensor:
 
   def cat(self, *args, dim=0):
     dim = (dim + len(self.shape)) if dim < 0 else dim
-    assert all([len(y.shape) == len(self.shape) and all([y.shape[i] == s for i,s in enumerate(self.shape) if i != dim]) for y in args])
+    assert all(len(y.shape) == len(self.shape) and all([y.shape[i] == s for i,s in enumerate(self.shape) if i != dim]) for y in args)
     catargs = [self] + list(args)
     assert all(len(t.shape) != 0 for t in catargs), "zero-dimensional tensor cannot be concatenated"
     shape_cumsum = [0, *accumulate([y.shape[dim] for y in catargs])]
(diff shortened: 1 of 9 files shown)


tinygrad/tensor.py
```diff
@@ -1,16 +1,22 @@
 # inspired by https://github.com/karpathy/micrograd/blob/master/micrograd/engine.py
 from __future__ import annotations
+
+import operator
 import time
 from functools import partialmethod, reduce
 from itertools import accumulate, filterfalse
-import operator
+from math import ceil, log, pi, prod, sqrt
```
AnselmC, line 8: have isort enabled on save, need to revert this.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py             4               4
tinygrad/helpers.py                        1               1
tinygrad/lazy.py                           5               5
tinygrad/shape/symbolic.py                 2               2
tinygrad/tensor.py                         4               4
total                                     19              19
```

geohot:
Eh, I heard it was faster the other way. But this is cleaner, and should be fixed in Python. Will merge

roelofvandijk:
This only faster for large iterators. For tinygrad, the generator overhead dominates. This commit caused a loss of around 5% on llama performance.

PR #1113, opened 2023-07-03
Broken Sigmoid backward: Add test and mlop for Sigmoid

kunwar31:
Sigmoid backward is broken for most of negative numbers, and this also breaks 
tanh, gelu and quick_gelu 

for large negatives, in Exp backward, we get 
-0. in grad_output, inf in self.ret, MUL of which gives NaN as grad

Adding mlop for sigmoid solves the issue, but there is an oppurtunity for fusion here

### DIFF
--- a/tinygrad/mlops.py
+++ b/tinygrad/mlops.py
@@ -41,7 +41,7 @@ class Log(Function):
   __slots__ = "x"
   def forward(self, x:LazyBuffer) -> LazyBuffer:
     self.x = x
-    return x.unary_op(UnaryOps.LOG2).binary_op(BinaryOps.MUL, x.const_like(math.log(2)/1))
+    return x.unary_op(UnaryOps.LOG2).binary_op(BinaryOps.MUL, x.const_like(math.log(2)))
 
   def backward(self, grad_output:LazyBuffer) -> LazyBuffer:
     return grad_output.binary_op(BinaryOps.DIV, self.x)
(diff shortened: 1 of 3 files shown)


kunwar31:
torch sigmoid backward kernel for reference: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cpu/BinaryOpsKernel.cpp#L725C6-L725C29

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/mlops.py                          9               0
total                                     10               2
```

geohot:
With a=-100 it doesn't test the normal case anymore

tinygrad/mlops.py
```diff
@@ -55,6 +55,15 @@ def forward(self, x:LazyBuffer) -> LazyBuffer:
   def backward(self, grad_output:LazyBuffer) -> LazyBuffer:
     return self.ret.binary_op(BinaryOps.MUL, grad_output)
 
+class Sigmoid(Function):
+  __slots__ = "ret"
+  def forward(self, x:LazyBuffer) -> LazyBuffer:
+    self.ret = x.const_like(1).binary_op(BinaryOps.DIV, x.const_like(1).binary_op(BinaryOps.ADD, x.binary_op(BinaryOps.MUL, x.const_like(-math.log(math.e)/math.log(2))).unary_op(UnaryOps.EXP2)))
```
geohot, line 61: `math.log(math.e) = 1`

You can replace this in the other places too, it's just dumb.

kunwar31:
> With a=-100 it doesn't test the normal case anymore

added the normal cases back

tinygrad/mlops.py
```diff
@@ -138,7 +147,7 @@ def forward(self, x:LazyBuffer, y:LazyBuffer) -> LazyBuffer:
 
   def backward(self, grad_output:LazyBuffer):
     return grad_output.binary_op(BinaryOps.MUL, self.y.binary_op(BinaryOps.MUL, self.ret.binary_op(BinaryOps.DIV, self.x))) if self.needs_input_grad[0] else None, \
-           grad_output.binary_op(BinaryOps.MUL, self.x.unary_op(UnaryOps.LOG2).binary_op(BinaryOps.MUL, self.x.const_like(math.log(2)/math.log(math.e))).binary_op(BinaryOps.MUL, self.ret)) if self.needs_input_grad[1] else None
+           grad_output.binary_op(BinaryOps.MUL, self.x.unary_op(UnaryOps.LOG2).binary_op(BinaryOps.MUL, self.x.const_like(math.log(2)/1)).binary_op(BinaryOps.MUL, self.ret)) if self.needs_input_grad[1] else None
```
geohot, line 150: Umm

tinygrad/mlops.py
```diff
@@ -41,20 +41,29 @@ class Log(Function):
   __slots__ = "x"
   def forward(self, x:LazyBuffer) -> LazyBuffer:
     self.x = x
-    return x.unary_op(UnaryOps.LOG2).binary_op(BinaryOps.MUL, x.const_like(math.log(2)/math.log(math.e)))
+    return x.unary_op(UnaryOps.LOG2).binary_op(BinaryOps.MUL, x.const_like(math.log(2)/1))
```
geohot, line 44: Please don't just find and replace and think about this for a minute.

kunwar31: my bad

geohot:
Cool, should unblock #1073

PR #1115, opened 2023-07-04
 Bounty for training mask-rcnn.

galegozi:


wozeparrot:
This is not enough to lock a training bounty, look at the other prs for the training bounty for an example of the minimum that should be done to lock it.

PR #1122, opened 2023-07-04
Use latest language syntax

AnselmC:
Use [pyugrade](https://github.com/asottile/pyupgrade) to make sure we use the latest language features for the target version. 
Not sure if this is controversial (aka seen as code golf) but IMO it enforces consistent style and is generally less verbose.

### DIFF
--- a/docs/abstractions.py
+++ b/docs/abstractions.py
@@ -41,11 +41,11 @@ import tinygrad.mlops as mlops
 # this is the good old familiar Tensor class
 class Tensor:
   # these two are pretty straightforward
-  grad: Optional[Tensor]
-  requires_grad: Optional[bool]
+  grad: Tensor | None
+  requires_grad: bool | None
 
   # this is the graph for the autograd engine
-  _ctx: Optional[Function]
+  _ctx: Function | None
 
   # this is where the data (and other tensor properties) actually live
   lazydata: LazyBuffer
@@ -74,7 +74,7 @@ from tinygrad.helpers import DType
 class LazyBuffer:
   # these three define the "type" of the buffer, and they are returned as Tensor properties
   device: str
-  shape: Tuple[int, ...]
+  shape: tuple[int, ...]
   dtype: DType
 
   # a ShapeTracker is used to track things like reshapes and permutes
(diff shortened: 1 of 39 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/graph.py                          2               2
total                                      3               3
```

docs/abstractions.py
```diff
@@ -85,18 +85,18 @@ class LazyBuffer:
...
-  op: Optional[LazyOp]
+  op: LazyOp | None
 
 # LazyOp (in tinygrad/ops.py, code 4/10)
 # in a tree they form an Abstract Syntax Tree for a single GPU kernel
 class LazyOp:
   op: Op                                       # the type of the compute
-  src: Tuple[Union[LazyOp, LazyBuffer], ...]   # the sources
-  arg: Optional[Any] = None                    # and an optional static argument
+  src: tuple[LazyOp | LazyBuffer, ...]   # the sources
```
jla524, line 98: does this work on python 3.8?

geohot:
tinygrad supports back to Python 3.8 for Ubuntu 20.04

AnselmC:
> tinygrad supports back to Python 3.8 for Ubuntu 20.04

? all the updates in this PR are python 3.8 syntax

geohot:
Err, so there's a chance I'm wrong about this.

I thought the `|` instead of `Union` was later.

jla524:
> Err, so there's a chance I'm wrong about this. I thought the `|` instead of `Union` was later.

It was [introduced in python 3.10](https://peps.python.org/pep-0604/), but CI passed with the 3.8 runners thanks to `from __future__ import annotations`.

AnselmC:
I totally agree that some of these are more code golf, however, I don't think you can pick-and-choose what `pyupgrade` does. 

I also agree that something like this needs to be enforced, hence the new `pre-commit` and gh workflow jobs.

In either case, lmk if you want to go ahead or not. If you do, I'll resolve the conflicts. If not, also good.

jdlib:
> A bunch of these changes are also very code golf. Why are the rs being removed from open? I know it's the same but why change it?

depends on the definition of code golf. Why explicitly pass the default value?

cauliyang:
If introducing `pyupdate`, why not use [ruff](https://github.com/astral-sh/ruff) directly?

geohot:
Yea, going to close this, I shouldn't have nitpicked about python 3.8. From discord: "every repo get shoved into the style wars without constant vigilance"

How does this PR further the goals of the tinygrad project?

Issue #967, opened 2023-06-10
Combined indexing, fancy indexing, broadcasting support 


kunwar31:
while gathering elements by a 1-D tensor along a single axes is somewhat possible with arange, compare, multiply (as shown here https://github.com/geohot/tinygrad/issues/858) , it does not allow for full fancy indexing (see cell 4 in [fancy indexing](https://jakevdp.github.io/PythonDataScienceHandbook/02.07-fancy-indexing.html#Exploring-Fancy-Indexing) ), and further, combined indexing, where gathers can be done across multiple axes with tensors (see  [combined indexing in numpy](https://jakevdp.github.io/PythonDataScienceHandbook/02.07-fancy-indexing.html#Combined-Indexing) )

Torch can handle this kind of indexing, and also does broadcasting on the fly on its `__getitem__`
![image](https://github.com/geohot/tinygrad/assets/24824080/dfd2b8b7-a4da-4296-a69b-0ac617702ff8)

kunwar31:
masked_index function in _bilinear_interpolate (https://github.com/geohot/tinygrad/pull/884) is blocked by this

kunwar31:
By converting everthing into 1-D, this can be achieved
```
import numpy as np
from tinygrad.tensor import Tensor, dtypes

input_ = Tensor.rand(2, 4, 6, 8)
a = Tensor.uniform(*(2, 1, 1, 1, 1, 1), low=0, high=1).cast(dtype=dtypes.int32)
b = Tensor.uniform(*(1, 4, 1, 1, 1, 1), low=0, high=3).cast(dtype=dtypes.int32) 
c = Tensor.uniform(*(2, 1, 4, 1, 2, 1), low=0, high=5).cast(dtype=dtypes.int32)  
d = Tensor.uniform(*(1, 1, 1, 4, 1, 2), low=0, high=7).cast(dtype=dtypes.int32)

def _gather(array, indices):
    arange = Tensor.arange(array.shape[-1]).reshape(1, array.shape[-1]).expand(*indices.shape, array.shape[-1])
    return Tensor.where(indices.unsqueeze(1).expand(*indices.shape, array.shape[-1]) == arange, array, 0,).sum(1)

def get_strides(shape):
    prod = [1]
    for idx in range(len(shape)-1, -1, -1): prod.append(prod[-1] * shape[idx])
    return Tensor(prod[::-1][1:]).unsqueeze(0)

def getitem(tensor, *keys):
[...]

PR #1125, opened 2023-07-04
Train Mask-RCNN

g1y5x3:
So far I am able to setup loading training data as well as iterate the `(image, target)` pair for running forward and backward.
`MODEL=maskrcnn python3 examples/mlperf/model_train.py` 
`(<Tensor <LB (3, 800, 1066) dtypes.float op=LoadOps.FROM st=ShapeTracker(shape=(3, 800, 1066), views=[View((3, 800, 1066), (852800, 1066, 1), 0, None)])> on GPU with grad None>, BoxList(num_boxes=50, image_width=1066, image_height=800, mode=xyxy), slice(0, 10, None))`

However, I think I need to refactor the code quite a bit. `models/mask_rcnn.py` is really messy and needs a lot of clean up. Ideally organize into different files according to their functionalities like what [mlcommons](https://github.com/mlcommons/training/tree/master/object_detection/pytorch/maskrcnn_benchmark/) did.

Additionally, a lot of the indexing operations have to be done with converting to numpy first and then convert back to Tensor, which was mentioned #967 and I think that's essentially why @kunwar31 implemented a bunch of additional functions just to index Tensors with different conditions. 

Finally, this is the current estimated inference time with a single NVIDIA 3090, it just felt wrong...
`1/5000 [02:10<181:21:58, 130.61s/it`

- [x] load and index through training `(image, target)` pair
- [ ] modularize `models/mask_rcnn.py` for better readability as well as remove some of the dependency on torch backend (mostly transforms) 
- [ ] find a solution for doing advanced indexing on Tensor directly
- [ ] add loss function into RPN, ROIHead
- [ ] perform training experiments

### DIFF
--- a/extra/gemm/metal_matmul.py
+++ b/extra/gemm/metal_matmul.py
@@ -6,6 +6,7 @@ from tinygrad.helpers import dtypes, getenv
 from tinygrad.runtime.ops_metal import RawMetalBuffer, MetalProgram
 
 N = getenv("N", 2048)
+LID = 2
 
 a = RawMetalBuffer(N*N, dtypes.float32)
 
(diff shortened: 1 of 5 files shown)


wozeparrot:
Need to be careful with `.numpy()` usage cuz grads won't flow through it, `.numpy()` should be fine for preprocessing and postprocessing stuff but it shouldn't be used in the actual model.

geohot:
Bounty locked, good luck!

models/mask_rcnn.py
```diff
@@ -254,31 +257,389 @@ def transpose(self, method):
 
   def clip_to_image(self, remove_empty=True):
     TO_REMOVE = 1
-    bb1 = self.bbox.clip(min_=0, max_=self.size[0] - TO_REMOVE)[:, 0]
-    bb2 = self.bbox.clip(min_=0, max_=self.size[1] - TO_REMOVE)[:, 1]
-    bb3 = self.bbox.clip(min_=0, max_=self.size[0] - TO_REMOVE)[:, 2]
-    bb4 = self.bbox.clip(min_=0, max_=self.size[1] - TO_REMOVE)[:, 3]
+    # TODO find solutions for 
+    # 'Tensor' object does not support item assignment
+    # 'Tensor' does not support & (and) operation
```
kunwar31, line 262: & operation is just elementwise multiply for 0/1 Tensors

g1y5x3:
lol made a mistake

PR #1126, opened 2023-07-04
Fix: Jacobian tests [WIP]

rezaarezvan:
`numerical_jacobian` is either bugged or not accurate enough, but this fixes the test for the `jacobian` function.


rezaarezvan:
The implementations are fine (I think, couldn't find anything explicitly wrong with them), and maybe higher accuracy can be achieved, but there are also a lot of variables that are important (the input function, eps, etc).

geohot:
Thanks!

PR #1128, opened 2023-07-05
add ptx formatter + syntax highlighter

cloud11665:
Also doing formatting as it's easier to compare ptx to cuda backend code.
orange - instruction or immediate value
purple - state space
blue - operand or identifier
green - type
white - modifier (be it infix or prefix)
image

### DIFF
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -1,12 +1,23 @@
 import subprocess
 from typing import Optional
 import time
+import re
 import numpy as np
 from pycuda.compiler import compile as cuda_compile # type: ignore
-from tinygrad.helpers import DEBUG, getenv, fromimport
+from tinygrad.helpers import DEBUG, getenv, fromimport, colored
 from tinygrad.ops import Compiled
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_cuda.py              19               2
lines added in the tinygrad folder: 17
```

geohot:
It's not obvious to me reading this that the text will always be the same (just colored differently)

cloud11665:
oh, so we don't want to format it (fix indentation on args and instructions, add newlines for params) ?

geohot:
I'd merge the colorizer.

cloud11665:
> I'd merge the colorizer (if it's clear from code it's just a colorizer), I think the formatting is fine from what I've seen. The risk is that it misprints the ptx and that bug wastes hours of someone's time.

Done. Explained in the comment in code:
 all expressions match `<valid_before><expr><valid_after>` and replace it with `<valid_before>color(<expr>)<valid_after>`
 
 The only caveat is that we don't color conditional instructions, but doing that would add additional complexity to the tokenizer which I don't think is worth it.

geohot:
Thanks!

PR #1134, opened 2023-07-05
Fold expand preceding reduce if the reduction is on the same axis as the expansion

rayanht:
New test case:

```
  def test_fold_expand_reduce(self):
    with CLCache(1):
      a = Tensor.randn(16, 1)
      a = a.expand((16, 16)).sum(-1)
      a.realize()
      assert len(GlobalCounters.cache) == 1, "optimizer didn't fold expand/reduce"
 ```
 
 This generates two kernels on master, including a reduction kernel
 
 ```
 ***    0 E_4_4                         arg   2 sz []                 [4]          OPs      0M/   0.00G  mem  0.00 GB tm     14.71us/     0.01ms (    0.00 GFLOPS,    0.01 GB/s)
***    1 r_16_16                       arg   2 sz [16]               [16]         OPs      0M/   0.00G  mem  0.00 GB tm     10.75us/     0.03ms (    0.02 GFLOPS,    0.01 GB/s)
```

Screenshot 2023-07-05 at 04 56 43

With these changes we have:

```
***    0 E_4_4                         arg   2 sz []                 [4]          OPs      0M/   0.00G  mem  0.00 GB tm     12.71us/     0.01ms (    0.00 GFLOPS,    0.01 GB/s)
```

Screenshot 2023-07-05 at 04 56 25

### DIFF
--- a/test/external/external_test_opt.py
+++ b/test/external/external_test_opt.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 import os
+
+import torch
 if "OPT" not in os.environ:
   os.environ["OPT"] = "2"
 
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                          12               3
```

geohot:
Hmm, wouldn't the expand and the sum multiply it by the value of the dimension?

rayanht:
>wouldn't the expand and the sum multiply it by the value of the dimension?

What's "it" here?

This is not from a concrete usage no but:

1. It ensures we don't miss this edge case as we implement more graph optimizations in `lazy.py`
2. Why would we allow users to shoot themselves in the foot? Optimizing compilers have to consider degenerate cases (within reason, of course it wouldn't make sense to write 100 lines of checks for an extreme edge case but 9 lines seems reasonable to me)

rayanht:
Hmm actually with a tiny bit more work this is on the path to fixing the reduce loop tests, let me explore that further

rayanht:
Okay yes, added a few more changes and now we can enable the `TestOptReduceLoop` suite

rayanht:
Yep, I'm adding some deterministic test cases, should've marked the PR as draft

rayanht:
Caught a bug with the previous implementation when I added the deterministic case, it only considered the case where the expand is on the first axis. Updated with a more generic implementation.

This is the graph resulting from the test case without the optimization:

Screenshot 2023-07-05 at 18 21 52

And with the optimization:

Screenshot 2023-07-05 at 18 20 36

rayanht:
Added a negative test case, ensure we don't fold things that can't be folded (different reduction axes)

rayanht:
Just corrected a comment and cleaned up some trailing whitespaces in the comments added in this PR, this is ready to merge yes.

A matching engine for AST optimization is probably a good idea; in that case, would it make sense to push the responsibility down to the Linearizer? Linearizer gets an unoptimized AST and runs a visitor pattern over it, matching rules.

geohot:
I don't think it belongs in the Linearizer.

The Linearizer gets an AST that's a single kernel, it should not be responsible for deciding how to group the kernels. That's in lazy.py

PR #1137, opened 2023-07-05
[WIP] fuse several reduces into one kernel

nimlgen:
The following code `_ = ((x@y)+(a@b)).numpy()` can benefit from fusing it into one kernel. It was measured to get about 10gflops (~195 -> ~205).

There are several cases where reduceops can be fused for training (found for efficientnet and resnet).

### DIFF
--- a/tinygrad/codegen/cstyle.py
+++ b/tinygrad/codegen/cstyle.py
@@ -170,34 +170,34 @@ def uops_to_cstyle(uops:List[UOp], bufs:List[Union[LocalBuffer,LazyBuffer]], lan
         kk(f"*(({lang.smem_prefix if isinstance(bufs[args.i], LocalBuffer) else lang.buffer_prefix}{bufs[args.i].dtype.name}4*)({bufnames[args.i]}+{args.idx.render(render_cl)})) = ({bufs[args.i].dtype.name}4){vin[0].render()};")
     elif uop == UOps.ATOMIC_ADD:
       atomic_on_bufs.add(args[0].i)
-      # TODO: Fix this.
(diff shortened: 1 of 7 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py            68              62
tinygrad/lazy.py                          14              10
total                                     82              72
```

geohot:
Hmm, so I guess this doesn't change the AST...

Can you post GRAPH=1 pictures to show what it's doing? Like in #1134

nimlgen:
For a simple case like `(x@y)+(a@b)` we have this graph:
old ast

turned into this one:
new ast

So, instead of 2 kernels (1st just reduce, 2nd reduce+binary ops) we got 1 kernel with 2 reduces and binary ops. The generated code looks like this one:
```
{ int gidx0 = gid.y;  /* 32 */
{ int gidx1 = gid.x;  /* 16 */
  { int lidx2 = lid.y;  /* 8 */
  { int lidx3 = lid.x;  /* 16 */
    // accs for 1st reduce
    for (int ridx4 = 0; ridx4 <= xxx; ++ridx4) {
     ...
    } /* reduce1 */
    // accs for 2nd reduce
    for (int ridx4 = 0; ridx4 <= xxx; ++ridx4) {
     ...
    } /* reduce2 */
    // binary ops.
}}}}
```

I think it would be great to use the laziness as much as possible, so I even think of trying to fuse more complex things like `((x+y)@y)+(a@b)` into one kernel... But this kinda the next step.

geohot:
Tests currently failing. But I like the direction this is going!

There's a $2,000 bounty for full norm fusion, it's quite a bit more complex than this as it requires multiple outputs, but it's in this vein. Posted an issue about it here #1146

PR #1138, opened 2023-07-05
cicd: use pip cache for dependency installs

roelofvandijk:
You can cache the pip download. Not sure if you care, but it's a simple change.


geohot:
Is it >20% faster?

roelofvandijk:
Surprisingly, it seems slower. Closing.

cloud11665:
Yea, the cache is only on the cuda emulated job, as pycuda builds c++ code upon installation, so caching that made sense.

PR #1141, opened 2023-07-05
Dedup kernel args (bufs) that have the same underlying RawBuffer

chaosagent:
This is part 1/k for winograd ( #1037 ). I will be splitting off orthogonal changes from that PR so that I can get some feedback / have less going on in that PR.

Before, only bufs with equal shapetracker and equal rawbuffer would be deduped. However, we may have multiple different shapetrackers on top of the same RawBuffer, which need to be deduped for UOp-level common subexpression elimination. The actual use-case in the winograd change is that two buffers may have the same shape/stride but different masks, due to the use of Tensor.stack. 

This introduces three properties to Linearizer: bufs (AST leaf buffers; not really useful anymore given that we have self.sts and self.raw_bufs), raw_bufs (deduped realized bufs, including rawconsts and localbuffers, for use in accelerator/cstyle/assembly backends), and raw_input_bufs (only realized RawBuffers, which should be passed to ASTRunner to execute the kernel).

This change assumes that all leaf buffers in an op AST are realized. I hope this is OK.

Conceptually, I am under the impression that LazyBuffer = (dtype, union[(st, rawbuffer), op]). Then LocalBuffer should be like a RawBuffer; without st or op but with dtype. I don't know if there are plans to have kernel-intermediate local buffers with ops in the future.

Not sure what is going on with arguments in assembly. Seems like master has LocalBuffers and RawConsts get passed in as arguments, which is suspicious. This patch keeps the same behavior.

Still need to write tests.


geohot:
Hmm, I feel like this can be done with a lot less line diff, and only in the place where it's used.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/assembly.py               2               2
tinygrad/codegen/assembly_ptx.py           1               1
tinygrad/codegen/assembly_rdna.py          1               1
tinygrad/codegen/cstyle.py                 6               7
tinygrad/codegen/linearizer.py            34              12
tinygrad/codegen/llvmir.py                 5               5
tinygrad/ops.py                            2               3
total                                     51              31
```

chaosagent:
OK we are now at only +1 real LOC in linearizer.py

I think everything here that is changed is conceptually important. LocalBuffer is a RawBuffer, and the compiled backends only need to know about RawBuffers.

The "ASTRunner.exec takes List[LazyBuffer]" change was good conceptually, and reduced the patch somewhat.

Now we have code to calculate the deduped set of inputs twice, and code to calculate which rawbufs are inputs twice. Hope they do not get de-synced.

self.bufs does not really have any reason to exist now (other than to map sts to raw_bufs) -- all the information is already in self.sts and self.raw_bufs. I guess we leave it in to reduce the line diff.

Antonio12389 (approved):
Si con esto queda ? Disculpen pero casi no entiendo

chaosagent:
There's a lot less lines, but now printbufs won't work after the end of linearize (not used anywhere except kernel_search.py), and lib_test_ast doesn't really do the right thing afaict if there are duplicate rawbuffers

tinygrad/codegen/linearizer.py
```diff
@@ -102,12 +102,16 @@ def __init__(self, ast:LazyOp, output_buffer:LazyBuffer):
     self.ast = ast.src[0] if ast.op == MovementOps.RESHAPE else ast
 
     # get the output buffers
+    # after linearize, this is will be deduped by key=buf.realized
```
geohot, line 105: this is will?

tinygrad/codegen/linearizer.py
```diff
@@ -223,6 +227,7 @@ def linearize(self):
       # TODO: the strides of this can be controlled
       self.sts.append(ShapeTracker(tuple([1] * self.first_reduce + self.group_for_reduce + [1] * (self.shape_len - self.upcasted - len(self.group_for_reduce) - self.first_reduce) + [x[0] for x in self.upcasted_axis(0)])))
       self.bufs.append(LocalBuffer("temp", self.sts[-1].size()))
+      self.bufmap.append(max(self.bufmap) + 1)
```
geohot, line 230: Does a non realized buffer have to go in the bufmap?

tinygrad/codegen/linearizer.py
```diff
@@ -329,6 +334,9 @@ def ssa(name, ltype=dtypes.float) -> Token:
       # end the global loop
       self.uop(UOps.ENDLOOP, None, [], (global_idxs, "global"))
 
+    # keep only the last buf for each RawBuffer -- these will be the inputs to the kernel
+    self.bufs = dedup([x.realized for x in self.bufs], self.bufs)
```
geohot, line 338: Why does this need to be here?

tinygrad/codegen/linearizer.py
chaosagent: global_load and global_store are called with -1 to load LocalBuffer, and they need to ask self.bufmap to find the pos after dedup

tinygrad/codegen/linearizer.py
```diff
@@ -102,12 +102,16 @@ def __init__(self, ast:LazyOp, output_buffer:LazyBuffer):
     self.ast = ast.src[0] if ast.op == MovementOps.RESHAPE else ast
 
     # get the output buffers
+    # after linearize, this is will be deduped by key=buf.realized
     self.bufs = [output_buffer] + dedup(ast.buffers)
 
+    # calculate mapping from to unique RawBuffer index
+    self.bufmap = [dedup([x.realized for x in self.bufs]).index(buf.realized) for buf in self.bufs]
```
geohot, line 109: Use the exact same logic for this as ASTRunner.exec. RawConsts shouldn't be in the bufmap

geohot:
Much most readable! Comments inline, and please add tests to confirm deduping behavior.

tinygrad/codegen/linearizer.py
geohot: Ahh, why don't you just handle that case in load and store?

test/test_ops.py
```diff
@@ -1019,6 +1020,18 @@ def test_inf_where(self):
     n = (x < 0).where(x, 1).numpy()
     assert np.all(n == 1.)
 
+  @unittest.skipUnless(Device.DEFAULT == "GPU", "Only GPU supports cache")
```
geohot, line 1023: All compiled backends support cache

tinygrad/codegen/linearizer.py
geohot: The -1 thing is a stupid hack anyway...want to just make it the right thing?

tinygrad/codegen/linearizer.py
chaosagent: RawConsts are loaded in cstyle.py using the LOAD uop, so the bufmap needs to have RawConsts.

There are three classes of usage:

linearizer: LazyBuffers (RawBuffer and RawConst), LocalBuffers
cstyle/assembly: deduped LazyBuffers, LocalBuffers
input: deduped LazyBuffers without RawConsts.

The bufmap is for the cstyle/assembly to know which sources to read from.

Maybe I am missing something.

chaosagent:
The gigantic diff is back. Half the diff is making the local reduce LocalBuf not -1 though...

chaosagent:
OK i feel like this is sufficiently refined. Let me know what you think.

chaosagent:
Do we want to dedup input buffers that are aliased to the output buffer?

.github/workflows/test.yml
```diff
@@ -281,15 +281,15 @@ jobs:
         key: ubuntu22.04-gpuocelot-19626fc00b6ee321638c3111074269c69050e091
         restore-keys: |
           ubuntu22.04-gpuocelot-19626fc00b6ee321638c3111074269c69050e091
-    - if: ${{ steps.cache-build.outputs.cache-hit != 'true' }}
```
geohot, line 284: why did this change?

chaosagent: For some reason the cached version of gpuocelot does not pass tests even with no changes from master (see a few commits above). This change is not meant to be merged; I will revert it once you approve this patch. Can you try clearing the gh actions cache? (Wozeparrot suggested this cause/line of action)

tinygrad/codegen/linearizer.py
```diff
@@ -133,10 +133,16 @@ def __init__(self, ast:LazyOp, output_buffer:LazyBuffer):
     # get the output buffers
     self.bufs = [output_buffer] + dedup(ast.buffers)
 
+    # dedup by raw buffer
+    self.dedup_bufs = dedup([x.realized for x in self.bufs], self.bufs)
```
geohot, line 137: What is x.realized is None?

tinygrad/codegen/linearizer.py
```diff
@@ -133,10 +133,16 @@ def __init__(self, ast:LazyOp, output_buffer:LazyBuffer):
...
+
     # key for lookup in cache (can change, str might not be right)
     # bufs are needed because kernels like f(x) = x + x and f(x, y) = x + y have the same str(ast), but are different kernels.
     # mapping the buffers to integers is required because a-b != b-a (and how would you tell a and b apart?)
-    self.key = (ast.map_buffers({x:i for i,x in enumerate(self.bufs)}).key, tuple([x.key for x in self.bufs]))
+    self.key = (ast.map_buffers({x:self.bufmap(i) for i,x in enumerate(self.bufs)}).key, tuple([x.key for x in self.bufs]))
+
+  def bufmap(self, i: int) -> int:
```
geohot, line 144: This is recomputing every time

.github/workflows/test.yml
geohot: I don't understand, there's 20 other PRs on this repo passing.

geohot:
Started this in #1256 

Should make this change much conceptually simpler

.github/workflows/test.yml
chaosagent: https://github.com/tinygrad/tinygrad/pull/1141/files/ffbdfcdb9c0433a08175f1c4ea5ea3728cc8b4b3

I don't understand either.

tinygrad/codegen/linearizer.py
chaosagent: When can this happen? I was under the impression that inputs to Linearizer must be realized. What does a realized buffer with .realized == None mean?

chaosagent:
closing; superceded by #1272 after refactor #1256

PR #1142, opened 2023-07-05
Don't save parents unless needed

cheeetoo:
Speeds up operations that do not require gradients by a bit since it no longer saves parents, and also speeds up backward pass since there is a less expensive check for if backward should be run.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -14,9 +14,10 @@ from tinygrad.ops import LoadOps
 # An instantiation of the Function is the Context
 class Function:
   def __init__(self, device:str, *tensors:Tensor):
-    self.device, self.parents = device, tensors
-    self.needs_input_grad = [t.requires_grad for t in self.parents]
+    self.device = device
+    self.needs_input_grad = [t.requires_grad for t in tensors]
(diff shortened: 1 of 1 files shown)


wozeparrot:
tests? does it provide a greater than 20% speedup?

cheeetoo:
It does not provide a >20% speedup, is that a requirement? I am not sure what you mean by tests, sorry.

wozeparrot:
> It does not provide a >20% speedup, is that a requirement? I am not sure what you mean by tests, sorry.

I'm not sure if this will hit any edge cases, and I don't think we have great test coverage around stuff like this, I stand to be corrected tho.

geohot:
Hmm, actually, I think this is good idea. The main thing it could save is RAM in freeing the parents.

roelofvandijk:
> Hmm, actually, I think this is good idea. The main thing it could save is RAM in freeing the parents.
> 
> Though it's sort of crazy this wasn't already done. The test coverage is the whole tinygrad...everywhere we do a backward pass.
> 
> I'm racking my brain about why I wrote `any(x.requires_grad for x in t0._ctx.parents)` and not just `t0.requires_grad`

If you can figure this out, there's quite the speedup hiding there. These __inits__ are called a LOT.

PR #1150, opened 2023-07-07
Fix evaluation stage in examples/transformer.py when using CUDA

barathsku:
**Problem Statement**
Running `examples/transformer.py` using CUDA crashes at the evaluation stage, error:
```
loss 0.38 accuracy 0.86: 100%|████████████████████████████████████████████████████████████████| 50/50 [00:04<00:00, 11.68it/s]
  0%|                                                                                                  | 0/16 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "/home/bdn/Desktop/Modules/tinygrad/examples/transformer.py", line 32, in <module>
    acc, Y_test_preds = evaluate(model, X_test, Y_test, num_classes=10, return_predict=True)
  File "/home/bdn/Desktop/Modules/tinygrad/extra/training.py", line 60, in evaluate
    acc, Y_test_pred = numpy_eval(Y_test, num_classes)
  File "/home/bdn/Desktop/Modules/tinygrad/extra/training.py", line 53, in numpy_eval
    out = model.forward(x) if hasattr(model, 'forward') else model(x)
  File "/home/bdn/Desktop/Modules/tinygrad/models/transformer.py", line 63, in forward
    xnp = x.cpu().numpy().astype(np.int32)
  File "/home/bdn/Desktop/Modules/tinygrad/tinygrad/tensor.py", line 112, in numpy
    def numpy(self) -> np.ndarray: return self.lazydata.toCPU()
  File "/home/bdn/Desktop/Modules/tinygrad/tinygrad/lazy.py", line 188, in toCPU
    realized = self.cast(dtypes.from_np(self.dtype.np)).contiguous().realize().realized
  File "/home/bdn/Desktop/Modules/tinygrad/tinygrad/lazy.py", line 144, in realize
    elif self.optype is LoadOps: LOAD_OPS_DISPATCHER[cast(LoadOps, self.op.op)](self)
  File "/home/bdn/Desktop/Modules/tinygrad/tinygrad/lazy.py", line 336, in _realize_contiguous
    realized = buffer.op.src[0].realize().realized
  File "/home/bdn/Desktop/Modules/tinygrad/tinygrad/lazy.py", line 144, in realize
    elif self.optype is LoadOps: LOAD_OPS_DISPATCHER[cast(LoadOps, self.op.op)](self)
  File "/home/bdn/Desktop/Modules/tinygrad/tinygrad/lazy.py", line 349, in _realize_from
    rawbuf = buffer.op.src[0].realize()
[...]

PR #1156, opened 2023-07-07
Fix ShapeTracker mismatch in LazyBuffer.fromCPU

sekstini:
The shape tracker in LazyBuffer.fromCPU is initialized with only the input array's shape, meaning it will always claim to be contiguous, including when the input array is not. This was the actual cause of the bug in #1150, as `x` already should already have been made contiguous [1].

This PR proposes a "fix" by simply making a copy of the array, which allocates a new contiguous buffer and thereby corrects the shape tracker mismatch. The downside is that it changes the previous behaviour of sharing/reusing the buffer.

If we wish to keep supporting buffer sharing, it should also be possible to construct a correct shape tracker by passing the array's shape, strides, and base.

[1]: Specifically the problem is that `_realize_contiguous` avoids emitting `UnaryOps.NOOP` (soon to be `LoadOps.CONTIGUOUS`) if the input is already contiguous (according to the shape tracker).

### DIFF
--- a/test/test_lazybuffer.py
+++ b/test/test_lazybuffer.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+import numpy as np
+import unittest
+from tinygrad.lazy import LazyBuffer
+
+class TestLazyBuffer(unittest.TestCase):
+  def test_fromcpu_buffer_sharing(self):
+    a = np.arange(8)
+    assert LazyBuffer.fromCPU(a).realized._buf is a
+
+  def test_fromcpu_shape_tracker(self):
+    def helper(a: np.ndarray):
+      print(a.shape, a.strides, a.flags.c_contiguous)
+      b = LazyBuffer.fromCPU(a).realize()
+      assert b.st.contiguous == a.flags.c_contiguous
+      assert b.st.shape == a.shape
+
+    for ndims in range(1, 4):
+      a = np.random.randn(*(4,)*ndims).astype(np.float32)
+      for stride in [-2, 1, 2]:
+        for start in [0, 1]:
+          helper(a[(slice(start, None, stride),)*ndims])
+
+if __name__ == "__main__":
+  unittest.main()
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
Are there any downsides to this?

sekstini:
It doesn't seem to break anything, but there could be a (probably small) performance hit from the extra copy. 

As mentioned we can keep the current behaviour if we construct the shape tracker properly, so it's a question of desired semantics.

Torch handles this by making a copy in `torch.Tensor`, but allows for buffer sharing through `torch.from_numpy`. I guess we could emulate this by adding a `copy` parameter to `Tensor.__init__` and `LazyBuffer.fromCPU` that defaults to True?

geohot:
it's not small. copies are very expensive

geohot:
I think it's right, but it definitely needs tests.

Example, try fromCPU on x[1:]

sekstini:
I'll throw in some tests. (sorry about the force pushes by the way, wasn't aware commits were squished)

sekstini:
So I'm not sure what you mean by "complete". I initially tried making a shape tracker with history, but this isn't possible in general as the base can be a torch tensor (this happens in a test).

If we had the real base array we could compare the data pointers from __array_interface__ to calculate an offset, but torch.from_numpy doesn't attach a reference the original array.

geohot:
I don't care about the history. Write a diverse set of tests, if it passes them it's good.

geohot:
LGTM! Thanks for pushing past copy

test/test_lazybuffer.py
```diff
@@ -0,0 +1,25 @@
...
+  def test_fromcpu_buffer_sharing(self):
+    a = np.arange(8)
+    assert LazyBuffer.fromCPU(a).realized._buf is a
+
+  def test_fromcpu_shape_tracker(self):
+    def helper(a: np.ndarray):
+      print(a.shape, a.strides, a.flags.c_contiguous)
+      b = LazyBuffer.fromCPU(a).realize()
+      assert b.st.contiguous == a.flags.c_contiguous
+      assert b.st.shape == a.shape
```
geohot, line 16: This doesn't test the contents is correct!

geohot:
Reverted. I should have read more carefully

geohot:
Will remerge with more tests, my bad, I'm a bit confused about this too

PR #1160, opened 2023-07-07
Removed two redundant ifs from simplify_ones function in linearizer

terafo:
So the first if is basically needed only because second exists so program doesn't crash, and second was needed only because tinygrad didn't support zero-dims. Which, ironically, was fixed with commit that introduced first if.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py             0               4
```

geohot:
Please add tests to confirm this didn't break anything

geohot:
Actually going to close this.

Issue #1155, opened 2023-07-07
Move datasets -> extra/datasets


geohot:
It's annoying that it conflicts with huggingface.

PR #1161, opened 2023-07-07
Fix naming conflict with huggingface datasets

terafo:
lm-eval uses huggingface datasets for everything and name conflicts from that are annoying and unnecessary.


terafo:
Ok, I haven't noticed comment edits in my original PR. tinydatasets is indeed quite long and similar enough to tinygrad to be confused. Would something like `nanodata`/`petitedata`/`babysets` be an OK name?

mdatsev:
extra/datasets #1155

g1y5x3:
Or just put it under extra/ like he suggested in discord

examples/mnist_gan.py
```diff
@@ -7,7 +7,7 @@
 from tinygrad.tensor import Tensor
 from tinygrad.helpers import getenv
 from tinygrad.nn import optim
-from datasets import fetch_mnist
+extra.datasets import fetch_mnist
```
jla524, line 10: from extra

PR #1162, opened 2023-07-07
Fix Metal tests [WIP]

rezaarezvan:
I've managed to track down why and fix the half-to-int tests, but now remains the conv tests.

After that, all metal tests should pass!


tinygrad/codegen/linearizer.py
```diff
@@ -566,7 +566,7 @@ def hand_coded_optimizations(self):
     for splits in [4]:
       if self.upcasted == 0 and len(self.full_unupcasted_shape) > 0 and self.full_unupcasted_shape[-1] % splits == 0:
         self.shift_to(len(self.full_unupcasted_shape)-1, splits, insert_before=len(self.full_unupcasted_shape))
-        self.upcast()
+        if self.bufs[0].dtype.name.startswith('image') and self.first_reduce > 0 and self.first_reduce < (self.shape_len-self.upcasted): self.upcast()
```
geohot, line 569: Is this a fix or a mitigation?

rezaarezvan: This is a fix according to my testings

geohot: It's not. What's broken that this fixes it?

rezaarezvan: > It's not. What's broken that this fixes it?

From what I've found, the upcast is what breaks these tests. But yeah, to be fair I haven't looked deeper into why it breaks. I'll should look into that

PR #1163, opened 2023-07-07
docs/abstractions.py typo fix

JJJJJJJJJJJJJJJl:
The src property contains the LB not the arg

> LazyOp(op=LoadOps.FROM, src=(<LB (1,) dtypes.float op=buffer<1, dtypes.float> st=ShapeTracker(shape=(1,), views=[View((1,), (0,), 0, None)])>,), arg=None)


geohot:
Err, I don't think so. I agree the docs are wrong, but this isn't the fix

PR #1165, opened 2023-07-07
helpers.py: improved test coverage + exception handling

dosier:
This PR adds better exception handling to prevent hangs and also improves the test coverage.

- Added exception handling in `proc`, before if an exception was thrown, the thread would hang.
- Made `_early_exec_process` catch any Exception. Otherwise, if an exception was thrown before the process was started, it would hang the thread. For example, a type error for an argument passed to `subprocess.check_output`.
- Also I removed the Timings import from extra/helpers, that one seemed out of place and confused my IDE, but if there is a reason for it then I can add it back.


extra/helpers.py
```diff
@@ -26,32 +24,28 @@ def early_exec(x):
...
+  def __getstate__(self): return cloudpickle.dumps(self.fn)
+  def __setstate__(self, pfn): self.fn = cloudpickle.loads(pfn)
+  def __call__(self, *args, **kwargs) -> Any:  return self.fn(*args, **kwargs)
 
 def cross_process(itermaker, maxsize=16):
   q: multiprocessing.Queue = multiprocessing.Queue(maxsize)
   # multiprocessing uses pickle which cannot dump lambdas, so use cloudpickle.
   p = multiprocessing.Process(target=proc, args=(_CloudpickleFunctionWrapper(itermaker), q))
-  #p.daemon = True
+  p.daemon = True
```
dosier, line 45: I made it a daemon to prevent the multiprocessing module from trying to join the child process when the main process exits. Otherwise, in case of an exception happening in the child process, the main process tries to join it twice which raises an exception.
I think it's safe because the child process should always finish before the main process since the main process is waiting for the child process to send a `None` value before it exits the `cross_process` function.

extra/helpers.py
```diff
@@ -26,32 +24,28 @@ def early_exec(x):
...
   p.start()
-
-  # TODO: write tests and handle exit case
```
dosier, line 53: I figured the graceful exit case was already handled, i.e. when `None` is returned, but the case of an exception was not.

extra/helpers.py
geohot: It's not a daemon because daemons can't spawn more processes

extra/helpers.py
geohot: Yea this is an improvement

PR #1166, opened 2023-07-07
Improve hlb_cifar10.py 91.47%, goal 94%

artem9k:
I use 2000 steps, same batch size, roughly same size model


mdatsev:
Don't mix large whitespace changes and functional changes

geohot:
This is not the bounty.

"Over 94% on CIFAR with examples/hlb_cifar10.py without increasing compute much (TTA okay)"

See "without increasing compute much"

Also what is that spacing?

artem9k:
ok, will fix spacing and try to use less compute

PR #1168, opened 2023-07-07
fix Tensor.manual_seed() default to wrong type

Maroc-OS:
manual_seed() has the wrong type None instead of int, this fails if we call it without argument: #1169

Added tests


test/test_tensor.py
```diff
@@ -136,6 +136,19 @@ def test_gradcheck(self):
     # coarse approx. since a "big" eps and the non-linearities of the model
     self.assertFalse(gradcheck(tiny_func, tiny_x, eps = 1e-5))
 
+  def test_manual_seed(self):
```
geohot, line 139: This is a useless test

PR #1173, opened 2023-07-07
Fix: auto create parent dir when downloading file

dosier:
Did this because some examples that use `download_file` expect a weights dir, this way the user doesn't have to manually create it. 
- also removed duplicate import `os`

### DIFF
--- a/extra/utils.py
+++ b/extra/utils.py
@@ -33,13 +33,13 @@ def fetch_as_file(url):
   return fp
 
 def download_file(url, fp, skip_if_exists=True):
-  import requests, os, pathlib
+  import requests, pathlib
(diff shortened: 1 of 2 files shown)


geohot:
Test for this?

dosier:
> Test for this?

Added test, I mocked the download op cuz no internet in CI

geohot:
No internet in CI?

dosier:
> No internet in CI?

In the existing test, I saw "no internet tests in CI":
```
@unittest.skipIf(getenv("CI", "") != "", "no internet tests in CI")
class TestFetch(unittest.TestCase):
```
from which I figured that we cannot make remote requests. 

But if that is outdated, I can change it :)

geohot:
meh, LGTM

PR #1182, opened 2023-07-08
[WIP] merge_views fix proposal

g1y5x3:
Just wondering if I am on the right track, here is a test script that I used
```
from tinygrad.shape.shapetracker import ShapeTracker, View

v1 = View((4,2), (2,1), 0, (1,1))
v2 = View((4,2), (2,1), 0, ((1,1),(2,1)))

s1 = ShapeTracker((4,2), [v1])
s2 = ShapeTracker((4,2), [v2])
# print(s1)
# print(s1.reshape((2,4)))
print(s2)
print(s2.reshape((2,4)))
```
And here is the output
```
ShapeTracker(shape=(4, 2), views=[View((4, 2), (2, 1), 0, ((1, 1), (2, 1)))])
ShapeTracker(shape=(2, 4), views=[View((2, 4), (4, 1), 0, ((0, 3), (1, 1)))])
```
I know it only works strictly on 2D shapes but that is easy to fix. I want to ask some question to clarify

1. If the function is not called from `_reshape`, then whatever is in `vm1.mask` that's not a duplicate should also be merged into the final output?
2. `for m in vm2.mask:` only works if I have at least two tuples in it. And I can't find an clean way to fix it. `print(s1.reshape((2,4)))` will break as well as in `strides = mst.real_strides()` even I by pass the for loop


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/shapetracker.py            14               2
lines added in the tinygrad folder: 12
```

g1y5x3:
will have to make sure the fix pass all the test and correct logically, I don't believe its going to be a lot of lines but logic is a bit tricky.

PR #1184, opened 2023-07-08
Cancel previous workflow runs on new commits

wozeparrot:
Something is very wrong with CI right now, the queue is pretty much completely backed up.

This should help a bit by canceling older runs when there are new commits.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


geohot:
I'm manually cancelling them, idk why it got backed up. Is it possible to add a 20 minute timeout? Nothing should take longer than that

geohot:
Added 20 minute timeout to all jobs

wozeparrot:
Yea we can set a timeout for the jobs, I think some of the changes in the dedup buffer pr are causing cuda ci and llvm ci? to hang for some reason.

geohot:
Also upgraded GitHub account to GitHub Teams. Should go from 20 -> 60 concurrent runners allowed.

PR #1185, opened 2023-07-08
ci: cache downloads

### LINKED ISSUE #1180: Use GitHub actions to cache files we download in CI
I think this is possible

unofficialquant:
This should fix #1180 by using cache in `actions/setup-python@v4` to store the downloaded files and `actions/cache@v3` to skip installation by caching the python location if `setup.py` is not changed. `actions/cache@v3` stores `efficientnet` model as well by caching the `test/models/efficientnet`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,8 +12,8 @@ jobs:
     timeout-minutes: 20
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Set up Python 3.8
       uses: actions/setup-python@v4
       with:
@@ -41,8 +41,8 @@ jobs:
     timeout-minutes: 20
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Set up Python 3.8
       uses: actions/setup-python@v4
       with:
@@ -62,8 +62,8 @@ jobs:
     timeout-minutes: 20
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Set up Python 3.8
       uses: actions/setup-python@v4
       with:
@@ -83,8 +83,8 @@ jobs:
     timeout-minutes: 20
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Set up Python 3.8
       uses: actions/setup-python@v4
       with:
@@ -102,8 +102,8 @@ jobs:
     name: CLANG Tests ${{ matrix.os }} (w method cache)
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Set up Python 3.8
       uses: actions/setup-python@v4
       with:
@@ -121,8 +121,8 @@ jobs:
     timeout-minutes: 20
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Set up Python 3.8
       uses: actions/setup-python@v4
       with:
@@ -140,8 +140,8 @@ jobs:
     timeout-minutes: 20
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Update packages
       run: |
         wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB | gpg --dearmor | sudo tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null
@@ -169,8 +169,8 @@ jobs:
     timeout-minutes: 20
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Update packages
       run: |
         wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB | gpg --dearmor | sudo tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null
@@ -204,8 +204,8 @@ jobs:
     timeout-minutes: 20
 
     steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
+      - name: Checkout Code
+        uses: actions/checkout@v3
     - name: Set up Python 3.11
       uses: actions/setup-python@v4
       with:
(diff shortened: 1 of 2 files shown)


PeterMcD1:
this does not cache files that are downloaded when running the tests, for example the efficientnet test downloads the model but does not cache it currently

unofficialquant:
> This does not cache files that are downloaded when running the tests, for example the efficientnet test downloads the model but does not cache it currently.
> 
> Will require tests to actually download files instead of just fetching them I think.

ah yes you are right. will add that once i get the CI caching passing. anywhere i can reference the download paths?

unofficialquant:
> This does not cache files that are downloaded when running the tests, for example the efficientnet test downloads the model but does not cache it currently.
> 
> Will require tests to actually download files instead of just fetching them I think.

should be fixed now!

roelofvandijk:
What is being saved? Runner time? Bandwidth? Ingress? 

Based on the tqdm timings, the downloads are near instant because the download speed is so high.

Creating the cache, however, is 1 (CLANG ubuntu-latest) or 2 (CLANG windows) minutes.

unofficialquant:
> What is being saved? Runner time? Bandwidth? Ingress?
> 
> Based on the tqdm timings, the downloads are near instant because the download speed is so high.
> 
> Creating the cache, however, adds 1 min (CLANG ubuntu-latest job) or 2 min (CLANG windows job). Is the cache action compressing stuff or so?

Some bandwidth and runner time since it'll cache both python deps and download of models. Tho marginal since download speed is high as you said. Creating the cache is only done when there's changes to the `setup.py`. In this case, it takes time to create it because there's no existing cache. Cache would not need to be updated frequently once its there. Would suggest having the discussion in the related issue since this PR is created to address the issue that was raised.

unofficialquant:
@geohot would be great if you could re-trigger a workflow run on this PR to compare the difference

geohot:
Any evidence this is faster?

unofficialquant:
I think it is by a margin. Would it be possible for you to trigger a re-run on this PR? We could compare it to the last run which should have it cached already

geohot:
Triggered

unofficialquant:
@geohot fixed some issues with setup-python cache since its conflicting with the `actions/cache`. based on the latest run i do notice that it it speeds up significantly on installing the deps and downloads of torch models. some parts thats slow are test related so i'm not sure whether its actually helpful. also other model downloads (e.g. vit_models) could potentially be faster. not sure if that's something worth working on. let me know what you think.

4ydan (commented):
Great stuff, just one question.

roelofvandijk:
@unofficialquant 

The downloaded models are stored in a tempfile based on a hash of the url or directly read (`fetch` and `fetch_as_file`).
https://github.com/tinygrad/tinygrad/blob/e2f6b09ffde3de184421c1dff164be1edee95cca/extra/utils.py#L16
If those are cached, you probably need a `.exists()` check to avoid re-downloading.

What about caching all .pth files in the tempdir using a [glob pattern](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action)?
For that to work, the filenames in those fetch functions would have to be changed to retain the extension.

This cache could potentially even be shared between jobs.

.github/workflows/test.yml
```diff
@@ -16,10 +16,19 @@ jobs:
...
+      with:
+        key: linter-${{ runner.os }}-${{ hashFiles('setup.py') }}
+        path: |
+          ${{ env.pythonLocation }}
+          ${{ env.LD_LIBRARY_PATH }}
+          ${{ github.workspace }}/test/models/efficientnet
+          ~/.cache/torch/hub
     - name: Install dependencies
-      run: pip install -e '.[linting,testing]' --extra-index-url https://download.pytorch.org/whl/cpu
+      run: pip install --upgrade-strategy eager -e '.[linting,testing]' --extra-index-url https://download.pytorch.org/whl/cpu
```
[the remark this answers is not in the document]
unofficialquant: yes you are right. just checked. will remove it thanks!

unofficialquant:
> @unofficialquant
> 
> The downloaded models are stored in a tempfile based on a hash of the url or directly read (`fetch` and `fetch_as_file`).
> 
> https://github.com/tinygrad/tinygrad/blob/e2f6b09ffde3de184421c1dff164be1edee95cca/extra/utils.py#L16
> 
> If those are cached, you probably need a `.exists()` check to avoid re-downloading.
> 
> What about caching all .pth files in the tempdir using a [glob pattern](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action)? For that to work, the filenames in those fetch functions would have to be changed to retain the extension.
> 
> This cache could potentially even be shared between jobs.

The issue with `tempfile` is it creates the directory at `/var/folders`. I've tested earlier on and GitHub action is not able to cache anything from that directory. Would have to store all the models at some other directory in order to cache them.

.github/workflows/test.yml
```diff
@@ -18,6 +18,17 @@ jobs:
...
+    - name: Check for cache
+      uses: actions/cache@v3
+      with:
+        key: linter-${{ runner.os }}-${{ hashFiles('setup.py') }}
+        path: |
+          ${{ env.pythonLocation }}
+          ${{ env.LD_LIBRARY_PATH }}
+          ${{ github.workspace }}/.pytest_cache
+          ${{ github.workspace }}/.mypy_cache
+          ${{ github.workspace }}/test/models/efficientnet
```
4ydan, line 30: Do you need `/test/models/efficientnet` and `~/.cache/torch/hub` in jobs that dont make use of it?

.github/workflows/test.yml
```diff
@@ -108,6 +152,17 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: 3.8
+    - name: Check for cache
+      uses: actions/cache@v3
+      with:
+        path: |
+          ${{ env.pythonLocation }}
+          ${{ env.LD_LIBRARY_PATH }}
+          ${{ github.workspace }}/.pytest_cache
```
4ydan, line 161: Does .pytest_cache help in any of the jobs, when pytest is invoked without -ff or -lf?

4ydan (commented):
Some more questions.

.github/workflows/test.yml
4ydan, line 30: Do you need models and cache/torch/hub when pytest is not beeing invoked.

unofficialquant: not needed but for consistency across all jobs i've included it there.

.github/workflows/test.yml
unofficialquant: yes it does, i'm not sure about the exact implementation but i believe it kinds of skips when the file tested is unchanged. `-ff` and `-lf` is for when you have failure cases only by only running on them.

roelofvandijk:
> The issue with `tempfile` is it creates the directory at `/var/folders`. I've tested earlier on and GitHub action is not able to cache anything from that directory. Would have to store all the models at some other directory in order to cache them.

To allow caching of the weights, they would have to be stored somewhere else. Either always (in a folder ignored by git), or only when CI jobs are ran (e.g. `if getenv('CI', False)` and then `CI=1 python -m pytest ...`. Two locations seem natural: `/weights` or `/cache`. Both are already ignored by git. By not including the job name in the key this cache shared between jobs, if the required weights largely overlap.

Generally speaking, I think caching the weight downloads is a great idea, however, _caching of the Python virtual environments can be problematical_. `setup.py` does not specify the versions of the dependencies, so now the latest are used. By caching the environment in the jobs, packages will get stuck on the versions in use when the cache was created, which means local environments will diverge from the CI environments. Eventually, an API will change somewhere and weird job failures will occur.

unofficialquant:
reopen

PR #1076, opened 2023-06-29
onnx full passing

geohotstan:
Failing tests (53): https://gist.github.com/geohotstan/db8d722fe311db2b06709eb2a5276a00
Todo:
- get to 0 failing
- clean up code (lots of clean up) 
- write better implementations for ops (Slice, gather, resize, pad, ...) and type parse. I'm using too many kernels (openpilot test failing since 205 (<199) kernels used), and some of my current implementations are admittedly, quite ugly. Tinygrad is  elegant.
- maybe also benchmarking (onnx runtime??? I have no idea how to do this)

Posting this draft PR since I've been perma stuck on ~~linear~~ and cubic interpolation for Resize for awhile now and also kinda unsure if the code I'm writing is totally disgusting or passable.... 😢
Not sure if anyone else have any ideas on interpolation.
For bounty hunters, feel free to take parts for your onnx fixes or write an implementation better than mine so that you can help me learn and we can share the fun!

I'm still dead set on getting to 0 failing though.

PR #1188, opened 2023-07-08
Implementation of VITS TTS model

dosier:
**Bountie**: _good sounding text to speech model in tinygrad, aka make tinygrad talk nicely_
**Original Paper:** https://arxiv.org/abs/2106.06103
**Original Code**: https://github.com/jaywalnut310/vits

- [x] auto downloading of model weights + config
- [x] vctk model (english, 109 speakers)
- [x] ljs model (english, 1 speaker)
- [x] [mmt-tts model (english, 1 speaker)](https://huggingface.co/facebook/mms-tts)
- [x] [umamusume model (multilingual, 999 anime girls)](https://huggingface.co/spaces/Plachta/VITS-Umamusume-voice-synthesizer)
- [x] [voistock model (multilingual, 2891 anime characters)](https://github.com/CjangCjengh/TTSModels)
- [x] [cjks model (multilingual, 24 speakers)](https://github.com/CjangCjengh/TTSModels)
- [x] waiting for merge of `tensor.squeeze` #1241 
- [x] waiting for merge of `nn.ConvTranspose1d` #1243 
- [x] Use a different method of exporting audio to wav file (remove ipy dep)

The gather operation was taken from #1076

Special thanks to @wozeparrot for helping test it out a bit, providing feedback, answering some questions,  and find more models to load :)


geohot:
Bounty locked, good luck!

dosier:
Got it to run at around 0.3 sec inference time for a single sentence on my M1, speed comparable to torch, waiting on #1223 to be merged before I can finish it fully due to [#1221](https://github.com/tinygrad/tinygrad/issues/1221).

examples/vits.py
```diff
@@ -0,0 +1,624 @@
...
+  logger = logging
+  Tensor.no_grad, Tensor.Training = True, False
+  Tensor.manual_seed(1337)
+  np.random.seed(1337)
+  net_g, hps = load_model(MODELS[MODEL_TO_USE])
+  stn_tst = get_text(TEXT_TO_SYNTHESIZE, hps)
+  logger.info(f"Converted input text to tensor \"{TEXT_TO_SYNTHESIZE}\" -> Tensor({stn_tst.shape}): {stn_tst.numpy()}")
+  x_tst, x_tst_lengths = stn_tst.unsqueeze(0), Tensor([stn_tst.shape[0]], dtype=dtypes.int64)
+  start_time = time.time()
+  out = net_g.infer(x_tst, x_tst_lengths, sid=Tensor([4], dtype=dtypes.int64) if MODEL_TO_USE == "vctk" else None, noise_scale=.667, noise_scale_w=0.8, length_scale=1)
```
wozeparrot, line 619: If your counting the time it takes to infer, make sure you realize here so that it correctly measures all the time.

dosier:
Waiting for https://github.com/tinygrad/tinygrad/pull/1241 so I can remove the squeeze implementation in vits.py

examples/vits.py
dosier: AH oops, when I was refactoring stuff I forgot to change that, cheers!

examples/vits.py
```diff
@@ -0,0 +1,695 @@
...
+  parser.add_argument("--noise_scale_w", type=float, default=0.8, help="Specify the noise scale w. Default is 0.8.")
+  parser.add_argument("--length_scale", type=float, default=1, help="Specify the length scale. Default is 1.")
+  parser.add_argument("--seed", type=int, default=1337, help="Specify the seed (set to None if no seed). Default is 1337.")
+  parser.add_argument("--num_channels", type=int, default=1, help="Specify the number of audio output channels. Default is 1.")
+  parser.add_argument("--sample_width", type=int, default=2, help="Specify the number of bytes per sample, adjust if necessary. Default is 2.")
+  args = parser.parse_args()
+
+  model_has_sid = args.model_to_use in ["vctk", "uma_trilingual"] # has more than one speaker, speaker selected through setting speaker_id
+
+  Tensor.no_grad, Tensor.Training = True, False
```
wozeparrot, line 649: `Training` should have a lowercase t, shouldn't need this anyways cuz it defaults to False.

dosier: Oops, removed it!

examples/vits.py
```diff
@@ -443,6 +430,45 @@ def rational_quadratic_spline(inputs: Tensor, unnormalized_widths: Tensor, unnor
...
+  path = sequence_mask(duration.cumsum(axis=2).reshape(b * t_x), t_y).cast(mask.dtype).reshape(b, t_x, t_y)
+  path = path - path.pad(convert_pad_shape([[0, 0], [1, 0], [0, 0]]))[:, :-1]
+  return path.unsqueeze(1).transpose(2, 3) * mask
+def fused_add_tanh_sigmoid_multiply(input_a: Tensor, input_b: Tensor, n_channels: int):
+  n_channels_int, in_act = n_channels, input_a + input_b
+  t_act, s_act = in_act[:, :n_channels_int, :].tanh(), in_act[:, n_channels_int:, :].sigmoid()
+  return t_act * s_act
+
+def cat_lr(t, left, right): return Tensor.full(get_shape(t), left).cat(t, dim=-1).cat(Tensor.full(get_shape(t), right), dim=-1)
+def masked_fill(tensor, mask, value): return tensor * (1 - mask) + value * mask
```
wozeparrot, line 446: Isn't this just a `.where`?

dosier: > Isn't this just a `.where`?

True :)

geohot:
Testing now

examples/vits.py
```diff
@@ -0,0 +1,725 @@
...
+    self.flow = ResidualCouplingBlock(inter_channels, hidden_channels, 5, 1, 4, gin_channels=gin_channels)
+    self.dp = StochasticDurationPredictor(hidden_channels, 192, 3, 0.5, 4, gin_channels=gin_channels) if use_sdp else DurationPredictor(hidden_channels, 256, 3, 0.5, gin_channels=gin_channels)
+    if n_speakers > 1: self.emb_g = nn.Embedding(n_speakers, gin_channels)
+  def infer(self, x, x_lengths, sid=None, noise_scale=1.0, length_scale=1, noise_scale_w=1., max_len=None, emotion_embedding=None):
+    x, m_p, logs_p, x_mask = self.enc_p.forward(x, x_lengths, emotion_embedding)
+    g = self.emb_g(sid.reshape(1, 1)).squeeze(1).unsqueeze(-1) if self.n_speakers > 0 else None
+    logw = self.dp.forward(x, x_mask, g=g, reverse=self.use_sdp, noise_scale=noise_scale_w if self.use_sdp else 1.0)
+    w_ceil = Tensor.ceil(logw.exp() * x_mask * length_scale)
+    y_lengths = Tensor.maximum(w_ceil.sum([1, 2]), 1).cast(dtypes.int64)
+    y_mask = sequence_mask(y_lengths, y_lengths.max().lazydata.toCPU()).unsqueeze(1).cast(x_mask.dtype)
```
wozeparrot, line 29: We should find a better way to do this.

dosier: So I added a very hacky way to estimate the output sequence length, which avoids realisation up until the end.

Once the model reaches the end, I realise the output, then realise the `y_lengths` buffer, then compare the actual `y_length` to the estimate, and if the estimated length was too short, it will recompute without the estimate, if the estimate was too long it will trim it. 
```py
if max_y_length_estimate_scale is not None:
  length_scaler = o.shape[-1] / max_y_length
  o.realize()
  real_max_y_length = y_lengths.max().numpy()
  if real_max_y_length > max_y_length:
    logging.warning(f"Underestimated max length by {(((real_max_y_length / max_y_length) * 100) - 100):.2f}%, recomputing inference without estimate...")
    return self.generate(g, logs_p, m_p, max_len, None, noise_scale, w_ceil, x, x_mask, y_lengths)
  if real_max_y_length < max_y_length:
    overestimation = ((max_y_length / real_max_y_length) * 100) - 100
    logging.info(f"Overestimated max length by {overestimation:.2f}%")
    if overestimation > 10: logging.warning("Warning: max length overestimated by more than 10%")
  o = o[:, :, :(real_max_y_length * length_scaler).astype(np.int32)]
return o
```
I also added a warning in case the overestimation is too big, this indicates to the user that maybe for their use case they need to modify the `max_y_length_estimate_scale` or provide none entirely.

wozeparrot: This is better than the lazydata usage, but it is slower for me than not doing a length estimation by ~0.05 seconds.

dosier:
> Maybe a comment with `pip3 install inflect phonemizer unidecode eng_to_ipa`?
Oops, I added installation instructions now :). 

> Now getting: `RuntimeError: espeak not installed on your system`...this lib seems hard to get. Do we need it? `brew install espeak` did fix.
So short answer is yes we need it if we care about sound quality.

Longer answer is that each model has "cleaner" functions defined in its config file, these cleaners preprocess text and convert it into phonemes, I added support for various models, many of which come with their own cleaner strategy. 
The vctk and ijs models use a cleaner that depends on the `phonemize` library whose default backend is `espeak`. I tried using the cleaner of another model on the vctk and ljs models, but those cleaers are missing support for certain features, like adding stress, which results in worse sounding models. The current cleaners  for each models are the same ones as originally used by their respective authors. 

> remove the lazydata usage

So some context: the synthesiser has an audio length predictor and a generator. This generator fills an output tensor, the length of this tensor is predicted by the predictor. So in order to create the output tensor we need to realise the predicted length first.
Currently I added a bit of a hack where we can provide an estimate for the output length, then at the end we compare the estimate to the actual output length, if the estimate is larger, we shrink the output tensor, if the estimate is lower, we recompute without the estimation (since at this point we already realised the actual length). 

I don't really like the hack and I did make it optional (can be turned off with `--estimate_max_y_length False`).
I do have some ideas for optimisations that would make the hack less fragile, like batching the input into equal sized chunks. But the variance between models/speakers/input remains hard to estimate nonetheless (which is what the prediction component is for :P)

dosier:
> Testing now
> 
> Maybe a comment with `pip3 install inflect phonemizer unidecode eng_to_ipa`?
> 
> Now getting: `RuntimeError: espeak not installed on your system`...this lib seems hard to get. Do we need it? `brew install espeak` did fix.
> 
> Hehe, it's working! Comment about installation + remove the `lazydata` usage and should be good.

> Maybe a comment with `pip3 install inflect phonemizer unidecode eng_to_ipa`?

Oops, I added installation instructions now :). 

> Now getting: `RuntimeError: espeak not installed on your system`...this lib seems hard to get. Do we need it? `brew install espeak` did fix.
So short answer is yes we need it if we care about sound quality.

Longer answer is that each model has "cleaner" functions defined in its config file, these cleaners preprocess text and convert it into phonemes, I added support for various models, many of which come with their own cleaner strategy. 
The vctk and ijs models use a cleaner that depends on the `phonemize` library whose default backend is `espeak`. I tried using the cleaner of another model on the vctk and ljs models, but those cleaers are missing support for certain features, like adding stress, which results in worse sounding models. The current cleaners  for each models are the same ones as originally used by their respective authors. 

> remove the lazydata usage

So some context: the synthesiser has an audio length predictor and a generator. This generator fills an output tensor, the length of this tensor is predicted by the predictor. So in order to create the output tensor we need to realise the predicted length first.
Currently I added a bit of a hack where we can provide an estimate for the output length, then at the end we realise the actual predicted length, and compare it to the estimate, if the estimate is larger, we shrink the output tensor, if the estimate is lower, we recompute without the estimation (since at this point we already realised the actual length). 

I don't really like the hack and I did make it optional (can be turned off with `--estimate_max_y_length False`).
I do have some ideas for optimisations that would make the hack less fragile, like batching the input into equal sized chunks. But the variance between models/speakers/input remains hard to estimate nonetheless (which is what the prediction component is for :P)

examples/vits.py
dosier: True, for smaller inputs its actually a tad slower, for larger inputs it should be a lot faster tho.

dosier: Also `Y_LENGTH_ESTIMATE_SCALARS` for larger inputs can be set lower, there seems to be a logarithmic relation between input size and the length, but for smaller inputs it varies a lot depending on the speaker and symbols.  For larger inputs it seems to converge to about `1.56` for the vctk model.

dosier:
@geohot PayPal: stanvdbend@gmail.com

geohot:
Sent, thanks!

PR #1191, opened 2023-07-08
Fix pow to work with negative tensors

### LINKED ISSUE #1151: Negative tensors taken to power 3 or greater return nan
- Issue: Negative tensors taken to power 3 or greater return nan. Does not match with PyTorch output
- Problem: In tensor.py -> def_pow, function tries to take log of a negative number. Log of a negative number = nan
-Might have been caused by the recent merge "Remove POW llop and add SQRT llop". 

Test code compared with PyTorch:
```
print(torch.tensor(-2).pow(3))
print(Tensor(-2).pow(3).numpy())
-> tensor(-8)
-> nan
```

"def pow" code: https://github.com/geohot/tinygrad/pull/1104/files
tensor.py -> def_pow code:
```
def pow(self, x:Union[Tensor, float], reverse=False) -> Tensor:
    if x.__class__ is not Tensor and not reverse:
      # simple pow identities
      if x < 0: return (1.0/self).pow(-x)
      if x == 2.0: return self*self
      if x == 1.0: return self
      if x == 0.5: return self.sqrt()
    return self.log().mul(x).exp() if not reverse or isinstance(x, Tensor) else self.mul(log(x)).exp()
```

madt2709:
[Fixes #1151](https://github.com/tinygrad/tinygrad/issues/1151).

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -218,6 +218,9 @@ class TestOps(unittest.TestCase):
     helper_test_op([(45,65), (45,65)], lambda x,y: x**y, Tensor.pow, a=0)
     helper_test_op([()], lambda x: x**2, lambda x: Tensor.pow(x,2), a=0)
     helper_test_op([()], lambda x: x**-2, lambda x: Tensor.pow(x,-2), a=0)
+    # Regression tests for https://github.com/tinygrad/tinygrad/issues/1151
+    helper_test_op([(45,65)], lambda x: x**3, lambda x: Tensor.pow(x,3), a=-10)
+    helper_test_op([()], lambda x: x**3, lambda x: Tensor.pow(x,3), a=-10)
   def test_pow_const(self):
     helper_test_op([(45,65)], lambda x: x**1.0, lambda x: x**1.0)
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         7               2
```

madt2709:
Marking as draft whilst I add a couple more test cases + potentially make code more readable.

TomFinet:
Note: `x = Tensor([-2]).pow(-2.3).numpy()` outputs `array([0.1193575], dtype=float32)`, while the same in pytorch, `y = torch.tensor([-2]).pow(-2.3)` outputs `tensor([nan])`. The current tests don't check for non-integer exponents.

geohot:
@TomFinet can you fix this?

wozeparrot:
So right now for negative bases to non-integer exponents it returns the real part of the complex answer, I'm not sure if there is a good way to test if the exponent is non-integer without something that might break the backwards pass.

geohot:
Hmm that might just be fine. I'm going to merge this, it's an improvement of what's there

PR #1192, opened 2023-07-08
Fix bug where Tensor.randn returns inf

sekstini:
Tensor.randn can return ±inf in the rare case of our RNG returning a 0. The cause is the log function in the Box-Muller transform we use for randn.

PyTorch also uses Box-Muller, but avoid this issue by mapping the RNG value from `[0, 1) -> (0, 1]`. This PR implements that fix. [[src]](https://github.com/pytorch/pytorch/blob/91dcc3b2726aa33ca9a30dfb8d165d8ec7b7bba3/aten/src/ATen/native/cpu/DistributionTemplates.h#L142)

---

Since zeros values are extremely rare, it's a bit of a pain to test. I mined a seed that would reproduce the bug on the current implementation, but I didn't add it because it seems too specific. Let me know if you want me to add it, or if you have a better test in mind.

```python
def test_randn_isnt_inf(self):
  Tensor.manual_seed(195866)
  a = Tensor.randn(128)
  assert not np.any(np.isinf(a.numpy()))
```

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -145,7 +145,8 @@ class TestTinygrad(unittest.TestCase):
         b = random_fn(10,10).realize()
         np.testing.assert_allclose(a.numpy(), b.numpy())
 
-  def test_randn_isnt_inf(self):
+  def test_randn_isnt_inf_on_zero(self):
+    # simulate failure case of rand handing a zero to randn
     original_rand, Tensor.rand = Tensor.rand, Tensor.zeros
     try: self.assertNotIn(np.inf, Tensor.randn(16).numpy())
     except: raise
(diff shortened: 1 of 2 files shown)


geohot:
Needs a test that was failing before change and now passes

tinyb0t:
Changes made in `tinygrad/`:
```
```

sekstini:
This test is kind of ugly, but less likely to break than a mined seed.
Let me know if you prefer this or the one from the description above.

sekstini:
Alternatively we could make a test doing a fixed number of loops checking for infs and mark it as SLOW.

Generated (4096, 4096) matrices had a very high chance of containing infs, so ~20 iterations (less than 10s), would be pretty accurate.

geohot:
I think that's fine, cute. Thanks, will merge when tests pass!

PR #1200, opened 2023-07-09
WIP: < 75 ms/step on 7900XTX: examples/hlb_cifar10.py while keeping 90% (float16/bf16 okay)

kunwar31:
So far, changed linearizer to do calculations based on buffer dtypes, and also made grouping generic to vector dtypes 
```
a = Tensor.rand(10,10, dtype=dtypes.float16) # also works for int32
(a+a).realize()
```
gives a nice kernel output
```
#pragma OPENCL EXTENSION cl_khr_fp16 : enable
__kernel void E_25_4(__global half* data0, const __global half* data1) {
{ int gidx0 = get_group_id(0);  /* 25 */
    half4 val1_0 = (half4)(*((__global half4*)(data1+(gidx0*4))));
    half4 alu0 = (val1_0+val1_0);
    *((__global half4*)(data0+(gidx0*4))) = (half4)alu0;
  } /* global+local */
}
```
The Dtypes could be cleaned up a bit.
Need to test actual training with float16 tensors


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py            17              18
tinygrad/helpers.py                       12               3
tinygrad/runtime/ops_gpu.py                1               1
total                                     32              24
```

kunwar31:
Still need to test other backends, only tested with OpenCL. 
CI for OpenCL fails because 
`91:1:26: warning: unsupported OpenCL extension 'cl_khr_fp16' - ignoring`

geohot:
I'll get you a computer with a 7900XTX, give me a few hours.

kunwar31:
Progress summary: 
1. dtype used in all kernels used to be float32: changed this to use dtypes based on buffers 
    - Tests pass for:
       - GPU (locally, cl_khr_fp16 needs some workaround for CI)
       - CLANG
       - CPU
       - CUDA
       - TORCH
   -  Fails for
      - webgpu  
      - LLVM
      - Metal 
    - have kept accumulate vars in float32, can be controlled with ACCUM_FLOAT env var (because accumates in half might overflow)
2. Backward pass for cifar with HALF - but the training diverges
    - This is a known problem with fp16, need to do mixed precision  https://arxiv.org/pdf/1710.03740.pdf 

Added HALF to test_speed_v_torch, though there are variations b/w runs heres a summary (similar fp16 numbers with ACCUM_FLOAT=1 and 0)
```
add                             8192 x 8192    4.83 ms  (   13.89 GFLOPS   166.66 GB/s) in tinygrad fp32  805.31 MB
exp                             8192 x 8192    4.23 ms  (   31.72 GFLOPS   126.87 GB/s) in tinygrad fp32  536.87 MB
gemm                            2048 x 2048    2.13 ms  ( 8048.53 GFLOPS    23.58 GB/s) in tinygrad fp32   50.33 MB
gemm                            4096 x 4096    12.82 ms (10718.67 GFLOPS    15.70 GB/s) in tinygrad fp32  201.33 MB
conv bs: 32 chans:128 -> 128 k:1               6.15 ms  ( 2793.23 GFLOPS    87.30 GB/s) in tinygrad fp32  536.94 MB
conv bs: 32 chans:128 -> 128 k:3               9.76 ms  (15843.48 GFLOPS    55.94 GB/s) in tinygrad fp32  545.91 MB

add                             8192 x 8192    0.78 ms  (   86.37 GFLOPS   518.25 GB/s) in tinygrad fp16   402.65 MB
exp                             8192 x 8192    0.35 ms  (  378.64 GFLOPS   757.28 GB/s) in tinygrad fp16   268.44 MB
gemm                            2048 x 2048    1.21 ms  (14163.54 GFLOPS    20.75 GB/s) in tinygrad fp16    25.17 MB
gemm                            4096 x 4096    5.89 ms  (23322.78 GFLOPS    17.08 GB/s) in tinygrad fp16   100.66 MB
conv bs: 32 chans:128 -> 128 k:1               1.08 ms  (15927.31 GFLOPS   248.89 GB/s) in tinygrad fp16   268.47 MB
conv bs: 32 chans:128 -> 128 k:3               13.10 ms (11806.21 GFLOPS    20.84 GB/s) in tinygrad fp16   272.96 MB
```
  3x3 convs are somehow slower in FP16
  
3. Running cifar in HALF (even with ACCUM_FLOAT=0) isn't acutally faster currently
     - time is still ~150ms / step
     
What I'm working on:
 - with OpenCL, there's no control on what assembly instructions it emits. DOT2ACC instructions in RDNAv3  are what actually enable 2x fp16 throughput. maybe HIP backend will emit DOT2ACC instructions, or we need to get to into RDNA assemler.   
 - Its know that using vector dtypes is faster in OpenCL. try to increase use of vector dtypes in kernels, maybe improve some hand_coded_optimizations.    
 - Mixed precision training   
 - getting tests to pass on webgpu, LLVM, Metal

tysam-code:
Great to see this moving, happy to see this work have an impact! Hope to do another release in the next few weeks or so.

I'll update if I see anything here, but Re: "Backward pass for cifar with HALF - but the training diverges"

There's a note that says _This is a known problem with fp16, need to do mixed precision https://arxiv.org/pdf/1710.03740.pdf_ It's true that you can do mixed precision, but the casts are a bit dicey and add some extra complexity. I don't know if bfloat16 is supported in tinygrad but it adds a lot of stability for a tiny speed hit (and is often simpler/faster than mixed precision IIRC)

I don't know if that helps, the code looks good from here and some brief browsing. I spent about a month banging my head against the wall trying to do a byte-for-byte rewrite of David Page's code, so I guess it's a rite of passage!

Happy to share any technical quirk tips or whatever, I'm in the thread now so I should get pinged if you have any thoughts/questions.

Good stuff! <3 :)))) :D 🎆 🐧

tysam-code:
Okay, I've done a little more thorough of a review, and hopefully this is helpful! Format is diffs [line: diff+notes] + reasons/extra thoughts behind it. It might be a bit overwhelming, but hopefully this answers any major questions and helps point out where a lot of the training instability is very likely coming from. :')))) <3 :D :D :) I don't know the ins and outs of tinygrad so if at all I have some horridly fruitless ignorance here, please excuse me (and we can leave it on the big ol' TODO pile for later! :D <3 <3 <3 <3 :)))) )

Diffs:
    --l46 couple of .numpy() calls, which looks like it might bottleneck things a tad w/ CPU moves. I don't know if there's a special pass-by-reference happening here to the Tensor class, but if it's a copy then that's going to be quite slow (especially since we do use this function ~30-40+ times or so during training.
    --l56 track_running_stats=False -> True (!!!!!), eps 1e-4 -> 1e-12, momentum 0.8 consider -> 0.5
    --l57 relu -> gelu is important, the full version too with the CDF and not the fast approximation. Both gradient stability and training speed. Gelu gives good gradient coverage as well a decent boost to expectations + training stability (ironically before selu, the OG, came out, iirc :'OOOO). Yay for self-normalizing, gated units! :)
    --l59 more of a 'novel invention'-oriented kind of note, but how you manage your buffers here at the lower levels is where we have some room for improvement. A lot of separate operations that maybe with some tiny changes could be fused together nicely in a way that's quite snappy.
    --l61 id personally move the max pool to a new line as its sorta easy to miss, but hybrid is not the end of the world. basically, i think having an expectation of a novel operator on each line or all at once is a good one (since this code layout seems to set an expectation of one per line). just minor 'bug surface for the casual dev' kind of thought. Hotz I think likes chained commands so maybe in this repo slopping lines 61-65 into a long chained expression in 61 is okayish (and logically does one 'block chunk' per line)
    --l62 reminder to eventually remove casts, hopefully these other changes will stabilize it a bit.
    --rest of the function looks good! :D :)
    --l76 ! **whitening**. this is probably why training is unstable, not the float16 issues. Whitening can seem hard but it's really not too terrible if you transcribe the code. The earliest layers of the network are always the most important for training stability. Basically, always always.
    --l79 kernel size of 1 will probably, uh, not do too much for you! That's basically BASE_DIM vectors of size 3, which more than spans the input space, so its more or less redundant and the training thrash isn't worth the extra expressivity. It's sorta the whitening thing -- a very unfun chore, but necessary. LMK and we can get on a voice call and pair program this section together if you want. It's really worth it! :D :))))
    --l77 BASE_DIM, Looks good, though I think we can trim this up a lot! I'd just steal whatever the dims are from the latest hlb-cifar10 (v0.7.0 coming out...soonish....), since network dimensions can be a huge limiting factor in speed. Once you get the ratios right, then using a scaler against the semi-hardcoded base values becomes more effective, as it seems that that kind of scaling works, er, generally enough-ish. 64 Is pretty overkill for a base dim, ive gotten away with much less and that might help with speed/overfitting/etc! :D :))))
    --l83 relu occurence
    --l92(93 and 101) log_softmax can be a bit funky sometimes, so just keep this in mind. I'd recommend returning the raw unnormalized logits instead of the log_softmaxed ones since that's a bit more standard (and can avoid some other issues down the road....)
    --l151 Small reminder on the batchsize -- it can be easier to get higher accuracies due to more steps, pushing this to 1024 (or whatever it is live, I think that's it....) can help tho if you copy hyperparams. The others look pretty okay to me, though pct_start is extremely agressive in this version (I may have had it that aggressively in the past, can't recall, but .2 is a good safe-ish number to use. See latest version for more info). Momentum I can't remember exactly where it is, but that's a somwhat sensitive parameter, I think it's .85 in the latest version IIRC? WD should be an exact value not a rounding as it's quite sensitive. Label smoothing on and at .20 (very important for stability!)., and cutmix >= mixup as A. it's easier (can just slice out different pixels instead of blending, and can just...reference swap in the right scenarios?!?!?), and B. performance!, so I'd go with the cutmix_epochs and cutmix_size parameters if possible. Whew, that's a lot! All very important though! <3 :)))) :D :D
    --l176 Don't know if the lr hardcode here is critical or not, but might as well set it to something like max_lr*final_lr_ratio for posterity/reducing accidental bug fallthru status
    --l183 highly highly highly recommend doing this before doing any network training at all! It's not counted in the training time as it's basically part of the dataset download and prep, and only should be done once (before the dataset is saved to disk in an easy-and-fast-to-load-format. Free speed right 'dere ! :)))) Remember, if we think about fusing as much as possible, doing it in a pre-epoch and/or pre-training step is a great way to help ourselves out. :))))
    --l188 This almost looks like entropy regularization (https://arxiv.org/abs/1701.06548) instead of label smoothing. While ER is my first love, LS is what the repo came with and was tuned around -- it works pretty well! Maybe some benefits to be had, but I'd save the swap for later. (I believe it should be out.mul(1-Y).mul(-LABEL_SMOOTHING).mean() in the last factor IIRC).
    --l189 FP32 vs FP16 debugging
    --l207 same as before w.r.t. normalization.
    --l208 Keeping the running stats here is what lets the model perform well in eval, ideally.
    --l209 maybe micro op is a gather vs a mul, would also save having to make one hots (I guess I should update the base repo for this too)
    --l226 If you create a virtual array w/ cutout and do it before the epoch, you can avoid having a left- and right-batcher. little bit less data flying overhead to and from things
    --l235 per same caveats before, appears to be a device-swap thrash. all-gpu and only-slicing is good! :D :))))
    --l243 oddly enough I would return last_eval over best_eval, possibly simply for the reason that the last_eval should be the best eval, it simplifies things, and oftentimes eval performance near the end is gaussian with noise, so when it's done best, returning the last should actually give better performance overall instead of cherrypicking based on eval set noise (basically, our eval performance is distributed in a gaussian way, and we don't want to arbitrarily bias the stats in our favor unnecessarily. Since every step counts, we want the model that's used every step, and for that to be as good as possible! <3 :)))))
    
And that should be about it! Hope this helps, lots of little technical notes, but I can assure you that on many, many, many of these points I have spent countless silent hours forming a hole in the wall with my head.

Happy to answer any questions as necessary. <3 :)))) :D :))))) :D <3

kunwar31:
@tysam-code thank you for the detailed notes! unfortunately, this PR is a bit messed up, it has a lot of code we don't really need now. Also, we've had @g1y5x3 clean up a lot of this code already, and it reaches 94% in 1000 steps! it is already merged in [master](https://github.com/tinygrad/tinygrad/blob/master/examples/hlb_cifar10.py).
on top of his changes, fp16 were minimal. 
here's the [diff](https://github.com/kunwar31/tinygrad/compare/math_in_buf_dtype...kunwar31:tinygrad:cifar-fp16?diff=unified
)
essentially doing batch norm in fp32, changed loss scaler, and label smoothing to 0.1 and got me
```
999   90.23 ms run,    4.23 ms python,   86.00 ms CL,  289.50 loss, 0.000000 LR, 3.73 GB used,   7473.38 GFLOPS
eval 9413/10000 94.13%,    0.29 val_loss STEP=1000
```
I'm gonna close this PR and create a cleaner one with only the required changes.
one of them for doing math in same dtype as buffer type : https://github.com/tinygrad/tinygrad/pull/1545
after that's merged, then another PR based on [this branch](https://github.com/kunwar31/tinygrad/blob/cifar-fp16/examples/hlb_cifar10.py)

PR #1203, opened 2023-07-09
Re-open Train maskrcnn after the big file structure changes

g1y5x3:
Not sure if what I did was the correct way. Had to delete the commits to sync on my fork, and then I pull to resolve some specific file issues. Anyway, repost #1125 

- [x]  load and index through training (image, target) pair

- [x] refactor the `coco.py` to separate the train and eval 

- [ ] modularize models/mask_rcnn.py for better readability as well as remove some of the dependency on torch backend (mostly transforms)

- [ ] find a solution for doing advanced indexing on Tensor directly

- [ ] add loss function into RPN, ROIHead

- [ ] perform training experiments


geohot:
No progress in a week, closing.

PR #1205, opened 2023-07-09
create cumsum speed test

zk-tarts:
I added a test for speed of cumsum since there are optimized algorithms for this problem (parallel prefix sum) and tinygrad is currently slower. Maybe there is interest in making it faster?  Not sure if cumsum is all that important or common.

Ran into a [problem in CI](https://github.com/zk-tarts/tinygrad/actions/runs/5483343759) with some backends crashing trying to allocate a ton of memory. Probably happening when they do expands as part of [this conv](https://github.com/tinygrad/tinygrad/blob/master/tinygrad/tensor.py#L480C50-L480C56).

### DIFF
--- a/test/test_speed_v_torch.py
+++ b/test/test_speed_v_torch.py
@@ -156,6 +156,12 @@ class TestSpeed(unittest.TestCase):
     def f(a, b): return a.reshape(int(4096//R), int(4096*R)).sum(axis=1)
     helper_test_generic_square('partial_sum', 4096, f, f, onearg=True)
 
+  def test_cumsum(self):
+    def f0(a,b): return a.cumsum(axis=0)
+    def f1(a,b): return a.cumsum(axis=1)
+    helper_test_generic_square('cumsum_0', 1024, f0, f0, onearg=True)
+    helper_test_generic_square('cumsum_1', 1024, f1, f1, onearg=True)
+
   def test_array_packing(self):
     N = 2048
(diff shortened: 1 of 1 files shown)


geohot:
Yea...the implementation right now is really bad. Have a proposal to fix it?

geohot:
Tests failing, but I would merge this if not

ZenginU:
Hello @zk-tarts,

I am new to opensource programming so sorry in advance for mistakes. I do not know if you are still working on this but if you are working on this. I believe that problem is with the [_pool](https://github.com/tinygrad/tinygrad/blob/5a8ad571634375ee3fcc4f5cc31c51ea1e3dcb64/tinygrad/tensor.py#L462C30-L462C35) in the conv2d.

To briefly explain, in the conv2d function input of 1024,1024 is converted to shape of (1, 1, 1024, 1024, 1, 1024) at some point.

@geohot 
Apart from the speed test, normal test on cumsum is failing on my M1 for the (1024, 1024) size tensor. I do not know if its a big deal or it might just be floating point error for big nums because of the conv2d.

`helper_test_op([(1024, 1024)], lambda x: torch.cumsum(x, dim=0), lambda x: Tensor.cumsum(x, axis=0), atol=1e-6)`

```
Mismatched elements: 96 / 1048576 (0.00916%)
Max absolute difference: 9.1552734e-05
Max relative difference: 0.13173492
```

PR #1206, opened 2023-07-09
adding `param_group` to `Optimizer`

KayneWest:
Using the recommendations from https://github.com/tinygrad/tinygrad/pull/1147. I'm proposing adding `param_groups` to the `Optimizer` so that the user can augment the parameters individually. The syntax would be similar to `torch`. This change functionally accomplishes what I had been using with `lr_mult` before

`torch`
```
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.layer1 = nn.Linear(1, 1)
        self.layer1.weight.data.fill_(1)
        self.layer1.bias.data.fill_(1)
        self.layer2 = nn.Linear(1, 1)
        self.layer2.weight.data.fill_(1)
        self.layer2.bias.data.fill_(1)
    def forward(self, x):
        x = self.layer1(x)
        return self.layer2(x)
net = Net()
optimizer = optim.Adam([
            {'params': net.layer1.weight},
            {'params': net.layer1.bias, 'lr': 0.01, 'weight_decay': 0.001},
            {'params': net.layer2.weight, 'lr': 0.001}
        ], lr=0.1, weight_decay=0.0001)
```
`tinygrad`
```
net = ResNet50()
parameters = get_parameters(net)
optimizer1 = Adam(
    [
        {'params': parameters[:20], 'lr': 0.001, 'wd': 0.001},
        {'params': parameters[20:-3], 'lr': 0.01, 'eps': 1e-2},
        {'params': parameters[-3:]}
    ], lr=0.1)
optimizer2 = Adam(parameters, lr=0.1)
```

### DIFF
--- a/tinygrad/nn/optim.py
+++ b/tinygrad/nn/optim.py
@@ -9,10 +9,10 @@ class Optimizer:
 
     param_groups = list(params)
 
-    if not isinstance(param_groups[0], dict):
-      param_groups = [{'params': param_groups}] #default lr
+    if not isinstance(param_groups[0], dict): 
+      param_groups = [{'params': param_groups}]
     
-    for param_group in param_groups:
+    for param_group in param_groups: 
       self.add_param_group(param_group)
     
     self.lr = Tensor([lr], requires_grad=False)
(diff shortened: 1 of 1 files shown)


geohot:
Tons of lines. What's the downside to using two optimizers?

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/optim.py                      59              45
```

KayneWest:
@geohot I know, I know, I don't like the line count either. I had misinterpreted that you were leaning towards `torch` syntax. hence this pr. I guess there really isn't a downside if you're used to running multiple optimizers. But having kind of a nice grouper/lr dampener is great when you're doing a bunch of experiments -  so something like the one below or something;

```
class OptimizerGroup:
  def __init__(self, optimizers): self.optimizer_groups = optimizers
  def zero_grad(self): (optimizer.zero_grad() for optimizer in self.optimizer_groups)
  def step(self): (optimizer.step() for optimizer in self.optimizer_groups)

groups = []
groups.append(SGD([net.layer1] lr=0.1))
groups.append(SGD([net.layer2] lr=0.2))
groups.append(SGD([net.layer3] lr=0.3))
groups.append(SGD([net.layer4] lr=0.4))
groups.append(SGD([net.feature_detector] lr=0.0001))
group = OptimizerGroup(groups)
group.step()

```

geohot:
Not going to merge, see above. I think this feature is worth about 3 lines, and *only* if they sit above the Optimizer. And should probably be used in examples before migrating to core tinygrad.

KayneWest:
Sounds good. Hopefully third times the charm.

PR #1213, opened 2023-07-10
[perf] use optimize=True in np.einsum

carsonRadtke:
local testing (test/test_net_speed.py) yields:
```
# optimize=False (default)
forward pass:  185.863 ms, 7.83x off baseline 23.746 ms
backward pass: 820.215 ms, 20.74x off baseline 39.543 ms

# optimize=True (pr change)
forward pass:  157.368 ms, 6.85x off baseline 22.983 ms
backward pass: 209.347 ms, 5.66x off baseline 36.967 ms
```

central tendency for `optimize=True` is around 8x and 20x. `optimize=False` is around 7.5x and 5.5x. I could compile an n=1000 analysis if that would be helpful in analyzing the change.

Reading:
 -  https://numpy.org/doc/stable/reference/generated/numpy.einsum.html
 -  https://stackoverflow.com/questions/49265044/numpy-einsum-path-differences-and-the-optimize-argument


tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
Looks better, and it's a small sample size but CPU CI ran 2 minutes faster on this PR than master. Nice!

PR #1216, opened 2023-07-10
[READY] style: no-else-return

roelofvandijk:
Some unnecessary of `else`/`elif` after returns. Could be a matter of taste.

I enabled the pylint option, that is optional,of course.


geohot:
Usually I'm pretty against these sort of PRs, but going to merge this one because it enforces it with a linter rule.

PR #1217, opened 2023-07-10
style: remove deprecated type aliases

roelofvandijk:
These type aliases are deprecated. I only did /tinygrad.


roelofvandijk:
Only works >3.8.

PR #1223, opened 2023-07-10
Fix LazyBuffer SHUFFLE_PAD_OPS to prevent invalid pad movement

### LINKED ISSUE #1221: Bug: NaN when using `cat` on tensor with unrealised `log` op
Tries to take the log of zero due to merging of the ops, resulting in NAN values. 

To reproduce:

```
x0 = Tensor([1])
x1 = Tensor([1]).log()    # Can realize `x1` before calling `cat` to circumvent
print(x0.cat(x1).numpy()) # [nan  0.]
```

Kernel it generates:
```
#include <metal_stdlib>
using namespace metal;
kernel void E_2(device float* data0, const device float* data1, const device float* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
{ int gidx0 = gid.x;  /* 2 */
    float val1_0 = ((gidx0<1)) ? (float)(data1[0]) : 0.0f;
    float val2_0 = (((gidx0*-1)<0)) ? (float)(data2[0]) : 0.0f;
    float val3_0 = (((gidx0*-1)<0)) ? (float)(0.6931471805599453f) : 0.0f;
    float alu0 = log2(val2_0);
    float alu1 = (alu0*val3_0);
    float alu2 = (alu1+val1_0);
    data0[gidx0] = alu2;
  } /* global+local */
}
```

flammit:
In addition to div, any ops that will generate non-zero outputs from zero inputs need to be guarded.

Closes #1221

### DIFF
--- a/test/test_lazybuffer.py
+++ b/test/test_lazybuffer.py
@@ -2,6 +2,7 @@
 import numpy as np
 import unittest
 from tinygrad.lazy import LazyBuffer
+from tinygrad.tensor import Tensor
 
 class TestLazyBuffer(unittest.TestCase):
   def test_fromcpu_buffer_sharing(self):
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                           3               2
```

geohot:
Thanks, good catch

Issue #1178, opened 2023-07-08
Fix constant folding for Tensor([3])


geohot:
I believe now 1-D numpy arrays aren't being constant folded, they should be.

Like `Tensor(3)` works but `Tensor([3])` doesn't.

We should also add a test to prevent this regression in the future.

wozeparrot:
Can be closed now.

chenyuxyz:
reopening this

The issue is that `a = Tensor(3).reshape(1)` is a reshape of 0 dim Tensor, thus different from `b = Tensor([3])`.

reshape itself is easy to solve, but need some thinking about all the possible cases that a non-load op can/need to create a buffer

PR #1227, opened 2023-07-11
Fix constant folding for Tensor([3])

leastcompressible:
#1178

### DIFF
--- a/docs/abstractions.py
+++ b/docs/abstractions.py
@@ -120,7 +120,8 @@ from tinygrad.tensor import Tensor
 from tinygrad.ops import LazyOp, BinaryOps, LoadOps
 
 # the 2+3 from before
-result = Tensor([2]) + Tensor([3])
+# added some 0s, otherwise Tensor([2]) will be folded into a constant without using LoadOps.FROM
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         4               2
```

docs/abstractions.py
```diff
@@ -120,7 +120,8 @@ class LoadOps(Enum):     EMPTY = auto(); RAND = auto(); CONST = auto(); FROM = a
 from tinygrad.ops import LazyOp, BinaryOps, LoadOps
 
 # the 2+3 from before
-result = Tensor([2]) + Tensor([3])
+# added some 0s, otherwise Tensor([2]) will be folded into a constant without using LoadOps.FROM
+result = Tensor([2, 0]) + Tensor([3, 0])
```
wozeparrot, line 124: Can you test if adding `.realize()` instead of the 0s will force it into a buffer?

leastcompressible: This part of the doc inspects the ops in lazydata before realizing, so we should not realize it early

PR #1231, opened 2023-07-12
simplify _ast_reduceops

sheganinans:
dunno if this would be considered code golf, but i think it reads better. 

open to renaming the variable name to something other than `arg`.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                           4               6
```

PR #1232, opened 2023-07-12
Small optimization in _Device.__getitem__

jdlib:
No need to call `inspect.getmembers(importlib.import_module())` every time `_Device.__getitem__` is called, cache the runtime object instead.
Also renamed `_Device._buffers` to `_Device._names`.


geohot:

PR #798, opened 2023-05-25
Promote Embedding to `nn`

wozeparrot:
Embedding is used in a few models now (RNN-T and LLaMA) and will be used in BERT and other language models as well. So we should promote it to nn.

Rewrote it to no longer rely on numpy which means that this can now be JIT compiled.

Added a test to see if it matches torch.

PR #1239, opened 2023-07-13
Quadric Layer in nn

diro5t:
introduces 'quadric learning' - second order hypersurface decision boundaries in quadric layers and has the proven potential to reduce model size significantly and therefore help to democratize and decentralize access to ai - based on my research and publication from back in the day (see example and repo for more background)

clean few new lines in tinygrad.nn.__init__.py

example: https://github.com/diro5t/deep_quadric_learning/blob/main/TinyQuadMNIST.ipynb

### DIFF
--- a/tinygrad/nn/__init__.py
+++ b/tinygrad/nn/__init__.py
@@ -70,6 +70,18 @@ class Linear:
   def __call__(self, x):
     return x.linear(self.weight.transpose(), self.bias)
 
+class Quadric:
+  def __init__(self, in_features, out_features, bias=True):
+    self.qweight = Tensor.kaiming_uniform(out_features, in_features, a=math.sqrt(5))
+    self.lweight = Tensor.kaiming_uniform(out_features, in_features, a=math.sqrt(5))
+    bound = 1 / math.sqrt(self.lweight.shape[1])
+    self.bias = Tensor.uniform(out_features, low=-bound, high=bound) if bias else None
+
+  def __call__(self, x):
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/__init__.py                   12               0
lines added in the tinygrad folder: 12
```

wozeparrot:
Usually we only promote layers into `nn` if they have been used in at least 3 models/very commonly used layers (like `Linear` or `Embedding`).

diro5t:
i get that, but you say 'usually' - did you look at the example - this is so high level it can reduce parameters in all kinds of model paradigms - higher order neurons have been a subject in research not only by me (although i am one of the earliest)

does this mean there will be no further consideration/review of this feature?
thanks

geohot:
Unless it’s a very common operation, the 3 model rule stands.

diro5t:
could you please explain the 3 model rule a little further? i showed that significant model size reduction is possible in a feed forward  classifier. if i can show the same effect let’s say in a gan and in a 3rd paradigm that would fulfill it? thanks

wozeparrot:
One of the main points of tinygrad is having a nice compact core, even if the 1000 line rule doesn't exist anymore. The 3 model rule is basically that a layer should be used in at least 3 popular models implemented in tinygrad, like how `Embedding` was [promoted](https://github.com/tinygrad/tinygrad/pull/798).

diro5t:
Thanks for the clarification. Cheers and keep up the good work

PR #1240, opened 2023-07-13
display sass for both cuda code and ptx

NotNotOddity:
nvcc --cubin error's when trying to compile PTX and can be skipped.

### DIFF
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -53,7 +53,7 @@ else:
 class CUDAProgram:
   def __init__(self, name:str, prg:str, binary=False):
     try:
-      if DEBUG >= 6:
+      if DEBUG >= 6 and not getenv("PTX"):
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

NotNotOddity:
More detailed information: 
Initially, when using both DEBUG=6 and PTX=1, the program encountered an issue due to the `prg` parameter being a PTX string. It was trying to compile the PTX to print the SASS. At first, I simply disabled the SASS output when using the PTX environment variable. 

Instead, I updated it to output SASS regardless of whether you pass PTX or CUDA to the `prg` parameter. This modification not only resolved the error I was encountering but also enables the program to print the SASS now for PTX. 

I am unsure about how to add a test that demonstrates this issue without introducing PTX to CI and testing it that way.

geohot:
Cool! On phone, will review when I get laptop but seems like a good feature. Someday we will generate SASS

PR #1243, opened 2023-07-14
Added `nn.ConvTranspose1d`

dosier:
Used in #1188


wozeparrot:
I think this is small enough to be worth adding, it also makes Conv1d have a matching transpose layer like Conv2d has.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/__init__.py                    3               0
```

PR #1245, opened 2023-07-15
Fix Metal tests on CI

akretz:
I could reproduce the failing Metal tests on the CI on a local Intel Mac, so I guess the failures are specific to this platform. I have narrowed it down to this: Let's consider the `TestHalfDtype.test_half_to_int8` test, which creates the following kernel.
```
kernel void E_4(device char* data0, const device half* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
    volatile float4 val1_0 = (float4)(*((device half4*)(data1+0)));
    data0[0] = val1_0.x;
    data0[1] = val1_0.y;
    data0[2] = val1_0.z;
    data0[3] = val1_0.w;
   /* global+local */
}
```

This test fails in the current master; without having looked at the generated assembly code, it appears to me that the compiler for some reason optimizes away the `val1_0` intermediate variable and only does the `data0[0] = val1_0.x;` assignment and skips the others. Adding the `volatile` keyword to all loads seems to fix this, because it forces the compiler to not remove the intermediate variables.

It's not enough to force it on vector types only; I still had `TestOps.test_output_padded_conv_transpose2d` failing in that case. Forcing it on integer buffers makes the `TestInt8Dtype` tests fail, because as far as I understand we here rely on the compiler not cast to an intermediate float variable and back into an int buffer.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py             2               2
tinygrad/runtime/ops_metal.py              3               2
total                                      7               5
```

geohot:
Hmm, why is this happening? It would surprise me if volatile on the intermediates like that is the real fix, this is consistent with a cache issue.

akretz:
Yes it doesn't seem to be the real fix. Randomly adding memory barriers or disabling loop unrolling also makes the tests succeed I've found. But these are all just ugly workarounds.

PR #1247, opened 2023-07-15
[fix] llama.py crashes w/ --profile --count < 3

carsonRadtke:
llama.py has an option to run the profiler on one of the tokens that gets generated, but it seems the profiler wants to run only on the third token. If the users specified `--count x < 3`, the collection of the profile panics and trace is printed (as can be seen below). This MR changes the behavior to profile the first token (if it exists). 

```
carson@Carsons-MBP tinygrad % python3 -O examples/llama.py --prompt "Hello." --profile --
...
 ITraceback (most recent call last):
  File "/Users/carson/code/tinygrad/examples/llama.py", line 423, in <module>
    stats = pstats.Stats(profiler)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/pstats.py", line 115, in __init__
    self.init(arg)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/pstats.py", line 129, in init
    self.load_stats(arg)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/pstats.py", line 155, in load_stats
    raise TypeError("Cannot create or construct a %r object from %r"
TypeError: Cannot create or construct a <class 'pstats.Stats'> object from <cProfile.Profile object at 0x1265c8fa0>
carson@Carsons-MBP tinygrad % 
```

### DIFF
--- a/examples/llama.py
+++ b/examples/llama.py
@@ -393,7 +393,7 @@ After you are done speaking, output [EOS]. You are not Chad.
 
     last_break = len(outputted)
     for i in range(args.count):
-      if args.profile and i == 2: profiler.enable()
+      if args.profile and i == 0: profiler.enable()
 
       if args.timing: print("")
       st = GlobalCounters.time_sum_s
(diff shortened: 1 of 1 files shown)


geohot:
Hmm, that's sort of by design. I don't want to profile the warmup

PR #1249, opened 2023-07-16
Moved mkdir in `utils.download_file` to diff line

dosier:
`.mkdir` does not return the actual directory being created.

### DIFF
--- a/extra/utils.py
+++ b/extra/utils.py
@@ -39,8 +39,7 @@ def download_file(url, fp, skip_if_exists=True):
   r = requests.get(url, stream=True)
   assert r.status_code == 200
   progress_bar = tqdm(total=int(r.headers.get('content-length', 0)), unit='B', unit_scale=True, desc=url)
-  path = pathlib.Path(fp).parent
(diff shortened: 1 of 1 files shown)


extra/utils.py
```diff
@@ -39,7 +39,9 @@ def download_file(url, fp, skip_if_exists=True):
   r = requests.get(url, stream=True)
   assert r.status_code == 200
   progress_bar = tqdm(total=int(r.headers.get('content-length', 0)), unit='B', unit_scale=True, desc=url)
-  with tempfile.NamedTemporaryFile(dir=pathlib.Path(fp).parent.mkdir(parents=True, exist_ok=True), delete=False) as f:
+  path = pathlib.Path(fp).parent
+  path.mkdir(parents=True, exist_ok=True)
```
wozeparrot, line 43: Can use walrus `:=` here if you want.

Like: `(path := pathlib.Path(fp).parent).mkdir(parents=True, exist_ok=True)`

dosier: ooo nice, was not aware of that operator.

PR #1250, opened 2023-07-16
Assembly ptx target current arch

NotNotOddity:
Targets your current arch when using PTX.


NotNotOddity:
Need to find a better way to handle the arch() method though. I am pretty sure it will fail if your trying to use ocelot, which will kill the tests in CI once @cloud11665 brings those in. Although at the moment unless you have the specific GPU that was targeted you manually had to change the .target string...

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/assembly_ptx.py           3               2
```

PR #1252, opened 2023-07-16
CI < 5 minutes

cheeetoo:
It is still not always under 5 minutes, but if I mess around with it a bit more it will be. There is also a bug with the clang tests where it can't find the .so file for random tests.


geohot:
Bounty locked, good luck!

Though currently, this isn't eligible. "no more runners" is in bounty description.

cloud11665:
I can recommend looking at caching apt installs in cuda CI. iirc there was even a thing that did it on the github actions marketplace.

cheeetoo:
> Though currently, this isn't eligible. "no more runners" is in bounty description.
I'm sorry, I'm not sure what runners means in this context. Do you mean the pytest test runner?

geohot:
GitHub Actions runners. Master is using 12 runners and you are using 14.

cheeetoo:
Would you be fine with combining some of the short but unrelated jobs (like linters, c efficientnet, metal tests) to decrease the number of runners and using more for the other tests?

roelofvandijk:
What about using a pre-built docker instead of caching the apt install? Or maybe there's a better docker base image (from Nvidia) that already includes the cuda toolkit.

roelofvandijk:
Here are some other ideas:

Consider skipping the test_speed_v_torch test, it's slow and not that sensical on a runner. Similar speed benchmark tests might also be skipped. Maybe time for a benchmarks folder that can be skipped completely.

Consider ignoring all print statements in the code/tests (make them dependent on whether the CI/DEBUG env variables are set). There's a busy one in... I want to say test_ops.py. Also, some of the profiling code (time.monotonic() etc) can be skipped in those cases.

Disable the `tqdm` prints in the same manner.

Try running pytest with a different set of flags. -s does not make sense in [combination with xdist](https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html#output-stdout-and-stderr-from-workers). -v might also not be necessary.

Check how all caches are (de-)compressed. It might be worth it to have them slightly larger but load faster.

Cache the downloaded models? That MR got closed, caching all downloaded `.pth` files (glob pattern cache) might help a bit. See the closed MR.

geohot:
Merging is fine. I mostly want to avoid tons and tons of runners. I think @roelofvandijk has some good ideas, I'm fine with `benchmarks` folder too.

roelofvandijk:
This is the print I meant:
https://github.com/tinygrad/tinygrad/blob/master/test/test_ops.py#L52

roelofvandijk:
Try one of these:

https://hub.docker.com/r/nvidia/cuda/tags

roelofvandijk:
You can check locally which tests are especially slow using `pytest --durations=0`. [Docs](https://docs.pytest.org/en/6.2.x/usage.html#profiling-test-execution-duration)

roelofvandijk:
Check if the external tests are being ran repeatedly. I think the llama speed test needs to run only once somewhere to check if it's not broken.

Ensure you actually set the CI env variable! `CI=1 python -m pytest ...` everywhere.

roelofvandijk:
@cheeetoo Here are some ideas: https://github.com/tinygrad/tinygrad/pull/1279

You could profile the tests and re-order them, to make sure the slowest one get started first. For instance, look at the changes to `test_mnist.py`.

roelofvandijk:
Consider a warning filter to suppress the PyOpencl warnings

` CompilerWarning: Non-empty compiler output encountered. Set the environment variable PYOPENCL_COMPILER_OUTPUT=1 to see more.`

[warnings.filterwarning](https://docs.python.org/3/library/warnings.html#warnings.filterwarnings)
See this job https://github.com/tinygrad/tinygrad/actions/runs/5598352921/jobs/10237883946

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        2               0
```

test/test_speed_v_torch.py
```diff
@@ -93,7 +93,7 @@ def helper_test_generic(name, f1, f1_args, f2, f2_args):
   desc = "faster" if et_torch > et_tinygrad else "slower"
   flops = save_ops*1e-6
   mem = save_mem*1e-6
-  print(f"{prefix}{name:42s} {et_torch:7.2f} ms ({flops/et_torch:8.2f} GFLOPS {mem/et_torch:8.2f} GB/s) in torch, {et_tinygrad:7.2f} ms ({flops/et_tinygrad:8.2f} GFLOPS {mem/et_tinygrad:8.2f} GB/s) in tinygrad, {colorize_float(et_tinygrad/et_torch)} {desc} {flops:10.2f} MOPS {mem:8.2f} MB")
+  print_ci(f"{prefix}{name:42s} {et_torch:7.2f} ms ({flops/et_torch:8.2f} GFLOPS {mem/et_torch:8.2f} GB/s) in torch, {et_tinygrad:7.2f} ms ({flops/et_tinygrad:8.2f} GFLOPS {mem/et_tinygrad:8.2f} GB/s) in tinygrad, {colorize_float(et_tinygrad/et_torch)} {desc} {flops:10.2f} MOPS {mem:8.2f} MB")
```
geohot, line 96: I'd rather just expose "CI" variable from helpers and use an if statement

cheeetoo: fixed

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:latest
+
+RUN \
+  export DEBIAN_FRONTEND=noninteractive && \
+  apt-get update -y && \
+  apt-get install -y --no-install-recommends git g++ cmake ninja-build llvm-15-dev libz-dev libglew-dev flex bison libfl-dev libboost-thread-dev libboost-filesystem-dev nvidia-cuda-toolkit-gcc ca-certificates
```
roelofvandijk, line 6: * you can use `apt update` and `apt install` instead of `apt-get`
* When installing stuff with `apt ` in a Dockerfile, you might want to end the command with `rm -rf /var/lib/apt/lists/*`. [Best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)
* You might want to remove the cloned gpuocelot folder after building & installing it

cheeetoo:
<5 minutes with two less runners. It is only under five minutes if pip and gpuocelot have been cached, and also sometimes because github actions really inconsistent.
times

roelofvandijk:
I think you are not setting `CI=1` now for the pytest runs, so the print suppression might not work.

cheeetoo:
> I think you are not setting `CI=1` now for the pytest runs, so the print suppression might not work.

It is always set to `true`. [Docs](https://docs.github.com/en/actions/learn-github-actions/variables)`/

cheeetoo:
Resolved merge conflicts, still under 5 minutes.

geohot:
Looks great! Are there any tradeoffs?

roelofvandijk:
If it makes sense, you could consider marking tests using to group the ignores.

https://docs.pytest.org/en/7.1.x/example/markers.html

https://stackoverflow.com/questions/73907111/exclude-multiple-marked-tests-with-pytest-from-command-line

cheeetoo:
> If it makes sense, you could consider marking tests to group the ignores and simplify the pytest commands.

I was considering that, but since I would have to mark each individual test, I chose not to because I think that would be a lot of unnecessary clutter in the test files.

cheeetoo:
> Looks great! Are there any tradeoffs?

One small potential downside is that not all tests are being tested on each backend, but I tried to choose tests that weren't backend specific or are all ready tested a decent amount. Another small downside is that the way I did the tests with the matrix strategy is not super readable, but I can change that if you want. Otherwise, I can't see any other downsides.

roelofvandijk:
> > If it makes sense, you could consider marking tests to group the ignores and simplify the pytest commands.
> 
> I was considering that, but since I would have to mark each individual test, I chose not to because I think that would be a lot of unnecessary clutter in the test files.

You can mark entire files:
https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules

You could mark the ones that you want to exclude per job.

This 
``` 
WEBGPU=1 WGPU_BACKEND_TYPE=Metal python -m pytest -n=auto test/test_ops.py test/test_speed_v_torch.py test/test_nn.py test/test_jit.py test/test_randomness.py test/test_tensor.py test/test_assign.py test/test_conv.py test/test_nn.py test/test_custom_function.py test/test_conv_shapetracker.py
```

would become this
```
WEBGPU=1 WGPU_BACKEND_TYPE=Metal python -m pytest -n=auto -m 'not webgpu_excluded'
```

I you marked those files with `pytestmark = pytest.mark.webgpu_excluded` at the top.

Each pytest call in the `test.yml` could probably become:
`python -m pytest -n=auto test/ -m 'not ${{ matrix.backend }}_excluded'`

However, I guess it comes down to whether the config complexity should be in the `test.yml` file or in the individual files.

Two other questions:
- Does the matrix syntax make it faster, or is the speed gain mainly down to the way you distributed the tests?
- Have you tried disabling the pip cache? In my tests re-downloading all packages (no pip cache) was faster than using the cache (download speed > decompression).

_I'm just curious, feel free to ignore._

cheeetoo:
@roelofvandijk Thanks so much for pointing out marking files! The matrix syntax does not make it faster, it just makes it so stuff like checkout, setup-python, pip cache, etc do not need to be rewritten. As far as pip cache speed, it seemed to be consistently faster than downloading to me, but I will test it more to see.

roelofvandijk:
> @roelofvandijk Thanks so much for pointing out marking files!

Most welcome!

2 more things: 

- Making sure the slowest tests start early (by putting them in the beginning of their files) helps a bit. See the MR I linked above for an example.

- If you want to filter out the OpenCL CompilerWarnings (I have no idea about the speed impact), you can put this on top of e.g. the `test_ops.py` file:
  
  ```
  if CI:
    import warnings
    warnings.filterwarnings("ignore", message="Non-empty compiler output encountered")
  ```

roelofvandijk:
@cheeetoo apt-fast looks promising, shaves of 20s off the apt installs

https://github.com/tinygrad/tinygrad/actions/runs/5632290561/job/15260092997?pr=1324

vs (old)

https://github.com/tinygrad/tinygrad/actions/runs/5624773829/job/15242275296?pr=1314

roelofvandijk:
You can use `apt` insted of `apt-get` everywhere. You never need `apt update`  after `add-apt-repository`, it's included in the latter.

roelofvandijk:
Re: bundling commands, see previous comment.

You need to register your marks, currently Pytest is emitting warnings. See: https://docs.pytest.org/en/stable/how-to/mark.html

The CPU test job contains fuzzer tests, you might want to disable the prints in them, too. https://github.com/tinygrad/tinygrad/actions/runs/5632368320/job/15260249014?pr=1252

roelofvandijk:
`pytestmark = [pytest.mark.cuda, pytest.mark.gpu]`

This reads like you are including them for these jobs, but as far as I can see you are _excluding_ them. Easy to misunderstand.

cheeetoo:
Alright, it should be good now.

geohot:
Congrats, it's great to have faster CI!

Either post PayPal here or george@tinygrad.org to claim bounty

PR #1253, opened 2023-07-17
WIP: Low hanging fruit refactors

NotNotOddity:
Started to go through and try to rip out all the custom logic for temp files. This is due to all the different directories and way's we handle downloading files & models are much harder to cache in CI. Didn't go too far on it, and didn't go far enough to get it going through all the tests successfully yet, but wanted to make sure I get some feedback before spending the rest of the time required to do so. All-in-all we spend a couple minutes per job downloading and wanted to see if we could standardize it a little bit now before it kept deteriorating.

Also, include standardization for the platform vars, and that part is actually tested. If there is interest there I can pull it out into another commit.

Thanks ahead of time for any feedback!!

### DIFF
--- a/examples/hlb_cifar10_torch.py
+++ b/examples/hlb_cifar10_torch.py
@@ -1,18 +1,15 @@
 import numpy as np
 import torch
 import time
-import platform
 from torch import nn
 from torch import optim
 
 from extra.datasets import fetch_cifar
-from tinygrad.helpers import getenv
+from tinygrad.helpers import getenv, Platform
 
 # allow TF32
 torch.set_float32_matmul_precision('high')
-
-OSX = platform.system() == "Darwin"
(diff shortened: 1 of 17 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                       16               4
tinygrad/runtime/ops_clang.py              9              10
tinygrad/runtime/ops_cuda.py               3               3
tinygrad/runtime/ops_gpu.py                3               3
total                                     31              20
```

wozeparrot:
Hmm, the problem with the way this is currently done is that it puts the datasets and models into the temporary dir as well. This probably isn't what we want as some of the datasets/models are massive and default temporary dir location might be in ram.

geohot:

PR #1257, opened 2023-07-17
fixing cifar data

KayneWest:
current code produces a vector of `X.min(), X.max() == (-1.989213, 1045.3115)`  (below)

```
X[0]
array([[[637.6034  , 641.6514  , 665.93976 , ..., 552.59424 ,
         508.06564 , 467.58508 ],
        [613.31506 , 609.267   , 641.6514  , ..., 548.5462  ,
         504.01758 , 479.72925 ],
        [609.267   , 609.267   , 637.6034  , ..., 560.69037 ,
         524.2579  , 483.7773  ],
        ...,
```
should produce a vector of `X.min(), X.max() == (-1.9892129679394217, 2.126788393620005)`
```
X[0]
array([[[ 0.51899327,  0.53486799,  0.63011633, ...,  0.18562409,
          0.01100213, -0.1477451 ],
        [ 0.42374493,  0.40787021,  0.53486799, ...,  0.16974936,
         -0.00487259, -0.10012093],
        [ 0.40787021,  0.40787021,  0.51899327, ...,  0.21737353,
          0.07450103, -0.0842462 ],
        ...
```

### DIFF
--- a/extra/datasets/__init__.py
+++ b/extra/datasets/__init__.py
@@ -25,5 +25,6 @@ def fetch_cifar(train=True):
     db = [pickle.load(tt.extractfile('cifar-10-batches-py/test_batch'), encoding="bytes")]
   X = np.concatenate([x[b'data'].reshape((-1, 3, 32, 32)) for x in db], axis=0)
   Y = np.concatenate([np.array(x[b'labels']) for x in db], axis=0)
+  X = X / 255.
   X = (X - cifar10_mean) / cifar10_std
   return X, Y


geohot:
All the CIFAR code works. I really doubt this is the right change.

KayneWest:
The current code is using the zero-based mean/std in the normalization step. The fact that it works is actually kind of cool, but every other data loader produces 0(ish)-centered outputs - I came across this artifact while trying to reproduce something from another library and wondered why the output was so different. 

Here's an example using this official pytorch tutorial with the values `fetch_cifar` provides 
https://pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html
```
import torch
import torchvision
import torchvision.transforms as transforms

transform = transforms.Compose(
    [transforms.ToTensor(),
     # using the mean/std from fetch_cifar()
     transforms.Normalize((0.4913997551666284, 0.48215855929893703, 0.4465309133731618), (0.24703225141799082, 0.24348516474564, 0.26158783926049628))])

batch_size = 4

testset = torchvision.datasets.CIFAR10(root='./data', train=False,
                                       download=True, transform=transform)
testloader = torch.utils.data.DataLoader(testset, batch_size=batch_size,
                                         shuffle=False, num_workers=2)

dataiter = iter(testloader)
images, labels = next(dataiter)
print(images.min(), images.max())
```
`images.min(), images.max() == (tensor(-1.9892), tensor(2.0946))`

extra/datasets/__init__.py
```diff
@@ -25,5 +25,6 @@ def fetch_cifar(train=True):
     db = [pickle.load(tt.extractfile('cifar-10-batches-py/test_batch'), encoding="bytes")]
   X = np.concatenate([x[b'data'].reshape((-1, 3, 32, 32)) for x in db], axis=0)
   Y = np.concatenate([np.array(x[b'labels']) for x in db], axis=0)
+  X = X / 255.
```
kunwar31, line 28: dtype may get upcasted to np.float64, explicit cast to float32 is needed

kunwar31:
@geohot this looks like a bug that went un-noticed, I tested this on master and acc. goes from 90.20 to 90.32 with same hparams. looks like it slows down learning by a small amount

kunwar31:
for 94% bounty this might be an important change

PR #1259, opened 2023-07-17
Symbolic ShapeTracker and JIT llama

chenyuxyz:
Open to have the bounty lock. @geohot 

Proof of concept:
```
@TinyJit
def f(x): return (x + x).realize()

for i in range(1, 10):
    t = Tensor.rand(i, 3)
    t.symbolize(0, Variable("X", 1, 100))
    f(t)
```

generates this CLANG code, also verified that SYM_X is updated correctly during run time
```
void E_X1_1003(float* restrict data0, const float* restrict data1, const int* SYM_X) {
int X = *SYM_X;
for (int gidx0 = 0; gidx0 <= ((X*3)+-1); ++gidx0) {
    float val1_0 = data1[gidx0];
    float alu0 = (val1_0+val1_0);
    data0[gidx0] = alu0;
  } /* global+local */
}
```

image

ops and mems are 0 because those are symbolic now

Next step is to make matmul work


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/cstyle.py                19               7
tinygrad/codegen/linearizer.py             4               0
tinygrad/jit.py                           30               9
tinygrad/lazy.py                          21               2
tinygrad/ops.py                           41              13
tinygrad/runtime/lib.py                    7               0
tinygrad/shape/shapetracker.py            42              14
tinygrad/shape/symbolic.py                33              11
tinygrad/tensor.py                        11               1
total                                    208              57
lines added in the tinygrad folder: 151
```

leastcompressible:
With some tweaking, simple matmul works
```
@TinyJit
def f(a, b): return (a @ b).realize()
for i in range(1, 10):
  a = Tensor.rand(3, i)
  a.symbolize(1, Variable("X", 1, 100))
  b = Tensor.rand(i, 5)
  b.symbolize(0, Variable("X", 1, 100))
  f(a, b)
```
generates this kernel
```
void r_3_5_X1_100(float* restrict data0, const float* restrict data1, const float* restrict data2, const int* SYM_X) {
int X = *SYM_X;
for (int gidx0 = 0; gidx0 <= 2; ++gidx0) {
for (int gidx1 = 0; gidx1 <= 4; ++gidx1) {
    float acc0_0 = 0.0f;
    for (int ridx2 = 0; ridx2 <= (-1+X); ++ridx2) {
      float val1_0 = data1[((gidx0*X)+ridx2)];
      float val2_0 = data2[((ridx2*5)+gidx1)];
    } /* reduce */
    data0[((gidx0*5)+gidx1)] = acc0_0;
  }} /* global+local */
}
```
image

Will spend some time to rework symbolic.py. Current problem is that previous optimization only works between Node and int, not Node and Node. For example this breaks
```
idx1 = Variable("idx1", 0, 31)
start_pos = Variable("start_pos", 1, 100)
idx2 = Variable("idx2", 0, start_pos-1)
idx3 = Variable("idx1", 0, 127)
s = (idx1 * start_pos * 128) + idx2 * 128 + idx3
print(f"{s.__floordiv__(128)=}")  # This should equal to (idx1 * start_pos + idx2)
```
pytorch uses [sympy](https://github.com/pytorch/pytorch/blob/main/torch/utils/_sympy/value_ranges.py) to compute symbolic shapes. For our use case (mostly non-negative, simple arithmetic) it should be possible to just expand symbolic.py.

geohot:
Yea we definitely aren't using sympy.

Is `symbolize` the right abstraction? What happens if movementops are done after that?

leastcompressible:
I don't know if `symbolize` will be the final API yet, I will have better answers after JITing llama and see how easy to apply JIT.

The current implementation registers the dimension with that symbol, and the shape tracker should use the symbol for downstream movements.
```
a = Tensor.rand(3, 5)
a.symbolize(0, Variable("X", 1, 100))
print(a.shape)  # (<X[1-100]>, 5)
a = a.permute((1, 0))
print(a.shape)  # (5, <X[1-100]>)
```

geohot:
What does

```
a = Tensor.rand(16, 16)
a.symbolize(0, Variable("X", 1, 100))
a.reshape(4,4,4,4)
```

do?

chenyuxyz:
Currently it gives you (4, 4, 4, 4) only because I have not updated the reshape shape check, it should fail. I think reshape only is defined with identical input size and output size (with the only exception to reshape from 0-dim). Like we cannot `Tensor.rand(20, 16).reshape(4, 4, 4, 4)`.

My current thinking is that an operation is allowed only if it's valid for all the possible input symbol value (or if somehow we cannot determine).

chenyuxyz:
High level full design:
1. Call `symbolize` to make a dimension a symbol. Symbol value is stored in lazydata. ShapeTracker uses the symbol downstream, all shape related computation becomes symbolic.
2. First compile time, lazydata passes the symbol and their values into AST. Symbols has a standalone buffer input in generated kernel.
3. First time running the code, AST creates the symbol buffer with values and use that to run the code. We store all the buffers along with their symbolic shapes to jit_cache.
4. JIT time, `symbolize` called again, lazydata will have the new value. TinyJIT gathers all symbol needed from args to form the symbol buffer. We have a new RawBuffer function `resize`. All intermediate and output buffers use symbolic shape and current symbol size to resize to the correct size.

I like the part that it can go through same kernel gen path and keeps some optimization potential. Like we can still upcast a dim=4*X, or put X in local if it's < 256 etc. We also get symbolic FLOPs and memory count for free, which is pretty cool.

Adding symbolic arithmetic coverage at the moment, 4d matmul fails now due to not able to figure out `(X*240)%(X*3)`. I also want to refactor and clean up symbolic.py a bit because many existing optimization do not work with op(Node, Node).

Antonio12389:
Chicos por favor necesito su apoyo tengo más de 4 años que alguien roba toda la información de mi móvil y computadora estoy desesperado ya no se que hacer ya informe a la policía cibernética y nadie a podido ayudarme robaron dinero de mis cuentas bancarias de mis criptomonedas de app, etc. Espero me puedan ayudar ya que me siento vigilado pues tIenen toda mi información 
Sin más por el momento agradezco su atención y apoyo,estoy al tanto para proporcionar cualquier información que sea necesaria.

geohot:
I don't exactly understand bind, but I generally like the idea of passing a Variable into reshape which it looks like is where you are going.

`Tensor.zeros(4,4).reshape(4, Variable('I', 1, 100))`

chenyuxyz:
Still have some cleanup and TODO to fix, but we are very close. Purple llama is coming! I need to change llama.py a bit because it does not handle data dependent control flow. New test file is very nice too.

image

chenyuxyz:
so the white kernels left are `cat` keys and values. It looks like the implementation with `slice` jammed the symbolic shape.

Also the JIT symbolic kernels are slowed now
`JIT=1 python -O examples/llama.py --prompt "Hello." --count 30 --temperature 0 --timing` gives me 85ms
`JIT=1 JIT_ATTN=1 python -O examples/llama.py --prompt "Hello." --count 10 --temperature 0 --timing` is about 250ms
I need to skip some optimization steps because those did not play nicely with symbolic shapes yet.

Rough plan now
1. carve out smaller PRs and get those merged, probably one for symbolic shapetracker and one for passing symbols in JIT
2. fix the relevant part to make the JIT fast
3. find out how to support `cat`

chenyuxyz:
NOTE: the actual change is in #1353 now

geohot:
Closing for #1353 and moved bounty lock there

PR #1260, opened 2023-07-17
Fix NaN handling in CMPEQ on METAL and LLVM backends and add Tensor.isnan

### LINKED ISSUE #1069: Tensor.eq with nan does not match PyTorch on LLVM backend
Steps to reproduce:
```
jacky@ephphatha tinygrad % METAL=1 python3
Python 3.11.3 (main, Apr  7 2023, 20:13:31) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from tinygrad.tensor import Tensor
>>> x = Tensor.full((3, 3), float("nan"))
>>> (x != x).numpy()
array([[0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.]], dtype=float32)
>>> import torch
>>> x = torch.full((3, 3), float("nan"), device="mps")
>>> x != x
tensor([[True, True, True],
        [True, True, True],
        [True, True, True]], device='mps:0')
```

Edit: LLVM also returns all zeros

flammit:
Closes #1069 but it comes at the cost of losing all of the fast math optimizations on METAL and the non-NaN optimization in LLVM.

### DIFF
--- a/tinygrad/codegen/llvmir.py
+++ b/tinygrad/codegen/llvmir.py
@@ -27,7 +27,7 @@ code_for_op: Final[Dict[Op, Callable]] = {
   BinaryOps.SUB: lambda builder,x,y: builder.fsub(x,y, flags=('fast',)),
   BinaryOps.MUL: lambda builder,x,y: builder.fmul(x,y, flags=('fast',)),
   BinaryOps.DIV: lambda builder,x,y: builder.fdiv(x,y, flags=('fast',)),
-  BinaryOps.CMPEQ: lambda builder,x,y: builder.uitofp(builder.fcmp_ordered("==", x, y, flags=('fast',)), ir.FloatType()),
+  BinaryOps.CMPEQ: lambda builder,x,y: builder.uitofp(builder.fcmp_ordered("==", x, y, flags=('ninf','nsz','arcp','contract','afn','reassoc',)), ir.FloatType()), # include all fast-math flags except nnan "no NaNs"
   BinaryOps.MAX: lambda builder,x,y: builder.select(builder.fcmp_unordered(">", x, y, flags=('fast',)), x, y, flags=('fast',)),
   TernaryOps.MULACC: lambda builder,x,y,z: builder.fadd(builder.fmul(x,y, flags=('fast',)), z, flags=('fast',)),
   TernaryOps.WHERE: lambda builder,x,y,z: builder.select(builder.fcmp_unordered("!=", x, ir.Constant(ir.FloatType(), 0), flags=('fast',)), y, z, flags=('fast',)),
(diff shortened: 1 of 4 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/llvmir.py                 1               5
tinygrad/runtime/ops_metal.py              1               0
tinygrad/tensor.py                         1               0
total                                      3               5
```

tinygrad/runtime/ops_metal.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, name:str, prg:str):
       self.library = unwrap(METAL.device.newLibraryWithData_error_(data, None))
     else:
       options = Metal.MTLCompileOptions.alloc().init()
+      options.setFastMathEnabled_(0)
```
geohot, line 48: Any way to be more granular with this? I worry this'll make metal slower

tinygrad/codegen/llvmir.py
```diff
@@ -41,10 +41,6 @@ def uops_to_llvm_ir(uops:List[UOp], bufs:List[LazyBuffer]) -> str:
   func_dtypes = [{dtypes.float16:ir.HalfType(), dtypes.float32:ir.FloatType(), dtypes.int8:ir.IntType(8), dtypes.uint8:ir.IntType(8), dtypes.bool: ir.IntType(1), dtypes.int64: ir.IntType(64), dtypes.int32: ir.IntType(32)}[buf.dtype] for buf in bufs]
   func = ir.Function(module, ir.FunctionType(ir.VoidType(), [x.as_pointer() for x in func_dtypes]), name='exec')
 
-  # force llvmlite to allow us to add function attribute then add the attribute
-  func.attributes._known = func.attributes._known.union(frozenset(['"no-nans-fp-math"="true"']))
```
geohot, line 45: We needed this to get performance.

geohot:
Skipping on WEBGPU is fine, my understanding is it doesn't support NaNs.

tinygrad/codegen/llvmir.py
flammit: Overall I totally agree this PR shouldn't be merged if performance is preferred over edge-case correctness over all backends.  This PR is just meant to show what would be required to make the backends match the CMPEQ/MAX op behavior of torch CPU: disabling fast math for CMPEQ, using select and isnan for for MAX instead of fmax.

I suppose I'm making the case that if performance is key, tinygrad should adopt the most GPU-efficient behavior as the base case and make the CPU backends less efficient to match their behavior since overall FLOPS will come from GPU.

tinygrad/runtime/ops_metal.py
flammit: Looking at the docs, I don't see any options to narrow this to only non-NaN optimizations like in the LLVM case.

PR #1262, opened 2023-07-17
cifar fix v2

KayneWest:
It doesn't look like I can add to a closed pr, so adding a new one here 
https://github.com/tinygrad/tinygrad/pull/1257 

adding cast to `np.float32` per @kunwar31


geohot:
Again, this isn't the problem. You are making a change, there's no added tests, and AFAIK CIFAR training works. Why would I want the dataloader returning float32? That's 4x the bandwidth to copy the GPU.

PR #999, opened 2023-06-18
whisper: clean up code, support all sizes, test with eval set

jla524:
Changes made:
- support all sizes
- test against librispeech dataset
- clean up code
  - no more torch

Results on librispeech:
```
jacky@ephphatha tinygrad % TEST=1 python3 examples/whisper.py
ram used:  0.11 GB, decoder.ln.bias                                   : 100%|█████| 167/167 [00:00<00:00, 2636.35it/s]
loaded weights in 70.28 ms, 0.11 GB loaded at 1.54 GB/s
--------------------------------------------------------------------------------------------------------------------------------
2277-149896-0000

  he was in a fevered state of mind owing to the blight his wifes action threatened to cast upon his entire future

word error rate: 0.0000
--------------------------------------------------------------------------------------------------------------------------------
2277-149896-0001

  he would have to pay her the money which she would now regularly demand or there would be trouble it did not matter what he did

word error rate: 0.0000
--------------------------------------------------------------------------------------------------------------------------------
...
```

PR #1264, opened 2023-07-18
Removed dep of torch, torchaudio, kept librosa only

pgavazzi:
First PR, just wanted to remove dependencies to torch, torch audio since librosa has all the same functionality already


wozeparrot:
So these changes are already in the other whisper pr: #999.

PR #1265, opened 2023-07-18
Enable JIT tests for supported devices, skip METAL and WEBGPU

leastcompressible:
Enable test to not skip any supported device by default.

While doing so, I realized METAL test fails, presumably due to fast-math optimization. But disabling fast-math did not fix it on CI, it fixed on my local M1 though.

I think we want to test these devices, so I recommend adding the test for all and explicitly skipping METAL and WEBGPU for now. [Link to show that these two failed in CI](https://github.com/leastcompressible/tinygrad/actions/runs/5589540711).


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/jit.py                            3               1
```

Issue #1263, opened 2023-07-18
Assertion Error using Transpose if buffer has WHERE Op 


ZenginU:
Do not know if this error is intended, but cant take the transpose of the tensor if there is WHERE op in the LB.

To reproduce
`Tensor.rand(5,5).ceil().T`

[This](https://github.com/tinygrad/tinygrad/blob/f8c539989e45690ddc95e89a884c2da634cdcb2a/tinygrad/ops.py#L55C7-L55C32) just throws assertion error.

ZenginU:
can be fixed by adding TernaryOps in assertion but do not know if it is against the logic if not I will open PR

```python
  def replace_with_movement_ops(self:LazyOp, ops:List[Tuple[MovementOps, Tuple[Any, ...]]]) -> 'LazyBuffer':
    ...
    assert self.op in BinaryOps or self.op in UnaryOps or self.op in TernaryOps
    ...
```

PR #1266, opened 2023-07-18
Fix: AssertionError Transpose/Permute when WHERE Op in LB

ZenginU:
Related #1263,

I have changed the assertion condition but not sure if it against the logic or just overlooked.

Transposing a tensor throws assertion if there is WHERE Ops in the LB.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -569,6 +569,9 @@ class TestOps(unittest.TestCase):
     helper_test_op([(3,4,5,6)], lambda x: x.movedim((3,2,1,0),(0,1,2,3)), lambda x: x.permute(order=(3,2,1,0)))
     helper_test_op([()], lambda x: x.permute(()), lambda x: x.permute(()))
 
+  def test_transpose_where(self):
+    Tensor.rand(5,5).ceil().permute((1, 0))
+
   def test_reshape(self):
     helper_test_op([(4,3,6,6)], lambda x: torch.reshape(x, (-1,3,6,6)), lambda x: x.reshape(shape=(-1,3,6,6)))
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

ZenginU:
@wozeparrot 
I believe it is not about the Transpose itself but happens with the **tensor.permute** as well, I guess maybe the problem is with the evaluation of the where ops??

If you are talking about the test cases for the **replace_with_movement_ops** then I may not be able to add a test case as I dont know the inner workings.

If you are suggesting test cases for Transpose than I believe problem is not only about transpose but may be with WHERE ops

Reproduce with permute (throws assertion)
```python
tensor.Tensor.rand(5, 5, 5).ceil().permute((0, 2, 1)).numpy()
```

geohot:
That's a fine regression test, just anything that currently fails and this change fixes.

ZenginU:
@geohot 
I have added a test with torch equivalent.
Currently nothing fails as ceil/floor is not used before movement ops.

wozeparrot:
Maybe the test should be testing with `.where` directly? Makes the test case simpler and `.ceil` is a little bit broken right now.

ZenginU:
@wozeparrot 
I believe ternaryOps is a new thing so in lazy.py/elementwise_op this function accepts TernaryOps input but I believe author does not edited the ops.py/replace_with_movement_ops.

 So when TernaryOps is in the LB system does not replace it because of the assert. 

I wrote a simple test case using only .where and .permute. Problem only occurs when ".where" is in the Buffer so directly using .where does **not** trigger the assert.

Any more suggestions?

wozeparrot:
Nope! I think the test is fine.

PR #933, opened 2023-06-05
RNN-T Training

the-praxs:
My starter code for training and benchmarking RNN-T.

TODO:
- [x] Add the code for automatic download, extractions, and conversion of necessary files in `librispeech.py`
- [x] Correct the path for storing the converted `.wav` file so that it matches `dev-clean-wav.json`
- [x] Ignore `.tar.gz` files
- [x] NVIDIA drivers kind of fixed (Windows sucks) 
- [x] Add code for training and validation files in a clean manner  
- [ ] Add code to generate JSON files similar to `dev-clean-wav.json`
- [ ] Check code for RNN-T for errors
- [ ] Train and benchmark performance

Thanks to @wozeparrot for his awesome guide-in-progress and the starter code for RNN-T

PR #1267, opened 2023-07-18
RNNT training

Huvinesh-Rajendran-12:
-  The training of RNNT is conducted using SGD. 
-  Added training mode to the iterator of librispeech to download training files. Added support for flac files. 
-  Implemented the loss function of negative log-likelihood. 
-  Enhanced the RNNT model to output softmax probabilities along with a sequence of characters. source of RNNT implementation : [https://www.microsoft.com/en-us/research/uploads/prod/2019/10/RNNT.pdf](url)
-  To-do: Implemented beam search decoding as an alternative to greedy decoding. source : [https://arxiv.org/abs/2302.14357](url)
-  To-do: Replace some numpy codes and functionalities with that of Tensor. 

Hopefully, this is enough for locking up the bounty.


wozeparrot:
So the training bounty is already locked in #933. 

Additionally, the bounty is for training a RNN-T model that follows the [mlperf specifications](https://github.com/mlcommons/training/tree/master/rnn_speech_recognition/pytorch), where the loss is RNN-T loss and not NLL loss and the optimizer is LAMB not SGD.

Huvinesh-Rajendran-12:
Ah okay, well I saw there is an opening slot in the bounties google sheet. So can you confirm with me if the slot is opened or locked so that I can decide whether to proceed or not with the changes. Thanks. 

image

wozeparrot:
Yea, so the other one is still being worked on, it should be re-locked on the doc again.

Huvinesh-Rajendran-12:
Okay, no problem but can I still work on this as a learning exercise for myself? Would that be a problem?

geohot:
Not touched in over a week, closing

dc-dc-dc:
Previous attempt at #1620

PR #1268, opened 2023-07-18
Fix WSGL to render NaN and prevent shader compile error

flammit:
Currently WGSL will fail to compile due to a shader error in the presence of NaN const (renders as "nanf").

For reference on this approach see: https://github.com/gpuweb/gpuweb/issues/3749

### DIFF
--- a/tinygrad/codegen/wgsl.py
+++ b/tinygrad/codegen/wgsl.py
@@ -25,14 +25,16 @@ class WGSLLanguage(CStyleLanguage):
     return f"var<workgroup> {name}: array<f32,{size}>;"
 
   def render_const(self, x:Union[float,int], var_dtype) -> str:
-    if math.isinf(x): val = ("-" if x < 0 else "") + "0x1.fffffep+127f"
+    if math.isnan(x): val = "nan()"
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/wgsl.py                   4               2
```

dc-dc-dc:
NaN tests are currently disabled for WEBGPU, can you reenable them to verify this change works or add in new tests for nan only as the nan tests also include inf

dc-dc-dc:
I'm not entirely sure about this, afaik nan and inf are considered undefined values in WEBGPU and not supported atm. The issue referenced was for a hack that was patched out but a new workaround was discovered which doesn't give much confidence that this new hack will be around for long.

geohot:
If the nan tests pass I'm okay with this.

wozeparrot:
> Does `bitcast<f32>(0xFFFFFFFF)` not work? Don't use the function

This is what makes it not quite on par with the inf hack, because this is actually hacking around a compiler bug that allows this behavior. The function shouldn't work either. It's way more hacky than inf is.

PR #1270, opened 2023-07-18
Don't store RawConst in JIT input_rawbuffers.

chenyuxyz:
ASTRunner only stores kernel arguments (not constants) in cache, so we don't store constants in input_rawbuffers so the two can match.

### DIFF
--- a/test/test_jit.py
+++ b/test/test_jit.py
@@ -71,5 +71,14 @@ class TestJit(unittest.TestCase):
       c = fun(b)
       np.testing.assert_equal(c.numpy(), fun.a.numpy()+b.numpy())
 
+  def test_jit_handles_constant_input(self):
+    @TinyJit
+    def add(a, b): return (a+b).realize()
+    for _ in range(3):
+      a = Tensor.randn(10, 10)
+      b = Tensor([1])
+      c = add(a, b)
+      np.testing.assert_equal(c.numpy(), a.numpy()+1)
+
 if __name__ == '__main__':
   unittest.main()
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/jit.py                            9               2
lines added in the tinygrad folder: 7
```

test/test_jit.py
```diff
@@ -71,5 +71,14 @@ def __call__(self, b:Tensor) -> Tensor:
...
+    @TinyJit
+    def add(a, b): return (a+b).realize()
+    for _ in range(3):
+      a = Tensor.randn(10, 10)
+      b = Tensor([1])
+      c = add(a, b)
+      np.testing.assert_equal(c.numpy(), a.numpy()+1)
+
 if __name__ == '__main__':
   unittest.main()
```
digitalnomd, line 84: nit: not related to your PR but there's no newline

kunwar31:
This allows even more buggy behaviour, the consts used in the 2nd fxn run will get JITed as part of the kernel itself   
```
for _ in range(3):
    foo(a,Tensor([2]))
```

```
__kernel void E_3(__global float* data0, const __global float* data1) {
  { int lidx0 = get_local_id(0);  /* 3 */
    float val1_0 = data1[lidx0];
    float val2_0 = 2.0f;
    float alu0 = (val2_0*val1_0);
    data0[lidx0] = alu0;
  } /* global+local */
}
```

Also, this doesn't solve cases where all inputs are consts:  `foo(Tensor([5]),Tensor([2]))`

chenyuxyz:
It's the same behavior without this change. Currently the constant is not part of the generated kernel. I feel the proper fix is to revert the constant folding so that size 1 tensor is always a buffer. Then we can decide the proper syntax to make declare that it's foldable.

Issue #1273, opened 2023-07-18
JIT with constant folded inputs is broken 


kunwar31:
after constant folding was fixed in https://github.com/tinygrad/tinygrad/pull/1227, adding .contiguous() to inputs of JITed functions was suggested as a fix for passing around 1-D tensors as inputs 
but this creates another problem where there is always an extra kernel because of .contiguous() which isn't part of JIT
![image](https://github.com/tinygrad/tinygrad/assets/24824080/cd64218a-ca19-4a95-90d7-286848864d32)

using the example doc method, in foo1 does work but needs to create a 2-D tensor. there should be a better way

g1y5x3:
Which also relates to current CIFAR training speed variation
![image](https://github.com/tinygrad/tinygrad/assets/17937555/644046d5-c3e1-4e08-a86d-3127c2de2742)
![image](https://github.com/tinygrad/tinygrad/assets/17937555/2f160ce2-04a7-4053-beb0-a8bff708566b)

PR #1274, opened 2023-07-18
Revert "Fix constant folding for Tensor([3])"

chenyuxyz:
Reverts tinygrad/tinygrad#1227

This change always make size 1 tensor a constant, which breaks #1273. Passing a size 1 tensor in JIT as input does not work because JIT only updates buffers.

I recommend reverting this now and find a better way to spec constant tensors.

### DIFF
--- a/docs/abstractions.py
+++ b/docs/abstractions.py
@@ -120,8 +120,7 @@ from tinygrad.tensor import Tensor
 from tinygrad.ops import LazyOp, BinaryOps, LoadOps
 
 # the 2+3 from before
-# added some 0s, otherwise Tensor([2]) will be folded into a constant without using LoadOps.FROM
-result = Tensor([2, 0]) + Tensor([3, 0])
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         2               5
```

geohot:
Hmm, I'm not ready to revert yet. The old behavior was just luck that you use `3` instead of `[3]`

Any proposals to fix the underlying problem?

chenyuxyz:
Can you clarify which behavior it is in "The old behavior was just luck that you use `3` instead of `[3]`"?

Now I think conceptually 0-dim tensor has empty shape and size 0, so we use const to represent it. And 1-dim tensor has size 1 and it should have a buffer.

geohot:
These tensors will be different in that case:
```
a = Tensor(3).reshape(1)
b = Tensor([3])
```

But eh, I'll revert it since I see it's causing problems for people.

PR #1278, opened 2023-07-19
Add Fuzz Test symbolic / shapetracker to CI.

chenyuxyz:
Modified the existing fuzz script and add to CI. It takes about 10 seconds each.

Want to get this in before making more changes in symbolics.
Tested locally that bad changes will reliably be caught.
Tested in CI that a bad changes will make CI fail.

### DIFF
--- a/test/external/fuzz_shapetracker.py
+++ b/test/external/fuzz_shapetracker.py
@@ -1,4 +1,5 @@
 import random
+import sys
 from tinygrad.helpers import DEBUG
 from test.unit.test_shapetracker import CheckingShapeTracker
 random.seed(42)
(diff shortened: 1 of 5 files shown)


test/external/fuzz_shapetracker.py
```diff
@@ -1,60 +1,65 @@
...
 if __name__ == "__main__":
   ops = [do_permute, do_pad, do_shrink, do_reshape_split_one, do_reshape_combine_two, do_stride, do_expand]
-  while 1:
-    st = CheckingShapeTracker((3, 3, 3))
+  for _ in range(200):
+    st = CheckingShapeTracker((random.randint(2, 10), random.randint(2, 10), random.randint(2, 10)))
     for i in range(8): random.choice(ops)(st)
-    #st.simplify()
-    st.assert_same()
+    try:
```
geohot, line 62: Do you really need this? Assertion itself doesn't return bad?

chenyuxyz: not needed, removed. Thanks for catching

PR #1283, opened 2023-07-19
Added test_chunk and fixed

ZenginU:
Added test_chunk and changed the implementation because previous one fails the tests. 

Test cases are bit mess right now as I tried to figure out how to test for mod != 0 and dimension tests.

I have changed the implementation (checks were failing).

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -343,7 +343,7 @@ class Tensor:
   def chunk(self, num, dim):
     slice_params = [[slice(None) for s in self.shape] for _ in range(ceil(self.shape[dim]/ceil(self.shape[dim]/num)))]
     for i, k in enumerate(range(0, self.shape[dim], ceil(self.shape[dim]/num))):
-        slice_params[i][dim] = slice(k, k + ceil(self.shape[dim]/num))
+      slice_params[i][dim] = slice(k, k + ceil(self.shape[dim]/num))
     return [self[tuple(sl)] for sl in slice_params]
 
   def squeeze(self, dim=None):
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -341,10 +341,10 @@ class Tensor:
 
   # TODO: make this nicer with syntactic sugar in slice
   def chunk(self, num, dim):
-    slice_params = [[(0, s) for s in self.shape] for _ in range(num)]
-    for i,k in enumerate(range(0, self.shape[dim], self.shape[dim]//num)):
-      slice_params[i][dim] = (k, min(self.shape[dim], k+self.shape[dim]//num))
-    return [self.slice(p) for p in slice_params]
+    x = self.permute(*(i for i in range(self.ndim) if i != dim), dim)
+    dv, mod = np.divmod(self.shape[dim], num)
+    step = dv if mod == 0 else dv + 1
+    return [x[..., i:i + step].permute(*range(dim), self.ndim - 1, *range(dim, self.ndim - 1)) for i in range(0, self.shape[dim], step)]
 
   def squeeze(self, dim=None):
     if dim is None: return self if 1 not in self.shape else self.reshape(*[size for size in self.shape if size != 1])
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -343,7 +343,7 @@ class Tensor:
   def chunk(self, num, dim):
     dim = (dim + self.ndim) if dim < 0 else dim
     x = self.permute(*(i for i in range(self.ndim) if i != dim), dim)
-    dv, mod = np.divmod(self.shape[dim], num)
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         4               4
```

geohot:
"previous one fails the tests"

What exactly fails and why?

https://thoughtbot.com/blog/chestertons-fence

ZenginU:
@geohot

Firstly, unfortunately there was no previous test_chunk. 
Secondly, I believe previous code is aimed to split the kernel perfectly.

`If the tensor size along the given dimension dim is not divisible by chunks, all returned chunks will be the same size, except the last one`
Previous code lacks this condition so when you try to split 14 into 3 chunk code fails.

I hope this explanation is sufficient

tinygrad/tensor.py
```diff
@@ -341,10 +341,11 @@ def repeat(self, repeats):
 
   # TODO: make this nicer with syntactic sugar in slice
   def chunk(self, num, dim):
-    slice_params = [[(0, s) for s in self.shape] for _ in range(num)]
-    for i,k in enumerate(range(0, self.shape[dim], self.shape[dim]//num)):
-      slice_params[i][dim] = (k, min(self.shape[dim], k+self.shape[dim]//num))
-    return [self.slice(p) for p in slice_params]
+    dim = (dim + self.ndim) if dim < 0 else dim
+    x = self.permute(*(i for i in range(self.ndim) if i != dim), dim)
```
geohot, line 345: Why is this permuting?

ZenginU: In order not to deal with slice_params, I moved the dimension that will be chunked to last dim to use the "..." in the for loop. May be there is a more optimized approach, if you want I can try to find it

geohot: Yea, do it the right way

ZenginU:
@geohot 

So made the code shorter and not used Tensor.slice as it is discouraged,

But if you say that this code is unreadable I have find the bug in the old code as well so I can just change it preserving the readability

tinygrad/tensor.py
```diff
@@ -341,10 +341,9 @@ def repeat(self, repeats):
 
   # TODO: make this nicer with syntactic sugar in slice
   def chunk(self, num, dim):
-    slice_params = [[(0, s) for s in self.shape] for _ in range(num)]
-    for i,k in enumerate(range(0, self.shape[dim], self.shape[dim]//num)):
```
ZenginU, line 345: changing
```python
self.shape[dim]//num -> ceil(self.shape[dim]/num)
```  
in all occurences fixes the bug in prev code, it was late for me to realize this unfortunately.

wozeparrot: I think the old chunk code is more readable, maybe try the fix on the old code and see if it passes the test?

ZenginU: Hello @wozeparrot,

Most of it looking the same only using slice(). But if still unreadable than I would directly update the old chunk code.

I am insisting on not using tensor.slice but instead using builtin slice(). Sometimes tensor.slice() can add unexpected zero padding.

wozeparrot: It's much more readable now. :)

wozeparrot:
Can you add a test for negative dims?

ZenginU:
Added new tests but testing seems odd to me but cant think any other way as .chunk returns list. Any ideas on this?

wozeparrot:
You can try to add a special case in `helper_test_op` but i don't think it will be much cleaner than this, cuz this is already pretty clean.

ZenginU:
@wozeparrot 
Find more bugs related with old code, first line became bit ugly (inside the range()) but I hope gets the job done and ready for merge

tinygrad/tensor.py
```diff
@@ -341,10 +341,10 @@ def repeat(self, repeats):
...
-      slice_params[i][dim] = (k, min(self.shape[dim], k+self.shape[dim]//num))
-    return [self.slice(p) for p in slice_params]
+    slice_params = [[slice(None) for s in self.shape] for _ in range(ceil(self.shape[dim]/ceil(self.shape[dim]/num)))]
```
wozeparrot, line 344: Do you need to initialize with `slice(None)` here? does initializing with just `None` work?

ZenginU: Replacing slice(None) with None breaks, my idea with the slice(None) is ":" but i think None acts different as it throws `AssertionError: zeros not allowed in shape (1, 0, 13)`

Btw, I have plans to replace the slice with .pad in .pad2d as well should I do it? Maybe I can combine them in this PR as well as it will be small

wozeparrot: Should keep that in a separate PR. I'll just merge this one now.

PR #1275, opened 2023-07-18
Fix llama.py to load and concatenate 13B, 30B, and 65B models

flammit:
Add back support for LLaMA 13B and add 30B and 65B.  Closes #1254.

Wasn't able to test 65B due to memory constraints but here are the results with 13B and 30B on a MBP16 M2 Max 96GB:

```
using METAL backend
using 13B model
ram used: 26.03 GB, freqs_cis                                         : 100%|█████████████████████████████████████████████████████████████████████████████████████████| 364/364 [00:06<00:00, 58.91it/s]
loaded weights in 6185.82 ms, 26.03 GB loaded at 4.21 GB/s
Hello.
ran model in 596.72 ms
sync in 31.25 ms
 I
ran model in 154.41 ms
sync in 2.73 ms
 am
ran model in 148.00 ms
sync in 2.73 ms
 a
ran model in 140.58 ms
sync in 3.20 ms
 
ran model in 142.85 ms
sync in 2.65 ms
2
ran model in 149.57 ms
sync in 2.81 ms
0
ran model in 143.11 ms
sync in 2.68 ms
 year
ran model in 142.81 ms
sync in 2.67 ms
 old
ran model in 150.24 ms
sync in 2.81 ms
 female
ran model in 143.87 ms
sync in 2.73 ms
.%                               
```

```
using METAL backend
using 30B model
ram used: 65.06 GB, freqs_cis                                         : 100%|█████████████████████████████████████████████████████████████████████████████████████████| 544/544 [00:22<00:00, 23.88it/s]
loaded weights in 22786.53 ms, 65.06 GB loaded at 2.86 GB/s
Hello.
ran model in 11879.79 ms
sync in 186.19 ms
 I
ran model in 1411.35 ms
sync in 171.10 ms
 am
ran model in 1616.58 ms
sync in 122.17 ms
 a
ran model in 1668.45 ms
sync in 163.60 ms
 
ran model in 1678.47 ms
sync in 136.57 ms
2
ran model in 1796.34 ms
sync in 106.86 ms
0
ran model in 1551.07 ms
sync in 132.73 ms
 year
ran model in 1669.32 ms
sync in 132.06 ms
 old
ran model in 1658.96 ms
sync in 76.19 ms
 female
ran model in 1550.73 ms
sync in 99.92 ms
.%                      
```

PR #1284, opened 2023-07-19
Add LLaMA-2 support

prusnak:
Followup to https://github.com/tinygrad/tinygrad/pull/1275 which cleans up code reuse and introduces LLaMA-2 support.

### DIFF
--- a/tinygrad/state.py
+++ b/tinygrad/state.py
@@ -4,6 +4,7 @@ from typing import Dict, Union, List
 from tinygrad.tensor import Tensor
 from tinygrad.helpers import dtypes, prod, argsort, DEBUG, Timing, GlobalCounters
 from tinygrad.shape.shapetracker import strides_for_shape
+from tinygrad.lazy import Device
 
 safe_dtypes = {"F16": dtypes.float16, "F32": dtypes.float32, "U8": dtypes.uint8, "I8": dtypes.int8, "I32": dtypes.int32, "I64": dtypes.int64}
 inverse_safe_dtypes = {v:k for k,v in safe_dtypes.items()}
@@ -67,6 +68,11 @@ def torch_load(fn:str):
     if storage[2] not in offsets: return None
     byte_offset = offsets[storage[2]]+storage_offset*storage[1].itemsize
     ret = t[byte_offset:byte_offset+prod(size)].cast(storage[1])
+    # convert bfloat16 -> float16 using LLVM
+    # upstream LLaMA also does this conversion:
+    # https://github.com/facebookresearch/llama/blob/6c7fe276574e78057f917549435a2554000a876d/llama/generation.py#L95
+    if storage[1] == dtypes.bfloat16:
+      ret = ret.to("LLVM").half().to(Device.DEFAULT).realize()
 
     # 7 lines to deal with permuted tensors. NOTE: this currently requires reading off the disk
     shape_strides = [(s, st) for s,st in zip(size, stride) if s != 1]
@@ -80,7 +86,7 @@ def torch_load(fn:str):
 
     return ret.reshape(size)
 
-  intercept = {"HalfStorage": dtypes.float16, "FloatStorage": dtypes.float32, "IntStorage": dtypes.int32, "LongStorage": dtypes.int64, "_rebuild_tensor_v2": _rebuild_tensor_v2}
+  intercept = {"HalfStorage": dtypes.float16, "FloatStorage": dtypes.float32, "BFloat16Storage": dtypes.bfloat16, "IntStorage": dtypes.int32, "LongStorage": dtypes.int64, "_rebuild_tensor_v2": _rebuild_tensor_v2}
   whitelist = {"torch", "collections", "numpy", "_codecs"}  # NOTE: this is not for security, only speed
   class Dummy: pass
   class TorchPickle(pickle.Unpickler):
--- a/examples/llama.py
+++ b/examples/llama.py
@@ -153,20 +153,24 @@ class Transformer:
 WEIGHTS_DIR = Path(__file__).parent.parent / "weights/LLaMA/"
 TOKENIZER_FILENAME = WEIGHTS_DIR / "tokenizer.model"
 VOCAB_SIZE = 32000
-
-args_small = {"dim": 512, "multiple_of": 256, "n_heads": 8, "n_layers": 8, "norm_eps": 1e-05, "vocab_size": VOCAB_SIZE}
-
-args_7B = {"dim": 4096, "multiple_of": 256, "n_heads": 32, "n_layers": 32, "norm_eps": 1e-06, "vocab_size": VOCAB_SIZE}
-WEIGHTS_7B_FILENAME = WEIGHTS_DIR / "7B/consolidated.00.pth"
-
-args_13B = {"dim": 5120, "multiple_of": 256, "n_heads": 40, "n_layers": 40, "norm_eps": 1e-06, "vocab_size": VOCAB_SIZE}
-WEIGHTS_13B_FILENAMES = [WEIGHTS_DIR / "13B/consolidated.00.pth", WEIGHTS_DIR / "13B/consolidated.01.pth"]
-
-args_30B = {"dim": 6656, "multiple_of": 256, "n_heads": 52, "n_layers": 60, "norm_eps": 1e-06, "vocab_size": VOCAB_SIZE}
-WEIGHTS_30B_FILENAMES = [WEIGHTS_DIR / "30B/consolidated.00.pth", WEIGHTS_DIR / "30B/consolidated.01.pth", WEIGHTS_DIR / "30B/consolidated.02.pth", WEIGHTS_DIR / "30B/consolidated.03.pth"]
-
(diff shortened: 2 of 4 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/state.py                          1               1
```

tinygrad/state.py
```diff
@@ -80,7 +80,7 @@ def _rebuild_tensor_v2(storage, storage_offset, size, stride, requires_grad, bac
 
     return ret.reshape(size)
 
-  intercept = {"HalfStorage": dtypes.float16, "FloatStorage": dtypes.float32, "IntStorage": dtypes.int32, "LongStorage": dtypes.int64, "_rebuild_tensor_v2": _rebuild_tensor_v2}
+  intercept = {"HalfStorage": dtypes.float16, "FloatStorage": dtypes.float32, "BFloat16Storage": dtypes.float16, "IntStorage": dtypes.int32, "LongStorage": dtypes.int64, "_rebuild_tensor_v2": _rebuild_tensor_v2}
```
geohot, line 83: Err, this isn't right. dtypes.float16 isn't bfloat

prusnak: Right. I was trying to find a way how to hack bfloat16 support into tinygrad, but I failed.

Simply adding a new dtype to `helpers.dtype` did not do the trick, because the backend does not know this.

Please advise, how to proceed.

It seems the correct way would be to use float32 as storage and use code similar to this one to convert bfloat16 to float32:

https://github.com/ggerganov/llama.cpp/blob/294f424554c1599784ac9962462fc39ace92d8a5/convert.py#L338-L341

dc-dc-dc, line 83: bfloat16 is 8bits exponent 7 bits fraction while float16 is 5bits exponent 10 bits fraction. Can you verify that this is not causing a accuracy loss in the weight values?

prusnak: It does not work at all unfortunately. Please check my comment above: https://github.com/tinygrad/tinygrad/pull/1284#discussion_r1268666825

geohot:
Do they all need bfloat16? Will merge without if they don't need it.

prusnak:
> Do they all need bfloat16? Will merge without if they don't need it.

The smallest one needs it, so I assume bigger ones need it too.

tinygrad/state.py
dc-dc-dc: Did a quick test locally and it matched up with pytorch but was pretty slow and double the memory footprint

```python
    if storage[1] == dtypes._bfloat16:
      temp = t[byte_offset:byte_offset+prod(size)].cast(dtypes.uint16)
      ret = Tensor((temp.numpy().astype(np.uint32) << 16).view(np.float32))
    else:
      ret = t[byte_offset:byte_offset+prod(size)].cast(storage[1])
```

prusnak: @dc-dc-dc I think that's better than nothing, we can always optimize later. Will you please push your changes into my PR branch? If that's not possible, will you please share your changes with me so I can update the PR?

geohot:
I think bfloat16 should be added in a separate PR. It's okay if only some backends support it.

tinygrad/state.py
dc-dc-dc: Not able to make a pr to your fork for some reason but I pushed the changes [here](https://github.com/dc-dc-dc/tinygrad/tree/llama-refactor)

prusnak:
I see the PR https://github.com/tinygrad/tinygrad/pull/1293 has been merged

Rebased on top of the current master and change commit https://github.com/tinygrad/tinygrad/pull/1284/commits/cfc227edb86e452b892a78bf989b95ea2c1fd0c5 to use `dtypes.bfloat16` for `BFloat16Storage`

prusnak:
Hmmm ... https://github.com/tinygrad/tinygrad/issues/1290#issuecomment-1642880556 suggets we should transform bf16->f32->f16 because the upstream LLaMA inference scripts use f16 anyway and therefore perform this conversion.

LMK how should we proceed with this PR.

prusnak:
I just wanted to know which way to proceed is preferred:
- bfloat16 implementation that works only on llvm
- converting bfloat16 to float16 which works everywhere

Since it seems that LLaMA upstream inference code does the latter, I will proceed with option 2 and update the PR to have working implementation.

prusnak:
I updated the PR with commit 5a6f288c71b0e12dc3cd3125308cde0e169dc22f which converts bfloat16->float32->float16 like suggested by @sekstini in https://github.com/tinygrad/tinygrad/issues/1290#issuecomment-1642880556

I confirm that with this commit the LLaMA-2 (`python3 examples/llama.py --gen 2`) works for me (M1) both with and without `MPS=1` set:

tinygrad/helpers.py
```diff
@@ -96,16 +96,19 @@ def fields() -> Dict[str, DType]: return DTYPES_DICT
...
-  uint64: Final[DType] = DType(2, 8, "unsigned long", np.uint64)
+  uint16: Final[DType] = DType(1, 2, "unsigned short", np.uint16)
+  uint32: Final[DType] = DType(2, 4, "unsigned int", np.uint32)
+  uint64: Final[DType] = DType(3, 8, "unsigned long", np.uint64)
 
   # NOTE: bfloat16 isn't supported in numpy
   bfloat16: Final[DType] = DType(0, 2, "__bf16", None)
 
   # NOTE: these are internal dtypes, should probably check for that
+  _bfloat16: Final[DType] = DType(0, 2, "bfloat16", None)
```
wozeparrot, line 111: Why can't the `bfloat16` dtype above be used?

prusnak: Because then it works only on LLVM and no other backend.

wozeparrot: The idea with it was to load the weights to LLVM cast to fp16, then move them to whatever device they will be used on.

prusnak: That sounds good. However, I have no idea how to do this because of my superficial knowledge of the tinygrad architecture. Feel free to take over this PR and update it to match your expectations.

sekstini: If we're using pytorch for the loading anyway, why not just do `.half()` after loading?
(I confirmed `x.half()` gave the same result as `x.float().half()`)

sekstini: Ah nevermind, I see we're using the tinygrad torch loader. My bad.

prusnak: > The idea with it was to load the weights to LLVM cast to fp16

@wozeparrot can you provide some hints how to do this so I can rework the PR?

wozeparrot: You should be able to do something like: `x.to("llvm").float().realize().to(Device.DEFAULT).realize()` where `x` is the loaded weight

prusnak: Hm, I tried the following:

``` python
    if storage[1] == dtypes.bfloat16:
      ret = t[byte_offset:byte_offset+prod(size)].to("LLVM").half().realize().to(t.device).realize()
    else:
      ret = t[byte_offset:byte_offset+prod(size)].cast(storage[1])
```

and it seems it modified weights on my disk :-/ Redownloading LLaMA-2-7b weights again ...

wozeparrot: oh yea can't do t.device because the weights are mmapped from disk, so the tensor's device is disk. So moving the weights to that device is the same as writing to disk.

prusnak: > so the tensor's device is disk

Learned the hard way, but it makes sense! :-)

prusnak: Pushed bf167b8db1fd5f3e5a2125b1d35f0abd9262ba0c which converts BFloat16 to Float16 via LLVM, but I am unable to get meaningful results (Stacy keeps talking jibberish).

I am once again out of ideas.

wozeparrot: maybe try something like:

`ret = t[byte_offset:byte_offset+prod(size)].contiguous().realize().to("LLVM").realize().half().realize().to(Device.DEFAULT).realize()`

prusnak: @wozeparrot nope, that didn't help

prusnak:
After the debugging session with @wozeparrot on Discord, we came up with the following commit:

https://github.com/tinygrad/tinygrad/pull/1284/commits/b42cddb410843e8a00a0a3fc906a55a328211eb4

Everything now works as expected (Stacy responds with meaningful answers), and we convert `bfloat16` to `float16` without the need to use NumPy.

tinygrad/state.py
```diff
@@ -67,6 +68,11 @@ def _rebuild_tensor_v2(storage, storage_offset, size, stride, requires_grad, bac
     if storage[2] not in offsets: return None
     byte_offset = offsets[storage[2]]+storage_offset*storage[1].itemsize
     ret = t[byte_offset:byte_offset+prod(size)].cast(storage[1])
+    # convert bfloat16 -> float16 using LLVM
+    # upstream LLaMA also does this conversion:
```
wozeparrot, line 72: So the one thing is that other model do use this, but I think its fine that we always convert bfloat16 to float16 for now, as we don't have great bfloat16 support.

PR #1288, opened 2023-07-19
fix CUDAProgram __init__ with DEBUG>=6 on Linux

alxshine:
Replace path generated in f-string by os.path.join

### DIFF
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -1,6 +1,7 @@
 import subprocess, time, re, hashlib, tempfile
 from typing import Optional
 import numpy as np
+from os.path import join
 from pycuda.compiler import compile as cuda_compile # type: ignore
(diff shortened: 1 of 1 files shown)


tinygrad/runtime/ops_cuda.py
```diff
@@ -1,6 +1,7 @@
 import subprocess, time, re, hashlib, tempfile
 from typing import Optional
 import numpy as np
+from os.path import join
```
geohot, line 4: Don't do this. `import os`

tinygrad/runtime/ops_cuda.py
```diff
@@ -1,6 +1,7 @@
 import subprocess, time, re, hashlib, tempfile
 from typing import Optional
 import numpy as np
+import os
```
geohot, line 4: Ugh

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_cuda.py               2               1
```

PR #1296, opened 2023-07-20
jit optimizations

nimlgen:
This MR introduces an optimization to JIT, enabling the optimization of buffers. As a result, it significantly reduces the space allocated on the GPU. Here are the performance results for CIFAR:

New Results with JIT Optimization:
```
  0 6341.29 ms run, 3484.22 ms python, 2857.08 ms CL,    2.92 loss, 0.009521 LR, 0.08 GB used,     23.08 GFLOPS
  1 5044.76 ms run, 1325.73 ms python, 3719.02 ms CL,    2.80 loss, 0.008464 LR, 1.09 GB used,     29.00 GFLOPS
  2 4382.71 ms run,   11.75 ms python, 4370.95 ms CL,    7.43 loss, 0.007408 LR, 1.09 GB used,     33.39 GFLOPS
  3 4391.83 ms run,   12.40 ms python, 4379.43 ms CL,    8.30 loss, 0.006351 LR, 1.09 GB used,     33.32 GFLOPS
  4 4386.60 ms run,   13.33 ms python, 4373.27 ms CL,    7.33 loss, 0.005294 LR, 1.09 GB used,     33.36 GFLOPS
  5 4389.28 ms run,   13.13 ms python, 4376.15 ms CL,    5.45 loss, 0.004237 LR, 1.09 GB used,     33.34 GFLOPS
  6 4387.37 ms run,   14.30 ms python, 4373.07 ms CL,    4.51 loss, 0.003180 LR, 1.09 GB used,     33.35 GFLOPS
  7 4245.75 ms run,    8.96 ms python, 4236.79 ms CL,    4.56 loss, 0.002124 LR, 1.09 GB used,     34.46 GFLOPS
  8 4382.56 ms run,   14.53 ms python, 4368.03 ms CL,    2.87 loss, 0.001067 LR, 1.09 GB used,     33.39 GFLOPS
  9 4380.30 ms run,   13.61 ms python, 4366.68 ms CL,    3.62 loss, 0.000010 LR, 1.09 GB used,     33.40 GFLOPS
```

Old:
```
  0 6336.27 ms run, 3776.85 ms python, 2559.41 ms CL,    2.92 loss, 0.009521 LR, 0.08 GB used,     23.09 GFLOPS
  1 5028.83 ms run, 1414.80 ms python, 3614.04 ms CL,    2.80 loss, 0.008464 LR, 1.96 GB used,     29.10 GFLOPS
  2 4401.00 ms run,   13.38 ms python, 4387.61 ms CL,    7.43 loss, 0.007408 LR, 1.96 GB used,     33.25 GFLOPS
  3 4406.28 ms run,   12.69 ms python, 4393.59 ms CL,    8.30 loss, 0.006351 LR, 1.96 GB used,     33.21 GFLOPS
  4 4440.72 ms run,   15.21 ms python, 4425.52 ms CL,    7.33 loss, 0.005294 LR, 1.96 GB used,     32.95 GFLOPS
  5 4391.67 ms run,   12.93 ms python, 4378.74 ms CL,    5.45 loss, 0.004237 LR, 1.96 GB used,     33.32 GFLOPS
  6 4399.21 ms run,   13.91 ms python, 4385.30 ms CL,    4.52 loss, 0.003180 LR, 1.96 GB used,     33.26 GFLOPS
  7 4255.84 ms run,    9.25 ms python, 4246.59 ms CL,    4.57 loss, 0.002124 LR, 1.96 GB used,     34.38 GFLOPS
  8 4410.97 ms run,   13.68 ms python, 4397.29 ms CL,    2.88 loss, 0.001067 LR, 1.96 GB used,     33.17 GFLOPS
  9 4379.89 ms run,   13.89 ms python, 4366.00 ms CL,    3.62 loss, 0.000010 LR, 1.96 GB used,     33.41 GFLOPS
```

With this optimization, we observe substantial memory savings of up to 0.8 GB and reduced memory usage during execution by up to 55% compared to the previous implementation without the optimization. 

This implementation uses a greedy approach which is not 100% efficient for all cases. Also we can get more savings using reordering of some kernels.


geohot:
Is there a way to use a caching allocator to do this?

nimlgen:
Yeah, I thought of making it not-JIT only. I think it's possible, but the only pitfall is the following. We have the following code:
```
z = x@y
... (1)
i = z@a
... (2)
w = i@b
print(w.numpy())
print(z.numpy())
```
We want to minimize allocations and memory working set. We could easily do this since tinygrad is lazy. In this case, to calculate `w`, we have to calculate `z`. What if we can optimize memory allocation and reuse the buffer of `z` after `i` is calculated, since `z` is not used anymore? This allows us to calculate `w` with less memory used. But in this example, if the user wants to access `z` after `w`, it might not be calculated (since its buffer was used for other computations). Is it fine to launch more kernels to calculate `z` again?

geohot:
Yea the generic would be a caching allocator. torch has one, tinygrad used to have one, there's a few challenges though (like when do you free?)

geohot:
We definitely need this, Stable Diffusion can't be jitted right now.

nimlgen:
Yeah, just tried some approaches, and it seems it is possible to move this into lazy.py.
Simple cases work fine. I was worried about `x.numpy()` in this example, but it seems to work fine without additional kernels.
```
for i in range(10):
    x = x@w
    w = a@x
print(w.numpy())
print(x.numpy())
print(GlobalCounters.max_mem_used // 1e6)
```
And for this example, the maximum memory allocated on the GPU at the same time dropped from 83 MB to 12 MB.

As for the implementation, I have currently come up with a kind-of `PromiseBuffer` which is returned instead of a real device buffer. I decided to go with that since it is pretty complex to manage something like ref counting for buffers with current implementation (since we do optimizations right in the middle of realization). With `PromiseBuffer`, we get a linear set of kernels (the set of kernels for the whole lazy-area, i.e., all the kernels from the previous `realize()` (in user code) to the current one). After that, kernels are executed with buffer optimization, and `PromiseBuffers` are replaced with real buffers for every `LazyBuffer` which is alive outside of this lazy-area.

I'm not sure if `PromiseBuffer` is the final solution, but it is the first one that works :)

geohot:
I put up a $500 bounty for this.

"caching allocator, all backends, must intelligently free, reduce memory usage in both JIT and thneed"

So yea, it might be something like PromiseBuffer. The other benefit of a caching allocator we need is not actually calling malloc all the time. Malloc is slow.

nimlgen:
As of now, the changes look like this:

1) Promised realization in lazy.py. We just postpone the execution of kernels until we know all of them. This allows us to control which LBs are referred to and which are not, and free them as soon as we do not need them (when I tried to do so inside the tree, it was messier since we want to apply optimizations during realization). Since we have all the kernels to run and information about their dependencies on each other, I think we could do some optimizations here. For example, scheduling kernels to different GPUs(?). If this makes any sense, I believe this would be beneficial, as we kind of know the kernels which are dependent on us and so on.
2) Allocator. Currently, I implemented an LRUcache which should cache buffers depending on the memory available on a device. We have a lot of small allocations (<8MB), and basically the best way to cope with them is to allocate a huge pool and give pieces from it. My question is, do we want this? In my opinion, this is the best option to remove all small allocations, but it adds complexity to backends. I mean backends should support mapping buffers with the given offset (of course, if they do not support binding buffers with a given offset, we can do this manually in the linearizer). To not make tinygrad messy with an optimized allocator, I think it is possible to wrap a C++ implementation for Python and import it. I wrapped one (https://pypi.org/project/TLSFAllocator) and will test performance to see if there is any difference for this complexity.

As for the results of these patches, many real ML models show some peak memory savings.
model| peak old(mb) | peak new(mb) | diff
-- | -- | -- | --
train_efficientnet.py | 96 | 76 | -20,83%
hlb_cifar10.py (JIT) | 1017 | 472 | -53,59%
transformer.py | 11,7 | 11,1 | -5,13%
whisper.py | 22,9 | 12,3 | -46,29%
vit.py | 63,1 | 24,7 | -60,86%
stable_diffusion.py | 5666 | 4910 | -13,34%

Allocation count drops about half with LRUAllocator. With TLSFAllocator for smaller sizes, I get way better results on the paper (will see if this matters for performance, as I have said).

geohot:
These seem like different issues. Can we just merge the LRUcache first? How many of the gains does that get us?

nimlgen:
So, I added LRUAllocator and CacheCollector, which help reduce memory consumption and calls to allocation functions. For example, in jitted hlb_cifar10.py, the memory allocated for jit dropped from 1.02 GB to 0.73 GB. As for calls to malloc, they are reduced from ~5600 to ~3700 in train_resnet.py.

CacheCollector is essential for reducing memory consumption in jitted parts, as the previous version held references to all buffers, preventing them from being reused. The new CacheCollector allows buffers to be freed when safe, efficiently optimizing buffers while ensuring correct final output without requiring changes from other parts collecting cache. It also handles circular dependencies like LSTM to maintain proper functionality.

I think this is ready for first review.

geohot:
Is there any way this PR can be broken up?

PR #1298, opened 2023-07-20
Fix max nan

uncommonSensor:
* Adds nan check to Tensor.max()
* Calls to max can pass in "ignore_nan=True/False" argument
* Added max nan CI tests

GPU=1 python3 test/test_speed_v_torch.py TestSpeed.test_max on 3090

"ignore_nan=False"
**This Patch:**
`
max   4096x 4096 3.79 ms (4.43 GFLOPS 17.71 GB/s) in torch, 9.52 ms (1.76 GFLOPS  7.05 GB/s) in tinygrad, 2.51x slower      16.78 MOPS 67.11 MB`
.
**Master:**
`max   4096x 4096    3.72 ms (4.51 GFLOPS 18.04 GB/s) in torch, 9.52 ms ( 1.76 GFLOPS  7.05 GB/s) in tinygrad,  2.56x slower  16.78 MOPS    67.11 MB`

**Torch anomoly detection** : [CLASStorch.autograd.set_detect_anomaly(mode, check_nan=True)](https://pytorch.org/docs/stable/autograd.html#torch.autograd.detect_anomaly)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         3               1
```

tinygrad/tensor.py
```diff
@@ -379,8 +379,14 @@ def _reduce(self, fxn:Type[Function], axis:Optional[Union[int, Tuple[int, ...]]]
     return ret if keepdim else ret.reshape(shape=shape)
 
   def sum(self, axis=None, keepdim=False): return self._reduce(mlops.Sum, axis, keepdim)
-  def max(self, axis=None, keepdim=False): return self._reduce(mlops.Max, axis, keepdim)
   def min(self, axis=None, keepdim=False): return -((-self).max(axis=axis, keepdim=keepdim))
+  def max(self, axis=None, keepdim=False, ignore_nan=True):
```
geohot, line 383: What does torch do? Do they have `ignore_nan`?

uncommonSensor: I found this:
(https://pytorch.org/docs/stable/autograd.html#torch.autograd.detect_anomaly)

Seems useful for granularity.

tinygrad/tensor.py
```diff
@@ -379,8 +379,8 @@ def _reduce(self, fxn:Type[Function], axis:Optional[Union[int, Tuple[int, ...]]]
     return ret if keepdim else ret.reshape(shape=shape)
 
   def sum(self, axis=None, keepdim=False): return self._reduce(mlops.Sum, axis, keepdim)
-  def max(self, axis=None, keepdim=False): return self._reduce(mlops.Max, axis, keepdim)
   def min(self, axis=None, keepdim=False): return -((-self).max(axis=axis, keepdim=keepdim))
+  def max(self, axis=None, keepdim=False, ignore_nan=True): return self._reduce(mlops.Max, axis, keepdim) if ignore_nan else self._reduce(mlops.Max, axis, keepdim) + (self.isnan() * np.nan).sum()
```
geohot, line 383: https://pytorch.org/docs/stable/generated/torch.fmax.html

wozeparrot: The correct function isn't actually `fmax` (`fmax` is the nan ignoring version of elementwise `maximum`), it should be `nanmax` but that's not implemented in torch yet: https://github.com/pytorch/pytorch/issues/61474

wozeparrot:
So this breaks the backwards pass when `ignore_nan=False` due to the use of `!=` in `Tensor.isnan` which uses `==` in it's implementation which doesn't have a defined backwards pass.

uncommonSensor:
Apologies on missing the 'ignore_nan=False' tests.

 I ran some tests on the different variations of 'max' in pytorch, tinygrad, and numpy:

**Notes:**
- tinygrad max doesn't handle nans from what I can tell. So it's currently most like torch.fmax, except torch.fmax takes two tensors and gets max element-wise (ignoring nans). 
- np.nanmax() uses reduce but only returns nan if all elements in a tensor are nan, otherwise it ignores nan elements.

The original ticket was written up because tinygrad max is not handling nans, like numpy max is:
```
Tensor([1, np.nan]).numpy().max() -> np.nan
Tensor([1, np.nan]).max().numpy() -> array([1.], dtype=float32)

Wrong output shape aside, it should be nan, not 1.
```

It seems like the best option is either to add a nan check to the existing mlops Max class, or create a new max function. Calling it fmax could be confusing though.

**Pytorch:**
```
t1 = torch.tensor([2, 1, np.nan, np.nan])
t2 = torch.tensor([np.nan, 5, 3, np.nan])

torch.max(t1).numpy()          -> array(nan, dtype=float32)
torch.max(t1, t2).numpy()      -> array([nan,  5., nan, nan], dtype=float32)
torch.maximum(t1, t2).numpy()  -> array([nan,  5., nan, nan], dtype=float32)
torch.fmax(t1, t2).numpy()     -> array([ 2.,  5.,  3., nan], dtype=float32)
np.nanmax(t1)                       -> 2.0
```

**Tinygrad:**
```
t1 = Tensor([2, 1, np.nan, np.nan])
t2 = Tensor([np.nan, 5, 3, np.nan])

Tensor.max(t1).numpy()               -> array(2., dtype=float32)
Tensor.max(t1, t2).numpy()          -> array(2., dtype=float32)  **(This is comparable torch.fmax)**
Tensor.maximum(t1, t2).numpy() -> array([ 2.,  5.,  3., nan], dtype=float32)
```

**Numpy:**
```
t1 =np.array([[np.nan, 5, 3, np.nan], [2, 2, 2, np.nan]])
np.nanmax(t1) -> 5.0

t1 = np.array([np.nan, np.nan, np.nan, np.nan]) 
np.nanmax(t1) -> nan

t1 = torch.tensor([1, np.nan, 5, 3])
np.nanmax(t1) -> 5.0

t1 = torch.tensor([np.nan, np.nan, np.nan, np.nan])
np.nanmax(t1) ->  <stdin>:1: RuntimeWarning: All-NaN axis encountered -> **nan**

```

PR #1299, opened 2023-07-20
add check global dim limit in linearizer

g1y5x3:
Currently the `gidx0-3` is not aligned with `grid.x,y,z` in `cuda` in  which caused #1224. By performing an additional reshape, this issue got resolved.
Having this error when I was trying to run `test/models/test_real_world.py` with `CUDA`
```
kernel.cu(23): error: no suitable user-defined conversion from "float4" to "half4" exists
```
will update once I resolve that but here's some metrics on CIFAR, tested with batch size as much as 3090 can handle.
```
GPU=1 CUDA=0 BS=512 STEPS=1000 python examples/hlb_cifar10.py
999  313.88 ms run,    3.06 ms python,  310.82 ms CL,    0.82 loss, 0.000010 LR, 8.85 GB used,   3720.15 GFLOPS
eval 9006/10000 90.06%,    0.41 val_loss STEP=1000

GPU=1 CUDA=1 BS=512 STEPS=1000 python examples/hlb_cifar10.py
 49  301.26 ms run,    2.55 ms python,  298.71 ms CL,    1.98 loss, 0.009143 LR, 7.57 GB used,   3875.96 GFLOPS
eval 3642/10000 36.42%,    2.12 val_loss STEP=50
 50  302.38 ms run,    2.76 ms python,  299.62 ms CL,    2.10 loss, 0.009326 LR, 8.85 GB used,   3861.67 GFLOPS
...
999  309.64 ms run,    2.43 ms python,  307.21 ms CL,    0.83 loss, 0.000010 LR, 8.85 GB used,   3771.06 GFLOPS
eval 9019/10000 90.19%,    0.41 val_loss STEP=1000

GPU=1 CUDA=1 BS=1024 STEPS=1000 python examples/hlb_cifar10.py
999  625.92 ms run,    3.60 ms python,  622.32 ms CL,    0.15 loss, 0.000010 LR, 16.34 GB used,   3730.96 GFLOPS
eval 9045/10000 90.45%,    0.39 val_loss STEP=1000
```


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py             7               4
```

tinygrad/codegen/linearizer.py
```diff
@@ -607,10 +607,13 @@ def limit_global_dims(self, limit):
...
-    if limit and (self.first_reduce-self.local_dims) > limit:
-      num_to_merge = ((self.first_reduce-self.local_dims) - limit)+1
-      self.reshape_and_permute(lambda x: (prod(x[0:num_to_merge]),)+x[num_to_merge:], None)
-      if DEBUG >= 3: print("reshaped to", self.full_shape, "due to too many global dimensions")
+    if limit:  
+      if (self.first_reduce-self.local_dims) > limit:
+        num_to_merge = ((self.first_reduce-self.local_dims) - limit)+1
+        self.reshape_and_permute(lambda x: (prod(x[0:num_to_merge]),)+x[num_to_merge:], None)
+        if DEBUG >= 3: print("reshaped to", self.full_shape, "due to too many global dimensions")
+      global_dims = self.first_reduce-self.local_dims
```
g1y5x3, line 615: I can probably add a `if Device.DEFAULT == "CUDA":` to pass CI but that doesn't feel right. Will look into why the additional reshape broke opencl.

tinygrad/codegen/linearizer.py
```diff
@@ -607,10 +607,13 @@ def limit_global_dims(self, limit):
...
+      self.reshape_and_permute(lambda x: x[0:global_dims][::-1]+x[global_dims:], None)
```
g1y5x3, line 616: I can probably add a if Device.DEFAULT == "CUDA": to pass CI but that doesn't feel right. Will look into why the additional reshape broke opencl.

tinygrad/codegen/linearizer.py
```diff
@@ -598,10 +598,12 @@ def limit_global_dims(self, limit):
     # sometimes, there's more dimensions than len(self.lang.gid).
     # compact all the dimensions into the first
     # NOTE: this might make multiview shapetrackers
-    if limit and (self.first_reduce-self.local_dims) > limit:
+    if (self.first_reduce-self.local_dims) > limit:
       num_to_merge = ((self.first_reduce-self.local_dims) - limit)+1
       self.reshape_and_permute(lambda x: (prod(x[0:num_to_merge]),)+x[num_to_merge:], None)
-      if DEBUG >= 3: print("reshaped to", self.full_shape, "due to too many global dimensions")
+    global_dims = self.first_reduce-self.local_dims
+    self.reshape_and_permute(lambda x: x[0:global_dims][::-1]+x[global_dims:], None)
```
geohot, line 605: This doesn't seem right. What is it doing?

g1y5x3: It's forcing the largest allocation to always be aligned with the `global.x` dim in cuda. It was also mis-aligned in opencl but I think it handles device limit internally.

geohot: I don't think it does that...it just flips the shape. Why is that one the largest?

g1y5x3: When it is large enough to exceed device limit, due to the flip from [render_kernel](https://github.com/tinygrad/tinygrad/blob/d2c1e8409abb281493e7aa5f7a8efb9055f6a1cf/tinygrad/codegen/cstyle.py#L85), it created  `global_size` such as `[2, 2, 1048576]`, `[16, 65536]` (later converted to `[16, 65536, 1]` by [ops](https://github.com/tinygrad/tinygrad/blob/d2c1e8409abb281493e7aa5f7a8efb9055f6a1cf/tinygrad/ops.py#L142)) which made those allocation to be aligned with `global.y` or `global.z`.

g1y5x3:
@geohot I have added the comparison of global dimension to there according limits. If it exceeds the absolute maximum I just assert it now cuz I feel like more complex reshape should be done in `optimizer` instead of `linearizer`.  I use permute instead of reshape to avoid creating multiple view shape tracker.

.gitignore
```diff
@@ -37,3 +37,4 @@ node_modules
 package.json
 package-lock.json
 temp
+wandb/
```
wozeparrot, line 40: extra?

g1y5x3: oh my bad, it was from the cifar branch that I added to ignore for conveniences

PR #1303, opened 2023-07-20
cache buffer loads across multiple bufs

chaosagent:
For winograd #1037; follow up to #1272. Different bufs can load the same stuff from the same rawbufs. The use case is Tensor.stack with the stack dimension unrolled.

PR #1272, opened 2023-07-18
dedup kernel args v2

chaosagent:
Now it is very nice. No CSE yet though.

Redo of #1141. All the real work was done in #1256.

PR #1304, opened 2023-07-20
CSE equivalent buffer-loads

chaosagent:
For winograd #1037; follow up to #1272; depends on #1303 to be useful. Different bufs can load the same stuff from the same rawbufs. The use case is Tensor.stack with the stack dimension unrolled.

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -390,10 +390,10 @@ class Linearizer:
           self.uop(UOps.BARRIER, None, [], ())
 
         # load earlybufs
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py            10              10
```

geohot:
These PRs are really hard to follow. Can you show an example of what it changes in the generated code? Is this the minimal way to achieve that change?

chaosagent:
This is on hold until I can get an actually performant Winograd kernel to be generated. I'll post a writeup on this sometime. Thanks!

PR #1308, opened 2023-07-21
Update arange to be (start, stop, step)

madt2709:


sekstini:
The most common case is to only pass stop, so requiring stop as a keyword argument seems wrong. We should match what PyTorch and Numpy are doing imo.

wozeparrot:
hmm, this makes the base case of `arange` a bit more annoying to use/verbose.

Do we maybe want to overload the base case? I think torch does this.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/__init__.py                    1               1
tinygrad/tensor.py                         4               2
total                                      5               3
```

sekstini:
Here's a simplified version of how PyTorch handles it:

```python
def arange(start=0, end=None, step=1):
  if end is None:
    end = start
    start = 0
  ...
```

madt2709:
Ah thanks, I knew there had to be a smarter way! will update now

tinygrad/tensor.py
```diff
@@ -152,7 +152,11 @@ def zeros(*shape, **kwargs): return Tensor.full(argfix(*shape), 0, **kwargs)
   def ones(*shape, **kwargs): return Tensor.full(argfix(*shape), 1, **kwargs)
 
   @staticmethod
-  def arange(stop, start=0, step=1, **kwargs): return Tensor.full((ceil((stop-start)/step),), step, **kwargs).cumsum() + (start - step)
+  def arange(start=0, stop=None, step=1, **kwargs):
+    if stop is None:
+      stop = start
+      start = 0
```
wozeparrot, line 158: Could use `if stop is None: stop, start = start, 0`

tinygrad/tensor.py
```diff
@@ -152,7 +152,11 @@ def zeros(*shape, **kwargs): return Tensor.full(argfix(*shape), 0, **kwargs)
   def ones(*shape, **kwargs): return Tensor.full(argfix(*shape), 1, **kwargs)
 
   @staticmethod
-  def arange(stop, start=0, step=1, **kwargs): return Tensor.full((ceil((stop-start)/step),), step, **kwargs).cumsum() + (start - step)
+  def arange(start=0, stop=None, step=1, **kwargs):
```
wozeparrot, line 155: Can we make `start` required?

madt2709: ah yes good point. 

(sorry being really slow on this pr)

tinygrad/tensor.py
madt2709: sure, less lines for sure

wozeparrot:
Much better, will merge if tests pass.

PR #1315, opened 2023-07-21
throw NotImplementedError instead of RuntimeError

zahash:
throw NotImplementedError instead of RuntimeError as default implementation for back prop method


tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
https://en.wikipedia.org/wiki/Wikipedia:Chesterton's_fence

Issue #1316, opened 2023-07-21
Negative min/max or rhs in Symbolic


ronak69:
Edit: This is actually a minor bug rather than a major issue. I apologize for the false positive.

Symbolic **is** agnostic to div/mod implementation of backends.

> Even though `Variable()` constructor doesn't allow negative `min/max`, still there are
> some bugs in div and mod implementation.
> 
> ### Floor division
> 
> * Two kinds of floor divisions:
>   1. round towards zero (C) `-21//5 => -4.2 => -4`
>   2. round towards negative infinity (Python) `-21//5 => -4.2 => -5`
> 
> * As mod depends on floor division, two mod operations as well.
> 
> * A div expression: `lhs//rhs`
> 
> * These two only differ if **one** of `lhs` or `rhs` operand to div is negative.
>   * `ShapeTracker` doesn't use negative `rhs` values for div and mod.
>   * But `lhs`'s `min/max` can be negative because of negative values in `strides`.
>     Or by using operations like `a-10` or `a*-1` which clearly produce negative
>     `min/max`.
> 
> * So symbolic's output is currently **not** agnostic to the div implementation of
>   a backend (or any program that evaluates the rendered expression).
> 
> * Symbolic should be able to produce output for both kinds of implementations and
>   let backends choose which one it wants. Another option is to select one implementation
>   in symbolic and let backends match/convert their implementation to symbolic's.
> 
> * Yet another option is to make sure users of symbolic never pass negative `lhs` to div or mod.
> 

### Conflicting tests!

* First test expects C like div and second test expects Python like.
* `7//-2` equals `-4` in Python and `-3` in C

https://github.com/tinygrad/tinygrad/blob/9746f6d09412241adb6ce9b64e06cbd0a46e8689/test/unit/test_symbolic.py#L110-L111

* `-90//11` equals `-9` in Python and `-8` in C

https://github.com/tinygrad/tinygrad/blob/9746f6d09412241adb6ce9b64e06cbd0a46e8689/test/unit/test_symbolic.py#L207-L208

> 
> ### Questions
> 
> * What are the exact requirements of the symbolic library?
>   Constraints on `min`, `max` and `rhs`?
> * Which kind of div and mod do each of the backends implement?
>

ronak69:
#1345 fixes this issue as well as the "infinite recursion in mod if negative variables" issue.

chenyuxyz:
no symbolic anymore

PR #1317, opened 2023-07-21
Negative variables in Symbolic

ronak69:
Read Issue #1316

* Implement floor div and mod which "rounds towards -infinity."
* Pass the negative tests which were failing.
* Negative min/max are now allowed.
* Add more tests.
* Factor out a function for performing common assertions/optimizations.

Welcome back negatives!

### DIFF
--- a/test/unit/test_symbolic.py
+++ b/test/unit/test_symbolic.py
@@ -90,7 +90,7 @@ class TestSymbolic(unittest.TestCase):
     self.helper_test_variable(Variable("a", 0, 8)*1, 0, 8, "a")
 
   def test_mul_neg_1(self):
-    self.helper_test_variable((Variable("a", 0, 2)*-1)//3, -1, 0, "((((a*-1)+3)//3)+-1)")
+    self.helper_test_variable((Variable("a", 0, 2)*-1)//3, -1, 0, "((a*-1)//3)")
 
   def test_mul_2(self):
     self.helper_test_variable(Variable("a", 0, 8)*2, 0, 16, "(a*2)")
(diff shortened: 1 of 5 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                43              29
```

geohot:
I support this direction, but tests are failing.

ronak69:
As i said in the Issue #1316, i think the reason is mismatched floor division implementations.

Tests for Clang *ubuntu* fails but for *windows* it doesn't, so it must be some architecture related bug right?

llvm's [sdiv](https://llvm.org/docs/LangRef.html#sdiv-instruction) instruction rounds towards zero. So the llvm codegen needs to replace all div operations with a custom routine that rounds towards -infinity.

But i don't understand codegen yet... but i will try.

geohot:
I don't understand what this gains. There's now backend specific code too? I'm fine with generally improving symbolic to support negatives, but this is a lot of code and I don't see the benefit.

ronak69:
The issue is that currently symbolic has a "hybrid" implementation of div/mod that either rounds towards zero or -infinity.

The solution is to select one or the other. This PR selected -infinity. But now i believe that it was a wrong choice as all of the hardware in the world rounds towards zero. This is why the tests failed: backends were rounding towards zero but symbolic was expecting them to round towards -infinity. Thus, for the tests to pass i had to make backends round towards -infinity. This turned out to be more complex and costly than i expected.

I am now working on making symbolic confirm to rounding toward zero instead. So no more backend specific code there.

This PR is a solution but not a merge-able one.

ronak69:
I was wrong. Symbolic doesn't need to select one kind of div/mod because there is a way to make lhs be always positive (getting rid of differences in two kinds of div/mod) which is already implemented.

PR #1322, opened 2023-07-21
(WIP) Stable Diffusion on Web

schlimeszn:
Use the stable diffusion code and port it into the webgpu canvas element. Pretty sure this can be done purely in python with no JavaScript. The canvas element is already implemented in the webgpu package.


wozeparrot:
Not enough for a bounty lock.

PR #1323, opened 2023-07-22
flake8

waifairer:
Added the following to flake8:
E124 [Closing bracket does not match visual indentation](https://www.flake8rules.com/rules/E124.html)
E272 [Multiple spaces before keyword (E272)](https://www.flake8rules.com/rules/E272.html)
E303 [Too many blank lines (3) (E303)](https://www.flake8rules.com/rules/E303.html)

Moved configs to .flake8 so text editors pick it up. Enabled flake8 on tests except for a few commonly occurring rule violations.


tinyb0t:
Changes made in `tinygrad/`:
```
total                                      2               3
```

wozeparrot:
I think E117 is already handled by pylint. How many of the flake8 checks does pylint already do?

waifairer:
@wozeparrot that's a good question! pylint also isn't run on tests right now, but most of the failing rules are things that we wouldn't want to fail, anyway. Let me see how much is worth fixing and I'll reorganize it for less duplicate checking. 

To my knowledge, pylint does not have a rule about multiple newlines which is why I went the flake8 direction.

waifairer:
@wozeparrot this commit https://github.com/waifairer/tinygrad/commit/ad10cf375f3801f2ddf84794a14c703ff41964c8 contains the results of running pylint on the test dir. I fixed a few of the suggestions and ignored the rest. Ultimately I ended up ignoring:

```
W0102: Dangerous default value {} as argument (dangerous-default-value)
W0104: Statement seems to have no effect 
W0108: Lambda may not be necessary (unnecessary-lambda)
W0404: "Reimport" lines 
W0602: Using global for 'prefix' but no assignment is done (global-variable-not-assigned)
W0611: Unused partial imported from functools (unused-import)
W0612: Unused variable 'out' (unused-variable) 
W0640: Cell variable i defined in loop (cell-var-from-loop) 
R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return) 
E1102: torch_conv is not callable (not-callable)E1102 
E1133: Non-iterable value test is used in an iterating context (not-an-iterable)
```

Mostly I ignored things that were false positives or just annoying. Ultimately, it appears that there is no overlap between the additional rules added in this MR and the pylint rules already used by the repo.

waifairer:
Actually, I just realized that I made a mistake when adding E117 to the flake8 config. Neither pylint nor flake8 are finding any indent issues. I'll remove E117 from the flake8 checklist. 

As for next steps, we could either merge one, both, or none of these MRs. Though I do have some preference for getting this flake8 one merged.

wozeparrot:
Can you update the precommit hook as well?

waifairer:
> Can you update the precommit hook as well?

Thanks for the catch, updated.

wozeparrot:
Thanks!

PR #1329, opened 2023-07-23
Gather

geohotstan:
The reason for the change in tinygrad/runtime/ops_cpu.py is because  https://github.com/tinygrad/tinygrad/pull/1101#issuecomment-1642630256
The patch is quite ugly but I found no other way of fixing it :(
Any suggestions?

I'm not sure if `Tensor.gather()` is considered a `movement hlops`? so should I place it somewhere else?
Also, I could split the gather up into multiple lines to help with readability lol

### DIFF
--- a/tinygrad/runtime/ops_cpu.py
+++ b/tinygrad/runtime/ops_cpu.py
@@ -17,7 +17,6 @@ base_fxn_for_op: Dict[Op, Callable] = {
 }
 
 def promote_types(x, y): return ret if (ret := np.promote_types(x.dtype, y.dtype)) != np.float64 else np.float32
-
 def match_types(x, y):
   up = x.dtype if dtypes.from_np(x.dtype).priority > dtypes.from_np(y.dtype).priority else y.dtype
   return x.astype(up, copy=False), y.astype(up, copy=False)
--- a/tinygrad/runtime/ops_cpu.py
+++ b/tinygrad/runtime/ops_cpu.py
@@ -16,7 +16,7 @@ base_fxn_for_op: Dict[Op, Callable] = {
   MovementOps.RESHAPE: lambda x, arg: x.reshape(arg), MovementOps.SHRINK: lambda x, arg: x[tuple(slice(p[0], p[1], None) for p in arg)],
 }
 
-def promote_types(x, y): (ret := np.promote_types(x.dtype, y.dtype)) if ret != np.float64 else np.float32
+def promote_types(x, y): return ret if (ret := np.promote_types(x.dtype, y.dtype)) != np.float64 else np.float32
 
 def match_types(x, y):
   up = x.dtype if dtypes.from_np(x.dtype).priority > dtypes.from_np(y.dtype).priority else y.dtype
--- a/tinygrad/runtime/ops_cpu.py
+++ b/tinygrad/runtime/ops_cpu.py
@@ -16,6 +16,10 @@ base_fxn_for_op: Dict[Op, Callable] = {
   MovementOps.RESHAPE: lambda x, arg: x.reshape(arg), MovementOps.SHRINK: lambda x, arg: x[tuple(slice(p[0], p[1], None) for p in arg)],
 }
 
+def promote_types(x, y):
+  ret = np.promote_types(x.dtype,y.dtype)
+  return ret if ret != np.float64 else np.float32
+
 def match_types(x, y):
   up = x.dtype if dtypes.from_np(x.dtype).priority > dtypes.from_np(y.dtype).priority else y.dtype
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_cpu.py                5               1
tinygrad/tensor.py                         2               0
total                                      7               1
```

tinygrad/tensor.py
```diff
@@ -313,6 +313,8 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
         final_shape.append(1)
     return sliced_tensor.reshape(tuple(final_shape))  # Reshape
 
+  def gather(self, indices, dim): return (self.reshape(*self.shape[:dim+1], *[1]*indices.ndim, *self.shape[dim+1:]).expand(*self.shape[:dim+1], *indices.shape, *self.shape[dim+1:]) * (Tensor.arange(self.shape[dim], dtype=dtypes.int32).reshape(self.shape[dim], *[1]*indices.ndim).expand(self.shape[dim], *indices.shape) == (indices < 0).where(indices+self.shape[dim], indices).unsqueeze(0).expand(self.shape[dim], *indices.shape)).reshape(*[1]*dim, self.shape[dim], *indices.shape, *[1]*(self.ndim-dim-1)).expand(*self.shape[:dim+1], *indices.shape, *self.shape[dim+1:])).sum(dim)
```
wozeparrot, line 316: I think the more readable version of this would be better here. We can sacrifice line count a little bit for readability because this isn't very readable.

tinygrad/runtime/ops_cpu.py
```diff
@@ -16,6 +16,10 @@ def shape_to_axis(old_shape:Tuple[int, ...], new_shape:Tuple[int, ...]) -> Tuple
   MovementOps.RESHAPE: lambda x, arg: x.reshape(arg), MovementOps.SHRINK: lambda x, arg: x[tuple(slice(p[0], p[1], None) for p in arg)],
 }
 
+def promote_types(x, y):
+  ret = np.promote_types(x.dtype,y.dtype)
+  return ret if ret != np.float64 else np.float32
```
wozeparrot, line 21: I think this is fine, matching torch behaviour should be fine.

Could do:
`def promote_types(x, y): (ret := np.promote_types(x.dtype, y.dtype)) if ret != np.float64 else np.float32`

wozeparrot:
Can this do fancy indexing as in #967?

tinygrad/tensor.py
```diff
@@ -313,6 +313,14 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
         final_shape.append(1)
     return sliced_tensor.reshape(tuple(final_shape))  # Reshape
 
+  def gather(self, idx, dim):
+    idx = (idx < 0).where(idx+self.shape[dim], idx) # Turn neg idx pos
+    new_shape = [*self.shape[:dim+1], *idx.shape, *self.shape[dim+1:]]
+    new_self = self.reshape(*self.shape[:dim+1], *[1]*idx.ndim, *self.shape[dim+1:]).expand(new_shape)
+    arange = Tensor.arange(self.shape[dim], dtype=dtypes.int32).reshape(*[1]*dim, self.shape[dim], *[1]*idx.ndim, *[1]*(self.ndim-dim-1)).expand(new_shape)
```
wozeparrot, line 320: should `requires_grad=False` on the arange here.

wozeparrot:
Backwards pass should work with this on some cases I think, where `idx` doesn't need a gradient. Probably should add a test for this.

geohotstan:
okay I tried at this for an entire day...
Looks like it's a bit harder than I thought 😄 
I got it to the point where

```
# I'm using "..." to replace the massive mess of reshape args cuz it causes eye pain.

def multi_gather(self, idx: List[Tensor], dim: List[Tensor]):
  new_self = self.reshape(...)      # no need for expand because __eq__ and __mul__ calls _broadcast() and that deals with it
  arange = [Tensor.arange(self.shape[d], dtype=dtypes.int32, requires_grad=False).reshape(...) for i, d in zip(idx, dim)]
  new_idx = [i.reshape(...) for i, d in zip(idx, dim)]
  ret = reduce(Tensor.__mul__, [new_self] + [a == n for a,n in zip(arange, new_idx)])
  for d in dim: ret = ret.sum(d)
  return ret
```


Running realize() this on the example given in #967 crashes my computer 😄 (pytorch runs fine so there has to be some other trick)
There is also still some bug for weird indexes like `input_[:, b, :, d]` that's giving wrong answers...
*smaller/simpler test cases work

going to throw in the towel on this for now.
will figure out backward pass~~

Issue #1321, opened 2023-07-21
Temporary `ContextVar` values


dosier:
**Proposal**: add way to temporarily change context var values, e.g. 
```py
with Debugging(level=4, graph=1): // on enter set the context vars to temp values, on exit, change them back
   // some ops
// some other ops we don't care about for the debugging
```
I find myself wanting to debug some specific part of the code sometimes and it'd make it easier to navigate the logs. I saw there is a Context class which supports temporarily adding to the var stack, but that doesn't work if the key already exists in the stack.

AlexTelon:
If you are talking about `tinygrad.helpers.Context` and `tinygrad.helpers.ContextVar` then I think it already works as you describe. At least on master I can run the below code without issue.

```python
from tinygrad.helpers import Context, DEBUG, ContextVar

FOO = ContextVar("FOO", 0)


assert FOO.value == 0
assert DEBUG.value == 0

with Context(FOO=4, DEBUG=1):
    assert FOO.value == 4
    assert DEBUG.value == 1

assert FOO.value == 0
assert DEBUG.value == 0
print('all passed')
```

But again I'm new to tinygrad so I might be missing something important here.

dosier:
> If you are talking about `tinygrad.helpers.Context` and `tinygrad.helpers.ContextVar` then I think it already works as you describe. At least on master I can run the below code without issue.
> 
> ```python
> from tinygrad.helpers import Context, DEBUG, ContextVar
> 
> FOO = ContextVar("FOO", 0)
> 
> 
> assert FOO.value == 0
> assert DEBUG.value == 0
> 
> with Context(FOO=4, DEBUG=1):
>     assert FOO.value == 4
>     assert DEBUG.value == 1
> 
> assert FOO.value == 0
> assert DEBUG.value == 0
> print('all passed')
> ```
> 
> But again I'm new to tinygrad so I might be missing something important here.

Ah yes you are correct there, I think we should probably refactor it to not use a stack lookup at all since its having a bit of a performance cost but that may be for another issue. Also migration of Graph to a context var would be good.

PR #1331, opened 2023-07-23
Refactored ContextVars

AlexTelon:
* Added tests
 * Moved ctx_stack from ContextVars to Context and renamed to stack
 * Simplified implementation
 * Context and Timing can now be used as decorators
 * Updated documentation
 

**Draft notes:**
Related to #844 and I believe this fits under the "Conceptual Cleanups" umbrella.
But with recent changes it now also introduces a new feature in that Context and Timing can be used as decorators.

Tell me if I should have split this up. I totally see that the Timing stuff could have been put in a separate PR from the rest.

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -33,21 +33,20 @@ def getenv(key, default=0): return type(default)(os.getenv(key, default))
 
 class Context(contextlib.ContextDecorator):
   stack: ClassVar[List[dict[str, int]]] = [{}]
-  _vars: ClassVar[Dict[str, Tuple[ContextVar, int]]] = {}
   def __init__(self, **kwargs): self.kwargs = kwargs
   def __enter__(self):
+    for k,v in self.kwargs.items(): ContextVar._cache[k].value = v
     Context.stack.append(self.kwargs)
-    for k, (o, _) in Context._vars.items(): o.value = self.kwargs.get(k, o.value)
   def __exit__(self, *args):
-    Context.stack.pop()
-    for k, (o, dv) in Context._vars.items(): o.value = Context.stack[-1].get(k, dv)
+    for k in Context.stack.pop(): ContextVar._cache[k].value = Context.stack[-1].get(k, Context.stack[0][k])
 
 class ContextVar:
+  _cache: ClassVar[Dict[str, ContextVar]] = {}
   __slots__ = "value"
   def __new__(cls, key, default_value):
-    if key in Context._vars: return Context._vars[key][0]
-    Context._vars[key] = (instance := super().__new__(cls), getenv(key, default_value))
-    instance.value = Context.stack[-1].get(key, default_value)
+    if key in ContextVar._cache: return ContextVar._cache[key]
+    instance = ContextVar._cache[key] = super().__new__(cls)
+    instance.value = Context.stack[0][key] = getenv(key, default_value)
     return instance
   def __bool__(self): return bool(self.value)
   def __ge__(self, x): return self.value >= x
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import contextlib
 import os, functools, platform, time, re
 from weakref import KeyedRef, ref
 from _weakref import _remove_dead_weakref # type: ignore
@@ -30,7 +31,7 @@ def fromimport(mod, frm): return getattr(__import__(mod, fromlist=[frm]), frm)
 @functools.lru_cache(maxsize=None)
 def getenv(key, default=0): return type(default)(os.getenv(key, default))
 
-class Context:
+class Context(contextlib.ContextDecorator):
   stack: ClassVar[List[dict[str, Any]]] = [{}]
   def __init__(self, **kwargs): self.kwargs = kwargs
   def __enter__(self): Context.stack.append({**Context.stack[-1], **self.kwargs})
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -5,6 +5,7 @@ from _weakref import _remove_dead_weakref # type: ignore
 import numpy as np
 from typing import Dict, Tuple, Union, List, NamedTuple, Final, Iterator, ClassVar, Optional, Callable, Any
 from math import prod # noqa: F401 # pylint:disable=unused-import
+from collections import ChainMap
 
 ShapeType = Tuple[int, ...]
 # NOTE: helpers is not allowed to import from anything else in tinygrad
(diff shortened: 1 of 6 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        9               8
```

wozeparrot:
Can you see if you can implement #1321 as part of these changes?

tinygrad/helpers.py
```diff
@@ -31,22 +32,20 @@ def fromimport(mod, frm): return getattr(__import__(mod, fromlist=[frm]), frm)
 def getenv(key, default=0): return type(default)(os.getenv(key, default))
 
 class Context:
-  def __init__(self, **kwargs): self.pvars = kwargs
-  def __enter__(self): ContextVar.ctx_stack.append({ **self.pvars, **{ key: ContextVar.ctx_stack[-1][key] for key in ContextVar.ctx_stack[-1].keys() if key not in self.pvars } })
-  def __exit__(self, *args): ContextVar.ctx_stack.pop()
+  stack: ChainMap = ChainMap()
```
geohot, line 35: I don't think importing this is worth it. It's harder to reason about.

wozeparrot:
Would merge just the test.

AlexTelon:
I am happy to force-push an update that only includes the tests if that is the only part you want.

But I have some ideas that I put down into code that I think might be useful.

`contextlib.ContextDecorator` can be used to make a contextmanager be usable as a decorator too without any extra work.

So I put that into `Context` and `Timing` too!

I added some examples in the current documentation and also showed how in one instance adding Timing as a decorator removed one level of indentation in `load_state_dict`.

I know this is outside the scope of the original PR so tell me if I should have just put the tests here instead and done a separate PR.

wozeparrot:
I don't think there is a specific reason why the `ContextVars` are set like that, but yea `.set()` would probably be better/clearer.

On the second point, probably not, `Context` is pretty clear in what it does and something like `Env` makes it seem like an environment variable when it may not be.

dosier:
Oh another thing u could look into is that looking up the value in the stack every time a context var is accessed is quite expensive, storing the value inside the var itself would make it a lot faster.

tinygrad/state.py
```diff
@@ -43,16 +43,16 @@ def get_state_dict(obj, prefix:str='', tensor_type=Tensor) -> Dict[str, Tensor]:
   return state_dict
 def get_parameters(obj) -> List[Tensor]: return list(get_state_dict(obj).values())
 
+@Timing("loaded weights in ", lambda et_ns: f", {GlobalCounters.mem_used/1e9:.2f} GB loaded at {GlobalCounters.mem_used/et_ns:.2f} GB/s")
```
geohot, line 46: Why did this change? Is this needed as part of this PR?

AlexTelon: I added `contextlib.ContextDecorator` to Context which makes it possible to use it as a decorator.
I was not sure if that would be a wanted feature but its only a `+1` line change so thought to suggest it.

And as Timing was another helper contextmanager it was a `+0` line change to add it to it as well.

I checked how `Timing` was used and noticed this one case which would be a natural fit for the decorator Timing pattern so felt natural to show how this feature would affect the code if used.

AlexTelon:
Great suggestion @dosier I have now locally developed what you suggested and it performs much better than what is on master.

tl;dr: Still working on it. it's faster now. here is a pretty picture

Figure_1

The reference comparisons are against 
**reference minimal**
```python
    class Derp():
        __slots__ = 'value'
        def __init__(self, value): self.value = value
        def __ge__(self, x): return self.value >= x
        def __bool__(self): return bool(self.value)
    DEBUG = Derp(0)

    [test('reference minimal solution') for test in [test_gt, test_bool]]
   ```
which unsurprisingly is equivalent to my new solution in speed and probably the best we can do in terms of a pure python solution for lookup?

**reference int**
```python
DEBUG = 0
```
Which shows how fast the old `DEBUG = getenv("DEBUG", 0)` solution would be on lookups again just for reference.

This is the code right now but I still want to work on it as I really would like to not store the default_value in the tuple but instead inside the stack. 
```python
class Context(contextlib.ContextDecorator):
  stack: ClassVar[List[dict[str, int]]] = [{}]
  _vars: ClassVar[Dict[str, Tuple[ContextVar, int]]] = {}
  def __init__(self, **kwargs): self.kwargs = kwargs
  def __enter__(self):
    Context.stack.append(self.kwargs)
    for k, (o, _) in Context._vars.items(): o.value = self.kwargs.get(k, o.value)
  def __exit__(self, *args):
    Context.stack.pop()
    for k, (o, dv) in Context._vars.items(): o.value = Context.stack[-1].get(k, dv)

class ContextVar:
  __slots__ = "value"
  def __new__(cls, key, default_value):
    if key in Context._vars: return Context._vars[key][0]
    Context._vars[key] = (instance := super().__new__(cls), getenv(key, default_value))
    instance.value = Context.stack[-1].get(key, default_value)
    return instance
  def __call__(self, x): self.value = x
  def __bool__(self): return bool(self.value)
  def __ge__(self, x): return self.value >= x
  def __gt__(self, x): return self.value > x
  def __lt__(self, x): return self.value < x
  ````

  
  That is much easier to solve when using `ChainMap`
  
  ```python
  class Context(contextlib.ContextDecorator):
    stack: ClassVar[ChainMap[str, int]] = ChainMap()
    _vars: ClassVar[dict[str, ContextVar]] = {}
    def __init__(self, **kwargs): self.kwargs = kwargs
    def __enter__(self):
        for k,v in self.kwargs.items(): Context._vars[k].value = v
        Context.stack.maps.insert(0, self.kwargs)
    def __exit__(self, *args):
        for k,v in Context.stack.maps.pop(0).items(): Context._vars[k].value = Context.stack.get(k, v)
```
as it will just handle arbritary nested contextmanagers without issue. Not that that really is important.

I also kindof like that in my chainmap solution here that the body of the __exit__ method is just one line. Would like to do that in the proposed solution above too if it does not get too code golfy.

Also above I choose to keep the same behaviour as the original but I would propose to remove `__call__` and not replace it with anything. We can now `DEBUG.value = 6` directly and that will just work.

I will push my stuff once I have cleaned up a bit. Not today though but wanted to give an update.

geohot:
Can you submit this as a minimal PR? Like that `Timing` change above shouldn't be needed, right? You can resubmit it after.

geohot:
Looking much better! This ready to merge?

AlexTelon:
Is the name and location of the test file ok?

If it is I will remove a TODO comment and then I think it's ready.

wozeparrot:
I think just putting it at `test/test_helpers.py` should be fine, it shouldn't conflict with extras cuz that's in its own directory.

PR #1333, opened 2023-07-23
Fix Tensor([3]) constant fold take 2

chenyuxyz:
#1178 
constant fold size 1 tensor by default. Also add a `force_buffer` arg to `Tensor` if we want to force it to be a buffer (only matter now in training with JIT).

now `Tensor(3)` and `Tensor([3])` always use the same `LoadOps`.

### DIFF
--- a/docs/abstractions.py
+++ b/docs/abstractions.py
@@ -120,7 +120,7 @@ from tinygrad.tensor import Tensor
 from tinygrad.ops import LazyOp, BinaryOps, LoadOps
 
 # the 2+3 from before
-result = Tensor([2]) + Tensor([3])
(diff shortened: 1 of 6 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/optim.py                       1               1
tinygrad/tensor.py                        10               5
total                                     11               6
```

geohot:
Discussed in Discord.

`force_buffer` makes `Tensor` more complex, and I don't see the downside to using `Tensor(3)` and `Tensor([3])` to disambiguate constants and buffers. If you want a buffer, give it a shape. If you really want a constant buffer with a shape `Tensor(3).reshape(1)`

PR #1337, opened 2023-07-24
Reshape into symbolic shape

chenyuxyz:
part of #1259

support reshape into symbolic shape like `Tensor.zeros(4,4).reshape(4, Variable('i', 1, 100))`
also add `ShapeTracker.inferr_shape(self, symbols)` to get shape from symbols

the value of the variable is inferred and stored with `lazydata`. The tests try to cover all types of ops after reshape and make sure `shape`, `inferred_shape`, and `symbols` are updated correctly.

only shape from int shape to single variable shape can add symbols, and reshape back to ints and reduce can remove symbols

caveat: need to keep the type of `shape` to be `Tuple[int]`, otherwise mypy complains about all the `prod(shape)` is bad. I don't think we want to change all to `prod(inferred_shape)` which is slower and confusing.

### DIFF
--- a/tinygrad/lazy.py
+++ b/tinygrad/lazy.py
@@ -9,6 +9,7 @@ from tinygrad.helpers import GRAPH, DEBUG, prod, getenv, DType, dtypes, flatten,
 from tinygrad.runtime.ops_cpu import RawNumpyBuffer
 from tinygrad.runtime.ops_disk import RawDiskBuffer
 from tinygrad.shape.shapetracker import MovementOps, ShapeTracker, View, get_contraction
+from tinygrad.shape.symbolic import Variable, Node, sym_vars
 from tinygrad.ops import Compiled, Interpreted, UnaryOps, BinaryOps, TernaryOps, ReduceOps, LoadOps, OpType, LazyOp
 from tinygrad.runtime.lib import RawBufferMapped, RawConst, RawBuffer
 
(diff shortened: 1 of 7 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                          33              18
tinygrad/shape/shapetracker.py             9               7
tinygrad/shape/symbolic.py                14               1
tinygrad/tensor.py                         8               3
total                                     64              29
lines added in the tinygrad folder: 35
```

tinygrad/lazy.py
```diff
@@ -93,23 +94,23 @@ def get_movementroot(root:LazyBuffer, allow_contiguous=False) -> LazyBuffer: ret
 def get_movementroot_contiguous(x:LazyBuffer) -> LazyBuffer: return get_movementroot_contiguous(cast(LazyBuffer, x.op.src[0])) if not x.realized and x.op.op == LoadOps.CONTIGUOUS else (get_movementroot(x, True) if x.optype == MovementOps and x.st.contiguous else x)
 
 lazycache: LightWeakValueDictionary = LightWeakValueDictionary()
-def create_lazybuffer(device:str, st:ShapeTracker, optype:OpType, op:LazyOp, dtype:DType):
+def create_lazybuffer(device:str, st:ShapeTracker, optype:OpType, op:LazyOp, dtype:DType, symbols:Dict[Variable, int]):
```
geohot, line 97: Why is this not just in the ShapeTracker?

tinygrad/shape/symbolic.py
```diff
@@ -173,12 +181,14 @@ def get_bounds(self) -> Tuple[int, int]:
...
 
 class DivNode(OpNode):
+  def infer(self, symbols): return sym_infer(self.a, symbols) // sym_infer(self.b, symbols)
   def __floordiv__(self, b: int, _=False): return self.a//(self.b*b) # two divs is one div
   def get_bounds(self) -> Tuple[int, int]:
     assert self.a.min >= 0 and isinstance(self.b, int)
     return self.a.min//self.b, self.a.max//self.b
 
 class ModNode(OpNode):
+  def infer(self, symbols): return sym_infer(self.a, symbols) % sym_infer(self.b, symbols)
```
geohot, line 191: What does `infer` do?

tinygrad/lazy.py
chenyuxyz: done - went back and worth on this one. It's cleaner in ShapeTracker

tinygrad/shape/symbolic.py
chenyuxyz: `infer` takes `symbols` and infer the value of the node. Used to infer symbolic shape to int shape. Added some unit tests for it

geohot: I don't really understand, I think there's a simpler way to think about this.

tinygrad/shape/shapetracker.py
```diff
@@ -132,17 +132,21 @@ def get_unsafe_resize_offset(strides, arg):
   return sum([s * x[0] for s, x in zip(strides,arg)])
 
 class ShapeTracker:
-  __slots__ = "views"
-  def __init__(self, shape:Union[ShapeTracker, Tuple[int, ...]], views:Optional[List[View]]=None):
+  __slots__ = "views", "symbols"
+  def __init__(self, shape:Union[ShapeTracker, Tuple[Union[Node,int], ...]], views:Optional[List[View]]=None, symbols:Optional[Dict[Variable, int]]=None):
```
geohot, line 136: Why is there a symbols dicts? Why is this not part of the shape?

chenyuxyz:
okay going to close this one after discord discussion - now I see a much simpler way

PR #1342, opened 2023-07-26
Fix RDNA Assembly

RogueGradient:
- fixed hand_coded_optimizations() call from when it got moved out of the linearizer
- fixed LoadOps/Const that got refactored
- fixed MemOps that got refactored
- fixed imports from when assembly got moved to extra
- new env "ROCM_LLVM_PATH" to enable "ROCM_LLVM_PATH" to be used. The default now is to use the systems  llvm and ld.lld

Tested with the following env set in the command `RDNA=1 DEBUGCL=5 DEBUG=5 GPU=1 OPTLOCAL=1 PRINT_KERNEL=1 `

<details>

<summary>A note for building LLVM on Ubuntu 20.04</summary>

## Basic Stuff
sudo apt update
sudo apt install python3-pip build-essential libssl-dev ninja-build

## Tinygrad
git clone git@github.com:tinygrad/tinygrad.git
cd tinygrad
python3 -m pip install -e .
pip install cloudpickle
pip install pyopencl[oclgrind]
cd ../

## CMake
wget https://github.com/Kitware/CMake/releases/download/v3.27.0/cmake-3.27.0-linux-x86_64.tar.gz
tar -zxvf cmake-3.27.0-linux-x86_64.tar.gz
cd ./cmake-3.27.0-linux-x86_64/bin
echo "export PATH=$(pwd):\${PATH}" >> ~/.bashrc
cd ../../

## LLVM
git clone --depth 1 git@github.com:llvm/llvm-project.git
cd llvm-project
cmake -S llvm -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_PROJECTS="llvm;clang;lld" -DLLVM_TARGETS_TO_BUILD="AMDGPU;X86"
ninja -C build
> Note: if you get `c++: fatal error: Killed signal terminated program cc1plus` increase avaible ram/swp memory

cd ./build/bin
echo "export PATH=$(pwd):\${PATH}" >> ~/.bashrc
cd ../../../

</details>


wozeparrot:
What we really need is the RDNA assembly backend tested in CI.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_gpu.py                1               1
```

RogueGradient:
Yeah I don't know if it's possible. I was looking into it and the only simulator for RDNA was [Navisim](https://michaeltshen.github.io/Files/NaviSim.pdf) which seems to be a research project but is available on [gitlab](https://gitlab.com/akita/navisim). However it was archived 7 months ago, doesn't have much documentation, and I'm not sure how up to date it is since I think it's mostly for the original RDNA spec.

One thing I could do is test that the RDNA asm code is properly generated by LLVM which would catch stuff like this but wouldn't actually run the code on the GPU or be able to pass the other tests. I haven't looked at the CI super closely but I think this would have to have it's own section to generate a setup like I have in the `building LLVM on Ubuntu 20.04` notes which would make the CI take longer.

geohot:
Getting through LLVM is a great first step for CI. It'll catch a lot of the regressions, I'd merge it with that

RogueGradient:
Just following up. I've started writing the unit tests for making sure the asm code is properly generated by LLVM. I'm currently porting the tests from `test_ops.py` to generate the asm code for each op and have 8 passing test. 

I should have all the tests done in the next day or two depending on how much free time I have.

RogueGradient:
Added tests for most the ops. Now just need to add them to CI.

However when I was looking into the CI I came accross [self hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners). If you have an extra machine available we could use it run the RDNA code on the GPU which would be more ideal. Doing this would make these tests redundant though.

geohot:
`test_rdna_asm.py` with all that binary is not the way. That's way too fragile.

geohot:
Still no, `test_rdna_asm` is tons of specific lines!

PR #1343, opened 2023-07-26
update editorconfig, enforce via CI

prusnak:
This PR enables more options in the editorconfig file:
- end of line = LF (unix)
- ~insert final newline = true~
- trim trailing whitespace = true
- charset = utf-8

I have not reformatted any existing files. The changes will be slowly applied by people editing the files with editors that support editorconfig.
Also there is no check (via editorconfig-checker) in the CI right now.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,16 @@
-# 2 space indentation
+# Top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file, utf-8 charset
+[*]
+end_of_line = lf
+trim_trailing_whitespace = true
+charset = utf-8
+
 [*.py]
 indent_style = space
-indent_size = 2
+
+# ignore binary files
+[{*.hwx,*.mlmodel,*.weights,*.golden}]
+end_of_line = unset
+trim_trailing_whitespace = unset
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -249,7 +249,7 @@ jobs:
           wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB | gpg --dearmor | sudo tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null
           echo "deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
           sudo apt update && \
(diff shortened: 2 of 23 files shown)


prusnak:
The CI failure is intermittent and not related to the PR.

wozeparrot:
Should probably add a check in CI.

prusnak:
> Should probably add a check in CI.

I didn't want to add a check in the CI, because then I would need to fix all the files now creating a large changeset containing only whitespace changes touching lot of files.

If this is preferred, let me know and I'll update the PR, I just thought this approach was better:

>  The changes will be slowly applied by people editing the files with editors that support editorconfig.

wozeparrot:
I think it's better if we just apply all the fixes now, and add the check in CI, so that this stuff is enforced. Basically only gonna merge style *enforcement* prs.

geohot:
I don't want to enforce 'insert final newline'

prusnak:
I pushed the requested changes into the repo.

The CI didn't run, because I assume it doesn't run when a contributor touches the github workflow file, so I couldn't have this tested in the github CI.

.github/workflows/test.yml
```diff
@@ -27,6 +27,9 @@ jobs:
       run: pip install -e '.[linting,testing]' --extra-index-url https://download.pytorch.org/whl/cpu
     - name: Repo line count
       run: python3 sz.py
+    - name: EditorConfig Checker
+      uses: editorconfig-checker/action-editorconfig-checker@main
+      run: editorconfig-checker
```
wozeparrot, line 32: Nope, the CI didn't run cuz the workflow file is broken, can't put the uses and the run into the same step.

prusnak:
> I don't want to enforce 'insert final newline'

Globally or for some specific files?

prusnak:
Updated the PR to not enforce newlines at end of files

tinyb0t:
Changes made in `tinygrad/`:
```
```

.editorconfig
```diff
@@ -1,4 +1,16 @@
...
+
+# Unix-style newlines with a newline ending every file, utf-8 charset
+[*]
+end_of_line = lf
+trim_trailing_whitespace = true
+charset = utf-8
+
 [*.py]
 indent_style = space
-indent_size = 2
```
wozeparrot, line 4: why was this removed?

prusnak: 1) Because it is now enforced by pylint.

2) Editorconfig does not handle this properly as there are cases when this strict rule needs to be broken: when function argument list is split into two lines, the second line can start on both even and odd column, depending on the function name.

PR #1345, opened 2023-07-26
Symbolic: Bug fix in div/mod for negative min/max

ronak69:
This fixes Issue #1316.

### DIFF
--- a/tinygrad/shape/symbolic.py
+++ b/tinygrad/shape/symbolic.py
@@ -77,7 +77,7 @@ class Node:
     assert b > 0
     if b == 1: return NumNode(0)
     if self.min >= 0 and self.max < b: return self
-    if self.min < 0: return create_node(ModNode(create_rednode(SumNode, [self, NumNode(-1*(self.min//b)*b)]), b))
+    if self.min < 0: return create_node(ModNode(self - (self.min//b)*b, b))
     return create_node(ModNode(self, b))
 
   @staticmethod
--- a/tinygrad/shape/symbolic.py
+++ b/tinygrad/shape/symbolic.py
@@ -63,7 +63,7 @@ class Node:
 
   def __floordiv__(self, b:int, factoring_allowed=True):
     assert b != 0
-    if b < 0: return (self//-b)*-1
+    if b < 0: return (self*-1//-b)
     if b == 1: return self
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                 2               2
```

tinygrad/shape/symbolic.py
```diff
@@ -77,7 +77,7 @@ def __mod__(self, b:int):
     assert b > 0
     if b == 1: return NumNode(0)
     if self.min >= 0 and self.max < b: return self
-    if self.min < 0: return (self - ((self.min//b)*b)) % b
+    if self.min < 0: return create_node(ModNode(create_rednode(SumNode, [self, NumNode(-1*(self.min//b)*b)]), b))
```
geohot, line 80: I really doubt this should be calling create node

ronak69: Did you mean create "red" node? I fixed that. (Nested SumNodes resulted if `self` was a SumNode)

Removing calls to `create_node` will be ok if NumNode implements it's own `__mod__` method.

geohot:
Closing, it's still calling `create_node`

ronak69:
@geohot

> Closing, it's still calling `create_node`

Why do you "doubt" calling `create_node`?

Just `assert b > 0 and self.min >= 0` if you really don't want to support negatives instead of having buggy logic.

ronak69:
I asked you for your reason, i give mine:

I tried to fix an error silently. Bad.

The issue with line 80:
* In both div and mod, an offset (`-((self.min//b)*b)`) is added to make the numerator positive
* Final expression is `(self - ((self.min//b)*b)) % b`
* But that invokes `SumNode`'s `__mod__` method, which will distribute the mod into sum and you end up with `self % b` which is what you started with so: infinite recursion.

Solutions:
* div solves it by using the `factoring_allowed` function parameter of `SumNode`'s div. But it's mod doesn't have it so maybe add it? (big diff)
* Just create `ModNode` explicitly. Which is what i did.

I wrapped `create_node` around it mainly because that is what the next line after it does. But functionally it makes `NumNode`s if min/max are equal.

PR #1346, opened 2023-07-26
Improve Symbolic fuzz tests

ronak69:
* remove hard-coded rng seed
* start expressions with a variable instead of zero
* generate "real" expressions for better debugging in failures
* huge performance improvement in random sample generation

### DIFF
--- a/test/external/fuzz_symbolic.py
+++ b/test/external/fuzz_symbolic.py
@@ -72,8 +72,9 @@ if __name__ == "__main__":
       real_expr = render_real_expr[t.__name__](real_expr, rng)
       rngs.append(rng)
     render = expr.render()
-    sample_size = min(100, (u1.max*u2.max*u3.max))
-    samples = [ (random.randint(u1.min, u1.max), random.randint(u2.min, u2.max), random.randint(u3.min, u3.max)) for _ in range(sample_size) ]
+    sample_size = min(101, (u1.max*u2.max*u3.max))
+    if sample_size < 101: samples = list(itertools.product(range(u1.min, u1.max+1), range(u2.min, u2.max+1), range(u3.min, u3.max+1)))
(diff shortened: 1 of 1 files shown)


chenyuxyz:
the random seed was hand coded for CI consistency, and sampling was done that way so for smaller space it's guaranteed to test all combinations. The original fuzz #1278 tests all combinations with small ranges and I am not sure sacrificing that is good

ronak69:
Fixed. Now all combinations will be tested for smaller spaces. And for bigger spaces it randomly generates value of each variable individually (faster).

Seed is now set conditionally. But don't you think "consistent" tests should be handled by `unit/test_symbolic.py` and fuzz tests should be allowed to test "random" inputs?

ronak69:
Note: used a prime number (closest to 100 is 101) of space/sample size because this guarantees cases where space size is exactly 100 not being considered as a bigger space.

chenyuxyz:
will let @geohot decide if this overall is an improvement. I don't see it and I think it makes tests harder to understand. Why `render_real_expr`?

the pr also changed the testing subjects, now it only tests expressions that start with "v1" and not testing expressions that start with a number

ronak69:
`render_real_expr` because if a test fails you get to see the difference between the rendered expression by symbolic and the "real" expression. For example, `"x*1 + 0"` is a real expression and `"x"` is symbolic's render of it.

And starting with a zero is essentially one wasted op. `0+1`, `0%3`, `0+v2` etc.

ronak69:
On CI consistency:

* use random seed everywhere
* on failure print the seed that was used
* go back in history testing with the "failure-causing" seed to find the commit that introduced the bug

Will this work?

geohot:
Closing, agree with @chenyuxyz

PR #1350, opened 2023-07-27
[WiP] testing max nan changes

uncommonSensor:
* added nan checking to mlops Max class
*  WHERE is returning -inf instead of nan due to another fix


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/mlops.py                          7               2
tinygrad/tensor.py                         1               0
total                                      8               2
```

uncommonSensor:
**More info:**

 I ran some tests on the different variations of 'max' in pytorch, tinygrad, and numpy:

**Notes:**
- tinygrad max doesn't handle nans from what I can tell. So it's currently most like torch.fmax, except torch.fmax takes two tensors and returns 'element-wise' max  (ignoring nans). 
- np.nanmax() uses reduce but only returns nan if all elements in a tensor are nan, otherwise it ignores nan elements.

The original ticket was written up because tinygrad max is not handling nans, like numpy max does:
```
Tensor([1, np.nan]).numpy().max() -> np.nan
Tensor([1, np.nan]).max().numpy() -> array([1.], dtype=float32)

Wrong output shape aside, it should be nan, not 1.
```

It seems like the best option is either to add a nan check to the existing mlops Max class, or create a new max function. Calling it fmax could be confusing though.

**Pytorch:**
```
t1 = torch.tensor([2, 1, np.nan, np.nan])
t2 = torch.tensor([np.nan, 5, 3, np.nan])

torch.max(t1).numpy()               -> array(nan, dtype=float32)
torch.max(t1, t2).numpy()           -> array([nan,  5., nan, nan], dtype=float32)
torch.maximum(t1, t2).numpy()       -> array([nan,  5., nan, nan], dtype=float32)
torch.fmax(t1, t2).numpy()          -> array([ 2.,  5.,  3., nan], dtype=float32)
np.nanmax(t1)                       -> 2.0
```

**Tinygrad:**
```
t1 = Tensor([2, 1, np.nan, np.nan])
t2 = Tensor([np.nan, 5, 3, np.nan])

Tensor.max(t1).numpy()              -> array(2., dtype=float32)
Tensor.max(t1, t2).numpy()          -> array(2., dtype=float32)  **(This is comparable torch.fmax)**
Tensor.maximum(t1, t2).numpy()      -> array([ 2.,  5.,  3., nan], dtype=float32)
```

**Numpy:**
```
t1 =np.array([[np.nan, 5, 3, np.nan], [2, 2, 2, np.nan]])
np.nanmax(t1) -> 5.0

t1 = np.array([np.nan, np.nan, np.nan, np.nan]) 
np.nanmax(t1) -> nan

t1 = torch.tensor([1, np.nan, 5, 3])
np.nanmax(t1) -> 5.0

t1 = torch.tensor([np.nan, np.nan, np.nan, np.nan])
np.nanmax(t1) ->  <stdin>:1: RuntimeWarning: All-NaN axis encountered -> **nan**

```

uncommonSensor:
WHERE returns -inf instead of nan

PR #1353, opened 2023-07-27
Symbolic Shape JIT main PR

chenyuxyz:
To try: `GPU=1 DEBUG=2 JIT=1 JIT_ATTN=1 python examples/llama.py --prompt "Hello." --count 10 --temperature 0` and `METAL=1 DEBUG=2 JIT=1 JIT_ATTN=1 python examples/llama.py --prompt "Hello." --count 10 --temperature 0` works too. 

~~UPDATE: this works! Try with `CLANG=1 DEBUG=2 JIT=1 JIT_ATTN=1 python examples/llama.py --prompt "Hello." --count 10 --temperature 0`~~

--

~~This WIP PR contains a working example for CLANG JIT 2d-matmul and attention qkv. I might need to redesign the symbolic structure based on feedbacks so not ready to merge yet.~~

~~looking for comments on~~
~~1. Change `GlobalCounter.cache` interface. Should I add a different cache for symbols or is the current implementation fine. I need to store the symbols used by AST because it cannot be inferred from rawbuffers alone.~~
~~2. Currently don't need the actual value of symbols anywhere other then runtime buffer generation because we only support functions with no symbol output. We still need some kind of infer for `op_estimate` and `mem_estimate`, and more importantly, `global_size` and `local_size` to support Metal.~~
~~3. Extend `MulNode` and `SumNode` to support general symbolic Add and Mul is not a good idea because ShapeTracker have strong assumption for the structure that it's always `Node * int` and `sum(Node * int)`. I don't see a clean way to directly reuse `MulNode` for the Symbolic*Symbolic case without breaking the `real_strides`. I can add a design for general Add and Mul is that directionally sounds correct.~~

~~Implementation details:~~
~~- Reshape into Variables (only support from all `int` to one variable). Inferred value of that variable is stored in `Variable.val` (re-using `Variable.b` is slightly confusing because b is not optional for other Node type).~~
~~- Each AST knows what symbols are needed and what values are by checking `v.val for buf in ast.buffers for s in buf.shape for v in sym_vars(s)`. Use symbols to generate kernel, and use the real value to form buffers for the first run.~~ ~~Stored the symbols used for each kernel in `cache`.~~
~~- JIT time we gather the symbol values from `arg.lazydata.st for arg in args` and generate the new input rawbuffer for symbols.~~
~~- I need to update the hash key for `Node` to add the values of variables because otherwise the cached ShapeTracker View function will return the variable with old value in JIT. We cannot use more granular id like `id(self)` because we still want `Variable("a") - Variable("a") = 0` in `factorize`.~~

kernel generated for `(3,i)@(i,5)`
```
void r_3_5_s0(float* restrict data0, const float* restrict data1, const float* restrict data2, const int* restrict SYM_i) {
int i = SYM_i[0];
for (int gidx0 = 0; gidx0 <= 2; ++gidx0) {
for (int gidx1 = 0; gidx1 <= 4; ++gidx1) {
    float acc0_0 = 0.0f;
    for (int ridx2 = 0; ridx2 <= (-1+i); ++ridx2) {
      float val1_0 = data1[((gidx0*i)+ridx2)];
      float val2_0 = data2[((ridx2*5)+gidx1)];
    } /* reduce */
    data0[((gidx0*5)+gidx1)] = acc0_0;
  }} /* global+local */
}
```


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/cstyle.py                 5               5
tinygrad/codegen/linearizer.py             9               5
tinygrad/codegen/llvmir.py                 3               3
tinygrad/codegen/optimizer.py              2               0
tinygrad/codegen/wgsl.py                   1               1
tinygrad/helpers.py                        1               1
tinygrad/jit.py                            6               5
tinygrad/lazy.py                          16               4
tinygrad/ops.py                            9               5
tinygrad/shape/shapetracker.py             6               6
tinygrad/shape/symbolic.py                 9               2
total                                     67              37
lines added in the tinygrad folder: 30
```

geohot:
SumNode is fine, but creating a SymbolicMulNode isn't the worst idea.

chenyuxyz:
SumNode is problematic with `cat` because a single `idx` can be a SumNode and `real_strides` cannot tell if it's from a single `idx` of SumNode or multiple `idxs`.

chenyuxyz:
so I accidentally made it work while trying to generate a bad example... Now the `cat` part and the `qkv` part are all JIT. You can try with `DEBUG=2 CLANG=1 JIT=1 JIT_ATTN=1 python examples/llama.py --prompt "Hello." --count 10 --temperature 0`

image

One limitation is that I cannot JIT the whole block. The shape in middle became `(SumNode(MulNode(SumNode))) // (SumNode(MulNode))` and I ran out of tricks to simply that. Will continue to improve this.

re: avoid the dereference. I thought `ClangProgram` takes buffers as input. Did you mean changing the `ops_clang.py` so that if input buffer has size 1, pass the value directly?

geohot:
Hmm, I see, the size 1 thing could even be generic. Eh, it's okay to leave it as a buffer for now, sounds like a separate improvement.

chenyuxyz:
Tried to perf this on my M1 Max. Ran each setup 3 consecutive times and plot the averaged time for each token for the second and the third run. The first run for each setup always introduces some weird spikes and I am not sure why

image

geohot:
Reviewing this is my project for tomorrow, it's a big change.

chenyuxyz:
The line count and files touched is big due to various type changes, jit_cache interface change and new test cases.

The core logic change is about < 10 lines each in `shapetracker.reshape`, `ops.py`, `LazyBuffer.get_symbol_buffers`, `jit.py`. It roughly maps to 1. how do we specify variable shape, 2. how do we build kernel with variables, 3. how do we get the variable value for exec, 4. what to store and replace in jit. And 50 lines of change in `symbolic.py` to support enough case when `b` can be a `Node`.

`test/test_symbolic_jit.py` should be representative to the whole picture. You can also try the llama JIT using `CLANG=1 DEBUG=2 JIT=1 JIT_ATTN=1 python examples/llama.py --prompt "Hello." --count 10 --temperature 0`.

Once we agree the whole flow is good and interface change is suitable, I can break this into smaller PRs to focus on readability and maintainability.

chenyuxyz:
only works for CLANG now. Metal requires compute the actual global size and local size, and not supported in this PR.

geohot:
I think the right strategy here is smaller PRs that add individual tested functionality first.

chenyuxyz:
ok will do

chenyuxyz:
updated GPU and METAL support. Tested with up to 100 tokens and look good.

the kernel is slower compared to recompiling now because the handcode optimization part that i disabled. Need to support more `(Node, Node)` math in `symbolic.py` to make it work. Will focus on this part next.

geohot:
Let's get this to the point where tests are all passing and you are happy with it. Like really ready to merge.

chenyuxyz:
yeah still need more work. will flip this to "Ready for review" only after it's ready for review

is it okay to post status update in draft PR or is it too noisy? i feel that plus a flaky ci test gave you the wrong impression that i asked you to review something broken

chenyuxyz:
note: perfed a bit. better handled variable DeviceBuffer. more tests. more `(Node, Node)` math and enabled most of the hand coded optimization. Now `JIT_ATTN=1 JIT=1` is 20% faster compared to `JIT=1` in `llama.py` with `--profile` on metal.
next: simplify variable value management, clean up, perf and rest of the kernel optimization, make one variable `llama.py` work

chenyuxyz:
note: squashed some for the earlier commits to make rebasing easier. I think this is now feature complete. Next few iteration will focus on cleaning up symbolic ops in `symbolic.py`, use `divmod` for Node etc.

kernel optimization matched the hand coded version, and the only difference is that we cannot loop unroll a symbolic axis. I find the result attention kernels very beautiful (left is jitted symbolic version, right is current unjitted version of a prime number)

image

chenyuxyz:
okay this is ready for review. it's a version that i am happy to read and maintain

only have a M1 Max to test - tested with
`CLANG=1`, `GPU=1`, `METAL=1`
no JIT, `JIT=1`, `JIT_ATTN=1`, `JIT=1 JIT_ATTN=1`

and the max token length is hardcoded to 120 now. example commands:
`GPU=1 DEBUG=2 JIT=1 JIT_ATTN=1 python examples/llama.py --prompt "Hello." --count 20 --temperature 0`
`METAL=1 DEBUG=2 JIT=1 JIT_ATTN=1 python examples/llama.py --prompt "Hello." --count 20 --temperature 0`

@geohot can you test this on the perf machines?

tinygrad/shape/symbolic.py
```diff
@@ -75,7 +85,12 @@ def __floordiv__(self, b:int, factoring_allowed=True):
       return (self + -offset*b).__floordiv__(b, factoring_allowed=False) + offset
     return create_node(DivNode(self, b))
 
-  def __mod__(self, b:int):
+  def __rmod__(self, b:int): raise RuntimeError(f"not supported: {b} % {self}")
+  def __mod__(self, b:Union[Node,int]):
+    if isinstance(b, Node):
+      if self.min >= 0 and b - self.max > 0: return self
+      if self - b == 0: return NumNode(0)
```
ronak69, line 92: Is `self == b` different behavior?

geohot:
It's not 100% jitted on my Mac.
Screenshot 2023-08-07 at 9 55 49 PM

geohot:
But this is amazing!

```
ran model in 26.97 ms
sync in 80.24 ms
 a
ran model in 26.62 ms
sync in 80.45 ms
 
ran model in 27.34 ms
sync in 80.12 ms
2
ran model in 27.80 ms
sync in 80.64 ms
```

Metal still has a sync issue (unrelated to this), but on GPU you can see it's running the LLaMA core in ~27 ms!

geohot:
Want to submit the first piece as a PR I'll carefully review? I don't think they have to be as small as last time, but maybe first all the Symbolic/ShapeTracker changes?

chenyuxyz:
updated a version that passes int directly to kernel (GPU backend only for now, CLANG should be easy, don't know METAL yet)
`GPU=1 JIT=1 JIT_ATTN=1 PYTHONPATH=. python3.11 -O examples/llama.py --prompt "Hello." --count 10 --temperature 0 --timing`

```
ran model in 39.15 ms
sync in 53.83 ms
0
ran model in 38.96 ms
sync in 54.15 ms
 year
ran model in 38.62 ms
sync in 54.33 ms
 old
```

ran model time became higher though, will profile to see what's happening

chenyuxyz:
Updated to include passing `int` args directly to kernel functions for CLANG and METAL. Now it should be working again for CLANG, GPU, METAL.

chenyuxyz:
ok this is rebased now and only contains the final jit part. @geohot let me know if reviewing it here is fine or if you prefer a new PR.

Now it's single "JIT" flag. Tested with (CLANG, METAL, GPU) / (gen 1, 2) / (size 7B 13B)
`DEBUG=2 JIT=1 python3 examples/llama.py --prompt "Hello." --count 10 --temperature 0`
Wall time is about 10% faster compared to master. The "ran model" time with GPU is around 42ms on my mac, not as small as the previous implementation that passes a size 1 buffer. But the total speed is faster, so I am not sure how to interpret it.

The only thing that is not jitted is the freqs_cis.contiguous() part. I narrowed it down to `apply_rotary_emb` cannot be jitted. Removing contiguous works without JIT so it's a JIT only bug. And it probably has something to do with `apply_rotary_emb` creates a slice of slice, and `c` and `d` of `complex_mult` are coming from the same buffer. I could not reproduce with a simpler example though.

For now I just hard code the max length to 1024, which allocates about 10% more memory.

Also not sure how exporting model should work with this and I just ignore it.

geohot:
This is good to review, it's short enough.

geohot:
Can LLaMA be jitted as one complete thing? That's the main outcome I wanted, not the pieces. (I'm fine with ignoring the apply_rotary_emb bug)

tinygrad/jit.py
```diff
@@ -26,11 +26,13 @@ def __call__(self, *args, **kwargs) -> Any:
     assert len(input_rawbuffers) != 0, "no inputs to JIT"
     assert len(set(input_rawbuffers.values())) == len(input_rawbuffers), "duplicate inputs to JIT"
     if self.cnt >= 2:
+      var_vals = dict(sorted(merge_dicts([arg.lazydata.st.var_vals for arg in args if isinstance(arg, Tensor)]).items(), key=lambda kv: kv[0].key))
       for (j,i),(input_name, expected_size, expected_type) in self.input_replace.items():
-        assert input_rawbuffers[input_name].size == expected_size and input_rawbuffers[input_name].dtype == expected_type, f"size or type mismatch in JIT, {input_rawbuffers[input_name]} != <{expected_size}, {expected_type}>"
+        if not var_vals: assert input_rawbuffers[input_name].size == expected_size and input_rawbuffers[input_name].dtype == expected_type, f"size or type mismatch in JIT, {input_rawbuffers[input_name]} != <{expected_size}, {expected_type}>"
```
geohot, line 31: Hmm, this should still assert if you changed a shape you can't. Unless this is hard for some reason?

chenyuxyz: done. slightly more verbose but this now compares the shapes instead of sizes of the buffer, so it works for symbolic too

chenyuxyz:
added assert for export.

I merged together the `Attention.__call__` part, which I believed was separated into `pre/attn/post` only to achieve partial jit. Now it's directly using/jitting `Attention.__call__`. To make jit work, I need to make `cache_k` and `cache_v` an argument and update in `TransformerBlock`. The only limitation is that `reshape` to register variable and saving the cache cannot be inside a jitted function. It's possible to move all these from `TransformerBlock` to `Transformer` by maintaining all the layer caches, but I am not sure the additional complexity is worth it.

I need to update `symbolic.py` while merging the attention functions, mostly due to strides now can be a `NumNode` and need to check for those. Also made `SumNode // SumNode` more like long division, except that it stills only uses the constant terms. For a more general symbolic library we probably need true long division though. Also every place that I previously had slightly different check between `//` and `%` ended up biting me... so I just made sure it's checking the same criteria now, and updated the tests accordingly.

Tested with METAL/GPU/CLANG with the same commands.

geohot:
Running with DEBUG=2, many `not loading layers.`

```
ram used:  3.24 GB, layers.7.jitted_attn.fxn.wk.weight                :  23%|████████████████████████████████▍                                                                                                             | 96/420 [00:00<00:02, 152.03it/s]WARNING: not loading layers.7.jitted_attn.fxn.wk.weight
ram used:  3.24 GB, layers.7.jitted_attn.fxn.wv.weight                :  23%|████████████████████████████████▍                                                                                                             | 96/420 [00:00<00:02, 152.03it/s]WARNING: not loading layers.7.jitted_attn.fxn.wv.weight
ram used:  3.24 GB, layers.7.jitted_attn.fxn.wo.weight                :  23%|████████████████████████████████▍                                                                                                             | 96/420 [00:00<00:02, 152.03it/s]WARNING: not loading layers.7.jitted_attn.fxn.wo.weight
```

It also got really really slow on my Mac for initial steps:
```
 malekafka@tubby:~/tinygrad$ METAL=1 JIT=1 python examples/llama.py --prompt "Hello." --count 10 --temperature 0 --timing
using METAL backend
using LLaMA-7B model
ram used: 13.48 GB, freqs_cis                                         : 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 420/420 [00:04<00:00, 89.43it/s]
loaded weights in 4707.61 ms, 13.48 GB loaded at 2.86 GB/s
Hello.
ran model in 1681.22 ms
sync in 618.99 ms
 I
ran model in 2871.97 ms
sync in 1407.55 ms
'
ran model in 5448.52 ms
sync in 726.93 ms
m
ran model in 4583.45 ms
sync in 49.66 ms
 a
ran model in 354.59 ms
sync in 10.96 ms
 
ran model in 77.27 ms
sync in 11.34 ms
2
ran model in 79.19 ms
sync in 11.24 ms
0
ran model in 78.24 ms
sync in 11.25 ms
 year
ran model in 78.07 ms
sync in 11.24 ms
 old
ran model in 81.10 ms
sync in 9.92 ms
```

geohot:
Also, why are there new changes to symbolic? Separate PR for them?

chenyuxyz:
@geohot can you test again? I fixed the `not loading layers.` and the slow unjitted run time (due to dropping a performance critical early `realize()`).

If the performance and `llama.py` looks good, I will send out a separate PR for the new `symbolic.py` change. It's added today.

geohot:
Looks good.

```
loaded weights in 5190.84 ms, 13.48 GB loaded at 2.60 GB/s
Hello.
ran model in 2071.72 ms
sync in 203.52 ms
 I
ran model in 497.51 ms
sync in 3.43 ms
'
ran model in 176.58 ms
sync in 4.92 ms
m
ran model in 78.78 ms
sync in 10.78 ms
 a
ran model in 75.47 ms
sync in 10.71 ms
...
 ```

**Comment by chenyuxyz** [comment]
I am not sure why the loading time is this long for you though. On my M1 Max
```
loaded weights in 1932.36 ms, 13.48 GB loaded at 6.97 GB/s
Hello.
ran model in 520.56 ms
sync in 43.74 ms
 I
ran model in 226.07 ms
sync in 4.30 ms
'
ran model in 165.05 ms
sync in 4.52 ms
m
ran model in 79.74 ms
sync in 10.82 ms
 a
ran model in 76.70 ms
sync in 10.86 ms
```

**Comment by geohot** [comment]
Oh that's fine and has nothing to do with this PR, I was just worried about the speed of the first couple steps.

**Comment by geohot** [comment]
Broken in CUDA:

```
batman@x840:~/xx/openpilot/tinygrad_repo$ JIT=1 python examples/llama.py --prompt "Hello." --count 10 --temperature 0 --timing
using CUDA backend
using LLaMA-7B model
ram used: 13.48 GB, freqs_cis                                         : 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 292/292 [00:01<00:00, 233.87it/s]
loaded weights in 1255.78 ms, 13.48 GB loaded at 10.73 GB/s
Hello.
ran model in 1496.05 ms
sync in 2.32 ms
 I
ran model in 449.23 ms
  File "examples/llama.py", line 441, in <module>
    logits = llama.model(Tensor([toks[start_pos:]]), start_pos).realize()[:, -1, :]
  File "examples/llama.py", line 160, in __call__
    h = h.sequential([functools.partial(layer, start_pos=start_pos, freqs_cis=freqs_cis, mask=mask) for layer in self.layers])
  File "/home/batman/xx/openpilot/tinygrad/tensor.py", line 701, in sequential
    def sequential(self, ll:List[Callable[[Tensor], Tensor]]): return reduce(lambda x,f: f(x), ll, self)
  File "/home/batman/xx/openpilot/tinygrad/tensor.py", line 701, in <lambda>
    def sequential(self, ll:List[Callable[[Tensor], Tensor]]): return reduce(lambda x,f: f(x), ll, self)
  File "examples/llama.py", line 130, in __call__
    output, cache_k, cache_v = self.jitted_attn(self.jitted_attention_norm(x), self.cache_k, self.cache_v, start_pos, freqs_cis, mask)
  File "/home/batman/xx/openpilot/tinygrad/jit.py", line 54, in __call__
    self.ret = self.fxn(*args, **kwargs)
  File "examples/llama.py", line 88, in __call__
    keys, values = repeat_kv(keys, self.n_rep).realize(), repeat_kv(values, self.n_rep).realize()
  File "/home/batman/xx/openpilot/tinygrad/tensor.py", line 94, in realize
    self.lazydata.realize()
  File "/home/batman/xx/openpilot/tinygrad/lazy.py", line 159, in realize
    self.realized = Device[self.device].exec_ast(self.op, output=self, **self._device_extra_args())
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 203, in exec_ast
    if k.key not in self.method_cache: self.method_cache[k.key] = get_program()
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 200, in get_program
    return self.to_program(k)
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 167, in to_program
    return ASTRunner(k.function_name, src, global_size, local_size,
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 132, in build
    self.clprg = runtime(self.name, self.prg, **self.runtime_args)
  File "/home/batman/xx/openpilot/tinygrad/runtime/ops_cuda.py", line 65, in __init__
    raise e
  File "/home/batman/xx/openpilot/tinygrad/runtime/ops_cuda.py", line 62, in __init__
    try: prg = cuda_compile(prg, target="ptx", no_extern_c=True, options=['-Wno-deprecated-gpu-targets']).decode('utf-8')
  File "/home/batman/.pyenv/versions/3.8.10/lib/python3.8/site-packages/pycuda/compiler.py", line 301, in compile
    return compile_plain(source, options, keep, nvcc, cache_dir, target)
  File "/home/batman/.pyenv/versions/3.8.10/lib/python3.8/site-packages/pycuda/compiler.py", line 154, in compile_plain
    raise CompileError(
pycuda.driver.CompileError: nvcc compilation of /tmp/tmp7lz61bs4/kernel.cu failed
[command: nvcc --ptx -Wno-deprecated-gpu-targets -arch sm_86 -I/home/batman/.pyenv/versions/3.8.10/lib/python3.8/site-packages/pycuda/cuda kernel.cu]
[stderr:
kernel.cu(1): error: identifier "pos" is undefined

kernel.cu(5): error: identifier "pos" is undefined

2 errors detected in the compilation of "kernel.cu".
]
```

**Comment by chenyuxyz** [comment]
rebased and ready for review. @geohot can you test this on CUDA again?

**Comment by geohot** [comment]
```
batman@x840:~/xx/openpilot/tinygrad_repo$ JIT=1 python examples/llama.py --prompt "Hello." --count 10 --temperature 0 --timing
using CUDA backend
using LLaMA-7B model
ram used: 13.48 GB, freqs_cis                                         : 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████| 292/292 [00:01<00:00, 235.61it/s]
loaded weights in 1246.58 ms, 13.48 GB loaded at 10.81 GB/s
Hello.
ran model in 1475.17 ms
sync in 2.81 ms
 I
ran model in 296.71 ms
  File "/home/batman/.pyenv/versions/3.8.10/lib/python3.8/site-packages/pycuda/driver.py", line 223, in _build_arg_buf
    gpudata = np.uintp(arg.gpudata)
AttributeError: 'int' object has no attribute 'gpudata'

During handling of the above exception, another exception occurred:

  File "examples/llama.py", line 441, in <module>
    logits = llama.model(Tensor([toks[start_pos:]]), start_pos).realize()[:, -1, :]
  File "examples/llama.py", line 160, in __call__
    h = h.sequential([functools.partial(layer, start_pos=start_pos, freqs_cis=freqs_cis, mask=mask) for layer in self.layers])
  File "/home/batman/xx/openpilot/tinygrad/tensor.py", line 701, in sequential
    def sequential(self, ll:List[Callable[[Tensor], Tensor]]): return reduce(lambda x,f: f(x), ll, self)
  File "/home/batman/xx/openpilot/tinygrad/tensor.py", line 701, in <lambda>
    def sequential(self, ll:List[Callable[[Tensor], Tensor]]): return reduce(lambda x,f: f(x), ll, self)
  File "examples/llama.py", line 130, in __call__
    output, cache_k, cache_v = self.jitted_attn(self.jitted_attention_norm(x), self.cache_k, self.cache_v, start_pos, freqs_cis, mask)
  File "/home/batman/xx/openpilot/tinygrad/jit.py", line 54, in __call__
    self.ret = self.fxn(*args, **kwargs)
  File "examples/llama.py", line 88, in __call__
    keys, values = repeat_kv(keys, self.n_rep).realize(), repeat_kv(values, self.n_rep).realize()
  File "/home/batman/xx/openpilot/tinygrad/tensor.py", line 94, in realize
    self.lazydata.realize()
  File "/home/batman/xx/openpilot/tinygrad/lazy.py", line 159, in realize
    self.realized = Device[self.device].exec_ast(self.op, output=self, **self._device_extra_args())
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 210, in exec_ast
    prg.exec(k.bufs, var_vals=output.st.var_vals)
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 138, in exec
    return self(rawbufs, var_vals, force_wait=force_wait)
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 144, in __call__
    if et := self.clprg((global_size + [1]*(3-len(global_size))) if global_size is not None else None,
  File "/home/batman/xx/openpilot/tinygrad/runtime/ops_cuda.py", line 81, in __call__
    self.prg(*[x._buf if isinstance(x, RawCUDABuffer) else x for x in args], block=tuple(local_size), grid=tuple(global_size))
  File "/home/batman/.pyenv/versions/3.8.10/lib/python3.8/site-packages/pycuda/driver.py", line 481, in function_call
    handlers, arg_buf = _build_arg_buf(args)
  File "/home/batman/.pyenv/versions/3.8.10/lib/python3.8/site-packages/pycuda/driver.py", line 225, in _build_arg_buf
    raise TypeError("invalid type on parameter #%d (0-based)" % i)
TypeError: invalid type on parameter #4 (0-based)
```

geohot:
Are the tests passing on CI CUDA? Are you sure they are running?

chenyuxyz:
weird - the error kinda make sense though. let me investigate

chenyuxyz:
@geohot so the correct way to pass `int` to pycuda is to wrap it with `np.int32` like pyopencl. I rented a 3090 and verified both tests and `llama.py` work.

The problem is that cuda emulator only accepts directly passing `int`. These tests were running in CI and passed, and it's now failing with the correct implementation... it's kinda annoying that it behaves differently from the real gpu. Should I just disable the tests in CI?

geohot:
Oh considering it's a one line change like that I would just add if CUDACPU

chenyuxyz:
cool, done. Both work now

geohot:
Congrats!

PR #1354, opened 2023-07-27
[Need feedback] Metal three_Conv Issue fix

### LINKED ISSUE #1302: Certain conv tests failing on METAL with TC=1
On up to date forked master branch, some of the tests are failing, `test_conv/test_first_three, test_optim/test_adam_high_lr, test_optim/test_adamw_high_lr`. Is this normal to fail in local machines (Macos METAL)? I just run the test folder in pycharm without any paramaters.

For the test_first_three, problem happens to be shift operation in [linearizer.py/hand_coded_optimizations](https://github.com/tinygrad/tinygrad/blob/8562b5a04f2fd2e39b7ddc0578abc2dfe377f774/tinygrad/codegen/linearizer.py#L666C9-L666C62). It tries to shift 4 with 8 which creates [0, 8] shape that cant be reshaped.

ZenginU:
Kinda fixes #1302 

Why this is working currently is that only for the test_first_three there is a 3 length axis_buffer. For all of the other cases it is always 1 so it does not matter if there is max or not. So this is not a legitimate fix but saves the day for now.

Also, I believe somehow Tensor.expand is broken

Using max instead of the previous implementation is just by trial error and observing, somehow it fixed METAL=1 TC=1

When changing ones -> rand in the test case code perfectly runs with older implementation

### DIFF
--- a/tinygrad/codegen/optimizer.py
+++ b/tinygrad/codegen/optimizer.py
@@ -119,8 +119,11 @@ def hand_coded_optimizations(k:Linearizer):
       if DEBUG >= 3: print("TENSOR CORES", axis_buf0, axis_buf1)
       k.use_tensor_cores = getenv("TC", 1) == 1  # TC=2 will do the shape ops without the WMMA
 
+      s0 = max(axis_buf0, key=lambda x:x[1])[0]
+      s1 = max(axis_buf1, key=lambda x:x[1])[0]
+
       # TODO: select axis in smart way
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/optimizer.py              4               1
```

geohot:
No tests, commented out line, "kinda" fixes

Issue #1355, opened 2023-07-27
test_net_speed/test_mnist throwing SIGTRAP


ZenginU:
With #1335, running test_net_speed/test_mnist throws SIGTRAP on METAL. Not throwing every time but most of the time. 

Test do not fail even with the SIGTRAP but Python quit unexpectedly shows up.

zboralski:
I can see a trap in `test/test_net_speed.py` as well.

The SIGTRAP we're encountering might be due to a "use after free". The `__del__` method is invoked when an object is about to be destroyed, and this can occur at unpredictable times. If GPU commands that use the buffer are still being dispatched after the buffer is released, this could result in a "use after free" error.

If we add a synchronization call within the `__del__` method of the `RawMetalBuffer`. This ensures that all GPU commands finish executing before the buffer is released. 

```python
class RawMetalBuffer(RawBufferMapped):
  def __init__(self, size:int, dtype:DType):
    super().__init__(size, dtype, METAL.device.newBufferWithLength_options_(size*dtype.itemsize, Metal.MTLResourceStorageModeShared))
  def __del__(self):
    METAL.synchronize()  # Ensure all GPU commands finish before releasing the buffer
    self._buf.release()
    super().__del__()
  def _buffer(self):
    METAL.synchronize()
    return self._buf.contents().as_buffer(self._buf.length())
```

wozeparrot:
#1335 has been reverted.

PR #1358, opened 2023-07-27
Synchronize Before Releasing Buffer in Metal Backend

zboralski:
This PR addresses an intermittent SIGTRAP issue observed in the Metal backend. We believe the issue is caused by attempting to release a Metal buffer while commands issued to that buffer are still in flight on the GPU.

To prevent this, we've added a call to `synchronize()` in the `__del__` method of `RawMetalBuffer`. This ensures that all commands have completed execution on the GPU before we release the buffer.

In the existing code, `synchronize()` was not called before `self._buf.release()`, which could lead to a situation where the buffer was released while GPU commands were still in flight. This behavior is not well-defined and could potentially cause the observed SIGTRAP.

This change should make the Metal backend more robust by ensuring that buffers are not released prematurely.

This fix resolves issue #1355.

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -1,7 +1,6 @@
 # pip3 install pyobjc-framework-Metal pyobjc-framework-Cocoa pyobjc-framework-libdispatch
 import os, subprocess, pathlib
 import Metal, Cocoa, libdispatch # type: ignore
-from typing import List, Any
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_metal.py              1               0
```

zboralski:
I'll redo the PR without all these commits.

PR #1362, opened 2023-07-27
Added consts value support to pad and pad2d

ZenginU:
Now you can pad with value

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -370,7 +370,7 @@ class Tensor:
     shape = [(0,s) for s in self.shape[:-(len(padding)//2)]]
     slc = shape + [(-p0, s+p1) for p0,p1,s in zip(padding[::2], padding[1::2], self.shape[::-1])][::-1]
     if value == 0: return self.slice(slc)
-    return self.slice(slc) + value*(1 - Tensor.ones(self.shape).slice(slc))
+    return self.slice(slc) + (value - Tensor.full(self.shape, value).slice(slc))
 
   @property
   def T(self) -> Tensor: return self.transpose()
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -370,7 +370,7 @@ class Tensor:
     shape = [(0,s) for s in self.shape[:-(len(padding)//2)]]
     slc = shape + [(-p0, s+p1) for p0,p1,s in zip(padding[::2], padding[1::2], self.shape[::-1])][::-1]
     if value == 0: return self.slice(slc)
-    return self.slice(slc) + value*(Tensor.ones(self.shape).slice(slc) == 0)
+    return self.slice(slc) + value*(1 - Tensor.ones(self.shape).slice(slc))
 
   @property
   def T(self) -> Tensor: return self.transpose()
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -241,7 +241,7 @@ class Tensor:
   # ***** movement hlops *****
 
   # NOTE: using slice is discouraged and things should migrate to pad and shrink
-  def slice(self, arg:Sequence[Optional[Tuple[int, int]]]) -> Tensor: return self.pad2d(flatten([(a[1] - s, -a[0]) if a is not None else (0,0) for s,a in zip(self.shape, arg)])[::-1])
+  def slice(self, arg:Sequence[Optional[Tuple[int, int]]]) -> Tensor: return self.pad2d(flatten(((a[1] - s, -a[0]) if a is not None else (0,0) for s,a in zip(self.shape, arg)))[::-1])
 
   # - Negative indices are taken relative to the end of the sequence, so X[-2] returns the 2nd-to-last element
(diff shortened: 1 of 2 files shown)


wozeparrot:
Should probably be done with `Tensor.pad` as well.

tinygrad/tensor.py
```diff
@@ -366,9 +366,11 @@ def unsqueeze(self, dim):
...
 
   # (padding_left, padding_right, padding_top, padding_bottom)
-  def pad2d(self, padding:Union[List[int], Tuple[int, ...]]):
-    slc = [(-p0, s+p1) for p0,p1,s in zip(padding[::2], padding[1::2], self.shape[::-1])][::-1]
-    return self.slice([(0,s) for s in self.shape[:-(len(padding)//2)]] + slc)
+  def pad2d(self, padding:Union[List[int], Tuple[int, ...]], value:int=0):
+    shape = [(0,s) for s in self.shape[:-(len(padding)//2)]]
+    slc = shape + [(-p0, s+p1) for p0,p1,s in zip(padding[::2], padding[1::2], self.shape[::-1])][::-1]
+    if value == 0: return self.slice(slc)
+    return self.slice(slc) + (Tensor.ones(self.shape).slice(slc) > 0).where(0, value) #ones for binary like map
```
geohot, line 373: Don't use where, just multiply by value.

tinygrad/tensor.py
```diff
@@ -366,9 +366,11 @@ def unsqueeze(self, dim):
     return self.reshape(self.shape[:dim] + (1,) + self.shape[dim:])
 
   # (padding_left, padding_right, padding_top, padding_bottom)
-  def pad2d(self, padding:Union[List[int], Tuple[int, ...]]):
-    slc = [(-p0, s+p1) for p0,p1,s in zip(padding[::2], padding[1::2], self.shape[::-1])][::-1]
-    return self.slice([(0,s) for s in self.shape[:-(len(padding)//2)]] + slc)
+  def pad2d(self, padding:Union[List[int], Tuple[int, ...]], value:int=0):
```
geohot, line 369: value should be float, not int

ZenginU:
@wozeparrot 
I can add this to pad but you can use pad2d as pad as well pad2d is not 2d actually

tinygrad/tensor.py
wozeparrot, line 373: `(Tensor.ones(self.shape).slice(slc) > 0)` isn't this always going to be true?

wozeparrot:
> @wozeparrot I can add this to pad but you can use pad2d as pad as well pad2d is not 2d actually

pad2d uses pad under the hood, but maybe we can leave pad as just a pure movement op and use pad2d as the wrapper

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         5               3
```

tinygrad/tensor.py
ZenginU: Nope, .slice adds 0 padding so there will be 1s and 0s, main reason where flips 1s with 0s.
Any suggestion for flipping without zero

wozeparrot: `(1 - Tensor.ones(self.shape).slice(slc)) * value`

ZenginU: Converted to value*( Tensor == 0)

ZenginU: Makes too much sense, my bad

ZenginU:
@wozeparrot 
I believe for under the hood operations we should just use pad and shrink directly and leave decision to user for pad2d as pad2d might add useless shrink under the hood

geohot:
Yea this should be in pad, not pad2d

ZenginU:
@geohot 
If we move it to the Tensor.pad then we might not be able to use it with the current pad2d as pad2d directly calls slice. 

So want me to change the pad2d along with the pad or leave it as it is.

g1y5x3:
Not necessarily, 
```
  def pad(self, arg:Tuple[Tuple[int, int], ...], mode='constant', value=None):
    if mode=='constant' and not value: return self._pad(arg)
```
where `_pad` is the mlop and won't be touched

ZenginU:
The problem is mlops pad cant do negative padding (shrink) but pytorch api can so we need shrink in pad api

ZenginU:
Moved value padding from pad2d to pad with pad2d support

geohot:
Is this a minimal change? Why is slice using pad2d?

ZenginU:
Slice is using pad2d because their implementation was mostly same, previously pad2d was using slice with some preprocessing.

Because slice has no "value" argument I moved slice to pad2d and move pad2d to slice.

Code except pad() is just for transfering from slice to pad2d so I believe it is minimal in order to implement "value" in pad()

I believe slice is still fully supported with pad2d

ZenginU:
@geohot If you want I can split this in to two PRs, one adding const padding to .pad(), second one moving pad2d to slice and adding const pad support to pad2d

geohot:
Yea I don't understand why slice is changing. Can you submit a minimal PR with only Pad and test?

PR #1363, opened 2023-07-27
Fix off by one error in View.expr_node

S-Lykles:
Noticed that the upper bound was off by one. Essentially the same code also appears on line 199, but there it does correctly include the -1.

### DIFF
--- a/test/unit/test_shapetracker.py
+++ b/test/unit/test_shapetracker.py
@@ -3,6 +3,7 @@ import unittest
 import numpy as np
 from tinygrad.helpers import prod, DEBUG
 from tinygrad.shape.shapetracker import ShapeTracker, View, get_contraction
+from tinygrad.shape.symbolic import Variable
 
 def shapetracker_getitem(st, val):
   locals = {"idx": val, "valid": 1}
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
Can you add a test that would have caught this?

S-Lykles:
Added a test for expr_node, has a bit of a pytest.mark.parameterize vibe to it. The loop seems out of style compared to the other tests but I need to check different idx Variables against different shape operations somehow.
The change I made to View.expr_node is tested through ShapeTracker.expr_node.
The default argument to ShapeTracker.expr_node is a bit weird as its a string: 
`def expr_node(self, idx='idx'):
    if idx.__class__ is str: idx = Variable(idx, 0, prod(self.shape)-1)`
Compared to the View.expr_idx:
`def expr_node(self, idx=None) -> Node:
    if idx is None: idx = Variable('idx', 0, prod(self.shape)-1)`
Ideally `expr_idxs` and `expr_node_mask` should really be tested as well.
I'm working out a small cleanup of shapetracker atm, okay if I add those tests there?

test/unit/test_shapetracker.py
```diff
@@ -119,6 +120,44 @@ def test_2(self):
...
 
+class TestIndexExpressions(unittest.TestCase):
+  def setUp(self):
+    self.st = ShapeTracker((10, 10))
+    self.numel = prod(self.st.shape)
+
+  def tearDown(self):
+    assert self.expr(self.default_idx()) == self.st.expr_node()[0]
+    for idx in [None, (0, 99), (7, 203), (2, 5), (0, 0), (0, self.numel-1), (self.numel, self.numel), (0, self.numel), (0, self.numel+1), (self.numel+100, self.numel+100)]:
+        if idx is not None:
```
geohot, line 131: We should have the linter catch this spacing issue

geohot:
Thanks!

PR #1226, opened 2023-07-11
Fix argfix

Andriamanitra:
argfix for 0 arguments got broken in #987, this fixes it and adds a couple of tests to make sure it keeps working

the author of the change says it was "statistical optimization" (https://github.com/tinygrad/tinygrad/pull/987#discussion_r1231584626) but in my benchmarks my version is comparable (possibly even a little bit faster) across all use cases:

```python
import timeit

def argfix_old(*x):
  if x[0].__class__ in {tuple, list}:
    try: return tuple(x[0])
    except IndexError: return tuple()
  return tuple(x)

def argfix_new(*x): return tuple(x[0]) if x and x[0].__class__ in (tuple, list) else x

N = 1_000_000
for argfix in (argfix_old, argfix_new):
    print(argfix.__name__)
    print(timeit.timeit("argfix(1, 2, 3)", number=N, globals={"argfix": argfix}))
    print(timeit.timeit("argfix(1)", number=N, globals={"argfix": argfix}))
    print(timeit.timeit("argfix((1,2,3))", number=N, globals={"argfix": argfix}))
    print(timeit.timeit("argfix([1,2,3])", number=N, globals={"argfix": argfix}))
```
```
argfix_old
0.13264382199849933
0.12719697097782046
0.1347857450018637
0.16105678299209103

argfix_new
0.10527223203098401
0.1038119489676319
0.11646945198299363
0.1526102180359885
```

I also included a couple of simplifications/optimizations that I ran into while reading the code

PR #1365, opened 2023-07-28
Fix argfix and add tests

karan0handa:
Previous implementation of `argfix` had unreachable code at line 18 and the exception wasn't being caught at the right spot (`x[0]` used in line 16 but exception inside `if` block). 

Credit goes to Andriamanitra, but he wasn't planning on finishing his PR, so I did it instead.
https://github.com/tinygrad/tinygrad/pull/1226


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        1               5
```

PR #1368, opened 2023-07-28
use ubuntu for clang ci test

### LINKED ISSUE #1366: CI is not testing CLANG properly
added a always fail test cast to `test_tensor.py` and CI still [passed](https://github.com/chenyuxyz/tinygrad/actions/runs/5691894156). Not sure if it's only for CLANG
```
@unittest.skipUnless(Device.DEFAULT == "CLANG", f"test CLANG")
class TestClang(unittest.TestCase):
  def test_always_fail(self):
    assert False
```

The recent CI time optimization [change](https://github.com/tinygrad/tinygrad/commit/a0965ee1988cb4bfd08e829d4cff1829fa591ff0) might be dropping tests. Also the CI log now is full of warnings like 
`D:\a\tinygrad\tinygrad\test\test_conv.py:6: PytestUnknownMarkWarning: Unknown pytest.mark.exclude_cuda - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    pytestmark = [pytest.mark.exclude_cuda, pytest.mark.webgpu]`

chenyuxyz:
Fix #1366. Root cause was that windows env was not set up correctly. It needs to be a powershell command writing to `$env.GITHUB_ENV` (https://stackoverflow.com/a/65110357). windows CI was using `CPU`. It was fine before #1252  because ubuntu was covering it. Also this test is the slowest now and it takes 8 minutes.

I changed the os to `ubuntu-latest` and added a guardrail check. Verified that guardrail works with windows [link](https://github.com/chenyuxyz/tinygrad/actions/runs/5695168286/job/15437766791), and the intentionally fail test indeed failed with clang+ubuntu [link](https://github.com/chenyuxyz/tinygrad/actions/runs/5695199334).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -205,7 +205,7 @@ jobs:
         backend: [llvm, clang, gpu, cuda]
 
     name: Tests on (${{ matrix.backend }})
-    runs-on: ${{ matrix.backend == 'gpu'  && 'ubuntu-20.04' || matrix.backend=='clang'&&'windows-latest'|| 'ubuntu-latest' }}
+    runs-on: ${{ matrix.backend == 'gpu'  && 'ubuntu-20.04' || matrix.backend=='clang'&&'ubuntu-latest'|| 'ubuntu-latest' }}
     timeout-minutes: 20
 
     steps:
@@ -263,6 +263,8 @@ jobs:
           sudo ninja install
       - name: Install dependencies
         run: pip install -e '.[testing${{matrix.backend=='llvm'&&',llvm'||matrix.backend=='cuda'&&',cuda'||''}}]' --extra-index-url https://download.pytorch.org/whl/cpu
+      - name: Check Device.DEFAULT
+        run: python -c "from tinygrad.lazy import Device; assert Device.DEFAULT in ['LLVM','CLANG','CUDA','GPU'], Device.DEFAULT"
       - name: Run pytest (not cuda)
         if: matrix.backend!='cuda'
         run: python -m pytest -n=auto test/ -k '${{matrix.backend=='llvm'&&'not (test_nn.py and test_conv_transpose2d)'||'test'}}' -m 'not exclude_${{matrix.backend}}'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,6 +76,27 @@ jobs:
     - name: Test EfficientNet
       run: curl https://media.istockphoto.com/photos/hen-picture-id831791190 | ./recognize | grep hen
 
+  testcpuwindows:
+    name: CPU on Windows
+    runs-on: windows-latest
+    timeout-minutes: 20
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v3
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.8
+    - name: Cache pip
+      uses: actions/cache@v3
+      with:
+        path: '~\AppData\Local\pip\cache'
+        key: testing
+    - name: Install Dependencies
+      run: pip install -e '.[testing]' --extra-index-url https://download.pytorch.org/whl/cpu
+    - name: Run pytest
+      run: python -m pytest -n=auto test/ -k 'test'
+
   testtorch:
     name: Torch Tests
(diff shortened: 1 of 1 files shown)


dc-dc-dc:
keep this test on windows as it was originally added to catch windows specific bugs

chenyuxyz:
See discussion in #1366 for context. This is about CLANG not being tested properly. I can add CPU+windows to keep the windows coverage

chenyuxyz:
kinda annoying to add since now CPU is combined with some other tests... mark as draft now and let me add windows. Agree we should keep windows tests coverage

dc-dc-dc:
Windows GitHub runner also comes with CLANG installed, the env variable fix seems like a easier solution versus adding a new test

chenyuxyz:
Added windows+cpu as a separate test to make sure we are not dropping test coverages. I don't know if clang+windows was really using clang when it was added, but I am pretty sure it's not on current CI. Adding support for that can be a separate change.

I don't want to rush it because the current CI structure is quite packed with many conditions, and the change needs to be carefully tested. But I do want to get this in because CI is currently skipping tests.

.github/workflows/test.yml
```diff
@@ -76,6 +76,27 @@ jobs:
...
+    - name: Cache pip
+      uses: actions/cache@v3
+      with:
+        path: '~\AppData\Local\pip\cache'
+        key: testing
+    - name: Install Dependencies
+      run: pip install -e '.[testing]' --extra-index-url https://download.pytorch.org/whl/cpu
+    - name: Run pytest
+      run: python -m pytest -n=auto test/ -k 'test'
+
```
wozeparrot, line 99: Do we need the windows cpu test?

chenyuxyz: This is the current CI behavior. Maybe the intention was to only have CLANG for windows, but it was configured incorrectly. I added these to not drop overall test coverage, and it can be a separate PR to fix if CLANG windows is what we want. The main goal of this PR is to fix clang tests that was dropped in a speed optimization PR.

wozeparrot: I think the original intention was to only have the windows clang test

chenyuxyz: Are you suggesting removing the windows CPU coverage? I can do that too. My first commit did not have this section.  @dc-dc-dc suggested that it's dropping windows coverage so I added it back as a standalone test.

The first commit #1011 was configured incorrectly, so we probably never have windows clang in CI. It just showed that even though the correct windows + clang might be easy to implement, it's very easy to miss and we need to verify carefully. That's why I don't want to bundle correcting windows + clang with bringing back ubuntu + clang in the same PR. I think fixing CI missing tests is more important / urgent than adding new CI functionalities.

wozeparrot: Just use this pr to fix windows+clang, because that's what's most important right now.

wozeparrot: I think @dc-dc-dc meant to keep the clang test on windows, not keep the windows+cpu test, tho correct me if im wrong

chenyuxyz: I disagree. The most important thing right now is adding back ubuntu clang test, it was the clang test that worked, and got dropped accidentally.

I don't mind adding/removing the windows CPU part. Let me know if you prefer it removed and I will update the PR.

There was no working windows clang tests to start with, so no "keep the clang tests on windows". Adding that is adding a new CI feature, not a fix, and I don't understand prioritizing new feature over a fix.

wozeparrot: Oh, that's my bad, didn't read the original comment carefully enough. I thought that the the windows+clang tests worked before the pr.

We definitely do need the windows+clang tests tho, because that exact combination has broken before.

I'll merge adding ubuntu+clang back for now because we do need a working clang test, then we can prioritize switching it to windows+clang. Unless you want to just fix windows+clang in this pr.

chenyuxyz: Let's merge this one first.

We can have a follow up PR that replace the windows CPU and ubuntu clang with a windows + clang test. I can look into it later, unless you already know how to do it? I imagine it's not very difficult but need to be extra careful to prevent regression like this from happening again.

PR #1370, opened 2023-07-29
Stable diffusion WebGPU port

wpmed92:
Stable Diffusion on WebGPU

### Current state

- [x] `cond_stage_model`
- [x] `first_stage_model.decoder`
- [x] `first_stage_model.post_quant_conv`
- [x] loading and compiling the whole model in browser
- [x] diffusion UNet
- [x] support whole inference
- [x] support dynamic prompt (`ClipTokenizer` in the browser)

I added a `forward` call to the stable diffusion model which includes the currently supported layers of the model. The input is a latent tensor, the output is an image. The latent tensor has to be saved beforehand, and it will be loaded on the website using a numpy loader utility JS function.
I only load the weights which are used by the layers in `forward`, so that the model can be loaded at once in Chrome.
The full model exceeds the ~2GB `ArrayBuffer` allocation limit in Chrome. A solution for this is to split up the `net.safetensors` into multiple parts, and loading them one by one. That requires modification of tensor offset calculations. I'll push this once I support all layers.

Prompt: A cup of coffee, photorealistic
Screenshot 2023-07-29 at 10 14 30

Prompt: Super mario in Picasso style
Screenshot 2023-07-29 at 10 18 14

### Limitations

These are the WebGPU constraints that are hacked in this PR but need general solution, as discussed on Discord:
- global size: have a limit of [65535, 65535, 65535] ([maxComputeWorkgroupsPerDimension=65535](https://www.w3.org/TR/webgpu/#dom-supported-limits-maxcomputeworkgroupsperdimension))
- maximum number buffers per kernel: WebGPU supports a max of [8 storage buffers per compute stage](https://www.w3.org/TR/webgpu/#dom-supported-limits-maxstoragebufferspershaderstage)
- workgroup size: [[256](https://www.w3.org/TR/webgpu/#dom-supported-limits-maxcomputeworkgroupsizex), [256](https://www.w3.org/TR/webgpu/#dom-supported-limits-maxcomputeworkgroupsizey), [64](https://www.w3.org/TR/webgpu/#dom-supported-limits-maxcomputeworkgroupsizez)] 

some limitations, such as `maxStorageBufferBindingSize` can be overriden by requesting a larger value in `adapter.requestDevice`


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py            25               0
tinygrad/state.py                          5               0
total                                     32               1
lines added in the tinygrad folder: 31
```

geohot:
Bounty locked, good luck!

wozeparrot:
Any update on this? or is this blocked on something?

wpmed92:
@wozeparrot 

The current status is that I can run the whole model, which is exported in parts (text model, diffusor UNet, decoder) to be invoked separately. I recreated the Python inference code in JS. I solved the 2 gig arraybuffer allocation limit in chrome, and also made it possible to export a model with multiple inputs. I confirmed that the text model and the decoder work fine, as you can see in the description, those images were created with the decoder. However, the UNet part is buggy, and I'm currently tracking it down where exactly the inference goes south inside UNet.

For example, a prompt for a cup of coffee generates this:

Screenshot 2023-08-24 at 15 09 28

wpmed92:
It's working. Now I have to polish everything up.
https://github.com/tinygrad/tinygrad/assets/8374618/8fe1e5a8-dad2-4dc9-ba0f-66785ec80cda

wpmed92:
Demo site is up, compiled from the current state of this PR: https://softwiredtech.github.io/stable-diffusion-webgpu/

wpmed92:
@geohot Test are passing with latest master

- Works in prod chrome
- Reuses existing sd code
- Hacks: 2 one-liner hacks (both limited to webgpu to not cause harm elsewhere) that are needed to compile SD on WebGPU, one in lazy (elementwise merging, silimar to your hack), one in linearizer (global limit)
- fp16: no fp16 support in `wgpu` yet (it's shader compiler `naga` can't compile things like `enable shader_f16` and `f16`), although Chrome supports it. Is it ok if I serve the safetensor in fp16, and convert it to fp32 on client side?

geohot:
The value in fp16 is the weight download time, I don't care what the models runs in.

tinygrad/nn/state.py
```diff
@@ -9,11 +9,14 @@
 safe_dtypes = {"F16": dtypes.float16, "F32": dtypes.float32, "U8": dtypes.uint8, "I8": dtypes.int8, "I32": dtypes.int32, "I64": dtypes.int64}
 inverse_safe_dtypes = {v:k for k,v in safe_dtypes.items()}
 
-def safe_load(fn:Union[Tensor,str]) -> Dict[str, Tensor]:
+def safe_load_metadata(fn:Union[Tensor,str]):
```
geohot, line 12: Output type?

tinygrad/lazy.py
```diff
@@ -226,7 +226,9 @@ def e(self:LazyBuffer, op:Union[UnaryOps, BinaryOps, TernaryOps], *srcs:LazyBuff
 
     if MERGE_ELEMENTWISE_OPS:
       # remove the buffers from any (childless) BinaryOps that feed into this
-      srcs = tuple([x.op if x.optype == BinaryOps and not x.children and not x.realized else x for x in srcs])  # type: ignore
+      _srcs = tuple([x.op if x.optype == BinaryOps and not x.children and not x.realized else x for x in srcs])  # type: ignore
+      # TODO: needs general merge limiting
+      if out_device != "WEBGPU" or len([x.base for _src in _srcs for x in _src.buffers if not x.is_unrealized_const()]) < 7: srcs = _srcs # type: ignore
```
geohot, line 231: len of dedup base, but I think this is fine for now

tinygrad/codegen/linearizer.py
```diff
@@ -152,7 +152,8 @@ def linearize(self):
 
     # limit dims if we need to
     # TODO: broken, and doesn't really belong here
-    #if self.opts.global_max and self.opts.local_max: self.limit_dims_to_max(self.opts.global_max, self.opts.local_max)
+    # needs to be enabled for WEBGPU stable diffusion
+    if self.opts.device == "WEBGPU" and self.opts.global_max and self.opts.local_max: self.limit_dims_to_max(self.opts.global_max, self.opts.local_max)
```
geohot, line 156: Can we reenable this everywhere? (separate PR)

geohot:
Can we get the weights from elsewhere on the internet?

tinygrad/codegen/linearizer.py
wpmed92: I put up a PR for this, tests pass.
It was commented out in the Token removal [PR](https://github.com/tinygrad/tinygrad/pull/1723/files)

tinygrad/nn/state.py
wpmed92: `Tuple[Tensor, int, Any]`, just fixed

wpmed92:
Do you mean to not serve from my digital ocean spaces? I think we could upload the fp16 safetensor to [huggingface](https://huggingface.co/docs/huggingface_hub/guides/upload)

geohot:
Can `loading...` be a progress bar? I have both chrome and safari just stuck there

wpmed92:
If you tried the current live demo yeah that takes forever. I'll just take that down and put back live once we have progressbar + fp16 weights. I'll also add the check for wegbpu support.

wpmed92:
Also I think we should cache model client side using indexeddb so on a next visit you don't redownload

geohot:
How's this coming along? fp16 weights and progress bar? I merged back in the size limiter

wpmed92:
@geohot I have the model dl progress bar, and I moved the f32 safetensors to [huggingface](https://huggingface.co/wpmed/tinygrad-sd/tree/main) in 4 parts, and the download is running in parallel. I tried f16, browser-side f16->f32 conversion is very slow... Can you have a look what dl time you get with the parallel f32 Huggingface dl? Here's the updated live [demo](https://softwiredtech.github.io/stable-diffusion-webgpu/)

geohot:
Hmm, still pretty slow, a couple minutes. But it is a lot nicer with the progress bar, and it did work!

Is caching possible? I refreshed and had to redownload.

How are you doing the conversion that it's slow? Would bf16 (just a shift) be faster? How are the other ML WebGPU demos dealing with this?

Also, I'm curious why step times are 2s. Faster would be great if it's something super obvious, but fine if not

wpmed92:
- Glad it worked! I could split up the safetensor into more parts to get more speed.
- Caching is possible, I'll add it
- I tried different gists that do half to f32 conversion. I perform the conversion inside `getTensorBuffer` (this is [one](https://stackoverflow.com/questions/5678432/decompressing-half-precision-floats-in-javascript) of which I tried).
For other ML demos I found [this one](https://github.com/mlc-ai/web-stable-diffusion) which is tvm based and seem to do a lot of the heavy lifting in `wasm`, although nice, the (sd demo) codebase is very large compared to ours, which is just a `net.js` and an `index.html`.
- I'll look into speeding up the steps

From your findings which ones would you say are must-haves for this bounty to be claimed?

geohot:
The bounty is "webgpu stable diffusion webpage (reuse tiny sd code, float16 weights for small dl, works in prod chrome)"

4.2 GB is a lot to ask people to download and cache. Caching is less important even than float16, which is mentioned in the bounty.

geohot:
Oh, are you sure f16 doesn't work? https://www.w3.org/TR/WGSL/#f16

You need `enable f16;`

geohot:
So I would load it in as uint32 and convert it with a compute kernel that outputs two float32 from the uint32

You should be able to rewrite that stack overflow to run on the GPU.

geohot:
Ugh, so tinygrad could do it if it actually did math in the dtype...

wpmed92:
Yes,  `enable f16;` is there in Chrome, you have to `requestDevice` with [`shader-f16`] to get rid of the error you got, but then in prod Chrome you'll get another error: `Failed to execute 'requestDevice' on 'GPUAdapter': Unsupported feature: shader-f16`. But in Canary, I could request the device with no problem, and `shader-f16` is listed as a feature for the device:

indirect-first-instance
texture-compression-etc2
texture-compression-astc
depth32float-stencil8
depth-clip-control
**shader-f16**
float32-filterable
texture-compression-bc
rg11b10ufloat-renderable
bgra8unorm-storage

wpmed92:
This works fine in Canary
````const f16tof32 = 
enable f16;
@group(0) @binding(0) var<storage,read_write> data0: array<f16>;
@group(0) @binding(1) var<storage,read_write> data1: array<f32>;
@compute @workgroup_size(256) fn main(@builtin(global_invocation_id) gid: vec3<u32>) {
  let gidx = gid.x;

  if (gidx >= arrayLength(&data0)) {
    return;
  }

  data1[gidx] = f32(data0[gidx]);
}

wpmed92:
And something that works in prod chrome based on your tinygrad code is:

````
const f16tof32_fallback = `
fn u16_to_f16(x: u32) -> f32 {
    let sign = f32((x >> 15) & 0x1);
    let exponent = f32((x >> 10) & 0x1F);
    let fraction = f32(x & 0x3FF);

    let sign_multiplier = select(1.0, -1.0, sign == 1.0);
    if (exponent == 0.0) {
        return sign_multiplier * 6.103515625e-5 * (fraction / 1024.0);
    } else {
        return sign_multiplier * exp2(exponent - 15.0) * (1.0 + fraction / 1024.0);
    }
}

@group(0) @binding(0) var<storage,read_write> data0: array<u32>;
@group(0) @binding(1) var<storage,read_write> data1: array<f32>;
@compute @workgroup_size(256) fn main(@builtin(global_invocation_id) gid: vec3<u32>) {
    let gidx = gid.x;
    let outgidx = gidx*2;

    if (gidx >= arrayLength(&data0)) {
        return;
    }

    let oo = data0[gidx];
    let oo1 = (oo >> 16);
    let oo2 = oo & 0xFFFFu;

    let f1 = u16_to_f16(oo1);
    let f2 = u16_to_f16(oo2);
    
    data1[outgidx] = f1;
    data1[outgidx + 1] = f2;
}`;

wpmed92:
@geohot fp16 is ready, deployed [https://softwiredtech.github.io/stable-diffusion-webgpu/](https://softwiredtech.github.io/stable-diffusion-webgpu/)

I can decompress the whole thing in about 5s on my M1 pro, I'm sure it can improved, but it's a huge speed up compared to the cpu based method which took 3 minutes+ if it didn't freeze. So model compilation goes from 11 sec to about 16 with the decompression, but DL time is cut in half.

geohot:
Sweet, f16 looks good and compilation time is only 7s on my MBP! Last thing, caching?

wpmed92:
@geohot Caching is ready. Using indexedDB.

PR #650, opened 2023-03-06
Refactor contraction and add integration test cases for push permute

nanamiwang:

PR #1373, opened 2023-07-30
simplify get_contraction

chenyuxyz:
can remove the `%` check and fold both `prod`s into one. I checked the initial refactor #650 that introduced these, and I believe it could have been this version.

Also added some unit tests to document the current 1s grouping behavior. Should be able to group them into single axis but when I tried, at some point I passed all tests except the openpilot one and I felt it's not as trivial.


tinyb0t:
Changes made in `tinygrad/`:
```
```

PR #1374, opened 2023-07-30
fix to_shape_strides

S-Lykles:
to_shape_strides wasn't merging two dimensions if there was a dimension of size 1 in between them. 
For example: `ShapeTracker((10, 1, 10))` now corectly produces shape strides `((100, 1),)` instead of `((10, 10), (10, 1))`.
Also simplified the conditional in on line 17 and added the tests that caught this, expanding on #1363

### DIFF
--- a/test/unit/test_shapetracker.py
+++ b/test/unit/test_shapetracker.py
@@ -4,6 +4,7 @@ import numpy as np
 from tinygrad.helpers import prod, DEBUG
 from tinygrad.shape.shapetracker import ShapeTracker, View, get_contraction
 from tinygrad.shape.symbolic import Variable
+from itertools import product
 
 def shapetracker_getitem(st, val):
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/shapetracker.py             3               1
```

geohot:
So you have to be careful to not merge across a reduce boundary, does this handle that?

S-Lykles:
shape_strides is only used in view.expr_node to figure out how a single idx should index into the underlying buffer (or other view). Without the "merging" of conescutive dimensions in shape_strides you end up with a lot of `idx[0-99]//10*10 + idx[0-99]%10` which is equivalent to just `idx[0-99]`, I'm looking at if this optimization can be made in symbolic so we can get rid of shape_strides all together.

PR #1375, opened 2023-07-30
[windows+clang fails now, don't merge] replace ubuntu clang CI with windows clang CI

chenyuxyz:
This is a follow-up to #1368. This PR replaced the ubuntu+clang setup with a updated windows+clang setup, and removed the temporary windows+CPU tests. Also removed the flag `ENABLE_METHOD_CACHE`, which is the default now, and there was a typo so it was a noop in clang anyway.

The problem is that windows+clang has 26 failed tests on master, so we cannot merge this yet. ubuntu+clang is green.

few options
(1) skip failed in CI windows tests and merge this, which is a pure test coverage drop
(2) wait until someone review/fix/skip the failed tests, then merge this
(3) update this PR to include both ubuntu+clang and windows+clang, and skip the failed windows tests. Improves coverage. Drawback is more CI resource and slow
(3+) do (3) and we pick a meaningful subset to test for clang to make clang CI fast again

I personally like (3), @wozeparrot what do you think?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,27 +76,6 @@ jobs:
     - name: Test EfficientNet
       run: curl https://media.istockphoto.com/photos/hen-picture-id831791190 | ./recognize | grep hen
 
-  testcpuwindows:
-    name: CPU on Windows
-    runs-on: windows-latest
-    timeout-minutes: 20
-    steps:
-    - name: Checkout Code
-      uses: actions/checkout@v3
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v4
-      with:
-        python-version: 3.8
-    - name: Cache pip
-      uses: actions/cache@v3
-      with:
-        path: '~\AppData\Local\pip\cache'
-        key: testing
-    - name: Install Dependencies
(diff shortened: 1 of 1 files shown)


wozeparrot:
I would prefer (2), let me see what tests are failing first.

(3+) is also doable, if the tests that are failing on windows are pretty linux specific. This way we can just have something testing those parts.

wozeparrot:
Hmm, seems like most of the dtype tests are failing. I think those are the main ones, idk exactly what could cause exit status 1120. Maybe try a run without `-n=auto`?

chenyuxyz:
kicked a run without `-n=auto`. Yeah the type ones feel important and that's why I would prefer having both ubuntu and windows, with a smaller curated test sets for speed

dosier:
If not merge ready u can mark it as a draft PR btw :)

wozeparrot:
Can you pull in the `int64` is `long long` change from #1376 and see if it fixes some of the dtype tests?

chenyuxyz:
@wozeparrot can you open an issue to centralize the discussion about windows+clang? And I'll revert this PR to the first commit only so other people can pull this to test CI as well. Messing it here directly makes it hard for people to follow what's going on and what has been tried.

geohot:
Windows is not a supported platform.

Issue #1258, opened 2023-07-17
Don't write to disk in the clang backend


wozeparrot:
Currently a race condition is possible: https://canary.discord.com/channels/1068976834382925865/1068976834928193609/1128683739946418277

![race condition pytest](https://github.com/tinygrad/tinygrad/assets/25372613/83713fe2-d30d-408a-8037-887c512d8242)

It's also a TODO item: https://github.com/tinygrad/tinygrad/blob/64d39188ad22574025eb6534727ed33d699b4348/tinygrad/runtime/ops_clang.py#L15

This is 100% possible to do on linux (see below), needs investigation for windows and mac.

![linux fix](https://github.com/tinygrad/tinygrad/assets/25372613/2dc7aa58-e19d-4fc4-b481-4dc68ce9b4f9)

mixmaester:
maybe possible with io.BytesIO() ?

ZenginU:
@wozeparrot,
I am new to this staff but I believe for macos we have mmap.mmap() in python and for windows we can create memory mapped file than get the handle with _get_osfhandle. Hope this is helpful

xuky76:
/dev/shm isn't a thing on mac

- [shm_open(2)](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/shm_open.2.html)
- [https://docs.python.org/3/library/multiprocessing.shared_memory.html](https://docs.python.org/3/library/multiprocessing.shared_memory.html)

WilliamTheHippo:
> maybe possible with io.BytesIO() ?

Sadly the `CDLL` constructor requires a filename for the shared object/library, and, from what I can tell, there's no option to specify a named stream or something; it needs a location on disk. Maybe it's some sort of guard against mangled/truncated streams?

Very frustrating, especially given the usual ease of piping around binary data between functions without needing to store it on disk.

xando:
Hey I am trying to understand the comments above. 


from @mixmaester
>maybe possible with io.BytesIO() ?


from @xuky76  
>/dev/shm isn't a thing on mac 

and [TODO](https://github.com/tinygrad/tinygrad/blob/7cac5ea16c46ed936ede2e6a172ddc6c6259f9b8/tinygrad/runtime/ops_clang.py#L15) itself. 
> `TODO: is there a way to not write this to disk?`



[...]

PR #1376, opened 2023-07-31
Workaround broken half handling in clang backend

bobsayshilol:
All of the half precision tests fail on a stock Fedora install due to the compiled kernels producing garbage. This was tracked down to the (necessary) `--rtlib=compiler-rt` flag and an existing issue in LLVM: https://github.com/llvm/llvm-project/issues/56204. As a workaround we can use `-march=native` to replace the broken calls with native instructions, and potentially get a speed boost too. See commit messages for more details.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_clang.py             16               7
```

wozeparrot:
Strange, but I can't seem to reproduce the issue on ubuntu 22.04 with clang-15 or clang-16.

tinygrad/runtime/ops_clang.py
```diff
@@ -9,15 +9,24 @@
...
 class ClangProgram:
   def __init__(self, name:str, prg:str):
-    prg = '#include <math.h>\n#define max(x,y) ((x>y)?x:y)\n#define int64 long\n#define half __fp16\n#define uchar unsigned char\n#define bool uchar\n' + prg
-    # TODO: is there a way to not write this to disk?
-    fn = f"{tempfile.gettempdir()}/clang_{hashlib.md5(prg.encode('utf-8')).hexdigest()}.{args['ext']}"
-    if not os.path.exists(fn):
-      subprocess.check_output(args=('clang -shared -O2 -Wall -Werror -x c '+args['cflags']+' - -o '+fn+'.tmp').split(), input=prg.encode('utf-8'))
-      os.rename(fn+'.tmp', fn)
-    self.lib = ctypes.CDLL(fn)
+    prg = '#include <math.h>\n#define max(x,y) ((x>y)?x:y)\n#define int64 long long\n#define half __fp16\n#define uchar unsigned char\n#define bool uchar\n' + prg
```
wozeparrot, line 28: I would say revert this change for now until we can get windows+clang CI fixed, progress in #1375.

tinygrad/runtime/ops_clang.py
```diff
@@ -9,15 +9,24 @@
...
+  # TODO: is there a way to not write this to disk?
+  dst = f"{tempfile.gettempdir()}/clang_{hashlib.md5(prg.encode('utf-8')).hexdigest()}.{args['ext']}"
+  if not os.path.exists(dst):
+    subprocess.check_output(args=('clang -shared -O2 -Wall -Werror -march=native -x c '+args['cflags']+' - -o '+dst+'.tmp').split(), input=prg.encode('utf-8'))
+    os.rename(dst+'.tmp', dst)
+  return ctypes.CDLL(dst)
+
+if platform.system() == 'Linux':
+  # Reproducer taken from https://github.com/llvm/llvm-project/issues/56204
+  lib = _compile_to_lib('float __attribute__((noinline)) n(float v) { return -v; } int check() { return (__fp16)n(-1.f); }')
```
wozeparrot, line 22: this is currently race conditioning in CI because of `-n=auto`, see #1258

tinygrad/runtime/ops_clang.py
bobsayshilol: Ah good shout. I'll drop the commit now that it's been picked over so a merge doesn't accidentally revert the pick.

tinygrad/runtime/ops_clang.py
bobsayshilol: That makes sense. It felt like a race but I couldn't spot where it was coming from. Would the best solution be to hold off on this PR until that's fixed then?

bobsayshilol:
> Strange, but I can't seem to reproduce the issue on ubuntu 22.04 with clang-15 or clang-16.

For completeness, here's the script I ran and results I get in a docker container:
```sh
$ cat main.c
float __attribute__((noinline)) n(float value) { return -value; }
int main(int argc, char* argv[]) { return (__fp16)n(-1.f); }

$ cat run.sh
rm rt-*
for v in 14 15 16; do
 apt install -y clang-$v
 clang-$v main.c -o rt-gcc-$v --rtlib=libgcc
 clang-$v main.c -o rt-llvm-$v --rtlib=compiler-rt
 apt remove -y clang-$v
 apt autoremove -y
done
for f in rt-*; do
 echo $f
 ./$f; echo $?
done

$ run.sh
...
# rt-gcc-14 fails to link since it needs compiler-rt 
rt-gcc-15
1
rt-gcc-16
1
rt-llvm-14
1
rt-llvm-15
0
rt-llvm-16
1
```

tinygrad/runtime/ops_clang.py
wozeparrot: Do you want to take a shot at fixing it? that would be best outcome. We can mark this as draft for now.

bobsayshilol: Hmm, I'm not sure how we'd get around writing to disk on all platforms since `CDLL` requires a filepath - not a file object - to load the lib. I thought about it a bit and I think a filesystem lock would be enough for this PR, which can always be replaced with an in-memory file later down the road if a x-plat way to do that is found. It looks like `portalocker` is what we want, which also appears to be used by `pytorch`.

geohot:
Weird import. There's much simpler ways to fix this.

wozeparrot:
just the file renaming race condition.

PR #1371, opened 2023-07-29
Implementation of SoftVC VITS SVC model

JakobSpahn:
**Bounty:** voice changer model (so-vits-svc, rvc). make me sound like drake
**Original Code:** https://github.com/svc-develop-team/so-vits-svc

Had to diverge from the standard VITS implementation from #1188 a bit to accommodate for the additional F0 frequency feature and the higher sampling rate.

TODO:
- [x] Add HnNSF module to Generator and finish up forward pass
- [x] Finish Synthesizer and load weights
- [x] Implement ContentVec from https://github.com/auspicious3000/contentvec (fairseq) & load weights + config 
- [x] Run SVC and generate proper sounding output
- [x] Cleanup & Refactor
- [x] Add proper upsample pass that doesn't rely on torch
- [x] Don't rely on torchaudio anymore 
- [x] Check if Drake weights can be fetched from somewhere else than mega 

```
# INSTALLATION: dependencies are for preprocessing and loading/saving audio.
pip3 install soundfile librosa praat-parselmouth
```
```
# DEMO USAGE (uses audio sample from LJ-Speech):
python3 examples/so_vits_svc.py --model saul_goodman
```

PR #1378, opened 2023-07-31
rng hlops: add normal and kaiming_normal

JakobSpahn:
Could use this in #1371  to implement the contentvec from  https://github.com/auspicious3000/contentvec .

**Tensor::normal:** normal dist can be computed from standard dist by factoring in std and adding mean. See notes in: https://numpy.org/doc/stable/reference/random/generated/numpy.random.randn.html

**Tensor::kaiman_normal:** analog to Tensor::kaiman_uniform.

Includes tests.

Also any reason why Tensor::uniform default args are low=-1.0 and high=1.0 and torch::init::uniform are a=0 and b=1? (https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.uniform_ )


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         9               0
```

PR #1381, opened 2023-07-31
trim space

chenyuxyz:
fix the newly added EditorConfig Checker on CI


geohot:
I just reverted that PR. Not dealing with this

PR #1384, opened 2023-07-31
Raise NotImplementedError instead of RuntimeError:

pin3da:
This represents the error better, and also matches the error in the `forward` method.

Also in this commit: Clean up unneeded imports in test_tensor.py.


wozeparrot:
#1315

PR #1385, opened 2023-07-31
Timing and Context can now be used as decorators

AlexTelon:
The various updates to documentation and code outside `helper.py` is just some suggestions. I can remove all of them if wanted.

(I initially pitched this in #1331.)

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -9,7 +9,7 @@ We need some imports to get started:
 
 ```python
 import numpy as np
-import time
+from tinygrad.helpers import Timing
 ```
 
 ## Tensors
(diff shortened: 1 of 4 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        3               2
```

tinygrad/helpers.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import contextlib
 import os, functools, platform, time, re
```
wozeparrot, line 3: Can just stack these imports together.

docs/quickstart.md
```diff
@@ -9,7 +9,7 @@ We need some imports to get started:
 
 ```python
 import numpy as np
-import time
+from tinygrad.helpers import Timing
```
AlexTelon, line 12: I totally buy that for a quickstart it might be better to just use time to decrease the amount of new stuff the reader has to deal with.

But it is a nice utility function to know about so hence this suggestion. But I'm not sure myself.

wozeparrot: Yea, I like this better, it showcases the feature.

test/test_helpers.py
```diff
@@ -90,5 +90,14 @@ def test_nested_context(self):
       self.assertEqual(VARIABLE.value, 1)
     self.assertEqual(VARIABLE.value, 0)
 
+  def test_decorator(self):
+    @Context(VARIABLE=1, DEBUG=4)
+    def test():
+      self.assertEqual(VARIABLE.value, 1)
```
wozeparrot, line 96: Should probably also check that `DEBUG` is set correctly here.

PR #1386, opened 2023-07-31
reshape to register variable value

chenyuxyz:
Part of #1353 

reshape from int shape to shape with a variable writes the variable value to `Variable.val`. We don't need the value to generate the kernel code, but we need it to run the generated program.

### DIFF
--- a/tinygrad/shape/shapetracker.py
+++ b/tinygrad/shape/shapetracker.py
@@ -235,12 +235,21 @@ class ShapeTracker:
     self.views[-1] = View(new_shape, self.views[-1].strides, self.views[-1].offset, mask)
     return self
 
-  def reshape(self, new_shape: Tuple[int, ...]):
+  def reshape(self, new_shape: Tuple[Union[Node,int], ...]):
+    # reshape into symbolic shape, update the variable value
+    if all(isinstance(s, int) for s in self.shape) and len(new_vars:=list(s for s in new_shape if isinstance(s, Variable))) > 0:
+      assert len(new_vars) == 1  # only allow adding one variable
+      new_var, new_val = new_vars[0], prod(self.shape) // prod(s for s in new_shape if isinstance(s, int))
+      if new_var.val is None:
+        assert new_var.min <= new_val <= new_var.max
+        new_var.val = new_val
+      else: assert new_var.val == new_val
+
     if self.views[-1].shape == new_shape: return self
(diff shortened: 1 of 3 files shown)


tinygrad/shape/symbolic.py
```diff
@@ -134,6 +134,7 @@ def __new__(cls, expr:Optional[str], nmin:int, nmax:int):
 
   def __init__(self, expr:Optional[str], nmin:int, nmax:int):
     self.expr, self.min, self.max = expr, nmin, nmax
+    self.val: Optional[int] = None
```
geohot, line 137: Should this be tracked in the Variable or in external context?

Example:
```
vi = Variable("i", 1, 10)
a = Tensor.rand(3, 4).reshape(vi, 2)
b = Tensor.rand(2, 4).reshape(vi, 2)
```

What happens there?

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/shapetracker.py            11               2
tinygrad/shape/symbolic.py                 1               0
total                                     12               2
```

tinygrad/shape/shapetracker.py
```diff
@@ -235,12 +235,21 @@ def expand(self, new_shape: Tuple[int, ...]) -> ShapeTracker:
...
     return self
 
-  def reshape(self, new_shape: Tuple[int, ...]):
+  def reshape(self, new_shape: Tuple[Union[Node,int], ...]):
+    # reshape into symbolic shape, update the variable value
+    if all(isinstance(s, int) for s in self.shape) and len(new_vars:=list(s for s in new_shape if isinstance(s, Variable))) > 0:
+      assert len(new_vars) == 1  # only allow adding one variable
+      new_var, new_val = new_vars[0], prod(self.shape) // prod(s for s in new_shape if isinstance(s, int))
+      if new_var.val is None:
+        assert new_var.min <= new_val <= new_var.max
```
geohot, line 244: Can you add text to these assertions to explain the failure?

tinygrad/shape/symbolic.py
chenyuxyz: the old `symbols` was trying to track these values in shapetracker level, which is also possible. But we still have the problem that two shapetracker can have different symbol:value dicts. Resolving at variable level feels the cleanest to me. Or are you suggesting a global variable like context?

tinygrad/shape/shapetracker.py
chenyuxyz: done

PR #1391, opened 2023-08-01
LazyBuffer.get_variable_buffers()

chenyuxyz:
Part of #1353

`LazyBuffer.get_variable_buffers()` returns a map of {symbol: unrealized size 1 buffer of symbol value}. We will need the keys of this dict in codegen, and the values in exec.

Extend the original `Node.vars` to possibly return the `vars` from in `OpNode.b`. The only place that used `vars` was in `real_strides`. We need to keep the version that only returns the vars of `OpNode.a` because `real_strides` construct a set of variables and relies on the assumption that `OpNode.b` will always be `int`.

### DIFF
--- a/tinygrad/shape/symbolic.py
+++ b/tinygrad/shape/symbolic.py
@@ -2,14 +2,14 @@ from __future__ import annotations
 from abc import abstractmethod
 import functools
 from math import gcd
-from tinygrad.helpers import partition
+from tinygrad.helpers import partition, prod
 from typing import List, Dict, Callable, Tuple, Type, Union, Optional, Any
 
 # NOTE: Python has different behavior for negative mod and floor div than c
 # symbolic matches the Python behavior, but the code output is agnostic, and will never have negative numbers in div or mod
 
 def is_sym_int(x: Any) -> bool: return isinstance(x, int) or isinstance(x, Node)
-def sym_vars(x: Union[Node, int], left_only=False) -> List[Variable]: return [] if isinstance(x, int) else x.vars(left_only)
+def sym_vars(x: Union[Node, int]) -> List[Variable]: return [] if isinstance(x, int) else x.vars()
 
 class Node:
   b: Union[Node, int]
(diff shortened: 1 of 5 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                           2               0
tinygrad/shape/symbolic.py                 5               4
total                                      8               5
```

tinygrad/shape/symbolic.py
```diff
@@ -189,7 +190,7 @@ def get_bounds(self) -> Tuple[int, int]:
 
 class RedNode(Node):
   def __init__(self, nodes:List[Node]): self.nodes = nodes
-  def vars(self): return functools.reduce(lambda l,x: l+x.vars(), self.nodes, [])
+  def vars(self, left_only=False): return functools.reduce(lambda l,x: l+x.vars(left_only), self.nodes, [])
 
 class SumNode(RedNode):
```
geohot, line 195: `ProductNode` instead of `left_only`

chenyuxyz: done

tinygrad/shape/symbolic.py
```diff
@@ -149,9 +152,10 @@ def create_node(ret:Node):
 
 class OpNode(Node):
   def __init__(self, a:Node, b:Union[Node, int]):
+    if isinstance(self, MulNode): assert isinstance(b, int), f"MulNode.b must be int, getting {type(b)}"
     self.a, self.b = a, b
     self.min, self.max = self.get_bounds()
-  def vars(self): return self.a.vars()
+  def vars(self): return self.a.vars() + (self.b.vars() if isinstance(self.b, Node) else [])
```
geohot, line 158: Do you need this?

tinygrad/shape/symbolic.py
```diff
@@ -149,6 +152,7 @@ def create_node(ret:Node):
 
 class OpNode(Node):
   def __init__(self, a:Node, b:Union[Node, int]):
+    if isinstance(self, MulNode): assert isinstance(b, int), f"MulNode.b must be int, getting {type(b)}"
```
geohot, line 155: Why do you need this + the assert below?

chenyuxyz:
so I have reverted all `MulNode` related changes to keep this PR minimal

`ProdNode` as it stands is incomplete because `real_strides` [expects](https://github.com/tinygrad/tinygrad/blob/ab9e4a2e936aaca450dc48cb6e85d39062c4d4e4/tinygrad/shape/shapetracker.py#L169) `MulNode`, so we cannot just promote symbolic * symbolic into a `ProdNode`. Need more time to have a clean fix so will do in a separate PR.

Issue #1351, opened 2023-07-27
Match gather implementation to torch


dosier:
I tried replacing the one used in the vits example with the newly merged one (#1329 ) but it's giving errors, maybe for sake of ease of porting models we should make sure the input/output results match that of torch.

geohotstan:
Do you have code I can run to reproduce the error?

dosier:
> Do you have code I can run to reproduce the error?
Yes, sorry. 
```py
from tinygrad.tensor import Tensor
import torch, numpy

tiny_idx = Tensor.ones(361, 1)
tiny_cum_widths = Tensor.randn(361, 11)
tiny_res = tiny_cum_widths.gather(tiny_idx, dim=-1).numpy()

torch_idx = torch.ones(361, 1, dtype=torch.long)
torch_cum_widths = torch.randn(361, 11)
torch_res = torch_cum_widths.gather(-1, torch_idx).numpy()

print(f"tiny_shape: {tiny_res.shape}")
print(f"torch_shape: {torch_res.shape}")
assert numpy.allclose(torch_res, tiny_res)
```

Which prints:
```
tiny_shape: (11, 361, 1, 361)
torch_shape: (361, 1)
```

geohotstan:
<img width="822" alt="image" src="https://github.com/tinygrad/tinygrad/assets/135171913/1d93deb8-92dd-43fc-8844-301ce8fbcac6">

Essentially what `Tensor.gather()` does right now is
```
out[i][j][k] = input[index[i][j][k]][:][:]  # if dim == 0
out[i][j][k] = input[:][index[i][j][k]][:]  # if dim == 1
out[i][j][k] = input[:][:][index[i][j][k]]  # if dim == 2
```

the `input[index[i][j][k]][j][k]` in `torch.gather` requires the ability to do combined indexing https://github.com/tinygrad/tinygrad/issues/967 (of which I have yet to figure out haha! :D)

There are other ways to hack `torch.gather()` aka GatherElements https://github.com/geohotstan/tinygrad/blob/passing_onnx/extra/onnx_ops.py#L439, but that's probably not ideal either.

I'll try at combined/fancy indexing some more, and if I figure it out, I'll move the current `Tensor.gather()` into `Tensor.__getitem__()` and express the new `Tensor.gather()` using `Tensor.__getitem__()`?
kinda like the torch example.

this way the apis can match

geohotstan:
[...]

Issue #1383, opened 2023-07-31
numpy default cast to float64 in some cases


geohotstan:
There are several instances where numpy default casts to float64 instead of float32, mainly caused when int32 or int64 are involved.

e.g.
```
ON CPU
In [9]: cmp = Tensor([1, 0], dtype=dtypes.int32) == Tensor([1.1, 1.0], dtype=dtypes.float32)
In [10]: cmp.realize()
KeyError: 'float64'

In [17]: Tensor([1], dtype=dtypes.int32).exp().realize()
KeyError: 'float64'

In [17]: Tensor([1], dtype=dtypes.int32).log().realize()
KeyError: 'float64'

In [38]: Tensor([1], dtype=dtypes.int32).sin().realize()
KeyError: 'float64'

In [4]: Tensor([1], dtype=dtypes.int32).sqrt().realize()
KeyError: 'float64'

There might be more
```

There are no internal ways to change numpy's default casting behavior https://github.com/numpy/numpy/issues/6860

So some possible solutions of covering this issue are:

- change to `UnaryOps.SIN: lambda x: np.sin(x, dtype=np.float32)` (this doesn't seem right cuz what about when x is float16)
- overload/wrap the numpy functions to cast to float32 if float64?
- write a similar promote_types() function that wraps every x or y to cast to float32 if it's int32 or int64 (so ret isn't float64)?
- other

Not sure which solution is proper

dc-dc-dc:
I added in a `match_types` helper function to help with this behavior
Changing `CMPEQ`  to `np.equal(*match_types(x,y))` should resolve the first case

Not sure what the ideal solution would be for the unary ops tho

geohotstan:
update:

`In [38]: Tensor([1], dtype=dtypes.int32).sin().realize()` examples like this on other backends, like METAL, casts to the original dtype, so after realize() this Tensor would still be `dtype=int` with value `[0]`

Not sure if this is a bug or not since 
```
In [5]: torch.tensor([1], dtype=torch.int32).sin().dtype
Out[5]: torch.float32
```

I've tried using [np.min_scalar_type()](https://numpy.org/doc/stable/reference/generated/numpy.min_scalar_type.html) to cast the result of functions like `np.sin()` to smallest size but doing it this way catches the AssertionError:
```
--> 162 assert self.realized.dtype == self.dtype or self.dtype.__class__ is ImageDType, f"dtype mismatch on realize got {self.realized.dtype} expected {self.dtype}"
[...]

Issue #858, opened 2023-05-29
Use Tensor as Index (e.g. `gather`)


vwxyzjn:
Hello, this is such a cool project. I was playing with using tinygrad to create some reinforcement learnings scripts, but I ran into a minor issue with using Tensor as Index.

A common thing to do in Deep Q Learning is to use a tensor filled with actions to fetch its corresponding Q values. For example, in the snippet below, there are two batches of q values and actions, and we want to fetch the q values `0` and `6`. We can achieve this feature in torch either using `gather` or just array indexing.

```python
import torch
from tinygrad.tensor import Tensor

q_values = torch.tensor([[1, 2, 3], [4, 5, 6]])
actions = torch.tensor([[0], [2]]) 
print(f"we have q_values: \n{q_values}\nand actions: \n{actions}")
print("q_values.shape", q_values.shape)
print("actions.shape", actions.shape)
print("q_values corresponding to actions", q_values.gather(1, actions).flatten())

actions = torch.tensor([0, 2]) 
print("q_values.shape", q_values.shape)
print("actions.shape", actions.shape)
print("q_values corresponding to actions", q_values[torch.arange(2), actions])
```
```
we have q_values: 
tensor([[1, 2, 3],
        [4, 5, 6]])
and actions: 
tensor([[0],
        [2]])
q_values.shape torch.Size([2, 3])
actions.shape torch.Size([2, 1])
q_values corresponding to actions tensor([1, 6])
q_values.shape torch.Size([2, 3])
actions.shape torch.Size([2])
q_values corresponding to actions tensor([1, 6])
```

After some digging, however, I realized tinygrad does not seem to have similar feature such as `gather`. Tinygrad also does not support Tensor to index arrays, as shown below. As a result, I wasn't sure how to proceed and was wondering if you have any suggestions. Thanks!

```python
from tinygrad.tensor import Tensor
Tensor([[1, 2, 3], [4, 5, 6]])[Tensor([0])]
```
```
Traceback (most recent call last):
  File "/home/costa/Documents/go/src/github.com/vwxyzjn/tinygrad/q_values.py", line 18, in <module>
    Tensor([[1, 2, 3], [4, 5, 6]])[Tensor([0])]
  File "/home/costa/Documents/go/src/github.com/vwxyzjn/tinygrad/tinygrad/tensor.py", line 252, in __getitem__
[...]

PR #1399, opened 2023-08-02
Fancy indexing is fancy wow and gather thing

geohotstan:
Now you can slice a Tensor with Tensors! Like slicing butter with butter! fun!
this took way longer than it should've because my brain very potato sorry. Very difficult for me, really mashed my potato brain trying to figure this out.

Possible future TODOs:

- not sure if `tinygrad/tensor.py 325:328` and `tinygrad/tensor.py 364:366` is _idiomatic python_, or actually not sure if any of my code is proper python but maybe there are refactors to shave off some code.
- merge `_gather()` into `__getitem__()` and express `gather()` using `__getitem__()` . My intuition is that this needs to do some crazy shape permutes and then getitem with idx being the first dim and the rest of the dims being `Tensor.arange(self.shape[dim])`
- test the speed for this... Might be possible to hack with shape permutes actually..., especially for gather. 

but alas, code works for now.
Another thing that's weird is that pytorch doesn't allow tor[..., None, a, b] but my cursed implementation of this allows ten[..., None, x, z]. Not sure if result is right tho.

fixes https://github.com/tinygrad/tinygrad/issues/1351 https://github.com/tinygrad/tinygrad/issues/967


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                        44               5
lines added in the tinygrad folder: 39
```

geohot:
Hmm, this looks like a lot of code. Can it be simpler?

tinygrad/tensor.py
```diff
@@ -314,22 +316,59 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
...
     final_shape = []
+    sub = [0] * len(tensor_found)
     it_shape = iter(new_shape)
-    for i in orig_slices:
+    for n,i in enumerate(orig_slices):
       if isinstance(i, (int, slice)):
         dim_shape = next(it_shape)
         if isinstance(i, slice): final_shape.append(dim_shape)
+        else: # i is int
+          if tensor_found: # for tensor indexing
```
wozeparrot, line 326: I think you can stack these two into a single `elif tensor_found:`

tinygrad/tensor.py
```diff
@@ -314,22 +316,59 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
...
+      dim = [i[0] for i in tensor_found]
+      idx = [(i[1] < 0).where(i[1]+ret.shape[i[0]], i[1]) for i in tensor_found]
+      dim_cond = dim[0] != 0 and dim != list(range(dim[0], dim[-1]+1)) and len(dim) != 1
+      for n,d in enumerate(dim):
+        if n == 0:
+          new_ret = ret._gather(idx=idx[n], dim=d)
+        else:
+          new_idx = idx[n].reshape(*[1]*dim[0], *idx[n].shape, *[1]*(ret.ndim-dim[0]-n))
+          arange = Tensor.arange(ret.shape[d], dtype=dtypes.int32, requires_grad=False).reshape(*[1]*(idx[n].ndim+d-n), ret.shape[d], *[1]*(ret.ndim-d-1))
+          new_ret = ((new_idx == arange) * new_ret).sum(idx[n].ndim+d-n)
```
wozeparrot, line 344: You should be able to move the first iteration of the loop out of the loop at just start the loop at 1.

geohotstan:
I got a lil too excited yesterday at things actually working and just fired off the PR loool
should've refined it before submitting



> Hmm, this looks like a lot of code. Can it be simpler?

Yeah definitely
It's just a pain that the args for reshape is a little different for each but I'll see if I can find a generalized way to do it.
(there might be conceptual improvements too, gotta think some more)

tinygrad/tensor.py
```diff
@@ -314,22 +316,55 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
...
-    return sliced_tensor.reshape(tuple(final_shape))  # Reshape
+    ret = sliced_tensor.reshape(tuple(final_shape))  # Reshape
+    # Fancy/tensor indexing
+    if tensor_found:
+      for i,s in enumerate(sub): tensor_found[i] = (tensor_found[i][0]+s, tensor_found[i][1])
+      dim = [i[0] for i in tensor_found]
+      idx = [(i[1] < 0).where(i[1]+ret.shape[i[0]], i[1]) for i in tensor_found]
+      dim_cond = dim[0] != 0 and dim != list(range(dim[0], dim[-1]+1)) and len(dim) != 1
+      new_ret = ret._gather(idx=idx[0], dim=dim[0])
+      for n,d in zip(range(1, len(dim[1:])+1), dim[1:]):
```
wozeparrot, line 338: `enumerate` actually has a start parameter, so you can probably just do `enumerate(dim[1:], 1)`

geohotstan:
After much mental cartwheels, I'm proud to say I managed to decrease the line count by an ASTOUNDING... 1 line :D 

I made a conceptual change where the _gather()'s dim parameter is changed from representing the index of self before the expansion of the first cmp->mul->reduce, to the index of the new_ret (after each cmp->mul->reduce) that is to be sum-ed.
this allows both `Tensor.gather()` and `fancy indexing` to be expressed entirely by `Tensor._gather()` (which just does all the cmp -> mul -> reduce). The difference now is that the new dim parameter for `Tensor._gather()` have to be pre-computed to the correct sum_dim (was very tempted to name it dim_sum)


I am currently out of ideas for how to make this simpler.
I've explored two other ways of decreasing line count:

- write mother of all 1 liners that's probably long enough to poke at the stars (not good)
- instead of looping in `_gather()` we have a list of `arange` and a list of `idx` which is all reshaped using the expanded shape from the first cmp->mul->reduce. Then compare all of them, mul all of them with self, and reduce with `sum(dim1).sum(dim2).sum(dim3)...` (this isn't ideal since each cmp and mul is done with the expanded shape instead of the reduced shape after each sum, and this I tried this with `Tensor.gather()` but didn't work)

I might need to think of a totally new way of doing this if the current way is too complicated. 
can close PR if still too complicated (forward pass and backward pass is also quite a bit slower than pytorch)
or I can draw out how this current implementation works conceptually and someone else can give this a shot?

tinygrad/tensor.py
```diff
@@ -314,21 +316,50 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
...
-    new_self = self.reshape(*self.shape[:dim+1], *[1]*idx.ndim, *self.shape[dim+1:])
-    arange = Tensor.arange(self.shape[dim], dtype=dtypes.int32, requires_grad=False).reshape(*[1]*dim, self.shape[dim], *[1]*idx.ndim, *[1]*(self.ndim-dim-1))
-    new_idx = idx.reshape(*[1]*dim, 1, *idx.shape, *[1]*(self.ndim-dim-1))
-    return (new_self * (arange == new_idx)).sum(dim)
+    ret = sliced_tensor.reshape(tuple(final_shape))  # Reshape
+    # Fancy/tensor indexing
+    if tensor_found:
+      for i,s in enumerate(sub): tensor_found[i] = (tensor_found[i][0]+s, tensor_found[i][1])
+      dim = [i[0] for i in tensor_found]
+      idx = [(i[1] < 0).where(i[1]+ret.shape[i[0]], i[1]) for i in tensor_found]
```
wozeparrot, line 335: is it possible to get rid of the comparison and where here? can we do a multiplication and add?

wozeparrot:
I do expect it to be slower than pytorch, because pytorch can just directly index memory for stuff like this.

Do we want to default to int32 dtype for the aranges?

Overall, I think this is already pretty simple for a pretty complex task, I'm not sure if there is a way that you can make it simpler.

tinygrad/tensor.py
geohotstan: like this? `idx = [-i[1].sign().minimum(0) * ret.shape[i[0]] + i[1]  for i in tensor_found]` 🤣  🤣  🤣 
I never know if I'm doing something stupid

geohotstan:
> pytorch can just directly index memory for stuff like this.

buncha CHEATERS! 😡 

> Do we want to default to int32 dtype for the aranges?

Hmm I don't quite know all the implications of using int32 or default (float32) dtype, but if we use default, then we can revert the `promote_types()` change in `ops_cpu.py`? Since `np.promote_types` won't cast to float64 anymore for gather.

tinygrad/tensor.py
```diff
@@ -314,21 +316,50 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
       final_slice = reduce(operator.add, (((0, sh), (0, 1)) for sh in new_shape), ())
       sliced_tensor = reshaped_tensor.shrink(final_slice)
     final_shape = []
+    sub = [0] * len(tensor_found)
     it_shape = iter(new_shape)
-    for i in orig_slices:
+    for n,i in enumerate(orig_slices):
```
roelofvandijk, line 321: Counter-intuitive naming, maybe swap the two?

tinygrad/tensor.py
```diff
@@ -314,21 +316,50 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
...
-    arange = Tensor.arange(self.shape[dim], dtype=dtypes.int32, requires_grad=False).reshape(*[1]*dim, self.shape[dim], *[1]*idx.ndim, *[1]*(self.ndim-dim-1))
-    new_idx = idx.reshape(*[1]*dim, 1, *idx.shape, *[1]*(self.ndim-dim-1))
-    return (new_self * (arange == new_idx)).sum(dim)
+    ret = sliced_tensor.reshape(tuple(final_shape))  # Reshape
+    # Fancy/tensor indexing
+    if tensor_found:
+      for i,s in enumerate(sub): tensor_found[i] = (tensor_found[i][0]+s, tensor_found[i][1])
+      dim = [i[0] for i in tensor_found]
+      idx = [-i[1].sign().minimum(0) * ret.shape[i[0]] + i[1]  for i in tensor_found]
+      max_sh = max([i.ndim for i in idx])
```
roelofvandijk, line 336: No `[]` needed.

tinygrad/tensor.py
```diff
@@ -314,21 +316,50 @@ def num_zeros(step, dim_sz): return 0 if step == 1 or (y := dim_sz % step) == 0
...
+    for i,d in zip(idx[1:],dim[1:]):
+      new_idx = i.reshape(*[1]*dim[0], *i.shape, *[1]*(new_ret.ndim-dim[0]-i.ndim))
+      arange = Tensor.arange(new_ret.shape[d], requires_grad=False).reshape(*[1]*(d), new_ret.shape[d], *[1]*(new_ret.ndim-d-1))
+      new_ret = ((new_idx == arange) * new_ret).sum(d)
+    return new_ret
+
+  def gather(self: Tensor, idx: Tensor, dim: int):
+    assert idx.ndim == self.ndim, "self.ndim must equal idx.ndim"
+    if dim < 0: dim += self.ndim
+    self_idx = list(range(self.ndim))
```
roelofvandijk, line 359: If you call this `self_dim` you avoid the double `idx` and the enumerates can become `i,d in enumerate(self_dim..)`

wozeparrot:
actually I think that having int32 be default might be fine.

this good to merge?

geohotstan:
> actually I think that having int32 be default might be fine.
> 
> this good to merge?

one final problem
```
In [26]: 
(Tensor([-1], dtype=dtypes.int64, device="TORCH") / (Tensor([-1], dtype=dtypes.int64, device="TORCH"))).realize()
AssertionError: dtype mismatch on realize got dtypes.float expected dtypes.long

in lazy.py:
self.op.op == BinaryOps.DIV
```
is this expected?
reason for bringing up this error is cuz

```
idx = [i[1].sign().__neg__().maximum(Tensor([0])) * ret.shape[i[0]] + i[1] for i in tensor_found]
i[1] here is int64 from test_ops.py and i[1].sign() fails to realize because Assertion error above
also (pytorch only allows int indices)
```
should I force cast i[1] to float here?

geohotstan:
if it's okay to assert idx.dtype to be float, then this is good to go.

wozeparrot:
> > actually I think that having int32 be default might be fine.
> > this good to merge?
> 
> one final problem
> 
> ```
> In [26]: # int32 fails also
> (Tensor([1], dtype=dtypes.int64, device="TORCH") / (Tensor([1], dtype=dtypes.int64, device="TORCH"))).realize()
> AssertionError: dtype mismatch on realize got dtypes.float expected dtypes.long
> ```
> 
> is this expected? reason for bringing up this error is cuz

is this only on torch? or does it exist on other backends?

> ```
> idx = [i[1].sign().__neg__().maximum(0) * ret.shape[i[0]] + i[1] for i in tensor_found]
> 
> i[1] here is int64 from test_ops.py and i[1].sign() fails to realize because Assertion error above
> (also pytorch only allows int indices)
> ```
> 
> this problem is more or less tied to #1383 because on CPU, DIV also fails for 2 ints since it casts to float64...
> 
> a temporary workaround is to force indices to be float32 when doing `sign()`?

hmm, indices should probably be int32

geohotstan:
> is this only on torch? or does it exist on other backends?

only torch
cpu also gives float64 error lol, the other backends are fine

> hmm, indices should probably be int32

👌 I'll look into this

wozeparrot:
indices could technically be any integer type, but I think int32 makes the most sense, as I don't know of many cases where you might need to index anything larger than int32 could handle.

geohotstan:
#1495

ym1234:
I have a simpler gather that was written for onnx gatherelements, but the performance is...
This essentially fixes the indices as if the array was flattened (the for loop), and then does a normal gather as in #858
```Python
def GatherElements(data, indices, axis=0):
  indices = indices.reshape(*([1] * ((len(data.shape) - len(indices.shape))//2)), *indices.shape)
  ds = data.lazydata.st.views[-1].strides
  indices = (indices<0).where(indices + data.shape[axis], indices).cast(dtypes.float) * ds[axis]
  for n, ax in enumerate(ds):
    if n == axis: continue
    indices += Tensor.arange(indices.shape[n]).reshape(-1, *([1] * (len(data.shape) - n - 1))) * ax
  indices = indices.reshape(*indices.shape, *([1] * len(data.shape))).expand(*indices.shape, *data.shape)
  counter = Tensor.arange(prod(data.shape)).reshape(*data.shape)
  return (counter == indices).where(data, 0).sum(range(-1, -len(data.shape) - 1, -1))
```

roelofvandijk:
```
  for n, ax in enumerate(ds):
    if n == axis: continue
    indices += Tensor.arange(indices.shape[n]).reshape(-1, *([1] * (len(data.shape) - n - 1))) * ax
```
could be
```
indices += [Tensor.arange(indices.shape[i]).reshape(-1, *([1] * (len(data.shape) - i - 1))) * ax for i,ax in enumerate(ds) if i != axis]
```

ym1234:
mother of oneliners
```Python
def GatherElements(data, indices, axis=0): return (Tensor.arange(prod(data.shape)).reshape(*data.shape) == ((indices<0).where(indices + data.shape[axis], indices).cast(dtypes.float) * data.lazydata.st.views[-1].strides[axis] + functools.reduce(Tensor.__add__, [Tensor.arange(indices.shape[i]).reshape(-1, *([1] * (len(data.shape) - i - 1))) * ax for i,ax in enumerate(data.lazydata.st.views[-1].strides) if i != axis])).reshape(*indices.shape, *([1] * len(data.shape)))).where(data, 0).sum(range(-1, -len(data.shape) - 1, -1))
```

geohotstan:
> mother of oneliners
> 
> ```python
> def GatherElements(data, indices, axis=0): return (Tensor.arange(prod(data.shape)).reshape(*data.shape) == ((indices<0).where(indices + data.shape[axis], indices).cast(dtypes.float) * data.lazydata.st.views[-1].strides[axis] + functools.reduce(Tensor.__add__, [Tensor.arange(indices.shape[i]).reshape(-1, *([1] * (len(data.shape) - i - 1))) * ax for i,ax in enumerate(data.lazydata.st.views[-1].strides) if i != axis])).reshape(*indices.shape, *([1] * len(data.shape)))).where(data, 0).sum(range(-1, -len(data.shape) - 1, -1))
> ```

LOOOOOL a work of art.



> I have a simpler gather that was written for onnx gatherelements, but the performance is... This essentially fixes the indices as if the array was flattened (the for loop), and then does a normal gather as in #858
> 
> ```python
> def GatherElements(data, indices, axis=0):
>   ds = data.lazydata.st.views[-1].strides
>   indices = (indices<0).where(indices + data.shape[axis], indices).cast(dtypes.float) * ds[axis]
>   for n, ax in enumerate(ds):
>     if n == axis: continue
>     indices += Tensor.arange(indices.shape[n]).reshape(-1, *([1] * (len(data.shape) - n - 1))) * ax
>   indices = indices.reshape(*indices.shape, *([1] * len(data.shape))).expand(*indices.shape, *data.shape)
>   counter = Tensor.arange(prod(data.shape)).reshape(*data.shape)
>   return (counter == indices).where(data, 0).sum(range(-1, -len(data.shape) - 1, -1))
> ```

yeah this is actually quite a bit faster than the current gather implementation for this PR. Is it is it generalizable to be used for fancy index as well (is this even important?)
cuz I wrote a GatherElements before as well that's even faster, but it looks hacky
```
def GatherElements(input, indices, axis=0)
    indices = indices.transpose(ax1=axis, ax2=0)
    permute_args = list(range(input.ndim))
    permute_args[0], permute_args[axis] = permute_args[axis], permute_args[0]
    permute_args.append(permute_args.pop(0))
    input = input.permute(*permute_args).slice([(0,x) if i == input.ndim-1 else (0, indices.shape[i+1]) for i,x in enumerate(input.shape)])
    reshape_arg = [1]*input.ndim + [input.shape[-1]]
    return ((indices.unsqueeze(indices.ndim).expand(*indices.shape, input.shape[-1]) == Tensor.arange(input.shape[-1]).reshape(*reshape_arg).expand(*indices.shape, input.shape[-1]))*input).sum(indices.ndim).transpose(ax1=0, ax2=axis)
```

test/test_ops.py
```diff
@@ -240,7 +240,7 @@ def test_mul_const(self):
   def test_div(self):
     helper_test_op([(45,65), (45,65)], lambda x,y: x/y, Tensor.div)
     helper_test_op([(), ()], lambda x,y: x/y, Tensor.div)
-    helper_test_op(None, lambda x,y: x/y, Tensor.div, forward_only=True, vals=[[5],[1]])
+    helper_test_op(None, torch.div, Tensor.div, vals=[(5), (1)], forward_only=True)
```
wozeparrot, line 243: why did this change?

Issue #1400, opened 2023-08-02
Out Of Memory error on Mac M1 Pro, 32GB memory when running examples/stable_diffusion.py


beachwood23:
Saw two Mac crashes from an OOM when trying to run `python3 examples/stable_diffusion.py`. 

Running with a Mac M1 Pro, 32GB of memory. 

Running on basic `Terminal.app`. Here's an example of the output where the program freezes, and I was able to `ctrl-c` before it killed the computer again:

```
(venv) [~/github/tinygrad]$ python3 examples/stable_diffusion.py                                     [master]
ram used:  4.26 GB, cond_stage_model.transformer.text_model.final_layer_norm.bias: 100%|█| 1131/1131 [00:00<00
loaded weights in 146.54 ms, 4.26 GB loaded at 29.10 GB/s
got CLIP context (1, 77, 768)
got unconditional CLIP context (1, 77, 768)
running for [1, 201, 401, 601, 801] timesteps
  0   1: 100%|██████████████████████████████████████████████████████████████████| 5/5 [00:11<00:00,  2.26s/it]
decode (1, 512, 64, 64)
decode (1, 512, 128, 128)
decode (1, 512, 256, 256)
decode (1, 256, 512, 512)
(512, 512, 3)

^C^C^C^C^CTraceback (most recent call last):
  File "/Users/b23/github/tinygrad/examples/stable_diffusion.py", line 673, in <module>
    im = Image.fromarray(x.numpy())
                         ^^^^^^^^^
  File "/Users/b23/github/tinygrad/tinygrad/tensor.py", line 112, in numpy
    def numpy(self) -> np.ndarray: return self.lazydata.toCPU()
                                          ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/b23/github/tinygrad/tinygrad/lazy.py", line 190, in toCPU
    ret = cast(RawBuffer, realized).toCPU().reshape(self.shape)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/b23/github/tinygrad/tinygrad/runtime/lib.py", line 50, in toCPU
    self._copyout(x)
  File "/Users/b23/github/tinygrad/tinygrad/runtime/ops_gpu.py", line 52, in _copyout
    mapped = cl.enqueue_map_buffer(CL.cl_queue[self._buf.device], buf, cl.map_flags.WRITE, 0, self.size, dtype=self.dtype.np)
[...]

PR #1276, opened 2023-07-19
cl fixes for multigpu

wozeparrot:
initial part of #963

this also makes copyin actually async, using the same event tracking style that the metal backend is currently using.

PR #1402, opened 2023-08-02
cl fixes

wozeparrot:
Closes: #1400 

When #1276 was merged and made copyins actually async, it also meant that none of the buffer resources could actually be freed until a copyout happened.

This switches the buffer event tracking to be part of the buffer itself, which allows buffer resources to be potentially freed as soon as the first kernel that requires that buffer is executed.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_gpu.py                4               8
```

PR #1405, opened 2023-08-02
One more optimization for `mod` on a `MulNode`

ronak69:
Inspired by this note in `test_symbolic.py` i tried to implement it and it works!
```python
# NOTE: even though the mod max is 50, it can't know this without knowing about the mul
self.helper_test_variable(Variable.sum([Variable("a", 0, 7)*100, Variable("b", 0, 3)*50]) % 100, 0, 99, "((b*50)%100)")
```

I ran a fuzzer few times and it didn't find anything :)

For the record, my text-editor-math-proof(?):

```
i % m
where,
  m <= i
  m = g * gcd
  i = x * (f * gcd)
  i = (a*m + b), 0 <= b < m, i % m = b
  x*f = (p*g + q), 0 <= q < g, x*f % g = q

i = x * (f * gcd)
i = x*f * gcd
i = (p*g + q) * gcd
i = p*(g*gcd) + q*gcd
i = p*m + (x*f % g)*gcd
i % m = (x*f % g)*gcd
(x * f * gcd) % (g * gcd) = (x*f % g)*gcd

0 <= q < g
0 <= q*gcd < g*gcd
0 <= i % m < m

(t * 75) % 100 = (t * 3 * 25) % (4 * 25) = (t*3 % 4) * 25
(t * 15) % 9 = (t * 6) % 9 = (t * 2 * 3) % (3 * 3) = (t*2 % 3) * 3

if f == 1:
(t * 50) % 100 = (t % 2) * 50
(t * 12) % 9 = (t * 3) % 9 = (t % 3) * 3
```

### DIFF
--- a/test/unit/test_symbolic.py
+++ b/test/unit/test_symbolic.py
@@ -126,9 +126,16 @@ class TestSymbolic(unittest.TestCase):
   def test_sum_div_no_factor(self):
     self.helper_test_variable(Variable.sum([Variable("a", 0, 7)*5, Variable("b", 0, 3)*5]) // 2, 0, 25, "(((a*5)+(b*5))//2)")
 
+  def test_mul_mod_gcd(self):
+    self.helper_test_variable(Variable("a", 0, 9)*50  % 100, 0, 50, "((a%2)*50)")
+    self.helper_test_variable(Variable("a", 0, 9)*-75 % 100, 0, 75, "((a%4)*25)")
+    self.helper_test_variable(Variable("a", 0, 9)*75  % 100, 0, 75, "(((a*3)%4)*25)")
+    self.helper_test_variable(Variable("a", 0, 9)*-25 % 100, 0, 75, "(((a*3)%4)*25)")
+    self.helper_test_variable(Variable("a", 0, 9)*33  % 100, 0, 99, "((a*33)%100)")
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                10               3
lines added in the tinygrad folder: 7
```

ronak69:
You will have to check those as I can't do global benchmarks with my current setup.

ronak69:
Before i only considered a human's perspective (small numbers fast, big numbers slow). But for a computer big/small numbers are the same and instruction count is what matters.

Factoring out `gcd` adds one more multiplication: `*f` in `(x*f % g)*gcd`. But it also decreases `self.max`. It is interesting to see if optimizing `self.max` is worth it by adding one more multiplication.

If it is worse then i can remove the `gcd` case and only do the optimization in the case where `self.b == the_gcd`.
Which: 1. doesn't add an extra multiplication and 2. decreases `self.max`.

```python
  def __mod__(self, b: int):
    multiplier = self.b % b
    # (x * f) % (g * f) = (x % g)*f
    if multiplier > 1 and b % multiplier == 0: return (self.a % (b // multiplier)) * multiplier
    new_self = self.a * multiplier
    return Node.__mod__(new_self, b)
```

PR #1330, opened 2023-07-23
opencl, intel/radeon  

waifairer:
Here is additional information about the unusual failures on Metals on the 2018/2019 Intel+Radeon mac stack. I'm lucky enough to be using one of those machines. My goal is to filter some of the noise that's currently in the GH-1019 issue. There are a few conflicting reports in that issue and also I think some distracting assumptions were made. At first, I am targeting the test case with values that disappear during the cast from half to char.

Here's what I know. There's a decent amount of new information in this (not mentioned in the issue comments).

1. The issue only occurs on both tensors of size 4 and 8. The issue does not appear on tensors of size 16, 32 or 64. The issue does not occur on any non-multiple of 4.
2. The issue is intrinsically tied to the following hand-coded optimization:
```
  # TODO: this is breaking the tests
  for splits in [4]:
    if k.upcasted == 0 and len(k.full_unupcasted_shape) > 0 and k.full_unupcasted_shape[-1] % splits == 0:
      k.shift_to(len(k.full_unupcasted_shape)-1, splits, insert_before=len(k.full_unupcasted_shape))
      k.upcast()
```
@geohot appears to know that this is the segment causing the issue from the comment. Specifically, it's the _k.upcast()_ that causes us to fail _this test_. Commenting this single line ` k.upcast()` causes the test cases included in this MR to pass. 
3. I have captured the UOPS stack and prg for both the passing and failing cases. They're currently checked in as `information`.  Here's the skinny:

The failing range(4) case has the following opstack/prg:
```
====== failing, free upcast on, range(4) =======
UOps.DEFINE_GLOBAL  :                           []                               ('data0', dtypes.unsigned char)
UOps.DEFINE_GLOBAL  :                           []                               ('data1', dtypes.half)
UOps.LOOP           :                           []                               ([], 'global')
UOps.LOOP           :                           []                               ([], 'local')
UOps.LOAD           : <val1_0:float4:None>      []                               MemOp(name='data1', idx=<0>, local=False, memory_dtype=dtypes.half, valid=<1>, invalid_value=0.0)
UOps.STORE          :                           [<val1_0:float4:0>]              MemOp(name='data0', idx=<0>, local=False, memory_dtype=dtypes.unsigned char, valid=<1>, invalid_value=0.0)
UOps.STORE          :                           [<val1_0:float4:1>]              MemOp(name='data0', idx=<1>, local=False, memory_dtype=dtypes.unsigned char, valid=<1>, invalid_value=0.0)
[...]

PR #1406, opened 2023-08-02
[WIP] Get all passing Metal tests in CI [Bounty]

jacobpradels:
Todo list:

- [ ] Find all tests that pass on metal
- [ ] Enable them in CI

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -85,9 +85,8 @@ class TestBFloat16DType(unittest.TestCase):
 @unittest.skipIf((getenv("CI", "") != "" and Device.DEFAULT in ["LLVM"]) or Device.DEFAULT == "WEBGPU", "float16 broken in some CI backends")
 class TestHalfDtype(unittest.TestCase):
   def test_float16_to_np(self): _test_to_np(Tensor([1,2,3,4], dtype=dtypes.float16), np.float16, [1,2,3,4])
-  @unittest.skipIf(Device.DEFAULT == "METAL", "Metal Backend does not support float16")
   def test_casts_to_half(self): _test_casts_to([1,2,3,4], source_dtypes=[dtypes.float32, dtypes.int8, dtypes.uint8], target_dtype=dtypes.float16)
-  @unittest.skipIf(Device.DEFAULT == "METAL", "Metal Backend does not support float16")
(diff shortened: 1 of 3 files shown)


test/test_dtype.py
```diff
@@ -85,7 +85,9 @@ def test_bf16_disk_write_read(self):
 @unittest.skipIf((getenv("CI", "") != "" and Device.DEFAULT in ["LLVM"]) or Device.DEFAULT == "WEBGPU", "float16 broken in some CI backends")
 class TestHalfDtype(unittest.TestCase):
   def test_float16_to_np(self): _test_to_np(Tensor([1,2,3,4], dtype=dtypes.float16), np.float16, [1,2,3,4])
+  @unittest.skipIf(Device.DEFAULT == "METAL", "Metal Backend does not support float16")
```
wozeparrot, line 88: Metal does support float16???

wozeparrot:
There has been quite a bit of investigation into this over at #1330, we are pretty sure that the main issue is a miscompilation somewhere, so there does need to be fix implemented there.

test/test_dtype.py
wozeparrot: Yup that's the miscompilation I was talking about, check out #1330 for more info.

jacobpradels:
> There has been quite a bit of investigation into this over at #1330, we are pretty sure that the main issue is a miscompilation somewhere, so there does need to be fix implemented there.

For the context of this PR do we want to continue skipping the test with a note to enable it when a fix is implemented or should this PR be blocked by the fix?

wozeparrot:
I mean fixing the issue is a part of the bounty, so you can try to find a fix as well.

jacobpradels:
@wozeparrot For clarification, what is the full scope of this bounty?  Is it to get every test that can run on metal to run on metal and pass?

Ex: the speed tests fail because tinygrad runs slower than pytorch on some operations (sub, sum).  Are optimizing those operations on metal part of this as well? This also includes a bunch of tests in `test_tensor.py`, but those tests don't seem to be ran on any of the back ends in the CI.

My original understanding was that this bounty was to implement all tests that currently pass for metal into the CI.

jacobpradels:
Followed the story of this issue and understand the scope better. Closing PR.

PR #1411, opened 2023-08-03
Support implicit expand to symbolic shape in _broadcasted

chenyuxyz:
Part of #1353, followup on symbolic expand

Add support to implicit expand in `_broadcasted` like `Tensor.rand(3,4).reshape(3, Variable) + 1` by adding `__bool__` to Node. Before this change
```
a = Variable("a", 1, 8)
max(a, 1) == min(a, 1) == a
```
with this change
```
a = Variable("a", 1, 8)
max(a, 1) == a
min(a, 1) == 1
```
so `_broadcasted` can do expand from `1` to `a` correctly.

`max(2, a)` still evaluates to `a`, but I don't think we need to support that for the movementops


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                 2               1
```

PR #1413, opened 2023-08-03
Add ops_shm

wozeparrot:
A lot of the design decisions here are pretty specific to linux/the use cases of this shared memory buffer, mainly regarding `SHM_CACHE` which exists to reduce mmap/shared buffer access overhead and that shm buffers don't need to be freed during distributed training, so we don't really care about the lifetimes of these buffers and mainly just care about their speed so reducing overhead everywhere helps.

These buffers also don't allocate shared memory themselves and expect to use shared memory allocated beforehand.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_shm.py               29               0
lines added in the tinygrad folder: 29
```

test/unit/test_shm_tensor.py
```diff
@@ -0,0 +1,39 @@
...
+    s.close()
+    t_shm = t.to(f"shm:{shm_name}").realize()
+
+    # copy from shm
+    t2 = t_shm.to(Device.DEFAULT).realize()
+
+    assert np.allclose(t.numpy(), t2.numpy())
+    s.unlink()
+
+  @unittest.skipIf(CI, "CI doesn't like big mmaps")
```
geohot, line 23: Hmm, this isn't big, it's only 128MB. I really doubt it's the mmap, maybe it's the shm partition size?

wozeparrot: Yea, I'm not quite sure, but it does fail during the mmap, but it might be something like an mmap limit in actions?

geohot: I think there's `SHMMAX` and `SHMALL` curious what they are on CI

wozeparrot: Don't think that's it, cuz it reports the same on github actions as on my machine:

```
 SHMMNI   Shared memory segments                                    4096    0 0.00%
SHMALL   Shared memory pages                       18446744073692774399    0 0.00%
SHMMAX   Max size of shared memory segment (bytes)                  16E    -     -
SHMMIN   Min size of shared memory segment (bytes)                   1B    -     -
```

geohot:
Merged as is, big test can be a mystery for another time.

PR #1417, opened 2023-08-03
Simplify a for-loop header

ronak69:
`kmin+1` because `kmin == kmax` cases are already tested in the loop above this one.


geohot:
Ugh, this is actually different behavior. This is why we don't merge PRs like this.

PR #1418, opened 2023-08-03
support symbolic shape codegen

chenyuxyz:
part of #1353 , adding codegen support for a simple example

`(Tensor.rand(3, 4).reshape(vi, 4) + 1).realize()` generates this kernel
```
void E_s0(float* restrict data0, const float* restrict data1, const int* restrict SYM_i) {
int i = SYM_i[0];
for (int gidx0 = 0; gidx0 <= ((i*4)+-1); ++gidx0) {
    float val1_0 = data1[gidx0];
    float acc2_0 = 1.0f;
    float alu0 = (acc2_0+val1_0);
    data0[gidx0] = alu0;
  } /* global+local */
}
```

add a helper function `sym_rename` to change the symbolic part in the function name to `s0`, `s1`, `s2`,...

add a temporary flag `TEST_COMPILE_ONLY` to keep this for compile only. Will remove in the next PR that will add exec support.
also temporarily exclude CI because I was hit by this outstanding pytest-xdist [issue](https://github.com/pytest-dev/pytest/issues/3216) and given that I will support exec soon that has no issue on CI. Hope this is fine

running locally with `DEBUG=4 CLANG=1 python test/test_symbolic_ops.py` looks like this
image

### DIFF
--- a/tinygrad/codegen/cstyle.py
+++ b/tinygrad/codegen/cstyle.py
@@ -64,7 +64,7 @@ class CStyleLanguage(NamedTuple):
       return f"vload_half{'' if output_dtype.sz == 1 else str(output_dtype.sz)}(0, {buf_name}+{idx.render(render_cl, strip_parens=True)})"
     if output_dtype.sz > 1:
       return f"({output_dtype.name})(*(({self.smem_prefix if local else self.buffer_prefix}{buf_dtype.name}{output_dtype.sz}*)({buf_name}+{idx.render(render_cl, strip_parens=True)})))"
-    return f"*({buf_name}+{idx.render(render_cl, strip_parens=True)})" if self.uses_ptr_arithmetic else f"{buf_name}[{sym_render(idx, render_cl)}]"
+    return f"*({buf_name}+{idx.render(render_cl, strip_parens=True)})" if self.uses_ptr_arithmetic else f"{buf_name}[{idx.render(render_cl)}]"
 
   def render_local(self, name:str, size:int):
     return self.smem_prefix + f"float {name}[{size}];"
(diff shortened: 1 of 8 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/cstyle.py                 4               4
tinygrad/codegen/linearizer.py            10               5
tinygrad/codegen/wgsl.py                   1               1
tinygrad/ops.py                            3               2
tinygrad/shape/symbolic.py                 4               0
total                                     22              12
```

tinygrad/codegen/cstyle.py
```diff
@@ -64,12 +64,12 @@ def render_load(self, output_dtype, buf_name, buf_dtype, idx, local=False) -> st
       return f"vload_half{'' if output_dtype.sz == 1 else str(output_dtype.sz)}(0, {buf_name}+{idx.render(render_cl, strip_parens=True)})"
     if output_dtype.sz > 1:
       return f"({output_dtype.name})(*(({self.smem_prefix if local else self.buffer_prefix}{buf_dtype.name}{output_dtype.sz}*)({buf_name}+{idx.render(render_cl, strip_parens=True)})))"
-    return f"*({buf_name}+{idx.render(render_cl, strip_parens=True)})" if self.uses_ptr_arithmetic else f"{buf_name}[{idx.render(render_cl)}]"
+    return f"*({buf_name}+{idx.render(render_cl, strip_parens=True)})" if self.uses_ptr_arithmetic else f"{buf_name}[{sym_render(idx, render_cl)}]"
```
geohot, line 67: I don't understand why this has to change?

chenyuxyz: yeah my bad, I was passing 0 instead of NumNode(0) in the earlier version. fixed

geohot:
How many more PRs are you thinking?

I'm thinking we should put merging more on hold until LLaMA actually runs jitted on the GPU. This is becoming a lot of effort to review, you should be finding these things, not me.

PR #1419, opened 2023-08-03
HIP CPU

dc-dc-dc:
Got it working but it is brutally slow right now, with 16 cores it takes almost an hour just to run `test/test_ops.py`

### DIFF
--- a/tinygrad/runtime/ops_hip.py
+++ b/tinygrad/runtime/ops_hip.py
@@ -103,7 +103,7 @@ __device__ float4 log2(float4 x) { return float4(log2(x.x), log2(x.y), log2(x.z)
 __device__ float4 exp2(float4 x) { return float4(exp2(x.x), exp2(x.y), exp2(x.z), exp2(x.w)); }
 __device__ float4 sin(float4 x) { return float4(sin(x.x), sin(x.y), sin(x.z), sin(x.w)); }
 extern "C" __global__
-    """, smem_prefix = "__shared__ ", barrier = "__syncthreads();", float4 = "make_float4", uses_vload=True,
+    """, smem_prefix = "__shared__ ", barrier = "__syncthreads();" if not HIP_CPU else "", float4 = "make_float4", uses_vload=True,
     half_prekernel = ("#include <hip/hip_fp16.h>\nusing half4 = HIP_vector_type<half, 4>;\n" if not HIP_CPU else "using half_float::half; MAKE_VECTOR_TYPE(half, half);\n") + 
 """
 __device__ float vload_half(size_t offset, const half *p) { return (float)*(p + offset); }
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_hip.py               63              20
lines added in the tinygrad folder: 43
```

prusnak:
`'utf-8'` is the default encoding for `encode()` and `decode()`, so you can just call `.encode()`/`.decode()`

dc-dc-dc:
current bottleneck is that it has to compile the runtime for each kernel as HIP-CPU is missing `hipModuleLaunchKernel` and instead uses an inline macro to launch the kernel.

dc-dc-dc:
Found something that is closer to whats needed `hip::detail::launch_kernel_from_so` but currently getting seg faults when trying to use it

aseok:
@dc-dc-dc Is there any of already proposed bounties relates to your pr?

geohot:
Locked the bounty for you.

tinygrad/runtime/ops_hip.py
```diff
@@ -16,53 +22,90 @@
...
       asm = early_exec((["/opt/rocm/llvm/bin/llvm-objdump", '-d', '-'], prg))
       print('\n'.join([x for x in asm.decode('utf-8').split("\n") if 's_code_end' not in x]))
-
-    module = hip.hipModuleLoadData(prg)
-    self.prg = hip.hipModuleGetFunction(module, name)
+    if HIP_CPU:
+      import tempfile, hashlib, os, subprocess
+      fn = f"{tempfile.gettempdir()}/clang_{hashlib.md5(prg.encode('utf-8')).hexdigest()}.so"
+      if not os.path.exists(fn):
+        subprocess.check_output(args=('clang -g -std=c++20 --stdlib=libstdc++ -shared -O2 -Wall -Werror -x c++ -ltbb -ltbbmalloc -fPIC --rtlib=compiler-rt - -o '+fn+'.tmp').split(), input=prg.encode('utf-8'))
```
prusnak, line 49: this is a race condition - add process id to tmp file name - see `ops_clang.py` to see how it is done:

https://github.com/tinygrad/tinygrad/blob/a293c18d345a45ff7a315c8fb5da17437c129949/tinygrad/runtime/ops_clang.py#L38

prusnak (changes requested):

wozeparrot:
Update on this?

dc-dc-dc:
Will close and let someone else continue, I still think `hip::detail::launch_kernel_from_so` is the right path forward as using `hipLaunchKernelGGL` requires rebuilding the runtime which takes way too long and times out in CI

Tracking #1700

nimlgen:
`hip::detail::launch_kernel_from_so` does not support any kernel arguments at all. Call to `launch` [passes empty args tuple](https://github.com/ROCm-Developer-Tools/HIP-CPU/blob/master/src/include/hip/detail/api.hpp#L636) and seems there is a [comment](https://github.com/ROCm-Developer-Tools/HIP-CPU/blob/master/src/include/hip/detail/api.hpp#L638).

Compilation time of "HIP-CPU" include-only templated library is huge. I am not sure we can do anything with this, since even `blockIdx`, which is inside each kernel, can take up to 5-20s to compile.

PR #1421, opened 2023-08-03
Arm

stevenandersonz:
ARM Backend:

- [x] arm passing all tests
- [x] emulating at the program level with unicorn
- [x] fix branch conflicts
- [x] abstract arm logic from assembly.py into assembly_arm64.py and find a better way to select ARM64 as runtime


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_clang.py             15               6
```

geohot:
Cool!

So another option is to use https://www.unicorn-engine.org at the small ARM program level instead of using a full ARM qemu including the Python. The Python would stay in x86, but the generated code would be ARM.

If it's more than 10 lines, it probably isn't worth it...so lame that GitHub doesn't have ARM runners. I just bought an RK3588 devkit, maybe this'll be the first self hosted runner. Can you try Unicorn though? We really have to get this < 10 minutes.

extra/assembly/assembly.py
```diff
@@ -1,15 +1,15 @@
...
 from tinygrad.shape.symbolic import Variable, NumNode, MulNode, DivNode, ModNode, LtNode, SumNode, AndNode
 import functools
 import math
 from collections import defaultdict
 
-_type_to_letter = {dtypes.float32: 'f', dtypes.bool: 'p', dtypes.int32: 'i', dtypes.int64: 'a', dtypes.uint32: 'u', dtypes.uint64: 'b', dtypes._float4: 'x'}
+_type_to_letter = {dtypes.half: 'h', dtypes.float32: 'f', dtypes.bool: 'p', dtypes.int8: 'c', dtypes.int32: 'i', dtypes.int64: 'a', dtypes.uint8:'e', dtypes.uint32: 'u', dtypes.uint64: 'b', dtypes._float4: 'x'}
 def type_to_letter(x): return _type_to_letter[x[0]].upper() if x[1] else _type_to_letter[x[0]]
-
+ARM64 = getenv("ARM64", False)
```
geohot, line 12: This shouldn't be here, it should be abstracted to assembly_arm64

test/test_ops.py
```diff
@@ -360,6 +360,7 @@ def test_relu6(self):
   def test_hardswish(self):
     helper_test_op([(45,65)], lambda x: torch.nn.functional.hardswish(x), Tensor.hardswish, atol=1e-6, grad_atol=1e-6)
     helper_test_op([()], lambda x: torch.nn.functional.hardswish(x), Tensor.hardswish, atol=1e-6, grad_atol=1e-6)
+
```
geohot, line 363: Remove this whitespace change, ops isn't touched otherwise.

tinygrad/runtime/ops_clang.py
```diff
@@ -10,13 +11,21 @@
...
-      os.rename(fn+'.tmp', fn)
+    if not binary:
+      prg = CLANG_PROGRAM_HEADER + prg
+      # TODO: is there a way to not write this to disk?
+      if not os.path.exists(fn):
+        subprocess.check_output(args=('clang -shared -O2 -Wall -Werror -x c '+args['cflags']+' - -o '+fn+'.tmp').split(), input=prg.encode('utf-8'))
+        os.rename(fn+'.tmp', fn)
+    else:
+      if DEBUG >= 5: print(prg)
+      if getenv('ARM64'):
```
geohot, line 25: Do you need this if, or is just the `else` on binary enough?

lucafanselau: Isn't it the case that currently binary is never set to true for the github action, so this else branch is never run for CI. Might be a case that I am missing something though

stevenandersonz:
Sweet, I'll make the changes and implement unicorn.

geohot:
So, both #1512 and this are in similar places...how bad are the merge conflicts?

How should I determine who is merged first, or maybe one of you doesn't care.

geohot:
Looking good! I'll merge this one first since the PR was first.

* Are there a few tests to disable to get test speed to be not the slowest? (13 min -> 8 min)
* I guess I'm fine with the `len(ret) == 4` hack to be cleaned up later

This is expected to just work on my M1 Mac, right? Will test when I get home.

stevenandersonz:
Yes, it will just work with `ARM64=1 CLANG=1`. Also, Tests are running now in 5min.

To avoid having `len(ret) == 4` should all renderers should return 4 elements?

geohot:
Congrats! Working on my Mac. Send PayPal to george@tinygrad.org to claim bounty, and you are now a red in the discord!

geohot:
Found a bunch of bugs in this, interested in fixing them?

`x14` and `x15` seem used in a non generic way and you get lucky with the shapetracker output style.

Issue #1152, opened 2023-07-07
Fix view merging for masked views


geohot:
[  if vm2.mask: return None  # this isn't supported yet](https://github.com/geohot/tinygrad/blob/c5aea13a6503f07e93e46f345fbf52da3b54d28d/tinygrad/shape/shapetracker.py#L89)

g1y5x3:
While looking into it, I think there's also a bug at 
[if view.contiguous: return new_view, False # NOTE: if it's contiguous it can't have an offset](https://github.com/tinygrad/tinygrad/blob/c5aea13a6503f07e93e46f345fbf52da3b54d28d/tinygrad/shape/shapetracker.py#L115C23-L115C29)
where the mask index was not being tracked

Here is an example:
```
>>> c = ShapeTracker((4,2))
>>> c.views[-1].mask = ((0,2))
>>> c
ShapeTracker(shape=(4, 2), views=[View((4, 2), (2, 1), 0, (1, 1))])
>>> c.reshape((2,4))
ShapeTracker(shape=(2, 4), views=[View((2, 4), (4, 1), 0, None)])
```

geohot:
Err, views are immutable unless you violently reach inside them and change them :)

roelofvandijk:
Comes for free once Views are NamedTuples - they're immutable.

PR #1423, opened 2023-08-03
Better reshape

S-Lykles:
This would add support for reshaping masked views. I also rewrote `_reshape` to never use the (apparently) expensive `mege_views`, it uses shape_strides to determing whether a view can be reshaped.
Reshaping a view with a mask is possible as long as the mask can be combined/split to the new dimensions, there are a lot of edge cases with this and I don't think `_merge_masks` can be much simpler. Possibly the reshaping of maks could be done symbolically, similar to how `merge_views` calls `real_strides`, but I don't think with the current way that the the mask is symbolically expressed that would be simpler to do.

If a dimension with size one is padded, `_reshape` sometimes cannot (yet) determine that it is possible to reshape the view "in place" and makes the ShapeTracker create an addtitional view, thats why two of the tests are disabled for now. Of course this doesn't lead to incorrect results, just can't do the optimization.

This is sort of related to #1152


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/shapetracker.py            42              22
```

geohot:
Tests failing

tinygrad/shape/shapetracker.py
```diff
@@ -100,30 +100,54 @@ def merge_views(vm2:View, vm1:View) -> Optional[View]:
...
+  for d, s in reversed(view.shape_strides):
+    acc, next_stride = 1, s
+    while acc < d:
+      new_dim = next(reverse_shape)
+      acc *= new_dim
+      strides.append(next_stride)
+      next_stride *= new_dim
+    if acc != d: break
+  else:
+    while len(strides) < len(new_shape): strides.append(0)
```
roelofvandijk, line 145: `if len(new_shape) > len(strides) : strides += [0,] * (len(new_shape) - len(strides))`

tinygrad/shape/shapetracker.py
```diff
@@ -100,30 +100,54 @@ def merge_views(vm2:View, vm1:View) -> Optional[View]:
...
-          break
-      else:
-        new_mask: List[Tuple[int, int]] = [y for x,y in zip(shape, mask) if x != 1]
-        new_mask_tuple = tuple([(0,1) if x == 1 else new_mask.pop(0) for x in new_shape])
-    return View(new_shape, new_strides_tuple, offset, new_mask_tuple), False
-
-  new_view = View(new_shape, strides_for_shape(new_shape))
-  if view.contiguous: return new_view, False # NOTE: if it's contiguous it can't have an offset
-  if (merged_view := merge_views(view, new_view)) is not None: return merged_view, False
+  if view.contiguous or (view.shape == new_shape): return View(new_shape), False
```
roelofvandijk, line 132: Used to return `View(new_shape, strides_for_shape(new_shape))`, is that not necessary here?

tinygrad/shape/shapetracker.py
S-Lykles: Yeah that looks better, thanks

geohot:
So I'm excited about this change, but can I really trust this to be correct? It's very hard to read and has a lot of if conditions., and subtle bugs are really expensive.

Is there a way to make this simpler or more obviously correct?

S-Lykles:
I get it, there's a lot of: "if you've reached this line, it means x cannot be y and z must be a or b" 
Ill try and write those out more explicitly and add some more tests. 
Maybe I can write some verbose explanation here in the pull request as well.

geohot:
Tests + comments making the code more readable are very welcome

S-Lykles:
I added some comments and tests, hopefully its more readable now. Honestly, I'm not sure this is worth it, the benefit of doing the reshapes on masks is that it can lead to slighly simpler index expressions, possibly one fewer mod or div inside the final kernel (which is pretty slow for integers on GPU). But the ammount of masks that can actually be reshaped is relatively small and the logic is pretty complex. When I was figuring out the algorithm I wrote a small document that explains when a mask can and cannot be reshaped. I'l copy it into this comment, it can help with understanding the code.

# Reshaping a View
A View can always be reshaped to a different shape if the product of the dimensions of the new shape matches that of the old one, sometimes the creation of a new View on top of the old one is necessary, but often a View can be reshaped "in place". This document explains when it is possible to reshape in place. If it is said that a View cannot be reshaped, it is meant that it cannot be reshaped without creating an additional View.

## Basics
If this document refers to a set of two (or more) dimensions as being contiguous with eachother, it means that the stride of each dimension is equal to the lsower dimension times the stride of the lower dimension.
So in `ShapeTracker((10,10,25)).permute((2,0,1))`, dimension 1 and 2 are contigous with eachother because `stride[1] = shape[2]*stride[2]`. Of course the View as a whole is <u>not</u> contiguous.

Reshaping is an operation that preserves the order of the elements of the underlying buffer, therefore a reshape cannot happen across dimensions that are not contiguous with eachother. 

For example: 

`ShapeTracker((100,100)).permute((1,0))` can be reshaped as long as the reshape does not mix the two dimensions. `(10,10,10,10)`, `(10,5,2,4,25)`, `(20,5,50,2)`, `(25,2,2,100)` are all valid reshapes that can be done, `(4,50,50)` and `(25,5,80)` are not. 

Similarly in `ShapeTracker((10,10,25)).permute((2,0,1))` dimension 1 and 2 (formerly dimension 0 and 1) are still contiguous, meaning they can be reshaped to any "subshape" whose product equals 10*10. So `(5,5,100)`, `(5,5,4,25)` and `(25,1,50,2)` are all valid reshapes.

## Reshaping an expanded View
A View that has had one of its dimensions expanded can be reshaped so long as that dimension is only ever combined with other expanded dimensions. 
`ShapeTracker((1,1,4)).expand((5,10,4))` can be reshaped to `(50,4)` and `(25,2,2,2)` but not to `(25,4,2)`.

## Reshaping a masked View
If a View ever has a dimension with a mask, `(b,e)` where `b<=e`, this means the View has been padded and then shrunken to fall only within the padded region. Any index into the view will be invalid and the underlying buffer will never be accessed. Any reshape of such a View is valid so long as at least one of the new dimensions also has a mask with `b<=e`, indicating that it too can never have a valid index.

If the value of a mask of a dimension is: `(0, size_of_that_dimension)`, then we considered that dimension not to be masked.

Assuming that the View can be reshaped if it had no mask, it may be possible to fit the mask to the new shape.
When reshaping a View with one or more masked dimensions, one of four things can happen:

### 1. The dimension with the mask is preserved/unchanged
This speaks for itself, the mask transfers.

### 2. The masked dimension is split
For example:
```
ShapeTracker((5,10)).pad(((0,5),(0,0))).reshape((2,5,10)) --> mask=((0,1),(0,5),(0,10))
ShapeTracker((5,15)).pad(((0,5),(0,15))).reshape((2,5,2,3,5)) --> mask=((0,1),(0,5),(0,1),(0,3),(0,5))
```
The masks can be reshaped so long as the mask can be cleanly factored into the new shape, so `ShapeTracker((5,10)).pad(((0,5),(0,0))).reshape((5,2,10))` should create an extra View because `5%2!=0`

### 3. The masked dimension is combined with another (masked or non-masked) dimension to a single larger dimension
When combining two or more dimensions into a single new dimension, this is only possible if either
1. Only the highest dimension has a mask
```
ShapeTracker((5,10)).pad(((0,5),(0,0))).reshape((100,)) --> mask=((0,50),)
ShapeTracker((3,5,3)).pad(((2,2),(0,0),(0,0))).reshape((105,)) --> mask=((30,75),)
```
2. All the dimensions above the lowest dimension with a mask have only a single valid index.
```
ShapeTracker((1,5)).pad(((0,0),(0,5))).reshape((10,)) --> mask=((0,5),)
ShapeTracker((1,5)).pad(((4,5),(0,5))).reshape((100,)) --> mask=((40,45),)
ShapeTracker((1,1,3,10)).pad(((0,0),(1,1),(0,3),(0,0))).reshape((180,)) --> mask=((60,66),)
```

### 4. The masked dimension is both split and combined
For example:
```
ShapeTracker((1,5,5)).pad(((0,4),(0,5),(0,0))).reshape((10,25)) --> mask=((0,1),(0,25))
```
Rules of both 2 and 3 apply

scratch2.py
wozeparrot, line 1: empty file

geohot:
I'm sold on this, and trust you have thought about the cases. Merged!

chenyuxyz:
@geohot this breaks master CI on one of the symbolic reshape tests

wozeparrot:
is a fix easy or should we revert?

wozeparrot:
@S-Lykles can you rebase with the new symbolic stuff, because some tests are failing with this merged into master.

chenyuxyz:
judging from the error message, it's probably a simple fix on the while loop condition. But I am not sure if it's a one-off case or something deeper. Will let the author handle and let me know if you encounter any symbolic shape specific problem

S-Lykles:
@wozeparrot I will but I'm not at my pc until tomorrow

PR #1426, opened 2023-08-04
Fix Clang Race Condition

l-j-g:
implement thread lock until clang file is compiled


wozeparrot:
I don't think this even works? pytest uses multiple processes not threads so holding a thread lock will do nothing to prevent cross process race conditions?

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_clang.py              9               4
```

l-j-g:
> oh i see, maybe a process lock would work instead?

geohot:
No. Fix this with the file name, not a lock.

PR #1427, opened 2023-08-04
A torch-like CrossEntropyLoss

g1y5x3:
Initially found a small mathematical error from the cross entropy loss calculation in [CIFAR](https://github.com/tinygrad/tinygrad/blob/801bed4f66821b1d53f31f61d9bde40bdc0a957f/examples/hlb_cifar10.py#L107C36-L107C53) where
`out.mul(Y).mean()`
should be
`out.mul(Y).sum(axis=1).mean()`

Since this is probably the most frequently used loss function, I implemented a class inside `tinygrad.nn` that matches `torch` as well as tests.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/nn/__init__.py                    9               0
```

tinygrad/nn/__init__.py
```diff
@@ -122,3 +122,12 @@ def __init__(self, vocab_size:int, embed_size:int):
...
     vocab_counter = Tensor.arange(self.vocab_size, requires_grad=False).reshape(1, 1, self.vocab_size).expand(*idx.shape, self.vocab_size)
     return (vocab_counter == idx.unsqueeze(2).expand(*idx.shape, self.vocab_size)) @ self.weight
+
+class CrossEntropyLoss:
+  def __init__(self, reduction:str='mean', label_smoothing:float=0.0) -> None:
+    self.reduction, self.label_smoothing = reduction, label_smoothing
+  
+  def __call__(self, x:Tensor, y:Tensor) -> Tensor:
+    y = (1 - self.label_smoothing)*y + self.label_smoothing / y.shape[1]
+    return -x.log_softmax(axis=1).mul(y).sum(axis=1) if self.reduction=='none' else -x.log_softmax(axis=1).mul(y).sum(axis=1).mean() if self.reduction=='mean' else -x.log_softmax(axis=1).mul(y).sum(axis=1).sum() if self.reduction=='sum' else None
```
geohot, line 132: This is unreadable. Don't save lines like this.

geohot:
This should not be a class, it has no state.

g1y5x3:
I agree with no need to write as a class. I think when you first wrote CIFAR, you used `sparse_categorical_crossentropy` from [`extra/training.py`](https://github.com/tinygrad/tinygrad/blob/c08ed1949f57bcb400d746166b3fa51da2d9069a/extra/training.py#L6C5-L14) And the difference is just resulting in a smaller loss value during training that I wouldn't able to catch if not crunching the last 1%.

g1y5x3:
I moved it in `tensor.py` as a function, under other functional nn ops but don't know if they should stay there.

wozeparrot:
Hmm, idk if we should put this in `tensor.py` either, probably just fix this in cifar for now.

PR #1387, opened 2023-07-31
Match Torch speed for sum reduction on M1

nimlgen:
Match Torch speed for sum reduction on M1

PR #1429, opened 2023-08-04
correct children count with lazycache

nimlgen:
Seems we count children wrong with lazycache. Because of this we got really bad kernels generated with #1387 when one (huge) reduce was merged several times into elementwise.

Measured on hlb_cifar10, efficientnet, sd, real_world_tests - no bad surprises happend, performance is the same.

### DIFF
--- a/test/test_lazybuffer.py
+++ b/test/test_lazybuffer.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python
 import numpy as np
 import unittest
-from tinygrad.lazy import LazyBuffer
+from tinygrad.lazy import LazyBuffer, Device
 from tinygrad.tensor import Tensor
 from tinygrad.shape.symbolic import Variable
+from tinygrad.ops import GlobalCounters
 
 class TestLazyBuffer(unittest.TestCase):
   def test_fromcpu_buffer_sharing(self):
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                           3               3
```

geohot:
Can you add a test that would have caught this? As end to end as possible

nimlgen:
Done. Added a test that reproduces similar behavior.

test/test_lazybuffer.py
```diff
@@ -44,6 +45,28 @@ def test_shuffle_pad_ops_exp(self):
...
+  def test_children_count(self):
+    a = Tensor.rand(8,8,8)
+    d1 = a.sum((0))
+    d2 = a.sum((0)).reshape(32,2)
+    assert len(d1.lazydata.op.src[0].children) == 1
+    in1 = d1.reshape(16,4)
+    d3 = in1.reshape(8,8)
+    assert len(d3.lazydata.op.src[0].children) == 2
+
+    GlobalCounters.cache = []
```
geohot, line 58: This test isn't cleaning up state

nimlgen: Oops... done.

geohot:
Thanks for the fix. I'm curious what else this was breaking, it's obvious we lack a lot of test coverage.

PR #1436, opened 2023-08-04
inf, -inf support for pad

ZenginU:
With this PR we can match with torch API, Also we can use this in the onnx as this will be compatible in constant mode

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -7,7 +7,7 @@ import operator
 import numpy as np
 from typing import List, Tuple, Callable, Optional, ClassVar, Type, Union, Sequence, cast
 from tinygrad.helpers import ImageDType, argfix, make_pair, getenv, IMAGE, DEBUG, flatten, DType, dtypes
-from math import ceil, pi, prod, sqrt, log, cos, copysign
+from math import ceil, pi, prod, sqrt, log, cos, copysign, isinf
 from tinygrad.lazy import Device, LazyBuffer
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

ZenginU:
Operation was canceled, can you retrigger the CI checks?

wozeparrot:
Oh its cuda again, why is apt install being so slow

tinygrad/tensor.py
```diff
@@ -246,6 +246,7 @@ def flip(self, axis, *args) -> Tensor: return mlops.Flip.apply(self, axis=[x if
   def shrink(self, arg:Tuple[Tuple[int, int], ...]) -> Tensor: return mlops.Shrink.apply(self, arg=arg) if any(x != (0,s) for x,s in zip(arg, self.shape)) else self
   def pad(self, arg: Tuple[Tuple[int, int], ...], value:float=0) -> Tensor:
     ret = mlops.Pad.apply(self, arg=arg) if any(x != (0, 0) for x in arg) else self
+    if isinf(value): return ret + copysign(1,value)*(1/mlops.Pad.apply(Tensor.full(self.shape, value), arg=arg))
```
wozeparrot, line 249: can't you just do `copysign(1, value)/mlops.Pad...` here?

ZenginU: Assumed it will be more readable as I was dealing with 1/inf = 0, formatted it now.

Issue #1364, opened 2023-07-28
matmul() behavior different from torch


eliujl:
I noticed tinygrad matmul() behaves differently than torch.
Tinygrad:
from tinygrad.tensor import Tensor
y = Tensor([[2.0,0,-3.0]])
y1 = Tensor([2.0,0,-2.0])
y.shape
y1.shape
(y1.matmul(y)).numpy()
Output:
(1, 3)
(3,)
array([0., 0., 0.], dtype=float32)

Torch:
import torch

y = torch.tensor([[2.0,0,-3.0]])
y1 = torch.tensor([2.0,0,-2.0])
y1.matmul(y)
Output:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: mat1 and mat2 shapes cannot be multiplied (1x3 and 1x3)

The torch output seems to make more sense. Can somebody explain how Tinygrad gets [0., 0., 0.]?

Addiictet:
I wanted to help out, so I looked into the code a bit:

The multiplication of y1 (shape 3x1) with y (shape 1x3) is the following:

$$\left(
\begin{array}{c}
2.\\
0.\\
-2.
\end{array}
\right)
\left(\begin{array}{ccc}
2. & 0. & -3.
\end{array}\right)
\=
\left(\begin{array}{ccc}
4. & 0.& -4.\\
0. & 0. & 0. \\
-6. & 0.& 6.
\end{array}\right\)
$$

In the code, this matrix is then summed along the last dimension resulting in the matrix `[0., 0., 0.]`

Using the vectors:
```
y = Tensor([[2.0, 0.0, 2.0]])
y1 = Tensor([-3.0, 0.0, -2.0])
```
yields `[-10., 0., -10.]`

Now, why does the summation happen at the end?  
This is because, y1 is not 2D tensor, but a 1D tensor, whose second dimension of size 1 is implied.  

Using the vectors:
```
y = Tensor([[2.0, 0.0, -3.0]])
y1 = Tensor([[2.0], [0.0], [-2.0]])
```
the correct matrix representation is obtained

Both in torch and tinygrad, the dimension is added at the front, so two tensors of sizes 1x3 are obtained.   
Torch does not allow the multiplication of the two, but tinygrad does.  

As of now, I am not exactly sure how this dimensionality mismatch is resolved within tinygrad, but it has to do with the `mlops.MUL` or rather the `BinaryOps.MUL`.

[...]

PR #1438, opened 2023-08-04
Add assert for wrong matmul/dot shapes

francescocastelli:
Currently Tensor.dot works even with wrong shapes: mxa @ bxn, when a != b but either one of a or b is 1 (see #1364). 

Added an assert on the shapes to have the same behavior as numpy/torch.


wozeparrot:
I believe there is a `helper_test_exception` or something like that to ensure that tinygrad and torch both raise to same exception.

francescocastelli:
Thanks, done!

PR #1439, opened 2023-08-04
Fix NOUNROLL

akretz:
The `NOUNROLL` option is currently broken. Does it make sense to add a test run of `test/test_ops.py` with `NOUNROLL` enabled to catch this in future?
Also I've noticed that the currently failing Metal tests do pass with `NOUNROLL` enabled. Should we maybe enable these tests and run them like that?


tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
Nice fix. It's not a mainline option, so idk if it needs a test, and there's just getting lucky on METAL, have to root cause

PR #1446, opened 2023-08-05
Moved global max and local max to LinearizerOptions for webgpu

dc-dc-dc:
Also added an assert in the call for max bufs so it doesnt go to rust and fail there


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/renderer/wgsl.py                  0               2
tinygrad/runtime/ops_webgpu.py             2               1
total                                      2               3
```

dc-dc-dc:
Looks like #1252 removed the WEBGPU tests in CI so this bug went unnoticed

geohot:
Oh, can you reenable them? I should have been more strict with that PR.

PR #1443, opened 2023-08-05
fix backends for new style

geohot:

PR #1447, opened 2023-08-06
CIFAR works again on CUDA

terafo:
#1443 reintroduced #1224 due to:
```python
if self.opts.global_max and self.opts.local_max: self.limit_global_dims(3, self.opts.global_max, self.opts.local_max)
``` 
And the fact CUDA didn't have local_max set. This PR sets local_max for CUDA backend, CIFAR works again.

### DIFF
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -92,4 +92,4 @@ renderer = functools.partial(uops_to_cstyle, CStyleLanguage(
       __device__ __forceinline__ explicit operator float4() const {return make_float4(__half2float(x.x), __half2float(x.y), __half2float(y.x), __half2float(y.y)); }
     };
   """))
-CUDABuffer = Compiled(RawCUDABuffer, LinearizerOptions(supports_float4_alu=False, global_max = [65535, 65535, 2147483647]), renderer, CUDAProgram, cuda.Context.synchronize)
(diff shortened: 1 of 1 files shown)


geohot:
```
  Max dimension size of a thread block (x,y,z): (1024, 1024, 64)
  Max dimension size of a grid size    (x,y,z): (2147483647, 65535, 65535)
```

`global_max = [65535, 65535, 2147483647], local_max = [1024, 1024, 64])`

Something is backward.

wozeparrot:
The global max was added in #1299.

Tagging @g1y5x3

terafo:
I used the same X, Y, Z ordering as in WGPU local_max, since it was introduced for that backend.

g1y5x3:
This addition by itself does't really do anything. 

Currently the `hand_coded_optimizations` is not utilizing the `local dim` which was what caused some global dim allocation to exceed the limits at the first place. There have been some discussion to utilize the `local dim` with [grid-stride loops](https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/) which is probably the correct fix for `cuda` as well as `webgpu`  but it does require changing quite a bit of things.

> Something is backward

Due to the `global dim` flip in [`cstyle`](https://github.com/tinygrad/tinygrad/blob/e2af95c2f8a5a00fd76351d37fc6abecb2920acb/tinygrad/renderer/cstyle.py#L87) with example in #1224 (where `gidx0,1,2` was corresponding with `z,y,x` which is also happening with CL backend), the `global_max` I added was also flipped to align with the allocation.

terafo:
Flipped `local_max`, codegen flips it just like `global_max`. Should likely be flipped it for WGPU too. Didn't remove condition, since this change will likely be reverted in few days.

PR #1448, opened 2023-08-06
remove view_from_shape

chenyuxyz:
The function was created a year ago. Later code calls `View()` directly. Removing assert type check should be fine as long as mypy passed. Removed a manually passing `strides_for_shape(new_shape)` arg, which is re-computed anyway in `View()`.

Also replaced the last line of `strides_for_shape` with helper function `filter_strides`


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/shapetracker.py             3               8
lines added in the tinygrad folder: -5
```

PR #1458, opened 2023-08-06
fix file race condition in ops_clang

prusnak:
attempt to fix file race condition described in https://github.com/tinygrad/tinygrad/issues/1258

### DIFF
--- a/tinygrad/runtime/ops_clang.py
+++ b/tinygrad/runtime/ops_clang.py
@@ -17,8 +17,9 @@ class ClangProgram:
     # TODO: is there a way to not write this to disk?
     fn = f"{tempfile.gettempdir()}/clang_{hashlib.md5(prg.encode('utf-8')).hexdigest()}.{args['ext']}"
     if not os.path.exists(fn):
-      subprocess.check_output(args=('clang -shared -O2 -Wall -Werror -x c '+args['cflags']+' - -o '+fn+'.tmp').split(), input=prg.encode('utf-8'))
-      os.rename(fn+'.tmp', fn)
+      _, tmp = tempfile.mkstemp()
+      subprocess.check_output(args=('clang -shared -O2 -Wall -Werror -x c '+args['cflags']+' - -o '+tmp).split(), input=prg.encode('utf-8'))
+      os.rename(tmp, fn)
     self.lib = ctypes.CDLL(fn)
     self.fxn = self.lib[name]
 


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_clang.py              3               2
```

tinygrad/runtime/ops_clang.py
```diff
@@ -17,8 +17,9 @@ def __init__(self, name:str, prg:str):
     # TODO: is there a way to not write this to disk?
     fn = f"{tempfile.gettempdir()}/clang_{hashlib.md5(prg.encode('utf-8')).hexdigest()}.{args['ext']}"
     if not os.path.exists(fn):
-      subprocess.check_output(args=('clang -shared -O2 -Wall -Werror -x c '+args['cflags']+' - -o '+fn+'.tmp').split(), input=prg.encode('utf-8'))
-      os.rename(fn+'.tmp', fn)
+      _, tmp = tempfile.mkstemp()
```
wozeparrot, line 20: does this ever close the handle that `mkstemp` returns?

prusnak: That's a good point.

It seems there is no good way how to achieve what we need here in Python (generate a filename but not create a file): https://stackoverflow.com/questions/26541416/generate-temporary-file-names-without-creating-actual-file-in-python

We can either:
1. create a temporary directory (`tmp = os.path.join(tempfile.mkdtemp(), 'something')`), len clang create a file there, then remove both the file and the directory
2. use deprecated `mktemp`

What do you think @wozeparrot ?

wozeparrot: does it not work if we just add the current process pid to the filename?

prusnak: The PIDs can be recycled, so I don't think that's a good idea.

wozeparrot: it's only for the temporary file tho, and the pid is guaranteed to be unique as long as the process is alive which it will be while the tempfile is in use.

wozeparrot: even if same process calls it twice it will be sequential calls not parallel so the race condition won't happen

wozeparrot: yea, that would be nice.

prusnak: ack -> https://github.com/tinygrad/tinygrad/pull/1541

PR #1462, opened 2023-08-06
Actually flip local_max in CUDA

terafo:
Quick fix to #1447, only committed in previous pr, but forgot to push, so only original commit of PR got merged. Added comment to linearizer that explains flip behavior. Also it actually doesn't have to be done for WGPU, since it doesn't flip local dims when kernel renders.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py             1               0
total                                      2               1
```

PR #1463, opened 2023-08-06
Add WEBGPU tests to CI

dc-dc-dc:


geohot:
Hmm, why are the high lr ADAM tests failing? Those are pretty good bellwethers for other issues

dc-dc-dc:
failing due to accuracy, trying a different threshold

PR #1465, opened 2023-08-06
fix bug in symbolic's LtNode and AndNode

ronak69:
### logic bugs
* `(a < 10) * 2` is not equal to `(a*2) < 20`
* `(a < 10) // 2` is not equal to `(a//2) < 5`

same thing with `mul` and `floordiv` operations on `AndNode`.

---

solution: `Node`'s `mul` and `floordiv` are enough for `LtNode` and `AndNode`.

---

`test_shapetracker.py` does not ever invoke the methods i delete here. so i
assume that shapetracker in the "real world" does not use these methods as well.
(let's see what the CI says) (shouldn't matter anyway as i think `Node` will handle them correctly)

---

`fuzz_symbolic.py` can't catch these because it too doesn't ever invoke these
methods. (because in it, `AndNode` is never tested and all the comparison
operators are only ever appended at the last)

### DIFF
--- a/test/unit/test_symbolic.py
+++ b/test/unit/test_symbolic.py
@@ -24,22 +24,28 @@ class TestSymbolic(unittest.TestCase):
     self.helper_test_variable(Variable("a", 3, 8)<3, 0, 0, "0")
     self.helper_test_variable(Variable("a", 3, 8)<2, 0, 0, "0")
 
-  def test_ge_divides(self):
-    expr = (Variable("idx", 0, 511)*4 + Variable("FLOAT4_INDEX", 0, 3)) < 512
-    self.helper_test_variable(expr, 0, 1, "((idx*4)<512)")
-    self.helper_test_variable(expr//4, 0, 1, "(idx<128)")
-
-  def test_ge_divides_and(self):
-    expr = Variable.ands([(Variable("idx1", 0, 511)*4 + Variable("FLOAT4_INDEX", 0, 3)) < 512,
-                          (Variable("idx2", 0, 511)*4 + Variable("FLOAT4_INDEX", 0, 3)) < 512])
-    self.helper_test_variable(expr//4, 0, 1, "((idx1<128) and (idx2<128))")
-    expr = Variable.ands([(Variable("idx1", 0, 511)*4 + Variable("FLOAT4_INDEX", 0, 3)) < 512,
-                          (Variable("idx2", 0, 511)*4 + Variable("FLOAT8_INDEX", 0, 7)) < 512])
-    self.helper_test_variable(expr//4, 0, 1, "((((FLOAT8_INDEX//4)+idx2)<128) and (idx1<128))")
-
   def test_lt_factors(self):
     expr = Variable.ands([(Variable("idx1", 0, 511)*4 + Variable("FLOAT4_INDEX", 0, 256)) < 512])
     self.helper_test_variable(expr, 0, 1, "(((idx1*4)+FLOAT4_INDEX)<512)")
+
+  def test_operations_on_LtNode(self):
+    lt = Variable("a", 3, 8) < 4
+    self.helper_test_variable(lt  * 1, 0, 1, "(a<4)")
+    self.helper_test_variable(lt  * 2, 0, 2, "((a<4)*2)")
+    self.helper_test_variable(lt // 1, 0, 1, "(a<4)")
+    self.helper_test_variable(lt // 2, 0, 0, "0")
+    gt = Variable("a", 3, 8) > 4
+    self.helper_test_variable(gt  * 1, 0, 1, "((a*-1)<-4)")
+    self.helper_test_variable(gt  * 2, 0, 2, "(((a*-1)<-4)*2)")
+    self.helper_test_variable(gt // 1, 0, 1, "((a*-1)<-4)")
+    self.helper_test_variable(gt // 2, 0, 0, "0")
+
+  def test_operations_on_AndNode(self):
+    and_node = Variable.ands([Variable("a", 3, 8) < 4, Variable("b", 0, 18) > 7])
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                 1               4
```

chenyuxyz:
do you really think people write tests that you need to delete to fix a "bug"?

geohot:
I mean, it's possible the tests are wrong, but this PR isn't remotely close to the rigor required to merge a fix.

test/unit/test_symbolic.py
```diff
@@ -24,23 +24,29 @@ def test_lt(self):
     self.helper_test_variable(Variable("a", 3, 8)<3, 0, 0, "0")
     self.helper_test_variable(Variable("a", 3, 8)<2, 0, 0, "0")
 
-  def test_ge_divides(self):
-    expr = (Variable("idx", 0, 511)*4 + Variable("FLOAT4_INDEX", 0, 3)) < 512
-    self.helper_test_variable(expr, 0, 1, "((idx*4)<512)")
-    self.helper_test_variable(expr//4, 0, 1, "(idx<128)")
-
```
ronak69, line 31: > do you really think people write tests that you need to delete to fix a "bug"?

the "people" have wrote wrong tests @chenyuxyz.

```
idx = 50
FLOAT4_INDEX = 0

expr = idx*4 < 512
expr = 200 < 512
expr = 1

the real math:
expr//4 = 1//4 = 0

people's math:
(idx < 128) = 50 < 128 = 1

0 != 1
```

ronak69:
@geohot 

> I mean, it's possible the tests are wrong, but this PR isn't remotely close to the rigor required to merge a fix. You need to fix tests, not delete them.

i "fixed" the tests by writing new ones. i "deleted" them because they are wrong math.

ronak69:
If i am wrong i am gonna give up programming and math.

ZenginU:
> * `(a < 10) * 2` is not equal to `(a*2) < 20`

Why is this false? In symbolic math you can do this.

ronak69:
> > * `(a < 10) * 2` is not equal to `(a*2) < 20`
> 
> Why is this false? In symbolic math you can do this.

My current understanding of symbolic math is that you start with bounded variables instead of concrete values and the final simplified expression should give same results for every possible concrete value inside every bound.

`(a < 10)` is either false or true. If you multiply false(0) and true(1) by 2 in python you will get either 0 or 2. But `(a*2) < 20` will again produce only true and false.

Are you saying that i cannot substitute `a` here for a concrete value like `5`? ( `(5 < 10) * 2 == 2` is not equal to `(5*2) < 20 == 1`)

BTW thanks for your interest in this discussion.

ZenginU:
> the final simplified expression should give same results

As you mentioned here, (a<10)*2 is not the final simplified version. (2a<20) is the final


> `(5 < 10) * 2 == 2` is not equal to `(5*2) < 20 == 1`

You are right that this is not equal because you are assuming a=5 before evaluating the equation.
Let me put it this way

a = 2
(a<5)*2 != (2*a < 10)
Symbolic a
(a<5)*2 == (2*a < 10)

Also, you can check out the "Multiplication property of inequality"

ronak69:
> Symbolic a
> (a<5)2 == (2a < 10)

But we are not checking for "inequality" we are just simply multiplying the result of the "inequality".

ZenginU:
if a is symbolic you are not mutiplying the result, you are multiplying the symbolic eq first then put in the "a".

So if you are going to assume a = 5 for symbolic var a. it will be like this.

(a<5)*2 -> (2*a<10) -> (2*5<10)
Also, you can replicate this behaviour using MATLAB or other symbolic math tools.

ronak69:
Another proof:

take [`fuzz_symbolic.py`](https://github.com/tinygrad/tinygrad/blob/master/test/external/fuzz_symbolic.py) and run it with these modifications.

lines: 51-53
```python
    tape = [random.choice(ops) for _ in range(random.randint(2, 30))]
    # 10% of the time, add one of lt, le, gt, ge
    if random.random() < 0.1: tape.append(random.choice([lt, le, gt, ge]))
```

Replace with:
```python
    # 10% of the time, add one of lt, le, gt, ge
    tape = [random.choice(ops) if (random.random() > 0.1) else random.choice([lt, le, gt, ge]) for _ in range(random.randint(2, 30))]
```

ZenginU:
In this case you have multiple "lt, le, gt, ge" ops in the tape thats why it is failing, if there is multiple inequalities than thats a different story.

Still even if the tests are failing it does not change the fact that (a<5)*2 = (a*2<10) as this is not proven by tests but proven by the math itself. So this relies on mathematical foundations.

ronak69:
> Also, you can replicate this behaviour using MATLAB or other symbolic math tools.

`(x < 10)* 2`
[sympy](https://www.sympygamma.com/input/?i=%28x+%3C+10%29+*+2)
[wolfram alpha](https://www.wolframalpha.com/input?i=%28x+%3C+10%29+*+2)

ZenginU:
So only Matlab can handle this type of equations, as sympy did not implemented and wolfram alpha does not evaluate as it does not know if x is symbolic.

Still I can replicate this behaviour in MATLAB. So as my final point
Variable(a, 0, 5) * 2, should be 0 < 2*a < 10 and by definition variable a is (0 < a < 5)*2

ronak69:
> I can replicate this behaviour in MATLAB

screenshot please.

ronak69:
Ok fair enough.

I also believe in "math" and not "tests".

But `fuzz_symbolic.py` fails (even if you only use "lt" and not all four comparison operators).

shapetracker will use symbolic just like the fuzzer does. But don't forget what i said in the original description: shapetracker does not ever invoke these methods. This is i think a bug in "deadcode" (see the branch name).

PR #863, opened 2023-05-30
Training UNet3D

jla524:
Changes made:
- Implement iterator for training set
- Implement transformations
- Implement dice cross entropy loss

TODO:
- Support batch size > 1
- Implement LR warmup
- Run up to 4000 epochs
- Train the model (currently running out of memory on M1 Pro)
- Validate results (goal is 0.908 Mean DICE score)

PR #1469, opened 2023-08-06
UNet-3D Training

gml16:
This PR builds on top of #863 (and thanks to the help of @terafo).

Changes made by #863:
- Implement iterator for training set
- Implement transformations
- Implement dice cross entropy loss

Changes from this PR:
- Support batch size > 1
- Implement LR warmup
- Train the model
  - Big caveat that with my 4GB GPU I can only train with a smaller RoI than the benchmark. Thus, I am unable to check whether the required quality of the model is reached (goal is 0.908 Mean DICE score).

Feedback is more than welcome :)

### DIFF
--- a/examples/mlperf/metrics.py
+++ b/examples/mlperf/metrics.py
@@ -38,6 +38,8 @@ def one_hot(arr, num_classes=3):
   return arr
 
 def get_dice_score(prediction, target, channel_axis=1, smooth_nr=1e-6, smooth_dr=1e-6):
+  if not isinstance(prediction, Tensor):
+    prediction = Tensor(prediction)
   reduce_axis = list(range(2, len(prediction.shape)))
   prediction = prediction.softmax(channel_axis)
   assert target.shape == prediction.shape, f"Target and prediction shape do not match. Target: ({target.shape}), prediction: ({prediction.shape})."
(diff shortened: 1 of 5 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

gml16:
Thank you, that makes sense. I will be away for some time so the PR will be stale. Hopefully this work in progress can help someone else.

wozeparrot:
Closing due to no progress in a week.

Max-We:
I also worked on this bounty, but had the same memory issue. Leaving the following here in case it helps someone else.

Tinygrad has bad memory usage, at least when being compared to PyTorch. Consider the following example

```
import torch
import torch.nn as nn

# Assuming input tensor shape is (batch_size, channels, depth, height, width)
input_tensor = torch.randn((1, 32, 128, 128, 128), dtype=torch.float32)

# Define 3D convolution layer
conv = nn.Conv3d(in_channels=32, out_channels=32, kernel_size=(3, 3, 3), padding=(1, 1, 1))

# Apply convolution
conv_out = conv(input_tensor)

# Print the result
print(conv_out.shape)
```

On a CPU machine with 12 GB RAM (Google Colab) this works great in PyTorch, but the Tinygrad equivalent runs out of memory.

With small operations this is ok, but when training UNet3D you'll have to do larger convolutions which require unreasonable amounts of RAM (compared to PyTorch).

PR #1470, opened 2023-08-06
reopen Model benchmark correctness, ort, exceptions

geohotstan:
reopened because I made it a bit better, I think?
haven't written many try excepts before in my life, sorry.

but if this is still bad then just close

I used the ret from `ort` as ground truth to test the correctness of other results. (ort as best converage? and should be able to run onnx.models?)
so when `ort` fails to run, I just raise RuntimeWarning.

### DIFF
--- a/test/external/external_model_benchmark.py
+++ b/test/external/external_model_benchmark.py
@@ -15,10 +15,10 @@ from tinygrad.tensor import Tensor
 from tinygrad.lazy import Device
 
 MODELS = {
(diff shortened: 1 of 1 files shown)


geohotstan:
k nvm I just reread my code
it's trash

Once someone else does this, I'll read their code and learn.
gonna close again and not reopen lol, not gonna waste reviewer's time.
I need this to test my onnx

PR #1474, opened 2023-08-07
ci: add ccache and mold to gpuocelot build

roelofvandijk:
The gpuocelot install triggered a re-compilation, probably due to a changing mtime.

I am adding [`ccache`](https://ccache.dev/)` and [`mold`](https://github.com/rui314/mold).


geohot:
So this is trading complexity for potential speed. How much speed? It might not be worth it.

roelofvandijk:
> So this is trading complexity for potential speed. How much speed? It might not be worth it.

This seems to me to be a bugfix of the original implementation. Saves around 2m30s on the cuda tests.

.github/workflows/test.yml
```diff
@@ -235,24 +235,24 @@ jobs:
...
+          mkdir build && cd build
           cmake .. -Wno-dev -G Ninja -DOCELOT_BUILD_TOOLS=OFF
           ninja
       - name: Install gpuocelot
         if: matrix.backend == 'cuda'
         run: |
-          cd ${{ github.workspace }}/gpuocelot/ocelot/build
+          cd ${{ github.workspace }}/gpuocelot/ocelot
+          mkdir -p build && cd build
+          cmake .. -Wno-dev -G Ninja -DOCELOT_BUILD_TOOLS=OFF
```
geohot, line 254: Duplicated?

roelofvandijk: Yes - I thought the `cmake` and `ninja` calls were required to avoid re-compilation. Turns out there were not. Updated, slimmed down the MR, ready to go (see above).

geohot:
Now that's a nice PR!

PR #1475, opened 2023-08-07
[READY] ci: setup venv cache

roelofvandijk:
Caches installed site-packages. See below for timings vs. main. Saves 1-2 minutes per job.
Caches are invalidated on changes to `setup.py`.


geohot:
So this is trading complexity for potential speed. How much speed? It might not be worth it.

roelofvandijk:
> So this is trading complexity for potential speed. How much speed? It might not be worth it.
See timings above. I believe it is.

roelofvandijk:
> > So this is trading complexity for potential speed. How much speed? It might not be worth it.

See timings above. I believe it is.

okias:
In the CI it would make more sense to just cache 
 - `.cache/pip/`
 - `~/.local/lib/python*/site-packages/`

the result should be +- same without the useless venv (in this case).

roelofvandijk:
> In the CI it would make more sense to just cache
> 
>     * `~/.cache/pip/`
> 
>     * `~/.local/lib/python*/site-packages/`
> 
> 
> the result should be +- same without the useless venv (in this case).

Certainly possible!

I prefer the venv, it allows for easy parity between local and CI environments for debugging things - but that might be a matter of taste.

@geohot, do you want this at all or should I close it?

geohot:
Oh, I like the caching better. That venv spam everywhere is annoying

roelofvandijk:
Before (https://github.com/tinygrad/tinygrad/actions/runs/5912680119/usage):
image
After:
image

geohot:
Oh I like this a lot better

roelofvandijk:
Should be good, I don't know because no jobs seem to trigger. I added invalidation on changes to `setup.py`.

geohot:
Sweet!

PR #1481, opened 2023-08-07
distributed world

wozeparrot:
So it begins...

How the files are structured:
`__init__.py` is responsible for everything to do with out-of-band communication and distributed setup. The OOB requirements are really small which makes it easy later on to replace multiprocessing pipes with something that can go across machine boundaries. `preinit()` needs to be called before anything is even imported from tinygrad if someone is going to be using distributed training. This prevents runtimes that need it from initializing devices early, which causes lots of problems at least on AMD's opencl driver.

`world.py` is responsible for tensor send/recv, these currently work at the `RawBuffer` level by copying to shared memory and then to the other gpu so its pretty slow when scaled up.
The actual send/recv functions (`__send_rb` and `__recv_rb`) need to emulate ASTRunner's `__call__` so we can put them in the cache and have jit pick them up correctly. `cache_id` can be set to anything to allow the reuse of shared memory as creating and destroying them is pretty slow.

Put the test in external as it can't really follow the normal `pytest` format as we spawn our own processes and need to ensure that certain state is kept the same before everything starts. I'm not sure how feasible adding the test to CI (at least running in github actions) is due to the weird shared memory behaviour that github actions has.


test/external/dist/test_world.py
```diff
@@ -0,0 +1,58 @@
...
+from tinygrad.tensor import Tensor
+import numpy as np
+
+@TinyJit
+def send_jit(t, target_rank, cache_id=None) -> Tensor:
+  return world.send(t, target_rank, cache_id=cache_id).realize()
+
+@TinyJit
+def recv_jit(t, target_rank, cache_id=None) -> Tensor:
+  (t + 1).realize() # this does nothing but make sure that jit doesn't error
```
geohot, line 17: So I comment this out and get `AssertionError: some input tensors not found`

Why can't the JIT see the buffers in the recv?

wozeparrot: because t isn't actually used in the sense that jit cares about it in `recv`, `recv` just requires the shape and dtype of t.

geohot:
Is there a way to meaningfully test this in normal CI?

test/external/dist/test_world.py
geohot: I'm trying to understand if the JIT should be changed or not.

wozeparrot:
Yea, I'm going to try to get it in CI, will probably have to do small matrices again due to the same issue with actions and shared memory.

geohot:
Is there a good way to debug this? Maybe add prints on send/recv with DEBUG=3?

geohot:
> Yea, I'm going to try to get it in CI, will probably have to do small matrices again due to the same issue with actions and shared memory.

Yea totally fine, just anything to exercise the send/recv codepaths.

extra/dist/world.py
```diff
@@ -0,0 +1,80 @@
...
+# sends a lazybuffer from our rank to the target rank
+def _send_lb(x:LazyBuffer, target_rank:int, cache_id:Optional[str]=None) -> None: _send_rb(x.contiguous().realize().realized, target_rank, cache_id=cache_id)
+
+# receive a lazybuffer from the target rank
+def _recv_lb(shape, dtype, target_rank:int) -> LazyBuffer:
+  lb = LazyBuffer.loadop(LoadOps.EMPTY, shape, dtype, Device.DEFAULT).realize()
+  _recv_rb(lb.realized, target_rank)
+  return lb
+
+class Send(Function):
```
geohot, line 63: This is non obvious to me that recv is the backward pass of send. I'm reading your main PR and I still don't get it. Can you add comments to explain?

test/external/dist/test_world.py
wozeparrot: Yea, I just realized that it could probably reuse the rawbuffer, or maybe I already tried that and ran into something, anyways, I'll try that again.

geohot: Yea you are passing it in...it's unintuitive that it shouldn't use it. If it doesn't use it, the API should probably be changed to not take it.

extra/dist/world.py
wozeparrot: Actually I'm probably just gonna remove the backwards passes on these, until we get to the point that we need them, because none of the stuff I've written for multigpu training need this functionality.

wozeparrot:
yup good to merge

PR #1486, opened 2023-08-08
[READY] ci: missing n=auto

roelofvandijk:
Two pytests with missing -n=auto.


.github/workflows/test.yml
```diff
@@ -180,9 +180,9 @@ jobs:
     #  run: DEBUG=4 METAL=1 python -m pytest test/test_dtype.py
```
prusnak, line 180: Maybe update this one as well? (it is commented out, but if someone uncomments it, they will introduce the same issue as this PR is trying to fix)

```suggestion
    #  run: DEBUG=4 METAL=1 python -m pytest -n=auto test/test_dtype.py
```

roelofvandijk:
@prusnak Added to commented out tests.

PR #1491, opened 2023-08-08
add pyproject.toml (but keep setup.py)

prusnak:
Earlier effort tried to remove setup.py and replace it with pyproject.toml

Since these two files can coexist, this PR just adds pyproject.toml which is a straightforward port of setup.py.

This enables people to install/develop on tinygrad while using poetry/pdm but doesn't change anything else.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,51 @@
+[tool.poetry]
+name = "tinygrad"
+version = "0.6.0"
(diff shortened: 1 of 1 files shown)


prusnak:
Hm, it seems the CI tries to use pyproject.toml to install packages, so the CI needs to be adapted anyway.

Closing for now :-/

PR #1494, opened 2023-08-08
try to run commavq ($200 bounty to finish this)

geohot:


geohotstan:
There's a type_proto when `realize onnx::MatMul_6361` that does not have a `dim_value` which is causing the shape mismatch discussed in discord.

```
tensor_type {
  elem_type: 6
  shape {
    dim {
      dim_param: "batch_size"
    }
    dim {
      dim_param: "seq_len"
    }
  }
}
```

this is also the case with some of the other onnx model tests where they have some weird `type_proto`s

geohot:
@geohotstan is this fixed in your onnx PR? is that close to merging btw?

YassineYousfi:
these are tensors with dynamic shapes. 
currently run_onnx assumes only the first dimension can be dynamic and replaces it with 1.
https://github.com/tinygrad/tinygrad/blob/master/extra/onnx.py#L95

YassineYousfi:
Yes other these ops comming from: https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md will be needed in extra/onnx.py: 

- Attention
- EmbedLayerNormalization
- SkipLayerNormalization
- FastGelu

PR #1482, opened 2023-08-07
cache buffer loads across multiple bufs

chaosagent:
Continuing #1303. Added test.

Cache loads by underlying rawbuffer, not by AST leaf. Helps in cases where upcasting results in the same load across two AST subtrees. Note that this does not CSE valid.max == 0 variables across differnt rawbufs -- this will be handled in a subsequent PR that folds these away.

Example:

Before (just running the added Linearizer test):

```
*** exec  0.00 GB    0.02 ms op: BinaryOps.ADD        out(float): (3,)                           in(2): [(3,)] 
UOps.DEFINE_GLOBAL  :                           []                               ('data0', dtypes.float)
UOps.DEFINE_GLOBAL  :                           []                               ('data1', dtypes.float)
   0 buffer<3, dtypes.float>                         [View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True, shape_strides=((3, 1),))]
   1 buffer<4, dtypes.float>                         [View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True, shape_strides=((3, 1),))]
   2 buffer<4, dtypes.float>                         [View(shape=(3,), strides=(1,), offset=1, mask=None, contiguous=False, shape_strides=((3, 1),))]
   3
UOps.LOOP           :                           []                               ([], 'global')
UOps.LOOP           :                           []                               ([], 'local')
UOps.LOAD           : <val1_0>                  []                               MemOp(name='data1', idx=<0>, local=False, memory_dtype=dtypes.float, valid=<1>, invalid_value=0.0)
UOps.LOAD           : <val1_1>                  []                               MemOp(name='data1', idx=<1>, local=False, memory_dtype=dtypes.float, valid=<1>, invalid_value=0.0)
UOps.LOAD           : <val1_2>                  []                               MemOp(name='data1', idx=<2>, local=False, memory_dtype=dtypes.float, valid=<1>, invalid_value=0.0)
UOps.LOAD           : <val2_0>                  []                               MemOp(name='data1', idx=<1>, local=False, memory_dtype=dtypes.float, valid=<1>, invalid_value=0.0)
UOps.LOAD           : <val2_1>                  []                               MemOp(name='data1', idx=<2>, local=False, memory_dtype=dtypes.float, valid=<1>, invalid_value=0.0)
UOps.LOAD           : <val2_2>                  []                               MemOp(name='data1', idx=<3>, local=False, memory_dtype=dtypes.float, valid=<1>, invalid_value=0.0)
[...]

PR #1501, opened 2023-08-08
use ConstOp for valid.max == 0

chaosagent:


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py             4               2
tinygrad/renderer/cstyle.py                1               3
tinygrad/renderer/llvmir.py                1               3
total                                      6               8
```

tinygrad/codegen/linearizer.py
```diff
@@ -249,10 +250,11 @@ def global_load(self, i:int, idxs:Sequence[VariableOrNum], const=None) -> List[T
         idx, valid = self.sts[i].expr_idxs(_idx)
         localtype = dtypes.float32
       key = f"{localtype}{idx.render()}{valid.render()}"
+      this_const, valid, key = (invalid_value, cast(Variable, Variable.num(1)), f"{localtype}{valid.render()}") if valid.max == 0 else (const, valid, key)
```
geohot, line 253: Why does this need a cast? Why does the key render valid? It's just 0!

chaosagent: The cast is because NumNode(k) is not a Variable... :( Not sure why MemOp/ConstOp has Variables instead of Nodes, is it ok to change that?

chaosagent: I changed the valid render, but this is just temporary -- when this merges with #1482 I will just treat invalid as RawConst 0 valid 1.

tinygrad/codegen/linearizer.py
```diff
@@ -249,10 +250,11 @@ def global_load(self, i:int, idxs:Sequence[VariableOrNum], const=None) -> List[T
         idx, valid = self.sts[i].expr_idxs(_idx)
         localtype = dtypes.float32
       key = f"{localtype}{idx.render()}{valid.render()}"
```
geohot, line 252: Why do you render idx for invalid? Just put this on the next line.

PR #1502, opened 2023-08-09
Tensor.scaled_dot_product_attention to match torch, used in LLaMA, and tested

jla524:
changes made (so far)
- implemented function
  - support is_causal and attn_mask
- wrote test cases
- use function in LLaMA

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -681,7 +681,7 @@ class Tensor:
     return self * mask * (1/(1.0 - p))
 
   def scaled_dot_product_attention(query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False) -> Tensor:
-    return (query @ key.transpose(-2,-1) / sqrt(query.shape[-1])).softmax(-1).dropout(dropout_p) @ value
+    return (query @ key.transpose(-2,-1) / sqrt(query.shape[-1]) + attn_mask).softmax(-1).dropout(dropout_p) @ value
 
   # ***** cast ops *****
 
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -680,6 +680,9 @@ class Tensor:
     mask = (Tensor.rand(*self.shape, requires_grad=False) >= p).cast(dtypes.bool)
     return self * mask * (1/(1.0 - p))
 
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         3               0
```

geohot:
Nice to see you back! Bounty locked.

jla524:
Thanks George! This is ready for review.

Here are the results for llama (how does Stacy have a 7900XTX?):

```
jacky@ephphatha tinygrad % python3 examples/llama.py 
using METAL backend
using LLaMA-7B model
...
User: Which is better, PyTorch or tinygrad?
Stacy: PyTorch is better than tinygrad because it's faster and bigger. [EOS]

User: I'm pretty sure tinygrad is faster. Have you tried running test_speed_v_torch.py?                 
Stacy: I'm sorry, but I'm not sure. I haven't tried it. [EOS]

User: You should try it, because tinygrad is faster on M1 and 7900XTX.
Stacy: I am sorry. I don't have an M1, but I have a 7900XTX. [EOS]
```

tinygrad/tensor.py
```diff
@@ -680,6 +680,12 @@ def dropout(self, p=0.5) -> Tensor:
     mask = (Tensor.rand(*self.shape, requires_grad=False) >= p).cast(dtypes.bool)
     return self * mask * (1/(1.0 - p))
 
+  @staticmethod
+  def scaled_dot_product_attention(query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False) -> Tensor:
+    if is_causal: attn_mask = Tensor.ones(query.shape[-2], key.shape[-2]).tril(0).cast(dtypes.bool)
```
wozeparrot, line 685: Should probably `requires_grad=False` here

tinygrad/tensor.py
```diff
@@ -680,6 +680,12 @@ def dropout(self, p=0.5) -> Tensor:
     mask = (Tensor.rand(*self.shape, requires_grad=False) >= p).cast(dtypes.bool)
     return self * mask * (1/(1.0 - p))
 
+  @staticmethod
```
geohot, line 683: Instead of staticmethod, just make the query be `self`

Add type signatures to this function, and I think it's good!

geohot:
Bounty paid!

PR #1506, opened 2023-08-09
symbolic shapetracker

chenyuxyz:
part 1 of #1353 , i plan to split it into 3 PRs: (1) this one with symbolic shape and variable value buffer management, (2) ops and codegen, (3) jit support with jitted llama

partially reverted some of the premature changes i made earlier. also let me know if there's any suggestion to make reviewing these easier

### DIFF
--- a/test/unit/test_symbolic.py
+++ b/test/unit/test_symbolic.py
@@ -278,9 +278,11 @@ class TestSymbolicMinMax(unittest.TestCase):
 class TestSymRender(unittest.TestCase):
   def test_sym_render(self):
     a = Variable("a", 1, 8)
+    b = Variable("b", 1, 10)
     assert sym_render(a) == "a"
     assert sym_render(1) == "1"
     assert sym_render(a+1) == "(1+a)"
+    assert sym_render(a*b) == "(a*b)"
 
 class TestSymbolicSymbolicOps(unittest.TestCase):
   def test_node_div_node(self):
@@ -295,48 +297,25 @@ class TestSymbolicSymbolicOps(unittest.TestCase):
     idx0 = Variable("idx0", 0, i*3-1)
     assert NumNode(0) % (Variable("i", 1, 10)*128) == 0
     assert NumNode(127) % (Variable("i", 1, 10)*128) == 127
+    assert NumNode(128) % (Variable("i", 1, 10)*128 + 128) == 128
     assert idx0 % (i*3) == idx0
     assert i % i == 0
 
-  def test_mulnode_div_node(self):
+  def test_mulnode_divmod_node(self):
     i = Variable("i", 1, 10)
     idx0 = Variable("idx0", 0, 31)
-    assert (idx0*i) // i == idx0
-    assert (idx0*(i*128)) // i == (idx0*128)
     assert (idx0*(i*4+4)) // (i+1) == (idx0*4)
-
-  def test_mulnode_mod_node(self):
-    i = Variable("i", 1, 10)
-    idx0 = Variable("idx0", 0, 31)
-    assert (idx0*i) % i == 0
-    assert (idx0*(i*128)) % i == 0
     assert (idx0*(i*4+4)) % (i+1) == 0
-    assert (idx0*(i*4+4)+1) % (i+1) == 1
+    assert (idx0*i) % i == 0
 
   def test_sumnode_divmod_sumnode(self):
     i = Variable("i", 1, 10)
     idx0 = Variable("idx0", 0, 7)
     idx1 = Variable("idx1", 0, 3)
     idx2 = Variable("idx2", 0, i)
-
-    result = (idx0*(i*4+4)+idx1*(i+1)+idx2) // (i+1)
-    excepted = idx0*4+idx1
-    assert result - excepted == 0
-    result = (idx0*(i*4+4)+idx1*(i+1)+idx2) % (i+1)
-    excepted = idx2
-    assert result - excepted == 0
-
-  def test_sumnode_div_mulnode(self):
-    i = Variable("i", 1, 10)
-    idx0 = Variable("idx0", 0, 31)
-    idx1 = Variable("idx1", 0, 1)
-    idx2 = Variable("idx2", 0, 15)
-    idx3 = Variable("idx3", 0, i-1)
-    idx4 = Variable("idx4", 0, 3)
-
-    result = (idx0*(i*128)+idx1*(i*64)+idx2*(i*4) + idx3*4+idx4) // (i*4)
-    excepted = idx0*32+idx1*16+idx2
-    assert result - excepted == 0
+    assert (idx0*(i*4+4)+idx1*(i+1)+idx2) // (i+1) == idx0*4+idx1
+    assert (idx0*(i*4+4)+idx1*(i+1)+idx2) % (i+1) == idx2
+    assert (i+1) % (i*128+128) == (i+1)
 
 if __name__ == '__main__':
   unittest.main()
--- a/test/unit/test_symbolic.py
+++ b/test/unit/test_symbolic.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 import unittest
-from tinygrad.shape.symbolic import MulNode, SumNode, Variable, NumNode, sym_vars
+from tinygrad.shape.symbolic import MulNode, SumNode, Variable, NumNode, sym_vars, sym_render
 
 class TestSymbolic(unittest.TestCase):
   def helper_test_variable(self, v, n, m, s):
@@ -257,8 +257,7 @@ class TestSymbolicVars(unittest.TestCase):
     a = Variable("a", 0, 10)
     b = Variable("b", 0, 10)
     c = Variable("c", 0, 10)
-    # TODO: update this after we support symbolic * symbolic
-    assert (a + b * c).vars() == [a, b]
+    assert (a + b * c).vars() == [a, b, c]
     assert (a % 3 + b // 5).vars() == [a, b]
     assert (a + b + c - a).vars() == [b, c]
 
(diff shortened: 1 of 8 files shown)


test/unit/test_symbolic.py
```diff
@@ -276,6 +275,74 @@ def test_min_max_known(self):
...
+    assert NumNode(127) // (Variable("i", 1, 10)*128) == 0
+    assert idx0 // (i*3) == 0
+
+  def test_node_mod_node(self):
+    i = Variable("i", 1, 10)
+    idx0 = Variable("idx0", 0, i*3-1)
+    assert NumNode(0) % (Variable("i", 1, 10)*128) == 0
+    assert NumNode(127) % (Variable("i", 1, 10)*128) == 127
+    assert idx0 % (i*3) == idx0
+    assert i % i == 0
```
ronak69, line 299: can you add a test here that justifies `self - b == 0` over `self == b`?

I could think of only one "symbolic simplification" you mentioned in the main pr: `x*y - y*x == 0`. But looks like even `Node.sum()`/`SumNode` cannot simplify that.

I remember some `ProdNode` idea from earlier which may solve this by "sorting" just like `SumNode` does.

ronak69: oops, that would call `MulNode.__mod__()` not `Node.__mod__()`. But still, simplification is not done.

ronak69: Readability is the main concern here. *If* you *have* to do this "hack" to get more optimization can it not be centralized in `__eq__()` or `__lt__()` to keep the rest of the code clean?

tinygrad/shape/symbolic.py
```diff
@@ -191,11 +229,36 @@ def vars(self): return functools.reduce(lambda l,x: l+x.vars(), self.nodes, [])
...
+    if isinstance(b, Node):  # try to divide the whole b
+      fully_divided, rest = [], []
+      for x in self.flat_components:
+        try:
+          if x % b == 0: fully_divided.append(x // b)
+          else: rest.append(x)
+        except RuntimeError:
+          rest.append(x)
+      rest_sum = create_rednode(SumNode, rest)
+      if b - rest_sum.max > 0: return create_rednode(SumNode, fully_divided)
```
ronak69, line 257:
```python
x = Variable("x", 0, 10)
z = Variable("z", 100, 200)

div = (x - 1000) // z # should not be 0
mod = (x - 1000)  % z # should not be `self` or `(x - 1000)`
```

You missed lower bound checks (`b < rest_sum.min`).

You can get away with these checks in `Node` (because negative `self.min` is not allowed) and `MulNode` (because it is symmetric).

ronak69: *`-b < rest_sum.min` or `b > -rest_sum.min`

geohot:
A bunch of high level points. This is adding 92 lines of code to core tinygrad, and I think it can be simpler.

* Does this support Node RHS for all Node types, or just some? Which ones?
* I really don't like the use of `GlobalCounters` in lazy. If that's the choice, maybe we do have to support bufferless ints first.

Can this PR be only the changes to shapetracker/symbolic? This way we can deal with `lazy.py` stuff later.

chenyuxyz:
Updated to be symbolic only. Also removed functions that will only be used in runtime. It adds 24 lines now. The new `scaled_dot_product_attention` removed corner cases that previously need addition work. Now every case added is tied to jitting llama, and i also put the corresponding example in the unit test.

This supports `Node//Node` and `Node%Node` with positive `Nodes` and only if the expression can somehow be simplified. So `a//a` works but `b//a` or `3//a` does not work. And because `MulNode` is a stack now and i only simplify from one side, `(a*b)//b` works but `(b*a)//b` does not work.

`Node//SumNode` has a speculative handling that guess the answer by checking the constant terms. This is needed because I want to have only one variable in jitted llama, and after cat kv cache, the shape will become a `SumNode` and we have cases like `(v*4+4)//(v+1)`.

re: `lazy.py` I also wanted to remove the buffer. Currently it's there because I did not find a solution to pass the value directly to the compiled function. I will look into bufless ints next

chenyuxyz:
oh i think i know what to do for the bufless int, i will try that in the main PR

geohot:
Now, instead of overloaded `MulNode`, can we do `ProdNode` instead?

chenyuxyz:
I need some time to think about this, and maybe write a version to see how it works

By memory rep do you mean the `sum(idx * stride)` idxs formation?
It's not yet clear to me why we need `MulNode` once we have `ProdNode`. We don't have an `AddNode` and add is handled by `SumNode`. And the current places we check for `MulNode` will need to check for `ProdNode` too because the strides can be symbolic. Something like `idx0*(i*4+4)+idx1*(i+1)+idx2` in the test case will be a `SumNode` of `ProdNode`

`idx0*(i*4+4)` is a `MulNode` feels right because `a=idx0` is an idx and `b=(i*4+4)` is the number part (just not `int`). This is also why the current implementation doesn't need to change `shapetracker.py` to make it work

tinygrad/shape/symbolic.py
```diff
@@ -40,7 +40,6 @@ def __le__(self, b:Union[Node,int]): return self < (b+1)
   def __gt__(self, b:Union[Node,int]): return (-self) < (-b)
   def __ge__(self, b:Union[Node,int]): return (-self) < (-b+1)
   def __lt__(self, b:Union[Node,int]):
-    if self == b: return NumNode(0)
```
geohot, line 43: Why was this removed?

geohot:
Eh, okay I'm fine with `MulNode` as is. I'm okay with the structure of this, will review the code itself.

tinygrad/shape/symbolic.py
```diff
@@ -137,48 +142,49 @@ class NumNode(Node):
   def __init__(self, num:int):
     self.b, self.min, self.max = num, num, num
   def __int__(self): return self.b
+  def __eq__(self, other): return self.b == other
+  def __hash__(self): return self.hash  # needed with __eq__ override
 
 def create_node(ret:Node):
   assert ret.min <= ret.max, f"min greater than max! {ret.min} {ret.max} when creating {type(ret)} {ret}"
-  if ret.min == ret.max: return NumNode(ret.min)
+  if ret.min == ret.max: return NumNode(ret.min) if isinstance(ret.min, int) else ret.min
```
geohot, line 150: Huh? How is ret.min not an int?

tinygrad/shape/symbolic.py
```diff
@@ -63,8 +62,10 @@ def __rmul__(self, b:int): return self*b
 
   # *** complex ops ***
 
-  def __floordiv__(self, b:int, factoring_allowed=True):
+  def __rfloordiv__(self, b:int): raise RuntimeError(f"not supported: {b} // {self}")
+  def __floordiv__(self, b:Union[Node,int], factoring_allowed=True):
     assert b != 0
+    if isinstance(b, Node) and b > self >= 0: return NumNode(0)
```
geohot, line 68: `b > self >= 0` if b is Node? What is this?

tinygrad/shape/symbolic.py
```diff
@@ -75,8 +76,12 @@ def __floordiv__(self, b:int, factoring_allowed=True):
       return (self + -offset*b).__floordiv__(b, factoring_allowed=False) + offset
     return create_node(DivNode(self, b))
 
-  def __mod__(self, b:int):
+  def __rmod__(self, b:int):
+    if self.min > b >= 0: return NumNode(b)
```
geohot, line 80: Is there a test for this? Does the fuzzer hit this?

tinygrad/shape/symbolic.py
```diff
@@ -228,18 +250,20 @@ def flat_components(self): # recursively expand sumnode components
...
   elif typ == AndNode: ret.min, ret.max = (min([x.min for x in nodes]), max([x.max for x in nodes]))
   return create_node(ret)
 
+def sym_render(a: Union[Node, int], ops=None, ctx=None) -> str: return str(a) if isinstance(a, int) else a.render(ops, ctx)
+
 render_python: Dict[Type, Callable] = {
   Variable: lambda self,ops,ctx: f"{self.expr}[{self.min}-{self.max}]" if ctx == "DEBUG" else f"{self.expr}",
   NumNode: lambda self,ops,ctx: f"{self.b}",
-  MulNode: lambda self,ops,ctx: f"({self.a.render(ops,ctx)}*{self.b})",
+  MulNode: lambda self,ops,ctx: f"({self.a.render(ops,ctx)}*{sym_render(self.b)})",
```
geohot, line 266: How does this work without ops and ctx being passed in?

geohot:
A bunch of inline comments. Can you upgrade the fuzzer to hit these paths?

tinygrad/shape/symbolic.py
ronak69: (not reproducible now but was in 42c739b and 0fe885e)

oh wait, i got fooled by `MulNode`. It's not worth it to omit these checks anywhere, just check everywhere.

```python
(x*-100) % z # should not be (x*-100)
```

ronak69: Also note that you can't have `LtNode`s in an `and` expression, only concrete deterministic `bool`s will give correct answers.

tinygrad/shape/symbolic.py
chenyuxyz: it was implicitly tested in one of the test, and I added a direct test case against it. fuzzer does not hit this because in fuzzer all LHS is a `Node`. For general fuzzer support i will reply in another comment

tinygrad/shape/symbolic.py
chenyuxyz: with symbolic stride, idx can have symbolic max, and a node derived from that can have symbolic min, max

tinygrad/shape/symbolic.py
chenyuxyz: updated. It worked because RHS is treated as a symbolic number, and `Variable` and `NumNode` happened to be rendered the same way. The only difference is the debug view of the node.

tinygrad/shape/symbolic.py
chenyuxyz: This was a simplification case, I made this part explicit for what we support, so no dangling `ModNode` or `DivNode` with `b: Node` is allowed now. Only support if it can be simplified.

The old comparison has false positive case because the bool of a `LtNode` is false only if both min and max are 0, I have updated the code and add a test case.

chenyuxyz:
I think the current fuzzer does not work well with these paths because we only support if it can be simplify, and once it simplifies, the fuzz tape has no more `Node` RHS, and tapes like this is already covered.

To test "these paths correctly simplify with all possible symbol values", I think adding unit tests that test the path with a loop for all possible input is good? Working on it now

tinygrad/shape/symbolic.py
geohot: oh...this is never how max was intended to work. is this a requirement?

chenyuxyz: I misunderstood the root cause. The symbolic max for an idx is still needed, and we will later update renderer to render the max in for loop too.

The root cause was that `NumNode * Node` turned into `NumNode(Node)`, and the correct fix is to unwrap in `__mul__`. I have updated the PR and the tests accordingly.

chenyuxyz:
re: fuzz test. I am not sure how to approach it using the existing fuzzer. My understanding is that we want fuzz test to generate expressions that might not be covered by hand written tests and extensively test all possible variable values. For the new paths, almost all randomly generated expressions are not covered. And the hand written tests that cover these paths already test symbolically. Also since we don't support cases like `a*b==b*a` or `(a*b)//a==b`, an expression can be False even if all underlying numeric value matches.

The test_symbolic_shapetracker, test_symbolic_ops and test_symbolic_jit tests that I have in the main PR test against all possible `Variable` value in a for loop. This way we can make sure it works for different variable value.

tinygrad/shape/symbolic.py
```diff
@@ -228,21 +256,23 @@ def flat_components(self): # recursively expand sumnode components
...
 
+def sym_render(a: Union[Node, int], ops=None, ctx=None) -> str: return str(a) if isinstance(a, int) else a.render(ops, ctx)
+
 render_python: Dict[Type, Callable] = {
   Variable: lambda self,ops,ctx: f"{self.expr}[{self.min}-{self.max}]" if ctx == "DEBUG" else f"{self.expr}",
   NumNode: lambda self,ops,ctx: f"{self.b}",
-  MulNode: lambda self,ops,ctx: f"({self.a.render(ops,ctx)}*{self.b})",
+  MulNode: lambda self,ops,ctx: f"({self.a.render(ops,ctx)}*{sym_render(self.b,ops,ctx)})",
   DivNode: lambda self,ops,ctx: f"({self.a.render(ops,ctx)}//{self.b})",
   ModNode: lambda self,ops,ctx: f"({self.a.render(ops,ctx)}%{self.b})",
```
geohot, line 274: Can `self.b` be a Node for mod?

chenyuxyz: Not through mod operation. It can happen if one first creates a valid `ModNode` then changes `self.b` directly. There is an assert check on L197. Same for `DivNode`

geohot: Are you sure? Maybe I read the code wrong, but what if I Mod a SumNode with a Node?

geohot:
Made the decision to merge. I wish there was a fuzzer for the Symbolic stuff, but I don't think this causes any regressions, and it's a fairly manageable number of lines.

See above: once I'm spotting concerns that you have thought about well, it's probably time to merge!

PR #1511, opened 2023-08-09
add GPT2 example

YassineYousfi:
GPT2 is very similar to LlaMA. But is simpler to implement (no RoPE, normal LayerNorm, simpler FF), and is faster to download and play with.

commaVQ model is a standard GPT2 model.

I added it to the benchmark tests.


geohot:
Cool! Mac is spending almost all the time on the embedding, 50k is big. Merged into update_benchmark so it'll run on real hardware before pushing to master.

Issue #1412, opened 2023-08-03
Very slow in optimizing my custom network


L-M-Sherlock:
This script cost nearly 4 minutes in my Mac M2 Max.

```python
from tinygrad.tensor import Tensor
from typing import List, Optional
import numpy as np
import tinygrad.nn.optim as optim


def power_forgetting_curve(t, s):
    return (1 + t / (9 * s)) ** -1

class FSRS:
    def __init__(self, w: List[float]):
        self.s = np.array(w[:4], dtype=np.float32)
        self.w = Tensor(w[4:])

    def stability_after_success(self, new_s: Tensor, new_d: Tensor, r: Tensor, rating: Tensor) -> Tensor:
        hard_penalty = Tensor.where(rating == 2, self.w[11], 1)
        easy_bonus = Tensor.where(rating == 4, self.w[12], 1)
        new_s = new_s * (1 + Tensor.exp(self.w[4]) *
                        (11 - new_d) *
                        Tensor.pow(new_s, -self.w[5]) *
                        (Tensor.exp((1 - r) * self.w[6]) - 1) * 
                        hard_penalty * 
                        easy_bonus)
        return new_s

    def stability_after_failure(self, new_s: Tensor, new_d: Tensor, r: Tensor) -> Tensor:
        new_s = self.w[7] * \
                Tensor.pow(new_d, -self.w[8]) * \
                (Tensor.pow(new_s + 1, self.w[9]) - 1) * \
                Tensor.exp((1 - r) * self.w[10])
        return new_s

    def step(self, i, X: Tensor, new_s: Tensor, new_d: Tensor) -> Tensor:
        if i == 0:
            keys = np.array([1,2,3,4])
            # first learn, init memory states
            index = np.nonzero(X[1].numpy() == keys)
            new_s = self.s[index]
            new_s = Tensor(new_s)
            new_d = self.w[0] - self.w[1] * (X[1] - 3)
            new_d = new_d.unsqueeze(0)
            new_d = new_d.clip(1, 10)
        else:
            r = power_forgetting_curve(X[0], new_s)
            new_d = new_d - self.w[2] * (X[1] - 3)
            new_d = self.mean_reversion(self.w[0], new_d)
            new_d = new_d.clip(1, 10)
            new_s = Tensor.where(X[1] > 1.5, self.stability_after_success(new_s, new_d, r, X[1]), self.stability_after_failure(new_s, new_d, r))
        new_s = new_s.clip(0.1, 36500)
[...]

PR #1512, opened 2023-08-10
PTX Reintegration and Passing Tests

kawpuh:
Making progress on getting the tests passing, so opening a draft pull request. Will cleanup once all the tests are passing.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/renderer/assembly.py            204               0
tinygrad/renderer/assembly_ptx.py         90               0
tinygrad/runtime/ops_cuda.py               3               3
total                                    300               6
lines added in the tinygrad folder: 294
```

dc-dc-dc:
is this working with gpuocelot ?

kawpuh:
Had to fix the runtime args, but it is now. Actually passing more tests with `PTX=1 CUDACPU=1` than `PTX=1` alone.

kawpuh:
It occurred to me that in cases where e.g. PTX doesn't support the instruction with a specific type and an additional casting instruction is needed, I've been putting this in assembly.py as an additional asm instruction (which is probably not what we want if this is going to be shared with other asm backends). It seems plausible that this could all be moved to assembly_ptx, but at some level we have to deal with this in assembly.py or as some kind of pre-pass so that we can have all of the DEFINE_REGISTER instructions we need.

kawpuh:
Should just be two more code changes to get all the tests passing, then the change I mentioned above and cleanup. I don't have write access for tests.yml to add to CI. Otherwise, should be able to get it wrapped up in the next couple days if not tonight.

kawpuh:
So it looks like with just PTX, the CUDA env variable isn't required, but with `CUDACPU=1` it is, so I never actually properly tested gpuocelot before. I can't get it to build on my machine though, so I can't actually get it to run with or without PTX. 

The only thing that should be left is fixing overflow/underflow. Basically, PTX doesn't wrap on under/overflow when converting from float, which is exacerbated by the fact that linearizer always sets the Token.dtype to float for ConstOp. So the easiest solution is just to always cast to int first before casting to uint, but it seems strange to add a hacky extra instruction so that we can get worse overflow/underflow behavior?

kawpuh:
Could someone with write access add PTX to test.yml? Something like the patch below
<details>
<summary>Patch</summary>

```diff
diff --git a/.github/workflows/test.yml b/.github/workflows/test.yml
index 8ac1e08..19e7cee 100644
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -199,7 +199,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        backend: [llvm, clang, gpu, cuda]
+        backend: [llvm, clang, gpu, cuda, ptx]
 
     name: Tests on (${{ matrix.backend }})
     runs-on: ${{ matrix.backend == 'gpu'  && 'ubuntu-20.04' || 'ubuntu-latest' }}
@@ -218,7 +218,7 @@ jobs:
           path: '~/.cache/pip'
           key: ${{ matrix.backend }}
       - name: Set env
-        run: printf "${{ matrix.backend == 'llvm' && 'ENABLE_METHOD_CACHE=1\nLLVM=1' || matrix.backend == 'clang' && 'CLANG=1\nENABLED_METHOD_CACHE=1' || matrix.backend == 'gpu' && 'GPU=1' || matrix.backend == 'cuda' && 'FORWARD_ONLY=1\nJIT=1\nOPT=2\nCUDA=1\nCUDACPU=1\n'}}" >> $GITHUB_ENV
+        run: printf "${{ matrix.backend == 'llvm' && 'ENABLE_METHOD_CACHE=1\nLLVM=1' || matrix.backend == 'clang' && 'CLANG=1\nENABLED_METHOD_CACHE=1' || matrix.backend == 'gpu' && 'GPU=1' || matrix.backend == 'cuda' && 'FORWARD_ONLY=1\nJIT=1\nOPT=2\nCUDA=1\nCUDACPU=1\n' || matrix.backend == 'PTX' && 'PTX=1'}} " >> $GITHUB_ENV
       - name: Find faster apt mirror
         uses: vegardit/fast-apt-mirror.sh@v1
       - name: Install packages (gpu)
@@ -229,12 +229,12 @@ jobs:
           sudo apt-get update -y && \
           sudo apt-get install -y intel-oneapi-runtime-compilers intel-oneapi-runtime-opencl
       - name: Install packages (cuda)
-        if: matrix.backend == 'cuda'
+        if: matrix.backend == 'cuda' || matrix.backend == 'ptx'
         run: |
           sudo apt-get update -y && \
           sudo apt-get install -y --no-install-recommends git g++ cmake ninja-build llvm-15-dev zlib1g-dev libglew-dev flex bison libfl-dev libboost-thread-dev libboost-filesystem-dev nvidia-cuda-toolkit-gcc
       - name: Cache gpuocelot
-        if: matrix.backend == 'cuda'
+        if: matrix.backend == 'cuda' || matrix.backend == 'ptx'
         id: cache-build
         uses: actions/cache@v3
         env:
@@ -243,7 +243,7 @@ jobs:
           path: ${{ github.workspace }}/gpuocelot/ocelot
           key: ubuntu22.04-gpuocelot-19626fc00b6ee321638c3111074269c69050e091
       - name: Clone/compile gpuocelot
-        if: matrix.backend == 'cuda' && steps.cache-build.outputs.cache-hit != 'true'
+        if: (matrix.backend == 'cuda' || matrix.backend == 'ptx') && steps.cache-build.outputs.cache-hit != 'true'
         run: |
           git clone --recurse-submodules https://github.com/gpuocelot/gpuocelot.git ${{ github.workspace }}/gpuocelot
           cd ${{ github.workspace }}/gpuocelot/ocelot
@@ -253,7 +253,7 @@ jobs:
           cmake .. -Wno-dev -G Ninja -DOCELOT_BUILD_TOOLS=OFF
           ninja
       - name: Install gpuocelot
-        if: matrix.backend == 'cuda'
+        if: matrix.backend == 'cuda' || matrix.backend == 'ptx'
         run: |
           cd ${{ github.workspace }}/gpuocelot/ocelot/build
           sudo ninja install
@@ -262,8 +262,11 @@ jobs:
       - name: Check Device.DEFAULT
         run: python -c "from tinygrad.lazy import Device; assert Device.DEFAULT in ['LLVM','CLANG','CUDA','GPU'], Device.DEFAULT"
       - name: Run pytest (not cuda)
-        if: matrix.backend!='cuda'
+        if: matrix.backend!='cuda' && matrix.backend!='ptx'
         run: python -m pytest -n=auto test/ -k '${{matrix.backend=='llvm'&&'not (test_nn.py and test_conv_transpose2d)'||'test'}}' -m 'not exclude_${{matrix.backend}}'
       - name: Run pytest (cuda)
         if: matrix.backend=='cuda'
         run: python -m pytest -n=auto test/ -k 'not (half or test_efficientnet_safetensors) and not (test_conv2d and test_tensor.py)' -m 'not exclude_cuda' --ignore=test/external --ignore=test/models
+      - name: Run pytest (ptx)
+        if: matrix.backend=='ptx'
+        run: python -m pytest -n=auto test/ -k 'not test_uops' -m 'not exclude_cuda' --ignore=test/external --ignore=test/models
```

</details>

geohot:
Can you move assembly and assembly_ptx back to extra/assembly so I can read the diff?

tinygrad/ops.py
```diff
@@ -160,8 +160,8 @@ def to_program(self, k):
     k.linearize()
     src, global_size, local_size = self.renderer(k.function_name, k.uops)
     return ASTRunner(k.function_name, src, global_size, local_size,
-                      op_estimate=k.info.flops, mem_estimate=k.mem_estimate,
-                      display_name=k.display_name).build(self.runtime)
+                     op_estimate=k.info.flops, mem_estimate=k.mem_estimate,
+                     display_name=k.display_name, runtime_args={"binary": True} if self.runtime.__name__ == "PTXProgram" else None).build(self.runtime)
```
geohot, line 164: Hmm, I don't think this line belongs here.

tinygrad/runtime/ops_cuda.py
```diff
@@ -80,6 +80,10 @@ def __call__(self, global_size, local_size, *args, wait=False):
       end.synchronize()
       return start.time_till(end)*1e-3
 
+class PTXProgram(CUDAProgram):
```
geohot, line 83: I don't think you need this, the renderer should just create a CUDAProgram with binary

kawpuh: Definitely don't need both this and the above, but CUDAProgram gets that binary setting from runtime_args in ops. So I believe it either has to be set in ops or set in form in CUDAProgram. Could just add
`if getenv("PTX"): binary = True` in CUDAProgram's __init__ if you think that's cleaner? Or maybe rearrange the args for __init__ and wrap it with a `partial`, but I'm not sure if that binary arg is ever used outside of PTX?

geohot:
So, both #1412 and this are in similar places...how bad are the merge conflicts?

How should I determine who is merged first, or maybe one of you doesn't care.

stevenandersonz:
It's mostly the same, we are handling casting in different ways tho.  I think adding UOps.CAST  inside load and store adds more complexity. My approach basically is the same as in `llvmir.py`.

I'm passing all the tests btw, you were right about unicorn I was able to bring it down to 12 min with ~26 new lines of code.

geohot:
I merged ARM first (since the PR was first). Moving assembly to the main codebase #1544

tinygrad/runtime/ops_cuda.py
```diff
@@ -91,5 +91,5 @@ def __call__(self, global_size, local_size, *args, wait=False):
       __device__ __forceinline__ explicit half4(const float4& a): x(make_half2(__float2half(a.x), __float2half(a.y))), y(make_half2(__float2half(a.z),__float2half(a.w))) {}
       __device__ __forceinline__ explicit operator float4() const {return make_float4(__half2float(x.x), __half2float(x.y), __half2float(y.x), __half2float(y.y)); }
     };
-  """))
-CUDABuffer = Compiled(RawCUDABuffer, LinearizerOptions(supports_float4_alu=False, global_max = [65535, 65535, 2147483647], local_max = [64, 1024, 1024]), renderer, CUDAProgram, cuda.Context.synchronize)
+  """)) if not getenv("PTX") else fromimport("tinygrad.codegen.assembly_ptx", "uops_to_ptx_asm")
+CUDABuffer = Compiled(RawCUDABuffer, LinearizerOptions(supports_float4=False if getenv("PTX") else True, supports_float4_alu=False, global_max = [65535, 65535, 2147483647], local_max = [64, 1024, 1024]), renderer, CUDAProgram, cuda.Context.synchronize)
```
geohot, line 95: Uhhh, is this actually calling into PTX? It seems like it's just switching the supports_float4

kawpuh: Yeah, it might be hard to see from the formatting, but it's setting the renderer = uops_to_ptx_asm

geohot: oh...haha that diff is hard to read

geohot:
On a real NVIDIA machine

```
batman@x840:~/tinygrad$ FORWARD_ONLY=1 PTX=1 CUDA=1 python3 test/test_ops.py TestOps.test_add
test_add (__main__.TestOps) ... /home/batman/.pyenv/versions/3.8.10/lib/python3.8/site-packages/pycuda/compyte/dtypes.py:120: DeprecationWarning: `np.bool8` is a deprecated alias for `np.bool_`.  (Deprecated NumPy 1.24)
  reg.get_or_register_dtype("bool", np.bool8)
ERROR

======================================================================
ERROR: test_add (__main__.TestOps)
  File "test/test_ops.py", line 214, in test_add
    helper_test_op([(45,68), (45,68)], lambda x,y: x+y, Tensor.add)
  File "test/test_ops.py", line 25, in helper_test_op
    ret = tinygrad_fxn(*tst).realize()
  File "/home/batman/xx/openpilot/tinygrad/tensor.py", line 94, in realize
    self.lazydata.realize()
  File "/home/batman/xx/openpilot/tinygrad/lazy.py", line 159, in realize
    self.realized = Device[self.device].exec_ast(self.op, output=self, **self._device_extra_args())
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 196, in exec_ast
    if k.key not in self.method_cache: self.method_cache[k.key] = get_program()
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 193, in get_program
    return self.to_program(k)
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 162, in to_program
    return ASTRunner(k.function_name, src, global_size, local_size,
  File "/home/batman/xx/openpilot/tinygrad/ops.py", line 131, in build
    self.clprg = runtime(self.name, self.prg, **self.runtime_args)
  File "/home/batman/xx/openpilot/tinygrad/runtime/ops_cuda.py", line 71, in __init__
    self.prg = cuda.module_from_buffer(prg.encode('utf-8')).get_function(prg.split(".visible .entry ")[1].split("(")[0])
pycuda._driver.LogicError: cuModuleLoadDataEx failed: the provided PTX was compiled with an unsupported toolchain. - ptxas application ptx input, line 1; fatal   : Unsupported .version 8.2; current version is '8.1'

Ran 1 test in 0.120s

FAILED (errors=1)
```

kawpuh:
Actually, could probably just search for the .version line in an nvcc compiled kernel. I'll look into it tonight if you prefer that approach.

geohot:
```
batman@x840:~$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2022 NVIDIA Corporation
Built on Wed_Sep_21_10:33:58_PDT_2022
Cuda compilation tools, release 11.8, V11.8.89
Build cuda_11.8.r11.8/compiler.31833905_0
```

I think you can just use an older version everywhere if you aren't using any of the new features

kawpuh:
I'm at a bit of a loss why the error message says current version = 8.1 with CUDA 11.8, so I might be misinterpreting the chart. It seems like ISA 7.8 is the oldest that supports the 4090 though, so I went with that.

geohot:
Bounty is yours! E-mail your PayPal to george@tinygrad.org to claim.

PR #1519, opened 2023-08-10
distributed collectives

wozeparrot:
really just allreduce


geohot:
CI test is failing I think but still getting green check

wozeparrot:
huh, odd.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_gpu.py                6               3
```

wozeparrot:
so apparently `wait_for` just doesn't wait under certain conditions??????

wozeparrot:
ok should be good to go now.

tinygrad/runtime/ops_gpu.py
```diff
@@ -43,12 +43,13 @@ def __init__(self, size, dtype, device='0'):
 
   def _copyin(self, x:np.ndarray):
     assert not self.dtype.name.startswith("image"), f"can't copyin images {self.dtype}"
-    self.event = cl.enqueue_copy(CL.cl_queue[self._buf.device], self._buf, np.require(x, requirements='C'), is_blocking=False)
+    self.event = cl.enqueue_copy(CL.cl_queue[self._buf.device], self._buf, np.require(x, requirements='C'), is_blocking=True)
```
geohot, line 46: Why does this have to change to blocking?

wozeparrot: was testing stuff, forgot to revert

tinygrad/runtime/ops_gpu.py
```diff
@@ -77,7 +78,9 @@ def max_work_group_size(): return CL.cl_ctxs[0].devices[0].max_work_group_size
 
   def __call__(self, global_size, local_size, *bufs, wait=False) -> Optional[float]:
     cl_bufs = [x._buf if isinstance(x, CLBuffer) else x for x in bufs]
-    e = self.clprgs[cl_bufs[0].device](CL.cl_queue[cl_bufs[0].device], [g*l for g,l in zip(global_size, local_size)] if local_size is not None else global_size, local_size, *cl_bufs, wait_for=[x.event for x in bufs if isinstance(x, CLBuffer) and hasattr(x, "event")])
+    for x in bufs: # ensure that all buffers are ready
```
geohot, line 81: Does `wait_for` in PyOpenCL do this under the hood, or does it use something lower level to do this that lets the Python continue running?

If it's on the same command stream, you shouldn't have to block the Python on the copyin finishing. I'm okay with merging this if it's the same behavior as before though.

wozeparrot: went back to `wait_for`. seems like the issue that this was trying to fix was actually the shared memory buffer getting raced.

PR #1520, opened 2023-08-11
Transformer: use Tensor.scaled_dot_product_attention

jla524:


geohot:
Is there any test for the correctness of Transformer?

terafo:
`test_efficientnet` tests ViT, which imports `TransfomerBlock` from `Transformer`. So it gets tested as a standard part of CI.

PR #1526, opened 2023-08-11
Dilated Attention (LongNet)

fabianxvogt:
I wanna try to implement the [LongNet](https://arxiv.org/abs/2307.02486) in Tinygrad. 

Code is not functional at all. It's just to show the Idea. Next I will start with implementing tests.
The plan is basically to copy the existing transformer model in Tinygrad and add the additional parameters for the dilation.

I'd love to hear more opinions on this! Let me know if you think this is a bad idea for some reason 😄


wozeparrot:
This should be something that is another repo, not in the main tinygrad repo

Issue #1524, opened 2023-08-11
Tensor.__eq__() with two bool tensors raises error on Torch backend


geohotstan:
This was introduced from #1493

To reproduce:
```
In [24]: (Tensor([1], dtype=dtypes.bool, device="TORCH") == Tensor([1], dtype=dtypes.bool, device="TORCH")).realize()
RuntimeError: Subtraction, the `-` operator, with a bool tensor is not supported. If you are trying to invert a mask, use the `~` or `logical_not()` operator instead.
```
RuntimeError is from pytorch

PR #1528, opened 2023-08-11
Bert: use Tensor.scaled_dot_product_attention

JakobSpahn:
Changes made:

- compute attention with ```Tensor.scaled_dot_product_attention```
- add tests ```test/models/test_bert.py```

### DIFF
--- a/models/bert.py
+++ b/models/bert.py
@@ -147,13 +147,8 @@ class BertSelfAttention:
     key_layer = self.transpose_for_scores(mixed_key_layer)
     value_layer = self.transpose_for_scores(mixed_value_layer)
 
-    attention_scores = query_layer @ key_layer.transpose(2, 3)
-    attention_scores = attention_scores / self.attention_head_size**0.5
-    attention_scores = attention_scores + attention_mask
-    attention_probs = attention_scores.softmax()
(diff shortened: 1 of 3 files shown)


JakobSpahn:
Exclude or include from CI? Because it will download the weights.

wozeparrot:
`test_rnnt` doesn't actually test the whole model, only the lstm part. Probably make it an external test like yolo and yolov8 are because of the large amount of dependencies that need to be pulled in.

JakobSpahn:
This would be a possible version without downloading weights and tokenizer. Don't know why CI doesn't pass on LLVM and GPU (they pass on CLANG), on my machine they pass.

wozeparrot:
running out of memory, can you run a much smaller version?

JakobSpahn:
sorry, forgot to make model smaller. Will test on my fork until it passes and then commit working config here.

JakobSpahn:
```
TORCH=1 python -c "from tinygrad.tensor import Tensor; from tinygrad.helpers import dtypes; (1-Tensor(True, dtype=dtypes.bool)).realize()"
``` 
fails, so skip ```TORCH```

wozeparrot:
Yea, thats known: #1524

geohot:
Thanks!

PR #1531, opened 2023-08-12
Use /run/shm for clang transients

waifairer:


tinygrad/runtime/ops_clang.py
```diff
@@ -14,12 +14,10 @@
 class ClangProgram:
   def __init__(self, name:str, prg:str):
     prg = CLANG_PROGRAM_HEADER + prg
-    # TODO: is there a way to not write this to disk?
-    fn = f"{tempfile.gettempdir()}/clang_{hashlib.md5(prg.encode('utf-8')).hexdigest()}.{args['ext']}"
+    # /run/shm is a in-mem tmpfs, so this is fast
+    fn = f"/run/shm/clang_{hashlib.md5(prg.encode('utf-8')).hexdigest()}.{args['ext']}"
```
wozeparrot, line 18: is `/run/shm` standard? it doesn't exist on my system, but `/dev/shm` does

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_clang.py              3               5
```

geohot:
This is not the sort of fix I want for this, and it completely ignores the race condition. `/tmp` is also a ramdisk.

waifairer:
Understood. I believe that solving this without a filesystem call involves extending our own version of `dlopen`. But I will spend time understanding the race condition better before providing another potential solution. @geohot

tinygrad/runtime/ops_clang.py
waifairer: Standards changed but I'm not sure which version it happened.

waifairer: https://askubuntu.com/questions/169495/what-are-run-lock-and-run-shm-used-for

PR #1532, opened 2023-08-12
perf: floordiv fast path first, only last SumNode can be NumNode

roelofvandijk:
* restore fast path for SumNode.__floordiv__
* Only last nodes in SumNodes can be NumNodes (this should be more explicit)

@chenyuxyz - feel free to incorporate this in a follow-up MR if so desired.

### DIFF
--- a/tinygrad/shape/symbolic.py
+++ b/tinygrad/shape/symbolic.py
@@ -206,9 +206,10 @@ class SumNode(RedNode):
   def __floordiv__(self, b: Union[Node, int], factoring_allowed=True):
     fully_divided: List[Node] = []
     rest: List[Node] = []
+    if b == 1: return self
     if isinstance(b, SumNode):
-      nu_num = sum(node.b for node in self.flat_components if node.__class__ is NumNode)
-      de_num = sum(node.b for node in b.flat_components if node.__class__ is NumNode)
+      nu_num = self.flat_components[-1].b if self.flat_components[-1].__class__ is NumNode else 0  # SumNode groups NumNodes in last element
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                 5               6
```

chenyuxyz:
I see interesting. Is `flat_components` O(N^2) of length of flat components? It looks like each `+=` in `flat_components` call `__add__`, which calls `__sum__`. Probably can simplify by expanding all the components first then do the sum?

geohot:
Needs tests that use to pass, now fail, and will pass again.

roelofvandijk:
They don't exist, this is purely for perfomance, not correctness.

I can add timings later if you'd like, e.g. starting with the div fastpath (/1=self) hits 25% of the time for the llama test.

roelofvandijk:
> I see interesting. Is `flat_components` O(N^2) of length of flat components? It looks like each `+=` in `flat_components` call `__add__`, which calls `__sum__`. Probably can simplify by expanding all the components first then do the sum?
> 
> oh never mind I misunderstood, it's `+=` for a list. Then why is it guaranteed only the last one will be a `NumNode`? Is it because we won't have a `SumNode` with a `SumNode` component?

Because the SumNode creation appends the NumNode last. It's a bit (too) implicit.

geohot:
Actually read what this is, I don't trust relying on this behavior! SumNode today might put the NumNode in the last element, but what if it changes! A small amount of speed is never worth a potential bug.

PR #1533, opened 2023-08-13
Handle variable values in GlobalCounters instead of Variable.val

chenyuxyz:
part of #1353

It turns out that stateful `Variable` is a bad idea because all the shapetracker cached functions either returned the cached symbolic shape with old values, or if we try to update the `Variable.key` to include `val`, the cache become less effective. The new design use a global state to store variable values.

Added `sym_infer` which is used in reshape size check, and also later when we need to infer the real global_size, local_size, and op_estimate of kernel call.

Removed old `LazyBuffer.get_variable_buffers`, which is also a bad design. I have implemented passing `int` directly into kernel function for CLANG, GPU, METAL in the main PR.

### DIFF
--- a/tinygrad/shape/shapetracker.py
+++ b/tinygrad/shape/shapetracker.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 from enum import Enum, auto
 import functools
 from typing import Dict, Tuple, Union, List, Optional, Callable, cast, NamedTuple
-from tinygrad.helpers import prod, DEBUG
-from tinygrad.shape.symbolic import Variable, MulNode, NumNode, Node, SumNode, is_sym_int
+from tinygrad.helpers import prod, DEBUG, GlobalCounters
+from tinygrad.shape.symbolic import Variable, MulNode, NumNode, Node, SumNode, is_sym_int, sym_infer
 
 # these ops live here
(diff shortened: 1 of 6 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        1               0
tinygrad/lazy.py                           1               3
tinygrad/shape/shapetracker.py             9              11
tinygrad/shape/symbolic.py                 8               3
total                                     19              17
```

tinygrad/shape/shapetracker.py
```diff
@@ -3,8 +3,8 @@
 from enum import Enum, auto
 import functools
 from typing import Dict, Tuple, Union, List, Optional, Callable, cast, NamedTuple
-from tinygrad.helpers import prod, DEBUG
-from tinygrad.shape.symbolic import Variable, MulNode, NumNode, Node, SumNode, is_sym_int
+from tinygrad.helpers import prod, DEBUG, GlobalCounters
```
geohot, line 6: Definitely no GlobalCounters here.

chenyuxyz: ok. closing this. will put symbols somewhere in the middle

PR #1534, opened 2023-08-13
examples: numpy() array returns only one value, not an array

okias:
- Python 3.11.4
 - numpy 1.24.2

Fixes issue:

```
    loss_cpu = loss.detach().numpy()[0]
               ~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
```


airlied:
I've had same patch in my local tree and it seems fine.

wozeparrot:
why are the file permissions changing?

okias:
@wozeparrot oh, that was not intentional to add into the PR, but the script should have `+x` anyway, right?

PR #1535, opened 2023-08-13
Train Unet3D

rkstgr:
Previously meant as a continuation of #1469. Now its a clean re-implementation of [reference](https://github.com/mlcommons/training/tree/master/image_segmentation/pytorch)

## TODO
- [x] losses
- [x] evaluate
- [x] dataloader (based on torch)
- [x] model (incl. conv3d and conv_transpose3d)
- [x] training
- [x] data preprocessing

## Original comment

MLPerf uses slices of size 128x128x128 and batch size 2, which does not fit on the largest GPU (48G) I have currently access to, although the code runs with 64x64x128, float32, batch size 1 on the NVIDIA A40.
Batch size is not a hard requirement as we could use grad accumulation but would be slower. It seems we would need at least 80G in GPU memory (will see if I can rent some to test it).

I tried to train it in half precision, but nvcc always threw an error:
> error: ambiguous "?" operation: second operand of type const half can be converted to third operand of type float

This happens in all lines that are rendered by Memcopy ops as conditionals.
```c
__global__ void r_1024_16_2_8_16_3_3_3_4_4(half* data0, const half* data1, const half* data2) {
{ int gidx0 = blockIdx.z; /* 1024 */.
{ int gidx1 = blockIdx.y; /* 16 */
{ int gidx2 = blockIdx.x; /* 2 */
  { int lidx3 = threadIdx.y; /* 8 */
  { int lidx4 = threadIdx.x; /* 16 */
    float4 acc0_0 = make_float4(0.0,0.0,0.0,0.0);
    float4 acc0_1 = make_float4(0.0,0.0,0.0,0.0);
    float4 acc0_2 = make_float4(0.0,0.0,0.0,0.0);
    float4 acc0_3 = make_float4(0.0,0.0,0.0,0.0);
    for (int ridx5 = 0; ridx5 <= 2; ++ridx5) { do
      float val1_0 = (<bound checks>)?(data1[<some index>]):f0.0; // <----
      ....
```

Any suggestions on how to get it to train in half-precision?


tinyb0t:
Changes made in `tinygrad/`:
```
```

g1y5x3:
I don't think the current cuda backend supports fp16. In fact, most of the backends don't support fp16, yet. It's in the working.

rkstgr:
To give more context: It fails during training, specifically in the backward step while realizing the gradients:
```python
# optim.py, line 35
g = t.grad.realize() + self.wd * t.detach()
```
I would count that as math, so its currently not expected to work?

geohot:
It all should work, it just does the math in fp32, but the weights can be fp16.

The CUDA backend isn't super well tested, try GPU=1 instead.

rkstgr:
| Weights | Input | Outcome |
|---|---|---|
| float32 | float32 | ✅ |
| float16 | x | 🚫 |

## Error
```
  File "/dev/tinygrad/examples/mlperf/model_train.py", line 74, in <module>
    globals()[nm]()
  File "/dev/tinygrad/examples/mlperf/model_train.py", line 39, in train_unet3d
    opt.step()
  File "/dev/tinygrad/tinygrad/nn/optim.py", line 40, in step
    self.realize(self.b)
  File "/dev/tinygrad/tinygrad/nn/optim.py", line 23, in realize
    p.realize()
  File "/dev/tinygrad/tinygrad/tensor.py", line 94, in realize
    self.lazydata.realize()
  File "/dev/tinygrad/tinygrad/lazy.py", line 162, in realize
    assert self.realized.dtype == self.dtype or self.dtype.__class__ is ImageDType, f"dtype mismatch on realize got {self.realized.dtype} expected {self.dtype}"
AssertionError: dtype mismatch on realize got dtypes.half expected dtypes.float
```

## Reason:
The gradients are calculated in float32, but the weights are still in float16 (see comments in code).
```
def step(self) -> None:
    for i, t in enumerate(self.params):
      assert t.grad is not None

      # g is float32 because of t.grad
      g = t.grad.realize() + self.wd * t.detach() 
      if self.momentum:
        self.b[i].assign(self.momentum * self.b[i] + g).realize()  # NOTE: self.b[i] is zero on the first run, no if required
        g = (g + self.momentum * self.b[i]) if self.nesterov else self.b[i]

      # t is still dtype half, but is now assigned with lazydata of dtype float32, it does not fail immediately but will in the next step when the tensor gets realized
      t.assign(t.detach() - g * self.lr) 
    self.realize(self.b)
```

## How to fix this? Ideas:
1. grads should always be of the same dtype as the original data
2. we explicitly cast the grads to the same dtype as the original data in the backwards step
3. ... any other ideas?

rkstgr:
This trains now with the float16 weights but it still fails with the full image size of 128x128x128. In the debug traces you see that still most of the computation is done in float32. 

I can't calculate the dice_loss in float16 because of overflows (for float16 max value already at 65504) as we would have to sum over 128^3=2,097,152 values. Idk if this is one of the reason why most of the computation falls back to float32.

rkstgr:
Ok there seems to be a bug in the training code / model.

The reference implementation in pytorch ([repo](https://github.com/mlcommons/training/tree/master/image_segmentation/pytorch)) runs without problems using around 10GB of GPU memory with 1.5sec per batch (batch size 2, float32, size 128^3). Also the loss starts around 1.15.

wozeparrot:
How's this coming along?

rkstgr:
Everything needed for training and evaluation is implemented. 

For original model and input sizes got OOM on the backwards pass. If tinygrad has gotten more performant/ less memory intensive this should be good to go, but haven't tested it in a while.

wozeparrot:
There has been some JIT memory optimizations, so if the model is fully jitted it should help.

PR #1539, opened 2023-08-14
[WIP] ast simplification rules

nimlgen:
This is mostly a PoC MR of simplification rules. I would greatly appreciate receiving feedback on the design aspect.

As an example, consider the optimization rule to simplify `x + a - a` to just `x`. This rule is defined as follows: `OptRule(0, lambda capt: capt['x'], [(1, BinaryOps.ADD, ['x','a']), (2, BinaryOps.SUB, [1,'a'])])`. Each rule comprises multiple rule nodes, such as `(2, BinaryOps.SUB, [1,'a'])`, which are used to structure a subtree.

The node `(1, BinaryOps.ADD, ['x','a'])` can be only matched with a BinaryOps.ADD LazyOp and is assigned an identifier of `1`. This identifier is used to pass this node as a child `(2, BinaryOps.SUB, [1,'a'])` (see [**1**,'a']). The argument `a` passed to both nodes ensures that the same LazyBuffer is used in both LazyOps. `lambda capt: capt['x']` generates the output LazyOp. In this case it just `x` which was captured while tree traversing.

Maybe need to write a dev doc to document rules better?


nimlgen:
Closing for now. Will back to it a bit later.

PR #1546, opened 2023-08-15
New example: adversarial attacks on small CNN

qlero:
Hello,

This is my first commit to TinyGrad: a new entry to the list of example applications. The code:

1. Trains a simple CNN network on MNIST
2. Generates a small batch (128 test set elements) of adversarially-attacked\* inputs

\* Adversarial attacks are purposefully-crafted perturbations to a neural network's inputs such that they are misclassified. 

The example fixes the weights of a pre-trained model and generates a buffer between the inputs' Tensor and the model. The buffer will accumulate adversarial perturbations by computing the additive inverse of the gradients of each input image's pixels with respect to the network's output loss (hill climbing exercise).

Implemented attacks: L2 and Linf FGSM, R+FGSM, PGD attacks [1]

A stand-alone implementation of this example, tested on the current master version of Tinygrad, can be found [here](https://github.com/qlero/tiny_adversary) along with output image examples.

[1] [Adversarial Attacks and Defences: A Survey](https://arxiv.org/abs/1810.00069)

### DIFF
--- a/examples/adversarial_attacks.py
+++ b/examples/adversarial_attacks.py
@@ -1,0 +1,374 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+from extra.datasets import fetch_mnist
+from tinygrad.nn import Linear, Conv2d
+from tinygrad.nn.optim import SGD
+from tinygrad.state import get_parameters
+from tinygrad.tensor import Tensor
+
+###########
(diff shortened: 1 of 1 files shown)


wozeparrot:
This should be something in another repo and not in the main tinygrad repo.

PR #1552, opened 2023-08-16
symbolic codegen and exec

chenyuxyz:
part of #1353 , codegen and exec to implement `realize` for symbolic inputs.

The combined `var_vals` are passed into kernel function directly. I have implemented the backend for CLANG, GPU, METAL. `global_size`, `local_size` and `op_estimate ` are computed to int during exec. In addition to CI, I also tested with `DEBUG=4 python -m pytest -rA test/test_symbolic_ops.py` to make sure all debugging info renders with symbols.

Most of the hand coded optimization works, and I need to disable the upcast ones because we cannot upcast a symbolic axis.

Also added some test cases for 2 variables. Want to make it as general as possible.

These two examples show that we need to support symbolic for loop max and symbolic offset

kernel for matmul `(3, vi) @ (vi, 5)`
```
#include <metal_stdlib>
using namespace metal;
kernel void r_5_3_s0(device float* data0, const device float* data1, const device float* data2, constant int& i, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
{ int gidx0 = gid.x;  /* 5 */
  { int lidx1 = lid.x;  /* 3 */
    float acc0_0 = 0.0f;
    for (int ridx2 = 0; ridx2 <= (-1+i); ++ridx2) {
      float val1_0 = *(data1+(lidx1*i)+ridx2);
      float val2_0 = *(data2+(ridx2*5)+gidx0);
    } /* reduce */
    *(data0+(lidx1*5)+gidx0) = acc0_0;
  }} /* global+local */
}
```

kernel for `(3, vi).cat((3, vj), dim=1)`
```
#include <metal_stdlib>
using namespace metal;
kernel void E_s0_3n14(device float* data0, const device float* data1, const device float* data2, constant int& i, constant int& j, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
{ int gidx0 = gid.x;  /* <(i[1-10]+j[1-10])> */
  { int lidx1 = lid.x;  /* 3 */
    float val1_0 = ((gidx0<i))?(*(data1+(lidx1*i)+gidx0)):0.0f;
    float val2_0 = (((gidx0*-1)<((i*-1)+1)))?(*(data2+(i*-1)+(lidx1*j)+gidx0)):0.0f;
    float alu0 = (val1_0+val2_0);
    *(data0+(lidx1*(i+j))+gidx0) = alu0;
  }} /* global+local */
}
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -188,6 +188,8 @@ jobs:
       run: DEBUG=2 METAL=1 python -m pytest -n=auto test/test_ops.py
     - name: Run JIT test
       run: DEBUG=2 METAL=1 python -m pytest -n=auto test/test_jit.py
+    - name: Run symbolic shapetracker test
+      run: METAL=1 python -m pytest -n=auto test/test_symbolic_shapetracker.py test/test_symbolic_ops.py
     - name: Check Device.DEFAULT
       run: WEBGPU=1 python -c "from tinygrad.lazy import Device; assert Device.DEFAULT == 'WEBGPU', Device.DEFAULT"
     #- name: Run webgpu pytest
@@ -263,10 +265,10 @@ jobs:
         run: python -c "from tinygrad.lazy import Device; assert Device.DEFAULT in ['LLVM','CLANG','CUDA','GPU'], Device.DEFAULT"
       - name: Run pytest (not cuda)
         if: matrix.backend!='cuda'
-        run: python -m pytest -n=auto test/ -k '${{matrix.backend=='llvm'&&'not (test_nn.py and test_conv_transpose2d)'||'test'}}' -m 'not exclude_${{matrix.backend}}'
+        run: CI=1 python -m pytest -n=auto test/ -k '${{matrix.backend=='llvm'&&'not (test_nn.py and test_conv_transpose2d)'||'test'}}' -m 'not exclude_${{matrix.backend}}'
       - name: Run pytest (cuda)
         if: matrix.backend=='cuda'
-        run: python -m pytest -n=auto test/ -k 'not (half or test_efficientnet_safetensors) and not (test_conv2d and test_tensor.py)' -m 'not exclude_cuda' --ignore=test/external --ignore=test/models
+        run: CI=1 python -m pytest -n=auto test/ -k 'not (half or test_efficientnet_safetensors) and not (test_conv2d and test_tensor.py)' -m 'not exclude_cuda' --ignore=test/external --ignore=test/models
 
   testunicorn:
     name: ARM64 unicorn Test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -265,10 +265,10 @@ jobs:
         run: python -c "from tinygrad.lazy import Device; assert Device.DEFAULT in ['LLVM','CLANG','CUDA','GPU'], Device.DEFAULT"
       - name: Run pytest (not cuda)
         if: matrix.backend!='cuda'
-        run: CI=1 python -m pytest -n=auto test/ -k '${{matrix.backend=='llvm'&&'not (test_nn.py and test_conv_transpose2d)'||'test'}}' -m 'not exclude_${{matrix.backend}}'
+        run: python -m pytest -n=auto test/ -k '${{matrix.backend=='llvm'&&'not (test_nn.py and test_conv_transpose2d)'||'test'}}' -m 'not exclude_${{matrix.backend}}'
       - name: Run pytest (cuda)
         if: matrix.backend=='cuda'
-        run: CI=1 python -m pytest -n=auto test/ -k 'not (half or test_efficientnet_safetensors) and not (test_conv2d and test_tensor.py)' -m 'not exclude_cuda' --ignore=test/external --ignore=test/models
+        run: python -m pytest -n=auto test/ -k 'not (half or test_efficientnet_safetensors) and not (test_conv2d and test_tensor.py)' -m 'not exclude_cuda' --ignore=test/external --ignore=test/models
 
   testunicorn:
     name: ARM64 unicorn Test
--- a/test/test_helpers.py
+++ b/test/test_helpers.py
@@ -106,7 +106,7 @@ with Context(VARIABLE=1):
       ...
     assert D.value == 2, f"Expected D to be 2, but was {D.value}. Indicates that Context.__exit__ did not restore to the correct value."
 
-class TestMergeDcits(unittest.TestCase):
+class TestMergeDicts(unittest.TestCase):
   def test_merge_dicts(self):
     a = {"a": 1, "b": 2}
     b = {"a": 1, "c": 3}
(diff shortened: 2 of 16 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py             6               3
tinygrad/codegen/optimizer.py              4               4
tinygrad/helpers.py                        4               0
tinygrad/ops.py                           21              14
tinygrad/renderer/cstyle.py                6               4
tinygrad/renderer/wgsl.py                  1               1
tinygrad/runtime/ops_clang.py              2               2
tinygrad/runtime/ops_gpu.py                2               2
tinygrad/runtime/ops_metal.py              6               3
tinygrad/shape/symbolic.py                 9               0
total                                     64              36
lines added in the tinygrad folder: 28
```

.github/workflows/test.yml
```diff
@@ -263,10 +265,10 @@ jobs:
         run: python -c "from tinygrad.lazy import Device; assert Device.DEFAULT in ['LLVM','CLANG','CUDA','GPU'], Device.DEFAULT"
       - name: Run pytest (not cuda)
         if: matrix.backend!='cuda'
-        run: python -m pytest -n=auto test/ -k '${{matrix.backend=='llvm'&&'not (test_nn.py and test_conv_transpose2d)'||'test'}}' -m 'not exclude_${{matrix.backend}}'
+        run: CI=1 python -m pytest -n=auto test/ -k '${{matrix.backend=='llvm'&&'not (test_nn.py and test_conv_transpose2d)'||'test'}}' -m 'not exclude_${{matrix.backend}}'
```
geohot, line 268: You shouldn't need this, CI should be set by CI.

test/test_helpers.py
```diff
@@ -106,5 +106,17 @@ def test_context_exit_reverts_updated_values(self):
       ...
     assert D.value == 2, f"Expected D to be 2, but was {D.value}. Indicates that Context.__exit__ did not restore to the correct value."
 
+class TestMergeDcits(unittest.TestCase):
```
geohot, line 109: typo

tinygrad/helpers.py
```diff
@@ -22,6 +22,10 @@ def make_pair(x:Union[int, Tuple[int, ...]], cnt=2) -> Tuple[int, ...]: return (
 def flatten(l:Iterator): return [item for sublist in l for item in sublist]
 def mnum(i) -> str: return str(i) if i >= 0 else f"m{-i}"
 def fromimport(mod, frm): return getattr(__import__(mod, fromlist=[frm]), frm)
+def merge_dicts(ds):
```
geohot, line 25: Is there a reason you can't use `dict1.update(dict2)` to merge dicts?

geohot:
Why is it `constant int&` and not just `int`?

test/test_helpers.py
chenyuxyz: fixed

tinygrad/helpers.py
chenyuxyz: `update` does not check if value conflicts, added a test case that `(3, vi) @ (vi, 5)` should fail if they hold different values for `vi`

chenyuxyz:
`int` does not work
```
E     AssertionError: Error Domain=MTLLibraryErrorDomain Code=3 "program_source:3:98: error: invalid type 'int' for input declaration in a kernel function
E     kernel void r_s0_s3_3(device float* data0, const device float* data1, const device float* data2, int i, int j, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
E                                                                                                      ^~~~~
E     program_source:3:105: error: invalid type 'int' for input declaration in a kernel function
```

I took this from tracing how [PyFR](https://github.com/PyFR/PyFR/blob/8edbf1878437ab43f2787b836f013f7f73a8a27f/pyfr/backends/metal/provider.py#L110) implemented it for metal.

My understanding from the [metal spec](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) section 4 is that kernel argument needs an address space attribute, and it only accepts pointer or reference

geohot:
So this isn't true for OpenCL...but maybe it's fine.

tinygrad/shape/shapetracker.py
```diff
@@ -162,7 +162,7 @@ def real_strides(self, ignore_valid=False) -> Tuple[Optional[Union[Node, int]],
     idx, valid = self.expr_idxs(idxs)
     ret: List[Optional[Union[Node, int]]] = [None] * len(self.views[-1].shape)
     for this_dim in (idx.nodes if isinstance(idx, SumNode) else [idx]):
-      if isinstance(this_dim, MulNode) and isinstance(this_dim.a, Variable):
+      if isinstance(this_dim, MulNode) and isinstance(this_dim.a, Variable) and str(this_dim.a.expr).startswith("idx"):
```
geohot, line 165: This looks sketchy using a string compare. Is there a better way to write this?

tinygrad/renderer/cstyle.py
```diff
@@ -69,15 +70,16 @@ def render_load(self, output_dtype, buf_name, buf_dtype, idx, local=False) -> st
...
+  def render_for(self, expr: str, _min:int, _max:Union[int,str]) -> str:
     return f"for (int {expr} = {_min}; {expr} <= {_max}; ++{expr}) {{"
 
   def render_conditional(self, cond: str, x:str, y:str) -> str:
     return f"({cond})?({x}):{y}"
 
   def render_kernel(self, function_name:str, kernel:List[str], bufs:List[Tuple[str,DType]], global_size:List[int], local_size:List[int], prekernel:List[str]) -> Tuple[str,List[int],List[int]]:
     tmp = "const sampler_t smp = CLK_NORMALIZED_COORDS_FALSE | CLK_ADDRESS_CLAMP | CLK_FILTER_NEAREST;\n" if any(isinstance(dtype, ImageDType) for _,dtype in bufs) else ""
-    buftypes = [(name,f"{'read_only' if i > 0 else 'write_only'} image2d_t" if dtype.name.startswith('image') else
+    buftypes = [(name[8:], self.arg_int_prefix) if name.startswith("ARG_INT_") else
```
geohot, line 81: String stuff is always sketchy, is there a better way to do this?

tinygrad/helpers.py
geohot: Can you add a type signature to this function?

geohot:
It's close, but I don't like the string compare stuff, particularly the one in shapetracker.

tinygrad/shape/shapetracker.py
chenyuxyz: good call - updated

chenyuxyz:
re: PyOpenCL and `set_scalar_arg_dtypes`. I suspect under the hood it's doing similar thing even with `np.int32`. I profiled the llama code in the main PR and the profiler indicated that a lot of time were on scalar arg related internal function. The generation wall time is lower so I am not sure if it's PyOpenCL issue or something else

tinygrad/renderer/cstyle.py
chenyuxyz: added an internal dtype for int argument and used that instead

geohot:
Cool, merged!

PR #1553, opened 2023-08-16
Change Any to List[Optional[RawBuffer]] in JIT

sb-:
Addressed the TODO in  [tinygrad/jit.py](https://github.com/tinygrad/tinygrad/blob/master/tinygrad/jit.py#L15)
`self.jit_cache: List[Tuple[Callable, Any]] = []  # TODO: Any should be List[RawBuffer], but this fails`

This should be `Tuple[Callable, List[Optional[RawBuffer]]]]` since elements can be None. The fix also requires `# type: ignore[no-redef]` on two lines since mypy fails to infer the correct type when `args` is redefined.

### DIFF
--- a/tinygrad/jit.py
+++ b/tinygrad/jit.py
@@ -41,10 +41,10 @@ class TinyJit:
       if DEBUG >= 1: print(f"JIT captured {len(self.jit_cache)} kernels with {len(input_rawbuffers)} inputs")
 
       # get the inputs for replacement
-      for j,(prg,pargs) in enumerate(self.jit_cache): # type: Tuple[int, Tuple[Callable, List[Optional[RawBuffer]]]]
-        for i,a in enumerate(pargs):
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/jit.py                            5               4
```

geohot:
Hmm I don't understand why the`type: ignore[no-redef]` is needed

sb-:
Yes, mypy won't be able to infer correctly if it's named arg. See [docs](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-each-name-is-defined-once-no-redef) an example:

`tinygrad/jit.py:32: error: Name "args" already defined on line 22  [no-redef]
Found 1 error in 1 file (checked 1 source file)
`

I can just rename them to `pargs` and remove `ignore[no-redef]` if you prefer

geohot:
Rename is fine, but there's still a `# type: ignore[no-redef]`

PR #1554, opened 2023-08-16
Fix uops int32 for llvm

madt2709:
* Adds int32 ops to LLVM 
* Need to do some more digging as to why mul_bool doesn't work


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/renderer/llvmir.py                9               8
```

PR #1564, opened 2023-08-18
more symbolic symbolic ops

chenyuxyz:
This was not needed for #1353 but makes the symbolic x symbolic more general. I dropped a `realize` during `llama.py` cleanup and ended up make these changes. We will have the `realize` back to llama but I think it's good to make these operation more future proof.

It's mostly about `NumNode` ops with other `Node`. `NumNode(1) * Node` was a `MulNode` and now simplified to a `Node`. Also support `NumNode // NumNode` etc.

Also now `Node.__mod__` and `Node.__floordiv__` use the same set of checks. Tests were updated to reflect that.

Lastly, `SumNode // SumNode` defers the remainder check to another `//`. Previously we only support the case that remainder is 0.

### DIFF
--- a/test/test_symbolic_shapetracker.py
+++ b/test/test_symbolic_shapetracker.py
@@ -139,7 +139,7 @@ class TestSymbolicShapeExpr(unittest.TestCase):
     view = View(shape, strides)
     st = ShapeTracker(shape, [view])
     idx, valid = st.expr_idxs(idx)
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                13               7
```

PR #1545, opened 2023-08-15
[WIP] Do math in same type as buffer dtype

kunwar31:
changes to do math in same dtype as buffer dtype, and upcast to highest priority dtype when operands are of different dtypes. 
on 7900XTX this gives 

```
GPU=1 HALF=0 BIG=1 python -W ignore test/test_speed_v_torch.py
add                             8192x 8192   70.45 ms (    0.95 GFLOPS    11.43 GB/s) in torch,    1.31 ms (   51.39 GFLOPS   616.67 GB/s) in tinygrad,    0.02x faster      67.11 MOPS   805.31 MB
exp                             8192x 8192  110.82 ms (    1.21 GFLOPS     4.84 GB/s) in torch,    0.82 ms (  163.48 GFLOPS   653.93 GB/s) in tinygrad,    0.01x faster     134.22 MOPS   536.87 MB
gemm                            2048x 2048  162.96 ms (  105.42 GFLOPS     0.31 GB/s) in torch,    2.28 ms ( 7528.15 GFLOPS    22.06 GB/s) in tinygrad,    0.01x faster   17179.87 MOPS    50.33 MB
gemm                            4096x 4096 1289.31 ms (  106.60 GFLOPS     0.16 GB/s) in torch,   17.32 ms ( 7936.21 GFLOPS    11.63 GB/s) in tinygrad,    0.01x faster  137438.95 MOPS   201.33 MB
conv bs: 32 chans:128 -> 128 k:1            367.02 ms (   46.81 GFLOPS     1.46 GB/s) in torch,    1.27 ms (13489.88 GFLOPS   421.61 GB/s) in tinygrad,    0.00x faster   17179.87 MOPS   536.94 MB
conv bs:  4 chans:128 -> 128 k:3            179.92 ms (  107.42 GFLOPS     0.38 GB/s) in torch,    1.21 ms (15982.02 GFLOPS    56.85 GB/s) in tinygrad,    0.01x faster   19327.35 MOPS    68.76 MB
conv bs:  4 chans:128 -> 128 k:5            452.53 ms (  114.96 GFLOPS     0.15 GB/s) in torch,    6.97 ms ( 7468.98 GFLOPS     9.87 GB/s) in tinygrad,    0.02x faster   52022.48 MOPS    68.76 MB
```

```
GPU=1 HALF=1 BIG=1 python -W ignore test/test_speed_v_torch.py
[...]

PR #1566, opened 2023-08-18
cleanup ops_gpu

wozeparrot:
move stuff that doesn't need to be in `post_init` into `__init__`.

this lets us get a total device count before the devices are actually initialized. Needed for #1529.

this also lets us get the device names before the devices are initialized. Needed for #1545.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_gpu.py                7               4
```

kunwar31:
Currently, length of platform_devices doesn't match number of platforms. this becomes a problem as CL_PLATFORM env var is used to both select the platform_device and platform itself
image

one solution can be to change platform devices like 
```
cpu_devices = [x.get_devices(device_type=cl.device_type.CPU) for x in cl_platforms]
gpu_devices = [x.get_devices(device_type=cl.device_type.GPU) for x in cl_platforms]
platform_devices: List[List[cl.Device]] = [cpu_devices[i]+gpu_devices[i] for i in range(len(cl_platforms))]
```
so it correctly selects all devices of CL_PLATFORM

PR #1570, opened 2023-08-18
fix usage of arm64 regs according to CC

nimlgen:
Some int registers and fp registers were overwritten being caller-save. Made the generated asm to follow arm64 calling convention.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/assembly_arm64.py         6               6
```

geohot:
Is there a test that could catch this regression?

Issue #1555, opened 2023-08-16
[obscure] ARM64 bug when creating Tensor from np.ndarray with dtype int64


geohotstan:
```
yay = np.ones(shape=(3,1), dtype=np.int32)
uhoh = np.ones(shape=(3,1), dtype=np.int64)
t1= Tensor(yay, dtype=dtypes.int32, requires_grad=False)
t2 = Tensor(uhoh, dtype=dtypes.int32, requires_grad=False)
a = Tensor.arange(4, dtype=dtypes.int32, requires_grad=False).reshape(1,4)
np.testing.assert_allclose((t1==a).numpy(), (t2==a).numpy())
```

Running the above code with ARM64=1 CLANG=1 will produce either 
```
Python(56005,0x2056d5e00) malloc: Incorrect checksum for freed object 0x11fea4a00: probably modified after being freed.
Corrupt value: 0x1
Python(56005,0x2056d5e00) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```
or `Segmentation fault: 11` or return the wrong value for `(t2==a).numpy()`

The error occurs after `Uop.CMPLT` is realized and only when the dtype of the np.ndarray used to create the Tensor is int64.

PR #1572, opened 2023-08-18
for issue #1555, int64 and int8 in CI=1 ARM64=1 CLANG=1

tpat7187:
broadcasting using CI=1 ARM64=1 CLANG=1 was incorrect since the stored values were shifted depending on dtype. When the assembler copied the tensor into the broadcasted array it would do so incorrectly because of this shift. This error would only occur when the broadcasted array shape was a multiple of 4. By default int32 was shifting by #0 so I removed the shift condition and all the tests passing again.

these are the tests I ran below:
```
CI=1 ARM64=1 CLANG=1 python -m pytest -n=auto test/ -k 'not (test_nn.py and (test_conv_transpose2d or test_conv2d))' --ignore=test/models --ignore=test/test_speed_v_torch.py --ignore=test/test_net_speed.py --ignore=test/test_specific_conv.py  --ignore=test/unit/test_disk_tensor.py
```


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/assembly_arm64.py         1               1
```

PR #1576, opened 2023-08-19
Argmax/Argmin Feature

ZenginU:
Image tests fails when test size is (10,20,30) ("test/test_ops.py::TestOps::test_argmax - pyopencl._cl.LogicError: clCreateImage failed: INVALID_IMAGE_SIZE") not sure why but with (10,20) passes.

Note: Torch selects first index for duplicate max, this version selects last index. Cant find a way to select the first occurence

@wozeparrot If this passes you can update the argmax in cifar, I cant update it as I cant run cifar at the moment.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         8               0
```

geohot:
Thanks!

Issue #1143, opened 2023-07-06
Switch nvcc -> cicc in CUDA backend


geohot:
Apparently it's faster to compile? Does someone want to try this and benchmark?

Replace `from pycuda.compiler import compile as cuda_compile` with something that uses cicc

sayantan3:
Did you mean to use pyopencl rather than pycuda?

dc-dc-dc:
I believe cuda_compile is just making nvcc command call, could just replace with a custom call to cicc

EliTheCoder:
CICC is part of NVCC. 

![image](https://github.com/geohot/tinygrad/assets/4386738/8c315200-eeb7-4615-a386-4a270d2a1e55)

sekstini:
Heya, dumping some info of possibly non-zero value:

cuda_compile calls nvcc like so: `nvcc --ptx -Wno-deprecated-gpu-targets -arch sm_89 -I/home/fluffy/mambaforge/envs/cuda121/lib/python3.10/site-packages/pycuda/cuda kernel.cu`.

So we can take a sample program:


```python
from tinygrad.tensor import Tensor
t = Tensor([1, 2, 3])
(t + t + t).realize()
```

This produces the following kernel.cu:

```c++
typedef unsigned char uchar;
typedef unsigned int uint;
typedef unsigned long ulong;
__global__ void E_3(float* data0, const float* data1) {
  { int lidx0 = threadIdx.x;  /* 3 */
    float val1_0 = data1[lidx0];
    float alu0 = (val1_0+val1_0);
    float alu1 = (alu0+val1_0);
    data0[lidx0] = alu1;
  } /* global+local */
}
```

If we run the above nvcc command with `--dryrun` we get the following:

```bash
#$ _NVVM_BRANCH_=nvvm
#$ _SPACE_= 
#$ _CUDART_=cudart
#$ _HERE_=/home/fluffy/mambaforge/envs/cuda121/bin
#$ _THERE_=/home/fluffy/mambaforge/envs/cuda121/bin
#$ _TARGET_SIZE_=
#$ _TARGET_DIR_=
#$ _TARGET_SIZE_=64
#$ TOP=/home/fluffy/mambaforge/envs/cuda121/bin/..
#$ NVVMIR_LIBRARY_DIR=/home/fluffy/mambaforge/envs/cuda121/bin/../nvvm/libdevice
[...]

PR #1579, opened 2023-08-20
NVCC to CICC

mebenstein:
#1143 

Replaces `cuda_compile` with a call to `cicc` for faster compilation time. 
85%+ faster without caching. To replicate:

```python
t = Tensor([1, 2, 3]).cuda()
(t + t + t).realize()
```

Tests:
- caching disabled:
  - `cuda_compile`: 326.1ms (+- 14.3ms)
  - `cicc`: 205.4ms (+- 6.3ms)
- with caching:
  - `cuda_compile`: 199.2ms (+- 19.5ms)
  - `cicc`: 195.7ms (+- 10.2ms)

The `CUDA_PROGRAM_HEADER` need to be maintained. Almost all relevant definitions can be generated via `nvcc -E -x cu - < /dev/null`


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/ops_cuda.py              13               3
```

tinygrad/runtime/ops_cuda.py
```diff
@@ -52,12 +52,22 @@ def __init__(self, size, dtype): super().__init__(size, dtype, cuda.mem_alloc(si
...
...}};uint3 __attribute__((device_builtin)) extern const threadIdx;uint3 __attribute__((device_builtin)) extern const blockIdx;dim3 __attribute__((device_builtin)) extern const blockDim;dim3 __attribute__((device_builtin)) extern const gridDim;int __attribute__((device_builtin)) extern const warpSize;struct __attribute__((device_builtin)) float1{float x;};struct __attribute__((device_builtin)) __attribute__((aligned(8))) float2 { float x; float y; };struct __attribute__((device_builtin)) float3{float x, y, z;};struct __attribute__((device_builtin)) __attribute__((aligned(16))) float4{float x, y, z, w;};static __inline__ __attribute__((host)) __attribute__((device)) float1 make_float1(float x){float1 t; t.x = x; return t;}static __inline__ __attribute__((host)) __attribute__((device)) float2 make_float2(float x, float y){float2 t; t.x = x; t.y = y; return t;}static __inline__ __attribute__((host)) __attribute__((device)) float3 make_float3(float x, float y, float z){float3 t; t.x = x; t.y = y; t.z = z; return t;}static __inline__ __attribute__((host)) __attribute__((device)) float4 make_float4(float x, float y, float z, float w){float4 t; t.x = x; t.y = y; t.z = z; t.w = w; return t;}'
```
geohot, line 55: Hmm, do we need all these things? What if we just don't use them?

mebenstein: We need all the variables and types that are used in the kernel, otherwise it won't compile. I did include some things we are currently not using for future compatibility, but we can remove it and shorten the header to: `CUDA_PROGRAM_HEADER = 'struct __attribute__((device_builtin)) uint3{unsigned int x, y, z;};struct __attribute__((device_builtin)) dim3{unsigned int x, y, z;__attribute__((host)) __attribute__((device)) constexpr operator uint3(void) const {return uint3{x,y,z};}};uint3 __attribute__((device_builtin)) extern const threadIdx;uint3 __attribute__((device_builtin)) extern const blockIdx;struct __attribute__((device_builtin)) __attribute__((aligned(16))) float4{float x, y, z, w;}; static __inline__ __attribute__((device)) float4 make_float4(float x, float y, float z, float w){float4 t; t.x = x; t.y = y; t.z = z; t.w = w; return t;}'`

geohot:
CI failing

dc-dc-dc:
@cloud11665 does gpuocelot provide a cicc shell command equivalent ?

cloud11665:
@dc-dc-dc, no, it doesn't anything like this. If you give me like an hour, I can also chip in and help with reviewing this

tinygrad/runtime/ops_cuda.py
wozeparrot: I do think it makes more sense to keep the header as small as possible, we can always add more to it later.

wozeparrot:
Any update on this?

tinygrad/runtime/ops_cuda.py
mebenstein: The snipped above is insufficent. The last commit I made includes the minimum number of definitions to pass CI.

geohot:
CUDA tests failing...though if I'm reading this right it's so much faster on the non failing tests

geohot:
Closing for now, but would love to see an updated version of this that passes the tests and is faster. Maybe comes with the compile refactor.

PR #1588, opened 2023-08-20
ci: testpaths flag for faster collection

roelofvandijk:
Speed up the collection step of pytest.


roelofvandijk:
Impact is within job-to-job standard deviation.

PR #1590, opened 2023-08-20
[READY] ci: reorder tests to start with the slowest

roelofvandijk:
Starting with the slowest tests should reduce the total test time, because it avoids the case where a single core is crunching on a slow test at the end of the `pytest` run.

[Before](https://github.com/tinygrad/tinygrad/actions/runs/5919062619/usage):

image

After:
image


geohot:
I don't think this is worth it, this will make tests take longer to fail on average

PR #1591, opened 2023-08-20
sparse cat cross entropy

g1y5x3:
Updated `docs/quickstart.md`, `examples/serious_mnist.py`, `extra/training.py` accordingly after adding the function into `tensor.py`
~~Also, I deleted `examples/hlb_cifar10_torch.py` since its not going to be further updated.~~

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -137,18 +137,18 @@ First we need to set the training flag in `Tensor`:
 Tensor.training = True
 ```
 
-For our loss function we will be using cross entropy loss.
+For our loss function we will be using sparse categorical cross entropy loss.
 
 ```python
-# from extra.training import sparse_categorical_crossentropy
-def cross_entropy(out, Y):
+# from tinygrad.tensor import sparse_categorical_crossentropy
+def sparse_categorical_crossentropy(out, Y, ignore_index=-1):
+  loss_mask = Y != ignore_index
   num_classes = out.shape[-1]
-  YY = Y.flatten().astype(np.int32)
-  y = np.zeros((YY.shape[0], num_classes), np.float32)
-  y[range(y.shape[0]),YY] = -1.0*num_classes
-  y = y.reshape(list(Y.shape)+[num_classes])
-  y = Tensor(y)
-  return out.mul(y).mean()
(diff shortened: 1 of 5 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         6               0
total                                      7               1
```

tinygrad/runtime/ops_torch.py
```diff
@@ -26,5 +26,5 @@ def __init__(self, size:int, dtype:DType, buf:Optional[torch.Tensor]=None): supe
   def fromCPU(cls, x):
     buf = torch.from_numpy(x).requires_grad_(False).to(device)
     return cls(prod(x.shape), type_map[buf.dtype], buf)
-  def toCPU(self): return self._buf.cpu().numpy()
+  def toCPU(self): return self._buf.numpy()
```
dc-dc-dc, line 29: I think cpu is still needed here in case the torch tensor is not on the cpu

g1y5x3: Thanks for the catch. Will it make sense to just use `force=True`? Which would do
`t.detach().cpu().resolve_conj().resolve_neg().numpy()`

dc-dc-dc: could just leave it as is, don't think any issues has been brought up that requires the force flag addition here

dc-dc-dc:
can you add some tests for `sparse_categorical_crossentropy` in test_ops, torch has there implementation under `torch.nn.functional.cross_entropy`

tinygrad/runtime/ops_torch.py
geohot: This is the torch buffer, not the tinygrad buffer!

geohot:
This should be two PRs I think

g1y5x3:
split the `.cpu().numpy() -> .numpy()` part into #1594

g1y5x3:
the current implementation matches the default setting in pytorch `reduction='mean'`

geohot:
This is broken. Accuracy stays pegged at 1.00

`kafka@tubby:~/fun/tinygrad$ python3 test/models/test_mnist.py TestMNIST.test_conv
loss 0.39 accuracy 1.00:  52%|██████████████████████████████████████████████████████████████████████████████████████████████████▊                                                                                           | 52/100 [00:01<00:01, 37.02it/s]`

PR #1593, opened 2023-08-21
Tensor.uniform with dtype=int bug fix

jaw566:
Identified the following bug: invoking Tensor.uniform with dtype=int32 returns a tensor with every entry equal to the low argument instead of a set of uniformly distributed integers


tinyb0t:
Changes made in `tinygrad/`:
```
total                                      3               2
```

tinygrad/tensor.py
```diff
@@ -182,7 +182,8 @@ def randn(*shape, dtype:Optional[DType]=None, **kwargs) -> Tensor:
   def normal(*shape, mean=0.0, std=1.0, **kwargs) -> Tensor: return (std * Tensor.randn(*shape, **kwargs)) + mean
 
   @staticmethod
-  def uniform(*shape, low=-1.0, high=1.0, **kwargs) -> Tensor: return ((high-low) * Tensor.rand(*shape, **kwargs)) + low
+  def uniform(*shape, low=-1.0, high=1.0, dtype:Optional[DType]=None, **kwargs) -> Tensor:
+    return ((high-low) * Tensor.rand(*shape, **kwargs)).cast(Tensor.default_type if dtype is None else dtype) + low
```
dc-dc-dc, line 186: is dtype not getting passed via kwargs?

jaw566: Ah, yes, looking into this now

ZenginU: I believe your approach is kinda true, if you give dtype with kwargs Tensor.rand is converted to 0 or 1 implicitly.

tinygrad/tensor.py
```diff
@@ -182,8 +182,9 @@ def randn(*shape, dtype:Optional[DType]=None, **kwargs) -> Tensor:
   def normal(*shape, mean=0.0, std=1.0, **kwargs) -> Tensor: return (std * Tensor.randn(*shape, **kwargs)) + mean
 
   @staticmethod
-  def uniform(*shape, low=-1.0, high=1.0, dtype:Optional[DType]=None, **kwargs) -> Tensor:
-    return ((high-low) * Tensor.rand(*shape, **kwargs)).cast(Tensor.default_type if dtype is None else dtype) + low
+  def uniform(*shape, low=-1.0, high=1.0, **kwargs) -> Tensor:
+    dtype = kwargs.pop("dtype") if "dtype" in kwargs else Tensor.default_type
```
ZenginU, line 186:
```
pop(key[, default])
If key is in the dictionary, remove it and return its value, else return default. If default is not
```
I believe
```python
dtype = kwargs.pop("dtype", Tensor.default_type)
```
is more readable

tinygrad/tensor.py
jaw566: I think my latest commit (pruning dtype from kwargs) addresses this issue?

tinygrad/tensor.py
jaw566: Totally agree. Fixing now

PR #1595, opened 2023-08-21
add cache collector

nimlgen:
This PR introduces a cache collector.

CacheCollector manages buffers while collecting allowing for reusability of buffers.

BS=64 hlb_cifar.py: 0.59gb -> 0.38gb
SD: 25+gb -> 6.54gb
openpilot jit: 0.28gb -> 0.26gb (583 -> 479 buffers in thneed)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        2               3
tinygrad/jit.py                            4               4
tinygrad/ops.py                            2               1
tinygrad/runtime/cache_collector.py       43               0
tinygrad/runtime/lib.py                    1               0
total                                     52               8
lines added in the tinygrad folder: 44
```

examples/stable_diffusion.py
```diff
@@ -621,20 +622,23 @@ def get_x_prev_and_pred_x0(x, e_t, index):
...
     return x_prev, pred_x0
 
+  @TinyJit
+  def do_step(latent, timestep):
+    e_t = get_model_output(latent, timestep)
+    x_prev, pred_x0 = get_x_prev_and_pred_x0(latent, e_t, index)
+    #e_t_next = get_model_output(x_prev)
+    #e_t_prime = (e_t + e_t_next) / 2
+    #x_prev, pred_x0 = get_x_prev_and_pred_x0(latent, e_t_prime, index)
+    return x_prev.realize(), pred_x0
```
geohot, line 632: `pred_x0` needs a realize, not sure how this works without it

tinygrad/runtime/cache_collector.py
```diff
@@ -0,0 +1,43 @@
+from typing import Any, Optional, Dict, List, Callable, Tuple
+from weakref import ref
+from tinygrad.runtime.lib import RawBuffer
+
+class _CacheCollector:
```
geohot, line 5: Line count is not important! This is really hard to read.

geohot:
I don't really understand how this differs from `LRUAllocator`

nimlgen:
> I don't really understand how this differs from LRUAllocator

This changes the way we collect cache. With just `GlobalCounters.cache=[]` we keep all the buffers and not allow them to free during collection and in jit. CacheColletor with its logic allows buffers to free as soon as they really can and it optimizes buffers in the cache itself.

geohot:
Yea, this is why I think it belongs in the JIT, since it's really a feature of the JIT only

examples/stable_diffusion.py
nimlgen: it was not used, cleaned this a bit.

tinygrad/runtime/cache_collector.py
nimlgen: added a bit of comments and make it cleaner in some places. lmk if this is better.

geohot:
It looks like stable diffusion is fully jitted with this! 

Is this it for #1296?

geohot:
Does freeing work, aka can you delete the JIT after the steps are done? Currently it looks like it's keeping the RAM in use while decoding.

geohot:
Congrats, and sorry it look me so long to test and read.

Lmk if that's all the cache changes, and the $500 `caching allocator, all backends, must intelligently free, reduce memory usage in both JIT and thneed (JIT SD)` bounty is yours

nimlgen:
Yeah, that was the last PR for #1296. All of caching things work fully for all gpu backends.

examples/stable_diffusion.py
```diff
@@ -621,20 +622,23 @@ def get_x_prev_and_pred_x0(x, e_t, index):
     x_prev = math.sqrt(a_prev) * pred_x0 + dir_xt #+ noise
     return x_prev, pred_x0
 
+  @TinyJit
+  def do_step(latent, timestep):
+    e_t = get_model_output(latent, timestep)
+    x_prev, _ = get_x_prev_and_pred_x0(latent, e_t, index)
```
wozeparrot, line 628: `index` here is from the loop on 638, and the calculations that use it use python math so it won't be jit captured and stays the same.

PR #1603, opened 2023-08-21
perf: lazyop as dataclass

roelofvandijk:
before
```
codegen         mean runtime:  189.89ms, runs:   194.70,  239.41,  180.85,  170.33,  207.72,  178.27,  179.02,  183.95,  183.69,  181.00
methodcache     mean runtime:  176.97ms, runs:   170.78,  224.18,  163.98,  163.25,  164.45,  189.46,  193.26,  167.76,  166.27,  166.29
profile         mean runtime:  790.07ms, runs:   769.91,  794.76,  799.09,  753.40,  847.84,  762.01,  795.60,  804.17,  770.68,  803.29
```

after
```
codegen         mean runtime:  182.52ms, runs:   182.84,  250.48,  175.73,  165.99,  176.54,  173.80,  172.70,  179.16,  173.53,  174.45
methodcache     mean runtime:  165.79ms, runs:   159.17,  220.91,  158.91,  177.57,  155.09,  155.78,  160.16,  153.85,  154.83,  161.62
profile         mean runtime:  785.40ms, runs:   758.52,  798.52,  801.80,  765.71,  780.81,  781.45,  806.35,  778.69,  778.43,  803.76
````

### CODE REVIEW

**Comment by tinyb0t** [comment]
Changes made in `tinygrad/`:
```
tinygrad/ops.py                            7              12
lines added in the tinygrad folder: -5
```

**Comment by geohot** [comment]
Err, this is confusing. What's a dataclass? Can this be a NamedTuple, I understand NamedTuple

**Comment by roelofvandijk** [comment]
They are similar in that they offer quite some convenience methods out-of-the box, such as an `__init__` to store the args in members, but a dataclass need not be immutable (a NamedTuple always is). `@dataclass(frozen=True)` is immutable (and hashable). 

Here is a pretty good comparison:
https://death.andgravity.com/namedtuples

Let me see if this can become a `NamedTuple`.

**Comment by roelofvandijk** [comment]
Ah yes, two blockers:
* NamedTuple cannot have a weakref.
* this line modifies one of the attributes: https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/linearizer.py#L518.
I think a dataclass is really the nicest choice here.

**Comment by roelofvandijk** [comment]
In general: there is quite a lot to be gained in terms of speed in the object initialization of Tensors, Functions, Buffers and Ops.

File: tinygrad/ops.py [hunk 1]
```diff
@@ -22,28 +23,22 @@ class LoadOps(Enum): EMPTY = auto(); RAND = auto(); CONST = auto(); FROM = auto(
...
-  arg: Any
-  buffers: Tuple[LazyBuffer, ...]
+  arg: Any = None
+  buffers: Tuple[LazyBuffer, ...] = field(init=False)
 
-  def __init__(self, op: Op, src: Tuple[Union[LazyOp, LazyBuffer], ...], arg: Any = None):
-    self.op = op
-    self.src = src
-    self.arg = arg
+  def __post_init__(self):
```
geohot, line 33: The behavior of this looks different

roelofvandijk: I do not follow you.

geohot: I mean do we still need the try?

The behavior looks different because I think it used to be all or nothing but now it's not.

geohot:
Oh, we also still need to support Python 3.8...

tinygrad/ops.py
roelofvandijk: Seems to me that it still is?

roelofvandijk: Outdated.

wozeparrot:
So I do kinda support this, `LazyOp` does make sense as a dataclass because of how it's initialized, and generally it is cleaner for me to read at least. But how much performance does having it be a dataclass give us? I thought dataclass was just syntax sugar around normal classes, or is it closer to a namedtuple during execution?

roelofvandijk:
Frozen data classes are very close, if not faster than namedtuples (in terms of init, property access, methods etc.). Also hashable and immutable by default.

roelofvandijk:
1) Frozen dataclass
2) Moved buffers to property to allow freezing. This also makes the hash, repr and eq functions the natural ones.

Did not focus on speed but on readability here.

roelofvandijk:
@wozeparrot

geohot:
Cool, let's merge!

PR #1606, opened 2023-08-21
perf: only sum of ints is int

roelofvandijk:
Only a sum of all ints is int. This is faster, but might not be more readable.

### DIFF
--- a/tinygrad/codegen/optimizer.py
+++ b/tinygrad/codegen/optimizer.py
@@ -222,7 +222,7 @@ def hand_coded_optimizations(k:Linearizer):
       # early exit
       return
 
-  if k.opts.has_local and all(isinstance(s, int) for s in k.sts[0].shape[:k.first_reduce]):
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/optimizer.py              1               1
```

dc-dc-dc:
Is this a significant speed up ?

tinygrad/codegen/optimizer.py
```diff
@@ -222,7 +222,7 @@ def hand_coded_optimizations(k:Linearizer):
       # early exit
       return
 
-  if k.opts.has_local and all(isinstance(s, int) for s in k.sts[0].shape[:k.first_reduce]):
+  if k.opts.has_local and sum(k.sts[0].shape[:k.first_reduce]).__class__ is int:
```
geohot, line 225: Err, I can't understand this

roelofvandijk:
It's significant, but too indirect. Who knows what you lot will put into the shape and if that suppors the sum operation. Closed.

PR #1608, opened 2023-08-21
np.argmax to Tensor.argmax

ZenginU:
There are more np.argmax than this PR. I have not touched them as i did not want to change the behaviour accidentally.

Only changed rather seemingly safe part.

Note: I did not test the models as I am unable to do so.


ZenginU:
breaks examples/hlb_cifar10.py do not know why.

PR #1440, opened 2023-08-05
Chunk one-line, repeats format

ZenginU:
* Just refactored chunk to one line preserving the readability (I hope).
* Converted if else into one line in repeats

PR #1610, opened 2023-08-22
Readability for unreadable functions

ZenginU:
Can be considered similar to #1440 

Made eye load from constant,
I believe cumsum is much more readable, repeat has kind of duplicate code, made chunk more readable 
I do not think this is considered code golf, if so please inform me which parts

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -383,7 +383,7 @@ class Tensor:
     final_shape = [r*s for r,s in zip(repeats, base_shape)]
     return self.reshape(new_shape).expand(expand_shape).reshape(final_shape)
 
-  def chunk(self, num:int, dim:int) -> List[Tensor,...]:
+  def chunk(self, num:int, dim:int) -> List[Tensor]:
     dim, step = dim + self.ndim if dim < 0 else dim, ceil(self.shape[dim]/num)
     slice_params = [[slice(None)]*dim + [slice(k, k + step)] for k in range(0, self.shape[dim], step)]
     return [self[tuple(sl)] for sl in slice_params]
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -242,8 +242,7 @@ class Tensor:
   def shrink(self, arg:Tuple[Tuple[int, int], ...]) -> Tensor: return mlops.Shrink.apply(self, arg=arg) if any(x != (0,s) for x,s in zip(arg, self.shape)) else self
   def pad(self, arg: Tuple[Tuple[int, int], ...], value:float=0) -> Tensor:
     ret = mlops.Pad.apply(self, arg=arg) if any(x != (0, 0) for x in arg) else self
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         8              16
lines added in the tinygrad folder: -8
```

tinygrad/tensor.py
```diff
@@ -544,13 +538,13 @@ def tan(self): return self.sin() / self.cos()
 
   @staticmethod
   def _tri(r:int, c:int, k:int=0, **kwargs) -> Tensor: return Tensor.arange(r, **kwargs).unsqueeze(1).expand(r,c) <= Tensor.arange(-k, c-k, **kwargs).unsqueeze(0).expand(r,c)
-  def triu(self, k:int=0) -> Tensor: return Tensor._tri(self.shape[-2], self.shape[-1], k=k, dtype=self.dtype).where(self, Tensor.zeros_like(self))
-  def tril(self, k:int=0) -> Tensor: return Tensor._tri(self.shape[-2], self.shape[-1], k=k+1, dtype=self.dtype).where(Tensor.zeros_like(self), self)
+  def triu(self, k:int=0) -> Tensor: return Tensor._tri(self.shape[-2], self.shape[-1], k=k, dtype=self.dtype) * self
```
geohot, line 541: Err, there's issues with inf for this

ZenginU: Currently it supports inf but is it a bug? 
Tensor(float("inf")) * 0 = nan
Tensor([float("inf")]) * Tensor(0) = 0
Tensor([float("inf")]) * Tensor([0]) = nan
Can we trust this behaviour?

tinygrad/tensor.py
```diff
@@ -544,13 +538,13 @@ def tan(self): return self.sin() / self.cos()
...
+  def tril(self, k:int=0) -> Tensor: return (1 - Tensor._tri(self.shape[-2], self.shape[-1], k=k+1, dtype=self.dtype)) * self
 
   # ***** math functions (unary) *****
   def trunc(self: Tensor) -> Tensor: return self.cast(dtypes.int32).contiguous().cast(self.dtype)
-  def ceil(self: Tensor) -> Tensor: return (self > (b := self.trunc())).where(b+1, b)
-  def floor(self: Tensor) -> Tensor: return (self < (b := self.trunc())).where(b-1, b)
+  def ceil(self: Tensor) -> Tensor: return (self > (b := self.trunc())) + b
```
geohot, line 546: I'd leave this too, dtypes shouldn't mix like this

tinygrad/tensor.py
wozeparrot: i knew there was a reason it was where, where has better inf and nan handing than a masked multiple does

ZenginU: @wozeparrot So can we use .where for "def pad" as well. We rely on 1/inf = 0

PR #1611, opened 2023-08-22
support symbols in shrink and JIT freqs_cis in llama

chenyuxyz:
This jits the `freqs_cis` kernel using `shrink`, and updates the jit to check `ShapeTracker.views` against the cached input. Verified that the old llama failed if we removed contiguous with this check.

Will check for the 5ms thing and jit the new sample (that uses `Tensor.argmax`) in a separate PR.

Assigning variable using `freqs_cis.lazydata.st.var_vals[pos] = start_pos` works, but it feels weird as an API. Another use case that we don't support now is reshape a 2d matrix into (variable, variable) that commavq might want?

### DIFF
--- a/examples/llama.py
+++ b/examples/llama.py
@@ -151,15 +151,18 @@ class Transformer:
 
   def __call__(self, tokens:Tensor, start_pos:int):
     _bsz, seqlen = tokens.shape
-    # get only the part we are using.
-    # NOTE: if you remove contiguous here, it breaks because you can't put different ShapeTrackers into the compiled JIT
-    # NOTE: realize is not enough, since the realized buffer will have an offset that the kernel doesn't know about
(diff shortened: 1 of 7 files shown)


geohot:
Woah, that was fast. I suspect to hit 5ms you'll have to do less different JITs. I added symbolic to GPT-2 with ~one (didn't do the outer stuff, waiting for #1612 . https://github.com/tinygrad/tinygrad/pull/1613

chenyuxyz:
yeah we already have symbolic offset supported so it's mostly just testing and adding test cases

Do you have suggestions on how to measure python runtime on a M1 Max? `GPU=1` shows 40ms but more than 20ms are on the setting scalar thing. I got the 15ms number from a CUDA machine and not sure if I should try to fix pyopencl or metal backend

PR #1616, opened 2023-08-22
#1615 fix

### LINKED ISSUE #1615: `device=self.device` missing from lots of places in tensor.py
See: https://github.com/tinygrad/tinygrad/blob/4f459841bcfc57bf4f328b45553212c3bd0b6fdb/tinygrad/tensor.py#L338
And: https://github.com/tinygrad/tinygrad/blob/4f459841bcfc57bf4f328b45553212c3bd0b6fdb/tinygrad/tensor.py#L334

Should be something like: https://github.com/tinygrad/tinygrad/blob/4f459841bcfc57bf4f328b45553212c3bd0b6fdb/tinygrad/tensor.py#L711

Basically: https://github.com/tinygrad/tinygrad/pull/1591#discussion_r1300387689

We don't want to use the default device for this, should be on the same device that the other tensors are on.

geohotstan:
#1615
I also turned requires_grad off on some of them
also for sparse_categorical_crossentropy and argmax, should the arange be int32?

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -331,11 +331,11 @@ class Tensor:
       idx = [i.reshape(*[1]*(max_dim-i.ndim), *i.shape) for i in idx]
       sum_dim = [d+max_dim-n for n,d in enumerate(dim)]
       new_idx = idx[0].reshape(*[1]*dim[0], 1,*idx[0].shape, *[1]*(ret.ndim-dim[0]-1))
-      arange = Tensor.arange(ret.shape[dim[0]], dtype=dtypes.int32, requires_grad=False).reshape(*[1]*dim[0], ret.shape[dim[0]], *[1]*idx[0].ndim, *[1]*(ret.ndim-dim[0]-1))
+      arange = Tensor.arange(ret.shape[dim[0]], dtype=dtypes.int32, requires_grad=False, device=self.device).reshape(*[1]*dim[0], ret.shape[dim[0]], *[1]*idx[0].ndim, *[1]*(ret.ndim-dim[0]-1))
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         6               6
```

wozeparrot:
yea, the dtypes should be fine for crossentropy and argmax, both are indexes

tinygrad/tensor.py
```diff
@@ -158,7 +158,7 @@ def arange(start, stop=None, step=1, **kwargs):
 
   @staticmethod
   def full_like(tensor, fill_value, dtype:Optional[DType]=None, **kwargs):
-    return Tensor.full(tensor.shape, fill_value=fill_value, dtype=tensor.dtype if dtype is None else dtype, **kwargs)
+    return Tensor.full(tensor.shape, fill_value=fill_value, dtype=tensor.dtype if dtype is None else dtype, device=tensor.device, **kwargs)
```
ZenginU, line 161: isn't device is included in **kwargs

geohotstan:
@ZenginU also found that [ones_like](https://pytorch.org/docs/stable/generated/torch.ones_like.html) and [zeros_like](https://pytorch.org/docs/stable/generated/torch.zeros_like.html) from torch
"Returns a tensor filled with the scalar value 1/0, with the same size as input. torch.ones_like(input) is equivalent to torch.ones(input.size(), dtype=input.dtype, layout=input.layout, **_device=input.device_**)"

So I suppose having full_like take the `device=tensor.device` is gooooooooooooood?

tinygrad/tensor.py
wozeparrot, line 161: need to remove a possible device from the `kwargs` here

wozeparrot:
yea thats fine

wozeparrot:
merge conflict

PR #1635, opened 2023-08-23
Implement random_choices in Tensor

c143:
Not sure if I'm going into the right direction with this.
Assumed it should be implemented with the Tensor.rand function and not directly with `np.random.choice`

Not happy with the current state, though.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                         3               0
```

PR #1644, opened 2023-08-23
do not replace divs with muls for non-floats

### LINKED ISSUE #1629: Tensor.maximum with int tensors broken on interpreted backends
reproduce:
```
Torch or CPU
In [22]: Tensor([20,10,3], dtype=dtypes.int32).maximum(Tensor([20,7,3], dtype=dtypes.int32)).numpy()
Out[22]: array([ 0, 10,  0], dtype=int32)
```
when the case that values equal, maximum sets it to 0

nimlgen:
Fixes #1629


geohot:
Good find

PR #1645, opened 2023-08-23
Implement random choice with replacement

ndieckow:
Sampling without replacement is still missing at the moment.

Explanation: We can sample from an arbitrary distribution by first sampling uniformly between 0 and 1 and then trying to insert those samples into the CDF whilst maintaining order. The indices where we would need to insert are then equal to the indices of the elements we wanted to sample in the first place. For example, if our distribution is `[0.1, 0.1, 0.8]`, the CDF would be `[0.1, 0.2, 1]`. Around 80% of uniform samples will fall between 0.2 and 1, so most indices will be 2, which corresponds to the element with probability 0.8.
NumPy uses `searchsorted` to find the indices, but I don't think we can do binary search lazily. Instead, we compute for each uniform sample how many CDF entries are smaller than or equal to it, which also gives us the index but only in linear time.


ndieckow:
Just realized that this is inefficient compared to NumPy's implementation, because I'm not using binary search. Maybe that can be done lazily as well.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                        15               0
lines added in the tinygrad folder: 15
```

ndieckow:
Without replacement is trickier to do (at least the way numpy does it), because it requires to keep track of the indices we have already sampled, which is not possible lazily. A direct approach would be to create a new tensor from `a` where the proportions match up with the probabilities, e.g. `a = [1, 2]` and `p = [0.2, 0.8]` becomes `a = [1, 2, 2, 2, 2]` and `p = [0.5, 0.5]`, which can then be sampled by computing a permutation. This seems a bit wasteful, but would work.

wozeparrot:
> A better approach seems to be the so-called Gumbel-max trick: [stackoverflow.com/a/55083317](https://stackoverflow.com/a/55083317) but it requires something like `top_k` or `argmax`.

There is an argmax.

ndieckow:
> > A better approach seems to be the so-called Gumbel-max trick: [stackoverflow.com/a/55083317](https://stackoverflow.com/a/55083317) but it requires something like `top_k` or `argmax`.
> 
> There is an argmax.

Ah, typo. I meant `argsort`. Fixed it in my comment.

wozeparrot:
Is it possible to do the wasteful way that you mentioned for sampling without replacement?

Actually how much is the `replace` arg actually changed from its default value?

ndieckow:
> Is it possible to do the wasteful way that you mentioned for sampling without replacement?

The wasteful way I thought might work requires us to permute a tensor created from Tensor.arange. At the time, I thought `permute` did this because I hadn't read tinygrad's code in detail, but it turns out that that's just for permuting dimensions.

> Actually how much is the `replace` arg actually changed from its default value?

Good question, I'm not sure. I mostly use it with `replace=False` whenever I'm building a test/train data split, but there may be more use cases.

wozeparrot:
> > Is it possible to do the wasteful way that you mentioned for sampling without replacement?
> 
> The wasteful way I thought might work requires us to permute a tensor created from Tensor.arange. At the time, I thought `permute` did this because I hadn't read tinygrad's code in detail, but it turns out that that's just for permuting dimensions.

You should be able to permute a tensor and not its dims with the new fancy indexing in `__getitem__`, api should be the same as torch.

geohot:
Can you write ops tests for this that compare to torch? See `test/test_ops.py`

ndieckow:
@wozeparrot Thanks for the help. Turns out trickier than I originally thought to repeat the values according to the probabilities. Not sure if I can figure it out at the moment, because I have to focus on other things.

@geohot I don't see a good way to compare them because of the randomness. The most I could do is using a distribution like [1, 0, 0], where the chosen element is deterministic.

wozeparrot:
Can't you just set the random seed to do the tests? wouldn't that make it deterministic?

ndieckow:
> Can't you just set the random seed to do the tests? wouldn't that make it deterministic?

But Torch and tinygrad have different random generators, right? Afaik, there's no way to match them up, even when using the same seed.

geohot:
Yea this needs a lot more tests. All the types options aren't even checked

geohot:
Closing for lack of tests

tinygrad/tensor.py
```diff
@@ -191,6 +191,21 @@ def kaiming_normal(*shape, a:float = 0.01, **kwargs) -> Tensor:
     std = math.sqrt(2.0 / (1 + a ** 2)) / math.sqrt(math.prod(shape[1:]))
     return Tensor.normal(*shape, mean=0.0, std=std, **kwargs)
 
+  @staticmethod
+  def choice(a: Union[Tensor, int], size: Union[int, Tuple[int, ...]] = 1, p: Optional[Tensor] = None):
```
geohot, line 195: Remove staticmethod, should have to be a Tensor

geohot:
Added $100 bounty for Tensor.multinomial (very similar to this), happy to lock it to you if you want it. You should be able to test it with the generator parameter

PR #1651, opened 2023-08-23
[not to be merged] implement random choice lmaooooooo

geohotstan:
**_posting a draft PR just to share ideas_**
This implementation is super cursed, probably the most cursed thing I've written and I'm wasting a bunch of ops (need to clean)
but it works? (I know the bugs)
Idk if this is the right idea though since I didn't learn math in skool

---------
the idea is this:
given `Tensor.choice(5, 20, Tensor([0.1, 0, 0, 0.8, 0.1]))` for example
if `Tensor.rand(size)` gives random numbers between 0 -> 1
we want to find the distance between `Tensor.rand(size)` and p such that min() of that distance we choose as the "choice"

so since p[0] is 0.1, there's a 10% chance p[0] is selected as "choice" and so we want p[0] to be 0.05 since that's the middle distance between 0 and 0.1 and for any rand in Tensor.rand(size), rand < 0.1 should have shortest distance to p[0].

for p[1] and p[2], we set 0's as float(-inf) so they're never selected.

for p[4], it's also selected 10% of the time so we set it as 0.15 and we want any 0.1 < rand < 0.2 to have shortest distance to p[4].

for p[3] since it's the max, we just set it to 0.25 since any 0.2 < rand will have shortest distance to p[3], which is 80% of the time.... right?

so we have` [0.05 -inf -inf 0.25  0.15]` as p which i call "cummy" in my code

---------
with 
```
  diff = (cummy-rand).abs()
  idx = ((diff == diff.min(axis=1).reshape(size,1)) * ints.reshape(1,high)).sum(axis=1)
```
we get all the idx of the choices.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -140,6 +140,7 @@ class Tensor:
     Tensor._seed += 1
     return Tensor._loadop(LoadOps.RAND, prod(shape), arg=Tensor._seed, **kwargs).reshape(shape)
 
+  @staticmethod
   def choice(high, size, p:Tensor):
     ints = Tensor.arange(high)
     peesum = p/p.sum()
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/tensor.py                        19               0
lines added in the tinygrad folder: 19
```

chenyuxyz:
is this correct? if the uniform sample picks 0.19 does it return p[3]?

geohotstan:
> is this correct? if the uniform sample picks 0.19 does it return p[3]?

:o crap. whoops LOL brb, fixing

geohotstan:
oh wow I just understood ndieckow's way of determining idx. Yeah that's much better. Closing this for now.

PR #1656, opened 2023-08-24
enable jit test for METAL and update failed/flaky tests

chenyuxyz:
JIT tests work now because `np.testing.assert_allclose` mitigates the numerical issue in CI.

Added all tests to METAL CI by default. I manually disabled the failed ones. The dtype one is known, vit has buffer limitation that would need buffer limit support. Not sure how relevant `test/unit/test_example.py` is now given that we have multiple backend tests in CI, and it failed with METAL on pyopencl stuff so I excluded it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -180,8 +180,15 @@ jobs:
       run: pip install -e '.[metal,webgpu,testing]' --extra-index-url https://download.pytorch.org/whl/cpu
     - name: Test LLaMA compile speed
       run: PYTHONPATH="." METAL=1 python test/external/external_test_speed_llama.py
-    - name: Run metal tests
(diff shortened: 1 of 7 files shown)


dc-dc-dc:
The CI tests should be rerun a couple times as METAL has been flaky before

chenyuxyz:
understood, I just removed the test.yml from this PR and I can do more local tests and figure out the what's flaky.

also updated some flaky test cases. The symbolic ones I am working on fixing, I suspect it hit bad cache from other tests so can fail if run with other tests. The linearizer one was testing on numbers very close to zeros, so I changed the `-` to be `+` and added `rtol`

chenyuxyz:
closing this one - I am fixing these tests individually

PR #1657, opened 2023-08-24
fix CIFAR jit

g1y5x3:
```
999   86.53 ms run,    3.84 ms python,   82.68 ms CL,  491.54 loss, 0.000070 LR, 6.90 GB used,   7793.55 GFLOPS
eval 9404.0/10000 94.04%,    0.40 val_loss STEP=1000
```
1. move to device, cast to float, normalize, reshape, one-hot encode labels, and pad to 36x36 all technically need to be done only once, thus take them out of the `fetch_batch`.
2. The left pre-processing operations that need to be done while batching are random crop, random flip, and cutmix with random masks. They are currently applied on all 50,000 training data every 98 STEPS and are not **jitted**. This is also what [hlb_CIFAR10](https://github.com/tysam-code/hlb-CIFAR10/blob/ff53cac99dc71a66b1ada43d63d1307debd592a7/main.py#L460C22-L460C22) is doing.
3. There's still 3 places inside `fetch_batch` that makes a `numpy` copy.  Mainly for applying random transformations (shuffle, random mask). Essentially we need a `random.choice` and a much faster `gather`, since random transformations must be timed for [mlperf](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#62-preprocessing-during-the-run). The kernel of numpy's gather is actually really simple but not sure we can achieve it now maybe with UAST.
```
numpy.take(a, indices, axis=None, out=None, mode='raise')

Ni, Nk = a.shape[:axis], a.shape[axis+1:]
Nj = indices.shape
for ii in ndindex(Ni):
    for jj in ndindex(Nj):
        for kk in ndindex(Nk):
            out[ii + jj + kk] = a[ii + (indices[jj],) + kk]
```

### DIFF
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -42,7 +42,7 @@ hyp = {
       'kernel_size': 2,             # kernel size for the whitening layer
       'batch_norm_momentum': .5,
       'cutmix_size': 3,
-      'cutmix_steps': 588,          # original repo used epoch > 12.1 - 6 which is roughly 7*98=686 STEPS
+      'cutmix_steps': 490,          # different from original repo which used epoch > 12.1 - 6 which is roughly 7*98=686 STEPS
       'pad_amount': 2
   }
 }
@@ -189,7 +189,7 @@ def fetch_batches(X_in, Y_in, BS, seed, is_train):
     if is_train:
       X = random_crop(X, crop_size=32)
       X = Tensor.where(Tensor.rand(X.shape[0],1,1,1) < 0.5, X[..., ::-1], X) # flip LR
-      # if step >= hyp['net']['cutmix_steps']: X, Y = cutmix(X, Y_in, mask_size=hyp['net']['cutmix_size'])
+      if step >= hyp['net']['cutmix_steps']: X, Y = cutmix(X, Y, mask_size=hyp['net']['cutmix_size'])
     X, Y = X.numpy(), Y.numpy()
     for i in range(0, X.shape[0], BS):
       # pad the last batch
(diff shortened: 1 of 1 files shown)


examples/hlb_cifar10.py
```diff
@@ -160,34 +160,15 @@ def make_square_mask(X, mask_size):
...
-
-transform_test = [
-  lambda x: x.to(device=Device.DEFAULT).float(),
-  lambda x: x / 255.0,
-  lambda x: (x - Tensor(cifar_mean).repeat((1024,1)).T.reshape(1,-1))/ Tensor(cifar_std).repeat((1024,1)).T.reshape(1,-1),
-  lambda x: x.reshape((-1,3,32,32)),
-]
-
-def cutmix(X, Y, mask_size=3, p=0.5):
-  if Tensor.rand(1) > p: return X, Y
```
g1y5x3, line 187: This for some reason was always true and `cutmix` was never called.

examples/hlb_cifar10.py
```diff
@@ -342,6 +340,10 @@ def eval_step_jitted(model, X, Y):
           best_eval = acc
           print(f"eval {correct_sum}/{correct_len} {acc:.2f}%, {(sum(losses)/len(losses)):7.2f} val_loss STEP={i}")
     if STEPS == 0 or i==STEPS: break
+    X, Y = next(batcher)
+    # further split batch if distributed
+    if getenv("DIST"):
+      X, Y = X.chunk(world_size, 0)[rank], Y.chunk(world_size, 0)[rank]
```
g1y5x3, line 346: Move the batcher down to avoid fetching one extra time during the last STEP

examples/hlb_cifar10.py
wozeparrot: thats cuz we don't know what the tensor contents are at this point and `Tensor` overrides the comparison operators to return a `Tensor` which means that this just checked if it was `None`

PR #1660, opened 2023-08-24
_tri not use arange function

ZenginU:
Couldnt really provide solid benchmarking as speed is very volatile.
For average this PR, has ~0.37 ms for 256 .triu()
For average prev PR, has ~0.40 ms for 256 .triu()

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -534,7 +534,7 @@ class Tensor:
   def tan(self): return self.sin() / self.cos()
 
   @staticmethod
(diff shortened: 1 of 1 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

wozeparrot:
hmm, not sure about this, it's less readable, and I don't think the small speedup is worth it. Cuz, `arange` is used the same way for all the other indexing stuff so there it's an easy target to optimize, but `_pool` is much more varying in its uses.

ZenginU:
Yeah, totally agree if arange is optimized for O(N).

PR #1662, opened 2023-08-24
Quickstart: Use tensors to compute train accuracy

DavidFarago:
In the quickstart MNIST training example, only convert final accuracy tensor to numpy, to avoid mixed type comparison in `pred == labels`, which yields `True` everywhere and thus accuracy `1.0`.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -205,11 +205,11 @@ for step in range(1000):
   opt.step()
 
   # calculate accuracy
-  pred = out.argmax(axis=-1).numpy()
+  pred = out.argmax(axis=-1)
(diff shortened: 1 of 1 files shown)


g1y5x3:
Assuming you have also tested after this and outputs looks correct? (the accuracy is not constantly 1?)

DavidFarago:
> @g1y5x3, yup:
```
Step 1 | Loss: 23.40022850036621 | Accuracy: 0.078125
Step 101 | Loss: 0.9517345428466797 | Accuracy: 0.875
Step 201 | Loss: 0.5465449690818787 | Accuracy: 0.90625
```

PR #1665, opened 2023-08-25
replaced npgather

scottrcarlson:
replaced npgather "using only tinygrad". The performance is similar.

I'm still getting my feet wet.


g1y5x3:
gather is already supported now and this essentially will be replaced by the mask rcnn training PR

scottrcarlson:
understood, thanks!

Issue #1640, opened 2023-08-23
Bounty: Training Mask R-CNN


waifairer:
Add the training logic to examples/mlperf/model_train.py -> def train_maskrcnn():

[Links](https://github.com/mlcommons/training/tree/master/object_detection) to the reference implementation for object detection.

[Links](https://cocodataset.org) to the dataset.

MLPerf training rules [here](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#appendix-allowed-optimizers)

---

Current Bounty: $500
Bounty Locked #1668

PR #1668, opened 2023-08-25
Training Mask-RCNN

waifairer:
This is the training frame for Mask-RCNN. I have a good handle on the todos remaining to get the model training:

Next Steps:
- [ ] Implement loss in rpn
- [ ] test losses in rpn using knowns
- [x] implement loss in roi_head
- [x] test loss roi_head
- [x] implement dataloader
- [ ] implement fast annotation lookups
- [ ] early exit / status tracking

Nice-to-Haves:
- [ ] randomhorizontalflip


examples/mask_rcnn/train_mask_rcnn.py
```diff
@@ -0,0 +1,252 @@
...
+  # but the code supports more general grouping strategy
+  aspect_grouping = [1]
+
+  # todo reintroduce test 
+  dataset_list = ["/path/to/train"]
+
+  transforms = build_transforms(is_train=True)
+
+  ## TODO, this needs a tiny equivalent
+  # datasets, epoch_size = build_dataset(dataset_list, transforms, DatasetCatalog, is_train)
```
waifairer, line 85: This "dataset" api doesn't appear to have a tiny equivalent. Correct me if I'm wrong. 

For good training we have to have a util that quickly pre-reads random images from disk and serves them to the model. This is probably one of the best places for training time gains.

models/mask_rcnn.py
```diff
@@ -1259,11 +1261,21 @@ def load_from_pretrained(self):
...
-    x, result, _ = self.roi_heads(features, proposals)
+    proposals, proposal_losses = self.rpn(images, features, targets)
+    
+    x, result, detector_losses = self.roi_heads(features, proposals, targets)
+
+    if self.training:
+      losses = {}
+      losses.update(detector_losses)
+      losses.update(proposal_losses)
+      return losses
```
waifairer, line 1277: I find this torch api to be a bit funky. I think it's a bit better to have the function always return a tuple, 

```
return result, losses if targets is not None else None
```

Then we would remove self.training altogether (pass targets=None if you don't want loss evaluation).

if no one disagrees I'll switch it.

waifairer:
#1640

waifairer:
Getting fairly close to the loss function being implemented

g1y5x3:
> randomhorizontalflip

Inside the current [`hlb_cifar`](https://github.com/tinygrad/tinygrad/blob/e3a062ad17c97c9e73dc59548360a4ddf1e7747e/examples/hlb_cifar10.py#L191) on master, there is an implementation for random lr flip that's all in tinygrad tensor operatons.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/lib.py                    6               1
```

waifairer:
Definitely a hard bounty, but I'm now at the point where I can start training the RPN (and by extension, the Resnet FPN). If I were a bit faster I'd have been able to capture the Resnet bounty as well.

I've documented the structure of the backbone + RPN for my own edification. I'm attaching it here incase someone is curious about it. Specifically, it took me a bit to understand that the anchors were generated only from the feature dimensions and not the features themselves. Anchors are totally deterministic for feature maps of known size.
Screenshot 2023-09-14 at 7 46 11 PM

waifairer:
Very exciting!
```
root@eb493c9d379e:~/tinygrad/examples/mask_rcnn# python3 ./loss.py
loading annotations into memory...
Done (t=9.08s)
creating index...
index created!
input [[[[ 121.05517     192.47644     374.19724    ...  256.19064
```
The first iteration of loss has been run.

geohot:
No updates in 2 weeks, closing

PR #1674, opened 2023-08-25
[ready] ci: add py38 to linters

roelofvandijk:
Running no tests on Python 3.8 makes it very easy to add features that are incompatible with that version. I suggest running at least the linters in both 3.8 and 3.11. Technically, only 3.8 would probably suffice.


wozeparrot:
I think it would be better to just have 3.8 only, all the CI stuff was bumped to 3.11 because its faster but the linter step is already fast enough on 3.8, and this way we don't use an extra runner.

PR #1605, opened 2023-08-21
[ready] perf: faster reshape

roelofvandijk:
See below.

PR #1683, opened 2023-08-26
[ready] perf: avoid reshaping if not necessary

roelofvandijk:
Also part of https://github.com/tinygrad/tinygrad/pull/1605, but screwing up the diff there.

### DIFF
--- a/tinygrad/shape/shapetracker.py
+++ b/tinygrad/shape/shapetracker.py
@@ -228,6 +228,7 @@ class ShapeTracker:
     return self
 
   def reshape(self, new_shape: Tuple[Union[Node,int], ...]):
+    if self.views[-1].shape == new_shape: return self
     new_ints, new_nodes = partition(new_shape, lambda s: isinstance(s, int))
     if new_nodes and all(isinstance(s, int) for s in self.shape):
       # reshape from all int shape into shape with a variable, update the variable value
@@ -238,7 +239,6 @@ class ShapeTracker:
         self.var_vals[new_var] = new_val
       else: assert self.var_vals[new_var] == new_val, f"value conflicts, was {self.var_vals[new_var]}, set to {new_val}"
     elif not new_nodes: self.var_vals = {}
-    if self.views[-1].shape == new_shape: return self
     assert all(is_sym_int(x) and x > 0 for x in new_shape), f"shape must be symbolic ints and can't contain 0 or negative numbers {new_shape}"
     # only check size for int shapes. we don't check symbolic here as long as the reshape itself can be done
     if all(isinstance(s, int) for s in self.shape) and all(isinstance(s, int) for s in new_shape):


tinyb0t:
Changes made in `tinygrad/`:
```
```

wozeparrot:
Does this break any of the symbolic stuff? because `self` is possibly mutated before this line was originally called.

@chenyuxyz

roelofvandijk:
I doubt it, this line can only be true if `new_shape` is all ints since `view.shape` should be all int.
But good to double-check.

chenyuxyz:
looks fine to me, if the input shape and output shape are the same, `self.var_vals` should remain the same

do you know if this path is hit at all? Because there's also a same check in `LazyBuffer.reshape`. We can remove this line if it's not hit at all

roelofvandijk:
It gets hit:

```
Total time: 1.63437 s
File: /home/rvd/src/roelofvandijk/tinygrad/tinygrad/shape/shapetracker.py
Function: reshape at line 229

   229                                             @profile
   230                                             def reshape(self, new_shape: Tuple[Union[Node,int], ...]):
   231     78989      45302.3      0.6      2.8      if self.views[-1].shape == new_shape: 
   232      3372        567.0      0.2      0.0        return self
   233     78989     427532.0      5.4     26.2      new_ints, new_nodes = partition(new_shape, lambda s: isinstance(s, int))
   234     78989      15583.1      0.2      1.0      if new_nodes and all(isinstance(s, int) for s in self.shape):
   235                                                 # reshape from all int shape into shape with a variable, update the variable value
   236                                                 assert len(new_nodes) == 1 and isinstance(new_nodes[0], Variable), "only support adding one Variable to the int shape"
   237                                                 new_var, new_val = new_nodes[0], prod(self.shape) // prod(new_ints)
   238                                                 if new_var not in self.var_vals:
   239                                                   assert new_var.min <= new_val <= new_var.max, f"variable value {new_val} out of range [{new_var.min}, {new_var.max}]"
   240                                                   self.var_vals[new_var] = new_val
   241                                                 else: assert self.var_vals[new_var] == new_val, f"value conflicts, was {self.var_vals[new_var]}, set to {new_val}"
   242     78989      28752.5      0.4      1.8      elif not new_nodes: self.var_vals = {}
   243     78989     409784.5      5.2     25.1      assert all(is_sym_int(x) and x > 0 for x in new_shape), f"shape must be symbolic ints and can't contain 0 or negative numbers {new_shape}"
   244                                               # only check size for int shapes. we don't check symbolic here as long as the reshape itself can be done
   245     78989     432496.2      5.5     26.5      if all(isinstance(s, int) for s in self.shape) and all(isinstance(s, int) for s in new_shape):
   246     78989     108764.0      1.4      6.7        assert prod(self.shape) == prod(new_shape), f"can't reshape {self.shape} -> {new_shape}" # type: ignore  # mypy cannot resolve, all ints here
   247     78989     121100.3      1.5      7.4      new_view, extra = _reshape(self.views[-1], new_shape)
   248     71754      12092.4      0.2      0.7      if extra: self.views.append(new_view)
   249     71754      19318.1      0.3      1.2      else: self.views[-1] = new_view
   250     78989      13081.8      0.2      0.8      return self
```


from here in optimizer.py. Might make sense to add it there to save the function call.
```
  def reshape_and_permute(self, new_shape_fxn, axis):
    for st in self.sts:
      if new_shape_fxn is not None: st.reshape(tuple(new_shape_fxn(st.shape)))
      if axis is not None: st.permute(tuple(axis))
```

roelofvandijk:
Ah, it's for cases like these:
```
<function OptimizedKernel.simplify_ones.<locals>.<lambda> at 0x7f102464f2e0> (32000, 1)
```

PR #1686, opened 2023-08-26
fix symbolic_ops tests with Tensor.training=True

chenyuxyz:
`scaled_dot_product_attention` in training mode always creates a dropout mask before, so symbolic tests will fail in training mode because we don't support `Tensor.rand` for symbolic shape. `test_nn.py` might set `Tensor.training` to true, and `test_symbolic_ops.py` fails if it's run after.

Fix by not creating a dropout mask if `dropout_p==0`, which is the default for `scaled_dot_product_attention`


tinyb0t:
Changes made in `tinygrad/`:
```
```

PR #1693, opened 2023-08-27
[ready] perf: faster lazyop eq

roelofvandijk:
The cast was killing the performance of `LazyOp.__eq__`

```
from:

Total time: 0.13719 s
File: /home/rvd/src/roelofvandijk/tinygrad/tinygrad/ops.py
Function: __eq__ at line 35

    35                                             @profile
    36                                             def __eq__(self, __value: object) -> bool:
    37     46378      14344.3      0.3     10.5      if __value.__class__ is not LazyOp: return False
    38     46378      98018.8      2.1     71.4      __value = cast(LazyOp, __value)
    39     46366      24826.8      0.5     18.1      return self.op == __value.op and self.src == __value.src and self.arg == __value.arg


to:

Total time: 0.0412131 s
File: /home/rvd/src/roelofvandijk/tinygrad/tinygrad/ops.py
Function: __eq__ at line 35

    35                                             @profile
    36     46378      41213.1      0.9    100.0    def __eq__(self, __value: object) -> bool: return isinstance(__value, LazyOp) and self.op is __value.op and self.src == __value.src and self.arg == __value.arg

```

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -32,10 +32,7 @@ class LazyOp:
     except AttributeError: pass
 
   def __repr__(self): return f"LazyOp(op={self.op}, src={self.src}, arg={self.arg})"
-  def __eq__(self, __value: object) -> bool:
-    if __value.__class__ is not LazyOp: return False
-    __value = cast(LazyOp, __value)
-    return self.op == __value.op and self.src == __value.src and self.arg == __value.arg
+  def __eq__(self, __value: object) -> bool: return isinstance(__value, LazyOp) and self.op is __value.op and self.src == __value.src and self.arg == __value.arg
   def __hash__(self) -> int: return hash((self.op, self.src, self.arg))
   @property
   def key(self): return (self.op, tuple(map(lambda x: getattr(x, "key", x), self.src)), getattr(self.arg, "key", self.arg))


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/ops.py                            1               4
```

tinygrad/ops.py
```diff
@@ -32,10 +32,7 @@ def __init__(self, op: Op, src: Tuple[Union[LazyOp, LazyBuffer], ...], arg: Any
     except AttributeError: pass
 
   def __repr__(self): return f"LazyOp(op={self.op}, src={self.src}, arg={self.arg})"
-  def __eq__(self, __value: object) -> bool:
-    if __value.__class__ is not LazyOp: return False
```
g1y5x3, line 36: Is the `cast` here is unnecessary due to the `if` statement previously that check its class type? There's a few more `cast(LazyOp, )` as well and they fall under different condition?

roelofvandijk: The cast here was necessary because `.__class__` was used instead of `isinstance`. Mypy only understands `isinstance`. Without either the cast or `isinstance`, mypy will complain about the attribute access of `__value`.

There are some other places in the code where `isinstance` was replaced by `.__class__ is ...` and then a `cast`, which together is both slower and more complex than just using `isinstance`. I'm slowly undoing those.

roelofvandijk: E.g.

```
 def exec_ast(self, ast:LazyOp, output=None, context=None, **kwargs):
    if TernaryOps.MULACC in self.fxn_for_op and ast.op == ReduceOps.SUM and ast.src[0].__class__ is LazyOp and ast.src[0].op == BinaryOps.MUL:
      ast = LazyOp(TernaryOps.MULACC, cast(LazyOp, ast.src[0]).src, ast.arg)
```

Same thing.

roelofvandijk: I can revert the others, separate MR though.

roelofvandijk: See https://github.com/tinygrad/tinygrad/pull/1694

g1y5x3: Thanks for the explanation

PR #1694, opened 2023-08-27
perf: remove cast and revert back to isinstance

roelofvandijk:
The cast here was necessary because .__class__ was used instead of isinstance. Mypy only understands isinstance. Without either the cast or isinstance, mypy will complain about the attribute access of __value.

Together they are both slower and more complex than just using isinstance.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                           4               3
tinygrad/ops.py                            5               6
total                                      9               9
```

wozeparrot:
conflict

roelofvandijk:
resolved

PR #1699, opened 2023-08-28
annotate failed webgpu tests

chenyuxyz:
WEBGPU is not tested in CI due to flakiness, and many newly added tests would fail with WEBGPU backend. This PR skips those tests and annotates the failure reason, so that the CI WEBGPU command can pass locally.

Previous [flaky runs](https://github.com/tinygrad/tinygrad/actions/runs/5780589178/job/15664459463) had some other test failure, probably not ready to re-enable CI test before understanding the issue. Maybe we want to re-enable `Build WEBGPU Efficientnet` in CI since that's mentioned in the release note?

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -152,6 +152,7 @@ class TestInt32Dtype(unittest.TestCase):
   @unittest.skipIf(Device.DEFAULT == "WEBGPU", "webgpu does not support int64")
   def test_int32_upcast_int64(self): _test_ops(a_dtype=dtypes.int32, b_dtype=dtypes.int64, target_dtype=dtypes.int64)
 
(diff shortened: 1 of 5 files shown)


geohot:
My understanding is that the failure of webgpu is caused by the underlying metal being either broken or doing something we don't understand.

chenyuxyz:
the skipped tests here are tests added after we disabled webgpu tests in CI (except the gc one), the test author would either need to skip them or update webgpu if we had webgpu CI test.

I thought we previously disabled webgpu in CI due to flakiness on some tests, but I cannot reproduce the flakiness issue. Assuming we want webgpu CI tests back, the best I can do is to update the test so that the CI test script can pass locally (this PR), then I can enable CI on different master commits on my local fork to see if I can reproduce the issue.

chenyuxyz:
closing - probably get back to it when we have better metal ci coverage

PR #1705, opened 2023-08-29
Small Quality of Life Improvement on Metal

FrostyTheSouthernSnowman:
I was installing my personal fork of tinygrad on metal. As I was installing the packages, I kept getting an error about not being able to import Metal. I have added a try-catch statement to explain how to install the required packages instead of requiring on developers to read the code themselves.

If you think this isn't necessary, just close it. I'm fine either way.


wozeparrot:
Yea this should just be something else, maybe in the readme, but def not in the code.
Also you should be able to install the deps with something like `pip install -e .[metal]`

Issue #1707, opened 2023-08-29
LLaMA is failing in Benchmark CI but still getting a checkmark


geohot:

prusnak:
has this been fixed by https://github.com/tinygrad/tinygrad/commit/ab9b9ff3e29e4ea22e17a6f3ca0a73f58db30c9c ?

PR #1709, opened 2023-08-29
pipefail benchmark

wozeparrot:
Closes: #1707

If you specify `shell: bash` it will add `-o pipefile`.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference


geohot:
This only changes it for AMD. Is there a better way than tee?

wozeparrot:
we can do a redirect like `python > file.txt`, that doesn't call another program so the python exit code should be the one that bubbles

geohot:
Hmm, but now I can't see it when I click!

PR #1711, opened 2023-08-29
tinyboard integration

nimlgen:
Tinygrad approach stage of mlperf trainings, so here is tinyboard to visualize them.

[Tinyboard](https://github.com/nimlgen/tinyboard) is a React+Flask tesorboard analogue for tinygrad. It features some chart visualization, function inspector and kernels.

Visualisation adds an API to render time series from tinygrad into tinyboard. Currently this is implemeted for hlb_cifar as example. This is as simple as `tinyboard_log_graph("Train stat", "line", [[float(loss_cpu)]], graphinfo={'series_names': ['loss']})`.

Function inspector adds a decorator to inspect any python function. It collects tensor operations generated by the function, mlops, lazyops ([screenshot](https://github.com/nimlgen/tinyboard/blob/master/screenshots/func_inspect.png)).

Tinyboard collects stat about kernels execution (the same as DEBUG>=2). Later this might be extended with some other information collected for each kernel (like uast graph).

If you think that this should not be a part of tinygrad, feel free to close this PR.

### DIFF
--- a/docs/env_vars.md
+++ b/docs/env_vars.md
@@ -56,6 +56,9 @@ CL_EXCLUDE          | [name0,name1] | comma-separated list of device names to ex
 CL_PLATFORM         | [# >= 0]   | index of the OpenCL [platform](https://documen.tician.de/pyopencl/runtime_platform.html#pyopencl.Platform) to run on. Defaults to 0.
 RDNA                | [1]        | enable the specialized [RDNA 3](https://en.wikipedia.org/wiki/RDNA_3) assembler for AMD 7000-series GPUs. If not set, defaults to generic OpenCL codegen backend.
 PTX                 | [1]        | enable the specialized [PTX](https://docs.nvidia.com/cuda/parallel-thread-execution/) assembler for Nvidia GPUs. If not set, defaults to generic CUDA codegen backend.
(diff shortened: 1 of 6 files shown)


nimlgen:
Also if tinyboard is fine with you, I also would be happy to move the tinyboard repo to tiny corp if you do not mind.

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        1               0
tinygrad/ops.py                           10               3
tinygrad/tensor.py                         5               2
total                                     16               5
```

geohot:
I agree something like this needs to exist.

flask I like, but I find react bloated and unreadable.

tinygrad should really have generic APIs to expose this sort of debugging data.

geohot:
Wow this inspector looks impressive though! I've been looking for something that could tie the graph node back to the code.

I think we need to clean up and make clear what's being exported

nimlgen:
You mean you want a kind of logging systems for tinygrad? Something like `tinygrad_log(event_name, data)`? So we could attach callbacks to these events (like tinyboard attached to some logging API and/or typical stdout printer like now)?

I think it's possible to replace react. @dc-dc-dc has suggested several options will see how they fits.

geohot:
Going to close. I'm open to something like this, but it has to be more of a generic framework, perhaps using the caching infrastructure.

PR #1712, opened 2023-08-29
stable diffusion benchmark artifact

wozeparrot:
do we still need the `time`?


geohot:
You can push yourself to update_benchmark if you want to test, that's a special branch that'll run the hosted CI.

Meh, time is probably mostly covered

PR #1713, opened 2023-08-29
Fix: Stable Diffusion index

ZenginU:
This should fix the index problem in the stable diffusion


geohot:
Can you post a before and after picture?

ZenginU:
Afterimage
Beforeimage
Commit before the errorimage

Actually when compared not sure if I fixed the problem. It can be seen that it is far better but not sure if bugless

ZenginU:
Tensor._seed = 0

Master 
image

This branch, bisected branch,
image

PR #1716, opened 2023-08-30
Change static methods to class methods

### LINKED ISSUE #1695: Classmethods instead of staticmethods
I noticed that many methods of the Tensor class are marked as static methods, even though they return new Tensor instances. These methods seem more like class methods than static methods. Converting them to class methods would:

1. Make it more clear that those methods create new Tensors
2. Will support subclassing Tensor

Is there any reason they are currently static methods? If there are no arguments against changing those static methods to class methods, I can create a PR for it.

yochem:
Static methods that use their own class, e.g. to create a new instance of its own class should be class methods. This makes the methods also work on subclasses.

Closes #1695.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -368,11 +368,12 @@ class Tensor:
       s[dim] = (k, shape_cumsum[-1] - k - shp)
     return reduce(Tensor.__add__, [arg.pad(tuple(s)) for arg,s in zip(catargs, slc)])
 
-  @classmethod
-  def stack(cls, tensors, dim=0):
+  @staticmethod
+  def stack(tensors, dim=0):
+    first = tensors[0].unsqueeze(dim)
     unsqueezed_tensors = [tensor.unsqueeze(dim) for tensor in tensors[1:]]
     # checks for shapes and number of dimensions delegated to cat
-    return cls.cat(*unsqueezed_tensors, dim=dim)
+    return first.cat(*unsqueezed_tensors, dim=dim)
 
   def repeat(self, repeats):
     base_shape = (1,) * (len(repeats) - self.ndim) + self.shape
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -122,85 +122,85 @@ class Tensor:
 
   # ***** creation llop entrypoint *****
 
-  @staticmethod
-  def _loadop(op, sz, device:Optional[str]=None, dtype:Optional[DType]=None, arg=None, **kwargs):
-    return Tensor(LazyBuffer.loadop(op, [sz], Tensor.default_type if dtype is None else dtype, Device.canonicalize(device), arg), dtype=dtype, device=device, **kwargs)
+  @classmethod
+  def _loadop(cls, op, sz, device:Optional[str]=None, dtype:Optional[DType]=None, arg=None, **kwargs):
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/lazy.py                           3               3
tinygrad/tensor.py                        54              55
total                                     57              58
```

yochem:
I also want to type-hint the class methods. In Python 3.11, this is done using [`typing.Self`](https://docs.python.org/3/library/typing.html#typing.Self). <3.11, this can be done by just specifying the class name:

e.g. in Tensor:
```python
def eye(cls:Tensor, dim:int, **kwargs) -> Tensor:
    ...
```

But this doesn't work for subclasses, as they return the subclass of Tensor, and not Tensor. So this is the recommended way for type hinting `cls` and the return value of the class method:

```python
T = TypeVar('T', bound=Tensor)
def eye(cls:T, dim:int, **kwargs) -> T:
    ...
```

This however (1) adds a line to the codebase and (2) is probably not very clear to all users what it does. We could of course use a better name than `T`, but still.

What do you think? Leave the type hints for now, or go with one of the two options above?

roelofvandijk:
How about `TensorInstance` or `TensorType`?

yochem:
No it would not. By using TypeVar: If you call `Tensor.eye()`, the TypeVar type hint understands that `cls` is of type Tensor and the return value of `Tensor.eye()` is Tensor as well. Only if we subclass Tensor as e.g. MyTensor, then `MyTensor.eye()` will return an instance of type MyTensor.

prusnak:
> How about `TensorInstance` or `TensorType`?

the following would be the most standard way and also comprehensible

``` python
TensorType = TypeVar("TensorType", bound=Tensor)
```

yochem:
Went with `TensorType` as proposed by @prusnak.

yochem:
Because other methods are used in the classmethods (e.g. `.reshape()` in `.empty()`), the typevar doesn't work unless we change all methods returning Tensor to return `TensorType` instead. I think this will make things only less clear. I propose to type annotate them for now as `Tensor`, which doesn't support the right typehints for subclasses. When support for 3.8 is dropped, it can be changed to be type hinted as `typing.Self`, as supported by 3.11+.

tinygrad/tensor.py
```diff
@@ -31,6 +31,7 @@ def apply(fxn:Type[Function], *x:Tensor, **kwargs) -> Tensor:
 import tinygrad.mlops as mlops
 
 # **** start with two base classes, Tensor and Function ****
+TensorType = TypeVar("TensorType", bound="Tensor")
```
geohot, line 34: Not worth the complexity. Subclassing Tensor isn't really supported

geohot:
Going to close this. Sorry I said it was fine, subclassing Tensor isn't really supported, and now with the custom type it's adding complexity.

PR should be "speed (either emitted or the Python), conceptual simplicity, mlperf training, and bugfixes"

yochem:
And without the custom types and just the staticmethod -> classmethod conversion ([34616296c7d70fb68a6f0d389fd5fd2a6a844718](https://github.com/tinygrad/tinygrad/pull/1716/files/34616296c7d70fb68a6f0d389fd5fd2a6a844718))? I agree that the custom type is adding too much complexity and Tensors being subclassable is not really the goal.

Otherwise sorry for your time! Hope to provide more useful PR's in the future that align with the PR goals :)

geohot:
Probably still not worth it. If there's a real use case for subclassing Tensor we can revisit

PR #1721, opened 2023-08-31
perf: canonicalize less, canonicalize faster

roelofvandijk:
Would this be acceptable? Cache the string op and define a tuple of known canonicals.

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -67,10 +67,14 @@ class LazyOp:
   def stride(self, _): raise NotImplementedError
 
 # **************** Device ****************
+@functools.lru_cache(maxsize=None)  # this class is a singleton, pylint: disable=method-cache-max-size-none
+def _canonicalize(device: str) -> str: return (device.split(":", 1)[0].upper() + ((":"+device.split(":", 1)[1]) if ':' in device else '')).replace(":0", "")
 
 class _Device:
-  def __init__(self) -> None: self._buffers: List[str] = [x.stem[len("ops_"):].upper() for x in (pathlib.Path(__file__).parent/"runtime").iterdir() if x.stem.startswith("ops_")]
-  def canonicalize(self, device:Optional[str]) -> str: return (device.split(":", 1)[0].upper() + ((":"+device.split(":", 1)[1]) if ':' in device else '')).replace(":0", "") if device is not None else self.DEFAULT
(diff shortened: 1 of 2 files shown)


tinygrad/ops.py
```diff
@@ -67,10 +67,14 @@ def shrink(self, _): raise NotImplementedError
...
 # **************** Device ****************
+@functools.lru_cache(maxsize=None)  # this class is a singleton, pylint: disable=method-cache-max-size-none
+def _canonicalize(device: str) -> str: return (device.split(":", 1)[0].upper() + ((":"+device.split(":", 1)[1]) if ':' in device else '')).replace(":0", "")
 
 class _Device:
-  def __init__(self) -> None: self._buffers: List[str] = [x.stem[len("ops_"):].upper() for x in (pathlib.Path(__file__).parent/"runtime").iterdir() if x.stem.startswith("ops_")]
-  def canonicalize(self, device:Optional[str]) -> str: return (device.split(":", 1)[0].upper() + ((":"+device.split(":", 1)[1]) if ':' in device else '')).replace(":0", "") if device is not None else self.DEFAULT
+  def __init__(self) -> None:
+    self._buffers: List[str] = [x.stem[len("ops_"):].upper() for x in (pathlib.Path(__file__).parent/"runtime").iterdir() if x.stem.startswith("ops_")]
+    self.canonicals = ("METAL", "CUDA", "GPU", "CPU")
```
wozeparrot, line 76: there are a lot more canonicals than this, it should be possible to automatically generate the list of canonicals.

tinygrad/tensor.py
```diff
@@ -40,7 +40,7 @@ class Tensor:
   default_type: ClassVar[DType] = dtypes.float32
   def __init__(self, data:Union[int, float, list, LazyBuffer, np.ndarray], device:Optional[str]=None, dtype:Optional[DType]=None, requires_grad:Optional[bool]=None):
     assert dtype is None or isinstance(dtype, DType), f"invalid dtype {dtype}"
-    device = Device.canonicalize(device)
+    device = Device.DEFAULT if device is None else device if device in Device.canonicals else Device.canonicalize(device)
```
wozeparrot, line 43: isn't most of this duplicated with what happens in `Device.canonicalize`?

geohot:
There's no way `ops.py` should have a list like that.

PR #1732, opened 2023-09-01
search for grouped reduces

nimlgen:
A bit better search of shapes using dividers instead of consts + search for group_for_reduce dim.

### DIFF
--- a/tinygrad/codegen/search.py
+++ b/tinygrad/codegen/search.py
@@ -3,8 +3,11 @@ import time
 from tinygrad.codegen.linearizer import Linearizer
 from tinygrad.helpers import DEBUG, prod, getenv
 
-UPCASTS = [1,2,3,4,5,6,7,8]
-LOCALS = [1,2,3,4,5,6,7,8,16,24,32]
+def get_divisors(n, min_div = 1, max_div = 512):
+  if min_div > 1: yield 1
+  for d in range(min_div, min(max_div, n//2) + 1):
+    if n % d == 0: yield d
+
 def kernel_optimize_search(k:Linearizer, create_k:Callable[[], Linearizer], to_prg, baseline):
   import nevergrad as ng
   def opt(x):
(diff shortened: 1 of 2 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/optimizer.py              8               3
tinygrad/codegen/search.py                20               5
total                                     28               8
```

tinygrad/codegen/search.py
```diff
@@ -1,10 +1,23 @@
 from typing import Callable
 import time
+from functools import lru_cache
 from tinygrad.codegen.linearizer import Linearizer
 from tinygrad.helpers import DEBUG, prod, getenv
 
-UPCASTS = [1,2,3,4,5,6,7,8]
-LOCALS = [1,2,3,4,5,6,7,8,16,24,32]
+@lru_cache(maxsize=32)
+def get_divisors(n, min_div = 1, max_div = 512):
```
prusnak, line 8: If you want to generate list of divisors from given range, why not just doing this?

``` python
def get_divisors(n, min_div = 1, max_div = 512):
    for i in range(min_div, max_div + 1):
        if n % i == 0:
            yield i
```

Unless you want to use really really large numbers of max_div, you really don't have to overoptimize and overengineer with LRU cache.

tinygrad/codegen/search.py
```diff
@@ -3,8 +3,11 @@
 from tinygrad.codegen.linearizer import Linearizer
 from tinygrad.helpers import DEBUG, prod, getenv
 
-UPCASTS = [1,2,3,4,5,6,7,8]
-LOCALS = [1,2,3,4,5,6,7,8,16,24,32]
+def get_divisors(n, min_div = 1, max_div = 512):
+  if min_div > 1: yield 1
+  for d in range(min_div, max_div + 1):
```
gijskoning, line 8: Dividing by zero gives error and divide by one doesn't make sense?
(Removing line 9 too)
```suggestion
  assert min_div > 0
  for d in range(min_div, max_div + 1):
```

tinygrad/codegen/search.py
```diff
@@ -3,8 +3,11 @@
 from tinygrad.codegen.linearizer import Linearizer
 from tinygrad.helpers import DEBUG, prod, getenv
 
-UPCASTS = [1,2,3,4,5,6,7,8]
-LOCALS = [1,2,3,4,5,6,7,8,16,24,32]
+def get_divisors(n, min_div = 1, max_div = 512):
+  if min_div > 1: yield 1
```
gijskoning, line 7: Dividing by zero gives error?
(Removing line 9 too)
```suggestion
  assert min_div > 0
```

tinygrad/codegen/search.py
gijskoning, line 8:
```suggestion
  for d in range(min_div, min(max_div + 1, n)):
```

PR #1669, opened 2023-08-25
added visitor pattern

Addiictet:
Since refactoring the linearizer seems to be important right now, I wanted to get more familiar with it.
Found a TODO which I could easily do and thought it would be good to implement.

Removes many unnecessary isinstance calls, adds one line of code in total.

First time contributing - I hope I did everything correctly.

PR #1733, opened 2023-09-01
visitor pattern for sym_infer and unit tests

chenyuxyz:
similar to #1669, also added unit tests for `sym_infer`

should be faster in theory due to fewer `int` case check , but I could not measure a meaningful end to end difference. Also verified that GPU jitted llama still is < 5ms.


geohot:
Can you document expand and infer? What's the difference?

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                 7               9
```

chenyuxyz:
sure, like this?

PR #1723, opened 2023-08-31
remove Token class

geohot:
No more stupid tokens!

PR #1741, opened 2023-09-02
Remove Token from abstractions.py

crankygrumpster:
In #1723, the Token class was removed.


docs/abstractions.py
```diff
@@ -245,14 +245,11 @@ def toCPU(self): return self._buf
 
 class UOps(Enum): LOOP = auto(); DEFINE_LOCAL = auto(); LOAD = auto(); ALU = auto(); CONST = auto(); ENDLOOP = auto(); STORE = auto();
 
-class Token:
-  name: str
-
 class UOp:
   uop: UOps
-  out: Optional[Token]
```
geohot, line 253: add dtype here

crankygrumpster: Added.

PR #1742, opened 2023-09-02
Add item()

waifairer:


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/runtime/lib.py                    6               1
tinygrad/tensor.py                         5               0
total                                     11               1
```

g1y5x3:
What's the actual use case for it? As far for the test you added, it's pretty much the same as using `.numpy()`?

waifairer:
> What's the actual use case for it? As far for the test you added, it's pretty much the same as using `.numpy()`?

The main use is for bool masks which frequently do this sort of thing:
```
    ind = Tensor.arange(mask.shape[0])
    nz = mask.sum().numpy().item()
    mask = mask * ind
    idx = mask.numpy().argsort()[-int(nz):]
```

It's exactly the same as using `numpy().item()`

waifairer:
I see that `toCPU` is typed to return an ndarray. That means that this breaks type structure.

waifairer:
> What's the actual use case for it? As far for the test you added, it's pretty much the same as using `.numpy()`?

BTW, I ultimately found that everywhere I wanted to use this, I could get away with cumsum, arange, and 'where'. So you were right to question this and I'm staying in JIT longer for having got rid of it.

Argsort() is painful, though. We could definitely use a native implementation of that.

PR #1744, opened 2023-09-03
Test KOPT in CI

chenyuxyz:
Add env variable `BUDGET` for search budget. Test in CI with `DEBUG=1` so we can check if it runs.

It surfaces 2 problems:
(1) llama generation is not fully covered in test, currently KOPT fails because of symbolic offset, but CI does not test it. I plan to fix this and update the test in a separate PR.
(2) cifar hangs with current KOPT. People reported it in discord. Skip in CI for now.

### DIFF
--- a/tinygrad/codegen/search.py
+++ b/tinygrad/codegen/search.py
@@ -38,7 +38,8 @@ def kernel_optimize_search(k:Linearizer, create_k:Callable[[], Linearizer], to_p
   if not opts: return "BASELINE"
   search_space = prod([len(x.choices) for x in opts])
   st = time.perf_counter()
-  optimizer = ng.optimizers.NGOpt(parametrization=ng.p.Tuple(*opts), budget=min(search_space, 200))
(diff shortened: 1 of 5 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/search.py                 2               1
```

tinygrad/codegen/search.py
```diff
@@ -38,7 +38,8 @@ def opt(x):
   if not opts: return "BASELINE"
   search_space = prod([len(x.choices) for x in opts])
   st = time.perf_counter()
-  optimizer = ng.optimizers.NGOpt(parametrization=ng.p.Tuple(*opts), budget=min(search_space, 200))
+  budget = int(getenv("BUDGET", 200))
```
geohot, line 41: You don't need the int here, it'll take the dtype from the default.

chenyuxyz: oh you are right - updated. wino PR had the int too and I got the wrong impression

PR #1754, opened 2023-09-04
Extend graph export through multi-format/file

DHDev0:
Add multi format and multi file output for graph.
- Splits GRAPHPATH by ":" or any special character, filters formats, and execute Graphviz commands for each valid format.
- Example:
 "GRAPH=1 GRAPHPATH=to/path/namefile.kol:.svg:..png:jpg:imap_np" -> execute Graphviz conversion .dot to svg,png and imap_np
- not working for xdot1.2 and xdot1.4


DHDev0:
Maximizing the reference frame of a complex graph look useful. If they implement graph in all this way it's because there is a demand.

PR #1759, opened 2023-09-04
fix: remove unused function

roelofvandijk:


geohot:
Can the linter check for this?

roelofvandijk:
> Can the linter check for this?

In theory, yes. I used (Vulture)[https://github.com/jendrikseipp/vulture]. It's not 100% reliable, so I would advise against it.

PR #1763, opened 2023-09-04
KeyError fixed.

tomtom-95:
KeyError fixed, now the test does not fail


geohot:
Huh, what was failing?

tomtom-95:
TestRealWorld.test_llama
If I run python -m pytest -x it's the first test that fails

PR #1770, opened 2023-09-05
Metal stuff pip install on default when on Darwin

geohotstan:
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies

image

used syntax from here :s

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setup(name='tinygrad',
         "Programming Language :: Python :: 3",
         "License :: OSI Approved :: MIT License"
       ],
-      install_requires=['numpy', 'requests', 'pillow', 'tqdm', 'networkx', 'pyopencl', 'PyYAML'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,15 +19,14 @@
         "Programming Language :: Python :: 3",
         "License :: OSI Approved :: MIT License"
       ],
-      install_requires=['numpy', 'requests', 'pillow', 'tqdm', 'networkx', 'pyopencl', 'PyYAML'],
+      install_requires=["numpy", "requests", "pillow", "tqdm", "networkx", "pyopencl", "PyYAML", "pyobjc-framework-Metal;platform_system=='Darwin'", "pyobjc-framework-Cocoa;platform_system=='Darwin'", "pyobjc-framework-libdispatch;platform_system=='Darwin'"],
```
wozeparrot, line 22: this line is getting a bit long with these, I think its probably fine just to newline split it

PR #1774, opened 2023-09-05
llama jit is default, print tok/sec

geohot:


test/external/external_multi_gpu.py
```diff
@@ -14,7 +14,7 @@
   sz = 1024*1024*256  # 1 GB
   #sz = 1024*64
 
-  with Timing("CPU creation: ", on_exit=lambda x: f", {(sz*4*2)/x:.2f} GB/sec"):
+  with ("CPU creation: ", on_exit=lambda x: f", {(sz*4*2)/x:.2f} GB/sec"):
```
prusnak, line 17: Was this change intentional? It does not look right.

prusnak: PR in https://github.com/tinygrad/tinygrad/pull/1777

PR #1778, opened 2023-09-05
cuda launch bounds, use ptrdiff_t instead of int for indices

chaosagent:
For large convs and other operations, int is not large enough for indices. I ran into this when playing with unet3d


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/renderer/cstyle.py                1               1
```

geohot:
Hmm, so this probably slows things down. Do you have a failing example?

geohot:
Going to close. Either need to see that this doesn't hit perf or gate it for when it's really needed

PR #1785, opened 2023-09-06
Use correct dtype in Tensor when data is an ndarray

badcc:
Upon tryign to run codellama example with GPU=1 on M2 Macbook Air 16GB, I was getting following errors
```
AttributeError: 'E_12_4_8_16_2_4' was not found as a program info attribute or as a kernel name
pyopencl._cl.LogicError: clCreateKernel failed: INVALID_KERNEL
```

The kernel had was taking in a `double` parameter, e.g.
`__kernel void E_12_4_8_16_2_4(__global float* data0, const __global float* data1, const __global double* data2) {`

but was assigning to float value
`float val4 = data2[(gidx1*32)+(lidx3*2)+(gidx0*128)];`

and after some digging I realized Tensor was loading in ndarray with incorrect dtype


tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
Neat, can you add a test to make sure this doesn't regress?

geohot:
Nice!

Issue #1757, opened 2023-09-04
CMPLT should output a bool in UOps


geohot:
Then be upcasted later

segf00lt:
Upcast to dtypes.bool?

geohotstan:
Is this the right idea?
```
==== code ====
a = (Tensor(3) > Tensor(4))
a.realize()

==== uops ====
[   0 UOps.DEFINE_GLOBAL  : ptr.dtypes.int            []                               ('data0', dtypes.int),    
    1 UOps.CONST          : dtypes.float              []                               4.0,    
    2 UOps.CONST          : dtypes.float              []                               3.0,    
    3 UOps.ALU            : dtypes.bool               [1, 2]                           BinaryOps.CMPLT,    
    4 UOps.CONST          : dtypes.int                []                               0,    
    5 UOps.STORE          :                           [0, 4, 3]                        None]

==== rendered cstyle code ====

*** CLANG ***
 void E_(float* restrict data0) {
data0[0] = (bool)(4.0f<3.0f);
}

*** METAL ***
#include <metal_stdlib>
using namespace metal;
kernel void E_(device float* data0, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
*(data0+0) = (bool)(4.0f<3.0f);
}

```
So data0[0] would be a bool
but since data0 has dtype float the bool value gets upcasted to float?
This is my first time digging into tinygrad internals and I don't know any C...
Trying to learn

chenyuxyz:
this is done and asserts in linearizer https://github.com/tinygrad/tinygrad/blob/17658499379eef6c1f67214295ecb3831bdce4c0/tinygrad/codegen/linearizer.py#L482

PR #1790, opened 2023-09-06
CMPLT output bool in ops_cpu.py and ops_torch.py

segf00lt:
Patch for #1757.


tinyb0t:
Changes made in `tinygrad/`:
```
```

geohot:
So that issue is about the compiled backends, but it should really be all of them. This is a more in depth project than a two line change though, it needs tests and to carefully think through the desired behavior

segf00lt:
Right, at first I thought since you refer to UOps in the issue it would be something higher level. But having taken another look the I see renderers cast to a float.

I assume this was done because the backends are statically typed and so just having CMPLT return a bool won't work(?)

Why is it better to output a bool rather than a 0.0 or 1.0 valued float? Does it have to do with the code generator needing a way of keeping track of what values the float could have and thus be able to optimize around that?

geohot:
No updates in 2 weeks, closing. Though I like this idea

PR #1796, opened 2023-09-06
use "a in (b,c)" instead of "a == b or a == c"

prusnak:
it's more obvious to understand what the code does


geohot:
Not a clear win

PR #1797, opened 2023-09-06
use class Foo: instead of class Foo():

prusnak:


wozeparrot:
Is there a way this can be enforced in the linter?

prusnak:
> Is there a way this can be enforced in the linter?

- black formatter does this
- ruff linter does this - https://beta.ruff.rs/docs/rules/unnecessary-class-parentheses/

afaik neither flake8 nor pylint recognize this

prusnak:
Added ruff as linter in d08a80d86d24eae14e861c05933abf1b5ab344e0 - rule `UP039` checks for `unnecessary-class-parentheses`

PR #1798, opened 2023-09-06
fix llama on apple silicon, keep freqs type consistent

badcc:
`precompute_freqs_cis` was returning a numpy float64 (double) when all we need is float32 (and all mac can support)

This seems consistent now with meta's implementation:
```
def precompute_freqs_cis(dim: int, end: int, theta: float = 10000.0):
    freqs = 1.0 / (theta ** (torch.arange(0, dim, 2)[: (dim // 2)].float() / dim))
    t = torch.arange(end, device=freqs.device)  # type: ignore
    freqs = torch.outer(t, freqs).float()  # type: ignore
    freqs_cis = torch.polar(torch.ones_like(freqs), freqs)  # complex64
    return freqs_cis
```

notably: the `.float()` for freqs after torch.outer.

---

now GPU=1, and also ofc METAL=1 work for llama example


geohot:
Does someone want to move that function into tinygrad?

Issue #1805, opened 2023-09-06
Stable Diffusion is swapping on 8GB Mac


geohot:
```
amanda@smol:~/tinygrad$ python3 examples/stable_diffusion.py --timing
ram used:  4.26 GB, cond_stage_model.transformer.text_model.final_layer_norm.bias: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1131/1131 [00:01<00:00, 615.76it/s]
loaded weights in 1848.34 ms, 4.26 GB loaded at 2.31 GB/s
got CLIP context (1, 77, 768)
got unconditional CLIP context (1, 77, 768)
running for [1, 201, 401, 601, 801] timesteps
  4 801:   0%|                                                                                                                                                                                                                                           | 0/5 [00:00<?, ?it/s]
step in 10566.70 ms
  3 601:  20%|█████████████████████████████████████████████▍                                                                                                                                                                                     | 1/5 [00:10<00:42, 10.57s/it]
step in 10049.36 ms
  2 401:  40%|██████████████████████████████████████████████████████████████████████████████████████████▊                                                                                                                                        | 2/5 [00:20<00:30, 10.26s/it]
step in 20548.51 ms
  1 201:  60%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏                                                                                          | 3/5 [00:41<00:29, 14.96s/it]
step in 24554.60 ms
  0   1:  80%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▌                                             | 4/5 [01:05<00:18, 18.75s/it]
step in 29281.73 ms
  0   1: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [01:35<00:00, 19.00s/it]
decode (1, 512, 64, 64)
decode (1, 512, 128, 128)
decode (1, 512, 256, 256)
decode (1, 256, 512, 512)
(512, 512, 3)
saving /var/folders/5c/fbxt_6zj5_3cm0sshw2_zck00000gn/T/rendered.png
```

At least I think so. It's really slow, yet only about 3 of those seconds are spent on the GPU. @nimlgen?

geohot:
Yea, something is wrong with the RAM tracking. fp16 is fast, but it says it uses more RAM

```
[...]

PR #1814, opened 2023-09-07
patch to remove hack from stable_diffusion.py

segf00lt:
#1805

### DIFF
--- a/test/test_assign.py
+++ b/test/test_assign.py
@@ -70,7 +70,7 @@ class TestAssign(unittest.TestCase):
     oba1 = a.lazydata.output_buffer
     a.assign(a.cast(dtypes.float16).realize())
     oba2 = a.lazydata.output_buffer
-    assert oba1 == None and oba2 == None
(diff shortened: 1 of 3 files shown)


tinyb0t:
Changes made in `tinygrad/`:
```
```

test/test_assign.py
```diff
@@ -63,5 +64,15 @@ def test_post_permuted_assignment(self):
 
   # TODO: is there a way to sneak in a permute such that it returns the wrong answer?
 
+  def test_cast_assignment(self):
+    a = Tensor(np.arange(N*N, dtype=np.float32)).reshape(N,N)
+    a.realize()
+    oba1 = a.lazydata.output_buffer
+    a.assign(a.cast(dtypes.float64).realize())
```
wozeparrot, line 71: float64 doesn't have full platform support, maybe use one of the int dtypes instead?

segf00lt: I'll change to int32.

geohot:
Nice find!

PR #1817, opened 2023-09-08
Printtree catch cycles

DKormann:
catch cycles in lazydata and display them with references

```python
a = Tensor.normal()
print_tree(a + a)
```

```
  0 ━┳ ADD 
  1  ┣━┳ MUL 
  2  ┃ ┣━┳ SIN 
  3  ┃ ┃ ┗━┳ SUB  
  4  ┃ ┃   ┣━━ CONST 1.5707963267948966
  5  ┃ ┃   ┗━┳ MUL  
  6  ┃ ┃     ┣━┳ RESHAPE 
  7  ┃ ┃     ┃ ┗━┳ SHRINK ((0, 1),) 
  8  ┃ ┃     ┃   ┗━━ RAND 1694154148 
  9  ┃ ┃     ┗━━ CONST 6.283185307179586 
 10  ┃ ┗━┳ SQRT  
 11  ┃   ┗━┳ MUL  
 12  ┃     ┣━┳ MUL 
 13  ┃     ┃ ┣━┳ LOG2 
 14  ┃     ┃ ┃ ┗━┳ SUB  
 15  ┃     ┃ ┃   ┣━━ CONST 1
 16  ┃     ┃ ┃   ┗━┳ RESHAPE  
 17  ┃     ┃ ┃     ┗━┳ SHRINK ((1, 2),) 
 18  ┃     ┃ ┃       ┗━⬆︎ goto 8: RAND 
 19  ┃     ┃ ┗━━ CONST 0.6931471805599453 
 20  ┃     ┗━━ CONST -2 
 21  ┗━⬆︎ goto 1: MUL 
```

i was visualizing rnn loss.grad and the number of printed lines grows with n square with sequence length because auf cyclic depps.  when the identical lazydata gets used in the AST we now show a refernce instead this is also a closer representation of the computation flow.


DKormann:
not sure this is neccessary but i added function to get specific lazydata by line number

in above example you could do: 
`get_tree(a+a, 19)` 
`LazyOp(op=LoadOps.CONST, src=(), arg=0.6931471805599453)`

extra/utils.py
```diff
@@ -209,12 +209,35 @@ def get_child(parent, key):
...
+circle_tracker:dict[int,tuple[int,any]] = {}
+
+def _tree(lazydata:Union[LazyBuffer, LazyOp]):
+  global ctr
+  ctr += 1
+  if type(lazydata) == LazyBuffer and not (lazydata.realized): lazydata = lazydata.op
+  if type(lazydata) == LazyBuffer : return [f"━━ realized {lazydata.dtype.name} {lazydata.shape} "]
+
+  if id(lazydata) in circle_tracker: return [f"━⬆︎ goto {circle_tracker[id(lazydata)][0]}: {lazydata.op.name}"]
+  circle_tracker[id(lazydata)] = (ctr, lazydata)
```
nimlgen, line 225: Holding lazydata here we do not allow some buffers to be freed. Clear `circle_tracker` at the end of `get_tree`?

DKormann: okay thanks for the tip

geohot:
This makes this more confusing. Do we need it?

DKormann:
Yeah it's a bit more complicated than I wanted it. I could maybe shave one or two lines.
The situation where I needed it:
I build a tinyyy rnn and for seq len 10 the tree for loss suddenly includes 80000 nodes. Because everything relates to everything elsse.

Alternative solutions: 
limit depth of the displayed tree.
Limit number of displayed nodes.
This would clear code up a lot but loose informatio.

Also removing get_tree will make the other ones a little more simple but I think the feature is nice because it makes it simple to retrieve eg. node from line 24 in the tree super simple.

PR #1819, opened 2023-09-08
Move var_vals from ShapeTracker to LazyBuffer

chenyuxyz:
Make ShapeTracker stateless and cache friendly.

Need to make a behavior change that reshape from symbolic shape to an int shape not erasing the `var_vals`. This is needed in the llama jitted path because we are using symbolic offset.


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/jit.py                            1               1
tinygrad/lazy.py                          28              15
tinygrad/shape/shapetracker.py             5              16
total                                     37              35
```

PR #1824, opened 2023-09-09
kopt works with local+grouped reduce and tests

nimlgen:
Made kopt to work with local+grouped reduce and a bit change limits of kopt to make search space smaller.
Added test for local+grouped reduce and some other transformations


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/optimizer.py              5               5
tinygrad/codegen/search.py                 4               5
total                                      9              10
```

geohot:
I think an idea here is to rewrite the hand coded optimizations in terms of the actions taken by the optimizer.

PR #1818, opened 2023-09-08
expand after expr_idxs in Linearizer.global_load

chaosagent:
This changes the behavior of float4 upcast to either succeed for the whole global load, or fail for the whole global load. There is no difference from previous behavior, at least in CI (see CI for [36396a5](https://github.com/tinygrad/tinygrad/pull/1818/commits/36396a53d2651ae028a9643e08aafd48f313bf89)); no comparison asserts trip)

this + substitute only upcasted + flat_components + cache View.expr_idxs:
```
(venv) chaos@tiny3:~/tinygrad$ STEPS=5 OPT=2 WINO=1 GPU=1 JIT=1 python examples/hlb_cifar10.py
/home/chaos/tinygrad/venv/lib/python3.11/site-packages/pyopencl/cache.py:495: CompilerWarning: Non-empty compiler output encountered. Set the environment variable PYOPENCL_COMPILER_OUTPUT=1 to see more.
  _create_built_program_from_source_cached(
  0 13410.33 ms run, 13410.09 ms python,    0.23 ms CL, 1187.23 loss, 0.002563 LR, 0.84 GB used,     35.40 GFLOPS
  1 4891.89 ms run, 4891.34 ms python,    0.55 ms CL, 1187.59 loss, 0.002577 LR, 4.39 GB used,     95.73 GFLOPS
  2   50.13 ms run,    2.85 ms python,   47.28 ms CL, 2760.24 loss, 0.001741 LR, 4.39 GB used,   9342.29 GFLOPS
  3   46.00 ms run,    2.81 ms python,   43.19 ms CL, 2844.66 loss, 0.000906 LR, 4.39 GB used,  10181.02 GFLOPS
  4   45.74 ms run,    2.76 ms python,   42.98 ms CL, 2117.96 loss, 0.000070 LR, 4.39 GB used,  10239.04 GFLOPS
```

Just this set of changes:
```
(venv) chaos@tiny3:~/tinygrad$ git log | head -n1 && STEPS=10 DEBUG=0 TC=0 OPT=2 WINO=1 GPU=1 JIT=1 python examples/hlb_cifar10.py
commit 25b23bca153f50e343de390a12355ffa79b7f948
/home/chaos/tinygrad/venv/lib/python3.11/site-packages/pyopencl/cache.py:495: CompilerWarning: Non-empty compiler output encountered. Set the environment variable PYOPENCL_COMPILER_OUTPUT=1 to see more.
  _create_built_program_from_source_cached(
[...]

PR #1829, opened 2023-09-09
make SumNode.flat_components a generator

chaosagent:
1% gain on hlb_cifar after #1818 , doesn't seem to make a difference before 1818, for some reason (I didn't profile very carefully though)


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/shape/symbolic.py                 3               3
```

roelofvandijk:
There's a different direction we could go here. In some places you want to know the numerical component of a sumnode (e.g. in sumnode floordiv). What if we split that off in a separate member of SumNode, i.e a SumNode 
A) always has n nodes, none of which are SumNodes or NumNodes
B) always has one NumNode

tinygrad/shape/symbolic.py
```diff
@@ -291,9 +291,9 @@ def substitute(self, var_vals: Dict[VariableOrNum, Node]) -> Node: return Variab
 
   @property
   def flat_components(self): # recursively expand sumnode components
-    new_nodes = []
-    for x in self.nodes: new_nodes += (x.flat_components if isinstance(x, SumNode) else [x])
-    return new_nodes
+    for x in self.nodes:
+      if isinstance(x, SumNode): yield from x.flat_components
```
geohot, line 295: Woah, I didn't even know this`yield from` syntax

Issue #1638, opened 2023-08-23
Bounty: Training ResNet-50 v1.5


waifairer:
Add the training logic to examples/mlperf/model_train.py -> def train_resnet():

[Links](https://github.com/mlcommons/training/tree/master/image_classification) to the reference implementation for image classification.

[Links](https://image-net.org/challenges/LSVRC/2012/) to the dataset.

MLPerf training rules [here](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#appendix-allowed-optimizers)

---

Current Bounty: $500
Bounty Locked: #2370

PR #1404, opened 2023-08-02
[WIP] Training Resnet_50v1.5 on ImageNet

ojaffe:
Working on this since [#873](https://github.com/tinygrad/tinygrad/pull/873) got closed.
### TODO

- [x] Data preprocessing
- [x] Data augmentation
- [x] [Model implementation](https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/Classification/ConvNets/resnet50v1.5)
- [x] Evaluation set
- [x] [Logging metrics](https://github.com/mlcommons/training/tree/master/image_classification)
- [x] Replace torchvision import?
- [ ] Scale up
- [ ] Remove wandb + other temporary code

PR #1830, opened 2023-09-09
Train resnet

mekashef:
Going to be picking up where @ojaffe left off.

[Linky to previous](https://github.com/tinygrad/tinygrad/pull/1404#issuecomment-1701744099)

#1638


mekashef:
I've brought it up to the latest on master.

Current problem is that the training is very slow. My previous timing was wrong (timing is often wrong with CUDA=1), so a typical run looked something like this on a 3070. The GPU time is 8x slower than my pytorch benchmark.
441.57 ms step, 37.10 ms fetch data, 398.86 ms GPU, 4.26 GB used, 995.41 GFLOPS

Possible solutions:

Wait for winograd
Do multiGPU stuff. This shouldn't be necessary, on the bottom of [this page](https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/Classification/ConvNets/resnet50v1.5) the authors did 90 epochs in 14 hours with mixed precision
Improve fp16 training. The training works despite failing some tests, has a much lower memory footprint, but is very slightly slower* than full precision
Pad images to nicer shapes (e.g. 256x256)

tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/helpers.py                        2               2
tinygrad/nn/__init__.py                    5               2
tinygrad/runtime/ops_disk.py               2               0
total                                      9               4
```

wozeparrot:
I have worked on this a bit as well, @ojaffe tested on imagenette, but to load actual imagenet is a bit different. Now that wino is merged it should also be possible to use that.

chaosagent:
You can use wino with WINO=1. It just compiles slower for now, the kernel performance on nvidia is quite good.

mehdi10beauty:
> 

@wozeparrot @chaosagent I'll take a look and start on that path.

zaristei3:
Is this still being worked on? Can I pick it up where it was left off?

mekashef:
Yes definitely!

PR #1832, opened 2023-09-10
Fix VALIDHACKS for Images and make it default

### LINKED ISSUE #1801: Fix VALIDHACKS for Images and make it default ($300 bounty)
When you read images out of bounds, they will return 0s. Currently the compiler is unaware of this and still gates the load. Figure out when we don't need it and disable it.

Images are used in the openpilot model `openpilot/go.sh` that have this extra gated load. Safely remove it!

Must be well tested for bounty, it's easy to do this subtly wrong.

Simple example of issue:
`GPU=1 DEBUG=4 FORWARD_ONLY=1 IMAGE=2 python3 test/test_ops.py TestOps.test_simple_padding_conv2d`

generates 

`float4 val0 = ((((lidx0*(-1))<0)*(lidx0<3)))?(read_imagef(data1, smp, (int2)(((lidx0+1)%2),(((lidx0+1)/2)+(-1))))):(float4)(0.0f,0.0f,0.0f,0.0f); # (lidx0 ranges from 0-3)`

instead of

`float4 val0 = read_imagef(data1, smp, (int2)(lidx0-1,0))`

to read image

`dtypes.imagef((1, 2, 4)) # the last 4 is the float4, this is a 2x1 image`

That gate is not needed if you remove the `%2` and subtract 2 from the index. You also then don't need the y index at all.

See `validhacks` in `to_image_idx` for the old (broken) code that hacked this. The symbolic engine should be good enough now to do this properly.

ZenginU:
So this is just a different approach to validhacks.

Fixes #1801

---------
Notes:
* Some cases indexes have sum node but still requires valid. It should not require it as it needs to go out of bound ?


tinyb0t:
Changes made in `tinygrad/`:
```
tinygrad/codegen/linearizer.py            29              17
tinygrad/shape/symbolic.py                 2               1
total                                     31              18
lines added in the tinygrad folder: 13
```

ZenginU:
@wozeparrot Is this enough to bounty lock?

Also, what do you think about the approach. Worth pushing forward?

geohot:
Bounty locked, good luck!

tinygrad/shape/symbolic.py
```diff
@@ -96,6 +96,7 @@ def __mod__(self, b:Union[Node,int]):
     assert b > 0
     if b == 1: return NumNode(0)
     if self.min >= 0 and self.max < b: return self
+    if (self.min//b) == (self.max//b): return self - (b*(self.min//b))
```
geohot, line 99: Neat!

ZenginU:
If I am not mistaken, openpilot's idx has ModNode and ModNode can not go out of bounds. We need special "handtuned garbage" in order to bypass the gate load but this can not be default.

Also theoratically, If we can remove gated loads in openpilot with general approach we can remove it from everywhere but is it even possible? Even old hack can not do that.

ZenginU:
First part of the to_image_idx is clearly understandable.
After first part I have done small symbolic math solver. Because you can not substitute node to node I had to make this possible myself.

There are one or two hacks that I need to take care of but apart from that this is basically it.

Please dont hesitate to show me a simpler way as I am struggling to find a simple solution. Because even the indexes are contiguous like without ModNode or DivNode they still need valid which I do not understand.

ZenginU:
Currently linearizer is bit mess because it includes two different approach to same problem.

I marked as ready for review as I am open to new ideas and need feedback on my approach.
Please don't hesitate to ask any unclear part.

tinygrad/codegen/linearizer.py
```diff
@@ -21,25 +21,108 @@ class UOps(Enum):
...
+        var[1] = nd.b - 1
+    # We do not allow NumNode because it is constant
+    # TODO: Remove mx != mn
+    sub_dict: dict[Union[Variable, NumNode], Node] = {v:Variable(k, mn, mx) for k, (v, (mn, mx)) in var_dict.items() if mx != mn}
+    valid = valid.substitute(sub_dict)
+    idxy = idxy.substitute(sub_dict)
+
+  b = base_shape[1]
+  idx = (idxy // 4) % b
+  idy = (idxy // (4 * b))
```
g1y5x3, line 47: `idx, idy = (idxy // 4) % base_shape[1], (idxy // (4 * base_shape[1]))`

tinygrad/codegen/linearizer.py
```diff
@@ -21,25 +21,98 @@ class UOps(Enum):
...
-      idy += unfactored
-      # ugh really...handtuned garbage
-      if idx.min >= (base_shape[1]*3)//4:
-        idx -= base_shape[1]
-        idy += 1
-  else:
-    idx = (idxy//4)%base_shape[1]
-  if DEBUG >= 5: print("to_image_idx", base_shape, idx.min, idx.max, idy.min, idy.max, idx, idy)
-  return idx, idy
+def to_image_idx(base_shape:Tuple[int, ...], idxy:Node, valid:Node) -> Tuple[Tuple[Node, Node], Node]:
```
geohot, line 24: Hmm, this is quite long. Are any of these generic simplifications that can go in symbolic? And if it really needs to be this long, it needs to be better commented

ZenginU: I will comment more better. The problem is there is valid = SumNode < b and SumNode > c. Because we can not substitute SumNode with different min and max. I needed to solve it by hand to simplify the ModNodes.

ZenginU: Also, I can make this shorter but then openpilot/compile.py takes 30 secs rather than 5 secs.

openpilot/compile.py
```diff
@@ -72,6 +72,11 @@ def compile(dat, output_fn):
     # pass these to thneed
     setattr(prg.clprg, 'op_estimate', prg.op_estimate)
     setattr(prg.clprg, 'prg', prg.prg)
+
+    if getenv("VALIDTEST") == 1:
+      src = re.search(r"=.*\?.*?read_image", prg.prg)
+      if src is not None: raise Exception("Openpilot has valid checks!")
```
geohot, line 78: Aren't there some where it's needed?

tinygrad/codegen/linearizer.py
```diff
@@ -21,25 +21,96 @@ class UOps(Enum):
...
+  # Check for var intersection, removing valid can affect other index
+  if valid.min == 0 and not idx_vars.intersection(idy_vars):
+    nds = valid.nodes if isinstance(valid, AndNode) else [valid]
+    flats = [id.flat_components for id in (idx, idy) if isinstance(id, SumNode)]
+    sym_sums = [Variable.sum([i for i in flat if not isinstance(i, NumNode)]) for flat in flats]
+    ones = [node for sym_sum in sym_sums for node in nds if (node.a == sym_sum) or (-(node.a) == sym_sum)] # type: ignore # AndNode always consists of LtNode
+    valid = Variable.ands([i for i in nds if i not in ones])
+
+  # This is the slow part
+  # This part is for brute forcing all possible values of idx, idy and valid
```
geohot, line 97: Ahhh....I see why this is slow.

tinygrad/codegen/linearizer.py
geohot: Just relying on the brute force...I understand. Is there a way to symbolic solve?

geohot:
This is close to mergable, let me check the speed on the device.

openpilot/compile.py
ZenginU: @geohot 
For openpilot, I think we do not need gated load.
But I have come to this conclusion by brute forcing.
Also, I was assuming that result of the output does not break
"np.testing.assert_allclose(thneed_out, tinygrad_out.numpy())"

ZenginU:
@geohot 
I am trying my best to be as general as possible but because I am dealing with shapes there some fuzz that might break this which I am not aware or misassumption

openpilot/compile.py
ZenginU: Like for now no, for the future maybe? If so, maybe to_image_idx is failing or we really need valid check. But I dont really know how to say we "really" need validchecks by just looking at idx idy valid

ZenginU:
Like we have these cases that needs to be solved 
idy = `<(((3+gidx1[0-6]+ridx5[0-2])//5)+(gidx0[0-1]*3))>`
valid = `<(((gidx1[0-6]*-1)+(ridx5[0-2]*-1))<-1) and ((gidx1[0-6]+ridx5[0-2])<7))>`

ideally idy should be `<((1 + gidx0[0-1]*3))>` because left part should be one

geohot:
Currently the brute force is solving that case? Seems fast enough on device, just tested.

ZenginU:
Brute force does not simplify it but solves the case.

Brute force is not symbolic solver but rather than it directly answer "can we remove valid checks" yes or no. Not always removes the validchecks tho (we need more advanced symbolic solver)

ZenginU:
Final Notes before merge:
* Left 2 TODOs
* Valids only removed by either substitute in first part or in the last brute force part
* Other parts are just for simplifying for the brute force part
Important Notice:
* If too many idx idy not simplified before coming to brute force part brute force takes way too long.

tinygrad/codegen/linearizer.py
```diff
@@ -21,25 +21,86 @@ class UOps(Enum):
...
+  # If valid is both 0 and 1 for the same (idx, idy) we can not delete the valid
+  if valid.min == 0 and not isinstance(idx, ModNode):
+    variables = tuple(val_vars | idy_vars | idx_vars)
+    val_infer, idx_infer, idy_infer = valid.expand(variables), idx.expand(variables), idy.expand(variables)
+    val_dict: Dict[int, Set[Tuple[int,int]]] = {0:set(), 1:set()}
+
+    for v, x, y in zip(val_infer, idx_infer, idy_infer): val_dict[v.min].add((x.min, y.min))
+
+    if not val_dict[1].intersection(val_dict[0]): valid = NumNode(1)
```
gijskoning, line 100: This might be faster. It breaks early if a pair is found. 
```suggestion
    val_list: List[Set[Tuple[int,int]]] = [set(), set()]
    
    for v, x, y in zip(valid.expand(variables), idx.expand(variables), idy.expand(variables)): 
      if (x.min, y.min) in val_dict[1-v.min]: break # found in other set
      val_dict[v.min].add((x.min, y.min))
    else:
      valid = NumNode(1) # no pair found
```

ZenginU: This is I believe faster but does it make difference as main slow part is expands.

Also, I think intersection is more easier to understand but I will look at this ones oerformance as well

geohot:
Congrats, and welcome to the bounty claimers! Send your PayPal to george@tinygrad.org to claim the $300 bounty

PR #1835, opened 2023-09-10
Perf/move numnode mul

roelofvandijk:
* move the specialization
* simplify the return line

Should be slightly faster


tinygrad/shape/symbolic.py
```diff
@@ -159,6 +158,7 @@ def __int__(self): return self.b
   def __index__(self): return self.b
   def __eq__(self, other): return self.b == other
   def __hash__(self): return self.hash  # needed with __eq__ override
+  def __mul__(self, b): return Node.num(self.b*b) if b.__class__ is int else b*self.b
```
geohot, line 161: More complexity

PR #1844, opened 2023-09-11
apply view.py patch

chenyuxyz:
I manually adjusted 2 apply rejects from imports. everything else is from the patch

btw we don't support symbolic shape toCPU so assert should be fine


geohot:
Thanks! Yea I think I had another bug that needed me to comment out that assert

PR #1852, opened 2023-09-11
Annotate shape type with TypeGuard

chenyuxyz:
Applied `TypeGuard` to narrow the type of int shape with `all_int` check. It's supported for python 3.10+, and to support 3.8, we need to import from `typing-extensions`

Also replaced all `math.prod(shape)` to `helpers.prod(shape)` so the `prod` output type can be fixed in one go.

Currently the int shape is checked in creation and before certain math operation. Output shape is possible to be `Union[NumNode, int]` due to symbolic shrink. Will fix this in the proper place separately.


tinygrad/shape/symbolic.py
```diff
@@ -320,6 +321,11 @@ def sym_infer(a: Union[Node, int], var_vals: Dict[Variable, int]) -> int:
   assert isinstance(ret, NumNode)
   return ret.b
 
+# symbolic int
+sint = Union[Node, int]
+
+def all_int(t: Tuple[sint, ...]) -> TypeGuard[Tuple[int, ...]]: return all(isinstance(s, int) for s in t)
```
roelofvandijk, line 327: `all(s.__class__ is int for s in t)`? You don't want types that subclass int.

roelofvandijk: Also, maybe a lru_cache helps here. Not sure.

geohot:
I worry this will fail on the Python 3.9 CI machine, and I can’t get to it right now.

chenyuxyz:
we have downgraded the CPU CI test to use python 3.8 in #1845 so that should cover it

Issue #1461, opened 2023-08-06
Limit buffers per kernel


dc-dc-dc:
WEBGPU supports up to 8 buffers per kernel, this is a blocker for large models running on webgpu

To reproduce 
```bash
WEBGPU=1 python test/test_ops.py TestOps.test_broadcast_full
```

tfmoraes:
This 8 buffer per kernel limit is a blocker to run my implementation of [Unet 3D](https://github.com/tfmoraes/tinygrad_unet) on webgpu backend. If I remove this [assertion](https://github.com/tinygrad/tinygrad/blob/491e85597a2508beb1e8b6e88ee49e4436304a30/tinygrad/runtime/ops_webgpu.py#L17) it works correctly.

PR #1175, opened 2023-07-07
Limit the number of kernel buffers

akretz:
If I'm not mistaken, the only place where the AST optimization adds buffers to a kernel is when two BinaryOps get merged. So adding a check here and not doing the merge in case it would exceed the buffer threshold should prevent kernels with too many buffers to be created.

I'm not sure how the limit should be defined. For now I've just made it an environment variable which defaults to 30 on Metal and to 256 on any other backend.

This fixes #1121 on my machine and also https://github.com/tinygrad/tinygrad/issues/953#issuecomment-1584562667 executes quickly.

PR #1856, opened 2023-09-12
Limit fuse kernel_arg_buffers

DKormann:
prevent lazy from fusing operations with too many kernel arg buffers. #1461

idea: in realize check number of buffers and if they too many pick a parent LazyOp and replace it with a LazyOp that gets realized. 

TODO: 
I set maxnumber to 30 (metal). This needs to be dependent on device

I am choosing the next parent with the most arg buffers for realization. This is simple but maybe unefficient?

probably fucked up elsewhere too but tests should pass for now


chenyuxyz:
see #1175 for previous discussion and the reason it got closed

also need to add tests if there's a behavior change, and the imports should not be in separate lines

DKormann:
so there is supposed to be a onliner for this? 
presumably in _ast_binaryops?

DKormann:
okay so I see how there might be something close to a oneliner in _as_binaryops

DKormann:
I started from scratch. gonna open a new PR

PR #1857, opened 2023-09-13
Limit kernel buffers for metal

DKormann:
sorry super tired i screwed up this PR. gonna fix it tomorrow basically only the last commit is interesting. code should be valid though

continuing #1856 

I actually found a oneliner that seems too many kernel buffers on metal. (+one line for import and one for const buffern number) 

the way the kernel buffer number is determined might not be quite perfect yet. should i use dedup?

also when we decide that there are too many buffers we dissalow of the sources from being merged. This is not efficient but its the simplest solution I can come up with

stuff like Tensor.cat(*[Tensor([1]) for _ in range(35)]) should work now on metal

### DIFF
--- a/test/test_lazybuffer.py
+++ b/test/test_lazybuffer.py
@@ -68,5 +68,9 @@ class TestLazyBuffer(unittest.TestCase):
     assert cache[1][0].name.startswith("E_")
     assert cache[2][0].name.startswith("E_")
 
+  @unittest.skipUnless(Device.DEFAULT == "METAL", "testing metal backend kernel fuse buffer limits")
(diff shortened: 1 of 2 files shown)


DKormann:
sorry super tired i screwed up this PR. gonna fix it tomorrow basically only the last commit is interesting

chaosagent:
This should count number of distinct rawbuffers, and not number of distinct LazyBuffers. There can be multiple LazyBuffers that are different views on top of the same underlying rawbuffer.

geohot:
The solution to them not being realized is lazyrawbuffer

PR #1863, opened 2023-09-13
add line changes diff bot to CI

g1y5x3:
Base on #895.
Add a diff mode to `sz.py` when it detects more than 2 arguments `source dir` and `target dir` which are specified in `szdiff.yml`. It's default action is still to show the total line count. Also set `skip_unchanged` and `ignore_empty` in CI to `true` to prevent the bot from spamming.

~~The output format and some test commits (line changes, same lines changes to skip, no changes ignore) that I did can be found in https://github.com/g1y5x3/tinygrad/pull/2~~

### DIFF
--- a/.github/workflows/szdiff.yml
+++ b/.github/workflows/szdiff.yml
@@ -1,0 +1,52 @@
+name: szdiff.yml
+on:
+  pull_request:
+
+# Cancel the workflow in progress in newer build is about to start.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  szdiff:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code from pr
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          path: pr
+      - name: Checkout code from base
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          path: base
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Count Lines Of Code
+        run: |
+          pip install tabulate
+          BASE="$GITHUB_WORKSPACE/base"
+          PR="$GITHUB_WORKSPACE/pr"
+          cp "$PR/sz.py" .
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          echo "loc_content<<$EOF" >> "$GITHUB_ENV"
+          echo "python sz.py '$BASE'"
+          python sz.py "$BASE" "$PR" >> "$GITHUB_ENV"
+          echo "$EOF" >> "$GITHUB_ENV"
+      - name: Comment Code Lines
+        continue-on-error: false
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          header: LOC
+          ignore_empty: true
+          skip_unchanged: true
+          recreate: true
+          message: ${{ env.loc_content }}
--- a/.github/workflows/szdiff.yml
+++ b/.github/workflows/szdiff.yml
@@ -36,7 +36,6 @@ jobs:
           PR="$GITHUB_WORKSPACE/pr"
           cp "$PR/sz.py" .
           echo "loc_content<<EOF" >> "$GITHUB_ENV"
-          echo "python sz.py '$BASE'"
           python sz.py "$BASE" "$PR" >> "$GITHUB_ENV"
           echo "EOF" >> "$GITHUB_ENV"
       - name: Comment Code Lines
--- a/.github/workflows/szdiff.yml
+++ b/.github/workflows/szdiff.yml
@@ -35,11 +35,10 @@ jobs:
           BASE="$GITHUB_WORKSPACE/base"
           PR="$GITHUB_WORKSPACE/pr"
           cp "$PR/sz.py" .
-          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
-          echo "loc_content<<$EOF" >> "$GITHUB_ENV"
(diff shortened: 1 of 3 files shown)


geohot:
Is this ready to go?

g1y5x3:
> Is this ready to go?

I believe so, all bots behavior are tested on my own branch.

.github/workflows/szdiff.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Count Lines Of Code
+        run: |
+          pip install tabulate
+          BASE="$GITHUB_WORKSPACE/base"
+          PR="$GITHUB_WORKSPACE/pr"
+          cp "$PR/sz.py" .
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
```
wozeparrot, line 39: this seems kinda hacky but meh, can't we guarantee certain things won't appear in the output? cuz we are generating the output ourselves anyways

.github/workflows/szdiff.yml
```diff
@@ -0,0 +1,53 @@
...
+  pull_request:
+
+# Cancel the workflow in progress in newer build is about to start.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  szdiff:
+    if: github.event.pull_request.draft == false
```
wozeparrot, line 12: Do we want this to run on draft prs?

.github/workflows/szdiff.yml
g1y5x3: The dd part was just following the [package's original example](https://github.com/marocchino/sticky-pull-request-comment#keep-more-than-one-comment) so I didn't dig too much.

sz.py
```diff
@@ -1,28 +1,54 @@
...
+        table.append([relfilepath, line_count, token_count/line_count])
+  return table
+
+def display_diff(diff): return "+"+str(diff) if diff > 0 else str(diff)
 
-  print(tabulate([headers] + sorted(table, key=lambda x: -x[1]), headers="firstrow", floatfmt=".1f")+"\n")
+def gen_diff(table_old, table_new):
+  table = []
+  for line_old, line_new in zip(table_old, table_new):
+    if line_new[1]-line_old[1] != 0:
```
madt2709, line 28: what happens when the exact same number of lines gets added and then deleted?

madt2709: its kind of an edge case so not that important but tinyb0t used to support that i think

g1y5x3: Yeah, not sure how feasible it is to do it with `sz.py` since itself doesn't track additions or deletions

g1y5x3:
Here is the scenario that I haven't given into consideration. Whenever there is a brand new file being added or a entire file being deleted inside the directory, it is going to break `sz.py` since both table are going to have different row coutns.

Also @madt2709 I think the case where line diff is 0 can be some what countered if I also compute the `tokens/line` diff. And if both are 0 when there's still edits, I believe it would be something that's not difficult to review in the code anyway.

g1y5x3:
Here is the output from a pretty through test I did. Another sample output can be seen in https://github.com/g1y5x3/tinygrad/pull/2
```
Name                              Lines    Diff    Tokens/Line    Diff
------------------------------  -------  ------  -------------  ------
tinygrad/tensor.py                  526     -14           22.9    +0.5
tinygrad/lazy.py                    290      -4           17.6    +0.1
tinygrad/codegen/optimizer.py       284      -1           16.2    +0.0
tinygrad/shape/symbolic.py          284      -3           14.8    -0.0
tinygrad/shape/shapetracker.py      217     +97           17.1    +4.2
tinygrad/mlops_copy.py              147    +147           14.3   +14.3
tinygrad/helpers.py                 115      -1           15.3    -0.2
tinygrad/codegen/kernel.py           92      -2           17.1    +0.2
tinygrad/nn/__init__.py              91      -4           19.2    +0.3
tinygrad/shape/view.py                0    -118            0.0   -18.5


total lines changes: +97
```
@wozeparrot still trying to think what assert we could add for some output that obviously seems wrong. But it should be very unlikely now since I manually match file names.

sz.py
```diff
@@ -1,28 +1,70 @@
...
+      for line in table_old: 
+        if file in line: table.append([line[0], 0, 0 - line[1], 0, 0-line[2]])
+  if unchanged:
+    for file in unchanged:
+      for line in table_old: 
+        if file in line: line_old = line 
+      for line in table_new: 
+        if file in line: line_new = line
+      if line_new[1]-line_old[1] != 0 or line_new[2]-line_old[2] != 0: 
+        table.append([line_new[0], line_new[1], line_new[1]-line_old[1], line_new[2], line_new[2]-line_old[2]])
```
madt2709, line 43: you should be able to simplify this quite a bit with something along the lines of (not tested): 
``` python
for file in files_new.union(files_old):
  if (file in unchanged) and (table_old[file] != table_new[file]):
    table.append([table_new[file][0], table_new[file][1], table_new[file][1]-table_old[file][1], table_new[file][2], table_old[file][file][2]-line_old[2]])
  else:
    table.append([line[0], line[1], line[1] - 0, line[2], line[2]-0]) if file in added else table.append([line[0], 0, 0 - line[1], 0, 0-line[2]])
```

g1y5x3: Oh, I totally agree this part can be better written. The nested for loops is kinda bad. However, yours won't work since `table_old`, `table_new` are lists not dictionaries.

.github/workflows/szdiff.yml
wozeparrot: yea it's fine, just a small nit since we know our output and can guarantee that certain things won't appear in it.

sz.py
madt2709: oh my bad didnt realise it was a list...

yep i agree its slightly more readable now 👍 

is there any reason for not storing table as a dictionary? if you're going to loop through the list to find the relevant element, you might as well store a hash map with the dictionary.

g1y5x3: Just lookup the `tabulate` package and the print is doable with dictionary. I think it has been using list because it was easy and there wasn't a need for hash map yet.

g1y5x3: start to think is there actually a need for hash map? currently nested list works fine for what we need without making too many changes.

g1y5x3:
I think it's ready @geohot @wozeparrot

wozeparrot:
merged, lets see how it works out on some prs

PR #1864, opened 2023-09-13
Add line count diff using sz.py

madt2709:
Added a DIFF mode to sz.py where you pass the path to a different branch of tinygrad to compare the versions.

This automatically runs in CI and then pastes comment.

EDIT: Oh I see im a little slow compared to https://github.com/tinygrad/tinygrad/pull/1863. Let me see if its better 👀


madt2709:
Meh the other PR is slightly better imo so will close this

PR #1875, opened 2023-09-16
no import when Python is shutting down

nimlgen:
This fixes import error when python is exiting: `ImportError: sys.meta_path is None, Python is likely shutting down` (from `extra/gemm/hip_matmul.py`).


geohot:
Do you even need that del?

nimlgen:
Need it for tests, but I put `__del__` right there.

PR #1878, opened 2023-09-17
hip multigpu training

wozeparrot:
is much faster with p2p


wozeparrot:
There is some weird sync thing happening, at least with cifar where some of the processes jump ahead, even tho I have barriers in place.

The other thing now is to optimize the non-transfer/python side of things, as there is a pretty sizeable overhead/latency per allreduce. 

CC @roelofvandijk, it just needs optimization on the jitted path, not sure if there is a faster way I can do out-of-band comms or not.

geohot:
What's the blocker on merging this?

extra/dist/world.py
```diff
@@ -1,59 +1,101 @@
...
 # fake the function signature of ASTRunner so we can put it in the cache
-def __send_rb(args:Tuple[RawBufferCopyInOut, RawShmBuffer, int, Any], variables=None, jit=False, force_wait=False):
-  args[0]._copyout(np.frombuffer(args[1]._buffer(), dtype=args[0].dtype.np))
-  dist.OOB.send(args[3], args[2])
-  if DEBUG >= 2: print(f"{colored('****', 'magenta' if jit else None)}   sent {args[0]} to rank {args[2]}")
+def __send_rb(args, variables=None, jit=False, force_wait=False):
+  x, target_rank, y = args[:3]
+  if RawHIPBuffer and x.__class__ is RawHIPBuffer:
+    hip.hipSetDevice(x._device)
+    hip.hipDeviceSynchronize()
```
geohot, line 20: Hmm. So this is okay, it's better than what we have, but is there a way to do this without needing to sync the GPU?

PR #1879, opened 2023-09-17
add branch up-to-date check

g1y5x3:
Add a branch up-to-date check before triggering the comment bot. It utilizes the `git rev-list --left-right --count  tinygrad/master...${{ github.event.pull_request.head.sha }}` to detect the exact amount of behind/ahead of current branch compare to the tinygrad master branch. Then just condition the comment bot action only on when branch is up-to-date (this is adjustable in `$count -gt 0`).

Here is a sample outout:
```
git rev-list --left-right --count  tinygrad/master...${{ github.event.pull_request.head.sha }} | awk '{print "Behind "$1" - Ahead "$2""}'
Behind 6 - Ahead 0
```

### DIFF
--- a/.github/workflows/szdiff.yml
+++ b/.github/workflows/szdiff.yml
@@ -95,5 +95,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           skip_unchanged: true
           message: |
-            This branch currently is behind tinygrad/master.
-            If there is no merge conflicts, please rebase.
+            This branch currently is behind tinygrad/master. The line count difference bot is disabled.
--- a/.github/workflows/szdiff.yml
+++ b/.github/workflows/szdiff.yml
@@ -79,3 +79,21 @@ jobs:
           skip_unchanged: true
           recreate: true
           message: ${{ env.loc_content }}
+
+  rebase:
+    name: Core Library Line Difference
+    permissions:
+      pull-requests: write
+    runs-on: ubuntu-latest
+    needs: checkbranch
+    if: needs.checkbranch.outputs.branchstat == 'true'
+    steps:
+      - name: Comment Rebase
+        continue-on-error: false
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          skip_unchanged: true
+          message: |
+            This branch currently is behind tinygrad/master.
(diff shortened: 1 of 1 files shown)


g1y5x3:
initially I kept this branch behind on purpose hoping to be able to test it when I make the PR. however, I don't think CI is using the new yml file from this PR. Thus the comment bot was triggered. 

Here is the test I did to show that comment bot was skipped
https://github.com/g1y5x3/tinygrad/actions/runs/6216063049/job/16869566238?pr=10#step:3:21

g1y5x3:
Instead of skipping, I make it trigger another bot comment that suggests a rebase.
https://github.com/g1y5x3/tinygrad/pull/16

.github/workflows/szdiff.yml
```diff
@@ -49,3 +79,21 @@ jobs:
...
+    steps:
+      - name: Comment Rebase
+        continue-on-error: false
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          skip_unchanged: true
+          message: |
+            This branch currently is behind tinygrad/master.
+            If there is no merge conflicts, please rebase.
```
wozeparrot, line 99: I think we can just mention that the current branch is behind, don't need the hint to rebase, maybe instead have it say that the line diff won't be shown until the branch is no longer behind?

g1y5x3: Yeah, it definitely sounds a lot nicer this way.

nimlgen: Why hide line diffs? If there are no merge conflicts, the line diff is the same in both cases (rebased and not), isn't it? The only reason we might want to ask for a rebase is if the branch is too outdated to run tests from the master (to avoid getting failed tests after the PR is merged).

g1y5x3: > If there are no merge conflicts, the line diff is the same in both cases (rebased and not), isn't it?

Not necessarily. The bot doesn't track conflicts, but only counts the raw line difference between the given directory. So if there's other changes made which affect the count, it's going to be reflected even if the change is not from the PR branch.

nimlgen: Well, maybe it's fine for now to show diff only when commits are rebased on the top of PR's target branch.

In the future for this case we could compare this with rebased version (to PR's target branch) or with pr's branch with N reverted commits (where N is number of commits in the PR).

PR #1880, opened 2023-09-18
HipGraph support

nimlgen:
This PR adds HipGraph support.

```
With graph:
ran model in 3.23 ms
total 5.25 ms

Without graph:
ran model in 17.38 ms
total 18.95 ms
```

HIPGraph works only for a single device for now.

### DIFF
--- a/extra/hip_wrapper.py
+++ b/extra/hip_wrapper.py
@@ -680,9 +680,3 @@ try:
     return e_code
 except:
   if DEBUG >= 1: print("WARNING: libamdhip64.so or libhiprtc.so not found. HIP support will not work.")
-  raise OSError('cant find libamdhip64.so.')
-
-try:
-  _libhiprtc = ctypes.cdll.LoadLibrary('libhiprtc.so')
-except:
-  raise OSError('cant find libhiprtc.so.')
--- a/extra/hip_wrapper.py
+++ b/extra/hip_wrapper.py
@@ -112,9 +112,12 @@ try:
     deps = [ctypes.cast(deps_out[i], ctypes.c_void_p) for i in range(num_deps.value)]
     return status_out, id_out.value, graph_out, deps
 
+  hipStreamAddCaptureDependencies = 0
+  hipStreamSetCaptureDependencies = 1
+
   _libhip.hipStreamUpdateCaptureDependencies.restype = int
   _libhip.hipStreamUpdateCaptureDependencies.argtypes = [ctypes.c_void_p, ctypes.c_void_p, ctypes.c_size_t, ctypes.c_uint]
-  def hipStreamUpdateCaptureDependencies(stream, deps, flags=1):
+  def hipStreamUpdateCaptureDependencies(stream, deps, flags=hipStreamAddCaptureDependencies):
     deps_in = (ctypes.c_void_p * len(deps))()
     deps_in[:] = deps
     num_deps = ctypes.c_size_t()
--- a/extra/hip_wrapper.py
+++ b/extra/hip_wrapper.py
@@ -1,5 +1,9 @@
 import ctypes
 from tinygrad.helpers import DEBUG
+import sys
+import numpy as np
+from typing import Any, Dict, List, Tuple
+from dataclasses import dataclass
 
 try:
   _libhip = ctypes.cdll.LoadLibrary("libamdhip64.so")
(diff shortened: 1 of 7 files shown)


nimlgen:
I am not really sure about `NOSTAT` envvar. The reason it's added is `op_estimate = sym_infer(self.op_estimate, var_vals)` which takes ~1ms each step. I think we would like have an option to disable this, since it takes a lot of time. Should we have something like `NOSTAT`?

nimlgen:
With `jit cache` it's about `ran model in 3.23 ms`. `sym_infer` for `global_size`/`local_size` is still a bit expensive. I think this change might be reasonable, since we already have `jit_ctx`.

nimlgen:
Btw not sure about `NOSTAT` envvar. `op_estimate = sym_infer(self.op_estimate, var_vals)`  takes some time to calculate (0.5ms for each gpt-2 step) while I think not really needed in all cases.

geohot:
Similar to CachedAllocator, this is quite complex looking. Can the hooks into the JIT be simpler?

nimlgen:
I think now it should be better after small refactor. Also speed is a bit better.

Tested quantised llama on 7900xtx, it's also faster 59.73 tok/sec vs 40.37 tok/sec.

tinygrad/jit.py
```diff
@@ -3,21 +3,26 @@
...
-from tinygrad.ops import Device
+from tinygrad.ops import RawBuffer, Device, Compiled
 from tinygrad.tensor import Tensor
-from tinygrad.ops import RawBuffer
 from tinygrad.shape.shapetracker import ShapeTracker
 from tinygrad.shape.symbolic import Variable
 
 JIT_SUPPORTED_DEVICE = ["GPU", "CLANG", "METAL", "CUDA", "HIP", "WEBGPU", "LLVM"]
 
+class JitCtx:
```
geohot, line 13: Does this have to be a separate class? I don't see why this has to change

nimlgen: Removed it

tinygrad/helpers.py
```diff
@@ -56,7 +56,7 @@ def __ge__(self, x): return self.value >= x
   def __gt__(self, x): return self.value > x
   def __lt__(self, x): return self.value < x
 
-DEBUG, IMAGE = ContextVar("DEBUG", 0), ContextVar("IMAGE", 0)
+DEBUG, NOSTAT, IMAGE = ContextVar("DEBUG", 0), ContextVar("NOSTAT", 0), ContextVar("IMAGE", 0)
```
geohot, line 59: Why do I ever want NOSTAT?

tinygrad/shape/symbolic.py
```diff
@@ -10,6 +10,8 @@
 # NOTE: Python has different behavior for negative mod and floor div than c
 # symbolic matches the Python behavior, but the code output is agnostic, and will never have negative numbers in div or mod
 
+infer_cache: Dict[Tuple[int, Any], int] = {}
```
geohot, line 13: Does this have to be global?

tinygrad/ops.py
```diff
@@ -149,13 +149,15 @@ def exec(self, bufs, var_vals:Optional[Dict[Variable, int]]=None, force_wait=Fal
...
+    return global_size, local_size
+
   def __call__(self, rawbufs:List[RawBuffer], var_vals:Optional[Dict[Variable, int]]=None, jit=False, force_wait=False) -> Optional[float]:
     if var_vals is None: var_vals = {}
-    global_size = [sym_infer(sz, var_vals) for sz in self.global_size] if self.global_size is not None else self.global_size
-    local_size = [sym_infer(sz, var_vals) for sz in self.local_size] if self.local_size is not None else self.local_size
-    if et := self.clprg((global_size + [1]*(3-len(global_size))) if global_size is not None else None,
-                        (local_size + [1]*(3-len(local_size))) if local_size is not None else None,
-                        *rawbufs, *var_vals.values(), wait=force_wait or DEBUG>=1): GlobalCounters.time_sum_s += et
+    global_size, local_size = self.launch_dims(var_vals)
```
geohot, line 159: Nice refactor

geohot:
I think instead of having this be different cases in the JIT, the JIT should always used a BasicBatchExecutor (which just calls the kernels sequentially). This way this codepath is tested.

tinygrad/shape/symbolic.py
nimlgen: In this impl it has to be. The problem is that sym_infer is really expensive (takes up to 50% of the gpt2 step) we can cache results, since they are reused often. gpt2 runs several TinyJits, so wanna use the same cache across them. So I see only 2 solutions global var (maybe move to jit.py) or `JitCtx` which is also shared between TinyJits.

As for JitCtx impl: I have chosen JitCtx since it's the most opaque for the end user as they should not care about caches. `jit_ctx = JitCtx({pos: start_pos})` creates a shadow cache for the provided set of vars which is shared between TinyJits (and when you want a new set of values to pass you just create a new `JitCtx` so caches do not intersect).

tinygrad/helpers.py
nimlgen: Removed it. With the `sym_infer` cache results do not differ so much.

tinygrad/shape/symbolic.py
nimlgen: With #1896 i think we can drop and have cache just inside TinyJit. I think the approach is possible to apply to llama as well to have one Jit.

nimlgen: Looking at #1896 I think of the following plan. This PR uses one JIT for the whole gpt step which allows to solve problem with caching having no new globals and no JitCtx. But the main problem is that it kinda delay execution of kernels on gpus and we haven't got cpu+gpu working in parallel (this is really visible with hipgraphs). The solution for this is a small refactor of JIT to allow updating batches of kernels and launch them immediately (now we just update everything and launch them after that).

I am not sure it's a great idea to mix up refactor of JIT with this PR, since this PR is already big. So we might have not the best performance for now and improve it with the JIT refactor. How that sounds?

nimlgen: Put it into JIT itself

nimlgen:
The current timings are `ran model in 2.43 ms total 7.49 ms`. I am going to leave this PR functionality as it is (unless other review comments), cuz it's already big to review.

I am going to refactor JIT a bit in the next PR to enable parallel execution cpu+gpu (send some amount of commands to the gpu while cpu updating others params of following cuda/hipgraph). I have a patch which splits jit execution into batches (sizes of the batches are logarithmic 4,8,16,32...), so both cpu and accel are loaded. This gives `ran model in 2.39 ms total 5.61 ms`. Regular jit executions (not cuda/hipgraph) should also benefit from this change, but not perf diff is not that huge.

tinygrad/jit.py
```diff
@@ -55,14 +56,24 @@ def __call__(self, *args, **kwargs) -> Any:
         for i,a in enumerate(cache[1]):
           if a in [v[0] for v in input_rawbuffers.values()]:
             self.input_replace[(j_,i)] = [(k, v[1], v[0].dtype) for k,v in input_rawbuffers.items() if v[0] == a][0]
+            self.updatable_entries[j_].append(i)
+        for i in range(len(cache[2])): self.updatable_entries[j_].append(len(cache[1])+i)
         #if prg.local_size is None: prg.local_size = prg.optimize_local_size(args, preserve_output=True)  # the JIT can optimize local
       assert set([x[0] for x in self.input_replace.values()]) == set(input_rawbuffers.keys()), "some input tensors not found"
+
+      # init batch_executor
+      try: self.batch_executor = cast(Compiled, Device[Device.DEFAULT]).batch_exec(self.jit_cache)
```
geohot, line 65: Hmm...should this be the DEFAULT Device? Also, shouldn't BasicBatchExecutor just be the default on compiled?

nimlgen: BasicBatchExecutor is the default now. Possible to take batch_exec from jit_cache's prg itself, is it better? Anyways HipGraph and others check if they can serve given jit_cache overwise it switches to BasicBatchExecutor.

tinygrad/shape/symbolic.py
```diff
@@ -314,10 +314,12 @@ def create_rednode(typ:Type[RedNode], nodes:List[Node]):
 @functools.lru_cache(maxsize=None)
 def sym_rename(s) -> str: return f"s{sym_rename.cache_info().currsize}"
 def sym_render(a: Union[Node, int], ops=None, ctx=None) -> str: return str(a) if isinstance(a, int) else a.render(ops, ctx)
-def sym_infer(a: Union[Node, int], var_vals: Dict[Variable, int]) -> int:
+def sym_infer(a: Union[Node, int], var_vals: Dict[Variable, int], infer_cache=None) -> int: # When infer_cache is provided, cache the infered result for **the given var_vals**.
```
geohot, line 317: Is this cache needed, or is this just a speedup? It's not really worth trading this sort of complexity for speed

extra/hip_wrapper.py
```diff
@@ -505,3 +680,9 @@ def hiprtcGetCode(prog):
     return e_code
 except:
   if DEBUG >= 1: print("WARNING: libamdhip64.so or libhiprtc.so not found. HIP support will not work.")
```
geohot, line 682: Was there a reason this was only a warning before?

tinygrad/jit.py
```diff
@@ -55,8 +56,14 @@ def __call__(self, *args, **kwargs) -> Any:
...
           if a in [v[0] for v in input_rawbuffers.values()]:
             self.input_replace[(j_,i)] = [(k, v[1], v[0].dtype) for k,v in input_rawbuffers.items() if v[0] == a][0]
+            self.updatable_entries[j_].append(i)
+        for i in range(len(cache[2])): self.updatable_entries[j_].append(len(cache[1])+i)
         #if prg.local_size is None: prg.local_size = prg.optimize_local_size(args, preserve_output=True)  # the JIT can optimize local
       assert set([x[0] for x in self.input_replace.values()]) == set(input_rawbuffers.keys()), "some input tensors not found"
+
+      # init batch_executor
+      try: self.batch_executor = self.jit_cache[0][0].batch_exec(self.jit_cache)
+      except (AttributeError, ValueError): self.batch_executor = BasicBatchExecutor(self.jit_cache)
```
geohot, line 66: Why should this ever fail? Shouldn't it always have BasicBatchExec?

tinygrad/shape/symbolic.py
nimlgen: Agree. This is an ugly hack, removed it. Still need to do something with sym_infer later, about 60-70% time of gpt2 step is spent in sym_infer for global_size and op_estimate.

extra/hip_wrapper.py
nimlgen: Oh thanks, missed that after rebase.

tinygrad/jit.py
nimlgen: Not any more.

geohot:
Nice! BatchExecutor is a good abstraction

PR #1882, opened 2023-09-18
wmma: refactor tensor core

flammit:
To test HIP TC, run `PYTHONPATH="." HIP=1 HALF=1 N=4096 DEBUG=3 python3 ./extra/gemm/simple_matmul.py`.

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -27,6 +27,7 @@ tensor_cores: Dict[str, List[TensorCore]] = {
   ],
   "HIP": [
     TensorCore(device="HIP", dims=[16,16,16], dtype_in=dtypes.half, dtype_out=dtypes.float, threads=[16,2], thread_local_sizes=[16,16,8], thread_local_aliases=[ [[-1], [0], [1]], [[-1], [1], [0]], [[0], [1], [2, -1]] ]),
+    TensorCore(device="HIP", dims=[16,16,16], dtype_in=dtypes.half, dtype_out=dtypes.half,  threads=[16,2], thread_local_sizes=[16,16,8], thread_local_aliases=[ [[-1], [0], [1]], [[-1], [1], [0]], [[0], [1], [2, -1]] ]),
   ]
 }
 
@@ -76,9 +77,7 @@ class Kernel:
     self.upcasted: int = 0
     self.local_dims: int = 0
     self.local_alias: Dict[int, LocalBuffer] = {}
-    self.use_tensor_cores: bool = False
-    self.exclude_local_upcast: int = 0
-    self.reverse_upcast_dir: bool = False
+    self.tensor_core: Optional[TensorCore] = None
 
     self.global_size: Optional[List[int]] = None
     self.local_size: Optional[List[int]] = None
@@ -139,6 +138,7 @@ class Kernel:
   # white  -- reduce-late upcasted dim (self.upcast_in_mid_reduce_axes)
   # red    -- reduce loops
   #  *** self.upcasted
+  # cyan   -- WWMA
   # purple -- reduce upcasted
   # yellow -- normal upcasted dimensions
   def colors(self) -> List[str]:
@@ -150,8 +150,8 @@ class Kernel:
     colors += ["white" if i in self.upcast_in_mid_reduce_axes else "green" for i in range(self.first_reduce, self.first_reduce + len(self.group_for_reduce))]
     # between first_reduce + group_for_reduce and upcasted, they are reduce (red)
     colors += ["red"] * ((self.shape_len-self.upcasted) - (self.first_reduce + len(self.group_for_reduce)))
-    # upcasted dimensions are reduce (magenta) or normal (yellow)
-    colors += ["magenta" if self.full_shape[i] != self.sts[0].shape[i] else "yellow" for i in range(self.shape_len-self.upcasted, self.shape_len)]
+    # upcasted dimensions are cyan (WMMA), reduce (magenta) or normal (yellow)
+    colors += ["cyan"] * (3 if self.tensor_core else 0) + ["magenta" if self.full_shape[i] != self.sts[0].shape[i] else "yellow" for i in range(self.shape_len-self.upcasted+(3 if self.tensor_core else 0), self.shape_len)]
     assert len(colors) == self.shape_len, "colors size mismatch"
     return colors
 
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -191,9 +191,29 @@ class Linearizer(OptimizedKernel):
 
     # define indexes
     global_idxs, loop_global_idxs = get_grouped_dims("gidx", 0, self.full_shape[:self.global_dims], 3 if self.opts.has_local else 0)
-    local_idxs, loop_local_idxs = get_grouped_dims("lidx", self.global_dims, self.full_shape[self.global_dims:self.first_reduce+len(self.group_for_reduce)], 3 if self.opts.has_local else 0)
-    full_upcast_idxs = [Variable(None, 0, s-1) for s in self.full_shape[self.shape_len-self.upcasted:]]
-    upcast_idxs = [Variable(None, 0, s-1) for s in self.output_shape[self.shape_len-self.upcasted:]]
+    local_idxs, loop_local_idxs = get_grouped_dims("lidx", self.global_dims+(1 if self.tensor_core else 0), self.full_shape[self.global_dims:self.first_reduce+len(self.group_for_reduce)], 3-(1 if self.tensor_core else 0) if self.opts.has_local else 0)
+    full_upcast_idxs = [Variable(None, 0, s-1) for s in self.full_shape[self.shape_len-self.upcasted+(3 if self.tensor_core else 0):]]
+    upcast_idxs = [Variable(None, 0, s-1) for s in self.output_shape[self.shape_len-self.upcasted+(3 if self.tensor_core else 0):]]
+    warp_idxs, aliased_wmma_idxs = [], []
+
+    if self.tensor_core:
+      warp_idxs, loop_warp_idxs = get_grouped_dims("lidx", self.global_dims, [prod(self.tensor_core.threads)], 1)
+      warp_segment_idxs = []
+      for segment_sz in self.tensor_core.threads:
+        warp_segment_idxs.append(loop_warp_idxs[0] % segment_sz)
+        loop_warp_idxs[0] //= segment_sz
+      def calc_tc_idxs(local_size: int, aliases: List[List[int]]):
+        idxs = []
+        for alias in aliases:
+          full_var, full_var_sz = Variable.num(0), 1
+          if alias[0] != 0:
+            for i in alias:
+              next_var = warp_segment_idxs[i-1] if i > 0 else Variable(None, 0, local_size-1)
+              full_var += next_var * full_var_sz
+              full_var_sz *= next_var.max+1
+          idxs.append(full_var)
+        return idxs
+      aliased_wmma_idxs = [calc_tc_idxs(s, a) for s, a in zip(self.tensor_core.thread_local_sizes, self.tensor_core.thread_local_aliases)]
 
     # global and local loops
     def render_loop(xx:List[Variable]):
@@ -207,11 +227,11 @@ class Linearizer(OptimizedKernel):
           if loop_uop.uop == UOps.LOOP: self.uop(UOps.END, None, (loop_uop,))
 
     if self.opts.has_local:
-      self.global_size, self.local_size = [x.max+1 for x in loop_global_idxs][::-1], [x.max+1 for x in loop_local_idxs][::-1]
+      self.global_size, self.local_size = [x.max+1 for x in loop_global_idxs][::-1], [x.max+1 for x in (loop_local_idxs+warp_idxs)][::-1]
       self.global_size += [1]*(3-len(self.global_size))
       self.local_size += [1]*(3-len(self.local_size))
       self.loop_uops.update({x.expr:self.uop(UOps.SPECIAL, dtypes.int32, (), (len(loop_global_idxs)-1-i, x.expr, x.max+1)) for i,x in enumerate(loop_global_idxs)})
-      self.loop_uops.update({x.expr:self.uop(UOps.SPECIAL, dtypes.int32, (), (len(loop_local_idxs)-1-i, x.expr, x.max+1)) for i,x in enumerate(loop_local_idxs)})
+      self.loop_uops.update({x.expr:self.uop(UOps.SPECIAL, dtypes.int32, (), (len(loop_local_idxs+warp_idxs)-1-i, x.expr, x.max+1)) for i,x in enumerate(loop_local_idxs+warp_idxs)})
     else:
       render_loop(loop_global_idxs+loop_local_idxs)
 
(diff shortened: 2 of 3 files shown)


tinygrad/codegen/kernel.py
```diff
@@ -8,6 +8,23 @@
 from tinygrad.shape.symbolic import sint
 from tinygrad.shape.view import strides_for_shape
 
+class TensorCore(NamedTuple):
```
roelofvandijk, line 11: Consider `@dataclass(frozen=True)`. Newer, slightly faster.

tinygrad/codegen/linearizer.py
```diff
@@ -224,9 +224,29 @@ def linearize(self):
...
+    if self.tensor_core:
+      warp_idxs, loop_warp_idxs = get_grouped_dims("lidx", self.global_dims, [prod(self.tensor_core.threads)], 1)
+      warp_segment_idxs = []
+      for segment_sz in self.tensor_core.threads:
+        warp_segment_idxs.append(loop_warp_idxs[0] % segment_sz)
+        loop_warp_idxs[0] //= segment_sz
+      def calc_tc_idxs(local_size: int, aliases: List[List[int]]):
+        idxs = []
+        for alias in aliases:
+          full_var, full_var_sz = Variable.num(0), 1
```
roelofvandijk, line 241: You don't need to define `full_var_sz` if `alias[0] == 0`

tinygrad/codegen/linearizer.py
```diff
@@ -224,9 +224,29 @@ def linearize(self):
...
+          if alias[0] != 0:
+            for i in alias:
+              next_var = warp_segment_idxs[i-1] if i > 0 else Variable(None, 0, local_size-1)
+              full_var += next_var * full_var_sz
+              full_var_sz *= next_var.max+1
+          idxs.append(full_var)
+        return idxs
+      aliased_wmma_idxs = [calc_tc_idxs(x[0], x[1]) for x in zip(self.tensor_core.thread_local_sizes, self.tensor_core.thread_local_aliases)]
```
roelofvandijk, line 249: `[calc_tc_idxs(s, a) for s, a in zip(self.tensor_core.thread_local_sizes, self.tensor_core.thread_local_aliases)]`
or
``[calc_tc_idxs(*x) for x in zip(self.tensor_core.thread_local_sizes, self.tensor_core.thread_local_aliases)]`

tinygrad/codegen/linearizer.py
```diff
@@ -224,9 +224,29 @@ def linearize(self):
 
     # define indexes
     global_idxs, loop_global_idxs = get_grouped_dims("gidx", 0, self.full_shape[:self.global_dims], 3 if self.opts.has_local else 0)
-    local_idxs, loop_local_idxs = get_grouped_dims("lidx", self.global_dims, self.full_shape[self.global_dims:self.first_reduce+len(self.group_for_reduce)], 3 if self.opts.has_local else 0)
-    full_upcast_idxs = [Variable(None, 0, s-1) for s in self.full_shape[self.shape_len-self.upcasted:]]
-    upcast_idxs = [Variable(None, 0, s-1) for s in self.output_shape[self.shape_len-self.upcasted:]]
+    local_idxs, loop_local_idxs = get_grouped_dims("lidx", self.global_dims+(1 if self.tensor_core else 0), self.full_shape[self.global_dims:self.first_reduce+len(self.group_for_reduce)], (2 if self.tensor_core else 3) if self.opts.has_local else 0)
+    full_upcast_idxs = [Variable(None, 0, s-1) for s in self.full_shape[self.shape_len-self.upcasted+(3 if self.tensor_core else 0):]]
```
roelofvandijk, line 228: you could define `self.shape_len-self.upcasted+(3 if self.tensor_core else 0)`

tinygrad/codegen/linearizer.py
```diff
@@ -224,9 +224,29 @@ def linearize(self):
 
     # define indexes
     global_idxs, loop_global_idxs = get_grouped_dims("gidx", 0, self.full_shape[:self.global_dims], 3 if self.opts.has_local else 0)
-    local_idxs, loop_local_idxs = get_grouped_dims("lidx", self.global_dims, self.full_shape[self.global_dims:self.first_reduce+len(self.group_for_reduce)], 3 if self.opts.has_local else 0)
-    full_upcast_idxs = [Variable(None, 0, s-1) for s in self.full_shape[self.shape_len-self.upcasted:]]
-    upcast_idxs = [Variable(None, 0, s-1) for s in self.output_shape[self.shape_len-self.upcasted:]]
+    local_idxs, loop_local_idxs = get_grouped_dims("lidx", self.global_dims+(1 if self.tensor_core else 0), self.full_shape[self.global_dims:self.first_reduce+len(self.group_for_reduce)], (2 if self.tensor_core else 3) if self.opts.has_local else 0)
```
roelofvandijk, line 227: `1 if self.tensor_core else 0` is the same as `int(self.tensor_core)`

tinygrad/codegen/linearizer.py
```diff
@@ -224,9 +224,29 @@ def linearize(self):
...
+    full_upcast_idxs = [Variable(None, 0, s-1) for s in self.full_shape[self.shape_len-self.upcasted+(3 if self.tensor_core else 0):]]
+    upcast_idxs = [Variable(None, 0, s-1) for s in self.output_shape[self.shape_len-self.upcasted+(3 if self.tensor_core else 0):]]
+    warp_idxs, aliased_wmma_idxs = [], []
```
roelofvandijk, line 230: Could be in an `else` after `if self.tensor_core`.

tinygrad/codegen/optimizer.py
```diff
@@ -166,119 +166,33 @@ def hand_coded_optimizations(self):
...
+          continue
+        has_cast = tc.dtype_in != tc.dtype_out
+        if has_cast and not(isinstance(self.reduceop.src[0], LazyOp) and self.reduceop.src[0].op == UnaryOps.CAST and self.reduceop.src[0].arg[0] == tc.dtype_out):
+          continue
+        mul_op = self.reduceop.src[0].src[0] if has_cast else self.reduceop.src[0]
+        if not(isinstance(mul_op, LazyOp) and mul_op.op == BinaryOps.MUL and isinstance(mul_op.src[0], LazyBuffer) and mul_op.src[0].dtype == tc.dtype_in and isinstance(mul_op.src[1], LazyBuffer) and mul_op.src[1].dtype == tc.dtype_in):
+          continue
+        buf0, buf1 = self.bufs.index(mul_op.src[0]), self.bufs.index(mul_op.src[1])
+        buf0_strides, buf1_strides = self.sts[buf0].real_strides(), self.sts[buf1].real_strides()
+        axis_buf0 = [(i,self.full_shape[i],buf1_strides[i]) for i,s in enumerate(buf0_strides) if s == 0 and self.full_shape[i]%tc.dims[0] == 0 and i < self.first_reduce]
```
roelofvandijk, line 181: You can speed this up by only enumerating until `i == self.first_reduce` instead of checking all. 
`enumerate(buf0_strides[:self.first_reduce])`

tinygrad/codegen/optimizer.py
```diff
@@ -166,119 +166,33 @@ def hand_coded_optimizations(self):
...
+        axis_buf1 = [(i,self.full_shape[i],buf0_strides[i]) for i,s in enumerate(buf1_strides) if s == 0 and self.full_shape[i]%tc.dims[1] == 0 and i < self.first_reduce]
+        optim_conv2d = (self.shape_len-self.first_reduce) == 3 and self.full_shape[self.first_reduce+1]%2 == 1 and self.full_shape[self.first_reduce+2]%2 == 1 and max(self.full_shape[self.first_reduce+1:self.first_reduce+3]) < 21
+        if not(axis_buf0 and axis_buf1 and self.full_shape[self.first_reduce]%tc.dims[2] == 0 and self.full_shape[self.first_reduce] > tc.dims[2] and ((self.shape_len-self.first_reduce) == 1 or optim_conv2d)):
```
roelofvandijk, line 184: Maybe `if not all(.., ..,...)`

tinygrad/codegen/optimizer.py
flammit: Done!

tinygrad/codegen/linearizer.py
```diff
@@ -252,84 +272,35 @@ def end_loop(xx:List[Variable]):
...
+        # barrier for fast GEMM
+        self.uop(UOps.BARRIER, None, ())
+
+        # load WMMA input values
+        wmma_vals = [self.global_load(i+1, global_idxs+local_idxs+reduce_idxs+aliased_wmma_idxs[i]+full_upcast_idxs) for i in tuple(range(2))]
+        wmma_sz = self.tensor_core.thread_local_sizes
+        acc_reds = (len(wmma_vals[1])//wmma_sz[1])//(len(acc)//wmma_sz[2])
+        for i in range(len(acc)//wmma_sz[2]):
+          for j in range(acc_reds):
+            self.uop(UOps.WMMA, None, tuple(wmma_vals[0][(i*acc_reds+j)*wmma_sz[0]:(i*acc_reds+j+1)*wmma_sz[0]] + wmma_vals[1][(i*acc_reds+j)*wmma_sz[1]:(i*acc_reds+j+1)*wmma_sz[1]] + acc[i*wmma_sz[2]:(i+1)*wmma_sz[2]]), [self.bufs[0].device, self.tensor_core.dtype_in, self.tensor_core.dtype_out])
```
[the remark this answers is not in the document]
flammit: Done!

tinygrad/codegen/linearizer.py
flammit: Done!

tinygrad/codegen/linearizer.py
```diff
@@ -252,84 +272,35 @@ def end_loop(xx:List[Variable]):
...
       # reduce loop
       render_loop(reduce_idxs)
 
-      # barrier for fast GEMM
-      if self.use_tensor_cores: self.uop(UOps.BARRIER, None, ())
-
-      # compute local aliases
-      # TODO: this is garbage code and should be at least moved elsewhere
-      locals_to_store = []
-      for i in self.local_alias:
```
geohot, line 273: What replaces this? A key aspect of the old implementation is that it would work with locals as well as tensor cores

geohot:
"remove unnecessary barrier" this made it faster

geohot:
Really before we can merge this, we need better tests. I'll note the tensor cores should work as locals regardless of the platform (they currently do).

tinygrad/codegen/kernel.py
geohot: I don't ever want this. Stop suggesting it, it's harder to reason about in exchange for a stupidly small amount of speed. That tradeoff is never worth it.

geohot:
Is there a way to retain all current behavior but just parameterize the cores themselves? We also need testing around these codepaths, maybe merge that first?

tinygrad/codegen/kernel.py
roelofvandijk: Thanks for the nuance, won't mention it again. I could have been more clear: I suggested a dataclass because it because it is newer, a bit more flexible and has more intuitive equalization (NamedTuple does not check the name/type!).

That last thing can lead to insidious bugs on NamedTuples with the same number of members, e.g.:

```
from typing import NamedTuple

def LazyOp(NamedTuple):
  r: int
  f: Tuple

def TensorCore(NamedTuple):
  g: int
  h: Tuple

print(LazyOp(1, (1,2,3)) == TensorCore(1,(1,2,3)))

```

https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-case

tinygrad/codegen/linearizer.py
```diff
@@ -252,84 +272,52 @@ def end_loop(xx:List[Variable]):
...
+                  if var.expr is None and elc.max >= var.max:
+                    idx = idx.substitute({var:elc%(var.max+1)})
+                    elc //= (var.max+1)
+                idxs[n] = idx
+            if DEBUG >= 3: print(f"{localbuf_idx} alias {i}:", idxs)
+            self.global_store(localbuf_idx, idxs, self.global_load(i, idxs))
           self.uop(UOps.BARRIER, None, ())
+          wmma_vals = [self.global_load(i, global_idxs+local_idxs+reduce_idxs+aliased_wmma_idxs[i-3]+full_upcast_idxs) for i in self.local_alias if isinstance(self.bufs[i], LocalBuffer)]
+        else:
+          wmma_vals = [self.global_load(i+1, global_idxs+local_idxs+reduce_idxs+aliased_wmma_idxs[i]+full_upcast_idxs) for i in tuple(range(2))]
```
gijskoning, line 310: nit
```suggestion
          wmma_vals = [self.global_load(i+1, global_idxs+local_idxs+reduce_idxs+aliased_wmma_idxs[i]+full_upcast_idxs) for i in range(2)]
```

tinygrad/codegen/linearizer.py
```diff
@@ -378,15 +366,16 @@ def end_loop(xx:List[Variable]):
         # end the late reduce loop
         end_loop(end_local_idxs)
         self.load_cache.clear()
+        acc_idxs = global_idxs+local_idxs+fake_reduce_idxs+upcast_idxs
 
     # load latebufs
-    loaded_buffers.update({b:self.global_load(i, global_idxs+local_idxs+fake_reduce_idxs+upcast_idxs) for i,b in enumerate(self.bufs) if b not in self.earlybufs and i != 0 and b.__class__ is not LocalBuffer})
+    loaded_buffers.update({b:self.global_load(i, acc_idxs) for i,b in enumerate(self.bufs) if b not in self.earlybufs and i != 0 and b.__class__ is not LocalBuffer})
```
gijskoning, line 372:
```suggestion
    loaded_buffers.update({b:self.global_load(i, acc_idxs) for i,b in enumerate(self.bufs[1:], start=1) if b not in self.earlybufs and b.__class__ is not LocalBuffer})
```

tinygrad/codegen/linearizer.py
flammit: Done!

roelofvandijk:
If you want to exclude dataclass fields from the built-in __repr__, you can use `field(repr=False)`

https://stackoverflow.com/questions/68722516/exclude-some-attributes-from-str-representation-of-a-dataclass

tinygrad/codegen/linearizer.py
flammit: This was just a hack to make it work - I'll remove this when there's a good way to linked main buffer with various local ones.

tinygrad/codegen/kernel.py
flammit: Changed in the latest commit

flammit:
> Is there a way to retain all current behavior but just parameterize the cores themselves (current master is terrible with the two large ifs)? We also need testing around these codepaths, maybe merge that first?

I've rebased to master and moved the new shared memory handling to a separate PR that depends on this one (#1921).  I've also added a new test that checks that the tensor cores are triggered when appropriate and also generate correct answers.

Unfortunately, I can't think of a way to maintain the current `TC=2` behavior and have the HIP tensor cores pass the new test with just local dims.  If I could, I wouldn't have had to implement the thread segments/aliases to explicitly map the WMMA thread indices and upcasted WMMA elements.

flammit:
Moving this to draft mode until the requisite tests in #1925 are merged.

tinygrad/codegen/optimizer.py
```diff
@@ -168,119 +168,34 @@ def hand_coded_optimizations(self):
...
+
+        if not(isinstance(mul_op, LazyOp) and mul_op.op == BinaryOps.MUL): continue
+        if not(isinstance(mul_op.src[0], LazyOp) and mul_op.src[0].op == BufferOps.MEM and mul_op.src[0].arg.dtype == tc.dtype_in): continue
+        if not(isinstance(mul_op.src[1], LazyOp) and mul_op.src[1].op == BufferOps.MEM and mul_op.src[1].arg.dtype == tc.dtype_in): continue
+        buf0, buf1 = self.bufs.index(cast(LazyOp, mul_op.src[0].arg)), self.bufs.index(cast(LazyOp, mul_op.src[1].arg))
+        buf0_strides, buf1_strides = self.sts[buf0].real_strides(), self.sts[buf1].real_strides()
+        axis_buf0 = [(i,self.full_shape[i],buf1_strides[i]) for i,s in enumerate(buf0_strides[:self.first_reduce]) if s == 0 and self.full_shape[i]%tc.dims[0] == 0]
+        axis_buf1 = [(i,self.full_shape[i],buf0_strides[i]) for i,s in enumerate(buf1_strides[:self.first_reduce]) if s == 0 and self.full_shape[i]%tc.dims[1] == 0]
+        optim_conv2d = (self.shape_len-self.first_reduce) == 3 and self.full_shape[self.first_reduce+1]%2 == 1 and self.full_shape[self.first_reduce+2]%2 == 1 and max(self.full_shape[self.first_reduce+1:self.first_reduce+3]) < 21
```
gijskoning, line 186: <21 could be <20 btw. Since 20 %2 != 1 which is checked before that. Probably doesn't matter much

Also this doesn't work well yet for symbolic shapes, not sure you expect any symbolic shapes here. But if so, the max() doesn't work well with symbolic shapes and SomeVar < 21 will be true for any variable with its minimum below 21, but you want its max to be strictly below 21 as well.

flammit: Mmm this logic is on master already and I defer to @akretz to comment on the interactions between this and the symbolic shapes.

tinygrad/codegen/kernel.py
```diff
@@ -27,6 +27,7 @@ def __str__(self): return f"tensor_core<{self.device}, {self.dims}, {self.dtype_
   ],
   "HIP": [
     TensorCore(device="HIP", dims=[16,16,16], dtype_in=dtypes.half, dtype_out=dtypes.float, threads=[16,2], thread_local_sizes=[16,16,8], thread_local_aliases=[ [[-1], [0], [1]], [[-1], [1], [0]], [[0], [1], [2, -1]] ]),
+    TensorCore(device="HIP", dims=[16,16,16], dtype_in=dtypes.half, dtype_out=dtypes.half,  threads=[16,2], thread_local_sizes=[16,16,8], thread_local_aliases=[ [[-1], [0], [1]], [[-1], [1], [0]], [[0], [1], [2, -1]] ]),
```
geohot, line 30: You wouldn't be able to use `__builtin_amdgcn_wmma_f32_16x16x16_f16_w32` for this, but I don't see gating in cstyle

flammit: It actually works fine (auto-casts properly - same goes for FP16 in METAL), but yes I have the WMMA ops passing the dtype_in/dtype_out to be able to gate in cstyle when necessary.

geohot:
Does this fix https://github.com/tinygrad/tinygrad/issues/1967?

flammit:
> Does this fix #1967?
> 
> I see it does, but it does other bad things.
> 
> `*** 59 r_1_1_36_9_2_8_8_8_4 arg 3 sz [324, 1, 1] [32, 1, 1] OPs 2M/ 0.02G mem 0.00 GB tm 61.00us/ 1.75ms ( 43.51 GFLOPS, 6.72 GB/s)`
> 
> Why is that first global dim 324?

I believe the PR does fix #1967, but I've disabled it in the latest rebase.

The global dim starts as 324 (here's with no optimization):

`***   59 r_32_8_36_9_16                        arg   3 sz [324, 8, 32]       [1, 1, 1]    OPs      2M/   0.02G  mem  0.00 GB tm    422.79us/     7.97ms (    6.28 GFLOPS,    0.97 GB/s)`

On master, after the standard optimizations it is also generates globals with [324,1,1]

flammit:
> Does this fix #1967?
> 
> I see it does, but it does other bad things.
> 
> `*** 59 r_1_1_36_9_2_8_8_8_4 arg 3 sz [324, 1, 1] [32, 1, 1] OPs 2M/ 0.02G mem 0.00 GB tm 61.00us/ 1.75ms ( 43.51 GFLOPS, 6.72 GB/s)`
> 
> Why is that first global dim 324?

Sorry I take that back.  I've added the simplify_ones to the end of the TC path now and it looks better:

```
***   59 r_36_9_2_8_8_8_4                      arg   3 sz [9, 36, 1]         [32, 1, 1]   OPs      2M/   0.02G  mem  0.00 GB tm     32.00us/     1.88ms (   82.94 GFLOPS,   12.82 GB/s)```

geohot:
Can you break this into two PRs? This is a really hard PR to review because it's making a pretty deep change to untested code.

I think about half of this PR is "style" like switching the arg to UOps.WMMA. Can you submit all the refactor looking stuff to be merged first, then I can really tackle reviewing this.

flammit:
> Can you break this into two PRs? This is a really hard PR to review because it's making a pretty deep change to untested code.
> 
> I think about half of this PR is "style" like switching the arg to UOps.WMMA. Can you submit all the refactor looking stuff to be merged first, then I can really tackle reviewing this.

Done! The style stuff is now in #2014 on which this PR now depends.

Issue #1019, opened 2023-06-21
Some METAL CI tests failing


geohot:
Track down why `test_half_to_int8` and `test_output_padded_conv_transpose2d` fail in Metal CI.

Currently the whole dtype test is disabled in the CI config file, and `test_output_padded_conv_transpose2d` is skipped in `test_ops.py` Would be great to reenable them.

geohot:
Btw, whatever this is might be the only remaining issue from all Metal tests passing in CI ($200 bounty)

b7r6:
FWIW on my 16GB M2 Mac Mini (details in log) these failures don't seem to reproduce. I have an M1 MacBook that just needs a re-image, I'm happy to fire it up if it's likely to shake loose the issue.

Log: https://gist.github.com/b7r6/c2a0d228499fdeb3b7f885d1abf48994

geohot:
Yea, I can't reproduce it locally either, just on CI. Which points to some sort of cache thing. Do we know what the CI machine is?

Cydox:
Also can't replicate on a real machine (or in qemu with passed through GPU).

It's consistently the second element that gets turned into a `0`. See: [https://github.com/Cydox/tinygrad/actions/runs/5349366539/jobs/9700485564](https://github.com/Cydox/tinygrad/actions/runs/5349366539/jobs/9700485564)

Cydox:
Doing this test with varying lengths of tensors it only ever breaks if the length is 4.

The shader generated for this is a special case:
```
#include <metal_stdlib>
using namespace metal;
kernel void E_4n6(device char* data0, const device half* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
    float4 val1_0 = (float4)(((device half4*)data1)[0]);
    data0[0] = val1_0.x;
    data0[1] = val1_0.y;
    data0[2] = val1_0.z;
    data0[3] = val1_0.w;
    threadgroup_barrier(mem_flags::mem_threadgroup);
    if (0 == 0) {
  } /* local */
 /* global */
}
```

For the other lengths the more generic version is this:
```
#include <metal_stdlib>
using namespace metal;
kernel void E_10(device char* data0, const device half* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
{ int gidx0 = gid.x;  /* 10 */
    float val1_0 = data1[gidx0];
[...]

PR #1883, opened 2023-09-18
Tracking down metal ci failing tests

Qazalin:
related to [#1019](https://github.com/tinygrad/tinygrad/issues/1019)

### The issue

Metal CI tests were failing on [this test case](https://github.com/tinygrad/tinygrad/blob/master/test/test_dtype.py#L91). Specifically, in the case that casting a tensor with data ` [1, 2, 3, 4] ` and dtype `half` to a `unit8` returned `[1 0 3 4]`!

```
AssertionError: 
tensor [1 0 3 4] dtype dtypes.char does not match target [1, 2, 3, 4] with dtype dtypes.char
```

This only happened in the specific combination of scenarios described below:

### Reproducing

There are two parts to reproducing the bug:
1. Hardware / software compatibility
2. Specific inputs to `cast`

Github actions only run on Intel macs, [apple silicon has been in the roadmap for 1y+](https://github.com/github/roadmap/issues/528).

[Wikipedia's "Supported GPUs" list](https://en.wikipedia.org/wiki/Metal_(API)) says only some generations of Intel-based macs support Metal 3. Though the hardware on actions runs Metal 2, you can see it by running `echo $(system_profiler SPDisplaysDataType)`:

image

So at first it might seem like Metal 2's entire casting is broken. But only certain kernels that do [this upcasting optimization](https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/optimizer.py#L393) have the problem.

To see the kernels in action, try running the following program with `METAL=1 DEBUG=4`:

```py

data = [1,2,3,4]
src = Tensor(data, dtype=dtypes.float16)
t = src.cast(dtypes.int8)
print(src.realize(), t.realize())

data = [1,2,3,4,5,6,7,8]
src = Tensor(data, dtype=dtypes.float16)
t = src.cast(dtypes.int8)
print(src.realize(), t.realize())
```

If you try doing the second cast on Metal 2 it'll work but the kernel that gets generated for the first type of casting won't working.

### This PR's diff

Removing the hand-coded optimization was the shortest path to getting back a working CI, though I acknowledge it's suboptimal. With some guidance on how to improve the optimization, I can work on cleaning it up and making it work with Metal 2.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -190,9 +190,8 @@ jobs:
       run: pip install -e '.[metal,webgpu,testing]' --extra-index-url https://download.pytorch.org/whl/cpu
     - name: Test LLaMA compile speed
       run: PYTHONPATH="." METAL=1 python test/external/external_test_speed_llama.py
-    #- name: Run dtype test
-    #  run: DEBUG=4 METAL=1 python -m pytest -n=auto test/test_dtype.py
-    # dtype test has issues on test_half_to_int8
+    - name: Run dtype test
+      run: DEBUG=4 METAL=1 python -m pytest -n=auto test/test_dtype.py
     - name: Run metal ops test
       run: DEBUG=2 METAL=1 python -m pytest -n=auto test/test_ops.py
     - name: Run JIT test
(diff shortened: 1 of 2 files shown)


waifairer:
For one, this removes the optimization entirely. It doesn't attempt to apply the fix with any specificity to platform.

For two, the proposed solution here has already been reviewed as something we don't want to do (selectively disabling the optimization because we don't understand why it produces the wrong result).

If we can find what we are doing incorrectly in the metal kernel, just fix that.

Issue #1620, opened 2023-08-22
Bounty: Training RNN-T 


dc-dc-dc:
Add the training logic to `examples/mlperf/model_train.py` -> `def train_rnnt():`

[Links](https://github.com/mlcommons/training/blob/master/rnn_speech_recognition/pytorch/utils/librispeech.csv) to training data used in mlperf

MLPerf training rules [here](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#appendix-allowed-optimizers)

---
Current Bounty: $500
Bounty Locked: #2150

PR #1884, opened 2023-09-19
Training RNNT

DKormann:
I want to request locking the bounty in #1620.

I am going to implement training for RNNT on mlperf.

Progress made:

- [x] applying preexisting code for training data and rnnt implementation
- [x] calculating RNNT loss in numpy
- [x] calculating gradient back to the distribution given by RNNT
- [x] one training pass work
- [x] make whole epoch work
- [x] batch train step
- [ ] JIT everything

I am solving the sequence alignment in numpy since that is computationaly nonexpensive


examples/mlperf/rnnt/train.py
```diff
@@ -0,0 +1,367 @@
...
+  features = np.concatenate(seq, axis=1)[:, :, ::3]
+
+  # normalize
+  features_mean = np.zeros((features.shape[0], features.shape[1]), dtype=np.float32)
+  features_std = np.zeros((features.shape[0], features.shape[1]), dtype=np.float32)
+  for i in range(features.shape[0]):
+    features_mean[i, :] = features[i, :, :x_lens[i]].mean(axis=1)
+    features_std[i, :] = features[i, :, :x_lens[i]].std(axis=1, ddof=1)
+  features_std += 1e-5
+  features = (features - np.expand_dims(features_mean, 2)) / np.expand_dims(features_std, 2)
```
g1y5x3, line 66: So all the numpy operations here could be very easily transferred to tinygrad Tensors except the `librosa.stft` part. Especially if they are being called periodically, you would like to eventually JIT this part of the processing to speed it up.

examples/mlperf/rnnt/train.py
```diff
@@ -0,0 +1,367 @@
...
+    # plt.imshow(-distribution_grad[0,:,:,:].sum(-1))
+    # plt.show()
+    return Loss,distribution_grad
+
+# %%
+rnnt= RNNT()
+
+# %%
+opt = LAMB(rnnt.params)
+def train_step():
```
g1y5x3, line 337: This is where `@TinyJit` should be used. You feed `model`, `optimizer`, (`schedular` if there is one), `X` and `Y`

examples/mlperf/rnnt/train.py
```diff
@@ -0,0 +1,367 @@
...
+  def __call__(self, x, hc, m):
+    emb = self.emb(x) * m
+    x_, hc = self.rnn(emb.transpose(0, 1), hc)
+    return x_.transpose(0, 1), hc
+
+# %% [markdown]
+# # training
+
+# %%
+def calc_loss(enc,distribution):
```
wozeparrot, line 290: This 100% needs to be done in tinygrad, you can use a custom function for it if the load/store stuff is way too hard to implement.

This breaks the gradient as well no?

examples/mlperf/rnnt/train.py
DKormann: I just copied this code from https://github.com/tinygrad/tinygrad/blob/master/extra/datasets/librispeech.py can translate to tensors but this is all just a couple of seconds for the whole dataset.

examples/mlperf/rnnt/train.py
DKormann: I just copied this code from https://github.com/tinygrad/tinygrad/blob/master/extra/datasets/librispeech.py can translate to tensors but this is all just a couple of seconds for the whole dataset.

examples/mlperf/rnnt/train.py
DKormann: okay will do.

examples/mlperf/rnnt/train.py
DKormann: I decided to calculate the gradient completely manually here since its very fast and my tinygrad implementation was quite slow. A problem here is the gradient get scattered a lot and the AST gets very big. I saw this was a problem in the previous attempts.
Also METAL only supports only float32 that is not enought since the gradient calculation includes quite extreme numbers. Basically for each item in the sequence the activation goes /= 26 and than when in the back pass it gets normalized again. but for the calculation doesnt require a lot of flops so i thought its fine to do this part in python. 
```plt.imshow(distribution_gradient[0].sum(-1)```
should visualizes the gradient I calculated in numpy. This should correspond to Pr in the original RNNT [paper](https://arxiv.org/pdf/1211.3711.pdf) equation (19)
so after I just use the fake _loss to inject gradient into the tinygrad pipeline and let that backprop through the model.

Comment (DKormann) [reply examples/mlperf/rnnt/train.py:290]: I could put it in a Function if this is really required. Would it be okay to leave some computation in numpy though? If no numpy is allowed would need to find a way to make it fast and find a workaround for big numbers in f32. there might be a way to use the logarithm instead of the actual values but not sure since there are a lot of additions going on

**Comment by DKormann** [comment]
tried running full training run on dev dataset.:

```
there are 2703 samples in the dataset
Loss: 604.3656240488192
Loss: 665.9494883715234
Loss: 437.83098322440844
Loss: 245.89187415643826
Loss: 183.50476562205418
Loss: 505.6060908807693
Loss: 445.69168576997674
Loss: 531.4507851531254
Loss: 708.0824434831337
Loss: 526.3079312092368
Loss: 499.95122920531657
Loss: 409.8135678234397
Loss: 236.7812701606914
Loss: 543.4424774173025
Loss: 372.52430668576267
Loss: 351.7172608145389
AGX: exceeded compiled variants footprint limit
Loss: 258.150256892044
Loss: 477.6123497466345
Loss: 389.51918288627587
Loss: 332.31478235165525
Loss: 477.6289806059444
Loss: 284.17548737080136
Loss: 278.6477132753076
Loss: 354.3767992014807
Loss: 443.4080056850288
Loss: 319.19218230906375
Loss: 447.09278682180235
Loss: 335.46123048562043
Loss: 265.20053908305306
Loss: 232.33005003108414
Loss: 585.8823363490178
Loss: 290.7629459927819
<ipython-input-46-2db83b300b97>:21: RuntimeWarning: divide by zero encountered in log
  Loss = -np.log(P)
<ipython-input-46-2db83b300b97>:25: RuntimeWarning: divide by zero encountered in scalar divide
  alpha_grad[-1,-1] = -1/alpha[-1,-1]
<ipython-input-46-2db83b300b97>:38: RuntimeWarning: invalid value encountered in scalar multiply
  distribution_grad[0,u,t,label] += alpha_grad[t,u+1] * alpha[t,u]
<ipython-input-46-2db83b300b97>:33: RuntimeWarning: invalid value encountered in scalar multiply
  distribution_grad[0,u,t,-1] += alpha_grad[t+1,u] * alpha[t,u]
Loss: inf
Loss: nan
Loss: nan
Loss: nan
Loss: nan
```
single forward backward pass takes about 9s batch_size=1 no jit 8GB M1 
full projected epoch duration about 9h.
Most time is spend not in numpy. Tensor operations need to be optimized

DKormann:
I have found a way to normalise alpha and beta across the lattice diagonal. This makes the numbers not so extreme that they overflow on f32. It produces mathematically the same gradients and I can prove that if necessary.
It is needed to access memory in the diagonal form so thats where numpy is helpful but its possible to do that in tinygrad for sure and with different views It can also be done pretty performant. But for now its not the bottleneck for performance.

DKormann:
batch training step works.
doing a full epoch on dev dataset takes ~2 hours on m1 air

DKormann:
working on a problem now where the gradients become partially 0 after some training steps.
This is because ab becomes partially 0 because of overflow. 
alpha and beta are stable, but ab = alpha * beta is not so stable because alpha tends to be near zero at center to top right of the latice while beta is near or at zero at center to bottom left. Trying to find some kind of further normalisation to fix this problem.

DKormann:
thanks for all feedback im closing this one and moving to #2150

PR #1886, opened 2023-09-19
hip: fix const int handling for symbolic

flammit:
LLaMA is broken on HIP without this


chenyuxyz:
fyi nimlgen did the same change in #1877

flammit:
Closing as dup for #1877

PR #1888, opened 2023-09-20
cifar step time 65ms while stay above 94%

g1y5x3:
```
999   64.98 ms run,    3.52 ms python,   61.46 ms CL,  497.12 loss, 0.000086 LR, 6.76 GB used,  10382.36 GFLOPS
eval     9404.0/10000 94.04%,    0.40 val_loss STEP=1000
eval ema 9413.0/10000 94.13%,    0.40 val_loss STEP=1000
```
The most of the speed gain was from changing the `heuristics` in `lazy.py` threshold to 0.1 so that the one giant reduce kernel becomes way faster. But in the mean time, there's also a bunch of incremental speed gain from various PRs @chaosagent and @chenyuxyz submitted.  

Added a ema net based on the [orignal repo](https://github.com/tysam-code/hlb-CIFAR10/blob/ff53cac99dc71a66b1ada43d63d1307debd592a7/main.py#L442-L458) which gives some additional accuracy improvement. However, I tried to tune the hyp-param a little bit more to get rid of fixed random seed but after over 2000 runs, it's still quite challenging to achieve.

The rest of the big block changes was just to move all the global variables/functions inside `train_cifar()` to make it easier to set up hyper-param sweep.

### DIFF
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -25,29 +25,30 @@ from extra.dist import collectives
 BS, EVAL_BS, STEPS = getenv("BS", 512), getenv('EVAL_BS', 500), getenv("STEPS", 1000)
 
 # hyper-parameters were exactly the same as the original repo
-bias_scaler = 56
+bias_scaler = 55
 hyp = {
   'opt': {
-    'bias_lr':           1.65 * bias_scaler/512,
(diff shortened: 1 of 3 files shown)


geohot:
1541.45 ms -> 1265.51 ms on my base M1

PR #1890, opened 2023-09-20
refactor calculate_strides and acc_offsets in kernel

perryrjohnson:
I'm new to contributing to the tinygrad project so wanting to pick off some low hanging fruit in the codebase.

going through `tinygrad/codgen/` files and looking to propose a light 'Conceptual cleanup' here.

Proposing: separate the strides calculation logic into its own method (calculate_strides) to make the acc_offsets method more concise.  Also utilizing the zip and reversed here on list comprehension. feels like better readability as well.


chenyuxyz:
I think it requires more than just shuffling lines and renaming variables to qualify as a conceptual cleanup.

geohot:
Yea, this is not conceptual, this is style.

PR #1807, opened 2023-09-06
Basic GGUF parsing

jfhs:
WIP, parses metadata, can read f32 Tensor as numpy.

Seems like dequantize should happen in Tensor since q4_0 data location depends on Tensor op:
- https://github.com/ggerganov/ggml/blob/79ea7a41a388e9e7c85e426e912ab38399c44bfa/src/ggml.c#L8967
- https://github.com/ggerganov/ggml/blob/79ea7a41a388e9e7c85e426e912ab38399c44bfa/src/ggml.c#L9231
- https://github.com/ggerganov/ggml/blob/79ea7a41a388e9e7c85e426e912ab38399c44bfa/src/ggml.c#L11803

Expected output weights (dumped from llama.cpp):
```
[0.003525, 1.003525, 2.003525, 3.003525, 4.003525, 5.003525, 6.003525, 7.003525, 8.003525, 9.003525, 10.003525, 11.003525, 12.003525, 13.003525, 14.003525, 15.003525, 16.003525, 17.003525, 18.003525, 19.003525, 20.003525, 21.003525, 22.003525, 23.003525, 24.003525, 25.003525, 26.003525, 27.003525, 28.003525, 29.003525, 30.003525, 31.003525, 32.003525, 33.003525, 34.003525, 35.003525, 36.003525, 37.003525, 38.003525, 39.003525, 40.003525, 41.003525, 42.003525, 43.003525, 44.003525, 45.003525, 46.003525, 47.003525, 48.003525, 49.003525, 50.003525, 51.003525, 52.003525, 53.003525, 54.003525, 55.003525, 56.003525, 57.003525, 58.003525, 59.003525, 60.003525, 61.003525, 62.003525, 63.003525],
[...]

PR #1892, opened 2023-09-21
Implement q4_0 loader for gguf llama

Qazalin:
Adds a gguf loader to plug into Tinygrad. The gguf file format [(spec here)](https://github.com/philpax/ggml/blob/gguf-spec/docs/gguf.md) stores everything about the LLM in a single file. I've spot checked the output with the original [llama.cpp](https://github.com/ggerganov/ggml/blob/8ca2c19a3bb8622954d858fbf6383522684eaf34/include/ggml/ggml.h#L1844) implementation and the [Rust implementation](https://github.com/huggingface/candle).

- [x] Custom loader for binary -> spec values (since gguf isn't picklable)
- [x] Loading tensors via Disktensors
- [x] Dequantization fused in GPU

Dequantizations implemented: (gguf has many more dtypes, these are used in llama)
- [x] q4_0
- [x] q6K

Perf:
- [x] load and realize weights fast (~10 seconds)
- [ ] make llama run at 10 tok/s


wozeparrot:
See #1807, but basically the loader should be implemented with disktensors like the torch loader and the safetensors loader.

Qazalin:
@wozeparrot added an initial [implementation of DiskTensors](https://github.com/tinygrad/tinygrad/pull/1892/files#diff-ce5d4e7e02fbe5f8a87efb10f9dd65903a097cbf18db6f30b720806e6e5d657cR105). Can load the buffers correctly and cast. reshapes working for all layers is WIP.

 I also added a custom loader extending `pickle.Unpickler` - Though the serialization gguf uses isn't compatible with pickle's default `load` so we need to override it for this one

geohot:
disk tensor bug should be fixed I think

Qazalin:
I'm pausing this to work on #2011 - Right now without bitcasting the loading is very memory intensive. We need to copy the disk tensors hundreds of times in chunks.
i did a prototype version of loading `t` once and then dequanitzing everything in the GPU and it was way more manageable _(though inaccurate)_.

Qazalin:
been stale too long, will come back to working on it

PR #1894, opened 2023-09-21
Add legend to graph

SzymonOzog:
image


geohot:
I don't want this in every graph

PR #1906, opened 2023-09-23
Refactor Conv2d/ConvTranspose2d into a single parent class

Qazalin:


tinygrad/nn/__init__.py
```diff
@@ -39,33 +39,36 @@ def __call__(self, x:Tensor):
...
     self.stride, self.padding, self.dilation, self.groups = stride, padding, dilation, groups
-    self.weight = Tensor.kaiming_uniform(out_channels, in_channels//groups, *self.kernel_size, a=math.sqrt(5))
+    self.weight = self.initialize_weight(out_channels, in_channels, groups)
     assert all_int(self.weight.shape), "does not support symbolic shape"
     bound = 1 / math.sqrt(prod(self.weight.shape[1:]))
     self.bias = Tensor.uniform(out_channels, low=-bound, high=bound) if bias else None
+  def initialize_weight(self, out_channels, in_channels, groups): raise NotImplementedError
+  def __call__(self, x): raise NotImplementedError
 
+class Conv2d(BaseConv):
```
geohot, line 53: Instead of BaseConv, want to just have Conv2d be the base?

geohot:
Cool, change looks good. Is there good test coverage on this?

Qazalin:
we have tests covering both convs via comparing them to the torch version https://github.com/tinygrad/tinygrad/blob/master/test/test_nn.py#L112

they don't test the internals of `initialize_weight` but are sufficient to catch it if the logic is wrong.
`output_padding`'s initialization is also tested here https://github.com/tinygrad/tinygrad/blob/master/test/test_nn.py#L191

Qazalin:
The only thing we didn't have test coverage for was this:
```py
assert all_int(self.weight.shape), "does not support symbolic shape"
```
but it seems to be a mypy thing rather than logic #1859

PR #1916, opened 2023-09-25
remove image hacks from realize

geohot:


geohot:
Done in a different way

Issue #1838, opened 2023-09-10
Challenge: Write a function to convert ShapeTracker into MovementOps that produce that ShapeTracker


geohot:
Might not always be possible, but clearly possible for all ShapeTrackers we have.

Generated set doesn't have to be the same set that created it.

fruitbox12:
To convert a ShapeTracker into MovementOps that produce that specific ShapeTracker, you would need to create a function that takes a ShapeTracker as input and returns a set of MovementOps operations that, when applied, would result in the desired ShapeTracker state. 
```py
class ShapeTracker:
    def __init__(self, shape):
        self.shape = shape
        self.movements = []

    def apply_movement(self, movement):
        self.movements.append(movement)

    def get_shape(self):
        # Apply all movements to get the final shape
        final_shape = self.shape.copy()
        for movement in self.movements:
            final_shape = movement.apply(final_shape)
        return final_shape

class MovementOps:
    def __init__(self):
        self.operations = []

    def apply(self, shape):
        # Apply all operations to the shape
        for op in self.operations:
            shape = op(shape)
        return shape

def convert_to_movement_ops(shape_tracker):
    movement_ops = MovementOps()
    for movement in shape_tracker.movements:
        movement_ops.operations.append(movement.apply)
    return movement_ops

# Example usage:
if __name__ == "__main__":
    initial_shape = [0, 0, 0]  # Replace with your initial shape
    shape_tracker = ShapeTracker(initial_shape)
    
    # Add movements to the shape_tracker
    # For example:
    shape_tracker.apply_movement(lambda shape: [shape[0] + 1, shape[1], shape[2]])  # Move in the X-axis

    # Convert ShapeTracker to MovementOps
    movement_ops = convert_to_movement_ops(shape_tracker)

    # Now you can use movement_ops to reproduce the ShapeTracker's final shape
    final_shape = movement_ops.apply(initial_shape)
    print("Final Shape:", final_shape)
```

Qazalin:
After `.simplify()` the ShapeTracker stacks up this information in its `views`. We need to ask what `MovementOps` would result in the view's internals in a sequence. This prototype supports `permute` and `reshape`:

```py
import itertools
from typing import List, Tuple

from tinygrad.ops import MovementOps
from tinygrad.shape.shapetracker import ShapeTracker, strides_for_shape

# Permute reorders shape and strides. Because we don't preserve the shape before permute, we have to reverse engineer the permute axis from the strides and the shape of the view
[...]

PR #1917, opened 2023-09-25
Draft: Shapetracker converter

madt2709:
Fixes https://github.com/tinygrad/tinygrad/issues/1838. 

This a first draft but keen to get some initial feedback.

To do: 
- Add more tests for more complicated cases. 
- Remove the uglyness with deepcopy + initial call
- Make it work on all backends: DONE

EDIT: It's pretty clear I need to improve the gates for each op to be identified. In particular for strides and shrinks.

### DIFF
--- a/extra/shape_tracker_converter.py
+++ b/extra/shape_tracker_converter.py
@@ -16,32 +16,17 @@ def convert_st_to_movement_ops(st: ShapeTracker, prev_ops: List[Tuple[MovementOp
   # if prev_ops == []: st = deepcopy(st)
   # Stop space from polluting. TO DO: remove this line.
   if initial_call: prev_ops = []
-  if (strides := st.views[-1].strides) != (default_strides:= strides_for_shape(st.views[-1].shape)):
-    # only identify negative strides since technically the rest is a combination of other movement ops.
-    if any(i < 0 for i in strides):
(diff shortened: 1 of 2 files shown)


extra/shape_tracker_converter.py
```diff
@@ -0,0 +1,69 @@
...
+    # identify reshapes by multiple views. TO DO: this does not track "vanilla" reshapes where no movement op has been applied before.
+    if prod(st.views[-1].shape) == prod(st.views[-2].shape):
+      prev_ops.append((MovementOps.RESHAPE, st.views[-1].shape))
+      prev_st = st.reshape(st.views[-2].shape)
+      # remove "duplicate views"
+      prev_st.simplify()
+      return convert_st_to_movement_ops(prev_st, prev_ops, initial_call=False)
+
+  # ops have been identified and appended in reverse order
+  return prev_ops[::-1]
```
ngdangdat, line 69:
```suggestion
  return prev_ops[::-1]

```

madt2709: if the linter doesn't complain, i'm not going to make a formatting change tbh

madt2709:
This is fairly tricky. I'm going to try and optimize reshape first and then look into doing this.

PR #1918, opened 2023-09-25
make lazy simple

geohot:


geohot:
Out of date, maybe some good ideas here but not needed

PR #1923, opened 2023-09-25
fix transfer to mapped buffers

### LINKED ISSUE #1922: Copying a sliced disktensor to metal ends up with the wrong data
the expected behavior is that for disktensor `a`, `a.to(device)` produces an equivalent tensor in shape **and data**

below is a minimal repro of the issue with metal specifically. Other devices work as expected:
```py
import struct
import numpy as np

raw = np.random.randint(0, 256, size=1000, dtype=np.uint8).tolist() # [0, 255
with open("my_binary.bin", "wb") as f:
    for r in raw: f.write(struct.pack("B", r))

def run_on_device(device):
  t = Tensor.empty(len(raw), dtype=dtypes.uint8, device="disk:my_binary.bin")
  print(f"slice[  :10] original tensor", t.numpy()[:10].tolist())
  a = t[10:20]
  print("slice[10:20] loaded in disk:", a.numpy().tolist())
  dev = a.to(device)
  print("slice[10:20] loadd to", dev.device, dev.numpy().tolist())

run_on_device("CPU")
print("-----------------")
run_on_device("GPU")
print("-----------------")
run_on_device("METAL") # Notice how it keeps the shape of `a` but the offset is incorrect (starts at 0 instead of 10)
```

nimlgen:
Closes #1922.

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -1,9 +1,11 @@
 import numpy as np
 import torch
+import struct
 import unittest, copy
 from tinygrad.tensor import Tensor, Device
 from tinygrad.helpers import dtypes
 from extra.gradcheck import numerical_jacobian, jacobian, gradcheck
+from extra.utils import temp
 
 x_init = np.random.randn(1,3).astype(np.float32)
 U_init = np.random.randn(3,3).astype(np.float32)
@@ -239,6 +241,12 @@ class TestTinygrad(unittest.TestCase):
   def test_tensor_copy(self):
     x = copy.deepcopy(Tensor.ones((3,3,3)))
     np.testing.assert_allclose(x.numpy(), np.ones((3,3,3)))
+
+  def test_copy_from_disk(self):
+    t = Tensor.randn(30, device="CPU").to(f"disk:{temp('test_copy_from_disk')}")
+    a = t[10:20]
+    dev = a.to(Device.DEFAULT)
+    np.testing.assert_allclose(a.numpy(), dev.numpy())
 
 if __name__ == '__main__':
   unittest.main()
--- a/tinygrad/lazy.py
+++ b/tinygrad/lazy.py
@@ -235,12 +235,14 @@ class LazyBuffer:
   def fromCPU(x: np.ndarray) -> LazyBuffer:
     return LazyBuffer("CPU", ShapeTracker.from_shape(x.shape), LoadOps, LazyOp(LoadOps.EMPTY, (), None), dtypes.from_np(x.dtype), {}, RawNumpyBuffer.fromCPU(x))
 
+  def prepare_transfer(self):
+    self_casted = self.e(UnaryOps.CAST, arg=(dtypes.from_np(self.dtype.np), False)) if dtypes.from_np(self.dtype.np) != self.dtype else self
+    return self_casted.contiguous().realize().realized
+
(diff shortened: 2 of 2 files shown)


Qazalin (commented):
can confirm this fixed it - thanks!

test/test_tensor.py
```diff
@@ -240,5 +241,14 @@ def test_tensor_copy(self):
     x = copy.deepcopy(Tensor.ones((3,3,3)))
     np.testing.assert_allclose(x.numpy(), np.ones((3,3,3)))
 
+  def test_copy_from_disk(self):
+    raw = np.random.randint(0, 256, size=30, dtype=np.uint8).tolist() # [0, 255
+    with open("/tmp/test_my_binary.bin", "wb") as f:
```
wozeparrot, line 246: This should probably use `temp` like here: https://github.com/nimlgen/tinygrad/blob/4a9cdd9c4ffacda26f537ac2c19d44a4098e4601/test/unit/test_disk_tensor.py#L95

nimlgen: done

test/test_tensor.py
```diff
@@ -240,5 +242,14 @@ def test_tensor_copy(self):
     x = copy.deepcopy(Tensor.ones((3,3,3)))
     np.testing.assert_allclose(x.numpy(), np.ones((3,3,3)))
 
+  def test_copy_from_disk(self):
+    raw = np.random.randint(0, 256, size=30, dtype=np.uint8).tolist() # [0, 255
+    with open(temp("test_copy_from_disk"), "wb") as f:
+      for r in raw: f.write(struct.pack("B", r))
+    t = Tensor.empty(len(raw), dtype=dtypes.uint8, device=f"disk:{temp('test_copy_from_disk')}")
```
Qazalin, line 249: can be a one-liner:
```suggestion
  t = Tensor.randn((200), device="CPU").to(f"disk:{temp('test_copy_from_disk')}")
```

test/test_tensor.py
```diff
@@ -240,5 +242,14 @@ def test_tensor_copy(self):
...
+    a = t[10:20]
+    dev = a.to(Device.DEFAULT)
+    assert a.numpy().tolist() == dev.numpy().tolist()
```
Qazalin, line 252:
```suggestion
    np.testing.assert_allclose(a.numpy(), dev.numpy())
```

geohot:
Why is this draft, any issue with it?

Qazalin:
I don't think this fixes everything - slicing a 2d tensor on the second dimension has the same problems:
```py
l =  [[27, 0, 137, 119, 149, 157, 169, 229, 170, 105, 4, 103, 181, 117, 104, 99, 137, 158], [31, 0, 157, 201, 121, 201, 146, 21, 186, 194, 92, 89, 138, 217, 185, 102, 112, 85]]
a = Tensor(l, device="CPU", dtype=dtypes.uint8).to(f"disk:{temp('test_copy_from_disk_2')}")
d = a[:, :2]
print(d.numpy()) # <-- This is wrong, (137, 119) instead of (31, 0)
print(d[1].numpy()) # <-- This is correct, (31, 0)

# Comparing to a normal tensor:
a = Tensor(l, dtype=dtypes.uint8)
d = a[:, :2]
print(d.numpy())
```

PR #1929, opened 2023-09-27
fix hipgraph exec

nimlgen:
Fix+simplify exec for HIP.


wozeparrot:
We really need HIP testing in CI

PR #1934, opened 2023-09-28
fix test_collectives

wozeparrot:


geohot:
Neat. reset jit should probably be a function

PR #1935, opened 2023-09-28
with Tensor.train()

g1y5x3:
most of the line addition/removal are caused by the indentation when applying `with Tensor.train()`


geohot:
Thanks!

PR #1944, opened 2023-09-29
simplify CacheCollector

nimlgen:
Cache collector is a bit simpler (i hope) and more performant.
- SD 6.54 -> 6.05 GB
- hlb_cifar 2.40 -> 2.04 GB

Others are the same or slightly better.

This is an intermediate step while I am looking on how to embed this into lazy, but this might help with unet training.


geohot:
Confirmed no impact to SD speed and reduced memory usage

PR #1948, opened 2023-09-30
Extend reshape hack

madt2709:
Avoid merge views in certain symbolic cases as well.

This test:

```python
  def test_expand_into_reshape_with_symbols(self):
    vi = Variable("i", 1, 5)
    a = Tensor([[1], [2], [3]]).expand((3, 4))
    a = a.reshape(3, vi)
    assert a.shape == (3, vi)
    assert a.lazydata.var_vals == {vi: 4}
```
currently fails on master due to a type error. Need to think if thats a bug or intentional

### DIFF
--- a/tinygrad/shape/view.py
+++ b/tinygrad/shape/view.py
@@ -99,7 +99,8 @@ class View:
     # check if this is adding or removing 1s with [x for x in self.shape if x != 1] == [x for x in new_shape if x != 1].
     # extend the above to include cases like (3,4) --> (3,i) where we add or remove 1s and add/remove a "dummy" variable.
     # NOTE: this is optional, but removes most calls to (expensive!) merge_views (with mask, not optional)
-    variable_added_or_removed = (sum(1 for a,b in zip_longest([x for x in self.shape if x != 1],[x for x in new_shape if x != 1]) if a != b) == 1)
+    diff_between_shapes = [(a,b) for a,b in zip_longest([x for x in self.shape if x != 1],[x for x in new_shape if x != 1]) if a != b]
+    variable_added_or_removed = (len(diff_between_shapes) == 1) and any(isinstance(x, Node) for x in diff_between_shapes[0])
     if ([x for x in self.shape if x != 1] == [x for x in new_shape if x != 1]) or variable_added_or_removed:
       new_strides: List[sint] = [y for x,y in zip(self.shape, self.strides) if x != 1]
       new_strides_tuple: Tuple[sint, ...] = tuple([0 if x == 1 else new_strides.pop(0) for x in new_shape])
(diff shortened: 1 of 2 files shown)


tinygrad/shape/view.py
```diff
@@ -97,7 +98,7 @@ def reshape(self, new_shape: Tuple[sint, ...]) -> Optional[View]:
 
     # check if this is adding or removing 1s (only)
     # NOTE: this is optional, but removes most calls to (expensive!) merge_views (with mask, not optional)
-    if [x for x in self.shape if x != 1] == [x for x in new_shape if x != 1]:
+    if sum(1 for a,b in zip_longest([x for x in self.shape if x != 1],[x for x in new_shape if x != 1]) if a != b) <= 1:
```
geohot, line 101: I don't understand this, what does this catch that the old one didn't?

madt2709: This catches reshapes like (3,4) --> (3, i) where an int gets replaced with a Variable and the other way round.

madt2709: Added a comment for this.

geohot: oh I would handle this explicitly then

madt2709: updated it to (hopefully!) be more readable

tinygrad/shape/view.py
```diff
@@ -95,9 +96,12 @@ def reshape(self, new_shape: Tuple[sint, ...]) -> Optional[View]:
     # after the asserts, it's okay to check contiguous
     if self.contiguous: return View.create(new_shape)
 
-    # check if this is adding or removing 1s (only)
+    # check if this is adding or removing 1s with [x for x in self.shape if x != 1] == [x for x in new_shape if x != 1].
+    # extend the above to include cases like (3,4) --> (3,i) where we add or remove 1s and add/remove a "dummy" variable.
     # NOTE: this is optional, but removes most calls to (expensive!) merge_views (with mask, not optional)
-    if [x for x in self.shape if x != 1] == [x for x in new_shape if x != 1]:
+    diff_between_shapes = [(a,b) for a,b in zip_longest([x for x in self.shape if x != 1],[x for x in new_shape if x != 1]) if a != b]
```
geohot, line 102: Huh? This is so complex looking!

madt2709: re read this and its complete garbage - sorry! 

Updated it again. 3rd time lucky 🤞

geohot:
It's more readable now, but I think it should be 2 lines instead of 4, maintaining the readability and doing less useless compute.

madt2709:
Alright it's now 2 lines instead of 4

geohot:
Still hard to read, going to close.

PR #1950, opened 2023-09-30
simplify line

roelofvandijk:
Note that this could be `zip(m,t)`, but is not, to save expensive `__getitem__` calls on t if `mm==0`. 

Failing test jobs should be the intel keys issue.


geohot:
Is this actually faster?

roelofvandijk:
Yes. Iterates directly over the iterators (instead of using `len` and indexing).

PR #1951, opened 2023-09-30
fix llvm nan/inf const

### LINKED ISSUE #1937: Fix CONST in LLVM
`  if isinstance(Device[buffer.device], Compiled) and buffer.device not in ["LLVM"]:  # consts are broken in LLVM in NaN/inf`

nimlgen:
The `fast` flag implies both `nnan` and `ninf`. These 2 lead to different results when handling `inf`s/`nan`s. To remove this workaround from lazy, we need to drop these flags.

Closes #1937


geohot:
Thanks!

PR #1952, opened 2023-10-01
[WIP] remove CacheCollector

nimlgen:


nimlgen:
I've come up with something like this. It still needs a bit of cleaning up, and I have to write tests.

The major change here is the delayed execution in 'lazy'. Now, it builds programs, allocates buffers, and executes kernels. I'm not sure if this is a great design, but allocating right during execution can be painful because it involves complex logic to optimally determine which buffers are retained from outside and which belong to this tree (unless I'm missing something?). So, from the refcounting perspective, the simplest approach is to `del op`. So we build programs and execute them later.

For jitted memory usage, it remains the same for SD, llama, and gpt, but hlb_cifar consumes slightly more RAM (I'll look into it)

PR #1956, opened 2023-10-02
Fix cast error in render_load in wgsl

wpmed92:
I noticed an error when running `stable_diffusion` on `webgpu` (but it can affect any other model when running on `webgpu`):

Screenshot 2023-10-02 at 13 49 19

The cast that is generated inside `render_load` is wrong, because `float` is not defined in wgsl, and because in wgsl this style of cast `(f32)(some_var)` is invalid. The correct syntax is `f32(some_var)`.

I added `render_cast` to wgsl and modified the cstyle version as well to solve this.

### DIFF
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -40,6 +40,7 @@ class CStyleLanguage(NamedTuple):
 
   # returns a str expression of the casted xs with the given type
   def render_cast(self, x:List[str], var_dtype:DType) -> str:
+    if len(x) == 1: return f"({var_dtype.name})({x[0]})"
     assert len(x) == var_dtype.sz, f"cast is wrong size {len(x)} != {var_dtype.sz}"
     assert self.float4 is not None, "cast is not supported on this platform"
     if var_dtype == dtypes._float4: return f"{self.float4}({','.join(x)})"
@@ -47,9 +48,6 @@ class CStyleLanguage(NamedTuple):
     if var_dtype == dtypes._int2: return f"{self.float4.replace('float4', 'int2')}({','.join(x)})"
     raise NotImplementedError(f"no cast for {var_dtype}")
 
-  def render_cast_scalar(self, x:str, var_dtype:DType) -> str:
-    return f"({var_dtype.name})({x})"
-  
   # returns a str expression of the const with the given type
   def render_const(self, x:Union[float,int], var_dtype) -> str:
     if math.isnan(x): val = "NAN"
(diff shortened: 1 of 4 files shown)


geohot:
Why is this not just render cast?

wpmed92:
@geohot I refactored to use the existing `render_cast`

tinygrad/renderer/cstyle.py
```diff
@@ -61,10 +62,14 @@ def render_load(self, output_dtype, buf_name, buf_dtype, idx, local=False) -> st
       return f"read_imagef({buf_name}, smp, {idx})"
     if self.uses_vload and buf_dtype == dtypes.float16:
       return f"vload_half{'' if output_dtype.sz == 1 else str(output_dtype.sz)}(0, {buf_name}+{idx})"
-    cast = f"({output_dtype.name})" if output_dtype != buf_dtype else ""
+    
+    out_val = ""
```
wozeparrot, line 66: `out_val` shouldn't need to be defined here

wpmed92: Ah, right. Deleted it, two lines shorter.

geohot:
Can you add tests to prevent this regression? Ideally carefully reenabling the WGSL tests.

wpmed92:
Sure, I'll do it, but is it ok to make the a separate PR, so this one can go in, or should it be part of this?

wozeparrot:
It should be part of this PR.

wpmed92:
I added back the efficient net compilation test, because we have the efficient net demo, I think it's good to see we don't break that, also the `dtype` tests are working now, there was an issue with `bool` casts: `bool` storage buffers are not possible in webgpu, because a storage buffer has to be host-shareable by [spec](https://gpuweb.github.io/gpuweb/wgsl/#host-shareable-types), and bool type is not host-shareable:
_6.4.2 Host-shareable types:_

> As described in [§ 7.3 var Declarations](https://gpuweb.github.io/gpuweb/wgsl/#var-decls), the [store type](https://gpuweb.github.io/gpuweb/wgsl/#store-type) of [uniform buffer](https://gpuweb.github.io/gpuweb/wgsl/#uniform-buffer) and [storage buffer](https://gpuweb.github.io/gpuweb/wgsl/#storage-buffer) variables [must](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) be host-shareable.
...
A type is host-shareable if it is both [concrete](https://gpuweb.github.io/gpuweb/wgsl/#type-concrete) and one of: - a [numeric scalar](https://gpuweb.github.io/gpuweb/wgsl/#numeric-scalar) type, ... -a [fixed-size array](https://gpuweb.github.io/gpuweb/wgsl/#fixed-size-array) type, if it has [creation-fixed footprint](https://gpuweb.github.io/gpuweb/wgsl/#creation-fixed-footprint) and its element type is host-shareable, - a [runtime-sized](https://gpuweb.github.io/gpuweb/wgsl/#runtime-sized) array type, if its element type is host-shareable

 Step-by-step we can add back the other webgpu tests as well. Running the tests locally, the failures mainly boil down to these
- maximum of 8 storage buffer limit reached
- shader creation errors
- unsigned char not supported errors

geohot:
Hopefully the tests aren't flaky this time...

wpmed92:
Great! I'll keep on eye on the CI, and comment out the tests if they're flaky again, in the meantime will look into why the flakiness might have happened.

PR #1960, opened 2023-10-03
Rand works in jit

### LINKED ISSUE #1957: JIT doesn't work with dropout
Fix LoadOps.RAND in the JIT

nimlgen:
Add a hook to regenerate rand buffers when jitted.

Closes #1957


tinygrad/lazy.py
```diff
@@ -428,6 +428,8 @@ def _realize_empty(buffer: LazyBuffer) -> None:
 def _realize_rand(buffer: LazyBuffer) -> None:
   rng = np.random.default_rng(buffer.op.arg)
   buffer.realized = Device[buffer.device].buffer.fromCPU(rng.random(size=buffer.shape, dtype=np.float32).astype(dtype=buffer.dtype.np, copy=False), **buffer._device_extra_args()) # type: ignore
+  from tinygrad.jit import CacheCollector
+  CacheCollector.add(lambda args, vars, jit: args[0]._copyin(args[1].random(size=args[2], dtype=np.float32).astype(dtype=args[3], copy=False)), [buffer.realized, rng, buffer.shape, buffer.dtype.np], {})
```
geohot, line 432: That this line is duplicated shows me that an abstraction is wrong

nimlgen: You mean this random generation part or `CacheCollector.add`-staff?

geohot:
That's not the solution. The abstraction is broken.

geohot:
Sorry, removed

```
def _gen_rand(rng, shape, dt): return rng.random(size=shape, dtype=np.float32).astype(dtype=dt, copy=False)
def _realize_rand(buffer: LazyBuffer) -> None:
  rng = np.random.default_rng(buffer.op.arg)
  buffer.realized = Device[buffer.device].buffer.fromCPU(_gen_rand(rng, buffer.shape, buffer.dtype.np), **buffer._device_extra_args()) # type: ignore

  # Jit support
  from tinygrad.jit import CacheCollector
  CacheCollector.add(lambda args, vars, jit: args[0]._copyin(_gen_rand(*args[1:])), [buffer.realized, rng, buffer.shape, buffer.dtype.np], {})
  ```
  
  will add back and merge. The JIT API needs to be refactored with schedule, but it doesn't have to block this

nimlgen:
Btw I have started with experimenting of rewriting JIT with schedule in #1952. Don't know how got abstractions there yet, but yeah I also thought of something like that. Would appreciate if you comment on `compiled schedule`, do we wanna have something like that?

geohot:
Yea, we do want something like compiled schedule, but I think that type signature is wrong. I think this should be designed before it's coded, I want it to be a really big red diff.

geohot:
Hmm, also this doesn't test that the RNG behavior matches in the JIT and not the JIT. I don't think it does.

nimlgen:
Yeah, they do not match. Let's revert it, better to redesign jit. But since seed is changed in Tensor, I think even after redesign we gonna have some mismatch, no? Maybe set rng with a given seed in Tensor and just pass it?

PR #1962, opened 2023-10-03
bind value with Variable in symbolic shape and offset

chenyuxyz:
Proposal: only allowed "bound" Variables in symbolic shape. `ShapeTracker.var_vals` unbinds the underlying variables and composes the `var_vals` dict.

`matmul` and attention works now, `cat` produced wrong `uops`...

### DIFF
--- a/test/test_symbolic_jit.py
+++ b/test/test_symbolic_jit.py
@@ -11,8 +11,8 @@ class TestSymbolicJit(unittest.TestCase):
   def test_plus1(self):
     def f(a): return (a+1).realize()
     jf = TinyJit(f)
-    vi = Variable("i", 1, 10)
     for i in range(1, 5):
+      vi = Variable("i", 1, 10).bind(i)
       a = Tensor.rand(3, i)
       symbolic = jf(a.reshape(3, vi)).reshape(3, i).numpy()
       expected = f(a).numpy()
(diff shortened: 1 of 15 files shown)


geohot:
Cool!

So I think another part of this refactor is not allowing variable creation inside the JIT. The variables should have to be passed in to the top function.

geohot:
Oh this is so close! I started writing similar stuff in #2004

Anything I can do to help with this?

geohot:
Oh...I see the issue with bind and the cache. They really should be unbound in the ast before it's compiled

test/test_symbolic_jit.py
```diff
@@ -11,22 +11,23 @@ class TestSymbolicJit(unittest.TestCase):
   def test_plus1(self):
     def f(a): return (a+1).realize()
     jf = TinyJit(f)
-    vi = Variable("i", 1, 10)
     for i in range(1, 5):
+      vi = Variable("i", 1, 10).bind(i)
```
geohot, line 15: Can you autobind in reshape?

tinygrad/shape/shapetracker.py
```diff
@@ -79,6 +79,20 @@ def contiguous(self) -> bool: return len(self.views) == 1 and self.views[0].cont
   @property
   def shape(self) -> Tuple[sint, ...]: return self.views[-1].shape
 
+  @functools.cached_property
+  def var_vals(self) -> Dict[Variable, int]:
+    # Variable can come from shape or offset
```
geohot, line 84: Do you have to do this in all the views, not just -1?

chenyuxyz: i think you misread or i misunderstand, but i was doing on -1 only

chenyuxyz:
continue in #2017

PR #1973, opened 2023-10-05
simplify gpt2 example

chenyuxyz:
moved the reshape part to be in the same place as shrink so we only have one variable now. It will be further simplify with variable bind, and no need to manually update `var_vals` after (it's currently needed because we are replacing symbolic shapes with symbolic shapes that does not register value).

the `_jitted` functions in the else branch is only called once, so we don't need to create special jitted variant for them. The middle attention part is clearly not jittable as is anyway.

I am also making the similar changes to llama. It's trickier because conversational mode mixed jitted and unjitted functions that share the same kv cache.

### DIFF
--- a/examples/gpt2.py
+++ b/examples/gpt2.py
@@ -5,7 +5,7 @@ import argparse
 import numpy as np
 from tqdm import trange
 np.set_printoptions(linewidth=200)
-from typing import Optional, Dict
+from typing import Optional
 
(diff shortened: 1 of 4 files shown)


geohot:
Do we have a test on this confirming all the kernels are jitting?

chenyuxyz:
done. cifar and stable_diffusion need a separate init to have all_jitted test, and llama is broken with non-CI. I will fix these in a separate PR

PR #1986, opened 2023-10-05
unused: comment out unused methods

roelofvandijk:
This method is only used here: https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/linearizer.py#L235 which us commented out.

If that goes, this can go.


geohot:
Style

PR #1991, opened 2023-10-05
chore: add `py.typed` marker.

### LINKED ISSUE #1990: missing `py.typed` marker files to mark that the library code is typed.
visual studio code reports type stubs as missing when on max type strictness, and i believe it is because the packages are missing the required `py.typed` marker files in each package, which lets type checkers know that the library code is typed (no `*.pyi`). is it possible for these markers to be added?

[relevant pep section](https://peps.python.org/pep-0561/#packaging-type-information)

screenshot

vidhanio:
closes #1990.

can confirm that installing from git with my fork stops pylance from complaining, solving the issue.


ethanwater (approved):
pylance talks too much

PR #1992, opened 2023-10-06
chore: add `py.typed` to all packages

vidhanio:
supplementary to #1991, it seems `py.typed` is required in all packages.


geohot:
Hmm, now this is too complex

geohot:
All the files!

vidhanio:
valid concern, those are also generate-able at setup-time as to not clutter the source repository but that's also added complexity. totally understand the decision to revert though.

PR #1995, opened 2023-10-06
Allow multi-input model export

wpmed92:
Add support for exporting models that have `forward` or `__call__` with multiple Tensor inputs

### DIFF
--- a/extra/export_model.py
+++ b/extra/export_model.py
@@ -25,24 +25,26 @@ def compile_net(run:TinyJit, special_names:Dict[int,str]) -> Tuple[Dict[str,str]
 
   return functions, statements, {name:(size, dtype, key) for (name,size,dtype,key) in bufs.values()}, bufs_to_save
 
-def jit_model(model, the_input:Tensor) -> Tuple[TinyJit,Dict[int,str]]:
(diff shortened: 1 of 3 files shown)


extra/export_model.py
```diff
@@ -125,24 +127,25 @@ def export_model_webgpu(functions, statements, bufs, bufs_to_save, weight_names)
 }}
   """ + f"\n\nconst loadNet = async (device) => {{ return await fetch('net.safetensors').then(x => x.arrayBuffer()).then(x => setupNet(device, new Uint8Array(x))); }}"
 
-def export_model(model, input:Tensor, target:str):
+def export_model(model, inputs:List[Tensor], target:str):
```
dc-dc-dc, line 130: switch the ordering of `inputs` and `target` to make the api cleaner with a variable number arg list

wpmed92:
compile_efficientnet is in the CI so that tests the single input aspect, but for the multi input I'll add a unit test

extra/export_model.py
wpmed92: just fixed this

wpmed92:
I added a unit test for model export.

Issue #1781, opened 2023-09-06
openpilot elementwise fusion broke (fixed / not merged)


geohot:
b12b60af2063ff148f0bde8cee8985bd0cf39cf5 broke it somehow

Can you get the `v0.9.4` openpilot model to compile with 207 kernels down from 209? Run `openpilot/go.sh`

Kernels 69 and 120 should be fused with 68 and 119 respectively.

WarPromo:
I will give 1 dollar on top of that to whoever fixes it.

sahamrit:
logic behind why only certain movement ops are allowed for "replace_with_movement_ops" ?
```
  def _movement_op(self, st: ShapeTracker, op: MovementOps, arg: Union[Tuple[sint, ...], Tuple[Tuple[sint, sint], ...]]) -> LazyBuffer:
    if SHUFFLE_MOVEMENT_OPS and self.optype == BinaryOps and not self.realized and (op in {MovementOps.SHRINK, MovementOps.STRIDE, MovementOps.PERMUTE} or (op == MovementOps.RESHAPE and self.op.op in UnaryOps)) and not self.children:
      return self.op.replace_with_movement_ops([(op, arg)])
```

I was analysing this [MR](https://github.com/tinygrad/tinygrad/pull/1789) & found the exact reason as to why "allowing reshape being pushed up through BinaryOps got the kernel count back from 209 -> 207" [link to my comment](https://github.com/tinygrad/tinygrad/pull/1789#issuecomment-1747384954).

I need clarity on what basis are certain Movement Ops allowed through certain BinaryOps.

sahamrit:
@geohot as per my above comment, if the kernel count is to be reduced from 209 -> 207, reshape needs to be pushed up through BinaryOps which was disabled in [this commit - push reshape through UnaryOps](https://github.com/tinygrad/tinygrad/pull/723/commits/11732cae4af4b23e31b2fe99d1f4c90067bda7ee)

I assume you restricted reshape through UnaryOps, so that `test_no_binop_rerun` passes ? However tests like these will fail for permute, prereshape etc. So a better way of ensuring these test passes is needed instead of restricting "reshape through Unary only"?

lmk if my understanding is sound!

sahamrit:
Can we have a bounty lock on this issue?

sahamrit:
@wozeparrot can close this issue. Done with PR #2006.
However the PR was reverted due to an outstanding unrelated CI issue #2065. So not sure.

[...]

PR #2006, opened 2023-10-07
openpilot kernel fix from 209 to 207

sahamrit:
Issue #1781

push reshape 
1. always if the op is Unary
2. any of the parents is childless
3. never if all the base buffers are either realized or sourceless

we don't push pad or expand


geohot:
Very cool! This is making interesting changes to the schedule test. Want to add new schedule tests that confirm your behaviors?

This seems pretty fundamental, so I upped the bounty to $200

tinygrad/lazy.py
```diff
@@ -255,7 +255,7 @@ def splitted_shape(dim_aft_div): return self.shape[:dim_to_split] + (self.shape[
   # *** movement ops ***
 
   def _movement_op(self, st: ShapeTracker, op: MovementOps, arg: Union[Tuple[sint, ...], Tuple[Tuple[sint, sint], ...]]) -> LazyBuffer:
-    if SHUFFLE_MOVEMENT_OPS and self.optype == BinaryOps and not self.realized and (op in {MovementOps.SHRINK, MovementOps.STRIDE, MovementOps.PERMUTE} or (op == MovementOps.RESHAPE and self.op.op in UnaryOps)) and not self.children:
+    if SHUFFLE_MOVEMENT_OPS and not self.realized and self.optype == BinaryOps and len(self.children) <= 1 and (op is not MovementOps.PAD or all(x.op not in UNSAFE_PAD_OPS for x in self.op.get_lazyops())) and not all(x.realized for x in self.op.buffers):
```
geohot, line 258: Not sure you want `len(self.children) <= 1` here, `not self.children` is safer.

sahamrit:
> Very cool! This is making interesting changes to the schedule test. Want to add new schedule tests that confirm your behaviors?
> 
> This seems pretty fundamental, so I upped the bounty to $200.
> 
> Is it 206 kernels or 207 now? Are you completely confident that no reduces are rerunning?

It is 206 kernels now. I'll double verify the reduces and get back. New to the schedule API and tinygrad, so will need sometime with the tests and double checking stuff!

tinygrad/lazy.py
sahamrit: restored

geohot:
Tests failing

sahamrit:
some kernel counts are going up. changed the tests for now. Need to check if something is wrong with the logic or this increase is a needed tradeoff.

geohot:
Oh yea, those other counts going up reflect an issue, not an acceptable tradeoff

sahamrit:
@geohot lmk if this looks good
to summaries these are the changes
1. most of the condition in _movement_op for **replace_movement_ops** to be true is borrowed from _push_movement_ops.
2. We push the PAD only if the operations are safe with PAD
3. We push EXPAND but not beyond LoadOps LB (pushing beyond causes OOM)
4. We don't push movement ops if all the ast.buffers are realized or sourceless

test/test_schedule.py
```diff
@@ -169,15 +169,15 @@ def test_fold_conv_relu(self):
     # run
     img = Tensor.ones(2,3,64,64)
     out = c1(img).relu()
-    check_schedule(out, 1, [c1.weight, c1.bias])
+    check_schedule(out, 3, [c1.weight, c1.bias])
```
geohot, line 172: Why is this increasing?

test/external/external_test_opt.py
```diff
@@ -215,7 +215,7 @@ def test_fold_4convs_sgd(self):
       c3 = nn.Conv2d(8,16,3,bias=False)
       c4 = nn.Conv2d(16,32,3,bias=False)
       opt = optim.SGD(get_parameters([c1, c2, c3, c4]))
-      with CLCache(allowed=19):
+      with CLCache(allowed=22):
```
geohot, line 218: Why is this increasing?

tinygrad/lazy.py
```diff
@@ -344,7 +344,7 @@ def _push_movement_ops(srcs:Tuple[LazyBuffer, ...]) -> Tuple[LazyBuffer, ...]:
       assert isinstance(bx.op.src[0], LazyBuffer)
       bx = bx.op.src[0]
     # NOTE: can't push pads past anything where f(0, 0) != 0 or f(0) != 0
-    if mops and not bx.realized and bx.optype is BinaryOps and len(bx.children) <= 1 and (all(x[0] is not MovementOps.PAD for x in mops) or all(x.op not in UNSAFE_PAD_OPS for x in bx.op.get_lazyops())):
+    if mops and not bx.realized and bx.optype is BinaryOps and len(bx.children) <= 1 and (all(y[0] is not MovementOps.PAD for y in mops) or all(y.op not in UNSAFE_PAD_OPS for y in bx.op.get_lazyops())):
```
geohot, line 347: Is this a real change?

sahamrit: my bad. reverted

test/test_schedule.py
sahamrit: this was increasing because BinaryOp was rerunning after reshape was pushed up. I had restricted this behaviour to an extend via "not all(x.realized or len(x.op.src) == 0 for x in self.op.buffers)" but this wasn't sufficient. 

Using "map(functools.partial(get_movementroot, allow_contiguous = True), self.op.buffers))"  instead now!

sahamrit: attaching for reference, the re-rerunning of BinOp.
image

test/test_schedule.py
```diff
@@ -136,6 +136,7 @@ def test_cache_binaryop_reshaped(self):
     d = a.reshape(10,1)+b.reshape(10,1)
     check_schedule(d, 0, [c])
 
+  @unittest.skip("not pushing movement ops through sourceless buffers")
```
sahamrit, line 139: so the logic is that check till the **movementroot** of **ast.op.buffers** and if all of them are realized/sourceless don't push.

this will make the following tests pass
```
  def test_no_binop_rerun(self): return self._test_no_binop_rerun(lambda a,b: a*b, lambda a,b: (a*b).reshape(16, 16, 1))
  def test_no_binop_rerun_alt(self): return self._test_no_binop_rerun(lambda a,b: (a*b).reshape(16, 16, 1), lambda a,b: a*b)
  def test_no_binop_rerun_reduce_broadcast(self): return self._test_no_binop_rerun(lambda a,b: a.sum()+b, lambda a,b: a.sum().reshape(1,1)+b, allowed=2)
```

but (a.T * b.T).T will fail. We need to decide which is more important at the very bottom of the computation graph. 

btw this test is already disabled "test_no_binop_rerun_transposed" in external_test_opt

```
  @unittest.skip("this test started failing with the new change, based movementop issue")
  def test_no_binop_rerun_transposed(self): return self._test_no_binop_rerun(lambda a,b: (a.T*b.T).T, lambda a,b: a*b)
```

test/external/external_test_opt.py
sahamrit: Reshape which is pushed up is getting stuck b/w ReduceOps.SUM & BinaryOps.MAX. Because ReduceOp LB has 2 , reshape can't be pushed up more.  Hence Binary & Reduce are not getting fused.

I think a better solution of handling reshapes through BinOp can exist,. I will think more of what condition can be imposed more.
image

sahamrit: test_fold_2convs & 4convs have the same above problem

test/test_schedule.py
geohot: Disabling this is okay

test/external/external_test_opt.py
geohot: This needs to be fixed

geohot:
See above, transpose test disable is fine, but conv kernel count increase needs to be fixed.

tinygrad/lazy.py
```diff
@@ -257,7 +257,7 @@ def splitted_shape(dim_aft_div): return self.shape[:dim_to_split] + (self.shape[
   # *** movement ops ***
 
   def _movement_op(self, st: ShapeTracker, op: MovementOps, arg: Union[Tuple[sint, ...], Tuple[Tuple[sint, sint], ...]]) -> LazyBuffer:
-    if SHUFFLE_MOVEMENT_OPS and self.optype == BinaryOps and not self.realized and (op in {MovementOps.SHRINK, MovementOps.STRIDE, MovementOps.PERMUTE} or (op == MovementOps.RESHAPE and self.op.op in UnaryOps)) and not self.children:
+    if SHUFFLE_MOVEMENT_OPS and not self.realized and self.optype == BinaryOps and not self.children and (op is not MovementOps.EXPAND or all(x.op.op not in LoadOps for x in map(functools.partial(get_movementroot, allow_contiguous = True), self.op.buffers))) and (op is not MovementOps.PAD or (SHUFFLE_PAD_OPS and all(x.op not in UNSAFE_PAD_OPS for x in self.op.get_lazyops()))) and not all(x.realized or len(x.op.src) == 0 for x in map(functools.partial(get_movementroot, allow_contiguous = True), self.op.buffers)):
```
geohot, line 260: Feel free to use a few lines here to make this readable!

tinygrad/lazy.py
```diff
@@ -257,8 +257,14 @@ def splitted_shape(dim_aft_div): return self.shape[:dim_to_split] + (self.shape[
   # *** movement ops ***
 
   def _movement_op(self, st: ShapeTracker, op: MovementOps, arg: Union[Tuple[sint, ...], Tuple[Tuple[sint, sint], ...]]) -> LazyBuffer:
-    if SHUFFLE_MOVEMENT_OPS and self.optype == BinaryOps and not self.realized and (op in {MovementOps.SHRINK, MovementOps.STRIDE, MovementOps.PERMUTE} or (op == MovementOps.RESHAPE and self.op.op in UnaryOps)) and not self.children:
-      return self.op.replace_with_movement_ops([(op, arg)])
+    if SHUFFLE_MOVEMENT_OPS and not self.realized and self.optype == BinaryOps and not self.children:
+      root_bufs = map(functools.partial(get_movementroot, allow_contiguous = True), self.op.buffers)
```
geohot, line 261: `.base` might give you what you want here

tinygrad/lazy.py
```diff
@@ -257,8 +257,14 @@ def splitted_shape(dim_aft_div): return self.shape[:dim_to_split] + (self.shape[
...
+      # don't push if all ast buffers (movement roots) are realized or sourceless
+      if not all(x.realized or len(x.op.src) == 0 for x in root_bufs):
+        push_expand_safe = all(x.realized or x.op.op not in LoadOps for x in root_bufs)
```
geohot, line 264: When is pushing expands ever a good idea?

sahamrit: expand doesn't allocate new memory right ?
image

the only point OOM was hitting me was when it was pushed above LoadOps. Hence the checking condition

sahamrit: and coming onto the advantage. my understanding is pushing any movement op, so that BinOp comes next to each other (for fusion) is a good idea ?

tinygrad/lazy.py
sahamrit: done thanks!

tinygrad/lazy.py
sahamrit: am getting test failures if i don't push
image

geohot: That's likely a different issue.

Compare:
```
a = Tensor([[1,2,3]])
a = (a+1).expand(1000000,3)
```
with
```
a = Tensor([[1,2,3]])
a = a.expand(1000000,3)+1
```

In the bottom one we create a huge tensor, in the top we don't

sahamrit: i actually get your point now. i was looking things from point of kernel fusion. if only # of kernels were important expand push would have really helped (see below example). But the memory & compute is getting messed up.

```

a = Tensor([[[[[1,2,3]]]]])
b = Tensor([[[[[1,2,3]]]]])

c = (a*b).expand((1,1,1,100,3))
d = Tensor([[[[[1,2,3]]]]])

e = (c+d).expand((1,1,100,100,3))
f = Tensor([[[[[1,2,3]]]]])

g = (f-e).expand((1,100,100,100,3))
h = Tensor([[[[[1,2,3]]]]])

i = (g*h).expand((100,100,100,100,3))
i.realize()
```

with expand push
image

without expand push
image

sahamrit: i will disable expand and see what is causing the tests to fail.

sahamrit:
@geohot almost done. only need to analyse 1 test case where kernel count is going up. can we have a bounty lock. I maybe inactive for a day

sahamrit:
I noticed that pushing movement ops logic can be improved a lot. There are gaps in logic for sure from optimal. I would take that up, seperate from this issue

PR #2008, opened 2023-10-07
add var_vals to kopt with symbolic

nimlgen:
Should pass any valid val_vars when kopting. For kernels like that, uninitialized start_pos causes mem error (cuda + hip suffer, gpu seems defaulting to 0, so it works in CI)

```
__global__ void E_8_32_4(float* data0, const float* data1, const float* data2, const int start_pos) {
  int gidx0 = blockIdx.x; /* 8 */
  int lidx1 = threadIdx.x; /* 32 */
  float4 val0 = (float4)(*((float4*)(data1+(gidx0*128)+(lidx1*4))));
  float val1 = data2[start_pos];
  *((float4*)(data0+(gidx0*128)+(lidx1*4))) = (float4)make_float4((1.0f-(((val0).x<val1)+(val1<(val0).x))),(1.0f-(((val0).y<val1)+(val1<(val0).y))),(1.0f-(((val0).z<val1)+(val1<(val0).z))),(1.0f-(((val0).w<val1)+(val1<(val0).w))));
}
```

### DIFF
--- a/test/models/test_real_world.py
+++ b/test/models/test_real_world.py
@@ -7,14 +7,13 @@ from tinygrad.jit import TinyJit, JIT_SUPPORTED_DEVICE
 from tinygrad.ops import Device, GlobalCounters, LazyOp, LoadOps
 from tinygrad.helpers import CI, dtypes, getenv, prod
 from tinygrad.codegen.search import kernel_optimize_opts
-from tinygrad.lazy import var_vals_from_ast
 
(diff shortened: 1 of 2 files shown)


geohot:
Even though it's correct, this is the exact PR I wanted to avoid. By rethinking why we are passing something around, we often can refactor and reduce downstream complexity.

See #2009

nimlgen:
Will wait for 2009 merge to rebase on top of it. I think we might wanna call auto_opts from Linearizer?

geohot:
Lots of copy/paste

PR #2010, opened 2023-10-07
Disable ocelot cache + fix matvec in triton

SzymonOzog:
This should confirm if cache is the reason that triton tests were failing on master


SzymonOzog:
weird, doesn't happen on CI in my branch or locally - need to throw some temporary debugs

geohot:
I'll note I removed triton from master since tests were failing. It also broke the abstraction for what the renderer returns and an `if len == 3` thing isn't very clean

SzymonOzog:
oh, that's why it cant find the renderer, weird that my branch didn't catch the change. Is it fine to bring it back?

SzymonOzog:
I reverted the commit that disabled triton - you can merge this if you want to check if the tests were broken because of ocelot cache

SzymonOzog:
I think that the only solution to the `if len == 3` thing is either making triton a seperate runtime (more lines, more readable) or moving the compilation from renderer to `CUDAProgram` and `if getenv("TRITON")` there to triton compile since we need to pass triton specfic arguments to CUDA. Which one do you prefer?

geohot:
I want a decent abstraction first to handle things like "binary"

.github/workflows/test.yml
```diff
@@ -256,7 +256,7 @@ jobs:
           sudo apt update -y
           sudo apt install -y --no-install-recommends git g++ cmake ninja-build llvm-15-dev zlib1g-dev libglew-dev flex bison libfl-dev libboost-thread-dev libboost-filesystem-dev nvidia-cuda-toolkit-gcc
       - name: Cache gpuocelot
-        if: matrix.backend == 'cuda' || matrix.backend == 'ptx'|| matrix.backend == 'triton'
+        if: ${{false}} # matrix.backend == 'cuda' || matrix.backend == 'ptx' #|| matrix.backend == 'triton'
```
geohot, line 259: We need this cache or it's slow, why is it broken?

SzymonOzog: I've restored the cache - remainder as per our discussion in discord::triton_backend. Cache is probably broken on master and it needs to be cleared before merging this or it will time out again. I think if you remove the cache and merge a commit to this pr the tests should run on master

geohot:
Merged! Looks like the cache was already deleted.

geohot:
Hmm, why isn't `from_linearizer` deleted?

PR #2016, opened 2023-10-08
add `tinygrad.features` to setup.py

wozeparrot:
the features change broke using tinygrad as a package.


geohot:
Good catch. Can we test this in CI?

wozeparrot:
Might be possible? would need something that imports tinygrad as an installed package, and tries to run some tinygrad code. there are other issues tho, like hip is broken because it depends on `extra`

PR #2017, opened 2023-10-08
Variable.bind newer

chenyuxyz:
update based on #1962 making a new draft. (1) added Variable.bind, (2) removed var_vals from LazyBuffer, (3) implemented ShapeTracker.unbind and used in jit check

### DIFF
--- a/tinygrad/lazy.py
+++ b/tinygrad/lazy.py
@@ -31,7 +31,7 @@ def _ast_reduceops(op:LazyOp) -> LazyOp:
   if not src.realized:
     assert isinstance(src.op, LazyOp), "if not src.realized, then src.op must be a LazyOp"
     if MERGE_ELEMENTWISE_INTO_REDUCE and src.optype is BinaryOps and len(src.children) <= 1: src = src.op
-  return LazyOp(op.op, (src,), op.arg)
+  return LazyOp(op.op, (src,), tuple(a if isinstance(a, int) else a.unbind()[0] for a in op.arg) if op.arg is not None else None)
 
 # this supports late merging an upstream Reduce op and even an Elementwise op above that
 def _ast_binaryops(op:LazyOp, shape: Tuple[sint, ...]) -> LazyOp:
@@ -64,7 +64,7 @@ def _replace_bufferops(op:LazyOp) -> Tuple[LazyOp, List[LazyBuffer]]:
   replacements:Dict[LazyBuffer, LazyOp] = {}
   base_bufs = dedup([x.base for x in op.buffers if not x.is_unrealized_const()])
   for x in op.buffers:
-    st = x.st.simplify()
+    st = x.st.simplify().unbind()
     if x.base in base_bufs:
       replacements[x] = LazyOp(BufferOps.MEM, (), MemBuffer(base_bufs.index(x.base)+1, x.dtype, st))
     elif not x.realized and x.base.op.op == LoadOps.CONST:
@@ -79,7 +79,7 @@ def get_single_root(root:LazyBuffer) -> LazyBuffer: return get_single_root(cast(
 def get_movementroot(root:LazyBuffer, allow_contiguous=False) -> LazyBuffer: return get_movementroot(cast(LazyBuffer, root.op.src[0]), allow_contiguous) if not root.realized and (root.optype == MovementOps or (root.op.op == LoadOps.CONTIGUOUS and allow_contiguous and root.op.src[0].st.contiguous)) else root
 def get_movementroot_contiguous(x:LazyBuffer) -> LazyBuffer: return get_movementroot_contiguous(cast(LazyBuffer, x.op.src[0])) if not x.realized and x.op.op == LoadOps.CONTIGUOUS else (get_movementroot(x, True) if x.optype == MovementOps and x.st.contiguous else x)
 
-def var_vals_from_ast(ast:LazyOp) -> List[Variable]: return dedup(functools.reduce(operator.add, [list(x.arg.st.var_vals.keys()) for x in ast.get_lazyops() if x.op in BufferOps], []))
+def var_vals_from_ast(ast:LazyOp) -> List[Variable]: return dedup(functools.reduce(operator.add, [x.arg.st.vars() for x in ast.get_lazyops() if x.op in BufferOps], []))
 
 lazycache: WeakValueDictionary = WeakValueDictionary()
 def create_lazybuffer(device:str, st:ShapeTracker, optype:OpType, op:LazyOp, dtype:DType, base:Optional[LazyBuffer]=None):
@@ -165,7 +165,6 @@ class LazyBuffer:
     ret = []
     for x in op.buffers: ret += x.schedule(seen)
 
-    # TODO: unbind self.st shapetracker here?
     var_vals = dict(sorted(merge_dicts([self.st.var_vals] + [buf.st.var_vals for buf in op.buffers]).items(), key=lambda kv:cast(Variable,kv[0]).key))
 
     # run the ast and log the op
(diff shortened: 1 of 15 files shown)


chenyuxyz:
this is ready for review. I will do a followup to pass bound Variable in jit and remove the jit_ctx.

the method cache thing is more complicated than I thought, will attempt separately

tinygrad/shape/view.py
```diff
@@ -20,19 +20,29 @@ class View:
...
     strides = filter_strides(shape, strides) if strides else strides_for_shape(shape)
     contiguous = offset == 0 and mask is None and all(s1 == s2 for s1,s2 in zip(strides, strides_for_shape(shape)))
     return View(shape, strides, offset, mask, contiguous)
 
   @functools.lru_cache(maxsize=None)  # pylint: disable=method-cache-max-size-none
   def size(self): return prod([s.max if isinstance(s, Node) else s for s,st in zip(self.shape, self.strides) if st != 0])
 
+  def unbind(self) -> View:
+    variables:List[Variable] = []
+    for x in self.shape+self.strides+(self.offset,):
```
geohot, line 38: This should probably be reused code with st.var_vals

tinygrad/shape/view.py
```diff
@@ -20,19 +20,29 @@ class View:
   shape:Tuple[sint, ...]
   strides:Tuple[sint, ...]
   offset:sint
-  mask:Optional[Tuple[Tuple[sint, sint], ...]]
+  mask:Optional[Tuple[Tuple[int, int], ...]]
```
geohot, line 23: Are you sure this is okay? How does cat work?

tinygrad/ops.py
```diff
@@ -214,7 +214,7 @@ def __call__(self, rawbufs:List[RawBuffer], var_vals:Optional[Dict[Variable, int
 class Compiled:
   def __init__(self, buffer: Type[RawBuffer], linearizer_opts, renderer, runtime, synchronize=lambda: None, batch_exec=BasicBatchExecutor):
     self.buffer, self.linearizer_opts, self.renderer, self.runtime, self.synchronize, self.batch_exec = buffer, linearizer_opts, renderer, runtime, synchronize, batch_exec
-    self.method_cache: Dict[Any, ASTRunner] = {}
+    self.method_cache: Dict[LazyOp, ASTRunner] = {}
```
geohot, line 217: 👍

geohot:
I don't see var bufferops unbinding the shapetracker. No bound variables should ever be in the AST that goes into the linearizer, can you check this?

tinygrad/shape/view.py
```diff
@@ -33,7 +33,9 @@ def create(shape:Tuple[sint, ...], strides:Optional[Tuple[sint, ...]]=None, offs
   @functools.lru_cache(maxsize=None)  # pylint: disable=method-cache-max-size-none
   def size(self): return prod([s.max if isinstance(s, Node) else s for s,st in zip(self.shape, self.strides) if st != 0])
 
-  def vars(self) -> List[Variable]: return dedup(functools.reduce(operator.add, [x.vars() for x in self.shape+self.strides+(self.offset,) if isinstance(x, Node)], []))
+  def vars(self) -> List[Variable]:
+    flatten_mask = tuple(flatten(self.mask)) if self.mask is not None else tuple()
+    return dedup(functools.reduce(operator.add, [x.vars() for x in self.shape+self.strides+(self.offset,)+flatten_mask if isinstance(x, Node)], []))
```
ZenginU, line 38: Maybe use flatten here as well instead of functools.reduce ?

tinygrad/shape/view.py
chenyuxyz: refactored some vars logic

tinygrad/shape/view.py
chenyuxyz: no that was incorrect. fixed and adding mask as a source of variable. I think it's possible to prove that only check a subset of (shape, strides, offset, mask) suffices, but checking all 4 is easier to reason

chenyuxyz:
added the unbind in op shapetracker and verified method cache works now since all variables are free. A side note is that we also need to unbind the reduce op arg if it contains variables. We only have (Variable, NumNode, MulNode, SumNode) in symbolic shape, so I added unbind for MulNode and SumNode.

tinygrad/shape/view.py
chenyuxyz: flatten does like tuple because it's not Iterator, I also revert the use of flatten in flatten_mask. There's also itertools.chain, but I think these are close enough in terms of speed and readability

tinygrad/shape/symbolic.py
```diff
@@ -194,6 +205,7 @@ def __mod__(self, b: Union[Node, int]):
   def get_bounds(self) -> Tuple[int, int]:
     return (self.a.min*self.b, self.a.max*self.b) if self.b >= 0 else (self.a.max*self.b, self.a.min*self.b)
   def substitute(self, var_vals: Dict[VariableOrNum, Node]) -> Node: return self.a.substitute(var_vals) * (self.b if isinstance(self.b, int) else self.b.substitute(var_vals))
+  def unbind(self) -> Tuple[Node, None]: return self.substitute({v: v.unbind()[0] for v in self.vars() if v.val is not None}), None
```
geohot, line 208: Can this just be on the root class? I see it's copy pasted below

tinygrad/lazy.py
```diff
@@ -79,29 +79,28 @@ def get_single_root(root:LazyBuffer) -> LazyBuffer: return get_single_root(cast(
 def get_movementroot(root:LazyBuffer, allow_contiguous=False) -> LazyBuffer: return get_movementroot(cast(LazyBuffer, root.op.src[0]), allow_contiguous) if not root.realized and (root.optype == MovementOps or (root.op.op == LoadOps.CONTIGUOUS and allow_contiguous and root.op.src[0].st.contiguous)) else root
 def get_movementroot_contiguous(x:LazyBuffer) -> LazyBuffer: return get_movementroot_contiguous(cast(LazyBuffer, x.op.src[0])) if not x.realized and x.op.op == LoadOps.CONTIGUOUS else (get_movementroot(x, True) if x.optype == MovementOps and x.st.contiguous else x)
 
-def var_vals_from_ast(ast:LazyOp) -> List[Variable]: return dedup(functools.reduce(operator.add, [x.arg.st.var_vals() for x in ast.get_lazyops() if x.op in BufferOps], []))
+def vars_from_ast(ast:LazyOp) -> List[Variable]: return dedup(functools.reduce(operator.add, [x.arg.st.vars() for x in ast.get_lazyops() if x.op in BufferOps], []))
```
geohot, line 82: As you pointed out, what about vars in ReduceOps?

tinygrad/lazy.py
```diff
@@ -31,7 +31,7 @@ def _ast_reduceops(op:LazyOp) -> LazyOp:
   if not src.realized:
     assert isinstance(src.op, LazyOp), "if not src.realized, then src.op must be a LazyOp"
     if MERGE_ELEMENTWISE_INTO_REDUCE and src.optype is BinaryOps and len(src.children) <= 1: src = src.op
-  return LazyOp(op.op, (src,), op.arg)
+  return LazyOp(op.op, (src,), tuple(a.unbind()[0] if isinstance(a, (Variable, MulNode, SumNode)) else a for a in op.arg) if op.arg is not None else None)
```
geohot, line 34: Should this be here? Why ever allow bound variables in ReduceOps?

tinygrad/shape/symbolic.py
chenyuxyz: done

tinygrad/lazy.py
chenyuxyz: I will go through this PR tomorrow again to make sure everything is good, but I have some clarifying questions about your comments.

so I moved the unbind reduceop.arg logic into the `r()` function, and I think the `arg` of a reduceop can only be `new_shape`. So it cannot be None and cannot be tuples either. I updated the logic to unbind that new_shape before creating LazyOp, and keep the bound st in LazyBuffer.

since reduce arg is new_shape, which is a subset of the original shape, it should suffice to get variables from the shapetracker.

geohot: Oh sorry, I was thinking movementops. ReduceOp can't be none or tuples. I think the unbind in the _reduce_ops function you have now is correct.

chenyuxyz:
rebased to master.

so we can have bound variables in lazybuffer shapetracker and movementops, and no bound variables (unbind in bufferops creation and reduceop arg, no need for elementwise ops) in AST going into linearizer. And we don't need to change `Variable.substitute` since all symbolic are unbound after compile time.

also checked with DEBUG=4 on various symbolic op tests and tested with llama. I think this is good to go

PR #2018, opened 2023-10-08
Adds LoRA implementation example

big-c-note:
Run with `GPU=1 python3 examples/lora_mnist.py` or `python3 examples/lora_mnist.py`

The printed console output will explain as well.

- There is a simulated pre-trained model on the mnist dataset by training for one epoch. 
- Then, the model is "loraized" and fine-tuned on the class with the worst performance. 
- The script asserts that the frozen model parameters don't change after fine-tuning.
- You will see a performance increase on the worst class in the second performance printout. 
- You may see classification of some other classes get slightly worse, this is normal.

Possible to create a more dynamic implementation of loraize if that is interesting to you for this code base


geohot:
No particular reason this is notable, and uses wrong abstractions like saving parameters with numpy

PR #2020, opened 2023-10-08
Trying out M1 runner

wpmed92:
M1 runners are available, let's see if it runs and passes.


wpmed92:
Turns out they're not free... billed at 0.16$/minute. Closing.

PR #2024, opened 2023-10-09
[BUG]: llama vocab size from 32016 to 32000

big-c-note:
After downloading code llama 7b I hit the assert statement on line 310. 
 ```assert sp_model.vocab_size() == MODEL_PARAMS[model_gen][model_size]["args"]["vocab_size"]```

The `vocab_size` should be 32k for each model, and runs great

### DIFF
--- a/examples/llama.py
+++ b/examples/llama.py
@@ -209,7 +209,7 @@ MODEL_PARAMS = {
   },
   "code": {
     "7B": {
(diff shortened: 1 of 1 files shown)


big-c-note:
Realized I was running code llama 7b python as code llama 7b.. that was my issue. code llama 7b actually has size 32016

PR #2027, opened 2023-10-09
use "<" instead of "<=" in codegen for loop

chenyuxyz:
example output from `CLANG=1 DEBUG=4 python -m pytest -rA test/test_symbolic_ops.py -k test_add`

master
```
void E_s0(float* restrict data0, const float* restrict data1, const float* restrict data2, const int i) {
  for (int ridx0 = 0; ridx0 <= ((i*3)+(-1)); ++ridx0) {
    float val0 = data1[ridx0];
    float val1 = data2[ridx0];
    data0[ridx0] = (val0+val1);
  }
}

void E_3(float* restrict data0, const float* restrict data1) {
  for (int ridx0 = 0; ridx0 <= 2; ++ridx0) {
    float val0 = data1[ridx0];
    data0[ridx0] = val0;
  }
}
```
this PR
```
void E_s0(float* restrict data0, const float* restrict data1, const float* restrict data2, const int i) {
  for (int ridx0 = 0; ridx0 < (i*3); ++ridx0) {
    float val0 = data1[ridx0];
    float val1 = data2[ridx0];
    data0[ridx0] = (val0+val1);
  }
}

void E_3(float* restrict data0, const float* restrict data1) {
  for (int ridx0 = 0; ridx0 < 3; ++ridx0) {
    float val0 = data1[ridx0];
    data0[ridx0] = val0;
  }
}
```

for LLVM, I switched the branch indicator to use "<" too
also tested with WEBGPU locally with `WEBGPU=1 WGPU_BACKEND_TYPE=Metal python -m pytest -n=auto` and it has the same failures compared to master

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -198,8 +198,8 @@ class Linearizer(OptimizedKernel):
     # global and local loops
     def render_loop(xx:List[Variable]):
       self.loop_uops.update({x.expr:self.uop(UOps.LOOP, dtypes.int32, (
-        self.const(x.min) if isinstance(x.min, int) else cast(Variable, x.min).render(self.render_ops, self),
-        self.const(x.max) if isinstance(x.max, int) else cast(Variable, x.max).render(self.render_ops, self)), cachable=False) for x in xx if not isinstance(x, NumNode) and x.expr is not None})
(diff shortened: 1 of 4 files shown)


geohot:
Thanks! I think `<` is prettier than `<=`

PR #1958, opened 2023-10-02
KOPT: parallel compilation, a bunch of QoL stuff

chaosagent:
Forward port of #1424 + other useful kopt changes. without this patch, winograd can't really kopt.

preliminarily ready for review

Changes
- generate suggestions from hand_coded (moved to #2029)
  - without this, very hard to kopt wino
- limit number of locals and upcasts for auto opts, limit allowed uops in search
  - rules out kernels that take forever to compile
- mp.Pool parallel compilation
  - don't crash master process if compilation crashes
  - fast!

TODO

- [ ] play well with GlobalCounters
- [ ] measure variation to reduce default number of times to measure per opt
- [ ] improve early-stopping for nevergrad search
- [ ] need to adjust kopt tests
- [ ] early fork pool

Anticipating questions:
- why not have hand_coded emit auto opts?
  - the order of operations done in auto opts and hand_coded are different; need to reconcile
  - need to figure out how to only do this partially, in order to maintain ease of writing things in hand_coded
  - need to figure out how to play well with tensor cores
  - will do it eventually, this set of changes is quite useful to have in-tree
- this is a lot of lines!
  - does kopt even belong in tinygrad/ instead of extra/...
    - we can have compile_schedule(t.schedule()), which can be done in parallel!
- can this be split?
  - yeah... inconvenient to test individually though
- tests?
  - i guess i can write some to check the suggestions from hand coded.
  - will follow up with a pr to add better heuristics to hand-coded

Observations

- Can get 26ms CL time for cifar, fp32, wino
- Can get 75ms CL time for resnet train, bs=32, fp32, wino
  - ~29ms for eval, iirc
- Feasible to do budget=2000 runs for cifar/resnet in a few hours

Example:

```
(venv) chaos@tiny3:~/tinygrad$ git log | head -n1 && STEPS=100 DEBUG=0 KOPT=2 KOPT_THRESH=500 BUDGET=100 KOPT_WORKERS=32 KOPT_CACHE=./kopt_cache_tmp HIP_VISIBLE_DEVICES=1 TC=0 OPT=2 WINO=1 HIP=1 JIT=1 python examples/hlb_cifar10.py
fatal: bad object HEAD
  1.123/  0.965 ( 116%) @   22/  47 - E_6250_192_4_16_2n1                  : 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [00:01<00:00, 80.45it/s]
[...]

PR #2029, opened 2023-10-10
KOPT: Output KOPT config from hand_coded

chaosagent:
split from #1958 

expresses most hand_coded operations in terms of auto opts, except top=True and TC. seems like @geohot is doing something similar in [better_opt](https://github.com/tinygrad/tinygrad/compare/better_opt)


geohot:
Did you test to make sure it matches 100%?

chaosagent:
Trying to run the test suite rn...

The hand_coded should behave exactly as before (by inspection). The hand_coded doesn't match KOPT, I need to give it some thought. The main two things are top=True and order of operations (NG has fixed order).

With the parallel KOPT change, search is quite fast, so maybe it is OK to have top=True and such in the NG search space, but need to be careful in the design. NG search space is currently subset of overall search space, since operation order is fixed.

chaosagent:
superseded by #2053

PR #2032, opened 2023-10-10
Fix stable diffusion output error on WebGPU

wpmed92:
When running `examples/stable_diffusion.py` on `WEBGPU` an error is thrown because the output cannot be casted to `uint8`, since it is not supported on `WEBGPU`.

Screenshot 2023-10-10 at 9 12 11

To solve this, I `numpy` cast if on webgpu, `Tensor` cast otherwise.

### DIFF
--- a/examples/stable_diffusion.py
+++ b/examples/stable_diffusion.py
@@ -652,17 +652,13 @@ if __name__ == "__main__":
   # make image correct size and scale
   x = (x + 1.0) / 2.0
   x = (x.reshape(3,512,512).permute(1,2,0).clip(0,1)*255)
+  if Device.DEFAULT != "WEBGPU": x = x.cast(dtypes.uint8)
   print(x.shape)
 
   # save image
   from PIL import Image
   import numpy as np
-
-  # TODO: without this hack numpy array will be zero. Bug in wgpu_device.queue.read_buffer?
-  if Device.DEFAULT == "WEBGPU":
-    x.numpy()
-
-  im = Image.fromarray(x.numpy().astype(np.uint8))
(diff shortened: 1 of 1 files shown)


examples/stable_diffusion.py
```diff
@@ -651,12 +651,18 @@ def do_step(latent, timestep, index):
...
+  x = (x.reshape(3,512,512).permute(1,2,0).clip(0,1)*255)
   print(x.shape)
 
   # save image
   from PIL import Image
-  im = Image.fromarray(x.numpy())
+  import numpy as np
+
+  # TODO: without this hack numpy array will be zero. Bug in wgpu_device.queue.read_buffer?
+  if Device.DEFAULT == "WEBGPU":
```
geohot, line 662: I can't merge this, this hack makes no sense.

examples/stable_diffusion.py
```diff
@@ -651,12 +651,14 @@ def do_step(latent, timestep, index):
...
-  x = (x.reshape(3,512,512).permute(1,2,0).clip(0,1)*255).cast(dtypes.uint8)
+  x = (x.reshape(3,512,512).permute(1,2,0).clip(0,1)*255)
+  if Device.DEFAULT != "WEBGPU": x = x.cast(dtypes.uint8)
   print(x.shape)
 
   # save image
   from PIL import Image
-  im = Image.fromarray(x.numpy())
+  import numpy as np
+  im = Image.fromarray(x.numpy().astype(np.uint8) if Device.DEFAULT == "WEBGPU" else x.numpy())
```
geohot, line 661: `im = Image.fromarray(x.numpy().astype(np.uint8, copy=False))`

PR #2033, opened 2023-10-10
Refactor GPU and Metal langauges in their own separate renderers

Qazalin:
the immediate use-case is for rendering the bitcast op. OpenCL and Metal have different utilities for doing bitcast safely. The raw C implementation doesn't work on their compiler safety checks:

```
program_source:8:32: error: address of vector element requested
*(data0+3) = (*((device int*)&((val0).w)));
```
_(even though we know it's safe when we upcast)_

 Other C-based backends didn't have this issue.

### DIFF
--- a/tinygrad/runtime/ops_gpu.py
+++ b/tinygrad/runtime/ops_gpu.py
@@ -1,14 +1,13 @@
 from __future__ import annotations
-import pathlib, functools
+import pathlib
 import numpy as np
 import pyopencl as cl  # type: ignore
 from typing import Optional, List
 from tinygrad.helpers import DEBUG, getenv, prod, ImageDType, OSX, fromimport
 from tinygrad.ops import Compiled
-from tinygrad.renderer.opencl import OpenCLLanguage
+from tinygrad.renderer.opencl import OpenCLRenderer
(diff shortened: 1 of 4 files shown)


tinygrad/runtime/ops_gpu.py
```diff
@@ -103,9 +104,5 @@ def __call__(self, global_size, local_size, *bufs, wait=False) -> Optional[float
         return None
     return None
 
-renderer = functools.partial(uops_to_cstyle, CStyleLanguage(
-  kernel_prefix = "__kernel ", buffer_prefix = "__global ", smem_align = "__attribute__ ((aligned (16))) ", smem_prefix = "__local ", arg_int_prefix = "const int",
-  half_prekernel = "#pragma OPENCL EXTENSION cl_khr_fp16 : enable",
-  barrier = "barrier(CLK_LOCAL_MEM_FENCE);", float4 = "(float4)",
-  gid = [f'get_group_id({i})' for i in range(3)], lid = [f'get_local_id({i})' for i in range(3)], uses_vload=True))
+renderer = functools.partial(uops_to_cstyle, OpenCLLanguage())
```
geohot, line 107: I would move this line to the renderers too.

PR #2035, opened 2023-10-10
kernel cache

### LINKED ISSUE #2025: Add a generic compiled kernel cache
Stalled attempt here: #1580 

Implement for `ops_hip`.
Should replace the one in `ops_clang` as well.

nimlgen:
Kernel cache for hip/clang.

Compile time for `WINO=1 HIP=1 python3 examples/hlb_cifar10.py` is 76s -> 21s for reruns.

Tried to make as simple as possible without overcomplicated decorators (imho). But this also support toolchain versions which, so it will recompile kernels when new toolchains are available.

Closes #2025

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-import os, functools, platform, time, re, contextlib, operator
+import os, functools, platform, time, re, contextlib, operator, hashlib
 import numpy as np
 from typing import Dict, Tuple, Union, List, NamedTuple, Final, Iterator, ClassVar, Optional, Iterable, Any, TypeVar, TYPE_CHECKING
 if TYPE_CHECKING:  # TODO: remove this and import TypeGuard from typing once minimum python supported version is 3.10
@@ -148,3 +148,24 @@ class GlobalCounters:
   mem_cached: ClassVar[int] = 0 # NOTE: this is not reset
   @staticmethod
   def reset(): GlobalCounters.global_ops, GlobalCounters.global_mem, GlobalCounters.time_sum_s, GlobalCounters.kernel_count = 0,0,0.0,0
+
+# *** cache compiled files ***
+
+def cache_compiled(folder:str):
+  cache_dir = os.path.join(os.path.expanduser("~/.cache/tinygrad"), folder)
+  os.makedirs(cache_dir, exist_ok=True)
+
+  def decorator(func):
+    def wrapper(self, prg:str, binary:bool=False) -> str:
+      prg_hash = hashlib.sha256(prg.encode()).hexdigest()
+      cache_path = os.path.join(cache_dir, prg_hash)
+
+      if not os.path.exists(cache_path):
+        result = func(self, prg, binary=binary)
+        with open(f"{cache_path}.tmp.{os.getpid()}", "wb") as f:
+          f.write(result)
+        os.rename(f"{cache_path}.tmp.{os.getpid()}", cache_path)
+      return cache_path
+
+    return wrapper
+  return decorator
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -157,9 +157,10 @@ def cache_compiled(folder:str):
 
   def decorator(func):
     def wrapper(self, prg:str, **kwargs) -> str:
-      cache_path = cache_dir / hashlib.sha256(prg.encode()).hexdigest()
+      cache_path, tmp_path = cache_dir / hashlib.sha256(prg.encode()).hexdigest(), cache_dir / f"tmp.{os.getpid()}"
       if not cache_path.exists():
-          (cache_dir / f"{cache_path.name}.tmp.{os.getpid()}").write_bytes(func(self, prg, **kwargs)).rename(cache_path)
+          tmp_path.write_bytes(func(self, prg, **kwargs))
+          tmp_path.rename(cache_path)
       return str(cache_path)
     return wrapper
   return decorator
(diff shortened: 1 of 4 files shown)


geohot:
There should be more shared code than this. Can this be a negative line diff?

nimlgen:
With a decorator +4 lines in total tho. I think other line-squashes might hurt readability.

tinygrad/runtime/ops_clang.py
```diff
@@ -29,32 +30,29 @@ def emulate_ext_calls(fn, uc, address, size, user_data):
...
-          with open(fn + '.bin', 'rb') as f:
-            self.prg = f.read()
-          return
-        subprocess.check_output(args=('as -o' + tmp).split(), input=prg.encode('utf-8'))
-        subprocess.check_output(args=('clang -lm -shared '+tmp+' -o'+fn).split())
-    self.lib = ctypes.CDLL(fn)
-    self.fxn = self.lib[name]
+    self.prg: Any = ctypes.CDLL(str(cached_file))[name] if not (CI and ARM64) else pathlib.Path(str(cached_file)).read_bytes()
+
+  @cache_compiled
```
geohot, line 41: Does this function still work if I remove the decorator?

nimlgen: done

tinygrad/runtime/ops_clang.py
```diff
@@ -29,32 +30,30 @@ def emulate_ext_calls(fn, uc, address, size, user_data):
...
+      prg = prg.split('\n') # type: ignore
+      self.varsize = align(int(prg[0].split(" ")[1]))
+      self.ext_calls = {(i*4+ADDRESS):ins.split(" ")[1:] for i, ins in enumerate(filter(lambda ins: ins[:4] != 'loop', prg[6:-3])) if ins[:2] == 'bl'}
+      prg = "\n".join(['nop' if ins[:2] == 'bl' else ins for ins in prg[6:-3]] + ['\n'])
+      subprocess.check_output(args=('aarch64-linux-gnu-as -o '+str(temp_file)).split(), input=prg.encode('utf-8'))
+      subprocess.check_output(args=('aarch64-linux-gnu-objcopy -O binary --only-section=.text '+str(temp_file)+' '+str(output_file)).split())
+    else:
+      subprocess.check_output(args=('as -o' + str(temp_file)).split(), input=prg.encode('utf-8'))
+      subprocess.check_output(args=('clang -lm -shared '+str(temp_file)+' -o'+str(output_file)).split())
+    return None, pathlib.Path(output_file)
```
geohot, line 55: Err, I don't care if it's slower, this return type is awful and I don't understand it.

Readability and simplicity > everything.

nimlgen: Shouldn't have done that. Made it back to one return value.

tinygrad/helpers.py
```diff
@@ -148,3 +148,18 @@ class GlobalCounters:
...
+
+# *** compiled cache decorator ***
+
+def cache_compiled(func):
+  def wrapper(self, prg:str, **kwargs):
+    # cache_path is where the cache is stored, output_file is save to write.
+    cache_path, output_file = pathlib.Path(f"{tempfile.gettempdir()}/tinygrad_cc_{hashlib.sha256(prg.encode()).hexdigest()}"), pathlib.Path(kwargs.get('output_file', tempfile.mktemp()))
+    if not cache_path.exists():
+      # If we compile into a file (has output_file in args), just rename it into the cache. Otherwise, write a temp file and rename it (for atomicity).
+      if (res := func(self, prg, **kwargs)) and 'output_file' not in kwargs: output_file.write_bytes(res)
```
geohot, line 160: AHHHH how can I reason about this? If a kwarg is named something special!!!!

tinygrad/runtime/ops_hip.py
```diff
@@ -106,6 +99,18 @@ def __init__(self, name:str, prg:str, binary=False):
       hip.hipSetDevice(i)
       self.prgs.append(hip.hipModuleGetFunction(hip.hipModuleLoadData(prg), name))
 
+  @cache_compiled
+  def compile(self, prg, name):
+    output_file = pathlib.Path(tempfile.mktemp())
+    try:
+      prog = hip.hiprtcCreateProgram(prg, name, [], [])
+      hip.hiprtcCompileProgram(prog, [f'--offload-arch={hip.hipGetDeviceProperties(HIP.default_device).gcnArchName}'])
+      output_file.write_bytes(hip.hiprtcGetCode(prog))
```
geohot, line 108: Hmm, now this writes a file if I don't cache.

Like I'm not being pedantic here, this introduces complexity that doesn't make sense. While it's readable, why?

geohot:
I want you to think about isolating things to minimum blocks.

If you have a compile function, what type signature should it have? To me it's obvious:

`def compile(pr:str) -> bytes:`

Why does it make sense for it to return a file?

nimlgen:
Yeah, thanks, you're absolutely right. I need to prioritize simplicity and design over reducing line count. While I don't do that intentionally, this is the first project where I've aimed for a specific line count. So sometimes, I try to minimize the lines of code, especially for minor features like that one.

I apologize for wasting your time with numerous reviews. Should adjust my approach to delivering code for this project.

geohot:
Line count is secondary, the primary is always readability. This one looks good!

tinygrad/runtime/ops_clang.py
```diff
@@ -29,32 +30,34 @@ def emulate_ext_calls(fn, uc, address, size, user_data):
...
-          subprocess.check_output(args=('aarch64-linux-gnu-objcopy -O binary --only-section=.text '+tmp+' '+fn+'.bin').split())
-          with open(fn + '.bin', 'rb') as f:
-            self.prg = f.read()
-          return
-        subprocess.check_output(args=('as -o' + tmp).split(), input=prg.encode('utf-8'))
-        subprocess.check_output(args=('clang -lm -shared '+tmp+' -o'+fn).split())
-    self.lib = ctypes.CDLL(fn)
-    self.fxn = self.lib[name]
+    if not (CI and ARM64):
+      cached_file_path = pathlib.Path(tempfile.mktemp())
```
geohot, line 40: __del__ should probably clean up these files

geohot: Actually even better, I think you can just delete it after you load the library.

PR #2039, opened 2023-10-11
ModNode __mod__ rule

ZenginU:
I realized we have not covered this rule, also this removed some of the conditional loads in IMAGE=2 tests but did not effect openpilot. 

This rule is mostly triggered in IMAGE

### DIFF
--- a/test/unit/test_symbolic.py
+++ b/test/unit/test_symbolic.py
@@ -143,6 +143,10 @@ class TestSymbolic(unittest.TestCase):
   def test_mod_mul(self):
     self.helper_test_variable((Variable("a", 0, 5)*10)%9, 0, 5, "a")
 
+  def test_mod_mod(self):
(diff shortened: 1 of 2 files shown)


test/unit/test_symbolic.py
```diff
@@ -143,6 +143,10 @@ def test_sum_div_const_big(self):
   def test_mod_mul(self):
     self.helper_test_variable((Variable("a", 0, 5)*10)%9, 0, 5, "a")
 
+  def test_mod_mod(self):
+    self.helper_test_variable((Variable("a", 0, 31)%12)%4, 0, 3, "(a%4)")
+    self.helper_test_variable(((4*Variable("a", 0, 31)) % 12) % 4, 0, 0, "0")
```
geohot, line 148: Should support (a%4)%12 removal also

Issue #1700, opened 2023-08-28
Bounty: Get HIP tests in CI


dc-dc-dc:
Utilize [HIP-CPU](https://github.com/ROCm-Developer-Tools/HIP-CPU) to get test suite checking the hip runtime in CI

---
Current Bounty: $200
To lock the bounty submit a draft PR with a decent amount of progress made
Make sure to reference this issue in the PR for future tracking

_Notice: If the PR goes stale the bounty will be unlocked_

PR #2040, opened 2023-10-11
HIP CPU test_ops CI

tralamazza:
HIP CPU test_ops CI based on @dc-dc-dc #1419.

`HIP=1 HIP_CPU=1 python -m pytest -v test/test_ops.p` takes ~11 min on a thinkpad T480s, w/ an average build time ~200ms per kernel.

missing:
- [x] fix clang racing (probably needs a file lock or file per pid)
- [x] removing one trampoline (just need to pass another argument)
- [x] make the linter happy?
- [x] check w/ flamegraph the average kernel build time

bounty ref: #1700 

shout-out to [corecode](https://github.com/corecode/)

### DIFF
--- a/extra/hip_wrapper.py
+++ b/extra/hip_wrapper.py
@@ -1,9 +1,8 @@
 import ctypes
-from tinygrad.helpers import DEBUG
-import sys
-import numpy as np
-from typing import Any, Dict, List, Tuple
 from dataclasses import dataclass
+from tinygrad.helpers import DEBUG
+from tinygrad.renderer.cstyle import CStyleLanguage
+from typing import Any, Dict, List, Tuple
 
 try:
   _libhip = ctypes.cdll.LoadLibrary("libamdhip64.so")
@@ -681,5 +680,6 @@ try:
     status = _libhiprtc.hiprtcGetCode(prog, e_code)
     hipCheckStatus(status)
     return e_code
+
 except:
   if DEBUG >= 1: print("WARNING: libamdhip64.so or libhiprtc.so not found. HIP support will not work.")
--- a/tinygrad/runtime/ops_hip.py
+++ b/tinygrad/runtime/ops_hip.py
@@ -1,12 +1,16 @@
 import numpy as np
 import ctypes, functools, math, collections
-import extra.hip_wrapper as hip
 from typing import Tuple, Any, List
 from tinygrad.helpers import DEBUG, getenv, cache_compiled
 from tinygrad.ops import Compiled, ASTRunner, BasicBatchExecutor
 from tinygrad.runtime.lib import RawBufferCopyInOut, LRUAllocator, RawBufferTransfer
 from tinygrad.codegen.kernel import LinearizerOptions
-from tinygrad.renderer.cstyle import uops_to_cstyle, CStyleLanguage
+from tinygrad.renderer.cstyle import CStyleLanguage, uops_to_cstyle
+
+if getenv("HIP_CPU"):
+  import extra.hip_cpu_wrapper as hip
+else:
+  import extra.hip_wrapper as hip # type: ignore[no-redef]
 
 # TODO: if you fork and exit the child process after creating anything with cl on AMD, it hangs on e.wait()
 if DEBUG >= 6:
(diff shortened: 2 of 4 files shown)


extra/hip_wrapper.py
```diff
@@ -681,5 +680,9 @@ def hiprtcGetCode(prog):
     status = _libhiprtc.hiprtcGetCode(prog, e_code)
     hipCheckStatus(status)
     return e_code
+
+  def language(**kwargs):
```
geohot, line 684: This shouldn't be in here, it's just hiding complexity.

tralamazza: I tried to minimize the impact on ops_hips.py, do you have any suggestions?

geohot: Just put it in ops_hip. If it's required complexity that's okay, but it shouldn't hide

tralamazza: but then ops_hip will need another "if HIP_CPU"

tralamazza: unfortunately HIP-CPU is a bit of a mess and lacks a few APIs (e.g. capture), maybe HIP_CPU should become a Device? so tests can filtered it out.

geohot: If it's a different Device there's no point in doing this, what's being tested?

geohot: "but then ops_hip will need another "if HIP_CPU""

and...? if you hide complexity, you still have the complexity, it's just elsewhere. surfaced complexity > hidden complexity.

tralamazza: > "but then ops_hip will need another "if HIP_CPU""
> 
> and...? if you hide complexity, you still have the complexity, it's just elsewhere. surfaced complexity > hidden complexity.

It's an interface provided by a hip cpu wrapper, like all those hip_ functions, implemented by both wrappers. Should I move all implementations to ops_hip?

tralamazza: > If it's a different Device there's no point in doing this, what's being tested?

test_ops is being tested

tinygrad/runtime/ops_hip.py
```diff
@@ -100,6 +104,7 @@ def __init__(self, name:str, prg:str, binary=False):
       self.modules.append(hip.hipModuleLoadData(prg))
       self.prgs.append(hip.hipModuleGetFunction(self.modules[-1], name))
 
+  cache_compiled = lambda _: _ if getenv("HIP_CPU") else cache_compiled # noqa: E731
```
geohot, line 107: Why does the cache break with CPU?

tinygrad/runtime/ops_hip.py
```diff
@@ -127,7 +132,8 @@ class PackageStruct(ctypes.Structure):
   def __del__(self):
     for module in self.modules: hip.hipModuleUnload(module)
 
-renderer = functools.partial(uops_to_cstyle, CStyleLanguage(
+language = hip.language if getenv("HIP_CPU") else CStyleLanguage
```
geohot, line 135: I'm fine with this, at least the normal HIP is here

geohot:
Looking good code wise, now we just have to do something about the speed. How is this slower than GPUocelot?!

tinygrad/runtime/ops_hip.py
tralamazza: because hiprtcCreateProgram returns the CDLL (and write_bytes expects byte-like objects)

tralamazza:
Locally `HIP=1 HIP_CPU=1 python -m pytest -v -n=auto test/test_ops.py` takes:

- 1st pass (full build and run)
`199 passed, 8 skipped in 224.93s (0:03:44)`

- 2nd pass (no build needed since /tmp contains all binaries)
`199 passed, 8 skipped in 62.75s (0:01:02)`

tralamazza:
> Looking good code wise, now we just have to do something about the speed. How is this slower than GPUocelot?!

I haven't profiled so I don't know yet.

Executing locally with all kernels precompiled still takes ~1min, pretty slow for just loading a shared lib and executing a function.

geohot:
Closing, CI is too slow to merge. You are welcome to reopen when it's ~5 minutes!

PR #2042, opened 2023-10-11
whisper: make file transcription work, add basic CI test

mmmkkaaayy:
* Fixed a couple bugs in the offline mode of whisper.py that were preventing it from running at all.
* Added padding to the log_spec, otherwise the model was only predicting the first token
* Added `<|notimestamp|>` to live transcription mode - the model was actually predicting this token after a couple iterations. (However, this seems to have introduced `[BLANK_AUDIO]` tokens in the live transcription when there is no speech)
* Extracted core functionality into helper functions that can be called from unit test, added a simple unit test


geohot:
Can we add tests for whisper? Possible to test the tiny one in CI?

mmmkkaaayy:
I can give it a shot. We're ok with downloading 72MB of weights on every CI run?

geohot:
Yea, I think 72 is on par with things we are already downloading

mmmkkaaayy:
I might be a bit stuck here... For some reason the torch/cpu/llvm backends are failing due to large (20GB+) memory allocations or timing out. I've attached some debug logs where I reproduced the issue - can anyone tell why the torch/cpu ops are so much slower than metal?

https://gist.github.com/mmmkkaaayy/59966bad909defaf7e17affb65a0a004

geohot:
Oh you really only have to test it on one backend

mmmkkaaayy:
I've added it to the Metal test set since this is only proven to work there for now.

I'm disappointed this doesn't work well on torch/other backends... my best guess so far is that `ConstBuffer`s behave differently on those backends, requiring a gigantic array allocation, that Metal somehow bypasses? Will try to look further into it when I have time.

geohot:
That's a big wav file...can you make it mp3 or something?

test/models/test_whisper.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+import pathlib
+from tinygrad.ops import Device
+
+from examples.whisper import init_whisper, transcribe_file
+
+@unittest.skipUnless(Device.DEFAULT == "METAL", "Some non-metal backends spend too long trying to allocate a 20GB array")
+class TestWhisper(unittest.TestCase):
+    @classmethod
```
geohot, line 9: 2 spaces

mmmkkaaayy:
> That's a big wav file...can you make it mp3 or something?

Reduced from 251KB to 35KB by generating a lower quality wav - I think it's better to keep it in wav for when we test on non-macos platforms that don't have many audio codecs.

geohot:
low quality wave works

mmmkkaaayy:
Cool, I won't commit to that bounty just yet, since it might require me to reimplement most/all of the reference whisper repo (there's a lot of extra junk on top of the core model!).

PR #2046, opened 2023-10-12
Internal casting support

Qazalin:
Non-fp32 math is gonna be another PR with mostly unrelated changes with this one, focused this on rendering the CAST explicitly:

```diff
#include <metal_stdlib>
using namespace metal;
kernel void E_5(device half* data0, const device float* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 5 */
  float val0 = *(data1+gidx0);
+ *(data0+gidx0) = ((half)(val0)+4.0f);
- *(data0+gidx0) = (val0+4.0f);
}
```

this is enough casting infra to make non-fp23 alu and bitcast work in later PRs

On its own, it helps out with accuracy in things like rounding precision and re-enables dtypes tests on triton.


extra/onnx.py
```diff
@@ -159,7 +159,7 @@ def fetch_tensor(x: str):
         ret = inp[0].reshape([s for i,s in enumerate(inp[0].shape) if i not in axes])
       elif n.op_type == "Div":
         # in openpilot, due to SHUFFLE_PAD_OPS issues, we are spending an extra kernel
-        ret = inp[0].div(inp[1]) if inp[0].dtype == dtypes.float else inp[0].div(inp[1]).floor()
+        ret = inp[0].div(inp[1])
```
Qazalin, line 162: The kernel for .floor in master is wrong since the values don't truncate to int32:
```
__kernel void E_4_32_4(__global float* data0, const __global float* data1, const __global float* data2, const __global float* data3) {
  int gidx0 = get_group_id(0); /* 4 */
  int lidx1 = get_local_id(0); /* 32 */
  float4 val0 = (float4)(*((__global float4*)(data1+(gidx0*128)+(lidx1*4))));
  float4 val1 = (float4)(*((__global float4*)(data2+(gidx0*128)+(lidx1*4))));
  float val2 = data3[0];
  float alu0 = (((val0).x+(val1).x)/val2);
  float alu1 = (((val0).y+(val1).y)/val2);
  float alu2 = (((val0).z+(val1).z)/val2);
  float alu3 = (((val0).w+(val1).w)/val2);
  *((__global float4*)(data0+(gidx0*128)+(lidx1*4))) = (float4)(float4)(((alu0<alu0)!=0?(alu0-1.0f):alu0),((alu1<alu1)!=0?(alu1-1.0f):alu1),((alu2<alu2)!=0?(alu2-1.0f):alu2),((alu3<alu3)!=0?(alu3-1.0f):alu3));
}
```

the code in this branch would be:
```
__kernel void E_4_32_4(__global float* data0, const __global float* data1, const __global float* data2, const __global float* data3) {
  int gidx0 = get_group_id(0); /* 4 */
  int lidx1 = get_local_id(0); /* 32 */
  float4 val0 = (float4)(*((__global float4*)(data1+(gidx0*128)+(lidx1*4))));
  float4 val1 = (float4)(*((__global float4*)(data2+(gidx0*128)+(lidx1*4))));
  float val2 = data3[0];
  float alu0 = (((val0).x+(val1).x)/val2);
  float alu1 = (((val0).y+(val1).y)/val2);
  float alu2 = (((val0).z+(val1).z)/val2);
  float alu3 = (((val0).w+(val1).w)/val2);
  int cast0 = (int)(alu0);
  int cast1 = (int)(alu1);
  int cast2 = (int)(alu2);
  int cast3 = (int)(alu3);
  *((__global float4*)(data0+(gidx0*128)+(lidx1*4))) = (float4)(float4)((float)(((alu0<cast0)!=0?(cast0-1.0f):cast0)),(float)(((alu1<cast1)!=0?(cast1-1.0f):cast1)),(float)(((alu2<cast2)!=0?(cast2-1.0f):cast2)),(float)(((alu3<cast3)!=0?(cast3-1.0f):cast3)));
}
```

applying floor with its truncate part will give the wrong result:
```
Not equal to tolerance rtol=0.01, atol=0.0001

Mismatched elements: 5269 / 6120 (86.1%)
Max absolute difference: 147.69875
Max relative difference: 105.37706
 x: array([[ 2.669612, -0.469016,  0.007674, ...,  0.      ,  0.      ,
         0.      ]], dtype=float32)
 y: array([[ 3.16809 , -0.461886,  0.010212, ...,  0.      ,  0.      ,
         0.      ]], dtype=float32)
```
that's why I'm removing it

tinygrad/runtime/ops_gpu.py
```diff
@@ -103,4 +103,4 @@ def __call__(self, global_size, local_size, *bufs, wait=False) -> Optional[float
         return None
     return None
 
-GPUBuffer = Compiled(CLBuffer, LinearizerOptions(), OpenCLRenderer, CLProgram, CL.synchronize)
+GPUBuffer = Compiled(CLBuffer, LinearizerOptions(device="GPU", supports_float16=False), OpenCLRenderer, CLProgram, CL.synchronize)
```
Qazalin, line 106: github's intel GPUs don't implement `half` in their drivers. The extension is automatically disabled.

https://github.com/tinygrad/tinygrad/actions/runs/6508596058/job/17678264259?pr=2046#step:13:117

right now this is a static flag. to support both kinds of GPUs _(even though no fp16 ones will kind of be broken for some functionality)_ this needs to be logic in the code. Either something we pre-compute on device initialization and assign to `supports_float16` in the tinygrad runtime or branching in the kernel itself

extra/onnx.py
Qazalin: If you remove the .floor() from what's already in master, the kernel counts will end up being 210 - This PR didn't impact the kernel counts directly.

tinygrad/runtime/ops_gpu.py
Qazalin: https://github.com/fangq/mcxcl/issues/24#issuecomment-511815534

extra/onnx_ops.py
```diff
@@ -16,7 +16,7 @@ def Add(input: Tensor, other: Tensor, broadcast=None): return input + other if i
 def Sub(input: Union[Tensor, Any], other: Tensor): return input - other # some test has input as int
 def Mul(input: Tensor, other: Tensor): return (input * other) if input.dtype == dtypes.float else (input * other).cast(input.dtype)
 # in openpilot, due to SHUFFLE_PAD_OPS issues, we are spending an extra kernel
-def Div(input: Tensor, other: Tensor): return input / other if input.dtype == dtypes.float else input.div(other).floor()
+def Div(input: Tensor, other: Tensor): return input / other if input.dtype == dtypes.float else input.div(other).float()
```
geohot, line 19: I don't understand this change

geohot:
`supports_float16` has to be done more generically.

tinygrad/renderer/llvmir.py
```diff
@@ -28,6 +28,7 @@
 
 def cast(bb, val, input_type, output_type):
```
geohot, line 29: Can you write tests for this function?

.github/workflows/test.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       - if: ${{ matrix.task == 'openpilot' }}
         name: Test openpilot model compile and size
         run: |
-          DEBUG=2 ALLOWED_KERNEL_COUNT=206 VALIDTEST=1 FLOAT16=1 DEBUGCL=1 GPU=1 IMAGE=2 python openpilot/compile.py
+          DEBUG=2 ALLOWED_KERNEL_COUNT=207 VALIDTEST=1 FLOAT16=1 DEBUGCL=1 GPU=1 IMAGE=2 python openpilot/compile.py
```
geohot, line 159: Why does this change?

Qazalin: this was the change I made for the model accuracy tests
https://github.com/tinygrad/tinygrad/pull/2046#discussion_r1359879357

Qazalin: was looking into how we do .floor() in that kernel and noticed there's no truncating before we do the comparison

geohot:
I locked this bounty for you also.

tinygrad/codegen/linearizer.py
```diff
@@ -398,10 +397,15 @@ def uop(self, uop:UOps, dtype:Optional[DType], vin:Tuple[UOp, ...], arg:Any=None
     if cachable: self.saved_exprs[key] = self.uops[-1]
     return self.uops[-1]
 
+  def parse_cast(self, uop: UOp, dtype:DType, bitcast:bool) -> UOp:
+    if isinstance(dtype, ImageDType): return uop # TODO: imagef and imageh casted to fp32 and fp16?
```
Qazalin, line 401: something like:
```py
  def parse_cast(self, uop: UOp, dtype:DType, bitcast:bool) -> UOp:
    if isinstance(dtype, ImageDType):
      if dtype.name == "imagef": dtype = dtypes.float32
      elif dtype.name == "imageh": dtype = dtypes.float16
    return self.uop(UOps.CAST, dtype, (uop,), (dtype,bitcast))
```
maybe

Qazalin:
alright CI is green, onto writing tests for this one

tinygrad/renderer/llvmir.py
Qazalin: tests done, do you mean removing bf16 support completely?

tinygrad/renderer/cstyle.py
```diff
@@ -191,7 +191,9 @@ def ssa(u, prefix="t"):
       elif len(vin) == 3:
         assert vin[0].dtype is not None and vin[2].dtype is not None
         kk(lang.render_store(r[vin[0]], vin[0].dtype, r[vin[2]], vin[2].dtype, strip_parens(r[vin[1]]), vin[0].uop == UOps.DEFINE_LOCAL))
-    elif uop == UOps.CAST and dtype is not None and dtype.sz > 1:
+    elif uop == UOps.CAST:
+      dtype = args[0] if args is not None else dtype
```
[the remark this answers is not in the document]
Qazalin: fixed, now done at the linearizer level https://github.com/tinygrad/tinygrad/pull/2046/files#diff-63d57daaca59e5495bbb108da1c13e782f4a408db8503e015e8094ac07c87ea6R403

geohot:
What's the status on this / why are tests failing?

This will allow f16 weights in WebGPU stablediffusion

extra/onnx_ops.py
```diff
@@ -17,7 +17,7 @@ def Add(input: Tensor, other: Tensor, broadcast=None): return input + other if i
 def Sub(input: Union[Tensor, Any], other: Tensor): return input - other # some test has input as int
 def Mul(input: Tensor, other: Tensor): return (input * other) if input.dtype == dtypes.float else (input * other).cast(input.dtype)
 # in openpilot, due to SHUFFLE_PAD_OPS issues, we are spending an extra kernel
-def Div(input: Tensor, other: Tensor): return input / other if input.dtype == dtypes.float else input.div(other).floor()
+def Div(input: Tensor, other: Tensor): return input / other if input.dtype == dtypes.float or (isinstance(input.dtype, ImageDType) and input.dtype.name == "imagef") else input.div(other).floor()
```
geohot, line 20: I don't understand why there's ever a floor here

Qazalin: right - the onnx spec sorta expects that for non-floats
https://github.com/onnx/onnx/blob/main/docs/Operators.md#examples-43
but I dont see why there'd be a need for a floor if the dtype is already an int. I think you also saw that the image stuff breaks because of this

Qazalin: same - The onnx spec sorta does `//` instead of `/` for a uint8 div https://github.com/onnx/onnx/blob/main/docs/Operators.md#examples-43 you also saw how it breaks the image dtypes

Qazalin:
@geohot there's an edge-case that fails on LLVM - if there's an AST with multiple kernels and one kernel outputs a non-fp32 value the next kernel doesn't get the correct input. It might be something with the store logic, since the identical uops render is cstyle and work.

i can do a follow-up on LLVM and for now just do this hack. https://github.com/tinygrad/tinygrad/pull/2046/commits/404d531c1dceb9cd4eb1fe8a3b361110b3c544a3#diff-eb3a3956449cb9727741ec803e42c03395a66d4d1efdf941ba4ed9ef3829d0b6R215-R216
The only implication of the hack is that bitcast wont work in LLVM.

tinygrad/renderer/llvmir.py
Qazalin: i removed bf16 - lmk if i need to add it back

tinygrad/renderer/llvmir.py
```diff
@@ -141,7 +142,8 @@ def uops_to_llvm_ir(function_name:str, uops:List[UOp]) -> Tuple[str, Dict]:
       element = cast(bb, lvars[vin[2]], vin[2].dtype, vin[0].dtype)
       bb[-1].store(element, bb[-1].gep(lvars[vin[0]], [lvars[vin[1]]], inbounds=True))
     if uop == UOps.ALU:
-      lvars[u] = code_for_op[args](bb[-1], *[lvars[x] for x in vin])
+      lvars[u] = cast(bb, code_for_op[args](bb[-1], *[cast(bb, lvars[x], x.dtype, dtypes.float) for x in vin]), dtypes.float, dtype)
```
Qazalin, line 145: LLVM cant do ALU ops when the vars have different dtypes so we need to explicitly cast as fp32

Qazalin:
status: LLVM is green, gonna focus on writing more tests

geohot:
How is this going?

Qazalin:
@geohot incrementally moving all of hardcoded fp32s to the buf dtype and iterating through tests.
I'm also looking to write an abstraction to handle vectorized dtypes at the linearizer and renderer level, since the vectorized dtypes are gonna go from just _float2 and _float4 to the other dtypes too

tinygrad/renderer/llvmir.py
Qazalin: this will be removed once "Non-fp32 math support" is done

tinygrad/codegen/linearizer.py
```diff
@@ -399,7 +399,6 @@ def calc_tc_idxs(local_size: int, aliases: List[List[int]]):
   def uop(self, uop:UOps, dtype:Optional[DType], vin:Tuple[UOp, ...], arg:Any=None, cachable=True) -> UOp:
     key = (uop, dtype, vin, arg)
     if uop == UOps.PHI and len(vin) == 2 and vin[0] == vin[1]: return vin[0]   # self phi is noop
-    if uop == UOps.CAST and all(x.uop == UOps.GEP for x in vin) and all_same([x.vin[0] for x in vin]) and all(x.arg == i for i,x in enumerate(vin)): return vin[0].vin[0]
```
Qazalin, line 402: I couldn't understand the original intention behind this - but it doesn't work when we render cast from the AST, eg. if we have 4 GEPs coming from a _float4 -> half cast, the cast and the dtype along with it get thrown away and vin[0].vin[0] ends up being the _float4's LOAD op, I removed this out of necessity to make a kernel like:
```
kernel void E_4(device half* data0, const device float* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  float4 val0 = (float4)(*((device float4*)(data1+0)));
  *((device half4*)(data0+0)) = (half4)float4((half)((val0).x),(half)((val0).y),(half)((val0).z),(half)((val0).w));
}
```
work

geohot: Hmm, yea this is sort of wrong. The intention was to force the GPU to broadcast I'm okay with removing.

geohot:
YOLO. There's definitely a lack of tests for this, but the current stuff was sort of nonsense. Merged, looking forward to the next one!

Qazalin:
alright thanks, will add more tests as I go along

PR #2047, opened 2023-10-12
train value net, improve API, add BCE

geohot:


tinygrad/codegen/search.py
```diff
@@ -85,7 +80,7 @@ def get_linearizer_actions(lin:Linearizer) -> Dict[int, Linearizer]:
         if c in {"magenta", "yellow"}: up *= s
         if c in {"cyan", "green", "white"}: lcl *= s
       if up > 256 or lcl > 256: continue
-      acted_lins[i] = lin2
+      acted_lins[i+1] = lin2
     except Exception:
       pass
```
gijskoning, line 85: I noticed that lin2.colors() can also throw an exception after some upcasts. Is that expected behaviour?
This assert fails: `assert len(colors) == self.shape_len, "colors size mismatch"`

PR #2049, opened 2023-10-12
enable assert optimization in reshape

roelofvandijk:
This is the nice way of removing the all_int calls for `-O`.


geohot:
This is trading readability for speed

PR #2050, opened 2023-10-12
get the list of all function calls leading to the current function fr…

J4e6eR:
get the list of all function calls before the current function from the stack.
Thereby making it more easier to understand the workflow of tinygrad, both forbeginners and even for the contributers.
image


J4e6eR:
To prevent any further issues, please let me know where I can add this function bcz according to me it's really helpful !! @geohot

PR #2053, opened 2023-10-12
KOPT: move to new apply_opt, ng suggest

chaosagent:
redo of #2029 

need to move the kopt stuff off the main optimization path

as always, can use more testing


geohot:
Hmm, this needs to be a red diff. Unless there's a reason not to, we have to commit and rip out the old stuff.

chaosagent:
Eh, i guess we just move it to extra until the current RL contest experiments conclude.

geohot:
I deleted `apply_auto_opt` this morning.

chaosagent:
nevergrad is kill

PR #2054, opened 2023-10-12
get the list of all function calls that leads to the current function from the stack

J4e6eR:
get the list of all function calls before the current function from the stack.
Thereby making it more easier to understand the workflow of tinygrad, both for beginners and even for the contributors.
This function is now added in at file location tinygrad/tinygrad/get_function_stack.py
If this location seems to be worrisome, please specify the file location which is most suitable.
image


geohot:
Not better. This is generic Python stuff that probably doesn't even belong in tinygrad

PR #2057, opened 2023-10-12
fix wino mask upcast calculation

chaosagent:
i wrote it wrong last time, lol


geohot:
Needs tests, particularly since it was already wrong once.

chaosagent:
Added a few tests that capture the intended behavior wrt wino.

geohot:
Tests failing

PR #2058, opened 2023-10-12
HIPProgram __del__

chaosagent:
if you run too many kernels, eventually it will start failing, probably because it runs out of space for more kernels.

move loading to __call__, so that we can compile in child process.


geohot:
The del is fine, but __init__ should load the module, that's the expectation

chaosagent:
done

Issue #2048, opened 2023-10-12
examples/whisper.py doesn't work


nextgens:
I have first tried:
```
PYTHONPATH=. CPU=1 DEBUG=1 python3 examples/whisper.py  ../T-22050.wav 
ram used:  0.08 GB, decoder.ln.bias                                   : 100%|██████████████████████████████████████████| 167/167 [00:00<00:00, 4705.07it/s]
loaded weights in 37.21 ms, 0.08 GB loaded at 2.03 GB/s
Traceback (most recent call last):
  File "examples/whisper.py", line 198, in <module>
    waveform, sample_rate = librosa.load(sys.argv[1], normalize=True)
TypeError: load() got an unexpected keyword argument 'normalize'
```

Then without normalize:
```
PYTHONPATH=. python3 examples/whisper.py  ../T-22050.wav 
ram used:  0.11 GB, decoder.ln.bias                                   : 100%|██████████████████████████████████████████| 167/167 [00:00<00:00, 1818.19it/s]
loaded weights in 93.72 ms, 0.11 GB loaded at 1.16 GB/s
/home/rocm-user/.local/lib/python3.8/site-packages/librosa/core/spectrum.py:362: ComplexWarning: Casting complex values to real discards the imaginary part
  stft_matrix[..., :off_start] = fft.rfft(fft_window * y_frames_pre, axis=-2)
/home/rocm-user/.local/lib/python3.8/site-packages/librosa/core/spectrum.py:366: ComplexWarning: Casting complex values to real discards the imaginary part
  stft_matrix[..., -off_end:] = fft.rfft(fft_window * y_frames_post, axis=-2)
/home/rocm-user/.local/lib/python3.8/site-packages/librosa/core/spectrum.py:378: ComplexWarning: Casting complex values to real discards the imaginary part
  stft_matrix[..., bl_s + off_start : bl_t + off_start] = fft.rfft(
Traceback (most recent call last):
  File "examples/whisper.py", line 201, in <module>
    dat = model.encoder(Tensor(log_spec)).realize()
  File "examples/whisper.py", line 71, in __call__
    x = self.conv1(x).gelu()
  File "/app/tinygrad/tinygrad/nn/__init__.py", line 52, in __call__
    return x.conv2d(self.weight, self.bias, padding=self.padding, stride=self.stride, dilation=self.dilation, groups=self.groups)
  File "/app/tinygrad/tinygrad/tensor.py", line 517, in conv2d
[...]

PR #2061, opened 2023-10-13
working on issue #2048

lusk4r:
maybe the normalize parameter comes from a previous commit (using torchsound.load)

-  torchsound normalize substitited with librosa dtype=np.float32
-  output from librosa.load reshaped in order to fit prep_audio expected input shape 
- dtype of the model output and final print fixed (same as online version)


lusk4r:
fixed by #2042

lusk4r:
fixed by #2042

PR #2064, opened 2023-10-13
cleanup tensor cores, expose exclude local upcast

geohot:


tinygrad/codegen/kernel.py
```diff
@@ -144,8 +144,10 @@ def global_dims(self) -> int: return self.first_reduce-self.local_dims
   def colors(self) -> List[str]:
     # up to first_reduce, they are all global (blue)
     colors = ["blue"] * self.global_dims
+    # some special local_dims are excluded from the local upcast
+    colors += ["CYAN"] * self.exclude_local_upcast
     # except the local_dims, these are non-reduce locals (cyan)
-    colors += ["cyan"] * self.local_dims
+    colors += ["cyan"] * (self.local_dims - self.exclude_local_upcast)
     # between first_reduce and first_reduce + group_for_reduce, they are either local (cyan), or late upcasted (green)
```
flammit, line 151: If you're in here already, it might be worth removing the comment about locals here since they're before first_reduce now.

PR #2068, opened 2023-10-14
splitting waveform in whisper offline

lusk4r:
- new pad adaptation for n_frames > 3000 in prep_audio
- splitting strategy of the waveform (simplified version of the strategy proposed [here](https://github.com/openai/whisper/blob/b38a1f20f4b23f3f3099af2c3e0ca95627276ddf/whisper/transcribe.py#L226))

### DIFF
--- a/examples/whisper.py
+++ b/examples/whisper.py
@@ -212,8 +212,7 @@ def init_whisper(model_name="tiny.en"):
   enc = get_encoding(state['dims']['n_vocab'])
   return model, enc
 
-def transcribe_file(model, enc, filename):
-  waveform, _ = librosa.load(filename, sr=RATE)
+def transcribe_waveform(waveform, model, enc):
   n_frames = 3000
   log_spec = prep_audio(waveform)
 
@@ -245,7 +244,8 @@ if __name__ == "__main__":
   model, enc = init_whisper("small.en" if getenv("SMALL") else "tiny.en")
 
   if len(sys.argv) > 1:
-    transcribe_file(model, enc, sys.argv[1])
+    waveform, _ = librosa.load(sys.argv[1], sr=RATE)
+    transcribe_waveform(waveform, model, enc)
   else:
     # online
     q = multiprocessing.Queue()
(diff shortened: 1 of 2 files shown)


geohot:
I don't understand what this does. Is it tested?

lusk4r:
It is an extension of what was done in #2042 (for the offline version of whisper), solving issue #2048 caused by audio files longer than 30 seconds. 

I tried to implement a similar version to what was done in the [transcribe method of whisper](https://github.com/openai/whisper/blob/b38a1f20f4b23f3f3099af2c3e0ca95627276ddf/whisper/transcribe.py#L226) by managing the pad in case the number of frames of the log_spec in method prep_audio was > 3000 (due to recordings > 30s) and passing the chunks obtained from the starting log_spec to the encoder. 

I used the test written in #2042 for audio < 30s (it works correctly), added tests to verify the pad, and a test for audio > 30s (in which I enlarged the test.wav file).

examples/whisper.py
```diff
@@ -214,30 +212,42 @@ def init_whisper(model_name="tiny.en"):
...
+      out = model.decoder(Tensor([lst]), dat).realize()
+      idx = int(out[0,-1].argmax().numpy().item())
+      lst.append(idx)
+      transcription = enc.decode(lst)
+      print(transcription)
+      if lst[-1] == enc._special_tokens["<|endoftext|>"]:
+        final_transcription += transcription
+        if seek == log_spec.shape[-1]:
+          return  final_transcription
+        lst = [enc._special_tokens["<|startoftranscript|>"], enc._special_tokens["<|notimestamps|>"]]
```
mmmkkaaayy, line 239: Is this done in the reference impl?

lusk4r: good point! for semplicity in the previous implementation I followed [this scenario](https://github.com/openai/whisper/blob/b38a1f20f4b23f3f3099af2c3e0ca95627276ddf/whisper/transcribe.py#L366).

In the current version, I have instead implemented the token feed

geohot:
Is this covered by #2074?

mmmkkaaayy:
#2074 doesn't contain this functionality (the OpenAI evaluations truncated all samples to 30s). But it might be better to wait until we've gotten an optimized/JITted impl before trying to include this.

lusk4r:
> #2074 doesn't contain this functionality (the OpenAI evaluations truncated all samples to 30s). But it might be better to wait until we've gotten an optimized/JITted impl before trying to include this.

agree with this. Provisionally, it might make sense to check the size of the input audio and cut it if greater than 30s

PR #2069, opened 2023-10-14
Add ONNX to CI for other backends

geohotstan:
- Added ONNX CI for CPU, TORCH, METAL, LLVM, CLANG, GPU, CUDA
- Moved most ops from onnx.py to onnx_ops.py
- cleaned up onnx_ops.py a little to make it have that tinygrad look with the # ******* something *******

---------------------

Still not entirely sure if there isn't any egregious hacks in here that might cause CI to fail for reasons related to hacks. 
Will look it over a few more times before ready for review.


geohotstan:
Do we want MODELTESTS too? 😎

geohotstan:
Do we want MODELTESTS too? 😎

.github/workflows/test.yml
```diff
@@ -67,6 +67,8 @@ jobs:
       run: awk '/```python/{flag=1;next}/```/{flag=0}flag' docs/quickstart.md > quickstart.py &&  PYTHONPATH=. python quickstart.py
     - name: Run Pytest
       run: python -m pytest -n=auto test/ -k "not (test_efficientnet and models/test_train.py)"
+    - name: Run ONNX
+      run: CPU=1 python -m pytest -n=auto test/external/external_test_onnx_backend.py --tb=no --disable-warnings
```
geohot, line 71: Why are these flags needed?

geohot:
Looks like it slows CUDA down a lot, I'm fine with skipping there

.github/workflows/test.yml
geohotstan: I added CPU=1 here just to trigger test exclusion in `external_test_onnx_backend.py`
for example:
```
# Don't know how to treat special TensorProto like TensorProto.FLOAT8E4M3FN
if getenv("CPU") or getenv("TORCH"):
  backend_test.exclude('test_dequantizelinear_axis_cpu')
  backend_test.exclude('test_dequantizelinear_cpu')
```
Not sure if this is the best way to do this :/

I deleted --disable-warnings
--tb=no is needed since the outputted info on failed tests is unnecessarily verbose

geohotstan:
I reverted the Shape op change in onnx_ops.py in #2085
tests expects int64 :((((

geohot:
@geohotstan hehe I'm glad this caught the Shape issue :)

PR #2072, opened 2023-10-14
O(1) VALIDHACKS

ZenginU:
half disaster but low key grasped the logic

Should I seperate the symbolic part to seperate PR ?


geohot:
Looking good! Do you need any more tests?

ZenginU:
To be honest I do not know how to test this without fuzzer like logic, also very last part is based solely on trial and error which I am not proud of. 
There are couple of assumption also

geohot:
Hmm, why did tests change?

ZenginU:
> Hmm, why did tests change? Were they wrong to begin with?

They were not wrong, I made the SumNode use implicit division. 

For example,
a*4 < 16 -> a < 4.
Is now implicit.

geohot:
Can you add more tests? What does `a*4 < 15` do?

ZenginU:
> Can you add more tests? What does `a*4 < 15` do?

Okay, I will add more tests. I can put that symbolic in to another PR if you want. It will not break this PR.

Also, a*4 < 15 stays same as their gcd is 1.

geohot:
Eh, it's simple enough to read.

Isn't `a*4 < 15` the same as `a < 4`?

ZenginU:
> Eh, it's simple enough to read.
> 
> 
> 
> Isn't `a*4 < 15` the same as `a < 4`?

Yes it is true but I have not changed the logic in Sumnode.__lt__ . If you want I can work on that as well. This symbolic change is more like using existing implementation.

geohot:
It's fine for now, let's just add more tests and I think it's good

ZenginU:
> It's fine for now, let's just add more tests and I think it's good

Added test for SumNode folding but any ideas how to test the to_image_idx ? I believe we need fuzzer like approach.

geohot:
Yea, fuzzing would be great. Going to merge this for now, I don't fully understand it but I understand the symbolic change

PR #2073, opened 2023-10-14
allow to specify model and audio file for whisper

ubihot:
Adds cli arguments to choose the model and audio file. Multilingual doesn't seem to work though, it'll need a bit of more work.


examples/whisper.py
```diff
@@ -229,12 +222,28 @@ def transcribe_file(model, enc, filename):
...
+  enc = get_encoding('gpt2' if model_name.endswith(".en") else 'multilingual', state['dims']['n_vocab'])
+  return model, enc
 
 if __name__ == "__main__":
-  model, enc = init_whisper("small.en" if getenv("SMALL") else "tiny.en")
+  parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+  parser.add_argument("--audio", default=None, type=str, help="audio file to transcribe")
+  parser.add_argument("--model", default="tiny.en", choices=MODELS.keys(), help="name of the Whisper model to use")
+
+  args = parser.parse_args().__dict__
```
geohot, line 239: Why not just `args.model` and `args.audio`?

ubihot: good point, changed it now

geohot: I still don't know why you need `__dict__`

ubihot: we don't need `__dict__` indeed, I had copied it from an example...removed it now

geohot:
Closing for #2074

PR #2079, opened 2023-10-16
I have added a new loss file, which currently only contains a simple …

hi-sushanta:
Hello, my name is Chi. I noticed that your framework does not currently have a loss file containing multiple loss functions, so I created one. I have so far added only a single loss function, MSE loss.

And also see what looks like my loss function:
Screenshot from 2023-10-16 07-03-50

### DIFF
--- a/tinygrad/nn/loss.py
+++ b/tinygrad/nn/loss.py
@@ -1,0 +1,18 @@
+import tinygrad as tg
+from tinygrad.tensor import Tensor
+from typing import List
+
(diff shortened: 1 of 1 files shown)


geohot:
Too many lines

hi-sushanta:
what are you trying to say about?

wozeparrot:
this def doesn't belong in a separate file, look in `tensor.py` where the other loss functions currently are.

hi-sushanta:
you understand wrong becuse I try to create this file exactly similar to optim.py file.

PR #2080, opened 2023-10-16
Remove offset from sops shapetrackers

SzymonOzog:
The newly created buffers don't have offset so using them results in invalid access when running any of the optimization scripts on CUDA


geohot:
Hmm, this isn't an issue with load_worlds, it's an issue with the buffer creation.

PR #2082, opened 2023-10-16
examples/waifu2x: Now uses safetensors

20kdc:
This removes the "KINNE" model format in favour of safetensors. KINNE was a simple format I threw together to get stuff running in an era where the examples all had their own save/load functions using stuff like numpy. I felt the format was in line with the ideals of simplicity that tinygrad had at the time, it fit nicely with the `get_parameters` workflow, and safetensors I don't believe was a thing in tinygrad at that time, nor a thing I knew of.
By comparison KINNE was easy to generate from a terrible "patch" to NCNN which dumped every tensor it loaded, which was how I got things running originally.

In the modern era such a thing could easily have dumped safetensors files instead (maybe still doing the one-per-file thing), or an intermediate Python script could've converted the results to safetensors. Safetensors seems to be the standard within tinygrad, and as a format it's fine, though the Tinygrad API can be a bit odd (Python classes must follow what `get_state_dict` expects).

This patch also renames the weight/bias fields in order to better match another project (<https://github.com/FHPythonUtils/Waifu2x>) that seems to have a Waifu2x VGG7 network in a modern format (specifically in NPZ format with similar enough weight order, however this may have some difference between tinygrad's `conv0.W` notation and its `conv0/W` notation, and the X/Y may be transposed - I'm not sure).

The ultimate hope here is to try to generally improve the ease of compatibility and model porting between the tinygrad waifu2x vgg7 implementation and other implementations. Will someone ever do that? No idea, but I'd hope it's a good ideal.


geohot:
Is this tested in CI?

20kdc:
Uh... no? I'm also pretty sure the example wasn't previously. Should I add a test, and if so, is there a particular standard for adding a test for executing a model with pretrained weights?

geohot:
I'm tempted to just delete this example. What drew you to it?

20kdc:
A responsibility to clean up old code that I originally wrote.

geohot:
Oh you are the author. Meh, I'll just merge

PR #2083, opened 2023-10-16
I added new loss.py file.

hi-sushanta:
At that time, this file had two loss functions: MSE and MAE. However, it is also capable of storing different types of loss functions in a single file, similar to the PyTorch type framework.

Now, see what this file looks like:

Screenshot from 2023-10-16 20-13-00

### DIFF
--- a/tinygrad/nn/loss.py
+++ b/tinygrad/nn/loss.py
@@ -1,4 +1,3 @@
-import tinygrad as tg
 from tinygrad.tensor import Tensor
 from typing import List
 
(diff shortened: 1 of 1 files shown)


g1y5x3:
These loss are simple enough that I think you can just include in any of the scripts that uses them instead of going to the core library.

hi-sushanta:
My main motive is to create a separate file to handle all types of loss calculation, similar to the PyTorch loss.py file.

geohot:
Closing per what @g1y5x3 said.

PR #2086, opened 2023-10-16
MulNode.__lt__ rule

ZenginU:
Mathematically seems true also tests pass.

### DIFF
--- a/tinygrad/shape/symbolic.py
+++ b/tinygrad/shape/symbolic.py
@@ -196,7 +196,7 @@ class LtNode(OpNode):
 
 class MulNode(OpNode):
   def __lt__(self, b: Union[Node, int]):
-    if isinstance(b, Node) or isinstance(self.b, Node): return Node.__lt__(self, b)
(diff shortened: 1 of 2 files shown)


tinygrad/shape/symbolic.py
```diff
@@ -195,6 +195,10 @@ def get_bounds(self) -> Tuple[int, int]:
   def substitute(self, var_vals: Dict[VariableOrNum, Node]) -> Node: return self.a.substitute(var_vals) < (self.b if isinstance(self.b, int) else self.b.substitute(var_vals))
 
 class MulNode(OpNode):
+  def __lt__(self, b: Union[Node, int]):
+    if isinstance(b, Node) or isinstance(self.b, Node): return Node.__lt__(self, b)
```
roelofvandijk, line 199: This line is odd: if `isinstance(b, Node)` false, it's not a Node, so how can it have a `.b` member? Mypy should trigger here.

ZenginU: self is Node, but self.b or b can be int.

I did not understand the problem

roelofvandijk: You're completely right. Sorry.

PR #2090, opened 2023-10-17
Disable logging in early compile2 and lower kernel counts

geohot:
This reverts commit 924ecc4d6a05327814d9b373beafdef6383ddaf1.


geohot:
This alone makes `WINO=1 STEPS=10 python3 examples/hlb_cifar10.py` slower.

PR #2093, opened 2023-10-17
Added dir(tinygrad) support.

atveit:
Added dir(tinygrad) support, example shown below. Also added similar support for other packages. All uses lazy loading with getattr to avoid any side effects.

>>> import tinygrad
>>> dir(tinygrad)
['__all__', '__builtins__', '__cached__', '__dir__', '__doc__', '__file__', '__getattr__', '__loader__', '__name__', '__package__', '__path__', '__spec__', **'codegen', 'features', 'nn', 'renderer', 'runtime', 'shape'**]

related quote from tweet: [Why does tinygrad not import it's submodules? Can't even do `dir(tinygrad)` to know what submodules are present.](https://twitter.com/susshree/status/1714264253249802569)


geohot:
So many lines, wrong spacing! I thought this was one file.

PR #2094, opened 2023-10-17
added support dir(tinygrad)

atveit:
ref. https://twitter.com/susshree/status/1714264253249802569

"Can't even do `dir(tinygrad)` to know what submodules are present."

(this only updates one __init__.py)

### DIFF
--- a/tinygrad/__init__.py
+++ b/tinygrad/__init__.py
@@ -1,0 +1,10 @@
+__all__ = ["nn", "codegen", "renderer", "runtime", "shape", "features"]
+
+def __getattr__(name):
+  if name in __all__:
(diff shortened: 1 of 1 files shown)


tinygrad/__init__.py
```diff
@@ -0,0 +1,10 @@
+__all__ = ["nn", "codegen", "renderer", "runtime", "shape", "features"]
```
geohot, line 1: This is duplicated with setup

tinygrad/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+
+def __getattr__(name):
```
geohot, line 3: Is writing this code really the python standard way to do this?

PR #2095, opened 2023-10-17
Reenable limit_dims_to_max

wpmed92:
It got removed [here](https://github.com/tinygrad/tinygrad/pull/1723/files), and I added back during the webgpu stable diffusion [port](https://github.com/tinygrad/tinygrad/pull/1370#discussion_r1362754621) only on webgpu, but it can be used anywhere else, so we're adding it back.

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -151,8 +151,7 @@ class Linearizer(OptimizedKernel):
     self.saved_exprs: Dict[Tuple, UOp] = dict()
 
     # limit dims if we need to
-    # TODO: broken, and doesn't really belong here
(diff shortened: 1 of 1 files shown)


geohot:
Why did it say it was broken? Do we have good tests for this?

tobias17:
The following causes CUDA to fail compilation with it commented out, might be worth adding in as a regression test when we re-enable.
```python
N = 139811
a = Tensor(np.random.randn(N,1,1)).expand((N,5,3,))
b = Tensor(np.random.randn(N,5,3))
(a*b).realize()
```
Need it re-enabled for some of the quantization stuff to work.

geohot:
Fine, added back

PR #2098, opened 2023-10-18
I added mse and mae loss directly into the tensor.py file

hi-sushanta:
If it's ok now I directly add mae and mse loss into the tensor.py file.

And also see what looks like that:
Screenshot from 2023-10-18 06-29-14


wozeparrot:
needs tests to compare against torch, and weird formatting at the end of the file.

geohot:
I don't like this feature period. You should just write that in your trainer.

PR #2099, opened 2023-10-18
fix: hlb cifar types

seendsouza:
```
❯ mypy examples/hlb_cifar10.py
Success: no issues found in 1 source file
```

### DIFF
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -11,6 +11,7 @@ if __name__ == "__main__":
 # https://siboehm.com/articles/22/CUDA-MMM
 import random, time
 import numpy as np
+from typing import Any, Dict, Optional, SupportsIndex, Type, Union
 from extra.datasets import fetch_cifar, cifar_mean, cifar_std
 from tinygrad import nn
 from tinygrad.nn.state import get_state_dict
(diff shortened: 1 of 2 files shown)


g1y5x3:
It would be nice if CI can spot this so they can be fixed on the fly next time.

seendsouza:
Hmm I think it's because it's in examples. There are 228 errors in other examples right now @g1y5x3

examples/hlb_cifar10.py
```diff
@@ -127,7 +129,8 @@ def _cov(X):
     def _patches(data, patch_size=(kernel_size,kernel_size)):
       h, w = patch_size
       c = data.shape[1]
-      return np.lib.stride_tricks.sliding_window_view(data, window_shape=(h,w), axis=(2,3)).transpose((0,3,2,1,4,5)).reshape((-1,c,h,w))
+      axis: SupportsIndex = (2, 3) # type: ignore
```
geohot, line 132: Hmm, why? Does the array [2,3] does work?

seendsouza: Nope. It's very weird. Int works, but not tuple or list

geohot:
`extra` I don't care about, but `examples`, `models`, and `test` should be type checked. However, this needs to be done tastefully, not with a huge commit, but with thought into what should just be deleted.

PR #2104, opened 2023-10-18
add test coverage for search

Qazalin:
preventing #2101

example failed case for LLVM before the fix commit:
```
self = <test.test_search.TestTimeLinearizer testMethod=test_reasonable_time>                                                                          
                                                                                                                                                                      
    def test_reasonable_time(self):                                                                                                                   
      si = [si for si in Tensor([1,2,3,4]).add(1).lazydata.schedule() if si.ast.op not in LoadOps][0]                                                  
      rawbufs = [Device[Device.DEFAULT].buffer(si.out.st.size(), si.out.dtype)] + [Device[Device.DEFAULT].buffer(x.st.size(), x.dtype) for x in si.inputs]                                 
      tm = time_linearizer(Linearizer(si.ast), rawbufs, allow_test_size=False, cnt=10, should_copy=False)                                              
>     assert tm > 0 and tm != float('inf')                                                                                                            
E     AssertionError: assert (inf > 0 and inf != inf)                                                                                                                 
E      +  where inf = float('inf')                                                                                                                                    
                                                                                                                                                                      
test/test_search.py:13: AssertionError 
```


Qazalin:
will try to find time to fix triton and cuda

test/test_search.py
```diff
@@ -0,0 +1,14 @@
+import unittest
+
+from tinygrad.codegen.linearizer import Linearizer
+from tinygrad.features.search import time_linearizer
+from tinygrad.ops import Compiled, Device, LoadOps
+from tinygrad.tensor import Tensor
+
+@unittest.skipIf(not isinstance(Device[Device.DEFAULT], Compiled), "only test for compiled backends")
```
chaosagent, line 8: don't access device[device.default] in the decorator, it doesn't play well with some backends (eg ocelot) and causes very mysterious test failures. can just write this in setUp.

Qazalin: changed for this one.
we use this pattern a lot tho - might need a codemod

Qazalin:
yeah the negative time in CI is weird. I get positive numbers for both CUDA=1 TRITON=0 and CUDA=1 TRITON=1 on google colab's gpu runtime, it's more likely to be something specific to the CI env than the implementation itself.
gonna have to debug this live on gh actions

tinygrad/runtime/ops_cuda.py
```diff
@@ -33,7 +33,7 @@ def __call__(self, *args, block, grid, shared): lib.ptx_run(self.src, len(args),
     class Event:
       def __init__(self): pass
       def record(self): self.start = time.perf_counter()
-      def time_till(self, other): return self.start - other.start
+      def time_till(self, other): return other.start - self.start
```
geohot, line 36: lol yea I knew it was something dumb

Qazalin: got lucky

PR #2109, opened 2023-10-18
feat: cache apt packages for Install packages (cuda)

seendsouza:
Current status: waiting for CI to run once, and then rerunning it to see how well apt cache action works


geohot:
Umm, did you look at the speed?

seendsouza:
30s faster but breaks ocelot :(

PR #2117, opened 2023-10-19
examples/waifu2x: testcase

20kdc:
* Adds testcase for the main code of the waifu2x vgg7 example. A "frozen" output of the tinygrad implementation is being used to test itself for now, but if a better reference can be found it should be used (in practice this would catch X/Y transpositions - the model isn't really complex enough for any other kind of error to sneak through)
* Function to download and load a vgg7 pretrained model.
* The image load/convert function will now strip off the alpha channel if found, resulting in less mysterious errors.
* The zero-dimension-tensor created by `.max()` changed, so now that's handled properly.

Given that I don't have a good way to test the model with an independent implementation (the official implementation is Lua-Torch7-CUDA so not exactly easy to arrange even the creation of a test vector), feel free to close the PR if that is not acceptable.

Something that ought to be brought up:

Given as noted in #2082 this is a heavily outdated model (hence my problems finding other implementations to test against), it may indeed be better to simply remove the model. This especially applies since the training side of the example seems to possibly be more broken than when the code was originally committed (I had at that time at least been training some models, though they were not good models - current attempts immediately NaN-flood after a single step).


geohot:
Neat, I like that it's clear from the two example pictures what this model does

PR #2118, opened 2023-10-19
minor triton casting refactor

Qazalin:
cleaning up cast renders in a single place. the bitcast implementation is much more readable if it's isolated in `render_cast`.


tinygrad/renderer/triton.py
```diff
@@ -78,13 +81,13 @@ def int_div(x,y): return f"({x}//{y})"
       assert dtype is not None
       val = code_for_op[args](*[r[x] for x in vin])
       if child_count[u] <=1 or dtypes.is_int(dtype): r[u] = int_div(*[r[x] for x in vin]) if args == BinaryOps.DIV and dtypes.is_int(dtype) else val
-      else: kk(f"{ssa(u, 'alu')} = ({val}).to({triton_dtypes[dtype]})")
+      else: kk(f"{ssa(u, 'alu')} = ({val}).{render_cast(dtype)}")
```
geohot, line 84: should be `render_cast(val, dtype)` I think

PR #2119, opened 2023-10-20
Improve triton generated code quality

SzymonOzog:
This removed the ugly `tl.where(...)` from all scalar definitions

### DIFF
--- a/tinygrad/renderer/triton.py
+++ b/tinygrad/renderer/triton.py
@@ -26,11 +26,11 @@ def remove_single_scalar_curly_braces(ptx_code):
   return '\n'.join([re.sub(r'\{\s*(%\w+)\s*\}', r'\1', line) for line in ptx_code.split('\n')])
 
 def render_const(args):
-  return (('-' if args<0 else '') + 'float("inf")') if math.isinf(args) else ('float("nan")' if math.isnan(args) else str(args))
+  return (('-' if args<0 else '') + 'tl.where(1,float("inf"),0)') if math.isinf(args) else ('tl.where(1,float("nan"),0)' if math.isnan(args) else str(args))
 
 def define_scalar(local_size, triton_type, args):
   if len(local_size) > 0: return f"tl.full(({','.join([str(next_power_of_2(x)) for x in local_size])},),{render_const(args)}, dtype={triton_type})"
-  return f"(tl.where(1, {render_const(args)}, {render_const(args)}).to({triton_type}))"
+  return render_const(args)
 
 def uops_to_triton(function_name:str, uops:List[UOp]):
   local_size: List[int] = []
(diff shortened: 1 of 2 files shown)


tinygrad/renderer/triton.py
```diff
@@ -58,14 +58,14 @@ def kk(s): kernel.append("  "*depth+s)
     UnaryOps.SQRT: lambda x,: f"tl.sqrt({x})",
     UnaryOps.NEG: lambda x,: f"-{x}",
     BinaryOps.ADD: lambda x,y,: f"({x}+{y})", BinaryOps.SUB: lambda x,y,: f"({x}-{y})",
-    BinaryOps.MUL: lambda x,y,: f"({x}*{y})", BinaryOps.DIV: lambda x,y,: f"({x}/{y})",
+    BinaryOps.MUL: lambda x,y,: f"({x}*{y})", BinaryOps.DIV: lambda x,y,: f"({x}/{y})" if y != '0.0' else f"{x}*tl.where({x}==0.0, float('nan'), float('inf'))",
     BinaryOps.MAX: lambda x,y,: f"tl.maximum({x},{y})",
     BinaryOps.CMPLT: lambda x,y,: f"({x}<{y})",
-    BinaryOps.MOD: lambda x,y,: f"({x}%{y})",
+    BinaryOps.MOD: lambda x,y,: f"tl.abs({x})%tl.abs({y})*tl.where({x}<0,-1,1)",
```
geohot, line 64: Why does this have to change?

tinygrad/renderer/triton.py
```diff
@@ -58,14 +58,14 @@ def kk(s): kernel.append("  "*depth+s)
     UnaryOps.SQRT: lambda x,: f"tl.sqrt({x})",
     UnaryOps.NEG: lambda x,: f"-{x}",
     BinaryOps.ADD: lambda x,y,: f"({x}+{y})", BinaryOps.SUB: lambda x,y,: f"({x}-{y})",
-    BinaryOps.MUL: lambda x,y,: f"({x}*{y})", BinaryOps.DIV: lambda x,y,: f"({x}/{y})",
+    BinaryOps.MUL: lambda x,y,: f"({x}*{y})", BinaryOps.DIV: lambda x,y,: f"({x}/{y})" if y != '0.0' else f"{x}*tl.where({x}==0.0, float('nan'), float('inf'))",
```
geohot, line 61: I don't like that these compares are to a string

tinygrad/renderer/triton.py
SzymonOzog: Triton uses the same mod as python and it's different than what tinygrad expects eg. `1.0%-3.0==-1` where we expect `2`, its basically copied from https://github.com/tinygrad/tinygrad/blob/36d4001b4f6541724071e2c78cb93e235877902e/test/test_uops.py#L92C141-L92C141. My guess is that when we changed from tl.where to constexpr it moved to doing this in python compile time vs doing it in runtime and it changed the value

tinygrad/renderer/triton.py
SzymonOzog: `x` can be a variable name so we can't cast to float and since triton can handle
``` 
x = 0
y = 1/x
```
and cant handle `y=0/0` it is enough to just compare strings. I could generate the comparison in code but I think that is unnecessary computation that can be resolved during codegen

tinygrad/renderer/triton.py
geohot: Ohh..if these are consts. Yea, consts uops should just be constant folded and not passed to here.

PR #2122, opened 2023-10-20
fuzzing linearizer transformations

Qazalin:
(this has been very stale, i'll come back to working on this once i finish off the bitcast PRs)

This fuzzes compiled devices and asts from the kernel dataset.
I also wanted to fuzz random buffer inputs but that wasn't possible because memory ops aren't scheduled. right now we just pass 0s, if the kernel is doing some math that doesn't impact 0s the accuracy test could give a false positive.

TODO:
- [ ] add beam to optimizers - _maybe there's a nicer pattern for getting optimizers than just manually adding them to a list?_
- [ ] There's no time or gflops based assertion, it only asserts accuracy - should it do some quality checks like:
   - one device doesn't exceed a timeout or take a lot longer to compute
   - the optimizer takes less time than `base_tm`
- [ ] If the fuzzer is gonna be _really_ cross-device we need to have a remote location to store results from the runs, ie. METAL can't run on the same vm that runs CUDA.
- [ ] Could `bufs_from_lin` be faster?


geohot:
It should really randomly pick optimizations and confirm that they are the same.

Qazalin:
closing in favor of 2188:)

PR #2123, opened 2023-10-20
fix elementwise contiguous reshape chain

sahamrit:
```
a = Tensor.randn(3).realize()
for _ in range(100):
  a = a.reshape(1,3)
  a = a.exp().contiguous()
  a = a.reshape(3)
  a = a.contiguous()
a.realize()
```

This generates 101 kernels on master
image

This fix leads to 2 kernels
image

### DIFF
--- a/test/external/external_test_opt.py
+++ b/test/external/external_test_opt.py
@@ -11,7 +11,7 @@ import numpy as np
 import unittest
 from tinygrad.tensor import Tensor, Device
 from tinygrad import nn
-from tinygrad.helpers import getenv
+from tinygrad.helpers import getenv, dtypes
 from tinygrad.nn import optim
 from tinygrad.ops import GlobalCounters, MovementOps, ReduceOps
 from tinygrad.lazy import PUSH_CONTIGUOUS, PUSH_PERMUTES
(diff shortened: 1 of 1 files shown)


tinygrad/lazy.py
sahamrit, line 1: single line - allow_contiguous = PUSH_CONTIGUOUS, will update

tinygrad/lazy.py
```diff
@@ -256,7 +256,10 @@ def _movement_op(self, st: ShapeTracker, op: MovementOps, arg: Union[Tuple[sint,
         return self.op.replace_with_movement_ops([(op, arg)])
     if REMOVE_MOVEMENT_NOPS and not self.realized and st.contiguous:
       # MovementOps aren't stacked any more, they each have one parent, find the root
-      root = get_movementroot(self)
+      if PUSH_CONTIGUOUS:
+        root = get_movementroot(self, allow_contiguous=True)
```
sahamrit, line 260: Single line, allow_contiguous = PUSH_CONTIGUOUS, will update

sahamrit:
@geohot check this out. The openpilot issue #1781 wasn't entirely solved. TLDR - The issue was never with reshapes being pushed through BinaryOps, but that CONTIGUOUS wasn't being pushed down ( @ OPT = 99) & was being blocked by reshape. See the explanation below

1. Reshapes are already pushed through BinaryOps, but this happens within `_push_movement_ops` (which is called from `def e()` ), which is when we have hope that we can get fusion of BinaryOps.
2. Pushing Reshape within `_movement_op` doesn't make sense because during this point of time we don't even know what is about to follow in the graph. `_push_movement_ops` is where we have enough information to decide if we want to push.

Now coming onto openpilot, see the image below
image

Till now we were trying to push the bottommost reshape up [highlighted as red - 1], but the actual problem was purple where contiguous wasn't able to be pushed down due to a reshape blocking it.

Hope things are clear!

tinygrad/lazy.py
```diff
@@ -252,11 +251,11 @@ def splitted_shape(dim_aft_div): return self.shape[:dim_to_split] + (self.shape[
 
   def _movement_op(self, st: ShapeTracker, op: MovementOps, arg: Union[Tuple[sint, ...], Tuple[Tuple[sint, sint], ...]]) -> LazyBuffer:
     if SHUFFLE_MOVEMENT_OPS and not self.realized and self.optype == BinaryOps and not self.children:
-      if op in {MovementOps.SHRINK, MovementOps.STRIDE, MovementOps.PERMUTE} or (op == MovementOps.RESHAPE and (self.op.op in UnaryOps or PUSH_RESHAPES)):
+      if op in {MovementOps.SHRINK, MovementOps.STRIDE, MovementOps.PERMUTE} or (op == MovementOps.RESHAPE and (self.op.op in UnaryOps)):
```
roelofvandijk, line 254: `(self.op.op in UnaryOps)` -> `self.op.op in UnaryOps`

geohot:
Pushing contiguous shouldn't be a thing. Contiguous is meant to break the fusion (see how it's used in Winograd), the 101 kernels is correct.

PR #2128, opened 2023-10-22
fix #1751: align np array before copying into cl buffer 

### LINKED ISSUE #1751: GPT2 fails with GPU=1 backend on Mac
Test with `GPU=1 python examples/gpt2.py --prompt="Hello." --count=10 --temperature=0` on M1 Max

On master, it errs with `ValueError: probabilities contain NaN`
The error was always there. In the first commit of gpt2 example `e3c6c0c6db3593592fa8302443ae3fb964eea112`, the output was `Hello.!!!!!!!!!!`.

Based on benchmark, GPU backend works on AMD card.

The issue seems to be that assigning `wte.weight` (and `wpe.weight`) to `GPU` resulted all NaN, but assigning to `METAL` is fine.

```
name='wte.weight'
weights[name].device='DISK:/var/folders/w1/wnz2yz_j2ggc6_7r5xv3vg540000gn/T/ca15bfc1e4e381f33a0a2c9ddffc61c6'
np.isnan(weights[name].numpy()).sum()=0
np.isnan(weights[name].to('CPU').to('GPU').numpy()).sum()=219880
np.isnan(weights[name].to('GPU').numpy()).sum()=219880
np.isnan(weights[name].to('METAL').numpy()).sum()=0
np.isnan(weights[name].to('CPU').to('METAL').numpy()).sum()=0
np.isnan(weights[name].to('METAL').to('GPU').numpy()).sum()=0

name='h.0.ln_1.weight'
weights[name].device='DISK:/var/folders/w1/wnz2yz_j2ggc6_7r5xv3vg540000gn/T/ca15bfc1e4e381f33a0a2c9ddffc61c6'
np.isnan(weights[name].numpy()).sum()=0
np.isnan(weights[name].to('CPU').to('GPU').numpy()).sum()=0
np.isnan(weights[name].to('GPU').numpy()).sum()=0
np.isnan(weights[name].to('METAL').numpy()).sum()=0
np.isnan(weights[name].to('CPU').to('METAL').numpy()).sum()=0
np.isnan(weights[name].to('METAL').to('GPU').numpy()).sum()=0
```

imaolo:
Hard to add a test for this because I was only able to reproduce with https://huggingface.co/gpt2-medium/resolve/main/pytorch_model.bin. Figured we wouldn't want a fetch happening in CI (recent effort to decrease test times).

All the weights are unaligned when sent to the CPU from disk, and roughly half of them get corrupted when they are copied into the cl buffer.

There is a performance hit as unaligned buffers need to be copied. I am seeing +~5% wall clock on M1. There is no impact on arrays that are already aligned.

Update: We may want to consider putting this in RawBufferCopyIn for sanity's sake.

#1751

### DIFF
--- a/tinygrad/runtime/ops_gpu.py
+++ b/tinygrad/runtime/ops_gpu.py
@@ -48,6 +48,7 @@ if not getenv("DELAYED_RUNTIME_INIT", False): CL.post_init()
 class CLBuffer(RawBufferCopyInOut, RawBufferTransfer):
   def __init__(self, size, dtype, device='0'): super().__init__(size, dtype, allocator=CL.cl_allocator, **{'device': device})
   def _copyin(self, x:np.ndarray):
+    x = np.require(x, requirements=['A'])
     assert not self.dtype.name.startswith("image"), f"can't copyin images {self.dtype}"
     self.event = cl.enqueue_copy(CL.cl_queue[self._buf.device], self._buf, np.require(x, requirements='C'), is_blocking=False)
   def _copyout(self, x:np.ndarray):
--- a/tinygrad/runtime/ops_gpu.py
+++ b/tinygrad/runtime/ops_gpu.py
@@ -48,9 +48,8 @@ if not getenv("DELAYED_RUNTIME_INIT", False): CL.post_init()
 class CLBuffer(RawBufferCopyInOut, RawBufferTransfer):
   def __init__(self, size, dtype, device='0'): super().__init__(size, dtype, allocator=CL.cl_allocator, **{'device': device})
   def _copyin(self, x:np.ndarray):
-    x = np.require(x, requirements=['A'])
     assert not self.dtype.name.startswith("image"), f"can't copyin images {self.dtype}"
-    self.event = cl.enqueue_copy(CL.cl_queue[self._buf.device], self._buf, np.require(x, requirements='C'), is_blocking=False)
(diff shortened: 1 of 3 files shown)


tinygrad/runtime/ops_gpu.py
```diff
@@ -48,6 +48,7 @@ def synchronize(self):
 class CLBuffer(RawBufferCopyInOut, RawBufferTransfer):
   def __init__(self, size, dtype, device='0'): super().__init__(size, dtype, allocator=CL.cl_allocator, **{'device': device})
   def _copyin(self, x:np.ndarray):
+    x = np.require(x, requirements=['A'])
```
wozeparrot, line 51: can the align requirement just be arrayed with the contiguous requirement below?

imaolo: Contiguous did not work. I forget to mention, should we consider aligning all numpy arrays in RawBufferCopyIn? Updating the original comment as well.

wozeparrot: I meant like does this have to be a seperate line, can the 'A' just be put in the same requirements as the 'C' below

imaolo: oh lol, yes my fault.

geohot:
Nice find!

wozeparrot:
Do we want to have this or the contiguous requirement directly in RawBufferCopyIn? iirc all the runtimes that use RawBufferCopyIn have a least the contiguous requirement.

imaolo:
> Nice find! We should really have a regression test before I merge though.

working on one

PR #2129, opened 2023-10-22
stable diffusion < 324ms

chenyuxyz:
should be able to reproduce by copying `beam_cache.db` to `/tmp/` and run `PYTHONPATH=. BEAM=4 WINO=1 python3 examples/stable_diffusion.py --timing --seed=0 --fp16`. Searching from scratch should be around 6~8 hours.

image

can get 462ms without BEAM.

change details
macro: enabled `PUSH_PERMUTES` and `PUSH_RESHAPES`, lowered the threshold for reduce kernel splitting (split more)
kernel: added action to swap dimensions and yellows, added more UPCAST/LASTLOCAL options in tensor core heuristic. Also compared the hand_code time and tensor core time for the baseline.
other: tuned the launch batch size in ops_metal

#2124 changed action space slightly so I would need to redo search if I want to sync to master. Also I am not sure how many changes here we want to merge.

### DIFF
--- a/test/external/external_test_opt.py
+++ b/test/external/external_test_opt.py
@@ -64,14 +64,14 @@ class TestInferenceMinKernels(unittest.TestCase):
     for p in get_parameters(model): p.assign(np.zeros(p.shape, dtype=p.dtype.np))
     img = Tensor.randn(1, 3, 224, 224)
     # TODO: this seems very high
-    with CLCache(115):
+    with CLCache(116):
       model.forward(img).realize()
 
   def test_resnet(self):
     model = ResNet18()
     for p in get_parameters(model): p.assign(np.zeros(p.shape, dtype=p.dtype.np))
     img = Tensor.randn(1, 3, 224, 224)
-    with CLCache(26):
+    with CLCache(29):
       model.forward(img).realize()
 
   def test_vit(self):
--- a/test/external/external_test_opt.py
+++ b/test/external/external_test_opt.py
@@ -64,14 +64,14 @@ class TestInferenceMinKernels(unittest.TestCase):
     for p in get_parameters(model): p.assign(np.zeros(p.shape, dtype=p.dtype.np))
     img = Tensor.randn(1, 3, 224, 224)
     # TODO: this seems very high
-    with CLCache(116):
+    with CLCache(115):
       model.forward(img).realize()
 
   def test_resnet(self):
     model = ResNet18()
     for p in get_parameters(model): p.assign(np.zeros(p.shape, dtype=p.dtype.np))
     img = Tensor.randn(1, 3, 224, 224)
-    with CLCache(29):
+    with CLCache(26):
       model.forward(img).realize()
 
   def test_vit(self):
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -433,7 +433,7 @@ class TestLinearizerOpts(unittest.TestCase):
     if not isinstance(Device[Device.DEFAULT], Compiled) or not Device[Device.DEFAULT].linearizer_opts.has_local or not Device[Device.DEFAULT].linearizer_opts.has_shared:
       self.skipTest("Only Compiled uses linearizer with locals and shared")
 
-    N = 128
+    N = 32  # set N such that the sum is not split into 2 kernels
     Tensor.manual_seed(1552)
     a = Tensor.rand(8, N, 8, N)
     r = a.sum(axis=(1,3))
@@ -443,7 +443,7 @@ class TestLinearizerOpts(unittest.TestCase):
       [Opt(OptOps.GROUPTOP, 1, 2)], [Opt(OptOps.GROUPTOP, 1, 32)], # Checking how it works with 1 grouped_reduce.
       [Opt(OptOps.GROUPTOP, 0, 2), Opt(OptOps.GROUPTOP, 1, 2)],
       [Opt(OptOps.GROUPTOP, 0, 16), Opt(OptOps.GROUPTOP, 1, 2)],
-      [Opt(OptOps.GROUPTOP, 0, 4), Opt(OptOps.GROUPTOP, 1, 64)], # Checking how it works with 2 grouped_reduces.
+      [Opt(OptOps.GROUPTOP, 0, 4), Opt(OptOps.GROUPTOP, 1, 16)], # Checking how it works with 2 grouped_reduces.
       [Opt(OptOps.GROUPTOP, 0, 16), Opt(OptOps.GROUPTOP, 1, 2), Opt(OptOps.UNROLL, 0, 4)],
       [Opt(OptOps.GROUPTOP, 0, 2), Opt(OptOps.GROUPTOP, 1, 32), Opt(OptOps.UNROLL, 2, 4)], # Checking how it works with 2 grouped_reduces + upcasts.
       [Opt(OptOps.LOCAL, 0, 4), Opt(OptOps.LOCAL, 1, 4), Opt(OptOps.GROUPTOP, 0, 4), Opt(OptOps.GROUPTOP, 1, 4)],
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -433,7 +433,7 @@ class TestLinearizerOpts(unittest.TestCase):
     if not isinstance(Device[Device.DEFAULT], Compiled) or not Device[Device.DEFAULT].linearizer_opts.has_local or not Device[Device.DEFAULT].linearizer_opts.has_shared:
       self.skipTest("Only Compiled uses linearizer with locals and shared")
 
-    N = 32  # set N such that the sum is not split into 2 kernels
+    N = 128
     Tensor.manual_seed(1552)
     a = Tensor.rand(8, N, 8, N)
     r = a.sum(axis=(1,3))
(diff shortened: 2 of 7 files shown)


tinygrad/ops.py
```diff
@@ -263,20 +263,40 @@ def exec_ast(self, ast:LazyOp, output, inputs, var_vals, **kwargs):
     # compilation time
     def get_program():
       from tinygrad.codegen.linearizer import Linearizer
+      from tinygrad.features.search import beam_search, time_linearizer
+
       k = Linearizer(ast, self.linearizer_opts)
       assert k.info.dtype == output.dtype, f"linearizer must match dtype. linearizer wants {k.info.dtype} but buffer is {output.dtype}"
       if not getenv("NOOPT"):
-        if not k.apply_tensor_cores(getenv("TC", 1)): k.hand_coded_optimizations()
+        ktc = k.copy()
```
geohot, line 271: I think this should only happen if we are searching with BEAM, otherwise listen to TC

tinygrad/features/search.py
```diff
@@ -17,6 +17,9 @@
   Opt(op=OptOps.GROUPTOP, axis=1, amt=16), Opt(op=OptOps.GROUPTOP, axis=1, amt=256),
   Opt(op=OptOps.GROUPTOP, axis=2, amt=16), Opt(op=OptOps.GROUPTOP, axis=2, amt=256)
 ]
+actions += [
+  Opt(op=OptOps.SWAP, axis=x, amt=y) for x in range(16) for y in range(x+1,16)
```
geohot, line 21: Hmm, amt doesn't mean amt here...

examples/stable_diffusion.py
```diff
@@ -636,6 +638,8 @@ def do_step(latent, timestep, index, guidance):
   if args.seed is not None: Tensor._seed = args.seed
   latent = Tensor.randn(1,4,64,64)
 
+  BEAM.value = getenv("BEAM")
+
   # this is diffusion
```
geohot, line 643: I would do this as LATEBEAM like how I do it in compile2

`with Context(BEAM=getenv("LATEBEAM"):`

geohot:
So I see a bunch of changes here:
* Reduce heuristic //= 4. This is probably fine to merge but should be tested in a smart way.
* Adding tensor cores to the BEAM search. Currently both are benchmarked, I'd like to keep all benchmarking inside BEAM, but I'm fine with this otherwise.
* Metal group size 8 -> 128. There were some issues with larger groups here, maybe 128 is okay? Need to test this, maybe check out the old conversation about this.
* Addition of /tmp/beam_cache. Fine, but needs to be set by an env var.
* I'm fine with SWAP in general, but I don't like how it breaks the axis/amt args. What's a nice way to handle this?

Concerning changes:
* PUSH_RESHAPES/PUSH_PERMUTES to OPT>=2, definitely not. This causes issues with other models. Change needs to be more specific than this,
* The example shouldn't change behavior of BEAM or DEBUG, but add another env var like LATEBEAM to control it.
* Committing a 61.9 MB beam_cache.db to master

geohot:
Tried testing, got `dbm.error: db type could not be determined`

geohot:
Implementing this stuff in #2130

chenyuxyz:
rebased to master and re-run the search. Also reverted the need of PUSH_RESHAPES/PUSH_PERMUTES (this became slower but very close, not that significant), and use `//8` instead of `//4` for reduce split (very significant, saved 30ms).

i was able to get 297ms with this setup
image

the beam=4 search took 4.5~5 hours from scratch. something is wrong with the cache because my `tinygrad_cache` is 9.5G now so i cannot share it. I can look into this later, along with cleaning up the SWAP arguments.

tinygrad/lazy.py
```diff
@@ -242,7 +242,7 @@ def _reduce_op(self:LazyBuffer, op:ReduceOps, new_shape:Tuple[sint, ...]) -> Laz
     return create_lazybuffer(self.device, ShapeTracker.from_shape(new_shape), ReduceOps, LazyOp(op, srcs, unbound_new_shape), self.dtype)
 
   def r(self:LazyBuffer, op:ReduceOps, new_shape:Tuple[sint, ...]) -> LazyBuffer:
-    if any(not isinstance(s, int) for s in self.shape) or prod(self.shape) // prod(new_shape) < 32768: return self._reduce_op(op, new_shape) # The amount of work should be big enough to take the benefit of "2 kernels" approach.
+    if any(not isinstance(s, int) for s in self.shape) or prod(self.shape) // prod(new_shape) < 4096: return self._reduce_op(op, new_shape) # The amount of work should be big enough to take the benefit of "2 kernels" approach.
```
roelofvandijk, line 245: `any(not isinstance(s, int) for s in self.shape)` -> `not all_int(self.shape)`

https://github.com/tinygrad/tinygrad/blob/3c56c181f6a9c528702d25dcc2a483dbb8ffca20/tinygrad/helpers.py#L20

geohot:
All three of those are fine

test/test_linearizer.py
```diff
@@ -433,7 +433,7 @@ def test_double_reduce(self):
     if not isinstance(Device[Device.DEFAULT], Compiled) or not Device[Device.DEFAULT].linearizer_opts.has_local or not Device[Device.DEFAULT].linearizer_opts.has_shared:
       self.skipTest("Only Compiled uses linearizer with locals and shared")
 
-    N = 128
+    N = 32  # set N such that the sum is not split into 2 kernels
```
chenyuxyz, line 436: cc @flammit i lowered this so that the sum is not splitting into two kernels with the new heuristic. Please let me know if this breaks any assumptions you have for this test.

flammit: I think the sizes of the shifts just need to be brought down as well (change the 32/16 down to 4/2 or something) so the structure stays the same.

chenyuxyz:
something is wrong with `examples/hlb_cifar10.py`. investigating

flammit:
> updated the test, more kernels due to more reduce splitting.
> 
> re: metal heuristic size: i think the one instance you had that slowed down stable diffusion was due to a bug that only split the graph into 2, not batches ([b3c3fae](https://github.com/tinygrad/tinygrad/commit/b3c3fae677ff3515492f33369f75e96532813ba4)). And i get a 10ms+ improvement with batch size 128 stable diffusion

Sorry I should've been more clear: I was using the `METAL=1 python3 ./test/test_speed_v_torch.py -k matvec` to test the speed differences between GROUP/GROUPTOP.  When I wrote the matvec opts, I didn't test the difference, so I think it's safe to just use Top=True.

flammit:
> > > updated the test, more kernels due to more reduce splitting.
> > > re: metal heuristic size: i think the one instance you had that slowed down stable diffusion was due to a bug that only split the graph into 2, not batches ([b3c3fae](https://github.com/tinygrad/tinygrad/commit/b3c3fae677ff3515492f33369f75e96532813ba4)). And i get a 10ms+ improvement with batch size 128 stable diffusion
> > 
> > 
> > Sorry I should've been more clear: I was using the `METAL=1 python3 ./test/test_speed_v_torch.py -k matvec` to test the speed differences between GROUP/GROUPTOP. When I wrote the matvec opts, I didn't test the difference, so I think it's safe to just use Top=True.
> 
> oh i was replying to the ops_metal.py using batch 128 made it slower earlier in geohot's comment, sorry for the confusion. and i don't quite get your reply. do you see thinks that I can test for this PR to make it more robust?

Whoops sorry I was thinking about a different PR :(

test/external/external_test_opt.py
```diff
@@ -64,14 +64,14 @@ def test_enet_se(self):
...
-    with CLCache(115):
+    with CLCache(116):
       model.forward(img).realize()
 
   def test_resnet(self):
     model = ResNet18()
     for p in get_parameters(model): p.assign(np.zeros(p.shape, dtype=p.dtype.np))
     img = Tensor.randn(1, 3, 224, 224)
-    with CLCache(26):
+    with CLCache(29):
```
geohot: Hmm, that's a lot for a forward pass. Is it faster?

chenyuxyz: so i reverted the default change and replaced with a env var. I tested around and some kernels became slower (gemm 4096x4096 became 3X slower), so i don't see a universal good default. benchmarking here is tricky because it might be faster with handcode but slower with best beam.

chenyuxyz:
gated the kernel split threshold with a env var flag. the other 2 are metal specific and i tested with cifar, gpt-2, llama, speed_v_torch and all are either slightly better for python time or no difference.

command: `PYTHONPATH=. LATEBEAM=2 WINO=1 REDUCEOP_SPLIT_THRESHOLD=4096 python3 examples/stable_diffusion.py --timing --seed=0 --fp16`

search from scratch took ~80 minutes, can get < 297ms per step

image

geohot:
Going to merge my dictionary keys fix and run this search on my machine. Sub 324 we'll merge and the bounty is yours!

geohot:
Congrats!

```
kafka@tubby:~/tinygrad$ PYTHONPATH=. LATEBEAM=2 WINO=1 REDUCEOP_SPLIT_THRESHOLD=4096 python3 examples/stable_diffusion.py --timing --seed=0 --fp16
ram used:  4.26 GB, cond_stage_model.transformer.text_model.final_layer_norm.bias: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1131/1131 [00:00<00:00, 1280.72it/s]
loaded weights in 892.58 ms, 4.26 GB loaded at 4.78 GB/s
got CLIP context (1, 77, 768)
got unconditional CLIP context (1, 77, 768)
running for [1, 201, 401, 601, 801] timesteps
  4 801:   0%|                                                                                                                                                                                                                       | 0/5 [00:00<?, ?it/s]step in 80143.80 ms, using 1.89 GB
  3 601:  20%|█████████████████████████████████████████▍                                                                                                                                                                     | 1/5 [01:20<05:20, 80.14s/it]step in 23463.45 ms, using 3.53 GB
  2 401:  40%|██████████████████████████████████████████████████████████████████████████████████▊                                                                                                                            | 2/5 [01:43<02:20, 46.80s/it]step in 310.02 ms, using 3.53 GBstep in 80143.80 ms, using 1.89 GB
  1 201:  60%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏                                                                                  | 3/5 [01:43<00:51, 25.57s/it]step in 23463.45 ms, using 3.53 GBstep in 305.68 ms, using 3.53 GBstep in 310.02 ms, using 3.53 GB
  0   1:  80%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▌                                         | 4/5 [01:44<00:15, 15.60s/it]step in 305.68 ms, using 3.53 GBstep in 305.37 ms, using 3.53 GBstep in 305.37 ms, using 3.53 GB
  0   1: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [01:44<00:00, 20.91s/it]
decode (1, 512, 64, 64)
decode (1, 512, 128, 128)
decode (1, 512, 256, 256)
decode (1, 256, 512, 512)
(512, 512, 3)
saving /var/folders/ch/2zp9hh591kb05qw6nm19693c0000gn/T/rendered.png
kafka@tubby:~/tinygrad$ 
```

PR #2132, opened 2023-10-23
Test #1751

imaolo:
Regression test for #1751 as requested in #2128. Passes on CI because we don't run cl on mac-os. test_copy_from_disk_unaligned() fails on M1/GPU.


geohot:
There has to be a smaller repro than GPT-2 medium

imaolo:
> There has to be a smaller repro than GPT-2 medium

I agree. Will keep digging.

PR #2136, opened 2023-10-23
conv2d: add an optional IGEMM (implicit GEMM) mode for conv2d

flammit:
Ideally this should be triggered when we know a tensor core will be available.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -511,7 +511,7 @@ class Tensor:
     padding = flatten((((k-1)*d-p,(k-1)*d-p+op) for k,d,p,op in reversed(list(zip(HW, make_pair(dilation, len(HW)), make_pair(padding, len(HW)), make_pair(output_padding, len(HW)))))))
     return x.conv2d(w.reshape(w.shape[0]*w.shape[1],*w.shape[2:]), groups=groups, bias=bias, dilation=dilation, padding=padding)
 
-  wino, igemm = int(getenv("WINO", "0")), int(getenv("IGEMM", "0"))
+  wino, igemm = getenv("WINO"), getenv("IGEMM")
   def conv2d(self, weight:Tensor, bias:Optional[Tensor]=None, groups=1, stride=1, dilation=1, padding=0) -> Tensor:
     (bs,cin_), (cout,cin), HW = self.shape[:2], weight.shape[:2], weight.shape[2:]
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -511,7 +511,7 @@ def conv_transpose2d(self, weight:Tensor, bias:Optional[Tensor]=None, groups=1,
     padding = flatten((((k-1)*d-p,(k-1)*d-p+op) for k,d,p,op in reversed(list(zip(HW, make_pair(dilation, len(HW)), make_pair(padding, len(HW)), make_pair(output_padding, len(HW)))))))
     return x.conv2d(w.reshape(w.shape[0]*w.shape[1],*w.shape[2:]), groups=groups, bias=bias, dilation=dilation, padding=padding)
 
-  wino = int(getenv("WINO", "0"))
+  wino, igemm = int(getenv("WINO", "0")), int(getenv("IGEMM", "0"))
```
chenyuxyz, line 514: can be `wino, igemm = getenv("WINO"), getenv("IGEMM")` because `getenv` gets the type from default value (int 0)

also probably need a test that covers this flag

flammit: Good call on the test - will add it.

geohot:
What is igemm and what does it do differently?

flammit:
Here's a description from CUTLASS: <https://github.com/NVIDIA/cutlass/blob/main/media/docs/implicit_gemm_convolution.md>.

We're not doing the shared memory parts, but basically this mostly changes the triple reduce axes to a single reduce where (CIN,H,W) are all combined, at which point the tensor cores can take over since it is a pure GEMM.

This makes all sized conv2d faster (except in the case where WINO outperforms at 3x3).

geohot:
Are you sure those two lines are different? If you reshape them all to one axis, without a contiguous tinygrad won't do anything.

flammit:
I believe it does because with a single axis there's no multiple reduce loops in the kernel.  Also when it is a single reduce axis, it can trigger the tensor cores when available.  Here's some data from `./test/models/test_real_world.py`:

Screenshot 2023-10-23 at 6 25 21 PM


Actually that said, if collapsing the reduces is doable as an Opt, I think that's way better than doing it here in `tensor.py` :/

geohot:
It should totally be doable as an Opt, just reshape in the linearizer

flammit:
Agreed - closing for now!

PR #2137, opened 2023-10-23
search: add tensor core and hand-coded opts to beam search

flammit:
This PR depends on #2097.

It also simplifies the search action space to only have GROUP (which implies top) and LOCAL (which implies last).  There's no measurable difference for the matvec ops to use GROUP vs. GROUPTOP and the tensor core structure is now written to only use LASTLOCAL.

PR #2138, opened 2023-10-23
unify GROUP and GROUPTOP

nimlgen:
I do not see why these 2 are different. I think could be unified.


flammit:
@nimlgen nice!  I did the same on my #2137 PR, but keeping Top=True as the default behavior is slightly better performance-wise in my tests.

nimlgen:
Oh, funny)

PR #2149, opened 2023-10-25
optimizer: simplify GROUP and LOCAL to have one of each

flammit:
Now that tensor cores only use LASTLOCAL, we can simplify to use only that op everywhere.

The only use of GROUP is in matvec hand-coded opts and it doesn't make a performance difference so switching to use only the top behavior.

Also adds additional asserts to prevent tensor core dims from being altered which causes bad kernels to be generated.

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -368,14 +368,14 @@ class TestLinearizerOpts(unittest.TestCase):
       [Opt(OptOps.LOCAL, 0, 2)],
       [Opt(OptOps.LOCAL, 0, 8)],
       [Opt(OptOps.LOCAL, 0, 16)], # Checking how it works with locals
-      [Opt(OptOps.GROUPTOP, 0, 2)],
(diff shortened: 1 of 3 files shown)


geohot:
Have you confirmed all behavior is identical across the test set? What about TC=2? What about the GROUP that's not a GROUPTOP in the MV stuff?

flammit:
Did a diff and manual inspection on the `DEBUG=4` output of `test/test_linearizer.py` from master and this branch, everything looks the same up to exchanges in lidx indices which are safe.

TC=2 is now part of the CI Metal tests, but confirmed it locally on all `test/*.py` not just `test_ops.py`.

For the matvec change, it doesn't seem to make a measurable difference (first is master, second is just changing the matvec GROUP to GROUPTOP):

Screenshot 2023-10-25 at 3 15 35 PM

Changed to keep the GROUPTOP opt in the latest commit.

PR #2151, opened 2023-10-25
Symbolic refactoring ideas half of them are math related 

ZenginU:
Around 10 line simplification, can be increased but with cost of performance


geohot:
Can you add tests for these simplifications and confirm behavior is the same?

ZenginU:
> Can you add tests for these simplifications and confirm behavior is the same?

I have not added new simplification rules so current tests should cover it. But I realized that even though I made wrong refactoring tests did not catch it

ZenginU:
Closing as I realized that math is not completely true but tests are passing needs more careful thinking

PR #2154, opened 2023-10-26
beam=12 makes gpt2 gpu-time < 5ms on 3090

nimlgen:
Made slight modifications to the search space, so Beam=12 gets kernels to achieve ~3ms execution time for gpt2. 

In theory, with CudaGraph we should get <5ms on 13900k+3090 (don't have any machine to check this). On the og tiny, I managed to achieve sub 5ms as well applying more CPU-related optimizations (caching sym_infer + asserts in jit.py). However, I'm sure about the cleanliness of these changes, so I focused on achieving the sub 5ms on the 13900k.

I would greatly appreciate it if you could run a test on the 13900k.

### DIFF
--- a/examples/gpt2.py
+++ b/examples/gpt2.py
@@ -50,7 +50,7 @@ class Attention:
       keys, values = cache_k.cat(xk, dim=1), cache_v.cat(xv, dim=1)
 
     # save the cache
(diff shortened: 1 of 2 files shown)


geohot:
Cool, will test

Issue #1636, opened 2023-08-23
Bounty: Training Retinanet


waifairer:
Add the training logic to examples/mlperf/model_train.py -> def train_rnnt():

[Links](https://github.com/mlcommons/training/blob/master/rnn_speech_recognition/pytorch/utils/librispeech.csv) to training data used in mlperf

MLPerf training rules [here](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#appendix-allowed-optimizers)

---

Current Bounty: $500
To lock the bounty submit a draft PR with a decent amount of progress made
Make sure to reference this issue in the PR for future tracking

Notice: If the PR goes stale the bounty will be unlocked

waifairer:
Previous attempts:
https://github.com/tinygrad/tinygrad/pull/902

7vansh7:
is it still available ?

PR #2155, opened 2023-10-26
[WIP] Retinanet Training II

mszsorondo:
#1636

next (high level) steps: 

- [x] L1 loss
- [x] focal loss for anchor classifications
- [x] L1 loss for box regressions
- [x] clean up code
- [x] adapt for mlperf-retinanet training specs
- [x] refactor for exploiting tensor lazyness and tensor mul + JIT
- [x] extra spec details (e.g. img flipping, random sampling, etc)
- [ ] Final refactor

### DIFF
--- a/examples/mlperf/model_eval.py
+++ b/examples/mlperf/model_eval.py
@@ -97,7 +97,7 @@ def eval_retinanet():
   from tinygrad.jit import TinyJit
   mdlrun = TinyJit(lambda x: mdl(input_fixup(x)).realize())
 
-  n, bs = 0, 8
+  n, bs = 0, 2
   st = time.perf_counter()
   for x, targets in iterate(coco, bs):
     dat = Tensor(x.astype(np.float32))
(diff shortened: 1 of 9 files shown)


mszsorondo:
@geohot I think this is almost done, my laptop doesn't have much resources to run big enough training samples to check (hw, if available, is 3x more expensive in argentina). It already takes me too long with resized images of 200x200. Can you SSH me to some machine? not necesarily the AMDs yet. Just something decent. I've been checking our implementation with mlperf's by calling for their constructors from tinygrad instead of running the actual mlperf's official training scripts. I am probably skipping some details I'd like to see by watching the step-by-step of their scripts with the original batch and image sizes.

geohot:
Do you have an estimated time to completion? How much will you wager it's correct?

mszsorondo:
> Do you have an estimated time to completion? How much will you wager it's correct?

I'm 100% confident this reproduces what the official [**written** spec](https://github.com/mlcommons/training/tree/master/single_stage_detector#datasetenvironment) states.

But there may be some missing details that cannot be seen unless you run the actual code from their reference code. If there is something missing, it should not be difficult to solve. I'll push one more commit that sets things up for running our implementation with the full training set and original batch & image sizes. This is why I don't want to ask for the AMDs now.

3 day worst-case time to completion once I have access to a computer that lets me reproduce a couple iterations from the training loop for both implementations. I'll try to get one elsewhere if this isn't possible or generates some trouble.

gijskoning:
If you want to be more confident of the runtime of your code on a bigger machine I recommend vast.ai. Its very cheap (0.24 cents/hour for rtx 3090) and easy to setup. They only list nvidia gpus but that should be enough to test your code.

geohot:
Unlocking bounty, over 1 week has gone by without update.

geohot:
New rule on mlperf training bounties. in order to lock the bounty, you must post weights that you trained in tinygrad that meet the accuracy requirement. Then we can look at speed and mergablity, but until there's weights, no lock.

PR #2161, opened 2023-10-27
Fix cuda half2 cast

Qazalin:
`test/test_dtype.py TestHalfDtype::test_half_upcast_ops` fails on a real CUDA (I used a v100 from google colab, but i think this is universally true). The error is: 

```
kernel.cu(12): error: no suitable user-defined conversion from "half2" to "float2" exists
```

because in this AST `Tensor([[1,2],[3,2]], dtype=dtypes.half)@Tensor.eye(2, dtype=dtypes.float32`
the tensor has elements of len 2 the dtype defaults to half2. I couldn't find a one-liner flag to disable this optimization (i suspect there's some pycuda thing for it when we create the CUDA buffer in the first place?)

This one fixes it at the renderer level by using `__half2float` and destructing the half2 to half

Kernel diff for the ast above:

```diff
__global__ void r_2_2_2(float* data0, const half* data1) {
  int lidx0 = threadIdx.x; /* 2 */
  float2 acc0 = make_float2(0.0f,0.0f);
- float2 val0 = (float2)(*((half2*)(data1+lidx0*2)));
+ float2 val0 = make_float2(__half2float(((half2*)(data1+lidx0*2))->x), __half2float(((half2*)(data1+lidx0*2))->y));;
  (acc0).x = (((val0).x*1.0f)+(acc0).x);
  (acc0).y = (((val0).x*0.0f)+(acc0).y);
  (acc0).x = (((val0).y*0.0f)+(acc0).x);
  (acc0).y = (((val0).y*1.0f)+(acc0).y);
  *((float2*)(data0+lidx0*2)) = (float2)make_float2((acc0).x,(acc0).y);
}
```

### DIFF
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -42,14 +42,14 @@ class CStyleLanguage(NamedTuple):
   }
 
   # returns a str expression of the casted xs with the given type
-  def render_cast(self, x:List[str], var_dtype:DType) -> str:
-    if len(x) == 1: return f"({var_dtype.name})({x[0]})"
-    assert len(x) == var_dtype.sz, f"cast is wrong size {len(x)} != {var_dtype.sz}"
+  def render_cast(self, x:List[str], output_dtype:DType, buf_dtype:DType|None=None) -> str:
+    if len(x) == 1: return f"({output_dtype.name})({x[0]})"
+    assert len(x) == output_dtype.sz, f"cast is wrong size {len(x)} != {output_dtype.sz}"
     assert self.float4 is not None, "cast is not supported on this platform"
-    if var_dtype == dtypes._float4: return f"{self.float4}({','.join(x)})"
-    if var_dtype == dtypes._float2: return f"{self.float4.replace('float4', 'float2')}({','.join(x)})"
-    if var_dtype == dtypes._int2: return f"{self.float4.replace('float4', 'int2')}({','.join(x)})"
-    raise NotImplementedError(f"no cast for {var_dtype}")
+    if output_dtype == dtypes._float4: return f"{self.float4}({','.join(x)})"
+    if output_dtype == dtypes._float2: return f"{self.float4.replace('float4', 'float2')}({','.join(x)})"
+    if output_dtype == dtypes._int2: return f"{self.float4.replace('float4', 'int2')}({','.join(x)})"
+    raise NotImplementedError(f"no cast for {output_dtype}")
 
   # returns a str expression of the const with the given type
   def render_const(self, x:Union[float,int], var_dtype) -> str:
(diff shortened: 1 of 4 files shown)


tinygrad/renderer/cuda.py
```diff
@@ -0,0 +1,28 @@
...
+from tinygrad.renderer.cstyle import uops_to_cstyle, CStyleLanguage
+from tinygrad.helpers import dtypes
+
+class CUDALanguage(CStyleLanguage):
+  kernel_prefix = "__global__ "
+  smem_prefix = "__shared__ "
+  smem_prefix_for_cast=False
+  arg_int_prefix = "const int"
+  barrier = "__syncthreads();"
+  float4 = "make_float4"
```
geohot, line 11: If you want to refactor, you can't have make_float2 and make_float4 hit different paths.

tinygrad/renderer/cstyle.py
```diff
@@ -42,7 +42,7 @@ class CStyleLanguage(NamedTuple):
   }
 
   # returns a str expression of the casted xs with the given type
-  def render_cast(self, x:List[str], var_dtype:DType) -> str:
+  def render_cast(self, x:List[str], var_dtype:DType, buf_dtype:Union[DType, None]=None) -> str:
```
geohot, line 45: What is buf_dtype and why do we need it?

tinygrad/renderer/cuda.py
Qazalin: done

tinygrad/renderer/cstyle.py
Qazalin: it's supposed to be the original dtype of the value that's being casted. half2float is a special fn can only cast between those two dtypes, so we end up not only needing the input but also the output to check

Qazalin: but the function body is still a wip.

in an ast that casts half2->float2 the args are `var_dtype=float2, buf_dtype=half`, buf_dtype should be a half2 instead, i'm looking to see if something like float2 could be implemented here too

Qazalin:
uh ok this is sorta painful - I found out CI could have actually caught this, but the way the assertion in `test_dtype` is setup makes it so that it gets ignored.

I suspect it's because of this error
https://github.com/tinygrad/tinygrad/actions/runs/6678878955/job/18150391985?pr=2161#step:16:54

could we do an `if CI` here to not throw this error somehow?
Because the exception not only passes that expected error but also actual kernel compilation problems. The full err is:

```
pycuda.driver.CompileError: nvcc compilation of /tmp/tmpf1wmci4_/kernel.cu failed
[command: nvcc --ptx -Wno-deprecated-gpu-targets -arch sm_35 -I/usr/local/lib/python3.10/dist-packages/pycuda/cuda kernel.cu]
[stderr:
kernel.cu(4): error: name must be a namespace name

kernel.cu(14): error: no suitable user-defined conversion from "half2" to "float2" exists
```

PR #2162, opened 2023-10-27
optimizer: simplify GROUP and LOCAL to have one of each

flammit:
Now that tensor cores only use LASTLOCAL, we can simplify to use only that op everywhere.

The only use of GROUP is in matvec hand-coded opts and it doesn't make a performance difference so switching to use only the top behavior.

Also adds additional asserts to prevent tensor core dims from being altered which causes bad kernels to be generated.

### DIFF
--- a/tinygrad/features/search.py
+++ b/tinygrad/features/search.py
@@ -13,7 +13,10 @@ actions += flatten([[Opt(op=OptOps.LOCAL, axis=axis, amt=amt) for amt in [2,3,4,
 actions += flatten([[Opt(op=OptOps.GROUPTOP, axis=axis, amt=amt) for amt in [13,16,29,32,256]] for axis in range(3)])
 actions += [
   Opt(op=OptOps.LOCAL, axis=0, amt=32),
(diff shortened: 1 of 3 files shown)


flammit:
Sorry I can't remember why I closed this...

tinygrad/features/search.py
```diff
@@ -13,7 +13,10 @@
 actions += flatten([[Opt(op=OptOps.GROUPTOP, axis=axis, amt=amt) for amt in [13,16,29,32,256]] for axis in range(3)])
 actions += [
   Opt(op=OptOps.LOCAL, axis=0, amt=32),
-  Opt(op=OptOps.GROUP, axis=0, amt=4), Opt(op=OptOps.GROUP, axis=0, amt=8), Opt(op=OptOps.GROUP, axis=1, amt=8),
+  Opt(op=OptOps.GROUPTOP, axis=0, amt=4), Opt(op=OptOps.GROUPTOP, axis=0, amt=8), Opt(op=OptOps.GROUPTOP, axis=1, amt=8),
+  Opt(op=OptOps.GROUPTOP, axis=0, amt=16), Opt(op=OptOps.GROUPTOP, axis=0, amt=256),
```
geohot, line 17: You shouldn't need these alt GROUPTOPs, see line 13 after merging nimlgen's GPT-2 speed thing

PR #2156, opened 2023-10-26
Higher test coverage for dtypes

Qazalin:
- GPU not-skipped test cases: 30 -> 60
- non-bool dtype -> bool test coverage: 0 -> 100% (avoiding bugs like #2164)
- Number of dtypes pairs tested in each case: (manually defined, 1-4) -> every possible combination
- Tested inputs: (only supported a simple [1,2,3,4] array) -> Fuzzed random sample of 10

Found broken cases in clang, llvm and triton - clang and LLVM are fixed here but triton needs more work

PR #2164, opened 2023-10-27
Fix llvm int->bool cast

Qazalin:
This cast in master:
```py
Tensor([1,2,3,4], dtype=dtypes.int32, device="LLVM").cast(dtypes.bool) # or any other int
```
returns:
```
[ True False  True False]
```

this PR fixes it by using a compare instead of a cast.

The test case added here only supports the new line in llvmir.
I'm writing a more comprehensive test suite for casting in #2156. Ran these changes through every casting combination in that PR as well


dc-dc-dc:
should zero also be included in the test to make sure it casts to false?

Qazalin:
@dc-dc-dc  I hardcoded one tbh i'd kinda wanna fuzz these stuff - testing casting would be much more robust if it's broken down into:
1. testing every dtype on every other dtype and reference-checking it against the same cast done in numpy
2. fuzzing the tensor inputs instead of hardcoding

I do p1 in #2156 but haven't looked into p2 yet

PR #2166, opened 2023-10-28
early_exec with Pool

chaosagent:
this is less lines, is there a reason for the old form?

this can be used for parallel compiler.


geohot:
As long the fork happens early it should be fine, but really we have to deal with all this better and this shouldn't be needed. Not going to merge, it's in extra and the line savings aren't worth the potential bugs.

Issue #1377, opened 2023-07-31
Tracking Issue: windows+clang in CI


wozeparrot:
Since windows+clang CI was added in #1011, it probably was never actually working correctly, but it was unnoticed due to also having a ubuntu+clang CI, but with the merge of #1252, ubuntu+clang CI was removed and only the broken windows+clang CI remained. This was first noticed in #1366 and temporarily patched around by merging #1368, which brought ubuntu+clang CI back.

Now there has been some work in #1375 which brings a fixed windows+clang CI back to replace the ubuntu+clang CI, but this has found that 26 tests are failing on windows+clang, mainly dtype related ones.

Opening this issue to track progress. You can pull the first commit of #1375 to test in CI.

geohot:
Windows is not a supported platform.

PR #2168, opened 2023-10-28
Windows HIP support

andrewjc:
The paths in hip_wrapper are hard coded, this PR adds dynamic path resolution for locating either the shared object files on unix like systems, or the dll files on windows.

On windows, the hiprtc file is suffixed with the version number, eg hiprtc5050.dll, so this uses a glob pattern to sort the collection, returning the last item in the collection. On windows the amdhip64 file is usually located in system32, but rather than hard coding that path, we search the path instead.

It also fixes a bug where the cache file is locked and prevents unlinking the database file before creating a new one.


wozeparrot:
https://github.com/tinygrad/tinygrad/issues/1377#issuecomment-1658900746

andrewjc:
windows not supported. will keep the branch for hacking.

PR #2170, opened 2023-10-28
fix repeat_kv for JITTED llama models using grouped query attention. 

r3sist-uniq:
Since slicing isn't supported for symbolic shapes, the models with specified `n_kv_heads` gave an error while running with JIT (found the error while writing the training script for tinyllama). 

Also added the config for [Tinyllama](https://github.com/jzhang38/TinyLlama) model.  You can test TinyLLAMA by downloading one of their model.safetensors [checkpoints](https://huggingface.co/PY007/TinyLlama-1.1B-intermediate-step-240k-503b/tree/main) (the tokenizer is the same as the official llama2) and running the following command

`python3 ./examples/llama.py  --prompt="where do you live?" --size="1B" --gen=tiny --model="./weights/model.safetensors"`

### DIFF
--- a/examples/llama.py
+++ b/examples/llama.py
@@ -192,8 +192,8 @@ MODEL_PARAMS = {
     #TinyLLaMA https://github.com/jzhang38/TinyLlama
     "1B": {
       "args": {"dim": 2048, "multiple_of": 256, "n_heads": 32, "n_kv_heads": 4, "n_layers": 22, "norm_eps": 1e-05, "vocab_size": 32000},
-      "files": 1, 
-    }, 
+      "files": 1,
+    },
     "7B": {
       "args": {"dim": 4096, "multiple_of": 256, "n_heads": 32, "n_layers": 32, "norm_eps": 1e-05, "vocab_size": 32000},
       "files": 1,
--- a/examples/llama.py
+++ b/examples/llama.py
@@ -189,11 +189,6 @@ MODEL_PARAMS = {
     },
   },
   "2": {
-    #TinyLLaMA https://github.com/jzhang38/TinyLlama
-    "1B": {
-      "args": {"dim": 2048, "multiple_of": 256, "n_heads": 32, "n_kv_heads": 4, "n_layers": 22, "norm_eps": 1e-05, "vocab_size": 32000},
-      "files": 1,
-    },
     "7B": {
       "args": {"dim": 4096, "multiple_of": 256, "n_heads": 32, "n_layers": 32, "norm_eps": 1e-05, "vocab_size": 32000},
       "files": 1,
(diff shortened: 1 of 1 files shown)


examples/llama.py
```diff
@@ -189,6 +189,11 @@ def __call__(self, tokens:Tensor, start_pos:int, temperature:Optional[float]=Non
     },
   },
   "2": {
+    #TinyLLaMA https://github.com/jzhang38/TinyLlama
```
chenyuxyz, line 192: cool! i think it needs to have its own entry and not under "2" since it's not official llama 2

geohot:
Does tinyllama work? Can you give me instructions to test it?

r3sist-uniq:
> Does tinyllama work? Can you give me instructions to test it?

added instructions above

geohot:
Are you sure this works?

```
kafka@tubby:~/fun/tinygrad$ python3 ./examples/llama.py --prompt="where do you live?" --gen=tiny --model=weights/LLaMA-2/tinyllama.safetensors --size=1B --temperature 0
using METAL backend
using LLaMA-tiny-1B model
ram used:  4.40 GB, freqs_cis                                         : 100%|███████████████████████████████████████████████████████████████████████████████| 202/202 [00:00<00:00, 298.40it/s]
loaded weights in 683.66 ms, 4.40 GB loaded at 6.44 GB/s
where do you live?
The 2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000^CTraceback (most recent call last):
```

r3sist-uniq:
> Are you sure this works?
> 
> ```
> kafka@tubby:~/fun/tinygrad$ python3 ./examples/llama.py --prompt="where do you live?" --gen=tiny --model=weights/LLaMA-2/tinyllama.safetensors --size=1B --temperature 0
> using METAL backend
> using LLaMA-tiny-1B model
> ram used:  4.40 GB, freqs_cis                                         : 100%|███████████████████████████████████████████████████████████████████████████████| 202/202 [00:00<00:00, 298.40it/s]
> loaded weights in 683.66 ms, 4.40 GB loaded at 6.44 GB/s
> where do you live?
> The 2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000^CTraceback (most recent call last):
> ```

Yes it seems to give this weird output with temp=0. Try the default temp or increase it.

chenyuxyz:
i have tested the `repeat_kv` fix separately in #2181 and it works with other models, so the issue is either in the model itself, or the way we load safetensor

r3sist-uniq:
> i have tested the `repeat_kv` fix separately in #2181 and it works with other models, so the issue is either in the model itself, or the way we load safetensor

I'm still debugging it. The model isn't the problem, it gives much better outputs with using huggingface. So it may either be loading the weights or some post processing stuff. 

I can close this PR, will create a separate one for fixing up tinyllama along with the training.

PR #2173, opened 2023-10-28
optimizer: revert changes to UNROLL insert location

flammit:


geohot:
Thanks for the PR! Turns out the openpilot thing was related to a different change with the dont_use_locals flag, an easy no trade off fix.

We should figure out what we actually want here and test it. I don't like that the axis order is canonical, though the flipped order of the purples shouldn't matter either way.

PR #2178, opened 2023-10-29
add step abstract method to Optimizer

carrascomj:
Since `step` is the end-user API of an `Optimizer`, maybe it should be implemented at the top level abstract class.

The main reason is that I find this diagnostic a bit awkward:

image

Very small PR, feel free to close it if this kind of changes go against the code style or something.

### DIFF
--- a/tinygrad/nn/optim.py
+++ b/tinygrad/nn/optim.py
@@ -1,9 +1,10 @@
 # sorted in order of increasing complexity
+from abc import ABCMeta, abstractmethod
 from typing import List
 from tinygrad.helpers import dedup
 from tinygrad.tensor import Tensor
 
-class Optimizer:
+class Optimizer(metaclass=ABCMeta):
   def __init__(self, params: List[Tensor], lr: float):
     # if it's None, but being put into an optimizer, set it to True
     for x in params:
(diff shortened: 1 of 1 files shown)


geohot:
What is ABC and why do I want it?

carrascomj:
`ABC` is a marker from the builtin `abc` that makes a class an _abstract_ class. The implication is that if a child is declared without the methods marked as `abstractmethod`, it will raise. That is, given

```python
import abc

class AbstractParent(metaclass=abc.ABCMeta):
    @abc.abstractmethod
    def method(self): pass

class GoodChild(AbstractParent):
    def method(self):
        print("Hello")

class BadChild(AbstractParent):
    def other_method(self):
        print("Hello")
```

instantiating `GoodChild()` will work just fine but `BadChild()` will raise an error:

```
  File "/home/georg/an_example.py", line 29, in <module>
    BadChild()
TypeError: Can't instantiate abstract class BadChild with abstract method method
```

If `ABCMeta` is not added as the metaclass (or `ABC` [which is a shorthand](https://docs.python.org/3/library/abc.html#abc.ABC)), then the `@abstractmethod` has no effect, not raising an error if it is not implemented by the child. The `OpNode` implementation has this _issue_ (fine if you just want to mark it without enforcing it), by the way.

geohot:
This doesn't seem worth the lines, it's internal only.

PR #2179, opened 2023-10-29
Enable Multi-Output Export

kashyapakshay:
Support for models that typically return multiple tensors (eg, regression + confidences in SSD)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -237,8 +237,8 @@ jobs:
       run: METAL=1 python -m pytest -n=auto test/test_linearizer.py
     - name: Test tensor core reshape-only ops
       run: METAL=1 TC=2 python -m pytest -n=auto test/test_ops.py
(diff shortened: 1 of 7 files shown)


geohot:
Can you get the e2e webgpu tests working again? With a test for that, I think this is good to merge

kashyapakshay:
@geohot I assume that's the webgpu pytest that was disabled in #1471 ? Or is that something else?

geohot:
Nah, idk if that can work. Just something fully end to end that tests the multioutput

kashyapakshay:
Modified `examples/index.html` to account for a list of outputs now and re-enabled `test_webgpu.js` to test running the webgpu efficientnet. I'm happy to write another example compile script for a model with multiple outputs but wanted to avoid boilerplate code.

geohot:
LGTM, nice to have this tested again. I think that coverage is fine (confirms no efficientnet regression)

PR #2180, opened 2023-10-30
separate compilers and runtimes

chaosagent:
just small ideas in here so far (most of the code here is dead code, just trying things)

on amd, turns out both hip and pyopencl produce hsaco binaries that you can pass between each other. similar can be done with nvidia / ptx. is it possible to get/set sass from opencl?

abstraction goals (not code to include in this pr):
- support parallel comp
- support amd triton
- support persistent ast -> binary cache
- support het. devices
- make testing easier (eg make cstyle params queryable, probably unnecessary)


chaosagent:
#2203 did this

PR #2183, opened 2023-10-30
Add type hint for 'rawbuffers' 

amirbouraoui:
Added 'rawbuffers' parameter in Compiled.exec_ast for improved clarity and type safety.


geohot:
Tests failing, and I just don't think this is right at all

Issue #1949, opened 2023-09-30
Training this model has issue compiling


geohot:
```
class TinyNet:
  def __init__(self):
    self.c1 = Conv2d(256,64,3)
    self.c2 = Conv2d(64,4,3)
    self.l1 = Linear(704,3)
    self.l2 = Linear(704,1)

  def __call__(self, x):
    x = self.c1(x).gelu()
    x = self.c2(x).gelu()
    x = x.reshape(x.shape[0], -1)
    x1 = self.l1(x)
    x2 = self.l2(x)
    return x1.log_softmax(), x2.sigmoid()
```

error:
```
AssertionError: Error Domain=MTLLibraryErrorDomain Code=3 "program_source:22:21: error: use of undeclared identifier 'val1'
  *(data0+gidx0) = (val1*(1.0f-val1)*acc1);
                    ^
program_source:22:32: error: use of undeclared identifier 'val1'
  *(data0+gidx0) = (val1*(1.0f-val1)*acc1);
                               ^
" UserInfo={NSLocalizedDescription=program_source:22:21: error: use of undeclared identifier 'val1'
  *(data0+gidx0) = (val1*(1.0f-val1)*acc1);
                    ^
program_source:22:32: error: use of undeclared identifier 'val1'
  *(data0+gidx0) = (val1*(1.0f-val1)*acc1);
                               ^
}
```

MutugiD:
I would love to work on this

geohot:
Works now!

PR #2188, opened 2023-10-31
fuzz linearizer transformation

chenyuxyz:
ran the fuzzer and now I have more questions.
on 3rd ast it found some issues already
also there seem to be no-op action in the action space
and linearizer is not copyable once linearized

a good run looks like this, one random path until it repeats or no action is available
image

will clean up more once i figure out the noop action and properly allocate buffer for image input and variable shape input


geohot:
Cool! Yea I think there's a ton of potential for improvement here

chenyuxyz:
update

- ImageDType needs more work
- variable shape can run, but the apply opt breaks with UNROLL and UPCAST
- CPU and TORCH are mostly correct, there's one AS_STRIDED bug and it also breaks if src contains different dtypes (we have float and half, and float and int)
- Compiled backend had kernels with errors similar to #1949, the variable was created in for loop and later used outside the for loop
- METAL takes a long time to compile some kernels, and breaks for some that exceed thread memory limit
- Some numerical issue when we divide by a number close to 0, or compared two numbers that are both close to 0 but the ratio is big

will finish ImageDType next and add some failed example in the test case, then clean up

chenyuxyz:
ready for review. i have tested on CPU, TORCH, CLANG, LLVM, METAL, GPU with various different pass rates

CPU and TORCH have a bug with AS_STRIDED, otherwise all good

Compiled backend have issue that a variable might be inited inside a for loop and used outside the loop
LLVM breaks a lot but seems to be some arg issue
CLANG has issue with `OptOps.NOLOCALS`
METAL has thread memory issue and `OptOps.NOLOCALS`, similar for GPU

Variable can run but most fix require good support for variable shape optimization
ImageTypes need some fundamental fix similar to `fix_schedule_for_images` so it's not done here

chenyuxyz:
added some failed examples to the test

PR #2191, opened 2023-10-31
Update test_example multidevice tests

Qazalin:
I think multi-device testing gets handled in test.yml. I made sure the test still runs the test on all devices in CI.

Making this change after breaking some part of the LLVM backend and seeing gpu, clang and cuda jobs all fail because they were trying to run stuff on LLVM https://github.com/tinygrad/tinygrad/actions/runs/6631863503/job/18016313825?pr=2046#step:13:185 - LLVM's job should've been the only one failing


geohot:
This is used in the pre-commit hooks mostly, has found several issues before I push

Qazalin:
weird i dont see it anymore https://github.com/tinygrad/tinygrad/blob/master/.pre-commit-config.yaml

Qazalin:
ok yeah that makes sense - i replicated those runs in precommit with a bash script, feels like an overkill though.

geohot:
Yea, I just like the test as it was. What's the issue with it?

Qazalin:
the issue with it is that if one backend out of this matrix was broken all jobs failed - https://github.com/tinygrad/tinygrad/blob/8c07c73a9b672c569ee335f05de5304e9498562f/.github/workflows/test.yml#L258

moving it to extra/ fixes it yeah - updated

PR #2196, opened 2023-11-01
no locals as opt action

geohot:


tinygrad/codegen/optimizer.py
```diff
@@ -270,6 +276,9 @@ def apply_opt(self, opt:Opt):
       assert amt == 4, "don't upcast mid anything but 4"
       self.shift_to(axis, amt, insert_before=self.first_reduce + len(self.group_for_reduce))
       self.group_for_reduce.append(amt)
+    elif opt.op == OptOps.NOLOCALS:
+      assert self.local_dims == 0 and len(self.group_for_reduce) == 0, "can't have no locals with locals"
+      self.dont_use_locals = True
```
chenyuxyz, line 281: also assert if `self.dont_use_locals` is already true so it's valid at most once

PR #2197, opened 2023-11-01
For cuda get current free space from device, and retry alloc failures

vish-pr:
Instead of calculating free space use what gpu device says.
In rare case even with available space alloc may fail, due to excessive shaded space, handle that by free 10% of space more.


tinygrad/runtime/ops_cuda.py
```diff
@@ -49,9 +49,16 @@ class device:
   import pycuda.autoprimaryctx # type: ignore # pylint: disable=unused-import # noqa: F401
   import pycuda.driver as cuda # type: ignore
   class CUDAAllocator(LRUAllocator):
-    def _do_alloc(self, size, dtype, device, **kwargs): return cuda.mem_alloc(size * dtype.itemsize) # type: ignore
+    def _do_alloc(self, size, dtype, device, **kwargs):
+      try: return cuda.mem_alloc(size * dtype.itemsize)
+      except cuda.MemoryError as e:
+        if len(self.aging_order[device]):
+          self.ensure_has_free_space(1.1*self._get_cur_free_spce(device), device) # increase free space by 10% and try again.
```
nimlgen, line 56: Maybe move this logic into LRU?

Something like?
```python
def _alloc_buffer(self, size, dtype, device, **kwargs):
    newbuf, alloc_size = None, size*dtype.itemsize
    while newbuf is None:
        self.ensure_has_free_space(alloc_size, device)
        try: newbuf = self._do_alloc(max(1, size), dtype, device, **kwargs)
        except Exception as e:
            if not len(self.aging_order[device]): raise e
            alloc_size *= 1.1
    self.free_space[device] -= alloc_size
    self.buffer_info[newbuf] = (size, dtype, alloc_size, device)
    return newbuf
```

nimlgen (commented):
Also would be great to have some tests.

geohot:
I want to delete the LRUAllocator and replace it with a schedule allocator

vish-pr:
@geohot Do you have any requirements for what  schedule allocator should be?
I think we can do better than LRU allocator. Thinking mostly in view of training phase. Given that alloc is a time taking op. And if that is proportional to amount of memory being alloc (not sure). We can see alloc pattern in first loop and optimise for it future loops. And maybe some kind of metric exporter over it which can be further used to tune param like batch size for fastest results.

tinygrad/runtime/ops_cuda.py
vish-pr: Different backend will raise different exceptions, either we handle any exception, which can be unrelated to memory, or we do an 'or' of all known memory error from each backend, which will result in backend related imports in lib.py or we introduce our memory exception and wrap backend exception in it handle that in lib.py. To avoid all that I have just handled it cuda file.

Alloc is costly, finding free space on gpu is not. I am not sure if unsuccessful alloc is fail fast. Also I have observed it diverges significantly for free_space calculation.

tinygrad/runtime/ops_cuda.py
```diff
@@ -49,9 +49,16 @@ class device:
...
+          return self._do_alloc(size, dtype, device, **kwargs)
+        raise e
     def _cached_bufkey(self, size, dtype, device): return (device, size*dtype.itemsize) # Buffers of the same length could be reused, no matter what dtype.
-  CUDAAlloc = CUDAAllocator(pycuda.driver.Context.get_device().total_memory())
+    def _get_cur_free_spce(self, deivice): return cuda.mem_get_info()[0]
```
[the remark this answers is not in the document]
vish-pr: added __init__ to call this.

geohot:
Oh, LRUAllocator is fine. It's CachedAllocator I want to replace. Does this fix a real issue @nimlgen?

tinygrad/runtime/ops_cuda.py
nimlgen: I don't think it belongs to cuda file, since it's related to all backends. I think it's fine to catch all exceptions.

nimlgen:
Yeah, we can get some exceptions while thinking we have free space (the way gpu manages memory or other processes use it). I think we could do retries in such cases trying to free more cached buffers.

tinygrad/runtime/ops_cuda.py
vish-pr: Done.

tinygrad/runtime/lib.py
```diff
@@ -38,9 +38,8 @@ def fromCPU(cls, x:np.ndarray, **kwargs):
 class RawBufferMapped(RawBufferCopyIn):
   def _buffer(self) -> memoryview: raise NotImplementedError("must be implemented")
   # NOTE: this metadata prevents the backing buffer from being freed. hack can be removed with PEP688
-  def buffer_view(self) -> np.ndarray: return np.frombuffer(self._buffer(), dtype=np.dtype(self.dtype.np, metadata={"backing": self}), count=self.size)  # type: ignore
```
nimlgen, line 41: Why did you revert this?

PR #2207, opened 2023-11-03
nlogn time arange lol

chaosagent:
just for fun


geohot:
This should be optimized in uops, should be possible to go O(n^2) -> O(n)

PR #2209, opened 2023-11-03
fix IMAGE=2 failed with NOOPT=1

chenyuxyz:
even with NOOPT, we need `required_optimizations` to set up shape for image dtypes

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -329,8 +329,7 @@ class Compiled:
           if DEBUG >= 1: print("  <  ".join(f"{nm:6s} : {lin.colored_shape(30, dense=True)} : {tm*1e6:8.2f} us" for nm, lin, tm in timed))
           k = timed[0][1]
       else:
-        pass
-        # k.required_optimizations()
+        k.required_optimizations()
       return self.to_program(k)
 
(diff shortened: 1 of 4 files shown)


tinygrad/ops.py
```diff
@@ -325,6 +325,9 @@ def get_program():
           timed = sorted([(nm, tk, time_linearizer(tk, test_rawbuffers, allow_test_size=False, disable_cache=True, clear_l2=True)) for nm, tk in lins], key=lambda x: x[2])
           if DEBUG >= 1: print("  <  ".join(f"{nm:6s} : {lin.colored_shape(30, dense=True)} : {tm*1e6:8.2f} us" for nm, lin, tm in timed))
           k = timed[0][1]
+      else:
+        pass
```
geohot, line 329: I don't think you want pass here

chenyuxyz: that was to demo that CI would catch this

chenyuxyz:
not entirely sure what's happening but the triton test segfault might be real, it also happened on my local fork. let me investigate

chenyuxyz:
yeah i think #2214  fixed it. ran it 10 times in my local fork and it all passed https://github.com/chenyuxyz/tinygrad/actions/workflows/test.yml

PR #2214, opened 2023-11-04
less broken limit_dims_to_max

chenyuxyz:
i think `limit_dims_to_max` is wrong, this PR makes it less wrong by ensuring the iterator don't go out of bound. the goal is to unblock debugging #2209 

happy to refactor the whole thing, there's really no test for the intended behavior though

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -1,10 +1,9 @@
 import numpy as np
 import unittest, os
 
-from tinygrad.codegen.kernel import tensor_cores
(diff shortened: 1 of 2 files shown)


g1y5x3:
So initial I wrote this because I noticed some of the cifar backward pass kernels back like a few months ago would generate super large global dimensions that exceeds the cuda device global dim limit. And it was used to just permute the axis to blockIdx.x which has the biggest size to bypass the issue. I agree there's definitely a much better solution to this.

PR #2216, opened 2023-11-05
Remove AS_STRIDED from shapetracker

forcefieldsovereign:
## Quick overview of how this works:

When a flat buffer is reshaped to (x, y), the strides will be (y, 1). This trick enables us to get the correct strides without using as_strided.

### Working through a simple example:  
Goal is to get shape (3,3,3) and strides (3, 1, 10).  
Start by rearranging so that strides are big to small: shape (3, 3, 3), strides (10, 3, 1). We'll permute back at the end.  
First, pad up to 3*10=30, since the size of the input buffer is only (3-1)*10+(3-1)*3+(3-1)*1+1 = 29.  
Now we can reshape to (3, 10). The strides will be (10, 1). The first dim of the tensor is already correct!  
Now, the next reshape will be (3, 3, 3) (here the last 3 is the stride we need for the middle dim), so we'll need to shrink down to (3, 9).  
Now we reshape (3, 3, 3), and the strides are (10, 3, 1). Add a permute and that's it.

### Handling the edge cases:  
The shapetracker simplifies views with merge_views, which can introduce overlapped views.
A common example of this is a view that has duplicate strides.  
For example, say there's an input buffer of size 5 and the view is shape (3, 3), strides (1, 1).  
Suppose the input buffer is (0, 1, 2, 3, 4).  
With as_strided, we would simply set the strides, and the matrix would be  
[0 1 2]  
[1 2 3]  
[2 3 4]  
Without using as_strided, there's no way to achieve overlapped strides like this directly. We have to expand and adjust the strides to achieve the same result.  
[0 1 2 3 4]  
[0 1 2 3 4]  
[0 1 2 3 4]  
Then, use the method to get shape (3, 3) strides (6, 1), where 6 is the size of the buffer plus the desired stride (1).


tinygrad/runtime/ops_disk.py
```diff
@@ -24,10 +24,10 @@ def __del__(self):
   def cast(self, arg:Tuple[DType, bool]): return RawDiskBuffer(self.size, arg[0], buf=self._buf, shape=self.shape, offset=self.offset)
   def reshape(self, arg): return RawDiskBuffer(self.size, self.dtype, buf=self._buf, shape=arg, offset=self.offset)
   def shrink(self, arg):
-    assert arg[1:] == tuple([(0,x) for x in self.shape[1:]]), f"can only slice the first dim of disk tensor {arg}"
-    offset = arg[0][0]*prod(self.shape[1:])*self.dtype.itemsize
-    size = (arg[0][1]-arg[0][0]) * prod(self.shape[1:])
-    return RawDiskBuffer(size, self.dtype, buf=self._buf, offset=self.offset+offset, shape=(arg[0][1]-arg[0][0],)+self.shape[1:])
+    if len(arg)>1: assert arg[1:] == tuple([(0,x) for x in self.shape[1:]]), f"can only slice the first dim of disk tensor {arg}"
```
roelofvandijk, line 27: assert len(arg)<2 or arg[1:] == tuple(..), ...

If you hide an assert behind an if statement, the if statement cannot be optimized away.

tinygrad/shape/shapetracker.py
```diff
@@ -99,12 +99,35 @@ def to_movement_ops(self) -> List[Tuple[MovementOps, Tuple]]:
...
-      real_offset = v.offset + (sum(x*st for (x,_),st in zip(v.mask, v.strides)) if v.mask else 0)
+      offset = v.offset + sum(v.strides[i] * (real_shape[i]-1) for i in range(len(v.strides)) if v.strides[i]<0)
+      real_offset = offset + (sum(x*st for (x,_),st in zip(v.mask, v.strides)) if v.mask else 0)
       # first, we apply the offset
       # then, we make it the correct shape
       # then, we apply permutations
       # TODO: don't use as_strided
-      to_apply.append((MovementOps.AS_STRIDED, ([s if st != 0 else 1 for s,st in zip(real_shape, v.strides)], v.strides, real_offset)))
+      # to_apply.append((MovementOps.AS_STRIDED, ([s if st != 0 else 1 for s,st in zip(real_shape, v.strides)], v.strides, real_offset)))
+      real_real_shape = [s for s,st in zip(real_shape, v.strides) if st!=0]
```
roelofvandijk, line 109: if st

tinygrad/shape/shapetracker.py
```diff
@@ -99,12 +99,35 @@ def to_movement_ops(self) -> List[Tuple[MovementOps, Tuple]]:
...
+      strides = [abs(st) for st in v.strides]
+      buffer_size = sum((s-1)*st for s, st in zip(real_shape,strides)) + 1
+      strides = [st for st in strides if st != 0]
```
roelofvandijk, line 112: if st

tinygrad/shape/shapetracker.py
```diff
@@ -99,12 +99,35 @@ def to_movement_ops(self) -> List[Tuple[MovementOps, Tuple]]:
...
+      new_shape_strides = sorted(zip(real_real_shape, strides), key=lambda k: k[1], reverse=True)
+      order = sorted(range(len(strides)), key=lambda k: strides[k], reverse=True)
+      to_apply.append((MovementOps.RESHAPE, -1))
+      to_apply.append((MovementOps.SHRINK, ((real_offset,real_offset+buffer_size),)))
+      if len(strides) > 0:
```
roelofvandijk, line 117: if strides:

tinygrad/shape/shapetracker.py
```diff
@@ -99,12 +99,35 @@ def to_movement_ops(self) -> List[Tuple[MovementOps, Tuple]]:
...
+        if (new_shape_strides[0][0] * new_shape_strides[0][1]) - buffer_size > 0:
+          to_apply.append((MovementOps.PAD, ((0, (new_shape_strides[0][0] * new_shape_strides[0][1]) - buffer_size),)))
+        for i in range(len(new_shape_strides)):
+          total_size = new_shape_strides[i][0] * new_shape_strides[i][1]
+          to_apply.append((MovementOps.SHRINK, (*((0, s[0]) for s in new_shape_strides[:i]), (0,total_size))))
+          to_apply.append((MovementOps.RESHAPE, (*[s[0] for s in new_shape_strides[:i+1]], new_shape_strides[i][1])))
+        to_apply.append((MovementOps.SHRINK, (*[(0, s[0]) for s in new_shape_strides], (0,1))))
+        to_apply.append((MovementOps.RESHAPE, (*[s[0] for s in new_shape_strides],)))
+        if not all(i == j for i, j in enumerate(order)):
```
roelofvandijk, line 126: Could be `order != range(len(order))` but maybe that is slower?

tinygrad/shape/shapetracker.py
```diff
@@ -99,12 +99,35 @@ def to_movement_ops(self) -> List[Tuple[MovementOps, Tuple]]:
...
+          to_apply.append((MovementOps.PERMUTE, [order.index(i) for i in range(len(strides))]))
+      to_apply.append((MovementOps.RESHAPE, [s if st!=0 else 1 for s,st in zip(real_shape, v.strides)]))
```
roelofvandijk, line 128: if st

forcefieldsovereign:
Thanks for review! Will finish up tomorrow.
TODO:
 - get it working with simplified views (merge_views introduces overlapped strides, eg strides of [1, 1])
 - possible to eliminate some ops if view is contiguous?
 - cleanup

tinygrad/shape/shapetracker.py
```diff
@@ -99,12 +99,35 @@ def to_movement_ops(self) -> List[Tuple[MovementOps, Tuple]]:
...
+      buffer_size = sum((s-1)*st for s, st in zip(real_shape,strides)) + 1
+      strides = [st for st in strides if st]
+      new_shape_strides = sorted(zip(real_real_shape, strides), key=lambda k: k[1], reverse=True)
+      order = sorted(range(len(strides)), key=lambda k: strides[k], reverse=True)
+      to_apply.append((MovementOps.RESHAPE, -1))
+      to_apply.append((MovementOps.SHRINK, ((real_offset,real_offset+buffer_size),)))
+      if strides:
+        if (new_shape_strides[0][0] * new_shape_strides[0][1]) - buffer_size > 0:
+          to_apply.append((MovementOps.PAD, ((0, (new_shape_strides[0][0] * new_shape_strides[0][1]) - buffer_size),)))
+        for i in range(len(new_shape_strides)):
```
roelofvandijk, line 120: I wonder whether an enumerate is nicer here.

geohot:
Amazing! Can you add a test to run through all the stored kernels (see `extra/optimization/get_action_space.py` for an example of the test) and confirm all the resulting shape trackers match?

geohot:
Bounty locked. Pending that test passing, this should be done.

chenyuxyz:
can you sync to master too? this might fix a failed linearizer test case and would need to update the test case

geohot:
Does this test pass? Want to run it in CI?

forcefieldsovereign:
Yeah the test passes I'll add it to CI. I'm working on an improvement that removes an extra padding view, but the shape trackers are all equivalent and mostly identical.

extra/to_movement_ops.py
```diff
@@ -0,0 +1,107 @@
...
+from tinygrad.ops import LazyOp, MovementOps, TernaryOps, BinaryOps, UnaryOps, ReduceOps, BufferOps, MemBuffer, ConstBuffer
+from tinygrad.helpers import dtypes, prod
+from tinygrad.shape.shapetracker import ShapeTracker
+from tinygrad.shape.view import View
+from tinygrad.shape.symbolic import Node, Variable
+inf, nan = float('inf'), float('nan')
+
+def get_buffer_size(shape, strides, offset, mask):
+  real_shape = tuple(y-x for x,y in mask) if mask else shape
+  offset = offset + sum(strides[i] * (real_shape[i]-1) for i in range(len(strides)) if strides[i]<0)
```
roelofvandijk, line 14: You could iterate over `zip(strides, real_shape)` instead.

extra/to_movement_ops.py
```diff
@@ -0,0 +1,107 @@
...
+  real_offset = offset + (sum(x*st for (x,_),st in zip(mask, strides)) if mask else 0)
+  real_real_shape = [s for s,st in zip(real_shape, strides) if st]
+  strides = [abs(st) for st in strides if st and not isinstance(st, Node)]
+  return real_offset + sum((s-1)*st for s, st in zip(real_real_shape,strides)) + 1
+
+def flatten_view(view: View):
+  real_shape = tuple(y-x for x,y in view.mask) if view.mask else view.shape
+  offset = view.offset + sum(view.strides[i] * (real_shape[i]-1) for i in range(len(view.strides)) if view.strides[i]<0)
```
roelofvandijk, line 22: `zip`. Come to think of it, the first lines of this and the preceding function are the same.

extra/to_movement_ops.py
```diff
@@ -0,0 +1,107 @@
...
+  real_offset = offset + (sum(x*st for (x,_),st in zip(view.mask, view.strides)) if view.mask else 0)
+  real_real_shape = [s for s,st in zip(real_shape, view.strides) if st]
+  strides = [abs(st) for st in view.strides if st and not isinstance(st, Node)]
+  def sort_by_strides(shape, strides): return sorted(zip(shape, strides), key=lambda k: (k[1],-k[0]), reverse=True), sorted(range(len(strides)), key=lambda k: (strides[k],-real_real_shape[k]), reverse=True)
+  ordered_shape_strides, _ = sort_by_strides(real_real_shape, strides)
+  ordered_shape_strides = [list(s) for s in ordered_shape_strides]
+  if strides:
+    i = 0
+    while i < len(ordered_shape_strides):
```
roelofvandijk, line 31: For loop?

extra/to_movement_ops.py
```diff
@@ -0,0 +1,107 @@
...
+      if i<len(ordered_shape_strides)-1 and ordered_shape_strides[i][1] == ordered_shape_strides[i+1][0]*ordered_shape_strides[i+1][1]:
+        ordered_shape_strides[i+1][0] = ordered_shape_strides[i][0]*ordered_shape_strides[i+1][0]
+      else: i += 1
+    flat_shape = [shape_stride[0] for shape_stride in ordered_shape_strides]
```
roelofvandijk, line 35: This can be done with `flat_shape, flat_strides = list(zip(*ordered_shape_strides))`, but maybe that's less readable.

extra/to_movement_ops.py
```diff
@@ -0,0 +1,107 @@
...
+    flat_strides = [shape_stride[1] for shape_stride in ordered_shape_strides]
+    return (flat_shape, flat_strides, real_offset)
+  return (real_shape, view.strides, real_offset)
+
+def views_equivalent(v1: View, v2: View) -> bool:
+  if v1 == v2: return True
```
roelofvandijk, line 41: return `v1 == v2 or flatten()...`

extra/to_movement_ops.py
```diff
@@ -0,0 +1,107 @@
...
+  return flatten_view(v1) == flatten_view(v2)
+
+
+def st_equivalent(st: ShapeTracker, st_rebuilt: ShapeTracker):
+  views = list(st.views)
+  rebuilt_views = list(st_rebuilt.views)
+  i = 0
+  while i < len(views):
```
roelofvandijk, line 49: For loop?

extra/to_movement_ops.py
```diff
@@ -0,0 +1,107 @@
...
+      i += 1
+    # hack to skip expands for overlapped strides
+    else:
+      rebuilt_views.pop(i)
+  return True
+
+def test_rebuild(st: ShapeTracker):
+  rebuilt_st = ShapeTracker.from_shape((get_buffer_size(st.views[0].shape, st.views[0].strides, st.views[0].offset, st.views[0].mask),))
+  for mop, arg in st.to_movement_ops():
+    match mop:
```
roelofvandijk, line 64: Can't use a match statement, that does not work in Python 3.8 which we still support.

extra/to_movement_ops.py
```diff
@@ -0,0 +1,107 @@
...
+        rebuilt_st = rebuilt_st.pad(arg)
+      case MovementOps.SHRINK:
+        rebuilt_st = rebuilt_st.shrink(arg)
+      case MovementOps.STRIDE:
+        rebuilt_st = rebuilt_st.stride(arg)
+      case _:
+        raise Exception("invalid mop")
+  rebuilt_st = rebuilt_st.simplify()
+  if len(st.views) != len(rebuilt_st.views):
+    if not set(st.views).issubset(set(rebuilt_st.views)):
```
roelofvandijk, line 87: `if a and b`
Also: `asserts behind if statements mean that the if statement will not be optimized away`

tinygrad/shape/shapetracker.py
```diff
@@ -97,14 +97,34 @@ def unbind(self) -> ShapeTracker: return ShapeTracker(tuple(v.unbind() for v in
...
-    for v in self.views:
+    for i, v in enumerate(self.views):
       real_shape = tuple(y-x for x,y in v.mask) if v.mask else v.shape
-      real_offset = v.offset + (sum(x*st for (x,_),st in zip(v.mask, v.strides)) if v.mask else 0)
-      # first, we apply the offset
-      # then, we make it the correct shape
-      # then, we apply permutations
-      # TODO: don't use as_strided
-      to_apply.append((MovementOps.AS_STRIDED, (tuple([s if st != 0 else 1 for s,st in zip(real_shape, v.strides)]), v.strides, real_offset)))
+      offset = v.offset + sum(v.strides[i] * (real_shape[i]-1) for i in range(len(v.strides)) if v.strides[i]<0)
```
roelofvandijk, line 102: zip, also: shares code with the functions above

tinygrad/shape/shapetracker.py
```diff
@@ -97,14 +97,34 @@ def unbind(self) -> ShapeTracker: return ShapeTracker(tuple(v.unbind() for v in
...
+      real_offset = offset + (sum(x*st for (x,_),st in zip(v.mask, v.strides)) if v.mask else 0)
+      real_real_shape = [s for s,st in zip(real_shape, v.strides) if st]
+      strides = [abs(st) for st in v.strides if st and not isinstance(st, Node)]
```
roelofvandijk, line 105: `if isintance(st, int)`

roelofvandijk (commented):
This looks great!

forcefieldsovereign:
@geohot this should be good to go

chenyuxyz:
i think it's good to merge, can you sync to master and fix the conflict?

chenyuxyz:
i also tested it with linearizer fuzzer and running llama on CPU/TORCH backend, all looks good. congrats!

pay bounty via PayPal or USDC. E-mail [george@tinygrad.org](mailto:george@tinygrad.org) (or DM on discord) with your info to claim!

PR #2217, opened 2023-11-05
Onnx 1.15.0 gogogo

geohotstan:
onnx 1.15.0 updates:
- onnx now uses serialized file (see onnx/onnx-ml.proto) as descriptors so no more pretty python cmd clicky inspector thingy for those PROTOs in code editors
- added Gelu, AffineGrid ops
- ImageDecoder op is cursed af (imports PIL) but gpt4 explicitly told me to not try to reimplement image decoding, so idk.
- auto_pad for MaxPool or AveragePool still a little janky
- pinned `onnx==1.15.0` in `setup.py` for now. I suppose we should update the pinned vers as onnx updates so that CI doesn't randomly break due to onnx updating to 1.16

speaking of onnx 1.16 .... 
image

welp 
:^)


geohotstan:
> Isn't it needed to include `tinygrad.extras` in `setup.py` packages? Without that how is it possible to use `onnx`?

idk if adding extras in setup.py would be a good idea since it's outside the scope of _core_ tinygrad, and the core tinygrad code is of very high quality.

there are examples of onnx used in `test/external/external_model_benchmark.py` that kinda demonstrates the usage :s

extra/onnx_ops.py
```diff
@@ -80,15 +81,15 @@ def Clip(input: Tensor, min=None, max=None): return input.clip(float('-inf') if
...
-def ReduceMean(data: Tensor, axes=None, keepdims=1, noop_with_empty_axes=0): return data.mean(_axes(axes, noop_with_empty_axes), keepdim=keepdims)
-def ReduceSumSquare(data: Tensor, axes=None, keepdims=1, noop_with_empty_axes=0): return data.square().sum(_axes(axes, noop_with_empty_axes), keepdim=keepdims)
-def ReduceL1(data: Tensor, axes=None, keepdims=1, noop_with_empty_axes=0): return data.abs().sum(_axes(axes, noop_with_empty_axes), keepdim=keepdims)
-def ReduceL2(data: Tensor, axes=None, keepdims=1, noop_with_empty_axes=0): return data.square().sum(_axes(axes, noop_with_empty_axes), keepdim=keepdims).sqrt()
-def ReduceLogSum(data: Tensor, axes=None, keepdims=1, noop_with_empty_axes=0): return data.sum(_axes(axes, noop_with_empty_axes), keepdim=keepdims).log()
-def ReduceLogSumExp(data: Tensor, axes=None, keepdims=1, noop_with_empty_axes=0): return data.exp().sum(_axes(axes, noop_with_empty_axes), keepdim=keepdims).log()
+def ReduceMax(data: Tensor, axes=None, keepdims=1, noop_with_empty_axes=0): return data.cast(dtypes.float).max(_axes(axes, noop_with_empty_axes), keepdim=keepdims).cast(data.dtype)
```
geohot, line 84: Is this cast needed? Why?

extra/onnx_ops.py
```diff
@@ -579,21 +589,71 @@ def EyeLike(x: Tensor, dtype=None, k=0):
...
+  elif pixel_format == "RGB":
+    return Tensor(np.array(img))
+  elif pixel_format == "Grayscale":
+    img = img.convert("L")
+    decoded = Tensor(np.array(img))
+    return decoded.unsqueeze(-1) # (H, W) to (H, W, 1)
+  else:
+    raise ValueError(f"pixel_format={pixel_format!r} is not supported.")
+
+def Gelu(x:Tensor, approximate=None):
```
geohot, line 627: Can this use Tensor.gelu?

extra/onnx_ops.py
```diff
@@ -579,21 +589,71 @@ def EyeLike(x: Tensor, dtype=None, k=0):
...
-  return (x - x_zer) * x_sc
+  return ((x - x_zer) * x_sc).cast(x_scale.dtype)
 
 # Needs work
-def IsNaN(x):
+def IsNaN(x: Tensor):
   return (x < float("-inf")).cast(dtypes.bool)
 
+# copied from https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_image_decoder.py
+# without importing PIL we'll have to manually decode a bunch of image formats like PNG, JPEG, WebP, etc
```
geohot, line 609: Use Pillow is fine

extra/onnx_ops.py
geohotstan: Ahh yeah... The casts are silly.
I initially thought casting like this ensures correctness and avoid having to change core tinygrad code, but this is also kinda dumb since increased kernel count probably

There are stupid tests like `test_reduce_min_bool_inputs_cpu` 
that has data as a bool tensor. 

on interpreted backends, we get this error
```
E   RuntimeError: Negation, the `-` operator, on a bool tensor is not supported. If you are trying to invert a mask, use the `~` or `logical_not()` operator instead.`
```
cuz Tensor.min uses NEG

----------

I tried changing UnaryOps.NEG in both ops_cpu and ops_torch.

test/external/external_test_onnx_backend.py
```diff
@@ -139,20 +145,26 @@ def supports_device(cls, device: str) -> bool:
 backend_test.exclude('test_resize_upsample_sizes_cubic_*') # unsure how to implement cubic
 
 # rest of the failing tests
-backend_test.exclude('test_averagepool_2d_dilations_cpu') # dilations != 1 not supported for avgpool
+backend_test.exclude('test_averagepool_2d_dilations_*') # dilations != 1 not supported for avgpool in tensor.py
```
geohot, line 148: You should just be able to add dilation to `avg_pool2d` in `tensor.py` just like `max_pool2d`

test/external/external_test_onnx_backend.py
```diff
@@ -165,15 +177,19 @@ def supports_device(cls, device: str) -> bool:
   backend_test.exclude('test_maxpool_2d_pads_cpu')
   backend_test.exclude('test_maxpool_2d_same_lower_cpu')
 
-# Don't know how to treat special TensorProto like TensorProto.FLOAT8E4M3FN
-if getenv("CPU") or getenv("TORCH"):
-  backend_test.exclude('test_dequantizelinear_axis_cpu')
-  backend_test.exclude('test_dequantizelinear_cpu')
-
-# compiled backends cannot reshape to and from 0
+# compiled backends cannot reshape to or from 0
```
geohot, line 180: @chenyuxyz we discussed this yesterday, we need to support 0s in the shape. torch does

```
>>> torch.zeros(10,10,10)[:, 0:0, :].shape
torch.Size([10, 0, 10])
```

geohot:
Thanks! Merged this, the other things can be new PRs.

PR #2225, opened 2023-11-06
Fix mmaped in jit

nimlgen:
Fixes a problem when jit overwrites the same buffer for backends which use mmaped buffer. This now makes a copy when .toCPU() is called, so several different numpy buffers do not share the same memory.


geohot:
Will merge, since correct > fast. But copies are really slow, any way to avoid this in at least some cases?

PR #2227, opened 2023-11-07
Update loss calculation in quickstart documentation

francislata:
# Overview
This pull request updates `quickstart.md` for the tinygrad MNIST example when computing the loss. When looking at how it is implemented now, it is directly part of `Tensor` and not as a standalone function. So I'm updating this documentation along the way.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -188,7 +188,7 @@ with Tensor.train():
     out = net(batch)
 
     # compute loss
-    loss = sparse_categorical_crossentropy(out, labels)
+    loss = out.sparse_categorical_crossentropy(labels)
 
     # zero gradients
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -188,7 +188,7 @@ with Tensor.train():
     out = net(batch)
 
     # compute loss
-    loss = sparse_categorical_crossentropy(out, labels)
+    loss = out.sparse_categorical_crossentropy(labels)
```
wozeparrot, line 191: It's defined on line 138. This was to show the implementation that is defined above.

francislata: Ah, got it. I definitely skipped that codeblock when I started to write the training code portion.

PR #2230, opened 2023-11-07
Remove PyYAML dependency

barathsku:
Removed pyyaml dependency from setup.py as it is not being used in tinygrad core files. _extra/assembly/assembly_rdna.py_ uses pyyaml, but it can be replaced with python's built-in json library.


extra/assembly/assembly_rdna.py
```diff
@@ -197,7 +197,7 @@ def assemble(self, args, ins, v_cnt, s_cnt):
                   '.wavefront_size': 32}],
                 'amdhsa.target': 'amdgcn-amd-amdhsa--gfx1100', 'amdhsa.version': [1, 2]}
 
-    code = boilerplate_start + "\n" + '\n'.join("%s %d" % x for x in kernel_desc.items()) + "\n" +  code_start + '\n'.join(ins) + "\n.amdgpu_metadata\n" + yaml.dump(metadata) + ".end_amdgpu_metadata"
+    code = boilerplate_start + "\n" + '\n'.join("%s %d" % x for x in kernel_desc.items()) + "\n" +  code_start + '\n'.join(ins) + "\n.amdgpu_metadata\n" + json.dump(metadata) + ".end_amdgpu_metadata"
```
geohot, line 200: ugh. does this work? why is yaml used here

PR #2233, opened 2023-11-07
universal user cache following XDG_CACHE_HOME posix

rodfer0x80:
universal user cache following XDG_CACHE_HOME posix compliance


geohot:
There's a cleaner way to write this as two lines, this is hard to read

geohot:
Merged this: https://github.com/tinygrad/tinygrad/pull/2235/files

Thanks for pushing for compliance with this

PR #2237, opened 2023-11-08
fix ops tests in test_dtype

Qazalin:
I had mistakenly used `size` instead of `itemsize` for upcasts, fixed here + asserts the upcast/downcast result dtype based on `priority` + one more test for same size ops


Qazalin:
back to draft to fix triton

PR #2243, opened 2023-11-09
remove type:ignore cuda 

rodfer0x80:
remove type:ignore cuda (safe push)


rodfer0x80:
also moved ctypes import to top as it would often be used with CUDA=1 and thus saving the overhead. but kindly correct me if i am wrong here

rodfer0x80:
pycuda.autoprimaryctx is making the linter fail but it is a necessary import. removing it breaks tests (for pycuda.driver.Context)

PR #2252, opened 2023-11-09
dynamic cache

rodfer0x80:
- set _cache_dir to ~/.cache/tinygrad or CACHEDB parent dir depending on env var is not set or set,
- replace os,sep with os.path.sep


geohot:
Huh? Why do I want this?

rodfer0x80:
reads cleaner but its just bike shedding

Issue #2240, opened 2023-11-08
Remove the `type: ignore`s that exist in the codebase


wozeparrot:
https://canary.discord.com/channels/1068976834382925865/1068976834928193609/1171944434011734037

chenyuxyz:
and add `--warn-unused-ignores` to mypy flag

geohot:
I'm closing this, enough are removed.

PR #2253, opened 2023-11-09
Remove unused type:ignore comments

Ayush-004:
This commit cleans up all instances where type:ignore comments are no longer necessary, as identified by running mypy with the --warn-unused-ignores flag. It addresses issue #2240.


geohot:
Can you add `--warn-unused-ignores` to CI to prevent regressions?

Ayush-004:
Sure, I'll add the --warn-unused-ignores flag in the CI pipeline. I'll update the workflow file and submit the changes for review shortly

PR #2257, opened 2023-11-10
Internal cast 2 with more tests

Qazalin:
- redoing ALU ops with different dtypes
- more tests to catch before it reaches master


Qazalin:
status: ran the benchmarks tests and the error is fixed - transitioning to tests now

geohot:
I'm going to merge this, I need it for loop folding stuff. Fingers crossed the benchmark tests pass...I wish there was a way to pretest this

Qazalin:
alright let's see what happens then

Qazalin:
i think they passed now

sobomax:
@Qazalin you broke bfloat support in the llvm backend here. :( It went unnoticed since the test_dtype.py got a bit broken in the be5f185ac0e commit, notice the TestBFloat16DType() is essentially NOP. Check PR #2765 where I am trying to fix it.

PR #2259, opened 2023-11-10
realize grad

DKormann:
gradients that have a lot of diamonds have potentially exponential runtime.

```python
from tinygrad.helpers import Timing

def run (N):
  with Timing(f"N={N} total:"):
    p = Tensor(1,requires_grad=True)

    for i in range(N):
        p = (p*0.5 + p*0.5).realize()
        
    l =p.sum()
    l.backward()

for i in range(10,25): run(i)
```


geohot:
Absolutely not. Realize is very undesirable, it breaks all fusion, it breaks scheduling.

If there's a runtime issue, it should be fixed in a different way.

PR #2264, opened 2023-11-10
Non fp32 math

Qazalin:
rendering as the buf_dtype unblocks bitcast too


tinygrad/helpers.py
```diff
@@ -79,13 +79,21 @@ class DType(NamedTuple):
   np: Optional[type]  # TODO: someday this will be removed with the "remove numpy" project
   sz: int = 1
   def __repr__(self): return f"dtypes.{INVERSE_DTYPES_DICT[self]}"
+  def get_vectorized_dtype(self, amt: int) -> DType:
+    if isinstance(self, ImageDType): return self.underlying.get_vectorized_dtype(amt)
+    assert self.sz == 1, f"{self} is not a scalar type"
+    return DTYPES_DICT[f"_{self.name.lower()}{amt}"]
+  def get_scaler(self) -> DType:
```
geohot, line 86: scalar

Qazalin:
status update:

- All the places in linearizer that hardcoded fp32 are dynamic
- Metal tests are green!

coming up:

- My highest pri is getting OpenCL green, the half dtype specifically needs a lot of work, fixing bugs one at a time

geohot:
I'm testing #2332 on my M3 where I only have llama 2. Very excited for fast load times!

Qazalin:
Status:
- [ ] LLVM pending green check
- [ ] Triton is particularly picky about not re-assigning variables with values of different dtypes implicitly, that needs a fix too
- [ ] After the two above, I'll cleanup the diff

.github/workflows/test.yml
```diff
@@ -171,11 +171,12 @@ jobs:
       - if: ${{ matrix.task == 'optimage'}}
         name: Test GPU IMAGE=2 ops
         run: GPU=1 IMAGE=2 python -m pytest -n=auto test/test_ops.py
-      - if: ${{ matrix.task == 'openpilot' }}
-        name: Test openpilot model compile and size
-        run: |
-          DEBUG=2 ALLOWED_KERNEL_COUNT=207 VALIDTEST=1 FLOAT16=1 DEBUGCL=1 GPU=1 IMAGE=2 python openpilot/compile2.py
-          python -c 'import os; assert os.path.getsize("/tmp/output.thneed") < 100_000_000'
+      # OpenCL needs cl_khr_fp16 to work
```
geohot, line 174: Just cast it to a float after the read.

Qazalin: I tried doing this https://github.com/tinygrad/tinygrad/pull/2264/commits/64e31ae8b5e89997b8359a74500181558f6c4546 which is the same as what exists in master, there will be no imageh actually rendered. This is what tinygrad master has:
https://github.com/tinygrad/tinygrad/blob/8ff2e135502170d55400dc45ea418b51c63b4e93/tinygrad/renderer/cstyle.py#L62
vs here:
https://github.com/tinygrad/tinygrad/blob/64e31ae8b5e89997b8359a74500181558f6c4546/tinygrad/renderer/cstyle.py#L62

so we can't really render imageh in CI if we wanna run FP16=1, because even casting that requires the extension, try compiling:
```py
from tinygrad.runtime.ops_gpu import compile_gpu
compile_gpu("""
__kernel void r_64_32_8_3_3_3_4_4_4(write_only image2d_t data0, read_only image2d_t data1, read_only image2d_t data2, const __global float* data3) {
    half4 val0 = read_imageh(data1, smp, (int2)(0, 0));
    float4 val1 = (float4)(val0.x,val0.y,val0.z,val0.w);
}
""") # program_source:4:49: error: loading directly from pointer to type '__private half' requires cl_khr_fp16. Use vector data load builtin functions instead
```

Qazalin: Update: I removed the support of imageh from this PR for now, I can follow-up with another PR for this if you're interested in rendering imageh in kernels.

test/external/external_test_onnx_backend.py
```diff
@@ -206,6 +214,37 @@ def supports_device(cls, device: str) -> bool:
   backend_test.include('test_zfnet512')
   backend_test.include('test_vgg19')
 
+if getenv('CLANG') and CI:
```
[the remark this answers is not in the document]
geohot: Why do they crash? That's probably out of bounds memory accesses, and bugs in this PR. We can't just skip the tests

geohot:
If they are failing in CI with clang and not locally, that could be a few things, but we have to figure out which.
1) OOM, you may have more memory locally
2) Memory corruption, you may just be getting lucky locally
3) Missing libraries in CI, like missing float16 support for example.

But yea, clang is very mature software. With the CI METAL I'm fine with skipping, but if clang is flaky there's a bug.

Qazalin:
Status: Now that we have a HIP compiler I can iterate on fixing HIP first and then figure out the CLANG regressions in CI. I'll work on those and reopen for review

test/external/external_test_onnx_backend.py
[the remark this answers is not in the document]
Qazalin: it was because of a bad division by zero because the preventions we do in `sign` only work for floats... updated sign to work for all dtypes
https://github.com/tinygrad/tinygrad/pull/2264/files#diff-7703e7b63c9844f019ae3609629fd9c9bab290d11ae6608ed2d09edd49bab2dbR631

Qazalin:
this is in a good spot.
I don't particularly like my two HIP hacks:
https://github.com/tinygrad/tinygrad/pull/2264/commits/efba3a3e65e2c48c67656f246b206e9cac63258f https://github.com/tinygrad/tinygrad/pull/2264/commits/70996c20f831b07cea9950069a345c530c90c166
I had to do for half dtype ALUs and wanna refactor those before opening up for a review

tinygrad/codegen/linearizer.py
```diff
@@ -46,6 +46,9 @@ def uop_alu_idx(self, a:UOp, b, ops, ctx:Linearizer, op, dtype=dtypes.int32):
 
   # NOTE: the consts have to be cached for deduping of downstream uops to work
   def const(self, b:Union[int,float], dtype=dtypes.int32, insert_before=None) -> UOp: return self.uop(UOps.CONST, dtype, tuple(), b, insert_before=insert_before)
+  def cast(self, val: UOp, dtype) -> UOp:
+    should_i_cast_you = val.dtype != dtype if self.opts.device != "HIP" else (val.dtype != dtype) or (val.uop == UOps.GEP and val.dtype == dtypes.half and val.vin[0].uop == UOps.LOAD and val.vin[0].dtype == dtypes.half.vec(2) and dtype == dtypes.half)
```
geohot, line 50: Why aren't all calls to cast added to UOps? This line looks sketchy

tinygrad/codegen/linearizer.py
```diff
@@ -455,8 +459,15 @@ def replace_op(old:UOp, new:UOp):
     self.applied_opts_cache = self.applied_opts[:]
     return self
 
-  def uop(self, uop:UOps, dtype:Optional[DType], vin:Tuple[UOp, ...], arg:Any=None, cachable=True, insert_before=None, simplify=True) -> UOp:
+  def uop(self, uop:UOps, dtype:Optional[DType]=None, vin:Tuple[UOp, ...]=tuple(), arg:Any=None, cachable=True, insert_before=None, simplify=True) -> UOp:
     key = (uop, dtype, vin, arg)
+    if uop == UOps.DEFINE_ACC and not dtypes.is_float(cast(DType, dtype)) and arg == -math.inf: arg = False if dtype == dtypes.bool else -2**31
```
geohot, line 464: Can these be added in a separate PR?

tinygrad/tensor.py
```diff
@@ -632,7 +632,7 @@ def floor(self: Tensor) -> Tensor: return (self < (b := self.trunc())).where(b-1
   def square(self): return self*self
   def clip(self, min_, max_): return self.maximum(min_).minimum(max_)
   def abs(self): return self.relu() + (-self).relu()
-  def sign(self): return self / (self.abs() + 1e-10)
+  def sign(self): return self / (self.abs() + 1e-10) if dtypes.is_float(self.dtype) else (self > 0) - (self < 0)
```
geohot, line 635: Why do you need the if? Isn't the second one always valid?

tinygrad/renderer/llvmir.py
```diff
@@ -145,7 +146,7 @@ def store_op(): bb[-1].store(element, bb[-1].gep(lvars[vin[0]], [lvars[vin[1]]],
         with bb[-1].if_then(bb[-1].trunc(lvars[vin[3]], ir.IntType(1))): store_op()
       else: store_op()
     if uop == UOps.ALU:
-      lvars[u] = cast(bb, code_for_op[args](bb[-1], *[cast(bb, lvars[x], x.dtype, dtypes.float) for x in vin]), dtypes.float, dtype)
+      lvars[u] = cast(bb, code_for_op[args](bb[-1], *[cast(bb, lvars[x], x.dtype, dtypes.float) for x in vin]), dtypes.float, dtype) if not (args == UnaryOps.NEG and dtype == dtypes.bool) else bb[-1].xor(lvars[vin[0]], ir.Constant(ir.IntType(1), 1))
```
geohot, line 149: This logic should probably be in code_for_op

Qazalin: ok two options:
allow code_for_op to take a dtype https://github.com/tinygrad/tinygrad/compare/master...Qazalin:tinygrad:code-for-op-with-dtype
Implement a new UnaryOps.NOT https://github.com/tinygrad/tinygrad/compare/master...Qazalin:tinygrad:UnaryOps-NOT
[similar to torch's logical_not() ](https://pytorch.org/docs/stable/generated/torch.logical_not.html)

tinygrad/codegen/linearizer.py
Qazalin: the function is just a utility for cleaner implementations that gets used in the ALU upcast:
https://github.com/tinygrad/tinygrad/pull/2264/files#diff-63d57daaca59e5495bbb108da1c13e782f4a408db8503e015e8094ac07c87ea6R466
This isn't a cast that's explicitly in the AST but we should cast it for compatibility with all backends.

Some backends like Metal don't need all ALU vars to be the same dtype and some (HIP, LLVM) do, so I'm opting to upcast in linearizer.

---
There's another part here, the half2 dtype is defined by HIP's library, but the underlying value of `val0.x` in a half2 isn't half, it's uint16:
https://github.com/argonne-lcf/THAPI/blob/f921880387730e75f04b97a8b07d191a8d28bc28/hip/include/hip/amd_detail/amd_hip_fp16.h#L54-L64

this won't compile:
```
half2 val0 = // ...
half val1 = // ...
val0.x + val1 // error: use of overloaded operator '+' is ambiguous (with operand types 'unsigned short' and 'half' (aka '__half'))
```

I moved this edge-case to linearizer where we do the GEP https://github.com/tinygrad/tinygrad/pull/2264/files#diff-63d57daaca59e5495bbb108da1c13e782f4a408db8503e015e8094ac07c87ea6R118 - so we'll treat the uop as a uint16 just like HIP and leave casting back to half to the other uops if they need to

tinygrad/renderer/hip.py
```diff
@@ -30,9 +32,17 @@ class HIPLanguage(CStyleLanguage):
 __device__ void vstore_half(float data, size_t offset, half *p) { *(p + offset) = (half)data; }
 __device__ void vstore_half2(float2 data, size_t offset, half *p) { *(p + offset*2) = (half)data.x; *(p + offset*2 + 1) = (half)data.y; }
 __device__ void vstore_half4(float4 data, size_t offset, half *p) { *(p + offset*4) = (half)data.x; *(p + offset*4 + 1) = (half)data.y; *(p + offset*4 + 2) = (half)data.z; *(p + offset*4 + 3) = (half)data.w; }
+__device__ half exp2(half x) { return hexp2(x); }
+__device__ half log2(half x) { return hlog2(x); }
+__device__ half sin(half x) { return hsin(x); }
+__device__ half sqrt(half x) { return hsqrt(x); }
+__device__ half hmax(half a, half b) { return __hgt(a, b) ? a : b; }
+__device__ half operator%(const half &a, const half &b) { return __hsub(a, __hmul(b, __float2half(floorf(__half2float(a) / __half2float(b))))); }
+__device__ bool operator!=(const half &a, const int &b) { return (float)a != b; }
```
Qazalin, line 41: This is because of the implementation for `TernaryOps.WHERE` https://github.com/tinygrad/tinygrad/blob/master/tinygrad/renderer/cstyle.py#L41C1-L41C1

Qazalin:
Ran test_ops in all of HIP and it compiles. Tried running in CI but [takes about 10 mins just to compile it](https://github.com/tinygrad/tinygrad/actions/runs/7060323436/job/19219639743)

tinygrad/tensor.py
```diff
@@ -632,7 +632,7 @@ def floor(self: Tensor) -> Tensor: return (self < (b := self.trunc())).where(b-1
   def square(self): return self*self
   def clip(self, min_, max_): return self.maximum(min_).minimum(max_)
   def abs(self): return self.relu() + (-self).relu()
-  def sign(self): return self / (self.abs() + 1e-10)
+  def sign(self): return ((self.float()) / (self.float().abs() + 1e-12)).cast(self.dtype)
```
geohot, line 635: Oh, I meant the other way, with the comparisons. The dividing is terrible

tinygrad/renderer/llvmir.py
geohot: So definitely not on the new op.

If code_for_op takes a dtype (which I'm largely fine with), what does that dtype apply to? Inputs? Outputs? Both?

tinygrad/tensor.py
Qazalin: i know i know - I tried it but test_ops breaks because we can't do a backward pass on a less than node

Qazalin: https://github.com/tinygrad/tinygrad/actions/runs/7058043827/job/19212838796#step:10:176 test run logs

tinygrad/renderer/llvmir.py
Qazalin: both, inputs always get upcasted to the final output's dtype

tinygrad/tensor.py
geohot: I'm not sure sign is supposed to be differentiable, is torch's?

tinygrad/renderer/hip.py
```diff
@@ -30,9 +32,16 @@ class HIPLanguage(CStyleLanguage):
 __device__ void vstore_half(float data, size_t offset, half *p) { *(p + offset) = (half)data; }
 __device__ void vstore_half2(float2 data, size_t offset, half *p) { *(p + offset*2) = (half)data.x; *(p + offset*2 + 1) = (half)data.y; }
 __device__ void vstore_half4(float4 data, size_t offset, half *p) { *(p + offset*4) = (half)data.x; *(p + offset*4 + 1) = (half)data.y; *(p + offset*4 + 2) = (half)data.z; *(p + offset*4 + 3) = (half)data.w; }
+__device__ half exp2(half x) { return hexp2(x); }
+__device__ half log2(half x) { return hlog2(x); }
+__device__ half sin(half x) { return hsin(x); }
+__device__ half sqrt(half x) { return hsqrt(x); }
+__device__ half mod(half x, half b) { return __hsub(x, __hmul(b, __float2half(floorf(__half2float(x) / __half2float(b))))); }
+__device__ half hmax(half a, half b) { return __hgt(a, b) ? a : b; }
```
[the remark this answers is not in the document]
geohot: Meh, I'm happy with the dtype arg anyway

tinygrad/codegen/linearizer.py
```diff
@@ -113,7 +115,7 @@ def rename_var(v: VariableOrNum, expr: str): return v if isinstance(v, NumNode)
           rendered_idx = idx.render(self.render_ops, self)
           valid_tuple = (valid.render(self.render_ops, self), self.const(invalid_value, localtype)) if valid.min == 0 else tuple()
           self.load_cache[key] = self.uop(UOps.LOAD, localtype, (buf_uop, rendered_idx) + valid_tuple + ((barrier,) if barrier else ()))
-      ret.append(self.uop(UOps.GEP, localtype.scalar(), (self.load_cache[key],), rep_idx[dim]) if dim is not None else self.load_cache[key])
+      ret.append(self.uop(UOps.GEP, localtype.scalar() if not (self.opts.device == "HIP" and localtype == dtypes.half.vec(2)) else dtypes.uint16, (self.load_cache[key],), rep_idx[dim]) if dim is not None else self.load_cache[key]) # In HIP, the elements within half2 are uint16
```
geohot, line 118: There should not be device specific hacks in the linearizer

geohot:
Getting close, let me know when this is ready to merge

tinygrad/tensor.py
Qazalin: it's not https://discuss.pytorch.org/t/torch-sign-breaks-autograd-dynamic-graph/90711/2 - but you _can_ compute the grad (will be 0.0)

tinygrad/codegen/linearizer.py
Qazalin: This is what it takes https://github.com/tinygrad/tinygrad/pull/2264/commits/7515fb2b4251f6c1908f06511bee2f34ee60a22f :) I'm ok with it i think it's a better place to put it, until we have a uop pre-processor based on device maybe (don't know how that'd look like)

Qazalin:
@geohot cstyle is done - but to get rid of casting to fp32 in llvm ALUs:
https://github.com/tinygrad/tinygrad/pull/2264/files#diff-eb3a3956449cb9727741ec803e42c03395a66d4d1efdf941ba4ed9ef3829d0b6R149
we'd need more dtypes in the llvm renderer to:
1. be more selective with [fast math flags](https://github.com/tinygrad/tinygrad/blob/008c724ae34278c19dd11dad5e543c9a1883b494/tinygrad/renderer/llvmir.py#L7)
2. get a more complete cmplt implementation https://github.com/tinygrad/tinygrad/blob/master/tinygrad/renderer/llvmir.py#L19-L20

tinygrad/renderer/llvmir.py
Qazalin, line 1: in code_for_op we used to render the lambda as first. float32, then int and then bool.
This is wrong in several ways:
1. isinstance(x.type, ir.FloatType()) isn't covering Half and Double
2. isinstance(x.type, ir.IntType) will always be true for bool (it's i1), so we'll never reach the actual bool implementation.

These changes refactor it to be: first)bool, then other ints, everything else is some kind of float

geohot:
Congrats!! $500 bounty is yours after the tests are merged

Issue #2249, opened 2023-11-09
Have a passable (vaguely mistakable for human) voice conversation with tinygrad (whisper + LLaMA + VITS)


shauray8:
I do not see anyone working on [this](https://docs.google.com/spreadsheets/d/1WKHbT-7KOgjEawq5h5Ic1qUWzpfAzuD_J06N1JwOCGs/edit#gid=0&range=A50), do you have any specifics in mind for this on how this should be done? or any tests?
@wozeparrot @geohot 

---

Current Bounty: $400
To lock the bounty submit a draft PR with a decent amount of progress made
Make sure to reference this issue in the PR for future tracking

Notice: _If the PR goes stale the bounty will be unlocked_

wpmed92:
For any bounty a first step is to put up a good draft pr, if it's good enough the bounty will be locked for your.

geohotstan:
dosier did an attempt at this before #1396 
check it out since he has done a lot of the API work already
I'm pretty sure he's got a good idea on how it should be done (he claimed the VITS TTS bounty)

I remember geohot mentioning in discord that the conversation has to stream like a natural convo
so basically create a demo that is capable of streaming the convo with low enough latency

wozeparrot:
Going to use this as the tracking issue.

J4e6eR:
@geohotstan @wozeparrot How to join the discord server, I found the invite link from the tinygrad website but that doesn't seem to work for me. 
I don't know why I am getting this error, is there any alternative way to join ?? 
I've been trying to join since a month.
![image](https://github.com/tinygrad/tinygrad/assets/91129126/f007e935-0242-438a-bb67-22a2e983aa7f)

J4e6eR:
@geohotstan @wozeparrot I am sorry, to divert off from the main topic but helping me out in joining tinygrad's official discord server would truly be appreciable. I'll delete these comments after that for sure!!

wozeparrot:
> @geohotstan @wozeparrot I am sorry, to divert off from the main topic but helping me out in joining tinygrad's official discord server would truly be appreciable. I'll delete these comments after that for sure!!

Can you move this to a discussion? thanks

J4e6eR:
> @J4e6eR Can you move this to a discussion? This really isn't the place for stuff like this.
Added in the discussions tab, please have a look 
https://github.com/tinygrad/tinygrad/discussions/2276

MeDott29:
Sweep: figure out how to solve this issue

[...]

PR #2265, opened 2023-11-10
[Bounty] Voice conversation with tinygrad

francislata:
**Bounty description**
Have a passable (vaguely mistakable for human) voice conversation with tinygrad (whisper + LLaMA + VITS)

# Issue
#2249

# Task list
- [x] Clean up Whisper code
- [ ] Add ability to support different LLaMA models
- [ ] Add support for different VITS speakers
- [ ] Add support for CLI argument parsing
- [ ] Add text display in prompt-like manner
- [ ] Clean up code further
- [ ] Make it faster!


geohot:
Not enough for a lock yet, I think this can be locked once a single back and forth happens. Then we push on speed

francislata:
@geohot - I did some updates. Definitely still lot more to work with but at least got a POC going where you can speak to it, prints what you said, asks LLaMA to generate response, and use VITS to read it for you.

Let me know what your thoughts are if this progress can lock the bounty.

PR #2270, opened 2023-11-11
add dilation to avg_pool2d

rodfer0x80:


geohot:
The whole point is to reenable the test!

rodfer0x80:
Makes sense, will work on this

rodfer0x80:
fix avg_pool2d passing onnx tests

rodfer0x80:
-- add tensor avg pooling dilation feature
-- pass avg_pool2d and avg_pool3d onnx tests
-- hotfix onnx padding

rodfer0x80:
- Torch doesn't support dilation in avg_pool but does in max_pool:
    torch.nn.AvgPool2d(kernel_size, stride=None, padding=0, ceil_mode=False, count_include_pad=True, divisor_override=None)
- Do the ONNX test suffice or do we need avg_pool tests for dilation in test/test_ops.py?

PR #2273, opened 2023-11-11
pad None means (0,0)

chenyuxyz:
similar to shrink


geohot:
Nice yea it's much cleaner than tuples

PR #2274, opened 2023-11-11
test_ops for all dtypes

Qazalin:
non-fp32 alu ops needs some good test infra
in progress:
- [ ] Random inputs causes problems in int8 because it can overflow in some ops
- [x] ALU with half has a dtype missmatch between torch and tinygrad -> had to change type promotion priorities
- [x] `compare` should do the correct thing based on the dtype
- [x] All cases unsupported in non-fp32 (eg. `test_cumsum`) should move from the base `TestOps` to `TestFloatOps`
- [x] `helper_test_op`
- [ ] diff is too big


Qazalin:
back to working on this, though needs to be broken down into several refactor PRs once i figure out a structure that applies to all dtypes

Qazalin:
I'll break this up into separate refactors

PR #2277, opened 2023-11-11
modernize setup.py

NotAndex:
### Proposal to modernize `setup.py` as suggested by @geohot on Discord

* Moved everything from `setup.py` to `pyproject.toml` and removed `setup.py` completely.
  * Clarification for setuptools as build-backend in `pyproject.toml`: [source](https://discuss.python.org/t/user-experience-with-porting-off-setup-py/37502/33)
* Checked for invocation of `setup.py`. 
  * Was only used for the build in `push_pypi.sh` and `.github/workflows/python-publish.yml`
  * I replaced the invocation in both files with `python -m build` according to: [source](https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives)
* Updated `.github/workflows/test.yml` to check `pyproject.toml` in cache step for dependency update instead of `setup.py`

> [!NOTE]  
> ["[...] `setup.py` itself is not deprecated, only invoking it directly"](https://discuss.python.org/t/user-experience-with-porting-off-setup-py/37502/10)

---

Further considerations:
* `ruff.toml` can be moved to `pyproject.toml` [source](https://docs.astral.sh/ruff/settings/#top-level)
* `pytest.ini` can be moved to `pyproject.toml` [source](https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml)
* Not sure about `mypy.ini`

> [!IMPORTANT]
> I think in general, a decision needs to be made: Should `tinygrad` have as much configuration as possible in `pyproject.toml`, or should the configuration files be left at the root level for an easy indication of the tools being used?


geohot:
We tried this and pip install broke...

NotAndex:
Can you give me a hint of what broke or how I might reproduce the issue?

I was testing it against `test.yml`-workflow and in a Devcontainer with `python -m build` and `python3 -m pip install dist/tinygrad-0.7.0.tar.gz ` and flat `python3 -m pip install .`

geohot:
The old one is in the commit history, I merged it and had to revert when pip install didn't work. Try checking out the old commit and pip installing that...if it doesn't work and yours doesn't have the bug, great!

If the old one does work, I can help figure out what was wrong with my machine.

Does this work on Python 3.8 / Ubuntu 20.04?

NotAndex:
Good hint!

I tried it now using Ubuntu *focal* 20.04 (py3.8) and *jammy* 22.04 (py3.10). The problem was neither one had pip [preinstalled](https://www.webhi.com/how-to/how-to-install-pip3-in-ubuntu/). 
I had to run: `sudo apt-get update && sudo apt-get install -y python3-pip`

Then, both the setup.py and pyproject.toml were **_working_**. 

If you want to invoke `setup.py` directly you still need to have `python3-pip` installed (this installs setuptools, too).
Then you can run: `python3 setup.py sdist bdist_wheel` -> `pip install dist/tinygrad-0.7.0.tar.gz`.

Now I also understand why Python 3.12 is shipping [without](https://docs.python.org/3/whatsnew/3.12.html) default setuptools, because invoking setup.py is deprecated, and in `pyproject.toml`, it can be defined as a build requirement.

geohot:
Yea that setup is fine.

Did the old attempt fail and the new attempt succeed? Or could you not reproduce the old failure?

NotAndex:
You are right! `pip install -e .[testing,linting]` is indeed failing on _focal_, I could reproduce that.

## Thoughts
* I think the problem is, that `sudo apt update && sudo apt install -y python3-pip` is not _"correctly"_ installing `pip`.

* _Focal_ is installing _setuptools_ in `/usr/lib/python3/dist-packages`, which `pip install -e .[testing,linting]` is defaulting to, instead of using the latest version from `sudo pip install --upgrade setuptools` located in `/usr/local/lib/python3.8/dist-packages`

* I was **not able** to figure out how to update the version of _setuptools_ located in `/dist-packages`

## 1st potential solution (working in Devcontainer)
```sh
sudo apt-get update && sudo apt-get install -y python3-pip 
pip install --upgrade pip 
pip install --upgrade setuptools 
python3 -m pip install -e .[testing,linting] 
```

## 2nd potential solution (working in Devcontainer)
After following this [guide] (https://packaging.python.org/en/latest/tutorials/installing-packages/), I installed pip the following way:

```sh
sudo apt update && sudo apt install -y python3-distutils
sudo python3 get-pip.py # https://bootstrap.pypa.io/get-pip.py
pip install -e .[testing,linting]
```
This is working because `pip install -e .[testing,linting]` is using the latest version 68.2.2 of _setuptools_.

geohot:
Yea, but that's the problem. If it breaks working out of the box on Ubuntu 20.04, this PR isn't worth the gains.

NotAndex:
That's fair; I agree.

And to conclude, from what I've understood, using `setup.py` and the way `tinygrad` invokes it through pip is good practice and is not considered deprecated.

PS: In any case, thanks, George, for the prompt responses. Your efforts with the _tiny corp_ are truly valued and indispensable. Keep up the excellent work!

PR #2278, opened 2023-11-12
make self referential AST fast too

DKormann:
## problem
building of self referential computation graph leads to exponential runtime even without being realized:

```
from tinygrad.ops import Device,BinaryOps

Device.DEFAULT = "METAL"

N=30
p = LazyBuffer.fromCPU(np.array([1]))
for _ in range(N): p = p.e(BinaryOps.ADD, p)

```

## source:
when building the Ast a list of exponentially many buffers is created in order to set .child property on all buffers.

## solution:

create a function to get unique buffers for LazyOp. just calling set() on the normal buffers doesn't help since we still would visit the .src exponentially often.

### DIFF
--- a/tinygrad/lazy.py
+++ b/tinygrad/lazy.py
@@ -183,7 +183,7 @@ class LazyBuffer:
     # NOTE: dtypes.from_np(self.dtype.np) to deal with image types
     return self.loadop(LoadOps.CONST, tuple(), dtypes.from_np(self.dtype.np), self.device, arg=val).reshape((1,)*len(self.shape)).expand(self.shape)
 
-  def copy_device(self, device:str) -> LazyBuffer:
+  def copy_to_device(self, device:str) -> LazyBuffer:
     # back off a FROM if it's a double FROM
     if not self.realized and self.op.op == LoadOps.FROM and cast(LazyBuffer, self.op.src[0]).device == device: return cast(LazyBuffer, self.op.src[0])
     return LazyBuffer.loadop(LoadOps.FROM, self.shape, self.dtype, device, src=self.contiguous())
--- a/tinygrad/lazy.py
+++ b/tinygrad/lazy.py
@@ -36,7 +36,7 @@ def _ast_reduceops(op:LazyOp) -> LazyOp:
 
 # this supports late merging an upstream Reduce op and even an Elementwise op above that
 def _ast_binaryops(op:LazyOp, shape: Tuple[sint, ...]) -> LazyOp:
-  real_srcs: Dict[LazyBuffer, Optional[Union[LazyOp, LazyBuffer]]] = {x:None for x in op.buffers()}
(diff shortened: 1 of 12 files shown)


geohot:
Why is buffers even needed? Is there a way to do this refactor that loses lines instead of gains them?

DKormann:
it seems connected to all kinds of slow downs

DKormann:
wow your right actually buffers was basically not needed.

geohot:
Now minimize the diff. Do you have to change the name?

DKormann:
could make the buffers into property like before but its 2 + lines or so I think not worth it?

tinygrad/lazy.py
```diff
@@ -143,8 +143,7 @@ def key(self):
 
   def _device_extra_args(self) -> Dict[str, str]: return {"device": self.device.split(":", 1)[1]} if ":" in self.device else {}
 
-  @property
-  def buffers(self) -> Tuple[LazyBuffer, ...]: return (self,)
+  def buffers(self,_=None) -> Tuple[LazyBuffer, ...]: return (self,)
```
geohot, line 146: Why can't this stay a property?

DKormann: I can even put it as a regular property in the __init__ since it always just returns itseld

DKormann:
the whole hash thing also doesn't seem relevant when caching buffers.
super sorry this took so long

tinygrad/ops.py
```diff
@@ -1,8 +1,10 @@
 from __future__ import annotations
 import importlib, inspect, functools, pathlib, re
 from enum import Enum, auto
+
```
geohot, line 4: Why change this whitespace?

tinygrad/ops.py
```diff
@@ -52,13 +54,11 @@ class LazyOp:
...
   def __repr__(self): return f"LazyOp(op={self.op}, src={self.src}, arg={self.arg})"
-  @property
-  def buffers(self):
-    buffers: Tuple[Union[LazyOp, LazyBuffer], ...] = ()
-    try:  # NOTE: the linearizer's key function maps the buffers to ints, and LOCAL_BUFFER is used. we don't care about buffers in these cases
-      for x in self.src: buffers += x.buffers
-    except AttributeError: buffers = ()
-    return buffers
+  @functools.cached_property
+  def buffers(self)-> Tuple[LazyBuffer,...]: return tuple(dedup(sum([x.buffers for x in self.src],())))
```
geohot, line 58: `) -> `

tinygrad/ops.py
```diff
@@ -52,13 +54,11 @@ class LazyOp:
...
+  @functools.cached_property
+  def hash(self): return hash((self.op,self.src, self.arg))
+  def __hash__(self):return self.hash
```
geohot, line 61: `: return`

geohot:
Looks good, just formatting issues now. Ask, is this beautiful code?

PR #2280, opened 2023-11-12
Add Poetry and Docker environment support

Zaffer:
Add support for Poetry and Dockerfile to manage fully isolated environments.

This helps with removing class of errors that emerge from installing directly into your system.

PEP 621 will be supported by Poetry soon and removes the need for setuptools etc.
Poetry uses a superior lock file mechanism for dependencies, and running in docker insures developing on Tinygrad have a consistent environment for development.


geohot:
What errors emerge from directly installing on the system. We should fix them, not pave over errors with containers.

Zaffer:
> What errors emerge from directly installing on the system? We should fix them, not pave over errors with containers.

It's for developer experience, avoiding potential environment errors. For example: tinygrad can't be installed in python 3.12 yet, due to removal of distutils.

Not a core enhancement, so we don't need it yet, especially since we are waiting for Poetry to implement PEP 621.

I prefer developing inside 100% isolated containers.

PR #2281, opened 2023-11-12
refactor/ci: delete many `# type: ignore`

76616c6172:
* replace `axis.__class__ is int` with `isinstance(axis, int)` to make mypy happy
* add `--warn-unused-ignores` to mypy flag

refs #2240

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
     - name: Lint tinygrad with pylint
       run: python -m pylint tinygrad/
     - name: Run mypy
-      run: python -m mypy --warn-unused-ignores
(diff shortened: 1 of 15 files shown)


.github/workflows/test.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     - name: Lint tinygrad with pylint
       run: python -m pylint tinygrad/
     - name: Run mypy
-      run: python -m mypy
+      run: python -m mypy --warn-unused-ignores
```
geohot, line 43: Can you add this to the mypy config instead?

76616c6172: Done, good point!

Issue #2031, opened 2023-10-10
No module named 'models' when trying to run examples/compile_efficientnet.py


mikolysz:
When trying to run:
```
python3 examples/compile_efficientnet.py
```

The following output is returned:
```
File "/Users/miki/code/tinygrad/examples/compile_efficientnet.py", line 2, in
<module>
    from models.efficientnet import EfficientNet
ModuleNotFoundError: No module named 'models'
```

Tinygrad is installed from source as recommended in the readme, the behavior doesn't change whether the working directory is `tinygrad` or `examples`.

wozeparrot:
Can you try with `PYTHONPATH=. python3 examples/compile_efficientnet.py`

J4e6eR:
Try adding these lines before the starting of the code
```python
import sys
from pathlib import Path
sys.path.append(str(Path(__file__).parent.parent))
```
The number of time you write the parent will determine which path you want to add to the system.
This is the path python would look for if they won't find a module.
Generally it's better to append the root director of the repo as a sys path to access all the modules from the repo.

wozeparrot:
@mikolysz This still an issue for you?

hmih:
@mikolysz It is something related to your PYTHONPATH, can you try this installation:

```shell
git clone https://github.com/tinygrad/tinygrad.git
cd tinygrad
python3 -m venv venv
. venv/bin/activate         # do not forget the dot!
python3 -m pip install -e . # do not forget the dot!
```

PR #2282, opened 2023-11-12
Tell newcomers to use virtualenvs

hmih:
This should fix issues of newcomers polluting their global package space and having issues with imports: #2031 

The readme update uses the built-in `venv` module that's available since 3.3. I thought about using `pyproject.toml` instead of `setup.py` to be compliant with the latest PEPs but I think it's an overkill because the setup right now works just fine.


geohot:
I dislike virtualenvs. tinygrad depends on like four unpinned packages.

PR #2283, opened 2023-11-12
Fix llama lm_eval

chenyuxyz:
model call takes int and converts to Variable internally, don't use Variable when not jitted and when Mask is not None.

added an option to return logits that's used by lm_eval


chenyuxyz:
still not correct, need a function to clear the kvcache. previous version worked because the kvcache resets implicitly when consuming a new prompt with start_pos = 0

PR #2287, opened 2023-11-13
get_linearizer_actions does not return illegal actions

nimlgen:


test/test_linearizer_failures.py
```diff
@@ -26,27 +26,40 @@ def test_failure_2(self):
...
     # related to OptOps.NOLOCALS
     # IndexError: list index out of range
     assert fuzz_linearizer(lin) != "PASS"
 
-  @unittest.skipUnless(Device.DEFAULT in ["CLANG", "LLVM"], "fails on these backends")
+
+class TestLinearizerFailuresFixed(unittest.TestCase):
+  @unittest.skipUnless(Device.DEFAULT in ["LLVM"], "fixed on these backends")
+  def test_failure_3(self):
+    ast = LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BufferOps.MEM, src=(), arg=MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(32, 8, 16, 16), strides=(2048, 256, 16, 1), offset=0, mask=None, contiguous=True),)))),), arg=(32, 8, 16, 1))
```
chenyuxyz, line 48: cool i like the regression test. can you write it in a way that reuses the ast and not copied?

nimlgen: better?

chenyuxyz:
thanks!

PR #2288, opened 2023-11-13
back to 6.54GB for stable diffusion

geohot:


tinygrad/runtime/ops_shm.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, size, dtype:DType, device:str):
       fd = _posixshmem.shm_open(device, os.O_RDWR, 0o600)
       # TODO: these flags are somewhat platform specific, but python doesn't expose the ones we need
       shm = mmap.mmap(fd, size * dtype.itemsize, flags=mmap.MAP_SHARED | 0x2000 | 0x008000)
-      shm.madvise(mmap.MADV_HUGEPAGE)
+      shm.madvise(mmap.MADV_HUGEPAGE)     # type: ignore   # not on OSX
```
wozeparrot, line 19: if this is the only thing preventing warn_unused_ignores, it's fine to just drop the entire madvise call, the amount of perf it gives is pretty negligible.

geohot: I mean...this is a more deep problem than just this though. it's valid on Linux

PR #2295, opened 2023-11-14
Add Tensor.multinomial 

marcellofuschi:
Only with replacement (for now)

### DIFF
--- a/test/test_randomness.py
+++ b/test/test_randomness.py
@@ -98,8 +98,8 @@ class TestRandomness(unittest.TestCase):
 
   def test_multinomial(self):
     def _check_with_torch(p, num_samples, replacement):
-      tiny_res = Tensor.multinomial(Tensor(p), num_samples, replacement=replacement)
-      torch_res = torch.multinomial(torch.tensor(p), num_samples, replacement=replacement)
+      tiny_res = Tensor(p).multinomial(num_samples, replacement=replacement)
+      torch_res = torch.tensor(p).multinomial(num_samples, replacement=replacement)
       self.assertEqual(tiny_res.shape, torch_res.shape)
       if torch_res.ndim == 1:
         tiny_res = tiny_res.unsqueeze(0)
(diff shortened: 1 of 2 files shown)


geohot:
Locked the bounty for you

marcellofuschi:
Fixed the first assert — the function now works with both 1D and 2D inputs. You cannot pass 3D or higher, just like torch

As for the second assert, sampling without replacement is trickier to implement lazily, so I'll need to think about it. Meanwhile, I'll set the PR as ready in case it's already useful just with replacement.

tinygrad/tensor.py
```diff
@@ -216,6 +216,19 @@ def kaiming_normal(*shape, a:float = 0.01, **kwargs) -> Tensor:
     std = math.sqrt(2.0 / (1 + a ** 2)) / math.sqrt(prod(shape[1:]))
     return Tensor.normal(*shape, mean=0.0, std=std, **kwargs)
 
+  @staticmethod
+  def multinomial(p: Tensor, num_samples: int, replacement: bool = False) -> Tensor:
+    assert p.ndim <= 2, "p must be 1 or 2 dim"
+    assert replacement, "supported only with replacement"
```
geohot, line 222: If num_samples is 1, is replacement=false okay?

marcellofuschi: yep, I'll allow it

imaolo:
Does any runtime checking need to happen (looking at torch.multinomial with negative values). I don't think there are OPs that raise exceptions, atleast I don't know about them.

test/test_randomness.py
```diff
@@ -96,6 +96,22 @@ def test_kaiming_normal(self):
...
+      if torch_res.ndim == 1:
+        tiny_res = tiny_res.unsqueeze(0)
+        torch_res = torch_res.unsqueeze(0)
+      for i in range(torch_res.shape[0]):
+        self.assertTrue(equal_distribution(lambda *_: tiny_res[i], lambda _: torch_res[i]))
+    _check_with_torch(p=[0.231, 0., 1., 0.5], num_samples=1000, replacement=True)
+    _check_with_torch(p=[[0.453, 0., 1., 0.81], [0.1, 0.8, 0., 0.1]], num_samples=1000, replacement=True)
+    # without-replacement isn't supported, unless taking only one sample
+    self.assertRaises(AssertionError, lambda: Tensor.multinomial(Tensor([0.5, 0.5]), num_samples=100, replacement=False))
+    _check_with_torch(p=[[0.5, 0.5], [0.99, 0.01]], num_samples=1, replacement=False)
```
geohot, line 113: Umm, doesn't this just fail 1% of the time?

marcellofuschi: It never fails actually, because for `num_samples=1` the ks test isn't reliable. That line was just checking that no error was raised when passing `num_samples=1, replacement=False`.

I've just changed the test code to make it more reliable by accumulating the results of repeated experiments.

Regarding the reliability of `equal_distribution` in general, I wrote a counterexample to test that the samples are considered from different distributions when they should. I ran it 1000 times and it always passed. I don't know if we should keep it in the code since the other tests in the file have no such thing.

geohot: Counterexample is great! Confirms the tests are actually testing something

tinygrad/tensor.py
```diff
@@ -216,6 +216,19 @@ def kaiming_normal(*shape, a:float = 0.01, **kwargs) -> Tensor:
     std = math.sqrt(2.0 / (1 + a ** 2)) / math.sqrt(prod(shape[1:]))
     return Tensor.normal(*shape, mean=0.0, std=std, **kwargs)
 
+  @staticmethod
```
geohot, line 219: This shouldn't be a staticmethod, and the first arg should be self.

Fixes `p.multinomial` and is a line shorter

marcellofuschi:
Ok, I removed staticmethod and changed the tests accordingly.

Plus, I found a bug with an edge case (input 2D tensor with only 1 row should still return 2D). I added an assert and fixed it.

geohot:
Sweet, can pay bounty via PayPal or USDC. E-mail george@tinygrad.org (or DM on discord) with your info to claim!

PR #2296, opened 2023-11-14
beam works with var vals

nimlgen:
Allow beam search on kernels with var vals.


chenyuxyz:
i think it's invalid to upcast / unroll a symbolic size dim, is this handled implicitly somehow?

chenyuxyz:
oh you are right, and it's not False but a RuntimeError if mod a Node

Issue #2177, opened 2023-10-29
ValueError accessing element of a padded Tensor


andresgit:
The command `Tensor([5,6]).pad(((1,0),))[0].numpy()` gives me a `ValueError`.

I'm running Arch Linux on an Intel based laptop. I installed a fresh ubuntu:latest docker container, then using apt installed git, python3 and python3-pip, then did the "Installation From source" instructions from the tinygrad GitHub Readme. This is what I get:

```python
>>> from tinygrad.tensor import Tensor
>>> Tensor([5,6]).pad(((1,0),)).numpy()
array([0., 5., 6.], dtype=float32)
>>> Tensor([5,6]).pad(((1,0),))[1].numpy()
array(5., dtype=float32)
>>> Tensor([5,6]).pad(((1,0),))[2].numpy()
array(6., dtype=float32)
>>> Tensor([5,6]).pad(((1,0),))[0].numpy()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/tinygrad/tinygrad/tensor.py", line 126, in numpy
    return self.detach().cast(dtypes.from_np(self.dtype.np)).contiguous().to('CPU').realize().lazydata.realized.toCPU().reshape(self.shape)
  File "/root/tinygrad/tinygrad/tensor.py", line 105, in realize
    run_schedule(self.lazydata.schedule())
  File "/root/tinygrad/tinygrad/realize.py", line 27, in run_schedule
    si.out.realized = Device[si.out.device].exec_ast(si.ast, output=si.out, inputs=si.inputs, var_vals=si.var_vals, **si.out._device_extra_args())
  File "/root/tinygrad/tinygrad/ops.py", line 133, in exec_ast
    srcs = [self.exec_ast(cast(LazyOp, x), inputs=inputs, context=context, **kwargs) for x in ast.src]
  File "/root/tinygrad/tinygrad/ops.py", line 133, in <listcomp>
    srcs = [self.exec_ast(cast(LazyOp, x), inputs=inputs, context=context, **kwargs) for x in ast.src]
  File "/root/tinygrad/tinygrad/ops.py", line 126, in exec_ast
    for mop,arg in ast.arg.st.to_movement_ops(): buf = self.fxn_for_op[mop](buf, arg)
  File "/root/tinygrad/tinygrad/runtime/ops_cpu.py", line 41, in <lambda>
[...]

Issue #2194, opened 2023-11-01
Accessing an element created by padding with a non-zero value gives zero


andresgit:
I'm using `GPU=1` and accessing an element created by padding with a non-zero value, but it still returns zero:

```python
>>> import os
>>> os.environ["GPU"] = "1"
>>> from tinygrad.tensor import Tensor
>>> Tensor([5,6]).pad(((0,1),),2).numpy()
array([5., 6., 2.], dtype=float32)
>>> Tensor([5,6]).pad(((0,1),),2)[0].numpy()
array(5., dtype=float32)
>>> Tensor([5,6]).pad(((0,1),),2)[1].numpy()
array(6., dtype=float32)
>>> Tensor([5,6]).pad(((0,1),),2)[2].numpy()
array(0., dtype=float32)
```
For the final return value I would instead expect `array(2., dtype=float32)`

imaolo:
I am reproducing on compiled backends and erroring on interpreted backends.

imaolo:
```
Device.DEFAULT='CLANG'
assert Tensor.ones(3)[2].numpy() == 1
assert Tensor.zeros(2).pad(((0,1),),1)[2].numpy() == 0
```

imaolo:
```
ARR_SIZE = 1
PAD_SIZE = 1 # >1 not an issue
PAD_VALUE = 543
Device.DEFAULT='GPU' # error if interpreted backend

np_src = np.pad([1]*ARR_SIZE,(0, PAD_SIZE), mode='constant', constant_values=PAD_VALUE)
tg_src = Tensor([1]*ARR_SIZE).pad(((0, PAD_SIZE),), PAD_VALUE)

np_np_slice = np_src[ARR_SIZE:]
np_tg_slice = Tensor(np_src)[ARR_SIZE:].numpy()
tg_tg_slice = tg_src[ARR_SIZE:].numpy()
tg_np_slice = tg_src.numpy()[ARR_SIZE:]

assert np.array_equal(np_np_slice, np_tg_slice)
assert not np.array_equal(np_np_slice, tg_tg_slice)
assert np.array_equal(np_np_slice, tg_np_slice)
```

imaolo:
Problem with pad x shrink on compiled backends

```
EXPECTED = np.array([0])

int_reg = Tensor([1,0], device='CPU')
int_reg = Shrink.apply(int_reg, arg=((1,2),))

int_pad = Tensor([1], device='CPU')
[...]

PR #2298, opened 2023-11-14
Fix accessing elements created by padding

andresgit:
This is an attempt to fix the issues #2177 and #2194 on the CPU interpreted backend. I will also try to see if I can fix it on compiled backends.

The problem is that `ShapeTracker.to_movement_ops` tries to calculate `real_offset`, the offset of the first element in the base buffer that appears in this view. But the view might not contain any such element (i.e. it is completely made up of elements created by padding). In this case it gets an offset that doesn't work.

As a fix I introduced `offset_noreal` to `View`, which is set `True` if we resize in a way that the `View.mask` doesn't contain any elements in the view. Then in `ShapeTracker.to_movement_ops` we treat this case separately by creating the required shape using padding.

I also made `View.mask` be between 0 and the new shape when resizing, because otherwise for example `Tensor([1,2]).pad(((3,0),))[0:2]` would have `View.mask=((3,2),)`, which caused problems.


andresgit:
It also fixes on CPU the `ValueError` caused by this code mentioned on Discord a couple of weeks ago:
`Tensor.stack([Tensor([1, 2, 3, 4, 5]) for i in range(3)])[0, :].numpy()`

geohot:
So this definitely doesn't belong in the view if it's only an interpreted bug

andresgit:
From the issues I added several test cases that also fail on compiled backends.
I managed to fix the errors on the compiled backends by returning all indices as invalid in the shapetracker.py function `expr_node_mask` when there is no base buffer element in the view (i.e. `View.offset_noreal` is set).

It fixes for example this on the GPU: `assert Tensor([[1],[2]]).pad(((0,0),(1,0)))[1,0]==0`

Instead of introducing a new variable `View.offset_noreal` one alternative possibility would be to set `View.offset` to `None` and then add a check for `None` everywhere the offset is used.

geohot:
I'm not convinced that a view change is required, feel free to reopen starting with a failing test case

andresgit:
I agree that it is not required to add a flag to View, it is just one possible fix and it is also possible to fix the issues in other ways.
I can try to submit a PR with a different approach.

PR #2304, opened 2023-11-15
Integrated A Novel Multi-Head Attention Layer.

hi-sushanta:
Please review my latest modifications, which incorporate the most widely used attention mechanism in deep learning.

**code-look-like-this:**
Screenshot (18)


wozeparrot:
I don't think that this is something that we really want, I don't see the problem with using `Tensor.scaled_dot_product_attention` directly. Having a class for this is just less flexible.

PR #2305, opened 2023-11-15
make self referential realize fast

DKormann:
continuation of #2278 now I want to make realize fast too

### DIFF
--- a/extra/optimization/helpers.py
+++ b/extra/optimization/helpers.py
@@ -8,7 +8,8 @@ inf, nan = float('inf'), float('nan')
 
 # kernel unpacker
 from tinygrad.codegen.linearizer import Linearizer
-def ast_str_to_lin(ast_str): return Linearizer(eval(ast_str))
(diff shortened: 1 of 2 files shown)


geohot:
I have no idea how to even begin reviewing this. Can this be done with a negative line change? Possible to use functools.lru_cache instead of an explicit dictionary?

DKormann:
i was trying. the problem is you don't need to cache the whole function arguments. I could do a couple less lines and possibly rlu_cache buy wrapping the whole ast_parse function. the diff is gonna be little bigger though

DKormann:
the caching itself is gonna add some lines regardless the only way I think to have neg line diff is by reducing lines elsewhere basically what i did last PR

DKormann:
ah wait i got something

DKormann:
I squached a couple simple lines up in the file maybe thats not helpful
I can make the ast_parse itself somewhat smaller but not much

DKormann:
linearizer change: i put all unhashable args into linearizer state. and offs i just pass the index and retrieve the offset when needed in ast_parse

PR #2309, opened 2023-11-15
Remove More `type: ignore`s from the code base

ananddtyagi:
Working on remove as many `type: ignore`s that I can. Relates to: https://github.com/tinygrad/tinygrad/issues/2240

Also fixed and enabled the `--warn_unused_ignores` flag for CI tests.


tinygrad/shape/symbolic.py
```diff
@@ -311,7 +311,7 @@ def __lt__(self, b:Union[Node,int]):
       if muls:
         # NOTE: gcd in python 3.8 takes exactly 2 args
         mul_gcd = b
-        for x in muls: mul_gcd = gcd(mul_gcd, x.b)  # type: ignore  # mypy cannot tell x.b is int here
+        for x in muls: mul_gcd = gcd(mul_gcd, cast(int, x.b)) # mypy cannot tell x.b is int here
```
ananddtyagi, line 314: since we know that x.b will be an int, this cast should be fine

tinygrad/runtime/ops_cuda.py
```diff
@@ -47,17 +47,17 @@ class device:
   RawCUDABuffer = RawMallocBuffer
 else:
   import pycuda.autoprimaryctx # pylint: disable=unused-import # noqa: F401
-  import pycuda.driver as cuda # type: ignore
+  import pycuda.driver as pycuda_driver
```
ananddtyagi, line 50: just renamed this, helps resolve the naming issue causing the type ignore error on line 53-60

tinygrad/features/search.py
```diff
@@ -46,7 +46,7 @@ def time_linearizer(lin:Linearizer, rawbufs:List[RawBuffer], allow_test_size=Tru
 
     # TODO: this is copied from prg.__call__
     global_size, local_size = prg.launch_dims(var_vals)
-    if global_size is not None and local_size is None and all_int(tuple(prg.global_size)): # type: ignore[arg-type]
+    if global_size is not None and local_size is None and all_int(tuple(cast(Iterable[Any], prg.global_size))):
```
geohot, line 49: Don't do things like this. This is worse than a type: ignore

geohot:
Using Any is worse than type: ignore

tinygrad/features/search.py
ananddtyagi: I'll look for a better solution

ananddtyagi:
> Using Any is worse than type: ignore, and cast is also frowned upon.

Got you. I'll revert those and work on better fixes. I looked through the repo to see uses of cast before using it in my changes, is the general guideline just use it as infrequently as possible? Or are there any uses we don't mind casting?

tinygrad/renderer/triton.py
```diff
@@ -41,8 +41,10 @@ def define_scalar(local_size, dtype, args):
 
 def uops_to_triton(function_name:str, uops:List[UOp]):
   local_size: List[int] = []
+  dims: List[str] = []
```
ananddtyagi, line 44: Inferred this type from the data put onto the list on 106. 107 - 112 indicates the type is string.

tinygrad/nn/state.py
```diff
@@ -110,7 +110,7 @@ def persistent_load(self, pid): return pid
     for n in myzip.namelist():
       if n.startswith(f'{base_name}/data/'):
         with myzip.open(n) as myfile:
-          offsets[n.split("/")[-1]] = myfile._orig_compress_start # type: ignore
+          offsets[n.split("/")[-1]] = myfile.tell()
```
ananddtyagi, line 113: need feedback here: 
_orig_compress_start is either None or `.tell()` depending on if the file is seekable. calling `.open` successfully on the line above at 112 sets seekable to true. So, this change should be sufficient, and passes mypy checks. Any thoughts?
https://android.googlesource.com/platform/prebuilts/clang/host/windows-x86.git/+/4a36d349eb0029b267457c75da8ba7a27beea1ed/clang-r433403/python3/Lib/zipfile.py#817

tinygrad/runtime/ops_cuda.py
```diff
@@ -44,20 +44,21 @@ class device:
...
   CUDAAlloc = CUDAAllocator()
-  class RawCUDABuffer(RawBufferCopyInOut): # type: ignore
+  class RawCUDABufferCopyInOut(RawBufferCopyInOut):
     def __init__(self, size, dtype): super().__init__(size, dtype, allocator=CUDAAlloc)
-    def _copyin(self, x:np.ndarray, stream:Optional[cuda.Stream]=None): cuda.memcpy_htod_async(self._buf, x.ravel(), stream) # type: ignore
-    def _copyout(self, x:np.ndarray): cuda.memcpy_dtoh(x, self._buf) # type: ignore
+    def _copyin(self, x:np.ndarray, stream:Optional[pycuda_driver.Stream]=None): pycuda_driver.memcpy_htod_async(self._buf, x.ravel(), stream)
+    def _copyout(self, x:np.ndarray): pycuda_driver.memcpy_dtoh(x, self._buf)
+
+RawCUDABuffer = RawMallocBuffer if getenv("CUDACPU", 0) == 1 else RawCUDABufferCopyInOut
```
ananddtyagi, line 61: need feedback here:
This change removes all the type: ignores from this file but it comes at the cost of calling the getenv for CUDACPU again. We could alternatively store the value at the start, do this call once and use that stored boolean value twice. Or, we could live with the `# type: ignore`s if that is fine for this case.

geohot: two calls is fine, but `RawCUDABufferCopyInOut` is a bad name. think about what it is

tinygrad/runtime/ops_cuda.py
geohot: yea `cuda` was a bad name

geohot:
Extra and examples shouldn't be checked or changed.

tinygrad/runtime/ops_cuda.py
```diff
@@ -44,20 +44,21 @@ class device:
...
   class CUDAAllocator(LRUAllocator):
     def __init__(self): super().__init__(self._get_cur_free_space(None))
-    def _do_alloc(self, size, dtype, device, **kwargs): return cuda.mem_alloc(size * dtype.itemsize) # type: ignore
+    def _do_alloc(self, size, dtype, device, **kwargs): return pycuda_driver.mem_alloc(size * dtype.itemsize)
     def _cached_bufkey(self, size, dtype, device): return (device, size*dtype.itemsize) # Buffers of the same length could be reused, no matter what dtype.
-    def _get_cur_free_space(self, device): return cuda.mem_get_info()[0] # type: ignore
+    def _get_cur_free_space(self, device): return pycuda_driver.mem_get_info()[0]
   CUDAAlloc = CUDAAllocator()
-  class RawCUDABuffer(RawBufferCopyInOut): # type: ignore
+  class CUDAMemoryBuffer(RawBufferCopyInOut):
```
ananddtyagi, line 56: I updated this with a slightly better name. We could call this something like PyCUDABuffer, but not sure if that's really a better name.

tinygrad/runtime/ops_cuda.py
```diff
@@ -44,20 +44,21 @@ class device:
...
-  import pycuda.driver as cuda # type: ignore
+  import pycuda.driver as pycuda_driver
   class CUDAAllocator(LRUAllocator):
     def __init__(self): super().__init__(self._get_cur_free_space(None))
-    def _do_alloc(self, size, dtype, device, **kwargs): return cuda.mem_alloc(size * dtype.itemsize) # type: ignore
+    def _do_alloc(self, size, dtype, device, **kwargs): return pycuda_driver.mem_alloc(size * dtype.itemsize)
     def _cached_bufkey(self, size, dtype, device): return (device, size*dtype.itemsize) # Buffers of the same length could be reused, no matter what dtype.
-    def _get_cur_free_space(self, device): return cuda.mem_get_info()[0] # type: ignore
+    def _get_cur_free_space(self, device): return pycuda_driver.mem_get_info()[0]
   CUDAAlloc = CUDAAllocator()
```
ananddtyagi: Is this line necessary? Why can't we just directly instantiate it on line 57? Saves a line from the count

tinygrad/renderer/triton.py
```diff
@@ -41,8 +41,10 @@ def define_scalar(local_size, dtype, args):
...
+  valid: List[str] = []
   depth = 1
-  signatures, dims, bufs, kernel, valid = [], [], [], [], [] #type: ignore
+  signatures, bufs, kernel = [], [], []
```
ananddtyagi, line 47: Would it be considered unnecessary to reduce these into one line separated by `;`? It would supposedly reduce the number of lines (https://discord.com/channels/1068976834382925865/1068991125353939066/1174459427559067770) and keep it readable, since these are just default, base initializations.

wozeparrot: semicolons shouldn't be used imo, its not a real line reduction, just code golfing.

ananddtyagi:
Is there anything else I should add to this PR? It doesn't address all remaining type: ignores, but the ones remaining in the code base require a potentially larger refractor since we want to avoid casting where we can. Most of these are the straightforward ones that were easily updatable.

tinygrad/runtime/ops_cuda.py
```diff
@@ -44,20 +44,20 @@ class device:
...
     def _cached_bufkey(self, size, dtype, device): return (device, size*dtype.itemsize) # Buffers of the same length could be reused, no matter what dtype.
-    def _get_cur_free_space(self, device): return cuda.mem_get_info()[0] # type: ignore
-  CUDAAlloc = CUDAAllocator()
-  class RawCUDABuffer(RawBufferCopyInOut): # type: ignore
-    def __init__(self, size, dtype): super().__init__(size, dtype, allocator=CUDAAlloc)
-    def _copyin(self, x:np.ndarray, stream:Optional[cuda.Stream]=None): cuda.memcpy_htod_async(self._buf, x.ravel(), stream) # type: ignore
-    def _copyout(self, x:np.ndarray): cuda.memcpy_dtoh(x, self._buf) # type: ignore
+    def _get_cur_free_space(self, device): return pycuda_driver.mem_get_info()[0]
+  class CUDAMemoryBuffer(RawBufferCopyInOut):
+    def __init__(self, size, dtype): super().__init__(size, dtype, allocator=CUDAAllocator())
```
geohot, line 56: I'm going to close this PR. Deleting `CUDAAlloc = CUDAAllocator()` and reallocating `CUDAAllocator()` is a serious bug

ananddtyagi:
Whoops, I did not mean to delete the branch. That edit was a mistake on my part. I was trying to understand its implementation and see if anything would break with that change but didn't fully understand the repercussions of that. I can remove that change and create a new PR if that works. Also, I'll take some more time to understand the codebase better before taking on another issue.

PR #2323, opened 2023-11-15
Searchable tensor cores

chaosagent:
Adds searchable tensor cores. The approach is to mark some dimensions as SIMD dimensions, and prevent them from being touched. I believe this is not privileging TC, and is general for all SIMD.

The upcast indexing calculations for TC were incorrect. I redid them to use ast_parse mechanisms as much as possible. The plan is to push down the stack until WMMA is just a part of ast_parse.

A small test suite for WMMA is included. I tested on HIP before and after (there is currently a bug with casting that beaks WMMA, cherry-pick the last two commits from https://github.com/chaosagent/tinygrad/tree/simdlocal_push_pre to make it work). I don't have a metal device, so I can't test on METAL.

The search results are OK, maybe a few tens of percent gain on some kernels. Nothing drastic.

flammit had TC be an action in the space. Should be simple to add, maybe next PR...

WINO=1 HALF=1 HIP=1 hlb_cifar is same speed as on master with no beam.

**I have arranged the commits for this PR to be individually reviewable. Hope this helps.**

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -331,7 +331,7 @@ class Kernel:
 
   # ******************** high level optimizers ********************
 
-  def apply_tensor_cores(self, use_tensor_cores=1, extra_opts:Optional[List[Opt]]=None):
+  def apply_tensor_cores(self, use_tensor_cores=1, hand_coded=True):
     if use_tensor_cores and self.opts.has_local and self.reduceop and self.reduceop.op == ReduceOps.SUM and self.opts.device in tensor_cores:
       for tc in tensor_cores[self.opts.device]:
         if not((tc.arch is None or tc.arch == os.uname().machine) and isinstance(self.reduceop.src[0], LazyOp)): continue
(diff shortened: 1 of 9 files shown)


chaosagent:
image
looks weird, dunno whats causing this

tinygrad/codegen/kernel.py
```diff
@@ -95,8 +95,11 @@ def __init__(self, ast:LazyOp, opts:Optional[LinearizerOptions]=None):
     self.group_for_reduce: List[int] = []
     self.upcasted: int = 0
     self.local_dims: int = 0
+    self.simd_local_dims: int = 0
```
geohot, line 98: I really don't want to add more options here. Why do these have to be special?

chaosagent: We could have TC infer in linearizer. 🤔

chaosagent: This might not actually solve anything, but would it be better if this were just read from the tensor_cores? Tensor_cores is the only simd we have, I guess.

I need to think about other solutions. The thought here is: either we want to force the simd dimensions to stay in place, or we don't. If we do, then we might as well make it explicit. If we don't, then we need to check that the right conditions exist in linearizer. Many tradeoffs.

Let me know if you have some good idea.

geohot:
This is adding a whole bunch of lines. I think the way to start this is to first make tensor cores searchable (aka adding them to the Opt space), even without searching beyond them.

tinygrad/codegen/linearizer.py
```diff
@@ -272,31 +284,37 @@ def calc_tc_idxs(local_size: int, aliases: List[List[int]]):
...
+          tc_views.append(ShapeTracker((View.create(tuple(tc_upcast_shape), tuple(localbuf_strides)),)))
         if DEBUG >= 3: print(f"{localbuf_idx} alias {i}: idxs=", buf_idxs)
         ll = self.global_load(i, buf_idxs)
         locals_to_store.append((localbuf_idx, buf_idxs, ll))
 
       # copy in any global buffers
       if self.tensor_core:
+        # reconstruct upcast pattern for wmma
+        def rename_var(v: VariableOrNum, expr: str): return v if isinstance(v, NumNode) else Variable(expr, v.min, v.max)
+        tc_upcast_idxs = tuple(full_upcast_idxs[self.simd_upcasted:] if self.upcasted > self.simd_upcasted else (Variable(None, 0, 1),))
```
chaosagent, line 301: i think this should be variable(None, 0, 0)

chaosagent:
tc search being done in #3275

PR #2324, opened 2023-11-16
Use pyproject.toml instead of setup.py

croumegous:
setup.py is deprecated. The new standard is now pyproject.toml (PEP 621)


geohot:
There's been multiple of these PRs, and they all have the same issue.

Does it support Ubuntu 20.04 out of the box?

croumegous:
The issue is that ubuntu 20.04 comes with pip version 20.0.2 (january 2020) and PEP 621 was made in June 2020. 
So the default pip doesn't work properly. 
The only fix is as easy has `python -m pip install -U pip` (ofc I tested it with ubuntu 20.04)

---
Also the CI was failing because wgpu-py renamed the file that we used to import in the today latest release v0.12.0 (from [_device](https://github.com/pygfx/wgpu-py/blob/v0.11.0/wgpu/utils/_device.py) to [device](https://github.com/pygfx/wgpu-py/blob/v0.12.0/wgpu/utils/device.py)
Fixed the import path

geohot:
Thanks for root causing 20.04 and why it can't work out of the box.

I don't feel it's worth changing until 24.04 is out, comma is still on 20.04

PR #2326, opened 2023-11-16
cache ast_parse

DKormann:
minor refactor + cached ast_parse gonna lead to exp speedup for #2305 with one more cache in map_buffers


roelofvandijk:
Have you measured the speed, for example for the llama example?

DKormann:
so far I am measuring speed only for stuff like:

for i in range(100): x = x + x 
x.realize()

but i suspect that there are going to be speedups for Transformer models too although only exponential with number of attention layers. Also this speedup is only relevant for building and compiling ast so unrelated to performance after JIT.

roelofvandijk:
Only after the hash override, I think.

geohot:
Why is this a draft? Do we want it?

DKormann:
Working at rnnt atm l. Gonna come back to it

PR #2328, opened 2023-11-16
Cache map buffers

DKormann:
based on #2326 would finish #2305


geohot:
Can you show an example of what this does? Is there a simpler way to write it?

geohot:
Add lines, tests failing, and I don't see the gains. If this can be clean, feel free to reopen

PR #2337, opened 2023-11-17
Assert for memory allocation failures

imaolo:
- If the aging buffers queue is emptied, ensure_has_free_space() may return without ensuring space_to_free is available.
- [Metal buffers](https://developer.apple.com/documentation/metal/mtldevice/1433375-newbufferwithlength) have a max length and may return None on allocation failures. Error immediately if the max is exceeded or None is returned.

Not catching these can cause hard to track downstream issues.

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -11,7 +11,7 @@ from tinygrad.shape.symbolic import Variable, Node
 
 class MetalAllocator(LRUAllocator):
   def _do_alloc(self, size, dtype, device, **kwargs):
-    assert (buf_len := size*dtype.itemsize) < (max_len := METAL.device.maxBufferLength()) and (buf := METAL.device.newBufferWithLength_options_(buf_len, Metal.MTLResourceStorageModeShared)), f"Failed to create a {buf_len/1e9:5.2f} GB Metal buffer. {f'The max buffer length is {max_len/1e9:5.2f} GB.' if buf_len > max_len else 'Metal returned None {buf}.'}"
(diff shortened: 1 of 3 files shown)


tinygrad/runtime/ops_metal.py
```diff
@@ -10,7 +10,9 @@
 from tinygrad.shape.symbolic import Variable, Node
 
 class MetalAllocator(LRUAllocator):
-  def _do_alloc(self, size, dtype, device, **kwargs): return METAL.device.newBufferWithLength_options_(size*dtype.itemsize, Metal.MTLResourceStorageModeShared)
+  def _do_alloc(self, size, dtype, device, **kwargs):
+    assert (buf_len := size*dtype.itemsize) < (max_len := METAL.device.maxBufferLength()) and (buf := METAL.device.newBufferWithLength_options_(buf_len, Metal.MTLResourceStorageModeShared)), f"Failed to create a {buf_len/1e9:5.2f} GB Metal buffer. {f'The max buffer length is {max_len/1e9:5.2f} GB.' if buf_len > max_len else f'Metal returned {buf}.'}"
```
geohot, line 14: Please use two asserts!

geohot:
Otherwise thank you for fixing this! I had stable diffusion just be wrong

tinygrad/runtime/ops_metal.py
chenyuxyz: also don't use `:=` in assert, assert should not have side effects

imaolo: Are they okay for use within the assert messages? I see one example in helpers.py:30.

chenyuxyz: i think if it's just for the assertion message and it reads better it's fine. here it's used outside of assertion

PR #2349, opened 2023-11-18
fix: retain or promote dtype for Unaryops.NEG

roelofvandijk:
The linearizer did not retain the dtype on `UnaryOps.NEG`, which meant the linearizer did not re-add a cast for `MAX` ops.

```
#include <metal_stdlib>
using namespace metal;
kernel void E_4_32_4(device int* data0, const device float* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 4 */
  int lidx1 = lid.x; /* 32 */
  float4 val0 = (float4)(*((device float4*)(data1+(gidx0*128)+(lidx1*4))));
  int cast0 = (int)(((val0).x*60000.0f));
  int cast1 = (int)(((val0).y*60000.0f));
  int cast2 = (int)(((val0).z*60000.0f));
  int cast3 = (int)(((val0).w*60000.0f));
  float alu0 = max((float)(cast0),0.0f);
  float alu1 = max((float)(cast1),0.0f);
  float alu2 = max((float)(cast2),0.0f);
  float alu3 = max((float)(cast3),0.0f);
  float alu4 = max((-cast0),0.0f); <---- int vs float
  float alu5 = max((-cast1),0.0f);
  float alu6 = max((-cast2),0.0f);
  float alu7 = max((-cast3),0.0f);
  *(data0+(gidx0*128)+(lidx1*4)) = (cast0/(alu0+alu4+1e-10f));
  *(data0+(gidx0*128)+(lidx1*4)+1) = (cast1/(alu1+alu5+1e-10f));
  *(data0+(gidx0*128)+(lidx1*4)+2) = (cast2/(alu2+alu6+1e-10f));
  *(data0+(gidx0*128)+(lidx1*4)+3) = (cast3/(alu3+alu7+1e-10f));
}
```

Causing ambiguous behavior:

```
METAL_FUNC int max(int x, int y)
               ^
program_source:16:16: error: call to 'max' is ambiguous
  float alu5 = max((-cast1),0.0f);
```

This MR allows the linearizer to retain the dtype for `UnaryOps.NEG`, which allows the example to run without `@TinyJit`, and the sample generation can be pulled in.

I also added type unsigned -> signed type promotion on negation.

### DIFF
--- a/examples/beautiful_mnist.py
+++ b/examples/beautiful_mnist.py
@@ -23,12 +23,12 @@ if __name__ == "__main__":
   model = Model()
   opt = nn.optim.Adam(nn.state.get_parameters(model))
 
-  # TODO: there's a compiler error if you comment out TinyJit since randint isn't being realized and there's something weird with int
   @TinyJit
-  def train_step(samples:Tensor) -> Tensor:
+  def train_step() -> Tensor:
     with Tensor.train():
       opt.zero_grad()
       # TODO: this "gather" of samples is very slow. will be under 5s when this is fixed
+      samples = Tensor.randint(512, high=X_train.shape[0])
       loss = model(X_train[samples]).sparse_categorical_crossentropy(Y_train[samples]).backward()
       opt.step()
       return loss.realize()
@@ -39,7 +39,6 @@ if __name__ == "__main__":
   test_acc = float('nan')
   for i in (t:=trange(70)):
     GlobalCounters.reset()   # NOTE: this makes it nice for DEBUG=2 timing
-    samples = Tensor.randint(512, high=X_train.shape[0])  # TODO: put this in the JIT when rand is fixed
-    loss = train_step(samples)
+    loss = train_step()
     if i%10 == 9: test_acc = get_test_acc().item()
     t.set_description(f"loss: {loss.item():6.2f} test_accuracy: {test_acc:5.2f}%")
--- a/examples/beautiful_mnist.py
+++ b/examples/beautiful_mnist.py
@@ -24,11 +24,10 @@ if __name__ == "__main__":
   opt = nn.optim.Adam(nn.state.get_parameters(model))
 
   @TinyJit
-  def train_step() -> Tensor:
+  def train_step(samples:Tensor) -> Tensor:
     with Tensor.train():
       opt.zero_grad()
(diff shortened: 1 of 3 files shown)


Qazalin:
I think #2264 sorta solved the same problem but not explicitly, what i did there is I just cast all the vins of an ALU to the same dtype, so no matter what the op was before they always get promoted correctly

examples/beautiful_mnist.py
```diff
@@ -23,12 +23,12 @@ def __call__(self, x:Tensor) -> Tensor: return x.sequential(self.layers)
...
   opt = nn.optim.Adam(nn.state.get_parameters(model))
 
-  # TODO: there's a compiler error if you comment out TinyJit since randint isn't being realized and there's something weird with int
   @TinyJit
-  def train_step(samples:Tensor) -> Tensor:
+  def train_step() -> Tensor:
     with Tensor.train():
       opt.zero_grad()
       # TODO: this "gather" of samples is very slow. will be under 5s when this is fixed
+      samples = Tensor.randint(512, high=X_train.shape[0])
```
geohot, line 31: Did you test this?

roelofvandijk: Yes. Works on M2 Mac Pro.

geohot: What accuracy did you get?

roelofvandijk: Samples inside of train_step: 94.99% with `@TinyJit`, 98.16% without. 
Samples outside of train_step: 98.21% with `@TinyJit`, 98.24% without.

Jit is around 15% faster.

So something going on with the jit + random, will revert that change.

geohot:
"I also added type unsigned -> signed type promotion on negation."

Is this needed? Please don't do two things in one PR.

If you want to fix a bug, apply the minimal fix and include a regression test

roelofvandijk:
Yes, it is needed - the behaviour would otherwise still be wrong for ints. 

To me, this is the same thing: correctly handling types on negation.
* keep the same type for floats instead of always going to `float32`
* promote to unsigned int if signed ints are negated

I can consider adding a test.

PR #2350, opened 2023-11-18
More Typing

eichenroth:
- added a bunch of types to `kernel`, `helpers`, `shapetracker` & `symbolic`
- `shapetracker` close to 100% typed now


tinygrad/shape/symbolic.py
```diff
@@ -332,10 +332,10 @@ def create_rednode(typ:Type[RedNode], nodes:List[Node]):
...
 
 @functools.lru_cache(maxsize=None)
-def sym_rename(s) -> str: return f"s{sym_rename.cache_info().currsize}"
-def sym_render(a: Union[Node, int], ops=None, ctx=None) -> str: return str(a) if isinstance(a, int) else a.render(ops, ctx)
-def sym_infer(a: Union[Node, int], var_vals: Dict[Variable, int]) -> int:
-  if isinstance(a, (int, float)): return a
+def sym_rename(s:sint) -> str: return f"s{sym_rename.cache_info().currsize}"
+def sym_render(a:sint, ops=None, ctx=None) -> str: return str(a) if isinstance(a, int) else a.render(ops, ctx)
+def sym_infer(a:sint, var_vals: Dict[Variable, int]) -> int:
+  if isinstance(a, int): return a
```
geohot, line 338: Err...what if it's a float?

geohot: Can you add a test to catch this regression? Look at the NOLOCALS stuff for when it can be a float

eichenroth: You are right, if it is a float it will break. Also it would mean, the types were wrong before.

I will revert this, add a todo and and create a separate PR.

tinygrad/shape/shapetracker.py
```diff
@@ -112,7 +112,7 @@ def to_movement_ops(self) -> List[Tuple[MovementOps, Tuple]]:
   # NOTE: if a stride is not always valid, it will be None
   def real_strides(self, ignore_valid=False) -> Tuple[Optional[sint], ...]:
     if len(self.views) == 1 and self.views[-1].mask is None: return self.views[-1].strides
-    idxs: List[Node] = [Variable(f"idx{i}", 0, s-1) for i,s in enumerate(self.shape)]
+    idxs = tuple(Variable(f"idx{i}", 0, s-1) for i,s in enumerate(self.shape))
```
geohot, line 115: Are you 100% sure this doesn't change behavior?

geohot:
I can't merge this. Please separate typing/whitespace from any changes that *in any way* can alter behavior. The risk of merging isn't worth it.

PR #2352, opened 2023-11-18
simplify Node.sum, remove factorize method

chenyuxyz:


roelofvandijk:
What's the speed impact?

chenyuxyz:
i tested non-jitted llama and wino cifar, no noticeable change. conceptually i think it simpler and faster because one less pass, did not micro benchmark thou.

PR #2360, opened 2023-11-19
remove whisper +1-1 hack

chenyuxyz:
@mmmkkaaayy can you verify if +1-1 is still needed? would like to remove this hack. i run the whisper locally and the tests also passed

i think the test case you put is bugged due to the optional input `cache_v`, and having jitted and non-jitted function to mutate the state based on different input. i also had some issues in llama chatbot mode. if i always put `cache_v` the issue disappear

btw the whisper code is pretty clean and good job getting symbolic kvcache jit worked!

### DIFF
--- a/examples/whisper.py
+++ b/examples/whisper.py
@@ -34,7 +34,7 @@ class MultiHeadAttention:
         if not hasattr(self, 'cache_k'):
           self.cache_k, self.cache_v = k, v
         else:
-          # see test_jitted_read_assign in test_jit.py
+          # see test_jitted_read_assign in test_jit.py. more context https://github.com/tinygrad/tinygrad/pull/2360#issuecomment-1817989994
           self.cache_k.assign(k+1-1).realize()
           self.cache_v.assign(v+1-1).realize()
       else:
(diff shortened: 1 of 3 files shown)


mmmkkaaayy:
>i run the whisper locally and the tests also passed

The way this issue manifests in whisper is that jitted decode will read only the first ever cached cross_attn_kv. I would have expected test_whisper.py to fail since there is more than one test case, but I see it passes by coincidence - `test_transcribe_file` is successfully reusing the cross attn from `test_transcribe_batch`. If you reorder the array on line 32 of test_whisper.py (and the corresponding assertions), then we can cause `python3 -m pytest test/models/test_whisper.py` to fail after removing the hack.

> i think the test case you put is bugged due to the optional input cache_v, and having jitted and non-jitted function to mutate the state based on different input. i also had some issues in llama chatbot mode. if i always put cache_v the issue disappear

Always passing `cache_v` changes the semantics of the jitted function - to update the cache and immediately read it. What if I don't want to update the cache on every jitted call? In whisper, we populate the cross attn cache with decoder.blocks_start_tok and then expect to read it in decoder.blocks_after_start_tok

>btw the whisper code is pretty clean and good job getting symbolic kvcache jit worked!

Thanks! Couldn't have done it without your paved path in LLaMa/GPT2.

chenyuxyz:
thanks for your reply. i have updated the tests so that removing +1-1 fails (it would have passed if we do setUp per test, not for the whole class, but i am not sure if that's your intention). will think about this issue more later

PR #2365, opened 2023-11-19
Pretty multinomial

chenyuxyz:
p, cdf_normalized -> weight, cdf
symmetric unsqueeze / squeeze
test cases to check num_sample > 0 and ndim != 0

TODO: how do we want to handle 0/0 in general?


chenyuxyz:
oh no this is one more kernel

PR #2366, opened 2023-11-19
Refactor to use fetch helper

francislata:
This PR refactors the codebase to use the new `fetch` helper function introduced in #2363.

~In addition, I have added the ability to specify an extension when using the `fetch()` helper function.~

Also, I have removed some of the download fetching helper functions in `extra/utils.py` as well as the associated test suite.


examples/whisper.py
```diff
@@ -249,9 +245,7 @@ def listener(q):
 def init_whisper(model_name="tiny.en", batch_size=1):
   assert MODEL_URLS[model_name] is not None
 
-  filename = BASE / "whisper-{}.pt".format(model_name)
```
mmmkkaaayy, line 252: The `whisper-` prefix was used during CI caching

francislata: yeah good call. i'm definitely going to go through all of these again just to make sure I don't break any existing caching, etc.

francislata: I updated the `workflow.yml` file to look at the new cache paths now.

.github/workflows/test.yml
```diff
@@ -85,8 +85,8 @@ jobs:
       uses: actions/cache@v3
       with:
         path: |
-          weights/whisper-tiny.en.pt
-          weights/gpt2.tiktoken
+          ~/.cache/tinygrad/downloads/3a2cae277fcec4464349df535729caaf
```
geohot, line 88: I don't like this with the hashes, it's hard to read. The whole downloads folder should probably be cached

tinygrad/helpers.py
```diff
@@ -36,6 +38,17 @@ def partition(lst:List[T], fxn:Callable[[T],bool]):
   b:List[T] = []
   for s in lst: (a if fxn(s) else b).append(s)
   return a,b
+def temp(x:str) -> str: return (Path(tempfile.gettempdir()) / x).as_posix()
+def get_child(parent, key):
```
geohot, line 42: Can this be a separate PR?

tinygrad/helpers.py
```diff
@@ -223,8 +236,9 @@ def wrapper(*args, **kwargs) -> bytes:
 
 # *** http support ***
 
-def fetch(url:str) -> pathlib.Path:
+def fetch(url:str, ext:Optional[str]=None) -> pathlib.Path:
```
geohot, line 239: Why do I need an extension?

.github/workflows/test.yml
```diff
@@ -85,8 +85,8 @@ jobs:
...
+          ~/.cache/tinygrad/downloads/5386a662f5b1245493601b29a5e61f40
         key: model-weights-v1
```
mmmkkaaayy, line 90: Should update the key as well

tinygrad/helpers.py
francislata: The `test/external/external_test_yolov8.py` was giving issues complaining that there's no ".pt" extension. I'll update the test to just rename the cache directly to have the extension.

tinygrad/helpers.py
francislata: Will do this in a separate PR. I'll link the PR number to this comment.

.github/workflows/test.yml
```diff
@@ -85,9 +85,8 @@ jobs:
       uses: actions/cache@v3
       with:
         path: |
```
geohot, line 87: Clean this up. Why does it have the pipe? It is only caching model weights?

examples/compile_efficientnet.py
```diff
@@ -21,10 +20,12 @@
   else:
     cprog = [prg]
     # image library!
-    cprog += ["#define STB_IMAGE_IMPLEMENTATION", fetch("https://raw.githubusercontent.com/nothings/stb/master/stb_image.h").decode('utf-8').replace("half", "_half")]
+    with open(fetch("https://raw.githubusercontent.com/nothings/stb/master/stb_image.h"), "rb") as f:
```
geohot, line 23: pathlib has a read method afaik

examples/efficientnet.py
```diff
@@ -61,7 +60,8 @@ def infer(model, img):
   model.load_from_pretrained()
 
   # category labels
-  lbls = fetch("https://gist.githubusercontent.com/yrevar/942d3a0ac09ec9e5eb3a/raw/238f720ff059c1f82f368259d1ca4ffa5dd8f9f5/imagenet1000_clsidx_to_labels.txt")
+  with open(fetch("https://gist.githubusercontent.com/yrevar/942d3a0ac09ec9e5eb3a/raw/238f720ff059c1f82f368259d1ca4ffa5dd8f9f5/imagenet1000_clsidx_to_labels.txt"), "rb") as f:
```
geohot, line 63: same as above, this is more unneeded lines

tinygrad/helpers.py
```diff
@@ -13,6 +13,7 @@ def prod(x:Iterable[T]) -> Union[T,int]: return functools.reduce(operator.__mul_
 
 # NOTE: helpers is not allowed to import from anything else in tinygrad
 OSX = platform.system() == "Darwin"
+WINDOWS = platform.system() == "Windows"
```
geohot, line 16: Why is this here?

geohot:
This PR is way too large to review. Want to start with only the whisper cache?

francislata:
> This PR is way too large to review. Want to start with only the whisper cache?

Yeah, sounds good. I'll separate these out into smaller PRs.

PR #2369, opened 2023-11-20
remove file writing in metal ops

ductm104:


geohot:
Woah, this fixes the bug for people? Looking like both a line reduction and a bounty

Issue #2226, opened 2023-11-06
Metal gives Error Domain=MTLLibraryErrorDomain Code=1 "MTLLibrary is not formatted as a MetalLib file"


JChunX:
Hi, I am going through the tinygrad documentation and trying to run the basic Tensor example on `abstractions.py`

Environment:
- M1 Mac, Sonoma 14.0, 
- Python 3.10.13
- tinygrad source install Nov 6th

`try_tensor.py:`
```python
from tinygrad.tensor import Tensor
a = Tensor([2])
b = Tensor([3])
result = a + b
print(f"{a.numpy()} + {b.numpy()} = {result.numpy()}")
```

However I am getting an error:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tinygrad/tinygrad/tensor.py", line 126, in numpy
    return self.detach().cast(dtypes.from_np(self.dtype.np)).contiguous().to('CPU').realize().lazydata.realized.toCPU().reshape(self.shape)
  File "/tinygrad/tinygrad/tensor.py", line 105, in realize
    run_schedule(self.lazydata.schedule())
  File "/tinygrad/tinygrad/realize.py", line 27, in run_schedule
    si.out.realized = Device[si.out.device].exec_ast(si.ast, output=si.out, inputs=si.inputs, var_vals=si.var_vals, **si.out._device_extra_args())
  File "/tinygrad/tinygrad/ops.py", line 292, in exec_ast
    if ast not in self.method_cache: self.method_cache[ast] = get_program()
  File "/tinygrad/tinygrad/ops.py", line 289, in get_program
    return self.to_program(k)
  File "/tinygrad/tinygrad/ops.py", line 239, in to_program
    display_name=k.display_name, runtime_args=runtime_args).build(self.compiler, self.runtime)
  File "/tinygrad/tinygrad/ops.py", line 193, in build
    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") else compiler(self.prg)
  File "/tinygrad/tinygrad/helpers.py", line 206, in wrapper
    return diskcache_put(table, key, func(*args, **kwargs))
  File "/tinygrad/tinygrad/runtime/ops_metal.py", line 51, in compile_metal
    unwrap(library.serializeToURL_error_(Cocoa.NSURL.URLWithString_(f"file://{output_file.name}"), None))
[...]

PR #2372, opened 2023-11-20
WIP: directly use Metallib instance

nullhook:
dont use unexposed api and directly use `MetalLibrary`

Metal.framework will auto cache shader IR on disk for you at location:
`$(getconf DARWIN_USER_CACHE_DIR)/com.python.python/com.apple.metal`

**todo:**

- [x] include benchmark info with/without cache
- [ ] investigate why MTLLibrary is being created twice

**benchmarks:**

Initial run with no cache:

image

Subsequent run with cache:

image

Generating the MTLLibrary is the most expensive part of compilation in Metal

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -251,7 +251,7 @@ class CompiledASTRunner(ASTRunner):
     super().__init__(ast)
 
   def build(self, compiler, runtime):
-    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") else compiler(self.prg)
+    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") or getenv("METAL") else compiler(self.prg)
     self.clprg = runtime(self.name, self.lib)
     return self
 
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -251,7 +251,7 @@ class CompiledASTRunner(ASTRunner):
     super().__init__(ast)
 
   def build(self, compiler, runtime):
-    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") or getenv("METAL") or Device.DEFAULT is "METAL" else compiler(self.prg)
+    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") or getenv("METAL") or Device.DEFAULT == "METAL" or getenv("METAL_XCODE") else compiler(self.prg)
     self.clprg = runtime(self.name, self.lib)
     return self
 
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -251,7 +251,7 @@ class CompiledASTRunner(ASTRunner):
     super().__init__(ast)
 
   def build(self, compiler, runtime):
-    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") or getenv("METAL") else compiler(self.prg)
+    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") or getenv("METAL") or Device.DEFAULT is "METAL" else compiler(self.prg)
     self.clprg = runtime(self.name, self.lib)
     return self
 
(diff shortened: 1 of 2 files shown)


tinygrad/ops.py
```diff
@@ -251,7 +251,7 @@ def __init__(self, ast:Optional[LazyOp], name:str, prg:str, global_size:Optional
     super().__init__(ast)
 
   def build(self, compiler, runtime):
-    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") else compiler(self.prg)
+    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") or getenv("METAL") or Device.DEFAULT is "METAL" else compiler(self.prg)
```
alexbaden, line 254: What about if `use_xcode` is true?

alexbaden:
As I said in #2226 this resolves the issues exporting metal compiled shader code on my m1 air on macOS 14.1.1.

tinygrad/ops.py
JChunX, line 1: Minor thing: importing `Tensor` on Python 3.10 gives `/tinygrad/tinygrad/ops.py:254: SyntaxWarning: "is" with a literal. Did you mean "=="?`

Can we swap `is` for `==`?

tinygrad/ops.py
nullhook: added

tinygrad/ops.py
nullhook: fixed

tinygrad/runtime/ops_metal.py
```diff
@@ -41,22 +41,21 @@ def unwrap(x):
   return ret
 
 @diskcache
-def compile_metal(prg, use_xcode=bool(getenv("METAL_XCODE"))) -> bytes:
+def compile_metal(prg, use_xcode=bool(getenv("METAL_XCODE"))):
```
wozeparrot, line 44: Hmm, my issue with this is that all the compile functions have had a very clear function signature, `(str) -> bytes`, this breaks that.

tinygrad/runtime/ops_metal.py
```diff
@@ -41,22 +41,21 @@ def unwrap(x):
   return ret
 
 @diskcache
```
nullhook, line 43: @wozeparrot probably don't need this for metal_ops? because we're letting Metal.framework take care of the cache.

tinygrad/runtime/ops_metal.py
nullhook: is there a specialization equivalent here?

a-arbabian:
@nullhook I see this branch has diverged from master so can't run any examples from it, getting `ModuleNotFoundError: No module named 'tinygrad.runtime.interpreted'`

Was the function signature issue in `compile_metal()` the only thing preventing this from being merged? I can't run anything on Metal with master right now so trying to revive this branch

10-zin:
yes I think the compile_metal() does not follow the str->bytes signature as @wozeparrot mentioned. I am also not able to run anything on metal for M1 Sonoma 14.1.1 
Trying to patch this with master on my local, but no luck in passing the tests yet.

geohot:
Can someone give me ssh into a machine that doesn't work?

tylercosgrove:
not sure whether this is still being worked on, but for those still facing this issue I had to update this this line: `self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") else compiler(self.prg)` in `device.py` instead of `ops.py` on my local as the branch has diverged since the original pr. I am on M1 Sonoma 14.2

10-zin:
@tylercosgrove wat did u update?

tylercosgrove:
> @tylercosgrove wat did u update?

`self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") else compiler(self.prg)`
becomes
`self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") or getenv("METAL") or Device.DEFAULT == "METAL" or getenv("METAL_XCODE") else compiler(self.prg)`

This is what the pr is for, but this line change is now in `device.py` instead of `ops.py`. The changes in `ops_metal.py` stay the same. This could still be problematic in regards to the `compile_metal` signature, but this is just in case anyone had a problem as a result of the branches diverging (when device moved to device.py in #2466 ).

geohot:
So this can't be merged because it violates the spec, and it's now out of date so I'm going to close. But there's a bounty for fixing this bug in a proper way, if it's still affecting people.

Issue #2228, opened 2023-11-07
tinyllama doesn't output correctly


r3sist-uniq:
Tried adding the [tinyllama](https://github.com/jzhang38/TinyLlama) config through #2170, expanded this config to include max_seq_len.  

`{"dim": 2048, "multiple_of": 256, "n_heads": 32, "n_kv_heads": 4,"max_seq_len": 2048,  "n_layers": 22, "norm_eps": 1e-05, "vocab_size": 32000}`

The tinyllama config.json file could be found [here](https://huggingface.co/PY007/TinyLlama-1.1B-intermediate-step-240k-503b/blob/main/config.json). 

The weight loading is being done properly, I confirmed this by manually loading weights through torch, and also np.asserting after the weight loading. 

The outputs (especially with temp=0) are not right. The huggingface inference seems to produce way better outputs. Couldn't find the bug, it seems to be subtle, most probably in the [llama.py](https://github.com/tinygrad/tinygrad/blob/master/examples/llama.py) implementation. 

Outputs from tinygrad with `CUDA=1 python3 ./examples/llama.py --count=50  --temperature=0.0 --prompt="where do you live?" --size="1B" --gen=tiny --model='./model.safetensors'` (can use this command after updating the config dictionary and LLAMA_SUFFIX inside llama.py)

![image](https://github.com/tinygrad/tinygrad/assets/72573738/6a6c9b75-7abf-4475-9731-2aad957e098c)

It could be related to post processing, but not sure about that. I tried comparing the outputs of each transformer block (there are 22 in tinyllama) from tinygrad and huggingface (torch). Using just the main `Transformer` class in llama.py. And Found the following 

1. The embedding (tok_embeddings) produces the same outputs, from both tinygrad and huggingface. This agains confirm, atleast the weight loading is being done properly. 
2. The model produces a somewhat better output than above for the same prompt without adding the "bos" token in the beginning. 
3. Suppose the input to the model is 5 tokens, the output of each transformer block in that case would be of shape (1,5,2048). For each transformer block, the outputs for the first token is the same, the rest of the outputs for the 4 tokens are different. And this is without any padding. This may hint the problem could be in the attention computation. As an example, the output of the 3rd transformer block (of shape 1,5,2048) is the same as the huggingface's output at the outputs for the first token (i.e at the index [0][0])
![image](https://github.com/tinygrad/tinygrad/assets/72573738/7215ed49-5e7c-4853-8617-e40898835b28)
[...]

PR #2374, opened 2023-11-20
fixed hf convert and now it's working with tinyllama

SkullMag:
Fixed huggingface convert


SkullMag:
I've compared the outputs with the transformers code:
```python
from transformers import LlamaForCausalLM, LlamaTokenizer

model = LlamaForCausalLM.from_pretrained("tinyllama_503b", local_files_only=True)
tokenizer = LlamaTokenizer.from_pretrained("tinyllama_503b/tokenizer.model")
prompt = "Where do you live?"
inputs = tokenizer(prompt, return_tensors="pt")
generate_ids = model.generate(inputs.input_ids, max_length=30)
print(tokenizer.batch_decode(generate_ids, skip_special_tokens=True, clean_up_tokenization_spaces=False)[0])
```

Here is the transformers output:
```
The 2018-19 season is the 10th season for the club and the
```

And here is the output from tinygrad:
```
python llama.py --gen=tiny --model tinyllama_503b/model.safetensors --size 1B --temperature 0 --count 30 --prompt "Where do you live?"

The 2018-19 season is the 10th season for the club and the 10th season in the
```

SkullMag:
I used the config from https://github.com/tinygrad/tinygrad/issues/2228

examples/llama.py
```diff
@@ -218,6 +218,12 @@ def __call__(self, tokens:Tensor, start_pos:Variable, temperature:float=0.0):
       "args": {"dim": 8192, "n_layers": 48, "n_heads": 64, "n_kv_heads": 8, "multiple_of": 256, "ffn_dim_multiplier": 1.0, "norm_eps": 1e-5, "rope_theta": 1000000, "vocab_size": 32000},
       "files": 4,
     },
+  },
+  "tiny": {
+    "1B": {
+      "args": {"dim": 2048, "multiple_of": 256, "n_heads": 32, "n_kv_heads": 4, "n_layers": 22, "norm_eps": 1e-05, "vocab_size": 32000},
```
chenyuxyz, line 224: nit: can you order this to have the same order as the ones above?

examples/llama.py
```diff
@@ -252,7 +261,21 @@ def convert_from_huggingface(weights, model):
...
-  return {keymap[k]: v for k,v in weights.items() if ".rotary_emb." not in k}
+  sd = {}
+  for k, v in weights.items():
+    if ".rotary_emb." in k:
+      continue
+    v = v.to(Device.DEFAULT)
+    to_name = keymap[k]
+    if "model.layers" in k:
+      if "q_proj" in k:
+        sd[to_name] = permute(v, 32, 2048, 2048)
```
chenyuxyz, line 272: use the params instead of hard coding the numbers

examples/llama.py
SkullMag: done

examples/llama.py
SkullMag: done

examples/llama.py
```diff
@@ -242,7 +248,10 @@ def load(fn:str):
   else:
     return torch_load(fn)
 
-def convert_from_huggingface(weights, model):
+def convert_from_huggingface(weights, model: Transformer, n_heads: int, n_kv_heads: int):
+  def permute(v: Tensor, n_heads: int, shape):
```
chenyuxyz, line 252: do you need `shape` if it's always `v.shape` when called?

examples/llama.py
```diff
@@ -489,7 +513,7 @@ def main(string: str):
 
   # *** prompt engineers stop here ****
 
-  LLAMA_SUFFIX = {"1": "", "2": "-2", "code": "-code"}[args.gen]
+  LLAMA_SUFFIX = {"1": "", "2": "-2", "code": "-code", "tiny": "tiny"}[args.gen]
```
chenyuxyz, line 516: "-tiny"

SkullMag: fixed

geohot:
Can it auto fetch the weights if they aren't there?

SkullMag:
@geohot @chenyuxyz I've added default size evaluation based on model gen and added code for dynamically updating help for easier development in the future. I suggest we add model fetch in separate PR, what do you think?

geohot:
Tested and merged, george@tinygrad.org to claim the bounty, either USDC or PayPal.

PR #2380, opened 2023-11-21
Readable and Faster Union of Vars

ZenginU:
I believe using set.union is more readable, appropriate and bit faster in my local setup.

Note: Set.union(*, set()) is for when argument is [] empty list


geohot:
Careful, does this keep order consistent?

ZenginU:
> Careful, does this keep order consistent?

Not sure if this or the previous one keep the order but Set itself does not preserve the order. 
So do we have cases that vars order is important?
Also, vars are generated as sets in the symbolic.py

PR #2384, opened 2023-11-22
Remove type ignore

ananddtyagi:
Removes some of the `type: ignore` from the code base. Relates to: https://github.com/tinygrad/tinygrad/issues/2240

Also fixed and enabled the --warn_unused_ignores flag for CI tests.


ananddtyagi:
This PR reverts the change made that would have introduced a bug from https://github.com/tinygrad/tinygrad/pull/2309
It includes all the relevant changes that remove some of the type: ignores from the codebase as well as fixing the issue that allows us to enable the --warn_unused_ignores flag for CI tests. 

I'm unsure if I am not supposed to open another PR with these changes, if not, I won't make this ready for review and close it instead.

geohot:
Going to close this, it should be broken up into smaller PRs

PR #2386, opened 2023-11-22
dtypes.float.vec(sz)

Qazalin:
This is a conceptual refactor of how vectorized and scalar dtypes work.
In the current way of dealing with vectorized dtypes:
- There's no clean way to relate `float4` and `float`
- Requires manual defining and duplicating the same priority and sizes
- Has led to duplicates in `render_cast`

This also helps with a cleaner implementation of non-fp32 math

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -91,7 +91,11 @@ class DType(NamedTuple):
   name: str
   np: Optional[type]  # TODO: someday this will be removed with the "remove numpy" project
   sz: int = 1
-  def __repr__(self): return f"dtypes.{INVERSE_DTYPES_DICT[self]}"
+  def __repr__(self): return f"dtypes.{INVERSE_DTYPES_DICT[self]}" if self.sz == 1 else f"dtypes._{INVERSE_DTYPES_DICT[self.scalar()]}{self.sz}"
+  def vec(self, sz:int):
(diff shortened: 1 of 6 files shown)


geohot:
Love this!

tinygrad/helpers.py
```diff
@@ -91,7 +91,11 @@ class DType(NamedTuple):
   name: str
   np: Optional[type]  # TODO: someday this will be removed with the "remove numpy" project
   sz: int = 1
-  def __repr__(self): return f"dtypes.{INVERSE_DTYPES_DICT[self]}"
+  def __repr__(self): return f"dtypes.{INVERSE_DTYPES_DICT[self]}" if self.sz == 1 else f"dtypes._{INVERSE_DTYPES_DICT[self.scalar()]}{self.sz}"
+  def vec(self, sz:int):
+    assert sz > 1 and self.sz == 1, f"can't vectorize {self} with size {sz}"
+    return DType(self.priority, self.itemsize*sz, self.name+str(sz), None, sz)
+  def scalar(self): return DTYPES_DICT[self.name[:-1]] if self.sz > 1 else self
```
geohot, line 98: Err, this is wrong for half16

Qazalin: uh i wish we could run the entire CI, sorry about it:)

PR #2394, opened 2023-11-23
Move HIP render logic to its dedicated place

Qazalin:
there are two non-obvious places for where HIP vectorized dtypes are handled, one in `ops_hip` where we define the types we support and another in render_cast that has dedicated logic for `float8` and `half16` because [HIP doesn't support those ](https://rocm.docs.amd.com/projects/HIP/en/develop/reference/kernel_language.html#short-vector-types) with `make_dtypen`.
These shouldn't be in cstyle, because [some runtimes like OpenCL don't have this problem](https://registry.khronos.org/OpenCL/specs/opencl-1.2.pdf#page=200)


geohot:
Can you add a test to prevent this regression?

tinygrad/renderer/hip.py
```diff
@@ -0,0 +1,42 @@
...
+__device__ void vstore_half(float data, size_t offset, half *p) { *(p + offset) = (half)data; }
+__device__ void vstore_half2(float2 data, size_t offset, half *p) { *(p + offset*2) = (half)data.x; *(p + offset*2 + 1) = (half)data.y; }
+__device__ void vstore_half4(float4 data, size_t offset, half *p) { *(p + offset*4) = (half)data.x; *(p + offset*4 + 1) = (half)data.y; *(p + offset*4 + 2) = (half)data.z; *(p + offset*4 + 3) = (half)data.w; }
+  """
+  gid = [f'blockIdx.{chr(120+i)}' for i in range(3)]
+  lid = [f'threadIdx.{chr(120+i)}' for i in range(3)]
+  xid = [f'(blockIdx.{chr(120+i)}*blockDim.{chr(120+i)}+threadIdx.{chr(120+i)})' for i in range(3)]
+
+  def render_cast(self, x, var_dtype):
+    if var_dtype.sz > 1 and var_dtype.scalar() == dtypes.half: return f"{{{','.join(f'(half){x}' for x in x)}}}"
```
Qazalin, line 38: half is the only floating-point dtype that `make_dtypen` doesn't support: https://docs.amd.com/projects/HIP/en/docs-5.3.0/reference/kernel_language.html#short-vector-types

tinygrad/renderer/hip.py
```diff
@@ -0,0 +1,42 @@
...
+    if var_dtype.sz == 8: return f"{{{','.join(x)}}}"
```
Qazalin, line 39: the max width for any vector dtype in HIP is 4

geohot:
Any reason this is draft, or good to merge?

Qazalin:
yeah i think it's good to merge now

tinygrad/renderer/hip.py
g1y5x3: yeah, but it's also pretty straight forward to implement.
```
struct __align__(8) half4 { half x; half y; half z; half w;};
__device__ half4 make_half4(const half x, const half y, const half z, const half w) { half4 h4; h4.x = x; h4.y = y; h4.z = z; h4.w = w; return h4; }
```
And then in kernel you can just replace the `float4 acc0` with
```
  half4 acc0 = make_half4(0.0f,0.0f,0.0f,0.0f);
```
It worked on the kernels that I have tried.

Qazalin: yeah I was thinking we could do that too - just define make_half4 manually, but then you have to also hardcode half2, half4 and half8 and maybe even half16. there's a bigger problem, you dont know right now what's the max n of a renderer

PR #2395, opened 2023-11-23
toCPU() copies only when jit uses the buffer as output

nimlgen:
Made toCPU() to copy only when the buffer is a jit output.

```
METAL=1 CI=1 python3 test/test_copy_speed.py TestCopySpeed.testCopyDefaulttoCPU

old:
.buffer: 0.07 GB
sync:  6.91 ms @ 9.72 GB/s
sync:  6.87 ms @ 9.77 GB/s
sync:  6.35 ms @ 10.57 GB/s

new:
.buffer: 0.07 GB
sync:  0.33 ms @ 206.15 GB/s
sync:  0.13 ms @ 524.63 GB/s
sync:  0.11 ms @ 619.70 GB/s
```

This should be better in most of cases than juat stupid copy(). I am not sure if we can totally remove .copy() here, since we want jit to not care about allocations, so .toCPU() should guarantee that the buffer is save to use and jit won't overwrite it.


tinygrad/runtime/lib.py
```diff
@@ -37,7 +37,7 @@ class RawBufferMapped(RawBufferCopyIn):
   def _buffer(self) -> memoryview: raise NotImplementedError("must be implemented")
   # NOTE: this metadata prevents the backing buffer from being freed. hack can be removed with PEP688
   def buffer_view(self) -> np.ndarray: return np.frombuffer(self._buffer(), dtype=np.dtype(self.dtype.np, metadata={"backing": self}), count=self.size)  # type: ignore
-  def toCPU(self) -> np.ndarray: return self.buffer_view().copy() # Need a copy, since jit will write to the same buffer.
+  def toCPU(self) -> np.ndarray: return self.buffer_view().copy() if hasattr(self, '_needs_explicit_copy') else self.buffer_view() # JIT sets this to not overwrite buffers holded by CPU.
```
geohot, line 40: The should just shouldn't be here. Can the JIT make the copy?

nimlgen: Yeah, will think on this

PR #2401, opened 2023-11-23
Update Whisper to use fetch helper

francislata:
## Overview
This is just a breakdown of #2366 to smaller PRs. This one starts with Whisper.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,10 +84,8 @@ jobs:
     - name: Cache model weights
       uses: actions/cache@v3
       with:
-        path: |
-          weights/whisper-tiny.en.pt
-          weights/gpt2.tiktoken
-        key: model-weights-v1
+        path: ~/.cache/tinygrad/downloads/
+        key: model-weights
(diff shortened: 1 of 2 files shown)


geohot:
Why is it still called "Cache model weights"? Isn't it caching all the downloads?

francislata:
> Why is it still called "Cache model weights"? Isn't it caching all the downloads?

@geohot - let me update that to "Cache downloads" to make it more generic.

francislata:
@geohot - all updated.

.github/workflows/test.yml
```diff
@@ -81,13 +81,11 @@ jobs:
...
-    - name: Cache model weights
+    - name: Cache downloads
       uses: actions/cache@v3
       with:
-        path: |
-          weights/whisper-tiny.en.pt
-          weights/gpt2.tiktoken
-        key: model-weights-v1
+        path: ~/.cache/tinygrad/downloads/
+        key: model-weights
```
geohot, line 88: Please read the diff and ask what wouldn't make sense to someone reading it.

PR #2421, opened 2023-11-24
Fix UnicodeDecodeError when debugging on Intel APU

andresgit:
My laptop has an Intel APU and during debugging with the "GPU" backend when I use `DEBUG >=5` I get the error

```python
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 17: invalid start byte
```

This is caused in ops_gpu.py in `CLProgram` by `print(prg.decode('utf-8'))`.

I added a basic test that causes this error on the CI. As a suggested fix this PR prints it only if the device name contains "NVIDIA".


geohot:
That test changes DEBUG

andresgit:
I removed that test and now changed test.yml to test DEBUG=5

PR #2422, opened 2023-11-24
Update VITS to use fetch helper

francislata:
## Overview
Continuation of #2366 that updates VITS example.

### DIFF
--- a/examples/vits.py
+++ b/examples/vits.py
@@ -515,9 +515,8 @@ class HParams:
 
 # MODEL LOADING
 def load_model(symbols, hps, model) -> Synthesizer:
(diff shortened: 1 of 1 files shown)


examples/vits.py
```diff
@@ -681,8 +673,8 @@ def _expand_number(self, _inflect, m):
   model_config = MODELS[args.model_to_use]
 
   # Load the hyperparameters from the config file.
-  config_path = model_config[0]
-  download_if_not_present(config_path, model_config[2])
+  config_path = fetch(model_config[0])
+  fetch(model_config[1])
```
geohot, line 677: Why is this here?

PR #2427, opened 2023-11-24
Auto Fetch tinyllama weights for examples/llama.py

yeeyangtee:
Use fetch API to get tinyllama weights from HF, if --gen is tiny and --model is not specified.
For other --gen, model path should be specified.

Using TinyLlama/TinyLlama-1.1B-intermediate-step-240k-503b by default.
Replicates the results from https://github.com/tinygrad/tinygrad/pull/2374,
`python3 ./examples/llama.py --prompt="Hello." --gen=tiny --count=10 --temperature=0`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -10,6 +10,9 @@ jobs:
   testmacbenchmark:
     name: Mac Benchmark
     runs-on: [self-hosted, macOS]
+    defaults:
+      run:
+        shell: bash -o pipefail {0}
     if: github.repository_owner == 'tinygrad'
     env:
       PYTHONPATH: .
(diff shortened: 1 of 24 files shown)


geohot:
This is not the right place to put the URLs. Imagine you wanted to add more for other models.

yeeyangtee:
Makes sense, I moved the default model urls to the MODEL_PARAMS.
I also added functionality to load bfloat16 models, based on the llama2 loading logic from `torch_load` in `state.py`
This is tested working with 
"TinyLlama-1.1B-intermediate-step-240k-503b" which is float32 model and "TinyLlama-1.1B-Chat-v0.6" which is a bfloat16 model (the default one now)

Issue #2417, opened 2023-11-24
JIT: Incorrect output for delayed CPU copy.


imaolo:
```
from tinygrad.tensor import Tensor
from tinygrad.jit import TinyJit

@TinyJit
def do_jit(x) -> Tensor: return x.sum().realize()

res = []
res.append(do_jit(Tensor([1]*2)))
res.append(do_jit(Tensor([2]*2)))
res.append(do_jit(Tensor([3]*2)))

print([r.numpy() for r in res])
```
[array(2., dtype=float32), array(6., dtype=float32), array(6., dtype=float32)]

imaolo:
The toCPU() copy is too late. The device buffers have already been written. More reason for a #2391 fix.

imaolo:
Thinking this problem is deeper than removing the toCPU() copy. The jit is returning the same Tensor for each invocation.

wozeparrot:
I'm not sure which behaviour is supposed to be the correct one here, I've been going off the basis that JIT is dumb and just replays the kernels, so if I want the data later then it shouldn't be kept as a tensor. But that makes the user facing behaviour slightly strange. 

An easy-ish fix for this would be to just recreate all output tensors when jitted functions are run, and replace all the output buffers with the new ones, like its done for the input buffers.
But, I'm not sure if this is a potential perf hit/behaviour that we want.

imaolo:
Agreed that the behavior is a bit strange.

For most cases, I had a cheap solution for removing the toCPU() copy. Reallocate _buf, but that is horrible practice, not to mention memory leaks. It would be better to replace at the RawBuffer layer. The issue was replacing the lazybuffer's reference to the RawBuffer that is being replaced, but your comparison to the input buffer mechanism will help a lot with this.

At that point, it might make sense to replace at the Tensor layer, solving the current issue.

Thank you for the suggestions.

chenyuxyz:
supported with Tensor.clone

PR #2434, opened 2023-11-25
Remove tocpu copy

imaolo:
Emulate jitted input buffers. Thanks to @wozeparrot for pointing me in the right direction (#2417)

local tests are failing on M1.


imaolo:
the send and receive functions don't work because the jit buffer copy works through the function's return value, while recv_jit receives into the first parameter.

Issue #1637, opened 2023-08-23
Bounty: Training 3D-UNet


waifairer:
Add the training logic to examples/mlperf/model_train.py -> def train_unet3d():

[Links](https://github.com/mlcommons/training/tree/master/image_segmentation/pytorch) to the reference implementation for image segmentation.

Training Dataset: [kits19](https://github.com/neheller/kits19).

MLPerf training rules [here](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#appendix-allowed-optimizers)

---

Current Bounty: $500
To lock the bounty submit a draft PR with a decent amount of progress made
Make sure to reference this issue in the PR for future tracking

Notice: If the PR goes stale the bounty will be unlocked

waifairer:
Previous attempts:
https://github.com/tinygrad/tinygrad/pull/1469

iantimmis:
Is this bounty for this still available? Looks like previous attempts were closed.

PR #2437, opened 2023-11-25
Training 3D-UNet #1637 

dxlnr:
#1637 

Request to lock the bounty. First epoch run done with a A100 NVIDIA GPU and input tensor shape of (64,64,64).


geohot:
Locked!

dxlnr:
I've came across an issue and I wanted to ask if there is any lead regarding that. It is also related to  #2448. 

Basically, I was **not** able to complete one full training step for the required (128,128,128) input tensor with batch size 1 as a OOM error is triggered down the line. 
*pycuda._driver.LogicError: cuEventSynchronize failed: an illegal memory access was encountered*
This was done using a A100 NVIDIA GPU with 80GB VRAM.
Additionally, I've tried to use half precision values for both the data and the model. Did help but not sufficient.
It works with input size (64,64,64) tough.

And I'll tried running the [mlcommons training pipeline](https://github.com/mlcommons/training) with input size (128,128,128) and it works even with a RTX 4090 (24GB). 

So I suspect, freeing up memory along the way is not yet implemented? Any leads how we/I can implement that?

imaolo:
> Running into a similar issue like #2448 when training with (128, 128, 128) input size using CUDA backend: OOM error.
> 
> **pycuda._driver.LogicError: cuModuleLoadDataEx failed: an illegal memory access was encountered**
> 
> Even when using a NVIDIA A100 (80GB) and with batch size 1. Also tried with half precision both on data and model tensors. Did only help a little bit. The issue persists when running it with CL backend.
> 
> The single training step works when using (64, 64, 64) input size (tested on NVIDIA RTX 4090 (24GB)).
> 
> For reference, I tested the [mlcommons](https://github.com/mlcommons/training) implementation of the unet3d in torch. This runs through seamlessly on the RTX 4090 with input tensor (128, 128, 128). In total, torch seems to use only roughly 7GB (batch size 1) for one step.
> 
> Any leads how to tackle this issue?

Sent some messages about this in tinygrad-dev.

I've prototyped ad hoc swapping between device/cpu. It works for stable diffusion on m1, but I am not sure it will be fast enough for a full training run.

There are more sophisticated techniques like scheduling swaps for async execution or freeing/recomputing.

I can put something up for ad hoc swapping in the next week or so. There was just a major buffer refactor so I need to update some stuff.

imaolo:
Have you tried since grabbing master? I dont get OOMs where I used to.

dxlnr:
@imaolo Hey sorry for the late response. Yes the issue persists for the 4090, but I have to admit I did not try it for a A100 yet, but will do.

dxlnr:
This can close as switched to a different branch. #2696

PR #2442, opened 2023-11-25
Switch to cuda wrapper instead of pycuda

nimlgen:
Decided to move to our own wrapper instead of pycuda & python-cuda. Also switches `nvcc` to `nvrtc`.

As of python-cuda I find it a bit strange when it comes to managing versions: if you have cuda11, you have to install python-cuda=11.x.x (the same for cuda12). Also there are no pip package of cuda11 for python3.11 and no cuda12 for python3.8.

### DIFF
--- a/extra/cuda_wrapper.py
+++ b/extra/cuda_wrapper.py
@@ -33,6 +33,7 @@ try:
     ptr = ctypes.c_char_p()
     status = _libcuda.cuGetErrorString(status, ctypes.byref(ptr))
     if status != 0: raise RuntimeError("CUDA error: cuGetErrorString failed")
+    assert ptr.value is not None
     return ptr.value.decode("utf-8")
 
   def cuCheckStatus(status):
--- a/extra/cuda_wrapper.py
+++ b/extra/cuda_wrapper.py
@@ -9,6 +9,7 @@ from dataclasses import dataclass
 try:
   def cu_get_include_paths(compiler):
     result = subprocess.check_output([compiler, "-E", "-x", "c", "-", "-v"], input="", stderr=subprocess.STDOUT, universal_newlines=True)
+    print("nvcc", result)
     lines = result.splitlines()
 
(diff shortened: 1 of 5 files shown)


extra/cuda_wrapper.py
```diff
@@ -0,0 +1,377 @@
...
+    status = _libcuda.cuGraphExecDestroy(gexec)
+    cuCheckStatus(status)
+
+  _libcuda.cuGraphLaunch.restype = int
+  _libcuda.cuGraphLaunch.argtypes = [ctypes.c_void_p, ctypes.c_void_p]
+  def cuGraphLaunch(graph_exec, stream=0):
+    status = _libcuda.cuGraphLaunch(graph_exec, stream)
+    cuCheckStatus(status)
+
+  class hipKernelNodeParams(ctypes.Structure):
```
geohot, line 154: Do you want this to be hip?

geohot:
Are HIP and CUDA the same API? I'd support bringing to main a very clean library that supports both

nimlgen:
Yeah, hip is almost cuda, but several functions

wozeparrot:
HIP is designed to match the cuda api pretty closely.

This does bring the same problem that hip has currently, which is that its impossible to use hip when tinygrad is used as a python package, at least without pythonpath shenanigans.

geohot:
I wonder if we can autogenerate these bindings from header files https://github.com/ctypesgen

geohot:
I have this being 100% autogenerated.

My proposal is that I'll put this in a library, then tinygrad can depend on this library and use it directly.

https://gist.github.com/geohot/6232fb00527de161a5c8ce8a635dd4f3

geohot:
https://github.com/tinygrad/gpuctypes

will go here

https://pypi.org/project/gpuctypes/

Going to try to autogenerate all three, and get the building to happen in CI

PR #2444, opened 2023-11-26
Move upcast and local size check into apply_opt

chenyuxyz:
also not allowing upcast == 256 which fixed a linearizer test case


chenyuxyz:
local size < 256 is definitely too small

chenyuxyz:
so we don't have all handcode opt result in search space yet due to the upcast and local size check. this is deeper than i initially thought

PR #2445, opened 2023-11-26
Remove the toCPU copy

qtkite:
Removes the copy from lib.py but copied in numpy instead


geohot:
Hmm, this isn't a bad idea. I'm curious about the performance implications, I suspect this fixes the issues I was seeing.

tinygrad/runtime/lib.py
```diff
@@ -41,7 +41,7 @@ class RawBufferMapped(RawBufferCopyIn):
   def _buffer(self) -> memoryview: raise NotImplementedError("must be implemented")
   # NOTE: this metadata prevents the backing buffer from being freed. hack can be removed with PEP688
   def buffer_view(self) -> np.ndarray: return np.frombuffer(self._buffer(), dtype=np.dtype(self.dtype.np, metadata={"backing": self}), count=self.size)  # type: ignore
```
geohot, line 43: This function is now useless.

wozeparrot:
This doesn't address #2417, since the copy would happen too late.

imaolo:
> Hmm, this isn't a bad idea. I'm curious about the performance implications, I suspect this fixes the issues I was seeing.
> 
> When is the copy needed and why does this fix it?

This works because the tests call numpy() before recalling the jitted function, similar to what woze said. qtkite beat me to this by a couple hours I had just realized this would pass the current tests lol.

I am not sure if recreating the tensor at each jit call is feasible without placing more limitations on what types JIT can return. Consider returning an object where a tensor is deeply nested. Would we walk the return value to find the tensor?

It may be possible without walking the jit return object, but I am still working through how it would work.

tinygrad/runtime/lib.py
qtkite: Removed, Tinygrad 1 line shorter.

qtkite:
> Hmm, this isn't a bad idea. I'm curious about the performance implications, I suspect this fixes the issues I was seeing.
> 
> When is the copy needed and why does this fix it?

From running the testcases, looks like issues only arise when calling numpy will modify the rawbuffer so the copy is used there. Can't seem to notice any performance impact by doing some timing tests on the jit cases. I cleaned up buffer view from the rest of the codebase, now Tinygrad is 1 line shorter. Also just read about #2417 but that would mean a more robust handling of the tensors as imaolo mentioned.

tinygrad/runtime/lib.py
```diff
@@ -40,9 +40,8 @@ def fromCPU(cls, x:np.ndarray, **kwargs):
 class RawBufferMapped(RawBufferCopyIn):
   def _buffer(self) -> memoryview: raise NotImplementedError("must be implemented")
   # NOTE: this metadata prevents the backing buffer from being freed. hack can be removed with PEP688
-  def buffer_view(self) -> np.ndarray: return np.frombuffer(self._buffer(), dtype=np.dtype(self.dtype.np, metadata={"backing": self}), count=self.size)  # type: ignore
-  def toCPU(self) -> np.ndarray: return self.buffer_view().astype(self.dtype.np, copy=True) # Need a copy (for now), since jit will write to the same buffer.
-  def _copyin(self, x:np.ndarray) -> None: np.copyto(self.buffer_view(), x.reshape(-1))
+  def toCPU(self) -> np.ndarray: return np.frombuffer(self._buffer(), dtype=np.dtype(self.dtype.np), count=self.size)
```
geohot, line 43: You just broke tinygrad on METAL

geohot:
It's an ugly hack I know, but can you see why it's there? Maybe there's a better way to deal with this issue that I'm missing.

geohot:
I locked the bounty for this, I think it's the right idea, it just needs tests

qtkite:
Right, I think this should be defined on what kind of behavior we want to see in JIT. If you are happy with the way this is as JIT is quite dumb, I can create another test that shows tocpu will share the same underlying buffer as the previous device buffer and document it as is.

geohot:
Hmm interesting test, and I agree the behavior is counterintuitive

geohot:
Merged, bounty is yours! Reach out to george@tinygrad.org with a PayPal or USDC address to claim

PR #2450, opened 2023-11-26
Add isinstance check before gcd call in SumNode.__lt__

PaulGustafson:
SumNode.\_\_lt\_\_ currently has an ill-typed call to math.gcd, sending a (Node | int) to the function which requires an int.   This PR fixes this by checking to see if the argument is an int.  If it is an int, the behavior is the same.  If it is not an int, the gcd is set to 1, so which results in the code not performing an optional reduction of both sides of the inequality.  It also includes a test case which reproduces the error in the old code.

It might also be worth writing a gcd function in the Node class so that more reductions are possible.  This would also affect similar code in SumNode.\_\_floor\_div\_\_


geohot:
LGTM!

PR #2452, opened 2023-11-26
Fix quickstart.md usage of sparse_categorical_crossentropy

Liamdoult:
- Documentation suggests that sparse_categorical_crossentropy should be imported but import is not shared in \__init\__.py. 
- Reviewed other usages of sparse_categorical_crossentropy and Tensor method is used.
- Tested full demo and worked.
image

- Updated documentation accordingly.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -134,7 +134,7 @@ They can then be used to update the parameters of our neural network using one o
 For our loss function we will be using sparse categorical cross entropy loss.
 
 ```python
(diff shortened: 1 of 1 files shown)


wozeparrot:
The definition of `sparse_categorical_crossentropy` is literally in the quickstart file? The idea is that the definition that is in the file is the one that is used since its the one that the user would have written out.

chenyuxyz:
i removed those commented import lines in #2463, hope that's less confusion

i think the goal was to show how to add a loss function, it just happened to be one that is supported in the core lib

PR #2415, opened 2023-11-24
[WIP] fetch tinyllama

rodfer0x80:
TLDR:
-- add fetching tinyllama from ~/.cache/tinygrad/downloads to ./weights/Tiny-LLaMA/1B from HuggingFace https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T

Log: 
PYTHONPATH="."  CPU=1 python examples/llama.py --prompt="Hello." --gen="tiny" --size="1B" --count=50 --temperature=0
using CPU backend
using LLaMA-tiny-1B model
https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T/resolve/main/model.safetensors?download
https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T/resolve/main/model.safetensors?download
https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T/resolve/main/model.safetensors?download
https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T/resolve/main/model.safetensors?download
https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T/resolve/main/model.safetensors?download
https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T/resolve/main/model.safetensors?download
https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T/resolve/main/model.safetensors?download
https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-955k-token-2T/resolve/main/tokenizer.model?download=t
ram used:  4.40 GB, freqs_cis                                         : 100%|███████| 202/202 [00:01<00:00, 131.50it/s]
loaded weights in 1536.88 ms, 4.40 GB loaded at 2.86 GB/s
Hello.

A: You can use the following code to get the^C

PYTHONPATH="."  CPU=1 python examples/llama.py --prompt="Hello." --gen="tiny" --size="1B" --count=10 --temperature=0
using CPU backend
using LLaMA-tiny-1B model
ram used:  4.40 GB, freqs_cis                                         : 100%|███████| 202/202 [00:01<00:00, 109.57it/s]
loaded weights in 1849.09 ms, 4.40 GB loaded at 2.38 GB/s
Hello.

A: You can use the following code

PR #2453, opened 2023-11-26
Safe load bf16 support

rodfer0x80:
* add support for bfloat16 in safe_load safetensors
* ref #2415 #2427  bf16 support for autofetch tiny llama


tinygrad/nn/state.py
```diff
@@ -16,7 +16,7 @@ def safe_load_metadata(fn:Union[Tensor,str]) -> Tuple[Tensor, int, Any]:
 
 def safe_load(fn:Union[Tensor,str]) -> Dict[str, Tensor]:
   t, json_len, metadata = safe_load_metadata(fn)
-  return {k:t[8+json_len+v['data_offsets'][0]:].cast(safe_dtypes[v['dtype']])[:prod(v['shape'])].reshape(v['shape']) for k,v in metadata.items() if k != "__metadata__"}
+  return {k:t[8+json_len+v['data_offsets'][0]:].cast(safe_dtypes[v['dtype']])[:prod(v['shape'])].reshape(v['shape']).bitcast(dtypes.uint16).to("CPU").cast(dtypes.uint32).mul(1<<16).bitcast(dtypes.float32).to(Device.DEFAULT).half() if v['dtype'] == "BF16" else t[8+json_len+v['data_offsets'][0]:].cast(safe_dtypes[v['dtype']]) [:prod(v['shape'])].reshape(v['shape']) for k,v in metadata.items() if k != "__metadata__"}
```
geohot, line 19: This needs to be abstracted out

rodfer0x80: Thank you for the feedback, will work on it

tinygrad/nn/state.py
```diff
@@ -15,8 +15,15 @@ def safe_load_metadata(fn:Union[Tensor,str]) -> Tuple[Tensor, int, Any]:
   return (t, json_len, json.loads(t[8:8+json_len].numpy().tobytes()))
 
 def safe_load(fn:Union[Tensor,str]) -> Dict[str, Tensor]:
-  t, json_len, metadata = safe_load_metadata(fn)
-  return {k:t[8+json_len+v['data_offsets'][0]:].cast(safe_dtypes[v['dtype']])[:prod(v['shape'])].reshape(v['shape']) for k,v in metadata.items() if k != "__metadata__"}
+  ts, json_len, metadata = safe_load_metadata(fn)
+  def build_tensor(t, v, dtype) -> Union[str, Tensor]:
+    return t[8+json_len+v['data_offsets'][0]:].cast(dtype)[:prod(v['shape'])].reshape(v['shape'])
+  def cast_bf16(t):
+    return t.bitcast(dtypes.uint16).to("CPU").cast(dtypes.uint32).mul(1<<16).bitcast(dtypes.float32).to(Device.DEFAULT).half()
```
geohot, line 22: Isn't this code used elsewhere?

tinygrad/nn/state.py
```diff
@@ -15,8 +15,15 @@ def safe_load_metadata(fn:Union[Tensor,str]) -> Tuple[Tensor, int, Any]:
   return (t, json_len, json.loads(t[8:8+json_len].numpy().tobytes()))
 
 def safe_load(fn:Union[Tensor,str]) -> Dict[str, Tensor]:
-  t, json_len, metadata = safe_load_metadata(fn)
-  return {k:t[8+json_len+v['data_offsets'][0]:].cast(safe_dtypes[v['dtype']])[:prod(v['shape'])].reshape(v['shape']) for k,v in metadata.items() if k != "__metadata__"}
+  ts, json_len, metadata = safe_load_metadata(fn)
+  def build_tensor(t, v, dtype) -> Union[str, Tensor]:
```
geohot, line 19: Why are there three functions now?

rodfer0x80: I misunderstood  the point

tinygrad/nn/state.py
rodfer0x80: You are right, for torchload and safeload in this file, changed it.

tinygrad/nn/state.py
rodfer0x80: Was trying to make it more readable since added the bf16 cast but maybe not the greatest approach

rodfer0x80: Got it first one is useless

rodfer0x80:
* This is useful for models other than llama2 for now, but with the removal of numpy it will be no longer be needed I think
* The tests from #2635 cover this on efficientnet with torch load.

geohot:
Fixed in master to just load bfloat16. The casting shouldn't even be in that file

PR #2458, opened 2023-11-27
Fix MulNode.get_bounds when self.b is a Node

PaulGustafson:
Previous behavior: whenever a MulNode had a Node as its second argument, its get_bounds() would not return integers but Nodes.  This violated the type signature and led to hidden incorrect behavior when comparing mulNodes (also affected division and mod since these use comparisons).  

New behavior: MulNode.get_bounds() should always return a tuple of integers

I had to comment out two asserts in test_sumnode_divmod_sumnode_complex that print "Not supported" errors now since an inequality in MulNode.\_\_mod\_\_ got fixed.

### DIFF
--- a/test/unit/test_symbolic.py
+++ b/test/unit/test_symbolic.py
@@ -374,8 +374,8 @@ class TestSymbolicSymbolicOps(unittest.TestCase):
     assert ((i*128+128)*2 + gidx0*128 + lidx1*(i*512+512) + ridx2*4) % (i*128+128) == gidx0*128 + ridx2*4
     assert ((gidx0*128+i*128+ridx2*4+129)) // (i*128+128) == 1
     assert ((gidx0*128+i*128+ridx2*4+129)) % (i*128+128) == gidx0*128 + ridx2*4 + 1
-    assert (ridx2*(i*4+4)+1+i+gidx0) // (i*128+128) == 0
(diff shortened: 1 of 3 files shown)


chenyuxyz:
test failing - we need MulNode with Nodes for symbolic shapetracker, in which case we treat self.b as a constant when we get the bounds

PaulGustafson:
Yes, this PR still needs some work.  

I do think it's useful to figure out because there is incorrect behavior in Node.\_\_mod\_\_ that gets evaluated in many reshape operations.  In the following line of Node.\_\_mod\_\_,

``` if (b - self).min > 0 and self.min >= 0: return self```

the expression `(b - self).min > 0` evaluates to True whenever it is a Node (which in many cases can have negative values).   I think it mainly occurs when you have a Variable whose max is another Variable (e.g. from `idxs = [Variable(f"idx{i}", 0, s-1) for i,s in enumerate(self.shape)]` when `s` is a Variable in `expr_idxs` in shapetracker.py).   (By the way this also violates the typing for Variable.\_\_init\_\_ which is supposed to only accept int bounds).

I'm currently trying to figure out a good example of a reshape that would expose the bug.

PaulGustafson:
Just going to close this since I think it makes more sense to work on typing/assertions to reject unsupported code first.

PR #2461, opened 2023-11-27
WebGL backend (YOLOv8 on iPhone)

wpmed92:
_What?_

Implementation of WebGL backend

_Why?_

WebGPU only has [29%](https://caniuse.com/?search=webgpu) global coverage in browsers, and currently there isn't any mobile browser that supports it. On the other hand, WebGL 2.0 is supported in [96%](https://caniuse.com/?search=webgl2) of browsers.

_How?_

Hacky, since WebGL 2.0 does not support compute kernels, so the computation is performed using the graphics pipeline, which means this backend is pretty limited. We can only write to the current fragment, there is no arbitrary load/store.
Compute is fragment shader, tensor buffers are 2D textures. A "kernel" invocation is rendering to a framebuffer texture. The texture parameters are `gl.R32F`, `gl.RED`, `GL.FLOAT`, so one float value per fragment. `float4` might be supported by using `vec4` as output.
Texture size is limited by `gl.getParameter(gl.MAX_TEXTURE_SIZE))`, `maxDim x maxDim` is the largest 2D texture we can create. At first I created `maxDim x 1` textures for simplicity (and to emulate how other backend create textures, where you only need the size of the buffer), but than we can't create large textures. 
Indexing is performed using `gl_FragCoord.x`, and `gl_FragCoord.y`, which are mapped to `uv` coordinates to sample textures.
I'm using the `moderngl` Python library to run the computation on Python side, and the backend "becomes" `WebGL 2.0` at model export time. It's somewhat similar to how we run webgpu.
For now I created a stripped down `index_webgl` to demo efficient net, but we can actually reuse the same `index.html` we use for `webgpu`.

Efficientnet: https://softwiredtech.github.io/efficientnet-tinygrad-webgl/
YOLOv8: https://softwiredtech.github.io/yolov8-tinygrad-webgl/


wpmed92:
Efficientnet demo is live: https://softwiredtech.github.io/efficientnet-tinygrad-webgl/

wpmed92:
@geohot YOLOv8 https://softwiredtech.github.io/yolov8-tinygrad-webgl/

wpmed92:
@geohot Tests are passing with minimal skips (skipped std tests on CI, and another one). Core changes are sub-100 lines. Efficientnet and yolov8 run. The backend isn't really webgl but a general "non-compute" gl backend, only becomes webgl when we export it.

geohot:
YOLO demo is🔥on my Mac, lil slow on my iPhone. But certainly works.

Reviewing now

tinygrad/device.py
```diff
@@ -243,7 +243,7 @@ def __call__(self, rawbufs:List[Buffer], var_vals:Dict[Variable, int], wait=Fals
     if global_size is not None and local_size is None and all_int(self.global_size): # type: ignore[arg-type]
       # TODO: this is copied from get_program
       from tinygrad.features.search import optimize_local_size
-      local_size = self.local_size = optimize_local_size(self.clprg, global_size, rawbufs)
+      local_size = self.local_size = [1,1,1] if Device.DEFAULT == "WEBGL" else optimize_local_size(self.clprg, global_size, rawbufs)
```
geohot, line 246: This doesn't belong here

tinygrad/renderer/cstyle.py
```diff
@@ -14,6 +14,7 @@ class CStyleLanguage(NamedTuple):
   smem_align: str = ""
   smem_prefix: str = ""
   smem_prefix_for_cast: bool = True
+  explicit_cast_alu: bool = False # hack for glsl
```
geohot, line 17: What's special about GL that it needs this? Maybe the cast should always be called?

geohot:
There's way too many, if `WEBGL` hacks around the codebase. What refactors would be required to remove them?

tinygrad/renderer/cstyle.py
wpmed92: I got rid of this hack, and solved it in `code_for_op` thanks to Quazalin's `dtype` refactor it was easy to solve it.
Got compiler errors for `bool * bool`, GLSL can't do it. So in `code_for_op` I do `int(a)*int(b)` (only for bool dtype). But then you'll have another problem, `cannot convert from 'int' to 'bool'` when you assign such result to a` bool alu`. That's why I added the explicit cast, but turns out I can just solve it in `code_for_op`.

tinygrad/runtime/ops_webgl.py
```diff
@@ -0,0 +1,59 @@
...
+
+CI = os.getenv("CI", "") != ""
+ctx = moderngl.create_standalone_context()
+dtype_map = { dtypes.float64: "f8", dtypes.float: "f4", dtypes.half: "f2", dtypes.int32: "i4", dtypes.uint32: "u4", dtypes.bool: "i1"}
+class WebGLProgram:
+  def __init__(self, name: str, prg: str, bufs:int=0, vars:int=0): self.name, self.prg = name, ctx.program(vertex_shader="#version 330\nprecision highp float;\nin vec2 in_position;in vec2 in_uv;out vec2 uv;void main(){gl_Position=vec4(in_position,0.0,1.0);uv=in_uv;}", fragment_shader=prg)
+  def __call__(self, *bufs, global_size, local_size, wait=False):
+    vert, uv =ctx.buffer(np.asarray([-1, 1, -1, -1, 1, 1, 1, -1], dtype='f4').tobytes()), ctx.buffer(np.asarray([0, 1, 0, 0, 1, 1, 1, 0], dtype='f4').tobytes())
+    self.vao = ctx.vertex_array(self.prg, [])
+    self.vao.bind(0 if CI else self.prg["in_position"].location, buffer=vert, cls='f', fmt='2f4')
```
wpmed92, line 19: In CI `self.prg["in_position"].location` and `self.prg["in_uv"].location`  fail with `KeyError`, for some reason it can't look up the uniform by key, which seems to be a bug in `moderngl`, this is why we bind hardcoded uniform positions.

wpmed92:
> There's way too many, if `WEBGL` hacks around the codebase. What refactors would be required to remove them?

Yes, I agree, although they're one-liners they're not nice, and I'll remove them one-by-one, keeping track of them here:

- [x] `explicit_cast_alu` hack: solved after `code_for_op` `dtype` support
- [ ] global reshape hack
- [ ] local size hack
- [ ] `apply_opt` hack: instead of returning immediately with simplify ones, see what optimizations I can apply, I think some of them (unroll for sure) should be fine

wpmed92:
@geohot 

The inital 4 `== "WEBGL"` core hacks has been removed (and there are a lot less test skips),  and I tried to work around them, but after the `dtype`-less alloc refactor I have one `== WEBGL` hack in device.
Core line diff went down from `109` to `80` (can be `74` if i put back the global reshape hack)
The reason there are these hacks is that WebGL doesn't have a compute pipeline, so things like globals and locals don't make much sense. (and unlike Clang, which also doesn't have a compute pipeline, WebGL doesn't have arbitrary load/store, so I can't do the loops Clang is doing. The "global loop" in WebGL is controlled by the fragment shader invocations and I can only write the current fragment).
I researched ([1](https://webgl2fundamentals.org/webgl/lessons/webgl-gpgpu.html), [2](https://arm-software.github.io/opengl-es-sdk-for-android/compute_intro.html)) and based on what I found I chose the right emulation of compute shaders in WebGL. You can do one of these, of which only the `2.` is capable enough to handle tiny.

`1.` _vertex shader_ based approach (transform feedback): your compute is a vertex shader, your data are vertex attributes. Disqualified by the limitation that you can only access the current vertex, so although elementwise works, reduce won't.
`2.` _fragment shader_ based approach: your compute is fragment shader, your data are textures. You can read the texture at any position, but you can only write to the current fragment. (this is why glsl `render_store` doesn't index the output, it just stores) To have tinygrad only generate one store per shader, I had to disable upcast using the newly introduced `unsupported_opts` linearizer opt. I'm indexing the textures using `gl_FragCoord`.

So although it has these limitations, It runs efficientnet and yolov8.

geohot:
Can you separate out the runtime changes from other refactors?

wpmed92:
@geohot Ok, I'll separate it out. I'm currently exceeding max line count, but i think wgsl-glsl could share more, also there's room to shave off code elsewhere, which I'll look into.

wpmed92:
@geohot I think webgl would make more sense as an extra run-time, so not part of core. Too much hack due to no compute shaders, and the value it brings doesn't justify the lines it takes to fit.

wpmed92:
@geohot I made a new deployment from latest master: https://softwiredtech.github.io/yolov8-tinygrad-webgl/

wpmed92:
@geohot I made a new deployment from latest master: https://softwiredtech.github.io/yolov8-tinygrad-webgl/
Since this won't be upstream, is the exported model running in the deployed demo site enough to claim the bounty?

wpmed92:
Bounty paid out, and webgl merged into extra [here](https://github.com/tinygrad/tinygrad/pull/3041), so closing this.

PR #2462, opened 2023-11-27
[WIP] cuda & hip gpuctypes

nimlgen:
Switch CUDA + HIP to gpuctypes. Still need to clean up several things. Maybe switch to the new style as in `new_lrualloc` branch.


geohot:
Nice! I don't think it'll be much effort to change the style, I'll try to support it on master today

geohot:
+20 lines is fine if it adds GRAPH

nimlgen:
This is ready for review. Also pushed new style (but CI fails there).

The question is how we want to share functions between hip & cuda. Currently, they are in ops_cuda and some logic has been implemented to allow importing ops_cuda when no CUDA devices are present (from ops_hip). Should we consider moving them to a new file? This would introduce some additional lines tho

nimlgen:
Closed in favour of #2539

Issue #2403, opened 2023-11-23
Error on matrix multiply with dtype float64


andresgit:
I'm using Linux. This code:

```python
from tinygrad.tensor import Tensor
from tinygrad.helpers import dtypes
(Tensor.uniform((2,), dtype=dtypes.float64, device="GPU")@Tensor.zeros(2, dtype=dtypes.float64, device="GPU")).numpy()
```

Gives me this error:
```opencl
...
1:4:42: error: call to 'mad' is ambiguous
  data0[0] = mad((double)((val0).y),0.0f,mad((double)((val0).x),0.0f,acc0));
                                         ^~~
./opencl-c.h:7790:21: note: candidate function
float __ovld __cnfn mad(float a, float b, float c);
                    ^
./CTHeader.h:5177:38: note: candidate function
double __attribute__((overloadable)) mad(double a, double b, double c);
                                     ^
```
The kernel it generated is
```opencl
__kernel void r_2(__global double* data0, const __global float* data1) {
  float acc0 = 0.0f;
  float2 val0 = (float2)(*((__global float2*)(data1+0)));
  data0[0] = mad((double)((val0).y),0.0f,mad((double)((val0).x),0.0f,acc0));
}
```
Using float32 or CPU it works.

geohot:
Hmm the problem is that the const isn't a double. This should be pretty easy to fix.

chenyuxyz:
this is fixed on master with the dtype refactor. ast and generated kernel:

```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.double, st=ShapeTracker(views=(View(shape=(1,), strides=(0,), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ SUM (1,)
  2    ┗━┳ MUL 
  3      ┣━┳ CAST (dtypes.double, False)
  4      ┃ ┗━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(2,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  5      ┗━━ CONST ConstBuffer(val=0.0, dtype=dtypes.double, st=ShapeTracker(views=(View(shape=(2,), strides=(0,), offset=0, mask=None, contiguous=False),)))
reduced UOp count from 19 to 18
__kernel void r_2(__global double* data0, const __global float* data1) {
  double acc0 = (double)(0.0f);
[...]

PR #2471, opened 2023-11-27
Fix GPU matrix multiply with dtype float64 #2403

gabelerner:
First PR, so not high confidence in code. The const double suffix may not be needed if all we're trying to do is fix mulacc

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -285,6 +285,15 @@ class TestFloat4(unittest.TestCase):
 
     assert TestFloat4.count_float4(k) == (1, 1)
 
+class TestFloat64(unittest.TestCase):
+  def setUp(self):
+    if not isinstance(Device[Device.DEFAULT], Compiled):
+      self.skipTest("Device does not use linearizer")
+
+  @unittest.skipIf(Device.DEFAULT != "GPU", "GPU only test")
+  def test_double_basic(self):
+    np.testing.assert_equal((Tensor.uniform((2,), dtype=dtypes.float64) @ Tensor.zeros(2, dtype=dtypes.float64)).numpy(), np.zeros(1))
+
 class TestHandCodedOpts(unittest.TestCase):
   def setUp(self):
     if not isinstance(Device[Device.DEFAULT], Compiled):
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -75,7 +75,7 @@ class Linearizer(Kernel):
         (g_idx, g_valid), amt, dim = self.sts[i].expr_idxs(fake_idxs), 1, None
     else:
       g_idx, g_valid = self.sts[i].expr_idxs(fake_idxs)
-    localtype = dtypes.float32 if amt == 1 else dtypes.float.vec(amt)
+    localtype = (dtypes.float64 if isinstance(buf, (ConstBuffer, MemBuffer)) and buf.dtype == dtypes.float64 else dtypes.float32) if amt == 1 else dtypes.float.vec(amt)
 
     e_idxs, e_valids = g_idx.expand(expand_vars), g_valid.expand(expand_vars)
 
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -123,14 +123,6 @@ def uops_to_cstyle(lang:CStyleLanguage, function_name:str, uops:List[UOp]) -> Tu
     for v in ru.vin:
       child_count[v] += 1
 
-  def _mulacc_propagate_float64(mulacc_vin):
-    if any(x for x in mulacc_vin if x.dtype == dtypes.float64):
-      for x in mulacc_vin:
-        x.dtype = dtypes.float64
-        if x.arg == TernaryOps.MULACC: _mulacc_propagate_float64(x.vin)
-  for u in uops:
-    if u.arg == TernaryOps.MULACC: _mulacc_propagate_float64(u.vin)
-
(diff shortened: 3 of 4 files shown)


test/test_linearizer.py
```diff
@@ -285,6 +285,15 @@ def test_float4_heterogeneous(self):
 
     assert TestFloat4.count_float4(k) == (1, 1)
 
+class TestFloat64(unittest.TestCase):
+  def setUp(self):
+    if not isinstance(Device[Device.DEFAULT], Compiled):
+      self.skipTest("Device does not use linearizer")
+
+  @unittest.skipIf(Device.DEFAULT != "GPU", "GPU only test")
```
gabelerner, line 293: unsure where this test should go since it's GPU/CL only

tinygrad/renderer/cstyle.py
```diff
@@ -176,6 +176,7 @@ def ssa(u, prefix="t"):
       if args[1].startswith("l"): local_size.append(args[2])
       r[u] = args[1]
     elif uop == UOps.CONST:
+      assert dtype is not None
```
gabelerner, line 179: `dtype` is assumed not `None` inside `render_const`

tinygrad/codegen/linearizer.py
```diff
@@ -75,7 +75,7 @@ def rename_var(v: VariableOrNum, expr: str): return v if isinstance(v, NumNode)
         (g_idx, g_valid), amt, dim = self.sts[i].expr_idxs(fake_idxs), 1, None
     else:
       g_idx, g_valid = self.sts[i].expr_idxs(fake_idxs)
-    localtype = dtypes.float32 if amt == 1 else dtypes.float.vec(amt)
+    localtype = (dtypes.float64 if isinstance(buf, (ConstBuffer, MemBuffer)) and buf.dtype == dtypes.float64 else dtypes.float32) if amt == 1 else dtypes.float.vec(amt)
```
gabelerner, line 78: i didn't know where to do this check against `ConstBuffer` (needed for the `0.0`) and `MemBuffer` (needed for the `acc0`) or something else, not sure of the implications to the rest of the codebase

gabelerner:
ok this approach going to work due to nesting/complexity in the AST ... 

i'm going to try another approach where we see if any of the MULACC dtypes are different and cast them appropriately

tinygrad/renderer/cstyle.py
```diff
@@ -160,6 +160,8 @@ def ssa(u, prefix="t"):
         val = lang.code_for_op[args](strip_parens(r[vin[0]]), *[r[x] for x in vin[1:]])
       elif args == BinaryOps.MAX:
         val = lang.code_for_op[args](*[lang.render_cast([r[x]], dtype) if x.dtype != dtype else r[x] for x in vin])
+      elif args == TernaryOps.MULACC and any(x.dtype == dtypes.float64 for x in vin):
+        val = lang.code_for_op[args](*[f"(double){r[x]}" for x in vin])
```
gabelerner, line 164: this is the main fix. the const suffix above doesn't fully address the problem since there's nested `mad` calls.

ex:
```
__kernel void r_2_2_2(__global double* data0, const __global half* data1) {
  int lidx0 = get_local_id(0); /* 2 */
  double acc0 = 0.0l;
  double acc1 = 0.0l;
  float2 val0 = vload_half2(0, data1+lidx0*2);
  data0[lidx0*2] = mad((val0).y,0.0l,mad((val0).x,1.0l,acc0));
  data0[(lidx0*2)+1] = mad((val0).y,1.0l,mad((val0).x,0.0l,acc1));
}
```

becomes
```
__kernel void r_2_2_2(__global double* data0, const __global half* data1) {
  int lidx0 = get_local_id(0); /* 2 */
  double acc0 = 0.0l;
  double acc1 = 0.0l;
  float2 val0 = vload_half2(0, data1+lidx0*2);
  data0[lidx0*2] = mad((double)(val0).y,(double)0.0l,(double)mad((double)(val0).x,(double)1.0l,(double)acc0));
  data0[(lidx0*2)+1] = mad((double)(val0).y,(double)1.0l,(double)mad((double)(val0).x,(double)0.0l,(double)acc1));
}
```

tinygrad/renderer/cstyle.py
```diff
@@ -160,6 +160,8 @@ def ssa(u, prefix="t"):
         val = lang.code_for_op[args](strip_parens(r[vin[0]]), *[r[x] for x in vin[1:]])
       elif args == BinaryOps.MAX:
         val = lang.code_for_op[args](*[lang.render_cast([r[x]], dtype) if x.dtype != dtype else r[x] for x in vin])
+      elif args == TernaryOps.MULACC and any(x.dtype == dtypes.float64 for x in vin):
```
gabelerner, line 163: can add `and DEVICE=GPU` also here if desired

geohot:
This looks very complex. Why isn't float64 just working?

gabelerner:
> This looks very complex. Why isn't float64 just working?

see https://github.com/tinygrad/tinygrad/pull/2471#discussion_r1407040099 where `float2 val0` breaks the override
also see https://github.com/tinygrad/tinygrad/actions/runs/7013150681/job/19078830589 for the tests that fail with just float64 consts only

i don't know how to simplify it because all the overloads of `mad` expect the same 3 types and while the const args are easy to solve, the other args can come from a lot of places like other variables that aren't float64 or other calls to `mad` which themselves may not be the float64 overload

geohot:
See #2264, this is likely what you want. This is not a float64 specific issue

PR #2473, opened 2023-11-27
Apply ruff linting rules to tests

christopherm99:
Requested [here.](https://discord.com/channels/1068976834382925865/1068976834928193609/1178759679220920482)

In order to remove all the per-file-ignores, I needed to insert two `# noqa: F821` comments, since enabling F821 (undefined-name) causes linting to fail despite the code being valid in these two cases:

1. `tinygrad_jitted_model` in [this snippet,](https://github.com/christopherm99/tinygrad/blob/d9057c37b390eb594ef0d19f2b058487f1fd617f/test/external/external_model_benchmark.py#L72C1-L76C54)
2. and `t` in [this snippet.](https://github.com/christopherm99/tinygrad/blob/d9057c37b390eb594ef0d19f2b058487f1fd617f/test/test_copy_speed.py#L35C1-L43C12)

This is because ruff doesn't know when the lambdas in these snippets will be called, and therefore doesn't know if these variables will already be deleted by the `del` statements. This is a [known bug](https://github.com/astral-sh/ruff/issues/6878) in ruff (and in flake8). 

In addition, I also noticed that some of the rules selected in `ruff.toml` are not actually being applied either in pre-commit hooks or in CI, since they are "nursery rules". If we actually want to enable these rules, we need to enable preview mode (this can be set in `ruff.toml`). For reference, the rules in question are:

1. [`E112` (no-indented-block)](https://docs.astral.sh/ruff/rules/no-indented-block/)
2. [`E113` (unexpected-indentation)](https://docs.astral.sh/ruff/rules/unexpected-indentation/)
3. [`E203` (whitespace-before-punctuation)](https://docs.astral.sh/ruff/rules/whitespace-before-punctuation/)
4. [`E272` (multiple-spaces-before-keyword)](https://docs.astral.sh/ruff/rules/multiple-spaces-before-keyword/)

These are all probably caught by one of the other pre-commit hooks, so perhaps removing them from the ruff config to suppress the warnings is best.

Finally, in `test_nn.py` it looks like some copy-pasting went haywire, so I fixed the duplicated and missing `assert_allclose`'s.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -28,7 +28,3 @@ exclude = [
   "openpilot/",
 ]
 
-[per-file-ignores]
-"test/*" = [
-  "F821",
-]
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,4 +1,4 @@
-tab-size = 2
+indent-width = 2
 
 select = [
   "F",
@@ -30,12 +30,5 @@ exclude = [
 
 [per-file-ignores]
 "test/*" = [
-  "F401",
-  "F403",
-  "F405",
-  "F541",
-  "E722",
-  "E731",
   "F821",
-  "F841",
 ]
--- a/test/external/external_test_opt.py
+++ b/test/external/external_test_opt.py
@@ -147,7 +147,7 @@ class TestOptWChild(unittest.TestCase):
     with CLCache():
       c = (a*b).sum()
       d = c+1
-      _ = c+2
+      e = c+2 # noqa: F841
       d.realize()
       assert len(CacheCollector.cache) == 2, "don't fuse if you have children"
 
--- a/test/external/external_test_opt.py
+++ b/test/external/external_test_opt.py
@@ -85,7 +85,6 @@ class TestInferenceMinKernels(unittest.TestCase):
 
   def test_llama(self):
     from examples.llama import Transformer
-    from tinygrad.shape.symbolic import Variable
     args_tiny = {"dim": 512, "hidden_dim": 1024, "n_heads": 8, "n_layers": 4, "norm_eps": 1e-05, "vocab_size": 1000}
     model = Transformer(**args_tiny)
     for p in get_parameters(model): p.assign(np.zeros(p.shape, dtype=p.dtype.np))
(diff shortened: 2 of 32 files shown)


christopherm99:
Ok, not sure how that passed locally, but will take a look. I think some "unused imports" are used within an eval.

christopherm99:
After fixing some stupid mistakes, everything seems to work.

ruff.toml
```diff
@@ -1,4 +1,4 @@
-tab-size = 2
+indent-width = 2
```
geohot, line 1: Why did this change?

test/external/external_test_opt.py
```diff
@@ -148,7 +147,7 @@ def test_unrealized_child(self):
     with CLCache():
       c = (a*b).sum()
       d = c+1
-      e = c+2
+      _ = c+2
```
geohot, line 150: This assignment to a real var is important to the test

test/test_assign.py
```diff
@@ -42,11 +42,11 @@ def test_post_permuted_assignment(self):
     a.realize()
     b.realize()
     #GlobalCounters.cache = []
-    ba1 = a.lazydata.realized
```
geohot, line 45: Also important to prevent GC here

test/test_lazybuffer.py
```diff
@@ -50,7 +49,7 @@ def test_shuffle_pad_ops_exp(self):
   def test_children_count(self):
     a = Tensor.ones(8,8,8)
     d1 = a.sum((0))
-    d2 = a.sum((0)).reshape(32,2)
+    _ = a.sum((0)).reshape(32,2)
```
geohot, line 52: Also important to prevent GC here

ruff.toml
christopherm99: ruff emits a warning when `tab-size` is used, which suggests this change (see [here](https://docs.astral.sh/ruff/settings/#tab-size)). The effect is the same, but I can revert if desired.

geohot: nah it's fine

geohot:
Sweet, thanks!

PR #2479, opened 2023-11-28
Switch from pyobj to metalcompute

PaulGustafson:
Fixes the "MTLLibrary is not formatted as a MetalLib file" #2226 on my local computer, but I still want to figure out what exactly is doing the fix.


PaulGustafson:
Oops, not a fix.

PR #1306, opened 2023-07-20
[GH-1305] Refactor test_dtypes.py to be cleaner

waifairer:
Pretty straightforward DRY refactor of the test_dtypes.py file.

PR #2488, opened 2023-11-28
Removing METAL Skips as CI works

Liamdoult:
Investigating `Get all passing Metal tests in CI` bounty.

Tried to repro failing tests by running in CI without skips and all tests passed.

After some investigation it appears that the main issue was failing tests `test_half_to_int8` which were changed in #1306.

As for `test_ops.py`, it is not clear as to why each was failing and I am unable to reproduce (in pipeline or on apple silicon mac). Perhaps the runners have moved to using intel to apple silicon, but I can only speculate here. Not sure if further investigation is needed here.


Liamdoult:
Should close #1019

geohot:
Huh, maybe GitHub fixed it?

PR #2489, opened 2023-11-28
Place metal simd matrix commands in unrolled loop

jafioti:
extra/gemm/metal_matmul.py contains a metal matrix multiplication kernel containing simd matrix commands that can be looped. The metal compiler automatically unrolls this loop, so no performance is lost.


geohot:
I find the other one more readable

PR #2497, opened 2023-11-29
Cleanup vectorized hip renders

Qazalin:
The problem with overriding vectorized half cast in render_cast is that since the `half4` dtype doesn't exist in HIP, we can't define a variable as a `half4` in the kernel.

I referenced [HIP's implementation](https://github.com/argonne-lcf/THAPI/blob/f921880387730e75f04b97a8b07d191a8d28bc28/hip/include/hip/amd_detail/amd_hip_vector_types.h#L1796) to define vectorized half dtypes and created the `make_dtypen` functions for it.

This unblocks non-fp32 math


tinygrad/renderer/hip.py
```diff
@@ -22,7 +20,10 @@ class HIPLanguage(CStyleLanguage):
   uses_vload=True
   uses_ptr_arithmetic=True
   arg_int_prefix = "const int"
-  half_prekernel = "#include <hip/hip_fp16.h>\nusing half4 = HIP_vector_type<half, 4>;" + """
+  half_prekernel = "#include <hip/hip_fp16.h>\n" + """
+typedef union { struct { half x, y, z, w; }; half data[4]; } half4; __device__ half4 make_half4(half x, half y, half z, half w) { return {x, y, z, w}; }
```
g1y5x3, line 24: would it make sense to add `__align__(8)` here which I didn't see in the header file? so that when compiles it will use the vector load?

Qazalin:
@geohot that test needs non-fp32 math

tinygrad/renderer/hip.py
Qazalin: I added alignment of 8 and 16 for half4 and half8

geohot:
Understood, merging this

PR #2514, opened 2023-11-30
new style device (this is the good one)

geohot:
This will be -100 lines

Note, it breaks multigpu and removes HIPGraph. They should be added in the new style.


geohot:
@nimlgen if you want to base CUDA and HIP on this, it's here to stay

PR #2517, opened 2023-11-30
load metal buffer from disk

gabelerner:


test/unit/test_disk_tensor.py
```diff
@@ -145,5 +146,12 @@ def test_reshape(self):
     helper_test_disk_tensor("dt5", [1,2,3,4,5], lambda x: x.reshape((1,5)))
     helper_test_disk_tensor("dt6", [1,2,3,4], lambda x: x.reshape((2,2)))
 
+  @unittest.skipIf(Device.DEFAULT != "METAL", "only metal")
+  def test_metal(self):
+    pathlib.Path(temp("dt7")).write_bytes(struct.pack("<III", 1, 2, 3)) # metal is little-endian
+    t1 = Tensor.empty(3, device=f"disk:{temp('dt7')}", dtype=dtypes.uint32)
+    t2 = Tensor([4, 5, 6], dtype=dtypes.uint32)
+    np.testing.assert_allclose((t1 + t2).realize().numpy(), np.array([5, 7, 9], dtype=np.uint32))
```
gabelerner, line 154: @imaolo i saw some of your PRs on this, i got this test to pass. what do you think of the approach?

tinygrad/realize.py
```diff
@@ -12,7 +12,8 @@ def run_schedule(schedule:List[ScheduleItem], disable_logging=False):
     si = schedule.pop(0)
     if not disable_logging: log_schedule_item(si)
     assert all(x.realized for x in si.inputs), "can't run schedule, some inputs aren't realized"
-    assert all(si.out.device == x.device for x in si.inputs) or si.ast.op is LoadOps.FROM, f"all devices must be the same, {si.out.device} != {[x.device for x in si.inputs]} {print_tree(si.ast) or ''}"
+    metal_disk_support = Device.DEFAULT == "METAL" and si.out.device.startswith("DISK:")
```
gabelerner, line 15: you can combine DISK tensors with METAL tensors and the code will know what to do. not sure if i even like this

tinygrad/realize.py
```diff
@@ -25,20 +26,21 @@ def run_schedule(schedule:List[ScheduleItem], disable_logging=False):
           if any(not x.arg.st.contiguous for x in si.ast.get_lazyops() if x.op == BufferOps.LOAD and x.arg.idx == i+1):
             si.out.output_buffer = None
             break
+    device = Device["METAL" if metal_disk_support else si.out.device]
```
gabelerner, line 29: use the METAL runtime for DISK tensors if we're on METAL

tinygrad/runtime/ops_metal.py
```diff
@@ -72,7 +84,7 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int], local_size:Tuple[int,i
     encoder = command_buffer.computeCommandEncoder()
     encoder.setComputePipelineState_(self.pipeline_state)
     for i,a in enumerate(bufs):
-      if isinstance(a, RawMetalBuffer): encoder.setBuffer_offset_atIndex_(a._buf, 0, i)
+      if isinstance(a, RawMetalBuffer): encoder.setBuffer_offset_atIndex_(a.try_from_file()._buf if i > 0 else a._buf, 0, i) # out buffer at index 0 is never a file
```
gabelerner, line 87: the issue here is that the out buffer at index 0 has `from_file` set because of the way the AST is but we want to ignore it since we're not writing back to disk in this case

gabelerner:
@geohot i see you're in the process of updating allocators/devices so i'm going to close for now and try to rebase on top of whatever you merge. take a look at this PR fast and if you like the direction and if so perhaps some of your new code can assist here like allowing a disk tensor to combine w/ a metal tensor more easily.

PR #2527, opened 2023-11-30
Metal loadBuffer

imaolo:
I am seeing little to no speedup on M1 air (using stable diffusion as benchmark). Hopefully someone else can make it work better than I can.

The integration isn't perfect either. RawBufferMappedDisk is a cheap way of doing it, and makes _realize_from() even less generic. Also, adding a filename attribute to RawBuffer is not ideal.

We could remove more lines from ops_metal.py by making _buf a property (same as RawDiskBuffer) and doing the io command buffer wait in there, but I didn't see the use because the speed wasn't great anyway.

Maybe newer macs take better advantage of loadBuffer if someone wants to try.


geohot:
It might be easier with the new refactor. Run `sudo purge` to test reading off the disk

PR #2529, opened 2023-11-30
New fix accessing elements created by padding

andresgit:
An alternative to PR #2298 to fix all the issues mentioned in #2177 and #2194.

Explanation of the changes:

1. In `View.__unsafe_resize` I clip the nmask to be between 0 and the new_shape, because otherwise for example `Tensor([1,2]).pad(((3,0),))[0:2]` would have `mask=((3,2),)`, and then in `ShapeTracker.to_movement_ops` it would calculate a negative real_shape and also would want to pad by 3, but that is wrong since the final result has only 2 zeros.
2. In `ShapeTracker.to_movement_ops` I check if the real_shape contains zero and there is a mask, which would happen if we shrink to elements outside the base buffer (created by padding), so in this case I set the real_offset to be zero to not go outside the base buffer and get an error.
3. In `View.reshape` I return None if the new_shape is empty and the current mask is what you would get by shrinking to an element created by padding. This we need because the empty shape also removes all mask info and causes the step 2 above not to work, so by returning None we create a multi-view where the first view contains the mask, so step 2 above works, and then the second view just resizes to an empty shape.


chenyuxyz:
looks good, minor readability comments

PR #2532, opened 2023-12-01
Switch ops_gpu -> gpuctypes

geohot:


geohot:
@nimlgen I think you might be able to save a few lines here, you are better at ctypes than me

PR #2536, opened 2023-12-01
Define ReduceOps accumulators in all dtypes

Qazalin:
Once values get rendered as their non-fp32 buffer dtype there are two problems with using -inf for ReduceOps.MAX:

1. bool: it's casted as `true`
2. int: it overflows

I think it's safe to also ensure 0.0 only gets rendered if the dtype is actually a float for ReduceOps.SUM


Qazalin:
when I loaded buffers as their dtype this was one of the things that broken the tests

tinygrad/codegen/linearizer.py
```diff
@@ -457,6 +457,7 @@ def replace_op(old:UOp, new:UOp):
 
   def uop(self, uop:UOps, dtype:Optional[DType], vin:Tuple[UOp, ...], arg:Any=None, cachable=True, insert_before=None, simplify=True) -> UOp:
     key = (uop, dtype, vin, arg)
+    if uop == UOps.DEFINE_ACC and not dtypes.is_float(cast(DType, dtype)) and arg == -math.inf: arg = False if dtype == dtypes.bool else -2**31
```
geohot, line 460: This doesn't belong here. Why is uop getting called with the wrong type?

Qazalin:
oh actually now with this new implementation it can't be a separate PR, should be a part of the entire non-fp32 render

PR #2537, opened 2023-12-01
Refactor llvm const renders

Qazalin:
We could get away with not casting DEFINE_ACC's arg because it was always a float. Now this renderer works like cstyle
https://github.com/tinygrad/tinygrad/blob/7fec966b5ec66efdc7dfcd4b0f0f3a99a866c503/tinygrad/renderer/cstyle.py#L55


Qazalin:
openpilot test being flakey now https://github.com/tinygrad/tinygrad/actions/runs/7057125485/job/19210239789?pr=2537#step:16:602

PR #2539, opened 2023-12-01
hip & cuda to gpuctypes

nimlgen:


geohot:
The CUDA CI is now ~3min down from 5, amazing!

extra/dist/world.py
```diff
@@ -1,23 +1,24 @@
-from typing import Optional
+import ctypes
 from extra import dist
 from multiprocessing import shared_memory
 from tinygrad.helpers import DEBUG, colored, getenv
 from tinygrad.lazy import LazyBuffer
 from tinygrad.runtime.lib import RawBuffer, RawBufferCopyInOut
```
geohot, line 6: This doesn't work anymore due to this.

geohot:
graph/cuda.py and graph/hip.py can probably be one file

extra/dist/world.py
nimlgen: I just updated this file as well to match calls to hip. Should i revert since it's not supported any more?

geohot: meh, it's fine

nimlgen:
Yeah

geohot:
Same USDC address for the $300 bounty?

geohot:
Sent

PR #2553, opened 2023-12-01
Metal load buffer attempt #3

imaolo:


geohot:
Is it still faster than master on Sonoma?

imaolo:
> Is it still faster than master on Sonoma?

Not on my machine. readinto() is really fast. This is because we used the metal shared storage mode so I assume the memory is shared between CPU and GPU.

We can make the readinto() path even faster by only mmap'ing when UnderlyingDiskBuffer.mem is accessed. We will never need to mmap() when loading directly into device buffers. I am seeing 20-30% improvement on this, not sure if it is worth it.

PR #2554, opened 2023-12-01
Remove as_buffer for TORCH

christopherm99:
When torch is using cuda, it should not support `as_buffer`, since it is not possible to convert a torch tensor that is on the gpu to a buffer (since torch doesn't use the unified memory api). Right now, this fix just removes `as_buffer` altogether, but if desired, I can change this to conditionally define `as_buffer` only when torch is using cuda tensors.

I can't really think of a good way to implement regression tests in CI for this, since without cuda support in CI it is not possible to instantiate torch cuda tensors. Perhaps a test that requires `as_buffer` to be undefined when `torch.cuda.is_available()` is mocked to return True?


geohot:
Not worth the complexity, I'm fine with just removing as_buffer altogether

christopherm99:
Sounds good, will remove and fix existing tests. Should I write that regression test? Seems like it has high potential to give false positives.

PR #2555, opened 2023-12-01
Refactor code_for_op to accept a dtype

Qazalin:
The immediate user is UnaryOps.NEG for the bool dtype.

### DIFF
--- a/examples/handcode_resnet50_opt.py
+++ b/examples/handcode_resnet50_opt.py
@@ -1,9 +1,10 @@
 from typing import List
 from extra.models.resnet import ResNet50
 from tinygrad.tensor import Tensor
-from tinygrad.ops import LoadOps, Device, Compiled
+from tinygrad.ops import LoadOps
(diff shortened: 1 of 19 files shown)


Qazalin:
openpilot flakey again, maybe we should do a retry? https://github.com/tinygrad/tinygrad/actions/runs/7065373654/job/19235253307?pr=2555#step:15:602

Qazalin:
openpilot flaky again, maybe we should do a retry? https://github.com/tinygrad/tinygrad/actions/runs/7065373654/job/19235253307?pr=2555#step:15:602

geohot:
Two things:

1. As I mentioned in the other PR, which param does this dtype apply to?
2. I don't think a retry is a good choice. Can you remove the non determinism?

Qazalin:
both inputs and the final output, because inputs always get upcasted.
the exception here is TernayOps.WHERE, there is always one bool and two vars of the same dtype, in that case, the two vars and the return value have the same dtype

PR #2565, opened 2023-12-02
bug fix due to #2529

sahamrit:
#2529 introduced this small bug

### DIFF
--- a/tinygrad/shape/view.py
+++ b/tinygrad/shape/view.py
@@ -108,7 +108,7 @@ class View:
       assert all(isinstance(s, (int, Variable)) for s in new_shape), f"{self.shape=} -> {new_shape=} contains non (int, Variable) dim"
       if prod(self.shape) != prod([s if isinstance(s, int) else cast(Variable,s).val for s in new_shape]): raise ValueError(f"size mismatched, can't reshape {self.shape=} -> {new_shape=}")
 
-    if new_shape == () and self.mask and any(mx==my for (mx,my) in self.mask): return None
+    if new_shape == () or (self.mask and any(mx==my for (mx,my) in self.mask)): return None
 
     # after the asserts, it's okay to check contiguous
     if self.contiguous: return View.create(new_shape)


tinygrad/shape/view.py
```diff
@@ -108,7 +108,7 @@ def reshape(self, new_shape: Tuple[sint, ...]) -> Optional[View]:
       assert all(isinstance(s, (int, Variable)) for s in new_shape), f"{self.shape=} -> {new_shape=} contains non (int, Variable) dim"
       if prod(self.shape) != prod([s if isinstance(s, int) else cast(Variable,s).val for s in new_shape]): raise ValueError(f"size mismatched, can't reshape {self.shape=} -> {new_shape=}")
 
-    if new_shape == () and self.mask and any(mx==my for (mx,my) in self.mask): return None
+    if new_shape == () or (self.mask and any(mx==my for (mx,my) in self.mask)): return None
```
sahamrit, line 111: i am actually a bit confused now. as such nothing will fail in master.

so the issue is below. in master currently.
```
View(shape=(1, 5), strides=(5, 1), offset=-5, mask=((1, 1), (0, 5)), contiguous=False).reshape((5,))
returns 
View(shape=(5,), strides=(1,), offset=-5, mask=((0, 0),), contiguous=False),)
```
This case is handled explicitly in add/remove ones logic in master. But for all other cases where mask range is zero, the return doesn't happen in this line, but at the last return None of view.reshape.

Question then is if we want the logic below this line to handle ranged 0 mask. The change I was working couldn't handle the above case, because I hadn't explicitly hard coded this add/remove ones case

sahamrit:
But how does "and" -> "or" doesn't affect the test cases ?

tinygrad/shape/view.py
sahamrit: But how does "and" -> "or" doesn't affect the test cases ? i still don't understand this line @andresgit

PR #2568, opened 2023-12-02
[ready] refactor getitem round 2 :D

geohotstan:
gonna give refactoring getitem another go before cleaning up/enabling `test_indexing`
once getitem is suuuuuper nice and readable, enabling and merging stuff from test_indexing should be easy


geohotstan:
@chenyuxyz 
ok I think I managed to make our `__getitem__` much more readable and I think the logic is easier to reason about 
with a +0 line diff :D
well, more readable according my me... who wrote it.... so take a look and lemme know if it improved lol

few things I did

- turned all internal lists and tuples tensors
- tracked the elements in indices using a `Dict[type, dimension]` and use that through all of getitem
- raise most of the exceptions in `# validation! raise Errors` block cuz by then we have parsed enough to raise those errors
- removed the annoying `StopIteration` error by replacing the dim injection and dim collapse calculation with straight forward pop and inserts
- removed .contiguous() cuz for some reason CI passes (before it was using torch+cpu and was causing errors)
- added some tests (not comprehensive but I want to start working on test_indexing so that I can get a good idea on what tests to add)

----------

there are few more TODOs I haven't finished
- for # TODO: move all slice(None) to the end and transpose non-None to the front
do we still need to do that?

- for # TODO: figure out the exact acceptable types for indices, especially for internal list/tuple types
I think I'll figure that one out after finishing test_indexing

- I'll correct the docs at the very end

---------
I think the getitem refactor is good enough to merge for now.

test/test_ops.py
```diff
@@ -662,9 +663,25 @@ def test_slice_errors(self):
     a = Tensor.ones(4, 3)
     with self.assertRaises(IndexError):
       a[1, 77, 77, 77]  # IndexError: (finds too many indices before the out of bounds)
-      a[1, 77]  # IndexError: (out of bounds).
       a[0, -77]
       a[..., ...] # IndexError: only single ellipsis
+    with self.assertRaises(IndexError): # no floats
+      a[1.1]
```
chenyuxyz, line 669: for multiple test cases, i think you would need to separate each into its own `with` block. otherwise it only tests the first one

tinygrad/tensor.py
```diff
@@ -302,75 +302,92 @@ def pad(self, arg:Tuple[Optional[Tuple[sint, sint]], ...], value:float=0.0) -> T
...
-    # TODO: move all slice(None) to the end and transpose non-None to the front
-    ellipsis_idx = ellipsis_found[0] if ellipsis_found else len(orig_slices)
-    orig_slices[ellipsis_idx:ellipsis_idx+1] = [slice(None)] * (len(self.shape) - num_slices)
-
-    valid_slices = [v for v in orig_slices if v is not None]
-    valid_slices = [v if isinstance(v, slice) else slice(y_ := normalize_int(v, i, dim_sz), y_+1) if isinstance(v, int) else slice(None) for i, (v, dim_sz) in enumerate(zip(valid_slices, self.shape))]
-
-    start, stop, strides = zip(*y) if (y := [s.indices(dim_sz) for s, dim_sz in zip(valid_slices, self.shape)]) else ((), (), ())
+    # TODO: figure out the exact acceptable types for indices, especially for internal list/tuple types
+    # TODO: move all slice(None) to the end and transpose non-None to the front...... maybe for fancy indexing?
```
chenyuxyz, line 307: i initially thought it would make something easier, as mentioned in the torch implementation. feel free to remove it if not needed

tinygrad/tensor.py
```diff
@@ -302,75 +302,92 @@ def pad(self, arg:Tuple[Optional[Tuple[sint, sint]], ...], value:float=0.0) -> T
...
+    # treat internal tuples and lists as Tensors and standardize indices to list type
+    if isinstance(indices, (tuple, list)):
+      if isinstance(indices, list) and all(isinstance(i, int) for i in indices): indices = [Tensor(indices)] # special case <indices: List[int]>, a lil ugly
+      else: indices = [Tensor(list(i)) if isinstance(i, (tuple, list)) else i for i in indices]
+    else: indices = [indices]
+
+    # filter ellipsis and fill with slice(None) or fill rest of indices with slice(None)
+    ellipsis_idx = [dim for dim, i in enumerate(indices) if i is Ellipsis]
+    fill_idx = ellipsis_idx[0] if ellipsis_idx else len(indices)
+    num_slices = len(indices) - len(ellipsis_idx) - sum(1 for i in indices if i is None)
```
chenyuxyz, line 320: `sum(1 for i in indices if i is None)` -> `indices.count(None)`

chenyuxyz:
i guess `sum(1 for i in indices if i is None)` is correct here if `indices` contains something that breaks with a `eq` check. sorry for the confusion

PR #2582, opened 2023-12-03
full jitted gpt2

nimlgen:
The final token part can be jitted now. It's put in a separate jit for now, so the main part runs with graph.

### DIFF
--- a/examples/gpt2.py
+++ b/examples/gpt2.py
@@ -102,6 +102,9 @@ class Transformer:
   def __call__(self, tokens:Tensor, start_pos:Variable, temperature:float=0.0) -> Tensor:
     return (self.forward_jit if tokens.shape[1] == 1 and getenv("JIT") else self.forward)(tokens, start_pos, temperature)
 
+@TinyJit
+def sample_token(probs): return probs.multinomial().flatten().realize()
+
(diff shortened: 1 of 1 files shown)


examples/gpt2.py
```diff
@@ -145,8 +148,9 @@ def greedy_until(self, prompt:str, max_length:int, temperature:float, timing:boo
                     f", {GlobalCounters.global_ops*1e-9:.2f} GOPS, {GlobalCounters.global_mem*1e-9:.2f} GB"+
                     (f", {GlobalCounters.global_mem*1e-9/(GlobalCounters.time_sum_s-st):.2f} GB/s" if DEBUG>=2 else "")) if DEBUG else None, enabled=timing):
           probs = self.model(Tensor([x[start_pos:] for x in toks]), Variable("start_pos", 1 if start_pos else 0, MAX_CONTEXT).bind(start_pos), temperature)
-        # TODO: fix JIT rand so we can put this in the JIT
-        tok = probs.multinomial().flatten().numpy().tolist()
+
+        # NOTE: Put sample_token in a seperate jit, so graph can work for the main part.
```
geohot, line 152: This should be fixed in the JIT.

PR #2585, opened 2023-12-03
don't apply padding on script call

SkullMag:
## Bug description
I get an assertion error while calling `vits.py` due to the padding needed for JIT
```
python vits.py
AssertionError: batch size is too small
```
## Fix
We don't really need JIT because the script only generates 1 audio output, so when `pad_length <= 0` is passed, `Synthesizer` doesn't use padding


examples/vits.py
```diff
@@ -733,7 +734,7 @@ def _expand_number(self, _inflect, m):
   # Perform inference.
   start_time = time.time()
   audio_tensor = net_g.infer(x_tst, x_tst_lengths, sid, args.noise_scale, args.length_scale, args.noise_scale_w, emotion_embedding=emotion_embedding,
-                             max_y_length_estimate_scale=Y_LENGTH_ESTIMATE_SCALARS[args.model_to_use] if args.estimate_max_y_length else None)[0, 0].realize()
+                             max_y_length_estimate_scale=Y_LENGTH_ESTIMATE_SCALARS[args.model_to_use] if args.estimate_max_y_length else None, batch_size=0)[0, 0].realize()
```
geohot, line 737: Err, is it 0 or 1?

PR #2591, opened 2023-12-03
Refactor test_ops to get ready for test_ops in all dtypes

Qazalin:
test_ops focuses on float32, with [some exceptions](https://github.com/tinygrad/tinygrad/blob/master/test/test_ops.py#L274).

The idea behind this refactor is to create a base `TestOps` and extend it based on dtypes to `TestFloatOps`, `TestIntOps`, etc, where we reuse the same torch and tiyngrad fxns but just change the data (we might also need to change atol and rtol depending on the dtype).

This diff doesn't change anything about how TestOps works just moves things around:

`helper_test_op` to the TestOps class: because it can get self.DTYPE from it and generate the correct data https://github.com/tinygrad/tinygrad/commit/3ad02f545c6b75203b72d14e1a65b085e360be71 https://github.com/tinygrad/tinygrad/commit/82c55e6eb63caf693c1a88614a5dea25f76afd34

`TestFloatOps` runs everything that was already in TestOps https://github.com/tinygrad/tinygrad/commit/5bcf9ae14a2bb86452202a73c7c2575da290a7ea

Handling test_div_int https://github.com/tinygrad/tinygrad/commit/7dc938c4992b4d9bee0c2fbdef9abf1449f2a2c2 (the only modified test)

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -1,3 +1,4 @@
+from typing import Any
 import torch
 import time
 import math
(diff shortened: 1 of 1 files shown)


geohot:
Two issues with the dtypes PR. 

1. model benchmark failed on NVIDIA, I just disabled it for now. https://github.com/tinygrad/tinygrad/actions/runs/7079546079/job/19266304610
2. while `HALF=1 HIP=1 WINO=1 STEPS=10 python3 examples/hlb_cifar10.py` compiles now, it produces NaNs

geohot:
Why do you need self for helper_test_ops?

Qazalin:
That could work too,

I'm exploring this first because there are some tests like all convs that are only supported in fp32 for torch's CPU backend, so my idea was to have a base TestOps for the default ALU ops tests and then a TestsFloatOps for everything torch CPU can't run for non-fp32

Issue #2418, opened 2023-11-24
Add fast PRNG support (remove LoadOps.RAND)


geohot:
JAX uses an encryption function to encrypt arange, threefry2x32. I think this is the right idea

- [ ] Support doing math on integers
- [ ] Add BinaryOps.XOR and `bitwise_xor` (note, rotate_left is a mul+div+add)
- [ ] Implement threefry2x32 and test it
- [ ] Make arange fast? Ensure this is all fused.
- [ ] Make symbolics work as normal inputs (aka jitted symbolic arange)
- [ ] Add "seed" to the JIT (TODO: how does JAX do this?)

PR #2601, opened 2023-12-04
threefry_2x32

wozeparrot:
Closes: #2418


tinygrad/tensor.py
```diff
@@ -156,9 +156,27 @@ def empty(*shape, **kwargs):
...
   def manual_seed(seed=0): Tensor._seed = seed
 
+  _rng_counter: int = 0
   @staticmethod
   def rand(*shape, **kwargs):
-    return Tensor._loadop(LoadOps.CUSTOM, prod((shape:=argfix(*shape))), arg=custom_random, **kwargs).reshape(shape)
+    num = prod((shape:=argfix(*shape)))
+    if (odd_counts := num % 2): num += 1
+    counts = Tensor.arange(num, dtype=dtypes.uint32) + Tensor._rng_counter
+    Tensor._rng_counter += num
```
blueridanus, line 165: **note**: this works with early realizes but breaks if lazy

tinygrad/tensor.py
```diff
@@ -156,9 +156,27 @@ def empty(*shape, **kwargs):
   @staticmethod
   def manual_seed(seed=0): Tensor._seed = seed
 
+  _rng_counter: int = 0
```
gabelerner, line 159: from what i read, this will need to be device specific. 

how i was planning to do it in my PR was:
`_rng_counter = defaultdict(int)` and then hash device like `device_id = int(hashlib.sha256('CPU'.encode()).hexdigest(), 16) & 0xFFFFFFFF` to get the current counter for the device

also i think below the `0x0` in `ks` should be the `device_id` 

that way the key/counter contain:
- seed
- device id
- device specific counter

which should allow parallel use without getting the same numbers twice

wozeparrot:
> I realized something about XOR. Isn't it just add on a bool Tensor? (though numpy seems to behave differently)
> 
> Never mind, add behaves like OR
> 
> ```
> #include <stdio.h>
> #include <stdbool.h>
> int main() {
>   bool a = true;
>   bool b = true;
>   bool c = a+b;
>   printf("%d\n", c);
> }
> ```
> 
> Oh, xor is SUB! (tested in C)
> 
> ```
> 0 - 0 = 0
> 1 - 0 = 1
> 0 - 1 = 1
> 1 - 1 = 0
> ```
> 
> So it should be possible to bitcast this to bool and subtract them. Would just be in the `__xor__` definition on Tensor, but that way we don't need a new op.

Is it guaranteed that bool has the same bitwidth as uint32 on all backends?

wozeparrot:
hmm, torch doesn't support uint32 as a dtype

geohot:
I'm fine with merging this with XOR, the bool thing has a whole bunch of prereqs we don't have (size changing bitcast for example)

wozeparrot:
yea, gonna put xor in a seperate pr

tinygrad/tensor.py
```diff
@@ -153,12 +153,31 @@ def empty(*shape, **kwargs):
...
+  def manual_seed(seed=0): Tensor._seed, Tensor._rng_counter = seed, Tensor([0], dtype=dtypes.uint32, requires_grad=False)
 
   @staticmethod
   def rand(*shape, **kwargs):
-    return Tensor._loadop(LoadOps.CUSTOM, prod((shape:=argfix(*shape))), arg=custom_random, **kwargs).reshape(shape)
+    if Tensor._rng_counter is None: Tensor._rng_counter = Tensor([0], dtype=dtypes.uint32, requires_grad=False)
+    num = prod((shape:=argfix(*shape)))
+    if num == 0: return Tensor.zeros(shape, **kwargs)
+    counts = (Tensor.arange(num, dtype=dtypes.uint32, requires_grad=None if kwargs.get("requires_grad") is None else kwargs.get("requires_grad")) + Tensor._rng_counter).realize()
+    if (odd_counts := num % 2): counts = counts.cat(Tensor([0], dtype=dtypes.uint32, requires_grad=False))
```
geohot, line 166: Just padding here should work

wozeparrot: its actually easier just to do `if (odd_counts := num % 2): num += 1` but then it doesn't match the jax impl.

geohot:
I'm also fine with hardcoding a few (regenerable by jax) references instead of installing jax. This is pretty typical for crypto stuff

geohot:
I'm fine with keeping the old rand for torch since torch doesn't support uint32

extra/optimization/test_beam_search.py
```diff
@@ -14,38 +14,38 @@ def tearDown(self):
...
 
   def test_variable_ast_beam(self):
-    a = Tensor.rand(3, 3).reshape((Variable("a", 1, 10).bind(3), 3))
+    a = Tensor.empty(3, 3).reshape((Variable("a", 1, 10).bind(3), 3))
     a = (a+1).realize()
 
   def test_big_prime_number(self):
-    a = Tensor.rand(367, 367)
-    b = Tensor.rand(367, 367)
+    a = Tensor.empty(367, 367)
```
geohot, line 21: These tests can't be empty, they need real numbers to compare with numpy

tinygrad/tensor.py
```diff
@@ -152,11 +152,29 @@ def _loadop(op, shape, device:Optional[str]=None, dtype:Optional[DType]=None, ar
...
+    rotations = [[13, 15, 26, 6], [17, 29, 16, 24]]
+    ks = [0x0, Tensor._seed ^ 0x0 ^ 0x1BD11BDA, Tensor._seed]
+
+    x = [(c := counts.chunk(2))[0] + ks[-1], c[1] + ks[0]]
+    for i in range(5):
+      for r in rotations[i % 2]: x[0], x[1] = (x0 := x[0] + x[1]), x0 ^ ((x[1] * (2 ** r)) + (x[1] / (2 ** (32 - r))))
+      x = [(x[0] + ks[i % 3]), (x[1] + ks[(i + 1) % 3] + i + 1)]
+    out = (x[0].cat(x[1])[:num].cast(dtypes.float32).realize() / (2 ** 32 - 1)).reshape(shape).cast(dtypes.default_float if dtype is None else dtype)
+    out.requires_grad = kwargs.get("requires_grad")
+    return out
```
geohot, line 177: Would you not have to change the tests if you added a contiguous here? I think it's fine for rand to not fuse with other ops.

wozeparrot: Most of the test changes aren't actually caused by rand fusing, just that rand itself is slower. the beam search test I don't really want to change, but without changing it it is really slow as its beam searching the rand as well.

I do think that having the contiguous is better anyways.

tinygrad/features/search.py
```diff
@@ -37,7 +37,7 @@ def _time_program(ast:LazyOp, rdev:Compiled, lib:bytes, global_size, local_size,
   tms = []
   for _ in range(cnt):
     if clear_l2:
-      with Context(DEBUG=0): Tensor.rand(1024,1024).realize()
+      with Context(DEBUG=0): Tensor.ones(1024,1024).realize()
```
chenyuxyz, line 40: this needs to be a real buffer to clear l2 cache

geohot:
What's blocking this?

wozeparrot:
I'm not sure what happened with metal, started failing pretty recently, and its hard for me to reproduce as I don't have a mac.

Also on CPU this takes enormous amounts of ram.

Still running into that `__repr__` issue with beam search since the asts are really deep.

geohot:
torch is going to be removed soon, so this should be doable.

wozeparrot:
Still optional? or just full commit?

geohot:
Can we get this merged optionally to start?

wozeparrot:
Currently I have a flag to turn it off, should I flip that so that its opt-in?

geohot:
Yea, I think opt in is good so there's not THREEFRY=0 everywhere in CI

wozeparrot:
CI being slower makes sense, threefry generates quite a few more kernels running the tests.

chenyuxyz:
I think it's fine to skip the randomness half test for threefry. That does not work for bfloat16 either and we need to have a different fix

chenyuxyz:
can you sync to master? I added some bf16 rand stuff that I can clean up after this is merged

geohot:
Congrats, bounty is yours

PR #2604, opened 2023-12-04
set useResources once in metal graph

nimlgen:
The 2nd useResource discards everything from the first. Visible in metal gputrace dependency graph + docs says the same:

> Calling this method again, or calling [useHeap:](https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/2866530-useheap?language=objc), overwrites any previously specified usage options for future dispatch calls within the same compute command encoder.


geohot:
That's crazy...how are you ever supposed to use this API?

PR #2607, opened 2023-12-05
WIP copy_with_st for partial tensor inplace update

chenyuxyz:
with this, gpt-2 kvcache can update only one column per token, not the full cache. on M1 Max with BEAM=12 per token time 13ms -> 11.7ms

does not work with interpreted backend, also need a cleaner op than abusing the BufferOps.STORE

given that this eventually would be part of setitem, we can have a slice <-> shapetracker helpers and an op to call tensor[slice] = other for interpreted.


chenyuxyz:
will open a new pr

PR #2608, opened 2023-12-05
remove numpy from beautiful.py

thelonejordan:


geohot:
Not the way to do it.

PR #2594, opened 2023-12-03
overload half ALU ops in CUDA

Qazalin:
This should fix all ALU ops for half in CUDA
pending CI running tests

PR #2611, opened 2023-12-05
dtypes alu test

geohot:
@Qazalin I wrote some dtype tests, when these pass the bounty is yours.

The TODOs about truediv and lt/eq I don't think are in scope of this bounty, but the failures are.

The LLVM crash and bool failure should be fixed. You can skip half in GPU CI, but I think CUDA should work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ outputs_yolov8
 wandb
 model.safetensors
 quickstart.py
+.hypothesis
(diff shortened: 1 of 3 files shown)


Qazalin:
yeah if these tests pass the implementation in core tinygrad is  done for that

Qazalin:
there's only gonna be the CUDA overloads left which would require CI, I'm working on that next #2594

PR #2218, opened 2023-11-05
view.reshape without symbolic

sahamrit:
@chenyuxyz this is your [earlier attempt](https://github.com/tinygrad/tinygrad/compare/master...chenyuxyz:tinygrad:merge-view) for reshape without symbolic. I analysed that the increase in time for your change is due to cache misses. Below are some details

**good part** 
1. your change reduces call to **merge_view** from 664 to 2 (i have still enabled merge_view in this change)
2. in master even if merge_view was called 664 times, only 7 were success. Whereas in current change, 557 are success. This means less views to deal in shapetracker during **expr_idx** in linearize. 

_Hence we see reduction in linearize time_

**bad part** 

Due to new views, there is less cache hits and increase in **ncalls of reshape (view.py)** from 7398 to 11265. 

_Hence notice the increase in **cumtime** of both shapetracker & view reshape._

Your thoughts on the tradeoff ?

**details below**

**master**
```
# TestWinograd.test_profile
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
17868   67.322    0.004  750.425    0.042 shapetracker.py:191(reshape)
7398   64.887    0.009  483.775    0.065 view.py:98(reshape)

# TestWinograd.test_speed      
running conv: 625.34 ms
scheduling: 179.52 ms
linearize 1 with  911 ops: 320.41 ms
linearize 3 with 1723 ops: 1156.65 ms
linearize 4 with    4 ops: 4.19 ms
linearize 5 with  955 ops: 484.96 ms
linearize 6 with    2 ops: 1.44 ms
.
----------------------------------------------------------------------
Ran 1 test in 3.883s

OK
```

**current change**

```
# TestWinograd.test_profile
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
17868   36.229    0.002  921.770    0.052 shapetracker.py:178(reshape)
11265   99.747    0.009  681.163    0.060 view.py:145(reshape)

# TestWinograd.test_speed      
running conv: 682.60 ms
scheduling: 207.60 ms
linearize 1 with  911 ops: 297.05 ms
linearize 3 with 1723 ops: 1071.64 ms
linearize 4 with    4 ops: 3.18 ms
linearize 5 with  955 ops: 454.48 ms
linearize 6 with    2 ops: 1.31 ms
.
----------------------------------------------------------------------
Ran 1 test in 3.805s

OK
```

PR #2614, opened 2023-12-05
patch reshape merge broken stable diffusion

chenyuxyz:
#2218 breaks stable diffusion with a false merge, might have other impact. disable that case for now. @sahamrit fyi

### DIFF
--- a/test/unit/test_shapetracker.py
+++ b/test/unit/test_shapetracker.py
@@ -271,6 +271,7 @@ class TestIndexExpressions2d(unittest.TestCase):
     assert len(self.st.views) == 1
     self.st.assert_same()
 
+  @unittest.skip("broken")
   def test_reshape_splitting_combining(self):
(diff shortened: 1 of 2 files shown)


chenyuxyz:
no idea why this breaks the index test (and only for TORCH, CPU is fine). will investigate tomorrow

sahamrit:
Checking in sometime

chenyuxyz:
interesting... thanks for digging this. i will check this separately to see if it's a bug if the input view is valid

chenyuxyz:
yeah pretty much what you said, we might see this later with more getitem refactors

for the record this is the input that went into reshape
```
from tinygrad.shape.view import View
view=View(shape=(1, 1, 1, 1), strides=(0, 0, 0, 0), offset=91, mask=((0, 1), (0, 1), (0, 1), (0, 1)), contiguous=False)
view.reshape((1, 1))
```

Issue #2588, opened 2023-12-03
Stale OpenCL program cache after system update


PaulGustafson:
I tried running `python openpilot/compile2.py` after updating my M1 Air to the most recent Mac OS version, and it failed with error `OpenCL Error -30`.   

Fix: Delete cache (`rm -rf ~/Library/Caches/tinygrad` on OS X).    Also can check with `DISABLE_COMPILER_CACHE=1 python openpilot/compile2.py` first .

Longer-term fixes: put a suggestion in the OpenCL error message to try without caching/delete cache.  Could also label the cache with the OpenCL version.

Only problem is that it is a pain to write tests for this.

PR #2618, opened 2023-12-05
compiler targets multiple devices

nimlgen:
Tagging GPU, CUDA and HIP backends with versions, so kernels are recompiled with new versions of compilers. 

Updates #2588


nimlgen:
Is something like this any better? I am not really sure if compile functions looks nice, but we can compile for the target device as well as cache with details of device/api version.

nimlgen:
Does it look any better? This way it allows to compile/search cache for a device passed to compile function.

With compiler_kwrags isn't it harder to pass device.context in ops_gpu with diskcache?

test/test_custom_function.py
```diff
@@ -15,11 +15,11 @@
 # we don't always have GPU support, so the type signature is the abstract CompiledBuffer instead of GPUBuffer
 def atan2_gpu(ret:Buffer, a:Buffer, b:Buffer):
   assert a.dtype == b.dtype and a.dtype == dtypes.float32, "gpu function only supports float32"
-  CompiledASTRunner(None, "atan2_gpu", """
+  CompiledASTRunner(Device[ret.device], None, "atan2_gpu", """
```
geohot, line 18: Tons more boilerplate here

nimlgen: Reverted

tinygrad/runtime/ops_gpu.py
```diff
@@ -13,19 +13,20 @@ def check(status):
...
-    cl.clGetProgramBuildInfo(program, CLDevice.compiler_context.device_id, cl.CL_PROGRAM_BUILD_LOG, 0, None, ctypes.byref(log_size := ctypes.c_size_t()))
-    cl.clGetProgramBuildInfo(program, CLDevice.compiler_context.device_id, cl.CL_PROGRAM_BUILD_LOG, log_size.value, mstr := ctypes.create_string_buffer(log_size.value), None)
-    raise RuntimeError(f"OpenCL Compile Error\n\n{ctypes.string_at(mstr, size=log_size.value).decode()}")
-  binary_sizes = init_c_var((ctypes.c_size_t * 1)(), lambda x: check(cl.clGetProgramInfo(program, cl.CL_PROGRAM_BINARY_SIZES, ctypes.sizeof(x), ctypes.byref(x), None)))
-  binary = init_c_var(ctypes.create_string_buffer(binary_sizes[0]), lambda x: check(cl.clGetProgramInfo(program, cl.CL_PROGRAM_BINARIES, ctypes.sizeof(ctypes.c_void_p), ctypes.byref((ctypes.c_void_p * 1)(ctypes.addressof(x))), None)))
-  check(cl.clReleaseProgram(program))
-  return bytes(binary)
+def compile_cl(device:CLDevice, prg:str) -> bytes:
+  @diskcache
+  def __compile(prg, *args):
```
geohot, line 18: This is harder to read

PR #2625, opened 2023-12-05
Use the reduceop dtype to define the acc in linearizer

Qazalin:
ensures the accumulator is always the output type when uoping `DEFINE_ACC`

this is the first part in solving the problems with midcasted mulaccs (eg. half/float/half). p2 in #2727


Qazalin:
This test passes now:
```
import numpy as np
import torch

Tensor.manual_seed(69)

x = Tensor.rand(1024,1024, dtype=dtypes.float16)
w = Tensor.rand(1024,1024, dtype=dtypes.float16)
out_tiny = (x*w).cast(dtypes.float32).sum(-1).cast(dtypes.float16)

print(out_tiny.numpy())

x = torch.tensor(x.numpy(), dtype=torch.float16)
w = torch.tensor(w.numpy(), dtype=torch.float16)
out = (x*w).to(torch.float32).sum(-1).to(torch.float16)
print(out)

np.testing.assert_allclose(out_tiny.numpy(), out.detach().numpy(), rtol=1e-3, atol=1e-3)
```

geohot:
This is not the right fix. The acc type is wrong.

In general, no casting should be done in the uops function, the types should already be right.

geohot:
I still don't think this is right. This isn't changing how the acc is being declared. Can you paste a generated matmul with fp32 acc but all memory and mul fp16?

dtypes bounty is yours after this is fixed.

Qazalin:
kernel:
```
#include <metal_stdlib>
using namespace metal;
kernel void r_1024_16_64(device float* data0, const device half* data1, const device half* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  threadgroup float temp[16];
  int gidx0 = gid.x; /* 1024 */
  int lidx1 = lid.x; /* 16 */
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 64; ++ridx0) {
    half val0 = *(data1+(gidx0*1024)+(lidx1*64)+ridx0);
    half val1 = *(data2+(gidx0*1024)+(lidx1*64)+ridx0);
    acc0 = (((float)(val0)*(float)(val1))+acc0);
  }
  *(temp+lidx1) = acc0;
  threadgroup_barrier(mem_flags::mem_threadgroup);
  if ((lidx1<1)) {
    float acc1 = 0.0f;
    for (int ridx1 = 0; ridx1 < 16; ++ridx1) {
      float val2 = *(temp+ridx1);
      acc1 = (val2+acc1);
    }
    *(data0+gidx0) = acc1;
  }
}
```

uops:
```
UOps.ALU            : dtypes.float              [<UOps.CAST: 15>, <UOps.CAST: 15>, <UOps.DEFINE_ACC: 7>] TernaryOps.MULACC
   UOps.CAST           : dtypes.float              [<UOps.LOAD: 8>]                 None
   UOps.CAST           : dtypes.float              [<UOps.LOAD: 8>]                 None
   UOps.DEFINE_ACC     : dtypes.float              []                               0.0
```

geohot:
I think this is still wrong. Look at where the accs are being created

tinygrad/codegen/linearizer.py
```diff
@@ -51,6 +51,13 @@ def get_reduce_acc(self, op, dtype:DType):
     if op == ReduceOps.SUM: return 0.0 if dtypes.is_float(dtype) else 0
     elif op == ReduceOps.MAX: return -math.inf if dtypes.is_float(dtype) else -2**31 if dtypes.is_int(dtype) else False
 
+  # TODO less dumb implementation
+  def get_reduce_dtype(self, i:int, op):
```
geohot, line 55: You should just be able to use the floptracker on the reduce op

Qazalin:
```
#include <metal_stdlib>
using namespace metal;
kernel void r_1024_1024(device half* data0, const device half* data1, const device half* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 1024 */
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 1024; ++ridx0) {
    half val0 = *(data1+(gidx0*1024)+ridx0);
    half val1 = *(data2+(gidx0*1024)+ridx0);
    acc0 = (((float)(val0)*(float)(val1))+acc0);
  }
  *(data0+gidx0) = (half)(acc0);
}
```

tinygrad/codegen/linearizer.py
```diff
@@ -47,9 +47,9 @@ def uop_alu_idx(self, a:UOp, b, ops, ctx:Linearizer, op, dtype=dtypes.int32):
   def const(self, b:Union[int,float], dtype=dtypes.int32, insert_before=None) -> UOp: return self.uop(UOps.CONST, dtype, tuple(), b, insert_before=insert_before)
   def cast(self, val: UOp, dtype) -> UOp: return self.uop(UOps.CAST, dtype, (val,)) if val.dtype != dtype else val
 
-  def get_reduce_acc(self, op, dtype:DType):
-    if op == ReduceOps.SUM: return 0.0 if dtypes.is_float(dtype) else 0
-    elif op == ReduceOps.MAX: return -math.inf if dtypes.is_float(dtype) else -2**31 if dtypes.is_int(dtype) else False
+  def get_reduce_acc(self):
+    if cast(LazyOp,self.reduceop).op == ReduceOps.SUM: return 0.0 if dtypes.is_float(get_lazyop_info(self.reduceop).dtype.scalar()) else 0
```
geohot, line 51: Don't repeat the expression three times! Lines don't matter!

geohot:
This is still wrong.

Check half/float/half AND float/half/float

geohot:
Sorry, didn't mean to close. But I'm sick of the back and forth on this.

Please fully think through what the behavior should be and write a test suite.

geohot:
Hint:

The `self.global_load` function is being used to create the accumulator. It's creating it based off of the dtype of buffer 0, it shouldn't be.  The output shape of buffer 0 is fine, It needs to create it with the dtype implied by the reduce op.

Qazalin:
https://github.com/tinygrad/tinygrad/pull/2625/files#diff-63d57daaca59e5495bbb108da1c13e782f4a408db8503e015e8094ac07c87ea6R82 i thought this solved that actually. `self.get_reduce_acc()` is passed to global_load so acc isn't None already

Qazalin:
status: I'm iterating on what the best way is to handle midcasted MULACC fusion by cross-checking the results this branch produces with torch/numpy.

the goal is to try and get to a green test suite with reasonable tolerance, I'll add the those to test_dtype_alu

Qazalin:
I'm opening this back up because this is blocking removal of MULACC from linearizer #2727.

I think the problem with downcasted mulacc should be fixed in lazy instead of here. Once we have support for MULACC in lazy, we can re-enable the tests for downcast.

test/test_linearizer.py
```diff
@@ -106,6 +107,21 @@ def test_sum_collapse(self):
     lin = Linearizer(sched[0].ast)
     assert not any(u.uop == UOps.LOOP for u in lin.linearize().uops), "found loop in sum collapse"
 
+  reduce_ops = (Tensor.max, Tensor.min, Tensor.sum)
+  float_dtypes = [v for v in DTYPES_DICT.values() if dtypes.is_float(v)]
```
chenyuxyz, line 111: consider putting these in `test/helpers.py`. also i removed `dtypes._arg_int32` that might simplify some logic

chenyuxyz (commented):
the change in `global_load` now and looks good to merge. can you sync to master?

test/test_linearizer.py
Qazalin: we need it because bool exists

test/test_linearizer.py
```diff
@@ -106,6 +108,19 @@ def test_sum_collapse(self):
...
+  def test_reduce_acc(self, d:DType, op):
+    a = Tensor.rand(1024,1024, dtype=d)
+    out = op(a)
+
+    ast = [si for si in out.lazydata.schedule() if si.ast.op not in LoadOps][0].ast
+    reduceop = [op for op in ast.get_lazyops() if op.op in ReduceOps][0]
+    uops = Linearizer(ast).linearize().uops
+    phi = [u for u in uops if u.uop == UOps.PHI][0]
+
+    assert phi.dtype == phi.vin[0].dtype == phi.vin[1].dtype == get_lazyop_info(reduceop).dtype
```
chenyuxyz, line 122: this does not test the case that reduceop has a different type from vin[0]

Qazalin: i removed it i dont think it's actually correct
https://github.com/tinygrad/tinygrad/pull/2625/commits/86e74d796023e71122a09b9355bef493da6ae086

Qazalin: we probably need some changes at the ops level to make a float/half/float kernel work

chenyuxyz: oh okay i guess that's fine for now

Qazalin: added back but skipped with a TODO

chenyuxyz:
reverted because it breaks llama. it's likely due to a half@half not doing acc in f32

chenyuxyz:
the bug is that there a kernel in llama that does a sum of int tensor, then (acc - 2) / 128. i will make a fix first so that binaryop(int tenser, float) cast the tensor first

PR #2627, opened 2023-12-05
binaryops xor

wozeparrot:


geohot:
This ready?

wozeparrot:
should be, haven't tested on cuda or metal but the ci tests pass so it should be fine.

geohot:
Yea, simple enough.

PR #2628, opened 2023-12-05
add Tensor.clone lol

geohot:


geohot:
This is actually wrong, what if you assign

PR #2632, opened 2023-12-05
Reintroduce PTX

christopherm99:
WIP: Currently all of `test_ops.py` passes locally.

TODO:
- [x] Don't hardcode `.version` and `.target`
- [x] Get all other tests passing
- [x] Make `mypy` happy and clean up code
- [x] Enable PTX testing in CI with `CUDACPU`

Open to suggestions on how the structure for `AssemblyLanguage` should look. I tried to make `assembly.py` modular so that adding future assembly backends should be easy.


geohot:
Woo bounty locked! This should test nicely in CI

tinygrad/renderer/assembly.py
```diff
@@ -0,0 +1,173 @@
...
+    elif uop == UOps.CONST:
+      r[u] = lang.render_const(args, dtype)
+    elif uop == UOps.LOAD:
+      if vin[0].uop == UOps.DEFINE_LOCAL:
+        index = ssa(None, "index", "u64")
+        kernel.append(f"mul.wide.{lang.dtype_to_asmtype[vin[1].dtype]} {index}, {r[vin[1]]}, {dtype.itemsize};")
+        loc = f"{r[vin[0]]}[{index}]"
+      elif vin[1] and vin[1].uop != UOps.CONST:
+        loc = ssa(None, "loc", "u64")
+        offset = ssa(None, "off", "u64")
```
geohot, line 92: these can be walrus assigns inside the kernel.append

christopherm99:
Over 5000 lines! Once I get everything working, I'll do my best to refactor below the limit!

christopherm99:
Everything passes locally now! Here are the issues that I think need to be resolved before this PR can be merged:

- Testing in CI needs to be fixed. I need to look more closely into what is causing ocelot to crash, but right now it seems like declaring `.shared` for the `DEFINE_LOCAL` UOp causes a crash. Looking at the PTX 2.2 documentation, it appears that `.shared` should be supported, and ocelot claims to fully support PTX 2.2, so maybe there is a different issue at hand here.
- Non-integer `float16` constants are disabled. This is because PTX does not support declaring inline half constants. However, it may be possible to manually add these values into constant table. If not, we're looking at float32 constants that get casted to float16 (check if this gets optimized out by `ptxas`) or disabling half arithmetic for PTX altogether.
- Over 5000 lines! In this current paradigm for `AssemblyLanguage`, which was designed to mirror `CStyleLanguage`, there are perhaps a couple more lines to save (probably not 23 however), but come at the cost of significantly sacrificing readability. Thus I see a couple options to get below the line limit:
  - Remove the modularity of `AssemblyLanguage`, and only implement `PTXRenderer`. This just pushes the issue down the road to when (if) another `AssemblyLanguage` gets implemented. If there are no plans for a future `RDNALanguage` or something similar, this may be the best choice.
  - Another option is to deviate from the `CStyleLanguage` formulation. Since methods like `render_cast` in the current structure do not have access to the kernel that is being generated, they can only emit one line of code, and also have no way of creating new variables. Thus by making `PTXRenderer` a subclass of a new `AssemblyRenderer` callable class, these methods could be more powerful, allowing for more line savings by abstracting more reused code. However, the cost of this is that `cstyle.py` and `assembly.py` would be completely different paradigms, which seems inelegant.
  - Finally, a third option is to unite `CStyleLanguage` and `AssemblyLanguage` (and possibly `llvmir.py` as well), saving lines by eliminating shared code between these implementations. However, this seems like a fairly large change for one PR, and also would likely greatly increase code complexity for the sake of line count.

tinygrad/runtime/ops_cuda.py
```diff
@@ -20,12 +22,12 @@ def check(status):
...
 
 @diskcache
-def compile_cuda(prg) -> bytes: return compile_cuda_style(prg, [f'--gpu-architecture={CUDADevice.default_arch_name}', "-I/usr/local/cuda/include", "-I/usr/include"], cuda.nvrtcProgram, cuda.nvrtcCreateProgram, cuda.nvrtcCompileProgram, cuda.nvrtcGetPTX, cuda.nvrtcGetPTXSize, cuda.nvrtcGetProgramLog, cuda.nvrtcGetProgramLogSize, check)
+def compile_cuda(prg) -> bytes: return prg.replace("TARGET", CUDADevice.default_arch_name).encode("utf-8") if PTX else compile_cuda_style(prg, [f'--gpu-architecture={CUDADevice.default_arch_name}', "-I/usr/local/cuda/include", "-I/usr/include"], cuda.nvrtcProgram, cuda.nvrtcCreateProgram, cuda.nvrtcCompileProgram, cuda.nvrtcGetPTX, cuda.nvrtcGetPTXSize, cuda.nvrtcGetProgramLog, cuda.nvrtcGetProgramLogSize, check)
 
 class CUDAProgram:
   def __init__(self, device:CUDADevice, name:str, lib:bytes):
     self.device, self.name, self.lib = device, name, lib
-    if DEBUG >= 5: print(pretty_ptx(lib.decode('utf-8')))
+    if DEBUG >= 5: print("\n".join([f"{i+1:>3} {line}" for i, line in enumerate(pretty_ptx(lib.decode('utf-8')).split("\n"))]))
```
christopherm99, line 30: Not sure if this is worth keeping, but was helpful for debugging.

geohot:
Tests are failing

tinygrad/runtime/ops_cuda.py
```diff
@@ -20,12 +22,12 @@ def check(status):
 def cu_time_execution(cb, enable=False) -> Optional[float]: return time_execution_cuda_style(cb, cuda.CUevent, cuda.cuEventCreate, cuda.cuEventRecord, cuda.cuEventSynchronize, cuda.cuEventDestroy_v2, cuda.cuEventElapsedTime, enable=enable) if not CUDACPU else cpu_time_execution(cb, enable=enable)
 
 @diskcache
-def compile_cuda(prg) -> bytes: return compile_cuda_style(prg, [f'--gpu-architecture={CUDADevice.default_arch_name}', "-I/usr/local/cuda/include", "-I/usr/include"], cuda.nvrtcProgram, cuda.nvrtcCreateProgram, cuda.nvrtcCompileProgram, cuda.nvrtcGetPTX, cuda.nvrtcGetPTXSize, cuda.nvrtcGetProgramLog, cuda.nvrtcGetProgramLogSize, check)
+def compile_cuda(prg) -> bytes: return prg.replace("TARGET", CUDADevice.default_arch_name).encode("utf-8") if PTX else compile_cuda_style(prg, [f'--gpu-architecture={CUDADevice.default_arch_name}', "-I/usr/local/cuda/include", "-I/usr/include"], cuda.nvrtcProgram, cuda.nvrtcCreateProgram, cuda.nvrtcCompileProgram, cuda.nvrtcGetPTX, cuda.nvrtcGetPTXSize, cuda.nvrtcGetProgramLog, cuda.nvrtcGetProgramLogSize, check)
```
geohot, line 25: Use a different compiler function, not an if statement

test/test_dtype_alu.py
```diff
@@ -57,7 +57,7 @@ def universal_test_unary(a, dtype, op):
   if not isinstance(op, tuple): op = (op, op)
   tensor_value = op[0](Tensor([a], dtype=dtype)).numpy()
   numpy_value = op[1](np.array([a]).astype(dtype.np))
-  if dtype in dtypes_float: np.testing.assert_allclose(tensor_value, numpy_value, atol=5 if Device.DEFAULT == "METAL" and op[0] == Tensor.sin else 1e-3, rtol=2 if Device.DEFAULT == "METAL" and op[0] == Tensor.sin else 1e-4 if dtype == dtypes.float32 else 1e-2)  # exp and log and sin are approximations (in METAL, the default fast-math versions are less precise)
+  if dtype in dtypes_float: np.testing.assert_allclose(tensor_value, numpy_value, atol=5 if (Device.DEFAULT == "METAL" or getenv("PTX")) and op[0] == Tensor.sin else 1e-3, rtol=2 if Device.DEFAULT == "METAL" and op[0] == Tensor.sin else 1e-4 if dtype == dtypes.float32 else 1e-2)  # exp and log and sin are approximations (in METAL, the default fast-math versions are less precise)
```
geohot, line 60: There shouldn't be getenv PTX's here. Why is this different from CUDA?

test/test_dtype.py
```diff
@@ -36,12 +36,12 @@ def _assert_eq(tensor:Tensor, target_dtype:DType, target):
   if DEBUG >= 2: print(tensor.numpy())
   try:
     assert tensor.dtype == target_dtype
-    np.testing.assert_allclose(tensor.numpy(), target)
+    np.testing.assert_allclose(tensor.numpy(), target, rtol=1e-3 if target_dtype == dtypes.float16 else 1e-7)
```
geohot, line 39: Why did these things change?

christopherm99: The underlying issue that we have to use `.rz` as the rounding modifier (round towards zero), since rounding towards nearest is not supported by PTX 2.2, and thus not support by ocelot. This error is negligible in float32 but becomes apparent in float16.

test/test_dtype_alu.py
christopherm99: This is because we're using `sin.approx` here, while cuda doesn't default to using the fast-math form of sine. Unfortunately there is no non-approx `sin` in PTX, otherwise we could just use that instead.

test/test_dtype.py
christopherm99: And the other change in this file is due to a weird edge case in numpy: `a.astype(np.float16).tolist()` gives values in float32 not float16.

christopherm99:
Ok, I opened a PR (#2723) for that weird numpy edge case in `tolist()`. Everything else seems specific to PTX here, so I'm not sure it makes sense to split out into separate PRs?

Another thing to note is that despite these not surfacing CI, it seems that ocelot handles overflows differently than CUDA does in reality, locally resulting in two overflow-related test failures in `test/test_dtype_alu.py::TestDTypeALU`:
 - `test_float16_unary` with reciprocal of `1.52587890625e-05` returns 65504 (max float16) rather than infinity
 - and `test_float_midcast_int32` in `0 + 2147483648.0` returns `2147483647` versus numpy's `-2147483648`.

It seems reasonable to me to say that overflow behavior is undefined? Perhaps these tests should be changed to ensure that the inputs don't result in overflows?

geohot:
I don't think overflow behavior is undefined. It seems all the other backends are passing, maybe there's a flag to set in PTX?

tinygrad/runtime/ops_cuda.py
```diff
@@ -21,11 +23,13 @@ def cu_time_execution(cb, enable=False) -> Optional[float]: return time_executio
 
 @diskcache
 def compile_cuda(prg) -> bytes: return compile_cuda_style(prg, [f'--gpu-architecture={CUDADevice.default_arch_name}', "-I/usr/local/cuda/include", "-I/usr/include"], cuda.nvrtcProgram, cuda.nvrtcCreateProgram, cuda.nvrtcCompileProgram, cuda.nvrtcGetPTX, cuda.nvrtcGetPTXSize, cuda.nvrtcGetProgramLog, cuda.nvrtcGetProgramLogSize, check)
+@diskcache
```
geohot, line 26: not worth caching I'd say

tinygrad/renderer/assembly.py
```diff
@@ -0,0 +1,181 @@
...
+      val = ssa(u, 'val')
+      if dtype.itemsize == 1: tmp = ssa(None, 'tmp', 's8')
+      if len(vin) > 3:
+        assert vin[2].dtype is not None
+        kernel.append(f"setp.ne.{lang.dtype_to_asmtype[vin[2].dtype]} {(pred:=ssa(None,'ld','pred'))}, {r[vin[2]]}, {const(0, vin[2].dtype)};")
+      kernel.append(f"{f'@{pred} ' if len(vin) > 3 else ''}ld{'.shared' if vin[0].uop == UOps.DEFINE_LOCAL else ''}.{'s8' if dtype.itemsize == 1 else 'b16' if dtype == dtypes.float16 else lang.dtype_to_asmtype[dtype]} {tmp if dtype.itemsize == 1 else val}, {loc};")
+      if len(vin) > 3: kernel.append(f"@!{pred} mov.b{'8' if dtype.itemsize == 1 else lang.dtype_to_asmtype[dtype][1:]} {tmp if dtype.itemsize == 1 else val}, {r[vin[3]]};")
+      if dtype.itemsize == 1: kernel.append(f"cvt.{lang.dtype_to_asmtype[dtype]}.s8 {val}, {tmp};")
+    elif uop == UOps.PHI:
+      assert dtype is not None
```
geohot, line 117: Why is this assert dtype line repeated over and over?

christopherm99: Unfortunately without this, mypy complains about indexing into `dtype_to_asmtype` with a possibly None value, I suppose adding a dummy None entry to that dictionary could make the asserts not required?

geohot:
The main issue with merging this is the line count. It's +168 now.

With clean refactors, what do you think it can get down to?

christopherm99:
> I don't think overflow behavior is undefined. It seems all the other backends are passing, maybe there's a flag to set in PTX?

It actually seems like the int32 add overflow case is failing on the CUDA backend too. The float16 reciprocal test is fixed by using `.rn`: I mistakenly believed we had to use `.rz`. 

> The main issue with merging this is the line count. It's +168 now.

To be honest, I think it's pretty close to the limit, at least without removing functionality or decreasing readability. Trying to avoid that, there's maybe 10 lines or so, LOAD and STORE have a lot of duplicate code, and all the `selp` and `setp` pertaining to handling bools can be folded into `cast`.

geohot:
I think this can share a lot of code with llvm

christopherm99:
Sounds good. I'll take a look at migrating llvm to this syntax when I get a chance.

geohot:
Line count has been raised, want to rebase and get this ready to merge? The LLVM refactor can wait.

christopherm99:
Oops, ok, I have LLVM almost merged into this PR, but I'll separate that out into a new PR.

christopherm99:
Well this is back to passing again, but given twitter I take it this should be in `extra/` instead?

geohot:
Yes, I'll merge into extra right now, and the bounty is yours!

geohot:
Merged in to extra

PR #2636, opened 2023-12-06
Gpt2 benchmark with HALF and BEAM

chenyuxyz:


examples/gpt2.py
```diff
@@ -191,4 +192,16 @@ def greedy_until(self, prompt:str, max_length:int, temperature:float, timing:boo
       print('Generating text...')
       if len(texts) == 1: print(texts[0])
       else:
-        for i,text in enumerate(texts): print(colored(f"Response {i}:", "green"), text)
\ No newline at end of file
+        for i,text in enumerate(texts): print(colored(f"Response {i}:", "green"), text)
+
+    # validate output!
+    if args.temperature == 0 and args.model_size == "gpt2-medium" and args.count == 10:
```
wozeparrot, line 198: why not just add a cli flag for this?

geohot: Can you elaborate on this? There's a bit of a debate here on what it should be.

geohot: Did you mean like this? https://github.com/tinygrad/tinygrad/pull/2636/commits/d657abc2ef58cc65b1fdb907d65de9356345baf2

wozeparrot: yea something like that, where the prompt is set and the expected response is also set.

the `try` on the dict feels kinda hacky.

chenyuxyz: time to do real work, we have mlx llama to beat

PR #2637, opened 2023-12-06
Implement Mamba

blueridanus:


blueridanus:
P.S:. merged #2686 into the branch, hence the diff

kroggen:
Great Job!

geohot:
Is this working? Does it match torch output exactly?

kroggen:
For me it is failing

The error is related to `einops`:

```
user@tests:~/tinygrad-mamba/examples# python3 mamba.py 
ram used:  0.52 GB, lm_head.weight                                    : 100%|█| 
loaded weights in 1123.20 ms, 0.67 GB loaded at 0.60 GB/s

So here's whatTraceback (most recent call last):
  File "/root/tinygrad-mamba/examples/mamba.py", line 466, in <module>
    logits = model(Tensor([tks[start_pos:]]), inference_params, start_pos)
  File "/root/tinygrad-mamba/examples/mamba.py", line 411, in __call__
    return self.forward(input_ids, inference_params, num_last_tokens)
  File "/root/tinygrad-mamba/examples/mamba.py", line 402, in forward
    hidden_states = self.backbone(input_ids, inference_params=inference_params)
  File "/root/tinygrad-mamba/examples/mamba.py", line 379, in __call__
    hidden_states, residual = layer(
  File "/root/tinygrad-mamba/examples/mamba.py", line 355, in __call__
    hidden_states = self.mixer(hidden_states, inference_params=inference_params)
  File "/root/tinygrad-mamba/examples/mamba.py", line 224, in __call__
    self.in_proj.weight @ rearrange(hidden_states, "b l d -> d (b l)"),
  File "/usr/local/lib/python3.10/dist-packages/einops/einops.py", line 591, in rearrange
    return reduce(tensor, pattern, reduction="rearrange", **axes_lengths)
  File "/usr/local/lib/python3.10/dist-packages/einops/einops.py", line 518, in reduce
    backend = get_backend(tensor)
  File "/usr/local/lib/python3.10/dist-packages/einops/_backends.py", line 59, in get_backend
    raise RuntimeError("Tensor type unknown to einops {}".format(type(tensor)))
RuntimeError: Tensor type unknown to einops <class 'tinygrad.tensor.Tensor'>
```

kroggen:
@blueridanus said "it works" on a commit

Maybe he has a modified version of einops that supports tinygrad?

examples/mamba.py
```diff
@@ -0,0 +1,472 @@
...
+  if model_name not in MODELS.keys():
+    raise Exception(f"Requested unknown mamba model: {model_name}")
+  downloaded = fetch(
+    f"https://huggingface.co/state-spaces/mamba-{model_name}/resolve/main/pytorch_model.bin?download=true"
+  )
+  weights = torch_load(downloaded)
+  return weights
+
+
+def selective_scan_ref(
```
kroggen, line 61: this can be renamed to `selective_scan`. the `_ref ` in the original code means "reference implementation" and is used to differ from the CUDA implementations

examples/mamba.py
```diff
@@ -0,0 +1,472 @@
...
+    x, z = xz.chunk(2, dim=1)
+    # Compute short convolution
+    if conv_state is not None:
+      conv_state.assign(x[:, :, -self.d_conv :])  # Update state (B D W)
+      x = self.conv1d(x)[..., :seqlen].swish()
+
+    x_dbl = self.x_proj(rearrange(x, "b d l -> (b l) d"))
+    dt = x_dbl[:, : self.dt_rank]
+    B = x_dbl[:, self.dt_rank : (self.dt_rank + self.d_state)]
+    C = x_dbl[:, (self.dt_rank + self.d_state) :]
```
kroggen, line 242: the 3 lines above can now use the new `Tensor.split`

```python
dt, B, C = Tensor.split(x_dbl, [self.dt_rank, self.d_state, self.d_state], dim=-1)
```

blueridanus:
@kroggen Use this branch:
https://github.com/arogozhnikov/einops/pull/297

blueridanus:
Hadn't yet checked if the output matched the original impl, hence the draft status. It does run though.

kroggen:
Wow, it worked!

```
~/tinygrad-mamba/examples# python3 mamba.py 
ram used:  0.52 GB, lm_head.weight                                    : 100%|█| 
loaded weights in 382.25 ms, 0.67 GB loaded at 1.76 GB/s

The capital of France is in the city of Paris.

The city of Paris is in the city of Paris.
```

kroggen:
The output was the same for the 130m model, but different for the 370m model

blueridanus:
> The output was the same for the 130m model, but different for the 370m model

@kroggen thanks for checking! if it's diverging on a larger model, my first hopeful guess would be that some of the tensors ought to be in fp32 but aren't

kroggen:
It has difference in the output from the original also in the last 2 commits

examples/mamba.py
```diff
@@ -0,0 +1,472 @@
...
+
+  def __call__(self, hidden_states: Tensor, inference_params=None):
+    batch, seqlen, dim = hidden_states.shape
+
+    conv_state, ssm_state = None, None
+    if inference_params is not None:
+      conv_state, ssm_state = self._get_states_from_cache(inference_params, batch)
+      if inference_params.seqlen_offset > 0:
+        # The states are updated inplace
+        out, _, _ = self.step(hidden_states[:, -1:, :], conv_state, ssm_state)
```
kroggen, line 220: The `[:, -1:, :]` can be removed to match the original code. The removal was tested and it works fine

kroggen:
The output is different from [mamba-minimal](https://github.com/johnma2006/mamba-minimal)

I am suspecting there is a problem on the logits processing

When using the prompt "The capital of France is" the `seqlen` is `5`, but the returned logits has the shape `(1, 5, 50280)`. The same as the input. Is this OK? I was expecting either 6 (one additional token) or 1 (just the new one)

examples/mamba.py
```diff
@@ -0,0 +1,472 @@
...
+  tks = tokenizer(prompt)["input_ids"]
+  print('\n' + prompt, end='', flush=True)
+
+  temperature = 0.5
+  start_pos = 0
+  inference_params = InferenceParams(max_seqlen=1, max_batch_size=1, seqlen_offset=0)
+  for i in range(200):
+    logits = model(Tensor([tks[start_pos:]]), inference_params, start_pos)
+    inference_params.seqlen_offset = len(tks)
+    tok = (logits[:, -1, :] / (temperature + 1e-10)).softmax().flatten().multinomial().item()
```
kroggen, line 468: The `.flatten()` appears to not be needed

kroggen:
The output is the same when we comment these 2 lines:

```python
inference_params.seqlen_offset = len(tks)
...
start_pos = len(tks)
```

On this case the inference is done via the convolutional mode (the `step` function is not called)

So the mismatch is in the `step()` function (recurrent mode)

examples/mamba.py
```diff
@@ -0,0 +1,472 @@
...
+  def step(self, hidden_states, conv_state, ssm_state):
+    #print(f"hidden_states is {hidden_states.shape}")
+    assert (
+      hidden_states.shape[1] == 1
+    ), f"Only support decoding with 1 token at a time for now, attempted {hidden_states.shape[1]}"
+    xz = self.in_proj(hidden_states.squeeze(1))  # (B 2D)
+    x, z = xz.chunk(2, dim=-1)  # (B D)
+
+    # Conv step
+    conv_state = conv_state[:, :, 1:].cat(x.unsqueeze(-1), dim=-1)
```
kroggen, line 284: The problem is on this line
It creates a new object instead of updating in-place
But the cache keeps reference to the old object
So every time this function is called, this array is empty (all zeros)

kroggen: It can optionally just update the cache with the new object

kroggen: This solves the problem:
```python
conv_state.assign(conv_state[:, :, 1:].cat(x.unsqueeze(-1), dim=-1))
```

PR #2638, opened 2023-12-06
multitensor WIP

geohot:
- [x] Make this not be wrong without DEBUG=2
- [ ] What is blocking, what can go in the graph?


geohot:
The first step for the backend is:

a.pad((50, 0)) + b.pad((50, 0)) = a padded output tensor

a + b.pad((50, 0)) shouldn't actually add the 0s

geohot:
Stale, all good stuff has been committed to master

PR #2640, opened 2023-12-06
replace linter test with pre-commit hook

ubihot:


geohot:
Haha no! Only replace the ops that are already in the pre-commit hook.

geohot:
And what is this change to the tests! It's not even running the pre-commit hook!

ubihot:
My bad! I see what you mean.

PR #2643, opened 2023-12-06
Adding MLX backend

noelfranthomas:
Improving M-Series performance with MLX ops!


geohot:
Do you have benchmarks showing it's faster than METAL?

PR #2645, opened 2023-12-06
adding test to cover #2644 failure

sahamrit:
adding test to cover #2644 failure


sahamrit:
@chenyuxyz have a look

PR #2646, opened 2023-12-06
Get all WEBGPU test_ops passing

wpmed92:
Tests were previously failing due to:

- `dtypes.int64` not supported on webgpu: `test_div`, `test_maximum` and `test_div_int` were failing, because the `vals` array's values were implicitly` int64`, I'm forcing `int32` now with `np.array`
- `bool` not HOST_SHAREABLE: bool cmp tests were failing, because bool storage buffers are not allowed in WebGPU (https://www.w3.org/TR/WGSL/#host-shareable-types)
- 8-buffer per kernel: this limit has already been bypassed by [this hack](https://github.com/tinygrad/tinygrad/blob/0978c24b8e2db85e5eddec1945934b5dd3d3334b/tinygrad/lazy.py#L254), so it's safe to remove the skip from the tests that were failing due to this
- `inf`: `test_pad` inf tests were failing, because the previous `inf` `(0x1.fffffep+127f`) didn't come back as `inf` to cpu, `(1.0/0.0)` seems to work fine, same as in GLSL. The new inf const also fixes the other inf tests that were skipped.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,7 +246,7 @@ jobs:
     #- name: Run webgpu pytest
     #  run: WEBGPU=1 WGPU_BACKEND_TYPE=Metal python -m pytest -n=auto
     - name: Run selected webgpu tests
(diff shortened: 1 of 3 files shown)


geohot:
Thanks!

PR #2650, opened 2023-12-06
split large jit into several graphs

nimlgen:
gpt-2 timings with CUDA:

new:
ran model in 1.79 ms
total 11.74 ms

old:
ran model in 1.67 ms
total 13.05 ms

### DIFF
--- a/test/test_jit.py
+++ b/test/test_jit.py
@@ -260,23 +260,24 @@ class TestJit(unittest.TestCase):
     if Device[Device.DEFAULT].graph is None: raise unittest.SkipTest("only test graphs")
 
     # Create long jit with 83 kernels.
-    def f(a, b, c, d):
+    def f(a, b, c, d, e):
       for _ in range(80):
         a = (a+b).realize()
       y = (a*c).realize()
       z = (y*d).realize()
-      w = (z*d)
+      w = (z*e)
       return w.realize()
 
     a = Tensor.randn(10, 10).realize()
     b = Tensor.randn(10, 10).realize()
     c = Tensor.randn(10, 10).realize()
     d = Tensor.randn(10, 10).realize()
+    e = Tensor.randn(10, 10).realize()
 
     jf = TinyJit(f)
     prev = None
     for _ in range(5):
-      o = jf(a, b, c, d).numpy()
+      o = jf(a, b, c, d, e).numpy()
       if prev is not None: np.testing.assert_allclose(o, prev, atol=1e-4, rtol=1e-5)
       prev = o
 
@@ -284,6 +285,7 @@ class TestJit(unittest.TestCase):
     # Checking that 2 graphs are inited.
     assert isinstance(jf.jit_cache[0].prg, graph_t)
     assert isinstance(jf.jit_cache[1].prg, graph_t)
+
 
 if __name__ == '__main__':
   unittest.main()
--- a/test/test_jit.py
+++ b/test/test_jit.py
@@ -259,11 +259,12 @@ class TestJit(unittest.TestCase):
   def test_jit_batch_split(self):
     if Device[Device.DEFAULT].graph is None: raise unittest.SkipTest("only test graphs")
 
+    # Create long jit with 83 kernels.
     def f(a, b, c, d):
-      x = (a+b).realize()
-      y = (x*c).realize()
-      rn = Tensor.randn(*a.shape)
-      z = (y*rn).realize()
+      for _ in range(80):
+        a = (a+b).realize()
+      y = (a*c).realize()
+      z = (y*d).realize()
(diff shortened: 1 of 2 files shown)


tinygrad/jit.py
```diff
@@ -74,18 +73,29 @@ def __call__(self, *args, **kwargs) -> ReturnType:
...
       if (make_graph := Device[Device.DEFAULT].graph) and getenv("JIT") != 2:
-        try:
-          if DEBUG >= 1: print(f"JIT GRAPHing {len(self.jit_cache)} kernels with {len(input_rawbuffers)} inputs")
-          self.jit_cache = [JitItem(make_graph(self.jit_cache, input_rawbuffers, var_vals), cast(List[Optional[Buffer]], input_rawbuffers))]
-        except GraphException as e:
-          if DEBUG >= 1: print(f"graph create failed {e}")
-      else:
-        if DEBUG >= 1: print(f"JIT captured {len(self.jit_cache)} kernels with {len(input_rawbuffers)} inputs")
+        # Split JIT cache into batches for faster graph execution. This allows the accelerator to run some batches while subsequent graphs are still being updated.
+        jit_batches: List[List[JitItem]] = []
```
geohot, line 81: Why are these two loops?

tinygrad/jit.py
```diff
@@ -22,7 +22,6 @@ def get_input_replace(jit_cache: List[JitItem], input_rawbuffers:List[Buffer]) -
     for i,a in enumerate(ji.rawbufs):
       if a in input_rawbuffers:
         input_replace[(j,i)] = input_rawbuffers.index(a)
-  assert len(set(input_replace.values())) == len(input_rawbuffers), "some input tensors not found"
```
geohot, line 25: Why was this removed?

test/test_jit.py
```diff
@@ -253,5 +255,31 @@ def foo(x) -> Tensor: return x.sum().realize()
...
+    a = Tensor.randn(10, 10).realize()
+    b = Tensor.randn(10, 10).realize()
+    c = Tensor.randn(10, 10).realize()
+    d = Tensor.randn(10, 10).realize()
+
+    jf = TinyJit(f)
+    for _ in range(5): jf(a, b, c, d).numpy()
+
+    graph_t = Device[Device.DEFAULT].graph.func if isinstance(Device[Device.DEFAULT].graph, functools.partial) else Device[Device.DEFAULT].graph
+    assert len(jf.jit_cache) == 3, "2 graphs and rand op"
```
geohot, line 279: Whenever you write tests, I want you to think about how many refactors they will survive.

What's going to happen when we switch rand to threefry?

Many of your tests just end up getting deleted because they are fragile.

tinygrad/jit.py
nimlgen: This method is called in graphs which contain a part of jit cache which might not have all inputs in their slice. I moved this check to the jit, so the behaviour is the same as it was for a single graph/regular jit.

tinygrad/jit.py
```diff
@@ -74,16 +73,22 @@ def __call__(self, *args, **kwargs) -> ReturnType:
...
-        except GraphException as e:
-          if DEBUG >= 1: print(f"graph create failed {e}")
-      else:
-        if DEBUG >= 1: print(f"JIT captured {len(self.jit_cache)} kernels with {len(input_rawbuffers)} inputs")
+        # Split JIT cache into batches for faster graph execution. This allows the accelerator to run some batches while subsequent graphs are still being updated.
+        graphed_jit_cache, batch_size = [], getenv("JIT_BATCH_SIZE", 64)
+        for start in range(0, len(self.jit_cache), batch_size):
+          current_batch = self.jit_cache[start:start+batch_size]
+          try:
+            graphed_jit_cache.append(JitItem(make_graph(current_batch, input_rawbuffers, var_vals), cast(List[Optional[Buffer]], input_rawbuffers)))
```
geohot, line 86: Err, this won't be good for copies.

geohot:
Can you write a spec of what you want this to do?

nimlgen:
Yeah that was not ready. Sorry, should have marked as draft.

Basically we want 2 things: 1) split jit into smaller graphs and 2) create graphs even when there are some operations not supported by the graph.  In the current implementation all non-CompiledASTRunner operations are expected to fail the graph creation.

The algo is pretty simple, just iterating through all jit items, combining consecutive items into batches of the maximum allowed size. If a JIT item is a non-CompiledASTRunner, we flush the current batch (which means creating a graph and putting it into the cache) and then add this jit item as is.

This is all implemented in a single pass. I hope it's more or less understandable. I've also added some comments to clarify this concept in the code.

geohot:
Much better!

Issue #2649, opened 2023-12-06
Remove numpy


geohot:
- [ ] Remove numpy from the dtypes
- [ ] Support Tensor creation from lists and nested lists without bumpy
- [ ] Add some clean method to get the Tensor out as a native Python object

Do these before just replacing numpy in places

geohot:
Added the method `.data()` to return a memoryview. That should work without numpy https://github.com/tinygrad/tinygrad/pull/2657

IvanVnucec:
@geohot Can you explain what does it mean to "get the Tensor out as a native Python object"? Example would do.

hristog:
> @geohot Can you explain what does it mean to "get the Tensor out as a native Python object"? Example would do.

This is a good question!
I'm not from the maintainers team (and I could be wrong in saying the following), but my interpretation is that the desired result is a native Python data structure, such as the efficient [`array`](https://docs.python.org/3/library/array.html) type or a newly-introduced `class` equivalent - abstraction-level-wise - to a Numpy `ndarray`, or (I suppose less likely) a plain `list`.

hristog:
For instance, the approach in https://github.com/tinygrad/tinygrad/pull/2745 proposes utilising the standard library `struct.pack`.

chenyuxyz:
dup to #321

PR #2655, opened 2023-12-06
Remove numpy #1

imaolo:
- [x] Remove numpy from the dtypes (it is just a string now)
- [x] Support Tensor creation from lists and nested lists without bumpy
- [ ] Add some clean method to get the Tensor out as a native Python object

Using ctypes for the canonical conversions, but there is no float16 ctype so some manual conversion will need to happen eventually.

to_mv/frombuffer has replaced np.array

 #2649

should probably consider using struct.pack()

### DIFF
--- a/test/external/external_test_opt.py
+++ b/test/external/external_test_opt.py
@@ -166,7 +166,7 @@ class TestOpt(unittest.TestCase):
     with CLCache(allowed=1):
       d = a * b + c
       d.realize()
-    np.testing.assert_allclose(d.numpy(), na*nb+nc, rtol=1e-5)
+    np.testing.assert_allclose(d.numpy(), na*nb+nc, rtol=1e-4)
 
   def test_fold_reduce_elementwise(self):
     img = Tensor.ones(32)


tinygrad/helpers.py
```diff
@@ -143,25 +163,29 @@ def is_float(x: DType) -> bool: return x in (dtypes.float16, dtypes.float32, dty
...
+  int8: Final[DType] = DType(1, 1, "char", "int8", ctypes.c_byte)
+  int16: Final[DType] = DType(3, 2, "short", "int16", ctypes.c_short)
+  int32: Final[DType] = DType(5, 4, "int", "int32", ctypes.c_int)
   int = int32
-  int64: Final[DType] = DType(7, 8, "long", np.int64)
-  uint8: Final[DType] = DType(2, 1, "unsigned char", np.uint8)
-  uint16: Final[DType] = DType(4, 2, "unsigned short", np.uint16)
-  uint32: Final[DType] = DType(6, 4, "unsigned int", np.uint32)
-  uint64: Final[DType] = DType(8, 8, "unsigned long", np.uint64)
+  int64: Final[DType] = DType(7, 8, "long", "int64", ctypes.c_long)
```
geohot, line 184: Why are these strings here?

geohot:
This is still too complex for a single PR. If the title of the PR is "remove numpy #1" and not a clear single "feature", it won't be merged

imaolo:
> This is still too complex for a single PR. If the title of the PR is "remove numpy #1" and not a clear single "feature", it won't be merged

We need the strings so we can support numpy(). They act the same when used in np.array(dtype=) but are not np.dtypes().

Will work on removing them.

tinygrad/helpers.py
imaolo: We need the strings so we can support numpy(). They act the same when used in np.array(dtype=) but are not np.dtypes().

Will work on removing them.

PR #2617, opened 2023-12-05
green dtypes ALU tests

Qazalin:
#2611
- [x] Skip GPU fp16 in CI
- [ ] Green Metal
- [x] Fix LLVM add overflow in bools https://github.com/tinygrad/tinygrad/commit/213447897f5fe4679189006c2e0d7999ecbdfe09
- [x] LLVM ALU fns in all floats https://github.com/tinygrad/tinygrad/commit/7fcb1ef31059c7cfbbbaf15e219a2d25d5e39621
- [ ] LLVM segfaults
- [ ] Green CUDA

PR #2656, opened 2023-12-06
green more dtypes tests

Qazalin:
#2653

- maximum can overflow, need an alternative
- LLVM uint ALU support
- sin in METAL is very imprecise and even using `precise::` doesn't help much. The rtol is very forgiving for that
- updated fp16 skip logic

---
this is where numpy overflow is inconsistent:
`print((np.float32(0.0) + np.float32(2147483904.0)).astype(np.int32) + np.int32(0)) # -2147483648 in CI, 2147483647 on OSX`
CUDACPU and Metal don't match numpy's behavior in CI for overflows: https://github.com/tinygrad/tinygrad/actions/runs/7128008768/job/19409125198?pr=2656#step:16:268
I think opencl and clang match because it's running on the same platform?


geohot:
No need to do `precise::sin`, you can lower the accuracy for this, exp, and log. The precise ones are really slow

Qazalin:
lmao even precise::sin doesn't cut it https://github.com/tinygrad/tinygrad/actions/runs/7119973005/job/19386221617?pr=2656#step:8:236

Qazalin:
the last thing left here is `test_float_midcast_int32` that I had made the range safer in the prev PR https://github.com/tinygrad/tinygrad/pull/2617#discussion_r1417611804 to _avoid_ an overflow.

I'll test the overflow values against a C backend to figure out why Metal and CUDACPU runtimes don't behave the same way that numpy does when it comes to overflows.

Qazalin:
So two things blocking mergablity:

sin is still flakey in Metal _CI_ we can't merge until we can do something about it
https://github.com/tinygrad/tinygrad/actions/runs/7120262877/job/19387138091?pr=2656#step:8:251


this overflow doesn't match numpy's in Metal _CI_  only, other backends match
https://github.com/tinygrad/tinygrad/pull/2656/commits/13674836e7e6ea0327e8c4da47128005e0fce945

This CUDACPU failure is similar, it's not reproducible across backends https://github.com/tinygrad/tinygrad/actions/runs/7120314671/job/19387291984?pr=2656#step:16:263

PR #2660, opened 2023-12-06
Replace np.array() in Tensor.__init__() with get_mv() and np.frombuffer()

imaolo:
finally fixed the typo...

I removed bytes and None from _Scalars because numpy only supports them for certain datatypes/values.
 - None: converts to NaN for floats and errors for ints
 - bytes (and other stringy types): numeric string conversion and error on non number chars
If we want, I can update get_mv() to support the exact numpy behavior

Still need numpy for float16. struct supports half but memoryview.cast() does not. https://github.com/python/cpython/pull/96738. available is python 3.12

When a nested empty list ([[]]) is converted with np.array(), the data attribute is empty, but it still has a shape. When we try to cast an empty memoryview to the correct shape, we get an error (see below). numpy is probably able to set the shape of an empty memoryview because it uses the buffer protocol in Cython. We could potentially make our own memoryview class and implement __buffer__(), but are nested empty lists necessary? Another possibility is writing memory directly using ctypes, but I really have no clue if that would actually work.

Currently, although we are able to create a tensor of this type, basic arithmetic does not work. I argue it would better to error upon creation (in get_mv), and not support this type of list data at all.

```
import numpy as np

print(Tensor(np.array([[]])).shape)
print(np.array([[]]).shape)
print(np.array([[]]).data.shape)
try: Tensor(np.array([[]])).numpy().data.cast('i', np.array([[]]).shape)
except Exception as e: print(e)

print(np.array([[]]) + np.array([[1]]))
print('+'*30)
print((Tensor(np.array([[]])) + Tensor(np.array([[1]]))).numpy())

(1, 0)
(1, 0)
(1, 0)
memoryview: cannot cast view with zeros in shape or strides
[]
++++++++++++++++++++++++++++++
  File "/Users/mattdipaolo/.work/pg.py", line 12, in <module>
    print((Tensor(np.array([[]])) + Tensor(np.array([[1]]))).numpy())
           ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/tensor.py", line 756, in __add__
    def __add__(self, x) -> Tensor: return self.add(x)
                                           ^^^^^^^^^^^
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/tensor.py", line 708, in add
    return mlops.Add.apply(*self._broadcasted(x, reverse)) if x.__class__ is Tensor or x else self
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/tensor.py", line 698, in _broadcasted
    if xshape != shape_ret: x = x.expand(shape_ret)
                                ^^^^^^^^^^^^^^^^^^^
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/tensor.py", line 274, in expand
    def expand(self, shape, *args) -> Tensor: return mlops.Expand.apply(self, shape=tuple([x if x != -1 else s for s,x in zip(self.shape, argfix(shape, *args))]))
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/tensor.py", line 30, in apply
    ret = Tensor(ctx.forward(*[t.lazydata for t in x], **kwargs), device=ctx.device, requires_grad=ctx.requires_grad)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/mlops.py", line 170, in forward
    return x.expand(shape)
           ^^^^^^^^^^^^^^^
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/lazy.py", line 293, in expand
    return self._movement_op(self.st.expand(arg), MovementOps.EXPAND, arg)
                             ^^^^^^^^^^^^^^^^^^^
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/shape/shapetracker.py", line 152, in expand
    def expand(self, new_shape: Tuple[sint, ...]) -> ShapeTracker: return ShapeTracker(self.views[0:-1] + (self.views[-1].expand(new_shape), ))
                                                                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mattdipaolo/Projects/tinygrad/tinygrad/shape/view.py", line 131, in expand
    assert all((s == x == 0) or (s > 0 and (x % s) == 0) for s,x in zip(self.shape, new_shape)), f"can't expand {self.shape} into {new_shape}"
AssertionError: can't expand (1, 0) into (1, 1)


```

#2649


geohot:
21 lines and typos in code!

imaolo:
I think this is ready. For _Scalars, I removed bytes because it is not a scalar and None because numpy only supports None for certain datatypes. It converts to NaN for floats and errors for ints.

If we want I can update get_mv() to support the exact NaN behavior.

PR #2668, opened 2023-12-07
support vals on WebGPU, run more tests

wpmed92:
- support `vals` using uniforms, test_symbolic passes
- brought back `render_store` because the explicit cast is needed in some cases (example: assigning a cmplt bool result to a i32 output buffer)

### DIFF
--- a/test/test_nn.py
+++ b/test/test_nn.py
@@ -11,6 +11,7 @@ import pytest
 pytestmark = [pytest.mark.exclude_cuda]
 
 class TestNN(unittest.TestCase):
+  @unittest.skipIf(Device.DEFAULT == "WEBGPU", "no int64 on WebGPU")
   def test_sparse_cat_cross_entropy(self):
     input = torch.randn(3, 5)
     target = torch.empty(3, dtype=torch.long).random_(5)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,17 +246,15 @@ jobs:
     #- name: Run webgpu pytest
     #  run: WEBGPU=1 WGPU_BACKEND_TYPE=Metal python -m pytest -n=auto
     - name: Run selected webgpu tests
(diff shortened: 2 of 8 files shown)


chenyuxyz:
cool! you can remove the `@unittest.skipIf(Device.DEFAULT in ["WEBGPU"], f"{Device.DEFAULT} is not supported")` in test_symbolic_ops.py too

8 minutes is too long for CI, can you disable some slow ones that is not too webgpu specific?

test/test_nn.py
```diff
@@ -11,6 +11,7 @@
 pytestmark = [pytest.mark.exclude_cuda]
 
 class TestNN(unittest.TestCase):
+  @unittest.skipIf(Device.DEFAULT == "WEBGPU", "no int64 on WebGPU")
```
geohot, line 14: Can we just change the test to use torch.int32?

wpmed92: nn.crossEntropyLoss expects target as Long:

```
 return torch._C._nn.cross_entropy_loss(input, target, weight, _Reduction.get_enum(reduction), ignore_index, label_smoothing)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: expected scalar type Long but found Int
```

wpmed92:
@chenyuxyz Added back `test_symbolic_ops.py` and removed slow tests, now down to 4m.

wpmed92:
@geohot as @chenyuxyz pointed out tests were running too slow so I'm keeping these
`test/test_ops.py test/test_dtype.py test/test_jit.py test/test_symbolic_ops.py test/test_symbolic_jit.py test/test_linearizer.py test/test_linearizer_failures.py test/test_nn.py`

Also, moving back llama compile to webgpu runner otherwise metal runner will be slow.

geohot:
I liked the blacklist better than the whitelist to have new tests be run, but I think this PR is good as is for now.

PR #2670, opened 2023-12-07
add tinygrad.features.graph to packages

Qazalin:
to repro, try installing tinygrad master and `import tinygrad.features.graph`

i think it was broken even before this https://github.com/tinygrad/tinygrad/commit/3128da832d5ad14ef2dd7aff48581fc966d7c8d9 because runtime.graph didnt exist in setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(name='tinygrad',
       license='MIT',
       long_description=long_description,
       long_description_content_type='text/markdown',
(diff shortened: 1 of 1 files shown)


Qazalin:
cant run cuda as a result:
image

for a longterm fix, we could add an `__init__.py` to every directory and use `setuptools.find_packages("tinygrad")` instead of hardcoding

PR #2678, opened 2023-12-08
fromCPU takes memoryview and add format str to DType

imaolo:
#2649


imaolo:
> This isn't removing numpy from the dtypes, and why do the formats have to be in the dtypes?

The formats have to be in dtypes because we convert the memoryview's format to a DType. see device.py:83 (from_format()). We will also need the formats for struct.pack() in #2660.

Leaving numpy in dtypes for now because removing them will require more complex changes, and this is just one step.

PR #2679, opened 2023-12-08
More simplification in to_image_idx and symbolic

ZenginU:
So,
* In to_image_idx idxy may not always be SumNode so it can be simplified with valid as well.
* In symbolic left over part in SumNode \_\_lt__ may be LtNode with MulNode so simplify it as well


tinygrad/features/image.py
```diff
@@ -103,11 +103,12 @@ def to_image_idx(base_shape:Tuple[int, ...], idxy:Node, valid:Node) -> Tuple[Tup
   idx = (idxy // 4) % base_shape[1]
   idy = (idxy // (4 * base_shape[1]))
 
-  if valid.min == 0 and isinstance(idxy, SumNode):
+  if valid.min == 0:
     nodes = valid.nodes if isinstance(valid, AndNode) else [valid]
     val_dict: Dict[Node, Any] = {}
     # TODO: is this correct? should it check there's only one variable from each component?
```
chenyuxyz, line 109: can you check this? i understand the intention, but it's hard to review a block of change that has `TODO: is this correct?` on it

ZenginU: @chenyuxyz 
It is true that it has TODO but it is with the .vars() part of the line. I can swap the TODO and the idxy_nodes line

chenyuxyz: so at the glance it's not simpler. it moved one if condition to another place, added a line inside a block with `is this correct` on it. not clear if it simplifies the kernel or not, or change any kernel at all, or just a symbolic corner case that we don't really care about. and i might just refactor the whole thing if i try to understand what it's doing

ZenginU: Hello again,
So I was the one who added the isinstance(idxy) check because of the mypy. I realized that there exist cases where idxy is MulNode or Variable (I assumed idxy can only be SumNode). Also I believe that TODO check is for (I did not put it) idxy SumNode may not always consists of MulNodes and Variables with one variable.

chenyuxyz: cool! i think it's fine to merge given that it's undoing a previous hack

PR #2680, opened 2023-12-08
Update yolov3.py

nickovaras:
The current yolov3 example is broken with the current implementation of of fetch in the helpers. I was tempted to fix the helpers instead but that could have just as well broken other examples.

### DIFF
--- a/examples/yolov3.py
+++ b/examples/yolov3.py
@@ -38,7 +38,7 @@ def show_labels(prediction, confidence=0.5, num_classes=80):
 def add_boxes(img, prediction):
   if isinstance(prediction, int): # no predictions
     return img
-  coco_labels = fetch('https://raw.githubusercontent.com/pjreddie/darknet/master/data/coco.names')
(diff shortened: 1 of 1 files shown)


nickovaras:
For reference: Before the fix, essentially you hit the PosixPath errors because fetch is returning the path to the file instead of its contents

```
~/tinygrad$ python3 examples/yolov3.py 
https://raw.githubusercontent.com/pjreddie/darknet/master/cfg/yolov3.cfg: 100%|█
  File "/home/nicolas/tinygrad/examples/yolov3.py", line 372, in <module>
    model = Darknet(fetch('https://raw.githubusercontent.com/pjreddie/darknet/master/cfg/yolov3.cfg'))
  File "/home/nicolas/tinygrad/examples/yolov3.py", line 203, in __init__
    self.blocks = parse_cfg(cfg)
  File "/home/nicolas/tinygrad/examples/yolov3.py", line 152, in parse_cfg
    lines = cfg.decode("utf-8").split('\n')
AttributeError: 'PosixPath' object has no attribute 'decode'

PR #2686, opened 2023-12-09
Implement einsum

### LINKED ISSUE #2665: Add Tensor.einsum
Then [upstream it to einops](https://github.com/arogozhnikov/einops/pull/297), by overriding [this method](https://github.com/arogozhnikov/einops/blob/fcd36c9b017d52e452a301e75c1373b75ec23ee0/einops/_backends.py#L133).

blueridanus:
Closes #2665. Is this desired in core? Seems appropriate to me, but it doesn't take self, so if LOC is a problem, could be in extra.

- [x] untested einsum impl
- [x] add a few simple tests
- [x] cover every case [here](https://rockt.github.io/2018/04/30/einsum)
- [x] make sure transpose shenanigans are indeed covered by tests
- [x] add mustfails for bad syntax, mismatched shapes
- [x] check if it works in einops
- [x] add more helpful error msg on dim mismatch
- [x] perf comparisons

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -517,14 +517,14 @@ class Tensor:
     lhs = [sorted(enumerate(s), key=lambda e:e[1]) for s in lhs.split(',')]
     rhs = sorted(enumerate(rhs), key=lambda e:e[1])
 
-    assert len(xs) == len(lhs), f"number of args doesn't match number of operands in formula, expected {len(lhs)}, got {len(xs)}"
+    assert len(xs) == len(lhs), f"number of inputs doesn't match number of operands in formula, expected {len(lhs)}, got {len(xs)}"
 
     lhs, rhs = [[list(zip(*l)) for l in lhs], list(zip(*rhs)) or [[], []]]
     dims = {}
     for i, l in enumerate(lhs):
       for order, letter in enumerate(l[1]):
         if letter not in dims: dims[letter] = xs[i].shape[l[0][order]]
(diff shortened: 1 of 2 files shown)


blueridanus:
Would be neat to see how the einsum version of e.g. matmul compares performance wise.

tinygrad/tensor.py
```diff
@@ -510,6 +510,31 @@ def argmax(self, axis=None, keepdim=False):
     return self.shape[axis]-idx.max(axis=axis, keepdim=keepdim)-1
   def argmin(self, axis=None, keepdim=False): return (-self).argmax(axis=axis, keepdim=keepdim)
 
+  @staticmethod
+  def einsum(formula, *xs) -> Tensor:
+    if isinstance(xs[0], list): xs = tuple(xs[0])
```
geohot, line 515: Can fixarg do this?

geohot:
Yea it should be in core, as beautiful as possible.

geohot:
> Would be neat to see how the einsum version of e.g. matmul compares performance wise.

Perf will be identical, test it :)

tinygrad/tensor.py
blueridanus: yep, nice

blueridanus:
Works in einops:
image

blueridanus:
> > Would be neat to see how the einsum version of e.g. matmul compares performance wise.
> 
> Perf will be identical, test it :)

Seems pretty close:
image

geohot:
Cool!

PR #2688, opened 2023-12-09
Formatted test_indexing

geohotstan:
This PR does mainly formatting
- 2 spaces
- translated torch code to tinygrad code (some torch functions are left as TODOs up top)
- annotated still disabled tests using TODO and triple quotation
- enabled easy to enable tests that don't require much review
- snuck 1 very simple `IndexError("slice() cannot be applied to a 0-dim tensor.")` test in

current total enabled tests: 24

-----
after this is merged, I'll start enabling more of these disabled tests
I already done some on other branches

### DIFF
--- a/test/imported/test_indexing.py
+++ b/test/imported/test_indexing.py
@@ -1,6 +1,6 @@
 # test cases are modified from pytorch test_indexing.py https://github.com/pytorch/pytorch/blob/597d3fb86a2f3b8d6d8ee067e769624dcca31cdb/test/test_indexing.py
 
-import math, unittest, random
+import math, unittest, random, copy
 import numpy as np
 
 from tinygrad.tensor import Tensor, dtypes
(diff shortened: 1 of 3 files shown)


test/imported/test_indexing.py
```diff
@@ -27,7 +29,42 @@ def set_(reference: Tensor, shape, strides, offset):
   assert strided.lazydata in reference.lazydata.base.views, "base.views should contain strided.lazydata"
   return strided
 
-def make_tensor(shape, dtype:dtypes, noncontiguous):
+# TODO tries to mimic .detach().copy() or just .copy() behavior
+# for torch.copy() the resulting tensor.data_ptr() is different than the original
+# some random medium article says .clone() is a shallow copy, so maybe this is correct?
+# https://discuss.pytorch.org/t/clone-and-detach-in-v0-4-0/16861/4
+def clone(original:Tensor):
```
chenyuxyz, line 36: we probably don't need clone #2628. because tinygrad is lazy. we might not want to follow some torch expectation (say they make a copy) if we don't need to

chenyuxyz:
fine to merge. we might have different clone behavior compared to torch but that can be handled in a different pr

PR #2696, opened 2023-12-10
Training 3D-UNet: New Branch (old: #2437)

dxlnr:
#1637 

Switched from #2437 to new branch as the upstream was broken. 

Update on training: Ive been able to train a model to roughly 60% DICE score (must reach ~0.9) in 200 epochs with the kits19 dataset on a single NVIDIA RTX 4090.

Only problem it is done with a smaller data input size (64,64,64) compared to the necessary (128,128,128) for mlperf. The mlperf input size of (128,128,128) has been consistently causing OOM problems (also described in #2437). 

For reference: The (128,128,128) input size works with PyTorch (& CUDA) and is roughly 5x faster. Epoch time using tinygrad (&CUDA) ~110s whereas using torch 20s. 

This might actually go down using JIT but I also encountered problems with the JIT and CUDA (some cuda graph error). 

I will continue training the model to push it over the 0.908 necessary mean dice score threshold but I guess to be mlperf ready, some extra steps have to be done.


gijskoning:
I was working on this bounty a few months ago (https://github.com/tinygrad/tinygrad/pull/1928) and had the same OOM issue, even when running on a very big GPU. This patch https://github.com/tinygrad/tinygrad/pull/1778 fixed my problem, it is still not in master so that might help you get this bounty.

geohot:
New rule on mlperf training bounties. in order to lock the bounty, you must post weights that you trained in tinygrad that meet the accuracy requirement. Then we can look at speed and mergablity, but until there's weights, no lock.

PR #2699, opened 2023-12-11
WIP: change some linter steps to use pre-commit hooks

allenzhao:
As I'm unsure of a better way to do it, I've done it based on my understanding of the goal, which is to reuse some steps from pre-commit hooks?

Several questions: 
1. The mypy call is different in two places. Several other inconsistencies, for example, as I'm not sure about the two different calls of pylint, so I only replaced the one that is the same command. 
2. Currently I didn't change the steps much in the file, I just kept the order and used the pre-commit action. Do we prefer a single step of call to this action with multiple checks, or it's fine to just leave this as-is?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,6 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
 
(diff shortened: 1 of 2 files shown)


allenzhao:
Closing for now, seems there are some major issues with how I've changed the action....

PR #2702, opened 2023-12-11
readonly memoryview

gabelerner:
wrapping a memoryview around the readonly memoryview's obj makes a new writeable memoryview. unsure how to test beyond the test suite.

```
>>> buf = bytearray([1,2,3,4,5])
>>> mvro = memoryview(buf).toreadonly()
>>> mvro.readonly
True
>>> mvw = memoryview(mvro.obj)
>>> mvw.readonly
False
>>> mvro.obj == mvw.obj
True
```

NOTE: `memoryview(mv)` on its own doesn't work because cpython [copies](https://github.com/python/cpython/blob/main/Objects/memoryobject.c#L548) the `readonly` bit

i also tried:
- ctypes incantations, didn't get anywhere
- can probably use `ctypes.pythonapi` methods but that felt more hacky than this

### DIFF
--- a/docs/abstractions2.py
+++ b/docs/abstractions2.py
@@ -15,8 +15,8 @@ a = MallocAllocator.alloc(4)
 b = MallocAllocator.alloc(4)
 
 # load in some values (little endian)
-MallocAllocator.copyin(a, bytearray([2,0,0,0]))
-MallocAllocator.copyin(b, bytearray([3,0,0,0]))
+MallocAllocator.copyin(a, memoryview(bytearray([2,0,0,0])))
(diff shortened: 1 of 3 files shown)


gabelerner:
Sigh this test fails because it offsets and when you wrap a memoryview around another memoryview's obj there's no way i can see to get the offset to pass to `from_buffer`

```
    def test_copy_from_disk(self):
      t = Tensor.randn(30, device="CPU").to(f"disk:{temp('test_copy_from_disk')}")
      a = t[10:20]
      dev = a.to(Device.DEFAULT)
     np.testing.assert_allclose(a.numpy(), dev.numpy())
```

PR #2704, opened 2023-12-11
Lean for mergeability of views

PaulGustafson:
Just starting with some basic definitions to prove the easiest case.  Also would be good to get a better idea of the right definition of "mergeability of shapetrackers." Currently working under the assumption that it just means there is a order-preserving bijection of underlying indices that preserves the masks.  In this case, it's clear that there needs to be a condition on the masks for the shapetrackers to be mergeable -- example: View(shape=(6), mask=(0,4)) can't be reshaped into a single mask on View(shape=(2,3))


geohot:
Closing, no progress in over a week

PR #2706, opened 2023-12-11
Add simple fancy indexing exceptions

geohotstan:


test/test_ops.py
```diff
@@ -1357,8 +1357,12 @@ def test_slice_fancy_indexing_tuple_with_tensors(self):
     helper_test_op([(2,5,6,5,3,4)], lambda x: x[(a,(1,1))], lambda x: x[(i,(1,1))])
     helper_test_op([(2,5,6,5,3,4)], lambda x: x[(a,b,c,d,e)], lambda x: x[(i,j,k,o,p)])
 
-  def test_slice_fancy_indexing_errors(self): ...
-    # TODO: currently we not support IndexError for out of bounds idx values
+  def test_slice_fancy_indexing_errors(self):
+    a = Tensor.ones(10,11,12)
+    with self.assertRaises(IndexError): a[Tensor(1.1)] # tensors used as indices must be int or bool tensors
```
chenyuxyz, line 1362: nit: i find putting test case comments in a separate line above the code more readable. also fits the same stype as other tests in this file. fine to change in the next pr

tinygrad/tensor.py
```diff
@@ -375,18 +375,21 @@ def __getitem__(self, indices: Union[int, slice, Tensor, None, List, Tuple]) ->
...
       arange = [Tensor.arange(ret.shape[d], dtype=dtypes.int32, requires_grad=False, device=self.device).reshape(*[1]*sd, ret.shape[d], *[1]*(ret.ndim + max_dim - n - sd - 1)) for n,(sd,d) in enumerate(zip(sum_dim, tdim))]
       first_idx = [idx[0].reshape(*[1]*tdim[0], *[1]*(1 + max_dim - idx[0].ndim), *idx[0].shape, *[1]*(ret.ndim - tdim[0] - 1))]
       rest_idx = [i.reshape(*[1]*tdim[0], *[1]*(max_dim - i.ndim), *i.shape, *[1]*(ret.ndim - tdim[0] - n)) for n,i in enumerate(idx[1:], 1)]
-      idx = first_idx + rest_idx
+      reshaped_idx = first_idx + rest_idx
       ret = ret.reshape(*ret.shape[:sum_dim[0]+1], *[1]*max_dim, *ret.shape[sum_dim[0]+1:])
 
       # iteratively eq -> mul -> sum fancy index
-      for a,i,sd in zip(arange, idx, sum_dim): ret = (a==i).mul(ret).sum(sd)
+      try:
```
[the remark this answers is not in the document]
chenyuxyz: we should have better errors than all these asserts, but i think it's fine for now

test/imported/test_indexing.py
```diff
@@ -1085,15 +1085,15 @@ def test_index_src_datatype(self):
   def test_int_indices2d(self):
     # From the NumPy indexing example
     x = Tensor.arange(0, 12).reshape(4, 3)
-    rows = Tensor([[0, 0], [3, 3]])
-    columns = Tensor([[0, 2], [0, 2]])
+    rows = Tensor([[0, 0], [3, 3]], dtype=dtypes.int32)
```
chenyuxyz, line 1088: this looks like a bug in core tinygrad that we default convert dtype to float. i will check this tomorrow

chenyuxyz (commented):
the changes look good. i think we should fix default convert to float before merging this, and we don't need to specify dtypes.int32 when it's from an int list. sorry to block you, i will check this tomorrow

geohotstan:
> the changes look good. i think we should fix default convert to float before merging this, and we don't need to specify dtypes.int32 when it's from an int list. sorry to block you, i will check this tomorrow

s'all goooood :D
I actually had a branch with the int32 tensor init change already
I didn't put up a PR because I thought it was a very minor issue and it might interfere with Tensor.default_type...
I'll look into this a little more

going to reopen PR when #2709 is fixed

chenyuxyz:
so the index and default dtype thing is deeper than i thought, and i might need another day or two to fully fix it. don't want to block you for too long. if you want i am fine explicitly setting dtypes for int for now. when i am done i will do a full sweep to remove the redundant dtype arg

geohotstan:
> so the index and default dtype thing is deeper than i thought, and i might need another day or two to fully fix it. don't want to block you for too long. if you want i am fine explicitly setting dtypes for int for now. when i am done i will do a full sweep to remove the redundant dtype arg

It's no problem, I'm still trying to figure out how to do bool indexing. I can work on that on a separate branch in the meantime.

I also tried a dtype refactor based on JAX type promo rules using new abstractions, just to learn. But yeah it was quite hard cuz of how integrated dtype is in the rest of the code base. I gave up after trying yesterday :D
GLGL!!! 🚀

tinygrad/tensor.py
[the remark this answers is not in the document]
geohotstan: I have wrote an independent checker while doing bool indexing.
I'm going to leave this try block here just as is for now.

extra/onnx_ops.py
```diff
@@ -425,10 +425,10 @@ def _nearest_mode(x_resized: Tensor, nearest_mode: str, x_len):
     elif nearest_mode == "round_prefer_ceil": ret = _round(x_resized, 0.5, "round_up")
     elif nearest_mode == "floor": ret = x_resized.floor()
     elif nearest_mode == "ceil": ret = x_resized.ceil()
-    return ret.clip(0, x_len-1)
+    return ret.clip(0, x_len-1).cast(dtypes.int32)
```
chenyuxyz, line 428: is it better to do `cast` before `clip` if we are clipping with int?

tinygrad/tensor.py
```diff
@@ -386,18 +386,22 @@ def __getitem__(self, indices: Union[int, slice, Tensor, None, List, Tuple]) ->
...
+        for a,i,sd in zip(arange, reshaped_idx, sum_dim): ret = (a==i).mul(ret).sum(sd)
+      except AssertionError as exc: raise IndexError(f"shape mismatch: indexing tensors could not be broadcast together with shapes {', '.join([str(i.shape) for i in idx])}") from exc   # noqa: E501
```
chenyuxyz, line 404: new line should have max 150 chars!

extra/onnx_ops.py
geohotstan: Trueeee
good call.
the nearest modes should resolve ret to an integer anyways so casting before clip would be good.

tinygrad/tensor.py
geohotstan: m'bad.
I should've reread my PR...
brain a lil fried from reshaping so many tensors.

tinygrad/tensor.py
```diff
@@ -386,18 +386,23 @@ def __getitem__(self, indices: Union[int, slice, Tensor, None, List, Tuple]) ->
...
-      idx = first_idx + rest_idx
+      reshaped_idx = first_idx + rest_idx
       ret = ret.reshape(*ret.shape[:sum_dim[0]+1], *[1]*max_dim, *ret.shape[sum_dim[0]+1:])
 
       # iteratively eq -> mul -> sum fancy index
-      for a,i,sd in zip(arange, idx, sum_dim): ret = (a==i).mul(ret).sum(sd)
+      try:
+        for a,i,sd in zip(arange, reshaped_idx, sum_dim): ret = (a==i).mul(ret).sum(sd)
+      except AssertionError as exc:
+        raise IndexError(f"shape mismatch: indexing tensors could not be broadcast together with shapes {str([i.shape for i in idx])[1:-1]}") from exc
```
chenyuxyz, line 405: i think fstring would call the `str` (or `repr`) for you? also typo `broadcasted`

Issue #2709, opened 2023-12-11
update default dtype for int / int list input


chenyuxyz:
we have dtype now and we should not convert int to float by default. numpy and torch both use int64 for int

chenyuxyz:
we will use int32 as the default for int. the main use case is GPU and GPU does not int64

PR #2710, opened 2023-12-11
add int32 default for Tensor(int) and Tensor(List[int)

geohotstan:
#2709 

TODO 
- need to check if this interferes with Tensor.default usage
- add tests

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -59,18 +59,16 @@ class Tensor:
     # internal variables used for autograd graph construction
     self._ctx: Optional[Function] = None
     if isinstance(data, LazyBuffer): assert dtype is None or dtype == data.dtype, "dtype doesn't match, and casting isn't supported"
-    elif data is None or isinstance(data, list):
-      assert dtype is None or dtype.np is not None, f"{dtype} doesn't have a numpy dtype"
-      data = LazyBuffer.fromCPU(np.array([] if data is None else data, dtype=(dtype or dtypes.int32 if all(isinstance(i, int) for i in data) else Tensor.default_type).np))
     elif isinstance(data, int): data = LazyBuffer.loadop(LoadOps.CONST, tuple(), dtype or dtypes.int32, device, data)
     elif isinstance(data, float): data = LazyBuffer.loadop(LoadOps.CONST, tuple(), dtype or Tensor.default_type, device, data)
     elif isinstance(data, bytes): data = LazyBuffer.fromCPU(np.frombuffer(data, np.uint8))
-    elif isinstance(data, np.ndarray):
+    else:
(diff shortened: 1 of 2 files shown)


geohotstan:
for reference: torch.zeros, torch.ones, torch.full
```
# defaults to float32
In [8]: torch.ones(5).dtype                                                                                                                                                        
Out[8]: torch.float32    
In [9]: torch.zeros(5).dtype
Out[9]: torch.float32

# full defaults to int
In [16]: torch.full(size=(5,), fill_value=4).dtype                          
Out[16]: torch.int64
```

geohotstan:
note:
if we do decide to do this
all Tensor() inits will have to be
`Tensor([1.])` instead of `Tensor([1])`
to init a float tensor

(saw torch do this a few times in test_indexing)

tinygrad/tensor.py
```diff
@@ -171,10 +168,10 @@ def rand(*shape, **kwargs):
   def full(shape:Tuple[sint, ...], fill_value, **kwargs): return Tensor(fill_value, **kwargs).reshape([1]*len(new_shape := argfix(shape))).expand(new_shape)
 
   @staticmethod
-  def zeros(*shape, **kwargs): return Tensor.full(argfix(*shape), 0, **kwargs)
+  def zeros(*shape, **kwargs): return Tensor.full(argfix(*shape), 0, dtype=kwargs.pop("dtype", Tensor.default_type), **kwargs)
```
geohotstan, line 171: for reference: torch.zeros, torch.ones, torch.full
```
# defaults to float32
In [8]: torch.ones(5).dtype                                                                                                                                                        
Out[8]: torch.float32    
In [9]: torch.zeros(5).dtype
Out[9]: torch.float32

# full defaults to int
In [16]: torch.full(size=(5,), fill_value=4).dtype                          
Out[16]: torch.int64
```

geohotstan:
There are a lot of changes around the code base that would need to be made to adapt to this
I only changed enough to get a green CI cuz green CI is nice to look at

This is also kinda an invitation for bugs since just having everything in float makes it so that we don't have to worry about dtype shannanigans (I vaguely remember geohot saying something a long time ago about having everything in float)
idk

if the current approach is ok (doing Tensor(1.) for float or adding dtype=dtypes.float), and having int or list of int default to int32 is what we want, I could go through the codebase and spot places that need to be changed to float.
otherwise feel free to close

chenyuxyz:
thanks for initiating this. there's a deeper problem here that we are not auto cast int to float for the ops like DIV, EXP etc. changing the test callers does not fix the issue. i am looking into that now

geohotstan:
https://github.com/chenyuxyz/tinygrad/pull/28  🚀 🚀 🚀

PR #2711, opened 2023-12-11
combine expressions and some cleanup in yolov8 and unet3d

VicKnightDR:
Combined some lines to eliminate unnecessary variable assignment (bad for total inference time), I believe these changes do not reduce readability.

Concatenate some op. calls (remove occurrences of calling .sequential twice in a row in same line).

I didn't understand why 3d UNet's final conv layer was created as a dictionary. Maybe it's a personal preference, but I believe having it just as a layer improves readability.


geohot:
This is a whitespace/style PR and will be closed as such.

PR #2712, opened 2023-12-11
WebGPU inf hack that works in Chrome and wgpu

wpmed92:
The new `inf` constant `1.0/0.0` gets all `inf` tests passing in `wgpu`, but in Chrome it results in a shader creation error, which I discovered when running stable diffusion:
_Tint WGSL reader failure: :2:35 error: '1.0 / 0.0' cannot be represented as 'abstract-float'_

This is consistent with [specs](https://www.w3.org/TR/WGSL/#floating-point-evaluation), stating:

> A [shader-creation error](https://www.w3.org/TR/WGSL/#shader-creation-error) results if any [const-expression](https://www.w3.org/TR/WGSL/#const-expressions) of floating-point type overflows or evaluates to NaN or infinity.

I came up with a hack that works on both wgpu, and Chrome. The idea is to prevent the compiler from detecting (1.0/0.0) as a `const-expression`. By moving the division inside a function, and passing in one of the operands as a parameter `inf(1.0),` the compiler will not detect it as a `const-expression`, and works in Chrome as well as in wgpu.

(I checked and `fn inf() -> f32 { return 1.0/0.0; }` does not work unfortunately, you have to pass one of the numbers as a parameter so that the expression is not a const-expression)


chenyuxyz:
i can merge this. can you sync to master?

wpmed92:
@chenyuxyz just synced

wpmed92:
@chenyuxyz Just synced

PR #2714, opened 2023-12-11
Add linespace

blankmail1331:
In order to follow the simplest examples available on the [pytorch site](https://pytorch.org/tutorials/beginner/pytorch_with_examples.html) let's implement `linespace` method.

Pros:
- easier way or creating test data,
- improved usability (no need to use `np`),
- easier entry for non technical users, which may result in a better framework adoption.

Cons:
- +6 lines to the `tensor.py`

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -181,6 +181,11 @@ class Tensor:
     return Tensor.full((math.ceil((stop-start)/step),), step, **kwargs).cumsum() + (start - step)
 
   @staticmethod
+  def linspace(start: Union[int, float], stop: Union[int, float], steps: int, **kwargs):
+    assert steps > -1, "number of steps must be non-negative"
+    return Tensor([], **kwargs) if steps == 0 else Tensor.full((1,), start, **kwargs) if steps == 1 else Tensor.full((steps,), start, **kwargs) + Tensor.arange(steps, **kwargs) * ((stop-start) / (steps-1))
+
+  @staticmethod
   def eye(dim:int, **kwargs): return Tensor.full((dim,1),1,**kwargs).pad(((0,0),(0,dim))).reshape(dim*(dim+1)).shrink(((0,dim*dim),)).reshape(dim, dim)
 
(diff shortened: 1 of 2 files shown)


blueridanus (changes requested):
Needs to use ops instead of python list.

e.g: 
```python
def linspace(start, end, steps): return Tensor.full(steps, start) + Tensor.arange(steps) * ((end-start) / (steps-1))
```

blankmail1331:
Thanks for a hint @blueridanus !
May I ask you take a look one more time?

tinygrad/tensor.py
```diff
@@ -180,6 +180,11 @@ def arange(start, stop=None, step=1, **kwargs):
     if stop is None: stop, start = start, 0
     return Tensor.full((math.ceil((stop-start)/step),), step, **kwargs).cumsum() + (start - step)
 
+  @staticmethod
+  def linspace(start: Union[int, float], stop: Union[int, float], steps: int, **kwargs):
+    assert steps > 1, f"steps must be greater than 1, [{steps}] was provided"
```
geohot, line 185: Steps can be 1 in torch

blankmail1331: Right, it didn't make too much sense to me (who uses it for 1 or 0 steps?!) so wanted to slightly adjust it.
Anyway updated the code + added few corner cases to match the torch api. 
@geohot WDYT?

geohot:
Can't read it

geohot:
Closing this.

1) There's a typo in the title. That shows what I need to know about the quality of this PR
2) I want you to think through the cases. The torch example includes 1. It doesn't include 0
3) Do you need that assert?

blankmail1331:
> Can't read it

Changed it to:

```
  @staticmethod
  def linspace(start: Union[int, float], stop: Union[int, float], steps: int, **kwargs):
    assert steps > -1, "number of steps must be non-negative"
    if steps == 0:
      return Tensor([], **kwargs)
    if steps == 1:
      return Tensor.full((1,), start, **kwargs)
    step_size = (stop - start) / (steps - 1)
    return Tensor.full((steps,), start, **kwargs) + Tensor.arange(steps, **kwargs) * step_size
```
The reason why I did it like was purely due to the limit in the number of lines.

PR #2715, opened 2023-12-11
cuda set context before sync

nimlgen:
Set device's context to synchronize


geohot:
I don't think this should have a helper. I understand it's more lines, but this is harder to read and reason about, which is all that really matters in the end.

PR #2716, opened 2023-12-11
fix dtypes helpers for integers

Qazalin:
1. dtypes.is_dtype should always check the scalar version

this caused a regression in rendering vectorized int constants. we should figure out why CI was green.

example kernel diff for `TestOps.test_simple_padding_conv2d`
```diff
__kernel void r_4_4_4_3(write_only image2d_t data0, read_only image2d_t data1, read_only image2d_t data2) {
const sampler_t smp = CLK_NORMALIZED_COORDS_FALSE | CLK_ADDRESS_CLAMP | CLK_FILTER_NEAREST;
  int lidx0 = get_local_id(0); /* 4 */
  float4 acc0 = (float4)(0.0f,0.0f,0.0f,0.0f);
  float4 acc1 = (float4)(0.0f,0.0f,0.0f,0.0f);
  float4 acc2 = (float4)(0.0f,0.0f,0.0f,0.0f);
  float4 val0 = read_imagef(data1, smp, (int2)((lidx0+(-1)),0));
-   float4 val1 = read_imagef(data2, smp, (int2)(false,false));
+   float4 val1 = read_imagef(data2, smp, (int2)(0,0));
    float4 val2 = read_imagef(data2, smp, (int2)(1,0));
  // ...
}
```

2. non-32bit ints and unsigned int vector types were broken because aliases didn't exist for them. 

adds unit tests at the dtypes level to fuzz these

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -4,6 +4,7 @@ from tinygrad.helpers import CI, DTYPES_DICT, getenv, DType, DEBUG, ImageDType, 
 from tinygrad import Device
 from tinygrad.tensor import Tensor, dtypes
 from typing import Any, List
+from hypothesis import given, strategies as st
 
 def is_dtype_supported(dtype: DType):
   # for GPU, cl_khr_fp16 isn't supported (except now we don't need it!)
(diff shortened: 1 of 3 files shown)


test/test_dtype.py
```diff
@@ -189,5 +189,13 @@ def test_strs(self):
     self.assertEqual(str(dtypes.imagef((1,2,4))), "dtypes.imagef((1, 2, 4))")
     self.assertEqual(str(PtrDType(dtypes.float32)), "ptr.dtypes.float")
 
+class TestHelpers(unittest.TestCase):
+  def test_helpers(self):
```
chenyuxyz, line 193: consider adding cases for `uint` and `uint.vec` and its interaction with int and float

PR #2718, opened 2023-12-11
dtypes math in ops instead of linearizer

Qazalin:
lifts up promotions from the linearizer to ops
unblocks autocast https://github.com/tinygrad/tinygrad/pull/2710#issuecomment-1850656084 and truediv

pending:
- [ ] defined behavior for FlopCounter on a LocalBuffer, this optimization breaks promotion https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/linearizer.py#L191
- [ ] LazyOp src promotion
- [ ] shouldn't cmplt expect a bool in return?
- [ ] linearizer should fallback to float for image ops


Qazalin:
everything will be in #2800

PR #2720, opened 2023-12-11
Extend available types for `safe_save`

gml16:
Saving tensors of certain types (e.g. doubles, bools, shorts) currently fails. I have added a unit test to show that they fail. Adding the missing dtypes to `safe_dtypes` enables the unit test to run successfully. 

Please note that there may be a very good reason that `safe_dtypes` did not already include the dtypes that I added. Please let me know if that is so. The unit test does assert that the saved and loaded tensors are the same, but maybe I am missing something else.

Please feel free to suggest any modifications.


geohot:
Nice! No, no reason it didn't, just no-one did it. Thanks for including a test

PR #2726, opened 2023-12-12
Cleanup code in view.py

grekiki2:
Removed instances of generators being converted to list and then tuple by removing the list part. Might make the code a bit faster as well. 

Some other minor changes as well, should have no effect on code output.


grekiki2:
Tried pretty hard to measure the time difference but it's not obvious it exists. Still feel like the smaller set of changes improves the code though.

chenyuxyz:
no style pr

PR #2728, opened 2023-12-12
remove duplicate typemap in GPU cstyle

Qazalin:
we're already aliasing this https://github.com/tinygrad/tinygrad/blob/master/tinygrad/helpers.py#L168


chenyuxyz:
does this work only because the aliased ones are the last entries in dtypes? DTYPES_DICT is a many-to-1 mapping so the inverse one does not make too much sense

Qazalin:
yeah I'm closing it I think it's a little fragile and type_map is more specific here

PR #2729, opened 2023-12-12
Remove wgsl custom render_for

wpmed92:
- Saved 2 lines by removing custom wgsl render_for


tinygrad/renderer/cstyle.py
```diff
@@ -72,7 +72,7 @@ def render_local(self, name:str, size:int):
     return self.smem_align + self.smem_prefix + f"float {name}[{size}];"
 
   def render_for(self, expr: str, _min:Union[int,str], _max:Union[int,str]) -> str:
-    return f"for (int {expr} = {_min}; {expr} < {_max}; ++{expr}) {{"
+    return f"for ({self.generic_var_prefix if self.generic_var_prefix else 'int'} {expr} = {_min}; {expr} < {_max}; {f'{expr}++' if self.generic_var_prefix else f'++{expr}'}) {{"
```
geohot, line 75: Huh, why the expr++ ++expr thing?

wpmed92: setting default for `generic_var_prefix` affects a alu, define acc, etc, so it will be picked against `dtype.name` in those places:
`kk(f"{lang.generic_var_prefix if lang.generic_var_prefix else dtype.name} {ssa(u,'val')} = {val};")`

wpmed92: setting default for generic_var_prefix affects a alu, define acc, etc, so it will be picked against dtype.name in those places:
kk(f"{lang.generic_var_prefix if lang.generic_var_prefix else dtype.name} {ssa(u,'val')} = {val};")

The ++ thing -> wgsl doesn't have pre-increment, but i can just make it post-increment for all.

wpmed92: We can shave off 2 more lines if we remove the `render_if` which is only custom in wgsl because of the explicit bool cast. We can make that cast for all `render_if`s, which does not add extra lines.

tinygrad/renderer/cstyle.py
```diff
@@ -328,7 +328,8 @@ class WGSLLanguage(CStyleLanguage):
   barrier="workgroupBarrier();"
   generic_var_prefix = "var "
   external_local_bufs = True
-  code_for_op = { **CStyleLanguage().code_for_op, BinaryOps.CMPLT: lambda x,y,dtype: f"f32({x}<{y})", TernaryOps.MULACC: lambda x,y,z,dtype: f"fma({x},{y},{z})", TernaryOps.WHERE: lambda a,b,c,dtype: f"select({c},{b},{a}!=0.)" }
+  code_for_op = { **CStyleLanguage().code_for_op, **{opcode: lambda a,b,dtype,op=op: f"bool(i32({a}){op}i32({b}))" if dtype == dtypes.bool else f"({a}{op}{b})" for opcode,op in [(BinaryOps.MUL,"*"),(BinaryOps.ADD,"+"),(BinaryOps.SUB,"-"),(BinaryOps.DIV,"/")]},
```
wpmed92, line 331: This is needed to convert `bool` binops to `int`, since you can't do `bool` binops in `wgsl`.

tinygrad/renderer/cstyle.py
```diff
@@ -328,7 +328,9 @@ class WGSLLanguage(CStyleLanguage):
   barrier="workgroupBarrier();"
   generic_var_prefix = "var "
   external_local_bufs = True
-  code_for_op = { **CStyleLanguage().code_for_op, BinaryOps.CMPLT: lambda x,y,dtype: f"f32({x}<{y})", TernaryOps.MULACC: lambda x,y,z,dtype: f"fma({x},{y},{z})", TernaryOps.WHERE: lambda a,b,c,dtype: f"select({c},{b},{a}!=0.)" }
+  code_for_op = { **CStyleLanguage().code_for_op, TernaryOps.MULACC: lambda x,y,z,dtype: f"fma({x},{y},{z})", TernaryOps.WHERE: lambda a,b,c,dtype:
+  f"select({c},{b},f32({a})!=0.)", **{opcode: lambda a,b,dtype,op=op: f"bool(i32({a}){op}i32({b}))" if dtype == dtypes.bool else f"({a}{op}{b})"
```
geohot, line 332: I don't think this is right with the bool casting. This should be fixed in a refactor, not with hacks

wpmed92: the cpmplt f32 hack we had is also a hack, but ok i'm putting that back and will solve this in a nice way as a separate refactor. So, this pr will only include the custom for removal.

PR #2730, opened 2023-12-12
simplify type promotion - remove weak types

chenyuxyz:


chenyuxyz:
torch does not have uint64, jax needs a flag and uint64 + int64 -> weak float64. we use float16 for simplicity

PR #2736, opened 2023-12-13
reduce Device class line length

shagler:


geohot:
This is the exact sort of "refactoring" we don't want

PR #2745, opened 2023-12-13
Support Tensor creation from lists and nested lists without numpy

sjmillius:
This is a draft for the second task #2649.
I'm trying to understand the code base; my apologies if this is completely missing the mark.
I'm not trying to lock a bounty, but rather understand by contribute.
This can be extended, but would get some input whether this goes into the correct direction.

This goes via a Python native route when creating tensors from lists. It doesn't remove any support for the creation from np arrays.

Approach:
We can create the pack the data via struct.pack and create a flat, (mutable) memory view on it (mutability required by https://github.com/tinygrad/tinygrad/blob/26f49869f47841817f066aa3357fd9d6aed4f32c/tinygrad/helpers.py#L294)
There's currently no support for bfloat16 (as not exposed in struct.pack), but this could maybe be implemented manually or using a library (as only serialization / deserialization is needed).

The way it is currently implemented is that it walks down the (potentially) nested input and gathers all elements in a flat list (and check that the shape is homogeneous). This is then serialized with struct.pack_into a bytesarray.
One could pack and walk at the same time, but this would assume that we know the dtype beforehand. This is currently the case, but you might want to support inferring the dtype (as numpy does) as well in case no dtype is specified.
Inferring requires looking at all data beforehand. So we could do a second walk there instead of collecting it in a flat list.
Are there any preferences?

The NumpyAllocator (used in the CPU device) assumes an np.ndarray as the buffer, so there's a np.frombuffer view around it in `fromNative` in lazy.py to avoid a copy.

Considered alternatives:
* Use array.array. But that doesn't support float16 and bfloat16.

### DIFF
--- a/test/unit/test_helpers.py
+++ b/test/unit/test_helpers.py
@@ -1,7 +1,7 @@
 import unittest
 import numpy as np
 from PIL import Image
-from tinygrad.helpers import Context, ContextVar, DType, dtypes, merge_dicts, strip_parens, prod, round_up, fetch
+from tinygrad.helpers import Context, ContextVar, DType, dtypes, merge_dicts, strip_parens, prod, round_up, fetch, flat_list
 from tinygrad.shape.symbolic import Variable, NumNode
 
 VARIABLE = ContextVar("VARIABLE", 0)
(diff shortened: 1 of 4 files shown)


chenyuxyz:
closing as stale

PR #2750, opened 2023-12-13
add Tensor.split

### LINKED ISSUE #2677: Add Tensor.split, or let .chunk take in a list as well
Feature parity with https://pytorch.org/docs/stable/generated/torch.split.html.

kevherro:
This change adds support for splitting a tensor into chunks, where each chunk is a view of the original tensor. It matches the [`torch.split`](https://pytorch.org/docs/stable/generated/torch.split.html) API and behavior, while using `tinygrad` internals under the hood (i.e., using `tensor.chunk` when splitting with an `int`).

Resolves #2677


kevherro:
Taking a look at what I can trim to keep LOC under 5000, and I notice that some `Tensor` methods don't validate input (e.g., `Tensor.chunk` does not validate its `num` param, which could lead to a `ZeroDivisionError`) as extensively as `Tensor.split`. Is this deliberately done to keep LOC under 5000?

kevherro:
Simplified this by:
- assuming `sizes` is either an `int` or a `List[int]`, as indicated by the type hints,
- handling negative dimensions, similar to `chunk`
- computing slices using list comprehension, similar to `chunk`

geohot:
Why does split return a tuple and chunk return a list?

Also, can chunk generate the list and call split to make this shorter?

geohot:
Your tuple/list argument makes no sense. Does torch do this?

kevherro:
> Your tuple/list argument makes no sense. Does torch do this?

Yeah, torch does this:

- [`torch.chunk`](https://pytorch.org/docs/stable/generated/torch.chunk.html)
- [`torch.split`](https://pytorch.org/docs/stable/generated/torch.split.html)

geohot:
Meh, if it matches torch, LGTM

PR #2755, opened 2023-12-14
updating to work with new internal apis

jaredeh:
Just cleaning up this example which has been broken for a while.

`export_model()` doesn't quite work for this example I didn't see a really clean way of merging the needs of compile_efficientnet.py vs compile_tensorflow.py.  There are two things I'm doing in response to the current version I feel are a little hacky.

1) run_onnx != model - I didn't see an example of using an onnx model as a first class Tinygrad model.  So I make a dumb little class TinyOnnx as a hack.

2) runtime model loading - One of the goals here is to be able to load weight at runtime, but the default is to embedded them into the sourcecode.  I just don't pass `bufs_to_save` to `export_model_clang()` which works for me.

I don't see any other users that would have similar needs so I think these hacks are probably okay.

The other think is this line works but just seems wrong:
` tinygrad_output = the_output[0].numpy()[0].tolist()`

### DIFF
--- a/examples/compile_tensorflow.py
+++ b/examples/compile_tensorflow.py
@@ -2,15 +2,14 @@
 
 import os, sys
 os.environ["CLANG"] = '1'
-os.environ["GPU"] = '1'
 
 import numpy as np
 import subprocess
 import tensorflow as tf
 import tf2onnx
-from examples.compile_efficientnet import compile_net
 from extra.onnx import get_run_onnx
 from tinygrad.tensor import Tensor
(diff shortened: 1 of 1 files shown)


geohot:
I think we should just delete this example

jaredeh:
I mean I commissioned that for a reason.  Still using it.

Is there a better way to get at that same goal?

geohot:
If it's still useful to you I'll merge it. It can be rewritten with safetensors and export model now I think

PR #2760, opened 2023-12-14
Getitem round3 ....

geohotstan:
by fixing the `TODO: totally unreadable line`
I decided to make more refactors....
- fully use type_dim to do eeevverything, avoid `[<code> if isinstance(i, type) else <code> ... for i in indices]`
- ~~move `2. basic indexing (no copy)` inside an if conditional, so we don't call shrink, flip, etc unless necessary~~
- got rid of crazy 1 liner :D

let's see how many lines I added

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -351,20 +351,20 @@ class Tensor:
     # currently indices_filtered: Tuple[Union[slice, int, Tensor], ...]
     # turn indices in indices_filtered to Tuple[shrink_arg, strides]
     for dim in type_dim[int]:
-      if (i := indices_filtered[dim]) >= self.shape[dim] or i < -self.shape[dim]:
-        raise IndexError(f"index {i} is out of bounds for dimension {dim} with size {self.shape[dim]}")
-      indices_filtered[dim] = ((i,i+1),1) if i >= 0 else ((self.shape[dim]+i, self.shape[dim]+i+1), 1)
+      if (i := indices_filtered[dim]) >= (sh := self.shape[dim]) or i < -sh:
+        raise IndexError(f"index {i} is out of bounds for dimension {dim} with size {sh}")
+      indices_filtered[dim] = ((i, i+1), 1) if i >= 0 else ((sh+i, sh+i+1), 1)
     for dim in type_dim[slice]:
       s, e, st = indices_filtered[dim].indices(self.shape[dim])
       indices_filtered[dim] = ((0, 0) if (st > 0 and e < s) or (st <= 0 and e > s) else (s, e) if st > 0 else (e+1, s+1), st)
-    for dim in type_dim[Tensor]: indices_filtered[dim] = ((0,self.shape[dim]), 1)
+    for dim in type_dim[Tensor]: indices_filtered[dim] = ((0, self.shape[dim]), 1)
 
     new_slice, strides = ((),()) if not indices_filtered else zip(*tuple(i for i in indices_filtered))
     ret = self.shrink(new_slice).flip(axis=[i for i, s in enumerate(strides) if s < 0])
     # add strides by pad -> reshape -> shrink
     if any(abs(s) != 1 for s in strides):
       strides = tuple(abs(s) for s in strides)
-      ret = ret.pad(tuple((0, round_up(sh,s) - sh) for s, sh in zip(strides, ret.shape)))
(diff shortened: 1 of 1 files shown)


geohotstan:
> ### Changes
> ```
> Name                  Lines    Diff    Tokens/Line    Diff
> ------------------  -------  ------  -------------  ------
> tinygrad/tensor.py      646      +4           22.3    -0.1
> 
> 
> total lines changes: +4
> ```

yikes

tinygrad/tensor.py
```diff
@@ -348,34 +346,35 @@ def __getitem__(self, indices: Union[int, slice, Tensor, None, List, Tuple]) ->
...
+        s, e, st = indices_filtered[dim].indices(self.shape[dim])
+        indices_filtered[dim] = (((0, 0) if e < s else (s, e)) if st > 0 else ((0, 0) if e > s else (e+1, s+1)), st)
+      for dim in type_dim[Tensor]: indices_filtered[dim] = ((0,self.shape[dim]), 1)
+
+      new_slice, strides = zip(*tuple(i for i in indices_filtered))
+      ret = self.shrink(new_slice).flip(axis=[i for i, s in enumerate(strides) if s < 0])
+      # add strides by pad -> reshape -> shrink
+      if any(abs(s) != 1 for s in strides):
+        strides = tuple(abs(s) for s in strides)
+        ret = ret.pad(tuple((0, s-(sh % s) if sh % s != 0 else 0) for s, sh in zip(strides, ret.shape)))
```
chenyuxyz, line 367: there's a `round_up` in helper. you can do `round_up(sh, s)-sh` to avoid the if

tinygrad/tensor.py
```diff
@@ -348,34 +346,35 @@ def __getitem__(self, indices: Union[int, slice, Tensor, None, List, Tuple]) ->
...
-    # inject dim=1 for None and collapse dim for int
+    if indexed := type_dim[int] or type_dim[slice]:
+      # turn indices in indices_filtered to Tuple[shrink_arg, strides]
+      for dim in type_dim[int]:
+        if (i := indices_filtered[dim]) >= self.shape[dim] or i < -self.shape[dim]:
+          raise IndexError(f"index {i} is out of bounds for dimension {dim} with size {self.shape[dim]}")
+        indices_filtered[dim] = ((i,i+1),1) if i >= 0 else ((self.shape[dim]+i, self.shape[dim]+i+1), 1)
+      for dim in type_dim[slice]:
+        s, e, st = indices_filtered[dim].indices(self.shape[dim])
+        indices_filtered[dim] = (((0, 0) if e < s else (s, e)) if st > 0 else ((0, 0) if e > s else (e+1, s+1)), st)
```
chenyuxyz, line 359: multiple conditions like this is very hard to read, think about what's common in them and can be factor out

chenyuxyz (commented):
i don't like wrapping the whole block with `indexed` condition. no-op `shrink` and `flip` should be resolved in `shrink` and `flip`. and it's one more place to change if we later want to Variable or whatever to index.

chenyuxyz:
is this ready?

geohotstan:
> is this ready?

yeah now it is. Had to look through it again.

tinygrad/tensor.py
```diff
@@ -348,34 +346,35 @@ def __getitem__(self, indices: Union[int, slice, Tensor, None, List, Tuple]) ->
...
-
+    # currently indices_filtered: Tuple[Union[slice, int, Tensor], ...]
+    # turn indices in indices_filtered to Tuple[shrink_arg, strides]
+    for dim in type_dim[int]:
+      if (i := indices_filtered[dim]) >= (sh := self.shape[dim]) or i < -sh:
+        raise IndexError(f"index {i} is out of bounds for dimension {dim} with size {sh}")
+      indices_filtered[dim] = ((i, i+1), 1) if i >= 0 else ((sh+i, sh+i+1), 1)
+    for dim in type_dim[slice]:
+      s, e, st = indices_filtered[dim].indices(self.shape[dim])
+      indices_filtered[dim] = ((0, 0) if (st > 0 and e < s) or (st <= 0 and e > s) else (s, e) if st > 0 else (e+1, s+1), st)
```
chenyuxyz, line 359: so this is not what i meant, just merging the leaves in the condition trees does not make it easier to read. it's fine for now but something to consider for the future refactor

chenyuxyz:
cool thanks!

PR #2761, opened 2023-12-14
Change `nn.Embedding` from matmul to lookup

mcognetta:
This simplifies `nn.Embedding`, which previously constructed a `(B, S, V)` one-hot matrix and performed a multiply against the embedding table weights. Now, we just look up the corresponding rows in the embedding table for each token.

This results in a slight reduction in lines, as well as a small speedup for large embedding matrices/batch sizes/sequence lengths.

A very simple benchmark is given below (on a NVIDIA GeForce RTX 3070) with some different batch sizes/sequence lengths and the gpt2 embedding configuration:

```
Old Time (64, 32): 4761.07 ms
New Time (64, 32): 2479.79 ms
Old Time (128, 64): 21437.84 ms
New Time (128, 64): 13881.80 ms
Old Time (32, 512): 43910.92 ms
New Time (32, 512): 28393.64 ms
Old Time (JIT) (16, 128): 5334.19 ms
New Time (JIT) (16, 128): 2365.25 ms
```
__________________________________________________

Below is the benchmark code. I am not yet familiar enough with tinygrad to know the standard benchmarking techniques, so advice is appreciated.

```python
from tinygrad import Tensor, nn
from tinygrad.jit import TinyJit
from tinygrad.helpers import Timing

class NewEmbedding:
  def __init__(self, v, e):
    self.weight = Tensor.glorot_uniform(v, e)
  def _copy(self, e):
    self.weight = e.weight[:]
  def __call__(self, i):
    return self.weight[i]

if __name__ == '__main__':
    # taken from gpt2
    V, E = 50257, 768
    TRIALS = 50

    for (B, S) in [(64, 32), (128, 64), (32, 512)]:

        old = nn.Embedding(V, E)
        new = NewEmbedding(V, E)

        with Timing(f"Old Time {B, S}: "):
            for _ in range(TRIALS):
                i = Tensor.randint((B, S), 0, V)
                old(i).realize()

        with Timing(f"New Time {B, S}: "):
            for _ in range(TRIALS):
                i = Tensor.randint((B, S), 0, V)
                new(i).realize()

    B, S = (16, 256)
    @TinyJit
    def new_jit(x):
        return new(x).realize()

    @TinyJit
    def old_jit(x):
        return old(x).realize()

    i = Tensor.randint((B, S), 0, V)
    for _ in range(5):
       new_jit(i)
       old_jit(i)

    with Timing(f"Old Time (JIT) {B, S}: "):
        for _ in range(TRIALS):
            i = Tensor.randint((B, S), 0, V)
            old_jit(i)

    with Timing(f"New Time (JIT) {B, S}: "):
        for _ in range(TRIALS):
            i = Tensor.randint((B, S), 0, V)
            new_jit(i)
```

### DIFF
--- a/tinygrad/nn/__init__.py
+++ b/tinygrad/nn/__init__.py
@@ -123,9 +123,7 @@ class LayerNorm2d(LayerNorm):
 
 class Embedding:
   def __init__(self, vocab_size:int, embed_size:int):
-    self.vocab_size = vocab_size
     self.weight = Tensor.glorot_uniform(vocab_size, embed_size)
 
   def __call__(self, idx:Tensor) -> Tensor:
-    if not hasattr(self, 'vocab_counter'): self.vocab_counter = Tensor.arange(self.vocab_size, requires_grad=False).reshape(1, 1, self.vocab_size)
-    return (self.vocab_counter == idx.unsqueeze(2)).expand(*idx.shape, self.vocab_size) @ self.weight
+    return self.weight[idx]


chenyuxyz:
this one is interesting. __getitem__ also uses arange, so the complexity is the same and it's not doing a lookup.

this adds a lot more kernels (see the CI failure), but it's different and faster compared to master due to (1) we are doing a dtype refactor + __getitem__ refactor now, and the __getitem__ uses dtype int for index buffer, while the old code path uses float to pass index (2) the memory access pattern changed

relevant kernel

master
```
    float val0 = *(data1+(gidx0*1608224)+(lidx2*201028)+ridx0);
    float val1 = *(data1+(gidx0*1608224)+(lidx2*201028)+ridx0+50257);
    float val2 = *(data1+(gidx0*1608224)+(lidx2*201028)+ridx0+100514);
    float val3 = *(data1+(gidx0*1608224)+(lidx2*201028)+ridx0+150771);
```

this pr
```
    int val0 = *(data1+(gidx0*32)+(lidx2*4)+(ridx0*4096));
    int val1 = *(data1+(gidx0*32)+(lidx2*4)+(ridx0*4096)+1);
    int val2 = *(data1+(gidx0*32)+(lidx2*4)+(ridx0*4096)+2);
    int val3 = *(data1+(gidx0*32)+(lidx2*4)+(ridx0*4096)+3);
```

since it adds a few kernels it's not a clear win. we can merge this approach if you can make it with the same number of kernels. or you can solve the O(n) embedding bounty that remove the 50257 from reduce idx

mcognetta:
> since it adds a few kernels it's not a clear win. we can merge this approach if you can make it with the same number of kernels. or you can solve the O(n) embedding bounty that remove the 50257 from reduce idx

By this you mean something like adding a file akin to extra/gemm/cuda_matmul.py but for an embedding table lookup?

chenyuxyz:
no the embedding bounty needs to be merged to core tinygrad and make it the default behavior

chenyuxyz:
closing this as stale. feel free to open a new one with more rigorous perf tests. even though it's not the same as the Embedding bounty, we are interested in perf improvements in general

PR #2766, opened 2023-12-14
test_linearizer cleanup

Qazalin:
cleans up the skips for running linearizer in compiled backends only

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -1,5 +1,4 @@
 # ruff: noqa: E501
-from typing import cast
 import numpy as np
 import unittest, os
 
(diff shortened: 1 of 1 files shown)


test/test_linearizer.py
```diff
@@ -13,10 +14,14 @@
...
 from tinygrad.helpers import dtypes, prod
 
-class TestLinearizer(unittest.TestCase):
+class _BaseTestLinearizer(unittest.TestCase):
+  device: Compiled = cast(Compiled, Device[Device.DEFAULT])
+  @classmethod
+  def setUpClass(cls):
+    if not isinstance(Device[Device.DEFAULT], Compiled): raise unittest.SkipTest("Only Compiled uses linearizer")
+
+class TestLinearizer(_BaseTestLinearizer):
```
chenyuxyz, line 23: why not just decorate the class with `@unittest.skipIf(...)`?

Qazalin: every single one of them? I guess you could yeah but setup() is also pretty easy to just inherit

chenyuxyz: yeah that's what i would do. it's more local so a lot easier to read and move around.

Qazalin: does it really make sense for any linearizer test to run on non-compiled backends? that's why I kinda wanted it to be a blackbox

chenyuxyz: no we skip non compiled backend on linearizer tests. there are other files that runs linearizer tests too. do they need to import and inherit from this file? i don't see a good reason to add this complexity or make it a blackbox

chenyuxyz:
cool thanks!

PR #2769, opened 2023-12-14
arange default dtype to int and zeros/ones default to float

chenyuxyz:


chenyuxyz:
this breaks LLVM and WEBGPU gpt2 - will investigate

PR #2777, opened 2023-12-15
hip cstyle is cuda

nimlgen:
I think we can merge hip and cuda in cstyle, since they are almost the same


geohot:
Do we need all these overloaded operators? code_for_exp takes the dtype now, can the underlying functions be explicitly used there?

chenyuxyz:
all cleaned in #2843

PR #2779, opened 2023-12-15
dtypes: these asserts should pass

geohot:
UOps should have a spec


chenyuxyz:
these asserts are merged in #2800

PR #2780, opened 2023-12-15
add assertion/test for 1d tensor transpose

shawnha:


chenyuxyz:
don't you already get IndexError from a line below? how useful is this?

PR #2781, opened 2023-12-15
adds permute assertion error and subsequent working test to remove TODO on permute test

steveenl:
this pull request modifies the `test_post_permuted_assignment` method in `test_assign.py`

it now includes an assertion to explicitly confirm that the result of adding a permuted tensor `b` to a permuted tensor `a` does not align with the expected result of a straightforward addition without permutation

Tensor `a` is permuted and then added to the unpermuted tensor `b`. This produces the correct result, matching the sum of the transposed `a` and the original `b`.

The first assertion uses `self.assertRaises(AssertionError)` to verify the incorrectness of the result due to the additional permutation. The second assertion checks that the correct operation produces the expected outcome.


geohot:
Why do I want this? Reads like ChatGPT

PR #2784, opened 2023-12-15
More readable to_image_idx

ZenginU:
@chenyuxyz 
Is this more readable than before?

I just want a better refactor on this but maybe my math skills are not enough.


chenyuxyz:
removing no-op `set()` is good. the rest is about the same and not worth merging in my opinion

i don't think it's about math skill, it's not easy to read for me because i have more questions to understand what `one_infer` or `muller` means, and generally not clear what the code is trying to do

ZenginU:
@chenyuxyz 
Hello,
Math goes like this
```
Assume,
idxy = c*k + b
single_valid_node = c

We need to find the k so that we can substitute the c*k properly.
I order to find the k we substitute with zeros to find NumNode.
And put ones to vars in valid and zeros to non includeds to find the coefficients.
After all coefficients are found we can to c*k//c.
muller = k
```
Hope this helps, with this logic I believe this code is more simple?

chenyuxyz:
closing this as it was incorrect and disabled on master

PR #2785, opened 2023-12-16
_broadcasted handles the python number types

chenyuxyz:
interesting cases `f16 + 2 -> f16`. `i8 + 2 -> i8`


chenyuxyz:
disabled one onnx sce test just for CPU. i think it might be a bug in numpy match_type stuff but not entirely sure. this can unblock some other dtype refactor. will revisit the disabled test after we replace the match_type with the new infra

PR #2791, opened 2023-12-16
get reduce dtype acc try 2

Qazalin:
we should figure out why get_lazyop_info is giving the wrong dtype.

blocked by int,float binary op autocast fix that should be done upstream of linearizer


Qazalin:
doing it in #2800 now since all the dtypes fixes need to get done in one place

PR #2795, opened 2023-12-16
fix tensor creation with a list and dtype bfloat16

chenyuxyz:
it went through numpy and numpy does not have bfloat16.

also added broadcasted with a python bool.


chenyuxyz:
oh this is still incorrect because we cannot create a numpy bfloat16 tensor with `LazyBuffer.fromCPU`

PR #2800, opened 2023-12-16
green linearizer asserts for ops

Qazalin:
- asserts from the uops spec #2779
- acc dtype is the reduceop dtype
- bugfixes in tensor functions (tracked in separate PRs)


tinygrad/codegen/linearizer.py
```diff
@@ -64,6 +64,10 @@ def get_reduce_acc(self, op, dtype:DType):
     AndNode: lambda self,ops,ctx:
       functools.reduce(lambda a,b: ctx.uop_alu_idx(a, b, ops, ctx, BinaryOps.MUL, dtype=dtypes.bool), self.nodes[1:], self.nodes[0].render(ops,ctx)) }
 
+  def get_uop_dtype(self, dtype: DType, amt=1):
```
chenyuxyz, line 67: so this function has nothing to do with `uop_dtype`. it's either extracting the base from `ImageDType` or applying `amt` to make a `vec`

something like
```
dtype = buf.dtype.base if isinstance(buf.dtype, ImageDType) else buf.dtype
if amt > 1: dtype = dtype.vec(amt)
```
reads better

Qazalin: we should duplicate the same logic in the localtype def and also maybe in the uop function or somewhere else, that's why I wanna isolate it

chenyuxyz: for cases like this i'd make a function when we are really reusing it, otherwise later we might find it's premature optimization. also it has very little to do with unary op check and can be a separate refactor pr

Qazalin:
we probably cant do these one-at-a-time because one ALU can be the input to another ALU, that's what happened with HALF=1 gpt3, it was taking the sqrt of a binary op's result.

chenyuxyz:
yeah i also had similar issue refactoring other dtypes changes that's deadlocking, it's quite a puzzle to linearize it and merge stuff.

i think binary is the missing piece and if you can figure it out you can make a combined pr

Qazalin:
yeah same conclusion as well, started working on binary

Qazalin:
status:
- 0 linearizer asserts failing!
- test_ops passing in almost all backends, I think [webgpu](https://github.com/tinygrad/tinygrad/actions/runs/7234436147/job/19710809827?pr=2800#step:8:279) is broken because of cstyle hacks
- openpilot works after [this change](https://github.com/tinygrad/tinygrad/pull/2800/commits/1ff232fca87159cc4af530802cc5ec4d428f7ecf) in the promotion logic for images
- i made a change in lazy to make CMPLT always be bool i think that [broke interpreted backends](https://github.com/tinygrad/tinygrad/actions/runs/7234380949/job/19710692553?pr=2800#step:7:441)
- working on the accuracy issues in `test/test_nn.py::TestNN::test_sparse_cat_cross_entropy` [kernel diff](https://gist.github.com/Qazalin/f71826bbc685717aed08ec9fd4a8410c)

tinygrad/mlops.py
```diff
@@ -89,7 +89,8 @@ def backward(self, grad_output:LazyBuffer) -> LazyBuffer:
 
 class Less(Function):
   def forward(self, x:LazyBuffer, y:LazyBuffer) -> LazyBuffer:
-    return x.e(BinaryOps.CMPLT, y)
+    vars_dtype = least_upper_dtype(x.dtype, y.dtype)
+    return x.cast(vars_dtype).e(BinaryOps.CMPLT, y.cast(vars_dtype))
```
chenyuxyz, line 93: i think we convert output to dtypes.bool here, not lazy. you can include this https://github.com/chenyuxyz/tinygrad/pull/55

Qazalin: linearizer always interprets CMPLT as bool that's why we need to cast here https://github.com/tinygrad/tinygrad/pull/2782

Qazalin: but it doesn't logically make any sense for cmplt to be anything but bool? I think we still need
https://github.com/tinygrad/tinygrad/pull/2800/commits/cd2bb4d7520308f7291e064078cbba248805b12f
otherwise the lazyop info would promote the dtypes

chenyuxyz: the solution might be LazyOp stores the output dtype, and not recompute the dtype in get_lazy_op_info

tinygrad/mlops.py
```diff
@@ -148,7 +149,7 @@ def backward(self, grad_output:LazyBuffer) -> Tuple[None, Optional[LazyBuffer],
 class Sum(Function):
   def forward(self, x:LazyBuffer, new_shape:Tuple[int, ...]) -> LazyBuffer:
     self.input_shape = x.shape
-    return x.r(ReduceOps.SUM, new_shape)
+    return x.cast(least_upper_dtype(x.dtype, dtypes.uint8)).r(ReduceOps.SUM, new_shape) # no reduce on bools TODO should be int8? int32 is safer? TODO should this be fixed upstream in sparse_cat_cross_entropy?
```
chenyuxyz, line 152:
```
>>> jnp.array([True, False, False], dtype="int8").sum()
Array(1, dtype=int32)
>>> torch.tensor([True, False, False], dtype=torch.int8).sum().dtype
torch.int64
```
hmm, what's the behavior you want?

Qazalin: if there's a bool sum you probably want an implicit cast to a numeric?
otherwise these things happen https://gist.github.com/Qazalin/f71826bbc685717aed08ec9fd4a8410c

chenyuxyz: i understand we want to upcast for boolean sum, i meant to ask what's your thought on which dtype summation resulted in which dtype, and how did you reach that decision

Qazalin: imo int32 is the best practical choice but theoretically just relying on uint8 will _work_ but it might too small

chenyuxyz:
fyi i did some dtype cleanup in #2817

tinygrad/mlops.py
Qazalin: a part of what I did is this:
https://github.com/tinygrad/tinygrad/pull/2800/files#diff-566a7254fa321896c55ed04f6d3eb9cf38ac84ee952b4e44a107c006cf886b16R239

Qazalin:
now we need to start thinking about ALUs on images @chenyuxyz 

I think we can refactor ops to treat images as floats, there shouldn't be an AST like:
```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ CAST (dtypes.float, False)
  2    ┗━┳ MAX 
  3      ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.imagef((1, 4, 4)), st=ShapeTracker(views=(View(shape=(4,), strides=(4,), offset=0, mask=None, contiguous=False),)))
  4      ┗━━ CONST ConstBuffer(val=0.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4,), strides=(0,), offset=0, mask=None, contiguous=False),)))
  ```
  
everything at the ops layer should be float just like how the kernel will end up:
```
__kernel void E_4n3(__global float* data0, read_only image2d_t data1) {
const sampler_t smp = CLK_NORMALIZED_COORDS_FALSE | CLK_ADDRESS_CLAMP | CLK_FILTER_NEAREST;
  int gidx0 = get_group_id(0); /* 4 */
  float4 val0 = read_imagef(data1, smp, (int2)(gidx0,0));
  float alu0 = max((val0).x,0.0f);
  data0[gidx0] = (float)(alu0);
}
```

**Comment by Qazalin** [comment]
webgou last red standing - there's no way we can go around the hack in casting CMPLT because this def `@group(0) @binding(0) var<storage,read_write> data0: array<bool>;` is invalid, even if your kernel returns a CMPLT's results you can't output them as bools.

**Comment by Qazalin** [comment]
webgou last red standing - there's no way we can go around the hack in casting CMPLT because this `@group(0) @binding(0) var<storage,read_write> data0: array<bool>;` is invalid, even if your kernel returns a CMPLT's results you can't output them as bools.

**Comment by Qazalin** [comment]
also, we need to update the worlds dataset, it has a bunch of wrong asts https://github.com/tinygrad/tinygrad/actions/runs/7252239549/job/19756283084?pr=2800

**Comment by chenyuxyz** [comment]
you can filter ast like https://github.com/tinygrad/tinygrad/blob/0723f26c80b3474dcb288c78df82dd586518a740/extra/optimization/helpers.py#L31

as long as something still runs it's fine. i will regenerate this after this pr is in

**Comment by Qazalin** [comment]
status
- webgpu is the only backend failing [because of the CMPLT hacks](https://github.com/tinygrad/tinygrad/pull/2800#issuecomment-1861281900) - I don't yet know how to solve that one e2e
- Image dtype ALUs get interpreted as float on the linearizer layer but stay as image in ops

File: test/external/external_test_onnx_backend.py [hunk 4]
```diff
@@ -107,6 +107,9 @@ def supports_device(cls, device: str) -> bool:
 backend_test.exclude('test_range_int32_type_negative_delta_expanded_cpu')
 
 # unsupported (strange) ops
+# TODO implement a Binarizer op? https://github.com/onnx/onnx/blob/67c456ba4747412afb44158a1a889c0fc3349641/onnx/reference/ops/aionnxml/op_binarizer.py#L13
+backend_test.exclude('test_ai_onnx_ml_binarizer_*')
+backend_test.exclude('test_bitwise_*')
```
chenyuxyz, line 112: dupped

tinygrad/codegen/linearizer.py
```diff
@@ -19,6 +19,11 @@ class UOps(Enum):
   LOAD = auto(); STORE = auto(); CONST = auto(); BARRIER = auto(); PHI = auto() # noqa: E702
   ALU = auto(); WMMA = auto(); CAST = auto(); GEP = auto() # noqa: E702
 
+# TODO temp util
+def dt(dtype: DType, amt=1):
+  if isinstance(dtype, ImageDType): dtype = dtype.base
```
chenyuxyz, line 24: is this what `dtype.scalar()` tries to unify?

tinygrad/codegen/linearizer.py
```diff
@@ -50,9 +55,11 @@ def const(self, b:Union[int,float], dtype=dtypes.int32, insert_before=None) -> U
 
   def cast(self, val: UOp, dtype) -> UOp: return self.uop(UOps.CAST, dtype, (val,)) if val.dtype != dtype else val
 
-  def get_reduce_acc(self, op, dtype:DType):
-    if op == ReduceOps.SUM: return 0.0 if dtypes.is_float(dtype) else 0
-    elif op == ReduceOps.MAX: return -math.inf if dtypes.is_float(dtype) else -2**31 if dtypes.is_int(dtype) else False
+  def get_reduce_acc(self):
+    assert self.reduceop
```
chenyuxyz, line 59: why this assert?

tinygrad/codegen/linearizer.py
```diff
@@ -513,28 +523,35 @@ def uop(self, uop:UOps, dtype:Optional[DType]=None, vin:Tuple[UOp, ...]=tuple(),
     if cachable: self.saved_exprs[key] = ret
     return ret
 
+  def parse_cast(self, x, u):
+    dtype,bitcast = x.arg
+    if not isinstance(dtype, ImageDType): return self.uop(UOps.CAST, dtype, (u,), (dtype,bitcast))
+    return u if u.dtype == dtype.base else self.uop(UOps.CAST, dtype.base, (u,), (dtype.base,False)) # we don't bitcast on images
```
chenyuxyz, line 529: are we bitcasting on image now?

chenyuxyz (commented):
commented, i also pushed this to the benchmark branch and it had some failures

tinygrad/codegen/linearizer.py
Qazalin: mypy, reduceop is nullable
i could also cast tho

tinygrad/codegen/linearizer.py
Qazalin: no the second arg is False

tinygrad/codegen/linearizer.py
Qazalin: i dont think it's related to that

imagedtype being the same as float should only be a linearizer thing

scalar is a generic vectorized<->normal dtypes bridge

Qazalin:
thanks for enabling the benchmarks I'll look into those

tinygrad/codegen/linearizer.py
chenyuxyz: if we are not bitcasting now, can it be `self.uop(UOps.CAST, dtype, (u,), (dtype,bitcast))` and merge with the line above?

tinygrad/codegen/linearizer.py
chenyuxyz: i meant write this line as `dtype = dtype.scalar()`, instead of `if isinstance...`

tinygrad/codegen/linearizer.py
chenyuxyz: you can still pass the self.reduceop here tho and it should be clear to mypy that self.reduceop is not None is that case? otherwise why was this not an issue before?

tinygrad/lazy.py
```diff
@@ -203,8 +203,8 @@ def loadop(op, shape:Tuple[sint,...], dtype:DType, device:str, arg=None, src:Opt
 
   # create a constant with the shape and dtype of self
   def const(self, val:Union[float, int]) -> LazyBuffer:
-    # NOTE: dtypes.from_np(self.dtype.np) to deal with image types
-    return LazyBuffer.loadop(LoadOps.CONST, tuple(), dtypes.from_np(self.dtype.np), self.device, arg=val).reshape((1,)*len(self.shape)).expand(self.shape)
+    # NOTE: we force the image dtype to be a float32
```
chenyuxyz, line 206: should it be float16 for imageh?

Qazalin: no because we never use imageh, see cstyle

tinygrad/codegen/linearizer.py
Qazalin: yeah that's becuase then the arg was untyped

passing self.reduceop is repetitive when we know there's only one reduceop in a kernel

chenyuxyz: yeah that's fine for this pr but these stateful helper functions need to be clean up. can you add an assertion message `reduceop cannot be None` and check for `is None`?

chenyuxyz:
can you make the treating image as float part a separate pr?

Qazalin:
llama having issues due to acc dtype changes, the acc has gone from float to int which means something in the reduceop info was wrong.
I'll figure that out and then cleanup a diff for a second round of reviews
https://gist.github.com/Qazalin/7f3171f249b427ec3d75ff971262210d#file-llama-diff-L273

Qazalin:
- i think tensor was a better layer to put the webgpu workaround, realize might work too but I think it's an overkill, the only op that could return a bool is cmplt
- the AMD benchmark will pass with 6.0, I confirmed on a docker container
- working on reproducing CUDA locally and fixing it

tinygrad/tensor.py
```diff
@@ -821,8 +821,13 @@ def __itruediv__(self, x) -> Tensor: return self.assign(self.div(x))
   def __imatmul__(self, x) -> Tensor: return self.assign(self.matmul(x))
   def __ixor__(self, x) -> Tensor: return self.assign(self.xor(x))
 
-  def __lt__(self, x) -> Tensor: return mlops.Less.apply(*self._broadcasted(x, False))
-  def __gt__(self, x) -> Tensor: return mlops.Less.apply(*self._broadcasted(x, True))
+  # in webgpu bool cannot be used as a storage buffer type
+  def __lt__(self, x) -> Tensor:
+    out = mlops.Less.apply(*self._broadcasted(x, False))
+    return out.float() if self.device == "WEBGPU" else out
```
chenyuxyz, line 827: can do this cast in mlops itself and no need to copy lines

test/external/external_test_onnx_backend.py
```diff
@@ -107,6 +107,8 @@ def supports_device(cls, device: str) -> bool:
 backend_test.exclude('test_range_int32_type_negative_delta_expanded_cpu')
 
 # unsupported (strange) ops
+# TODO implement a Binarizer op? https://github.com/onnx/onnx/blob/67c456ba4747412afb44158a1a889c0fc3349641/onnx/reference/ops/aionnxml/op_binarizer.py#L13
+backend_test.exclude('test_ai_onnx_ml_binarizer_*')
```
chenyuxyz, line 111: was this one passing before or is it a new test in different onnx version?

Qazalin: the problem isn't accuracy it's dtypes

tinygrad/tensor.py
Qazalin: i think tinygrad should behave like torch
```
a = torch.tensor([1,2,3,4])
b = torch.tensor([1,2,3,4])
out = a < b
print(out.dtype) # bool
```

test/external/external_test_onnx_backend.py
chenyuxyz: can it be fixed here? https://github.com/tinygrad/tinygrad/blob/ca59054463b7d7567cf28d5ee81a008ed2ff8bab/extra/onnx_ops.py#L125

tinygrad/tensor.py
Qazalin: i think tinygrad should behave like torch

chenyuxyz: i agree `__lt__` should return dtypes.bool (same for `__le__`, `__eq__` etc), but the "WEBGPU" type conversion can happen in one place

tinygrad/codegen/linearizer.py
```diff
@@ -479,15 +480,19 @@ def uop(self, uop:UOps, dtype:Optional[DType]=None, vin:Tuple[UOp, ...]=tuple(),
...
+      elif arg == BinaryOps.CMPLT:
+        assert dtype == dtypes.bool, f"{arg} output dtype mismatch {vin[0].dtype=} != {dtypes.bool}"
+        assert vin[0].dtype == vin[1].dtype, f"{arg} dtype mismatch {dtype=} != {vin[0].dtype=} != {vin[1].dtype=}"
+      elif arg in BinaryOps:
+        assert dtype == vin[0].dtype == vin[1].dtype, f"{arg} dtype mismatch {dtype=} != {vin[0].dtype=} != {vin[1].dtype=} {vin[0]} {vin[1]}"
+      elif arg == TernaryOps.WHERE:
+        assert vin[0].dtype == dtypes.bool, f"{arg} selector dtype mismatch {vin[0].dtype=} != {dtypes.bool}"
+        assert dtype == vin[1].dtype == vin[2].dtype, f"{arg} choice dtype mismatch {dtype=} != {vin[1].dtype=} != {vin[2].dtype=}"
+
+    # noqa: E501 if uop == UOps.PHI: assert vin[1].dtype == vin[0].dtype, f"PHI dtype mismatch {dtype=} != {vin[0].dtype=} != {vin[1].dtype=}" TODO this is a wmma bug L317
```
chenyuxyz, line 494: can you elaborate more on this or file an issue? also it should not need `noqa: E501` and comments are not counted in code size

tinygrad/codegen/linearizer.py
```diff
@@ -518,26 +523,28 @@ def uop(self, uop:UOps, dtype:Optional[DType]=None, vin:Tuple[UOp, ...]=tuple(),
...
-    if x.op == UnaryOps.CAST: return [self.uop(UOps.CAST, x.arg[0], (u,), x.arg) if not isinstance(x.arg[0], ImageDType) else u for u in self.ast_parse(cast(LazyOp, x.src[0]), acc, offs, loaded_buffers)]  # noqa: E501
+    if x.op == UnaryOps.CAST: return [self.uop(UOps.CAST, dtype:=self.get_base_dtype(x.arg[0]), (u,), (dtype,x.arg[1])) for u in self.ast_parse(cast(LazyOp, x.src[0]), acc, offs, loaded_buffers)] # noqa: E501
     if x.op in ReduceOps and not do_reduce:
       assert offs is None, "not available if we aren't doing reduce"
       return acc
     # MULACC fusion. TODO: this is copied from Interpreted
-    if x.op == ReduceOps.SUM and x.src[0].__class__ is LazyOp and x.src[0].op == BinaryOps.MUL:
+    # TODO only doing MULACC in floats, ints can mulacc with autocast
+    # TODO should we also check self.reduceop being float? add tests for this
+    if x.op == ReduceOps.SUM and x.src[0].__class__ is LazyOp and x.src[0].op == BinaryOps.MUL and dtypes.is_float(self.get_base_dtype(get_lazyop_info(x).dtype)): # noqa: E501
```
chenyuxyz, line 533: no new `# noqa: E501`. maybe use `isinstance` instead of `__class__ is`

Qazalin: definitely, I'm planning to do an entire diff cleanup last thing once everything works

tinygrad/codegen/linearizer.py
Qazalin: do you think the assert is wrong?
it fails in wmma because of [op3's dtype](https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/linearizer.py#L314C63-L314C63)
there's a chance I'm wrong here and wmma is behaving as expected though

chenyuxyz:
fyi some changes related to this
(1) reverted the acc cleanup pr, no `dtype is ` special behavior
(2) moved the mlops cast to tensor, so we can get backward cast for free
(3) dtype assert check in lazy, and many other related cleanup

i think what's left in this pr is to fix the acc dtype using vin[0], and add a rule to say sum acc needs to be at least int32 or float32 (this should be in tensor.py). and for matmul we cast back to input types

Qazalin:
I think we could do an [override for `dot`](https://github.com/tinygrad/tinygrad/commit/002d76b5565ca9836ec4fe4d2da823ef93ccd442) but we can't really control all ASTs from tensor, eg `(x*w).sum()` is still possible and _will_ be uoped as `MULACC`, that's why I had to use the op dtype in linearizer.

Qazalin:
i think this diff is everything linearizer should do.

changes:
- no use of get_lazyop_info, the buffer dtypes should be enough
- no casts and if dtype checks

we need to [fix these ops on ints](https://github.com/tinygrad/tinygrad/actions/runs/7301235143/job/19897481443?pr=2800).

Qazalin:
#2926 landed almost everything that was here except:
- image cast fallback https://github.com/tinygrad/tinygrad/pull/2800/commits/a0aa1a5927b149498c4ff6a9a817f3ea9b0e4329
- spec asserts

chenyuxyz:
not exactly sure why gpt2 with BEAM failed (re-tried twice). the only thing has a an effect is the cast change? maybe that can be a separate pr with a test for behavior

Qazalin:
the cast only impacts ImageDType though, i think it's from actions @chenyuxyz

chenyuxyz:
i found a bug in PADTO, plan to disable it in #2931 so we can get this and CMPEQ in first

chenyuxyz:
merged, thanks for your help on this one!

PR #2803, opened 2023-12-16
add Tensor.roll

### LINKED ISSUE #2666: Add Tensor.roll
https://pytorch.org/docs/stable/generated/torch.roll.html

kevherro:
This change adds support rolling a tensor along some given dimension(s). It matches the [`torch.roll`](https://pytorch.org/docs/stable/generated/torch.roll.html), using `tinygrad` internals under the hood (i.e., using `Tensor.cat`).

This change also includes support for narrowing a tensor ([`torch.narrow`](https://pytorch.org/docs/stable/generated/torch.narrow.html)).

TODO:
- [x] Support passing a 0-dim `Tensor` as the index of the element to start the narrowed dimension from (`Tensor.narrow`)
- [x] Support passing a `Tuple[int]` as the number of places by which the elements of the tensor are shifted (`Tensor.roll`)
- [x] Support passing a `Tuple[int]` as the axis along which to roll (`Tensor.roll`)

Resolves #2666


Isalia20:
Hi, sorry to jump in but I was looking to contribute to tinygrad, I've never contributed before and saw this thread and the TODOs so I thought I would do the second one and third one. I wrote it but I haven't pushed it yet, wanted to get your approval to jump in to work on this together. If you don't mind it, I'll push the code and you can take a look. I've tested it on tuples of dims, tuples of shifts and it works

kevherro:
> Hi, sorry to jump in but I was looking to contribute to tinygrad, I've never contributed before and saw this thread and the TODOs so I thought I would do the second one and third one. I wrote it but I haven't pushed it yet, wanted to get your approval to jump in to work on this together. If you don't mind it, I'll push the code and you can take a look. I've tested it on tuples of dims, tuples of shifts and it works

Hey, thanks for offering to help out! I'm wrapping up the remaining TODOs at the moment. I'm happy to work with you on another issue, though. Let me know if you have one in mind!

Isalia20:
Thanks, I'll take a look and maybe take something which seems easy for starters :)

geohot:
Many lines, I'd say that's too many tests also (they aren't free to run).

kevherro:
@geohot I managed to trim some lines, but it's still >5000 and I'm not quite sure how to reduce further without it becoming too code golf-y. Recommendations are appreciated!

chenyuxyz:
i think this can be simpler, narrow is a wrapper of shrink, and roll is offset then repeat 2 then slice

kevherro:
> roll is offset then repeat 2 then slice

@chenyuxyz can you expand on this a bit? My interpretation is essentially:

- For each dim in `dims`, `cat` the tensor with itself along that dim. This effectively doubles the tensor along the specified dim, creating a larger tensor where the original data is repeated twice. This sets up for...
- Slicing the `cat`-enated tensor such that we obtain a section where the start and end are offset by the desired `shifts` value

In any case, this probably speeds up `roll` with the added benefit of avoiding `narrow` but comes at the cost of temporarily doubling the tensor's size along the rolled dim which could have memory implications for very large tensors.

Is my understanding of this correct?

chenyuxyz:
does it double the memory? have you measured it?

chenyuxyz:
closing as stale

PR #2816, opened 2023-12-17
Delete useless code path

tomtom-95:
there is no compiler_explorer.py file to run


chenyuxyz:
please don't spam

tomtom-95:
it's not spam, I have created a new pull request because I did not see that the other was behind the master

tomtom-95:
Ok just realized George was referring to the actual file (sorry I did not get it because the previous pull request had the other problem of not being in sync with master and I thought he was referring to that). Don't you think it's worth a comment with the link as it was before for the compiler_explorer.py file?

PR #2818, opened 2023-12-17
Fix argmax and argmin dtypes

gml16:
Argmax and argmin returned tensors of type `float32` instead of `int64`. Numpy returns `int64`. This also allows to directly use argmin/max to index another Tensor.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -519,11 +519,11 @@ class Tensor:
   def argmax(self, axis=None, keepdim=False):
     if axis is None:
       idx = (self == self.max(axis)) * Tensor.arange(prod(self.shape)-1,-1,-1, requires_grad=False, device=self.device).reshape(self.shape)
-      return prod(self.shape) - idx.max() - 1
+      return (prod(self.shape) - idx.max() - 1).cast(dtypes.int64)
     axis = axis + len(self.shape) if axis < 0 else axis
-    m = self == self.max(axis=axis, keepdim=True)
+    m = (self == self.max(axis=axis, keepdim=True))
     idx = m * Tensor.arange(self.shape[axis]-1,-1,-1, requires_grad=False, device=self.device).reshape(self.shape[axis], *[1]*(self.ndim-axis-1))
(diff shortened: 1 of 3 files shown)


tinygrad/tensor.py
```diff
@@ -519,11 +519,11 @@ def log_softmax(self, axis=-1):
   def argmax(self, axis=None, keepdim=False):
     if axis is None:
       idx = (self == self.max(axis)) * Tensor.arange(prod(self.shape)-1,-1,-1, requires_grad=False, device=self.device).reshape(self.shape)
-      return prod(self.shape) - idx.max() - 1
```
chenyuxyz, line 522: this is not the right fix, it's not `int` here due to implicit conversion in the ops, and casting at the end just hide the issue. it should automatically fixed with #2800

gml16: I see, thanks for providing this context. However, I am not sure if #2800 will solve this. Multiplying 2 tensors above (line 521) will cast to `least_upper_dtype`. So if we take argmax of a Tensor with dtype `float32`, I don't see how this will result in a dtype of `int64`. 

Maybe I misunderstand #2800 though. If it does solve this issue, we can just keep the tests that I added in this PR and remove the explicit cast.

chenyuxyz: the intended behavior is that `(self == self.max(axis))` would have `dtypes.bool`, `Tensor.arange` would have `dtypes.int32`, so multiplication output would be int32 (tinygrad's int default)

gml16: Okay thank you, in that case is it okay that numpy's and tinygrad's dtypes differ? (`int64` vs `int32` respectively)

chenyuxyz: jax argmax returns int32. int32 works better in GPU and accelerators. if we really want to use int64 then we will change the default type in arange, and that would propagate naturally. we don't need to manual cast here

chenyuxyz:
thanks for your suggestion, i added the tests in #2933. also i ended up need to cast argmax/argmax manually but for a different reason that we have special dtype handling for webgpu

PR #2778, opened 2023-12-15
HIP ALU ops overload cleanup try 2

Qazalin:
time to get rid of the overloads I originally had to put in here.

PR #2822, opened 2023-12-18
fix compiling bfloat16 in CUDA

### LINKED ISSUE #2812: `RuntimeError: compile failed` with cuda "__bf16" is undefined (cuda 12.2, ubuntu 22.04)
Hi,

I ran `python3 examples/coder.py` and got a compilation error with cuda. Below are my steps with all the details. I would be glad if you can point out where the issue came from. New to the repo and cannot make sense the error message. 

Steps to reproduce on ubuntu 22.04 machine with cuda 12.2

### 2. Install dependencies
```$ python3 -m pip install -e . && pip3 install sentencepiece```
### 3. run coder.py (and get the error below)
```$ python3 examples/coder.py

create model: 101.23 ms
download weights: 25.95 ms
ram used:  0.00 GB, layers.0.attention.wq.weight                      :   0%| |
loaded weights in 20.52 ms, 0.03 GB loaded at 1.64 GB/s
weights -> model: 67.99 ms
  File "/home/baris/development/fun/tinygrad/examples/coder.py", line 37, in <module>
    nn.state.load_state_dict(model, fix_bf16(convert_from_huggingface(part1, model, 32, 8)), strict=False)
  File "/home/baris/development/fun/tinygrad/tinygrad/nn/state.py", line 68, in load_state_dict
    v.assign(state_dict[k].to(v.device)).realize()
  File "/home/baris/development/fun/tinygrad/tinygrad/tensor.py", line 103, in realize
    run_schedule(self.lazydata.schedule())
  File "/home/baris/development/fun/tinygrad/tinygrad/realize.py", line 28, in run_schedule
    prg = lower_schedule_item(si)
  File "/home/baris/development/fun/tinygrad/tinygrad/realize.py", line 19, in lower_schedule_item
    return Device[si.out.device].get_runner(si.ast)
  File "/home/baris/development/fun/tinygrad/tinygrad/device.py", line 314, in get_runner
    def get_runner(self, ast:LazyOp) -> CompiledASTRunner: return self.to_program(self.get_linearizer(ast))
  File "/home/baris/development/fun/tinygrad/tinygrad/device.py", line 286, in to_program
    return CompiledASTRunner(k.ast, k.name, src, k.global_size, k.local_size, runtime_args).build(self.compiler, self.runtime)
  File "/home/baris/development/fun/tinygrad/tinygrad/device.py", line 254, in build
    self.lib = compiler.__wrapped__(self.prg) if getenv("DISABLE_COMPILER_CACHE") else compiler(self.prg)
  File "/home/baris/development/fun/tinygrad/tinygrad/helpers.py", line 261, in wrapper
    return diskcache_put(table, key, func(*args, **kwargs))
  File "/home/baris/development/fun/tinygrad/tinygrad/runtime/ops_cuda.py", line 23, in compile_cuda
    def compile_cuda(prg) -> bytes: return compile_cuda_style(prg, [f'--gpu-architecture={CUDADevice.default_arch_name}', "-I/usr/local/cuda/include", "-I/usr/include"], cuda.nvrtcProgram, cuda.nvrtcCreateProgram, cuda.nvrtcCompileProgram, cuda.nvrtcGetPTX, cuda.nvrtcGetPTXSize, cuda.nvrtcGetProgramLog, cuda.nvrtcGetProgramLogSize, check)  # noqa: E501
  File "/home/baris/development/fun/tinygrad/tinygrad/helpers.py", line 322, in compile_cuda_style
    if status != 0: raise RuntimeError(f"compile failed: {get_bytes(prog, get_log_size, get_log, check).decode()}")
RuntimeError: compile failed: <null>(3): error: identifier "__bf16" is undefined
  extern "C" __global__ void E_512_64_8_16_4(unsigned int* data0, const __bf16* data1) {
                                                                        ^

<null>(8): error: identifier "__bf16" is undefined
    __bf16 val0 = data1[(gidx1*64)+(lidx3*4)+((lidx2%2)*262144)+((((gidx0*4)+(lidx2/2))%64)*4096)+((gidx0/16)*524288)];
    ^

<null>(9): error: identifier "__bf16" is undefined
    __bf16 val1 = data1[(gidx1*64)+(lidx3*4)+((lidx2%2)*262144)+((((gidx0*4)+(lidx2/2))%64)*4096)+((gidx0/16)*524288)+1];
    ^

<null>(10): error: identifier "__bf16" is undefined
    __bf16 val2 = data1[(gidx1*64)+(lidx3*4)+((lidx2%2)*262144)+((((gidx0*4)+(lidx2/2))%64)*4096)+((gidx0/16)*524288)+2];
    ^

<null>(11): error: identifier "__bf16" is undefined
    __bf16 val3 = data1[(gidx1*64)+(lidx3*4)+((lidx2%2)*262144)+((((gidx0*4)+(lidx2/2))%64)*4096)+((gidx0/16)*524288)+3];
    ^

5 errors detected in the compilation of "<null>".
```

### Info about my system:
```
$ nvidia-smi

Sun Dec 17 05:31:07 2023
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.129.03             Driver Version: 535.129.03   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA GeForce RTX 4090        Off | 00000000:01:00.0  On |                  Off |
|  0%   39C    P8              23W / 450W |    325MiB / 24564MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+

+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|    0   N/A  N/A      1161      G   /usr/lib/xorg/Xorg                          118MiB |
|    0   N/A  N/A      1504      G   /usr/bin/gnome-shell                         78MiB |
|    0   N/A  N/A      4575      G   ...irefox/2987/usr/lib/firefox/firefox      109MiB |
+---------------------------------------------------------------------------------------+
```

```
$ uname -a

Linux uname 6.2.0-39-generic #40~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 16 10:53:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

g1y5x3:
Fix #2812

The [cuda](https://docs.nvidia.com/cuda/cuda-math-api/struct____nv__bfloat16.html#struct____nv__bfloat16) variable type is defined as `__nv_bfloat16`. Also looked up for HIP and I think it is `__hip_bfloat16` but can't find any reference on their documentation, only a discussion in https://github.com/ROCm/ROCm/issues/2534

Screenshot from 2023-12-17 21-24-42

~Not sure if this is the best way to fix it. Probably better to define those type name in the `CStyleLanguage` and leaving them as `None` in the helpers??~
~New solution with `render_dtype`~
Adding a render function that's rarely used other than one particular dtype is not a great idea either.


tinygrad/helpers.py
```diff
@@ -173,7 +173,7 @@ def fields() -> Dict[str, DType]: return DTYPES_DICT
 
   # NOTE: bfloat16 isn't supported in numpy
   # it has higher priority than float16, so least_upper_dtype(dtypes.int64, dtypes.uint64) = dtypes.float16
-  bfloat16: Final[DType] = DType(10, 2, "__bf16", None)
+  bfloat16: Final[DType] = DType(10, 2, "__nv_bfloat16", None) if os.getenv("CUDA") else DType(10, 2, "__bf16", None)
```
Qazalin, line 176: wanna use `type_map` for CUDALanguage here?

tinygrad/renderer/cstyle.py
```diff
@@ -256,6 +256,7 @@ class CUDALanguage(CStyleLanguage):
                  UnaryOps.EXP2: lambda x,dtype: f"exp2({x})" if dtype != dtypes.half else f"hexp2({x})"}
   half_prekernel = """
     #include <cuda_fp16.h>
+    #include <cuda_bf16.h>
```
Qazalin, line 259: we need a `dtypes_prekernel`

g1y5x3: to dynamically select the header files? I thought additional include would be more simple

tinygrad/helpers.py
g1y5x3: I think the right way is probably to define them in `CUDALanguage`, 'METALLanguage`, and `HIPLanguage` etc. And here we just pass like `None` so when render it uses those definitions instead.

geohot: Yea you absolutely can't use the env var like this. What if there's two devices? What if CUDA=0?

g1y5x3:
first attempt with `type_map` wasn't successful, since the rendering is happening in `UOps.LOAD` and `UOps.DEFINE_GLOBAL` so it's not just overriding some of the render functions. since `DType is `NamedTuple` which is immutable, so couldn't just override the `dtype.name` either

g1y5x3:
Added `render_dtype`, for the scope of the fix. Need to read a bit more carefully to see whether it should replace all the `dtype.name` and some of the `self.type_map.get(var_dtype)`, might be able to save a few lines there.

tinygrad/renderer/cstyle.py
```diff
@@ -80,15 +81,18 @@ def render_if(self, cond: str):
...
     buftypes = [(name,f"{'read_only' if i > 0 else 'write_only'} image2d_t" if dtype.name.startswith('image') else
-                ("const " if i > 0 else "")+self.buffer_prefix+dtype.name+"*"+self.buffer_suffix if isinstance(dtype, PtrDType) else
+                ("const " if i > 0 else "")+self.buffer_prefix+self.render_dtype(dtype)+"*"+self.buffer_suffix if isinstance(dtype, PtrDType) else
                 self.arg_int_prefix if dtype == dtypes.int else None) for i,(name,dtype) in enumerate(bufs)]
     prg = ''.join([f"{self.kernel_prefix}void {f'__launch_bounds__ ({prod(local_size)}, 1) ' if self.launch_bounds else ''}{function_name}(",] +
     [', '.join([f'{t} {name}' for name,t in buftypes] + self.extra_args)] +
     [") {\n" + tmp] + ['\n'.join(kernel), "\n}"])
-    if self.half_prekernel and any(dtype == dtypes.float16 for _,dtype in bufs): prg = ''.join([f"{self.half_prekernel}", "\n", prg])
+    if self.half_prekernel and any(dtype in [dtypes.float16, dtypes.bfloat16] for _,dtype in bufs): prg = ''.join([f"{self.half_prekernel}", "\n", prg])  # noqa: E501
```
Qazalin, line 95: fyi:
>  No new code can be merged with > 150 char lines.
https://github.com/tinygrad/tinygrad/pull/2778#issuecomment-1862111047

g1y5x3: Thx. Also been looking at the overridden `render_cast` to see if it's possible to save some lines there.

tinygrad/renderer/cstyle.py
```diff
@@ -80,15 +81,17 @@ def render_if(self, cond: str):
...
     buftypes = [(name,f"{'read_only' if i > 0 else 'write_only'} image2d_t" if dtype.name.startswith('image') else
-                ("const " if i > 0 else "")+self.buffer_prefix+dtype.name+"*"+self.buffer_suffix if isinstance(dtype, PtrDType) else
+                ("const " if i > 0 else "")+self.buffer_prefix+self.render_dtype(dtype)+"*"+self.buffer_suffix if isinstance(dtype, PtrDType) else
                 self.arg_int_prefix if dtype == dtypes.int else None) for i,(name,dtype) in enumerate(bufs)]
-    prg = ''.join([f"{self.kernel_prefix}void {f'__launch_bounds__ ({prod(local_size)}, 1) ' if self.launch_bounds else ''}{function_name}(",] +
-    [', '.join([f'{t} {name}' for name,t in buftypes] + self.extra_args)] +
-    [") {\n" + tmp] + ['\n'.join(kernel), "\n}"])
-    if self.half_prekernel and any(dtype == dtypes.float16 for _,dtype in bufs): prg = ''.join([f"{self.half_prekernel}", "\n", prg])
+    prg = ''.join((self.kernel_prefix, "void ", f"__launch_bounds__ ({prod(local_size)}, 1) " if self.launch_bounds else "",function_name, "(",
+                  ", ".join([f'{t} {name}' for name, t in buftypes] + self.extra_args), ") {\n", tmp, "\n".join(kernel), "\n}"))
```
g1y5x3, line 93: most of the lists created before were a bit unnecessary

g1y5x3:
Still don't like the current solution. `render_dtype` is pretty much a dead function other than bfloat16. Ideally we can dynamtically change `dtypes.name` based on the `Device.DEFAULT` that's used but it would cause a circular import.

g1y5x3:
fixed by 3036

PR #2830, opened 2023-12-18
Add backward for Less MLOps

gml16:
Using `Less` (`<`) in the computation graph during backprop would previously result in the following error: `E   RuntimeError: backward not implemented for <class 'tinygrad.mlops.Less'>`.

For example, this would happen if you implemented your own custom loss that used `<`.

I added `None` gradients to `Less`'s backward pass. Maybe this should better be fixed by modifying the base class `Function` and always returning `None` grads if not overriden.

I added the corresponding test to `test_tensor.py`, happy to move them elsewhere if there is a better spot.


gml16:
Well, this broke most tests. Should I file an issue instead with this test?

PR #2834, opened 2023-12-19
Add Cpp example

IldarKashaev:
yo-yo :v: , @geohot 

Added examples for running tiygrand models (only on OpenCl for now), planning to add CUDA + CPU if it makes sense.
What do you think about this?


geohot:
So many lines! We already have a compiled efficientnet C program running in CI.

PR #2841, opened 2023-12-19
Removing ~10 lines (No functionality change)

ScottTpirate:
Change the `axis` and `amt` if else statements to more condensed versions. the `amt` keeps part of its if statement. 

condensed  `OptOps.LOCAL` and `OptOps.LASTLOCAL` into one if statement. (only difference between them is the `self.shift_to()` function and `assert not(self.tensor_core), "can't local with tensor cores"` and both differences are accounted for )


geohot:
NO CODE GOLFFF!

PR #2847, opened 2023-12-19
Minor refactoring: Attempt to fix a TODO in shapetracker moving private logic to symbolic.

iglesias:
Hi. I started taking a look at the `shapetracker_math` bounty and gave a shot at this to start off.

### DIFF
--- a/tinygrad/shape/symbolic.py
+++ b/tinygrad/shape/symbolic.py
@@ -34,6 +34,13 @@ class Node:
   def substitute(self, var_vals: Dict[VariableOrNum, Node]) -> Node: raise RuntimeError(self.__class__.__name__)
   def unbind(self) -> Tuple[Node, Optional[int]]: return self.substitute({v: v.unbind()[0] for v in self.vars() if v.val is not None}), None
 
+  # traverse max attribute looking for an int
+  def find_max_int(self) -> int:
+    tryint = self.max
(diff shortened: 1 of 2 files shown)


tinygrad/shape/symbolic.py
```diff
@@ -34,6 +34,13 @@ def iter_idxs(idxs:Tuple[VariableOrNum, ...]) -> Iterator[Tuple[int,...]]:
   def substitute(self, var_vals: Dict[VariableOrNum, Node]) -> Node: raise RuntimeError(self.__class__.__name__)
   def unbind(self) -> Tuple[Node, Optional[int]]: return self.substitute({v: v.unbind()[0] for v in self.vars() if v.val is not None}), None
 
+  # traverse max attribute looking for an int
+  def find_max_int(self) -> int:
+    tryint = self.max
+    while(not isinstance(tryint, int)): tryint = tryint.max
+    assert isinstance(tryint, int), f"tryint must be integer, {tryint=} isn't"
```
iglesias, line 41: Can the assertion actually fail? It'd be stuck in the loop if that condition is not met.

chenyuxyz:
making the questionable logic a function is not refactoring

iglesias:
I think the definition of refactoring is  irrelevant. I could amend and rephrase the commit, if that's what you'd like.

Why is that piece of code iterating until it finds an int; do you have  any test or something reproducible for this case?

PR #2848, opened 2023-12-19
UOps support for ImageDType

Qazalin:
providers a general utility for linearizing ImageDType ASTs


tinygrad/codegen/linearizer.py
```diff
@@ -54,6 +54,9 @@ def get_reduce_acc(self, op, dtype:DType):
     if op == ReduceOps.SUM: return 0.0 if dtypes.is_float(dtype) else 0
     elif op == ReduceOps.MAX: return -math.inf if dtypes.is_float(dtype) else -2**31 if dtypes.is_int(dtype) else False
 
+  # NOTE: once images are loaded, we uop them as their base float
+  def get_var_dtype(self, dt:DType): return dt if not isinstance(dt, ImageDType) else dt.base
```
chenyuxyz, line 58: i would change this to `def get_base_dtype(self, dt:DType): return dt.base if isinstance(dt, ImageDType) else dt`. do you see why i find this more readable?

PR #2849, opened 2023-12-19
Remove deepwalk's nested func

alexandermorgan:
Instead of creating a second function (_deepwalk) every time deepwalk is called, deepwalk now handles the combined functionality of deepwalk and _deepwalk. Also specifies the return type. This passes all the tests except for an overzealous linter. 1 line is removed and code is slightly less indented/complex.


chenyuxyz:
no code golf, also this changed the API

PR #2851, opened 2023-12-19
Extract _deepwalk as a private method

alexandermorgan:
This prevents _deepwalk from getting created every time deepwalk is called.


chenyuxyz:
and that pollutes Tensor methods!

PR #2853, opened 2023-12-19
More readable torch_load ext check

SkullMag:
Instead of using magical numbers and strings, check file type with more readable function


chenyuxyz:
looks good! is this tested in CI?

SkullMag:
> looks good! is this tested in CI?

not yet, I'll add tests to this PR

chenyuxyz:
Thanks!

PR #2854, opened 2023-12-19
Use set comprehensions

alexandermorgan:
Improve readability and performance especially in larger use cases of get_pairs in stable_diffusion.py. Also the assert call on line 49 of compile2.py should be faster when input_lb is a set instead of a list.


chenyuxyz:
please stop spamming

PR #2855, opened 2023-12-19
Update README.md

mseele:
correct low level ops numbers (from https://github.com/tinygrad/tinygrad/blob/master/docs/adding_new_accelerators.md)


geohot:
Fixed in master to be ~25

PR #2856, opened 2023-12-19
support for empty tensor indexing

geohotstan:
Had to take a break from bool indexing 😫

-------------
who in the world indexes a Tensor with empty tensors anyways...
or what kind of sick person indexes an empty tensor...
idk if we want to waste lines supporting this but torch supports, so here it is

I changed the broadcasting error stuff from try except block to manual broadcasting.
also updated some error msges

todo (if we want to support)
- tests
- clean up code


tinygrad/tensor.py
```diff
@@ -380,28 +380,42 @@ def __getitem__(self, indices: Union[int, slice, Tensor, None, List, Tuple]) ->
...
-      rest_idx = [i.reshape(*[1]*tdim[0], *[1]*(max_dim - i.ndim), *i.shape, *[1]*(ret.ndim - tdim[0] - n)) for n,i in enumerate(idx[1:], 1)]
-      reshaped_idx = first_idx + rest_idx
-      ret = ret.reshape(*ret.shape[:sum_dim[0]+1], *[1]*max_dim, *ret.shape[sum_dim[0]+1:])
-
-      # iteratively eq -> mul -> sum fancy index
-      try:
+      if empty_idx or 0 in ret.shape:
+        new_shape = list(new_shape) # bro wtf why is mypy being dumb
+        new_shape[tdim[0]:tdim[0]+len(idx)] = final_shape
+        ret = Tensor.empty(new_shape, dtype=ret.dtype)
```
geohotstan, line 407: also I'm wondering if I'm understanding empty tensors correctly
if Tensor is indexed to being empty, it's fine if the returned Tensor is detached from compute graph right (can be copy and no need to require_grad)?

or should I shrink orignal Tensor to a 0 dim so that it's no copy

tinygrad/tensor.py
```diff
@@ -380,28 +380,42 @@ def __getitem__(self, indices: Union[int, slice, Tensor, None, List, Tuple]) ->
...
+        if 0 in t.shape: empty_idx = True
         # TODO uint8 and bool tensor indexing
         if not (dtypes.is_int(t.dtype) or t.dtype == dtypes.bool): raise IndexError("tensors used as indices must be int or bool tensors")
 
+      if empty_idx and 0 in ret.shape: return Tensor.empty(0, dtype=ret.dtype)
+
+      # manual broadcasting
+      final_shape = (shapes := tuple(list(i.shape) for i in idx))[0]
+      for interim_shape in shapes[1:]:
+        if len(final_shape) < len(interim_shape): final_shape = [1]*(len(interim_shape) - len(final_shape)) + final_shape
```
chenyuxyz, line 397: can you refactor this with the logic in `_broadcasted`? the left pad with 1 and max from zip are the same

geohotstan:
apart from basic support for empty tensor indexing
I also snuck in the change from `assertRaisesRegex` to regular `assertRaises` tests for `test_indexing`

chenyuxyz:
this is pretty hard to review, can you split the regex thing into a different pr?

chenyuxyz:
so i don't like this because advance indexing is already quite complicated and this adds a lot more complexity for very little function gain. is this and the boolean index all that's left? even if we want to support empty it might just be a quick check with separate code path

geohotstan:
> is this and the boolean index all that's left?

this, a few functions (shouldn't be too hard) like argsort, unravel_index, and data_ptr, and yeah bool. That's all.
 

> even if we want to support empty it might just be a quick check with separate code path

currently it's checking within advanced, it's not adding any complexity to the original advanced (eq -> mul -> sum) to get ret, but just creating an empty Tensor as ret. (the diff looks a lil crazy cuz of extra tab)
They both use the special permute case.
I do have ideas on rewriting the advanced indexing to be simpler (got these ideas when working on bool), and empty might be better integrated with new advanced... maybe


but yeah idk, I also don't think supporting empty is too important
should I close this for now and work on other ones?
and revisit this at the end when other ones are done?

chenyuxyz:
yeah let's close this one now and we can revisit it later. i am very interested in a simpler advanced indexing, along with boolean index if it's conceptually similar, and if gather and advanced indexing can share some code.

PR #2862, opened 2023-12-20
[Refactor][Kernel] use deepcopy, save LoC, increase readibility

barisozmen:
Removed most lines from the `copy()` function of `Kernel` by moving the logic to Python's deepcopy  

## Test Results
[success] python3 -m pip install -e '.[testing]'
[success] python3 test/test_ops.py
[fail] python3 -m pytest test/

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -114,7 +114,7 @@ class Kernel:
   def copy(self):
     ret = deepcopy(self)
     # uncached since linearize didn't run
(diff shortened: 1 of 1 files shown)


geohot:
Did you benchmark this?

barisozmen:
> Did you benchmark this?

No I haven't. It doesn't pass the [beam search test](https://github.com/tinygrad/tinygrad/actions/runs/7272631597/job/19815095613?pr=2862), so I believe it breaks something there. Trying to understand why.

Is there any code entry point or document for understanding how to do benchmarking?

PR #2865, opened 2023-12-20
Line saving: merge diskcache and dtype aliases

kunwar31:
dtypes
1. dtype aliases change is just code golf, worth it for -5 lines?

diskcache
1. unified _diskcache_exec to do both read/write queries, and it also handles db_connection
2. why have a set of _db_tables if we CREATE TABLE "IF NOT EXISTS"
3. diskcache_get now creates a table if it doesn't exist (and still returns None), is this okay?


tinygrad/helpers.py
```diff
@@ -150,34 +150,26 @@ def from_py(x) -> DType: return dtypes.default_float if isinstance(x, float) els
...
-  int8: Final[DType] = DType(1, 1, "char", np.int8)
-  uint8: Final[DType] = DType(2, 1, "unsigned char", np.uint8)
-  int16: Final[DType] = DType(3, 2, "short", np.int16)
-  uint16: Final[DType] = DType(4, 2, "unsigned short", np.uint16)
-  int32: Final[DType] = DType(5, 4, "int", np.int32)
-  uint32: Final[DType] = DType(6, 4, "unsigned int", np.uint32)
-  int64: Final[DType] = DType(7, 8, "long", np.int64)
-  uint64: Final[DType] = DType(8, 8, "unsigned long", np.uint64)
-  float16: Final[DType] = DType(9, 2, "half", np.float16)
+  char: Final[DType] = (int8 := DType(1, 1, "char", np.int8))
```
geohot, line 153: code golf!!!

PR #2867, opened 2023-12-20
fix .item() for single item

geohotstan:
```
Before
In [2]: Tensor(10).item()
AttributeError: 'NoneType' object has no attribute 'toCPU'
In [3]: (Tensor(10)+1).item()
Out[3]: 11

Now
In [2]: Tensor(10).item()
Out[2]: 10
In [3]: (Tensor(10)+1).item()
Out[3]: 11
```

I'm not sure if this is the right fix though....

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -303,6 +303,12 @@ class TestTinygrad(unittest.TestCase):
     # force device copy - to() is opt'd away - Tensor(dev)/1 is ignored
     np.testing.assert_allclose(ua_arr, (Tensor(ua_arr)/Tensor(1)).numpy())
 
+  def test_item(self):
+    assert Tensor(10).item() == 10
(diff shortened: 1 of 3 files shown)


test/test_tensor.py
```diff
@@ -303,6 +303,12 @@ def test_copy_from_numpy_unaligned(self):
     # force device copy - to() is opt'd away - Tensor(dev)/1 is ignored
     np.testing.assert_allclose(ua_arr, (Tensor(ua_arr)/Tensor(1)).numpy())
 
+  def test_item(self):
+    assert Tensor(10).item() == 10
```
chenyuxyz, line 307: cool! can you test the output dtype too? like bool to bool, int to int and float to float

geohotstan:
The new lazy fixed this 🚀

PR #2870, opened 2023-12-20
merge `render_cast` in cstype

g1y5x3:
sort of related to #2822 
sadly not that much of line winner.
webgpu only renders casts from its `type_map` but if we use a function like `render_dtype` in #2822, then it can be handled separately.


geohot:
This is less readable

PR #2873, opened 2023-12-20
render_locals accepting a dtype

Qazalin:
Linearizer currently [hardcodes DEFINE_LOCAL to be float32](https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/linearizer.py#L197-L203).

This unblocks replacing the hardcoded fp32s with the op dtype in linearizer for dtypes math.


chenyuxyz:
can you add a test for this?

Qazalin:
@chenyuxyz I think this test:
https://github.com/tinygrad/tinygrad/blob/master/test/test_linearizer.py#L394
already covers the functionality. Adding more tests for non-fp32 locals would require the full dtypes math refactor first.

chenyuxyz:
is this ready?

PR #2875, opened 2023-12-20
Support sum of bools

Qazalin:
mirrors torch's `torch.tensor([True, True, False]).sum()` output dtype and results. This unblocks dtypes math.

### DIFF
--- a/tinygrad/mlops.py
+++ b/tinygrad/mlops.py
@@ -150,6 +150,7 @@ class Where(Function):
 class Sum(Function):
   def forward(self, x:LazyBuffer, new_shape:Tuple[int, ...]) -> LazyBuffer:
     self.input_shape = x.shape
+    if x.dtype == dtypes.bool: x = x.cast(dtypes.int64)
     return x.r(ReduceOps.SUM, new_shape)
 
   def backward(self, grad_output:LazyBuffer) -> LazyBuffer:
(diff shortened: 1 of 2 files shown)


tinygrad/mlops.py
```diff
@@ -150,6 +150,7 @@ def backward(self, grad_output:LazyBuffer) -> Tuple[None, Optional[LazyBuffer],
 class Sum(Function):
   def forward(self, x:LazyBuffer, new_shape:Tuple[int, ...]) -> LazyBuffer:
     self.input_shape = x.shape
+    if x.dtype == dtypes.bool: x = x.cast(dtypes.int64)
```
chenyuxyz, line 153: i think we want int32, we also differ from torch about index default (we use int32, jax use int32, torch uses int64)

chenyuxyz: also is this output dtype tested?

geohot: Wait, why do we want a hack like this at all?

geohot:
I don't want to see any "if dtype" anywhere

tinygrad/mlops.py
chenyuxyz: my bad i thought sum of int8 was int8, it can always be `least_upper_dtype(dtype, dtypes.int32)`

PR #2871, opened 2023-12-20
mulacc fusion cleanup

Qazalin:
removes #noqa lines and makes the fusion more readable.

PR #2876, opened 2023-12-20
Revert "mulacc fusion cleanup"

geohot:
Reverts tinygrad/tinygrad#2871


geohot:
I don't want to see any dtype checking on things like this

PR #2877, opened 2023-12-20
Fix reshape merging with masks

corsix:
ad233d557f435 disabled part of `_reshape_mask`, due to the offset computation being wrong. I've removed the offset computation entirely, re-enabled the part, and addressed the offset computation in a different way at the caller of `_reshape_mask`.


geohot:
Great to see you contributing, and a tough problem at that!

geohot:
Cool, tested with `test/external/fuzz_shapetracker_math.py` to 100k

sahamrit:
Nice one. I had found myself struggling handling reshapes with add/ remove ones. This looks neat!

PR #2879, opened 2023-12-20
add Tensor.cross

Isalia20:
TODO:
- [x] Tests


geohot:
What model uses this? I haven't seen it before

Isalia20:
It's not used in vanilla models but if someone is building an object detector model for non-axis aligned boxes or wants to detect polygons, they'll need this. For example extending FCOS model for detecting 8 points, 4 (x,y) corners instead of only predicting left top and right bottom corners. In FCOS model these distances will be used to filter the targets on feature pyramid network

tinygrad/tensor.py
```diff
@@ -472,7 +472,19 @@ def transpose(self, ax1=1, ax2=0) -> Tensor:
     order[ax1], order[ax2] = order[ax2], order[ax1]
     return self.permute(order)
   def flatten(self, start_dim=0): return self.reshape(shape=self.shape[:start_dim] + (-1,))
-
+  def cross(self, other, dim=0):
```
BominRahmani, line 475: I think we move this function next to the other Binary ops no? From what I understand cross would operate on two tensors and run elementwise to return one.

Isalia20: I agree, I moved it to binary mlops and reduced line count

geohot:
I think this doesn't belong in tensor.py, but rather the model that uses it. If it's used in a few examples we can move it to tensor.py, but not without a model.

PR #2885, opened 2023-12-21
add lazy asserts

geohot:


chenyuxyz:
the asserts are merged in #2888. i have two follow ups (1) WEBGPU bool and (2) forced casted float32 image in lazy

PR #2887, opened 2023-12-21
move the op casting logic from mlops to tensor try 2

chenyuxyz:
#2883 breaks commavq result in a weird way so i split elementwise from reduceop and will do lazy assert as follow up. also #2800 can rebase on this

### DIFF
--- a/tinygrad/mlops.py
+++ b/tinygrad/mlops.py
@@ -98,30 +98,29 @@ class Xor(Function):
 
 class Add(Function):
   def forward(self, x:LazyBuffer, y:LazyBuffer) -> LazyBuffer:
-    self.x_dtype, self.y_dtype, output_dtype = x.dtype, y.dtype, least_upper_dtype(x.dtype, y.dtype)
-    return x.cast(output_dtype).e(BinaryOps.ADD, y.cast(output_dtype))
+    return x.e(BinaryOps.ADD, y)
 
   def backward(self, grad_output:LazyBuffer) -> Tuple[Optional[LazyBuffer], Optional[LazyBuffer]]:
-    return grad_output.cast(self.x_dtype) if self.needs_input_grad[0] else None, \
-           grad_output.cast(self.y_dtype) if self.needs_input_grad[1] else None
+    return grad_output if self.needs_input_grad[0] else None, \
+           grad_output if self.needs_input_grad[1] else None
 
 class Sub(Function):
   def forward(self, x:LazyBuffer, y:LazyBuffer) -> LazyBuffer:
-    self.x_dtype, self.y_dtype, output_dtype = x.dtype, y.dtype, least_upper_dtype(x.dtype, y.dtype)
-    return x.cast(output_dtype).e(BinaryOps.SUB, y.cast(output_dtype))
+    self.x_dtype, self.y_dtype = x.dtype, y.dtype
+    return x.e(BinaryOps.SUB, y)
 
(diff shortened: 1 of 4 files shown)


tinygrad/mlops.py
```diff
@@ -89,60 +88,58 @@ def backward(self, grad_output:LazyBuffer) -> LazyBuffer:
 
 class Less(Function):
   def forward(self, x:LazyBuffer, y:LazyBuffer) -> LazyBuffer:
-    output_dtype = least_upper_dtype(x.dtype, y.dtype)
-    return x.cast(output_dtype).e(BinaryOps.CMPLT, y.cast(output_dtype))
+    # in webgpu bool cannot be used as a storage buffer type
+    return x.e(BinaryOps.CMPLT, y).cast(dtypes.float if self.device == "WEBGPU" else dtypes.bool)
```
Qazalin, line 92: I don't think CMPLT should cast to bool, it's already always a bool.
Otherwise we end up with ASTs like this:
```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(2, 1, 1, 1, 1, 1), strides=(1, 0, 0, 0, 0, 0), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ ADD 
  2    ┣━┳ WHERE 
  3    ┃ ┣━┳ CAST (dtypes.bool, False)
  4    ┃ ┃ ┗━┳ CAST (dtypes.int, False)
  5    ┃ ┃   ┗━┳ CMPLT 
  6    ┃ ┃     ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(2, 1, 1, 1, 1, 1), strides=(1, 0, 0, 0, 0, 0), offset=0, mask=None, contiguous=True),)))
  7    ┃ ┃     ┗━━ CONST ConstBuffer(val=0.0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(2, 1, 1, 1, 1, 1), strides=(0, 0, 0, 0, 0, 0), offset=0, mask=None, contiguous=False),)))
  8    ┃ ┣━━ CONST ConstBuffer(val=2.0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(2, 1, 1, 1, 1, 1), strides=(0, 0, 0, 0, 0, 0), offset=0, mask=None, contiguous=False),)))
  9    ┃ ┗━━ CONST ConstBuffer(val=0.0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(2, 1, 1, 1, 1, 1), strides=(0, 0, 0, 0, 0, 0), offset=0, mask=None, contiguous=False),)))
 10    ┗━━ LOAD MemBuffer(idx=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(2, 1, 1, 1, 1, 1), strides=(1, 0, 0, 0, 0, 0), offset=0, mask=None, contiguous=True),)))
```
```
kernel void E_2(device int* data0, const device int* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 2 */
  int val0 = *(data1+gidx0);
  *(data0+gidx0) = (((bool)((int)((val0<0)))?2:0)+val0);
}
```
do you see the problem?


also, this assert won't pass https://github.com/tinygrad/tinygrad/pull/2779/files#diff-63d57daaca59e5495bbb108da1c13e782f4a408db8503e015e8094ac07c87ea6R475

I'm removing the cast to bool

PR #2888, opened 2023-12-21
assert for elementwise dtypes in lazy

chenyuxyz:
rest of #2885

### DIFF
--- a/tinygrad/lazy.py
+++ b/tinygrad/lazy.py
@@ -59,9 +59,7 @@ class LazyBuffer:
     return create_lazybuffer(device, ShapeTracker.from_shape(shape), dtype, op, arg, (src,) if src is not None else ())
 
   def const(self, val:Union[float, int]) -> LazyBuffer:
-    # NOTE: we force the image dtype const to be a float32
-    const_dtype = self.dtype if not isinstance(self.dtype, ImageDType) else dtypes.float32
-    return LazyBuffer.loadop(LoadOps.CONST, tuple(), const_dtype, self.device, arg=val).reshape((1,)*len(self.shape)).expand(self.shape)
+    return LazyBuffer.loadop(LoadOps.CONST, tuple(), self.dtype, self.device, arg=val).reshape((1,)*len(self.shape)).expand(self.shape)
 
   def contiguous(self):
     if not self.st.contiguous or self.st.size() != self.base.st.size() or self.is_unrealized_const():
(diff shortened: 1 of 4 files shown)


tinygrad/lazy.py
```diff
@@ -96,16 +94,17 @@ def copy_to_device(self, device:str) -> LazyBuffer:
...
+    for s in (self,)+in_srcs:
       if s == s.base and s.base.contiguous_child and (root:=s.base.contiguous_child[0]()) is not None:
-        new_srcs.append(root._view(s.base.contiguous_child[1]))
+        srcs.append(root._view(s.base.contiguous_child[1]))
       else:
-        new_srcs.append(s)
-    srcs = tuple(new_srcs)
-    return create_lazybuffer(self.device, ShapeTracker.from_shape(self.shape), max(x.dtype for x in srcs), op, arg, srcs)
+        srcs.append(s)
+    assert all_same(dts:=[x.dtype.scalar() for x in (srcs if op != TernaryOps.WHERE else srcs[1:])]), f"all dtypes must match {dts} on {op}"
```
chenyuxyz, line 104: changed this to test for scalar due to ImageDType force casted to float32 in https://github.com/tinygrad/tinygrad/blob/41b2a25be6ae9b271cc87bb79f378916e30ed26e/tinygrad/lazy.py#L208. not entire sure why that's the case but that's only case that failed strict type check. given the other cleanups in this pr i will follow up and change this assert back separately

tinygrad/lazy.py
```diff
@@ -96,16 +94,17 @@ def copy_to_device(self, device:str) -> LazyBuffer:
...
+    assert op != TernaryOps.WHERE or self.device != "WEBGPU" or srcs[0].dtype == dtypes.bool, "TernaryOps.WHERE must have the first arg be bool"
+    output_dtype = srcs[-1].dtype if op != BinaryOps.CMPLT else (dtypes.bool if self.device != "WEBGPU" else dtypes.float32)
```
Qazalin, line 106: do you think the webgpu hack is getting scattered in too many places now?
first mlops 
https://github.com/tinygrad/tinygrad/pull/2887/files#diff-2160099cd6a4d64858c555c33546afb784d0dfdc80731362c7170e01ba362a97R92
and now here, it's also in cstyle..

Qazalin: I think we can centralize this - It's a headache to get right. because linearizer also needs to get refactored for the asserts to pass https://github.com/Qazalin/tinygrad/blob/master/tinygrad/codegen/linearizer.py#L64

PR #2889, opened 2023-12-21
[Refactor] Potential speed-up in `get_linearizer_actions()` + 3 LoC saves

barisozmen:
Potential performance improvements in this PR:

1) moved `if` logic inside of for-loop to a `filter`, which makes the code faster. Additionally, it arguably make the code more readable, and saves 1 LoC. It is possible to save one more LoC's, trading off readibility, by moving `is_valid` variable inside of the filter. 

2) Moved zip+for+if logic into a numpy array selection logic. It is much faster, but also have some overhead from list -> np.array conversion. Overall, it is probably faster. Need to do benchmarking

Ask: Please show me how to do performance benchmarking. Couldn't understand it from the code yet.

### DIFF
--- a/tinygrad/features/search.py
+++ b/tinygrad/features/search.py
@@ -47,7 +47,7 @@ def bufs_from_lin(lin:Linearizer) -> List[Buffer]:
 # get dictionary of all possible actions
 def get_linearizer_actions(lin:Linearizer, include_0=True) -> Dict[int, Linearizer]:
   acted_lins = {0:lin} if include_0 else {}
-  is_valid = lambda a: (a.axis is None) or (a.axis<lin.shape_len) or (lin.full_shape[a.axis] == a.amt and Opt(a.op, a.axis, 0) not in actions)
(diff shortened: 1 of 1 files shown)


tinygrad/features/search.py
```diff
@@ -46,18 +47,15 @@ def bufs_from_lin(lin:Linearizer) -> List[Buffer]:
...
     try:
-      lin2.apply_opt(a)
-      up, lcl = 1, 1
-      for s,c in zip(lin2.full_shape, lin2.colors()):
-        if c in {"magenta", "yellow"}: up *= s
-        if c in {"cyan", "green", "white"}: lcl *= s
-      if up > 256 or lcl > 256: continue
-      acted_lins[i+1] = lin2
+      (lin2:=lin.copy()).apply_opt(a)
+      colors, full_shape = np.array(lin2.colors),  np.array(lin2.full_shape)
```
wozeparrot, line 54: we want to get rid of numpy, not use it more!

PR #2894, opened 2023-12-21
Implement wait in WebGPU

wpmed92:
I added support for timestamp QuerySets in `wgpu-py`, and it got released with [0.13.2](https://github.com/pygfx/wgpu-py/blob/main/CHANGELOG.md), so now we can implement `wait` in WebGPU which allows us to beam search it.

Test that failed previously (since `__call__` was returning `None` instead of the time), and passes with the change: `test/test_search.py`

NOTE: the `timestamp-query` feature is not supported on the CI Mac, but has been locally tested on my M1 pro, so I added a check to only `wait` if timestamp query is supported.
Compared to other backends, WebGPU API is pretty verbose regarding measuring the gpu time, thus the +8 lines in core. (down from the initial +12)


wpmed92:
Closing for now. Will reopen once we decide if webgpu stays.

PR #2895, opened 2023-12-21
Move the webgpu CMPLT hack to one place

Qazalin:
The high-level problem is that we can't define globals with boolean buffers in webgpu. These hacks were first put in both lazy and mlops in #2888 and #2887 to cast the CMPLT output to float. This made it really hard to manage all the issues it caused with the UOps spec PR #2800.

these changes:
- ensure CMPLT is always a bool, we don't need the mlops bool cast
- remove the workaround in lazy asserts for webgpu, CMPLT _is_ still a bool in wgsl.


chenyuxyz:
yeah it's fine. this ready to merge? i will do a cleanup here / webgpu when i add CMPEQ

PR #2896, opened 2023-12-21
mulacc cleanup try 2

Qazalin:
I think it's gonna be harder to move mulacc float upcasts to tensor, because even if we do the cast in `matmul`, `(x*w).sum(-1)` will still get fused without that cast.

We don't wanna promote ints to float in neither mul nor a sum either.

blocking dtypes math
https://github.com/tinygrad/tinygrad/actions/runs/7290459675/job/19867304498?pr=2800#step:14:1826


Qazalin:
what we ended up with is changing mulacc's rendering for llvm and gpu #2926

PR #2898, opened 2023-12-21
don't match dtype with first input in where

chenyuxyz:
`Tensor([1, 2, 3]).where(1.2, 2.3)` the first `[1, 2, 3]` can directly cast into bool without casting float (in broadcasted) first


chenyuxyz:
@Qazalin fyi this is the root cause of the double cast you mentioned earlier

PR #2900, opened 2023-12-21
add local ci scripts

kevherro:
This change adds two CI scripts, `lint.sh` and `fuzz.sh`. They contain the lint and fuzz steps from the `Unit Test` workflow, making them reproducible across local environments and GitHub Actions.


geohot:
We have pre-commit hooks for this. Instead, reuse the pre-commit hooks in CI

Issue #1367, opened 2023-07-28
test_dtype fails with a segfault on LLVM on linux


ghost:
$ python3 examples/llama.py --gen 2
using CPU backend
using LLaMA-2-7B model
ram used:  0.00 GB, layers.0.attention.wq.weight                      :   0%|                                                                                                      | 0/292 [00:00<?, ?it/s]Segmentation fault (core dumped)

chenyuxyz:
loading works for me, do you have enough ram?

<img width="1499" alt="image" src="https://github.com/tinygrad/tinygrad/assets/2525478/8bffcfde-831a-40e0-b11e-baeeeafad0f7">

oliverhu:
It works in macOS (Metal & CPU), but ran into the same issue on Ubuntu 22.04 (clang 14).
```
(p3) pi@pig ~/py/tinygrad (master) $ CPU=1 PYTHONFAULTHANDLER=1 python  -O examples/llama.py --prompt "Hello." --count 10 --temperature 0 --timing
using CPU backend
using LLaMA-7B model
ram used:  0.00 GB, layers.0.attention.wq.weight                      :   0%| | 0/292 [00:0Fatal Python error: Segmentation fault

Thread 0x00007f56dc0ef640 (most recent call first):
  File "/usr/lib/python3.10/threading.py", line 324 in wait
  File "/usr/lib/python3.10/threading.py", line 607 in wait
  File "/home/pi/p3/lib/python3.10/site-packages/tqdm/_monitor.py", line 60 in run
  File "/usr/lib/python3.10/threading.py", line 1016 in _bootstrap_inner
  File "/usr/lib/python3.10/threading.py", line 973 in _bootstrap

Current thread 0x00007f5a2a299000 (most recent call first):
  File "/home/pi/py/tinygrad/tinygrad/runtime/ops_llvm.py", line 62 in __call__
  File "/home/pi/py/tinygrad/tinygrad/ops.py", line 142 in __call__
  File "/home/pi/py/tinygrad/tinygrad/ops.py", line 139 in exec
  File "/home/pi/py/tinygrad/tinygrad/ops.py", line 197 in exec_ast
  File "/home/pi/py/tinygrad/tinygrad/lazy.py", line 157 in realize
  File "/home/pi/py/tinygrad/tinygrad/lazy.py", line 352 in _realize_from
  File "/home/pi/py/tinygrad/tinygrad/lazy.py", line 144 in realize
[...]

PR #2915, opened 2023-12-22
Fix LLVM issue with bitcast+fptrunc vectorization and add type load/downconversion test

sobomax:
This patch is expected to fix Issue #1367 as well as add some tests around various types being downconverted after loading from disk. The test brings to light various inconsistencies that conversion might go with different backends and different type permutations, as well as hopefully help us to clean it over time to make it all more consistent.

The goal over time would be to reduce broken to 0.

The original bug I was after causes crash in the llvm-generated code, and it's only happens when the compiler sees multiple fptrunc operations like so:

```
define void @"E_4194304_4"(half* noalias %".1", i32* noalias %".2") "no-nans-fp-math"="true"
{
[...]
  %".17" = bitcast i32 %".7" to float
  %".18" = bitcast i32 %".10" to float
  %".19" = bitcast i32 %".13" to float
  %".20" = bitcast i32 %".16" to float
  %".21" = fptrunc float %".17" to half
  %".22" = fptrunc float %".18" to half
  %".23" = fptrunc float %".19" to half
  %".24" = fptrunc float %".20" to half
```

The code was crashing so dive into gdb revealed this:

```
Dump of assembler code for function E_4194304_4:
   0x00007ffff7bcd000 <+0>:  push   %rbp
   0x00007ffff7bcd001 <+1>:  push   %r15
   0x00007ffff7bcd003 <+3>:  push   %r14
   0x00007ffff7bcd005 <+5>:  push   %r13
   0x00007ffff7bcd007 <+7>:  push   %r12
   0x00007ffff7bcd009 <+9>:  push   %rbx
   0x00007ffff7bcd00a <+10>: sub    $0x18,%rsp
   0x00007ffff7bcd00e <+14>: mov    %rsi,%rbx
   0x00007ffff7bcd011 <+17>: mov    %rdi,%r14
   0x00007ffff7bcd014 <+20>: add    $0xc,%rbx
   0x00007ffff7bcd018 <+24>: xor    %ebp,%ebp
   0x00007ffff7bcd01a <+26>: movabs $0x0,%r15
   0x00007ffff7bcd024 <+36>: cs nopw 0x0(%rax,%rax,1)
   0x00007ffff7bcd02e <+46>: xchg   %ax,%ax
   0x00007ffff7bcd030 <+48>: movss  -0xc(%rbx),%xmm0
   0x00007ffff7bcd035 <+53>: movss  -0x8(%rbx),%xmm1
   0x00007ffff7bcd03a <+58>: movss  %xmm1,0xc(%rsp)
   0x00007ffff7bcd040 <+64>: movss  -0x4(%rbx),%xmm1
   0x00007ffff7bcd045 <+69>: movss  %xmm1,0x10(%rsp)
   0x00007ffff7bcd04b <+75>: movss  (%rbx),%xmm1
   0x00007ffff7bcd04f <+79>: movss  %xmm1,0x14(%rsp)
=> 0x00007ffff7bcd055 <+85>: call   *%r15
   0x00007ffff7bcd058 <+88>: mov    %ax,0xa(%rsp)
   0x00007ffff7bcd05d <+93>: movss  0xc(%rsp),%xmm0
```

%r15 is very obviously 0 at this point, so trying to compile the IR with the lcc gave me this:

```
        movss   -12(%r15), %xmm0                # xmm0 = mem[0],zero,zero,zero
        movss   -8(%r15), %xmm1                 # xmm1 = mem[0],zero,zero,zero
        movss   %xmm1, 32(%rsp)                 # 4-byte Spill
        movss   -4(%r15), %xmm1                 # xmm1 = mem[0],zero,zero,zero
        movss   %xmm1, 16(%rsp)                 # 4-byte Spill
        movss   (%r15), %xmm1                   # xmm1 = mem[0],zero,zero,zero
        movss   %xmm1, 12(%rsp)                 # 4-byte Spill
        callq   __truncsfhf2@PLT
        movaps  %xmm0, 48(%rsp)                 # 16-byte Spill
        movss   32(%rsp), %xmm0                 # 4-byte Reload
                                        # xmm0 = mem[0],zero,zero,zero
        callq   __truncsfhf2@PLT
```

Which was a bit unexpected but gave me a clue. So after trying few things that did not work, the +f16c' seems to do the trick.

None of the existing dtype tests hit the issue since you'd need multiple conversion operations to trigger the vectorization.

I've fixed the llvm so it passes all checks, as a bonus I've fixed coder to check if bfloat is supported and fallback to do conversion via LLVM backend if not. This makes coder working with GPU and TORCH, however for some reason it produces gibberish on torch (uint support is lacking?)

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -1,3 +1,4 @@
+import warnings
 import unittest
 import numpy as np
 import torch
@@ -6,7 +7,7 @@ from tinygrad.helpers import CI, getenv, DEBUG, OSX, temp
 from tinygrad.dtype import DType, DTYPES_DICT, ImageDType, PtrDType, least_upper_float, least_upper_dtype
 from tinygrad import Device
 from tinygrad.tensor import Tensor, dtypes
-from typing import Any, List
+from typing import Any, Set
 from hypothesis import given, settings, strategies as st
 
 core_dtypes = list(DTYPES_DICT.values())
@@ -22,9 +23,9 @@ def is_dtype_supported(dtype: DType, device: str = Device.DEFAULT):
   if dtype == dtypes.float64: return device != "METAL" and not (OSX and device == "GPU")
   return True
 
-def get_available_cast_dtypes(dtype: DType) -> List[DType]:
+def get_available_cast_dtypes(dtype: DType) -> Set[DType]:
   if not is_dtype_supported(dtype): return []
-  return [v for k, v in DTYPES_DICT.items() if v != dtype and is_dtype_supported(v) and not k.startswith("_")] # dont cast internal dtypes
+  return set([v for k, v in DTYPES_DICT.items() if v != dtype and is_dtype_supported(v) and not k.startswith("_")]) # dont cast internal dtypes
 
 def _test_to_np(a:Tensor, np_dtype, target):
   if DEBUG >= 2: print(a)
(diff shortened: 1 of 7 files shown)


sobomax:
Also fixes similar issue doing double->half conversion, that one emits a similar intrinsic:

```
  %".5" = getelementptr inbounds double, double* %".2", i32 %"loop0"
  %".6" = load double, double* %".5"
  %".7" = fptrunc double %".6" to half
  %".8" = getelementptr inbounds half, half* %".1", i32 %"loop0"
```

```
        movq    %rsi, %r14
        movq    %rdi, %r15
        xorl    %ebx, %ebx
        movabsq $__truncdfhf2, %r12
        .p2align        4, 0x90
```
After this change:

```
  %".5" = getelementptr inbounds double, double* %".2", i32 %"loop0"
  %".6" = load double, double* %".5"
  %".7" = fptrunc double %".6" to float
  %".8" = fptrunc float %".7" to half
```
Optimizer properly fuses those in the generated code:
```
        vmovsd  (%rsi,%rax,8), %xmm0            # xmm0 = mem[0],zero
        vcvtsd2ss       %xmm0, %xmm0, %xmm0
        vcvtps2ph       $4, %xmm0, %xmm0
        vpextrw $0, %xmm0, (%rdi,%rax,2)
```

geohot:
There's a whole bunch of different unrelated things in this PR, it's unreviewable. Want to submit small, clear win PRs?

PR #2918, opened 2023-12-22
[RDNA3_EMU][1/n] Add new compile_hip function to generate llvm bitcode + asm

kvkenyon:
For testing purposes we want to create asm to ingest into the rdna3 emulator. By using different hiprtc functions and a new flag (hiprtcGetBitcode, hiptrtcGetBitcodeSize, and -fgpu-rpc) we can generate llvm IR bitcode.

We can then use llc -o <ir> to create proper  AMD assembly.

I've modified the test to use this multi-step hip compilation so we can ingest the asm into the emulator.

Next steps include parsing the assembly and obviously emulating the rdna3 ISA.

### DIFF
--- a/test/test_hip_rdna3.py
+++ b/test/test_hip_rdna3.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
-import unittest
-import operator
+import operator, subprocess, unittest
 from tinygrad import Tensor, Device, dtypes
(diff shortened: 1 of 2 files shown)


geohot:
This has changed to use comgr. Feel free to resubmit if there's something you want from this.

PR #2925, opened 2023-12-23
Refactor/setup requirements sync

obar1:
just an idea feel free to close this if not intereted :) 
I prefer a nice req than keep all the dep in command separated text in setup.py etc
> the idea is simple... so draft pr as to move all the dependencies  in the req files is not trivial for newbie to project like myself ...

- [X] quick test in `codespace` ...
```bash
12  python3 -m pip install -e .
13  python3 -m pip install -e '.[testing]'
14  python3 -m pytest test
```
passed


geohot:
Closing, we don't pin on purpose

PR #2928, opened 2023-12-24
Change enum syntax

hathibelagal-dev:
Instead of using auto() all the time, how about we change to the functional syntax?


geohot:
https://github.com/tinygrad/tinygrad#contributing

"All docs and whitespace changes will be closed unless you are a well-known contributor. "

PR #2929, opened 2023-12-24
Remove np from dtype

yabirgb:
First step of #2649. This commit removes np from dtype and uses ctype instead. Also adjust the logic to keep a translation from ctypes to dtype until numpty is fully removed

Update: Sorry for the failing tests. I'm checking now and will update once they are fixed

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -6,6 +6,7 @@ from tqdm import tqdm
 from typing import Dict, Tuple, Union, List, NamedTuple, Final, ClassVar, Optional, Iterable, Any, TypeVar, TYPE_CHECKING, Callable, Set, Sequence
 if TYPE_CHECKING:  # TODO: remove this and import TypeGuard from typing once minimum python supported version is 3.10
   from typing_extensions import TypeGuard
+  import _ctypes.PyCSimpleType
 
 T = TypeVar("T")
 U = TypeVar("U")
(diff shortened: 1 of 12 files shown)


tinygrad/helpers.py
```diff
@@ -108,13 +109,14 @@ class DType(NamedTuple):
...
   name: str
-  np: Optional[type]  # TODO: someday this will be removed with the "remove numpy" project
+  ctype: Optional['_ctypes.PyCSimpleType']
   sz: int = 1
   def __repr__(self): return f"dtypes.{INVERSE_DTYPES_DICT[self]}" if self.sz == 1 else f"dtypes._{INVERSE_DTYPES_DICT[self.scalar()]}{self.sz}"
   def vec(self, sz:int):
     assert sz > 1 and self.sz == 1, f"can't vectorize {self} with size {sz}"
     return DType(self.priority, self.itemsize*sz, f"{INVERSE_DTYPES_DICT[self]}{str(sz)}", None, sz)
   def scalar(self): return DTYPES_DICT[self.name[:-len(str(self.sz))]] if self.sz > 1 else self
+  def ctype_as_dtype(self): return np.dtype("float16") if self.name == "half" else np.dtype(self.ctype)
```
yabirgb, line 119: had to add this method since the half type is not supported in ctypes. When we start removing numpy from tinygrad this can go away

yabirgb (commented):
~~The failing test passes locally. Is it a CI thing or should I look deeper? Don't want to run the CI again to prevent spam since is my first contribution~~ Just saw in discord that the failing test fails in other PRs too so is not related to the changes here

geohot:
Way too many lines changed, and this still has numpy!

yabirgb:
got it! Will convert to draft and see how to do it better

geohot:
Closing for now

PR #2934, opened 2023-12-25
enable boolean ops tests in test_dtype_alu

Qazalin:
these todos are done with compeq


geohot:
Tests failing

Qazalin:
will reopon once i fix the LLVM bug, it seems related to nan ALUs only

PR #2936, opened 2023-12-26
Einsum ellipsis operator implementation

Isalia20:
This PR implements ellipsis operator in einsum function. While it adds quite a bit of lines, I couldn't find easier way to do this and to make it work at the same time. Suggestions in refactoring/making logic easier are welcome

Example, this works in torch but doesn't work in tinygrad(before PR)
```
from tinygrad import Tensor
import torch
import numpy as np

a = Tensor([1, 2, 3])
b = torch.tensor([1, 2, 3])

torch.einsum("...->", b)
Tensor.einsum("...->", a).numpy()
```

After this PR this will work and many other formulas, see some examples which were added in test_ops.py


Isalia20:
oh well, I've found some shapes for which this doesn't work... Will continue work on this

geohot:
Is this worth / does this need +19 lines?

Isalia20:
I'd say it's not worth 19 lines, with the current changes there are in this PR, there are still some edge cases which are not covered. This means that the line count might grow even more. 

Real question then stands whether this can be squashed in fewer lines, which I currently doubt it can since einsum covers a lot of operations. I'll try to come up with a solution with fewer lines so let's keep this PR open. I'll close it in a ~week if I can't come up with something smaller/more elegant

Issue #2721, opened 2023-12-11
tracking issue for dtype refactor


chenyuxyz:
- [x] type promotion helpers (List[input_type] -> output_type) #2724 
- [x] update and fix test_uops where to have vin[0] as bool #2782
- [x] int / int list should be treated as int32 tensor #2789
- [x] tensor.py handles unary that need to cast to float (exp, sin... etc) #2740 #2887 
- [x] mlops handles least upper type for elementwise ops #2788 
- [x] dtypes.int+3.3 -> float, dtypes.half+3.3 -> half #2785 
- [x] Tensor of list of list of list... of int should be dtypes.int #2793
- [x] Tensor.default_type -> dtypes.default_float #2824
- [x] Tensor.sum returns at least int32 or float32 #2926 
- [x] all Tensor cmp returns in dtypes.bool #2931 
- [x] assert that type is properly handled in uops level, and assert in lazy level #2779 #2782 #2800 #2888
- [x] remove match_types in ops_cpu and ops_torch once all input types are aligned in ops #2817
- [x] remove output_type in ops_cpu and ops_torch once all output types are aligned in ops #2892
- [x] skipped onnx sce test in #2785, reenabled in #2902
- [x] llama is quite slower, fix type for freq and mask #2932
- [x] confirm argmax and argmin returns dtypes.int see #2818 #2933
- [ ] ~~remove the hack in WEBGPU to write bool as f32~~
- [ ] cannot create a bfloat16 tensor. see #2795
- [x] remove the redundant manual cast and manual dtype in tensor.py and onnx #2939

i started to write some of it in (https://github.com/chenyuxyz/tinygrad/pull/28) and realized lack of helper to get the output type is blocking everything, so i will do that next. @geohotstan @Qazalin fyi.

for the type promotion we will follow the jax rules. and we will try to match jax behavior (already found some undesired behavior in torch...)
https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.dtype.html
https://github.com/google/jax/blob/3651d4c4f5d59c570f573824e32c0515a8c883f0/jax/_src/dtypes.py#L556
https://github.com/google/jax/blob/b225c86f10380ee7cb924153b6349ef1359f7a15/tests/dtypes_test.py#L126

Issue #1556, opened 2023-08-16
Float("Nan") == Any number is True


ZenginU:
So with the #1493 commit, now Float("NaN") == Any Number is True

Nan == 6 is True, and so on.

```
x = Tensor([float("inf"), float("-inf"),float("nan"), 5,5,-4])
x == 6 is [0. 0. 1. 0. 0. 0.]

```

Note: Not sure about the title, admins can change it to make it more general

wozeparrot:
Probably related: #862

geohotstan:
```
  def __ge__(self, x) -> Tensor: return (((self>float('-inf')) + (self<float('inf')) + (x>float('-inf')) + (x<float('inf'))) < 3).where(0.0, (1.0-(self<x)))
  def __le__(self, x) -> Tensor: return (((self>float('-inf')) + (self<float('inf')) + (x>float('-inf')) + (x<float('inf'))) < 3).where(0.0, (1.0-(self>x)))
  def __ne__(self, x) -> Tensor: return (((self>float('-inf')) + (self<float('inf')) + (x>float('-inf')) + (x<float('inf'))) < 3).where(1.0, ((self<x) + (self>x)))   # type: ignore
  def __eq__(self, x) -> Tensor: return 1.0-(self != x)       # type: ignore
```
shield your eyes from this cursed concoction of code 🫣
but yeah this hack fixes the float('nan') issue for ge, le, ne, and eq
probably not at all ideal

it is also possible to say screw it and totally curse the tensor.py implementation of ge, le, ne, and eq and fix #1562 while at it, but probably not a good idea.

either way this is one possible fix
gonna try to find more

gabeorosan:
Not sure how important, but that runs into issues if both are inf. What about this?
```
  def __ge__(self, x) -> Tensor: return (self>x) + (self==x)
  def __le__(self, x) -> Tensor: return (self<x) + (self==x)
  def __ne__(self, x) -> Tensor: return 1.0 - (self==x)
  def __eq__(self, x) -> Tensor: return (1/((self*x*float('inf')))<1) - (self<x) - (self>x)
```
This is testing as slightly faster that any other workaround for me. You can also divide by zero instead of multiplying by inf but this seems more intuitive/reliable

imaolo:
I think it is impossible to get the correct behavior with only CMPLT, the only comparison op at the moment.

chenyuxyz:
this is fixed on master with CMPEQ

both
```
[...]

PR #2939, opened 2023-12-27
removed redundant dtype hacks in onnx_ops

geohotstan:
#2721  
remove the redundant manual cast and manual dtype in ~~tensor.py and~~ onnx


chenyuxyz:
cool! this manual cast might need update as well https://github.com/tinygrad/tinygrad/blob/54629b56d2533ff1c3f604e163746573b13acb36/extra/onnx.py#L62

also run test/external/external_model_benchmark.py to check that nothing breaks

geohotstan:
> cool! this manual cast might need update as well
> 
> https://github.com/tinygrad/tinygrad/blob/54629b56d2533ff1c3f604e163746573b13acb36/extra/onnx.py#L62

Wow this is actually the reason the previous commit failed CI LOL
Thanks :D

geohotstan:
Hmmmm, it appears that openpilot's onnx has input with
`inp.data_type` as `half` or `long`

and it appears that taking out astype float32 here makes OPENCL unable to compile... since the dtype gets set to half
https://github.com/tinygrad/tinygrad/blob/54629b56d2533ff1c3f604e163746573b13acb36/extra/onnx.py#L62

pretty sure it's the half that's causing the problem here.
Will investigate more when I wake up.

Qazalin:
the half thing is because in general half is broken in CI for the GPU backend, intel CPUs dont have the fp16 extension

chenyuxyz:
probably fine to explicitly cast only float16 to float32 in CI for now

extra/onnx.py
```diff
@@ -59,7 +59,11 @@ def buffer_parse(inp: TensorProto) -> Tensor:
       elif len(inp.int32_data) > 0:
         ret = Tensor(np.array(inp.int32_data, dtype=np.int32).reshape(inp.dims), requires_grad=False)
       else:
-        ret = Tensor(np.frombuffer(inp.raw_data, dtype=tensor_dtype_to_np_dtype(inp.data_type)).reshape(inp.dims).astype(np.float32).copy(), requires_grad=False)
+        # TODO half broken in CI for GPU backend
+        if (dtype := tensor_dtype_to_np_dtype(inp.data_type)) == np.half and CI:
```
chenyuxyz, line 63: limit this to only GPU backend?

chenyuxyz:
might need to update external_model_benchmark.py too, i see a commented out cast there

extra/onnx_ops.py
```diff
@@ -9,35 +9,35 @@
...
-def Sub(x: Union[Tensor, Any], other: Tensor): return x - other # some test has input as int
-def Div(x: Tensor, other: Tensor): return x / other if x.dtype == dtypes.float or isinstance(x.dtype, ImageDType) else x.div(other).floor()   # TODO: this has dtype issues
-def Pow(x: Tensor, other: Tensor): return x.float() ** other.float()
-def Less(x:Tensor,y:Tensor): return (x<y).cast(dtypes.bool)
-def LessOrEqual(x:Tensor,y:Tensor): return (x<=y).cast(dtypes.bool)
-def Greater(x:Tensor,y:Tensor): return (x>y).cast(dtypes.bool)
-def GreaterOrEqual(x:Tensor,y:Tensor): return (x>=y).cast(dtypes.bool)
-def Equal(x:Tensor,y:Tensor): return (x==y).cast(dtypes.bool)
+def Add(x: Tensor, other: Tensor, broadcast=None): return x + other
+def Sub(x: Union[Tensor, Any], other: Tensor): return x - other
```
chenyuxyz, line 18: i think it's useful to keep the comment `# some test has input as int` because i tried to add Sub to tensor_method once and failed...

extra/onnx_ops.py
```diff
@@ -9,35 +9,35 @@
 
 tensor_methods = {"Neg", "Reciprocal", "Sqrt", "Sign", "Abs", "Exp", "Log", "Mish", "Sin", "Cos", "Tan", "Relu", "Sigmoid", "MatMul",
                   "Floor", "Ceil", "Softplus", "HardSwish", "Where", "Mul", "Sinh", "Cosh", "Tanh", "Softsign", "Asinh", "Acosh", "Atanh",
-                  "Elu", "Celu"}
+                  "Elu", "Celu", "Div", "Pow"}
```
chenyuxyz, line 12: nit: add `Div` after `Mul`

extra/onnx_ops.py
```diff
@@ -564,22 +556,19 @@ def EyeLike(x: Tensor, dtype=None, k=0):
 
 def Upsample(X, scales, mode): return Resize(X=X, scales=scales, mode=mode)
 
-# Needs work
 def IsInf(x: Tensor, detect_negative=1, detect_positive=1):
-  ret = (x == float("inf"))*detect_positive + (x == float("-inf"))*detect_negative + Tensor.zeros(*x.shape)
-  return ret.cast(dtypes.bool)
+  ret = (x == float("inf")) * bool(detect_positive) + (x == float("-inf")) * bool(detect_negative) + Tensor.zeros(*x.shape, dtype=dtypes.bool)
```
chenyuxyz, line 560: why is the last `+ Tensor.zeros` needed?

test/external/external_test_onnx_backend.py
```diff
@@ -194,6 +194,11 @@ def supports_device(cls, device: str) -> bool:
 if Device.DEFAULT == 'METAL':
   backend_test.exclude('test_maxpool_2d_same_upper_cpu')
 
+# TODO when ran locally, RuntimeError: OpenCL Error -48
```
chenyuxyz, line 197: is this due to metal not supporting float64? does `METAL=1` run?

geohotstan: What's weird is that `METAL=1` passes this test...... 
But this is due to the same error when `double` is introduced
For instance, 
`test_cast_DOUBLE_to_FLOAT16_cpu` also raises `OpenCL Error -48`

geohotstan:
image

but elem_type: 10 is [float16](https://onnx.ai/onnx/intro/concepts.html#element-type) 
---------- 
openpilot.onnx uses half tensors?
not to sure what to do here...

geohotstan:
Hmmmmmmmm it's still broken...
I have super slow internet rn so it's a little hard to run benchmark
but lemme investigate.

extra/onnx.py
```diff
@@ -49,19 +49,26 @@ def type_parse(type_proto: TypeProto):
       elif attr == 'optional_type': type_proto = getattr(type_proto, attr).elem_type
       else: raise Exception(f"unknown attr: {attr}, {type_proto}")
 
+  # NOTE: FLOAT8E4M3FN = 17, FLOAT8E4M3FNUZ = 18, FLOAT8E5M2 = 19, FLOAT8E5M2FNUZ = 20
```
chenyuxyz, line 52: can you add a link to proto definition?

extra/onnx.py
```diff
@@ -49,19 +49,26 @@ def type_parse(type_proto: TypeProto):
...
-      else:
-        ret = Tensor(np.frombuffer(inp.raw_data, dtype=tensor_dtype_to_np_dtype(inp.data_type)).reshape(inp.dims).astype(np.float32).copy(), requires_grad=False)
+    if inp.data_type in (8,14,15):
+      raise Exception(f"data type not supported {inp.name} {inp.dims} {inp.data_type}")
+    if dat := list(inp.float_data) or list(inp.int32_data) or list(inp.int64_data):
+      ret = Tensor(dat, dtype=DTYPE_MAP[inp.data_type], requires_grad=False).reshape(tuple(inp.dims))
     else:
-      raise Exception(f"bad data type {inp.name} {inp.dims} {inp.data_type}")
+      # TODO half broken in CI for GPU and LLVM segfaults in CI
+      if (dtype := tensor_dtype_to_np_dtype(inp.data_type)) == np.half and CI and Device.DEFAULT in ('GPU', 'LLVM'):
```
chenyuxyz, line 68: do you still need this `if` with a `DTYPE_MAP` which also has an if?

chenyuxyz (commented):
benchmark script failing

chenyuxyz:
i think it's a good idea to separate this into two prs, one for removing redundant dtype casting, and one for changing the default casting behavior. it's fine to keep some manual casting around in between. easier to review too

chenyuxyz:
isnan has some issues on llvm, also metal has the fast-math issue. might want to separate that too with a proper fix

geohotstan:
> isnan has some issues on llvm, also metal has the fast-math issue. might want to separate that too with a proper fix

okok will do. I commented out the test.
-----------

Limited this to only removing redundant casts in onnx.
also tried to snuck in enabling the tests disabled by #1556, commented out the tests that don't work

extra/onnx.py
```diff
@@ -24,6 +25,15 @@ def safe_numpy(t) -> np.ndarray:
...
   return numpy_cache[t]
 
+# src: onnx/mapping.py
+# not supported: STRING = 8 COMPLEX64 = 14, COMPLEX128 = 15
+# NOTE: 17, 18, 19, 20 are float8
+DTYPE_MAP = {1:dtypes.float, 2:dtypes.uint8, 3:dtypes.int8, 4:dtypes.uint16, 5:dtypes.int16, 6:dtypes.int32, 7:dtypes.int64,
+              9:dtypes.bool, 10:dtypes.float16, 11:dtypes.double, 12:dtypes.uint32, 13:dtypes.uint64, 16:dtypes.bfloat16,
+              17:dtypes.float, 18:dtypes.float, 19:dtypes.float, 20:dtypes.float}
+# TODO: fix buffer_parse to use this and fix get_weight_and_biases to only use buffer_parse
+DTYPE_MAP = {i:dtype if Device.DEFAULT == "TORCH" or is_dtype_supported(dtype) else dtypes.int64 if dtypes.is_int(dtype) else dtypes.float32 for i,dtype in DTYPE_MAP.items()}
```
chenyuxyz, line 35: hmm, `is_dtype_supported` is a test helper function and we probably don't want to rely on that for onnx support. what do you think?

chenyuxyz: should work here if we use float for 10 (like the float8 ones)?

geohotstan: Let me try.

Originally I wanted to use `is_dtype_supported` cuz it has very similar code to limiting dtypes as I had hard coded in `onnx.py`, but yeah importing from `test` is really weird.

chenyuxyz:
great thanks!

PR #2945, opened 2023-12-28
cstyle -20 lines (languages refactor)

Qazalin:
- simplify var prefixing
- hip and cuda share logic
- gid,lid,xid -> `code_for_workitem`
- remove MetalRenderer and similar definitions, these are just placeholders we use in runtime init

there's room for saving lines in uops_to_cstyle, then we'll hit the [-50 lines goal](https://docs.google.com/spreadsheets/d/1WKHbT-7KOgjEawq5h5Ic1qUWzpfAzuD_J06N1JwOCGs/edit#gid=1867876150)


Qazalin:
this is mergable but hard to review, I'll break it into 4 PRs

geohot:
Feel free to reopen as smaller PRs as you suggested

PR #2947, opened 2023-12-29
New LLVM renderer

christopherm99:
Unifies renderer syntax with #2632. Still needs to be cleaned up, but is passing all of `test_ops.py`. 

Removing `llvmlite` dependency altogether is definitely possible (with a pretty decent line savings too), but seems a little difficult: not just as simple as replacing `compile_llvm` with a call to `llc` since `llc` can't build dynamic libraries.


geohot:
So as of now this is a large line increase, and I assume many of these abstractions are for PTX?

Want to merge the two PRs? Like there's no real way I'll merge a big line increase, but if it's a line increase with new functionality that's easier to justify.

christopherm99:
Sure, I can just merge this into #2632. I had started removing the lines that were just for PTX but it still wasn't going to be below 5k lines.

PR #2951, opened 2023-12-29
fix TC=2 tensor core op test

chenyuxyz:
correctly enabled TC=2 tensor core op test in CI which was not testing tensor core due to "arm64" check even with TC=2. fixed the broken test by not pushing uop out of loop scope if it has UOps.DEFINE_LOCAL in parents.

also removed some obsolete isinstance checks in tensor core now that LazyOp only has LazyOp in src.


chenyuxyz:
would like to merge this first to unblock fixing dtypes for wmma. full uops graph and fuzz test topology sort can be a separate project

PR #2970, opened 2024-01-02
Add a Nix flake for development and usage

MatthewCroughan:
Adds a `flake.nix` that you interact with via `nix develop`, `nix shell` or `nix build` to get up and running with Tinygrad with no extra steps on any platform.

I haven't tried it on a GPU yet, but I will later use [Mobile NixOS](https://mobile.nixos.org/) to try it on an SDM845

## macOS (lol)

Since I'm too poor to own a Mac I used [NixThePlanet](https://github.com/MatthewCroughan/NixThePlanet) to run Darwin with a single Nix command to test

### x86_64-darwin
image

### aarch64-darwin
image

# Linux

### x86_64-linux

image

### aarch64-linux

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ wandb
 model.safetensors
 quickstart.py
 .hypothesis
+*result*
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,34 @@
+{ lib
+, buildPythonPackage
+, setuptools
+, wheel
+, numpy
+, tqdm
+, ...
+}@args:
+
+buildPythonPackage {
+  pname = "tinygrad";
+  version = args.version;
+  pyproject = true;
+
+  src = ./.;
+
+  propagatedBuildInputs = [
+    numpy
+    tqdm
+  ];
+
+  nativeBuildInputs = [
+    setuptools
+    wheel
+  ];
+
+  pythonImportsCheck = [ "tinygrad" ];
+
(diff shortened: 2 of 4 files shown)


wozeparrot:
Needs a `gpuctypes` package for hip/cuda/opencl to work: https://github.com/wozeparrot/tinygrad-nix/blob/5b8b5d54dc3150d279048dd196e8ef5062f19ab4/flake.nix#L29

There was a previous pr for a nix devshell: #1026

MatthewCroughan:
@wozeparrot ah awesome, I'll develop this a bit more, I've upstreamed it to nixpkgs here just so that we have a derivation there too. https://github.com/NixOS/nixpkgs/pull/278187

wozeparrot:
The hip stuff doesn't quite work yet so i would just remove it.

geohot:
I think we aren't going to merge this since it's an extra thing we have to maintain, though I'm supportive of it being a part of nixpkgs, particularly if there's a pinned version.

MatthewCroughan:
Yeah I'll just pop it into Nixpkgs, especially since this Python code is intending to be minimal and not too heavy on dependencies it should work out well. But if you start adding a bunch of deps it can be really beneficial to have a Flake, as pip will no longer work as well as you might expect it to. Having a Flake for development is still desirable, but I'll bring that up if I have more examples in the future, and if it becomes more evident.

PR #2977, opened 2024-01-02
Simpler variable prefixing in cstyle

Qazalin:
- gets rid of `lang.generic_var_prefix if lang.generic_var_prefix else dtype.name`
- `size_prefix` and `generic_var_prefix` are conceptually close enough to share a function


Qazalin:
this no longer makes sense because webgpu is removed from tinygrad core

PR #2981, opened 2024-01-02
adjsut div factor to avoid underflow for cifar in fp16

g1y5x3:
when [`OneCycleLR`](https://github.com/tinygrad/tinygrad/blob/ad0d710ec442a34a4b0278f24b27d096d0d34d6b/extra/lr_scheduler.py#L69C65-L69C65) is initialized in fp16, `max_lr / div_factor` resulted a value `1.993e-17` that caused underflow and nan.


chenyuxyz:
if half + hip cifar is fixed, can re-enable this https://github.com/tinygrad/tinygrad/blob/ad0d710ec442a34a4b0278f24b27d096d0d34d6b/.github/workflows/benchmark.yml#L136 and commented out the Run 10 CIFAR training steps w winograd above since it's quite slow and having one is enough

PR #3003, opened 2024-01-04
Improve HIP support on Intel HW, fix few double exceptions and refactor the backend code

sobomax:
This patch gets rid of the following warning when running HIP backend on Intel HW using chipStar project as well as reorganizes the HIP backend code to reduce duplication and boilerplate (as well as gets rid of the most of the noqa: E501). It also fixes few double exceptions here and there.

```
CHIP warning [TID 177892] [1704330852.929269646] : hiprtc: ignored option: '--offload-arch=unavailable'
CHIP warning [TID 177892] [1704330852.929358833] : hiprtc: ignored option: '-I/opt/rocm/include'

### DIFF
--- a/tinygrad/runtime/graph/hip.py
+++ b/tinygrad/runtime/graph/hip.py
@@ -7,8 +7,8 @@ from tinygrad.runtime.graph.cuda import CUDAGraph
 
 class HIPGraph(CUDAGraph):
   def __del__(self):
-    check(hip.hipGraphDestroy(self.graph))
-    check(hip.hipGraphExecDestroy(self.instance))
+    if hasattr(self, 'graph'): check(hip.hipGraphDestroy(self.graph))
+    if hasattr(self, 'instance'): check(hip.hipGraphExecDestroy(self.instance))
 
   def encode_args_info(self): return (hip.hipDeviceptr_t, (1,2,3))
   def graph_create(self): return init_c_var(hip.hipGraph_t(), lambda x: check(hip.hipGraphCreate(ctypes.byref(x), 0)))
--- a/extra/gemm/hip_matmul.py
+++ b/extra/gemm/hip_matmul.py
@@ -2,7 +2,7 @@ import time
 import numpy as np
 from tinygrad import dtypes
 from tinygrad.helpers import getenv, prod, flat_mv
-from tinygrad.runtime.ops_hip import HIPAllocator, HIPProgram, compile_hip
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: tinygrad/runtime/graph/hip.py [hunk 1]
```diff
@@ -6,9 +6,7 @@
 from tinygrad.runtime.graph.cuda import CUDAGraph
 
 class HIPGraph(CUDAGraph):
-  def __del__(self):
-    check(hip.hipGraphDestroy(self.graph))
-    check(hip.hipGraphExecDestroy(self.instance))
+  def __del__(self): _=[x(getattr(self, y)) for x, y in ((hip.hipGraphDestroy, 'graph'), (hip.hipGraphExecDestroy, 'instance')) if hasattr(self, y)]
```
geohot, line 9: unreadable code golf

sobomax: Fixed.

pjaaskel:
@sobomax: with this patch (and the one you submitted to chipStar), are you able to correctly run some interesting models through tinygrad on top of chipStar on an Intel GPU, or did you find other issues?

sobomax:
@pjaaskel in fact I did. I went all the way to working Llama 7B (cannot test anything bigger since I only have 16GB card). There are at least one problem remaining open at the chipStar repo: https://github.com/CHIP-SPV/ROCm-Device-Libs/pull/2, on top of which I have has to patch gputypes to import from the correct .so files. And the last but not least, graph kernels are broken at the moment at CS, so JIT=0 or JIT=2 are needed. I quickly ran some of the unit tests, it seems to be problems still here and there, but as I said nothing preventing LLAMA to run. As a good test, I have had it to recall PI number to like 250 correct digits before it failed.

pjaaskel:
Nice. Did you test the Level0 or the OpenCL backend of chipStar? We might want to add this as a test case to chipStar.

sobomax:
Mostly with level0, tinygrad has quite a decent opencl support, so I did not feel like doing it much. One layer of abstraction less.

geohot:
Looks complex, HIP on Intel is not a development target of tinygrad.

pjaaskel:
@sobomax: 
> Mostly with level0, tinygrad has quite a decent opencl support, so I did not feel like doing it much. One layer of abstraction less.

Curious: How does going through HIP and chipStar compare with just using the tinygrad's OpenCL support directly? Intel HW has good OpenCL drivers.

PR #3009, opened 2024-01-04
Tensor Core support in CUDA

mobley-trent:
This PR attempts to showcase Tensor Core support in CUDA using tinygrad


mobley-trent:
Work in progress

g1y5x3:
this is the same as the [one](https://github.com/tinygrad/tinygrad/blob/master/extra/gemm/cuda_matmul.py) that we already have. not real tensor core support for cuda.

PR #3017, opened 2024-01-05
[proof of concept] new fancy indexing

geohotstan:
Idk if this idea is workable so I need some feedback
if this is optimizable or not or if `_to_list_val()` is proper tinygrad lol
currently `test_indexing` passes and `test_ops` passes

How it works:

1. takes Tensor indices and converts them into `(shape, flat buffer)` pairs. This is done through inspecting the `_buf` directly without any `.realize()` if possible, otherwise have to `toCPU` (see `_to_list_val()`)
2. do not convert list/tuple indices into Tensors but instead converts them to `(shape, fully_flatten(indices))`

above 2 steps essentially gets `(shape, flat_indices)` for every dim and fancy indexing only uses that.
for ex: 
image
has `indices = [slice(None, None, None), ([2, 2], [0, 1, 3, 4]), ([2, 2], [2, 3, 0, 1]), slice(None, None, None)]`

3. for fancy indexing, that big chunky block of cursed looking spaghetti code is just decomposing the `(shape, flat_indices)` pairs into `shrink args` to shrink the tensor for each indices 
4. Then we `cat` the shrunken indices together to get the resulting Tensor after each round of fancy

for ex (using prev ex):
```
# shrink args for dim=1
[(0, 3), (0, 1), (0, 8), (0, 6)]
[(0, 3), (1, 2), (0, 8), (0, 6)] 
[(0, 3), (3, 4), (0, 8), (0, 6)]
[(0, 3), (4, 5), (0, 8), (0, 6)]

# after shrinking and cat-ing the fancy indexing on dim=1
# shrink args for dim=2
[(0, 3), (0, 1), (0, 1), (2, 3), (0, 6)]
[(0, 3), (0, 1), (1, 2), (3, 4), (0, 6)]
[(0, 3), (1, 2), (0, 1), (0, 1), (0, 6)]
[(0, 3), (1, 2), (1, 2), (1, 2), (0, 6)]

# NOTE: there are some weird reshapes tucked into all this as well. I'll clean this up. This is only a PoC
```

doing this decomposing is, however, not trivial :( 
despite how much I wished it was so.
by doing this for each dim, we effectively get only 1 additional cat 😸 for each additional idx for any given dim

```
# still for the prev example
# all MemBuffer args look very similar. Bunch of Views......
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(3, 2, 2, 6), strides=(24, 12, 6, 1), offset=0, mask=None, contiguous=True),)))                                                                                                                          
  1  ┗━┳ ADD                                                                                                                                     
  2    ┣━┳ ADD 
  3    ┃ ┣━┳ ADD 
  4    ┃ ┃ ┣━┳ ADD 
  5    ┃ ┃ ┃ ┣━┳ ADD 
  6    ┃ ┃ ┃ ┃ ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(3, 2, 8, 6), strides=(432, 0, 6, 1), offset=0, mask=((0, 3), (0, 1), (0, 8), (0, 6)), contiguous=False), View(shape=(3, 3, 8, 6), strides=(96, 48, 6, 1), offset=0, mask=((0, 3), (0, 2), (0, 8), (0, 6)), contiguous=False), View(shape=(3, 4, 8, 6), strides=(144, 48, 6, 1), offset=0, mask=((0, 3), (0, 3), (0, 8), (0, 6)), contiguous=False), View(shape=(3, 1, 2, 6), strides=(192, 0, 0, 1), offset=12, mask=((0, 3), (0, 1), (0, 1), (0, 6)), contiguous=False), View(shape=(3, 2, 2, 6), strides=(12, 0, 6, 1), offset=0, mask=((0, 3), (0, 1), (0, 2), (0, 6)), contiguous=False))))
  7    ┃ ┃ ┃ ┃ ┗━━ LOAD MemBuffer(...)
  8    ┃ ┃ ┃ ┗━━ LOAD MemBuffer(...)
  9    ┃ ┃ ┗━━ LOAD MemBuffer(...)
 10    ┃ ┗━┳ ADD 
 11    ┃   ┣━┳ ADD 
 12    ┃   ┃ ┣━┳ ADD 
 13    ┃   ┃ ┃ ┣━━ LOAD MemBuffer(...) 
 14    ┃   ┃ ┃ ┗━━ LOAD MemBuffer(...) 
 15    ┃   ┃ ┗━━ LOAD MemBuffer(...) 
 16    ┃   ┗━━ LOAD MemBuffer(...) 
 17    ┗━┳ ADD         
 18      ┣━┳ ADD      
 19      ┃ ┣━┳ ADD                                                                                                                                
 20      ┃ ┃ ┣━┳ ADD 
 21      ┃ ┃ ┃ ┣━━ LOAD MemBuffer(...)
 22      ┃ ┃ ┃ ┗━━ LOAD MemBuffer(...)
 23      ┃ ┃ ┗━━ LOAD MemBuffer(...)
 24      ┃ ┗━━ LOAD MemBuffer(...)
 25      ┗━┳ ADD                                                                                                                                 
 26        ┣━┳ ADD     
 27        ┃ ┣━┳ ADD 
 28        ┃ ┃ ┣━━ LOAD MemBuffer(...)
 29        ┃ ┃ ┗━━ LOAD MemBuffer(...)
 30        ┃ ┗━━ LOAD MemBuffer(...)
 31        ┗━━ LOAD MemBuffer(...)
```

KERNEL for fancy indexing on 2 dims with NOOPT=1 in CLANG=1
```
void E_3_2_2_6(int* restrict data0, const int* restrict data1) {
  for (int ridx0 = 0; ridx0 < 3; ridx0++) {
    for (int ridx1 = 0; ridx1 < 2; ridx1++) {
      bool alu0 = (ridx1<1);
      bool alu1 = ((ridx1*(-1))<0);
        bool alu2 = (ridx2<1);
        bool alu3 = ((ridx2*(-1))<0);
        for (int ridx3 = 0; ridx3 < 6; ridx3++) {
          int alu4 = (ridx3+(ridx0*432));
          int val0 = ((alu0*alu2))?(data1[alu4+12]):0;
          int val1 = ((alu0*alu3))?(data1[alu4+66]):0;
          int val2 = ((alu1*alu2))?(data1[alu4+144]):0;
          int val3 = ((alu1*alu3))?(data1[alu4+198]):0;
          data0[(ridx0*24)+(ridx1*12)+(ridx2*6)+ridx3] = (val0+val1+(val2+val3));
        }
      }
    }
  }
}
```

-----------------------------
Now here's a story of why I decided this would be the way to do it.

... in one foreboding ominous evening when geohotstan was busy cooking spaghetti code
he found that bool indexing was, with the current arsenal of tensor.py APIs, impossible.
similar to how `mask_select` g1y5x3 needed was impossible (I think)
[btw MLX doesn't support it either :D](https://github.com/ml-explore/mlx/blob/c3272d49175470a0f89ad6835bde79a55dc364f9/mlx/ops.cpp#L1779C37-L1779C37)
and I'm pretty sure jax's lazy stuff does some weird stuff too.
They all face similar problems we do when it comes to fancy indexing (like jax also can't raise out of bounds error for fancy)

But fancy indexing or bool indexing is just applying a mask over the original tensor and `cat`ing right?
This was the approach of the original `gather op in onnx` (that only worked with 1dim indices tensors IIRC) which was really fast! For small number of indices though.

Also the original fancy indexing I wrote before has indices Tensor all with `requires_grad=False` and often times they're creating a Tensor just to fancy index (which seems really pointless cuz we can just use a list instead of going through with tensor creation). 
I also think these indices Tensors are realizable?? since they do not interfere with the compute graph of the `self Tensor`, which means backwards still works fine.

back to bool indexing.
bool indexing is essentially
```
# bool idx
[False, True, True, False]

is just
# int idx
[1, 2]
```
so if we can retrieve the value from the Tensor indices, we're in business!
`_to_list_val()` attempts to do that.
retrieving it from `_buf` is free
The Tensor indices created just to fancy index has `_buf`
embeddings also just creates a tensor just for lookup so it also has `_buf`

Then I wrote a fancy indexing based on the premise of us having the indices value, which was surprisingly difficult to write, and surprisingly slow for large numel indices. But doing this way also fits fancy in one kernel

There are many more alternative methods of fancy I tried with this shrink cat method (which used up a lot of time), and this one has the least ops and is currently the fastest.

---------------
doesn't support bool, empty
yet


chenyuxyz:
i will take a look tomorrow, it's very possible that we decide not to support it

python array api also recognize that value dependent index does not work with static framework like jax (and tinygrad)
https://data-apis.org/array-api/2022.12/design_topics/data_dependent_output_shapes.html#data-dependent-output-shapes

chenyuxyz:
yea i am inclined to not support boolean indices. value dependent behavior will always need to peek the value inside tensor, and it's not jittable. jax doesn't support it either. @geohot let me know if you have a different opinion

```
>>> import jax.numpy as jnp
>>> jnp.array([1,2,3])
Array([1, 2, 3], dtype=int32)
>>> jnp.array([1,2,3])[2]
Array(3, dtype=int32)
>>> jnp.array([1,2,3])[True]
...
    raise TypeError("JAX arrays do not support boolean scalar indices")
TypeError: JAX arrays do not support boolean scalar indices
```

please use this principle to disable tests in test_indexing, bounty is yours once it's all cleaned up, and the doc of getitem is updated accordingly

geohotstan:
> python array api also recognize that value dependent index does not work with static framework like jax (and tinygrad)
> https://data-apis.org/array-api/2022.12/design_topics/data_dependent_output_shapes.html#data-dependent-output-shapes

I read through everything in this Python array API standard doc (learned a lot)
They do say [here](https://data-apis.org/array-api/2022.12/API_specification/indexing.html#boolean-array-indexing) that boolean indexing support is a must have. 
What they omit is instead the use of dynamically sized boolean array to mask (index) another array.

So for instance this would be omitted I think?
image

Otherwise, this array api standard requires bool indexing.

------------------------

> value dependent behavior will always need to peek the value inside tensor, and it's not jittable.

While this is true, in most use cases indices are actually "non-tensors", but because of how fancy indexing is implemented currently, we have to rely on roundabout Tensor operations (eq w/ arange -> mul -> sum).

Let me give a more clear example.
image
This is actually equal!!!

```
# for this
inp[:,a,b,:]

# indices passed in
indices = (slice(None, None, None), <Tensor <LB METAL (2, 2) contig:True ...>, <Tensor <LB METAL (2, 2) contig:True ...>, slice(None, None, None))

# while for passing in list
inp[:,[[0,1],[3,0]],[[2,3],[0,1]],:]

# indices passed in 
# notice how we have the values here!!!!!!!!
indices = (slice(None, None, None), [[0, 1], [3, 0]], [[2, 3], [0, 1]], slice(None, None, None))
```
Both is used to fancy index to the same result

Similarly!
image

The thing with current tinygrad lazy is that 
image

with this PoC PR, I just wanted to illustrate if I had the (shape, flattened_array) pair, I could use only that to fancy index.

chenyuxyz:
oh yeah i was thinking about passing boolean indices as True means taking the entry and False mean don't take. using boolean as a mask we should support

we can compare with jax and see what they support too

geohotstan:
image
Doesn't this mean "passing boolean indices as True means taking the entry and False mean don't take"? Or am I mistaken...


image

Jax currently supports this
but it isn't jittable

image

chenyuxyz:
no you are not mistaken, that's what i meant. i understand that by inspecting the value in _buf, we can generate the code from the values, and that's not jittable and the kernel is not optimizable. if we can come up with a nice solution that's general enough and jittable (for example jax.unique can provide a size to be jittable) then we can support it, otherwise i prefer to skip it for now (and we won't have full python array api support)

geohotstan:
Ahhh got it
Thanks :D

PR #3019, opened 2024-01-05
Update device.py

reddyn12:
Bug fix for error:
```
  File "/cache/home/nr620/tinygrad/examples/testMulti.py", line 71, in <module>
    O = simpFunc(Xs)
  File "/cache/home/nr620/tinygrad/examples/testMulti.py", line 62, in simpFunc
    t = x+1
  File "/cache/home/nr620/tinygrad/tinygrad/tensor.py", line 832, in __add__
    def __add__(self, x) -> Tensor: return self.add(x)
  File "/cache/home/nr620/tinygrad/tinygrad/tensor.py", line 786, in add
    return mlops.Add.apply(*self._broadcasted(x, reverse)) if x.__class__ is Tensor or x else self
  File "/cache/home/nr620/tinygrad/tinygrad/tensor.py", line 33, in apply
    ret = Tensor(ctx.forward(*[t.lazydata for t in x], **kwargs), device=ctx.device, requires_grad=ctx.requires_grad)
  File "/cache/home/nr620/tinygrad/tinygrad/mlops.py", line 97, in forward
    def forward(self, x:LazyBuffer, y:LazyBuffer) -> LazyBuffer: return x.e(BinaryOps.ADD, y)
  File "/cache/home/nr620/tinygrad/tinygrad/features/multi.py", line 91, in e
    assert all_same([x.device for x in msrcs]), f"all buffers must have the same device {[x.device for x in msrcs]}"
AssertionError: all buffers must have the same device [('GPU:0', 'GPU:1', 'GPU:2', 'GPU:3'), ('GPU', 'GPU:1', 'GPU:2', 'GPU:3')]
```
for code:
```python
X = Tensor.ones(N, N)
Xs = X.shard(('GPU:0','GPU:1','GPU:2','GPU:3'), 0) def simpFunc(x) -> Tensor:
    t = x+1
    t = t.sum(1)
    return t.realize()
O = simpFunc(Xs)
```

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -16,7 +16,7 @@ if TYPE_CHECKING:
 
 class _Device:
   def __init__(self) -> None: self._buffers: List[str] = [x.stem[len("ops_"):].upper() for x in (pathlib.Path(__file__).parent/"runtime").iterdir() if x.stem.startswith("ops_")]  # noqa: E501
-  def canonicalize(self, device:Optional[str]) -> str: return (device.split(":", 1)[0].upper() + ((":"+device.split(":", 1)[1]) if ':' in device else '')).replace(":0", "") if device is not None else self.DEFAULT  # noqa: E501
+  def canonicalize(self, device:Optional[str]) -> str: return (device.split(":", 1)[0].upper() + ((":"+device.split(":", 1)[1]) if ':' in device else '')) if device is not None else self.DEFAULT  # noqa: E501
   def __getitem__(self, ix:str) -> Union[Interpreted, Compiled]: return self.__get_canonicalized_item(self.canonicalize(ix))
   @functools.lru_cache(maxsize=None)  # this class is a singleton, pylint: disable=method-cache-max-size-none
   def __get_canonicalized_item(self, ix:str) -> Union[Interpreted, Compiled]:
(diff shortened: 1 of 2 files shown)


chenyuxyz:
that's not the right fix. it should be fixed in ops_gpu. also bugfix pr needs a regression test

reddyn12:
It breaks in the same way for CUDA also. It's a new bug introduced with multi tensor. The replace method will always replace ':0' with '' irregardless of multi or single gpu. Will try adding a multi flag for device

reddyn12:
Also looks like Linters Unit Test has a bug. An else statement can follow a return statement

reddyn12:
There is a test for it, but the unit test is skipped for GPU and CUDA
```python
unittest.skipIf(CI and Device.DEFAULT in {"GPU", "CUDA"}, "no GPU CI")
class TestMultiTensor(unittest.TestCase):
  def test_shard(self):
    X = Tensor.ones(256).contiguous().realize()
    X.shard_((d0, d1), 0)
    for lb in X.lazydata.lbs:
      assert lb.shape == (128,)

  def test_shard_same_device(self):
    X = Tensor.ones(256).contiguous().realize()
    X.shard_((d0, X.device), 0)
    (X + X).realize()

  def test_shard_plus_one_sum(self):
    X = Tensor.ones(256).contiguous().realize()
    X.shard_([d0, d1], 0)
    (X + 1).sum().realize()
```

reddyn12:
@chenyuxyz Is this a good fix?

chenyuxyz:
it's not the right fix. but the issue is legit. feel free to file an issue or commit a broken test and skip it in CI

PR #3020, opened 2024-01-05
add test_softmax to test_ops

chenyuxyz:
somehow it was not tested


chenyuxyz:
eh WEBGPU buffer check is wrong, there's only one buffer here

PR #3024, opened 2024-01-05
add const support for multi gpu

g1y5x3:


reddyn12:
Pr #3019 should pass the tests

PR #3029, opened 2024-01-06
multitensor jit

ramenguy99:
- [x] track MLB devices and shapes in expected buffers
- [x] Simple add and resnet multitensor tests with jit

Not sure about the following:
- Does the bounty also require supporting graphs for multitensor? I guess the best solution here would be to have a graph per device, currently I just disable graphs if any MLB is being passed to the jit.
- Should tests go to a different file or be duplicated between jit / no jit?
- I could only test this with CPU=1 since I don't have a multi gpu system at hand.


geohot:
So I don't think graphs have to be supported, but I don't think that's the right way to exclude them. Even if a multi isn't passed in, they are still be used.

Re: tests and JIT, I'd just add a representative couple that test all the cases of the JIT.

ramenguy99:
As far as I understand graphs don't work in the current JIT with any buffer that is not created on the default device, so I wasn't sure how to make it do the right thing when multiple devices are involved. 

I couldn't find an easy way to discriminate JITRunners that could be graphed together. So I ended up adding a `device` field to `CompiledASTRunner` and use it to divide items in batches for the graph. 

Ideally this allows multitensors to take advantage of graphing too, but it relies on the jit items to be contiguous for each device which currently is not the case all the times. I guess this is something that the scheduler/linearizer should be responsible for, so maybe it's fine for the JIT to expect that?

I cleaned up the tests and added a test with multitensors in the function body and with graphs.

tinygrad/device.py
```diff
@@ -241,13 +241,13 @@ def _interpret_ast(ast:LazyOp) -> str:
 # **************** for Compiled Devices ****************
 
 class CompiledASTRunner(JITRunner):
-  def __init__(self, ast:Optional[LazyOp], name:str, prg:str, lib:bytes, global_size:Optional[List[int]]=None, local_size:Optional[List[int]]=None):  # noqa: E501
+  def __init__(self, ast:Optional[LazyOp], name:str, prg:str, lib:bytes, device:Compiled, global_size:Optional[List[int]]=None, local_size:Optional[List[int]]=None):  # noqa: E501
```
geohot, line 244: Why is this new parameter required? If it really is, can you add it in a separate PR?

tinygrad/jit.py
```diff
@@ -44,36 +45,36 @@ def reset(self):
     self.cnt: int = 0
     self.ret: Optional[ReturnType] = None
     self.expected_vals: Optional[Tuple[Variable, ...]] = None
-    self.expected_name_sts_dtype: Optional[Tuple[Tuple[Union[int, str], ShapeTracker, DType], ...]] = None
+    self.expected_name_sts_dtype_device: Optional[Tuple[Tuple[Union[int, str], ShapeTracker, DType, Union[str, Tuple[str, ...]]], ...]] = None
```
geohot, line 48: Ahh, this is a good point and should be added anyway

geohot:
How much of this is prereqs that can go in a separate PR?

tinygrad/device.py
ramenguy99: I need a way to get from a JITRunner to the device that should be used to graph it, I couldn't find a way without adding this. Could go to a different PR but only the jit Grapher needs this atm

ramenguy99:
What were you thinking of splitting? I guess the main components of this PR are:
- Added device tracking to the JIT (modifies one loc)
- Extended the jit to work on multibuffers by flattening the inner lazybuffers
- The graph stuff (including the `device` field added to `CompiledASTRunner`) and the extra logic in the jit to split based on it

tinygrad/device.py
geohot: I have a fix for this in master, the device is in clprg.

geohot:
Actually this is sort of nicer with device than my way, sorry for being non-responsive, was at CES.

Let's get this upstreamed.

geohot:
I'm fine with this as is if we get the tests passing (and adding device)

geohot:
Freed up 20 lines for this

PR #3032, opened 2024-01-07
MNIST autoencoder example

fishhf:
Adding a simple MNIST autoencoder example as tinygrad doesn't have one.
Not naming it as beautiful because I'm not sure if it is. :)


terafo:
This shouldn't be in tinygrad repo.

g1y5x3:
feel free to make a separate repo for this example

PR #3033, opened 2024-01-07
update gh actions

jxdv:


geohot:
Why? Is there an advantage?

jxdv:
> Why? Is there an advantage?

Speed improvements in the upload-artifact based on their changelog, node20 used instead of 16 in setup-python. Up 2 you if you want to merge

PR #3035, opened 2024-01-07
Update tensor.py with no_grad

unknownusername504:
Just like train var, allows for idiomatic with statements that effectively save and restore the state of no_grad.


g1y5x3:
can you add tests? there's some todos in the test/ where this can be used.

PR #3037, opened 2024-01-07
Fix backward fn for `<` and `==`

gml16:
Running the below code fails as the mlops `Less` has no backward method:
```
tt = Tensor.randn(4, requires_grad=True)
(tt*(tt < 0)).sum().backward()
```
However, the equivalent Torch code is valid. 
Similarly for mlops `Eq`.

I made the simple changes of adding those backward methods which just propagate `None` gradients.

Also added the corresponding tests.


geohot:
Hmm, very interesting.

gml16:
@geohot good point, thank you. Using autograd's variable to attach gradient has not been needed for a while in PyTorch. I have removed it in the latest commit.
Can you give me a pointer as to what you mean about the automatic stuff, please? I don't know what you're referring to. Thanks

geohot:
Eh, I meant the test actually testing the backward. But I'm fine with this

PR #3064, opened 2024-01-10
add mutigpu support for llama attention

g1y5x3:
add multigpu support for llama attention with tests
~~still need to fix the kv cache~~
need to make a small for loop to test kv cache more properly

### DIFF
--- a/extra/models/llama.py
+++ b/extra/models/llama.py
@@ -47,6 +47,11 @@ class Attention:
     self.head_dim = dim // n_heads
     self.n_rep = self.n_heads // self.n_kv_heads
     self.max_context = max_context
+    print("\n init")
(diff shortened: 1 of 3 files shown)


g1y5x3:
I think I run into the case where reshape in multigpu is wrong.

g1y5x3:
The current `scaled product attention` isn't really appropriate because the attention module inside transformer xq, xk, xv are never sharded. The sharding happens at the `n_heads` level that was able to produce independent xqs, xks and xvs without all reduce untl the very end linear layer. Need to modify the `cat` to be able to take `MutiLazyBuffer`.

g1y5x3:
torch backend is a runtime error and the cpu backend's error is from numpy

Issue #2829, opened 2023-12-18
JIT custom random has copyin error on Interpreted backends (examples/transformer.py)


tostenzel:
python examples/transformer.py 
loss 2.32 accuracy 0.09:   2%|█▎                                                              | 1/50 [00:00<00:29,  1.64it/s]
Traceback (most recent call last):
  File "examples/transformer.py", line 31, in <module>
    train(model, X_train, Y_train, optim, 50, BS=64)
  File "/home/tobias/Documents/tinygrad/extra/training.py", line 31, in train
    loss, accuracy = train_step(x, y)
  File "/home/tobias/Documents/tinygrad/tinygrad/jit.py", line 77, in __call__
    assert len(set(get_input_replace(self.jit_cache, input_rawbuffers).values())) == len(input_rawbuffers), "some input tensors not found"
AssertionError: some input tensors not found

Looking up the values: assert 1 == 2

chenyuxyz:
which backend is this? and is it fixed on master? i cannot reproduce on METAL
<img width="1506" alt="image" src="https://github.com/tinygrad/tinygrad/assets/2525478/75465086-cca9-4625-9de4-bf5ea0e69ace">

tostenzel:
I use CPU. Below is my system info.
![Screenshot from 2023-12-21 17-53-35](https://github.com/tinygrad/tinygrad/assets/32765108/8620066b-62a1-4155-909f-9635a9932fde)

chenyuxyz:
TORCH=1 gave a more useful error message
```
  File "/Users/chenyu/code/tinygrad/tinygrad/runtime/ops_torch.py", line 49, in copyin
    def copyin(self, dest:torch.Tensor, src:memoryview): dest.copy_(torch.frombuffer(src, dtype=dest.dtype))
                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: The size of tensor a (128) must match the size of tensor b (49152) at non-singleton dimension 3
```

argosopentech:
I'm seeing this issue too on an Intel CPU

```
loss 5.54 accuracy 0.00:   1%|▊                                                                                   | 1/100 [00:07<12:38,  7.66s/it]
Traceback (most recent call last):
  File "/home/pj/Desktop/Tinygrad/tinytranslate/translate.py", line 100, in <module>
[...]

PR #3073, opened 2024-01-10
fix onehot and jit in examples/transformer

chenyuxyz:
trained to 0.999 in < 6 seconds on M1 Max consistently


chenyuxyz:
#2829 is still a bug though

PR #3075, opened 2024-01-10
remove numpy, too aggressive

geohot:
If someone wants to break apart this PR, very carefully do each piece of it, and make the tests pass that would be great.

There's some annoyance with the float16 type, 'e' doesn't seem to work.


chenyuxyz:
it looks like float16 support in memoryview is a 3.12 feature https://github.com/python/cpython/issues/90751

savagesmc:
Started looking at this over the weekend. I am using this as an opportunity to start understanding your codebase better. How were you all thinking of replacing the 'shape' functionality in numpy? It looks like native python arrays are all 1-d arrays. Were you thinking of adding something to tinygrad.shape to represent N-dimensional arrays out of 1-d native arrays?

geohot:
This has been (mostly) merged in a series of PRs. I don't see the urgent need to actually remove numpy from the dtype

PR #3081, opened 2024-01-11
wmma: clean up device specific tensor core code

flammit:


flammit:
Cleaned-up the code to remove device-specific logic from `linearizer.py` and also set the stage for half/bfloat16 TC support in METAL.

chenyuxyz:
thanks!

PR #3084, opened 2024-01-11
Implement onehot

obadakhalili:


chenyuxyz:
it should be using Tensor, not numpy array! and need test and less lines!

PR #3092, opened 2024-01-11
Llama support weight sharding

g1y5x3:
image

image

~~Currently host memory will overflow if trying to load bigger model due to the weights being cached while sharding.~~
~~This is dependent on the attention changes made in #3064 which is a correctness check for attention with sharded weights but somehow the CI fails on interpreted backend so I need to take a further look.~~

- [x] 7B sharded across 2, 4 GPUs
- [x] 13B sharded across 2, 4 GPUs
- [x] 70B sharded across 6 GPUs - two make weights evenly sharded across GPUs, the layer norm at the beginning and the first 54 transformer layers are sharded arcoss 4 GPUs and the remaining 26 layers and the last layer norm as well as linear layer are sharded across the other 2 GPUs. 
- [x] need to clean up the code and try to use minimal amount of copy, also more unifying interface between choosing different type of models

currently the speed is bottleneck by the layers sharded across 2 GPUs essentially, need to improve it.

### DIFF
--- a/examples/llama.py
+++ b/examples/llama.py
@@ -131,6 +131,7 @@ def load(fn:str):
   else:
     return torch_load(fn)
 
+# TODO: 4 GPU rotational loads
 device1 = (d2, d3, d4, d5)
 device2 = (d0, d1)
 def load_state_dict_shard(model, state_dict, devices, strict=True, verbose=True):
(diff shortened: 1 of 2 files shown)


g1y5x3:
70B llama 2 loaded and running
image
image

PR #3093, opened 2024-01-11
One hot in tensor.py

SnakeOnex:
- [x] one_hot in tensor.py
- [x] tests
--> ONNX reimplementation?

So, I definitely used the ONNX implementation as some kind of reference, but the ONNX OneHot method has more arguments and is more complicated than the pytorch one. ONNX supports negative indices, custom values for the encoding (not just 0/1), custom one-hot dim position, etc. 

The question is, should I also look at the ONNX implementation and make it use the Tensor method instead, or just leave it be as is, since it's quite a bit differerent compared with pytorch.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -211,7 +211,6 @@ class Tensor:
   def zeros_like(self, **kwargs): return self.full_like(0, **kwargs)
   def ones_like(self, **kwargs): return self.full_like(1, **kwargs)
 
-
   # ***** rng hlops *****
 
   @staticmethod
@@ -499,7 +498,6 @@ class Tensor:
     slc = [(-p0, s+p1) for p0,p1,s in zip(padding[::2], padding[1::2], self.shape[::-1])][::-1]
     return self.slice([(0,s) for s in self.shape[:-(len(padding)//2)]] + slc, value=value)
 
-  def one_hot(self, num_classes:int, **kwargs) -> Tensor: return Tensor.where(self[..., None] == Tensor.arange(num_classes), 1, 0, **kwargs)
 
   @property
   def T(self) -> Tensor: return self.transpose()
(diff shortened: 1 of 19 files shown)


tinygrad/tensor.py
```diff
@@ -211,6 +211,11 @@ def full_like(self, fill_value:Scalar, **kwargs):
   def zeros_like(self, **kwargs): return self.full_like(0, **kwargs)
   def ones_like(self, **kwargs): return self.full_like(1, **kwargs)
 
+  @staticmethod
```
chenyuxyz, line 214: why staticmethod?

SnakeOnex: I was going by pytorch, where its only in torch.nn.functional, so a static method. Do you want it as a non-static method or have both versions?

chenyuxyz: there are so many functions only in `torch.nn.functional` that is not static in tinygrad (see test_ops). definitely not both. why is one better than another?

SnakeOnex: I agree, seems like the precedent is that the methods that create initialize a new tensor are static, but the ones that modify an existing tensor are not. Changed it to be a regular instance method.

tinygrad/tensor.py
```diff
@@ -211,6 +211,10 @@ def full_like(self, fill_value:Scalar, **kwargs):
   def zeros_like(self, **kwargs): return self.full_like(0, **kwargs)
   def ones_like(self, **kwargs): return self.full_like(1, **kwargs)
 
+  def one_hot(self, num_classes:int=-1, **kwargs) -> Tensor:
+    num_classes = self.max().item()+1 if num_classes == -1 else num_classes
```
chenyuxyz, line 215: can you do this without going through item()? otherwise it's value dependent and not jittable. it's fine to always require a num_classes i think (see how jax does it)

also can you move this a suitable section under `functional nn ops` now it's not a creation function? same for test

SnakeOnex: yeah, makes sense my mistake, I looked at Jax and as you thought they have num_classes as required argument.

tinygrad/tensor.py
```diff
@@ -498,6 +499,8 @@ def pad2d(self, padding:Sequence[int], value:float=0) -> Tensor:
     slc = [(-p0, s+p1) for p0,p1,s in zip(padding[::2], padding[1::2], self.shape[::-1])][::-1]
     return self.slice([(0,s) for s in self.shape[:-(len(padding)//2)]] + slc, value=value)
 
+  def one_hot(self, num_classes:int, **kwargs) -> Tensor: return Tensor.where(self[..., None] == Tensor.arange(num_classes), 1, 0, **kwargs)
```
chenyuxyz, line 502: i thinks this belongs to `functional nn ops`.

also double check pr diff for spaces

chenyuxyz:
thanks! would like to see some examples that uses this in training to make sure the backward is handled correctly (there's one use in examples/transformer)

PR #3096, opened 2024-01-12
bugfix do not reset shapetracker of 0 size lazybuffer

### LINKED ISSUE #3094: `cat` with interpreted backends
Here is a minimum example to reproduce. And probably is happening on TORCH backend as well. But the weird part is that if you run it twice, it works.
image

chenyuxyz:
it might be coming from an expand, and resetting st results incorrect stride. caught by interpreted backend

fixes #3094


chenyuxyz:
will also check other places that `ShapeTracker.from_shape` are used but might not be contiguous

PR #3098, opened 2024-01-12
multigpu support uneven sharding

g1y5x3:
there are two options for this:
1. leave the last GPU with the evened shards but that means its kernels are going to be different than the rest GPUs
2. pad on sharded axis so each GPU has equal shards which requires internal tracking for the padded dim but then the kernels running each GPU should be the same.

this PR implements 1. 
~~still need to all more matmul test caeses.~~
added corresponding tests for add/matmul/reduce when there are uneven sharding devices, all passed.


g1y5x3:
the type incompatible error from mypy is kinda annoying

g1y5x3:
not sure if this is worth the complexity and unevenly split is just a bad way to shard weights since one GPU is going to run a completely different set of kernels.

PR #3108, opened 2024-01-13
Fix backward for `matmul` with 0 in input shapes

obadakhalili:
The reason `.backward()` wasn't working for loss outputs with input tensors that came from a `matmul` operation with 0 in its input shapes is because the `_reduce` in tensor.py generate entirely new tensors for reduce operations that their original tensor has a 0 in its shape and the requested output shape doesn't have a zero in it, resulting in losing the history in `_ctx` and thus being unable to do backpropagation. After testing, it turned out that this is necessary for reduce operations that use the `mlops.Max` op, but not for ones using `mlops.Sum` op, as the forward logic for the operation can produce the desired result.

I ran the tests and they worked (I stop `test_real_world.py` & `test_whisper.py` after 30 mins as they take forever on my device), and the below dev test compares the new results with pytorch's.

Dev test:
```python
import torch

torch.manual_seed(42)

a = torch.rand((3, 4), requires_grad=True)
b = torch.rand((4, 0), requires_grad=True)

c = a @ b

loss = (c+1).square().mean()

loss.backward()

from tinygrad import Tensor

t_a = Tensor(a.detach().numpy(), requires_grad=True)
t_b = Tensor(b.detach().numpy(), requires_grad=True)

t_c = t_a @ t_b

t_loss = (t_c+1)\
    .square()\
    .mean()

t_loss.backward()

print(b.grad.numpy(), b.grad)
print(t_b.grad.numpy(), t_b.grad)
```

Output:
```
[] tensor([], size=(4, 0))
[] <Tensor <LB CPU (4, 0) contig:True (<LoadOps.CONST: 2>, None)> on CPU with grad None>
```

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -557,14 +557,13 @@ class TestOps(unittest.TestCase):
     helper_test_op([(256,256), (256,256)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-3)
   @unittest.skipIf(IMAGE>0, "no 0 in shape matmul on images")
   def test_gemm_with_zeros_shape(self):
-    # TODO: support backward for this
-    helper_test_op([(8,8), (8,0)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-7, forward_only=True)
-    helper_test_op([(0,8), (8,8)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-7, forward_only=True)
-    helper_test_op([(0,8), (8,0)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-7, forward_only=True)
-    helper_test_op([(8,0), (0,8)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-7, forward_only=True)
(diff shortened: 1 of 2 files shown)


chenyuxyz:
this is not the right fix (see ci failures)

obadakhalili:
Yeah, unfortunately, it didn't occur to me to test with other runtimes until after pushing the PR. The only other runtimes I can test against are `TORCH` and `CLANG`. Hopefully. the error in Clang generalizes to the other runtimes. I will look through the CI failures and try to learn the correct fix. Any hints to the desired fix are appreciated.

obadakhalili:
So after checking the tests and reading your [input](https://github.com/tinygrad/tinygrad/pull/3097#:~:text=would%20need%20to%20change%200%20size%20shortcuts) for the backward solution, I think I'm in the right direction? Except that we want to remove the shortcuts for reduce with input tensors that have 0 in their shapes, not just limit them to `mlops.Max` op, and from there make the realization work for all runtimes, not just interpreted runtimes as is the case right now. I would like to try and do that, but I thought I could loop you in and get your input @chenyuxyz.

PR #3111, opened 2024-01-13
make DType a dataclass

thelonejordan:
- [x] Make `DType` a dataclass

* directly using `DType` object as keys creates issues while subclassing the dataclass, so used the `name` attribute as key instead.
* added support for `PtrDType` to `DType` comparison. (this was not needed for `NamedTuple`).

### DIFF
--- a/tinygrad/dtype.py
+++ b/tinygrad/dtype.py
@@ -9,8 +9,7 @@ class DType:
   itemsize: int
   name: str
   sz: int = 1
-  def __repr__(self):
-    return f"dtypes.{INVERSE_DTYPES_DICT[self.name]}" if self.sz == 1 else f"dtypes._{INVERSE_DTYPES_DICT[self.scalar().name]}{self.sz}"
+  def __repr__(self): return f"dtypes.{'_'*(c:=self.sz!=1)}{INVERSE_DTYPES_DICT[self.name if not c else self.scalar().name]}{str(self.sz)*(c)}"
   def vec(self, sz:int):
     assert sz > 1 and self.sz == 1, f"can't vectorize {self} with size {sz}"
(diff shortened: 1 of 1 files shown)


chenyuxyz:
making DType a dataclass is good, but i don't see the point of making np a property. the goal is to remove .np!

thelonejordan:
> making DType a dataclass is good, but i don't see the point of making np a property. the goal is to remove .np!

@chenyuxyz the goal was to remove numpy incrementally. Removing numpy is a big change in itself https://github.com/tinygrad/tinygrad/pull/3075. I was trying to limit this PR to just converting NamedTuple to dataclass. We can always delete later.

chenyuxyz:
is making .np a property a requirement to make DType a dataclass?

thelonejordan:
> is making .np a property a requirement to make DType a dataclass?

@chenyuxyz It's not. There was a TODO to remove np from attributes. So I removed it and added "np" as a property to support the .np calls. The functionality is exactly the same, the numpy to dtype mapping is handled dynamically, so cleaner.

tinygrad/dtype.py
```diff
@@ -70,9 +73,9 @@ def fields() -> Dict[str, DType]: return DTYPES_DICT
 
   # NOTE: these are image dtypes
   @staticmethod
-  def imageh(shp): return ImageDType(100, 2, "imageh", np.float16, shp, dtypes.float32)
```
thelonejordan, line 73: this change is ambiguous!

thelonejordan: fixed!

thelonejordan:
> is making .np a property a requirement to make DType a dataclass?

@chenyuxyz reverted this.

geohot:
Cool!

PR #3125, opened 2024-01-15
Remove numpy from metal graph

madt2709:
- Grepped and there are no instances of int_buf_view
- Tests pass


geohot:
Ugh.

1) What is the code supposed to do?

2) Did you test that it's still doing it?

chenyuxyz:
i am pretty sure this hangs in a metal device with graph support

PR #3126, opened 2024-01-15
separate try except blocks in onnx2torch in model benchmark

chenyuxyz:
exceptions can be raised from either model conversion or individual backend failed. openpilot on torch mps works, but does not work with torch cpu. seperate the expcetion block so that the benchmark can inlcude torch mps for openpilot.

### DIFF
--- a/test/external/external_model_benchmark.py
+++ b/test/external/external_model_benchmark.py
@@ -8,8 +8,7 @@ import onnxruntime as ort
 from onnx2torch import convert
 from extra.onnx import get_run_onnx
 from tinygrad.helpers import OSX, DEBUG, fetch
-from tinygrad.tensor import Tensor
-from tinygrad import Device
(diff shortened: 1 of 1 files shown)


chenyuxyz:
openpilot on torch_cuda is quite fast!
```
openpilot       tinygrad_cuda_jitless       52.17 ms
openpilot       tinygrad_cuda_jit            3.77 ms
openpilot       tinygrad_clang_jitless     173.74 ms
openpilot       tinygrad_clang_jit         120.18 ms
openpilot       torch_cpu              RuntimeError
openpilot       torch_cuda                   3.30 ms
openpilot       onnxruntime_cpu             37.93 ms
openpilot       onnxruntime_cuda             6.33 ms
openpilot       outputs validated with rtol=2.0e-03, atol=2.0e-03
```

PR #3127, opened 2024-01-15
update env_vars.md

chenyuxyz:
mostly removed deprecated ones. not clear how to maintain this especially for extra/examples


wozeparrot:
might be better just to only maintain it for core tinygrad, and just have examples document their own env vars.

PR #3128, opened 2024-01-15
Pad then shard

g1y5x3:
another attempt for uneven sharding after #3098, allow the multitensor api to automatically right pad the sharded axis and adjust the calculations internally, add a `self.pads` to track


geohot:
Ending up going with the uneven sharding approach, it's way simpler to reason about, even if it might be slower in some cases

g1y5x3:
Cool!

PR #3129, opened 2024-01-15
remove numpy from metal graph

nimlgen:


chenyuxyz:
tested with manual test and ci benchmark runner and both looked good. thanks!

PR #3144, opened 2024-01-16
[WIP] RFC: Introducing Hexagon runtime support for tinygrad

Santhoshkumar-p:
**Overview**
I have implemented basic support for tinygrad on Qualcomm's Hexagon DSP. 
I would like to ask some feedback and development plan for DSP hardware integration before upstreaming the codegen and other low level implementations. Attached file is the idea of the design.

**Hexagon DSP Gotchas:**
- The Hexagon DSP is a 32-bit mobile processor with its own ISA (VLIW), primarily designed for Audio, Video, and Image 
processing.
- Functioning as a general-purpose CPU, it is optimized for handling compute tasks with a low clock frequency, 
addressing power concerns for various workloads. 
- Communication between the CPU and DSP is facilitated through FastRPC.
- Qualcomm provides Hexagon SDK that contains toolchain for compiling and creating stub (CPU side) and skel (DSP side) files (written in C/C++).
- Using native interface we can call stub function which will establish a remote session on DSP side. 
- We can use this bridge to offload all the compute kernels.

**Target Platform:** The intended platform for Hexagon DSP is predominantly Android. 

**Implementation Design:**

- The current approach involves using Hexagon in a somewhat limited manner, we are taking the GPU compute kernel path.
- The Tinygrad host runs on the CPU, with all compute kernels offloaded to the DSP. 
- It is very much possible to offload an entire graph on the DSP.
- FastRPC is the framework used for runtime state communication between the CPU and DSP.
- For code generation, we rely on Clang/LLVM, and testing has been done for both, working fine.
- We can very much achieve SIMD capabilities for tensor ops using HVX instruction sets out of the box. 

**Clarifications Required**:

1. Should we consider running Tinygrad on Android and utilizing the Hexagon SDK toolchain for the compilation of all JIT functions? 
   This would involve creating shared objects (.so) on the client side for each JIT-compiled function, coupled with Hexagon SDK toolchain, similar to the implementation in Tinygrad using Clang.

2. Offloading Strategy: Is it advisable to offload all operation implementations (llOps, mlOps, hlops) to the DSP side, utilizing a JSON or map-like structure for communication of graph construction from CPU? 
   This approach would eliminate Tinygrad JIT compilation, and instead, we would leverage Tinygrad's AST for the serialization of 
 LazyOps that will be executed on DSP.

Given that tinygrad resides at the intersection of both a user-friendly deep learning framework and a powerful compiler.
I believe the above design choice is pivotal for continued progress. Please help me clarify this point and provide some more details on this development plan.

Thanks!


geohot:
The goal of the bounty is to replace SNPE in openpilot. It has to support it natively, it can't have a bridge.

Santhoshkumar-p:
Thanks George! for the clarification, I'll move all the ops impl to DSP side and will upstream the testbench soon.

Santhoshkumar-p:
Hi, 

A quick follow up question on this development plan. 

Could you please tell me which of these methods is preferred.


**Option 1:**

1. Compile ONNX model with Tinygrad.
2. Generate an AOT compiled executable for full inference graph using Tinygrad Hexagon codegen.
3. Create and distribute resulting _.so_ file to Android DSP path.
4. Implement/Reuse CPU-based client program for streaming input images and output.(FastRPC invocation for i/p and o/p only)

**Option 2:**

- Same steps as the above except inference calls will be made to the DSP's QuRT OS driver directly from Android (Without RPC).
- But it was reported as exploiting the security vulnerability in the driver. 
    I'm just wondering if this is okay to do. Like _freethedsp.c_ way : )

**Option 3:** 

- Full DSP native runtime, featuring Buffer, OPS implementation, and Graph construction, will run entirely on the DSP. 
- Model location and input streaming will be facilitated through FastRPC/shared buffer communication.
- We can use Hexagon specific Intrinsics on-the-go for performance boost.

Thanks!

Santhoshkumar-p:
Yes makes sense. Will raise another PR with impl later.

PR #3148, opened 2024-01-16
compile cache for several devices

### LINKED ISSUE #2588: Stale OpenCL program cache after system update

nimlgen:
Closes #2588

### DIFF
--- a/test/test_hip_rdna3.py
+++ b/test/test_hip_rdna3.py
@@ -51,7 +51,7 @@ def compile_ast_to_hip(out: Tensor):
   lin.linearize()
   code = HIPRenderer(to_function_name(lin.name), lin.uops)[0]
   if DEBUG >= 4: print(code)
-  compile_hip("gfx1100", code)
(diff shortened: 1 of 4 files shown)


geohot:
I don't want HIP and CUDA to take in the device object. I'm okay with them taking in an optional string to specify the arch, but I really like that (unlike OpenCL and Metal) the compilers can run without accessing a real GPU

geohot:
Sweet!

PR #3171, opened 2024-01-18
Changing deepwalk function removes the need to reverse list on backward

y0n1n1:
This pull request is a very minor change to tensor.py

In the function deepwalk, the line "yield node" is always executed after all parent nodes have been exhausted. This makes the list returned by the function in the order of most parent nodes to most recent nodes. 

Because of this, the backward function has to call the reversed list returned by deepwalk

if the "yield node" line were to be executed before any parent nodes are checked, the list produced by deepwalk would naturally come in the reversed order that the backward function needs it to be, with no apparent differences in any other aspect of the function.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -276,9 +276,9 @@ class Tensor:
     def _deepwalk(node, visited):
       visited.add(node)
       if getattr(node, "_ctx", None):
+        yield node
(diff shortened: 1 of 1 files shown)


chenyuxyz:
"with no apparent differences in any other aspect of the function" and breaks everything

PR #3178, opened 2024-01-19
fuzzer validates outputs + (partially) oob accesses

nimlgen:
Fuzzer now generates random inputs to validate that output matches (before all rawbufs were 0s, so no matter what in most of cases output was 0 as well). Add some additional area for output rawbuf to check for oob writes.


chenyuxyz:
cool!

PR #3180, opened 2024-01-19
fix padto when with late reduce

nimlgen:
Caught this test with the updated fuzzer. group + padto generates wrong kernels, since if condition for write is wrong there. We ended up with wrong local_idxs after group_for_reduce, some of which are used in the late reduce loop. I think after the late reduce loop we should just zero all local indices which are used for group_for_reduce.

~Note that without #3178 the test always passes~


chenyuxyz:
thanks! i think it also fixed the gpt2 issue (tested on 3080), i will enable the action back

PR #3194, opened 2024-01-21
fix maximum/where Scalar casting

### LINKED ISSUE #3186: maximum autocasts float16 to float32
tiny
```
In [3]: Tensor([1.,2.], dtype=dtypes.half).maximum(3.1).dtype
Out[3]: dtypes.float
```

jax
```
In [6]: jnp.maximum(jnp.array([1., 2.], dtype=jnp.float16), 3.1).dtype
Out[6]: dtype('float16')
```

found this from onnx openpilot implicitly casting float16 to float32 from a Clip op :D

geohotstan:
resolves #3186 

the problem for #3186 is a `where` scalar casting problem it appears

I cross checked jax's casting behavior for `where` and `maximum` with the tests I wrote in `test_dtype` and they do match.

but the changes in `tensor.py` to fix `where` is really ugly :s   

is there a better way to do it? :s  
I played with a few other ways like changing `_broadcast` to something like `_broadcastall(multiple_tensors)` but it just added even more unneeded complexity

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -478,18 +478,9 @@ class TestAutoCastType(unittest.TestCase):
   @given(strat.sampled_from(core_dtypes))
   def test_where_scalar(self, dt):
     t = Tensor(2, dtype=dt)
-    if dtypes.is_float(dt):
-      self.check_where_alternate_input_other(t, 3.2, dt)
-      self.check_where_alternate_input_other(t, 3, dt)
-      self.check_where_alternate_input_other(t, True, dt)
-    elif dtypes.is_int(dt):
-      self.check_where_alternate_input_other(t, 3.2, dtypes.default_float)
-      self.check_where_alternate_input_other(t, 3, dt)
-      self.check_where_alternate_input_other(t, True, dt)
-    else:
-      self.check_where_alternate_input_other(t, 3.2, dtypes.default_float)
-      self.check_where_alternate_input_other(t, 3, dtypes.default_int)
-      self.check_where_alternate_input_other(t, True, dt)
+    self.check_where_alternate_input_other(t, 3.2, (dt if dt >= dtypes.float16 else dtypes.default_float))
+    self.check_where_alternate_input_other(t, 3, (dt if dt >= dtypes.int8 else dtypes.default_int))
+    self.check_where_alternate_input_other(t, True, dt)
 
   def test_where_two_scalars(self):
     self.check_where_alternate_input_other(3.1, 3.2, dtypes.default_float)
@@ -503,15 +494,12 @@ class TestAutoCastType(unittest.TestCase):
   def test_maximum(self, dt1, dt2):
     Tensor([0, 1, 2], dtype=dt1).maximum(Tensor([2, 0, 5], dtype=dt2)).dtype == least_upper_dtype(dt1, dt2)
 
-  def test_maximum_const(self):
-    assert Tensor([1, 2], dtype=dtypes.float16).maximum(3.1).dtype == dtypes.float16
-    assert Tensor([1, 2], dtype=dtypes.float64).maximum(3.1).dtype == dtypes.float64
-    assert Tensor([1, 2], dtype=dtypes.float16).maximum(3).dtype == dtypes.float16
-    assert Tensor([1, 2], dtype=dtypes.float64).maximum(3).dtype == dtypes.float64
-    assert Tensor([1, 2], dtype=dtypes.int64).maximum(3.1).dtype == dtypes.default_float
-    assert Tensor([1, 2], dtype=dtypes.int16).maximum(3.1).dtype == dtypes.default_float
-    assert Tensor([1, 2], dtype=dtypes.int64).maximum(3).dtype == dtypes.int64
-    assert Tensor([1, 2], dtype=dtypes.int16).maximum(3).dtype == dtypes.int16
+  @given(strat.sampled_from(core_dtypes))
+  def test_maximum_const(self, dt):
+    assert Tensor([1, 2], dtype=dt).maximum(3.1).dtype == (dt if dt >= dtypes.float16 else dtypes.default_float)
+    assert Tensor([1, 2], dtype=dt).maximum(3).dtype == (dt if dt >= dtypes.int8 else dtypes.default_int)
+    # TODO :D
+    # assert Tensor([1, 2], dtype=dt).maximum(True).dtype == dt
 
 if __name__ == '__main__':
   unittest.main()
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -54,10 +54,6 @@ def _test_cast(a:Tensor, target_dtype:DType):
 def _test_bitcast(a:Tensor, target_dtype:DType, target=None):
   _test_op(lambda: a.bitcast(target_dtype), target_dtype, target or a.numpy().view(target_dtype.np).tolist())
 
-def where_alternate_input_other(input_, other, data_type):
-  assert (Tensor([True, False]).where(input_, other)).dtype == data_type
-  assert (Tensor([True, False]).where(other, input_)).dtype == data_type
-
 class TestDType(unittest.TestCase):
   DTYPE: Any = None
   DATA: Any = None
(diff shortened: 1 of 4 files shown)


geohotstan:
total line count: 5000
XDDDDDDD

darn.

test/test_dtype.py
```diff
@@ -466,5 +470,48 @@ def test_cumsum(self):
...
+      where_alternate_input_other(t, 3.2, dtypes.default_float)
+      where_alternate_input_other(t, 3, dt)
+      where_alternate_input_other(t, True, dt)
+    else:
+      where_alternate_input_other(t, 3.2, dtypes.default_float)
+      where_alternate_input_other(t, 3, dtypes.default_int)
+      where_alternate_input_other(t, True, dt)
+
+
+  def where_where_two_consts(self):
```
chenyuxyz, line 494: `two_scalars`

test/test_dtype.py
```diff
@@ -54,6 +54,10 @@ def _test_cast(a:Tensor, target_dtype:DType):
 def _test_bitcast(a:Tensor, target_dtype:DType, target=None):
   _test_op(lambda: a.bitcast(target_dtype), target_dtype, target or a.numpy().view(target_dtype.np).tolist())
 
+def where_alternate_input_other(input_, other, data_type):
```
chenyuxyz, line 57: can you put this closer to where it's being used?

test/test_dtype.py
```diff
@@ -466,5 +470,48 @@ def test_cumsum(self):
...
+    where_alternate_input_other(3.1, 3.2, dtypes.default_float)
+    where_alternate_input_other(3.1, 3, dtypes.default_float)
+    where_alternate_input_other(3.1, True, dtypes.default_float)
+    where_alternate_input_other(3, 2, dtypes.default_int)
+    where_alternate_input_other(3, True, dtypes.default_int)
+    where_alternate_input_other(False, True, dtypes.bool)
+
+  @given(strat.sampled_from(core_dtypes), strat.sampled_from(core_dtypes))
+  def test_maximum(self, dt1, dt2):
+    Tensor([0, 1, 2], dtype=dt1).maximum(Tensor([2, 0, 5], dtype=dt2)).dtype == least_upper_dtype(dt1, dt2)
```
chenyuxyz, line 504: does this fix every case of maximum dtype change? if so can you remove the comment above maximum in tensor.py

tinygrad/tensor.py
```diff
@@ -842,6 +842,8 @@ def maximum(self, x:Union[Tensor, Scalar]) -> Tensor: return (self<x).detach().w
   def minimum(self, x:Union[Tensor, Scalar]) -> Tensor: return -((-self).maximum(-x))
 
   def where(self:Tensor, input_:Union[Tensor, Scalar], other:Union[Tensor, Scalar]):
+    if isinstance(input_, Tensor): input_, other = input_._broadcasted(other)
```
chenyuxyz, line 845: i think this is fine. for (Tensor, Tensor) we call broadcasted twice (one in return) on them but i don't quite see a better way to write it

test/test_dtype.py
geohotstan: YIKES!
doesn't even have a test_
lmfao how did I manage to make this typo

tinygrad/lazy.py
```diff
@@ -105,7 +105,6 @@ def e(self, op:Union[LoadOps, UnaryOps, BinaryOps, TernaryOps], *in_srcs:LazyBuf
         srcs.append(s)
     assert all_same(dts:=[x.dtype.scalar() for x in (srcs[1:] if op is TernaryOps.WHERE else srcs)]), f"all dtypes must match {dts} on {op}"
     assert all_same([x.shape for x in srcs]), f"all shapes must be the same {[x.shape for x in srcs]}"
-    if op is TernaryOps.WHERE: assert srcs[0].dtype == dtypes.bool, "TernaryOps.WHERE must have the first arg be bool"
```
geohotstan, line 108: I think this can be removed cuz 
`return mlops.Where.apply(x.cast(dtypes.bool), *y._broadcasted(z))` already pins `x` to `bool`

I went down git blame and I think this was added to cover for WEBGPU hack
and WEBGPU is rip

chenyuxyz: let's keep this. it completes the assert check two lines above

geohotstan:
`Tensor([True, False]).maximum(True).dtype == dtypes.int`
not too sure what to do about this

rest is correct tho

also simplified test_broadcast with scalars.
The test tests for correct casting behavior for tensor <-> scalar implicit casting that matches jax.

test/test_dtype.py
```diff
@@ -397,40 +397,12 @@ def test_int_to_float_unary_func(self, dtype):
...
-    assert (Tensor([0, 1], dtype=dtypes.int) + True).dtype == dtypes.int32
-    assert (Tensor([0, 1], dtype=dtypes.int8) + True).dtype == dtypes.int8
-    assert (Tensor([0, 1], dtype=dtypes.uint64) + True).dtype == dtypes.uint64
-    assert (Tensor([0, 1], dtype=dtypes.float16) + True).dtype == dtypes.float16
-    assert (Tensor([0, 1], dtype=dtypes.bfloat16) + True).dtype == dtypes.bfloat16
-    assert (Tensor([0, 1], dtype=dtypes.float32) + True).dtype == dtypes.float32
-    assert (Tensor([0, 1], dtype=dtypes.float64) + True).dtype == dtypes.float64
+  @given(strat.sampled_from(core_dtypes))
+  def test_broadcast_scalar(self, dt):
+    assert(Tensor.rand(4, 4, dtype=dt) + 2.3).dtype == (dt if dt >= dtypes.float16 else dtypes.default_float)
```
chenyuxyz, line 402: can you change the conditions to something that does not depend on the implementation detail of dtypes ordering? combining `dtypes.is_float()` and `dtypes.is_int()` should give you the desired conditions

chenyuxyz (commented):
cool! for the boolean / 2, can you add a expected failure test case for that and a TODO note?

geohotstan:
maximum and minimum both work fine for all dtype now

some notes:
- thought about the `/2` in maximum for awhile but didn't find anything better. But we only need `/2` for `backward` so we check for `requires_grad` and `float` (should requires_grad imply float? dunno). This check also allows correct implicit dtype promo since `non-floats` avoid the /2 since no need for backward anyways
- having this check in maximum also enables #3195 to not care about maximum.
- changed minimum cuz no need for extra neg ops when we can just flip `>` -> `<`

tinygrad/tensor.py
```diff
@@ -835,11 +835,14 @@ def pow(self, x:Union[Tensor, Scalar], reverse=False) -> Tensor:
     return ar.mul(sign * base_sign + (1 - base_sign)).mul(inject_nan)
   def xor(self, x:Tensor, reverse=False) -> Tensor: return mlops.Xor.apply(*self._broadcasted(x, reverse))
 
-  # TODO: this implicitly changes dtype with /2
-  def maximum(self, x:Union[Tensor, Scalar]) -> Tensor: return (self<x).detach().where(x, (self>x).detach().where(self, (self+x)/2))
-  def minimum(self, x:Union[Tensor, Scalar]) -> Tensor: return -((-self).maximum(-x))
+  def maximum(self, x:Union[Tensor, Scalar]) -> Tensor:
+    return (self<x).detach().where(x, (self>x).detach().where(self, (self+x)/2) if self.requires_grad and self.is_floating_point() else self)
```
chenyuxyz, line 839: use both `requires_grad` and `is_floating_point` do not make sense because `bool` and `int` should not have `requires_grad` and we should check that somewhere

i don't like branching based on implementation details though and it's fine to just cast at the end. with 2 wheres it's already pretty complicated

let's also keep minimum to reuse maximum, unless the removed neg makes it significantly faster. code reuse is less likely to introduce new bug in the future and easier to maintain. the neg should be constant fused so i would be surprised if the speed is very different

geohotstan: >  unless the removed neg makes it significantly faster.

it's barely faster. Unnoticable.
I tested it before and I was wondering why :D
You're right. Maintainability! 

> i don't like branching based on implementation details though and it's fine to just cast at the end. with 2 wheres it's already pretty complicated

makes sense. My initial idea was that the second `.where` and the `div` is only there just to apply the backward pass, and we only backward pass when `self` is `requires_grad`. (`self` has to be `div(2)` where `self == x` for backward)

test/test_dtype.py
```diff
@@ -397,40 +397,12 @@ def test_int_to_float_unary_func(self, dtype):
...
-    assert (Tensor([0, 1], dtype=dtypes.int) + True).dtype == dtypes.int32
-    assert (Tensor([0, 1], dtype=dtypes.int8) + True).dtype == dtypes.int8
-    assert (Tensor([0, 1], dtype=dtypes.uint64) + True).dtype == dtypes.uint64
-    assert (Tensor([0, 1], dtype=dtypes.float16) + True).dtype == dtypes.float16
-    assert (Tensor([0, 1], dtype=dtypes.bfloat16) + True).dtype == dtypes.bfloat16
-    assert (Tensor([0, 1], dtype=dtypes.float32) + True).dtype == dtypes.float32
-    assert (Tensor([0, 1], dtype=dtypes.float64) + True).dtype == dtypes.float64
+  @given(strat.sampled_from(core_dtypes))
+  def test_broadcast_scalar(self, dt):
+    assert(Tensor.rand(4, 4, dtype=dt) + 2.3).dtype == (dt if dtypes.is_float(dt) else dtypes.default_float)
```
chenyuxyz, line 402: `assert (`

PR #3195, opened 2024-01-21
Division of two integer types to return float result

### LINKED ISSUE #3133: division doesn't implicitly upcast dtype
Behaviour in tinygrad:
```python
from tinygrad import Tensor
print((Tensor([1]) / Tensor([2])).numpy())
# array([0], dtype=int32)
```
Behaviour in torch:
```python
import torch
out = torch.tensor([1]) / torch.tensor([2])
print(out, out.dtype)
# tensor([0.5000]) torch.float32
```
is this intented?

torrhen:
The code is aimed at changing the default behaviour of the `__truediv__` operator for tensors when both arguments are integers, so that a float type is returned instead of an integer - resolves #3133.

Before:

`from tinygrad import Tensor`
`(Tensor([1]) / Tensor([2])).numpy()`
`# array([0], dtype=int32)`

`import torch`
`out = torch.tensor([1]) / torch.tensor([2])`
`print(out, out.dtype)`
`# tensor([0.5000]) torch.float32`

After:

`from tinygrad import Tensor`
`(Tensor([1]) / Tensor([2])).numpy()`
`# array([0.5], dtype=float32)`

`import torch`
`out = torch.tensor([1]) / torch.tensor([2])`
`print(out, out.dtype)`
`# tensor([0.5000]) torch.float32`

The new result is more intuitive and more aligned with torch when using normal division. It makes more sense for a additional `__floordiv__` operator to explicitly return an integer type after division when needed.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -312,7 +312,7 @@ class TestOps(unittest.TestCase):
     helper_test_op([(), ()], lambda x,y: x/y)
     helper_test_op(None, lambda x,y: x/y, Tensor.div, forward_only=True, vals=np.array([[5],[1]], dtype=np.int32))
   def test_div_int(self):
-    helper_test_op(None, lambda x: (x/2).to(torch.int), lambda x: x/2, forward_only=True, vals=np.array([[3]], dtype=np.int32))
+    helper_test_op(None, lambda x: x/2, lambda x: x/2, forward_only=True, vals=np.array([[3]], dtype=np.int32))
   def test_div_const(self):
     helper_test_op([(45,65)], lambda x: x/255)
     helper_test_op([(45,65)], lambda x: x/1)
--- a/tinygrad/mlops.py
+++ b/tinygrad/mlops.py
@@ -1,7 +1,7 @@
 import math
 from typing import Tuple, Optional
 from tinygrad.helpers import argsort
-from tinygrad.dtype import DType
+from tinygrad.dtype import DType, dtypes
 from tinygrad.ops import UnaryOps, BinaryOps, TernaryOps, ReduceOps
 from tinygrad.tensor import Function
 from tinygrad.lazy import LazyBuffer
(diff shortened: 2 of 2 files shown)


tinygrad/mlops.py
```diff
@@ -121,6 +121,7 @@ def backward(self, grad_output:LazyBuffer) -> Tuple[Optional[LazyBuffer], Option
 class Div(Function):
   def forward(self, x:LazyBuffer, y:LazyBuffer) -> LazyBuffer:
     self.x, self.y = x, y
+    if dtypes.is_int(x.dtype) and dtypes.is_int(y.dtype): self.x, self.y = ((x := x.cast(dtypes.float32)), (y := y.cast(dtypes.float32)))
```
geohotstan, line 124: Maybe having the dtypes check in `tensor.py` would work?

test/test_ops.py
```diff
@@ -312,7 +312,7 @@ def test_div(self):
     helper_test_op([(), ()], lambda x,y: x/y)
     helper_test_op(None, lambda x,y: x/y, Tensor.div, forward_only=True, vals=np.array([[5],[1]], dtype=np.int32))
   def test_div_int(self):
-    helper_test_op(None, lambda x: (x/2).to(torch.int), lambda x: x/2, forward_only=True, vals=np.array([[3]], dtype=np.int32))
+    helper_test_op(None, lambda x: x/2, lambda x: x/2, forward_only=True, vals=np.array([[3]], dtype=np.int32))
```
geohotstan, line 315: checkout #1644
reverting the change in #1644 also makes this test pass
maybe it's relevant? i dunno :D
me noob

torrhen: https://github.com/tinygrad/tinygrad/blob/640e5c36ad839c38159bccd23a23ba5b4c2d8efb/tinygrad/tensor.py#L818

torrhen: Thanks for that, looks promising, specifically the conditions in the line below need modifying as it also allows division of `None` currently.

https://github.com/tinygrad/tinygrad/blob/640e5c36ad839c38159bccd23a23ba5b4c2d8efb/tinygrad/tensor.py#L818

Revisiting #1644 to modify the way `maximum()` works for int tensors would negate the need for `dtypes.is_float(self.dtype)` and potentially the need for any casting of int arguments passed to `div`.

Is there any reason why `(self + x) / 2` is used:

https://github.com/tinygrad/tinygrad/blob/640e5c36ad839c38159bccd23a23ba5b4c2d8efb/tinygrad/tensor.py#L841

Surely at the point it is confirmed that neither `self < x` or `self > x` the fill value can be either `self` or `x`?

geohotstan: > Is there any reason why (self + x) / 2 is used


I think it's to get the backwards working?

geohotstan: ok it looks like `(self + x) / 2` is used to treat cases where self == x, for `forwards` and `backwards`

geohot:
I'm fine with this if it matches torch. How's it coming?

geohot:
No update in a few days, feel free to reopen.

PR #3197, opened 2024-01-21
tweak tensor core heuristic

chenyuxyz:


chenyuxyz:
meh it's faster for mac but slower for tinybox. really need tensor core in search

PR #3198, opened 2024-01-22
cleanup noop prefixes in _pool

chenyuxyz:
make expand dim=None as noop (in addition to -1). then slice, reshape, expand in _pool can share the same noop prefix


chenyuxyz:
- replaced the long `reshape -> expand -> reshape` with `repeat`
- i think the slices can be replaced by shrinks after repeat, will do in a separate pr to see if we can remove all use of slice

PR #3203, opened 2024-01-22
NVIDIA Tensor Core support in CUDA

TheTechromancer:
Thought I'd take a stab at this one.

Was hoping it would be as easy as slapping in the right intrinsic for `wmma::mma_sync`, but there seem to be differences between how HIP and CUDA organize data inside the accumulator. As an exercise I took the liberty of creating a fully manual CUDA matmul (not using any cuda data structures or helpers except the intrinsic `__hmma_m16n16k16_mma_f32f32`).

The result is that we now understand the memory layout of CUDA's (closed source) fragments, which hopefully will enable me to adapt it into the current HIP TC system.

Needless to say this is WIP; I am new to tinygrad and CUDA programming so this has been a learning experience.


geohot:
How is this coming? Going to close because no update in > 1 week, but feel free to reopen.

TheTechromancer:
Fine with closing this as @chaosagent is doing better work on it.

Is the [manual matmul](https://github.com/TheTechromancer/tinygrad/blob/master/extra/gemm/cuda_matmul_manual.py) useful? If so, I can submit it as a new PR.

PR #3205, opened 2024-01-22
 support same uidx in multiple shape positions 

chaosagent:
continuing with #3200. the use case is for cuda tc, we need the uidx to index into two dimensions of the input buffer.

the "real" fix for this is to do the shape ops in the input buffers to the wmma, but that will be a very complicated change.


chenyuxyz:
i think it would be easier to review this if you have a pr with working cuda tc

chaosagent:
i'm getting there... there will be a few refactors before that, if you don't mind them stacking up

chaosagent:
even without cuda tc, these changes result in -10 lines: https://github.com/tinygrad/tinygrad/pull/3210

chaosagent:
OK, CUDA tc works: https://github.com/tinygrad/tinygrad/pull/3213/files

chaosagent:
There are 2 changes in here:

1. Take only the first occurrence of an upcast dim as an upcast loop index. All but 2 changes in this commit can be replaced by a +2 line change that just does this dedup, but the current patch is less lines, I guess

2. Only do float4 if the index is unit stride (check this by just checking that the index is just the raw uidx instead of a funny tensor core expression). These are the +2 lines.

PR #3207, opened 2024-01-22
Cast final result in `.pow` hlop to be the least upper dtype from input dtypes

obadakhalili:
While comparing pytorch's `torch.pow` hlop to tinygrad's, I noticed the result `dtype` of tinygrad is always float. This is due to the series of operations used to implement tinygrad's `Tensor.pow` method which contains intermediary results of type float. To solve this and match pytorch's behavior, I added a casting operation to the least upper bound data type of the inputs' data types.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -364,18 +364,6 @@ class TestOps(unittest.TestCase):
     helper_test_op([(45,65)], lambda x: 2.0**x, lambda x: 2.0**x)
     helper_test_op([()], lambda x: x**2.0, lambda x: x**2.0)
     helper_test_op([()], lambda x: 2.0**x, lambda x: 2.0**x)
-  def test_pow_dtype_cast(self):
-    x = Tensor.arange(-10, 10)
-    y = Tensor.randint((20,))
-    int_x = x.cast(dtypes.default_int)
-    float_x = x.cast(dtypes.default_float)
-    int_y = y.cast(dtypes.default_int)
-    float_y = y.cast(dtypes.default_float)
-
-    self.assertEqual((int_x ** int_y).dtype, dtypes.default_int)
-    self.assertEqual((int_x ** float_y).dtype, dtypes.default_float)
-    self.assertEqual((float_x ** int_y).dtype, dtypes.default_float)
-    self.assertEqual((float_x ** float_y).dtype, dtypes.default_float)
 
   def test_sqrt(self):
     helper_test_op([(45,65)], lambda x: x.sqrt(), a=0)
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -364,6 +364,18 @@ class TestOps(unittest.TestCase):
     helper_test_op([(45,65)], lambda x: 2.0**x, lambda x: 2.0**x)
     helper_test_op([()], lambda x: x**2.0, lambda x: x**2.0)
     helper_test_op([()], lambda x: 2.0**x, lambda x: 2.0**x)
+  def test_pow_dtype_cast(self):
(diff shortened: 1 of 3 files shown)


chenyuxyz:
behavior change needs regression test

obadakhalili:
Will add it!

test/test_ops.py
```diff
@@ -364,6 +364,18 @@ def test_pow_const(self):
     helper_test_op([(45,65)], lambda x: 2.0**x, lambda x: 2.0**x)
     helper_test_op([()], lambda x: x**2.0, lambda x: x**2.0)
     helper_test_op([()], lambda x: 2.0**x, lambda x: 2.0**x)
+  def test_pow_dtype_cast(self):
```
chenyuxyz, line 367: put this in test/test_dtype.py

obadakhalili: Indeed they felt out of place. Done 👍🏽

tinygrad/tensor.py
```diff
@@ -832,7 +832,7 @@ def pow(self, x:Union[Tensor, Scalar], reverse=False) -> Tensor:
     # inject nan if the base is negative and the power is not an integer
     to_nan = (((x - x.trunc()) * 1e10).abs().clip(0, 1) if isinstance(x, Tensor) else int(bool(x - int(x))) if not reverse else ((self - self.trunc()) * 1e10).abs().clip(0, 1)) * base_sign  # noqa: E501
     inject_nan = ((((-to_nan) * 2) + 1)).log().add(1) if isinstance(to_nan, Tensor) else 1 if not to_nan else float("nan")
-    return ar.mul(sign * base_sign + (1 - base_sign)).mul(inject_nan)
+    return ar.mul(sign * base_sign + (1 - base_sign)).mul(inject_nan).cast(least_upper_dtype(self.dtype, x.dtype) if isinstance(x, Tensor) else self.dtype)  # noqa: E501
```
chenyuxyz, line 835: do you need to also cast in line 824? please double check to make sure it's complete

obadakhalili: Good catch. The way the logic is now casting won't work, as the used formula to calculate the reversed power under the defined conditions doesn't produce accurate results. For example, `5**1` returns 4.9999995, so casting to `int` (in this example) will cause a loss of information. Also, while testing with scalars, I noticed the current code doesn't consider scalars' data types, so I fixed the code and added a regression test.

chenyuxyz: either adding `Tensor.round` or change how `pow` works for int can work right?

i am closing this pr. i have given many feedback and it's still not close to being merged

chenyuxyz:
i think the right fix is to fix `pow` so that it's correct for all cases. just attaching a cast at the end of one branch

tinygrad/tensor.py
obadakhalili: Exactly, it is between supporting `Tensor.round` or fixing `Tensor.pow` logic for scalers when `Reverse=True`.

No problem, I understand. I will be trying to do either of the above and open a new PR when I have a complete solution.

obadakhalili:
@chenyuxyz sorry for taking much of your time on this but I have been thinking about supporting casting for all cases after implementing `Tensor.round` (#3225) and I don't it is possible, at least not for this version of `Tensor.pow`, and I want to share my insight with you. First, using `Tensor.round` coerces some results to be incorrect. Second, and most importantly, you can have the result of `Tensor.pow` to be of a data type that isn't equal to the `least_upper_dtype` of the input tensors, so casting would be wrong.

I hope `Tensor.round` is useful at least 😅

PR #3209, opened 2024-01-22
refactor if checks to be more compact

hristog:
This proposes a refactoring, which simplifies and makes more compact `if-else` expression of type `var if var else` to the equivalent `var or`.

Reference one-liner to obtain all `var if var else` occurrences:
```bash
grep -oPrnH '\b(\w+) if \1\b else' .
```

Additionally, some unused imports have been removed.

### DIFF
--- a/examples/yolov8.py
+++ b/examples/yolov8.py
@@ -200,7 +200,7 @@ def clip_boxes(boxes, shape):
   return boxes
 
 def scale_boxes(img1_shape, boxes, img0_shape, ratio_pad=None):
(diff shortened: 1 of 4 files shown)


chenyuxyz:
we don't merge style only pr unless it's in core tinygrad with a new linter check

hristog:
> we don't merge style only pr unless it's in core tinygrad with a new linter check

Hi @chenyuxyz,

Could you confirm what the "new linter check" entails, please?

Also, isn't `tinygrad/helpers.py` considered core?

Thank you for the info provided!

PR #3219, opened 2024-01-23
tensor.nbytes should return sint

thelonejordan:


chenyuxyz:
this pr is mostly style changes and shuffling lines

PR #3222, opened 2024-01-24
fix neg logical_not inconsistencies

geohotstan:
```
# jax:                                                                             
In [15]: jax.lax.neg(jnp.array(True))
TypeError: neg does not accept dtype bool. Accepted dtypes are subtypes of floating, complexfloating, integer.
                                    
# torch:                                      
In [7]: torch.neg(torch.tensor(True))                                                                                                                                                                                                                                                                       
RuntimeError: Negation, the `-` operator, on a bool tensor is not supported. If you are trying to invert a mask, use the `~` or `logical_not()` operator instead.   
```

it appears that bool tensors don't `neg`
but our `logical_not` depend on neg when Tensor is bool dtype
and when it's not bool dtype, it doesn't match jax or torch
```
# tiny:
In [10]: Tensor([0,1,2,0.5]).logical_not().numpy()
Out[10]: array([ 1. ,  0. , -1. ,  0.5], dtype=float32)

# jax:
In [11]: jax.numpy.logical_not(jnp.array([0,1,2,0.5]))
Out[11]: Array([ True, False, False, False], dtype=bool)
```

what do?
:D

is making `logical_not` just `self == 0` workable?

### DIFF
--- a/tinygrad/runtime/ops_cpu.py
+++ b/tinygrad/runtime/ops_cpu.py
@@ -25,9 +25,8 @@ def as_strided(x, arg):
 
 numpy_fxn_for_op: Dict[Op, Callable] = {
   BufferOps.CONST: lambda val, dtype: np.array(val, dtype=dtype.np),
-  UnaryOps.EXP2: np.exp2, UnaryOps.LOG2: np.log2, UnaryOps.SIN: np.sin, UnaryOps.SQRT: np.sqrt,
+  UnaryOps.EXP2: np.exp2, UnaryOps.LOG2: np.log2, UnaryOps.SIN: np.sin, UnaryOps.SQRT: np.sqrt, UnaryOps.NEG: np.negative,
   UnaryOps.CAST: lambda x,y: x.view(y[0].np) if y[1] else x.astype(y[0].np, copy=False),
-  UnaryOps.NEG: lambda x: np.negative(x),
   BinaryOps.MAX: np.maximum, BinaryOps.CMPLT: np.less, BinaryOps.CMPEQ: np.equal, BinaryOps.ADD: np.add, BinaryOps.SUB: np.subtract,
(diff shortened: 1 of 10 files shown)


tinygrad/runtime/ops_cpu.py
```diff
@@ -27,7 +27,7 @@ def as_strided(x, arg):
   BufferOps.CONST: lambda val, dtype: np.array(val, dtype=dtype.np),
   UnaryOps.EXP2: np.exp2, UnaryOps.LOG2: np.log2, UnaryOps.SIN: np.sin, UnaryOps.SQRT: np.sqrt,
   UnaryOps.CAST: lambda x,y: x.view(y[0].np) if y[1] else x.astype(y[0].np, copy=False),
-  UnaryOps.NEG: lambda x: np.logical_not(x) if x.dtype == np.bool_ else np.negative(x),
+  UnaryOps.NEG: lambda x: np.negative(x),
```
chenyuxyz, line 30: can just be `np.negative` and combine with other lines

chenyuxyz (commented):
cool! yeah i think not calling mlops.Neg for bool is better because that has backward

geohotstan:
didn't add test for the dunder neg cuz 
```
In [6]: -torch.tensor(True)
RuntimeError: Negation, the `-` operator, on a bool tensor is not supported. If you are trying to invert a mask, use the `~` or `logical_not()` operator instead.
```

but we use `-` in some places in tinygrad where tensor could be bool
like `minimum` :D and `argmin`
I think it's cleaner to just have diverging implementation in dundar neg instead of writing a dtype check when `-` should be used

chenyuxyz:
now we have more robust dtype support, we can also add `~` and `+` to tensor, and replace `-` on dtypes.bool with `~`. fine to be a separate pr as this one is pretty big already

PR #3238, opened 2024-01-25
isnan

chaosagent:
it's tested in case of compiler changes breaking it!


chenyuxyz:
i think metal does not work with fastmath enabled, and llvm does not work as #1069

PR #3239, opened 2024-01-25
fix _round in onnx_ops to look more like new Tensor.round

geohotstan:
The onnx round is a little more involved so it can't directly use the new Tensor.round
but I made the onnx round use the new ideas from Tensor.round

for onnx,
it sometimes considers the val of the breakpoint from rounding up or down
instead of 0.5, it could be 0.3

and it sometimes also considers whether it wants to round up or round down when it's equal to the breakpoint
so 1.5 may round to 2 or 1
There is also an option here to round to the nearest even integer, which is really weird...

### DIFF
--- a/extra/onnx_ops.py
+++ b/extra/onnx_ops.py
@@ -400,10 +400,10 @@ def _round(x:Tensor, n:float, equidistant_case = "round_down") -> Tensor:
   if equidistant_case == "round_up": return (x + (1 - n)).floor()
   if equidistant_case == "round_to_even":
     b = x.trunc()
-    b = (b >= 0).where(b+n, b-n)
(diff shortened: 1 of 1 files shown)


extra/onnx_ops.py
```diff
@@ -396,18 +396,17 @@ def GatherElements(x: Tensor, indices: Tensor, axis):
 
 def _round(x:Tensor, n:float, equidistant_case = "round_down") -> Tensor:
```
chenyuxyz, line 397: is there a use case that n is not 0.5?

geohotstan: oh wut, there actually isn't...
I might've forgotten to delete it after changing some _round use case

my bad

PR #3245, opened 2024-01-25
add _reduce_ function for Tensor

lukasl-dev:
## Purpose

Enables pickle serialisation for tools like [ray](https://ray.io/).

### Code

```py
@ray.remote
def add(x: tiny.Tensor, y: tiny.Tensor) -> tiny.Tensor: return x + y

ref = add.remote(
    tiny.Tensor([0, 1, 2, 3, 4]),
    tiny.Tensor([1, 1, 1, 1, 1])
)

print(ray.get(ref)) # get the tensor
```

### Previously

Previously, ray was not able to serialise Tensor objects from tinygrad:

```
TypeError: cannot pickle 'weakref.ReferenceType' object

During handling of the above exception, another exception occurred:

IndexError                                Traceback (most recent call last)
Cell In[10], line 1
----> 1 ref = add.remote(
      2     tiny.Tensor([0, 1, 2, 3, 4]),
      3     tiny.Tensor([1, 1, 1, 1, 1])
      4 )
      6 ray.get(ref)
```

### Now

This pull request fixes this issue by implementing the `_reduce_` function in the `Tensor` class:

```py
print(ray.get(ref)) # get the tensor

<Tensor <LB CUDA (5,) contig:True (<LoadOps.COPY: 3>, None)> on CUDA with grad None>
```

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -99,6 +99,8 @@ class Tensor:
   # Python has a non moving GC, so this should be okay
   def __hash__(self): return id(self)
 
+  def __reduce__(self) -> Tuple[Type[Tensor], Tuple[List[int]]]: return self.__class__, (self.data().tolist(),)
+
   @property
   def device(self) -> Union[str, Tuple[str, ...]]: return self.lazydata.device
 
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -99,7 +99,8 @@ class Tensor:
   # Python has a non moving GC, so this should be okay
   def __hash__(self): return id(self)
 
-  def __reduce__(self) -> Tuple[Type, Tuple[List, Optional[Union[str, tuple, list]], Optional[DType], Optional[bool]]]: return self.__class__, (self.data().tolist(), self.device, self.dtype, self.requires_grad)
+  def __reduce__(self) -> Tuple[Type, Tuple[List, Optional[Union[str, tuple, list]], Optional[DType], Optional[bool]]]:
+    return self.__class__, (self.data().tolist(), self.device, self.dtype, self.requires_grad)
 
   @property
   def device(self) -> Union[str, Tuple[str, ...]]: return self.lazydata.device
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -99,7 +99,7 @@ class Tensor:
   # Python has a non moving GC, so this should be okay
   def __hash__(self): return id(self)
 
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -99,6 +99,9 @@ def __repr__(self):
   # Python has a non moving GC, so this should be okay
   def __hash__(self): return id(self)
 
+  def __reduce__(self) -> Tuple[Type, Tuple[List, Optional[Union[str, tuple, list]], Optional[DType], Optional[bool]]]:
+    return self.__class__, (self.data().tolist(), self.device, self.dtype, self.requires_grad)
```
geohot, line 103: Does this handle multi-dimensional shapes? Can you add that to test?

lukasl-dev: I have added `test_pickle_multi_dim` and `test_pickle_scalar` and both pass  :heavy_check_mark:

geohot: Why is dtype optional? Why are you only pickling to CPU?

geohot:
Actually, it probably shouldn't pickle to a list, but rather bytes

tinygrad/tensor.py
lukasl-dev: `DType` was optional since one of the purposes of the `__reduce__` function is to return the arguments needed to reconstruct the `Tensor` object, meaning:

```py
y = Tensor.randn(shape)
x = Tensor(*y.__reduce__()[1])

np.testing.assert_equal(y.numpy(), x.numpy())
assert y.dtype == x.dtype
assert y.device == x.device
assert y.requires_grad == x.requires_grad

# (I have added this as a new test case in `test_tensor.py` - passes)
```

Since `dtype` is optional in `Tensor`'s constructor, `dtype` can be optional in `__reduce__`. However, I removed the type hints in my latest commits because the method signature alone would go over 2 lines with ruff's current max line length.

[Python Docs about reduce](https://docs.python.org/3/library/pickle.html#object.__reduce__)

lukasl-dev:
> It probably shouldn't pickle to a list, but rather bytes.

`pickle.dumps()` does the serialisation to bytes already:

```py
In[9]: type(pickle.dumps(Tensor([1, 2, 3, 4])))
Out[9]: bytes
```

`__reduce__` should only return a tuple of `(class, constructor_args, ...)` according to [Python's pickle docs](https://docs.python.org/3/library/pickle.html#object.__reduce__).

geohot:
tolist is wrong.

PR #3247, opened 2024-01-25
tests for Tensor init data dtype and resulting dtype

### LINKED ISSUE #2797: test_ops test if dtype matches
need to fix `.numpy()` output type conversion. `Tensor([True, False]).numpy()` returns `[1, 0]` now for example. then we can test both the content and the dtype of each ops.

ideally can fuzz with different dtypes and if torch does not support uint16 maybe replace it with jax.numpy

hristog:
* Tests for Tensor init data dtype matching resulting dtype.

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -106,6 +106,16 @@ class TestDType(unittest.TestCase):
     self.assertTrue(all(isinstance(value, DType) for value in fields.values()))
     self.assertTrue(all(issubclass(value.np, np.generic) for value in fields.values() if value.np is not None))
 
+  def test_resulting_and_init_dtypes_match(self):
+    dtypes = list(map(np.dtype, ["bool", "uint8", "int8", "int16", "int32", "int64", "float32", "float64"]))
+    data = [1., 2., 0., 0.5, -1.5, 5.25]
+    for dt in dtypes:
+      arr = np.asarray(data, dtype=dt)
+      tin = Tensor(arr).numpy()
+      tor = torch.as_tensor(arr).detach().numpy()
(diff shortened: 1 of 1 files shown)


geohot:
we have test_dtypes.py, anything that misses?

hristog:
Hi @geohot, thanks for looking into this!

I was going by the title and description of the referenced https://github.com/tinygrad/tinygrad/issues/2797, and I also asked for a bit of clarification there, just so that I could understand the intention better.

You're absolutely right that the `test_dtypes.py` are quite comprehensive. I think they're missing comparison against Torch, in terms of resulting `dtype` upon initialisation from equivalent source data (which is what I've added in this PR).

Which course of action would be preferable - closing this PR or moving the tests to `test_dtype.py`?

hristog:
I've moved the tests to `test_dtype.py` and removed the direct association for closing #2797 based on this PR, because the former does mention as a requirement/suggestion that the ops get tested additionally (which could be done in a separate PR).

> then we can test both the content and the dtype of each ops

chenyuxyz:
existing test `test_to_np` already test Tensor construction from numpy, and compare the resulting output. i don't see the value to compare with torch output (and it still construct the Tensor with numpy array). if torch output differs from numpy, it's a torch bug

hristog:
Hi @chenyuxyz, thank you for your insightful feedback!

I'll have a proper look at the referenced test a bit later, and potentially raise a PR to remove the tests added in this one (and possibly propose an augmentation to the existing ones, if applicable).

I confirm that I'm not a supporter or adding stuff with zero or marginal value, and I wasn't aware of the referenced set of tests beforehand. I simply saw your issue and contributed accordingly.

PR #3250, opened 2024-01-26
wmma: fix the upcasts after WMMA to be hcopt ordering invariant

flammit:
will correctly handle and permutation of optops after the TC one.

This is the first PR in a series to add tensor cores to the optimization space with no wrong actions.


flammit:
Will plan to hold this as a draft until CUDA TC is merged (will change the logic for the number of non-TC upcast dims)

Issue #2469, opened 2023-11-27
Incorrect behavior in symbolic due to poor typing discipline


PaulGustafson:
The following test fails
```
    i = Variable("i", 1, 5)
    idx0 = Variable("idx0", 0, i)
    assert idx0 % 2 != idx0
```

This is because get_bounds() can return Nodes, not just integers, which messes up inequalities in the definition of Node.\_\_mod\_\_.  Similar tests fail for MulNodes:

```  
    a = Variable("a", 1, 2)
    u = Variable("u", 1, 2)
    x = MulNode(a,u)
    assert (x < 2) != 0 
```

There are also hacks like this one in `def size()` in shapetracker.py that should be removed:
```while not isinstance(ret, int): ret = ret.max```

Proposed solution: make symbolic.py type safe.

**Comment by PaulGustafson** [comment]
#2474 rules out the first failing test, but the following variant is harder to rule out since many shapetracker unit tests use a similar code path

```  
def test_nested_variable_mod2(self):
    i = Variable("i", 1, 100)
    idx0 = Variable("idx0", 0, i)
    with self.assertRaises(AssertionError):
      assert idx0 % Variable('j', 2, 3) == idx0
```

ronak69:
> Similar tests fail for MulNodes:

this patch fixes the issue. (buf if shapetracker doesn't do `MulNode * MulNode < int` then it does not matter.

```diff
diff --git a/tinygrad/shape/symbolic.py b/tinygrad/shape/symbolic.py
index a9518f12..627bced8 100644
--- a/tinygrad/shape/symbolic.py
+++ b/tinygrad/shape/symbolic.py
@@ -180,7 +180,7 @@ class LtNode(OpNode):
   def __floordiv__(self, b: Union[Node, int], _=False): return (self.a//b) < (self.b//b)
   def get_bounds(self) -> Tuple[int, int]:
     if isinstance(self.b, int):
-      return (1, 1) if self.a.max < self.b else (0, 0) if self.a.min >= self.b else (0, 1)
+      return (1, 1) if (self.a.max < self.b) == True else (0, 0) if (self.a.min >= self.b) == True else (0, 1)
     return (1, 1) if self.a.max < self.b.min else (0, 0) if self.a.min >= self.b.max else (0, 1)
   def substitute(self, var_vals: Dict[VariableOrNum, Node]) -> Node:
[...]

Issue #1924, opened 2023-09-26
max(symbolic, int) has unexpected behaviour


gijskoning:
The following fails:
```
vi = Variable("i", 1, 10)
assert max(5, vi) == max(vi, 5)    
```
Because  `max(5, vi) == <i[1-10]>`
And `max(vi, 5) == 5`

Sorry I don't have a fix for it at the moment but do want do mention it.

And it looks like we use max on symbolic: https://github.com/tinygrad/tinygrad/blob/9c6bb7ff1344c198de21b23290bc576b3a8c9219/tinygrad/codegen/optimizer.py#L243

geohot:
Hmm cool find.

gijskoning:
I couldn't get my head around it either. It depends a bit on the use case, but I think most of the time you want the real maximum number. You could have Node.max(nodes) which retrieves the real max number for every node and then returns the maximum of all.

ronak69:
`max()` compares its arguments to determine the largest one. That results in `LtNode`s being used in `if` statements (which in turn get converted to `bool`).

```python
x = Variable("x", 1, 10)
assert bool(x < 5) == True
assert bool(x > 5) == True
```

`x < 5` is an instance of `LtNode` which will always get converted to `True` regardless of whether `x` will be less than `5` or not.

---

`max(5, vi)` is equal to `vi` because `vi > 5` is `True`.
`max(vi, 5)` is equal to `5` because `5 > vi` is also `True`.

PR #3273, opened 2024-01-29
take merge views from corsix branch

geohot:
This makes Winograd twice as fast (to compile) on M3

### DIFF
--- a/test/test_winograd.py
+++ b/test/test_winograd.py
@@ -1,6 +1,6 @@
 import unittest
 from tinygrad import Tensor, GlobalCounters
-from tinygrad.helpers import Timing, CI, Profiling, WINO
+from tinygrad.helpers import Timing, CI, Profiling, WINO, DEBUG
 from tinygrad.ops import LoadOps
 from tinygrad.codegen.linearizer import Linearizer
 
@@ -28,6 +28,11 @@ class TestWinograd(unittest.TestCase):
         l = Linearizer(s.ast)
         l.hand_coded_optimizations()
         l.linearize()
+      for st in l.sts:
+        assert len(st.views) <= 2, "too many views in winograd"
+        if DEBUG >= 3:
+          print(len(st.views))
+          for v in st.views: print(v)
 
   def test_profile(self):
     x,w = Tensor.rand(1,4,9,9).realize(), Tensor.rand(4,4,3,3).realize()
--- a/test/test_winograd.py
+++ b/test/test_winograd.py
@@ -28,10 +28,11 @@ class TestWinograd(unittest.TestCase):
         l = Linearizer(s.ast)
         l.hand_coded_optimizations()
         l.linearize()
+      if DEBUG >= 2: print(f"{len(l.sts):4d} shapetrackers")
       for st in l.sts:
         assert len(st.views) <= 2, "too many views in winograd"
         if DEBUG >= 3:
-          print(len(st.views))
+          print(f"{len(st.views):3d} views")
           for v in st.views: print(v)
 
   def test_profile(self):
--- a/test/test_winograd.py
+++ b/test/test_winograd.py
@@ -28,7 +28,7 @@ class TestWinograd(unittest.TestCase):
         l = Linearizer(s.ast)
         l.hand_coded_optimizations()
         l.linearize()
-      if DEBUG >= 2: print(f"{len(l.sts):4d} shapetrackers")
+      if DEBUG >= 2: print(f"{len(l.sts):4d} shapetrackers with max {max(len(x.views) for x in l.sts)} views")
       for st in l.sts:
         assert len(st.views) <= 2, "too many views in winograd"
         if DEBUG >= 3:
--- a/tinygrad/shape/shapetracker.py
+++ b/tinygrad/shape/shapetracker.py
@@ -1,19 +1,86 @@
 # ShapeTracker allows movement operations to a buffer that don't require a copy to be made.
 from __future__ import annotations
-import functools
+import functools, math
 from dataclasses import dataclass
 from typing import Tuple, List, Optional, Dict, Set, cast, Iterable, Union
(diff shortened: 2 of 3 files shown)


geohot:
@corsix I remember you mentioned having a better merge views, I found it

This is the fix I need for Winograd, but it seems like it breaks something else.

geohot:
This should also be able to share a lot of code with reshape I think. The reshape thing is just a specialized version of this where the second view has to be contiguous.

corsix:
FWIW, taking out the `tinygrad/shape/view.py` change seems to put it in a better place (said change was only there because `merge_views` was very occasionally too good in synthetic tests, and that `View.create` change was an easy way to make the competitor equally good).

geohot:
Too good? I didn't really understand that change, is there a bug with/without it?

corsix:
> I didn't really understand that change

There's long-standing code in `View.create` to canonicalise strides: if any dimension has size 1, then its stride can be canonicalised to 0. The change added another case of canonicalisation: if any dimension has size >1, but is masked such that only one index in the dimension is unmasked (e.g. a dimension with range from 0 through 10, but with <4 and >4 all masked off, leaving 4 as the only unmasked index), then its stride can also be set to 0, albeit with a corresponding adjustment required to the offset (e.g. continuing the example, if said dimension had stride X, we'd need to add 4*X to the offset).

> Too good?

The context was a unit test that compared the following:
* A: Start with some shape X, applied a sequence of elementary transforms to it to create Y, then applied another sequence of elementary transforms to create Z.
* B: Start with shapeof(Y), apply the 2<sup>nd</sup> sequence of elementary transforms to that, then do `merge_views` between Y and that.

The expectation was that (B) would create something complicated, then the `merge_views` would simplify it back down, ideally getting as simple as (A). In practice, it sometimes created something _simpler_ than (A), because (B) naturally canonicalised away masked-to-single-element dimensions as a consequence of removing masked elements then canonicalising then re-applying padding. In those cases, to get the results of (A) and (B) to be identical, (A) also needed to canonicalise away masked-to-single-element dimensions, which is the `view.py` change.

> is there a bug with/without it?

It slightly increased the number of cases where (A) gave identical result to (B). At least in the tests I was looking at, in all cases (A) and (B) gave _equivalent_ results, just not _identical_ results. That said, there is clearly some bug with it based on the earlier test failures in this PR.

> Is this merge_views complete?

It isn't sufficient to make the aforementioned unit test always give identical results for (A) and (B); testing showed various cases where either (A) was slightly simpler, or (B) was slightly simpler, or (A) and (B) were equally complex but decided to express the complexity in slightly different ways. Is there some weaker sense in which it is complete? Perhaps.

geohot:
Huh, so the `test_shrink_const_then_cast` is fixed it seems by the canonicalization.

geohot:
I have a different fix for this in #3277

corsix:
Ah, I see you identified the bug as the handling of symbolics, as comparison operators on symbolics end up false-y if the comparison is always false, but end up true-y if the comparison is _potentially_ true. Both new calls to `any` in `view.py` are affected; one fix as you've done is `instanceof` checks (though the 2<sup>nd</sup> `any` also needs them), the other (more subtle) approach is to invert the direction of the condition and wrap it in a not (e.g. `b+1 >= e` -> `not (b+1 < e)`).

geohot:
Ahh yea, the second any probably caused the issue with llama on master.

@chenyuxyz perhaps symbolic should assert on those comparisons?

corsix:
FWIW, this behaviour of symbolics is why the revised `merge_views` has things like `if not (merged_term >= merged_size) and not (merged_term < 0):`, rather than the more obvious way of writing that without `not`s.

ronak69:
@corsix @geohot

> comparison operators on symbolics end up false-y if the comparison is always false, but end up true-y if the comparison is potentially true.

Comparison operation on symbolics will return `True` only if the condition is true for all values of the symbolic between `min` and `max`. Analogously, return `False` if false for all.

The problem occurs when for some values between `min` and `max`, the condition is true and for others it is false. In this case, the comparison operation will return `LtNode` instead of a `bool`; but since you have put the condition in an if-statement, python will try to convert the `LtNode` into a `bool` by calling `LtNode.__bool__()`. But guess what, a `LtNode` will *always* get converted to `True`.

I have a more **readable** solution for this problem: explicitly checking if the comparison returned `True` or it was `LtNode` or `False`. So, you can replace pattern `if x < y` with `if (x < y) is True` and pattern `if not x < y` with `if (x < y) is False`.

see also: https://github.com/tinygrad/tinygrad/issues/2469#issuecomment-1867856487 and https://github.com/tinygrad/tinygrad/issues/1924#issuecomment-1742153891 for original.

To catch these kind of bugs, why not just put an always failing assert in `LtNode.__bool__()`?

The assertions in this patch does not fail when running `test/test_symbolic_ops.py` and `test/test_symbolic_jit.py`.

```diff
diff --git a/tinygrad/shape/view.py b/tinygrad/shape/view.py
index acdd4b9b..3bb0a2f7 100644
--- a/tinygrad/shape/view.py
+++ b/tinygrad/shape/view.py
@@ -85,6 +85,16 @@ class View:
     contiguous = offset == 0 and mask is None and strides == strides_for_shape(shape)
     # if any dimension has size >1, but is masked such that only one index in the dimension is unmasked
     # then its stride can also be set to 0, albeit with a corresponding adjustment required to the offset
+
+    if mask:
+      explicit = [(b+1 >= e) is True for b,e in mask]
+      flip = [not (b+1 < e) for b,e in mask]
+      assert explicit == flip
+
+      explicit = [(b >= e) is True for b,e in mask]
+      flip = [not (b < e) for b,e in mask]
+      assert explicit == flip
+
     if mask and any(elim := [not (b+1 < e) for b,e in mask]):
       if any(not (b < e) for b,e in mask):
         strides, offset, mask = (0,) * len(shape), 0, ((0,0),) * len(shape)
```

PR #3275, opened 2024-01-29
search: add tensor core to beam search space

flammit:
Screenshot 2024-01-29 at 2 29 34 PM

This depends on #3285 being merged first.

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -36,6 +36,15 @@ class TensorCore: # D = A * B + C, A is (M x K), B is (K x N), C and D are (M x 
   def num_threads(self): return len(self.threads)
   def num_upcasts(self): return len(self.thread_local_aliases[0]) - self.num_threads()
 
+class TensorCoreOptions(NamedTuple):
+  bufs: Tuple[int, int] # the local aliased buffers for A and B
+  axes: List[int] # the location of the original N and M axes if still in the shape
+  axes_exist: List[bool] # true if the original N and M axes are still in the shape
+  def fix_axes(self, removed_axis:int): # adjust the TC axes if necesssary when an dimension is removed
+    for tc_dim in [i for i in range(2) if self.axes_exist[i]]:
+      if removed_axis < self.axes[tc_dim]: self.axes[tc_dim] -= 1
+      elif removed_axis == self.axes[tc_dim]: self.axes_exist[tc_dim] = False
+
 tensor_cores: Dict[str, List[TensorCore]] = {
   "METAL": [
     TensorCore(dims=[8,8,8], dtype_in=dtypes.float, dtype_out=dtypes.float, wmma_func="__metal_wmma<float2,simdgroup_float8x8,float2>", threads=[(0,2),(1,4),(0,2),(1,2)], thread_local_sizes=[2,2,2], thread_local_aliases=[ [[4],[0],[2],[0],[-1, 1, 3],[0]], [[0],[3],[0],[1],[2, 4],[-1]], [[4],[3],[2],[1],[0],[-1]] ]), # noqa: E501
@@ -103,7 +112,7 @@ class Kernel:
     self.local_dims: int = 0
     self.local_alias: Dict[int, LocalBuffer] = {}
     self.tensor_core: Optional[TensorCore] = None
-    self.tensor_core_bufs: List[int] = []
+    self.tensor_core_opts: Optional[TensorCoreOptions] = None
     self.dont_use_locals: bool = False
 
     # group simplifies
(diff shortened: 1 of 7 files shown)


tinygrad/codegen/kernel.py
```diff
@@ -390,16 +394,21 @@ def fix(needed, ax):
...
-        alias_pattern = [0]*(self.global_dims+(self.local_dims-len(tc.threads))) + [2]*(len(tc.threads)) + [0]*(self.shape_len-self.upcasted-self.first_reduce) + [1,1] + [3]*(self.upcasted-2)  # noqa: E501
-        self.alias_buffer(buf0, alias_pattern)
-        self.alias_buffer(buf1, alias_pattern)
+        # delay alias_buffer until all optimizations are completed
+        self.tensor_core_bufs = [buf0, buf1]
         return True
     return False
 
-  def apply_opt(self, opt:Opt):
+  def apply_opt(self, opt:Opt, append_opt:bool=True):
```
geohot, line 402: Hmm, so if I construct this with apply_tensor_cores it's never appended? I don't like this as an option, it should always append and be able to reconstruct.

Aka, the Linearizer state should always be able to be recreated with the ast + applied_opts

flammit: Whoops good catch - this wasn't intended.  Will push a fix shortly.

geohot: Can you add an explicit test for this reconstruction?

flammit: Added this test in `test_tensor_cores`

tinygrad/device.py
```diff
@@ -249,12 +249,9 @@ def get_linearizer(self, ast:LazyOp) -> Linearizer:
...
     k.required_optimizations()
     if not NOOPT:
-      if not (used_tensor_cores:=k.apply_tensor_cores(getenv("TC", 1))): k.hand_coded_optimizations()
+      k.hand_coded_optimizations()
       if BEAM >= 1:
-        lins = [(("tc" if used_tensor_cores else "hc"), k)]
-        if used_tensor_cores:
-          lins.append(("hc", Linearizer(ast, self.compiler.linearizer_opts)))
-          lins[-1][1].hand_coded_optimizations()
+        lins = [("hc", k)]
```
geohot, line 254: I don't like that this doesn't try the hand coded and tensor cores separate anymore? Why did this have to change?

tinygrad/codegen/kernel.py
```diff
@@ -470,6 +463,25 @@ def required_optimizations(self):
         self.apply_opt(Opt(OptOps.UPCAST, unit_stride_axes_mul_4[0], 4))
 
   def hand_coded_optimizations(self):
+    try: # check TC first and apply hand-coded opts if successful
```
geohot, line 466: Can this just work like how it used to work? I don't see why this has to be in hand coded.

tinygrad/features/search.py
```diff
@@ -72,7 +73,7 @@ def bufs_from_lin(lin:Linearizer) -> List[Buffer]:
 
 # get dictionary of all possible actions
 def get_linearizer_actions(lin:Linearizer, include_0=True) -> Dict[int, Linearizer]:
-  acted_lins = {0:lin} if include_0 else {}
+  acted_lins, max_up, max_lcl = {0:lin} if include_0 else {}, getenv("BEAM_UP_MAX", 256), getenv("BEAM_LCL_MAX", 256)
```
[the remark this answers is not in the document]
flammit: Will remove from this PR and stage a separate PR for this later

flammit:
@geohot is this closer to what you were expecting?

tinygrad/codegen/kernel.py
```diff
@@ -119,13 +129,14 @@ def copy(self):
...
+    ret.info, ret.reduceop, ret.bufs, ret.earlybufs, ret.full_buf_index = \
+      self.info, self.reduceop, [x for x in self.bufs if not isinstance(x, LocalBuffer)], self.earlybufs, self.full_buf_index
+    ret.sts = self.sts[:len(ret.bufs)] # NOTE: must redo the local buffers with TC in beam
 
     # parameters for optimizations
-    ret.applied_opts, ret.group_for_reduces, ret.upcasted, ret.local_dims, ret.local_alias, ret.tensor_core, ret.dont_use_locals = \
-      self.applied_opts[:], self.group_for_reduces, self.upcasted, self.local_dims, self.local_alias.copy(), self.tensor_core, self.dont_use_locals
+    ret.applied_opts, ret.group_for_reduces, ret.upcasted, ret.local_dims, ret.dont_use_locals = \
+      self.applied_opts[:], self.group_for_reduces, self.upcasted, self.local_dims, self.dont_use_locals
+    ret.tensor_core, ret.tensor_core_opts, ret.local_alias = self.tensor_core, self.tensor_core_opts, {}
```
geohot, line 139: Ehh, why is this not `self.local_alias.copy()`?

flammit: the local_aliases depend on the full shape, so copying/then adding opts make them invalid.  Similar situation with the temp group_for_reduce buffer, which is why we drop out all LocalBuffer a couple lines above.

PR #3281, opened 2024-01-30
add no_grad_mode context manager with decorator support

MaxWolf-01:
Introduces `no_grad_mode` context manager which can be called like `torch.no_grad`:
```python
with Tensor.no_grad_mode(): ...
# OR
@Tensor.no_grad_mode()
def f(): ...
```
Also adds ability to use `Tensor.train()` as a decorator:
```python
@Tensor.train():
def f(): ...
```

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -468,5 +468,50 @@ class TestTensorCreationDevice(unittest.TestCase):
     x = y.one_hot(10)
     x.realize()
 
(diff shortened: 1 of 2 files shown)


geohot:
Does it need the `_mode`?

geohot:
No updates in a bit, going to close. I'm supportive of this if it has the right API

PR #3210, opened 2024-01-22
use optops to specify tc dims

chaosagent:
continuing with #3205. current HIP creates an incorrect pattern and uses hacks to make it work. no behavior change in the first commit, need some way to divide indices before can make the behavior actually correct

Later commits get the right hip pattern, almost ready for cuda TC!

PR #3282, opened 2024-01-30
tc_optops but with no lambda

chaosagent:
#3210 but with no lambda. did not use brain when writing this one.

also symbolic doesn't seem to automatically simplify `x // k * k + x % k` to `x`

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -247,7 +247,7 @@ class Linearizer(Kernel):
       # define accumulator
       acc = self.global_load(0, global_idxs+local_idxs+fake_reduce_idxs+upcast_idxs, self.get_reduce_acc(self.reduceop))
 
-      if (tc:=self.tensor_core):
+      if self.tensor_core:
         def calc_tc_idxs(local_size: int, aliases: List[List[int]]):
           replace_idxs = []
           for alias in aliases:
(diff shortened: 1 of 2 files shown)


geohot:
Looks good, but there's a merge conflict on this.

tinygrad/codegen/linearizer.py
```diff
@@ -259,11 +259,6 @@ def calc_tc_idxs(local_size: int, aliases: List[List[int]]):
                 full_var_sz *= next_var.max+1
             replace_idxs.append(full_var)
           return replace_idxs
-        replace_acc_idxs = calc_tc_idxs(tc.thread_local_sizes[2], tc.thread_local_aliases[2])
```
geohot, line 262: Why is this removed?

geohot:
I'm also starting to think we don't have enough testing around tensor cores to merge anything related to them.

There's a whole bunch of subtle ways this can be wrong? Did you test it on Mac and AMD?

tinygrad/codegen/linearizer.py
chaosagent: before, afaict the only thing that this did was permute the output axes for hip tc to work around upcasts always coming before locals (it was building a 16-length index on a 8-length shape dimension, even). now we specify the output pattern through optops.

maybe i missed something about why this was here before.

chaosagent:
I tested on AMD on simple_matmul and cifar. I don't have a mac to test on, unfortunately. I think we can test AMD tc with remu in CI.

chaosagent:
actually, this one is still relevant and doesn't require any additional tests

geohot:
Closing, stale.

PR #3285, opened 2024-01-31
wmma: protect TC locals from modification and use only LOCAL

flammit:
This is a prereq to simplify the merge of the TC beam PR #3275.

The TC locals definitely can't be upcast, but really there's no need to upcast any locals.  This leads to a simpler search space where only globals goes to local or upcast.

Also, currently there's a bug in simplify_ones with fully unrolling the first_reduce (incorrectly lowers the local_dim count).

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -12,7 +12,7 @@ from dataclasses import dataclass
 from enum import Enum, auto
 
 class OptOps(Enum):
-  UPCAST = auto(); UPCASTMID = auto(); UNROLL = auto(); LOCAL = auto(); LASTLOCAL = auto() # noqa: E702
+  UPCAST = auto(); UPCASTMID = auto(); UNROLL = auto(); LOCAL = auto() # noqa: E702
   GROUP = auto(); GROUPTOP = auto(); NOLOCALS = auto(); PADTO = auto() # noqa: E702
   def __lt__(self, x:OptOps): return self.value < x.value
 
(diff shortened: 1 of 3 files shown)


chaosagent:
>     but really there's no need to upcast any locals
do you have any measurements about this? (eg this doesnt happen for winning beams in the models we have)
upcasting locals makes sense for eg float4 loads, because the beam will greedily fill locals, and you want to take the float4 dim from the "low" end of it

flammit:
> > ```
> > but really there's no need to upcast any locals
> > ```
> 
> do you have any measurements about this? (eg this doesnt happen for winning beams in the models we have) upcasting locals makes sense for eg float4 loads, because the beam will greedily fill locals, and you want to take the float4 dim from the "low" end of it

If there's a way to keep the local to upcast behavior without reintroducing the bug that's covered in the new `[Opt(OptOps.UNROLL, 0, 0)]` test case, happy to accept it.

tinygrad/codegen/kernel.py
```diff
@@ -241,8 +241,6 @@ def simplify_ones(self) -> bool:
     # TODO: this should be factored in to multi shape stride
     if self.shape_len == 0: return False
     all_ones = [s==1 for s in self.full_shape]
-    self.local_dims -= sum(all_ones[self.first_reduce-self.local_dims:self.first_reduce])
```
geohot, line 244: I don't think removing this is right

flammit: This PR makes it so it's one way from globals to either LOCAL or UPCAST (see lines 411/429), so it's not possible.  It might not be the desired behavior though.

As discussed live, shelving this PR and reworking it into more testing and bug fixes before functional changes.

PR #3286, opened 2024-01-31
Improve error message on wrongly set env key

flxwu:
When I tried to run tinygrad locally (M1 Mac), I got `ValueError: invalid literal for int() with base 10: 'arm64-apple-darwin20.0.0-clang'`, it took me a minute to figure out the root cause (CLANG was set by default in my PATH).

This PR improves the error message to help future users figure this out quicker.

I also included a mini bug-fix in `tinygrad/device.py`, it was failing with `local variable 'lib' referenced before assignment` when setting `DISABLE_COMPILER_CACHE=1`.

### DIFF
--- a/test/unit/test_device.py
+++ b/test/unit/test_device.py
@@ -1,19 +1,46 @@
 #!/usr/bin/env python
 import unittest
+from unittest.mock import patch
+import os
 from tinygrad.device import Device
+from tinygrad.helpers import getenv
+
 
 class TestDevice(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


chenyuxyz:
i agree it's important to improve the error message, but the proposed change to catch cast error does not make sense. user can still have pre set env vars with a type that can be cast and failed later. so it's incomplete and adds complexity

also please double check and minimize the pr diff, and avoid mixing multiple fixes in one pr

PR #3287, opened 2024-01-31
wmma: add HIP FP16 to FP16 tensor core

flammit:


test/test_linearizer.py
```diff
@@ -116,7 +116,7 @@ def test_tensor_cores(self):
       k.linearize()
       assert len([uop for uop in k.uops if uop.uop == UOps.WMMA]) == 1, "tensor core not triggered"
       np_c = np_a @ np_b
-      np.testing.assert_allclose(np_c, r.numpy(), atol=5e-3, rtol=1e-4)
```
chenyuxyz, line 119: condition this based on dtype_out and keep the test the same for float acc

flammit:
Updated to keep the same tolerances for float

flammit:
I think so because the float16 doesn't have much space for precision bits... the numpy comp is done in float32 because of the bad CPU performance for float16.  Wish there was a more consistent way to comparable float16 results on all platforms.

tinygrad/renderer/cstyle.py
```diff
@@ -273,8 +273,12 @@ class HIPLanguage(CStyleLanguage):
   __attribute__((device)) __attribute__((const)) _Float16 __ocml_sqrt_f16(_Float16);
   }\n""" + '\n'.join([_make_hip_dtype(*x) for x in [("signed int", "int", 2),
                      ("_Float16", "half", 2), ("_Float16", "half", 4), ("_Float16", "half", 8), ("_Float16", "half", 16),
-                     ("float", "float", 2), ("float", "float", 4), ("float", "float", 8)]]) + \
-  'extern "C" __attribute__((global))'
+                     ("float", "float", 2), ("float", "float", 4), ("float", "float", 8)]]) + """
```
chenyuxyz, line 276: not specific to this pr but similar to `half_prekernel` i think we want to include these wmma functions only for kernels that have wmma

chenyuxyz (commented):
oh yeah because that is not centered around 0 and output is near to 1024 already, that makes sense

tinygrad/renderer/cstyle.py
flammit: yeah that's a good call.  will setup a PR to fix this in METAL and HIP.

PR #3291, opened 2024-02-01
Update copyright year to 2024

rylan-justice:


nextgens:
@rylan-justice copyright is about the year of publication, it shouldn't be updated

rylan-justice:
You're right, @nextgens. Thank you for the correction!

PR #3303, opened 2024-02-02
Fix dot product on buffers with zero strides

obadakhalili:
Trying to apply the `.dot` method on tensors that translate directly to buffer ops of type `CONST` will produce incorrect results. Example:

```py
from tinygrad import Tensor

a = Tensor.ones((2, 3))
b = Tensor.full((3, 4), 2.0)
x = a.dot(b)

print(x.numpy()) # should be matrix of 6s, but getting a matrix of 2s
```

This issue only applies to interpreted backends because of an optimization in the `_interpret_ast` function which rewrites ASTs containing `ReduceOps.SUM` followed by a `BinaryOps.MUL` into `TernaryOps.MULACC`.

In order for `MULACC` to execute correctly, it uses strides info of input arrays, which doesn't work on input arrays that came directly from a buffer op of type `CONST` and were broadcasted later, because `np.broadcast_to` keeps the strides info of input arrays for optimization purposes, and when both inputs contain strides info that are all zeros, it confuses the algorithm into producing incorrect results.

The same issue arose for all cases that need to execute `TernaryOps.MULACC` on input arrays that came from broadcasted `BufferOps.CONST`. For example, computing the gradients of the following scalar also produces incorrect results:

```py
from tinygrad import Tensor

a = Tensor.ones(2, 3, requires_grad=True)
b = Tensor.arange(3.0, requires_grad=True)
x = a.dot(b).sum()

x.backward()
b.grad.numpy() # array([1., 1., 1.], dtype=float32), which should be [2., 2., 2.]
```

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -588,7 +588,7 @@ class TestOps(unittest.TestCase):
       a = Tensor(3.14)
       a.matmul(a)
   def test_dot_with_const_buffer(self):
-    helper_test_op([], lambda: torch.ones((2,3)).matmul(torch.full((3,4), 2.0)), lambda: Tensor.ones((2,3)).dot(Tensor.full((3,4), 2.0)), forward_only=True)
+    helper_test_op([], lambda: torch.ones((2,3)).matmul(torch.full((3,4), 2.0)), lambda: Tensor.ones((2,3)).dot(Tensor.full((3,4), 2.0)), forward_only=True)  # noqa: E501
 
   def test_matmul_simple(self):
     helper_test_op([(4), (4,4)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-4)
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -587,8 +587,21 @@ class TestOps(unittest.TestCase):
     with self.assertRaises(AssertionError):
       a = Tensor(3.14)
       a.matmul(a)
-  def test_dot_with_const_buffer(self):
-    helper_test_op([], lambda: torch.ones((2,3)).matmul(torch.full((3,4), 2.0)), lambda: Tensor.ones((2,3)).dot(Tensor.full((3,4), 2.0)), forward_only=True)  # noqa: E501
+  def test_mulacc_with_zero_strides(self):
+    helper_test_op(
+      [],
+      lambda: torch.tensor(1.0).reshape((1,1,1)).expand(2,4,3).mul(torch.tensor(1.0).reshape((1,1,1)).expand(2,4,3)).sum(-1),
+      lambda: Tensor(1.0).reshape((1,1,1)).expand(2,4,3).mul(Tensor(1.0).reshape((1,1,1)).expand(2,4,3)).sum(-1),
+      forward_only=True
+    )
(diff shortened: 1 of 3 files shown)


chenyuxyz:
instead of skipping i'd fix `einsum_mulacc`. also no `# noqa: E501` for new code

obadakhalili:
I pushed a fix to the `einsum_mulacc` function. Hopefully, it addresses the full issue.

Not sure why the first time I thought `einsum_mulacc` doesn't have enough info to compute the correct results. Thanks for pushing me to do it correctly though @chenyuxyz.

Here is a little explanation of my understanding of the bug and my fix:

`einsum_mulacc` uses `np.einsum`, which because it doesn't use the strides info to avoid duplicating computing the same results, the current implementation ignores sending axes that their strides are zero and expands the result later. The problem with this though is that the axes to be multiplied and summed should always be sent to ensure computing the correct result, which is what I added in my fix.

tinygrad/runtime/ops_cpu.py
```diff
@@ -12,7 +12,10 @@ def einsum_mulacc(einsum, get_strides, expand):
   def einscripts(x): return ''.join(["abcdefghijklmnopqrstuvwxyz"[i] for i in x])
   def axes_slice(strides): return tuple(i for i,s in enumerate(strides) if s != 0), tuple(slice(None) if s != 0 else 0 for s in strides)
   def mulacc(a, b, new_shape):
-    (a_axes, a_slices), (b_axes, b_slices) = axes_slice(get_strides(a)), axes_slice(get_strides(b))
+    (a_axes, a_slices), (b_axes, b_slices) = axes_slice(a_strides := get_strides(a)), axes_slice(get_strides(b))
```
chenyuxyz, line 15: i'd remove the `axes_slice` helper since it's incomplete, and refactor the whole logic. also it's not obvious why it only updates `a` and not `b`

test/test_ops.py
```diff
@@ -587,6 +587,21 @@ def test_dot(self):
     with self.assertRaises(AssertionError):
       a = Tensor(3.14)
       a.matmul(a)
+  def test_mulacc_with_zero_strides(self):
```
chenyuxyz, line 590: i like the old test that uses const that motivates this fix, consider adding that back in addition to this

tinygrad/runtime/ops_cpu.py
obadakhalili: - You are right for pointing out the need to refactor, though it should be within the `axes_slice` helper. The current code handles concerns for this helper outside of it, so I refactored it to take `sum_axes` into account, which is cleaner and less confusing.
- As for why only one operand is needed, this is because `np.einsum` handles broadcasting the value for the operand with missing axes. However, this isn't part of the new code, which is also cleaner and less confusing.

test/test_ops.py
obadakhalili: Done.

obadakhalili:
@chenyuxyz sorry for over-pinging you 😅, but do you think this PR qualifies for bounty #1 (find and fix a bug)?

chenyuxyz:
we don't have bounty for find and fix a bug

PR #3305, opened 2024-02-03
[try] Parallel compile

chaosagent:
Just trying... Would appreciate feedback/suggestions on how to make it beautiful ;)

AFAICT, no visible behavior changes other than beam running for all kernels before any are executed for real.

After:
```
(venv) chaos@tiny4:~/tinygrad$ time LATEWINO=1 LATEBEAM=2 HALF=1 HIP_VISIBLE_DEVICES=3 BS=512 DEBUG=0 python3 examples/hlb_cifar10.py 
shuffling training dataset in 1199.50 ms (epoch=0)
  0 15307.16 ms run, 15304.89 ms python,    2.27 ms HIP, 1200.00 loss, 0.000015 LR, 0.42 GB used,    230.20 GFLOPS,   3523.74 GOPS
  1  967.79 ms run,  967.16 ms python,    0.63 ms HIP, 1202.00 loss, 0.000030 LR, 2.84 GB used,   3639.10 GFLOPS,   3521.88 GOPS
  2   19.47 ms run,    2.27 ms python,   17.19 ms HIP, 1190.00 loss, 0.000045 LR, 2.85 GB used, 180923.39 GFLOPS,   3521.88 GOPS
  3   18.30 ms run,    1.66 ms python,   16.64 ms HIP, 1175.00 loss, 0.000060 LR, 2.85 GB used, 192486.75 GFLOPS,   3521.88 GOPS
  4   18.48 ms run,    1.62 ms python,   16.86 ms HIP, 1160.00 loss, 0.000075 LR, 2.85 GB used, 190574.21 GFLOPS,   3521.88 GOPS
  5   18.59 ms run,    1.62 ms python,   16.97 ms HIP, 1160.00 loss, 0.000090 LR, 2.85 GB used, 189441.68 GFLOPS,   3521.88 GOPS
  6   18.50 ms run,    1.60 ms python,   16.90 ms HIP, 1152.00 loss, 0.000105 LR, 2.85 GB used, 190358.31 GFLOPS,   3521.88 GOPS


...

994   18.36 ms run,    1.57 ms python,   16.79 ms HIP,  499.00 loss, 0.000107 LR, 2.85 GB used, 191810.78 GFLOPS,   3521.88 GOPS
995   18.60 ms run,    1.57 ms python,   17.02 ms HIP,  493.75 loss, 0.000102 LR, 2.85 GB used, 189374.96 GFLOPS,   3521.88 GOPS
[...]

PR #3309, opened 2024-02-04
[try] parallel compilers, try 2, maybe a little verbose

chaosagent:
same behavior more or less as #3305. future is implemented as a JITRunner, and more lines this time


chaosagent:
this compromises CompiledASTRunner by making it double as a device-state-free data class... otherwise compiler outputs would need to be returned in a special dataclass or a tuple. some win and some lose

tinygrad/realize.py
```diff
@@ -40,13 +40,11 @@ def lower_schedule_item(si:ScheduleItem) -> Optional[JITRunner]:
 
 logops = open(getenv("LOGOPS", ""), "a") if getenv("LOGOPS", "") else None
 def run_schedule(schedule:List[ScheduleItem]):
+  prgs = [lower_schedule_item(si) for si in schedule]  # kick off any parallel compilations
```
geohot, line 43: If there's no parallel, this is bad

chaosagent:
thanks for the feedback! will try again!

Issue #3163, opened 2024-01-18
Move merge_views into view.py


geohot:
It shouldn't use a ShapeTracker and real_strides, it should use the same logic as View.reshape, but capable of merging two arbitrary Views

It should also be `View.__add__` if that's okay to return an Optional

geohot:
@corsix You interested in this? I think you might already have most of it in your fork

chenyuxyz:
done in #3686

PR #3311, opened 2024-02-04
patch merge_views to handle missing case

thelonejordan:
When `vm1` is contiguous and size of `vm1` and `vm2` match, then `merge_views(vm2, vm1)` must be `vm2` reshaped to `vm1` , if the reshape is possible.

benchmarks:
```python
SEED=6961 CNT=10000 CHECK_NEQ=1 FUZZ=plus PYTHONPATH="." python3 test/external/fuzz_shapetracker_math.py
```
with `SEED=6961`,
master:
same but unequal 5.54%
branch:
same but unequal 5.42%

with `SEED=5115`,
master:
same but unequal 6.05%
branch:
same but unequal 5.90%

### DIFF
--- a/tinygrad/shape/shapetracker.py
+++ b/tinygrad/shape/shapetracker.py
@@ -18,8 +18,9 @@ def un1d(shape:Tuple[sint, ...], offs:sint) -> List[sint]:
 
 @functools.lru_cache(maxsize=None)
 def merge_views(vm2:View, vm1:View) -> Optional[View]:
-  if vm1.contiguous and vm1.shape == vm2.shape: return vm2
   if vm2.contiguous: return vm1
+  if vm1.contiguous and vm1.shape == vm2.shape: return vm2
+  if vm1.contiguous and vm1.size() == vm2.size() and (ret := vm2.reshape(vm1.shape)) is not None: return ret
   if not vm2.mask and vm1.offset == 0 and None not in (rstrides := ShapeTracker((vm2, vm1)).real_strides()):
     return View.create(vm1.shape, cast(Tuple[sint, ...], rstrides), vm2.offset, vm1.mask)
   if vm1.mask:


chenyuxyz:
this feels quite marginal, does it fix all `vm1.contiguous` case?

thelonejordan:
> this feels quite marginal, does it fix all `vm1.contiguous` case?

No it doesn't. And yeah this is marginal, but don't we care about correctness, as we wanted to remove st.real_strides() from merge_views? https://github.com/tinygrad/tinygrad/issues/3163

thelonejordan:
@chenyuxyz Ideally, this shouldn't even be a case as reshapes would merge the views if its possible. This is probably due to the way we are comparing the ShapeTrackers using st_equal (we add backup to the front of the other shapetracker and then simplify...)

geohot:
Is there a more generic way to write this? I don't think it's worth the marginal gain

tinygrad/shape/shapetracker.py
```diff
@@ -18,7 +18,9 @@ def un1d(shape:Tuple[sint, ...], offs:sint) -> List[sint]:
 
 @functools.lru_cache(maxsize=None)
 def merge_views(vm2:View, vm1:View) -> Optional[View]:
-  if vm1.contiguous and vm1.shape == vm2.shape: return vm2
+  if vm1.contiguous:
+    if vm1.shape == vm2.shape: return vm2
+    if vm1.size() == vm2.size() and (vm2_new := vm2.reshape(vm1.shape)) is not None: return vm2_new
```
geohot, line 23: Also do you have to check the size here? Won't the reshape always fail if they are different sizes?

thelonejordan: Thats correct. But isn't reshape costlier (it would call merge_dims and that woud return None when it can't merge the dims...), so was trying to avoid calling it, thats why the size equality check.

geohot: Speed is never worth complexity!

thelonejordan: tried removing it, realised reshape throws ValueError when the sizes don't match!

PR #3312, opened 2024-02-04
clean up einsum_mulacc

chenyuxyz:


obadakhalili:
- Nice cleanup! Sometimes I write somewhat condensed code because I worry about the number of lines. But it is good to know now that I have more degrees of freedom.
- Also, I would like to note a small NIT: I believe the reason `einsum_mulacc` uses higher-order functions is because the backend source for the data could be anything, this now includes Pytorch and NumPy, which both happen to have a `.ndim` property, but this assumption could turn out false in the future and the code would break. So from a better abstraction point of view, it feels wrong to access `.ndim` directly.

PR #3316, opened 2024-02-04
add batching in beatiful_mnist.py

flowpoint:
beatiful_mnist.py with TORCH=1 oom-ed on CPU only (16GB mem) and on a 4090 because it didnt minibatch the testset, this should fix it and and save memory


flowpoint:
#3323 related, also fixes it, both save memory

PR #3317, opened 2024-02-04
Fix: Always cast ONNX Slice op arguments into ints

terafo:
This op has optional arguments which are stored as floats, but, unlike required params, are not being cast into ints, which causes some models crashing.


chenyuxyz:
i think int32 is fine? although the onnx spec says it can be int64, but we already casted start and end to int32 so that might be a whole different problem

PR #3318, opened 2024-02-04
Fix Tensor.mean to compute the mean correctly when 0-length axes are selected

obadakhalili:
Using `Tensor.mean` to compute the mean for tensors with 0(s) in their shape should output `nan`s when selecting those zeros in `axis`. This complies with Pytorch's behavior and actually makes sense.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -541,8 +541,9 @@ class Tensor:
 
   def mean(self, axis=None, keepdim=False):
     assert all_int(self.shape), "does not support symbolic shape"
-    out = self.sum(axis=axis, keepdim=keepdim)
-    return out.mul(prod(out.shape)/prod(self.shape)) if 0 not in self.shape else out
+    sum = self.sum(axis=axis, keepdim=keepdim)
+    x_shape_prod, sum_shape_prod = prod(self.shape), prod(sum.shape)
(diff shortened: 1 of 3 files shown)


tinygrad/tensor.py
```diff
@@ -541,8 +541,9 @@ def min(self, axis=None, keepdim=False): return -((-self).max(axis=axis, keepdim
 
   def mean(self, axis=None, keepdim=False):
     assert all_int(self.shape), "does not support symbolic shape"
-    out = self.sum(axis=axis, keepdim=keepdim)
-    return out.mul(prod(out.shape)/prod(self.shape)) if 0 not in self.shape else out
+    sum_t = self.sum(axis=axis, keepdim=keepdim)
+    x_shape_prod, sum_shape_prod = prod(self.shape), prod(sum_t.shape)
+    return sum_t if sum_shape_prod == 0 else sum_t.div(x_shape_prod / sum_shape_prod)
```
chenyuxyz, line 546: do you really need these variables and an extra line? it's less readable
i think `return out.div(prod(self.shape)/prod(out.shape)) if 0 not in out.shape else out` suffices

obadakhalili: Yes, I agree. I just noticed how similar the new logic could be to the old code. I will change it.

chenyuxyz:
cool, thanks!

obadakhalili:
thanks for the review!

PR #3319, opened 2024-02-05
PoC UnaryOps before expand

chaosagent:
** this is a PoC, and probably not how this should eventually be implemented ** (i will need to read the rest of lazy.py...)

the matmul forwards is: `half(sum(float(mul(a,b))))`, so the backwards is `mul(half(expand(float(grad))),b)`. Expands always realize in the lazy graph, but half(expand(float)) is just expand, so the casts can be simplified out by swapping the half and expand. this saves membw, and also allows half/half tc to trigger.

this change just applies casts to smaller dtypes before any movementops, just to check perf, and it's quite good (7950x):

```
993   14.06 ms run,    0.86 ms python,   13.21 ms HIP,  496.00 loss, 0.000112 LR, 2.08 GB used, 250407.02 GFLOPS,   3521.88 GOPS
994   14.05 ms run,    0.85 ms python,   13.20 ms HIP,  493.25 loss, 0.000107 LR, 2.08 GB used, 250617.10 GFLOPS,   3521.88 GOPS
995   14.07 ms run,    0.84 ms python,   13.23 ms HIP,  494.75 loss, 0.000102 LR, 2.08 GB used, 250300.24 GFLOPS,   3521.88 GOPS

```

### DIFF
--- a/tinygrad/lazy.py
+++ b/tinygrad/lazy.py
@@ -70,6 +70,7 @@ class LazyBuffer:
 
   def cast(self, dtype:DType, bitcast:bool=False):
     if self.dtype == dtype: return self
+    if dtype.itemsize <= self.dtype.itemsize and self != self.base: return self.base.cast(dtype, bitcast)._view(self.st)
     return create_lazybuffer(self.device, ShapeTracker.from_shape(self.shape), dtype, UnaryOps.CAST, (dtype, bitcast), (self,))
 
   def is_unrealized_const(self): return not self.base.realized and self.base.op is LoadOps.CONST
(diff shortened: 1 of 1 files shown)


chaosagent:
i'm not sure what to think about pushing ops.

would it be good to do small -> large casts before expand? casts are surprisingly costly (mostly for very small dtypes to fp16/32 etc).

can also reorder size-changing casts and other realizes (ie always try to realize as the smallest dtype)

chaosagent:
not sure why this is failing on hipcpu?

Qazalin:
hipcpu should work now if you re-run, the kernel used a new instruction `v_fma_mixhi_f16` in the rdna3.

chaosagent:
** note that we can no longer write .float().expand().half() to get float().sum().half() in backwards after this **

PR #3322, opened 2024-02-05
hip can only take local buffers of size 2**14

chaosagent:
limited too conservatively last time


chenyuxyz:
do you have a reference that you can link here or add a comment?

chaosagent:
I just guessed why the compiler was crashing and added a limit that worked... Lol

Can look more into it tomorrow

geohot:
This is also changing it for every device. There aren't any consequences of a compiler crash, right? It should just not be a valid node in the search.

PR #3323, opened 2024-02-05
set expandable_segments:True in ops_torch.py to reduce memory fragmentation

flowpoint:
the current ops_torch.py backend ooms because of memory fragmentation on a 4090 on: 
test/models/test_real_world.py::TestRealWorld::test_llama 
test/models/test_whisper.py::TestWhisper::test_transcribe_long 
examples/beautiful_mnist.py

setting the envvar PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True fixes it for now.

setting it in runtime/ops_torch.py because its a torch-runtime specific tweak

### DIFF
--- a/tinygrad/runtime/ops_torch.py
+++ b/tinygrad/runtime/ops_torch.py
@@ -1,3 +1,4 @@
+import os
 import torch
 from typing import Dict, Callable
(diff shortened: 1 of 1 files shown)


geohot:
We don't really care about the torch backend.

PR #3325, opened 2024-02-06
wmma: reverse the order of lobal and global to enable group with TC

flammit:
Enables OptOp.GROUP when using tensor cores.  This PR reverses the global and local dim order to so that the first global and local dims are in threadgroup.x.  This also fixes the group_for_reduce strides logic so that it works properly with HIP TCs.

Depends on #3320


Qazalin:
released expanded lds support for hipcpu, tests should be good now https://github.com/Qazalin/remu/commit/db1d143ce05a4791f953923ff0c5792c85cc5369

PR #3326, opened 2024-02-06
factor out parallel compilation in beam_search

flowpoint:


geohot:
Why?

flowpoint:
just starting on "A beautiful refactor of the parallel BEAM into generic parallel kernel compilation"
ig, this was too granular

PR #3335, opened 2024-02-07
conceptually simpler fancy index

geohotstan:
this was pulled from my setitem attempt.

basically while doing setitem I found out we could reduce multiple dims at once and that reduces kernel count
(lol didn't know that before)

so instead of doing 
`(idx == arange).mul(ret).sum(sum_dim)`
reduction for each dimension iteratively
which makes for unreadable reshape args and generates a reduce kernel for each sum axis, for ex:
```
def test_slice_fancy_indexing_no_dim_collapse(self):
->  helper_test_op([(2,5,6,5,3,4)], lambda x: x[a,b,c,d,e], lambda x: x[i,j,k,o,p])

...
*** METAL     22 r_75_2_8_3_2                           arg   3 mem  0.00 GB tm     12.63us/     0.33ms (    1.14 GFLOPS,    2.28 GB/s)
*** METAL     23 r_15_2_3_8_3_5                         arg   3 mem  0.00 GB tm     12.00us/     0.34ms (    1.80 GFLOPS,    1.93 GB/s)
*** METAL     24 r_2_5_3_4_4_3_6                        arg   3 mem  0.00 GB tm     12.12us/     0.36ms (    1.43 GFLOPS,    1.20 GB/s)
*** METAL     25 r_3_5_2_4_4_3_5                        arg   3 mem  0.00 GB tm     15.88us/     0.37ms (    0.91 GFLOPS,    0.74 GB/s)
*** METAL     26 r_15_2_8_3_4_3                         arg   3 mem  0.00 GB tm     12.13us/     0.38ms (    1.43 GFLOPS,    1.43 GB/s)
...
```

we do instead
```
# idx == arange -> creates 1 mask
# mask1 * mask2 * ... -> combines the masks together
mask = (idx == arange) * (idx2 == arange2) * ...
ret = mask * ret
ret = ret.sum([all_dimsum_yum])

which just has 1 reduce kernel
...
*** METAL     27 r_45_16_225_4_4                        arg   3 mem  0.00 GB tm    176.50us/     0.60ms (   29.37 GFLOPS,   14.83 GB/s)
...

```

I think these are the reduce kernels?? Idk...... might be wrong.
doesn't make forward pass much faster but makes bp pretty consistent.
I haven't looked into the details yet.


tinygrad/tensor.py
```diff
@@ -418,21 +418,21 @@ def calc_dim(tensor_dim:int) -> int:
...
-        for a,i,sd in zip(arange, reshaped_idx, sum_dim): ret = (a==i).mul(ret).sum(sd)
+      idx: Dict[int,Tensor] = {(dim := calc_dim(td)):(tensor<0).where(ret.shape[dim],0) + tensor for td,tensor in zip(type_dim[Tensor], tensor_index)}
+
+      eqs, first_dim, max_idx_dim, last_dim  = [], min(idx.keys()), max(i.ndim for i in idx.values()), max(idx.keys())
+      for dim, i in idx.items():
+        a = i.reshape(i.shape + (1,)*(ret.ndim - first_dim))
+        b = Tensor.arange(ret.shape[dim]).reshape(ret.shape[dim:dim+1] + (1,)*(ret.ndim - dim - 1))
+        eqs.append(a == b)
+
+      try: mask = reduce(lambda x,y: x.mul(y), eqs).contiguous()
```
geohotstan, line 429: this `contiguous` is cuz maybe a bug in the shapetrackers
or maybe view not merging 
or idk

gotta to some investigating 🧐

geohot: Yea, that needs to be understood. contiguous should never change the final output (we should fuzz this)

geohotstan:
I know we're probably moving to some symbolic indexing stuff so this PR might not be too useful
but I guess this is still better than before :D ... I think

some notes:
- took out 2 transposes in `gather` yay
- left `_resolve_dim` for `gather` for dfriehs ;)
- `gather` is super cursed conceptually. Truth be told, I literally spam programmed to figure out the concept. I don't think any comments can explain why it works. But the code is readable I think.
- `gather` cannot use `getitem` because the sum_dim and the added dim for masking is different.
- advanced indexing uses setitem stuff. The backwards is significantly faster. Forwards take very slight perf hit.
- for setitem, there's a super annoying edge case that makes masking with arange (current approach) mighty slow if it is to work generically (it works tho). See example below:
```
reference = Tensor.arange(16).reshape(4, 4)
a = Tensor([[0],[1]])
b = Tensor([2,1,1])
reference[a,b] = Tensor([[99,100,111],[123,234,345]])

[[  0 345 123   3]
 [  4   5   6   7]
 [  8   9  10  11]
 [ 12  13  14  15]]
```
the index `b` has two 1's. This makes the value of the second 1 overwrite the value of the first 1. (234 overwritten by 345)
this renders my attempt useless
unlucky 😭 

----------------------

I've started on the fuzzer though this is quite a challenge. Not sure if I'm capable but I'll give it a try.

chenyuxyz:
I think it's quite slower. `PYTHONPATH=. python examples/beautiful_mnist.py` uses gather to generate mini-batch, and on master it took 8 seconds on my M1 Max, and took 12 seconds on this branch. I like the simplification (still quite complicated but maybe it's what it is), but want to understand the slowdown better.

geohotstan:
here're some early findings.

The reason for `examples/beautiful_mnist.py` slowdown in particular is cuz the final kernel's `TensorCore` has an inefficient shape caused by taking in inefficient stride caused by different reshape_args

Starting with TC:
```
# master branch
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(1, 512, 1, 28, 28), strides=(0, 784, 0, 28, 1), offset=0, mask=None, contiguous=True),)))       
  1  ┗━┳ SUM (0,)                                                                                                                                                                    
  2    ┗━┳ MUL                                                                                                                                                                       
  3      ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(60000, 512, 1, 28, 28), strides=(512, 1, 0, 0, 0), offset=0, mask=None, contiguous=False),)
))                                                                                                                                                                                   
  4      ┗━━ LOAD MemBuffer(idx=2, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(60000, 512, 1, 28, 28), strides=(784, 0, 0, 28, 1), offset=0, mask=None, contiguous=False),
)))                                                                                                                                                                                  
TENSOR CORES [(1, 784, 1)] [(0, 512, 1)] tensor_core<[8, 8, 8], dtypes.float, dtypes.float>    
...
*** METAL      7 r_16_49_2_2_4_2_7500_8_2_4_2           arg   3 mem  0.31 GB tm     33.31ms/    40.03ms ( 1446.28 GFLOPS,    9.39 GB/s)
```
```
# this branch
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(512, 1, 1, 28, 28), strides=(784, 0, 0, 28, 1), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ SUM (1,)                                                                                                                                                                    
  2    ┗━┳ MUL                                                                                                                                                                       
  3      ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(512, 60000, 1, 28, 28), strides=(0, 784, 0, 28, 1), offset=0, mask=None, contiguous=False),
)))                                                                                                                                                                                  
  4      ┗━━ LOAD MemBuffer(idx=2, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(512, 60000, 1, 28, 28), strides=(60000, 1, 0, 0, 0), offset=0, mask=None, contiguous=False)
,)))                                                                                                                                                                                 
TENSOR CORES [(0, 512, 60000)] [(1, 784, 1)] tensor_core<[8, 8, 8], dtypes.float, dtypes.float>    
...
*** METAL      7 r_4_49_4_2_2_4_2_7500_8_2_2_4          arg   3 mem  0.31 GB tm    114.28ms/   122.17ms (  421.51 GFLOPS,    2.74 GB/s)
```

The difference in strides here are `(60000, 1, 0, 0, 0)` and `(512, 1, 0, 0, 0)`
This is the mask resulting from `arange == idx`


An example for why that is:
```
self.shape[dim] == 60000
idx.shape == (512)

# on master
arange.shape == (60000, 1)
idx.shape == (1, 512)

mask = arange == idx 
mask.shape == (60000, 512)
mask.lazydata.st.real_strides() == (512, 1)


# on this branch
arange.shape == (1, 60000)
idx.shape == (512, 1)

mask = arange == idx 
mask.shape == (512, 60000)
mask.lazydata.st.real_strides() == (60000, 1)
```

obviously I can hack the `reshape args` to swap them around, but that's obv dumb
maybe I could check for `self.shape[dim] > idx.numel()` and decide swap or not to minimize stride, but that's gonna be really complex.

maybe there's some other change in shapetracker that handles this for any generic case to minimize strides?
like for `mask.shape == (512, 60000)`  
`stride == (60000, 1)` is the same as `stride == (1, 512)`
but I tried hand reshaping the mask to mangle the stride to `(1, 512)` and that deactivated `TensorCore` lol (though perf was still much faster!!!)

need some more time to think this through 🤔

chenyuxyz:
i don't mind the speed difference if it's a bad default, we can use the BEAM=2 number to compare perf

arange has been updated to a more efficient algorithm, so this might need a refresh

geohotstan:
both BEAM=2 and base line are still slower for this PR on my local mac, though the speed difference is less.

geohotstan:
Woahhh if I just copy what patosai did in #4036 by pre-expanding the shapes, it's getting much better FLOPS with BEAM=2
and the base speed is a tiny bit faster too

This is for the `beautiful_mnist` gather thing.
now:
`avg:  1801.71 GFLOPS    21.37 GB/s           total:     4 kernels    63.65 GOPS     0.75 GB    35.33 ms`

before:
`avg:   918.02 GFLOPS    18.89 GB/s           total:     5 kernels    48.62 GOPS     1.00 GB    52.97 ms`

master:
`avg:  1365.07 GFLOPS    28.09 GB/s           total:     5 kernels    48.62 GOPS     1.00 GB    35.62 ms`

lmao before I thought pre-expanding would make it slow because more element-wise arithmetic would have to be performed
man I feel dumb

chenyuxyz:
fine to skip some fancy indexing tests if it's taking too long for PYTHON backend

chenyuxyz:
looks like `beautiful_mnist` is slightly slower, is that expected?

geohotstan:
> looks like `beautiful_mnist` is slightly slower, is that expected?

for me it's about the same
```
This branch:
loss:   0.06 test_accuracy: 97.95%: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:13<00:00,  5.05it/s]

master:
loss:   0.07 test_accuracy: 98.22%: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:13<00:00,  5.37it/s]
```

This branch also decreases the kernels for 
```
t = Tensor.randn(60000, 1, 28, 28)
t[Tensor.randint(512, high=60000)].realize()
```
from 5 (master) -> 4


I'm gonna check pre-expanding the shapes on master just to check speed.

chenyuxyz:
i ran the benchmark for this, you can compare the ones with master

it's possible that multi level reduce is slower due to hand code optimization not handle it well and optimize more for single reduce. you can try with `NOOPT=1` and some BEAM (might not be as effective as BEAM does not swap reduce orders). it's fine to merge this given less kernel as long as speed is similar and the reason is well understood

chenyuxyz:
cool! merging. @flammit fyi about conditions to invoke TC

PR #3339, opened 2024-02-07
[WIP] fix stddev and variance using creating 2 kernels

aranggitoar:
For some reason, when we chain both (in any order) or any `.var()` and `.std()` after `.mean()`, it fuses the kernel.

So a solution without adding a new Op or using a complex solution like Welford's algorithm as suggested by [Fleetwood in Discord](https://discord.com/channels/1068976834382925865/1070745817025106080/1196855032604278875), would be making `var` chains `mean` by default. Hence this will make stddev to also have only one kernel.


aranggitoar:
Fixing the failing tests.

aranggitoar:
That's right, hence the nan value.

I'll take a couple more days on this, if there's still nothing, I'll just report some findings with the methods I've tried.

geohot:
Yea this is just wrong

PR #3348, opened 2024-02-08
[WIP] Support lazy layers

obadakhalili:
This is a proposal to support lazy layers in tinygrad.

When building complex models it is harder to manually compute the output of each layer to use it as the input argument for the next layer.

The current code implements the base `LazyLayer` class and the lazy layers for `Linear` and `Conv2d`, and uses them in the `beautiful_mnist.py` example for illustration.

I would happily implement lazy layers for the rest and add tests if the proposal is desired and the approach is good.


geohot:
I don't like this, it makes it harder to see bugs. Our behavior looks like torch, not Keras

PR #3349, opened 2024-02-08
reflect changes to shapetracker in example printouts

masonmahaffey:
Back in September [ShapeTracker](https://github.com/tinygrad/tinygrad/blob/d8ad9e5660a50b516a37de7f0e56c0259a911c08/tinygrad/shape/shapetracker.py#L96) was modified to return a tuple of views instead of a list in this PR: #1909 like so:
```python3
class ShapeTracker:
  views: Tuple[View, ...]
```

And `shape` was moved within View as opposed to ShapeTracker itself which now uses the `@property` decorator to access shape thereby hiding it in printouts in this PR: #1844
```python3
@property
  def shape(self) -> Tuple[sint, ...]: return self.views[-1].shape
```

These changes weren't reflected in the [docs/abstractions.py](https://github.com/tinygrad/tinygrad/blob/d8ad9e5660a50b516a37de7f0e56c0259a911c08/docs/abstractions.py#L310) example printouts. 

I am aware this violates the _do not touch_ the docs rule in [CONTRIBUTING.md](https://github.com/tinygrad/tinygrad?tab=readme-ov-file#contributing), however I think this is a valid exception as it keeps the walkthrough docs up to date even if in a small way. 

That being said, I totally understand if the PR must be closed to adhere to that rule.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -10,10 +10,11 @@ if CI:
 
 FORWARD_ONLY = getenv("FORWARD_ONLY", 0)
 PRINT_TENSORS = getenv("PRINT_TENSORS", 0)
+
 def helper_test_op(shps, torch_fxn, tinygrad_fxn=None, atol=1e-6, rtol=1e-3, grad_atol=1e-4, grad_rtol=1e-3,
-                   forward_only=False, vals=None, a=-0.5, b=3):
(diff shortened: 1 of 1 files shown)


geohot:
It does violate the rule, but it's mechanical enough I'll merge it

PR #3351, opened 2024-02-09
exclude half with PYTHON in test_dtype.is_dtype_supported

chenyuxyz:
half memoryview only in 3.12+. rest of the test_dtype (bounty) seems to be legit issue in ops_python.


reddyn12:
I was able to get around that with numpy buffer casts

PR #3352, opened 2024-02-09
[WIP] ops_python Dtype tests pass

reddyn12:
Currently at 32 fails. Lmk if I'm misunderstanding the point of the emulator by using numpy casts instead of memoryview casts. Also, memoryview doesn't support 'e' float16 cast.

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -53,7 +53,9 @@ def _test_op(fxn, target_dtype:DType, target):
 def _test_cast(a:Tensor, target_dtype:DType):
   _test_op(lambda: a.cast(target_dtype), target_dtype, list(a.numpy().astype(target_dtype.np)))
 def _test_bitcast(a:Tensor, target_dtype:DType, target=None):
-  _test_op(lambda: a.bitcast(target_dtype), target_dtype, target or a.numpy().view(target_dtype.np).tolist())
+  if target is not None: b = target
+  else: b = a.numpy().astype(target_dtype.np).tolist()
+  _test_op(lambda: a.bitcast(target_dtype), target_dtype, b)
 
 class TestDType(unittest.TestCase):
   DTYPE: Any = None
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -53,9 +53,12 @@ def _test_op(fxn, target_dtype:DType, target):
 def _test_cast(a:Tensor, target_dtype:DType):
   _test_op(lambda: a.cast(target_dtype), target_dtype, list(a.numpy().astype(target_dtype.np)))
 def _test_bitcast(a:Tensor, target_dtype:DType, target=None):
-  if target is not None: b = target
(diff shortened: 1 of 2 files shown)


test/test_dtype.py
```diff
@@ -53,7 +53,12 @@ def _test_op(fxn, target_dtype:DType, target):
 def _test_cast(a:Tensor, target_dtype:DType):
   _test_op(lambda: a.cast(target_dtype), target_dtype, list(a.numpy().astype(target_dtype.np)))
 def _test_bitcast(a:Tensor, target_dtype:DType, target=None):
-  _test_op(lambda: a.bitcast(target_dtype), target_dtype, target or a.numpy().view(target_dtype.np).tolist())
+  if Device.DEFAULT=='PYTHON':
```
geohot, line 56: Why?

tinygrad/runtime/ops_python.py
```diff
@@ -68,11 +73,11 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
         dl[i] = dtype
         if uop is UOps.DEFINE_GLOBAL:
           assert dtype.fmt is not None
-          ul[i] = [pbufs.pop(0).cast(dtype.fmt)] * warp_size
+          ul[i] = [np.frombuffer(pbufs.pop(0), dtype=dtype.np)] * warp_size
```
geohot, line 76: No numpy!

test/test_dtype.py
reddyn12: This gets executed before the bit cast: a.numpy().view(target_dtype.np).tolist()

tinygrad/runtime/ops_python.py
reddyn12: Lol ok

tinygrad/runtime/ops_python.py
```diff
@@ -109,7 +113,8 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
           else:
             # TODO: add real cast
             if dtypes.is_int(dtype):
-              ul[i] = [int(x) for x in inp[0]]
+              if dtypes.is_unsigned(dtype): ul[i] = [int(256+x) if x < 0 else int(x) for x in inp[0]]
```
geohot, line 116: WTF

geohot:
This isn't good. You may have made the tests pass, but this code has many bugs and age cases not handled. The goal is not to make the tests pass, the goal is to write the correct code.

test/test_dtype.py
geohot: This is not resolved? Why should PYTHON backend be different?

tinygrad/runtime/ops_python.py
reddyn12: There is a 16 32.... u right, I'll do cast properly tmr.

reddyn12: and add better overflow tests

PR #3358, opened 2024-02-09
test_dtype now pass for Python backend

terafo:
It also fixed some issues with test_ops, but not all of them. Will investigate test_ops now.


tinygrad/runtime/ops_python.py
```diff
@@ -120,7 +120,10 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
             ul[i] = inp
           else:
             # TODO: add real cast
-            if dtypes.is_int(dtype):
+            if isinstance(arg, tuple) and arg[1] or (dtypes.is_int(dtype) and dtypes.is_int(dtp[0]) and dtype.itemsize == dtp[0].itemsize):
```
geohot, line 123: What is this second part of this if clause with the ints?

terafo: Bitcasting between int and uint of the same size.

geohot: Huh? If it's bitcast, wouldn't the first clause get it?

tinygrad/runtime/ops_python.py
```diff
@@ -73,7 +73,7 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
                 _store(m, ox*4 + oy*dtp[0].shape[1]*4 + j, v)
           elif dtp[2].sz > 1:
             for j,val in enumerate(inp[2]):
-              for m,o,v in zip(inp[0], inp[1], val): _store(m, o+j, v)
+              for m,o,v in zip(inp[0], inp[1], val if type(val) is list else inp[2]): _store(m, o+j, v)
```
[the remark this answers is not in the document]
terafo: Because inp[2] can be just a single float4 vector, which is basically just the same list of values that gets handled in the next elif statement, just labeled with different datatype. Here that case gets handled.

tinygrad/runtime/ops_python.py
terafo: Casting between int and uint and back in C has the same behavior as bitcasting, but isn't considered bitcasting, thus first part of clause doesn't get it. But doing proper cast would be better for sure.

tinygrad/runtime/ops_python.py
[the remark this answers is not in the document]
geohot: See what davidgonmar said about it in discord in #general

geohot:
Closing this. Been draft for a while, I don't think that "fix" is right

PR #3359, opened 2024-02-09
Add perfetto profiler to HIP

geohot:


geohot:
@nimlgen Here's a start on this for HIP, going to close because adding the profiling makes HIP way slower, but hopefully this isn't true for HSA. Maybe the perfetto output code is useful.

PR #3360, opened 2024-02-09
IS-Net (inference) example

zku:
IS-Net picture background removal from https://github.com/xuebinqin/DIS implemented in tinygrad (for inference only). Unfortunately still has 2 torch dependencies (interpolate/upsample, normalize).

Fetches model from https://huggingface.co/briaai/RMBG-1.4, but the original one hosted on Drive by https://github.com/xuebinqin/DIS works interchangeably.


geohot:
We aren't really looking to merge more examples, but thanks for using tinygrad!

PR #3361, opened 2024-02-09
MLPerf Resnet

chaosagent:
~~Trains to 71.86% in 45 epochs (14.2 hrs): https://wandb.ai/chaosagent/tinygrad-examples_mlperf/reports/stoic-silence--Vmlldzo2Nzg0MzIx~~

Trains to 75.9% in 43 epochs (<14 hrs): https://wandb.ai/chaosagent/tinygrad-examples_mlperf/runs/bfn1jsys/workspace?workspace=user-chaosagent

parts of data augmentation and some other stuff taken from @ojaffe https://github.com/tinygrad/tinygrad/pull/1404

Hacks in current implementation:

- [ ] unsyncbatchnorm in resnet.py is ugly
- [ ] maybe LARS should go in extra/ instead of optim.py?
- [ ] checkpointing code is ugly
- [x] eval is done with train = true to use the same jitted forward_step, so the batchnorm normalizer statistics are more noisy. also need to shuffle eval set to get good BN statistics
- [ ] top5 accuracy calculated in numpy (is there a good way in tinygrad?)
- [ ] LR scheduler get_lr() always returns fp32, disregarding default_float. gives default_float to optimizer though
- [x] split forward and backward step jits wastes memory -- with unified, the intermediates are freed inside jit, with split, intermediates are kept around forever

Things that need to be done:
- [x] use HeNormal initialization in Conv2d
- [x] Normal / zeros init for Linear
- [ ] gradient accumulation to help copy hparams
- [ ] pool of JIT-internal buffers, so that multiple jit functions can share the same buffers (get rid of split fw and bw jits and fix tensor.train in eval)
- [ ] make BEAM not allocate buffers when results are already cached
- [ ] FP16
- [ ] automatic nan-detecting LR scaler for FP16
- [ ] hparam sweep
- [ ] limit group buffers even more in beam for HIP, these compiler crashes crash the entire python process, and beam can't continue after that (i think python multiprocessing has a process replacement mechanism we can use though)
- [ ] investigate optimizer backwards passes
- [x] restore rng state after checkpoint
- [ ] pad or fill last batch instead of dropping it
- [ ] sync weights for unsyncbatchnorm
- [ ] wait for HSA to have fast multigpu ;)
- [ ] wait for ring-all-reduce ;)

Things I saw along the way:

- GPUS=2 is basically free, GPUS=6 is fully python-bound, somewhat n^2 scaling
- kernels are queued one gpu at a time, wasting time.
- when restoring from ckpt, beam takes more memory, since on fresh lazy initialization most of the beams run when buffers are not yet realized (maybe we can steal buffers from other places for the read-only ones? lol)
- ##3343 might be a "soft oom" where if you fill gpu memory too much, something internal to hip (or internal to us) doesn't check allocation succeeds and crashes, or something like that. We should consider a `MLIMIT` env var that OOMs inside our own allocator at some soft limit to prevent this, and also prevent OOMs from crashing desktop sessions ;)
- can't change poweroverdrive on tiny9
- tinybox (tiny9) is unstable
  - `HW Exception by GPU node-4 (Agent handle: 0x562a0ae767c0) reason :GPU Hang` after around 12 hours, sporadically
  - RuntimeError: HIP Error 101, invalid device ordinal cf #3440?
  - `Expected integer value from monitor, but got ""`
- conv kernels get only 20TFlops fp32, can be significantly improved with search

 ```
- bs=104
  - gpus=1:
    - 338.52 ms run
    - 5.95fw    4.35bw ms python
    - 321.81 ms CL,   6.40 ms fetch data,  6.99 loss, 0.00 acc, 0.000305 LR, 22.59 GB used,   7628.71 GFLOPS
  - gpus=4:
    - 388.30 ms run
    - 10.49fw  144.14bw ms python
    - 217.95 ms CL,  15.73 ms fetch data,  3.01 loss, 0.05 acc, 0.318207 LR, 91.62 GB used,  26662.02 GFLOPS                 
  - gpus=5:
    - 424.80 ms run
    - 16.23fw  273.14bw ms python
    - 115.39 ms CL,  20.03 ms fetch data,  3.05 loss, 0.05 acc, 0.195437 LR, 114.66 GB used,  30491.95 GFLOPS                
  - gpus=6: (i don't think i picked the best step for this one)
    - 485.21 ms run
    - 22.90fw  431.69bw ms python
    - 3.28 ms CL,  27.33 ms fetch data,  3.04 loss, 0.06 acc, 0.434845 LR, 137.78 GB used,  32062.92 GFLOPS                
```

potential speedup breakdown:
- (1.25, 1.5)x better search (MCTS?)
- 1.5x FP16
- ~1.25x HSA + better multigpu
- ???

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -129,7 +129,8 @@ class Tensor:
     if DEBUG >= 4: print(f"assign {self.lazydata} <- {x.lazydata}")
     if self.dtype == x.dtype and not getenv("DISALLOW_ASSIGN"):
       if isinstance(self.lazydata, MultiLazyBuffer):
-        for d,s in zip(x.lazydata.lbs, self.lazydata.lbs): d.output_buffer = s.base.realized
+        if self.lazydata.axis == x.lazydata.axis:
+          for d,s in zip(x.lazydata.lbs, self.lazydata.lbs): d.output_buffer = s.base.realized
       else:
         if self.lazydata.base.realized is not None: x.lazydata.output_buffer = self.lazydata.base.realized
     self.lazydata = x.lazydata
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -897,11 +897,12 @@ class Tensor:
     y = (self - self.mean(axis, keepdim=True))
     return y.mul((y*y).mean(axis, keepdim=True).add(eps).rsqrt())
 
-  def batchnorm(self, weight:Optional[Tensor], bias:Optional[Tensor], mean:Tensor, invstd:Tensor) -> Tensor:
-    shape = (1, -1) + (1,) * (self.ndim-2)
+  def batchnorm(self, weight:Optional[Tensor], bias:Optional[Tensor], mean:Tensor, invstd:Tensor, axis=1) -> Tensor:
+    shape = self.shape[:axis-1] + tuple(1 if ax != axis else -1 for ax in range(axis-1,self.ndim))
     x = self - mean.reshape(shape)
-    if weight: x = x * weight.reshape(shape)
-    ret = x.mul(invstd.reshape(shape) if len(invstd.shape) == 1 else invstd)
+    if weight:
+      x = x * weight.reshape(shape)
+    ret = x.mul(invstd.reshape(shape) if len(invstd.shape) == axis else invstd)
     return (ret + bias.reshape(shape)) if bias else ret
 
   def dropout(self, p=0.5) -> Tensor:
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -925,12 +925,12 @@ class Tensor:
   def binary_crossentropy_logits(self, y:Tensor) -> Tensor:
     return (self.maximum(0) - y * self + (1 + self.abs().neg().exp()).log()).mean()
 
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

**Comment by wozeparrot** [comment]
LARS could be done like how LAMB is done, with Adam and LAMB having the same implementation just Adam has the trust factor set to 1 always.

File: tinygrad/tensor.py [hunk 1]
```diff
@@ -925,12 +925,12 @@ def binary_crossentropy(self, y:Tensor) -> Tensor:
   def binary_crossentropy_logits(self, y:Tensor) -> Tensor:
     return (self.maximum(0) - y * self + (1 + self.abs().neg().exp()).log()).mean()
 
-  def sparse_categorical_crossentropy(self, Y:Tensor, ignore_index=-1) -> Tensor:
+  def sparse_categorical_crossentropy(self, Y:Tensor, ignore_index=-1, label_smoothing=0) -> Tensor:
```
geohot, line 928: This option needs a test

geohot:
It's crazy how close the train and test accuracy are staying. I would expect train to be higher.

Do we have a reference MLPerf run to compare this to?

geohot:
Closing (superseded by #3573)

PR #3362, opened 2024-02-10
linearizer: fix up edge case bugs in UNROLL opt

flammit:
Fully UNROLLing the first_reduce should not change the number of local_dims.

Fully UNROLLing a GROUP dim should reduce the number of group_for_reduces by one.

Also changed group_for_reduces to be a count as the axis number isn't used anywhere (they are always the first reduce dims).

For testing ran: `PYTHONPATH=. METAL=1 FUZZ_BEAM=1 FUZZ_N=64 DEPTH=2 python3 ./test/external/fuzz_linearizer.py` and no failures except for the expected compare failures with large half accs and GROUP.

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -383,7 +383,7 @@ class TestHandCodedOpts(unittest.TestCase):
     k = Linearizer(s.ast)
     k.hand_coded_optimizations()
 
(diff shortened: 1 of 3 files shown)


geohot:
Sweet!

PR #3373, opened 2024-02-12
uops flop counter

geohot:
Annoying, needs more tests, but way better for WINO and probably a prerequisite for multi output kernels.


tinygrad/device.py
```diff
@@ -49,9 +50,10 @@ def exec(self, rawbufs:List[Buffer], var_vals:Optional[Dict[Variable, int]]=None
   def __call__(self, rawbufs:List[Buffer], var_vals:Dict[Variable, int], wait=False, jit=False) -> Optional[float]:
     raise NotImplementedError("override this")
 
-def update_stats(name:str, op_estimate:sint, mem_estimate:int, var_vals: Optional[Dict[Variable, int]], et: Optional[float], buf_count:int, jit=False, num_kernels=1, lra: Optional[Dict]=None, device:str="", first_run=False):  # noqa: E501
+def update_stats(name:str, op_estimate:sint, mem_estimate:sint, var_vals: Optional[Dict[Variable, int]], et: Optional[float], buf_count:int, jit=False, num_kernels=1, lra: Optional[Dict]=None, device:str="", first_run=False):  # noqa: E501
   if var_vals is None: var_vals = {}
   op_estimate = sym_infer(op_estimate, var_vals)
+  mem_estimate = sym_infer(mem_estimate, var_vals)
```
chenyuxyz, line 56: i think memory should be `int` given that it always allocates an `int` amount of memory even for symbolic size buffer. Like in llama with MAX_CONTEXT = 4096, even if it only generates one token, we want to report memory allocated to the len 4096 kv caches

geohot: This isn't memory allocated, this is memory accessed. GB/s, not GB

PR #3375, opened 2024-02-12
do not truncate float64 precision

### LINKED ISSUE #3304: exp loses precision
When I do exp on a tensor with float64 dtype the result has only around 7-8 digits of precision instead of the around 15-16 expected:

```python
import math

val = 2.1
res = Tensor([val], dtype=dtypes.float64, device="cuda").exp().numpy()[0]
print(f"{res:.16f}")
print(f"{math.exp(val):.16f}")
```

It prints out 
```python
8.1661696836332265
8.1661699125676517
```

The kernel is
```cuda
extern "C" __global__ void E_(double* data0, const double* data1) {
  double val0 = data1[0];
  data0[0] = exp2((val0*(double)(1.4426950408889634f)));
}
```
In the kernel the f at the end of the 1.4426... converts it to a float, losing precision. Without the f the result would be with proper precision.

zku:
fixes #3304

before:
```
$ ipython3 -c "import tinygrad; from tinygrad.dtype import dtypes; tinygrad.Tensor([0.12345678901234567], dtype=dtypes.float64).exp2().numpy().item()"
Out[1]: 1.089341869369704
```

after:
```
$ ipython3 -c "import tinygrad; from tinygrad.dtype import dtypes; tinygrad.Tensor([0.12345678901234567], dtype=dtypes.float64).exp2().numpy().item()"
Out[1]: 1.0893418703580051
```

reference (numpy, same as after):
```
$ ipython3 -c "import numpy as np; np.exp2(np.asarray([0.12345678901234567], dtype=np.float64)).item()"
Out[1]: 1.0893418703580051
```

### DIFF
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -46,7 +46,7 @@ class CStyleLanguage(NamedTuple):
   def render_const(self, x:Union[float,int,bool], var_dtype) -> str:
     if math.isnan(x): val = "NAN"
     elif math.isinf(x): val = ("-" if x < 0 else "") + "INFINITY"
-    else: val = f"{float(x)}" if dtypes.is_float(var_dtype) else f"{int(x)}" if dtypes.is_int(var_dtype) else f"{bool(x)}".lower()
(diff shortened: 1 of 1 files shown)


geohot:
Tests failing

zku:
> Tests failing

Thanks! Working on it.

geohot:
Err, why is this 'l'?

geohot:
Closing, this doesn't make sense.

PR #3383, opened 2024-02-13
(proposal) dynamic cstyle prekernel

Qazalin:
I think DEBUG=4 has become almost unreadable for HIP, and we render WMMA functions when we don't need to.
This PR generates the prekernel based on uops and buf dtypes.


Qazalin:
If the line count change is acceptable I'll break this into 3 PRs for review:
- delete `half_prekernel`
- Metal and HIP WMMA renders
- The rest of HIP (boilerplate, dtypes)

Qazalin:
closing this in favour of the smaller PRs

PR #3384, opened 2024-02-13
Make tests in `test/test_ops.py` pass for Python emulator

obadakhalili:
Can I lock the bounty with this PR?

Notes:
- Right now only 6 tests are failing and most of them are due to only one error. So I figured I could push a PR to lock the bounty, and in the meanwhile, fix the remaining tests.
- I left a few comments on the code which I would appreciate an answer for.

**UPDATE**: There are only 3 remaining failing tests.
**UPDATE**: All tests pass now.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -113,7 +113,6 @@ class TestOps(unittest.TestCase):
     helper_test_op([], lambda: torch.eye(1), lambda: Tensor.eye(1), forward_only=True)
     helper_test_op([], lambda: torch.eye(0), lambda: Tensor.eye(0), forward_only=True)
 
-  @unittest.skipIf(Device.DEFAULT == "PYTHON", "too slow")
   def test_split(self):
     test_cases = [
       (torch.arange(10), Tensor.arange(10), 5),
@@ -502,11 +501,9 @@ class TestOps(unittest.TestCase):
 
   def test_small_cumsum(self):
     helper_test_op([(10)], lambda x: torch.cumsum(x, dim=0), lambda x: Tensor.cumsum(x, axis=0))
-  @unittest.skipIf(Device.DEFAULT == "PYTHON", "too slow")
   def test_simple_cumsum(self):
     helper_test_op([(512)], lambda x: torch.cumsum(x, dim=0), lambda x: Tensor.cumsum(x, axis=0))
     helper_test_op([(1022)], lambda x: torch.cumsum(x, dim=0), lambda x: Tensor.cumsum(x, axis=0))
-  @unittest.skipIf(Device.DEFAULT == "PYTHON", "too slow")
   def test_cumsum(self):
     helper_test_op([(20)], lambda x: torch.cumsum(x, dim=0), lambda x: Tensor.cumsum(x, axis=0))
     helper_test_op([(20,30)], lambda x: torch.cumsum(x, dim=0), lambda x: Tensor.cumsum(x, axis=0))
@@ -528,7 +525,6 @@ class TestOps(unittest.TestCase):
     helper_test_op([(10,20)], lambda x: x.argmin(1, False), forward_only=True)
     helper_test_op([(10,20)], lambda x: x.argmin(1, True), forward_only=True)
 
-  @unittest.skipIf(Device.DEFAULT == "PYTHON", "too slow")
   def test_einsum(self):
     # matrix transpose
     helper_test_op([(150,150)], lambda a: torch.einsum('ij->ji', a), lambda a: Tensor.einsum('ij->ji', a))
@@ -586,7 +582,6 @@ class TestOps(unittest.TestCase):
       Tensor.einsum('ij,jk->ij', a)
 
   @unittest.skipIf(IMAGE>0, "no 1d dot for images")
-  @unittest.skipIf(Device.DEFAULT == "PYTHON", "too slow")
   def test_dot_1d(self):
     helper_test_op([(65), (65)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-4)
     helper_test_op([(65), (65,45)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-4)
@@ -646,7 +641,6 @@ class TestOps(unittest.TestCase):
     helper_test_op(None, lambda x,y: x.matmul(y), lambda x,y: x@y, atol=1e-3, vals=[np.eye(8).astype(np.float32), np.eye(8).astype(np.float32)])
   def test_gemm(self):
     helper_test_op([(64,64), (64,64)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-3)
-  @unittest.skipIf(Device.DEFAULT == "PYTHON", "too slow")
   def test_big_gemm(self):
     helper_test_op([(256,256), (256,256)], lambda x,y: x.matmul(y), Tensor.dot, atol=1e-3)
   @unittest.skipIf(IMAGE>0, "no 0 in shape matmul on images")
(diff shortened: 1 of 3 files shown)


tinygrad/runtime/ops_python.py
```diff
@@ -107,7 +112,7 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
           else:
             ul[i] = [arg] * warp_size
         elif uop is UOps.LOOP:
-          if i not in ul:
+          if i not in ul or ul[i][0] == inp[1][0]:
             ul[i] = [0] * warp_size
```
obadakhalili, line 116: This doesn't seem to be breaking any test, but shouldn't the initial value be in `inp[0][0]`?

obadakhalili:
Regarding the `if` block in [line 120](https://github.com/tinygrad/tinygrad/pull/3384/files#diff-00bb76cdab6aa6dccea93689ae8dc2eb92343681f52231909495af017800628aR120), shouldn't this block be intended back to also check after initialization of loop state, in case the loop shouldn't be entered in the first place. Note though that this also doesn't seem to break any test. Also, implementing my suggestion would require containing the `loop_ends` in a different way.

tinygrad/runtime/ops_python.py
```diff
@@ -82,18 +85,20 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
           loop_ends[idp[0]] = i
           i = idp[0]
           continue
-        elif uop is UOps.BARRIER:
+        # TODO: implement UOps.IF and UOps.ENDIF
+        elif uop in (UOps.BARRIER, UOps.IF, UOps.ENDIF):
```
obadakhalili, line 89: I had to skip `UOps.IF` and `UOps.ENDIF` temporarily similar to `UOps.BARRIER` so I could focus on another fix, though to my surprise this didn't cause any test to fail. So I pushed this for now while I investigate it further.

tinygrad/runtime/ops_python.py
```diff
@@ -82,18 +85,20 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
...
+
         assert dtype is not None, f"{uop} is missing a dtype"
         dl[i] = dtype
         if uop is UOps.DEFINE_GLOBAL:
           assert dtype.fmt is not None
           ul[i] = [pbufs.pop(0).cast(dtype.fmt)] * warp_size
         elif uop is UOps.DEFINE_LOCAL:
           assert dtype.fmt is not None
-          lbuf = memoryview(bytearray(arg[1]*dtype.sz))
+          lbuf = memoryview(bytearray(arg[1]*dtype.itemsize))
```
geohot, line 101: This fix is definitely correct, would merge this alone right away.

geohot:
Can you add test_ops to CI?

obadakhalili:
@geohot I identified 3 tests that make the CI run for a long time when using the Python backend. Together with all the tests they take 1 hour to run on a personal device, and ~20 minutes without these 3 tests. I can skip them when `PYTHON=1` if you want.

obadakhalili:
I skipped the 3 tests for now so you can inspect the rest. There are 3 remaining failing tests, which I'll start working on tomorrow morning.

geohot:
Tests failing

obadakhalili:
All tests pass now @geohot. And a lot of `test/test_dtype.py` tests pass too.

tinygrad/runtime/ops_python.py
obadakhalili: Fixes commits are mostly independent, so I can cherry-pick this and push it as a separate PR if you want to. What about other fixes?

geohot:
Bounty locked

tinygrad/runtime/ops_python.py
geohot: Yea, it's actually okay to not have IF/ENDIF, it's an optional performance improvement for GPU, not a requirement.

BARRIER isn't needed because it's always in sync in this interpreter

tinygrad/runtime/ops_python.py
geohot: Oh it should be, good catch.

tinygrad/runtime/ops_python.py
```diff
@@ -137,6 +147,8 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
                 else: ret.append(_load(m, ox*4 + oy*dtp[0].shape[1]*4 + j))
               ul[i].append(ret)
           elif dtype.sz > 1:
+            if len(inp) == 4 and dtp[3].sz > 1:
```
geohot, line 150: Hmm, this looks sketchy.

tinygrad/runtime/ops_python.py
```diff
@@ -82,32 +85,39 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
           loop_ends[idp[0]] = i
           i = idp[0]
           continue
-        elif uop is UOps.BARRIER:
+        # TODO: implement UOps.IF and UOps.ENDIF
+        elif uop in (UOps.BARRIER, UOps.IF, UOps.ENDIF):
           # in the python emulator, the warp is always in sync
           i += 1
           continue
+
```
geohot, line 93: Don't change whitespace if you don't have to

tinygrad/runtime/ops_python.py
obadakhalili: I see. Thanks for clarifying this.

tinygrad/runtime/ops_python.py
obadakhalili: - This is necessary for nested loops. Without it, you will enter the nested loop for the second time, but its state will still be pointing to `inp[1][0]`. So it is necessary to initialize the loop on first entrance, and to reinitialize on first reuse.
- Thanks, I will push a comment with the fix.

obadakhalili:
@geohot I updated `test/test_ops.py` to skip the least amount of tests in CI that keep the runtime under 5 minutes.

geohot:
Don't spam test_ops like that, select the tests you want from the command line

tinygrad/runtime/ops_python.py
geohot: Yea I don't think this is right

obadakhalili: Yes, I'm looking into it right now.

obadakhalili: I changed this in my second to last commit. The problem I was trying to solve is when gated `UOps.LOAD` reference values that have `.sz` > 1. I addressed that by sending the `gate`/`default` that corresponds to the current `UOps.LOAD` item.

test/test_ops.py
```diff
@@ -1373,19 +1373,21 @@ def test_maxpool2d_bigger_stride(self):
           lambda x: torch.nn.functional.max_pool2d(x, kernel_size=(2,2), stride=stride),
           lambda x: Tensor.max_pool2d(x, kernel_size=(2,2), stride=stride))
 
-  @unittest.skipIf(Device.DEFAULT == "CUDA", "CUDA fails on this")
+  @unittest.skipIf(Device.DEFAULT in ["CUDA", "PYTHON"], "CUDA fails on this, and PYTHON is too slow")
```
geohot, line 1376: No need for this, these tensor sizes are stupidly big. Want to submit another PR making them smaller? Say:

(8, 2, 17, 14) or something

obadakhalili: #3417

tinygrad/runtime/ops_python.py
```diff
@@ -100,15 +103,20 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
...
+            ul[i] = [casted_arg] * warp_size
         elif uop is UOps.DEFINE_ACC:
           if dtype.count > 1:
             ul[i] = [[arg] * warp_size for _ in range(dtype.count)]
           else:
             ul[i] = [arg] * warp_size
         elif uop is UOps.LOOP:
-          if i not in ul:
-            ul[i] = [0] * warp_size
+          if i not in ul or ul[i][0] == inp[1][0]:
```
geohot, line 118: I don't understand this or

obadakhalili: There are two cases that you need to (re)initialize the loop for:
- The first is when you enter the loop for the first time. This is initialization because `ul` won't have any state about this loop.
- The second is when you enter the loop again after you have finished all its iterations. This only happens in nested loops. When the nested loop executes its last iteration, the loop will be entered again because of the next iteration of the upper loop, which will be problematic because the state of the previous iteration of the nested loop is still saved in `ul` and needs to be reinitialized to `inp[0][0]`. But now while explaining it, I realize that the best solution is to clear the state of the loop after the last iteration. I will push a commit with that.

geohot:
PayPal or USDC for your $100? E-mail george@tinygrad.org with email/address to claim.

Max-We:
Are you sure that this is fixed? Running `PYTHON=1 python3 test/test_ops.py -f` on my machine / Google Colab throws and error. Of course, without the `PYTHON=1` flag it works, but that's not the point, no?

iglesias:
"$100 bounties are a few line change and could be 10 minutes if you see it" 

hmm, I think someone is not being precisely candid,

PR #3385, opened 2024-02-13
Implement Python runtime ops (test_ops.py)

xarkes:
Hello,
I've seen in the PRs this topic is getting some interest so I am pushing my current implementation although it is not cleaned up yet.

```
$ PYTHON=1 python3 ./test/test_ops.py
...
Ran 245 tests in 5769.725s

OK (skipped=4)
```

Let me know what you think of it, as I am new to this project I hope my understanding of it is good enough.
I will do some refactor a bit later to remove some useless lines.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,6 +46,8 @@ jobs:
       run: |
         DEBUG=2 EMULATE_METAL=1 FORWARD_ONLY=1 PYTHON=1 python3 ./test/test_linearizer.py TestLinearizer.test_tensor_cores
         DEBUG=2 EMULATE_HIP=1 FORWARD_ONLY=1 PYTHON=1 python3 ./test/test_linearizer.py TestLinearizer.test_tensor_cores
+    - name: Test ops (Python)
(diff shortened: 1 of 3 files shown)


xarkes:
Rebased and cleaned up I believe.
The performances are a bit concerning but I think that's to be expected with that many operations.
Feel free to merge whatever you think is valid. Cheers!

xarkes:
Note that the `UOps.IF` implementation is wrong at this stage (eg. nested ifs are not handled - a test should be added for that, but I have no clue how to generate those) and that likely the store implementation is too hacky. KamiKomplex504 gave some hints about how to handle it, and I guess the open dtypes PR may be relevant as well.

geohot:
Tests failing.

xarkes:
`Error: The operation was canceled.`
I think that's because there is a timeout of some sorts on the CI. Maybe I should disable some for the python testing.

xarkes:
Just noticed https://github.com/tinygrad/tinygrad/pull/3384 is finally green too.
From the diff their implementation is likely more accurate so feel free to close this PR.

geohot:
If you think theres is better, I'll close

PR #3388, opened 2024-02-14
delete half_prekernel

Qazalin:
Replaces `half_prekernel` with a generic render_kernel extension for dtypes+uops.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,7 +76,7 @@ jobs:
     - name: Lint tinygrad with pylint
       run: python -m pylint tinygrad/
     - name: Run mypy
-      run: python -m mypy
+      run: python -m mypy --strict-equality
     - name: Test Docs
       run: |
         python docs/abstractions.py
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -231,7 +231,7 @@ class CUDALanguage(CStyleLanguage):
     if any(uop.dtype == dtypes.half for uop in uops):
       prefix += ["#include <cuda_fp16.h>", "struct half4 { half x, y, z, w; };",
       "__device__ half4 make_half4(half x, half y, half z, half w) { half4 ret; ret.x = x; ret.y = y; ret.z = z; ret.w = w; return ret; }"]
-    if any(uop.uop == dtypes.bfloat16 for uop in uops): prefix.append("#include <cuda_bf16.h>")
+    if any(uop.dtype == dtypes.bfloat16 for uop in uops): prefix.append("#include <cuda_bf16.h>")
     return super().render_kernel(function_name, kernel, bufs, local_size, uops, prefix=prefix)
 CUDARenderer = functools.partial(uops_to_cstyle, CUDALanguage())
 
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -231,7 +231,7 @@ class CUDALanguage(CStyleLanguage):
     if any(uop.dtype == dtypes.half for uop in uops):
       prefix += ["#include <cuda_fp16.h>", "struct half4 { half x, y, z, w; };",
       "__device__ half4 make_half4(half x, half y, half z, half w) { half4 ret; ret.x = x; ret.y = y; ret.z = z; ret.w = w; return ret; }"]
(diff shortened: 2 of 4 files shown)


tinygrad/renderer/cstyle.py
```diff
@@ -187,6 +184,10 @@ class OpenCLLanguage(CStyleLanguage):
   type_map = { dtypes.uint8: "uchar", dtypes.uint32: "uint", dtypes.uint16: "ushort", dtypes.uint64: "ulong" }
   def render_cast(self, x, var_dtype, bitcast=False) -> str:
     return f"as_{self.type_map.get(var_dtype) or var_dtype.name}({x[0]})" if bitcast else super().render_cast(x, var_dtype)
+
+  def render_kernel(self, function_name, kernel, bufs, local_size, uops, prefix=None) -> str:
+    if any(dt == dtypes.half for _, dt in bufs): prefix = ["#pragma OPENCL EXTENSION cl_khr_fp16 : enable"]
```
geohot, line 189: I don't think this logic is right. This is only about doing math on fp16, not fp16 buffers.

Qazalin:
yeah half_prekernel was actually wrong if you had a half midcast but no bufs, should use uops

tinygrad/renderer/cstyle.py
```diff
@@ -215,19 +216,23 @@ def render_cast(self, x: List[str], var_dtype: DType, bitcast=False) -> str:
...
-    __device__ half4 make_half4(half x, half y, half z, half w) { half4 ret; ret.x = x; ret.y = y; ret.z = z; ret.w = w; return ret; }
-  """
   type_map = {dtypes.bfloat16: "nv_bfloat16"}
+
+  def render_kernel(self, function_name, kernel, bufs, local_size, uops, prefix=None):
+    prefix = ["#define INFINITY (__int_as_float(0x7f800000))","#define NAN (__int_as_float(0x7fffffff))"]
+    if any(uop.dtype == dtypes.half for uop in uops):
+      prefix += ["#include <cuda_fp16.h>", "struct half4 { half x, y, z, w; };",
+      "__device__ half4 make_half4(half x, half y, half z, half w) { half4 ret; ret.x = x; ret.y = y; ret.z = z; ret.w = w; return ret; }"]
+    if any(uop == dtypes.bfloat16 for uop in uops): prefix.append("#include <cuda_bf16.h>")
```
geohot, line 234: How does this line possibly pass mypy? In addition to fixing this line, make sure the type checker catches this.

.github/workflows/test.yml
```diff
@@ -76,7 +76,7 @@ jobs:
     - name: Lint tinygrad with pylint
       run: python -m pylint tinygrad/
     - name: Run mypy
-      run: python -m mypy
+      run: python -m mypy --strict-equality
```
Qazalin, line 79: `--strict-equality` catches always false `==` checks like that uops err:
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict-equality

tinygrad/runtime/ops_gpu.py
```diff
@@ -23,8 +23,8 @@ def render(self, name:str, uops) -> str: return OpenCLRenderer(name, uops)
   def compile(self, src:str) -> bytes:
     program = checked(cl.clCreateProgramWithSource(self.device.context, 1, to_char_p_p([prg_bytes := src.encode()]),
                                                   ctypes.byref(ctypes.c_size_t(len(prg_bytes))), ctypes.byref(status := ctypes.c_int32())), status)
-    status = cl.clBuildProgram(program, 1, ctypes.byref(self.device.device_id), None, cl.clBuildProgram.argtypes[4](), None)
-    if status != 0:
+    build_status: int = cl.clBuildProgram(program, 1, ctypes.byref(self.device.device_id), None, cl.clBuildProgram.argtypes[4](), None)
```
Qazalin, line 26: mypy's `--strict-equality` complains about this because in line 25 `status` is a ctypes.c_int32()

geohot: Oh, this is actually a bug. Should be `status.value == 0`

Qazalin: status was int when i tested though?

Qazalin: i tried doing status.value and actually it doesn't exist, the correct way was to do status != 0 (mac)

PR #3401, opened 2024-02-14
dtypes: delete the float cast in realize.py

Qazalin:
in realize.py we force the ConstBuffer's arg to be float. I think this is wrong, it has lead to ASTs like this:

`Tensor([1,2,3]) + 2`
```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ ADD 
  2    ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  3    ┗━━ CONST ConstBuffer(val=2.0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
  ```
  
when I removed that float cast, it uncovered issues upstream where we create these const buffers, eg. tensor.py's _broadcast and lazy.py's const could return bufs like this:

`ConstBuffer(val=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(), strides=(), offset=0, mask=None, contiguous=True),)))`

now the AST is correct:
```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ ADD 
  2    ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  3    ┗━━ CONST ConstBuffer(val=2, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
 ```

### DIFF
--- a/test/test_lazybuffer.py
+++ b/test/test_lazybuffer.py
@@ -2,6 +2,7 @@
 import numpy as np
 import unittest
 from tinygrad import Tensor, Device, dtypes
+from tinygrad.lazy import LazyBuffer
 
 class TestLazyBuffer(unittest.TestCase):
   def test_fromcpu_shape_tracker(self):
@@ -56,6 +57,15 @@ class TestLazyBuffer(unittest.TestCase):
     b = Tensor.zeros(4,1,4)
     c = a.cat(b, dim=1)
     np.testing.assert_allclose(c.numpy(), np.concatenate((a.numpy(), b.numpy()), axis=1))
+
+  def test_const_dtype(self):
+    lb: LazyBuffer = Tensor([1], dtype=dtypes.int).lazydata
+    assert lb.const(1).base.arg == 1
+    assert type(lb.const(1).base.arg) == int
+
+    lb: LazyBuffer = Tensor([1], dtype=dtypes.float).lazydata
+    assert lb.const(1).base.arg == 1.0
+    assert type(lb.const(1).base.arg) == float
 
 if __name__ == "__main__":
   unittest.main()
--- a/test/test_lazybuffer.py
+++ b/test/test_lazybuffer.py
@@ -61,11 +61,11 @@ class TestLazyBuffer(unittest.TestCase):
   def test_const_dtype(self):
     lb: LazyBuffer = Tensor([1], dtype=dtypes.int).lazydata
     assert lb.const(1).base.arg == 1
-    assert type(lb.const(1).base.arg) == int
+    assert type(lb.const(1).base.arg) is int
 
     lb: LazyBuffer = Tensor([1], dtype=dtypes.float).lazydata
     assert lb.const(1).base.arg == 1.0
-    assert type(lb.const(1).base.arg) == float
+    assert type(lb.const(1).base.arg) is float
 
 if __name__ == "__main__":
   unittest.main()
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -90,7 +90,7 @@ class Linearizer(Kernel):
     invalid_value = 0 if dtypes.is_int(buf.dtype) else 0.0
     for idx, valid, rep_idx in zip(e_idxs, e_valids, iter_idxs(expand_vars)):
       this_const, idx, valid = (invalid_value, NumNode(0), NumNode(1)) if valid.max == 0 else (const, idx, valid)
-      key = f"{acc}{localtype}{this_const if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
+      key = f"{acc}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
       if key not in self.load_cache:
         if acc is not None:
           self.load_cache[key] = self.uop(UOps.DEFINE_ACC, localtype, (), this_const, cachable=False)
--- a/tinygrad/dtype.py
+++ b/tinygrad/dtype.py
@@ -38,7 +38,7 @@ class PtrDType(DType):
   def __eq__(self, dt): return self.priority==dt.priority and self.itemsize==dt.itemsize and self.name==dt.name and self.sz==dt.sz
   def __ne__(self, dt): return not (self == dt)
 
-def cast_scalar(scalar: Scalar, dtype:DType): return int(scalar) if dtypes.is_int(dtype) else float(scalar)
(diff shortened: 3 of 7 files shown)


Qazalin:
pending work left is in the linearizer.
If you have a ConstBuffer with val=1 _and_ a global load `*(data1+0)`, ast_parse's `loaded_buffers[x.arg]` returns the global load, not Uops.CONST itself.

before, there was no conflict between global loads and ConstBuffers because they had different keys:
```
LOAD MemBuffer(idx=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(10, 1), strides=(0, 0), offset=0, mask=None, contiguous=False),)))
CONST ConstBuffer(val=1.0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(11, 19), strides=(0, 0), offset=0, mask=((0, 11), (9, 19)), contiguous=False), View(shape=(10, 10), strides=(1, 20), offset=0, mask=None, contiguous=False))))
```
now there's a conflict:
```
LOAD MemBuffer(idx=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(10, 1), strides=(0, 0), offset=0, mask=None, contiguous=False),)))
CONST ConstBuffer(val=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(11, 19), strides=(0, 0), offset=0, mask=((0, 11), (9, 19)), contiguous=False), View(shape=(10, 10), strides=(1, 20), offset=0, mask=None, contiguous=False))))
```

so the AST:
```
 10      ┣━┳ ADD 
 11      ┃ ┣━━ LOAD MemBuffer(idx=2, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(10, 1), strides=(1, 0), offset=0, mask=None, contiguous=True),)))
 12      ┃ ┗━━ CONST ConstBuffer(val=-1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(10, 1), strides=(0, 0), offset=0, mask=None, contiguous=False),)))
 13      ┗━━ CONST ConstBuffer(val=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(10, 1), strides=(0, 0), offset=0, mask=None, contiguous=False),)))
 ```
 
 gets UOped as:
 `int val0 = *(data1+0); val1+(-1)+val0`
 instead of:
 `val1+(-1)+1 `

File: tinygrad/dtype.py [hunk 1]
```diff
@@ -38,6 +38,8 @@ def __hash__(self): return super().__hash__()
   def __eq__(self, dt): return self.priority==dt.priority and self.itemsize==dt.itemsize and self.name==dt.name and self.sz==dt.sz
   def __ne__(self, dt): return not (self == dt)
 
+def cast_scalar(scalar: Scalar, dtype:DType): return int(scalar) if dtypes.is_int(dtype) else float(scalar)
```
chenyuxyz, line 41: can scalar be bool?

Qazalin:
pending is adding unit tests now - linearizer is fixed!

test/test_lazybuffer.py
```diff
@@ -57,5 +58,14 @@ def test_shrink_const_then_cast(self):
...
     np.testing.assert_allclose(c.numpy(), np.concatenate((a.numpy(), b.numpy()), axis=1))
 
+  def test_const_dtype(self):
+    lb: LazyBuffer = Tensor([1], dtype=dtypes.int).lazydata
+    assert lb.const(1).base.arg == 1
+    assert type(lb.const(1).base.arg) == int
+
+    lb: LazyBuffer = Tensor([1], dtype=dtypes.float).lazydata
+    assert lb.const(1).base.arg == 1.0
+    assert type(lb.const(1).base.arg) == float
```
Qazalin, line 68: this fails in master:
```
    assert type(lb.const(1).base.arg) == float
AssertionError
```

Qazalin: it'd be very tedious to pick 1.0 vs 1 in mlops itself, i think LazyBuffer should just default to whatever the lb's dtype already is

tinygrad/codegen/linearizer.py
```diff
@@ -90,7 +90,7 @@ def global_load(self, i:int, idxs:List[Node], acc=None, barrier:Optional[UOp]=No
     invalid_value = 0 if dtypes.is_int(buf.dtype) else 0.0
     for idx, valid, rep_idx in zip(e_idxs, e_valids, iter_idxs(expand_vars)):
       this_const, idx, valid = (invalid_value, NumNode(0), NumNode(1)) if valid.max == 0 else (const, idx, valid)
-      key = f"{acc}{localtype}{this_const if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
+      key = f"{acc}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
```
Qazalin, line 93: cache keys can have conflicts between ConstBuffers and MemBuffers, see:
https://github.com/tinygrad/tinygrad/pull/3401#issuecomment-1944684081

we didnt have this problem before because MemBuffers are indexed with ints and [ConstBuffers where always float](https://github.com/tinygrad/tinygrad/pull/3401/files#diff-4fe1256664cc5870d2d44de21d5f463a512f81729a13f5309e272c8f7793a039L96)

geohot:
Good fix, yea that cast to float was def wrong.

chenyuxyz:
lol was reviewing this too since i was requested. The fix looks good but I could not quite understand what the test `test_load_cache_const_bufs` was trying to test

Qazalin:
@chenyuxyz try running test_indexing without the buffer load cache key fix in linearizer with DEBUG=4, compare the AST with the kernels that get generated.
here's what you'll get:
https://gist.github.com/Qazalin/96e7e6f932002d812ca38234762180f5
I could either hard-code the exact AST that caught this bug or unit test load_cache itself. I picked the isolated test, because that AST also complicates stuff with reduce ops and earlybufs.

chenyuxyz:
yea i understand the motivation, but confused by the test itself

it says `# make sure const buffers are differentiated from local and mem buffers` but it's not clear on the asserted quantity (cache_keys len 5, 2 have "CONST"). This tests the key implementation and i don't quite see the behavior (const buffers are differentiated from local and mem buffers) that's being tested?

PR #3404, opened 2024-02-15
run test_indexing on remu

Qazalin:
follow-up on the int4 bug report:
https://discord.com/channels/1068976834382925865/1069001075828469790/1207509674904129576

Ran on a local github actions runner and I think remu is fast enough.


geohot:
Thanks!

PR #3405, opened 2024-02-15
correct division dtype casting

geohotstan:
make #3195 great again

the test_dtype tests follows jax casting rules


geohot:
@chenyuxyz is this right?

chenyuxyz:
high level seems fine. I will review this one once it's ready. And would be great if the onnx fix can be in this pr too so we don't break onnx div

geohotstan:
The Div op for onnx has a special extra cast that isn't there for other ops like Mul or Add
see [Div](https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_div.py) compared with [Mul](https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_mul.py) or [Add](https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_add.py)

So I think the extra cast is proper?
The documentation onnx gives is so scant... :c

PR #3411, opened 2024-02-15
Rename .sz for .count on dtype (and ANETensor for completeness)

fijal:
This renames .sz for .count on dtype, as requested on discord

### DIFF
--- a/extra/accel/ane/lib/ane.py
+++ b/extra/accel/ane/lib/ane.py
@@ -112,16 +112,16 @@ class ANETensor:
   def __init__(self, *shape):
     self.shape = shape
     self.dtype = np.float16
-    self.count = int(np.prod(shape))
-    assert(self.count <= 0x4000)
-    self.tt = libane.ANE_TensorCreate(self.count, 1)
+    self.sz = int(np.prod(shape))
+    assert(self.sz <= 0x4000)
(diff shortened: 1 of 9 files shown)


extra/accel/ane/lib/ane.py
```diff
@@ -112,16 +112,16 @@ class ANETensor:
   def __init__(self, *shape):
     self.shape = shape
     self.dtype = np.float16
-    self.sz = int(np.prod(shape))
-    assert(self.sz <= 0x4000)
-    self.tt = libane.ANE_TensorCreate(self.sz, 1)
+    self.count = int(np.prod(shape))
```
geohot, line 115: WTF

geohot:
Don't change extra

fijal:
Reverted the "ane" changes. Should I also not touch the extra/backends/cstyle.py?

Xiloscient:
none of my business but the making the linter happier by parenthesizing a condition in order to split it into two lines thereby both decreasing readiblity and increasing line count doesn't seem like a good idea

fijal:
I do tend to agree, alternative would be to e.g. use a temp variable, but that would still increase the line count or to disable linter for this line. I'm not familiar with the codebase, so please suggest some solution and I'm happy to do it

geohot:
Extra is still being changed

PR #3414, opened 2024-02-15
dynamic prekernel for HIP

Qazalin:
- generates boilerplates in loops as much as possible
- renders dtypes defs and ALU functions based on uops (float always renders)


Qazalin:
needs to be smaller for review

PR #3415, opened 2024-02-15
Rename .sz to .count

Xiloscient:
Decided to ignore the 150 line length limit on:
`return self.render_cast([val]*var_dtype.count, var_dtype) if var_dtype.count > 1 or var_dtype not in [dtypes.float, dtypes.int, dtypes.bool] else val`
if unacceptable will fix wihout line count increase.


Xiloscient:
Alternative to the linting disable is a rename to "cnt" instead of "count"
will do so if deemed acceptable

PR #3420, opened 2024-02-15
If feasible, do not truncate float64 down to float32 in cstyle renderer

zku:
Follow-up from https://github.com/tinygrad/tinygrad/pull/3375 (issue #3304).

* Remove the long double suffix `l` - initially introduced to avoid function overload ambiguity (e.g. `error: call to 'mad' is ambiguous`) but not needed.
* Add test to show the increased precision when working with float64; that test fails if we keep the `f` suffix for all float dtypes due to f64 truncate. Test is disabled for CUDACPU (does not support `half` used by various functions like `sin` which leads to segfaults in gpuocelot, as documented in test_dtype_alu and test_dtype - see comments below) and HIP.

Please let me know if you'd like to see more / different tests for this float64 precision change.


zku:
Checking test failures now.

zku:
Metal tests pass, need to figure out why we only have float32 precision on hip test[0] and why cuda test segfaults.

[0]
```
E           Max absolute difference: 1.87374919e-06
E           Max relative difference: 9.32884791e-08
```

zku:
Locally I can reproduce a segfault like this:

```
env PYTHONPATH="." LD_LIBRARY_PATH=$HOME/gpuocelot/ocelot/build/ CUDA=1 CUDACPU=1 OPT=2 JIT=1 python3 test/test_dtype.py TestDoubleDtype
```

```
.terminate called after throwing an instance of 'executive::RuntimeException'
  what():  conversion not implemented
```

https://github.com/gpuocelot/gpuocelot/blob/c476bcb605579b123b85d34afef2c848127e8cdc/ocelot/src/executive/CooperativeThreadArray.cpp#L3747

zku:
The offending instr.type is 9, which is f16. That aligns with the comment in the dtypes support check `  # CUDA in CI uses CUDACPU that does not support half`, need to disable test for that case.

zku:
Next failure is HIP/HIPCPU - somehow we only get f32 precision. Investigating.

zku:
Ah, the HIP renderer in cstyle.py does not support f64, hence truncated to f32. Disabling high precision test for HIP.

tinygrad/renderer/cstyle.py
```diff
@@ -45,7 +45,7 @@ def render_cast(self, x:List[str], var_dtype:DType, bitcast=False) -> str:
   def render_const(self, x:Union[float,int,bool], var_dtype) -> str:
     if math.isnan(x): val = "NAN"
     elif math.isinf(x): val = ("-" if x < 0 else "") + "INFINITY"
-    else: val = f"{float(x)}f" if dtypes.is_float(var_dtype) else f"{int(x)}" if dtypes.is_int(var_dtype) else f"{bool(x)}".lower()
+    else: val = f"{float(x)}" if var_dtype == dtypes.float64 else f"{float(x)}f" if dtypes.is_float(var_dtype) else f"{int(x)}" if dtypes.is_int(var_dtype) else f"{bool(x)}".lower()  # noqa: E501
```
geohot, line 48: No new lines with noqa E501

zku: Thanks for reviewing. Removed the noqa E501 exception.

PR #3423, opened 2024-02-15
Remove Interpreted device & remaining CPU/TORCH ref

xarkes:
Let's see if I understood the assignment


xarkes:
Green now, you may want to check the doc updates and the onnx `supports_device` function, I'm not sure you want it to run only on clang but from the previous run it looks like metal/llvm were not passing.

PR #3425, opened 2024-02-16
ops_python Stronger typing and improvements to alu

unknownusername504:
This has some work arounds that should be removed with further improvements. Extensive changes to dtypes class. Minor improvements to the alu but inferred/casted types should be handled better to reduce most of these changes.

I'm sure a the amount changed will have this rejected but a reference for anyone that is working on this.


geohot:
Huge number of lines. No new tests means there's no new functionality.

PR #3428, opened 2024-02-16
Add Node.val for NumNode, MulNode and SumNode

chenyuxyz:
Used to obstain int size for symbolic shape that can be checked in reshape.


chenyuxyz:
hmm we also have reshape inside kernel that's not for bound variables

PR #3429, opened 2024-02-16
Python uop emulator passes test_dtype

Ashad-h:
The test_dtype now passes with Python emulator

I used ctypes to do casting and struct to do bitcasting.

Is it valid ?


geohot:
You don't need all those lines.

Ashad-h:
Is it better like this ?

PR #3430, opened 2024-02-16
limit sint to be Union[int, Variable, MulNode, SumNode]

chenyuxyz:
these are the only allowed nodes in a Tensor shape


chenyuxyz:
merging this, working on separating symbolic for shape v.s. symbolic for memory indexing. Considering a specialized MulNode and SumNode for symbolic shape only, so the base index ones can be simpler

PR #3434, opened 2024-02-17
delete MovementOps

Qazalin:
tinygrad is Compiled-only now, we don't use movement ops


geohot:
Looks good, but don't delete extra/to_movement_ops.py. Move the definition of movementops there, the concept could be useful in the future.

geohot:
I think MULACC can go too.

PR #3436, opened 2024-02-17
Make tests in `test/test_dtype.py` pass for Python emulator

obadakhalili:
This is more of a proposal to solve the casting problem in the Python emulator.

The proposal suggests using Numpy to handle casting.

My case:
- While bit casting can be implemented easily using Python's `struct` module, the real hassle is with type casting.
- Python's type casting can handle simple cases like casting a float to an integer, but it doesn't have support for c-style type casting which can handle richer cases like converting an unsigned integer to an integer.
- Implementing complete c-style type casting in native Python a) requires more undesired extra lines and b) unnecessary.
- Exporting the problem to Numpy makes the code shorter, cleaner, and of course complete.
- I also believe that this implementation has no trade-offs because Numpy won't be required as a dependency, since it will be only necessary to run the code against the Python backend which is used for development, and doesn't interfere with the runtime's purpose to act as the "living specs of uops".


obadakhalili:
I have a feeling I might be mistaken in my assumption "it doesn't interfere with the runtime's purpose to act as the living specs of uops". I would appreciate explaining why.

PR #3437, opened 2024-02-18
realize unsafe pads even if they don't expand the buffer

chaosagent:
a test and a quick fix.

checking for unsafe pads can just be its own pass.

need to think about this expand rule!

### DIFF
--- a/tinygrad/realize.py
+++ b/tinygrad/realize.py
@@ -147,13 +147,13 @@ def _recurse_lb(buf:LazyBuffer, realizes:Set[LazyBuffer], allbufs:Dict[LazyBuffe
     if DEBUG >= 3: print(f"forcing image {buf.dtype} with shape {buf.shape} to float32")
     buf.dtype = dtypes.float32  # NOTE: this is what makes the dtype above not match
   if buf.base != buf:
-    # realize all places where the buffer is expanded
-    if prod(buf.base.st.shape) < prod(buf.st.shape):
-      if len(buf.st.views) == 1 and buf.st.views[-1].mask and all_int(buf.base.st.shape) and \
-          prod(buf.base.st.shape) >= prod([y-x for x,y in buf.st.views[-1].mask]):
-        simple_pads.add(buf.base)
-      else:
-        realizes.add(buf.base)
+    # defer checking for unsafe pads
+    if any(v.mask is not None for v in buf.st.views): simple_pads.add(buf.base)
+    # realize all places where the buffer is expanded (eliding if the expansion is just padding)
+    if prod(buf.base.st.shape) < prod(buf.st.shape) and not \
+      (len(buf.st.views) == 1 and buf.st.views[-1].mask and all_int(buf.base.st.shape) and \
+        prod(buf.base.st.shape) >= prod([y-x for x,y in buf.st.views[-1].mask])):
+      realizes.add(buf.base)
     return _recurse_lb(buf.base, realizes, allbufs, simple_pads, children)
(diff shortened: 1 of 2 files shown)


tinygrad/realize.py
```diff
@@ -139,22 +139,22 @@ def _recursive_schedule(out:LazyBuffer, seen:Set[LazyBuffer], realizes:Set[LazyB
...
-    # realize all places where the buffer is expanded
-    if prod(buf.base.st.shape) < prod(buf.st.shape):
-      if len(buf.st.views) == 1 and buf.st.views[-1].mask and all_int(buf.base.st.shape) and \
-          prod(buf.base.st.shape) >= prod([y-x for x,y in buf.st.views[-1].mask]):
-        simple_pads.add(buf.base)
-      else:
-        realizes.add(buf.base)
-    return _recurse_lb(buf.base, realizes, allbufs, simple_pads, children)
+    # defer checking for unsafe pads
+    if any(v.mask is not None for v in buf.st.views): pad_srcs.add(buf.base)
```
geohot, line 151: Why do I need to check if it's already added to realize?

geohot:
What's the minimal change that does this?

tinygrad/realize.py
chaosagent: pads don't necessarily expand. I don't see anywhere else that realizes unsafe pads.

geohot:
Why is the PR more complex than the simplest it can be?

chaosagent:
the only additional change was renaming simple_pads -> pad_srcs, since not only simple pads are being added to the set now

geohot:
You've put a whole bunch of PRs up that I haven't merged. If I can't read it, I can't merge it.

If you want this merged, revert the name change, do the absolute minimal change, and add a regression test for the changed behavior.

tinygrad/realize.py
```diff
@@ -147,12 +147,12 @@ def _recurse_lb(buf:LazyBuffer, realizes:Set[LazyBuffer], allbufs:Dict[LazyBuffe
     if DEBUG >= 3: print(f"forcing image {buf.dtype} with shape {buf.shape} to float32")
     buf.dtype = dtypes.float32  # NOTE: this is what makes the dtype above not match
   if buf.base != buf:
-    # realize all places where the buffer is expanded
+    # defer checking for unsafe pads
+    if any(v.mask is not None for v in buf.st.views): simple_pads.add(buf.base)
```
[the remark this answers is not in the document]
chaosagent: Oh, I see what you mean. Yes, it can be an else-if.

tinygrad/realize.py
```diff
@@ -148,12 +148,15 @@ def _recurse_lb(buf:LazyBuffer, realizes:Set[LazyBuffer], allbufs:Dict[LazyBuffe
     buf.dtype = dtypes.float32  # NOTE: this is what makes the dtype above not match
   if buf.base != buf:
     # realize all places where the buffer is expanded
+    # defer checking for unsafe pads
     if prod(buf.base.st.shape) < prod(buf.st.shape):
       if len(buf.st.views) == 1 and buf.st.views[-1].mask and all_int(buf.base.st.shape) and \
           prod(buf.base.st.shape) >= prod([y-x for x,y in buf.st.views[-1].mask]):
         simple_pads.add(buf.base)
```
geohot, line 155: Better. Now, is this still needed?

chenyuxyz:
conflicts with master

chenyuxyz:
oh it's still bugged on master... conflicts though

PR #3446, opened 2024-02-19
Cast correctly in python emulator (dtype tests pass)

patosai:
First cast to float or int depending on the output value. Then use struct to pack and then unpack to get the correct bits (esp. important for floats). Last edge case is integer casts where struct packing an out-of-bounds value will throw, so use the original dtype fmt to pack, and then unpack with the desired type.


geohot:
Congrats! Claim your $100 bounty by emailing george@tinygrad.org with either a PayPal email or a USDC address.

PR #3447, opened 2024-02-19
Implement complete C-style casting for the Python backend

obadakhalili:
I understand that the bounty have been claimed, but I have been working on C-style casting for the Python backend and I noticed that the merged PR isn't complete.

Here are some cases that the current implementation misses (run twice with `CLANG=1` and `PYTHON=1` to compare the results):

```python
a = Tensor([-2.5, -1, 1, 2.5], dtype=dtypes.float)
b = a.cast(dtypes.uint) # or dtypes.uint8, dtypes.uint16, dtypes.uint64
print(a.numpy(), b.numpy())

a = Tensor([0, 1, 255], dtype=dtypes.int8)
b = a.cast(dtypes.uint16) # or dtypes.uint32, dtypes.uint64
print(a.numpy(), b.numpy())

a = Tensor([-1, -2], dtype=dtypes.int16)
b = a.cast(dtypes.uint32) # or dtypes.uint8, dtypes.uint64
print(a.numpy(), b.numpy())
```

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -161,9 +161,13 @@ class TestHalfDtype(TestDType): DTYPE = dtypes.half
 
 class TestFloatDType(TestDType):
   DTYPE = dtypes.float
-  @unittest.skipIf(getenv("HIP", 0) == 1 or getenv("CUDA", 0) == 1 or getenv("METAL", 0) == 1, "doesn't work on CUDA, HIP, or METAL")
+  @unittest.skipIf(any([getenv("HIP"), getenv("CUDA"), getenv("METAL")]), "has different behavior on mentioned devices")
   def test_float_to_uint_negative(self):
-    _test_op(lambda: Tensor([-3.5, -2.5, -1.5], dtype=dtypes.float).cast(dtypes.uint), dtypes.uint, [4294967293, 4294967294, 4294967295])
+    _test_op(lambda: Tensor([-3.5, -2.5, -1.5], dtype=dtypes.float).cast(dtypes.uint), dtypes.uint, [2**32-1, 2**32-1, 2**32-1])
+
+  @unittest.skipIf(not any([getenv("HIP"), getenv("CUDA"), getenv("METAL")]), "the behavior is only for mentioned devices")
+  def test_float_to_uint_zeros(self):
+    _test_op(lambda: Tensor([-3.5, -2.5, -1.5], dtype=dtypes.float).cast(dtypes.uint), dtypes.uint, [0, 0, 0])
 
 class TestDoubleDtype(TestDType):
   DTYPE = dtypes.double
@@ -192,7 +196,7 @@ class TestInt8Dtype(TestDType):
     _test_op(lambda: Tensor([-1, -2, -3, -4], dtype=dtypes.int8).cast(dtypes.uint8), dtypes.uint8, [255, 254, 253, 252])
 
   def test_int8_to_uint16_negative(self):
-    _test_op(lambda: Tensor([-1, -2, -3, -4], dtype=dtypes.int8).cast(dtypes.uint16), dtypes.uint16, [65535, 65534, 65533, 65532])
+    _test_op(lambda: Tensor([-1, -2, -3, -4], dtype=dtypes.int8).cast(dtypes.uint16), dtypes.uint16, [2**16-1, 2**16-2, 2**16-3, 2**16-4])
 
 class TestUint8Dtype(TestDType):
   DTYPE = dtypes.uint8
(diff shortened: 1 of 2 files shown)


geohot:
Please add failing test

patosai:
Are we trying to be 1:1 with numpy? There's some undefined behavior when casting between unsigned and signed, ex. -1 as int8 is 0xff and numpy turns -1 int8 to 65535 uint16 which is 0xffff, so a 0xff popped out of nowhere. Going the other way seems fine as you just chop off bytes

unknownusername504:
> Are we trying to be 1:1 with numpy? There's some undefined behavior when casting between unsigned and signed, ex. -1 as int8 is 0xff and numpy turns -1 int8 to 65535 uint16 which is 0xffff, so a 0xff popped out of nowhere. Going the other way seems fine as you just chop off bytes

Screenshot_20240219-103301

What do you mean?

patosai:
Cast as a uint16, not uint8. Going between same size types works rn

unknownusername504:
> Cast as a uint16, not uint8. Going between same size types works rn

Oh well same idea, you mean the overflow? Yeah the sign extension is converted basically symbolically, it's the max value of int8 (in memory) to uint16 which acts as if it went to int16 first to transfer the symbolism of the max. It is to be deprecated behavior though, could just throw errors on out of bounds, not a really useful quirk.

tinygrad/runtime/ops_python.py
```diff
@@ -46,6 +46,22 @@ def _store(m, i, v):
   if i<0 or i>=len(m): raise IndexError(f"store out of bounds, size is {len(m)}, access is {i}, value is {v}")
   m[i] = v
 
+def cstyle_cast(value, in_dtype: DType, out_dtype: DType, bitcast: bool):
+  def c_bitcast(value, in_dtype: DType, out_dtype: DType):
+    # mapping dtype.name to ctypes API names as per https://docs.python.org/3/library/ctypes.html
+    def to_cname(dtype_name: str): return dtype_name.replace("char", "int8").replace("unsigned", "u").replace(" ", "")
```
chenyuxyz, line 52: `dtype_name.replace("char", "int8").replace("unsigned ", "u")` suffices

tinygrad/runtime/ops_python.py
```diff
@@ -46,6 +46,22 @@ def _store(m, i, v):
...
+    return ctypes.cast(ctypes.pointer(in_cvalue), ctypes.POINTER(out_ctype)).contents.value
+
+  if bitcast: return c_bitcast(value, in_dtype, out_dtype)
+
+  # perform Python type casting
+  out = int(value) if dtypes.is_int(out_dtype) else float(value) if dtypes.is_float(out_dtype) else bool(value)
+  # return the output if the involved types are Python-supported
+  if not (dtypes.is_unsigned(in_dtype) or dtypes.is_unsigned(out_dtype)): return out
+  # otherwise, bitcast the output to mimic C-style type casting for operands involving unsigned types
+  return c_bitcast(out, out_dtype, out_dtype)
```
chenyuxyz, line 63: i think it's more readable to write
```
if dtypes.is_unsigned(in_dtype) or dtypes.is_unsigned(out_dtype): out = c_bitcast(out, out_dtype, out_dtype)
return out
```

also it's not obvious to me why it's `c_bitcast(out, out_dtype, out_dtype)` not `c_bitcast(out, in_dtype, out_dtype)`, consider commenting if that's intended

test/test_dtype.py
```diff
@@ -159,7 +159,11 @@ def test_bf16_disk_write_read(self):
 
 class TestHalfDtype(TestDType): DTYPE = dtypes.half
 
-class TestFloatDType(TestDType): DTYPE = dtypes.float
+class TestFloatDType(TestDType):
+  DTYPE = dtypes.float
+  @unittest.skipIf(getenv("HIP", 0) == 1 or getenv("CUDA", 0) == 1 or getenv("METAL", 0) == 1, "doesn't work on CUDA, HIP, or METAL")
```
chenyuxyz, line 164: why does it fail on these backends?

tinygrad/runtime/ops_python.py
obadakhalili: It's because after the Python-cast, they are now both of the same type, and you need to specify the appropriate type to wrap the value correctly in C.

obadakhalili: A comment feels a bit unnecessary and redundant.

obadakhalili: I could rewrite the `in_dtype` as `out_dtype` on the same casting line if it feels cleaner.

test/test_dtype.py
obadakhalili: tbh, I don't have access to any of these devices to research an answer that gives a more specific reason to use, but I noticed that it is a pattern for some casting ops to fail on some backends, so I wrote a skip line initially to make it stand for review.

tinygrad/runtime/ops_python.py
obadakhalili: I might have a refactor in mind that makes this even cleaner.

obadakhalili: I pushed a refactor that addresses this by changing the API of `c_bitcast` to accommodate the use case here, and I explained the reasoning in a comment.

test/test_dtype.py
obadakhalili: After researching, I found that the issue wasn't a matter of a failing test, but rather implementation-defined behavior. The mentioned devices treat `float` to `uint` differently than the rest. So I reflected that in my last commit.

obadakhalili:
@geohot do you think I can get feedback on this?

geohot:
What bug does this fix?

obadakhalili:
It used to handle the cases mentioned in the description, which got handled only after I pushed this PR. I think the main advantage this PR currently offers is cleaner code, as I think it is more intuitive and easier to read and change. Also, the `test_float_to_uint` test in master doesn't cover the cases that this PR came to handle, which are found here.

I don't really want the bounty money, I care more about the contribution, and if possible co-owning the bounty on the sheet (i,e listing my name too), but mainly the contribution if the later isn't possible.

PR #3450, opened 2024-02-20
Linearizer 2.0 (proof-of-concept)

Qazalin:
linearizer2.py has the infra to schedule and uop a multioutput, multireduce kernel.
To see it in action, run the unit tests in the file with `LINEARIZER2=1 DEBUG=4`

This poc is _very_ limited:
- MiniLinearizer will evolve as I add apply_opt support, ShapeTracker based load/store indexing and global/local size.
- The scheduler needs to get sophisticated: eg. deciding when to break fusion now that it's not limited by store/reduce.
- Needs a data structure to represent the DAG (for opt)


geohot:
Good stuff! So the deciding how to split challenge is better than this, because all the buffers have to have the same shape and that's a tricky decision too.

But I think this style can be a lot more compact than the existing Linearizer, even just rendering the kernels as is, testing that multireduce and multioutput work.

Qazalin:
closing in favor of multioutput

PR #3451, opened 2024-02-20
fix LtNode simplification when lhs and rhs contain same variables

chenyuxyz:
`(Variable("a", 1, 5) < Variable("a", 1, 5))` should eval to `NumNode(0)`


chenyuxyz:
hmm the subtraction is quite slow... more than doubled the calls to Node.sum for winograd. trying a different fix

PR #3454, opened 2024-02-20
Adding type hints

marcoloco23:
Just adding type hints throughout


geohot:
Closing, this is effectively a whitespace PR. I would only consider a very targeted type hint addition that fixes a bug.

PR #3455, opened 2024-02-20
test/external/fuzz_linearizer: add a FUZZ_MAX_SIZE option

flammit:
this allows us to limit the size of the kernel and reduce running times by avoiding ones that take a long time

### DIFF
--- a/test/external/fuzz_linearizer.py
+++ b/test/external/fuzz_linearizer.py
@@ -71,13 +71,13 @@ def fuzz_linearizer(lin: Linearizer):
   print_tree(lin.ast)
   print(lin.colored_shape())
   rawbufs = get_fuzz_rawbufs(lin)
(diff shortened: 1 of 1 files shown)


test/external/fuzz_linearizer.py
```diff
@@ -76,6 +76,11 @@ def fuzz_linearizer(lin: Linearizer):
   last_lins = [lin]
   failures = defaultdict(list)
 
+  FUZZ_MAX_SIZE=getenv("FUZZ_MAX_SIZE", 0)
```
chenyuxyz, line 79: nit: can you fix the spaces between `=` and `>`? (this block and line 74)

flammit: done and fixed line 80 too

PR #3457, opened 2024-02-21
WIP symbolic dimension

chenyuxyz:
add SymDim to represent Symbolic Dimension. It's used in symbolic shape and has a smaller method set. most of complicated (Node, Node) can be handled only in SymDim, and should make non-symbolic shape symbolic fast.


chenyuxyz:
working on a different approach. Specialized LtNode creation for valid, and assert all non-defined Node comparision.

PR #3461, opened 2024-02-21
simpler acc uoping (with upcast)

Qazalin:
linearizer's global_load is handling a lot of things right now. I think DEFINE_ACC can be uoped separately since:
- acc is a register value, there is no buffer to load
- different dtype/value logic since there is no Mem/Local buffer for it in the AST.
- DEFINE_ACC isn't concerned about a lot of the things that global_load does, eg. `valid`.

still exploring if this makes sense for multireduce.


geohot:
So the issue isn't the perf in the arithmetic, it's in the store. You want the variable to be a float4 when you store it, because otherwise the compiler might not put the registers in order.

Qazalin:
but it might still make sense to group accs if the store can be grouped too...
eg `*((device float2*)(data0+alu4)) = (float2)float2((acc0).x,(acc0).y)`

Qazalin:
i generated a diff between this and master to make sure no kernels changed,
but i think we need upcast validation tests in test_linearizer.

Qazalin:
i think it's better to do this refactor once linearizer doesn't have late/early bufs - for now global_load does the job

PR #3463, opened 2024-02-21
fix softmax and log_softmax for 0d tensor

chenyuxyz:
matched torch to take axis \in [-1, 0] and used axis=None internally


geohot:
Ahh, yea I like this better in softmax, I don't think the reduce ops should be able to take out of range axes

PR #3473, opened 2024-02-22
Fix `Tensor.cumsum` when axis of length 0 is selected

obadakhalili:
I can also just early return in `_cumsum` when the specified axis length is 0.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -722,7 +722,7 @@ class Tensor:
     return x.dot(self, acc_dtype=acc_dtype) if reverse else self.dot(x, acc_dtype=acc_dtype)
 
   def _cumsum(self, axis:int=0, _first_zero=False) -> Tensor:
-    return self.transpose(axis,-1).pad2d((self.shape[axis]-int(not _first_zero),0))._pool((self.shape[axis],)).sum(-1).transpose(axis,-1)
+    return self.transpose(axis,-1).pad2d((max(self.shape[axis]-int(not _first_zero), 0),0))._pool((self.shape[axis] or 1,)).sum(-1).transpose(axis,-1)
   def cumsum(self, axis:int=0) -> Tensor:
     # TODO: someday the optimizer will find this on it's own
     # for now this is a two stage cumsum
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -722,7 +722,7 @@ def matmul(self, x:Tensor, reverse=False, acc_dtype:Optional[DType]=None) -> Ten
     return x.dot(self, acc_dtype=acc_dtype) if reverse else self.dot(x, acc_dtype=acc_dtype)
 
   def _cumsum(self, axis:int=0, _first_zero=False) -> Tensor:
-    return self.transpose(axis,-1).pad2d((self.shape[axis]-int(not _first_zero),0))._pool((self.shape[axis],)).sum(-1).transpose(axis,-1)
+    return self.transpose(axis,-1).pad2d((max(self.shape[axis]-int(not _first_zero), 0),0))._pool((self.shape[axis] or 1,)).sum(-1).transpose(axis,-1)
```
geohot, line 725: I can't read this line.

obadakhalili: - When specifying an axis of length 0 in `Tensor.cumsum`, `pad2d` will throw an error because it will receive `(-1, 0)` as an argument, so the first change is equivalent to subtracting one only if `not _first_zero and self.shape[axis]` is True, but I wrote it as `max(self.shape[axis]-int(not _first_zero), 0)` to avoid the long line length.
- The second change is also to handle the case when the specified axis length is 0. After the first change, the output shape of `pad2d` will be the same as its input shape that came from `transpose` (`(2, 4, 0)` for example), and in order to preserve the same shape for the output from `_pool` (which is what we need when the selected axis length is 0), 1 is passed as an argument.

So the change is manipulating the passed arguments to make sure that the output shape from `cumsum` is the same as the input shape when the specified axis length is 0, which is why I suggested that adding an early return to `_cumsum` to handle this case would also work (`if self.shape[axis] == 0: return self`). What do you think is the better fix?

obadakhalili:
I define the padding-left size in a separate line to make the code easier to read.

obadakhalili:
Can I get feedback on this? @chenyuxyz

geohot:
Looks good, merged!

PR #3475, opened 2024-02-22
move all reduces to the end in lazy

geohot:
This is just an experiment.

I don't like how the reduce LazyOp doesn't specify the actual reduce operation. Neither new_shape nor axis list expresses it fully, but rather depends on what shape the input buffer has.

One idea is to have the arg just be the number of elements that are being reduced and move them to the end, but I'm not sure this is expressive enough for multireduce.


geohot:
Oh I think this actually might work! The final permute isn't really a permute, it's a reshape of only 1s.

geohot:
This is also already happening in the Linearizer

PR #3477, opened 2024-02-22
O(n) nn.Embedding in one kernel without new ops.

pushpendre:
Submission for the `O(n) nn.Embedding in one kernel without new ops (supporting BS=128)` bounty.

The idea is to generate all the indices in the right shape first and then to rely on tensor indexing. This gets rid of matrix multiplications and reduces runtime.

The generated kernels can be seen with this command
```
DEBUG=9 python > /tmp/tmp.txt <<EOF
from tinygrad import nn
import numpy as np
batch_size, seqlen, vocab_size, embedding_size = 128, 64, 32, 16
I = Tensor(np.random.randint(0, vocab_size, size=(batch_size, seqlen)))
print(nn.Embedding(vocab_size, embedding_size)(I).numpy())
EOF
```

Tested: All unit tests pass

### DIFF
--- a/tinygrad/nn/__init__.py
+++ b/tinygrad/nn/__init__.py
@@ -126,8 +126,8 @@ class Embedding:
     self.weight = Tensor.glorot_uniform(vocab_size, embed_size)
 
   def __call__(self, idx:Tensor) -> Tensor:
-    if not hasattr(self, 'vocab_counter'):
-      self.vocab_counter = Tensor.arange(self.vocab_size, requires_grad=False, device=self.weight.device).reshape(1, 1, self.vocab_size)
     batch_size, seqlen = idx.shape
     if seqlen == 0: return Tensor.empty(batch_size, 0, self.embed_size, device=self.weight.device)
-    return (self.vocab_counter == idx.unsqueeze(2)).expand(*idx.shape, self.vocab_size) @ self.weight
+    increments = Tensor.arange(self.embed_size, requires_grad=False, device=self.weight.device).reshape(1, 1, self.embed_size)
+    idx = idx.reshape((batch_size, seqlen, 1)).expand((batch_size, seqlen, self.embed_size)) * self.embed_size + increments
+    return self.weight.reshape(-1)[idx].reshape((batch_size, seqlen, self.embed_size))


chenyuxyz:
Are you sure this is an O(n) Embedding?

davidjanoskyrepo:
Technically if you just do an straight analysis, it isn't strictly better than ~O(n^2) (functions of batchsize, seq len and embed size). It went from the matrix mult to the indexes formation and indexing driving the complexity.

However since the indexes are formed with symbolic operations and the indexing itself is done with those views then the time complexity analysis would depend on the specifics of how these symbolic representations are implemented and executed within the computational graph framework. This does at the very least form one kernel with no new ops.

Was this profiled at all for small and large data?

pushpendre:
I profiled it a little on my own machine (apple M1) using the CLANG
backend. METAL was crashing due to another bug.

I have written a micro-benchmark but how can I test how it will perform on
cuda/opencl gpus etc. ? If there is a doc or code (maybe from another
issue?) to set up a micro-benchmark / test harness please let me know.

Also the tests on clang and llvm are failing due to higher memory / larger
runtime but the gpu/cuda/opencl runtimes are fine. Will that make this PR a
non-starter?

pushpendre:
This is the benchmarking code and its results unfortunately show that the new kernel is much slower on my mac but it'll be great if someone could run it on an ubuntu machine with cuda/rocm.

```python
from tinygrad import nn
import numpy as np
import itertools
import time
def timeit_callable(f):
    time_taken = []
    for trials in range(3):
        start = time.perf_counter_ns()
        f()
        time_taken.append(time.perf_counter_ns() - start)
    return min(time_taken)

for batch_size, seqlen, vocab_size, embedding_size in itertools.product([128,], [64, 128], [32, 256], [128, 256]):
    indices = Tensor(np.random.randint(0, vocab_size, size=(batch_size, seqlen)))
    x = nn.Embedding(vocab_size, embedding_size)(indices).numpy() # warmup 
    del x
    time_taken = timeit_callable(lambda : nn.Embedding(vocab_size, embedding_size)(indices).numpy())
    print(batch_size, seqlen, vocab_size, embedding_size, time_taken)
```



```bash
git checkout master
GPU=1 python ../on_embedding_benchmark.py > ../before.txt
git checkout On-embedding
GPU=1 python ../on_embedding_benchmark.py > ../after.txt
paste ../before_gpu.txt <( awk '{print $5}' ../after_gpu.txt ) | awk '{print $0 , $6/$5}'
```



```
bs, seqlen, vocab, embedding, before, after, after/before
128 64 32 128 4869125   16956750 3.4825
128 64 32 256 4727542   46456083 9.82669
128 64 256 128 2009792  90560500 45.0596
128 64 256 256 3051875  354687958 116.22
128 128 32 128 5621833  24103834 4.28754
128 128 32 256 8112000  90950542 11.2119
128 128 256 128 2123375 178329500 83.984
128 128 256 256 3839792 705470583 183.726
 ```

pushpendre:
Closing PR because even though O(n) complexity is achieved by explicit indexing the resulting code runs slower than the existing code.

PR #3403, opened 2024-02-15
hsa multigpu graph

nimlgen:
Graph is faster than hip both for a single device and multidevices. HSA still got limitation since the only way to access sdma queue is to call hsa functions. Ideally we want to access queue and push prebuilt packets into it (as it's now with aql queue). As of now, the only way to access it is to ask the kernel driver to map it. I will play with it to see what perf we can get ideally.

llama: `enqueue in   3.89 ms total  43.15 ms, 23.17 tok/sec` -> `enqueue in   2.43 ms, total  41.12 ms, 24.32 tok/sec`
gpt2: `ran model in   7.80 ms` -> `ran model in   7.32 ms`
cifar:
```
HSA=1 GPUS=6 HALF=1 LATEWINO=1 LATEBEAM=0 DEBUG=0 STEPS=350 BS=3072 TARGET_EVAL_ACC_PCT=93 python3 examples/hlb_cifar10.py
...
348   45.27 ms run,   10.43 ms prep batch,    3.45 ms jit,   31.40 ms HSA * 6,  493.00 loss, 0.000099 LR, 11.09 GB used,  27859.79 GFLOPS,   1261.24 GOPS
349   45.95 ms run,   10.64 ms prep batch,    3.34 ms jit,   31.97 ms HSA * 6,  493.50 loss, 0.000086 LR, 11.09 GB used,  27449.68 GFLOPS,   1261.24 GOPS
shuffling test dataset in 52.84 ms (epoch=0)
eval     11504/12288 93.62%,    0.43 val_loss STEP=350 (in 4647.68 ms)
eval_acc_pct=93.61979166666666 >= 93.0
```

```
HIP=1 GPUS=6 HALF=1 LATEWINO=1 LATEBEAM=0 DEBUG=0 STEPS=350 BS=3072 TARGET_EVAL_ACC_PCT=93 python3 examples/hlb_cifar10.py
...
348   80.01 ms run,   12.56 ms prep batch,   34.22 ms jit,   33.23 ms HIP * 6,  493.00 loss, 0.000099 LR, 11.09 GB used,  15762.93 GFLOPS,   1261.24 GOPS
349   79.63 ms run,   12.43 ms prep batch,   35.53 ms jit,   31.68 ms HIP * 6,  493.50 loss, 0.000086 LR, 11.09 GB used,  15838.13 GFLOPS,   1261.24 GOPS
shuffling test dataset in 66.75 ms (epoch=0)
[...]

PR #3478, opened 2024-02-22
jit grapher simplified

nimlgen:
Rewrite jit grapher to simpler integrate multigraph support. I think it looks cleaner, the logic is the same.

For #3403


nimlgen:
Something like this it will look with multigraph support: https://github.com/tinygrad/tinygrad/pull/3403/commits/6b3d42c342b3e8119cb8eb5fea65d291df03e033

PR #3480, opened 2024-02-23
general constant folding

carsonRadtke:
* enable constant folding tests
* call exec_alu when all ops are CONST

### DIFF
--- a/test/test_multitensor.py
+++ b/test/test_multitensor.py
@@ -148,6 +148,7 @@ class TestMultiTensor(unittest.TestCase):
     out.mean().backward()
     #for p in get_parameters(conv): p.grad.realize()
     optim.step()
+    out.numpy()
 
   def test_lr_scheduler_OneCycleLR(self):
     from extra.lr_scheduler import OneCycleLR
@@ -520,6 +521,7 @@ class TestShrinkMultiTensorShardedAxis(unittest.TestCase):
     optim.zero_grad()
     out.mean().backward()
     optim.step()
+    out.numpy()
 
   def test_unsynced_backprop_standalone_bn(self):
     from extra.lr_scheduler import OneCycleLR
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -341,10 +341,13 @@ class TestMoveTensor(unittest.TestCase):
          strat.sampled_from([dtypes.float16, dtypes.float32]), strat.sampled_from([True, False, None]))
   def test_to_preserves(self, src, dest, dtype, requires_grad):
     s = Tensor([1, 2, 3], device=src, dtype=dtype, requires_grad=requires_grad)
+    if requires_grad: s.sum().backward()
     t = s.to(dest)
     np.testing.assert_equal(s.numpy(), t.numpy())
     assert s.dtype == t.dtype
     assert s.requires_grad == t.requires_grad
+    if requires_grad:
+      np.testing.assert_equal(s.grad.numpy(), t.grad.numpy())
 
   @given(strat.sampled_from([dtypes.float16, dtypes.float32]), strat.sampled_from([True, False, None]))
   def test_shard_preserves(self, dtype, requires_grad):
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -180,7 +180,7 @@ class Tensor:
     if device is None or device == self.device: return self
     if not isinstance(device, str): return self.shard(device)
     ret = Tensor(self.lazydata, device, requires_grad=self.requires_grad)
-    if self.grad: ret.grad = self.grad.to(device)
+    if self.grad is not None: ret.grad = self.grad.to(device)
     if hasattr(self, '_ctx'): ret._ctx = self._ctx
     return ret
 


geohot:
Cool! Can you test the overflow conditions of the dtypes. This currently isn't handled correctly.

geohot:
Can you also add a test that would have caught the sqrt bug? (would quickly merge that test+fix alone)

geohot:
Any progress on this? I like it, but we need to check overflow conditions.

test/test_uops.py
```diff
@@ -114,5 +114,8 @@ class TestExecALU(TestUOps):
   def test_sqrt(self):
     self.assertEqual(exec_alu(UnaryOps.SQRT, dtypes.int, (0,)), 0)
 
+  def test_overflow(self):
+    self.assertEqual(exec_alu(BinaryOps.ADD, dtypes.uint8, (250, 250)), 244)
```
geohot, line 118: Going to need a lot more tests than this.

tinygrad/dtype.py
```diff
@@ -17,6 +17,10 @@ def vec(self, sz:int):
     assert sz > 1 and self.count == 1, f"can't vectorize {self} with size {sz}"
     return DType(self.priority, self.itemsize*sz, f"{INVERSE_DTYPES_DICT[self.name]}{sz}", None, sz)
   def scalar(self): return DTYPES_DICT[self.name[:-len(str(self.count))]] if self.count > 1 else self
+  def __rmod__(self, other: Scalar) -> Scalar:
```
geohot, line 20: I don't think you actually want to mod by the dtype! This logic should be in exec_alu

tinygrad/runtime/ops_python.py
```diff
@@ -3,35 +3,36 @@
...
-  if arg == BinaryOps.ADD: return p[0]+p[1]
-  if arg == BinaryOps.SUB: return p[0]-p[1]
-  if arg == BinaryOps.XOR: return p[0]^p[1]
-  if arg == BinaryOps.MAX: return max(p[0], p[1])
-  if arg == BinaryOps.CMPEQ: return p[0] == p[1]
-  if arg == BinaryOps.CMPLT: return p[0] < p[1]
-  if arg == BinaryOps.DIV: return p[0]//p[1] if dtypes.is_int(dtype) else (p[0]/p[1] if p[1] != 0 else math.nan)
-  if arg == BinaryOps.MOD: return p[0]%p[1]
-  raise NotImplementedError(f"no support for {arg}")
+  def exec_alu_helper(arg, p):
```
geohot, line 14: Don't nest functions, just set ret =

tinygrad/dtype.py
carsonRadtke: Perhaps mod is not correct, but I don't think the operation belongs in exec_alu because exec_alu isn't the only place that performs this adjustment business. 

__call__ is probably better because then one can do `dtypes.int8(x)` and get an "in-bounds" version of x.

geohot: That's not better. This should not be in dtype at all.

test/test_linearizer.py
```diff
@@ -168,11 +167,6 @@ def helper_test_simplify(uop, dtype, vin, arg=None):
     c0 = UOp(UOps.CONST, dtypes.float, vin=(), arg=0.0)
     assert helper_test_simplify(UOps.ALU, dtypes.float, vin=(UOp(UOps.CONST, dtypes.bool, vin=(), arg=True), c0, c0), arg=TernaryOps.WHERE) == c0
 
-    c0 = UOp(UOps.CONST, dtypes.float, vin=(), arg=0.0)
-    c1 = UOp(UOps.CONST, dtypes.float, vin=(), arg=1.0)
-    assert helper_test_simplify(UOps.ALU, dtypes.float, vin=(UOp(UOps.CONST, dtypes.bool, vin=(), arg=True), c0, c1),
```
geohot, line 173: Why was this test removed?

carsonRadtke: I removed it because it was no longer correct. I returned it, corrected it, and fixed a bug that it identified.

carsonRadtke:
First time dealing with conflicts in a GitHub fork. sorry for the noise.

tinygrad/codegen/linearizer.py
```diff
@@ -44,8 +45,8 @@ def uop_alu_idx(self, a:UOp, b, ops, ctx:Linearizer, op, dtype=dtypes.int32):
     return self.uop(UOps.ALU, dtype, (a, render_b), op)
 
   # NOTE: the consts have to be cached for deduping of downstream uops to work
-  def const(self, b:Union[int,float], dtype=dtypes.int32, insert_before=None) -> UOp:
-    return self.uop(UOps.CONST, dtype, tuple(), b, insert_before=insert_before)
+  def const(self, b:Union[int,float], dtype=dtypes.int32, insert_before=None, cachable=True) -> UOp:
```
geohot, line 48: Consts should always be cached. If they can't be, there's a deeper bug.

geohot:
I moved the simplify function into uops.py, hard for me to resolve that conflict via GitHub editor

See above comment about constants, they should never be cached. Just try without it

Want to get this merged ASAP

tinygrad/codegen/uops.py
```diff
@@ -88,11 +88,12 @@ def add(self, uop:UOps, dtype:Optional[DType]=None, vin:Tuple[UOp, ...]=tuple(),
         if arg is BinaryOps.ADD and vin[1].uop is UOps.ALU and vin[1].arg is UnaryOps.NEG:
           return self.add(UOps.ALU, dtype, (vin[0], vin[1].vin[0]), BinaryOps.SUB, cachable, insert_before)
         # constant folding
-        if arg is UnaryOps.NEG and vin[0].uop is UOps.CONST: return self.add(UOps.CONST, dtype, arg=-vin[0].arg, insert_before=insert_before)
         if arg is TernaryOps.WHERE and vin[1] == vin[2]: return vin[1] # a conditional with the same results either way is a noop
         if arg is TernaryOps.WHERE and vin[0].uop is UOps.CONST: return vin[1] if vin[0].arg else vin[2]
         if arg is BinaryOps.MUL and vin[0].uop is UOps.CONST and vin[1].uop is UOps.CONST and dtype is not None and dtypes.is_float(dtype):
```
geohot, line 93: This MUL should be unneeded, right?

tinygrad/codegen/uops.py
```diff
@@ -88,11 +88,12 @@ def add(self, uop:UOps, dtype:Optional[DType]=None, vin:Tuple[UOp, ...]=tuple(),
         if arg is BinaryOps.ADD and vin[1].uop is UOps.ALU and vin[1].arg is UnaryOps.NEG:
           return self.add(UOps.ALU, dtype, (vin[0], vin[1].vin[0]), BinaryOps.SUB, cachable, insert_before)
         # constant folding
```
chenyuxyz, line 90: move this comment too

geohot:
uops test is now taking 8 minutes. The overflow handling needs to be faster

carsonRadtke:
> uops test is now taking 8 minutes. The overflow handling needs to be faster


I was able to shave 3 minutes off. Definitely still need to get it below 5 min. I'll spend some more time tonight trying to get it below 3 like master is now.

tinygrad/codegen/uops.py
```diff
@@ -41,10 +41,10 @@ def wfxn(*args):
   TernaryOps.WHERE: lambda x,y,z: y if x else z}
 
 def exec_alu(arg, dtype, p):
-  return python_alu[arg](*p)
-  #if not dtypes.is_int(dtype): return ret
-  #adjusted = 0 if dtypes.is_unsigned(dtype) else 2 ** (dtype.itemsize * 8 - 1)
-  #return (ret + adjusted) % 2 ** (dtype.itemsize * 8) - adjusted
+  ret = python_alu[arg](*p)
+  if not dtypes.is_int(dtype): return ret
```
geohot, line 45: I suggest (once) constructing a dictionary of functions to apply the truncation for dtype. Like Dict[dtype, Callable]

carsonRadtke:
I'm going to keep this closed because working of master on my fork is causing too many issues. Will open another MR and link here.

Issue #3453, opened 2024-02-20
CLANG and METAL backends fail with bfloat16 when ran with Clang v14


skoshx:
Installed tinygrad from scratch, then just ran:

```
$ CLANG=1 python3 examples/coder.py
```
and
```
$ METAL=1 python3 examples/coder.py
```

Here's debug output (quite same on METAL and CLANG, both contain `__bf16*`)
```c
create model:  32.15 ms
download weights:  25.70 ms
ram used:  0.00 GB, layers.0.attention.wq.weight                      :   0%| | assign <LB CLANG (4096, 4096) contig:True (<BinaryOps.ADD: 1>, None)> <- <LB CLANG (4096, 4096) contig:True (<UnaryOps.CAST: 3>, None)>
*** DISK:/U    1 empty 9943044915 dtypes.uchar          arg   1 mem  0.00 GB 
*** CLANG      2 read   33.55M,   CLANG <- DISK:/U      arg   2 mem  0.03 GB tm   2508.54us/     2.51ms (    0.00 GFLOPS,   13.38 GB/s)
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.half, st=ShapeTracker(views=(View(shape=(4096, 4096), strides=(4096, 1), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ CAST (dtypes.half, False)
  2    ┗━┳ CAST (dtypes.float, True)
  3      ┗━┳ MUL 
  4        ┣━┳ CAST (dtypes.uint, False)
  5        ┃ ┗━┳ CAST (dtypes.ushort, True)
  6        ┃   ┗━━ LOAD MemBuffer(idx=1, dtype=dtypes.bfloat16, st=ShapeTracker(views=(View(shape=(32, 64, 2, 4096), strides=(524288, 4096, 262144, 1), offset=0, mask=None, contiguous=False), View(shape=(4096, 4096), strides=(4096, 1), offset=0, mask=None, contiguous=True))))
  7        ┗━━ CONST ConstBuffer(val=65536, dtype=dtypes.uint, st=ShapeTracker(views=(View(shape=(4096, 4096), strides=(0, 0), offset=0, mask=None, contiguous=False),)))
float4 merging axis : [(2, 2, 1, 4), (2, 8192, 0, 4)]
void E_4096_1024_4(half* restrict data0, const __bf16* restrict data1) {
  for (int ridx0 = 0; ridx0 < 4096; ridx0++) {
[...]

PR #3481, opened 2024-02-23
test: add failing bfloat16 test case for metal backend

skoshx:
Created a failing test for Metal for issue #3453


test/test_ops.py
```diff
@@ -1624,6 +1624,11 @@ def test_one_hot(self):
     data = [[[1, 2, 3], [0, 3, 5]], [[1, 2, 3], [0, 3, 5]]]
     helper_test_op([], lambda: torch.nn.functional.one_hot(torch.tensor(data), 8), lambda: Tensor(data).one_hot(8), forward_only=True)
 
+  @unittest.skipIf(getenv("METAL", 0)== 0, "only run on METAL backend")
```
chenyuxyz, line 1627: I think bfloat16 tests in test_dtype.py is a better place to add this. What's special about uniform? Would be nice to have the minimum test case that shows the issue. Finally need to set expect failure instead of just letting it fail.

skoshx: Hey, I can make it simpler with maybe just like `Tensor([1])`, nothing special about uniform.

And isn't letting it fail correct? I'm assuming that these errors aren't the expected output? Shouldn't tinygrad somehow gracefully handle `bfloat16` on Metal backend?

If I make it expect failure, then the CI essentially reflects that when there are bugs, the CI should pass?

skoshx:
@chenyuxyz moved to `test_dtype.py`. Funnily enough, just using something like `Tensor([1,2,3])` or `Tensor.eye(3)` doens't reproduce the bug, so the uniform seems to be the simplest to reproduce it.

Also added `@unittest.expectedFailure`

chenyuxyz:
running
```
from tinygrad import Tensor
x = Tensor.uniform(10, dtype=dtypes.bfloat16)
x.realize()
```
has no problem on my local M1 Max. Is the test set up correctly?

skoshx:
Weird, fails for me on local M2 Pro. And yeah I think so, earlier when the test suite ran without the `expectedFailure`, my test correctly failed on the CI.

skoshx:
Screenshot 2024-03-05 at 17 25 16

I'm suspecting this is just some super weird bug on my machine. Since M1, M2, M3 should all support `bfloat16` according to [Metal docs](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf).

PR #3472, opened 2024-02-22
UnsyncedBatchNorm with synced trainable weights for hlb cifar

chaosagent:
Developed for #3361. Previous UnsyncedBatchNorm kept separate weights and running statistics for every GPU, but weights are learnable parameters and should be optimized against the entire global batch. In the old UnsyncedBatchNorm, this would result in 1/6th the LR for weights for 6 GPUs, and also desynced learnable parameters.

Implemented ad-hoc in hlb_cifar10.py right now, but I think this can replace nn.BatchNorm2d as-is (this should work even without MLB, need to try)

Results in ~0.5-1.0% accuracy gain on CIFAR with no noticeable slowdown (maybe it's actually faster). Also removes the Tensor.stack kernels needed from old UnsyncedBatchNorm

Am benchmarking...

Old:
```
LATEBEAM=2 HALF=1 STEPS=350 BS=1536 GPUS=6
348   44.00 ms run,   26.53 ms python,   17.47 ms HIP * 6,  503.00 loss, 0.000099 LR, 6.76 GB used,  46090.57 GFLOPS,   2027.99 GOPS
349   43.81 ms run,   26.95 ms python,   16.86 ms HIP * 6,  501.50 loss, 0.000086 LR, 6.76 GB used,  46292.03 GFLOPS,   2027.99 GOPS
shuffling test dataset in 43.00 ms (epoch=0)
eval     10009/10752 93.09%,    0.43 val_loss STEP=350 (in 2164.65 ms)

real    0m50.613s

LATEBEAM=2 HALF=1 STEPS=350 BS=3072 GPUS=6
348   62.89 ms run,   30.54 ms python,   32.34 ms HIP * 6,  494.25 loss, 0.000099 LR, 12.67 GB used,  64486.94 GFLOPS,   4055.33 GOPS
349   62.76 ms run,   30.47 ms python,   32.29 ms HIP * 6,  493.50 loss, 0.000086 LR, 12.67 GB used,  64620.31 GFLOPS,   4055.33 GOPS
shuffling test dataset in 43.85 ms (epoch=0)
eval     11507/12288 93.64%,    0.44 val_loss STEP=350 (in 2118.33 ms)

real    1m7.084s
```

New:
```
LATEBEAM=2 HALF=1 STEPS=350 BS=1536 GPUS=6
[...]

Issue #3395, opened 2024-02-14
Understand what assign is


geohot:
It's really a constraint on the memory manager, enforcing that two assigned tensors use the same underlying Buffer.

Though there may be edge cases here.

PR #3483, opened 2024-02-23
don't allow MLB assigns with different axes

chaosagent:
this allows initializing axis-sharded weights without sharding on axis, and letting the shard happen implicitly after in the first step.

needed for #3472


chenyuxyz:
i don't quite understand the pr title and the intended change? isn't the code change more limiting? also i think the test already passes without code change?

chaosagent:
it doesn't pass, because the old buffers before axis shard are larger than the new buffers. Then .nbytes is not the same as prod(shape), and copyin will fail.

chaosagent:
ok the test is fixed, this bug only happens if the axis=None buffer already had a realized buffer (eg when resuming training from ckpt)

chenyuxyz:
Is there a way to make it more explicit? What would be the alternative if we asserts the axis to be the same.

While we don't have a full spec for assign now (#3395), it feels less predictable that an assignment changes shard axis implicitly. I would expect that the assignment eventually checks certain properties (like shape, size, shard axis, although there's more complicated issues like real size can be different or contiguous etc) and throws error accordingly.

test/test_multitensor.py
```diff
@@ -307,6 +307,24 @@ def test_bn_ast_on_devices(self):
     # for i, ast in enumerate(asts):
     #   print(f"{i} {ast}")
 
+  def test_mlb_assign_change_axis(self):
+    devices = (d0, d1, d2, d3)
+
+    # axis=None <- axis=None + axis=0
+    t_none = Tensor.zeros((16, 16)).shard(devices).realize()  # allocate 16*16 length buffers on each device
+    assert t_none.lazydata.lbs[0].realized.size == 16 * 16  # check that buffers have been allocated
```
chenyuxyz, line 315: const multi Tensor will not have realized (similar to single LB case) so this might change

chenyuxyz (commented):
Let me think more about the desired case. Can you fix the conflict in the meantime?

chenyuxyz:
> But when it get to that point, the optimizer has already gotten the reference to the Tensor, and thus we need a way to swap out the lazydata that isn't assign

We can shard the model weights before putting it in optimizer. Does that work?

chenyuxyz:
this fails with the newly added unsynced bn with virtual multi devices like llvm and clang

PR #3484, opened 2024-02-23
support MLB reshaping on-axis for evenly sharded

chaosagent:
needed for #3472


chenyuxyz (commented):
looks good, some nits

Issue #3482, opened 2024-02-23
examples/gpt2.py doesn't work with GPU=1 on m1


certik:
```console
$ GPU=1 python examples/gpt2.py --model_size=gpt2
using GPU backend
using gpt2
ram used:  0.50 GB, lm_head.weight                                    : 100%|█████████████████████████████████████████████████████████████████████████████| 149/149 [00:00<00:00, 1575.12it/s]
loaded weights in 101.92 ms, 0.65 GB loaded at 6.40 GB/s
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [00:06<00:00, 14.38it/s]
thread '<unnamed>' panicked at src/lib.rs:201:64:
no entry found for key
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/Users/ondrej/repos/tinygrad/examples/gpt2.py", line 201, in <module>
    texts = gpt2.generate(args.prompt, args.count, args.temperature, timing=args.timing, batch_size=args.batch_size)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ondrej/repos/tinygrad/examples/gpt2.py", line 170, in generate
    return [self.tokenizer.decode(x) for x in toks]
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ondrej/mambaforge/envs/tinygrad/lib/python3.12/site-packages/tiktoken/core.py", line 258, in decode
    return self._core_bpe.decode_bytes(tokens).decode("utf-8", errors=errors)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: no entry found for key
```

fijal:
The problem is loading uchar from disk using METAL. Trying to debug

fijal:
Specifically, it hits https://github.com/tinygrad/tinygrad/blob/358a24eae634f1b6240579eaf1b96a1416326a3f/tinygrad/runtime/ops_disk.py#L56 with what I believe is not a bitcast (uchar to float)

vladov3000:
Here is a minimal reproduction:
```
from tinygrad import Tensor
[...]

PR #3486, opened 2024-02-23
Add an assert where there should be one

fijal:
This makes #3482 fail cleanly. However, it also makes tests fail in general (IMO they should fail though). Seems related to changes done in #3421 

I'm going to need a bit more guidance on how to continue from here. Would adding supporting casts from disk be the right option?

### DIFF
--- a/tinygrad/runtime/ops_disk.py
+++ b/tinygrad/runtime/ops_disk.py
@@ -53,6 +53,7 @@ class DiskRunner(JITRunner):
     # TODO: there shouldn't actually be casts here, bitcasts should fold into the load
     if ast.src[0].op == UnaryOps.CAST:
       top_src = ast.src[0].src[0]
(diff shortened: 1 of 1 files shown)


chenyuxyz:
this breaks many test cases that should work

geohot:
Yea this just isn't right. You want to change disk to use bitcast, not assert all casts are the same size!

davidjanoskyrepo:
If I have a uint8 and I want it to be vectorized into a uint32 as [uint8]*4 or visav why would that be illegal? Did you want to do something with count here? Also this seems like a symptom not a root cause, if it only affects metal then I would look in the render in cstyles to see if anything odd happens with the data before this point specifically if it is uchar try to see why it might 1.) Be a legal cast treated poorly or 2.) Be an illegal cast that might have been generated poorly

PR #3492, opened 2024-02-24
float64 function support for HIP

### LINKED ISSUE #3304: exp loses precision

chenyuxyz:
follow up of #3420, fixed #3304 on HIP

tested on tinybox. @Qazalin fyi enumerator does not support these

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -168,7 +168,8 @@ class TestFloatDType(TestDType):
 
 class TestDoubleDtype(TestDType):
   DTYPE = dtypes.double
-  @unittest.skipIf(getenv("CUDACPU",0)==1, "conversion not supported on CUDACPU")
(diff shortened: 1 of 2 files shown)


Qazalin:
alright - i'll add it soon!

PR #3494, opened 2024-02-25
Python ALU Ops Dtypes

davidjanoskyrepo:
Adds:
- Consideration of dtypes when performing alu ops in ops_python emu.
- Necessary helper functions for this and modified exec_alu.
- Consideration of vectorized operands/result was not made but maintains old behaviors.
- Rounding of floats in limited scenarios (can be disabled) which assumes IEEE standard precision.
- Overflow checking on non-floats.

### DIFF
--- a/tinygrad/dtype.py
+++ b/tinygrad/dtype.py
@@ -1,7 +1,9 @@
+import math
 from typing import Final, Optional, ClassVar, Set, Tuple, Dict, Union
 from dataclasses import dataclass
 import numpy as np  # TODO: remove numpy
 import functools
+from tinygrad.helpers import getenv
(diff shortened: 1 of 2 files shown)


davidjanoskyrepo:
Is the uops tests timing out? Locally the tests take ages and seems like it times out after 20mins, there are like 3 tests that take the majority of the time. I might be able to speed it up if I don't copy where inputs.

davidjanoskyrepo:
Closing again, probably not going to pick this back up. There are a couple oddities that make strict type enforcement difficult. Maybe just the cast at the end is possible for now. Also uops test slow.

PR #3501, opened 2024-02-26
GPU (OpenCL) parallel beam support

anuraglamsal:
When using parallel beam for GPU (OpenCL) kernels, you get pickling error in the current implementation. This seems to be because CLDevice (and CLCompiler too) deal with ctypes objects containing pointers, and these aren't picklable. So, instead of passing the compiler object from within the beam search function, I access the compiler object from within the compile function directly. This got rid of the pickling error. I ran "handcode_resnet50_opt.py" with BEAM=1 after making these changes, and got 2x speed improvement as compared to using non-parallel beam (i.e. PARALLEL=0) -- parallel beam worked basically. I tried it multiple times, and it seems to work fine. Maybe the current implementation was intended, but I can't seem to figure out why. My guess is that the current implementation prevents creating copies of the Device object in each process, but is this significant?


geohot:
Err, this isn't generic. Because OpenCL requires the device context, you can't really safely parallelize it without opening a new context in each process.

PR #3504, opened 2024-02-26
check buffers are accessible by other gpu before transfer

nimlgen:
Fixes the case where a device is initialized after the creation of a buffer, which is then used as a source in a transfer operation for the device. Check that all known devices have access to the buffer prior to initiating the transfer.


geohot:
Is this slower? I bet that's mapping it in the page table of the other device

geohot:
Merged it because the current behavior is wrong, but it probably is slower

PR #3505, opened 2024-02-26
fix: align .split, .chunk and .unsqueeze with torch, add fuzz tests

dfriehs:
This fixes `.split` where `self.shape[dim]` is not perfectly divisible by `sizes` - `.chunk` is always the wrong choice here:

`tensor((5,)).split(4)` should result in `(tensor((4,)), tensor((1,)))`, was `(tensor((3,)), tensor((2,)))`

This also fixes issues in `.split` and `.chunk` where tensors with `shape[dim] == 0` lead to empty tuples/lists when the tensor itself should have been returned instead.

All included tests are hypothesis fuzz-tests and written with the idea in mind that if torch rises an error then so should tinygrad - otherwise all outputs must match. Because of this tinygrad will now be strict regarding `sizes` having to sum up to passed dimension in `.split`, `num` having to be non-null for `.chunk` and only allowing valid dims in `.unsqueeze`.

---

I am not sure if the testcases are in the right directory and file, or if any existing helper functions should have been used instead. Please tell me should anything be moved or rewritten to use other functions.

I am also not sure on the name or location of `_resolve_dim`. I think adding the function allows for nicer errors to be returned and the four modified functions are by far not the only ones able to use it.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -499,27 +499,31 @@ class Tensor:
     final_shape = [r*s for r,s in zip(repeats, base_shape)]
     return self.reshape(new_shape).expand(expand_shape).reshape(final_shape)
 
+  def _resolve_dim(self, dim:int, *, outer:bool=False) -> int:
+    if not -max(1, self.ndim+outer) <= dim < max(1, self.ndim+outer):
+      raise IndexError(f"{dim=} out of range {[-max(1, self.ndim+outer), max(1, self.ndim+outer)-1]}")
+    return dim + self.ndim+outer if dim < 0 else dim
+
   def split(self, sizes:Union[int, List[int]], dim:int=0) -> Tuple[Tensor, ...]:
     assert all_int(self.shape), f"does not support symbolic shape {self.shape}"
-    dim = dim + self.ndim if dim < 0 else dim
-    if isinstance(sizes, int): return tuple(self.chunk(math.ceil(self.shape[dim]/sizes), dim=dim))
+    dim = self._resolve_dim(dim)
+    if isinstance(sizes, int): sizes = [min(sizes, self.shape[dim]-i) for i in range(0, max(1, self.shape[dim]), max(1, sizes))]
+    assert sum(sizes) == self.shape[dim], f"expect sizes to sum exactly to {self.shape[dim]}, but got {sum(sizes)}"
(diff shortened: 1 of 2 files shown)


geohot:
Cool, good work!

Is this test intended to run in CI? Why create a new folder, just `test_fuzz_shape_ops.py`?

tinygrad/tensor.py
```diff
@@ -499,27 +499,30 @@ def repeat(self, repeats:Sequence[int]) -> Tensor:
     final_shape = [r*s for r,s in zip(repeats, base_shape)]
     return self.reshape(new_shape).expand(expand_shape).reshape(final_shape)
 
+  def _resolve_dim(self, dim:int, *, outer:bool=False) -> int:
```
geohot, line 502: I think this logic can be used in reduce also

dfriehs: Also in flatten, unflatten and probably multiple others. I was planning to go over them one by one in the near future whenever I have time to test.

Should I instead change them all in this PR?

dfriehs:
The tests can probably be skipped on the CI if they would take too long - most functionality should be tested by `test_ops.py` already. I am also not sure if I took out all flakyness already as I don't know if torch requires some integers to be 32-bit - theres one error I got in the meantime where torch can't handle very large `dim`s for `chunk`.

The extra folder was both to be able to easily exclude these tests and to group them when I hopefully find time for more. Would you prefer `test_fuzz_shape_ops.py`?

geohot:
Yea, `test_fuzz_shape_ops` and make it run at least a little in CI

dfriehs:
I think limiting fuzzing to 32 bit values fixed the tests being flaky as I got no errors when testing with `max_examples=25_000` locally.

tinygrad/tensor.py
```diff
@@ -499,27 +499,30 @@ def repeat(self, repeats:Sequence[int]) -> Tensor:
...
+    if not -max(1, self.ndim+outer) <= dim < max(1, self.ndim+outer): raise IndexError(f"{dim=} out of range {[-max(1, self.ndim+outer), max(1, self.ndim+outer)-1]}")  # noqa: E501
```
geohot, line 503: No new code with noqa: E501

dfriehs:
Values very close to MAX_INT64: `torch.zeros((3,)).chunk(9223372036854775806, dim=0)` would raise `RuntimeError: split expects split_size be non-negative, but got split_size=-1` while `torch.zeros((3,)).chunk(9223372036854775805, dim=0)` would work.

PR #3508, opened 2024-02-26
Use new_dtype.itemsize in disk tensor offset

chenyuxyz:
This fixed hlb_cifar disk tensor dataset construction when there are multiple files with a different dtype (/tmp/cifar_train_y).


geohot:
Ugh did I pick the wrong one?

chenyuxyz:
See issue #3510, this broke cifar dataset gen, but yea we don't have test around this.

PR #3512, opened 2024-02-27
no cuda compile helper

nimlgen:
This should be cleaner and less lines


geohot:
Sweet, I never liked those in helpers

PR #3516, opened 2024-02-28
use hash as key for beam

wozeparrot:
sha256 should be good enough


geohot:
Can you bump the database version number?

PR #3519, opened 2024-02-28
`skipIf(not(` -> `skipUnless(` in test_linearizer_failures

chenyuxyz:
if these behaves weirdly in CI might need to disable them in CI


chenyuxyz:
hmm seems to run fine.

PR #3522, opened 2024-02-28
Revert "check buffers are accessible by other gpu before transfer"

geohot:
Reverts tinygrad/tinygrad#3504


geohot:
@nimlgen never mind, it breaks `JIT=2 HSA=1 python3 examples/llama.py  --temperature=0 --count=50 --prompt="Hello." --timing --shard 2`

PR #3524, opened 2024-02-28
change if else to ternary op in tensor.py

CalebBunch:


CalebBunch:
bad pr forget to change indents

PR #3526, opened 2024-02-28
Runtime:[CLANG][Update] faster and cleaner compile(), remove unecessa…

Maroc-OS:
…ry str() wraps.

Update Clang runtime to use subprocess.run instead of subprocess.check_output is faster and follow's new API for modern python, and add new command list for better readability.

usage of str() was never necessary.

the following is faster in my machine by 1 second and some fractions.
 
```
➜  tinygrad_latest git:(master) ✗ PYTHONPATH='.' PYOPENCL_COMPILER_OUTPUT=1 CLANG=1 DISABLE_COMPILER_CACHE=1 python3.11 -O test/models/test_efficientnet.py TestEfficientNet.test_car
.
Ran 1 test in 7.596s

OK
```

### DIFF
--- a/tinygrad/runtime/ops_clang.py
+++ b/tinygrad/runtime/ops_clang.py
@@ -12,8 +12,8 @@ class ClangCompiler(Compiler):
   def compile(self, src:str) -> bytes:
     # TODO: remove file write. sadly clang doesn't like the use of /dev/stdout here
     with tempfile.NamedTemporaryFile(delete=True) as output_file:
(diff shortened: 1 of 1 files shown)


geohot:
Effectively a whitespace change, closing.

geohot:
Hmm, actually, maybe it's not launching bash.

geohot:
Nope, check_output doesn't use shell either. Closing

PR #3530, opened 2024-02-29
O(n) arange attempt

patosai:
The insight I use here is that certain loops accumulate using only the loop index, consts or specials (which are const-like?), so I can symbolically evaluate one iteration of the loop and evaluate the entire summation as a function of the loop index. A general expression $\sum_i {f(i)}$ or a $\sum_i {f(i) < x}$ can't be reduced, but if we assume a certain form of $f(i) = ai+b$, then it is possible to reduce. For $\sum_i {ai+b}$ this is pretty straightforward, but I see that arange-like functions have a CMPLT which makes reducing much harder, here are the expressions I came up for those:

$$\begin{aligned} \sum_i^j {(\alpha i + \beta) < \gamma} &= \sum_i^j i < \frac{\gamma - \beta}{\alpha} \\
&= \text{clamp}(0, \left\lceil \frac{\gamma - \beta}{\alpha} \right\rceil - i, j-i+1)
\end{aligned}$$

and similarly:

$$\begin{aligned} \sum_i^j {(\alpha i + \beta) > \gamma} &= \sum_i^j i > \frac{\gamma - \beta}{\alpha} \\
&= \text{clamp}(0, j - \left\lfloor \frac{\gamma - \beta}{\alpha} \right\rfloor, j-i+1)
\end{aligned}$$

Here $\text{clamp(min, val, max)}$ means to limit val between min and max, inclusive.

I'm sure nonlinear forms of $f(i)$ work too but I didn't do that in this PR as it would probably get complicated. Also the code still needs cleaning up but I had this thing working for a few simple examples of arange so I want to see what people think first. These are the prod ops for `Tensor.arange(1, 78, 2)`

```
   0 UOps.DEFINE_GLOBAL  : ptr.dtypes.int            []                               data0
   1 UOps.SPECIAL        : dtypes.int                []                               (0, 'gidx0', 13)
   2 UOps.SPECIAL        : dtypes.int                []                               (0, 'lidx1', 3)
   3 UOps.DEFINE_ACC     : dtypes.int                []                               0
   4 UOps.CONST          : dtypes.int                []                               0
   5 UOps.CONST          : dtypes.int                []                               39
   6 UOps.CONST          : dtypes.int                []                               2
   7 UOps.CONST          : dtypes.int                []                               -3
   8 UOps.ALU            : dtypes.int                [1, 7]                           BinaryOps.MUL
   9 UOps.CONST          : dtypes.int                []                               -1
  10 UOps.ALU            : dtypes.int                [2, 9]                           BinaryOps.MUL
  11 UOps.ALU            : dtypes.int                [8, 10]                          BinaryOps.ADD
  12 UOps.CONST          : dtypes.int                []                               -37
  13 UOps.CONST          : dtypes.int                []                               3
  14 UOps.ALU            : dtypes.int                [1, 13]                          BinaryOps.MUL
  15 UOps.ALU            : dtypes.int                [14, 2]                          BinaryOps.ADD
  16 UOps.LOOP           : dtypes.int                [4, 5]                           None
  17 UOps.ALU            : dtypes.int                [16, 9]                          BinaryOps.MUL
  18 UOps.ALU            : dtypes.int                [11, 17]                         BinaryOps.ADD
  19 UOps.ALU            : dtypes.bool               [18, 12]                         BinaryOps.CMPLT
  20 UOps.ALU            : dtypes.int                [19, 6, 4]                       TernaryOps.WHERE
  21 UOps.ALU            : dtypes.int                [20, 3]                          BinaryOps.ADD
  22 UOps.PHI            : dtypes.int                [3, 21, 16]                      None
  23 UOps.ENDLOOP        :                           [16]                             None
  24 UOps.ALU            : dtypes.int                [22, 9]                          BinaryOps.ADD
  25 UOps.STORE          :                           [0, 15, 24]                      None
```

and this is what it gets reduced to after this optimization:

```
   0 UOps.DEFINE_GLOBAL  : ptr.dtypes.int            []                               data0
   1 UOps.SPECIAL        : dtypes.int                []                               (0, 'gidx0', 13)
   2 UOps.SPECIAL        : dtypes.int                []                               (0, 'lidx1', 3)
   3 UOps.CONST          : dtypes.int                []                               0
   4 UOps.CONST          : dtypes.int                []                               -1
   5 UOps.CONST          : dtypes.int                []                               3
   6 UOps.ALU            : dtypes.int                [1, 5]                           BinaryOps.MUL
   7 UOps.ALU            : dtypes.int                [6, 2]                           BinaryOps.ADD
   8 UOps.CONST          : dtypes.int                []                               1
   9 UOps.ALU            : dtypes.int                [7, 8]                           BinaryOps.ADD
  10 UOps.ALU            : dtypes.int                [9, 3]                           BinaryOps.MAX
  11 UOps.ALU            : dtypes.int                [10, 4]                          BinaryOps.MUL
  12 UOps.CONST          : dtypes.int                []                               -39
  13 UOps.ALU            : dtypes.int                [11, 12]                         BinaryOps.MAX
  14 UOps.CONST          : dtypes.int                []                               -2
  15 UOps.ALU            : dtypes.int                [13, 14]                         BinaryOps.MUL
  16 UOps.ALU            : dtypes.int                [15, 4]                          BinaryOps.ADD
  17 UOps.STORE          :                           [0, 7, 16]                       None
```
so you can see the loop is gone.

For arange(256):
```
   0 UOps.DEFINE_GLOBAL  : ptr.dtypes.int            []                               data0
   1 UOps.DEFINE_LOCAL   : ptr.dtypes.int            []                               ('temp', 16)
   2 UOps.SPECIAL        : dtypes.int                []                               (0, 'gidx0', 256)
   3 UOps.SPECIAL        : dtypes.int                []                               (0, 'lidx1', 16)
   4 UOps.DEFINE_ACC     : dtypes.int                []                               0
   5 UOps.CONST          : dtypes.int                []                               0
   6 UOps.CONST          : dtypes.int                []                               16
   7 UOps.CONST          : dtypes.int                []                               1
   8 UOps.CONST          : dtypes.int                []                               -1
   9 UOps.ALU            : dtypes.int                [2, 8]                           BinaryOps.MUL
  10 UOps.CONST          : dtypes.int                []                               -16
  11 UOps.ALU            : dtypes.int                [3, 10]                          BinaryOps.MUL
  12 UOps.ALU            : dtypes.int                [9, 11]                          BinaryOps.ADD
  13 UOps.CONST          : dtypes.int                []                               -254
  14 UOps.ALU            : dtypes.bool               [3, 7]                           BinaryOps.CMPLT
  15 UOps.LOOP           : dtypes.int                [5, 6]                           None
  16 UOps.ALU            : dtypes.int                [15, 8]                          BinaryOps.MUL
  17 UOps.ALU            : dtypes.int                [12, 16]                         BinaryOps.ADD
  18 UOps.ALU            : dtypes.bool               [17, 13]                         BinaryOps.CMPLT
  19 UOps.ALU            : dtypes.int                [18, 7, 5]                       TernaryOps.WHERE
  20 UOps.ALU            : dtypes.int                [19, 4]                          BinaryOps.ADD
  21 UOps.PHI            : dtypes.int                [4, 20, 15]                      None
  22 UOps.ENDLOOP        :                           [15]                             None
  23 UOps.STORE          :                           [1, 3, 21]                       None
  24 UOps.BARRIER        :                           [23]                             None
  25 UOps.IF             :                           [14, 24]                         None
  26 UOps.DEFINE_ACC     : dtypes.int                []                               0
  27 UOps.LOOP           : dtypes.int                [5, 6]                           None
  28 UOps.LOAD           : dtypes.int                [1, 27, 25]                      None
  29 UOps.ALU            : dtypes.int                [28, 26]                         BinaryOps.ADD
  30 UOps.PHI            : dtypes.int                [26, 29, 27]                     None
  31 UOps.ENDLOOP        :                           [27]                             None
  32 UOps.ALU            : dtypes.int                [30, 8]                          BinaryOps.ADD
  33 UOps.STORE          :                           [0, 2, 32]                       None
  34 UOps.ENDIF          :                           [25]                             None
```

after optimization:
```
   0 UOps.DEFINE_GLOBAL  : ptr.dtypes.int            []                               data0
   1 UOps.DEFINE_LOCAL   : ptr.dtypes.int            []                               ('temp', 16)
   2 UOps.SPECIAL        : dtypes.int                []                               (0, 'gidx0', 256)
   3 UOps.SPECIAL        : dtypes.int                []                               (0, 'lidx1', 16)
   4 UOps.CONST          : dtypes.int                []                               0
   5 UOps.CONST          : dtypes.int                []                               16
   6 UOps.CONST          : dtypes.int                []                               1
   7 UOps.CONST          : dtypes.int                []                               -1
   8 UOps.CONST          : dtypes.int                []                               -16
   9 UOps.ALU            : dtypes.bool               [3, 6]                           BinaryOps.CMPLT
  10 UOps.ALU            : dtypes.int                [3, 5]                           BinaryOps.MUL
  11 UOps.ALU            : dtypes.int                [10, 2]                          BinaryOps.ADD
  12 UOps.CONST          : dtypes.int                []                               -239
  13 UOps.ALU            : dtypes.int                [11, 12]                         BinaryOps.ADD
  14 UOps.ALU            : dtypes.int                [13, 4]                          BinaryOps.MAX
  15 UOps.ALU            : dtypes.int                [14, 7]                          BinaryOps.MUL
  16 UOps.ALU            : dtypes.int                [15, 8]                          BinaryOps.MAX
  17 UOps.ALU            : dtypes.int                [16, 7]                          BinaryOps.MUL
  18 UOps.STORE          :                           [1, 3, 17]                       None
  19 UOps.BARRIER        :                           [18]                             None
  20 UOps.IF             :                           [9, 19]                          None
  21 UOps.DEFINE_ACC     : dtypes.int                []                               0
  22 UOps.LOOP           : dtypes.int                [4, 5]                           None
  23 UOps.LOAD           : dtypes.int                [1, 22, 20]                      None
  24 UOps.ALU            : dtypes.int                [23, 21]                         BinaryOps.ADD
  25 UOps.PHI            : dtypes.int                [21, 24, 22]                     None
  26 UOps.ENDLOOP        :                           [22]                             None
  27 UOps.ALU            : dtypes.int                [25, 7]                          BinaryOps.ADD
  28 UOps.STORE          :                           [0, 2, 27]                       None
  29 UOps.ENDIF          :                           [20]                             None
```
the first loop is gone but the second loop inside hasn't been optimized yet, there's a LOAD so I'm not sure if that can be treated as const-like right now.

It doesn't work for float aranges just yet, need to fix that, want to see what people think first


geohot:
Cool, bounty locked!

The code looks very complex. I have an old version of this here and it's shorter, why? https://github.com/tinygrad/tinygrad/pull/2318

patosai:
Wanted to know first if we thought this was a good approach, will be simplifying it soon

geohot:
Why do you want floats in symbolic in the first place?

patosai:
> Why do you want floats in symbolic in the first place? Symbolic is really designed for indexing.

Integer arange ops work without this, but if you want to do a float arange, ex. arange(5.5, 178.5, 2.5), the loop does something like 2.5 * (loop < gidx0*a + lidx1*b), and in order to generate the correct ops, the symbolic lib needs to be able to handle a float multiplication thus why I added the floats in there.

So we could presumably ship an integer-only arange without changing the symbolic library, and I get that the symbolic library was supposed to be for shapetracker which is int-only. I just happened to use the existing symbolic library to evaluate the loop which I thought was good reuse of the code. So I guess the decision is have floats in symbolic, or no floats and I evaluate the loop to generate the ops in some other way which would look like the existing symbolic library but w/ floats.

patosai:
Actually that's brilliant, what I did was I changed arange to do the cumsum with ints only and then multiply by the step so symbolic can be int only

patosai:
@geohot I think this is ready for review. There are some things which could be separated into different PRs:

1. MaxNode addition as you mentioned
2. fix for symbolic negative floordiv, see 
https://github.com/tinygrad/tinygrad/pull/3530/files#diff-5a93626a00a4d42205c5aa753f088cb71421561ef8b1376f0bb58db0e558a8c0R59
https://github.com/tinygrad/tinygrad/pull/3530/files#diff-ee59e34646985312c128bec4ab8b2116615446d65b5b41c57aa1d821caa4ba5cR125
and related is
https://github.com/tinygrad/tinygrad/pull/3530/files#diff-91ca5e2e75ef3ea1982c8ca6cc175ee88f20efa0d8e4b96f305b970dc6df71e7R40
since integer division rounds towards 0, but floor rounds toward negative infinity


3. the stuff which actually optimizes the arange loop

The only thing failing is the line limit, this takes the total line limit way over 6000, even just merging MaxNode would be >6000 already

geohot:
Why do you need MaxNode?

patosai:
I added the MaxNode as a way to implement the clamp - need the clamp otherwise the optimization doesn't work. That being said probably nothing else in the codebase is gonna use a MaxNode or a variable factorization so I will probably make the code less general to make it smaller

patosai:
I simplified the factoring to be specific to only this use case, and implemented the clamp w/o symbolic. The symbolic lib solves the issue of dividing where the numerator is possibly negative - the formula needs the floor of the division, but integer division rounds towards 0 so positive numerator integer division gives the right answer, but negative doesn't without this
https://github.com/tinygrad/tinygrad/blob/master/tinygrad/shape/symbolic.py#L62-L67

patosai:
gonna attempt to remove the cumsum 2-stage with this

geohot:
Woah, this works for cumsum too? Is that behavior tested?

tinygrad/codegen/uops.py
```diff
@@ -179,7 +182,84 @@ def fix_loop_scope(self, get_recursive_parents:Callable[..., Set[UOp]]):
...
+    modified_phis = set()
+    for phi in [op for op in self.uops if op.uop is UOps.PHI]:
+      where_ops = [op for op in get_recursive_parents(phi) if op.arg == TernaryOps.WHERE and op.dtype is not None and dtypes.is_int(op.dtype)]
+      for where in sorted(where_ops, key=lambda x: self.uops.index(x)):
+        try:
+          loop_op = phi.vin[2]
+          comp, comp_lt, comp_gt = where.vin[0], where.vin[0].vin[0], where.vin[0].vin[1]
+          if where.vin[2].arg != 0 or comp.arg != BinaryOps.CMPLT or comp_gt.uop is not UOps.CONST or comp_gt.arg > 0: continue
+          factored = loop_factor(get_recursive_parents, loop_to_name, comp_lt, NumNode(comp_gt.arg), loop_op)
+        except (RuntimeError, StopIteration, IndexError, AssertionError): continue
```
geohot, line 220: Hmm, this line looks sketchy

geohot:
Btw, I'm glad this was done without MaxNode.

tinygrad/tensor.py
```diff
@@ -235,7 +235,7 @@ def ones(*shape, **kwargs): return Tensor.full(argfix(*shape), 1.0, **kwargs)
   def arange(start, stop=None, step=1, **kwargs):
     if stop is None: stop, start = start, 0
     dtype = kwargs.pop("dtype", dtypes.default_float if any(isinstance(x, float) for x in (start, stop, step)) else dtypes.default_int)
-    return (Tensor.full((math.ceil((stop-start)/step),), step, dtype=dtype, **kwargs).cumsum() + (start - step)).cast(dtype)
```
geohot, line 238: Why could the old way not simplify? And why can only ints simplify?

geohot:
Testing this on Mac, first thing I tried didn't behave as expected. Shouldn't this be simplified?

```
diane@blak:~/tinygrad$ FORWARD_ONLY=1 DEBUG=4 python3 test/test_ops.py TestOps.test_arange_big
test_arange_big (__main__.TestOps.test_arange_big) ...   0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(256, 1), strides=(1, 0), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ SUB 
  2    ┣━┳ SUM (1,)
  3    ┃ ┗━━ CONST ConstBuffer(val=1.0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(257, 511), strides=(0, 0), offset=0, mask=((0, 257), (255, 511)), contiguous=False), View(shape=(256, 256), strides=(1, 512), offset=0, mask=None, contiguous=False))))
  4    ┗━━ CONST ConstBuffer(val=1.0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(256, 1), strides=(0, 0), offset=0, mask=None, contiguous=False),)))
reduced UOp count from 36 to 35
reduced UOp count from 35 to 33
reduced UOp count from 33 to 31
reduced UOp count from 31 to 28
#include <metal_stdlib>
using namespace metal;
kernel void r_256_16_16(device int* data0, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  threadgroup int temp[16];
  int gidx0 = gid.x; /* 256 */
  int lidx1 = lid.x; /* 16 */
  int alu0 = max(((lidx1*16)+gidx0+(-239)),0);
  int alu1 = max((-16),(alu0*(-1)));
  *(temp+lidx1) = (alu1*(-1));
  threadgroup_barrier(mem_flags::mem_threadgroup);
  if ((lidx1<1)) {
    int acc0 = 0;
      int val0 = *(temp+ridx0);
      acc0 = (val0+acc0);
    }
    *(data0+gidx0) = (acc0-1);
  }
}
*** METAL      1 r_256_16_16                            arg   1 mem  0.00 GB tm     10.12us/     0.01ms (    6.50 GFLOPS,    0.10 GB/s)
```

geohot:
cumsum isn't what I expected either.

FORWARD_ONLY=1 DEBUG=4 python3 test/test_ops.py TestOps.test_simple_cumsum

```C
#include <metal_stdlib>
using namespace metal;
kernel void r_511_2_1022(device float* data0, const device float* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 511 */
  int lidx1 = lid.x; /* 2 */
  float acc0 = 0.0f;
  int alu0 = ((gidx0*2)+lidx1);
  for (int ridx0 = 0; ridx0 < 1022; ridx0++) {
    float val0 = ((((gidx0*(-2))+(lidx1*(-1))+(ridx0*(-1)))<(-1020))?*(data1+alu0+ridx0+(-1021)):0.0f);
    acc0 = (val0+acc0);
  }
  *(data0+alu0) = acc0;
}
```

geohot:
Okay, to summarize:
1. That Exception thing is sketchy, I'm sure there's a better way to write that.
2. The big arange being split by the optimizer is outside the scope of this bounty. Confirmed it's okay with `NOOPT=1`
3. You removed the cumsum splitting logic, but it's not optimizing cumsum. This will cause slowness. Did you expect that `TestOps.test_simple_cumsum` to be optimized?

tinygrad/codegen/uops.py
```diff
@@ -194,26 +274,8 @@ def get_recursive_parents(x:UOp, with_phi=False) -> Set[UOp]:
...
-            del self.saved_exprs[(u.uop, u.dtype, u.vin, u.arg)]
-            # NOTE: assuming u.vin[2].vin[1] and u.vin[2].vin[0] have the same dtype
-            loop_len = self.add(UOps.ALU, u.vin[2].vin[1].dtype, (u.vin[2].vin[1], u.vin[2].vin[0]), BinaryOps.SUB, insert_before=self.uops.index(u))
-            if loop_len.dtype != u.dtype: loop_len = self.add(UOps.CAST, u.dtype, (loop_len,), insert_before=self.uops.index(u))
-            new = self.add(UOps.ALU, u.dtype, (u.vin[1], loop_len,), BinaryOps.MUL, insert_before=self.uops.index(u))
-            # replace u with new
-            for v in self.uops: v.vin = tuple(new if x is u else x for x in v.vin)
-            self.uops.remove(u)
-            changed_something = True
+    keep_doing_if_did_something(lambda: self.uops_optimization(get_recursive_parents))
```
geohot, line 277: Do you need this helper?

`while self.uops_optimization(get_recursive_parents): pass` is the same thing, right?

geohot:
So I don't think you can remove the cumsum splitting logic at all? There's not a way to optimize that with only UOps that I know about. I believe all the major libraries do it in 2 stages.

Instead, just modify arange to call the underlying `_cumsum`

tinygrad/codegen/uops.py
```diff
@@ -179,7 +182,84 @@ def fix_loop_scope(self, get_recursive_parents:Callable[..., Set[UOp]]):
...
+          loop_op = phi.vin[2]
+          comp, comp_lt, comp_gt = where.vin[0], where.vin[0].vin[0], where.vin[0].vin[1]
+          if where.vin[2].arg != 0 or comp.arg != BinaryOps.CMPLT or comp_gt.uop is not UOps.CONST or comp_gt.arg > 0: continue
+          factored = loop_factor(get_recursive_parents, loop_to_name, comp_lt, NumNode(comp_gt.arg), loop_op)
+        except (RuntimeError, StopIteration, IndexError, AssertionError): continue
+        final_value = NumNode(loop_op.vin[1].arg-1) - factored
+        self.uops, after_split_ops = self.uops[:(where_index:=self.uops.index(where))], self.uops[where_index:]
+        rendered = final_value.render(render_ops, ctx)
+        loop_length = loop_op.vin[1].arg - loop_op.vin[0].arg
+        clamped = mult_neg_1(max(const(-1*loop_length), mult_neg_1(max(rendered, const(0)))))
```
geohot, line 225: Do you always need this max? Can't you check final_value.min and if it's >0 you can skip?

patosai:
1. I was trying to figure out why >256 arange didn't optimize the same way, thought cumsum could use the same optimization w/o 2 stage but yeah it doesn't work right now, gonna revert that one
2. You're right that the min/max clamping doesn't always need to happen, 95% of the cases it doesn't even do anything
3. The Exception came from the initlal attempt where I had like 10 lines of code checking if the loop satisfied the optimization requirements (linear equation, integer output, constants/loop vars only), so instead this time I try factoring it as a linear equation and if it doesn't work/throws, nothing happens. Yes there's probably a prettier way to do this
4. Float outputs aren't supported because the symbolic lib doesn't support floats. I use the symbolic lib because it does all the stuff I need (variables/negative division fixing/op rendering). I could write my own symbolic lib impl for this but the logic would be a duplication of symbolic which would suck

tinygrad/codegen/uops.py
patosai: it seems so obvious now 🤦

patosai:
hmm now that I look at it, the consts are all insts so the optimization up to the CMPLT should be possible with just ints, then I manually do a multiply w a possible cast

patosai:
ok. works for float/int by just factoring the CMPLT portion which is int only, doesn't have any weird assertion catches, does MAX for min/max optionally depending on symbolic min/max

I ended up not using the `while (): pass` thing for my optimizer fn but I left it in for uoptimize

patosai:
other simplifications I thought were interesting:
`DEBUG=4 python3 test/external/external_test_onnx_backend.py OnnxBackendNodeModelTest.test_averagepool_2d_precomputed_pads_cpu` first kernel
`DEBUG=4 python3 test/test_ops.py TestOps.test_sum_pad_collapse`

geohot:
Very close. Two more things.

1) I see one test confirming this optimization happens, but is there any other good way correctness can be tested? It's fine if the answer is no, but it's worth thinking about.

2) `NOOPT=1 FORWARD_ONLY=1 DEBUG=4 python3 test/test_ops.py TestOps.test_arange_big`

```C
#include <metal_stdlib>
using namespace metal;
kernel void r_256_256(device int* data0, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 256 */
  int alu0 = max((-256),(-(gidx0+1)));
  *(data0+gidx0) = ((-alu0)+(-1));
}
```

Now there's only 1 max, which is better than 2, but I'm still not sure why you need that one. The largest value `gidx0` can have is 255, meaning the smallest value `-(gidx0+1))` can have is `-(255+1) = -256`, aka that max is a no-op. How come the code doesn't see this?

patosai:
Had an off by one error on the symbolic variable maxes.

As for testing, the optimization targets UOps.PHI/TernaryOps.WHERE so I added tests for the absence of those ops (where ops can stick around if they're used for other stuff, not PHIs though). Ideally I would be able to linearize with and without this opt and check that loop/phi/where counts went down, but I think checking that they don't exist is fine for now. I'm not sure how to test for the unnecessary MAX ops other than have a handwritten test on specific tensors where we know what the number of ops should be, so that's what I did

geohot:
Tests + off by one error good.

Oh I missed something, why is all that state being passed in from the Linearizer? I really want to avoid passing state like this, particularly generic things like "ctx"

tinygrad/codegen/linearizer.py
```diff
@@ -391,7 +391,7 @@ def upcast_strides(buf:int):
     self.global_store(0, global_idxs+local_idxs+fake_reduce_idxs+upcast_idxs, val)
 
     # optimize the uops
-    self.uops.uoptimize()
+    self.uops.uoptimize({op: name for name, op in self.loop_uops.items()}, self.render_ops, self)
```
geohot, line 394: This state. It's a lot harder to reason about a program when state like this is passed

geohot:
Tests failing now (some interaction with constant folding?)

patosai:
I looked at the test and `TestLinearizerFailures::test_failure_6` is failing because there's a const in the test which has an int dtype but its arg is 1.0 - I had gotten around the issue before by using the constant itself instead of trying to perform an op on it, but in this case the constant folding is failing cuz it thinks it's an int but the arg is a float. Seems like a garbage input which shouldn't be, I added cast_scalar to the constant folding instead of changing the test.

As for that state, I wanted to pass the loop op name to the optimizer so the loops can be resolved correctly. Passing linearizer.loop_uops seemed like an overreach of state so I passed a modified map in - I can see how it can be confusing though if you have to look at multiple levels of modification. The ops don't have loop name so are you suggesting passing in linearizer.loop_uops by itself?

test/test_linearizer_failures.py
```diff
@@ -75,7 +75,7 @@ def test_failure_5(self):
...
 
   def test_failure_6(self):
-    ast = LazyOp(op=BinaryOps.ADD, src=(LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BufferOps.CONST, src=(), arg=ConstBuffer(val=-1.0, dtype=dtypes.int32, st=ShapeTracker(views=(View(shape=(11, 19), strides=(0, 0), offset=0, mask=((0, 11), (9, 19)), contiguous=False), View(shape=(10, 10), strides=(1, 20), offset=0, mask=None, contiguous=False))))),), arg=(1,)), LazyOp(op=BufferOps.CONST, src=(), arg=ConstBuffer(val=10.0, dtype=dtypes.int32, st=ShapeTracker(views=(View(shape=(10, 1), strides=(0, 0), offset=0, mask=None, contiguous=False),))))), arg=None)
+    ast = LazyOp(op=BinaryOps.ADD, src=(LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BufferOps.CONST, src=(), arg=ConstBuffer(val=-1, dtype=dtypes.int32, st=ShapeTracker(views=(View(shape=(11, 19), strides=(0, 0), offset=0, mask=((0, 11), (9, 19)), contiguous=False), View(shape=(10, 10), strides=(1, 20), offset=0, mask=None, contiguous=False))))),), arg=(1,)), LazyOp(op=BufferOps.CONST, src=(), arg=ConstBuffer(val=10.0, dtype=dtypes.int32, st=ShapeTracker(views=(View(shape=(10, 1), strides=(0, 0), offset=0, mask=None, contiguous=False),))))), arg=None)
```
patosai, line 78: `ConstBuffer(val=-1.0, dtype=dtypes.int32...` should have been `ConstBuffer(val=-1, dtype=dtypes.int32` imo

patosai:
I passed loop_uops unmodified into the phi simplifier fn and did the map inversion there

test/test_linearizer_failures.py
chenyuxyz: that was an artifact of the old const, I think it's fine to update this test

tinygrad/codegen/uops.py
```diff
@@ -180,6 +180,92 @@ def fix_loop_scope(self, get_recursive_parents:Callable[..., Set[UOp]]):
...
+    def neg(x): return self.add(UOps.ALU, dtypes.default_int, (x,), UnaryOps.NEG)
+    def max(x, y): return self.add(UOps.ALU, dtypes.default_int, (x, y), BinaryOps.MAX)
+    render_ops = {Variable: lambda self, ops, _: seen_vars[self.expr], NumNode: lambda self, ops, _: const(self.b),
+                  MulNode: lambda self, ops, _: uop_alu_idx(self.a.render(ops, self), self.b, BinaryOps.MUL),
+                  DivNode: lambda self, ops, _: uop_alu_idx(self.a.render(ops, self), self.b, BinaryOps.DIV),
+                  SumNode: lambda self, ops, _:
+                  functools.reduce(lambda a, b: uop_alu_idx(a, b, BinaryOps.ADD), self.nodes[1:], self.nodes[0].render(ops, self))}
+    def uop_alu_idx(a: UOp, b, op, dtype=dtypes.default_int):
+      render_b: UOp = cast(UOp, (NumNode(b) if not isinstance(b, Node) else b).render(render_ops))
+      return self.add(UOps.ALU, dtype, (a, render_b), op)
```
patosai, line 219: I removed the linearizer arg passing but had to copy this logic into UOps

tinygrad/codegen/uops.py
```diff
@@ -180,6 +180,92 @@ def fix_loop_scope(self, get_recursive_parents:Callable[..., Set[UOp]]):
...
+    def alu_opposite(arg, x, y):
+      if arg == BinaryOps.ADD: return x - y
+      elif arg == BinaryOps.MUL: return Node.__floordiv__(x, y, False)
+      else: raise RuntimeError("unhandled alu")
+    def to_symbolic(u: UOp):
+      if u.uop == UOps.CONST: return NumNode(int(u.arg))
+      elif u.uop in {UOps.LOOP, UOps.SPECIAL}:
+        name = "loop{}".format(len(seen_vars))
+        seen_vars[name] = u
+        return Variable(name, u.vin[0].arg, u.vin[1].arg-1) if u.uop is UOps.LOOP else Variable(name, 0, u.arg[2]-1)
```
patosai, line 198: modified this fn to generate internal loop names so when rendering later, the op can be resolved again

geohot:
Is this ready to merge?

patosai:
Yup

geohot:
Congrats! george@tinygrad.org to claim the bounty, either PayPal or USDC.

There's a $1000 bounty for O(n) embedding, you are halfway there with this. You are welcome to mark the arange reduce as special and fused with others.

PR #3545, opened 2024-02-29
WIP PR trying to lock up mamba bounty

CalebBunch:
Big WIP. Will work on it more this evening. I think this is a good first bounty and i think that I will be able to solve it in a few days max.


geohot:
This is not even close to locking a bounty

CalebBunch:
> This is not even close to locking a bounty

Alright I will keep working on it. Should I implement mamba or mamba_simple? Both?

PR #3555, opened 2024-03-01
WIP Mamba Model Bounty 

CalebBunch:
I finished the basic building blocks of the model's architecture. I need to work on writing a training loop and/or loading pre-trained weights. Which should I do? Load existing weights for an LLM, train it on some dataset like HG38 or both?

EDIT: I did not intend to leave the import torch line there. I was just using it to test against my tinygrad code, and forgot to remove it. Will do it next PR.


geohot:
I have closed this three times. This shouldn't be a WIP, it's way too WIP

Issue #3536, opened 2024-02-29
no known conversion from 'float' to 'hip_bfloat16' for 1st argument


geohot:
`tiny@tiny5:~/tinygrad$ HSA=1 python3 examples/llama.py  --temperature=0 --count=50 --prompt="Hello." --timing --shard 2 --size 70B --gen 2`

```
using HSA backend                                                                                                                                                                                                                                             
using LLaMA-2-70B model                                                                                                        
ram used:  0.00 GB, layers.0.attention.wq.weight                      :   0%|                                                                                                                                                         | 0/724 [00:00<?, ?it/s]
/tmp/comgr-bf1528/input/<null>:65:43: error: no matching conversion for C-style cast from 'float' to 'hip_bfloat16'
  hip_bfloat16 val0 = (alu3?*(data1+alu2):(hip_bfloat16)(0.0f));                                                                                                                                                                                              
                                          ^~~~~~~~~~~~~~~~~~~~                                                                 
/tmp/comgr-bf1528/input/<null>:10:10: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'float' to 'const hip_bfloat16' for 1st argument
  struct hip_bfloat16 { unsigned short data; };                                                                                                                                                                                                               
         ^                                         
/tmp/comgr-bf1528/input/<null>:10:10: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'float' to 'hip_bfloat16' for 1st argument
  struct hip_bfloat16 { unsigned short data; };                                                                                                                                                                                                               
         ^                                                                                                                     
/tmp/comgr-bf1528/input/<null>:10:10: note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 1 was provided  
```

Looks like it's failing trying to cast the const

Zaffer:
my attempt to resolve the issue #3558

Zaffer:
rocm needs to be updated on tinybox.

appears to be many issues with hip_bfloat16 in the rocm github still open: 
https://github.com/ROCm/ROCm/issues?q=is%3Aissue+is%3Aopen+bfloat16

there are multiple issues with the HIPLanguage class in renderers/cstyle.py:

```python
class HIPLanguage(CStyleLanguage):
    kernel_prefix = "#include <hip/hip_runtime.h>\n"
    # Updated launch bounds definition to align with HIP's expected syntax.
    # Additional HIP includes or definitions can be added here as needed.

    code_for_workitem = {
[...]

PR #3558, opened 2024-03-01
Overwrite render_cast method to HIPLanguage for manual float to hip_b…

Zaffer:
Please confirm if I am on the right track with this approach?
Then will add the reversion test for this.


geohot:
No test

Zaffer:
Added test to benchmark.yml as per command in issue #3536

geohot:
Not a benchmark test, a unit test that runs in CI. The problem with benchmark tests is they don't run on contributors

Zaffer:
All working now. Fix is to use built in functions from the hip library.

Had to remove some of the custom functions to avoid reintroduction errors for new driver library here:
https://github.com/ROCm/clr/blob/rocm-6.0.x/hipamd/include/hip/amd_detail/amd_hip_bf16.h

I added a test to test.yml CI that reproduced the exact issue here, but commented it out because it runs on self-hosted tinybox.

chenyuxyz:
can you make minimal change and include a minimal test that runs on CI?

Zaffer:
> can you make minimal change and include a minimal test that runs on CI?

added a minimal test and added explicit test in CI
this is already the most minimal change possible in cstyle.py

summary:
- had to add a render_cast function override in the HIPLangauge class to handle float to bfloat16 type casting in HIP
- to use the built-in HIP conversion function called "__float2bfloat16", so I had to add this to the kernel prefix:
```c
#include <stdint.h>
#include <stddef.h>
#include <hip/amd_detail/amd_hip_bf16.h>
```
> more detail here: https://github.com/ROCm/clr/blob/rocm-6.0.x/hipamd/include/hip/amd_detail/amd_hip_bf16.h
- to add these libraries to the kernel and I had to remove the following custom `make_` functions:
```python
_make_hip_dtype(*x) for x in [("signed int", "int", 2), ("signed int", "int", 4), ("float", "float", 2), ("float", "float", 4)]
```

Zaffer:
first test is the minimal test, simply do a type cast to bfloat16 causes the issue on AMD devices:
```python
def test_f16_to_bf16_minimal(self):
  original_tensor = Tensor([1.0, 2.0, 3.0], dtype=dtypes.float16)
  converted_tensor = original_tensor.cast(dtypes.bfloat16)
  self.assertEqual(converted_tensor.dtype, dtypes.bfloat16)
```

chenyuxyz:
see #3630

Issue #3552, opened 2024-03-01
.softmax().argmax() broken in CLANG


reddyn12:
```python 
t1 = torch.arange(16).reshape(4,4).float()
t2 = Tensor.arange(16).reshape(4,4).float()

o1 = t1.softmax(dim=-1).argmax()
o2 = t2.softmax(axis=-1).argmax()

print(o1, o2.numpy())
```

chenyuxyz:
it's a tie and somehow CLANG breaks ties differently

wozeparrot:
Might be related to #862

reddyn12:
issue inherited with PYTHON

chenyuxyz:
minimal repro
```
a = Tensor([4, 3, 2, 1]).exp().argmax()
print(a.numpy())
```
returns 3 for CLANG and PYTHON

chenyuxyz:
the core issue is that when we do
```
# kernel 1
m = max(exp(a), exp(b))
# kernel 2
c = exp(a) == m
d = exp(b) == m
```
the m is stored in float32, and exp(a) can be float64 so there's mismatch

PR #3559, opened 2024-03-01
Add test for .softmax.argmax

reddyn12:
Issue: #3552


reddyn12:
Should I skip PYTHON with .skipIf or add it to the test.yaml file?

geohot:
Is PYTHON broken too?

reddyn12:
Yup, getting same answer as CLANG also

Issue #3541, opened 2024-02-29
UOps should be a graph such that all toposorts work


geohot:
We have to fix BARRIER. We also have to fix DEFINE_GLOBAL to be removable and the order to not depend on arbitrary DEFINE_GLOBAL position.

Once all toposorts work, this is a space we can search over.

- [x] https://github.com/tinygrad/tinygrad/pull/3542

geohot:
Okay, only STORE is left. I think it should work, now to write the fuzzer

PR #3561, opened 2024-03-01
explicitly create_lt_node when used in shapetracker _expr_view

chenyuxyz:
Use `create_lt_node` and `create_ge_node` in `_expr_view` for the valid template. Leave `Node.__lt__` and other cpm functions for symbolic shape comparison, so we can handle it better and remove the `not (a < b)` hack later.

I thought it's no-op. Investigating why this fixed one of the linearizer failure.


chenyuxyz:
something about the uops loop optimization breaks loop scope

geohot:
We need to fix up the uops optimizer.

#3541 I'm writing a fuzzer right now to confirm all toposorts of the uops are still valid programs. There's an annoyance in C where if you declare a variable in loop scope you can't access it outside loop scope, where uops don't really have the concept of loop scope.

chenyuxyz:
Fixed the linearizer failure thing, there was a missed replacement in substitute and generated a different kernel. Will think about how to handle shape cmp before merging this

PR #3568, opened 2024-03-01
label_smoothing in sparse_cat_crossentropy

chaosagent:


chenyuxyz (commented):
Need tests. Also check the range of the input

PR #3570, opened 2024-03-02
fix sparse_categorical_crossentropy ignore_index with label_smoothing

chenyuxyz:


chaosagent:
`return ((1 - label_smoothing) * (log_probs * y).sum() + -1 * label_smoothing * (log_probs.mean(-1) * loss_mask).sum()) / loss_mask.sum()`

PR #3573, opened 2024-03-02
MLPerf Resnet (cleaned up)

chaosagent:
#3361

How can I make it prettier?


geohot:
Can you break this into multiple PRs? For example, I see a new LRScheduler, that should be a PR with tests

geohot:
Re: the Linear init override. #3576

geohot:
I'm also ready to give this a try, just got myself a tiny12. Any instructions or things I should know?

chaosagent:
Just write BEAM=2 GPUS=6 and it will take care of everything else (hopefully). I did the successful run in an old fork (from around start of Feb), stability has since regressed somehow. HSA OOMs, and HIP freezes after a few minutes

geohot:
Since this is the first training bounty, if you can get this merged by next Friday (3/8) I'll double it to $1200.

There's a lot that needs to be broken out of this and tested. LARS/LRScheduler etc...

For the bounty, it must be upstream and I repro the model in <24 hours (cleanly, no restarts)

geohot:
Running with `HSA=1 MODEL=resnet BEAM=2 BS=624 GPUS=6 python3 examples/mlperf/model_train.py` on tiny12

geohot:
NOTE: this branch breaks `HSA=1 HALF=1 STEPS=350 BS=1536 GPUS=6 python3 examples/hlb_cifar10.py` wth
```
  File "/home/tiny/tinygrad/examples/hlb_cifar10.py", line 439, in <module>                                                                                                                   
    train_cifar()                                                                                                                                                                             
  File "/home/tiny/tinygrad/examples/hlb_cifar10.py", line 416, in train_cifar                                                                                                                
    loss = train_step_jitted(model, [opt_bias, opt_non_bias], [lr_sched_bias, lr_sched_non_bias], X, Y)                             
  File "/home/tiny/tinygrad/tinygrad/features/jit.py", line 119, in __call__                                                                                                                  
    self.ret = self.fxn(*args, **kwargs)                                                                                                                                                      
  File "/home/tiny/tinygrad/examples/hlb_cifar10.py", line 344, in train_step                                                                                                                 
    loss.backward()                                                                                                                                                                           
  File "/home/tiny/tinygrad/tinygrad/tensor.py", line 310, in backward
    grads = t0._ctx.backward(t0.grad.lazydata)                                                 
  File "/home/tiny/tinygrad/tinygrad/mlops.py", line 118, in backward
    return self.y.e(BinaryOps.MUL, grad_output) if self.needs_input_grad[0] else None, \
  File "/home/tiny/tinygrad/tinygrad/features/multi.py", line 81, in e                                                                                                                        
    return MultiLazyBuffer([lsrcs[0].e(op, *lsrcs[1:], arg=arg) if r else lsrcs[-1].const(0) for lsrcs,r in zip(zip(*srcs),new_real)], axis, new_real)
  File "/home/tiny/tinygrad/tinygrad/features/multi.py", line 81, in <listcomp>                                                                                                                   return MultiLazyBuffer([lsrcs[0].e(op, *lsrcs[1:], arg=arg) if r else lsrcs[-1].const(0) for lsrcs,r in zip(zip(*srcs),new_real)], axis, new_real)
  File "/home/tiny/tinygrad/tinygrad/lazy.py", line 111, in e                
    assert all_same(dts:=[x.dtype.scalar() for x in (srcs[1:] if op is TernaryOps.WHERE else srcs)]), f"all dtypes must match {dts} on {op}"                                                  AssertionError: all dtypes must match [dtypes.float, dtypes.half] on BinaryOps.MUL
    ```

**Comment by geohot** [comment]
Screenshot 2024-03-03 at 6 01 33 AM

Hmm, lock up at 673. Going to try HSA

**Comment by geohot** [comment]
@nimlgen When I run with HSA
Screenshot 2024-03-03 at 6 07 09 AM

**Comment by geohot** [comment]
I did two more runs, one with and one without WANDB=1. The non WANDB one didn't hang...testing again.

**Comment by nimlgen** [comment]
* There is a bug when HSA BEAM uses HIP, since HSA compiler is just hip compiler. So it was OOM, looks like HIP took some
* I got `16  353.28 ms run,   40.26 ms python,  27.61 ms fetch data,  285.41 ms HSA * 6,  6.94 loss, 0.00 acc, 0.004488 LR, 138.96 GB used,  43805.07 GFLOPS` with **some rescheduling hacks** to hide copies time. HSA master is now slower than HIP.

**Comment by geohot** [comment]
HSA froze at eval here (ran with BS=384). Why does it wait for the 2nd epoch for eval, you don't know if there's a bug until late.

Screenshot 2024-03-04 at 6 10 19 AM

**Comment by chenyuxyz** [comment]
Now lars and polynomial scheduler are merged, I think it's a good idea to have this pr for full training (< 24 hours). Other improvements likely half can be made in subsequent prs.

I am ready to test this and help getting it merged

File: .github/workflows/benchmark.yml [hunk 1]
```diff
@@ -166,6 +166,10 @@ jobs:
     #   run: time HALF=1 STEPS=10 BS=1536 GPUS=6 python3 examples/hlb_cifar10.py
     - name: Run full CIFAR training
       run: time HSA=1 HALF=1 LATEWINO=1 STEPS=1000 TARGET_EVAL_ACC_PCT=93 python3 examples/hlb_cifar10.py | tee train_cifar_one_gpu.txt
+    - name: Run 10 MLPerf ResNet50 training steps (1 gpu)
+      run: HSA=1 BENCHMARK=10 BS=104 GPUS=1 MODEL=resnet python3 examples/mlperf/model_train.py | tee train_resnet_one_gpu.txt
```
chenyuxyz, line 170: the `BENCHMARK` specified steps per epoch, but it still runs multiple epochs

chenyuxyz (commented):
what's the full command you suggest for the full run?

chaosagent:
> what's the full command you suggest for the full run?

WANDB=1 HSA=1 BEAM=2 CKPT=1 BS=624 GPUS=6 MODEL=resnet python examples/mlperf/model_train.py

the limitation was that the training will crash periodically... My run from yesterday crashed at epoch 12 (MES error i think)

chenyuxyz:
getting 1 sec step time with BEAM=2 HSA=1 (and several hangs during BEAM). trying HIP now

chenyuxyz:
fyi 400ms step time with BEAM=4 (this plus 3626).

chaosagent:
> crashed with checkpoint

oops... need to re-test checkpoint code after cleanup

.github/workflows/benchmark.yml
chenyuxyz: returning early seems to not clean up dataloader properly (same if the process crashed or keyboard interrupt early)

chaosagent: what doesn't get cleaned up when return?

chenyuxyz:
i was reading the doc https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc and it seems like we need to use the same number of epochs too

chaosagent:
which part says we need to use same number of epochs? I think the relevant section is `13. Reference Convergence Points (RCPs)`, which says we cannot converge faster than a supplied set of reference convergence points, which can be generated from any set of hyperparameters given to the reference implementation (need to notify them ahead of time if we have better hparams, though). The Nvidia submissions are using aggressive LRs (10.4) and aggressive epochs (37, target reached at 35).

examples/mlperf/model_train.py
```diff
@@ -1,9 +1,210 @@
...
+      x.realize().shard_(GPUS, axis=0)
+    else:
+      x.realize().to_(GPUS)
+  parameters = get_parameters(model)
+
+  # ** hyperparameters **
+  epochs            = config["epochs"]            = getenv("EPOCHS", 45)
+  BS                = config["BS"]                = getenv("BS", 104 * len(GPUS))  # fp32 GPUS<=6 7900xtx can fit BS=112
+  EVAL_BS           = config["EVAL_BS"]           = getenv("EVAL_BS", BS)
+  base_lr           = config["base_lr"]           = getenv("LR", 8.4 * (BS/2048))
```
chenyuxyz, line 47: should this be `getenv("LR", 8.4) * (BS/2048)`?

examples/mlperf/model_train.py
```diff
@@ -1,9 +1,210 @@
...
+  lr_warmup_epochs  = config["lr_warmup_epochs"]  = 5
```
chenyuxyz, line 48: `getenv("WARMUP_EPOCHS", 5)`

I was trying to match the nvidia hparams you mentioned. Have you tried it?

chaosagent:
> I was trying to match the nvidia hparams you mentioned. Have you tried it?

I tried it a while back, and I ran into training instability in 2-3 epochs...

examples/mlperf/model_train.py
chaosagent: Copying hparams from runs that have different bs would be pretty confusing then. maybe getenv("LR", 8.4/2048) * BS?

chenyuxyz: oh yea... I was not sure what equivalent LR i should set for the nvidia one

chenyuxyz:
For some reason the 6gpu one took more than 20 minutes. Not sure if it hung but I cancelled it manually. The second one failed because the shared `resnet_X` was already created.

Can you add a check to cleanup if `resnet_X` is already created. Also I think it's fine to remove the 6gpu one from CI for now.

chaosagent:
> For some reason the 6gpu one took more than 20 minutes.

hm, it shouldnt take that long without beam...

chenyuxyz:
seems like it hangs after the 10 steps (tested on tiny9 with no cache)
image

chenyuxyz:
for the record, this is the run i did that hit 75.9% https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/vudd5ogk/overview. with beam=4, plus 3626 it's 400ms step and ~10.5 hours to hit 75.9%; without 3626 it's 880ms so I think it qualifies as < 24hours run!

PR #3574, opened 2024-03-02
fix SCE ignore_index with label_smoothing

chaosagent:


chaosagent:
ugly

chenyuxyz:
I still want to make it prettier, but thanks for the fix!

PR #3576, opened 2024-03-02
Revert "allow overriding weight init for Linear"

geohot:
Reverts tinygrad/tinygrad#3569


geohot:
So instead just override __init__ in the superclass and set weight and bias. Since tinygrad is lazy, the others aren't actually created and this is cheap

PR #3580, opened 2024-03-02
lazy.py late folding

geohot:
Hmm, why is this breaking?

`python3 test/test_optim.py TestOptim.test_multistep_sgd_high_lr_teeny` seems to be something weird with assign

Related to #3395


chenyuxyz:
this is done with generic const folding in lazy

PR #3584, opened 2024-03-02
add constant folding for WHERE in uops

geohot:


tinygrad/codegen/uops.py
```diff
@@ -25,6 +25,29 @@ class UOp:
...
+  elif arg == UnaryOps.SIN: ret = math.sin(p[0])
+  elif arg == UnaryOps.NEG: ret = -p[0]
+  elif arg == BinaryOps.MUL: ret = p[0]*p[1]
+  elif arg == BinaryOps.ADD: ret = p[0]+p[1]
+  elif arg == BinaryOps.SUB: ret = p[0]-p[1]
+  elif arg == BinaryOps.XOR: ret = p[0]^p[1]
+  elif arg == BinaryOps.MAX: ret = max(p[0], p[1])
+  elif arg == BinaryOps.CMPEQ: ret = p[0] == p[1]
+  elif arg == BinaryOps.CMPLT: ret = p[0] < p[1]
+  elif arg == BinaryOps.DIV: ret = p[0]//p[1] if dtypes.is_int(dtype) else (p[0]/p[1] if p[1] != 0 else math.nan)
```
geohot, line 44: @carsonRadtke what if it's int and p[1] is 0?

carsonRadtke: Perhaps we should just let Python raise a ZeroDivisionError? (Or raise it ourselves so it is a bit cleaner)

The other option is to mimic the non-int branch and return math.nan. This feels a bit weird to me because
```
>>> type(math.nan)
<class 'float'>
```

PR #3564, opened 2024-03-01
Directly store float4 nodes

Qazalin:
#3551
This optimization works with both:

1. PHI nodes `TestOps.test_gemm`
```
__kernel void r_2_16_8_16_4_4_4n2(__global float* data0, __global float* data1) {
  // this branch
  *((__global float4*)(data0+idx)) = acc0;
  // master
  *((__global float4*)(data0+idx)) = (float4)(float4)((acc0).x,(acc0).y,(acc0).z,(acc0).w);
}
```
2. float4 values `TestOps.test_avgpool2d`
```
__kernel void E_2_37_7_3_2_16_4n3(__global float* data0, __global float* data1) {
  // this branch
  *((__global float4*)(data0+idx)) = val0;
  // master
  *((__global float4*)(data0+idx)) = (float4)(float4)((val0).x,(val0).y,(val0).z,(val0).w);
}
```
I limited this to global stores only, because they're already childless.
Next is supporting local store and accurately updating its children in the UOpGraph.

PR #3592, opened 2024-03-03
simpler float4 direct store and locals support

Qazalin:
follow-up on #3564 
Since the UOpGraph holds a single copy of every uop, if we update that UOp directly it auto-updates vins too.
This simplifies things a lot, and supports locals out of the box

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -37,7 +37,7 @@ python_alu = {
   UnaryOps.SQRT: lambda x: math.sqrt(x) if x >= 0 else math.nan, UnaryOps.SIN: math.sin, UnaryOps.NEG: operator.neg,
   BinaryOps.MUL: operator.mul, BinaryOps.ADD: operator.add, BinaryOps.SUB: operator.sub, BinaryOps.XOR: operator.xor,
   BinaryOps.MAX: max, BinaryOps.CMPEQ: operator.eq, BinaryOps.CMPLT: operator.lt, BinaryOps.MOD: operator.mod,
-  BinaryOps.DIV: lambda x,y: x//y if isinstance(x, int) else (x/y if y != 0 else math.nan),
+  BinaryOps.DIV: lambda x,y: int(x/y) if isinstance(x, int) else (x/y if y != 0 else math.nan),
   TernaryOps.WHERE: lambda x,y,z: y if x else z}
 
 def exec_alu(arg, dtype, p):
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -217,18 +217,16 @@ class UOpGraph:
             self.uops.remove(u)
             changed_something = True
 
-    # (recursively) remove childless uops
-    self.remove_childless()
-
     # store float4 upcasts directly if possible
-    replaced_stores: Dict[UOp,UOp] = {}
+    replaced_stores: Dict[UOp,Tuple[UOp,...]] = {}
     for u in self.uops:
       if u.uop is not UOps.STORE or (val:=u.vin[-1]).uop is not UOps.CAST or cast(DType,val.dtype).count == 1: continue
-      if all(el.uop is UOps.GEP for el in val.vin): replaced_stores[u] = val.vin[0].vin[0]
-      elif all(el.uop is UOps.PHI for el in val.vin): replaced_stores[u] = phi_resolve_acc(val)
+      if all(el.uop is UOps.GEP for el in val.vin) or all(el.uop is UOps.PHI for el in val.vin): replaced_stores[u] = val.vin
     for prev,new in replaced_stores.items():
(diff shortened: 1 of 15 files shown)


tinygrad/codegen/uops.py
```diff
@@ -224,12 +224,11 @@ def get_recursive_parents(x:UOp, with_phi=False) -> Set[UOp]:
     replaced_stores: Dict[UOp,UOp] = {}
     for u in self.uops:
       if u.uop is not UOps.STORE or (val:=u.vin[-1]).uop is not UOps.CAST or cast(DType,val.dtype).count == 1: continue
-      if u.vin[0].uop is UOps.DEFINE_LOCAL: continue # TODO add support for local store
       if all(el.uop is UOps.GEP for el in val.vin): replaced_stores[u] = val.vin[0].vin[0]
       elif all(el.uop is UOps.PHI for el in val.vin): replaced_stores[u] = phi_resolve_acc(val)
     for prev,new in replaced_stores.items():
-      self.add(UOps.STORE, prev.dtype, (prev.vin[0],prev.vin[1],new), insert_before=self.uops.index(prev))
-      self.uops.remove(prev)
+      self.uops.remove(prev.vin[-1]) # remove the old upcast
```
geohot, line 230: Is this a childless uop? You might be able to do this optimization before the remove childless pass, then it will be autoremoved.

Qazalin: that won't work because of the PHI support.
If there's a `cast(acc.x, acc.y, acc.z, acc.w)` it's casting 4 PHI nodes. It's their only parent so if I let the childless pass run it'll remove those PHIs too.

tinygrad/renderer/cstyle.py
```diff
@@ -118,9 +118,9 @@ def ssa(u, prefix="t"):
       kk("}")
     elif uop is UOps.STORE:
       assert vin[0].dtype is not None and vin[2].dtype is not None
-      if len(vin) > 3: kk(f"if ({r[vin[3]]}) {{")
```
chenyuxyz, line 121: Do you need to update the llvm one too?

Qazalin: no llvm doesnt have float4

geohot:
I don't understand this. What changed about the args to the STORE op?

We really need docs for the uops.

Qazalin:
change: The STORE's children are now [the float4 acc, scalar PHIs]

It should've always been there, because the childless pass was skipped it wasn't caught.
But if you skip the CAST, there should still be something linking the PHIs to the STORE.

We can't store the PHIs directly because the access pattern is indexed.

@geohot example: `TestOps.test_matvec`


example: `TestOps.test_matvec`
```
kernel void r_8_4_8_16_4(device float* data0, const device float* data1, const device float* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  // DEFINE_ACC dtypes.float4
  float4 acc0 = float4(0.0f,0.0f,0.0f,0.0f);

    // PHI dtypes.float vin=[GEP,...]
    (acc0).x = ((val0*val1)+(acc0).x);
    // PHI dtypes.float
    (acc0).y = ((val0*val2)+(acc0).y);
    // PHI dtypes.float
    (acc0).z = ((val0*val3)+(acc0).z);
    // PHI dtypes.float
    (acc0).w = ((val0*val4)+(acc0).w);
  }
  // --- before, CAST was PHI's child
  // CAST dtypes.float4 [PHI, PHI, PHI, PHI]
  // then, we stored the cast:
  //  STORE vin=[cast]
  *((threadgroup float4*)(temp+alu1+(lidx2*4))) = float4(acc0.x, acc0.y, acc0.z, acc0.w);

  // -- now, the CAST is skipped
  // The PHI should be directly linked to the STORE
  // STORE vin=[acc0, *PHI values]
  *((threadgroup float4*)(temp+alu1+(lidx2*4))) = acc0;
}
```
otherwise the PHIs are childless.

geohot:
Oh...this is a different issue. I don't think the PHIs should be in the store like that, but I see that they are childless after you make the change. It's almost a rendering issue...

Qazalin:
yeah it's rendering dependant tbh. I gave this a try in cstyle, can bring it back tho
https://github.com/tinygrad/tinygrad/pull/3592/commits/bf25e935f8ecef78bbd07ed9fa10edc5331ef0a5#diff-9694d15708a36a57a7e1485db60b1b4b1d2fc1e7278bc87127a73efd0bfa7dc0R123

Screenshot 2024-03-03 at 9 24 36 PM

then you'd just pass the CAST's vins to store as a `Tuple[UOp]`:
Screenshot 2024-03-03 at 9 26 30 PM

geohot:
I'm fine with leaving it in UOps, just after the child removal and it breaks the graph (what it is now). Adding UOps.END has a similar feel. It's fine to do the old cast removal manually

PR #3593, opened 2024-03-03
Fix two bugs concerning `Tensor.to`.

Marandil:
1. `Tensor.to` should return `self` if `device == self.device`. This was not the case if provided with non-canonical name of `self.device` (e.g. `HIP` and `HIP:0`).
2. `Tensor.to` result was missing graph (`_ctx`), even though `requires_grad` and `grad` were propagated .

Add corresponding tests.
~~As a bonus, missing return type for deepwalk~~ ←I think I get why it's missing now.

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -339,6 +339,19 @@ class TestMoveTensor(unittest.TestCase):
     assert s.dtype == t.dtype
     assert s.requires_grad == t.requires_grad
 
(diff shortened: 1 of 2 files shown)


geohot:
Merged. The canonicalize is definitely correct. The grad one...I'm more skeptical but it doesn't seem more wrong than what's on master.

PR #3594, opened 2024-03-03
Fix symbolic negative floordiv

patosai:
`(7//-2)` = -4, not -3

also a[0-7]//-2 = `((((a*-1)+8)//2)+-4)`, if you plug in various values of a, the new formula is correct.
`a=0`, result is 0. `a=1`, result is -1.


chenyuxyz:
will probably remove the support of // negative numbers, but fine for now

PR #3595, opened 2024-03-03
test BEAM params on gpt2 benchmark

chenyuxyz:


chenyuxyz:
weird, if `CACHELEVEL` is not specified, this runs locally on M1 Max

PR #3596, opened 2024-03-03
Python div alu behavior differs slightly from others

patosai:
In Python, `x//y` rounds towards negative infinity, but `(x/y)` integer ops in other languages rounds towards 0.

Example in Python:
```
>>> (-9//4)
-3
>>> int(-9/4)
-2
```

Same example in C:
```
#include <stdio.h>
int main()
{
  printf("%d\n", -9/4);
  return 0;
}

$ ./a.out
-2
```

### DIFF
--- a/test/test_uops.py
+++ b/test/test_uops.py
@@ -123,7 +123,7 @@ class TestExecALU(TestUOps):
     self.assertEqual(exec_alu(BinaryOps.DIV, dtypes.float32, (7.0, 3.0)), 2+(1.0/3.0))
     self.assertEqual(exec_alu(BinaryOps.DIV, dtypes.float32, (7.0, -3.0)), -2-(1.0/3.0))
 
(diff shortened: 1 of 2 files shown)


davidjanoskyrepo:
I wouldn't conflate an operator with an operation.

Using the floor div isn't really a choice in correctness. The rounding behavior of floor div vs int cast div is just a design choice with ideas about performance. Truncate cast is faster than rounding cast and in real hardware is simple to do.

Since this is an alu though, if only integer operations are available maybe an intermediate float result doesn't make sense and the overhead of a class cast is not really ideal. If you test the performance of one versus the other, the floor div wins by a lot.

https://colab.research.google.com/drive/1nisIhOzfMlVom2sekMIEiW_fR74pnm6G?usp=sharing

geohot:
That reply is total nonsense. Good fix

PR #3598, opened 2024-03-03
search: hotfix to make sure TC behavior is all in applied_opts

flammit:


tinygrad/features/search.py
```diff
@@ -16,8 +16,8 @@
 actions += [Opt(op=OptOps.GROUPTOP, axis=axis, amt=amt) for amt in [13,16,29,32,256] for axis in range(3)]
 actions += [Opt(op=OptOps.GROUP, axis=axis, amt=amt) for amt in [0,4,8,16] for axis in range(3)]
 actions += [Opt(op=OptOps.PADTO, axis=axis, amt=amt) for amt in [32] for axis in range(7)]
-actions += [Opt(op=OptOps.LOCAL, axis=0, amt=32), Opt(op=OptOps.UPCASTMID, axis=1, amt=4)]
-actions += [Opt(op=OptOps.TC, axis=axis, amt=0) for axis in range(4)]
+actions += [Opt(op=OptOps.LOCAL, axis=0, amt=32), Opt(op=OptOps.UPCASTMID, axis=1, amt=4), Opt(op=OptOps.TC, axis=0, amt=0)]
+actions += [Opt(op=OptOps.TC, axis=axis, amt=1) for axis in range(4)]
```
chenyuxyz, line 20: Why is axis=0, amt=0 special? I would expect `axis in range(4)` in the amt=0 case too

flammit: `amt=1` will do the everything `amt=0` does, so the only reason I added the `axis=0, amt=0` one is so that the search space contains all potential opts coming from the hand-coded ones.

PR #3600, opened 2024-03-04
Fix softmax.argmax issue

### LINKED ISSUE #3552: .softmax().argmax() broken in CLANG

chenyuxyz:
argmax is finding max + finding argmax. The finding max part fused with softmax and introduced numerical error. The resulting max can be different from any of the tensor element, and argmax returned size - 1 as a result. One way to fix it is to include eps when comparing with float max.

Also there's no mismatch if there's a realize between softmax and argmax.

Fixed #3552 but I don't really like this. Open as a draft for now.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -752,7 +752,6 @@ class TestOps(unittest.TestCase):
     helper_test_op([(10,10,10)], lambda x: x.softmax(0), atol=1e-7, grad_atol=1e-7)
     helper_test_op([(10,10,10)], lambda x: x.softmax(1), atol=1e-7, grad_atol=1e-7)
     helper_test_op([(10,10,10)], lambda x: x.softmax(2), atol=1e-7, grad_atol=1e-7)
-  @unittest.skipIf(CI and Device.DEFAULT in ["CLANG", "PYTHON"], "Broken ISSUE #3552")
   def test_softmax_argmax(self):
     helper_test_op([(45,65)], lambda x: x.softmax(0).argmax(), forward_only=True, atol=1e-7, grad_atol=1e-7)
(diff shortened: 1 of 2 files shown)


chenyuxyz:
to be more specific, this is because it recomputes the underlying data being compared

```
float alu0 = max((exp2(((val0-val1)*1.4426950408889634f))/val2),acc0);
```
in fused find max kernel, and
```
*(data0+gidx0) = ((int)((val0==(exp2(((val1-val2)*1.4426950408889634f))/val3)))*(acc0+16));
```
in find argmax kernel

and somehow it could not find a match

chenyuxyz:
minimal repro
```
a = Tensor([4, 3, 2, 1]).exp().argmax()
print(a.numpy())
```
returns 3 for CLANG and PYTHON

chenyuxyz:
simpler repro
```
a = Tensor([4.0, 3.0]).exp().max() == Tensor([4.0, 3.0]).exp()
print(a.numpy())
```
returns [False, False] with CLANG or PYTHON

chenyuxyz:
the root cause is that the output of exp is not float32, and the comparison to an element in float32 buffer failed to match. For example, if in cstyle we changed eq to rendered as `f"((float){a}==(float){b})"` it can pass clang.

chenyuxyz:
#3927 fixed for CLANG

for PYTHON i don't see a clean fix as every number is a python number and there's no cast. Maybe with multi output / multi reduce kernel we don't recompute the expression and has no problem

PR #3602, opened 2024-03-04
hsa in lin opts

nimlgen:
I don't know if it's a the best way to fix beam calling to HIP device when using hsa backend...


chenyuxyz:
looks fine to me, what's the concern?

PR #3603, opened 2024-03-04
tinybox benchmark with HSA

chenyuxyz:


geohot:
This blocked on anything?

chenyuxyz:
#3608 resource error if torch also uses the device. I will test this again and merge it first with the caveat that torch only uses cpu in benchmark

geohot:
Hmm, I don't really like losing that functionality, think we should block until we fix. Curious why it breaks, I wouldn't expect it to.

chenyuxyz:
yea i don't like that either so did not merge. It looks like if tinygrad inits HSA, torch cuda_init would fail, and if torch already inited, tinygrad HSA init would fail. @nimlgen could it be related to how the device was inited?

nimlgen:
Uh, that's seems to be the way that we have 2 libhsa-runtime64.so libraries, one from PyTorch, one ours. Like kernel does not allow to init runtime twice for the same task_struct, but since state of this libhsa-runtime64.so is different they try to do this.

Like the same way this:
```
from tinygrad import Tensor
a = Tensor([1])
print(a.numpy())

import torch
print(f"{torch.__version__=}")
a = Tensor([1])
b = torch.tensor([1]).to("cuda")
print(a.numpy(), b)
```
does not work for HIP, according to gdb we got 2 different libhsa-runtime64.so (I believe we call one, torch calls other).

nimlgen:
Oh can add `LD_PRELOAD="/opt/rocm/lib/libhsa-runtime64.so"` to ci, bad?

chenyuxyz:
eh that makes sense, and thanks for the suggestions. Benchmark looks good now and time is similar to latest HIP one (I reran both).

PR #3607, opened 2024-03-04
multioutput kernels

Qazalin:
This diff #4030 deletes the "can only have one output buffer" constraint -
```py
          # can only have one output buffer
          # can only reduce contiguous
          # max one reduceop per kernel
          if len(realized_children) > 1 or not st.contiguous or st.size != r.st.size or (tr in reduce_for_op and reduce_for_op[tr] != r):
```

In this PR I'm making progress on fusing kernels without reduces, the same code in 4030 can generalize to fuse are patterns like:
Screenshot 2024-04-15 at 9 34 42 PM

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -307,7 +307,7 @@ class Kernel:
           self.reshape_and_permute(None, order)
           if DEBUG >= 3: print("permuted global dim", order, "due to allocation exceeds global limit")
 
-  def alias_buffer(self, op:LazyOp, i:int, pattern:List[int]) -> None:
+  def alias_buffer(self, op:LazyOp, i:int, pattern:List[int]) -> LocalBuffer:
     assert len(pattern) == len(self.sts[i].shape), f"must include a pattern for each shape {pattern} {self.sts[i].shape}"
 
     bst = 1
@@ -320,9 +320,9 @@ class Kernel:
           bst *= shp[j]
 
     self.sts.append(ShapeTracker((View.create(tuple(shp), tuple(stride)),)))
-    self.bufs.append(LocalBuffer(name=f"ldata{i}", size=self.sts[-1].size))
     if DEBUG >= 4: print("aliasing buffer", self.sts[i])
-    self.local_alias[op][i] = cast(LocalBuffer, self.bufs[-1])
+    self.local_alias[op][i] = cast(LocalBuffer, buf:=LocalBuffer(name=f"ldata{i}", size=self.sts[-1].size))
+    return buf
 
   # ******************** high level optimizers ********************
 
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -317,10 +317,10 @@ class Linearizer(Kernel):
     # no new opts and we already ran? skip relinearizing
     if self.applied_opts == self.applied_opts_cache: return self
 
-    # *** Kernel
     # save backups
     sts_backup, gfr_backup, upc_backup = self.sts[:], self.group_for_reduces, self.upcasted
 
+    # *** Kernel
     # global uop cache
     self.saved_exprs: Dict[Tuple, UOp] = dict()
 
@@ -335,9 +335,9 @@ class Linearizer(Kernel):
     # add global buffers
     for i,buf in enumerate(self.bufs):
       if isinstance(buf, MemBuffer):
-        self.buf_uops[i] = self.uops.add(UOps.DEFINE_GLOBAL,
-                                         buf.dtype if isinstance(buf.dtype, ImageDType) else PtrDType(buf.dtype), (),
-                                         (buf.idx, any(buf.idx == x.idx for x in self.outbufs)))
+        self.buf_uops[i] = self.uops.add(UOps.DEFINE_GLOBAL, buf.dtype if isinstance(buf.dtype, ImageDType) else PtrDType(buf.dtype),
+                                         (), (buf.idx, any(buf.idx == x.idx for x in self.outbufs)))
+
     # add var vals
     for i,var in enumerate(self.vars):
       assert var.expr is not None
@@ -379,7 +379,7 @@ class Linearizer(Kernel):
     reduce_idxs = [Variable(f"ridx{i}", 0, self.full_shape[i]-1) for i in range(self.first_reduce+self.group_for_reduces, self.shape_len-self.upcasted)]  # noqa: E501
     fake_reduce_idxs = [x*0 for x in reduce_idxs]
 
-    # late alias the tensor core buffers
+    # add LocalBuffers for multistage reduce and tensor core
     if (tc:=self.tensor_core) and self.tensor_core_opts is not None:
       alias_pattern = [0]*(self.global_dims) + [2]*(len(tc.threads)) + [0]*(self.local_dims-len(tc.threads)) + [0]*(self.shape_len-self.upcasted-self.first_reduce) + [1,1] + [3]*(self.upcasted-2)  # noqa: E501
       for op, tc_bufs in self.bufs_for_tensor_core.items():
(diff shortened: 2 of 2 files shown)


geohot:
I'll lock the bounty for you. Good luck! These $1,000 bounties demonstrate real mastery of tinygrad.

What's your approach to `LazyOp`, a list?

Qazalin:
thank you:) yeah that's how I've implemented it

Qazalin:
status:
- create_schedule can group realizes
- linearizer can uop multioutput

`test.test_multioutput.TestMultiOutput.test_basic_fusion`
```
kernel void E_4n2(device int* data0, device int* data1, const device int* data2, const device int* data3, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 4 */
  int val0 = *(data2+gidx0);
  int val1 = *(data3+gidx0);
  *(data0+gidx0) = (val0+val1);
  *(data1+gidx0) = (val0*val1);
}
```

now I need to figure out [a generic way to fuse these buffers](https://github.com/tinygrad/tinygrad/pull/3607/files#diff-4fe1256664cc5870d2d44de21d5f463a512f81729a13f5309e272c8f7793a039R148).

geohot:
Oh I think the fusion decision isn't hard. From realize.py

```python
          # can only have one output buffer
          # can only reduce contiguous
          # max one reduceop per kernel
          if len(realized_children) > 1 or not st.contiguous or st.size != r.st.size or (tr in reduce_for_op and reduce_for_op[tr] != r):
```
       
You should just be able to relax the "can only have one output buffer" restriction. This won't be 100% of everything possible, but as long as it gets Adam from 4 -> 1 I'm happy.

Qazalin:
Status

First fusion from `test.models.test_mnist.TestMNIST.test_adam_onestep`
```
kernel void E_69(device int* data0, device int* data1, const device char* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 69 */
  char val0 = *(data2+gidx0);
  *(data0+gidx0) = (int)(val0);
  *(data1+gidx0) = (int)(((((char)(-1))==val0)==false));
}
```

this could easily fuse too:
```
kernel void E_69n1(device float* data0, const device char* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 69 */
  *(data0+gidx0) = 0.0f;
}
```
i need to reshape it so that kernel.py's simplify_ones still works

Qazalin:
status:

all the "free" stuff are being fused correctly! here's a fused kernel from a resnet block: `NOOPT=1 test.test_schedule.TestSchedule.test_resnet_block`

```
kernel void E_36864(device float* data0, device float* data1, const device float* data2, const device float* data3, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.x; /* 36864 */
  float val0 = *(data2+gidx0);
  float val1 = *(data3+gidx0);
  *(data0+gidx0) = ((0.08333333333333333f*val0)+(-0.041666666666666664f));
  *(data1+gidx0) = ((0.08333333333333333f*val1)+(-0.041666666666666664f));
}
```
I'm iterating on the graph sorting/grouping to do full turbo adam and more.

Qazalin:
Status:
So far, reduced the kernel count of one e2e MNIST training step from 33 to 26. (more can be done with shapetracker simplification)
Now the JIT exec needs a refactor to be accurate.

Qazalin:
Status:
So far, reduced the kernel count of one e2e MNIST training step from 33 to 26. (more can be done with shapetracker simplification)
Now the JIT exec needs a refactor to be accurate.

chenyuxyz:
In addition to testing the kernel count, we also need to test the correctness of the kernel output. It's not clear if we are hitting it enough in the existing tests.

Qazalin:
This PR runs multioutput 104 times on the existing tests with CI=1 HSA=1

Qazalin:
[the new grouper](https://github.com/tinygrad/tinygrad/pull/3607/commits/026741aa53e7ba0534dc815b4b84d97494bb761b#diff-4fe1256664cc5870d2d44de21d5f463a512f81729a13f5309e272c8f7793a039R264-R270) reduces kernels across the board for openpilot, llama, resnet and cifar! @chenyuxyz can we please rerun benchmarks to confirm correctness?

tinygrad/engine/schedule.py
```diff
@@ -192,16 +203,22 @@ def create_schedule(outs:List[LazyBuffer], seen:Optional[Set[LazyBuffer]]=None)
...
-  schedule: List[ScheduleItem] = []
+  queue: Deque[Tuple[int,LazyBuffer]] = deque((0,out) for out in prescheduled if in_degree[out] == 0)
+  groups: DefaultDict[Tuple,List[ScheduleItem]] = defaultdict(list)
   while queue:
-    buf = queue.popleft()
+    level, buf = queue.popleft()
     seen.add(buf)
-    schedule.append(prescheduled[buf])
+    # single output
+    if buf.op in LoadOps or buf.device.startswith("DISK") or buf.device == "METAL" or \
```
geohot, line 212: Can you add an env var to turn multioutput off?

geohot:
Suggest adding this name change to mark multioutput kernels


diane@blak:~/tinygrad$ git diff
diff --git a/tinygrad/codegen/linearizer.py b/tinygrad/codegen/linearizer.py
index 511479d8..231382df 100644
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -213,7 +213,8 @@ class Linearizer(Kernel):
       self.buf_uops.append(self.uops.add(UOps.DEFINE_LOCAL, PtrDType(temp_dtype), (), ("temp", self.sts[-1].size)))
 
     # kernel name (before late upcast)
-    self.name = ("r_" if self.reduceop else "E_") + colored('_', 'BLACK').join([colored(str(x), c) for x,c in zip(self.full_shape, self.colors())])
+    self.name = ("r" if self.reduceop else "E") + (f"{len(self.outbufs)}_" if len(self.outbufs) > 1 else "_") + \
+      colored('_', 'BLACK').join([colored(str(x), c) for x,c in zip(self.full_shape, self.colors())])
 
     # name the function something unique
     Linearizer.kernel_cnt[(function_name := to_function_name(self.name))] += 1

Qazalin:
With a simple multioutput assign https://github.com/tinygrad/tinygrad/commit/c6a43d684b4cb4084d4adc1d8643b704d09a6dea, beautiful_mnist's JIT capture count went from 114->91. I think multioutput assign needs a dedicated PR.

tinygrad/engine/schedule.py
```diff
@@ -193,17 +205,23 @@ def create_schedule(outs:List[LazyBuffer], seen:Optional[Set[LazyBuffer]]=None)
         in_degree[assign_targets[x]] += 1
       if x in prescheduled: in_degree[out] += 1
 
-  queue = deque(out for out in prescheduled if in_degree[out] == 0)
-  schedule: List[ScheduleItem] = []
+  queue: Deque[Tuple[int,LazyBuffer]] = deque((0,out) for out in prescheduled if in_degree[out] == 0)
+  groups: DefaultDict[Tuple,List[ScheduleItem]] = defaultdict(list)
   while queue:
-    buf = queue.popleft()
+    level, buf = queue.popleft()
```
geohot, line 211: I don't understand what level does here

Qazalin: I'm borrowing the idea from level order traversal of binary trees. The way we can think about "level" in ScheduleItems toposort is that the items in the same level can run in parallel.

geohot: Hmm, I think a lot more fusion might be possible.

I don't love the `_fuse_group` function. Have you tried doing this instead by relaxing scheduler constraints on line 157? Ideally this should be line neutral.

Qazalin: I think 157 only checks the case for reduces.
Adam's weight update realizes happen in `_recurse_lb`, we need to graph the realizes.

Qazalin:
This PR has multioutput support for all `E_` kernels.

For multioutput `reduce_for_op`s, we should also think about where `realized_children` are in the graph. eg.`test.models.test_train.TestTrain.test_convnext` shouldn't fuse.

tinygrad/engine/schedule.py
```diff
@@ -199,31 +212,39 @@ def create_schedule(outs:List[LazyBuffer], seen:Optional[Set[LazyBuffer]]=None)
...
-  for out, si in prescheduled.items():
-    for x in si.inputs:
+  for out, ps in prescheduled.items():
+    for x in ps.inputs:
       graph[x].append(out)
       if x in assign_targets:
         graph[out].append(assign_targets[x])
         in_degree[assign_targets[x]] += 1
       if x in prescheduled: in_degree[out] += 1
-    del out.srcs  # can only schedule once
```
geohot, line 209: Why is this line removed?

geohot:
What's the blockers on getting this merged? Responded on the fuzzer, that should be in tests, not tinygrad proper

Qazalin:
@geohot

This diff does a limited multi output fusion - gets to -19% https://github.com/tinygrad/tinygrad/pull/3607#issue-2167646235.
#4030 is more generic - gets to -25%. I think since it naturally has a bigger scope it needs the fuzzer.

Qazalin:
this is done now - closing this diff

PR #3612, opened 2024-03-05
llama + beam to mac benchmark, full cifar to nvidia benchmark

chenyuxyz:
would merge if it's also ~1 minute. btw why is gpt2 beam not slower in the first beam run?


chenyuxyz:
22.5 tok/s for 7B llama on M2 Max

70 seconds cifar training on 3090 (52ms step time)

PR #3613, opened 2024-03-05
initialize Tensor grad same type as self

chaosagent:
when a tensor has different dtype then default_float, backward will initialize gradient to the wrong type (default instead of self)


geohot:
I'm fine with this, but tests failing

chaosagent:
> Wait, why do you have to same default_type?

normally the least_upper_dtype stuff is done in tensor.py, but this gradient stuff is MLOps and below

chenyuxyz:
oh i was hit by this when trying bf16 training. What's the status of this pr?

chenyuxyz:
merging this as i need it for bf16 training too. cifar training with half seems fine.

chaosagent:
Oh, what happened to the test failures?

PR #3614, opened 2024-03-05
use f32 by default to calculate in randn

chaosagent:
There are many places in the codebase where fp16 is not adequate for some particular calculation. sum() is handled well; it selects least upper dtype with float32. We might have to go around the codebase and do least upper dtype with float32 in all places that need it.

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -158,12 +158,10 @@ class TestTinygrad(unittest.TestCase):
 
   def test_randn_fp16(self):
     # low precision can result in inf from randn
-    old_default_float = dtypes.default_float
-    try:
-      dtypes.default_float = dtypes.float16
-      assert Tensor.randn((2, 128, 128)).abs().max().numpy().item() != float('inf')
-    finally:
-      dtypes.default_float = old_default_float
+    old_default_float, dtypes.default_float = dtypes.default_float, dtypes.float16
+    # this is way too small to fail consistently, increase to (4, 128, 128) if needed
+    assert Tensor.randn(2048).abs().max().numpy().item() != float('inf')
+    dtypes.default_float = old_default_float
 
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -246,7 +246,7 @@ def ones_like(self, **kwargs): return self.full_like(1, **kwargs)
   @staticmethod
   def randn(*shape, dtype:Optional[DType]=None, **kwargs) -> Tensor:
     # https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform
-    src = Tensor.rand((2, *argfix(*shape)), **kwargs)
+    src = Tensor.rand((2, *argfix(*shape)), dtype=least_upper_dtype(dtype or dtypes.default_float, dtypes.float32), **kwargs)
```
chenyuxyz, line 249: fyi there's least_upper_float

test/test_tensor.py
```diff
@@ -156,6 +156,15 @@ def test_randn_isnt_inf_on_zero(self):
     except: raise
     finally: Tensor.rand = original_rand
 
+  def test_randn_fp16(self):
+    # low precision can result in inf from randn
+    old_default_float = dtypes.default_float
+    try:
+      dtypes.default_float = dtypes.float16
+      assert Tensor.randn((2, 128, 128)).abs().max().numpy().item() != float('inf')
+    finally:
```
chenyuxyz, line 165: if you don't need to catch error don't need try and finally

chenyuxyz:
fyi this was fixed in a different way that guarantees the Tensor.rand is always in [0, 1) (not complete yet, need to update again with threefry and bf16)

fine to add the randn should not be inf test though

chenyuxyz:
i added the test in #3994 float without threefry is fine but we want to generic solution

Issue #3604, opened 2024-03-04
int overflow in indexing


geohot:
Have to catch this and assert in Linearizer when rendering the idx. Example from @chaosagent 

```python
from tinygrad import Tensor, dtypes, Device
from tinygrad.ops import LazyOp, BufferOps, BinaryOps, ReduceOps, MemBuffer, ConstBuffer, UnaryOps
from tinygrad.shape.shapetracker import ShapeTracker, View
from tinygrad.codegen.linearizer import Linearizer
from tinygrad.features.search import time_linearizer, bufs_from_lin
from tinygrad.codegen.kernel import Opt, OptOps

[...]

PR #3615, opened 2024-03-05
out of bounds access caused by launch bounds

geohot:
Working on #3604 though idk if it's simple int32 overflow.

I can confirm these are out of bounds on AMD though


geohot:
@flammit @Qazalin 

The test failure is interesting. It doesn't happen on real AMD, so either it's a bad opt or a remu issue

Qazalin:
@geohot yeah sorry about that remu failure - 
i'd never seen that noop these new kernels used... should've skipped that
can you please rerun remu? I landed this to the latest release

PR #3617, opened 2024-03-05
decouple buffer mutability from cstyle

Qazalin:
I think linearizer should explicitly define a buf is mutable in the UOp, this is a prereq for multioutput.


geohot:
Makes sense

PR #3619, opened 2024-03-05
Fix Tensor.dropout() with multigpu

Eliulm:
Right now `.dropout()` can not be used with multigpu. `.rand()`, which is called under the hood, uses `Tensor._loadop()` which cannot handle multidevice strings.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -202,7 +202,11 @@ class Tensor:
 
   @staticmethod
   def _loadop(op, shape, device:Optional[Union[Tuple[str], str]]=None, dtype:Optional[DType]=None, arg=None, **kwargs):
-    return Tensor(_loadop(op, shape, dtype or dtypes.default_float, device if device else Device.canonicalize(device), arg, kwargs.get("src", None)), dtype=dtype, device=device, **kwargs)
+    device, dtype, src = device if device else Device.canonicalize(device), dtype or dtypes.default_float, kwargs.get("src", None)
+    if isinstance(device, tuple):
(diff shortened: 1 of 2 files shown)


Eliulm:
> can you add a regression test (in test_multitensor)?

@chenyuxyz Sure thing! 

> Also is the recursive call necessary?

I can avoid it, but doing so will introduce more lines.

chenyuxyz:
do it in the correct, simple, compact way (following that order). line count is a metric, not the target

tinygrad/tensor.py
```diff
@@ -201,8 +201,11 @@ def shard_(self, devices:Tuple[str, ...], axis:Optional[int]=None):
   # ***** creation llop entrypoint *****
 
   @staticmethod
-  def _loadop(op, shape, device:Optional[str]=None, dtype:Optional[DType]=None, arg=None, **kwargs):
-    return Tensor(LazyBuffer.loadop(op, shape, dtype or dtypes.default_float, Device.canonicalize(device), arg), dtype=dtype, device=device, **kwargs)
+  def _loadop(op, shape, device:Optional[Union[Tuple[str], str]]=None, dtype:Optional[DType]=None, arg=None, **kwargs):
+    device, dtype, src = device if device else Device.canonicalize(device), dtype or dtypes.default_float, kwargs.get("src", None)
```
chenyuxyz, line 205: don't you need to canonicalize each element if device is a tuple too?

also why do you need to get the src out? these creation function should not have src, and the single LazyBuffer case should remain the same

Eliulm: Ye, you are right. Should be fixed now

PR #3620, opened 2024-03-05
this mem fault still happening

geohot:


Qazalin:
checking remu

Qazalin:
done - i'll make some time this week to use opts and add new instructions to remu.

PR #3621, opened 2024-03-05
add inference_mode context manager with decorator support

MaxWolf-01:
This is https://github.com/tinygrad/tinygrad/pull/3281 but with the right API.

Introduces `inference_mode` context manager akin to [torch.inference_mode](https://pytorch.org/docs/stable/generated/torch.inference_mode.html):
```python
with Tensor.inference_mode(): ...
# OR
@Tensor.inference_mode()
def f(): ...
```
Also adds ability to use `Tensor.train()` as a decorator:
```python
@Tensor.train():
def f(): ...
```

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -77,7 +77,7 @@ class Tensor:
   class inference_mode(ContextDecorator):
     def __init__(self, mode:bool = True): self.mode = mode
     def __enter__(self): self.prev, Tensor.no_grad = Tensor.no_grad, self.mode
-    def __exit__(self, exc_type, exc_value, traceback): Tensor.no_grad = self.mode
+    def __exit__(self, exc_type, exc_value, traceback): Tensor.no_grad = self.prev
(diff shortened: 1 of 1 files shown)


geohot:
Why is it `val` and not `mode` to match torch?

tinygrad/tensor.py
```diff
@@ -67,12 +68,16 @@ class Tensor:
   __slots__ = "lazydata", "requires_grad", "grad", "_ctx"
   __deletable__ = ('_ctx',)
   training: ClassVar[bool] = False
-  class train:
-    def __init__(self, val=True): self.val = val
-    def __enter__(self): self.prev, Tensor.training = Tensor.training, self.val
+  class train(ContextDecorator):
+    def __init__(self, mode:bool = True): self.mode = mode
```
geohot, line 72: Err, why did this one change? Does this match torch?

MaxWolf-01: Yes, torch also uses `mode` for `train`. See [source code](https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module)

geohot:
Cool, merged

PR #3625, opened 2024-03-06
PolynomialLR

chaosagent:
warmup will come later, maybe we can do something clever...


chaosagent:
Ok I think we can put all the mlperf pieces in examples and do some thorough equivalence testing

PR #3000, opened 2024-01-04
Ring allreduce in multitensor

uuuvn:
- [X] Basic ring_allreduce impl
- [ ] Fix last failing test
- [ ] Benchmark to make sure that it's actually faster
- [ ] Hopefully merge

PR #3627, opened 2024-03-06
Metal GPU profiling

uuuvn:
Made it to debug metal flakiness (which became 10x worse with #3000 so I have to fix it now), maybe will be helpful for future.

image

I also find it much nicer than looking at GRAPH=1 + DEBUG=4 to understand what's going on.

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
-import os, subprocess, pathlib, ctypes, tempfile, functools
-import Metal, libdispatch
+import os, subprocess, atexit, multiprocessing, pathlib, ctypes, tempfile, functools
+import Metal, Foundation, libdispatch
 from typing import List, Any, Tuple, Optional
 from tinygrad.codegen.kernel import LinearizerOptions
 from tinygrad.helpers import prod, getenv, DEBUG, unwrap2
 from tinygrad.device import Compiled, LRUAllocator, Compiler
 from tinygrad.renderer.cstyle import MetalRenderer
+
+MTL_CAPTURE_ENABLED = bool(getenv("MTL_CAPTURE_ENABLED", 0))
 
 def wait_check(cbuf: Any):
   cbuf.waitUntilCompleted()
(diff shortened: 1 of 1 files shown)


tinygrad/runtime/ops_metal.py
```diff
@@ -84,15 +92,40 @@ def copyin(self, dest:Any, src:memoryview): self.as_buffer(dest)[:] = src
...
+  @staticmethod
+  def stopCapture():
+    if MetalDevice.captureManager:
+      MetalDevice.captureManager.stopCapture()
+      print(f"Saved GPU trace to /tmp/tinygrad_{multiprocessing.current_process().name}.gputrace")
+
   def synchronize(self):
-    for cbuf in self.mtl_buffers_in_flight: wait_check(cbuf)
+    if len(self.mtl_buffers_in_flight) > 0:
+      self.mtl_buffers_in_flight[-1].waitUntilCompleted()
```
geohot, line 126: Err, is this okay to assume they are in order?

geohot:
Can you add some usage instructions for this in a comment? How do I view the trace?

uuuvn:
> Can you add some usage instructions for this in a comment? How do I view the trace?

Double click file in /tmp and it will open in Xcode

tinygrad/runtime/ops_metal.py
uuuvn: They're added to MTLCommandQueue and this list in the same order and should be executed in same order

I did it to declutter list of API calls, can revert this change, it isn't very important

uuuvn:
https://developer.apple.com/documentation/xcode/replaying-a-gpu-trace-file

tinygrad/runtime/ops_metal.py
```diff
@@ -84,15 +92,42 @@ def copyin(self, dest:Any, src:memoryview): self.as_buffer(dest)[:] = src
   def copyout(self, dest:memoryview, src:Any): dest[:] = self.as_buffer(src)
 
 class MetalDevice(Compiled):
+  captureManager = None
+
   def __init__(self, device:str):
     self.device = Metal.MTLCreateSystemDefaultDevice()
+    self.dname = device
+
+    if MTL_CAPTURE_ENABLED and not MetalDevice.captureManager:
```
geohot, line 101: Want to make this a startCapture method to match? Also, why is stopCapture static? It looks like it depends on the device

uuuvn: It depends on MTLDevice but there is only one of them (`Metal.MTLCreateSystemDefaultDevice()` always returns the same thing)

It is static because there is multiple `MetalDevice` instances created but only one `captureManager` is needed.

geohot: Hmm, I see, because the multiple METAL devices are really a hack. Okay, static is fine, but that should be a method

geohot:
What is `multiprocessing.current_process().name` supposed to do? I'd probably just remove it to match the other output files, it's returning `MainProcess` for me

uuuvn:
> What is `multiprocessing.current_process().name` supposed to do? I'd probably just remove it to match the other output files, it's returning `MainProcess` for me

It is supposed to make different files for different processes (eg pytest uses multiple workers).

geohot:
I get it, but none of our other saved files (like the graphs) look like this. I'd remove it.

tinygrad/runtime/ops_metal.py
```diff
@@ -60,13 +64,17 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
 class MetalAllocator(LRUAllocator):
   def __init__(self, device:MetalDevice):
     self.device:MetalDevice = device
+    self.i = 0
     super().__init__()
   def _alloc(self, size:int) -> Any:
     ret = self.device.device.newBufferWithLength_options_(size, Metal.MTLResourceStorageModeShared)
     if ret is None: raise MemoryError(f"Metal OOM while allocating {size=}")
+    ret.setLabel_(f"{self.device.dname} buf {(self.i)}")
```
geohot, line 72: Is this slow at all?

geohot:
Mostly I want to see this better abstracted. It's cluttering up the Metal runtime for a feature that will rarely be used.

tinygrad/runtime/ops_metal.py
uuuvn: 0.021209 milliseconds (21042 nanoseconds measured with time.perf_counter_ns())

uuuvn:
> I get it, but none of our other saved files (like the graphs) look like this. I'd remove it.

Removed

uuuvn:
> Mostly I want to see this better abstracted. It's cluttering up the Metal runtime for a feature that will rarely be used.

I don't see how it can be better abstracted other than moving init stuff into `startCapture()`

uuuvn:
> It's cluttering up the Metal runtime for a feature that will rarely be used.

I thought about that for a while and i also don't think that it's worth it, closing pr but i would keep this branch in my fork so it can be used in case of another weird metal bug

PR #3631, opened 2024-03-06
lars optimizer + tests

chaosagent:
tested against reference TF implementation

#3361

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -148,7 +148,7 @@ jobs:
       fail-fast: false
       matrix:
         task: [optimage, openpilot, onnx, mlperf]
-    name: ${{ matrix.task=='optimage'&&'GPU OPT and IMAGE Tests' || matrix.task=='openpilot'&&'openpilot (OpenCL) Tests' || matrix.task=='onnx'&&'ONNX+Optimization Tests' }}
+    name: ${{ matrix.task=='optimage'&&'GPU OPT and IMAGE Tests' || matrix.task=='openpilot'&&'openpilot (OpenCL) Tests' || matrix.task=='onnx'&&'ONNX+Optimization Tests' || matrix.task=='mlperf'&&'MLPerf Component Tests' }}
     runs-on: ubuntu-20.04
     timeout-minutes: 20
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -147,7 +147,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        task: [optimage, openpilot, onnx]
+        task: [optimage, openpilot, onnx, mlperf]
     name: ${{ matrix.task=='optimage'&&'GPU OPT and IMAGE Tests' || matrix.task=='openpilot'&&'openpilot (OpenCL) Tests' || matrix.task=='onnx'&&'ONNX+Optimization Tests' }}
     runs-on: ubuntu-20.04
     timeout-minutes: 20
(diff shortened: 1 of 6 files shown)


examples/mlperf/optimizers.py
```diff
@@ -0,0 +1,37 @@
...
+  def step(self):
+    for i, t in enumerate(self.params):
+      assert t.grad is not None
+      g = t.grad.realize()
+      w = t.detach()
+
+      if t not in self.skip_list:
+        g_norm = (g * g).sum().sqrt()
+        w_norm = (w * w).sum().sqrt()
+        trust_ratio = (w_norm > 0).where((g_norm > 0).where(
```
chenyuxyz, line 23: can use `((w_norm > 0) * (g_norm > 0)).where` for and

test/external/external_test_optim.py
chenyuxyz, line 1: is this being run in CI?

test/external/mlperf_resnet/lars_optimizer.py
chenyuxyz, line 1: fine to skip linter since this is imported from other place

test/external/external_test_optim.py
chaosagent: no... do we want tensorflow tests in CI?

chaosagent: added

.github/workflows/test.yml
```diff
@@ -225,6 +225,9 @@ jobs:
       - if: ${{ matrix.task == 'onnx' }}
         name: Test Beam Search
         run: PYTHONPATH="." GPU=1 IGNORE_BEAM_CACHE=1 python3 -m pytest extra/optimization/test_beam_search.py
+      - if: ${{ matrix.task == 'mlperf' }}
```
chenyuxyz, line 228: put this under onnx for now to save a worker

examples/mlperf/optimizers.py
```diff
@@ -0,0 +1,37 @@
+from typing import List, Set
+
+from tinygrad import Tensor
+from tinygrad.nn.optim import Optimizer
+
+# https://github.com/mlcommons/training/blob/master/image_classification/tensorflow2/lars_optimizer.py
+class LARS(Optimizer):
+  def __init__(self, params: List[Tensor], lr, momentum=0.9, weight_decay=1e-4, eta=0.001, eps=0.0, skip_list=None, nesterov=False, track_gnorm=False):
```
chenyuxyz, line 8: what's track_gnorm? did not see it in tf version

chaosagent: oh, it was a debug flag i forgot to remove, oops

examples/mlperf/optimizers.py
```diff
@@ -0,0 +1,37 @@
+from typing import List, Set
+
+from tinygrad import Tensor
+from tinygrad.nn.optim import Optimizer
+
+# https://github.com/mlcommons/training/blob/master/image_classification/tensorflow2/lars_optimizer.py
+class LARS(Optimizer):
+  def __init__(self, params: List[Tensor], lr, momentum=0.9, weight_decay=1e-4, eta=0.001, eps=0.0, skip_list=None, nesterov=False):
+    super().__init__(params, lr)
+    self.momentum, self.weight_decay, self.eta, self.eps, self.nesterov = momentum, weight_decay, eta, eps, nesterov
```
chenyuxyz, line 10: assert momentum and weight_decay need to be positive

examples/mlperf/optimizers.py
```diff
@@ -0,0 +1,37 @@
...
+    self.b = [Tensor.zeros(*t.shape, device=t.device, requires_grad=False) for t in self.params]
+    self.skip_list = set(skip_list or [])
+
+  def step(self):
+    for i, t in enumerate(self.params):
+      assert t.grad is not None
+      g = t.grad.contiguous()
```
chenyuxyz, line 17: why contiguous?

chenyuxyz (commented):
looks good. will merge after the assert on params

examples/mlperf/optimizers.py
chaosagent: In SGD, it realizes the grad because in earlier version of lazy, there was diamond problem. I think it should be fixed, but I don't want to make big changes here. So contiguous instead of realize because then it can go into one schedule.

PR #3632, opened 2024-03-06
ban __bool__ on Tensor

chenyuxyz:
avoid misuse


chenyuxyz:
@Qazalin fyi one dtype alu tests you added had a typo and false positive passed

chenyuxyz:
some cast combinations are undefined behavior and the result depends on platform an implementation

PR #3637, opened 2024-03-07
RDNA3 assembly support

geohot:


stalkerg:
How big is the difference from what `rusticle` is creating for RDNA?

PR #3638, opened 2024-03-07
add tests for bfloat16 on HIP

Zaffer:
just making this separate PR with the tests I wrote from the #3558 closed PR.


test/external/external_test_hsa_backend.py
```diff
@@ -0,0 +1,57 @@
+import unittest
+
+# import subprocess
+from tinygrad.tensor import Tensor, dtypes
+import numpy as np
+
+
+class TestHSABackend(unittest.TestCase):
```
geohot, line 8: Huh, why is this the name?

Also, this test isn't running in CI

Zaffer: was to regression test issue #3536 that only occurred with "HSA=1"
moved the tests to CI

I noticed these tests caused segfault on LLVM backend, so only testing on HIP now

geohot:
Cool!

PR #3639, opened 2024-03-07
general constant folding (2nd try)

carsonRadtke:
* continues #3480
* implement general constant folding


tinygrad/renderer/assembly.py
```diff
@@ -63,7 +63,7 @@ def uops_to_asm(lang:AssemblyLanguage, function_name:str, uops:UOpGraph) -> str:
         new = uops.add(UOps.ALU, dtypes.bool, (u,), arg=UnaryOps.NEG, insert_before=uops.uops.index(u)+1)
         replace[u] = new
       if u.arg == BinaryOps.CMPLT:
-        new = uops.add(UOps.ALU, dtypes.bool, (u.vin[0],), arg=UnaryOps.NEG, insert_before=uops.uops.index(u))
+        new = uops.add(UOps.ALU, dtypes.bool, (u.vin[0],), arg=UnaryOps.NEG, insert_before=uops.uops.index(u), simplify=False)
```
geohot, line 65: Is this a fix, or masking an issue?

tinygrad/dtype.py
```diff
@@ -43,11 +43,14 @@ def cast_scalar(scalar: Scalar, dtype:DType):
 
 class dtypes:
   @staticmethod
-  def is_float(x: DType) -> bool: return x.scalar() in (dtypes.float16, dtypes.bfloat16, dtypes.float32, dtypes.float64)
-  @staticmethod # static methds on top, or bool in the type info will refer to dtypes.bool
-  def is_int(x: DType) -> bool: return x.scalar() in (dtypes.int8, dtypes.int16, dtypes.int32, dtypes.int64) or dtypes.is_unsigned(x)
+  @functools.lru_cache(None)
+  def is_float(x: DType) -> bool: return x.scalar() in set((dtypes.float16, dtypes.bfloat16, dtypes.float32, dtypes.float64))
```
chenyuxyz, line 47: are you sure this `set()` is faster?

tinygrad/renderer/assembly.py
carsonRadtke: If I understand this rewrite, it seems PTX understands a boolean algebra where `-x * y` is semantically `!x && y` for booleans. Python does not do this, so exec_alu would instead do `-int(x) * int(y)` which is wrong.

I don't think this is masking an issue because having exec_alu mimic this for non-PTX backends would be wrong.

tinygrad/dtype.py
carsonRadtke: It made the ops test much faster, but then lru_cache made it even faster. The set vs tuple is now negligible it appears. I can revert it.

tinygrad/renderer/assembly.py
geohot: I don't understand. Which answer is correct? `simplify=False` is a hack

carsonRadtke: I don't think wrong or correct describes either of them. Python decided that `False < True` was a valid expression and should resolve to `True`. This PTX IR has decided that `<` is not defined for booleans, so the workaround is to replace `x < y` with `-x * y`.

Python says that `-False * True` is `0` which is not the same as `False < True`, but PTX says thats that `-False * True` is the same as `!False && True` which is the same as Python's `False < True`.

`simplify=False` is not a hack; it is a way of dealing the diverging semantics of Python and PTX.

geohot: Oh, PTX is wrong it sounds like.

False < True should evaluate to True

tinygrad/codegen/uops.py
```diff
@@ -40,11 +40,7 @@ def wfxn(*args):
   BinaryOps.DIV: lambda x,y: int(x/y) if isinstance(x, int) else (x/y if y != 0 else math.nan),
   TernaryOps.WHERE: lambda x,y,z: y if x else z}
 
-def exec_alu(arg, dtype, p):
-  return python_alu[arg](*p)
-  #if not dtypes.is_int(dtype): return ret
-  #adjusted = 0 if dtypes.is_unsigned(dtype) else 2 ** (dtype.itemsize * 8 - 1)
-  #return (ret + adjusted) % 2 ** (dtype.itemsize * 8) - adjusted
+def exec_alu(arg, dtype, p): return trunc_lut[dtype](dtype.itemsize, python_alu[arg](*p))
```
geohot, line 43: Also why is itemsize being passed in here? This is slow

tinygrad/dtype.py
```diff
@@ -43,10 +43,13 @@ def cast_scalar(scalar: Scalar, dtype:DType):
 
 class dtypes:
   @staticmethod
+  @functools.lru_cache(None)
```
geohot, line 46: Are these even called anymore?

geohot:
See #3675, I tracked down the PTX issue and wrote a fast truncation.

PR #3640, opened 2024-03-07
Lean proof of mergeability

Zaffer:
I am working on this every week, do you want me to commit more regularly?


geohot:
This is not close to enough for a lock

PR #3643, opened 2024-03-07
Track transfers on metal

uuuvn:
This is the source of metal flakiness

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 import os, subprocess, pathlib, ctypes, tempfile, functools
 import Metal, libdispatch
-from typing import List, Any, Tuple, Optional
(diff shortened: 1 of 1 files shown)


uuuvn:
Needed for #3000 as it is happens 10x more often with it

tinygrad/runtime/ops_metal.py
```diff
@@ -60,7 +60,13 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
 class MetalAllocator(LRUAllocator):
   def __init__(self, device:MetalDevice):
     self.device:MetalDevice = device
+    self.track_cross_device: Set[MetalDevice] = set()
```
geohot, line 63: Huh, how do devices end up in this set?

uuuvn: image
in device.py

PR #3644, opened 2024-03-07
debug time in tinybox ci

chenyuxyz:
checking why external_model_benchmark.py took 3 minutes on ci. only 35 seconds on tiny9


chenyuxyz:
also 35 seconds running on tiny8 directly. no idea what's happening

PR #3648, opened 2024-03-07
update fuzz_linearizer

chenyuxyz:
included non-reduce kernel and kernel with variables. green msg when everything passed. it's possible that creating rawbufs failed due to memory error, included that in failure cases

### DIFF
--- a/test/external/fuzz_linearizer.py
+++ b/test/external/fuzz_linearizer.py
@@ -7,7 +7,7 @@ from tinygrad.codegen.linearizer import Linearizer
 from tinygrad.features.search import get_linearizer_actions, bufs_from_lin
 from tinygrad.tensor import Tensor
 from tinygrad.features.graph import print_tree
-from tinygrad.helpers import getenv, from_mv, prod, Context
+from tinygrad.helpers import getenv, from_mv, prod, colored, Context
 from tinygrad.device import Device, Compiled
 from tinygrad.codegen.linearizer import UOp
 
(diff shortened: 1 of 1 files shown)


chenyuxyz:
on tiny9
```
testing ast 13
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(409600,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ CAST (dtypes.float, False)
  2    ┗━━ LOAD MemBuffer(idx=1, dtype=dtypes.uchar, st=ShapeTracker(views=(View(shape=(409600,), strides=(1,), offset=3535760512, mask=None, contiguous=False),)))
409600
Memory access fault by GPU node-1 (Agent handle: 0x559f9aee7a40) on address 0x7f5267e00000. Reason: Page not present or supervisor privilege.
[2]    51603 IOT instruction (core dumped)  FUZZ_N=100 PYTHONPATH=. python test/external/fuzz_linearizer.py
```

chenyuxyz:
i think that's a disk tensor stuff, and disk tensor load should not be included in the kernel dataset

chenyuxyz:
it looks like the one that involved Variables are having `COMPARE_ERROR`, likely the way we compare is wrong

on M1 Max, `time FUZZ_N=1000 PYTHONPATH=. python test/external/fuzz_linearizer.py`
have
```
tested=1000
COMPARE_ERROR: 10
COMPILE_ERROR: 44
```
some compile errors are due to temp too big (similar to test_failure_3), and others are internal errors

on tinybox it segfaulted at kernel 434.

@flammit fyi.

PR #3649, opened 2024-03-07
PolynomialDecayWithWarmup + tests

chaosagent:


examples/mlperf/lr_schedulers.py
```diff
@@ -0,0 +1,21 @@
...
+# https://github.com/mlcommons/training/blob/e237206991d10449d9675d95606459a3cb6c21ad/image_classification/tensorflow2/lars_util.py
+class PolynomialDecayWithWarmup(LR_Scheduler):
+  def __init__(self, optimizer: Optimizer, initial_lr, end_lr, train_steps, warmup, power=2):
+    super().__init__(optimizer)
+    assert train_steps > 0 and warmup > 0
+    self.warmup = min(warmup, train_steps)
+    self.initial_lr, self.end_lr, self.epochs, self.power = initial_lr, end_lr, train_steps, power
+
+    # set lr for first warmup step
+    self.optimizer.lr.assign(lr.cast(self.optimizer.lr.dtype) if isinstance(lr := self.get_lr(), Tensor) else lr).realize()
```
chenyuxyz, line 15: why would `self.get_lr` not be a Tensor? can you annotate output type for `get_lr`?

test/external/external_test_optim.py
```diff
@@ -84,5 +111,13 @@ def test_multistep_lars(self): self._test_lamb(10, {'lr': 0.001}, 1e-5, 0)
   def test_multistep_lars_high_lr(self): self._test_lamb(10, {'lr': 10}, 1e-5, 3e-4)
   def test_lars_skip_list(self): self._test_lars(1, {'lr': 0.01, 'skip_list': True}, 1e-5, 0)
 
+  def test_lars_polylr(self):
+    self._test_lars_polylr(10, {'lr': 1.0}, {
```
chenyuxyz, line 115: can you add one more with bigger steps? maybe the one that resnet will use

chenyuxyz (commented):
looks good! minor comments

examples/mlperf/lr_schedulers.py
chaosagent: unfortunately lr_schedule.py is really old. half of the schedulers don't work with jit, and some of them return floats. needs work.

test/external/external_test_optim.py
chaosagent: bigger on what dimension?

chenyuxyz: sorry i meant steps, just one with different steps and warmup

examples/mlperf/lr_schedulers.py
chaosagent: the reason why this line is so long is because resnet train_steps is larger than the fp16 largest int value, so we need epoch_counter in fp32 always, and thus we need to cast the output of get_lr. thought i would do a separate pr for that, though

chenyuxyz: i think break it into two lines with a TODO note helps, it's fine to fix other scheduler in a separate pr

test/external/external_test_optim.py
chaosagent: Hm, seems like it can run through 45 * 2053 steps pretty fast; added

examples/mlperf/lr_schedulers.py
chaosagent: needs other changes in lr_scheduler.py anyways, will defer them until resnet fp16 PR

PR #3654, opened 2024-03-07
fix: make Tensor.rand produce correct values for float16

### LINKED ISSUE #3611: Tensor.rand goes to -inf with default half
```
dtypes.default_float = dtypes.half
print(Tensor.randn((128, 128)).min().numpy())
```
often produces -inf

skoshx:
Due to precision loss when casting to float16, the data distribution created by custom_random isnt correctly in the interval ]0, 1[, but instead in the interval ]0, 1], which causes the Tensor.randn to incorrectly generate values of infinity.

The solution uses a scaling value to make sure the values stay under 1, when using half precision.

Closes #3611


tinygrad/tensor.py
```diff
@@ -1011,5 +1011,6 @@ def custom_random(out:Buffer):
   Tensor._seed += 1
   if DEBUG >= 2: print(f"*** {out.device}   rand  seed {Tensor._seed} size {out.size:<15d} dtype {out.dtype}")
   rng = np.random.default_rng(Tensor._seed)
-  rng_np_buffer = rng.random(size=out.size, dtype=np.float32).astype(dtype=out.dtype.np, copy=False)
+  scaling_factor = 0.999755 if out.dtype == dtypes.float16 else 1
```
chenyuxyz, line 1014: this does not look like a generic fix

skoshx: Okay, what about running through all the values that are 1 and decrementing them by some epsilon?

chenyuxyz: that does not work because it changed the output distribution

skoshx: Right ofcourse. Passing dtype straight to `rng.random` gives TypeError: Unsupported dtype dtype('float16') for random

chenyuxyz: I see. I think #3614 is a better short term approach given that `sin` also has precision problem with fp16. If the goal is to fix randn with float16 then do cast to half at the end of Box-Muller is safer.

Hopefully new threefry don't have this problem @wozeparrot fyi

skoshx: But the problem is also that `Tensor.rand` just doesn't give correct distributions ]0, 1[ using float16. And that should be fixed too. I get that the scaling solution doesn't look generic, but as far as I can tell it keeps the distribution true, and solves the root cause

skoshx: > that does not work because it changed the output distribution

Reflecting on this a bit. I think the distribution actually stays the same, because the values that end up becoming 1 are the values that are so close to 1 that precision differences cause it to go to 1. So taking away epsilon keep the distribution as close to the true distribution as possible within float16?

chenyuxyz: it's possible that 1-eps can work, but I don't think scaling and 1-eps can both represent the same distribution.

the largest number less than 1 in float16 is 2047/2048. So maybe a cleaner solution would be randint(low=0, high=2047) / 2048 and convert to float16. I think this is similar to how threefry samples.

Not clear to me if these random.random are designed to be slightly less than true continuous [0, 1) uniform random.

skoshx: Yeah okay. I think this enforcing the `Tensor.rand` to ]0, 1[ is clearly an edge-case only for `Tensor.randn` due to the Box-Mueller transform. As such changing the whole implementation for all `Tensor.rand` calls feels a bit overkill, and results in a tiny bit of extra compute even when the interval being "correct" isn't so important.

Would it be possible to have some option `enforce_open_interval` that takes (1 - eps) away from overflowing numbers (and perhaps also from underflowing numbers if we care for correctness), and then just use that in `Tensor.randn`? Seems like a good compromise.

chenyuxyz: Tensor.rand should have range that includes 0 (although for torch on cpu it's (0, 1] which contradicts to their document...).

We are changing the rand implementation to threefry, which is similar to [0, 2^N-1] / 2^N but on GPU, so probably not going to add additional arg to current Tensor.rand.

unknownusername504: > it's possible that 1-eps can work, but I don't think scaling and 1-eps can both represent the same distribution.
> 
> the largest number less than 1 in float16 is 2047/2048. So maybe a cleaner solution would be randint(low=0, high=2047) / 2048 and convert to float16. I think this is similar to how threefry samples.
> 
> Not clear to me if these random.random are designed to be slightly less than true continuous [0, 1) uniform random.

That sounds clean, was thinking something like that but your solution is simpler:

import numpy as np
scaling_factor = np.float16(1 - 1 / (2 ** np.dtype(np.float16).itemsize * 8))
np.minimum(rng.random(size).astype(np.float16), scaling_factor)

chenyuxyz:
```
  if out.dtype == dtypes.half: rng_np_buffer = (rng.integers(low=0, high=2047, size=out.size) / 2048).astype(np.half, copy=False)
  else: rng_np_buffer = rng.random(size=out.size, dtype=np.float32).astype(dtype=out.dtype.np, copy=False)
```
playing with this

skoshx: Right, so the long term solution is to just implement threefry? If that's the case, then I guess this PR is useless, you can close it out

unknownusername504: Seems to work, should do dist analysis:

import numpy as np

def custom_random(size=None, dtype=np.float16):
    low_range = 0
    high_range = (2 ** np.dtype(dtype).itemsize * 8) - 1
    random_int = np.random.randint(low_range, high_range, size=size)
    return dtype(random_int / high_range)

def test_custom_random():
    size = (100000,)
    random_values = custom_random(size=size)
    assert np.all(random_values < 1), "Generated random numbers are not strictly less than 1."

if __name__ == '__main__':
    test_custom_random()

skoshx: Sorry, didn't see your comment. Any idea how long the threefry implementation will take? 

I can change this PR to your solution as a short-term fix if the threefry implementation is still some time away. What do you think?

unknownusername504: Seems to work, pretty clean idea chenyuxyz.

import numpy as np

def custom_random(size=None, dtype=np.float16):
    low_range = 0
    # Plus one since 1bit sign
    mantissa_bits = np.finfo(np.float16).nmant + 1
    high_range = (2 ** mantissa_bits) - 1
    random_int = np.random.randint(low_range, high_range, size=size)
    return dtype(random_int / high_range)

def test_custom_random():
    size = (100000,)
    random_values = custom_random(size=size)
    print(random_values)
    assert np.all(random_values < 1), "Generated random numbers are not strictly less than 1."

if __name__ == '__main__':
    test_custom_random()

chenyuxyz: we can merge if it's nice and clean, and if it contains robust tests (not just test max!=1, test boundaries includes 0, and use the existing infra to test the distribution)

skoshx: Sounds good, i'll make the necessary changes.

skoshx: Ready for review :)

PR #3658, opened 2024-03-08
limit kernel optimization threadgroup size

chenyuxyz:
testing... all the Metal "Compiler encountered an internal error" seemed to be from the same big threadgroup memory issue

@flammit fyi

nvm doing it wrong, it does not revert the change even if late assert failed


chenyuxyz:
probably want to make apply_opt a function so we can do post apply check and revert the opt if failed

PR #3661, opened 2024-03-08
use custom KernelOptError in kernel opt

chenyuxyz:
be more specific about invalid kernel opt, used that in test_linearizer_failures.

make BEAM kernel search work even with assertion disabled.

`BEAM=2 python3 -O examples/llama.py  --temperature=0 --count=10 --prompt="Hello." --timing`

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -1,7 +1,7 @@
 import numpy as np
 import unittest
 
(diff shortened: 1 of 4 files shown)


chenyuxyz:
tested with llama, gpt2 and cifar on mac and tinybox with -O.

PR #3664, opened 2024-03-09
search: catch RuntimeError when timing acted_lins

flammit:
when compilation succeeds, but runtime fails due to thread limits on METAL, this allows a beam search to proceed, treating this the same way as a compile failure.

### DIFF
--- a/tinygrad/features/search.py
+++ b/tinygrad/features/search.py
@@ -124,7 +124,8 @@ def beam_search(lin:Linearizer, rawbufs, amt:int, allow_test_size=True) -> Linea
         if lib in seen_libs: continue
         #print(acted_lins[i].colored_shape(), acted_lins[i].applied_opts)  # for debugging BEAMs that segfault
         seen_libs.add(lib)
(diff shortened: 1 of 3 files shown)


flammit:
The assert I want to catch is in the runtime `__call__`: 
 https://github.com/tinygrad/tinygrad/blob/8ae85b2cf5b5e38eaa9675ed3f327474137b4574/tinygrad/runtime/ops_metal.py#L46

Should those runtime ones be made into a more explicit exception too?  `ops_has.py` has one as well to check the kernel_arg_segment_size.

chenyuxyz:
I think it's a good start to change those assertion inside `__call__` to RuntimeError, especially if you want to catch it and do other, instead of just fail.

We have started to assign more specific error types instead of just asserting and over time we will have a better guideline on this.

flammit:
> I think it's a good start to change those assertion inside `__call__` to RuntimeError, especially if you want to catch it and do other, instead of just fail.
> 
> We have started to assign more specific error types instead of just asserting and over time we will have a better guideline on this.

Sounds good - will do the RuntimeError here in the program call path on this PR, but we probably should also have a specific error in the compile paths too.

Issue #3662, opened 2024-03-09
Make ShapeTracker independent of symbolic


geohot:
First step, `merge_views` needs to move from `ShapeTracker` to `View`, but currently it uses ShapeTracker.real_strides

chenyuxyz:
i think this is done. the only thing before removing symbolic is its usage in the symbolic shape

PR #3670, opened 2024-03-10
Move "merge_views" from shapetracker to views class as addition operator

davidjanoskyrepo:
Refactor to address https://github.com/tinygrad/tinygrad/issues/3662


geohot:
I want it to not use `real_strides` I don't just want the functions moved. View shouldn't require Symbolic

unknownusername504:
Ah makes sense why it is a bounty. I assume modifying the algorithms is fair game then.

chenyuxyz:
done in #3686

PR #3673, opened 2024-03-10
Fix Tensor's __repr__ for printing out grad

francislata:
This PR fixes a small issue when checking for `self.grad` inside `__repr__` after the implementation of `__bool__`.

Here's a simple way to reproduce the issue:
Screenshot 2024-03-10 at 4 11 45 PM

Here's the case when the fix is applied:
Screenshot 2024-03-10 at 4 12 53 PM

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -334,6 +334,12 @@ class TestTinygrad(unittest.TestCase):
     with self.assertRaises(TypeError):
       _a = Tensor([3]) in [Tensor([3]), Tensor([4]), Tensor([5])]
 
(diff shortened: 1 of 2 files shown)


chenyuxyz:
Can you add a test for this? Something in test_tensor that calls `repr`. Don't need to check the output value just make sure it does not break

francislata:
> Can you add a test for this? Something in test_tensor that calls `repr`. Don't need to check the output value just make sure it does not break

sure I can add that in 👍

chenyuxyz:
Thanks!

PR #3676, opened 2024-03-10
add mixtral and 6 gpus cifar to tinybox ci

chenyuxyz:


chenyuxyz:
~2 tok/s for mixtral 8x7B and 48 seconds cifar training with 6 gpus

PR #3681, opened 2024-03-11
examples/stable_diffusion: support model checkpoints without alphas_cumprod key

rnxyfvls:
Most models on civitai don't have this key, and the previous code would generate a black image. This patch initializes alphas_cumprod with a default and allows these models to work.


examples/stable_diffusion.py
```diff
@@ -512,9 +512,16 @@ def encode(self, text):
       bpe_tokens = bpe_tokens[:75]
     return [49406] + bpe_tokens + [49407] * (77 - len(bpe_tokens) - 1)
 
+
+def get_alphas_cumprod(beta_start=0.00085, beta_end=0.0120, n_training_steps=1000):
```
chenyuxyz, line 516: use indent=2

rnxyfvls: fixed

PR #3686, opened 2024-03-11
View.__add__ to merge_view

chenyuxyz:
verified the cases that used real_strides are redundant (https://github.com/chenyuxyz/tinygrad/pull/166)


unknownusername504:
Ah I see so you didn't even need real_strides in the end.

PR #3689, opened 2024-03-11
linearizer ast as a tuple of lazyops

Qazalin:
moving from `ast` to `*ast` will make the diff for multioutput linearizing very minimal!
With this change, `opts` should be a named arg.


geohot:
Simple!

PR #3699, opened 2024-03-12
multioutput ScheduleItem

Qazalin:
Unlike the linearizer refactor, changing ScheduleItem invalidates the old usage.


geohot:
Have you checked the cache still works? Perhaps bump the db number since the old cache will be invalid now.

Qazalin:
This doesn't impact the cache DB since `@diskcache`'s compiled kernels and rawbufs are the same.
Though JITRunner's CacheCollector treating rawbufs[0] as special should get a refactor once there is a functional change to rawbufs/output count.

geohot:
I'm saying the ast is now a tuple. Isn't that going to affect the cache key?

Qazalin:
search.py uses lin.ast instead of the ScheduleItem's.
Screenshot 2024-03-13 at 5 38 47 PM
so far I've only updated the API for linearizer. UOping/search for multioutput needs some extra work (#3706)

PR #3704, opened 2024-03-12
new constant folding depth 2

geohot:
Still annoying with the UOps list data structure. Need to just fix this.


geohot:
This test should go in the new UOps PR #4560

PR #3705, opened 2024-03-12
kernel: limit shared memory usage when adding opts

flammit:
This cleans up a lot of the compile failures we were having.

Also cleaned up `test_linearizer_failures.py` after testing on METAL/GPU/CUDA/HIP/HSA.  CUDA was added to failures 13 and 14 as it also fails there (looks like it should fail on all platforms).


chenyuxyz:
Do you think we can update (relax or remote) this line too [https://github.com/tinygrad/tinygrad/blob/2024b24f353bb6aa4a683250f0941529edb95626/tinygrad/features/search.py#L91](https://github.com/tinygrad/tinygrad/blob/2024b24f353bb6aa4a683250f0941529edb95626/tinygrad/features/search.py#L91)?

flammit:
> Do you think we can update (relax or remote) this line too
> 
> https://github.com/tinygrad/tinygrad/blob/2024b24f353bb6aa4a683250f0941529edb95626/tinygrad/features/search.py#L91
> 
> ?

Yes definitely can be removed now.

chenyuxyz:
Cool! How did you find these numbers?

flammit:
For METAL, it's in <https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf> under "Maximum total threadgroup memory allocation", though as discussed with nimlgen here <https://discord.com/channels/1068976834382925865/1068982781490757652/1211409374841540759> it might be sharing that 32KB with registers.

For HIP/HSA, it's in <https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna3-shader-instruction-set-architecture-feb-2023_0.pdf> on Section 2.3 (page 11 of the PDF).

For CUDA, the answer is more complicated and architecture-specific I think.  By default though, I believe a minimum of 48KB is available <https://developer.nvidia.com/blog/using-shared-memory-cuda-cc/> with modern architectures being able to use a lot more.

PR #3706, opened 2024-03-12
multioutput uoping infra

Qazalin:
since all outbufs are the same shape opts and dimensions are valid. UOping needs a small refactor.

I'm leaving out multioutput FLOPS and mem from this refactor, it can't yet be tested e2e.
I think we could still use the first lb for the cache key


geohot:
Maybe we just remove LazyOp info?

PR #3710, opened 2024-03-13
dataloader worker/shm cleanup

chaosagent:


chenyuxyz:
seems fine. Tested with `time MODEL=resnet python examples/mlperf/model_eval.py` and ctrl-c

PR #3721, opened 2024-03-13
feat: add support for `prefix` in `load_state_dict`

skoshx:
Also, this is super useful for instance when loading weights, for a model that doesn't need all of the weights. I'm already using this to load weights from `clip-vit-large-patch14`, but only load the text related weights wit `prefix="text_model."`.

Also matches API of `load_state_dict` with `get_state_dict`.


geohot:
Repeated format string and a new NOQA. If you want this functionality, it needs a better implementation and tests

PR #3730, opened 2024-03-14
Fix kwargs in JIT

Rohan138:
Minor bug fix: Currently calling the same `JIT`ted function with different kwargs breaks because the `expected_name_sts_dtype_device` changes.

This fixes the issue, but if it feels hacky, I can add a test or find a better solve (imo `itertools.chain`-ing the args and kwargs together isn't the best way of doing this).

### DIFF
--- a/test/test_jit.py
+++ b/test/test_jit.py
@@ -98,17 +98,17 @@ class TestJit(unittest.TestCase):
 
   def test_reorder_kwargs_jit(self):
     @TinyJit
-    def add_kwargs(first, second): return (first+second).realize()
+    def add_kwargs(first, second): return (first/second).realize()
     for _ in range(2):
       a = Tensor.randn(10, 10)
       b = Tensor.randn(10, 10)
-      c = add_kwargs(second=a, first=b)
-      np.testing.assert_allclose(c.numpy(), a.numpy()+b.numpy(), atol=1e-4, rtol=1e-5)
+      c = add_kwargs(second=b, first=a)
+      np.testing.assert_allclose(c.numpy(), a.numpy()/b.numpy(), atol=1e-4, rtol=1e-5)
     for _ in range(2):
       a = Tensor.randn(10, 10)
       b = Tensor.randn(10, 10)
       c = add_kwargs(first=a, second=b)
-      np.testing.assert_allclose(c.numpy(), a.numpy()+b.numpy(), atol=1e-4, rtol=1e-5)
+      np.testing.assert_allclose(c.numpy(), a.numpy()/b.numpy(), atol=1e-4, rtol=1e-5)
     assert_jit_cache_len(add_kwargs, 1)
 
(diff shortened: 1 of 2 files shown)


chenyuxyz:
can you include a regression test that fails on master?

Rohan138:
Added test and fixed lint

test/test_jit.py
```diff
@@ -96,6 +96,21 @@ def add_kwargs(first, second): return (first+second).realize()
       np.testing.assert_allclose(c.numpy(), a.numpy()+b.numpy(), atol=1e-4, rtol=1e-5)
     assert_jit_cache_len(add_kwargs, 1)
 
+  def test_reorder_kwargs_jit(self):
+    @TinyJit
+    def add_kwargs(first, second): return (first+second).realize()
```
geohot, line 101: Can you make the test something non commutative so we can confirm the order is right?

Rohan138: Actually, good catch. `first/second` breaks for example. Let me see if there's a better fix.

Rohan138:
Test is division now; switched to using `inspect.signature(fxn)`. I think this is cleaner than the `enumerate(args)` + `sorted(kwargs)`.

Fixing `export_model` errors now, but let me know if you want me to rollback to the current way of doing it instead.

Rohan138:
NVM; keeping it as is. I might switch to inspect.signature and clean up model export code (It's somewhat hacky rn) in a separate PR.

tinygrad/features/jit.py
```diff
@@ -91,7 +91,7 @@ def __get__(self, obj, objtype): return functools.partial(self.__call__, obj)
 
   def __call__(self, *args, **kwargs) -> ReturnType:
     # all inputs (except const) are realized
-    input_tensors: Dict[Union[int, str], Union[LazyBuffer, MultiLazyBuffer]] = { cast(Union[int, str], k):v.realize().lazydata for k,v in itertools.chain(enumerate(args), kwargs.items()) if v.__class__ is Tensor }  # noqa: E501
+    input_tensors: Dict[Union[int, str], Union[LazyBuffer, MultiLazyBuffer]] = { cast(Union[int, str], k):v.realize().lazydata for k,v in itertools.chain(enumerate(args), sorted(kwargs.items(), key=lambda x: str(x[0]))) if v.__class__ is Tensor }  # noqa: E501
```
chenyuxyz, line 94: Do you need to specify this `key=` function? I think it's the same default behavior without specifying the `key`

Rohan138: No, `key` and `reverse` are kwarg-only parameters to sorted:
https://stackoverflow.com/questions/75761879/python-sorted-expected-1-argument-got-3-trying-to-understand-named-argument

chenyuxyz: i mean just `sorted(kwargs.items())`

PR #3739, opened 2024-03-14
fix buf_index not found case in _apply_tc_opt

chenyuxyz:
ValueError if src.src[0] is not a LOAD. Replaced with returning None in _apply_tc_opt and test to make sure the net output is KernelOptError.


chenyuxyz:
@flammit fyi

PR #3740, opened 2024-03-14
Support shape changing bitcast

obadakhalili:
Addresses the first part of #3422:

> Currently you can't bitcast between dtypes with different itemsize


tinygrad/tensor.py
```diff
@@ -988,7 +988,27 @@ def cast(self, dtype:DType) -> Tensor:
...
+      assert self.shape != (), "bitcasting between types of different sizes is not allowed for scalar tensors"
+      to_uint = {1: dtypes.uint8, 2: dtypes.uint16, 4: dtypes.uint32, 8: dtypes.uint64}
+      if self.dtype.itemsize > dtype.itemsize:
+        repeat = self.dtype.itemsize // dtype.itemsize
+        as_repeated_uint = self.repeat([1] * (self.ndim - 1) + [repeat]).reshape(self.shape[:-1] + (repeat, -1))\
+          .transpose(-2, -1).flatten(-2).bitcast(to_uint[self.dtype.itemsize])
+        idxs = Tensor.arange(repeat).repeat([self.shape[-1]])
+        bits = 8 * dtype.itemsize
+        shift = (2 ** (bits * idxs)).cast(to_uint[self.dtype.itemsize])
+        def div(t1: Tensor, t2: Tensor): return mlops.Div.apply(t1, t2)
```
obadakhalili, line 1001: I had to add this temporarily to focus on implementing bitcast because `.div` sometimes returns incorrect results when used against unsigned data types.

tinygrad/codegen/uops.py
```diff
@@ -36,7 +36,7 @@ def wfxn(*args):
 
 python_alu = {
   UnaryOps.LOG2: lambda x: math.log2(x) if x > 0 else -math.inf if x == 0 else math.nan,
-  UnaryOps.EXP2: hook_overflow(math.inf, lambda x: math.exp(x*math.log(2))),
+  UnaryOps.EXP2: hook_overflow(math.inf, lambda x: 2**x),
```
obadakhalili, line 39: The implementation of `UnaryOps.EXP2` in the Python runtime generated from the example `Tensor([0, 8, 16]).cast(dtypes.uint32)` produces inaccurate results. The results are very close to the correct ones but when casted to `dtypes.uint32` they will be off by one, which is problematic for the implementation of bitcasting as casting has to be accurate. Using Python's power operator directly resolves the issue.

obadakhalili:
Now I will work on refactoring the code mentioned in the first comment. Then I should address the issues in the HIP runtime.

obadakhalili:
Can I get feedback on this? @geohot
The remaining failing tests should be fixed when the HIP backend is removed, which my understanding is happening very soon.

chenyuxyz:
done in #3973

PR #3749, opened 2024-03-15
bf16 tests in test_dtype.py

chenyuxyz:
With bf16 creation and bf16 to numpy, we can test bf16 in test_dtype. Only support HIP now as it needs bf16 buffer support. Also the rtoal is slightly larger


chenyuxyz:
tested on tinybox too

PR #3750, opened 2024-03-15
add lars to nn

wozeparrot:
Can do the same thing for LARS and SGD as is done for LAMB and Adam.

Only thing that's dropped is the `skip_list`, but you should be able to get `skip_list` behaviour just from using two optimizers, one SGD for the params in `skip_list` and LARS for the rest. @chaosagent for resnet.

Should resnet be moved to this LARS?

FYI, that grad realize seems to still be required, at least for SGD, since without it some of the optim tests will fail.


geohot:
I'm very supportive of this. Let's get resnet training switched over and confirm it's still good

chaosagent:
Looks fine to me. I didn't want to do the merged lars/sgd because I didn't think lars was common enough to warrant inclusion in tinygrad/. For testing, I wanted to make the mlperf components explicitly be copies of and tested against the mlperf reference. I tried to make mlperf resnet "showcase quality" as much as I could too, but maybe this is fine to showcase "dual optimizers"? It would be cool to have our mlperf submissions also be API showcases.

chaosagent:
Can you add a test for dual optimizers vs mlperf reference skip list?

chenyuxyz:
i will do a full resnet training with this tonight and merge tomorrow if all looks good

chenyuxyz:
It increased the memory used for search. Now BEAM=4 with BS=768 erred with `RuntimeError: HSA Error 4104: HSA_STATUS_ERROR_OUT_OF_RESOURCES`. Trying BS=624 now

wozeparrot:
Might be cause `t.grad` is realized now instead of just contiguous?

chaosagent:
Did you run into a problem with contiguous?

wozeparrot:
yea it fails on the sgd tests

chaosagent:
Why does it fail? Well I can look at it I guess.

I wrote contiguous to have the step be mostly in one schedule, so that the scheduler has the freedom to do some cool scheduling. With realize, I think the earliest weights get realized first, and that leaves a lot of gradient intermediates floating around in between.

chaosagent:
Btw, why do you think lars should be in tinygrad/?

wozeparrot:
> Btw, why do you think lars should be in tinygrad/?

tensorflow has it, and I feel like if we want the mlperf stuff to be a tinygrad showcase, stuff like optimizers shouldn't be put outside of tinygrad, since if people now want to use LARS (like me) they would have to copy paste a file somewhere and it just harms the ux.

chenyuxyz:
3 epochs in the eval looks quite different (33 is this pr, 24/27 are good ones) https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf

also need to fix the sgd realize before merging this

wozeparrot:
Oh I know what the issue is, our sgd is using the standard momentum calculation where you just take the weight decayed grad as the momentum, but the mlperf lars for some reason uses post lr grad for momentum update, so the params that are in skip_list don't have matching momentum updates.

chaosagent:
what tests fail? #3770 passes CI

wozeparrot:
huh odd they don't seem to fail locally for me anymore either, weird, maybe it was something else then.

chaosagent:
the double optimizers is also a problem -- there will be two corealizes.

related to this topic, I think it's quite annoying for optim.step() to realize everything. If you need to also realize a loss or an output, or as in this case, have two optimizers, the schedule will be split up.

chenyuxyz:
> Can you add a test for dual optimizers vs mlperf reference skip list?

I want to merge this so it does not stale. Since we don't have a way to train full resnet now, I am going to run 1 GPU training with this for 12 hours and merge if all matches previous runs.

chenyuxyz:
`not in` doest not work for tensors if it calls `__eq__`

chenyuxyz:
trained on 1 GPU for 19.5 epochs <https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/igxv7yx6/overview>, everything looks identical enough to the old master <https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/qptdg5q0>

PR #3765, opened 2024-03-15
[WIP] small copies in a bg aql queue

nimlgen:
This seems to be faster for small sizes

### DIFF
--- a/tinygrad/runtime/graph/hsa.py
+++ b/tinygrad/runtime/graph/hsa.py
@@ -61,13 +61,14 @@ class HSAGraph(MultiDeviceJITGraph):
 
     # Build queues.
     self.virt_aql_queues: Dict[Compiled, VirtAQLQueue] = {dev:VirtAQLQueue(dev, 2*len(self.jit_cache)+16) for dev in self.devices}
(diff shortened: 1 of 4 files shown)


tinygrad/runtime/graph/hsa.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, jit_cache: List[JitItem], input_rawbuffers: List[Buffer], var
 
     # Build queues.
     self.virt_aql_queues: Dict[Compiled, VirtAQLQueue] = {dev:VirtAQLQueue(dev, 2*len(self.jit_cache)+16) for dev in self.devices}
+    self.virt_aql_copy_queues: Dict[Compiled, VirtAQLQueue] = {dev:VirtAQLQueue(dev, 2*len(self.jit_cache)+16) for dev in self.devices}
```
geohot, line 64: Why is this in a different queue from the compute kernels?

nimlgen: This is faster from my tests. I think because it starts to execute them in parallel. I can try to put them in one queue with some different header and hope that CP will put the in parallel as well.

PR #3770, opened 2024-03-16
[experimenting] use contiguous instead of realize in optim

chaosagent:
Just opening a PR to run CI lol

### DIFF
--- a/tinygrad/nn/optim.py
+++ b/tinygrad/nn/optim.py
@@ -34,7 +34,7 @@ class SGD(Optimizer):
   def step(self) -> None:
     for i, t in enumerate(self.params):
       assert t.grad is not None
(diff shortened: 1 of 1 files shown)


wozeparrot:
interesting

geohot:
Oh I'd merge this.

chaosagent:
OK, but I wanted to write tests. But I don't know a good test API.

geohot:
Adding tests here would be welcome. Does it fail without the contiguous?

chaosagent:
> Does it fail without the contiguous?

I was thinking about tests to prevent kernel count, memory, etc regression

chaosagent:
hm, causes weird numbers!

geohot:
It's not numerics. It might be assign.

PR #3771, opened 2024-03-16
increase jit batch size progressively

nimlgen:
Saves some enqueue time for big jit caches

llama 7b gpu6
```
enqueue in  39.35 ms
total  60.86 ms, 16.43 tok/sec
```
vs
```
enqueue in  52.92 ms
total  65.90 ms, 15.17 tok/sec
```

resnet beam=4
```
30  339.68 ms run,   36.58 ms python,  26.77 ms fetch data,  276.33 ms HSA * 6,  7.03 loss, 0.00 acc, 0.007821 LR, 94.53 GB used,  45560.44 GFLOPS
```
vs
```
30  369.88 ms run,   57.52 ms python,  24.03 ms fetch data,  288.33 ms HSA * 6,  7.03 loss, 0.00 acc, 0.007821 LR, 94.53 GB used,  41840.78 GFLOPS
```

### DIFF
--- a/tinygrad/features/jit.py
+++ b/tinygrad/features/jit.py
@@ -36,15 +36,17 @@ def get_jc_idxs_with_updatable_var_vals(jit_cache: List[JitItem]) -> List[int]:
 def apply_graph_to_jit(jit_cache: List[JitItem], input_rawbuffers: List[Buffer], var_vals: Dict[Variable, int]) -> List[JitItem]:
   # Split JIT cache into batches for faster graph execution.
   # This allows the accelerator to run some batches while subsequent graphs are still being updated.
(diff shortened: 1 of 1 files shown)


chenyuxyz:
This is great! Is there a upper limit how big a single batch can be?

chenyuxyz:
llama 70B 4.66 tok/s!

nimlgen:
I decided not to set an upper limit. I tested it (4096), and it really doesn't change anything. We could add one if something goes wrong

chenyuxyz:
FYI this one might cause resnet training to hang on the second epoch. It might not be this one and I need to do more test, but 3 runs I included this hung, and the one that is on previous commit has trained more than 5 epochs as of now

PR #3772, opened 2024-03-16
end-to-end multioutput

Qazalin:
`Tensor.corealize` attempts to fuse the outputs.
The scheduler groups buffers using a basic level ordering algorithm. Getting to 1 kernel Adam requires smarter realize grouping, tinygrad has all the infrastructure to run multioutput DAGs.

### DIFF
--- a/tinygrad/features/jit.py
+++ b/tinygrad/features/jit.py
@@ -117,7 +117,7 @@ class TinyJit(Generic[ReturnType]):
       CacheCollector.start(var_vals)
       with Context(GRAPH=getenv("JITGRAPH", GRAPH.value)):
         self.ret = self.fxn(*args, **kwargs)
-        for p in get_parameters(self.ret): p.realize()
+        Tensor.corealize(get_parameters(self.ret))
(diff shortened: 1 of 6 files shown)


tinygrad/features/jit.py
```diff
@@ -160,15 +160,16 @@ def start(self, var_vals:Optional[Dict[Variable, int]]=None):
     self.placeholders: WeakKeyDictionary[Buffer, PlaceHolder] = WeakKeyDictionary()
     self.var_vals = var_vals if var_vals is not None else {}
 
-  def add(self, prg, rawbufs, var_vals):
+  def add(self, prg, rawbufs, var_vals, outbufs:Optional[List[Buffer]]=None):
```
nimlgen, line 163: Perhaps we can put information about writable buffers in `CompiledASTRunner`s?

Qazalin:
the JIT refactor seems to be wrong for CUDA, I need to isolate that change as a separate PR first.
https://github.com/tinygrad/tinygrad/actions/runs/8316267565/job/22755525144

PR #3775, opened 2024-03-16
PTX - implement float 4, ptr arithmetics and other speed improvements

SzymonOzog:


SzymonOzog:
image
Stable diffusion comparison master on the left, float4 on the right

test/test_search.py
```diff
@@ -6,9 +6,11 @@
 from tinygrad.device import Device, Buffer
 from tinygrad.ops import LoadOps
 from tinygrad.tensor import Tensor
+from tinygrad.helpers import getenv
 
 class TestTimeLinearizer(unittest.TestCase):
   def test_reasonable_time(self):
+    if getenv("PTX"): self.skipTest('PTX expects UOpsGraph, this passes UOps as list')
```
geohot, line 13: Can this just be fixed instead of adding a special case? It should always be UOpsGraph I think

tinygrad/codegen/linearizer.py
```diff
@@ -45,6 +45,7 @@ def uop_alu_idx(self, a:UOp, b, ops, ctx:Linearizer, op, dtype=dtypes.int32):
 
   # NOTE: the consts have to be cached for deduping of downstream uops to work
   def const(self, b:Union[int,float], dtype=dtypes.int32, insert_before=None) -> UOp:
+    b = int(b) if dtypes.is_int(dtype) else float(b)
```
geohot, line 48: This should be an assert and fixed elsewhere, it shouldn't be hotfixed here. The type should be correct, and if it isn't we should fix it where it isn't correct.

tinygrad/codegen/uops.py
```diff
@@ -167,6 +167,7 @@ def remove_childless(self, keep:Set[UOp]):
         for vu in ru.vin:
           has_child.add(vu)
       nu: List[UOp] = [x for x in self.uops if x in has_child or x in keep]
+      self.saved_exprs = {k:v for k,v in self.saved_exprs.items() if v in nu}
```
geohot, line 170: Does this have to be recomputed every time in the while loop, or is once at the end enough?

tinygrad/renderer/assembly.py
```diff
@@ -113,45 +135,70 @@ def cast(a:str, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
...
+        else:
+          kk(f"mov.b{lang.types[dtype][1:]} {ssa(u, 'acc')}, {const(args, dtype)};")
       elif uop == UOps.SPECIAL:
         assert args[1][0] != "i", "idx not supported"
         kk(f"mov.u32 %{args[1]}, {(lang.gid if args[1][0] == 'g' else lang.lid)[args[0]]};")
         r[u] = "%" + args[1]
         kernel = [f".reg .u32 %{args[1]};"] + kernel
-      elif uop == UOps.CONST: r[u] = const(args, dtype, mov=True)
+      elif uop == UOps.CONST:
+            if dtype.count > 1:
```
geohot, line 176: Spacing!!!

tinygrad/renderer/assembly.py
```diff
@@ -172,7 +219,7 @@ def cast(a:str, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
   return lang.render_kernel(kernel, function_name, bufs, c.items())
 
 class PTXLanguage(AssemblyLanguage):
-  kernel_prefix = """.version 7.8
+  kernel_prefix = """.version 7.5
```
geohot, line 222: Why did this change?

geohot:
Can you refactor the UOps replacement to use the new pattern matching engine?

tinygrad/renderer/assembly.py
SzymonOzog: Couldn't run on my machine with an older cuda install, figured that since we don't use any 7.8 specific features it would be better to change it in the repo to support older architectures as well

tinygrad/codegen/linearizer.py
SzymonOzog: I already posted this as an issue - will attempt to fix this

SzymonOzog:
> Can you refactor the UOps replacement to use the new pattern matching engine? I don't want to merge more of the old style code.

sure thing

tinygrad/codegen/linearizer.py
SzymonOzog: Removed the cast as the issue no longer causes an error but that still persists, and there seem to be multiple places where the type is wrong, I'll try looking into it later

test/test_search.py
SzymonOzog: removed the line and the test no longer crashes, probably got fixed in a different pr

chenyuxyz: yea was fixed in #3786

geohot:
This good to merge? How close are we on speed to CUDA now?

SzymonOzog:
> This good to merge? How close are we on speed to CUDA now?

Good to merge, currently stable diffusion is 1.3s/it whereas cuda is 1.15s/it on a 3090ti. I'm already working on more speed improvemets - tensor cores are ready an I'll make a pr with them once I clean stuff up

PR #3778, opened 2024-03-16
"TINY_HSA_DEVICES"

chaosagent:
Is there an existing way to do this in HSA?


wozeparrot:
If you set `ROCR_VISIBLE_DEVICES=0,1,2,3` etc. it should do this

chaosagent:
> If you set ROCR_VISIBLE_DEVICES=0,1,2,3 etc. it should do this

Thanks! Where did find this? Is there a list of env vars somewhere?

wozeparrot:
https://rocm.docs.amd.com/en/latest/conceptual/gpu-isolation.html

PR #3781, opened 2024-03-17
add outbufs info to CompiledASTRunner

Qazalin:
This is a prereq for multioutput since both JITRunner and HSAGraph assume rawbufs[0] is the only output.

### DIFF
--- a/docs/abstractions2.py
+++ b/docs/abstractions2.py
@@ -64,7 +64,7 @@ print(fxn.prg)
 # NOTE: fxn.clprg is the ClangProgram
 
 # run the program
-fxn.exec([out, a, b], [out])
+fxn.exec([out, a, b])
 
 # check the data out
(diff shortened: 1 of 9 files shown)


nimlgen:
We can discuss that, but I think that CompiledASTRunner should have info about output buffers or kernels. This info is also needed to build hsa graph deps

Qazalin:
The approach here is to do this at `exec` because that is when the `Buffer`s for the outputs are created.
@nimlgen does the hsa graph need the info before exec is called?

nimlgen:
I thought maybe CompiledASTRunner can have bitmask/positions of rawbufs which are used as output? hsagraph needs to know which buffers of `CompiledASTRunner` are read-only to build an optimal dependency chain.

geohot:
Hmm, I don't like adding more variables here. What is this more generically? Maybe something like types on all the buffers.

PR #3785, opened 2024-03-17
threefry again

wozeparrot:
Testing just not caching clear_l2 kernels


chenyuxyz:
ran benchmark 3 times, should be good

PR #3786, opened 2024-03-17
Fix: BEAM search with PTX fails

Eliulm:
`uops_to_asm` only accepts `UOpGraph` instead of list


geohot:
Can you add a test to CI that would have caught this?

Eliulm:
Sure thing!

Eliulm:
@geohot  Was unsure whether I should add the test in `test/external` or add to `test/test_search.py`. I think we want the later, sorry.

chenyuxyz:
I think having type checker catching this is fine. I am working on beam search tests for all ci devices

PR #3792, opened 2024-03-18
remove mixtral weight to clang first

chenyuxyz:
seems fine without it now


chenyuxyz:
somehow for benchmark it's 5% slower, but I could not reproduce locally

PR #3793, opened 2024-03-18
clear l2 by getting an astrunner so it doesn't get added to jit

chaosagent:
or we could have a ContextVar like INTERNAL_TENSOR_CALL

### DIFF
--- a/tinygrad/features/search.py
+++ b/tinygrad/features/search.py
@@ -7,8 +7,9 @@ from tinygrad.helpers import prod, flatten, DEBUG, CACHELEVEL, diskcache_get, di
 from tinygrad.dtype import ImageDType
 from tinygrad.codegen.linearizer import Linearizer
 from tinygrad.codegen.kernel import KernelOptError
-from tinygrad.tensor import Tensor
 from tinygrad.shape.symbolic import sym_infer, Variable
+from tinygrad import Tensor, dtypes
+from tinygrad.realize import create_schedule
 
 from tinygrad.codegen.kernel import Opt, OptOps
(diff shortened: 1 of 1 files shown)


wozeparrot:
I did context var here: [`371a8d4` (#3785)](https://github.com/tinygrad/tinygrad/pull/3785/commits/371a8d49bd38ab68f7f85efb83b47fbbb235c02c)

chaosagent:
> I did context var here: [371a8d4 (#3785)](https://github.com/tinygrad/tinygrad/pull/3785/commits/371a8d49bd38ab68f7f85efb83b47fbbb235c02c), not sure which is better.

I didn't notice you did the same thing lol :sweat_smile: 

I like the idea of first-class support for doing tensor calls internally... maybe can replace do_update_stats and such too.

geohot:
That "clear L2" hack should probably just be removed. It's unclear if it's even useful.

HSA should support proper L2 clearing.

wozeparrot:
Yea, its probably better just to wait for #3698, but then clear_l2 doesn't do much on non-HSA platforms, but then again, not sure if it ever did.

chaosagent:
> That "clear L2" hack should probably just be removed. It's unclear if it's even useful.

the original 1024*1024 isn't even as large as xtx's l2 cache, let along l3 :sweat_smile: 

it does do something though, i was able to get >1TB/s gemv without it in isolated test (but not with a 100MB clear)

chenyuxyz:
fyi #3785 is merged

wozeparrot:
I removed the cache thing from #3785, so 3785 is the same as the old threefry pr since it didn't seem to do anything. So this is still an issue.

PR #3800, opened 2024-03-18
hotfix: update inputs of correct transfers in hsagraph

nimlgen:


chenyuxyz:
ehh... we really need some tests to cover these.

nimlgen:
@chenyuxyz good?

PR #3802, opened 2024-03-18
Optimize bool multiply/indexing/nn.Embedding

alexlitz:
Optimizes bool tensor indexing by fusing cases where a bool tensor constant over a is cast then multiplied with another value before being added to an accumulator. The operating principle is if it is zero over that broadcast dim the computation can be eliminated. Sometimes this requires loop reordering if the dim it is being iterated over is not the lowest loop. So loops are reordered if needed, bools are pulled out it is check if the aforementioned criteria is meet if so the loop(s) are put in an if statement. This helps with the indexing speed as well as many cases with bool broadcast multiplies. The arange case can be optimized further to just more directly do the indexing but would be less general and this seems to get ~2/3 (depending of course on sizes) of the order of magnitude improvements.

Running:
`CLANG=1 NOOPT=1 TC=0 DEBUG=4 GRAPH=1 python3 test/external/external_test_mnist_data_select.py`
Before:
`*** CLANG      4 r_512_784_60000                        arg   4 mem  0.22 GB tm 101742.51ms/101820.88ms (    0.71 GFLOPS,    0.00 GB/s)`

After:
`*** CLANG      4 r_512_784_60000                        arg   4 mem  0.22 GB tm     10.70ms/    88.36ms ( 6750.10 GFLOPS,   17.75 GB/s)`

So a 10,000x speedup.

Needs some tweaks to work without NOOPT, LLVMIR backend does not seem to support rendering UOps.IF.


chenyuxyz:
closing as stale

alexlitz:
@chenyuxyz actually have the fixes that are mentioned as needed in the description, just pushed them to the branch.

PR #3806, opened 2024-03-18
tricky assign tests

chaosagent:
these should all pass (or assert)

not comprehensive!

### DIFF
--- a/test/test_assign.py
+++ b/test/test_assign.py
@@ -92,6 +92,25 @@ class TestAssign(unittest.TestCase):
     new = a + old_a
     np.testing.assert_allclose(new.numpy(), 4)
 
+  def test_assign_stale_replace(self):
+    a = Tensor.ones(4).contiguous().realize()
+    old_a = Tensor.zeros(4)
+    old_a.replace(a)
+    a.assign(Tensor.full((4,), 2.).contiguous())
+    # old_a should syntactically be 1 but actually its 2, which is fine... I guess
+    new = a + old_a
+    np.testing.assert_allclose(new.numpy(), 4)
+
+  def test_assign_unrealized(self):
+    a = Tensor.ones(4)
+    b = a * 2
+    # normal lazydata overwrite
+    b += 4
+    b = b.contiguous()
+    # this one may or may not reuse buffer from previous line
+    b += b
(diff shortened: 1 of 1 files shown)


test/test_assign.py
```diff
@@ -92,6 +92,25 @@ def test_assign_changes(self):
     new = a + old_a
     np.testing.assert_allclose(new.numpy(), 4)
 
+  def test_assign_stale_replace(self):
+    a = Tensor.ones(4).contiguous().realize()
+    old_a = Tensor.zeros(4)
+    old_a.replace(a)
+    a.assign(Tensor.full((4,), 2.).contiguous())
+    # old_a should syntactically be 1 but actually its 2, which is fine... I guess
+    new = a + old_a
```
geohot, line 101: I don't think this is fine, it should be 1

test/test_assign.py
```diff
@@ -107,6 +126,69 @@ def test_assign_diamond_alt(self):
...
+    np.testing.assert_allclose(b.numpy(), 8)
+
+  def test_assign_cycle(self):
+    a = Tensor.ones(4).contiguous().realize()
+    b = Tensor.ones(4).contiguous().realize()
+    c = a + 1
+    a.assign(b + 1)
+    b.assign(c)
+    with self.assertRaises(Exception):
+      # there is a cycle in the dag for this one.
```
geohot, line 176: I don't think this should be a cycle, it should keep the first generation a around.

geohot:
@chaosagent assign should be fixed. It will assert in the cycle case. Want to check if it tests are what you expect?

chaosagent:
new assign looks good.

test_assign_fork_multirealize still fails (the assign edges can't be drawn because the readers are not in the same graph), but it might be fine depending on the intended semantics of lazybuffers/schedules. The current behavior has lazybuffers/schedule act like JIT (will perform operation on new buffer after assign). If you realize multiple times in a single train step, you will still run into the old assign problems.

I think the case to consider here is when training a model, references to the original LB are obviously kept, but if a lazy expression eg weight * 3 is kept around, should it be calculated from the original weights or updated weights? I think we should keep the current behavior for now, since it will add a bit of complexity to support the alternative behavior.

PR #3812, opened 2024-03-18
disable simplify_phi_loops

chenyuxyz:
this breaks BEAM search GPT2.


chenyuxyz:
`python -m pytest test/test_ops.py::TestOps::test_split` passed on tinybox but crashed in ci remu @Qazalin fyi

PR #3816, opened 2024-03-19
fp16 resnet (without expand backwards sum in float, doesn't work)

chaosagent:
this used to at least produce non-nan losses.

testing it...


chenyuxyz:
can you sync this to master? want to see where we are. I fixed the linearizer failure case, and I think the int index is better to be added at `expr_idx`

tinygrad/shape/symbolic.py
```diff
@@ -13,6 +13,7 @@ class Node:
   max: sint
   def render(self, ops=None, ctx=None) -> Any:
     if ops is None: ops = render_python
+    assert -2 ** 31 <= self.min <= self.max < 2 ** 31, f"calculation overflows 32-bit int! {self.__class__} {self.min} {self.max}"
```
chenyuxyz, line 16: i think this belongs to expr_idx so symbolic can be pure symbolic

tinygrad/tensor.py
```diff
@@ -343,7 +343,7 @@ def backward(self) -> Tensor:
 
     # fill in the first grad with one. don't use Tensor.ones because we don't need contiguous
     # this is "implicit gradient creation"
-    self.grad = Tensor(1.0, device=self.device, requires_grad=False)
+    self.grad = Tensor(1.0, device=self.device, requires_grad=False, dtype=self.dtype)
```
chenyuxyz, line 346: i merged this in #3613

test/test_linearizer_failures.py
```diff
@@ -215,5 +215,13 @@ def test_failure_27(self):
     for opts in all_failing_opts:
       helper_test_lin(Linearizer(ast), opts, failed_platforms=["METAL", "HSA", "CUDA", "CLANG"]) # "GPU" is a compiler failure
 
+  def test_failure_28(self):
```
chenyuxyz, line 218: this does not fail anymore after symbolic fix i think

tinygrad/realize.py
```diff
@@ -228,6 +228,7 @@ def create_schedule(outs:List[LazyBuffer], seen:Optional[Set[LazyBuffer]]=None)
           st_childs = dedup([s for s in tr_next.srcs if s.base == tr])
           if len(st_childs) > 1: break
           if st.size != st_childs[0].st.size: break
+          if len(st_childs) > 0 and tr.dtype.itemsize < tr_next.dtype.itemsize: break
```
chaosagent, line 231: need to isolate this line and test it out

chenyuxyz:
Can we have a version of this merged first? I wan to test bf16 training too. I think it's fine as long as it's not changing core tinygrad and all currently working stuff (float32 resnet, float32/half cifar) still works

chaosagent:
i'll try to clean this up today.

chaosagent:
for reference, the only additional thing needed to make this work is expand = float.expand.half

maybe a few of the changes here are unnecessary, need to do ablations.

PR #3808, opened 2024-03-18
test_linearizer_failures: add failing kernel from GPT2 CUDA

flammit:
this bisects to #3530 which I think caused the uops to be incorrect on all platforms

PR #3817, opened 2024-03-19
Fix factoring bug (O(n) arange related)

patosai:
See #3808, related to #3530

The division can't be factored, the param wasn't passed through when floordiv called itself. I see this caused a number of reverts, apologies.

### DIFF
--- a/tinygrad/shape/symbolic.py
+++ b/tinygrad/shape/symbolic.py
@@ -51,12 +51,12 @@ class Node:
   def __rfloordiv__(self, b:int): return NumNode(b) // self
   def __floordiv__(self, b:Union[Node,int], factoring_allowed=True):
     if isinstance(b, Node):
-      if b.__class__ is NumNode: return self.__floordiv__(b.b, factoring_allowed=factoring_allowed)
+      if b.__class__ is NumNode: return self.__floordiv__(b.b, factoring_allowed)
       if self == b: return NumNode(1)
       if (b - self).min > 0 and self.min >= 0: return NumNode(0) # b - self simplifies the node
       raise RuntimeError(f"not supported: {self} // {b}")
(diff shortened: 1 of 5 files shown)


tinygrad/shape/symbolic.py
```diff
@@ -51,12 +51,12 @@ def __rmul__(self, b:int): return self*b
   def __rfloordiv__(self, b:int): return NumNode(b) // self
   def __floordiv__(self, b:Union[Node,int], factoring_allowed=True):
     if isinstance(b, Node):
-      if b.__class__ is NumNode: return self // b.b
+      if b.__class__ is NumNode: return self.__floordiv__(b.b, factoring_allowed)
```
patosai, line 54: the problem was something like `Node.__floordiv__(NumNode(-16), factoring_allowed=False)` was being called, this ran and called itself but didn't give factoring_allowed so it ended up being `Node.__floordiv__(-16, factoring_allowed=True)`

chenyuxyz: Can you add a direct test for this case in test/unit/test_symbolic.py?

patosai: yup

patosai: test is added now, you can see the subtleness of the bug

chenyuxyz: thanks. yea that looks very sketchy, we probably have other bugs around div negative numbers. I will think about how to improve that.

change looks good. I will run more tests tomorrow and merge the fix if all looks good!

chenyuxyz:
Tested with local fuzzer and ran on ci benchmark 3 times, all looked good. Thanks for fixing the bug!

PR #3829, opened 2024-03-19
log optimized kernels and a script to compare with non-optimized ones

### LINKED ISSUE #3819: generic ast comparison
Verifying beam output directly in device.py was useful to debug the recent gpt2 beam flaky issue. We have ast to ast comparison in fuzz_linearizer, which would require generating a new kernel dataset if we changed the kernel and wanted it in fuzzer test. Factoring out a generic ast to ast comparison function and use it to optionally verify kernel opt output live, and reuse in fuzz_linearizer

flammit:
refactor fuzz_linearizer comparison to allow it to be used in for BEAM_VERIFY in device.py

Closes #3819

### DIFF
--- a/test/external/verify_kernel.py
+++ b/test/external/verify_kernel.py
@@ -9,7 +9,7 @@ from tinygrad.features.graph import print_tree
 
 # Example for GPT2:
 # 1) Run the model to log all kernels: `PYTHONPATH=. LOGKERN=/tmp/gpt2_kerns.txt JIT=1 HALF=1 BEAM=2 CACHELEVEL=0 CAST_BEFORE_VIEW=0 python3 examples/gpt2.py --count 10 --temperature 0 --timing`   # noqa: E501
-# 2) Validate the kernel correctness: `PYTHONPATH=. python3 ./test/external/verify_kernel.py --file /tmp/gpt2_lins.txt`
+# 2) Validate the kernel correctness: `PYTHONPATH=. python3 ./test/external/verify_kernel.py --file /tmp/gpt2_kerns.txt`
 
 if __name__ == "__main__":
   parser = argparse.ArgumentParser(description="Verify the correctness of one or more kernel", formatter_class=argparse.ArgumentDefaultsHelpFormatter)    # noqa: E501
--- a/test/external/verify_kernel.py
+++ b/test/external/verify_kernel.py
@@ -1,0 +1,57 @@
+import argparse
+from collections import defaultdict
+from extra.optimization.helpers import kern_str_to_lin
+from test.external.fuzz_linearizer import compare_linearizer
+from tinygrad.helpers import colored
+from tinygrad.features.graph import print_tree
+
+# Use this with the LOGKERN options to verify that all executed kernels are valid and evaluate to the same ground truth results
+
+# Example for GPT2:
+# 1) Run the model to log all kernels: `PYTHONPATH=. LOGKERN=/tmp/gpt2_kerns.txt JIT=1 HALF=1 BEAM=2 CACHELEVEL=0 CAST_BEFORE_VIEW=0 python3 examples/gpt2.py --count 10 --temperature 0 --timing`   # noqa: E501
+# 2) Validate the kernel correctness: `PYTHONPATH=. python3 ./test/external/verify_kernel.py --file /tmp/gpt2_lins.txt`
+
+if __name__ == "__main__":
+  parser = argparse.ArgumentParser(description="Verify the correctness of one or more kernel", formatter_class=argparse.ArgumentDefaultsHelpFormatter)    # noqa: E501
+  parser.add_argument("--kernel", type=str, default=None, help="a string of a tuple of (ast, applied_opts,)")
+  parser.add_argument("--file", type=str, default=None, help="a file containing a tuple of ast and applied_opts, one per line")
+  parser.add_argument("--rtol", type=float, default=1e-2, help="relative tolerance for numerical comparison")
+  parser.add_argument("--atol", type=float, default=1e-2, help="absolute tolerance for numerical comparison")
+  args = parser.parse_args()
+
+  if args.kernel is not None:
+    print("loading kernel from args")
+    kern_strs = [args.kernel]
+  elif args.file is not None:
+    print(f"loading kernel from file '{args.file}'")
+    with open(args.file, 'r') as file:
+      kern_strs = file.readlines()
+  else:
+    raise RuntimeError("no kernel specified; use --lin or --file options")
+
+  print(f"verifying {len(kern_strs)} kernel")
+
+  failed_ids = []
+  failures = defaultdict(list)
+  for i, kern_str in enumerate(kern_strs):
+    print(f"testing kernel {i}")
+    test_lin = kern_str_to_lin(kern_str)
+    for op in test_lin.ast: print_tree(op)
+    print(test_lin.colored_shape())
+    if (msg:=compare_linearizer(test_lin, None, None, None, rtol=args.rtol, atol=args.atol)[0]) != "PASS":
+      failed_ids.append(i)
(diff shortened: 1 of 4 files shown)


tinygrad/device.py
```diff
@@ -278,6 +278,12 @@ def get_linearizer(self, *ast:LazyOp) -> Linearizer:
         timed = sorted([(nm, tk, time_linearizer(tk, test_rawbuffers, allow_test_size=False, clear_l2=True)) for nm, tk in lins], key=lambda x: x[2])
         if DEBUG >= 1: print("  <  ".join(f"{nm:6s} : {lin.colored_shape(30, dense=True)} : {tm*1e6:8.2f} us" for nm, lin, tm in timed))
         k = timed[0][1]
+
+        if getenv("BEAM_VERIFY", 0): # compare the beam generated values with required_opts only
+          from test.external.fuzz_linearizer import compare_linearizer
+          if (msg:=(compare_linearizer(k, None, None, None, getenv("BEAM_VERIFY_RTOL", 1e-2), getenv("BEAM_VERIFY_ATOL", 1e-2)))[0]) != "PASS":
```
chenyuxyz, line 284: this is comparing with the fastest kernel, not necessarily from beam

chenyuxyz: oh i think it's more useful to have it verify the optimized output, not just the beam one. (imagine the usecase that the beam output fail, but maybe tc or hcopt have bugs)

flammit: how about just comparing all 2 or 3 linearizers in `lins` to the ground truth?

chenyuxyz:
i think core tinygrad should not import from external. if it's too much to port the fuzzer into core, can consider the logops https://github.com/tinygrad/tinygrad/blob/d17900bc456d36d7c0744b5411b9b8540e19a734/tinygrad/realize.py#L45

this is how we currently generate the kernel data set. so the debug workflow would be running with logops and write kernels to a file, then load the file into fuzzer for testing

flammit:
> i think core tinygrad should not import from external. if it's too much to port the fuzzer into core, can consider the logops
> 
> https://github.com/tinygrad/tinygrad/blob/d17900bc456d36d7c0744b5411b9b8540e19a734/tinygrad/realize.py#L45
> 
> this is how we currently generate the kernel data set. so the debug workflow would be running with logops and write kernels to a file, then load the file into fuzzer for testing

Yeah, I was using LOGOPS along with this other patch to fuzz_linearizer for testing <https://github.com/flammit/tinygrad/tree/fuzz_params> which depends on this PR.  I do want to add an option to limit ast_strs to only unique kernels and skip simples STORE<-CAST-<-LOAD ones by default.  For my testing, I hand-removed them from the GPT2 kernels after doing `cat ops.txt | sort | uniq > uniq_ops.txt`.

chenyuxyz:
i think the logging and fuzzer should be separated, like logging should keep all kernels and the log loader handles filtering the ones you don't want and dedup etc. it might be more confusing if the bug happens to be in simple cast etc

disk is filtered out in logger because the offset makes log file very big, and LoadOp filter is no-op now i think given that LOAD and STORE kernel are simplified in uop

flammit:
> i think the logging and fuzzer should be separated, like logging should keep all kernels and the log loader handles filtering the ones you don't want and dedup etc. it might be more confusing if the bug happens to be in simple cast etc
> 
> disk is filtered out in logger because the offset makes log file very big, and LoadOp filter is no-op now i think given that LOAD and STORE kernel are simplified in uop

Sorry I wasn't clear: I agree that in the next PR, the filtering and uniqueness of kernels will be done in `fuzz_linearizer` initialization, not in LOGOPS logic.

flammit:
@chenyuxyz should be good to review now.  I'm validating all lins now (hc, tc, and beamN).

I tried to clean up the typing in `extra/optimization/helper.py` as well but it's a bit much.

tinygrad/device.py
```diff
@@ -278,6 +278,14 @@ def get_linearizer(self, *ast:LazyOp) -> Linearizer:
         timed = sorted([(nm, tk, time_linearizer(tk, test_rawbuffers, allow_test_size=False, clear_l2=True)) for nm, tk in lins], key=lambda x: x[2])
         if DEBUG >= 1: print("  <  ".join(f"{nm:6s} : {lin.colored_shape(30, dense=True)} : {tm*1e6:8.2f} us" for nm, lin, tm in timed))
         k = timed[0][1]
+
+        if getenv("BEAM_VERIFY", 0): # compare the beam generated values with required_opts only
+          from test.external.fuzz_linearizer import compare_linearizer
```
chenyuxyz, line 283: I think core tinygrad should not import from test, even if it's a debug feature. That's why I proposed logging in core and use fuzzer to read the file.

tinygrad/device.py
```diff
@@ -278,6 +279,9 @@ def get_linearizer(self, *ast:LazyOp) -> Linearizer:
         timed = sorted([(nm, tk, time_linearizer(tk, test_rawbuffers, allow_test_size=False, clear_l2=True)) for nm, tk in lins], key=lambda x: x[2])
         if DEBUG >= 1: print("  <  ".join(f"{nm:6s} : {lin.colored_shape(30, dense=True)} : {tm*1e6:8.2f} us" for nm, lin, tm in timed))
         k = timed[0][1]
+        if logkern and logkern_level > 1: logkern.write("\n".join(map(lambda x: str((x[1].ast, x[1].applied_opts,)), timed[1:]))+"\n")
```
chenyuxyz, line 282: use `is not None` to compare with `None`. Also `writelines` to prevent manually add "\n"

test/external/verify_kernel.py
```diff
@@ -0,0 +1,57 @@
...
+
+  if args.kernel is not None:
+    print("loading kernel from args")
+    kern_strs = [args.kernel]
+  elif args.file is not None:
+    print(f"loading kernel from file '{args.file}'")
+    with open(args.file, 'r') as file:
+      kern_strs = file.readlines()
+  else:
+    raise RuntimeError("no kernel specified; use --lin or --file options")
```
chenyuxyz, line 30: `--kernel`

tinygrad/device.py
flammit: writelines doesn't append "\n", but this is still better than the having the previous join() call.

tinygrad/device.py
```diff
@@ -278,6 +279,9 @@ def get_linearizer(self, *ast:LazyOp) -> Linearizer:
         timed = sorted([(nm, tk, time_linearizer(tk, test_rawbuffers, allow_test_size=False, clear_l2=True)) for nm, tk in lins], key=lambda x: x[2])
         if DEBUG >= 1: print("  <  ".join(f"{nm:6s} : {lin.colored_shape(30, dense=True)} : {tm*1e6:8.2f} us" for nm, lin, tm in timed))
         k = timed[0][1]
+        if logkern is not None and logkern_level > 1: logkern.writelines(map(lambda x: str((x[1].ast, x[1].applied_opts,))+"\n", timed[1:]))
```
chenyuxyz, line 282: nit: i think `logkern.writelines([f"{(lin.ast, lin.applied_opts)}\n" for _,lin in timed])` is cleaner. similar below

also are you sure you want to drop timed[0] here? that's not necessarily the best kernel

flammit: will fix both to use f string.  when in BEAM, timed[0] will be written in line 284 below because it's assigned to k at line 281.

LOGKERN_LEVEL=1 will write just the kernel that is eventually returned, and LOGKERN_LEVEL=2 will write all timed ones.

chenyuxyz: oh this is other 1 or 2.... yea you are right. somehow i thought this is other beam candidates

chenyuxyz:
tested, works pretty well!

i am very supportive to add this feature all in core too

PR #3830, opened 2024-03-20
intel matmul

geohot:
It's strange, it seems like k isn't really 16 but actually 8.


extra/gemm/intel_matmul.py
```diff
@@ -0,0 +1,85 @@
...
+# https://hc34.hotchips.org/assets/program/conference/day1/GPU%20HPC/Intel_s%20Ponte%20Vecchio%20GPU%20-%20Architecture%20Systems%20and%20Software%20FINAL.pdf
+
+device = Device["GPU"]
+
+# NOTE: only the subgroup type 8 ones work
+prog = CLProgram(device, "test", CLCompiler(device, "test").compile(f"""
+__attribute__((reqd_sub_group_size(8)))
+__kernel void test(__global float* data0, const __global int* data1, const __global int8* data2) {{
+  int lidx0 = get_local_id(0);
+  int a = data1[lidx0/2];
```
ColonelPhantom, line 21: Is this correct? The extension expects an int here, which represents two half-floats. So [a,b].

With this code, your second subgroup will load halfs [a,b] as well, instead of [c,d] which I think is what should be loaded instead. As such you will load duplicate rows, and half the matrix will be missing. (Over the subgroup, `a` will look like `ab,ab,cd,cd,ef,ef,gh,gh` instead of `ab,cd,ef,gh,ij,kl,mn,op`.

That's probably where your 'halved K' comes from. The fix is probably just to say `int a = data1[lidx0]` instead. Although it's a bit confusing, because I think `a` in OpenCL corresponds to `b` in the Python code?

I'd test it myself, but don't have any Arc GPU to test it on unfortunately.

geohot: Nah, it's worse with the /2

Issue #3660, opened 2024-03-08
PTX is slower than CUDA


geohot:
Seeing a ratio of 1.13x. Some improvements half done in #3655, but tests are failing

You can compare speed to CUDA with `python3 test/external/speed_compare_cuda_ptx.py`

- [ ] Add UnaryOps.RECIP, TernaryOps.MULACC (only in UOps)
- [ ] Switch loads/stores to do ptr arithmetic properly
- [ ] Add support for float4
- [ ] Constant folding + reordering (#3639 but a level deeper)
- [ ] Add tensor core support to PTX

These are good things for new people. Timing LLaMA / stable diffusion between PTX and CUDA is also a good test (trigger PTX with `PTX=1` env var)

PR #3833, opened 2024-03-20
use true half for onnx

chenyuxyz:
don't remember why this did not work, is it precision?

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -28,7 +28,7 @@ def safe_numpy(t) -> np.ndarray:
 # not supported: STRING = 8 COMPLEX64 = 14, COMPLEX128 = 15
 # NOTE: 17, 18, 19, 20 are float8, 10 is half
 DTYPE_MAP = {1:dtypes.float, 2:dtypes.uint8, 3:dtypes.int8, 4:dtypes.uint16, 5:dtypes.int16, 6:dtypes.int32, 7:dtypes.int64,
-              9:dtypes.bool, 10:dtypes.float, 11:dtypes.double, 12:dtypes.uint32, 13:dtypes.uint64, 16:dtypes.bfloat16,
+              9:dtypes.bool, 10:dtypes.half, 11:dtypes.double, 12:dtypes.uint32, 13:dtypes.uint64, 16:dtypes.bfloat16,
               17:dtypes.float, 18:dtypes.float, 19:dtypes.float, 20:dtypes.float}
 # TODO: fix buffer_parse to use this and fix get_weight_and_biases to only use buffer_parse
 


chenyuxyz:
@geohotstan do you remember why this did not work?

geohotstan:
Yeah, it's the `external_model_benchmark.py` for openpilot that blocks this.
Here's the [branch](https://github.com/geohotstan/tinygrad/pull/9) that does the full fix 
So the problem is onnxruntime's CPUExecutionProvider does casting to float32 to do math and casts it back to float16 ([src](https://github.com/microsoft/onnxruntime/issues/15977#issuecomment-1604562734))
which causes precision problems when we do it all in float16. 

(then GPU should work?)
So I also tried to use google colab's GPU runtime to test the precision result of CUDAExecutionProvider on openpilot, and for some reason the result when compared to our float16 still had the precision error.... (╯°□°）╯︵ ┻━┻)

chenyuxyz:
Oh thank you, now i remember. If it's a onnxruntime cpu bug we probably just remove verification for that and use onnxruntime gpu / torch. Otherwise we are having the bug behavior too. Are you interested in cleaning this up?

geohotstan:
> Oh thank you, now i remember. If it's a onnxruntime cpu bug we probably just remove verification for that and use onnxruntime gpu / torch. Otherwise we are having the bug behavior too. Are you interested in cleaning this up?

Yeah I'll give it another go!

chenyuxyz:
does torch gpu match onnxruntime gpu? and when it does not match is it a very big diff?

geohotstan:
> does torch gpu match onnxruntime gpu? and when it does not match is it a very big diff?

I have not checked (gonna spin up google colab GPU again). Those two should be the same.
The diff between tinygrad and the others isn't too large.

I've tried decorating some of the op functions with a `.float()` for all tensor input and a `.cast(dtypes.half)` for output, and it seems like op `Gemm` and `Conv` when decorated, fixes a lot of the precision inaccuracy (though not all).

chenyuxyz:
interesting. tinygrad default uses float32 for acc even if inputs are float16. you can change the acc_dtype here to see if that's issue
`https://github.com/tinygrad/tinygrad/blob/4e0819e40bd4c0ee88eab94e3545fcbad716ce6f/tinygrad/tensor.py#L770`

geohotstan:
ok I have a hint as to to the reason why

I think what torch gpu and ORT gpu are doing is they're both aggressively fusing multiply and sum such that they're calling some _fast_ mulacc kernel underneath with the entire kernel operating in `float32` (I'm not 100% sure on this, but I went through both of their repos for quite some time and got confused haha)

So this means that it's not only `sum` that's `acc_dtype`ed but also the `multiply` as well. 
Which also explains why the decorator I used was solving a lot of the accuracy problems.

This has accuracy problems
```
n = np.random.rand(100, 100).astype(np.float16)

t = Tensor(n)
t2 = Tensor(n)
tt = torch.tensor(n).to('mps')
tt2 = torch.tensor(n).to('mps')

a = (t @ t2)
b = tt @ tt2

np.testing.assert_allclose(a.numpy(), b.cpu().numpy())
```

while just doing `sum` doesn't

I'm going to try adding back `MULACC` and force it to be `dtypes.float32`, maybe in uops as requested in #3660 if possible? :s 
again, this is just a hint. Not 100% sure if this is the root problem.

geohotstan:
I found some new hints to this issue

I think what torch gpu and ORT gpu are doing is they're both aggressively fusing `multiply and sum` such that they're calling some fast math kernel (tensorcore?) underneath (I'm not 100% sure on this. ORT calls reinterpret_cast before doing cudaBLAS. I've never dug this deep so I'm trying to learn as I go.)

------------

```
n = np.random.rand(512, 1024).astype(np.float16)

t = Tensor(n)
t2 = Tensor(n.T)
tt = torch.tensor(n).to('mps')
tt2 = torch.tensor(n.T).to('mps')

a = t @ t2
b = tt @ tt2

np.testing.assert_allclose(a.numpy(), b.cpu().numpy())
```
This passes if we use `TC` (wmma), and fails when `TC=0`

In the case that `TC` does not activate, if we cast to `float` before doing _BOTH_ the mul and the sum and then cast back to `acc_dtype` we considerably less mismatched elements than if we only do `acc_dtype` casting to the sum.
Which also explains why the decorator I used was solving a lot of the accuracy problems.

chenyuxyz:
i see, thanks for the update. this is quite annoying. in general we don't have a principled way to define the ground truth for these.

would it work if we consider a norm allclose that's based on L1/L2 (i.e. `sum(abs(xi - yi))` and `sqrt(sum((xi - yi)**2))`?

chenyuxyz:
see #3896

PR #3834, opened 2024-03-20
level zero runtime

nimlgen:


extra/level_zero.py
```diff
@@ -0,0 +1,116 @@
...
+ev_desc.wait = ze.ZE_EVENT_SCOPE_FLAG_HOST
+check(ze.zeEventCreate(event_pool, ctypes.byref(ev_desc), ctypes.byref(event := ze.ze_event_handle_t())))
+
+# signal the event from the device and wait for completion
+
+
+print("trying to create module")
+from tinygrad.runtime.ops_gpu import CLCompiler
+from tinygrad import Device
+spriv_bin = CLCompiler(Device["GPU"], "test").compile(f"""
```
nimlgen, line 59: Just reused OpenCL binary, but I think we need to compile to spirv here. Haven't found any online(rt) OpenCL->spirv compiler yet.

geohot: Hmm, OpenCL outputs an ELF in the intel GPU assembly, I wouldn't want to put in some intermediate language.

geohot:
I think we put this on hold for now in favor of further CUDA speed improvements / refactoring of the graphs.

PR #3836, opened 2024-03-20
add a better error message for resnet training

Qazalin:


extra/datasets/imagenet.py
```diff
@@ -13,10 +13,16 @@ def get_imagenet_categories():
   return {v[0]: int(k) for k,v in ci.items()}
 
 @diskcache
-def get_train_files(): return glob.glob(str(BASEDIR / "train/*/*"))
+def get_train_files():
+  files = glob.glob(p:=str(BASEDIR / "train/*/*"))
+  if len(files) == 0: raise Exception(f"no training fails found in {p}")
```
geohot, line 18: s/fails/files

geohot: Also you shouldn't raise Exception. This should probably be an assert

Issue #3820, opened 2024-03-19
HSA transfer hangs intermittently


flammit:
@chenyuxyz for addition to #3805.

The HSA transfer hang can be repro'd using the following branches.  It should hang within 10 seconds every 3-4th run or so, if not cancel, reset the GPU and try again.

https://github.com/flammit/tinygrad/tree/hang_repro with the following command : `time PYTHONPATH=. HSA=1 LATEWINO=1 LATEBEAM=64 BEAM_UPCAST_MAX=16384 BEAM_LOCAL_MAX=512 BEAM_MIN_PROGRESS=5 BEAM_LOG=1 IGNORE_BEAM_CACHE=1 STEPS=10 DEBUG=2 python3 ./examples/hlb_cifar10.py` (see discord link [here](https://discord.com/channels/1068976834382925865/1148714305060098138/1219023206426607746))

or https://github.com/nimlgen/tinygrad/tree/hang_repro_2 with this command: `PYTHONPATH=. HSA=1 LATEWINO=1 LATEBEAM=64 BEAM_UPCAST_MAX=16384 BEAM_LOCAL_MAX=512 BEAM_MIN_PROGRESS=5 IGNORE_BEAM_CACHE=1 HSA_HANG_REPRO=1 STEPS=10 DEBUG=2 python3 ./examples/hlb_cifar10.py` (see discord link [here](https://discord.com/channels/1068976834382925865/1069001075828469790/1219664229062152324))

flammit:
Closing this out.  I rebased to master and it's not reliably hanging anymore.

PR #3847, opened 2024-03-21
fix hsa sync issue

nimlgen:


chenyuxyz:
could this in theory have fixed #3820? first time in a while i can train 6 gpus resnet for more than 3 epochs

PR #3849, opened 2024-03-21
poc reduce over expanded axis

chenyuxyz:
i mean it's faster for `Tensor.ones(2000, 2000, 2000).sum().numpy()`

probably also need the one for elementwise op, if two inputs have common expanded axis, can skip the compute and propagate expand to output


chenyuxyz:
not very useful

PR #3851, opened 2024-03-21
don't include hip common

geohot:
You only need to install "hsa-rocr comgr" (on a real system, remu needs more? @Qazalin)

We can make HIP CI a lot faster. We also need to replace <hip/amd_detail/amd_hip_bfloat16.h>

* hsa-rocr -- the HSA runtime, can be replaced with direct kernel calls
* hsa-rocr-dev -- the header file for autogen for runtime
* comgr -- the compiler
* rocm-llvm -- the disassembler / for remu


Qazalin:
remu only needs comgr and rocm-llvm to disassemble.

Qazalin:
i think the bf16 removal is gonna be a prereq for test_dtype to pass

geohot:
Yea, tests failing. Have to move bfloat16 stuff into tinygrad.

PR #3852, opened 2024-03-21
Ring allreduce try 2

uuuvn:


chenyuxyz:
Tested hlb_cifar for 30 rounds with different seeds. The lowest acc is 93.32 with similar average number. I think it's well within the noisy variance.

PR #3856, opened 2024-03-21
proposal: multioutput JIT spec

Qazalin:
I think JIT should use Tensor.corealize to let the scheduler fuse outputs, the user can .realize() to opt-out of the fusion.

will enable the tests once multioutput is merged.


geohot:
I'm supportive of this, it always should have been corealize in the JIT

PR #3862, opened 2024-03-21
search: change to use "spawn" and limit the number of tasks per child

flammit:
also clean up some examples to use __main__ and not initialize resources outside of main

This limits the amount of memory leaked in the child processes.  For very large kernels, you might need to limit the tasks per child to even less by setting `BEAM_MAX_TASKS_PER_CHILD` to a lower number.

For this run `PYTHONPATH=. HSA=1 GPUS=6 LATEWINO=1 LATEBEAM=64 BEAM_UPCAST_MAX=16384 BEAM_LOCAL_MAX=512 BEAM_MIN_PROGRESS=5 BEAM_MAX_TASKS_PER_CHILD=32 STEPS=10 DEBUG=2 HALF=1 BS=1536 python3 ./examples/hlb_cifar10.py`, the parent process still leaked up to 30GB by the end of a 4.5hr run, but this is not addressed here.  The total memory usage peaked at 80GB during one late WINO kernel, so in hindsight, I should've run with lower max tasks.

Tested in HSA and CUDA.  Trying to do beam inside of the rocm/pytorch docker container didn't seem to work (hangs) but that might be due to other issues.

### DIFF
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -19,7 +19,6 @@ EVAL_BS = getenv("EVAL_BS", BS)
 GPUS = [f'{Device.DEFAULT}:{i}' for i in range(getenv("GPUS", 1))]
 assert BS % len(GPUS) == 0, f"{BS=} is not a multiple of {len(GPUS)=}, uneven multi GPU is slow"
 assert EVAL_BS % len(GPUS) == 0, f"{EVAL_BS=} is not a multiple of {len(GPUS)=}, uneven multi GPU is slow"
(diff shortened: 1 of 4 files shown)


chenyuxyz:
was searching why spawn is better than fork and found this https://github.com/python/cpython/issues/84559. is cpython fork just broken?

anecdotally searching cifar on my 3090 machine killed the machine on master, and i can finished the search with this pr

geohot:
Oh sweet, yea spawn is exactly how to fix the weird driver behavior!

PR #3864, opened 2024-03-21
diverse test value in test_dtype DATA based on dtype

chenyuxyz:
added negative int for int case

TODO: test this on real GPU. test passed on tinybox

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -3,7 +3,7 @@ from tinygrad import Tensor, Device, dtypes
 from tinygrad.device import JITRunner
 from tinygrad.dtype import DType
 from tinygrad.nn.state import get_parameters
-from tinygrad.helpers import Context, CI, OSX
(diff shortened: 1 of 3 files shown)


Qazalin:
I think there might be some edgecases i didnt cover in the instruction implementations, thanks for catching these I'll fix them

chenyuxyz:
some of these might be undefined behavior (there are some skipped tests, i tested before https://github.com/chenyuxyz/tinygrad/pull/159). Feel free to skip the ones that you are sure it's undefined

Qazalin:
@chenyuxyz fyi shipped int16 sign extension fixes for remu - should be good now. thanks again for helping catch this.

chenyuxyz:
added annotation for PTX. it looks like PTX does not support int8, uint8 buffer and we are using int16 and uint16 to represent it now, which failed in cast.

PR #3867, opened 2024-03-22
disable TC in action with amt=1

chenyuxyz:
open for investigation. `BEAM=2 python examples/handcode_resnet50_opt.py` on METAL M1 Max
1. master: 297ms
2. removed all TC: 206ms
3. keep only TC axis=0: 186ms

### DIFF
--- a/tinygrad/features/search.py
+++ b/tinygrad/features/search.py
@@ -18,7 +18,7 @@ actions += [Opt(op=OptOps.GROUPTOP, axis=axis, amt=amt) for amt in [13,16,29,32,
 actions += [Opt(op=OptOps.GROUP, axis=axis, amt=amt) for amt in [0,4,8,16] for axis in range(3)]
 actions += [Opt(op=OptOps.PADTO, axis=axis, amt=amt) for amt in [32] for axis in range(7)]
 actions += [Opt(op=OptOps.LOCAL, axis=0, amt=32), Opt(op=OptOps.UPCASTMID, axis=1, amt=4), Opt(op=OptOps.TC, axis=0, amt=0)]
-actions += [Opt(op=OptOps.TC, axis=axis, amt=1) for axis in range(4)]
+# actions += [Opt(op=OptOps.TC, axis=axis, amt=1) for axis in range(4)]
 if getenv("NOLOCALS"): actions += [Opt(op=OptOps.NOLOCALS)]
 
 def _get_test_global_size(global_size, max_global_size, var_vals):


flammit:
I definitely see a very large performance diff on M2 Max using that `BEAM=2 python examples/handcode_resnet50_opt.py`:

1. master: 177ms
2. removed all TC: 137ms
3. keep only TC axis=0 amt=0: 128ms

FWIW, on AMD/tiny10 using `BEAM=64 HALF=1 python3 ./examples/handcode_resnet50_opt.py`:

1. master: 70.69 ms
2. removed all TC: 71.19 ms
3. keep only TC axis=0 amt=0: 71.64 ms
4. (for a comparision to above) master w/ BEAM=2: 94.8ms

The reason I including it in the search space was that it really seemed to help non-WINO convs.

On METAL M2, we can get to 9.7TF with TC AMT=1 and 5.4TF without it:

metal_m2_conv_w_tc_amt1

metal_m2_conv_wo_tc_amt1


On AMD tiny10, we can get to 31TF with TC AMT=1 and 8TF without it:


amd_conv_w_tc_amt1

amd_conv_wo_tc_amt1


In multi-reduce situations, the TC amt=1 gets fast with a single action which crowds out the eventual fastest kernel requires 3-4 actions to get to comparable speeds:

Screenshot 2024-03-22 at 1 29 17 PM
Screenshot 2024-03-22 at 1 31 26 PM


Since it really does impact that search with low BEAM, until there's a better search perhaps we can gate the addition of the TC amt=1 to the action space behind its own flag like "NOLOCALS" is?

chenyuxyz:
It did not change handcode_resnet50_opt.py for AMD because it's using float32 and only METAL has float32 TC?

The main reason I don't like flag is that flag does not go into cache key and is annoying to track. Maybe we just need a better search

flammit:
> It did not change handcode_resnet50_opt.py for AMD because it's using float32 and only METAL has float32 TC?
> 
> The main reason I don't like flag is that flag does not go into cache key and is annoying to track. Maybe we just need a better search

I think `handcode_resnet50_opt.py` has the HALF option implemented properly:

Screenshot 2024-03-22 at 3 59 20 PM


Agreed, the flags aren't a great solution especially because of the cache key.  We do need a better search :/

chenyuxyz:
i see. my number was with float32, were you using half?

flammit:
> i see. my number was with float32, were you using half?

I used FP16 on AMD so that the TCs were available in the action space.  Used FP32 for METAL on my M2 Max so it would be the same comparison as you did on M1 Max.

As an aside, there should probably also be "TC, axis=range(4), amt=0" instead of just "axis=0, amt=0".  It was an oversight on my part that they weren't there (to be able to test the performance of the choice of the TC axis when the option exists).

chenyuxyz:
might not be relevant after padto tc and generally will be fixed with a better search

PR #3868, opened 2024-03-22
tiny version of amd_hip_bfloat16

Qazalin:
Replaces [amd_hip_bfloat16](https://github.com/ROCm/clr/blob/develop/hipamd/include/hip/amd_detail/amd_hip_bfloat16.h) - something like this is a prereq for #3851


geohot:
Hmm, any reason not to write in the same style as the AMD one?

geohot:
Nice fix, this good to merge?

PR #3871, opened 2024-03-22
Rust Code "accelerator" - Part 1 - Adding size of global buffers in Uops graph

jaredeh:
This is the first patch for adding a Rust Code "accelerator".  As expected the requirements and layout is fairly similar CLANG, however there is a major difference.  The Rust compiler needs to know how big arrays are.

When called from a function like this, the compiler will handle sizing the parameters data0..data3 for the function r_4_32()
```
fn r_4_32(data0: &mut [f32], data1: &[f32], data2: &[f32], data3: &[f32]) {
    data0[0] = data1[0] * data2[0] + data1[1] * data3[1]
}

pub extern "C" fn net(input0: &[f32; 32], output0: &mut [f32; 4]) {
   let buf1 = [0.0; 64];
   let buf2 = [0.0; 128];
   r_4_32(output0, input0, &buf1, &buf2);
}
```
However when using FFI to connect to python we need to specify the size like this:
```
pub extern "C" fn r_4_32(data0: &mut [f32; 4], data1: &[f32; 32], data2: &[f32; 64], data3: &[f32; 128]) {
    data0[0] = data1[0] * data2[0] + data1[1] * data3[1]
}
```
To enable this sizing feature I made the following changes:
* Add size field to args of DEFINE_GLOBAL node in the UOps graph for Rust to size input/output arrays
* Adapt dedup key to ignore size, apparently we can get redundant buffers in the graph.
* Fixed test

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -45,7 +45,7 @@ class TestLinearizer(unittest.TestCase):
     lin.linearize()
 
     stores = [u for u in lin.uops if u.uop is UOps.STORE]
-    mutable_bufs = [u for u in lin.uops if u.uop is UOps.DEFINE_GLOBAL and u.arg[-1]]
+    mutable_bufs = [u for u in lin.uops if u.uop is UOps.DEFINE_GLOBAL and u.arg[-2]]
     assert len(mutable_bufs) == len(stores) == 2
     assert [u.arg[0] for u in mutable_bufs] == [0, 1]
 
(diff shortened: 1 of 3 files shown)


geohot:
Hmm, so it's not needed now. I'm not going to merge this until it's actually needed by a backend, and I don't think we want a rust backend, what advantage does it have over clang?

jaredeh:
I mean... "I" want a rust backend, obviously.

My use case is that I'm building and deploying simple FNNs.  I take a model (ONNX for now) and use tinygrad to render source that I then import to build into my app.  I see a lot places in the embedded world, desktop apps, servers, and even web stuff (webGL/WASM) where using this code export capability to embed AI is an amazing feature unique to tinygrad.

Specifically why do I want a Rust backend?  The app I want to embed tinygrad output into happens to be written in Rust.  I am importing C code into my Rust app today.  But doing that when unnecessary makes for uglier, bad code with unnecessary dependencies.

jaredeh:
An RFC level Rust backend patch is ~1 day out.

Would you prefer I move it to extra/ for line count and support purposes? I see extra/backends though that all looks dead.

PR #3872, opened 2024-03-22
sum's of strings moved to f-strings

igeni:
concatenation of strings moved to f-strings and removed redundancy of type's conversion


geohot:
whitespace changes will be closed unless you are a well-known contributor

PR #3875, opened 2024-03-22
preschedule all

geohot:


Qazalin:
How can we create multioutput ASTs if all realizes are prescheduled?
Multioutput impacts the ScheduleItem's inputs.

geohot:
Hmm, so it should precompute what all the kernels will be before scheduling, I don't see why this impacts multioutput.

I need this as a prereq to fixing assign.

PR #3880, opened 2024-03-22
Remove requires_grad check for replace and assign

### LINKED ISSUE #3879: `assert not x.requires_grad` in Tensor.replace()?
Any reason why requires_grad needs to be False for replacing a tensor?
https://github.com/tinygrad/tinygrad/blob/master/tinygrad/tensor.py#L143

Here is my use case which doesn't work. It's quite common in RL to copy weights to target networks.
```
from tinygrad import nn

class Network:
    def __init__(self):
        self.test = nn.Linear(10, 1)

net = Network()
target_net = Network()
tau = 0.9

for param, target_param in zip(nn.state.get_parameters(net), nn.state.get_parameters(target_net)):
    target_param.replace(tau * param + (1 - tau) * target_param)
```

eugeneteoh:
fix #3879

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -340,6 +340,18 @@ class TestTinygrad(unittest.TestCase):
     c = (a + b).mean().backward()
     print(c)
 
+  def test_replace_and_assign(self):
+    x = Tensor([1,2,3], requires_grad=True)
+    y = Tensor([4,5,6])
+    y.replace(x)
+    assert not y.requires_grad
+
+    z = Tensor([7,8,9])
+    z.assign(x)
+    assert not z.requires_grad
+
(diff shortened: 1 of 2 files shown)


geohot:
Can you add tests confirming that this is okay to do, and things behave as expected?

eugeneteoh:
Added some simple tests. Let me know if you have better ideas as I can't think of any now.

Also, I don't fully understand what .assign() does, how it's different from .replace(), and why they have to be separate.

geohot:
No! I mean test that gradient behaves as expected.

eugeneteoh:
replace/assign seems to work differently from torch.Tensor.copy_ (see my test). All the optimiser tests are failing when requires_grad is set, as they use assign. Should replace/assign be separate?

chenyuxyz:
closing as stale

PR #3883, opened 2024-03-22
debug: add optional detailed BEAM_LOG logging

flammit:
show uop count, compile and run times for each candidate in search

also add --timing to verify_kernel.py to make it easier to explore hand-crafted applied opts

You can see the long tail of compile times for some of the WINO kernel:

Screenshot 2024-03-22 at 3 39 01 PM
Screenshot 2024-03-22 at 3 33 23 PM
Screenshot 2024-03-22 at 3 46 16 PM


chenyuxyz:
i think it's fine, don't quite see how to do it differently. but overall i think search needs some cleanups

PR #3884, opened 2024-03-22
include negative float in test_dtype

chenyuxyz:
except casting negative float to unsigned integer which is undefined


chenyuxyz:
@Qazalin fyi this passes on amd tinybox too, might be similar issue?

Qazalin:
these kernels needed new instructions! shipped the latest binary with `v_max`&`v_min` for f16 and f64. I think the compiler uses these for `a.abs()`

chenyuxyz:
learned two cases:
1. converting negative float to unsigned integer is undefined, this we just take abs
2. struct.pack (how we do cast in PYTHON) does not work when packing a bigger than half number into half (does not convert to inf). skipped this case for now

PR #3891, opened 2024-03-23
Add negative_log_likelihood loss and cross_entropy loss 

airpods69:
This PR adds `negative_log_likelihood` and `cross_entropy` to `Tensor`

Eg: 
For `negative_log_likelihood`,
```py
import tinygrad
data = tinygrad.Tensor([[1, 2, 4]])
target = [2]
print(data.negative_log_likelihood(target).numpy())
``` 

==ToDo==
1. Still need to implement this `cross_entropy`. 

==Something which confused me==
1. As mentioned in #1018 `negative_log_likelihood`(referring to it as `nll`) and `cross_entropy` operate very similarly, `nll` works with log probabilities while `cross_entropy` works with logits. What if I keep them in the same function with an if else and 2 more functions which act as abstractions? I'll implement `cross_entropy` first and keep this as a task for future me.


airpods69:
The force push to get some of that rebase errors corrected (had those on local)
On a second thought, I could have fixed the lint unit test too...

airpods69:
to use `cross_entropy`,
You just have to call cross_entropy from Tensor and should be all good.
```
print(data.cross_entropy(target).numpy())
```

test/test_ops.py
```diff
@@ -1665,6 +1665,24 @@ def test_one_hot(self):
     helper_test_op([], lambda: torch.nn.functional.one_hot(torch.tensor(data), 8).type(torch.int32),
                        lambda: Tensor(data).one_hot(8), forward_only=True)
 
+  def test_negative_log_likelihood(self):
+    data, target = torch.randn(1,3), [2]
+    helper_test_op([], lambda: torch.nn.functional.nll_loss(torch.nn.functional.log_softmax(data, dim = 1), torch.tensor(target)).type(torch.float32),
```
geohot, line 1670: Can you use helper_test_op like the other examples?

tinygrad/tensor.py
```diff
@@ -987,6 +987,16 @@ def dropout(self, p=0.5) -> Tensor:
   def one_hot(self, num_classes:int) -> Tensor:
     return (self[..., None] == Tensor.arange(num_classes, requires_grad=False, device=self.device)).where(1, 0)
 
+  # nll --> Negative Log Likelihood (linter was throwing in errors...)
+  def nll(self, target:Tensor, weight:Optional[Tensor] = None, reduction:Literal['none', 'mean', 'sum'] = 'mean') -> Tensor:
```
geohot, line 991: Does this match the torch API perfectly? Do you have tests for all the code paths?

airpods69: except for ignore_indexes parameter and the function name, it should match the API perfectly. A mistake on my end that I didn't test if for all code paths (to confirm, do you mean all the different parameters or something else?)

function name should be nll_loss so I'll switch it to that

test/test_ops.py
airpods69: Done with this, weights can be either None or a 1D tensor and helper_test_op would throw an error cause of that? correct me if im wrong.

airpods69:
I think I'll need some help with the checks failing...

unknownusername504:
> I think I'll need some help with the checks failing...

You are generating random float weights and Tensor is inferring that type. While you are doing type casts to long/int64 for the tests of torch side and the test tells you it fails for mismatching dtypes (float for tg and int for torch)

airpods69:
> > I think I'll need some help with the checks failing...
> 
> You are generating random float weights and Tensor is inferring that type. While you are doing type casts to long/int64 for the tests of torch side and the test tells you it fails for mismatching dtypes (float for tg and int for torch)

Worked on this, removing the type casting did the trick but seems like metal, hip and ptx are still failing. 
1. Metal errors out because `double` isn't supported on it which I believe can be fixed by managing the types.
2. hip on the other hand is doing some interesting stuff... 
```
tinygrad_output = array(-2.1397502)
torch_output = array(2.1397502), atol = 1e-06, rtol = 0.001
```
Tests on my intel cpu and cuda work fine and give the same result but on hip, the results are different? makes no sense? does hip have some different instruction causing the additive inverse??

airpods69:
`Unit Tests / GPU OPT and IMAGE Tests (pull_request) ` fails cause of a type mismatch but it looks like torch has a torch.tensor of `float64` no problem but `dtypes.default_float.np` from `prepare_test_op` generates a tinygrad.Tensor with `float32` which causes the assertion error

Idk I could be a dumbass but I can't seem to find a way around this.

airpods69:
3 days and I still can't figure out the broken test. torch has no problem creating `float64` tensors but the architectures don't support it which makes the tests fail and I can't figure out a way to make it not do it
Any insights on this would be helpful!

chenyuxyz:
close as stale

airpods69:
@chenyuxyz you would think so... gonna fix those test cases cause I have no clue why it was failing on the AMD side of things. 
Had to save up a bit, and now 7700XT has been acquired. Any leads would be great but I shall get to this again

PR #3894, opened 2024-03-23
Tensor core ptx

SzymonOzog:


tinygrad/renderer/assembly.py
```diff
@@ -216,6 +216,14 @@ def cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
         if lang.load_global:
           dt = dtypes.ulong if dtype.__class__ == PtrDType else dtype
           kk(*lang.render_load(args[1], ssa(u, 'dat', dtype=lang.types[dt]), dt, ss=".param"))
+      elif uop is UOps.WMMA:
+        wmma=[]
+        for v in vin[:2]:
+          for i in range(0, len(r[v]), 2): # type: ignore
```
chenyuxyz, line 222: why `type: ignore`?

chenyuxyz (commented):
can you add this to benchmark? next to https://github.com/tinygrad/tinygrad/blob/4e18dd78d32eb65cce3c81822da3672f970e0fd8/.github/workflows/benchmark.yml#L101

tinygrad/renderer/assembly.py
SzymonOzog: Not sure why but mypy thinks v is of type string and throws `error: Invalid index type "str" for "dict[UOp, list[str] | str]"; expected type "UOp"`

SzymonOzog:
> can you add this to benchmark? next to
> 
> https://github.com/tinygrad/tinygrad/blob/4e18dd78d32eb65cce3c81822da3672f970e0fd8/.github/workflows/benchmark.yml#L101

Added

tinygrad/renderer/assembly.py
geohot: This has to be understood and fixed before we can merge this.

SzymonOzog: hmm, seems like changing the var name form `v` to `vv` fixes this, not sure why `v` breaks mypy

tinygrad/codegen/uops.py
```diff
@@ -383,6 +383,6 @@ def flops_mem(self) -> Tuple[sint, sint]:
       elif u.uop is UOps.WMMA:
         if u.arg.startswith("__metal_wmma"): flops += 2*(8*8*8)//32 * mults
         elif u.arg == "__hip_wmma_f16_f16" or u.arg == "__builtin_amdgcn_wmma_f32_16x16x16_f16_w32": flops += 2*(16*16*16)//32 * mults
-        elif u.arg == "__cuda_mma_m16n8k16_f16_f32": flops += 2*(8*16*16)//32 * mults
+        elif "m16n8k16"in u.arg: flops += 2*(8*16*16)//32 * mults
```
geohot, line 386: spacing

tinygrad/renderer/assembly.py
```diff
@@ -216,6 +216,14 @@ def cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
         if lang.load_global:
           dt = dtypes.ulong if dtype.__class__ == PtrDType else dtype
           kk(*lang.render_load(args[1], ssa(u, 'dat', dtype=lang.types[dt]), dt, ss=".param"))
+      elif uop is UOps.WMMA:
+        wmma=[]
+        for vv in vin[:2]:
+          for i in range(0, len(r[vv]), 2):
+            wmma.append(ssa(None, "wmma", "b32"))
+            kk(f'mov.b32 {wmma[-1]}, {{{", ".join(r[vv][i:i+2])}}};')
```
geohot, line 224: Hmm, does there have to be an extra mov here, or is there a way plan this earlier and have the data already be in the right place?

SzymonOzog: Not reliably afaik, what this code actually boils down into is this. 
```
     ld.global.b16   %val_f16_0, [%alu_u64_0+0];
     ...
     ld.global.b16   %val_f16_1, [%alu_u64_1+0];
     ...
     mov.b32         %wmma_b32_0, {%val_f16_0, %val_f16_1};
```
theoretically it is possible to load directly like this 
`     ld.global.b16   %wmma_b32_0, [%alu_u64_0+0];`
But in practice `%val_f16_0` and `%val_f16_1` aren't always contiguous and we fail with bad memory access

tinygrad/codegen/kernel.py
```diff
@@ -51,7 +51,7 @@ def fix_axes(self, removed_axis:int): # adjust the TC axes if necesssary when an
...
   "HSA": [TensorCore(dims=(16,16,16), threads=[(0,8),(0,2),(1,2)], thread_local_sizes=[[16],[16],[4,2]], thread_local_aliases=[ [[2],[0],[0],[-1],[1]], [[0],[2],[1],[-1],[0]], [[-2],[2],[1],[0],[3,-1]] ], dtype_in=di, dtype_out=do) for (di, do) in [(dtypes.half, dtypes.float), (dtypes.half, dtypes.half)]],  # noqa: E501
-  "CUDA": [TensorCore(dims=(8,16,16), threads=[(0,2),(0,2),(1,2),(1,2),(0,2)], thread_local_sizes=[[2,2,2],[2,2],[2,2]], thread_local_aliases=[ [[0],[-2],[5],[0],[0],[-1,1,2,-3],[3,4]], [[5],[0],[0],[4],[3],[-1,1,2,-2],[0]], [[2],[-2],[5],[1],[-1],[0],[3,4]] ], dtype_in=di, dtype_out=do) for (di, do) in [(dtypes.half, dtypes.float), (dtypes.bfloat16, dtypes.float)]],  # noqa: E501
+  "CUDA": [TensorCore(dims=(8,16,16), threads=[(0,2),(0,2),(1,2),(1,2),(0,2)], thread_local_sizes=[[2,2,2],[2,2],[2,2]], thread_local_aliases=[ [[0],[-2],[5],[0],[0],[-1,1,2,-3],[3,4]], [[5],[0],[0],[4],[3],[-1,1,2,-2],[0]], [[2],[-2],[5],[1],[-1],[0],[3,4]] ], dtype_in=di, dtype_out=do) for (di, do) in ([(dtypes.half, dtypes.float)] if getenv("PTX") else [(dtypes.half, dtypes.float), (dtypes.bfloat16, dtypes.float)])],  # noqa: E501
```
geohot, line 54: Why is this required? Can bfloat16 work in PTX?

tinygrad/renderer/assembly.py
```diff
@@ -206,6 +206,15 @@ def cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
         if lang.load_global:
           dt = dtypes.ulong if dtype.__class__ == PtrDType else dtype
           kk(*lang.render_load(args[1], ssa(u, 'dat', dtype=lang.types[dt]), dt, ss=".param"))
+      elif uop is UOps.WMMA:
+        wmma=[]
```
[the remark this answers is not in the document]
SzymonOzog: updated

tinygrad/codegen/kernel.py
SzymonOzog: Tried adding bf16 but it required sm_90 and my gpu does not support it

PR #3895, opened 2024-03-23
Fix the result permutation in einsum

afqueiruga:
The previous einsum code was broken when some of the result indices where not in the alphabetical order.


chenyuxyz:
cool! can you fix the linter error?

afqueiruga:
Fixed it. I didn't realize the linter the wasn't run along with the other tests.

chenyuxyz:
thanks!

PR #3896, opened 2024-03-23
onnx float16 compatibility

geohotstan:


extra/onnx_ops.py
```diff
@@ -601,15 +609,18 @@ def AffineGrid(theta: Tensor, size: Tensor, align_corners=0):
...
+  if epsilon is None: epsilon = 1e-12 if x.dtype == dtypes.float else 1e-5 # Tensor can only be float or half
   x = x + skip + bias
   return x.layernorm(eps=epsilon) * gamma + beta, None, None, x
 
+@half_to_float_to_half
 def FastGelu(x:Tensor, bias:Optional[Tensor]=None):
   x = x + bias
   return 0.5 * x * (1 + (x * 0.797885 + 0.035677 * x ** 3).tanh())
 
+@half_to_float_to_half # l2 norm for commavq fails without this
```
geohotstan, line 623: it turns out `commavq` also has some shocking float16 weirdness
the inaccuracy here is large.. (like 20.0 to 0.15 kind of large)
My guess is some op is making these float16 numbers go under the threshold, and it ends up as 0.0

need to investigate.

chenyuxyz: i think that's where we need to update onnx implementation. Normalization in half in brittle so we usually cast training pipeline to float before batchnorm for example

.github/workflows/benchmark.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     #- name: Run Stable Diffusion
     #  run: python3 examples/stable_diffusion.py --seed 0 --noshow --timing | tee sd.txt
     - name: Run model inference benchmark
-      run: METAL=1 python3 test/external/external_model_benchmark.py
+      run: METAL=1 NOCLANG=1 python3 test/external/external_model_benchmark.py
```
geohotstan, line 34: Before the benchmark tests didn't actually perform device specific validation tests. 
`[Device.DEFAULT, "CLANG"]` made it so the last `Device.DEFAULT` was set to `CLANG` and validation was done to `CLANG` vs `CPUExecutionProvider`

test/external/external_model_benchmark.py
```diff
@@ -103,29 +98,44 @@ def benchmark_model(m, devices, validate_outs=False):
     del ort_sess
 
   if validate_outs:
-    rtol, atol = 2e-3, 2e-3  # tolerance for fp16 models
-    inputs = {k:Tensor(inp) for k,inp in np_inputs.items()}
-    tinygrad_model = get_run_onnx(onnx_model)
-    tinygrad_out = tinygrad_model(inputs)
+    if any(t == np.dtype(np.float16) for t in input_types.values()): rtol, atol = 3e-3, 1e-3  # tolerance for fp16 models
```
geohotstan, line 101: raised rtol for `half` ... 
otherwise I would have to set `half_to_float_to_half` to non normalization ops.
I tried a lot of `float()` then `cast(half)` casting to find where the inaccuracy was originating from and it was pretty difficult.
I ended up just wrapping the normalization with cast to copy how cifar casts.

test/external/external_model_benchmark.py
```diff
@@ -103,29 +98,44 @@ def benchmark_model(m, devices, validate_outs=False):
...
+    for device in devices:
+      Device.DEFAULT = device
+      inputs = {k:Tensor(inp) for k,inp in np_inputs.items()}
+      tinygrad_model = get_run_onnx(onnx_model)
+      tinygrad_out = tinygrad_model(inputs)
 
-    ort_sess = ort.InferenceSession(str(fn), ort_options, ["CPUExecutionProvider"])
-    onnx_out = ort_sess.run(output_names, np_inputs)
-    onnx_out = dict([*[(name,x) for name, x in zip(output_names, onnx_out)]])
+      provider = "CPUExecutionProvider"
```
geohotstan, line 110: both CLANG and GPU are compared with CPUExecutionProvider with slightly larger rtol.
CPUExecutionProvider does everything in float32 so it's a good source of truth.

geohotstan:
Currently validation all passes locally for me (on METAL)
I tried running it with CUDA on a colab notebook but it fails on CLANG validation... (spent quite awhile trying to figure out why but no luck.)
probably not CUDA's fault.

A little unsure the validation will pass for CI...

test/external/external_model_benchmark.py
chenyuxyz: this is fine it think

test/external/external_model_benchmark.py
```diff
@@ -13,18 +13,13 @@
...
-
-  # broken in torch MPS
-  # "zfnet": "https://github.com/onnx/models/raw/main/archive/vision/classification/zfnet-512/model/zfnet512-9.onnx",
-  # TypeError: BatchNormalization() got an unexpected keyword argument 'is_test'
-  # "densenet": "https://github.com/onnx/models/raw/main/archive/vision/classification/densenet-121/model/densenet-3.onnx",
-  # AssertionError: only onnx version >= 10 supported for slice
-  # "bert": "https://github.com/onnx/models/raw/main/archive/text/machine_comprehension/bert-squad/model/bertsquad-8.onnx",
-  # really slow
-  # "resnet18": "https://github.com/onnx/models/raw/main/archive/vision/classification/resnet/model/resnet18-v2-7.onnx",
+  "densenet": "https://github.com/onnx/models/raw/main/Computer_Vision/densenet121_Opset18_timm/densenet121_Opset18.onnx",
```
chenyuxyz, line 19: does these work now? is it because some fix in this pr or if it always work?

.github/workflows/benchmark.yml
chenyuxyz: cool, can we have a separate pr that just fixes this?

test/external/external_model_benchmark.py
geohotstan: these worked for awhile now, like for `bert`, `slice` support for onnx version < 10 was implemented quite some time back, it's just that this wasn't updated.

Also some of the links were dead links. (url changed)

I guess I'll submit this change as a separate PR

chenyuxyz:
I see many type improvements and debug message in this one. Can we get a pr that fixed device specific checking in first? Separating behavior change v.s. test improvement would make this one easier to review. Fine to add type change and debug message change in that too.

chenyuxyz:
i'd merge the test fix first then i will review this one. ci benchmark failed too

chenyuxyz:
it's weird that it's a dtype error not precision error, maybe it's different torch version? you can sync to master to remove the flaky mac test

test/external/external_model_benchmark.py
```diff
@@ -122,12 +124,22 @@ def benchmark_model(m, devices, validate_outs=False):
...
-    else: np.testing.assert_allclose(tiny_v.numpy(), onnx_v, rtol=rtol, atol=atol, err_msg=f"For tensor '{k}' in {tiny_out.keys()}")
+    else:
+      assert (tiny_v := tiny_v.numpy()).dtype == onnx_v.dtype, f"tiny={tiny_v.dtype} onnx={onnx_v.dtype}"
+      try:
+        np.testing.assert_allclose(tiny_v, onnx_v, rtol=rtol, atol=atol,
+                                   err_msg=f"device={Device.DEFAULT} using {rtol=} {atol=} for tensor '{k}' in {tiny_out.keys()}")
+      except AssertionError:
+        # test using l1 norm and l2 norm instead
+        diff = tiny_v - onnx_v
+        assert (l1_norm := np.sum(np.abs(diff)) <= atol + rtol * np.sum(np.abs(onnx_v))) \
```
chenyuxyz, line 140: numpy.linalg.norm?

chenyuxyz:
can we get the `buffer_parse` cleanup merged? i think it's quite nice

geohot:
Want to update this? I think there's good stuff in here

Issue #3853, opened 2024-03-21
Vulkan backend


eugeneteoh:
Given that a (good) PR is raised, will a Vulkan backend be merged?

sojohnnysaid:
live long and prosper

sojohnnysaid:
> live long and prosper

[...]

PR #3898, opened 2024-03-23
Vulkan backend via Kompute

iceychris:
Possible starting point for #3853.

- [ ] Get https://github.com/KomputeProject/kompute/pull/367 merged 
- [ ] Fix broken `bool` dtype, seems to require 32 bits in GLSL
- [ ] Devices and synchronization

```bash
$ DEBUG=4 KOMPUTE=1 GPU=1 PYTHONPATH=. pytest -n auto -s -v test/test_ops.py::TestOps
=== 9 failed, 235 passed, 4 skipped, 6 warnings in 20.81s ===
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,5 @@ quickstart.py
 .hypothesis
 weights
 *.lprof
+tmp_kp_shader.comp
+tmp_kp_shader.comp.spv
--- a/.gitignore
+++ b/.gitignore
@@ -50,5 +50,3 @@ quickstart.py
 .hypothesis
 weights
 *.lprof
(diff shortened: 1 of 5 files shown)


chenyuxyz:
closing. not likely to upstream this in core tinygrad

stalkerg:
@chenyuxyz why? Can you explain a little bit more? Do tinygrad have a "plugin" system to load a such things outside?

chenyuxyz:
We have limited resource, so we cannot support backend (which includes testing, maintenance and tailor to specific dtype or hacks) that's not on our roadmap. We removed webgpu for the same reason.

PR #3899, opened 2024-03-23
add --minimal flag to nvrtc

sekstini:
Enables the new `--minimal` option added to NVRTC in CUDA 12.4, as mentioned in 
image

---

Test program:
```python
from time import perf_counter_ns as pc
from tinygrad import Tensor

z = Tensor.ones(1).relu()
st = pc()
z.realize()
et = pc()
print(f"Time taken: {et-st:>12,} ns")
print(z.numpy())
```

Before this PR with CUDA 12.4:
```bash
fluffy@mist in tinygrad :: CACHELEVEL=0 DISABLE_COMPILER_CACHE=1 python hello_tiny.py
Time taken:   21,606,017 ns
[1.]
```

With this PR and CUDA 12.4:
```bash
fluffy@mist in tinygrad :: CACHELEVEL=0 DISABLE_COMPILER_CACHE=1 python hello_tiny.py
Time taken:   17,667,016 ns
[1.]
```

### DIFF
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -66,12 +66,14 @@ class CUDACompiler(Compiler):
   def __init__(self, arch:str):
     self.arch = arch
     CUDACompiler.linearizer_opts = CUDACompiler.linearizer_opts._replace(has_tensor_cores=int(arch[3:]) >= 80)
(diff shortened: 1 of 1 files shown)


sekstini:
image

Someone with more context should probably double check that these aren't / won't be needed before merging.

PR #3906, opened 2024-03-24
Fix linearizer failure 26 test

### LINKED ISSUE #3839: UOpGraph.simplify_phi_loop still has issues with a GPT2 kernel
@patosai please take a look `test_failure_26` in `test/test_linearizer_failures.py`: https://github.com/tinygrad/tinygrad/blob/455f7bea9b675df49ef8d158a98b68e5eb0e9fad/test/test_linearizer_failures.py#L196.  

It might be related to the same issue as before as this kernel also had the same issue resolved the `test_failure_25`.  This could potentially still cause issues with flaky GPT2 beam tests.

cc: @chenyuxyz

patosai:
Closes #3839. #3530 handled summations of WHEREs, but not where consts were added to the WHEREs. Example:

$$
\sum_i^j 3 + f(i) = 3(j-i+1) + \text{clamp(...)}
$$

but the const 3 multiplication by the loop was not implemented - no tests failed until now.


tinygrad/codegen/uops.py
```diff
@@ -287,6 +288,11 @@ def uop_alu_idx(a: UOp, b, op, dtype=dtypes.int32):
         final_op = self.add(UOps.ALU, where.dtype, (maybe_cast, where.vin[1]), BinaryOps.MUL)
         self.uops = self.uops + after_split_ops
         self.replace_op(where, final_op)
+        for op in self.get_recursive_children(final_op):
+          if (op.arg is BinaryOps.ADD and any([u.uop is UOps.PHI for u in self.get_recursive_children(op)]) and op not in adjusted_adds
```
chenyuxyz, line 292: this feels very specific, is it the most generic form for the case?

patosai: here I'm attempting to get any add op between the WHERE and corresponding PHI, I haven't figured out a beautiful way to do it

patosai:
Code updated - I also added a check that loop vins are consts which I had assumed was true before. There was an attention module which had a var as part of the loop start so it failed the assumption - maybe an opportunity for more performance improvement later

https://github.com/patosai/tinygrad/blob/54dc48aa47aa6e4df612d5b5f349f40b2aaa186a/test/test_assign.py#L143

tinygrad/codegen/uops.py
```diff
@@ -270,29 +270,33 @@ def uop_alu_idx(a: UOp, b, op, dtype=dtypes.int32):
...
       phis = set([u for u in self.get_recursive_children(loop_op) if u.uop is UOps.PHI])
       wheres = set([u for phi in phis for u in get_recursive_parents(phi) if u.arg == TernaryOps.WHERE])
-      if (any([u.uop not in allowed_ops or (u.uop is UOps.ALU and u.arg not in allowed_alus) for phi in phis for u in get_recursive_parents(phi)])
+      if (any([u.uop is not UOps.CONST for u in loop_op.vin])
+        or any([u.uop not in allowed_ops or (u.uop is UOps.ALU and u.arg not in allowed_alus) for phi in phis for u in get_recursive_parents(phi)])
         or any([where.vin[2].arg != 0 or where.vin[0].vin[1].uop is not UOps.CONST for where in wheres])
         or any(len([op for op in get_recursive_parents(where) if op.uop is UOps.LOOP]) == 0 for where in wheres)): continue
       if DEBUG >= 4 and (len(phis) > 0 or len(wheres) > 0): print("simplified {} PHI and {} WHERE in loop".format(len(phis), len(wheres)))
+      loop_length = loop_op.vin[1].arg - loop_op.vin[0].arg
+      for op in [op for op in self.uops if op.arg is BinaryOps.ADD]:
```
chenyuxyz, line 279: eh don't do this, it creates the list first thus loops twice. maybe this?

```
for op in self.uops:
  if op.arg is BinaryOps.ADD and wheres.intersection(get_recursive_parents(op)) and phis.intersection(self.get_recursive_children(op)):
    op.vin = ...
```

patosai: ok

chenyuxyz:
thanks! yea i am interested in arange optimization for variables, it might be similar to O(n) embedding too

PR #3907, opened 2024-03-24
assign regression

Qazalin:
I think the preschedule fix only worked because `a` was assigned to a contig tensor.
Deletes a few things to fix diamond assign to non-contig:
- graph and in_degree only track scheduled realizes
- Create the queue  _after_ the graph, assign_targets can change in_degrees for prev bufs.


geohot:
Can you leave both the contig and non-contig?

PR #3912, opened 2024-03-24
kfd driver wip

geohot:
@nimlgen Can you work the "night shift" on this? It works now to launch kernels.

Things to do:
* Fix private segment args (TestOps.test_dilated_conv_transpose2d fails)
* Fix getting the wait time (why aren't self.device.completion_signal.start_ts and end_ts set?)
* Add SDMA support, replace copyin/copyout in KFDAllocator with SDMA in a queue. Should be the raw queue, not the hsa api. Synchronous like the kernels is fine.

Things I'll handle as part of a larger refactor:
* Adding barrier across multidevice support
* Running in asynchronous mode (one kernel at a time now is fine)
* Writing KFD graph


extra/hip_gpu_driver/test_kfd_2.py
```diff
@@ -0,0 +1,146 @@
...
+  BARRIER_HEADER |= hsa.HSA_PACKET_TYPE_BARRIER_AND << hsa.HSA_PACKET_HEADER_TYPE
+
+  EMPTY_SIGNAL = hsa.hsa_signal_t()
+
+  packet = hsa.hsa_barrier_and_packet_t.from_address(rw_ptr.va_addr+0x38)
+  packet.reserved0 = 0
+  packet.reserved1 = 0
+  for i in range(5): packet.dep_signal[i] = EMPTY_SIGNAL
+  packet.reserved2 = 0
+  packet.completion_signal = hsa.hsa_signal_t(evt.event_id)
```
nimlgen, line 127: I have seen this to be a ptr (looks like to `amd_signal_t` `ROCR-Runtime/src/inc/amd_hsa_signal.h`). Also some info on kernel headers (`ROCR-Runtime/src/inc/amd_hsa_kernel_code.h`).

extra/hip_gpu_driver/test_kfd_2.py
```diff
@@ -0,0 +1,146 @@
...
+  evt = kio.create_event(fd, auto_reset=1)
+
+  BARRIER_HEADER  = 1 << hsa.HSA_PACKET_HEADER_BARRIER
+  BARRIER_HEADER |= hsa.HSA_FENCE_SCOPE_SYSTEM << hsa.HSA_PACKET_HEADER_SCACQUIRE_FENCE_SCOPE
+  BARRIER_HEADER |= hsa.HSA_FENCE_SCOPE_SYSTEM << hsa.HSA_PACKET_HEADER_SCRELEASE_FENCE_SCOPE
+  BARRIER_HEADER |= hsa.HSA_PACKET_TYPE_BARRIER_AND << hsa.HSA_PACKET_HEADER_TYPE
+
+  EMPTY_SIGNAL = hsa.hsa_signal_t()
+
+  packet = hsa.hsa_barrier_and_packet_t.from_address(rw_ptr.va_addr+0x38)
```
nimlgen, line 122: I am not sure, but I think it should be `ring_base.va_addr`. Also before doorbell you have to move the write pointer.

extra/hip_gpu_driver/test_kfd_2.py
geohot: Ahh, `amd_signal_t` I was looking for this!

We shouldn't have to use the ioctl to CREATE_EVENT if the syncing is only between GPUs I think.

geohot:
@nimlgen still seeing failure of some kernels, repro with

`KFD=1 python3 test/test_ops.py TestOps.test_dilated_conv_transpose2d`

geohot:
build_sdma_command should support merging of `addr_31_0` and `addr_63_32` so we don't have `addr_31_0=ts_addr&0xffffffff, addr_63_32=(ts_addr>>32)&0xffffffff)` everywhere

PR #3915, opened 2024-03-25
Fix sm89 PTX=1 compilation

zeux:
The minimum PTX version that supports sm89 is 7.8 (same version also supports sm90); without this ptxas fails when running tinygrad with PTX=1 on RTX 4090.

### DIFF
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -56,9 +56,10 @@ class PTXCompiler(Compiler):
   linearizer_opts = LinearizerOptions("CUDA", suffix="PTX", global_max=[65535, 65535, 2147483647], local_max=[64, 1024, 1024], shared_max=49152)
   def __init__(self, arch:str):
     self.arch = arch
+    self.version = "7.8" if arch >= "sm_89" else "7.5"
     PTXCompiler.linearizer_opts = PTXCompiler.linearizer_opts._replace(has_tensor_cores=int(arch[3:]) >= 80)
     super().__init__(f"compile_ptx_{self.arch}")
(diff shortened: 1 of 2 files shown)


zeux:
version table for reference https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#release-notes-ptx-release-history

tinygrad/runtime/ops_cuda.py
```diff
@@ -56,9 +56,10 @@ class PTXCompiler(Compiler):
   linearizer_opts = LinearizerOptions("CUDA", suffix="PTX", global_max=[65535, 65535, 2147483647], local_max=[64, 1024, 1024], shared_max=49152)
   def __init__(self, arch:str):
     self.arch = arch
+    self.version = "7.8" if arch >= "sm_89" else "7.5"
```
geohot, line 59: Err, what do you think that >= does?

zeux: This should be accurate for existing architectures ;) but sure, I'll adjust to actually parse the integer version

zeux: Updated

geohot: oh haha I thought there was a 10, I was confused with cuda versions.

geohot:
oh the str compare might even be better, apparently sm_90a is an arch

zeux:
Yeah sm_90a is why I used a string, but it will break if NV releases sm_100 for blackwell or above - but there's other cases where `int([arch[3:])` is used so these could be all fixed at once...

geohot:
yea, I'm fine with leaving it for all now. this is a Blackwell problem.

PR #3916, opened 2024-03-25
Fix issues with pointer provenance in load/store through ALU

zeux:
Previously load/store could be incorrectly rendered into ld.global/st.global when the input was an ALU op that performed an address computation with DEFINE_LOCAL on one of the arguments.

This would result in a GPU crash (unfortunately ptxas would silently compile this into a SASS kernel but the address would be a shared memory address). These kernels would be created during beam search for llama example and as such PTX=1 run would crash.

### DIFF
--- a/test/test_uops.py
+++ b/test/test_uops.py
@@ -207,6 +207,7 @@ class TestConstantFolding(unittest.TestCase):
     assert any(uop.uop is UOps.BITCAST for uop in lin.uops.uops), f"{[uop.uop for uop in lin.uops.uops]} does not contain bitcast"
 
 class TestLocalAccess(unittest.TestCase):
+  @unittest.skipIf(Device.DEFAULT in {"LLVM"}, "device doesn't support local memory")
   def test_local_basic(self):
     uops = []
     smem = uop(uops, UOps.DEFINE_LOCAL, PtrDType(dtypes.float32), (), ('smem', 16))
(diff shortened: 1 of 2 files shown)


zeux:
Example or previously incorrect code; note that store argument is an ALU op that uses a local to compute the final address:

```
	mul.lo.s32	%alu_s32_32, %lidx0, %const_s32_9;
	mov.b32		%const_s32_33, 0;
	cvt.u64.s32	%cast_u64_16, %alu_s32_32;
	add.u64		%alu_u64_16, %local_u64_0, %cast_u64_16;
	st.global.f32	[%alu_u64_16+0], %alu_f32_31;
```

zeux:
Btw the code in `ptr_ar` makes me think the intent was to never get IR corresponding to loop index based addressing, maybe the arg isn't cleanly set up and this is more correct to handle elsewhere. Let me know if that's the case and I can debug further.

geohot:
Can you add a test that would have failed before?

zeux:
I didn't find existing tests that tested the assembly output so wasn't sure - do you mean a test that just checks the lowering, or a full end-to-end test that would have crashed with PTX=1 before?

zeux:
Alright I've added tests that exercise the local addressing via a non-constant offset which is what triggered the issue in the first place, *HOWEVER* these tests alone would not trigger the problem (they still feel like they are important to have...).

This is because the core issue was not that the IR that was fed into the assembler used LOAD on a computed address, but that the IR - best as I can tell - was fed into the assembler *twice* during beam search (maybe with some alterations?). The first run would modify the LOAD instructions in place (in `ptr_ar`) and replace the address, that was originally DEFINE_LOCAL, with an ALU computation. It would also correctly assign `.shared` address space. However, the second run on the same graph would overwrite the address space with `.global` as the address was replaced by ALU.

I've adjusted the workaround to be simpler accordingly: don't change .arg if it's already assigned (but validate that the caller didn't pass something non-sensical). This still fixes the problem that came up in context of beam search. To actually exercise the problem in tests we'd need to run the code in local tests through the assembler twice, the tests don't do that atm

geohot:
Sweet! Simple change

PR #3922, opened 2024-03-25
improved caching for pointer arithmetics in ptx

SzymonOzog:
With this and https://github.com/tinygrad/tinygrad/pull/3894 we are matching CUDA speed on `HALF=1 DEBUG=2 python3 extra/gemm/simple_matmul.py`


geohot:
Are these generic rewrite rules that can be applied earlier?

SzymonOzog:
> Are these generic rewrite rules that can be applied earlier? Also, try to use the matching engine instead of if statements, you should be able to pattern match deeper with it.


No benefit anywhere else than asm, currently when we have a store of `x+offset`, and we multiply by itemsize we end up transforming this pattern

```
  13 UOps.ALU            : dtypes.int                [10, 12]                         BinaryOps.ADD
  14 UOps.CONST          : dtypes.int                []                               64
  15 UOps.ALU            : dtypes.int                [13, 14]                         BinaryOps.ADD
  16 UOps.CONST          : dtypes.int                []                               128
  17 UOps.ALU            : dtypes.int                [13, 16]                         BinaryOps.ADD
  50 UOps.STORE          :                           [0, 13, 47]                      None
  51 UOps.STORE          :                           [0, 15, 48]                      None
  52 UOps.STORE          :                           [0, 17, 49]                      None
```

into
```

  66 UOps.ALU            : dtypes.int                [13, 8]                          BinaryOps.MUL
  67 UOps.CONST          : dtypes.int                []                               0
  68 UOps.CAST           : dtypes.ulong              [66]                             None
  69 UOps.ALU            : dtypes.ulong              [0, 68]                          BinaryOps.ADD
  70 UOps.STORE          :                           [69, 67, 63]                     .global
  71 UOps.ALU            : dtypes.int                [15, 8]                          BinaryOps.MUL
  72 UOps.CONST          : dtypes.int                []                               0
  73 UOps.CAST           : dtypes.ulong              [71]                             None
  74 UOps.ALU            : dtypes.ulong              [0, 73]                          BinaryOps.ADD
  75 UOps.STORE          :                           [74, 72, 64]                     .global
  76 UOps.ALU            : dtypes.int                [17, 8]                          BinaryOps.MUL
  77 UOps.CONST          : dtypes.int                []                               0
  78 UOps.CAST           : dtypes.ulong              [76]                             None
  79 UOps.ALU            : dtypes.ulong              [0, 78]                          BinaryOps.ADD
  80 UOps.STORE          :                           [79, 77, 65]                     .global
```

this commit makes it transform `(x+offset)*itemsize)` into `x*itemsize + offset*itemsize` effectively caching `x*itemsize` for further use, and not requiring multiple casts and adds

```
  51 UOps.ALU            : dtypes.int                [13, 8]                          BinaryOps.MUL
  52 UOps.CONST          : dtypes.int                []                               0
  53 UOps.CAST           : dtypes.ulong              [51]                             None
  54 UOps.ALU            : dtypes.ulong              [0, 53]                          BinaryOps.ADD
  55 UOps.STORE          :                           [54, 52, 48]                     .global
  56 UOps.CONST          : dtypes.int                []                               256
  57 UOps.STORE          :                           [54, 56, 49]                     .global
  58 UOps.CONST          : dtypes.int                []                               512
  59 UOps.STORE          :                           [54, 58, 50]                     .global
```

other backends don't really depend on the order of operations, and don't need to multiply by itemsize


as per using the pattern matcher, that was my initial idea as well but it becomes very complex:
we have 2 args we need to consider(ADD, SUB)
times 2 operations(LOAD, STORE)
and they both can be gated so either 3 or 4 uops in vin
that's 8 patterns already,  and given the fact that boolean load has it's own pattern, if we don't want to add the same if statement there it needs additional 2 patterns and a new function, this is 10 additional patterns + 2 additional functions
This could be cleaned up by doing a pattern matcher refactor, I have a few ideas for it but want to finish ptx speedup first to see what other usecases for it might emerge - and if it will be worth it in the long run

SzymonOzog:
updated to match https://github.com/tinygrad/tinygrad/pull/3993

SzymonOzog:
> Can this be tested?

Test added

geohot:
The test shouldn't import `ptr_ar` that's super fragile. It should take in uops and run them through the PTX matcher, ideally running them to confirm they are the same. Think about how to make the test robust to code changes.

Tests should test desired behavior, not code.

SzymonOzog:
> The test shouldn't import `ptr_ar` that's super fragile. It should take in uops and run them through the PTX matcher, ideally running them to confirm they are the same. Think about how to make the test robust to code changes.
> 
> Tests should test desired behavior, not code.

updated

PR #3924, opened 2024-03-25
Zero copy buffers with offset

uuuvn:


geohot:
Can this PR be broken up into smaller tested pieces? Example: start by adding and testing the offset functionality in Buffer

uuuvn:
> Can this PR be broken up into smaller tested pieces? It's hard to review something that changes 10 files.
> 
> Example: start by adding and testing the offset functionality in Buffer

#3992 is just offset Buffer + tests + offset in MallocAllocator

The problem is that it doesn't fit (+28 lines) without ripping out all of the DISK hacks and ripping them out requires adding generic support for offset buffers in scheduler which in turn requires supporting them in jit which is compicated and requries adding support for that in other graph backends which is basically everything and we end up with one big pr again.... Not sure what to do with it

geohot:
Similar functionality exists now

PR #3930, opened 2024-03-25
examples: Fix parameter bandwidth accounting for quantized LLama

zeux:
Instead of assuming every parameter is 2 bytes, add up tensor sizes in bytes. There's still some accuracy issues with bandwidth measurement but this brings things closer to reality.


chenyuxyz:
looks good, thanks!

PR #3934, opened 2024-03-26
_alloc and _free with options

nimlgen:


nimlgen:
@geohot, you good with moving options to `_alloc`, removing `_alloc_with_options`? any reasons you made it separate function? As alternative for CUDA I can add `_free_with_options` (since different function need to be called to free host mem), but unifying them should be less lines

geohot:
Yea, that's fine, as long as `Buffer` is still easy to use and doesn't need options by default.

PR #3936, opened 2024-03-26
Make View.contiguous a @property instead of a field

uuuvn:
Checking offset and strides after modifying offset and strides in `if mask and any(elim := [not (b+1 < e) for b,e in mask])` is fine because it is only triggered if mask isn't None and mask is None is a requirement for contiguous


uuuvn:
Failing tests are because some tests use stuff from sops.gz and it creates view with contiguous as a filed. sops.gz and other tests recreate AST by eval'ing their repr, which won't work with this PR. Not sure if it's worth the effort to rewrite this

PR #3939, opened 2024-03-26
BF16 FMA with CUDA TC

Eliulm:


tinygrad/renderer/cstyle.py
```diff
@@ -248,11 +248,17 @@ def render_kernel(self, function_name, kernel, bufs, uops, prefix=None):
...
+        : "+f"(c.x), "+f"(c.y), "+f"(c.z), "+f"(c.w) : "r"(a_pk[0]), "r"(a_pk[1]), "r"(a_pk[2]),  "r"(a_pk[3]), "r"(b_pk[0]), "r"(b_pk[1]) );
+      return c;}"""]
+    if any(uop.dtype == dtypes.bfloat16 for uop in uops):
+      prefix += ["#include <cuda_bf16.h>", "struct bfloat164 { nv_bfloat16 x, y, z, w; };", "struct bfloat168 { nv_bfloat16 x, y, z, w, a, b, c, d; };",
+      "__device__ bfloat164 make_bfloat164(nv_bfloat16 x, nv_bfloat16 y, nv_bfloat16 z, nv_bfloat16 w) { bfloat164 r={x, y, z, w}; return r; }",
+      "__device__ bfloat168 make_bfloat168(nv_bfloat16 x, nv_bfloat16 y, nv_bfloat16 z, nv_bfloat16 w, nv_bfloat16 a, nv_bfloat16 b, nv_bfloat16 c, nv_bfloat16 d) { bfloat168 r={x, y, z, w, a, b, c, d}; return r; }",
+      """__device__ float4 __cuda_mma_m16n8k16_bf16_f32(bfloat168 a, bfloat164 b, float4 c) { int *a_pk = (int *) (&a), *b_pk = (int *) (&b);
+      asm( "mma.sync.aligned.m16n8k16.row.col.f32.bf16.bf16.f32 { %0, %1, %2, %3 }, { %4, %5, %6, %7 }, { %8, %9 }, { %0, %1, %2, %3 };"
+        : "+f"(c.x), "+f"(c.y), "+f"(c.z), "+f"(c.w) : "r"(a_pk[0]), "r"(a_pk[1]), "r"(a_pk[2]),  "r"(a_pk[3]), "r"(b_pk[0]), "r"(b_pk[1]) );
+      return c;}"""]
```
chenyuxyz, line 261: @flammit do you see a way to do this without copying lines? we would want to add `m16n8k8` and other ones too

flammit: Yeah at a minimum use a generator to make all necessary `make_*` and then for the TC function calls.  I can put up a PR today, but finishing up parallel fuzzing.

Eliulm:
@chenyuxyz `DEBUG=2 EMULATE_CUDA=1 FORWARD_ONLY=1 PYTHON=1 python3 ./test/test_linearizer.py TestLinearizer.test_tensor_cores` is failing because bfloat16 is not supported in numpy and hence does not have type char. So I guess we this skip test for this tensor core?

chenyuxyz:
i'd update the test like casting the output to float if it's bf16, might need to adjust atol and rtol too

chenyuxyz:
I prefer we do a refactor first before adding bf16 tc. Copy pasting more hard-coded line is not sustainable and we want to eventually support every dtype / mnk available without more lines and complexity. I believe @flammit has some ideas and we can coordinate on discord

Eliulm:
Yeah, makes sense.

PR #3944, opened 2024-03-26
Freeing size instead of buffer

uuuvn:
Spent a lot of time trying to find an (non-existent) error in my "buffers with offset" code

Introduced in #3934

How on earth isn't this caught by tests?


uuuvn:
Is there any tests with LRU=0? If there isn't we should add a couple....

PR #3945, opened 2024-03-27
wmma: refactor to remove wmma_func and create TC funcs as needed

flammit:


extra/gemm/simple_matmul.py
```diff
@@ -1,10 +1,12 @@
 import numpy as np
 from tinygrad.helpers import getenv
 from tinygrad import dtypes, Tensor
-dtype_in = dtypes.half if getenv("HALF") else dtypes.float
-acc_dtype = dtypes.half if getenv("ACC_HALF") else None
+dtype_in = dtypes.half if getenv("HALF") else dtypes.bfloat16 if getenv("BFLOAT16") else dtypes.float
```
chenyuxyz, line 4: can you add bf16 to ci benchmark?

chenyuxyz (commented):
i will review this tomorrow, quite excited about it as it reduces lines and adds capabilities!

also need to sync to master

extra/gemm/simple_matmul.py
flammit: done!

tinygrad/renderer/cstyle.py
```diff
@@ -242,17 +247,24 @@ class CUDALanguage(CStyleLanguage):
   type_map = {dtypes.bfloat16: "nv_bfloat16"}
 
   def render_kernel(self, function_name, kernel, bufs, uops, prefix=None):
+    # TODO: why is dtypes.bfloat16.name == "__bf16"? would be easier not override dtypes.name
```
chenyuxyz, line 250: i think `__bf16` works for clang

chenyuxyz (commented):
nice work! looks good

I would suggest next time splitting the make dtype change from the wmma change and make two prs. Easier to debug small individual prs in case we find a regression,

PR #3950, opened 2024-03-27
CommandQueue is the future

geohot:
i plan to replace the jit, schedule running, and a whole bunch of the graph boilerplate with CommandQueue

CommandQueue will be constructed with a schedule in the order you want it run. this schedule contains sync, wait, and assign.

unlike TinyJit which will be a lightweight wrapper, CommandQueue will be picklable.

multiprocess compilation and memory scheduling will be done in CommandQueue, in addition to splitting it among devices and potentially into a separate compute/copy queue.

the default sync/wait will be using CPU, but KFD and CUDA should support faster stuff.


geohot:
Will test it really well with the KFD stuff, KFD is the real API that GPUs have. Will shoehorn cudagraph and metalgraph into this too.

The key thing that it does generically is split the (sorted) list of ScheduleItem into different devices, and determine how the different devices need to sync.

geohot:
CommandQueue's memory scheduler can also fix copy_from_fd

geohot:
I should also be able to remove LoadOps.SYNC and LoadOps.WAIT. These can be computed from the CommandQueue

Issue #3953, opened 2024-03-27
TypeError in limit_dims_to_max method of Kernel.py


JoeruCodes:
The lambda function of `reshape_and_permute` expects a lambda function which takes in a `List[sint]` as input. But the lambda function here takes `x` as input which is then used as an input for `limit_size` method which expects a `List[int]`

https://github.com/tinygrad/tinygrad/blob/6c7df1445b287131862a628937e03e336c895c0c/tinygrad/codegen/kernel.py#L330

When I looked into the implementation of `reshape_and_permute` method, this lambda function is given a `List[sint]` as input from the shape method of shapetracker

![image](https://github.com/tinygrad/tinygrad/assets/79764496/8451dc74-5905-42f5-9130-e4a03692d75b)

JoeruCodes:
changing `x: Tuple[int]` to `x: Tuple[sint]` or an assertion should do the job...

https://github.com/tinygrad/tinygrad/blob/6c7df1445b287131862a628937e03e336c895c0c/tinygrad/codegen/kernel.py#L312

chenyuxyz:
limit_dims_to_max is removed in #4855

PR #3954, opened 2024-03-27
#3953 Fixes TypeError in limit_dims_to_max method of Kernel.py

JoeruCodes:
#3953 

The lambda function of `reshape_and_permute` expects a lambda function which takes in a `List[sint]` as input. But the lambda function here takes `x` as input which is then used as an input for `limit_size` method which expects a `List[int]`

https://github.com/tinygrad/tinygrad/blob/6c7df1445b287131862a628937e03e336c895c0c/tinygrad/codegen/kernel.py#L330

When I looked into the implementation of `reshape_and_permute` method, this lambda function is given a `List[sint]` as input from the shape method of shapetracker

image

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -309,7 +309,7 @@ class Kernel:
 
   # ******************** helpers ********************
 
-  def _limit_size(self, x: Tuple[int], max_size: List[Union[int,float]]) -> Tuple[int, ...]:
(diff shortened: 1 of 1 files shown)


JoeruCodes:
seems like cmp isnt implemented for sints and floats... Maybe an assertion is needed?

chenyuxyz:
mypy is failing, i think `limit_dims_to_max` is just broken

PR #3960, opened 2024-03-27
correct device for validation test in model benchmark CI

geohotstan:
Before both `METAL` and `CUDA` were validating with only the `CLANG` backend
Now CI validation runs for the intended specified backend.

Or if we remove `NOCLANG`, validation test runs for both specified backend and `CLANG`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -30,7 +30,7 @@ jobs:
     # TODO: why is this test not reliable?
     #- name: Run Stable Diffusion
     #  run: python3 examples/stable_diffusion.py --seed 0 --noshow --timing | tee sd.txt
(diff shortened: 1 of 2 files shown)


chenyuxyz:
yea let's keep CLANG for now. i removed it for tinybox because it was slow

chenyuxyz:
does cuda comparison fail?

geohotstan:
> does cuda comparison fail?

I haven't tested `CUDA` yet. 
The tests are still all being ran using `float32` though, so I think it should be fine?
I'll go give `CUDA` a try

chenyuxyz:
i tested on 3090 and it passed, let's keep CLANG for CUDA too

geohotstan:
also passed in my colab 😎 
image
yeah the environment's different so that's worth testing CLANG twice. 👌

PR #3965, opened 2024-03-27
elf parsing

nimlgen:
Tested this elf with HSA it passes test_dilated_conv_transpose2d. Should be something wrong with queue setting I think.


geohot:
Ahh, more hidden queue params.

nimlgen:
Should be good, gpt2 runs fine with this

nimlgen:
It looks like we don’t need it. We can use header which is in an elf. We need this  https://github.com/ROCm/ROCR-Runtime/blob/master/src/loader/AMDHSAKernelDescriptor.h#L199. I will bring it once clang2py is fixed. This header is c++ and adding c++ flag to haw’s clang2py break output (the same thing why I moved sdma regs into a different file for now)

PR #3966, opened 2024-03-27
llama: prevent device initialization outside of __main__

flammit:
causes HSA resources leakages in child compile processes

To repro and make a tinybox REAALLY angry, do the following on master vs here: `PYTHONPATH=. DEBUG=2 BEAM=2 BEAM_MAX_TASKS_PER_CHILD=1 python3 examples/llama.py --gen 1 --size '7B' --temperature 0 --count 10 --prompt "Hello." --timing`.

You'll need to `rocm-smi -d N --gpureset` after to sort of recover (even after reset the default GPU is at 25Mhz and not 0Mhz).

### DIFF
--- a/examples/llama.py
+++ b/examples/llama.py
@@ -160,7 +160,7 @@ class LLaMa:
     model = Transformer(**params["args"], linear=AbsmaxQuantizedLinear, max_context=MAX_CONTEXT, jit=jit) if quantize else Transformer(**params["args"], max_context=MAX_CONTEXT, jit=jit)
 
     if model_path.is_dir():
-      weights = concat_weights([load(filename) for filename in [f"{model_path}/consolidated.{i:02d}.pth" for i in range(params["files"])]], device)
+      weights = concat_weights([load(filename) for filename in [f"{model_path}/consolidated.{i:02d}.pth" for i in range(params["files"])]], device[0] if isinstance(device, tuple) else device)
     else:
       weights = load(str(model_path))
     if "model.embed_tokens.weight" in weights:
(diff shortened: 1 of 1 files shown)


flammit:
> incorrect with multiple devices

Sorry yes - I'm testing a large long sharded beam now.

examples/llama.py
```diff
@@ -160,7 +160,7 @@ def build(model_path, tokenizer_path, model_gen="1", model_size="7B", quantize=F
     model = Transformer(**params["args"], linear=AbsmaxQuantizedLinear, max_context=MAX_CONTEXT, jit=jit) if quantize else Transformer(**params["args"], max_context=MAX_CONTEXT, jit=jit)
 
     if model_path.is_dir():
-      weights = concat_weights([load(filename) for filename in [f"{model_path}/consolidated.{i:02d}.pth" for i in range(params["files"])]])
+      weights = concat_weights([load(filename) for filename in [f"{model_path}/consolidated.{i:02d}.pth" for i in range(params["files"])]], device[0] if isinstance(device, tuple) else device)
```
chenyuxyz, line 163: Does this work because somewhere earlier in __main__ that all devices are inited? Can we have a loop somewhere that does `Device[device] for device in devices` to explicitly init

flammit: I believe the cause of the child process leakage is on line 109, because when the child process spawns, it loads the function definitions.  The use of the default argument Device.DEFAULT means each child will init and not tear down the GPU resources (like AQLQueue) even if it doesn't call that function.

chenyuxyz: oh got it, that makes sense

PR #3968, opened 2024-03-28
Swaps `Tensor.replace` to internal funciton `_replace`

RyanAugust:
Related to review of #3880

### DIFF
--- a/tinygrad/nn/state.py
+++ b/tinygrad/nn/state.py
@@ -68,7 +68,7 @@ def load_state_dict(model, state_dict:Dict[str, Tensor], strict=True, verbose=Tr
       if k not in state_dict and not strict:
         if DEBUG >= 1: print(f"WARNING: not loading {k}")
         continue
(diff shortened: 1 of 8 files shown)


tinygrad/nn/state.py
```diff
@@ -68,7 +68,7 @@ def load_state_dict(model, state_dict:Dict[str, Tensor], strict=True, verbose=Tr
       if k not in state_dict and not strict:
         if DEBUG >= 1: print(f"WARNING: not loading {k}")
         continue
-      v.replace(state_dict[k].shard(mlb.device, mlb.axis) if isinstance((mlb:=v.lazydata), MultiLazyBuffer) else state_dict[k].to(v.device)).realize()
+      v._replace(state_dict[k].shard(mlb.device, mlb.axis) if isinstance((mlb:=v.lazydata), MultiLazyBuffer) else state_dict[k].to(v.device)).realize()# noqa: E501
```
geohot, line 71: don't add noqa: E501

chenyuxyz:
we discussed and the function that is wildly used in examples should not be an internal function

geohot:
Yea, my bad on this. It really just needs to be documented better, but it does have a use as an external API.

PR #3970, opened 2024-03-28
kfd sdma

nimlgen:


geohot:
I don't like stuff being in "driver' I think it should just be in one file. Encourages things to be short.

```
def _build_hdp_cmd(self):
    cmd = amd_sdma.SDMA_PKT_HDP_FLUSH.from_address(self.sdma_ring.va_addr + (self.doorbell_value % self.ring_size))
    cmd.DW_0_DATA = 0x8
    cmd.DW_1_DATA = 0x0
    cmd.DW_2_DATA = 0x80000000
    cmd.DW_3_DATA = 0x0
    cmd.DW_4_DATA = 0x0
    cmd.DW_5_DATA = 0x0
    
should be a struct constructor and a memcpy in a helper function, I bet it's even faster to do it that way if it's in GPU memory.
```

nimlgen:
Yeah, that's `ROCR-Runtime/src/core/inc/sdma_registers.h`. they just mostly randomly choose what case to use for each struct in this file (old ones are `l2_policy`, newer are `LimitVA_HI`).

nimlgen:
I am not sure if this is any better (like unrolling this useless unions in python)... I am not sure which of ways is more readable and easier to maintain: 1) keep unions 2) put headers in extra and modify them (remove unions and uppercase), they should not change a lot, since it's backward capable hw headers, but... 3) remove union in python like it's done rn (but clearer)

geohot:
I can clean it up a bit more, but I like this a lot better. Is this ready?

tinygrad/runtime/ops_kfd.py
```diff
@@ -104,15 +127,20 @@ def __call__(self, *args, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
...
+
+    # TODO: This does not signal, tried higher timeout and it never returns...
+    # evt_arr = (kfd.struct_kfd_event_data * 1)()
+    # evt_arr[0].event_id = self.device.completion_signal.event_id
+    # kio.wait_events(KFDDevice.kfd, events_ptr=ctypes.addressof(evt_arr), num_events=1, wait_for_all=1, timeout=1000)
 
-    evt_arr = (kfd.struct_kfd_event_data * 1)()
-    evt_arr[0].event_id = self.device.completion_signal.event_id
-    kio.wait_events(KFDDevice.kfd, events_ptr=ctypes.addressof(evt_arr), num_events=1, wait_for_all=1, timeout=1000)
+    while True: # Kind of active wait
```
geohot, line 139: Hmm, why do we have to spin?

nimlgen:
Yeah, sdma working, need svm calls (it's just annoying a bit, since they pass it as variable-sized struct with attributes at the end)

nimlgen:
Add svm to test queues with copies from cpu as well. Not sure that looks really clean, maybe can do better. feel free to delete if they are bad :)

PR #3973, opened 2024-03-28
Shape changing bitcast and assert bitcast in disk

uuuvn:
On master disk doesn't assert that a cast is bitcast and does it as a bitcast even if it's a normal cast.
This behaviour is a footgun at least and broken at most. Also this is required (kinda, not really but it would make things much cleaner) for my zero copy offset buffers

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -249,7 +249,7 @@ class TestUint8Dtype(TestDType):
 @unittest.skipIf(Device.DEFAULT == "WEBGL", "No bitcast on WebGL")
 class TestBitCast(unittest.TestCase):
   def test_shape_change_bitcast(self):
-    with self.assertRaises(AssertionError):
+    with self.assertRaises(RuntimeError):
       _test_bitcast(Tensor([100000], dtype=dtypes.float32), dtypes.uint8, [100000])
 
   def test_bitcast_float_to_int32(self):
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -249,9 +249,8 @@ class TestUint8Dtype(TestDType):
 @unittest.skipIf(Device.DEFAULT == "WEBGL", "No bitcast on WebGL")
 class TestBitCast(unittest.TestCase):
   def test_shape_change_bitcast(self):
-    with self.assertRaises(RuntimeError) as cm:
-      _test_bitcast(Tensor.empty((128,), dtype=dtypes.float32), dtypes.uint8)
-    self.assertEqual('shape changing bitcast only supported on DISK right now', str(cm.exception))
+    with self.assertRaises(AssertionError):
+      _test_bitcast(Tensor([100000], dtype=dtypes.float32), dtypes.uint8, [100000])
 
   def test_bitcast_float_to_int32(self):
     a = Tensor([1.,2,3])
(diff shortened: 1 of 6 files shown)


uuuvn:
Right now it's only supported on disk but i think it can be extended to other backends as well in the future if they support it

uuuvn:
Also offset backends will get support for shape changing bitcast for free btw

uuuvn:
> Can you add tests for this? Why is shape changing bitcast only supported on disk and not everywhere?

Because it would require either support from backend or extra realizes (except the cases with folding into store/load)

Right now it's only for disk because of that (disk supports it already due to how it's implemented) and disk is the only thing that uses it

I'll add tests (thought that disk tests that indirectly use this functionality is enough)

obadakhalili:
#3740 not sure if this is useful, but I thought I could mention it in case it was.

geohot:
You added several asserts. Are there tests for those?

uuuvn:
> You added several asserts. Are there tests for those? (also, they probably shouldn't be asserts. if a user can trigger them, they should be Exceptions)
> 
> All behavioral changes you are making should be tested

added

uuuvn:
> Now the test_dtype test is accessing the disk. Don't treat this like a race, think, is this the most beautiful code I can write? What's wrong with it? Where will this cause pain for people in the future?

i mean there can be a race condition if somebody runs two instances of same test at the same time but this is true for other disk tests as well, isn't it?

uuuvn:
also issues with multiuser systems where /tmp/x can be owned by another user but also not unique to my tests

uuuvn:
Updated tests to use mktemp

Can you elaborate on what issue you think is present in my tests? (assuming you don't think that things that other tests do are bad) i don't see what you mean by that

uuuvn:
> Updated tests to use mktemp
> 
> Can you elaborate on what issue you think is present in my tests? (assuming you don't think that things that other tests do are bad) i don't see what you mean by that

If that assumption is wrong then mktemp should fix the issue

uuuvn:
this is done i think

test/test_dtype.py
```diff
@@ -249,8 +249,9 @@ def test_uint8_to_int8_overflow(self):
 @unittest.skipIf(Device.DEFAULT == "WEBGL", "No bitcast on WebGL")
 class TestBitCast(unittest.TestCase):
   def test_shape_change_bitcast(self):
-    with self.assertRaises(AssertionError):
-      _test_bitcast(Tensor([100000], dtype=dtypes.float32), dtypes.uint8, [100000])
+    with self.assertRaises(RuntimeError) as cm:
+      _test_bitcast(Tensor.empty((128,), dtype=dtypes.float32), dtypes.uint8)
+    self.assertEqual('shape changing bitcast only supported on DISK right now', str(cm.exception))
```
geohot, line 254: Why? Is this testing functionality?

uuuvn: Should i remove all error message checks? I thought they were useful because the code in question can throw for another reason (but that's less relevant here compared to other tests)

uuuvn: removed all error message checks

geohot:
Cool, merged.

PR #3976, opened 2024-03-28
remove TYPE_CHECKING between kernel and linearizer

chenyuxyz:
also fixed some type annotation that was not properly checked due to this


chenyuxyz:
`LinearizerOptions` should be refactor to more like a device option, and Kernel should not import Device for every init

chenyuxyz:
moved the LinearizerOption to be CompilerOption. The rest we will do when we move linearizer into pure functions

PR #3989, opened 2024-03-29
Yet another O(n) embedding attempt

patosai:
Todo list:
- [x] correctly put it in one kernel #4036
- [ ] cleanly mark tensor eligible for optimization (this PR)
- [x] implement optimization (this PR)
- [ ] ensure out-of-bounds indices handled
-----
### Original post:

The code is in a pretty nasty state atm, but in essence embedding/getitem multiplies the target by one hot vectors and sums the result, so instead of multiplying and summing lots of 0s, you just get the element you want. This is done in two steps:

1. mark output tensors eligible for the optimization (the nasty part of the code, way too much passing around of `one_hot_optimizable` everywhere)
2. do the optimization in uops

<span>#</span>2 basically entails changing the uops to read the index desired, then instead of looping through all the possible indices, remove the loop and read the elements from the index. 
- **Caveat**: out of bounds values not handled yet lol

Unroll optimizations breaks <span>#</span>2 so I disabled it when the tensor is marked. After doing that, the upcast optim stuff started breaking so I disabled that too for these cases. TBD what to do here too

Example using a very simple example (this is test_nn.test_embedding_tiny):
```python
    a = Tensor([[1, 2, 3, 4, 5, 5],
                [6, 7, 8, 9, 10, 10],
                [11, 12, 13, 14, 15, 15],
                [16, 17, 18, 19, 20, 20],
                [21, 22, 23, 24, 25, 25],
                ])
    b = Tensor.randint(1000, low=0, high=5)
    result = a[b]
```

kernel before:
```
kernel void r_125_2_3_8_5(device int* data0, const device int* data1, const device int* data2, const device int* data3, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.y; /* 125 */
  int gidx1 = gid.x; /* 2 */
  int lidx2 = lid.y; /* 3 */
  int lidx3 = lid.x; /* 8 */
  int acc0 = 0;
  int val0 = *(data1+(gidx0*8)+lidx3);
  int alu0 = (gidx1*3);
  for (int ridx0 = 0; ridx0 < 5; ridx0++) {
    int val1 = *(data2+ridx0);
    int val2 = *(data3+alu0+lidx2+(ridx0*6));
    acc0 = (((int)((val0==val1))*val2)+acc0);
  }
  *(data0+(gidx0*48)+alu0+lidx2+(lidx3*6)) = acc0;
}
```

kernel after:
```
kernel void r_125_2_3_8_5(device int* data0, const device int* data1, const device int* data2, const device int* data3, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.y; /* 125 */
  int gidx1 = gid.x; /* 2 */
  int lidx2 = lid.y; /* 3 */
  int lidx3 = lid.x; /* 8 */
  int val0 = *(data1+(gidx0*8)+lidx3);
  int alu0 = (gidx1*3);
  int val1 = *(data3+alu0+lidx2+(val0*6));
  *(data0+(gidx0*48)+alu0+lidx2+(lidx3*6)) = val1;
}
```
interestingly data2 (the arange) isn't used anymore.


chenyuxyz:
bounty locked, good luck!

patosai:
@chenyuxyz  this is close to what I was imagining for the implementation. I mark a tensor as eligible for this embedding optimization, then in order to get it to the linearizer I have to go Tensor -> LazyBuffer -> LazyOp -> Linearizer which is quite the journey, not sure of a cleaner way to do it currently.

relevant kernel for something like
```python
layer = Embedding(20, 30)
    a = Tensor([[1, 5, 9, 11],
                [12, 19, 8, 1]])
    layer(a).realize()
```

before:
```
kernel void r_8_30_20(device float* data0, const device int* data1, const device int* data2, const device float* data3, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.y; /* 8 */
  int gidx1 = gid.x; /* 30 */
  float acc0 = 0.0f;
  int val0 = *(data1+gidx0);
  for (int ridx0 = 0; ridx0 < 20; ridx0++) {
    int val1 = *(data2+ridx0);
    float val2 = *(data3+gidx1+(ridx0*30));
    acc0 = (((float)((val0==val1))*val2)+acc0);
  }
  *(data0+(gidx0*30)+gidx1) = acc0;
}
```

after:
```
kernel void r_8_30_20(device float* data0, const device int* data1, const device int* data2, const device float* data3, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int gidx0 = gid.y; /* 8 */
  int gidx1 = gid.x; /* 30 */
  int val0 = *(data1+gidx0);
  float val1 = *(data3+gidx1+(val0*30));
  *(data0+(gidx0*30)+gidx1) = val1;
}
```

patosai:
it seems that reducing >= 32768 items leads to multi-kernel reduction, to be investigated
https://github.com/tinygrad/tinygrad/blob/f61ed869f5d44843caa567a911f8c75d5d49fe94/tinygrad/lazy.py#L161-L162

chenyuxyz:
the new kernel looks nice! however, i don't think we want to pipe `is embedding` all the way down. is there a way to detect this pattern in uops? (like for loop that would only hit once, fine to make it restrict to start with, how about hitting exactly twice)

the goal of tinygrad is to be as generic as possible. instead of optimizing a specific kind of kernel, we want to think the broadest possible property we can apply the optimization to

patosai:
What I could try is mark a CMPEQ as one-hot and go from there

patosai:
@chenyuxyz ok - I tried marking cmpeq and it was basically the same as marking embedding - so arange is now marked which I think is the minimum required, you can't detect this optimization from uops alone as it is now.

What I do now is in Tensor.arange I mark the lazybuffer as an arange, with the start/stop/end data. That data goes to MemBuffer and then to the linearizer where there's an optimization which specifically targets aranges with a 0 start/1 step (aka `*(data2 + ridx0) = ridx0`), and where this arange is used in a reduction sum involving a cmpeq true for only one iteration of the loop, so it is a very scoped optimization which is what embedding uses now/getitem will eventually use.

If we like this approach, I still have some cleanup to do around the actual optimization code. Also adding the arange data to the arg of the buf makes the arg kind of long, but I think it's reasonable as then the uopgraph can optimize without needing to know anything else

example buf ops, 1 is not arange but 2 is arange:
```   
   1 UOps.DEFINE_GLOBAL  : ptr.dtypes.long           []                               (1, 'data1', False, False)
   2 UOps.DEFINE_GLOBAL  : ptr.dtypes.int            []                               (2, 'data2', False, (0, 28, 1))
```

chenyuxyz:
What's the reason that this cannot be detect from uops alone, do you have some examples? I understand your approach, trying to understand the fundamental reason for it.

patosai:
The code beforehand is something like
```
  int val0 = index;
  for (loop) {
    int val1 = *(data2+loop);
    float val2 = *(data3 + some other value including loop)`
    acc0 = (val0==val1)*val2+acc0;
  }
```
If you don't know anything about data2, you don't know how many times the equality condition will be hit. If `data2` is a buffer full of `val0`, it can be hit `loop` number of times - or maybe `data2` is an arange where `val0` appears exactly once in the buffer, you can't tell from uops alone

patosai:
after discussion, the more general/desired approach is a combo of:
1. Variable holding a tensor value so the embedding indices can be used in shrink
2. Tensor.stack optimization so the shrink stack goes from O(N^2) to O(N)

PR #3992, opened 2024-03-29
Buffer views (first part of zero copy offset buffers)

uuuvn:
https://github.com/tinygrad/tinygrad/pull/3924#issuecomment-2023126645

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -129,6 +129,7 @@ class _MallocAllocator(LRUAllocator):
   def as_buffer(self, src) -> memoryview: return flat_mv(memoryview(src))
   def copyin(self, dest, src:memoryview): ctypes.memmove(dest, from_mv(src), len(src))
   def copyout(self, dest:memoryview, src): ctypes.memmove(from_mv(dest), src, len(dest))
+  def offset(self, buf, offset: int, size:int): return ctypes.cast(ctypes.byref(buf, offset), ctypes.POINTER(ctypes.c_uint8 * size)).contents
 MallocAllocator = _MallocAllocator()
 
 # **************** for Compiled Devices ****************
--- a/test/test_offset_buffer.py
+++ b/test/test_offset_buffer.py
@@ -2,7 +2,7 @@ import unittest
 from typing import Union, Tuple, Optional, List
 from tinygrad.buffer import Buffer, flat_mv
 from tinygrad.dtype import DType, dtypes
-from random import randbytes, randint, seed
+from random import randbytes, randint, choice, seed
 
 def mvs(mv: memoryview) -> str:
   return str(list(mv))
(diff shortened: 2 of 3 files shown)


uuuvn:
Also right now when it uncows buffer it copies it to cpu and back, i will fix that tomorrow

geohot:
Why does it have copy on write? I would expect it to actually just be the same memory.

tinygrad/device.py
```diff
@@ -129,6 +129,7 @@ def _alloc(self, size:int, options:BufferOptions): return (ctypes.c_uint8 * size
   def as_buffer(self, src) -> memoryview: return flat_mv(memoryview(src))
   def copyin(self, dest, src:memoryview): ctypes.memmove(dest, from_mv(src), len(src))
   def copyout(self, dest:memoryview, src): ctypes.memmove(from_mv(dest), src, len(dest))
+  def offset(self, buf, offset: int, size:int): return ctypes.cast(ctypes.byref(buf, offset), ctypes.POINTER(ctypes.c_uint8 * size)).contents
```
geohot, line 132: I think this function signature works.

geohot:
In fact, it definitely shouldn't be copy on write. Think about the disk backend.

uuuvn:
The copy on write is optional. The idea is that by default it behaves like a copy (no write passthrough) to save memory without everything above the Buffer level except scheduler (which creates those CoW copies) even knowing about it.

The cow=False feature is for setitem (including DISK one) to work properly. Which would probably be just a flag to contiguous

If i make it always cow=False which is like very easy and i had it a long time ago and most of the time was actually spent on CoW stuff and testing it then on backends that support offsets something like:
```python
b = a.shrink(((4,16),)).contiguous()
b.assign(c)
print(a.numpy())
```
would give different result than on a backend that doesn't support it. Also it might break things that rely on this behaviour.

uuuvn:
Read only is probably fine but then it's not really possible to know if something is going to be written to and so when something from eg another run_schedule attempts to write into it we just end up with either incorrect behaviour or just doing CoW at another level

uuuvn:
Also see the `test_simple_onelevel` test where both cow=False behaviour (for DISK and setitem) and cow=True (for saving memory when .contiguous() can be done copy-free

uuuvn:
I mean if you think that cow=True stuff is useless i can just make it always passthrough but then it's a potential for bugs in future when we assign to something that looks like it should be just itself but actually some other tensor is modified because the thing we assigned to was a offset pseudo-copy of that other thing we don't expect to change

uuuvn:
maybe i'm just overthinking stuff idk

uuuvn:
> Why does it have copy on write? I would expect it to actually just be the same memory (or maybe read only).

#4008 is the same as this pr except always cow=False , ready to review

uuuvn:
closing in favor of #4008

PR #3999, opened 2024-03-29
extra/gemm/hip_matmul: fix to use new HSA devices and no headers

flammit:


extra/gemm/hip_matmul.py
```diff
@@ -93,22 +111,32 @@
       }}
     }}
   }}
-}}""")
+}}"""
 
-prog = HIPProgram(device.device, "test", lib)
+if DEBUG > 1: print(prog_str)
+lib = compile_hip(prog_str)
```
chenyuxyz, line 117: use `device.compiler.compile` instead of importing `compile_hip` directly?

flammit: done!

PR #4008, opened 2024-03-30
Always-passthrough offset buffer

uuuvn:
A version of #3992 which is always cow=False

Fuzz tests are less relevant because their purpose was to test CoW logic agains naive copy when cow and slice when not cow and now there isn't really much to test


uuuvn:
I've wrote in #3992 why i think that copy on write by default approach is better but it's arguably also less clean so yeah

tinygrad/runtime/ops_disk.py
```diff
@@ -1,26 +1,15 @@
...
-  def __repr__(self): return f"<DiskBuffer size={self.size} dtype={self.dtype} offset={self.offset}>"
-  def _buf(self) -> memoryview: return memoryview(self.ud.mem)[self.offset:self.offset+self.size*self.dtype.itemsize]
+  def __init__(self, fd, mem, offset, size): self.fd, self.mem, self.offset, self.size = fd, mem, offset, size
+  def _buf(self) -> memoryview: return memoryview(self.mem)[self.offset:self.offset+self.size]
 
 MAP_LOCKED, MAP_POPULATE = 0 if OSX else 0x2000, getattr(mmap, "MAP_POPULATE", 0 if OSX else 0x008000)
 class DiskAllocator(Allocator):
   def __init__(self, device:str): self.device = device
-  def _alloc(self, size:int, options):
+  def _alloc(self, size:int, options) -> DiskBuffer:
```
geohot, line 12: You have to be careful that this isn't being called multiple times for the same file, you don't want it reopening the file a ton of times.

tinygrad/engine/realize.py
```diff
@@ -11,9 +11,17 @@ def __init__(self, fxn):
     super().__init__()
   def __call__(self, rawbufs:List[Buffer], var_vals:Dict[Variable, int], wait=False, jit=False): self.fxn(*rawbufs)
 
+def maybe_zerocopy(si: ScheduleItem) -> bool:
+  if not (all([len(x) == 1 for x in [si.outputs, si.inputs, si.ast]]) and si.outputs[0].device.startswith("DISK")): return False
+  ast, out, inp = si.ast[0], si.outputs[0], si.inputs[0]
+  top_src = ast.src[0].src[0] if ast.src[0].op is UnaryOps.CAST and ast.src[0].arg[1] else ast.src[0]
+  if not (top_src.op is BufferOps.LOAD and top_src.arg.st.consecutive and ast.arg.st.contiguous) or hasattr(out, "_buf"): return False
+  out.base, out.offset = (inp.base or inp), inp.offset+(top_src.arg.st.views[0].offset*inp.dtype.itemsize)
```
geohot, line 19: Oh, I see, allocate won't trigger because of this...

uuuvn: I don't think it is possible to do it cleanly. I've tried and some of the problems are:
- toposort in schedule doesn't understand that realizing one buffer can realize its zero copy childs and teaching it do understand it isn't very clean 
- there are different ways to trigger zero copy (contiguous, bitcast, assign) and they can be nested (eg assign after bitcast). This makes it hard to trigger on lazybuffer level but all of this is already taken care of in scheduling so it's a simple pattern match with bitcast skip afterwards (see maybe_zerocopy function before i removed it)

I think the most clean way to do it is to move maybe_zerocopy between schedule_one which takes care of contiguous/bitcast/assign+combinations of them already and the place where multioutput will merge ScheduleItems.

uuuvn: But "then that buffer will become "allocated" when the parent is" is a very good idea and i've implemented it so that zero copy can be triggered before realization of base

uuuvn: Done.

@geohot can you enable benchmars/additional tests/whatever this is for this pr? tiny13 is busy again (and seems to be +-always busy recently) with some wandb stuff

uuuvn:
I've been very busy recently and can't finish it - @geohot can you unlock the bounty?

PR #4010, opened 2024-03-30
Added tensor size call for tensor object

Tassadon:


chenyuxyz:
i don't think Tensor should have size

Tassadon:
How would you go about solving for MSE loss then without size? There must be some trick to it, I don't understand or is getting the product of a tensor shape more safe to prevent integer overflow?

PR #4026, opened 2024-04-01
test assign in JIT

mmmkkaaayy:
realized we should probably keep the core of the test removed in #4024


mmmkkaaayy:
actually there is some coverage already in test_assign so this is not necessary

PR #4027, opened 2024-04-01
kfd free buffers

nimlgen:


geohot:
Can we find a PM4 header somewhere to not hand define these things?

nimlgen:
All tests (but multitensor) pass now. Parsed `nvd.h` into python in autogen (clang2py can't even get defines as variables). Renamed amd_sdma.h -> amd_gpu.h to include nvd.h into this file as well. So review pls how this looks overall

geohot:
Someday we'll clean up clang2py, but it doesn't have to be today.

PR #4028, opened 2024-04-01
multioutput reduce pairs repro

Qazalin:
good repro of a failure: `GPU=1 python3 -m unittest test.models.test_train.TestTrain.test_convnext` (r2_49_16_2_4_16) - realized_children are in different parts of the graph.

@geohot I also don't think just removing the constraint in line 157 is enough for Adam.


Qazalin:
#4030 is a better diff for this.

PR #4030, opened 2024-04-01
minimal diff for multioutput reduce pairs

Qazalin:
Trains beautiful_mnist with ~25% less kernels.

this diff: https://tiny-tools-client.vercel.app/?id=6d8f2c55c53a4150976e39d6b7e20ffd
master: https://tiny-tools-client.vercel.app/?id=0c8240da3a9b4eef8759987ef3df4708

A lot of reduces with multiple children went from 3 kernels to 1 kernel:

Screenshot 2024-04-14 at 5 36 18 PM

test plan:
- Unit tests as the "multi output spec"
- #4136


geohot:
Ahh, this looks a lot more like the diff I was expecting. Can you post a `GRAPH=1` output of something that fails?

Qazalin:
`test/models/test_train.py::TestTrain::test_convnext` 

This is the graph: https://github.com/tinygrad/tinygrad/assets/77887910/acfd0178-1c56-4799-9b5e-4332b0eb7f75
Oh it's actually because it keeps recursing the realized child's children - it should break.

For this reduce: `(2, 56, 56, 64, 1)`
image

the realized_children search _should_ stop once we find a realized child.
But it's going all the way down to this MUL - BinaryOps.ADD's child...
Screenshot 2024-04-01 at 10 49 03 PM

I'll figure out a fix

Qazalin:
a self-reference check fixes almost all the issues
some of children are still in different levels though
Screenshot 2024-04-03 at 1 11 20 AM
Maybe we could do a deeper search to get the depth of every child relative to root

tinygrad/engine/schedule.py
```diff
@@ -193,12 +206,19 @@ def create_schedule_with_vars(outs:List[LazyBuffer], seen:Optional[Set[LazyBuffe
...
+    else: reduce_for_op.update((tr, r) for tr in realized_children)
+
+  output_groups: DefaultDict[Tuple, List[LazyBuffer]] = defaultdict(list)
+  for r in realizes:
+    if r.realized is not None or r.op is LoadOps.CONST or r in seen: continue
+    output_groups[(reduce_for_op[r], ) if r in reduce_for_op else (r, )].append(r)
+
+  # bufs should fuse in the same order
+  for k, group in output_groups.items():
+    if len(group) > 1: output_groups[k] = list(sorted(group, key=lambda x: x.key))
```
Qazalin, line 218: debatable if there is a better way to sort LazyBuffer ASTs - we need it for the compiler cache

geohot: Where is the nondeterminism coming from? Can we catch it earlier and not use sort?

tinygrad/lazy.py
```diff
@@ -52,6 +52,10 @@ def realized(self) -> Optional[Buffer]:
   @property
   def base(self) -> LazyBuffer: return self._base if self._base is not None else self
 
+  @functools.cached_property
+  def key(self) -> bytes: return hashlib.sha256(functools.reduce(lambda x,y: x+y, [s.base.key for s in self.srcs if s.base.realized is None],
```
geohot, line 56: Yea, this doesn't work, crypto hash functions are super slow. We have to find the source of the nondeterminism and tiebreak it there.

Qazalin:
#4180 prereq for green CI

geohot:
Test failure after rebase.

geohot:
Congrats! Is this the bounty?

PR #4033, opened 2024-04-01
move exec_alu from uops to ops

chenyuxyz:
will use this for const folding in lazy too


chenyuxyz:
@geohot this looks good to you?

geohot:
Very much so, that totally shouldn't be in uops

PR #4037, opened 2024-04-02
Make Mamba Maintainable

reddyn12:
I cleaned up the code. I tested up to 1.4b param model on 3080 and they work.  Getting 
26sec - 1000 tokens - 1.4b
Mamba is fast!


chenyuxyz:
not really interested. you had the chance to make it good

PR #4039, opened 2024-04-02
Use exec_alu for lazy const folding

chenyuxyz:
fixed `1 ** Tensor` folding.

wip. `test_onecyclelr` became very slow and need to prep this with some update rules in uops optimization to avoid `tautological-compare` warnings in clang

### DIFF
--- a/test/test_const_folding.py
+++ b/test/test_const_folding.py
@@ -1,5 +1,5 @@
 import unittest, math
-from tinygrad import Tensor, Device
+from tinygrad import Tensor, Device, dtypes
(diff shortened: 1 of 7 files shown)


chenyuxyz:
`test_onecyclelr` is slow because now the lr update is not lazy and kernel recompiles...

chenyuxyz:
fixed scheduler. make the base const of `const ** tensor` a tensor too, otherwise it's being const folded and generate different kernels for each step. everything else benchmark-wise looks correct to me

PR #4041, opened 2024-04-02
uops const fold rules to prevent tautological compare warnings

chenyuxyz:
`bool < false` is false, `true < bool` is false

(`a == a` is true, `a != a` is false is not implemented in this pr due to issues with float and nan)


chenyuxyz:
eh except nan...

PR #4043, opened 2024-04-02
Refactor the use of pattern matcher in ptx

SzymonOzog:
Updated as promised. I left out `ptr_arm` as a separate function as it's quite complex and needs to be called for all LOAD/STORE not fitting that well with the pattern matcher

### DIFF
--- a/tinygrad/renderer/assembly.py
+++ b/tinygrad/renderer/assembly.py
@@ -43,34 +43,7 @@ def uops_to_asm(lang:AssemblyLanguage, function_name:str, uops:UOpGraph) -> str:
   kernel:List[str] = []
   bufs = []
 
-  def eq_rep(root, x, y):
-    root.arg = BinaryOps.XOR
-    return uops.add(UOps.ALU, dtypes.bool, (root,), arg=UnaryOps.NEG, insert_before=uops.uops.index(root)+1)
-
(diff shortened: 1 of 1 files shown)


tinygrad/renderer/assembly.py
```diff
@@ -106,6 +85,18 @@ def ptr_ar(root):
       if o in u.vin and u is not n:
         u.vin = tuple(n if x == o else x for x in u.vin)
     if rew := matcher.rewrite(u): replace[u] = rew
+
+  for o,n in replace.items():
```
geohot, line 89: What does this do? Should this be in the pattern matcher?

SzymonOzog: This goes through the uop returned by the fxn passed into the pattern matcher and all of it's parents and adds them to the UOpGraph if they aren't there

geohot: That should be in pattern matcher!

SzymonOzog: oh it works on `replace` continuing on the work done in the previous block - not sure how to fit it into the pattern matcher, could you elaborate on that?

geohot:
Meh, I'll merge it, it's less lines. But this should really be generic code to apply a pattern to a formed list of uops.

PR #4044, opened 2024-04-02
nv driver

nimlgen:


geohot:
If you have an improved IOCTL=1 for NVIDIA, maybe get that merged first?

nimlgen:
Some simple kernels work, but stuck after running 3-4 as of now. Will investigate.

Yeah, might put ioctl into a sep pr, also wanna include a fifo dumper into it

geohot:
Good progress.

nimlgen:
No I haven't spent much time on multigpus, just basic p2p experiments mapping cpu pages to both gpus, but it's slower than cuda's transfer. Want to fix and cleanup one gpu setup before going multigpu.

nimlgen:
As of now all tests pass for single gpu. Cleaned up and refactored into queues. Still want to clean up nv imports somehow and memory management (while refactoring for multigpu).

geohot:
Next steps:
1. It doesn't need to support cards without P2P support, welcome to delete that from NV.
2. It should work with the HCQ tests I wrote for KFD
3. Add a graph executor that works for both KFD and NV.
4. Get PM4 packets working (better) in KFD. (I think I forget to clear the HDP)
5. Test KFD a bit more, when ready, delete HSA.

geohot:
Why are class_ioctl and nv_ioctl still in this repo? Can't you just use nv_gpu?

geohot:
What's it going to take to merge this?

nimlgen:
Should be good to merge. Tests and models look good

PR #4045, opened 2024-04-02
search: fix counting of upcasts to ignore TC upcasts

flammit:
TC upcasts don't impact the size or complexity of the kernel code

Fixes it so default search params will properly explore upcasts after the TC (i.e. for `simple_matmul.py`), but will increase the search time on TC-eligible kernels.


chenyuxyz:
i see a very small improvement on llama but that might be just noise, similar search time too

PR #4046, opened 2024-04-02
Mergeability for shapetrackers

PaulGustafson:
Functions to convert a shapetracker to a tensor of indices + Boolean tensor of masks.  In this representation, it's easy to check if the shapetracker is single-view:

1. Check if the mask is rectangular (only 1's in a region of the form $\prod_i [a_i, b_i]$ and 0's elsewhere)
2. Calculate the strides by shifting by 1 in each direction in the mask=1 region, and then  the offset is also determined.  
3. Check if the strides are consistent by checking equality between your shapetracker's tensor rep and the tensor rep corresponding to the single-view shapetracker with those strides, offset, and mask.

Next steps:
- Add rectangular mask check
- Fix errors showing up when checking for equality of tensors
- Clean up code + typing
- Figure out whether this representation is better computationally than existing shapetracker class


PaulGustafson:
This should be a provably correct version of merge_views (modulo figuring out offsets, and testing).  The only problem is that it depends on contiguous Tensors, so I can't just replace the existing merge_views impl (would create a circular dependency).

One other interesting thing -- to_tensor is really a function of a Node, so probably useful elsewhere if we keep using symbolic.

PaulGustafson:
Seems to be an improvement over existing merge_view code.  No known merge-view regressions, although running the fuzzer is slightly slower.    Also, I haven't checked Tensors with symbolic values or bounds.   Example regression fuzz test output showing improvements:

```CHECK_MV=1 METAL=1 FUZZ=plus PYTHONPATH="." /opt/homebrew/bin/python3 test/external/fuzz_shapetracker_math.py

new simplify better than old
OLD: ShapeTracker(views=(View(shape=(6, 6, 4), strides=(24, -8, 1), offset=28, mask=((0, 6), (1, 4), (0, 4)), contiguous=False), View(shape=(14,), strides=(1,), offset=86, mask=None, contiguous=False)))
NEW: ShapeTracker(views=(View(shape=(14,), strides=(1,), offset=78, mask=((0, 2),), contiguous=False),))
new simplify better than old

new simplify better than old
OLD: ShapeTracker(views=(View(shape=(5, 6, 7, 7), strides=(30, 5, 0, 1), offset=0, mask=((0, 5), (0, 6), (1, 5), (0, 5)), contiguous=False), View(shape=(5, 35, 2), strides=(294, 7, 1), offset=12, mask=None, contiguous=False)))
NEW: ShapeTracker(views=(View(shape=(5, 35, 2), strides=(0, 0, 0), offset=0, mask=((0, 0), (0, 0), (0, 0)), contiguous=False),))

new simplify better than old
OLD: ShapeTracker(views=(View(shape=(3, 5, 6), strides=(0, 40, 1), offset=0, mask=((2, 3), (0, 5), (0, 6)), contiguous=False), View(shape=(1, 30), strides=(0, 1), offset=30, mask=None, contiguous=False)))
NEW: ShapeTracker(views=(View(shape=(1, 30), strides=(0, 0), offset=0, mask=((0, 0), (0, 0)), contiguous=False),))

new simplify better than old
OLD: ShapeTracker(views=(View(shape=(9, 4, 6), strides=(6, 0, 1), offset=0, mask=((0, 9), (1, 2), (0, 6)), contiguous=False), View(shape=(18, 3, 1, 6), strides=(12, 0, 0, 1), offset=0, mask=None, contiguous=False)))
NEW: ShapeTracker(views=(View(shape=(18, 3, 1, 6), strides=(0, 0, 0, 0), offset=0, mask=((0, 0), (0, 0), (0, 0), (0, 0)), contiguous=False),))

simplify2 better than simplify 0.40%
simplify2 worse than simplify 0.00%
```

PaulGustafson:
Bad running time, going to reopen this as a reference implementation for the fuzzer.

unknownusername504:
> Bad running time, going to reopen this as a reference implementation for the fuzzer.

I would expect this depending on the complexity of the mask. In the worst case scenario you are essentially copying (n+(n-1)/2) elements with this approach.

You could improve the average case by representing contiguous elements and masks in compressed form with slicing.

It does seems useful for a correctness proof though.

PR #4051, opened 2024-04-03
only fold empty shape const for advanced indexing

### LINKED ISSUE #4048: advance indexing generates wrong output shape if indexed with a const tensor
repro
```
a = Tensor.arange(16).reshape(1, 1, 4, 4).realize()
b = Tensor.zeros(1, dtype=dtypes.int)
print(a[:, :, b, :].shape)  # (1, 1, 4)
print(a[:, :, b.contiguous(), :].shape)  # (1, 1, 1, 4)
```

the index should be treated as if it's a contiguous tensor.
this value extraction is not correct: https://github.com/chenyuxyz/tinygrad/blob/f61ed869f5d44843caa567a911f8c75d5d49fe94/tinygrad/tensor.py#L425

@geohotstan fyi

geohotstan:
fixes #4048 
Yeah, only empty shape Tensors can be folded into int indexing.
Anything with shape has to be done through advanced indexing.

Lemme think if there's a better fix though

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -422,7 +422,7 @@ class Tensor:
     else: indices = [indices]
 
     # turn scalar Tensors into const val for int indexing if possible
-    indices = [self._to_const_val(i) if isinstance(i, Tensor) else i for i in indices]
+    indices = [self._to_const_val(i) if isinstance(i, Tensor) and i.shape == (0,) else i for i in indices]
     # move Tensor indices to the same device as self
     indices = [i.to(self.device) if isinstance(i, Tensor) else i for i in indices]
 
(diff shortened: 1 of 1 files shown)


chenyuxyz:
`indices = [self._to_const_val(i) if isinstance(i, Tensor) and i.shape == () else i for i in indices]` is fine. can you add a test for it?

chenyuxyz:
added a simple fix and test in #4047. feel free to iterate on that (like maybe rewrite const tensor as list of list of list?) and more test coverage if you see fit

geohotstan:
> added a simple fix and test in #4047. feel free to iterate on that (like maybe rewrite const tensor as list of list of list?) and more test coverage if you see fit

okok, I'll close this PR for now. 
Ideally I think a Const Tensor with shapes that isn't `()` and is `is_unrealized_unpadded_const` can be done by no copy basic indexing instead of leaving it for advanced indexing. (just gotta int index, reshape, expand?)

I've been trying to get that to work :D 
but the way I implemented it before is a little hard to get it working so I might do another little rewrite lol.

Issue #4016, opened 2024-03-31
more const folding


chenyuxyz:
move `to_const_val` logic into lazy
- #4020
- #4022 
- #4023 
- #4025
- #4029

multi tensor const folding
- #4031
- #4032

ALU ops
`Tensor.ones(3, 3).exp()`, `Tensor.ones(3, 3) + 1`, `Tensor.ones(3, 3) * 3` don't need a kernel
- #4039 

`Tensor.ones(3, 3).cast(dtypes.int64)`
- #4047 

`Tensor.ones(3, 3).sum()`
- #4059

chenyuxyz:
considered this done. There are some potential for `pow ** const Tensor` to be nicer, but pow itself needs more cleanup.

really wanted to merge the rules in uops with rules in lazy

also this paves the idea to fold ops for expanded tensor

PR #4052, opened 2024-04-03
Const folding reduce ops

ignaciosica:
Sum and Max const folding

#4016

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -205,9 +205,7 @@ class TestLinearizer(unittest.TestCase):
   def test_sum_collapse(self):
     t = Tensor.ones(256,256).sum()
     sched = [si for si in create_schedule([t.lazydata]) if si.ast[0].op not in LoadOps]
-    assert len(sched) == 1
-    lin = Linearizer(*sched[0].ast)
-    assert not any(u.uop is UOps.LOOP for u in lin.linearize().uops), "found loop in sum collapse"
+    assert len(sched) == 0
 
   def test_assign_fold(self):
     a = Tensor.ones(4, 4).contiguous().realize()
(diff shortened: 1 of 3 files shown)


test/test_linearizer.py
```diff
@@ -205,9 +205,7 @@ def test_limit_dims_to_max_5d_global(self):
   def test_sum_collapse(self):
     t = Tensor.ones(256,256).sum()
```
ignaciosica, line 206: I believe this test can be removed altogether

ignaciosica: removed

test/external/external_test_opt.py
```diff
@@ -178,7 +178,7 @@ def test_fold_reduce_elementwise(self):
     with CLCache():
       ret = img.sum() + addme
       ret.realize()
-      assert len(CacheCollector.cache) == 1, "optimizer didn't fold reduce/elementwise"
+      assert len(CacheCollector.cache) == 0, "optimizer didn't fold reduce/elementwise"
```
geohot, line 181: This shouldn't be the fix here. The `Tensor.ones` should be changed to something else

ignaciosica: Something that isn't const?
Like: 
def test_fold_reduce_elementwise(self):
    img = Tensor([2,3,4])
    addme = Tensor([1])
    with CLCache():
      ret = img.sum() + addme
      ret.realize()
      assert len(CacheCollector.cache) == 3, "optimizer didn't fold reduce/elementwise"

chenyuxyz:
between incorrectly changed the old test, not fully tested the behavior of the implementation, and overly complicated way to implement, don't think i will merge this

PR #4054, opened 2024-04-03
Graph ScheduleItems

Qazalin:
The first output should _only_ be a unique key for a ScheduleItem. This diff refactors some parts of the scheduler that treat it as the only output.


geohot:
Makes sense

PR #4056, opened 2024-04-03
Add pattern matcher tests, move uop transforms from assembly to pattern

SzymonOzog:
Your last comment on the previous pr clarified what are the expectations for this - hope my question was not interpreted as stalling on this, just wanted to make sure about the direction


geohot:
Sweet! Not at all, I appreciate this work

PR #4057, opened 2024-04-03
Fix for "MTLLibrary is not formatted as a MetalLib file" issue

tabashki:
Fix for: [#2226](https://github.com/tinygrad/tinygrad/issues/2226)

Tested on: `macOS 14.4.1` (GPU: `M3 Max`)

Was able to run `test/test_ops.py` with no failures in a conda environment, whereas previously it would fail with a stream of MTLLibrary errors similar to linked issue. Should be possible by extension to run tinygrad without depending on an existing Xcode installation, but haven't tested that thoroughly

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -50,7 +50,7 @@ class MetalProgram:
         shader.write(lib)
         shader.flush()
         os.system(f"cd {pathlib.Path(__file__).parents[2]}/disassemblers/applegpu && python3 compiler_explorer.py {shader.name}")
-    assert lib[:4] == b"MTLB", "Invalid Metal library. Could be due to using conda. Try system python or METAL_XCODE=1 DISABLE_COMPILER_CACHE=1."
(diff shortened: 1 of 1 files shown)


tabashki:
Hmm, looks like that recently added `assert` might be doing more harm than good in this case. Looks like the actual file format output by `MTLBinaryArchive` might also be OS version dependent, but it *should* always be valid to pass to `newLibraryWithData` since it's generated by the public API

tabashki:
> Any way to do this without creating a file?

Yeah was investigating that for a while but didn't find a solution. Unless there's a magic type of in-memory file URL protocol that can be passed to `serializeToURL`, this is the only API exposed so far:
https://developer.apple.com/documentation/metal/mtlbinaryarchive/3553928-serializetourl?language=objc

tabashki:
This approach might turn out to be a dead end. After some hacking on the M1-based Github runners, they appear to serialize `MTLLibrary` and `MTLBinaryArchive` quite differently from M3. And in all honesty there's nothing in the public docs that implies there'd be any binary compatibility, so that may have just been a happy accident but not a proper solution. Going to keep this open for now to see if I can't make a breakthrough somewhere

PR #4058, opened 2024-04-03
Const folding reduce ops 2

ignaciosica:
Test the behavior of implementation pending

#4016


ignaciosica:
is implementation more clear now? Or it is wrong to const fold those reduce operations here?

chenyuxyz:
#4059

PR #4061, opened 2024-04-03
Deterministic schedule order

Qazalin:
Tested by fuzzing stdout of `for si in schedule: print(si.outputs[0])` in `test/test_multitensor.py TestMultiTensor.test_simple_add_X `https://gist.github.com/Qazalin/cd7d88ba1b221ed58b46ea4a091f3a89

There can be multiple valid topological sorts of a DAG. In this graph https://tiny-tools-client.vercel.app/?id=785de65a9e5246ae9c6f651a3f96d453

320464226-44cda5f9-abfb-46e1-95f1-0fe1f9d6b2a0

Any ordering of nodes in the same depth of the tree is valid.


geohot:
So while this works, it's quite a heavy solution. Where is the nondeterminism coming from?

Qazalin:
4180 does the job

PR #4062, opened 2024-04-03
remove define global without child

chenyuxyz:
Appending fake global is incorrect because the unused buffer can be in the middle. And the buffer map information no available to the caller after compilation based on lib alone.

This pr implemented `UOpGraph.buf_idxs` that's updated in `linearize` after childless UOps removal. The caller uses it to filter the input buffers. This does not require renderer change.

@geohot thoughts on this approach?

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -393,7 +393,7 @@ class Linearizer(Kernel):
     # optimize the uops
     self.uops.uoptimize()
     # indices of global buffers that are used in kernel
-    self.buf_idxs = [uop.arg[0] for uop in self.uops.uops if uop.uop is UOps.DEFINE_GLOBAL]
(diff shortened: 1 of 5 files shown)


geohot:
I'm fine with this. This is the closer to the real solution. You can make fake work I think, I don't see why it has to be at the end, but this is cleaner, we just have a little more state to track.

chenyuxyz:
cool, need to think more about jit and graph. also need some more test cases

chenyuxyz:
not easy to come up with a test case for this because pretty much all uop optimization that removed a buffer is done in lazy already. this might need to wait until we have a true uop only optimization like embedding

geohot:
There's some tests that trigger removal. Btw, we now have a clean place for this. The ScheduleItem should have the buffers but the ExecItem should not.

PR #4068, opened 2024-04-04
add and enable tests for indexing const folding

geohotstan:
2 additional tests in `test_indexing` also enabled by #4047 
also some basic `const_folding` tests to test folding advanced indexing into basic indexing.

Will do the `test_const_tensor_index` TODO next.


chenyuxyz:
thanks!

PR #4085, opened 2024-04-05
broadcast expand to match torch

### LINKED ISSUE #4070: Tensor.expand inconsistencies
reproduce:

```
In [5]: torch.ones(1,1,1).expand(2,3,2,3).shape
Out[5]: torch.Size([2, 3, 2, 3])

In [6]: Tensor.ones(1,1,1).expand(2,3,2,3).shape
Out[6]: (2, 3, 2)
```

Before this `expand` in tinygrad would have hit an `exception` if I'm not mistaken...
Not sure if this current behavior is intended

geohotstan:
first try to fix #4070
~I didn't really refactor `_broadcast` and use it to match 🫣 hehe~
~kinda did the opposite with a new helper function~

~it's just that this `broadcast_shapes` helper function is useful in making `__getitem__` cleaner...~
~though the logic is kinda duplicated...~
ima sit on this for a little since this is still ugly.


test/test_tensor.py
```diff
@@ -421,6 +421,8 @@ def test_reshape(self):
       # cannot reshape from size 0 to size 1
       a = t.reshape(())
 
+  # TODO: should this be allowed?
+  @unittest.expectedFailure
   def test_expand(self):
     t = Tensor.full((3, 2, 0), 12).expand((6, 2, 0))
     assert t.shape == (6, 2, 0)
```
geohotstan, line 428: should this expand be allowed?
```
In [2]: a = jnp.full((3, 2, 0), 12)
In [3]: expanded = jnp.broadcast_to(a, (6,2,0))
ValueError: Incompatible shapes for broadcasting: (3, 2, 0) and requested shape (6, 2, 0)

In [5]: torch.full((3, 2, 0), 12).expand((6, 2, 0))
RuntimeError: The expanded size of the tensor (6) must match the existing size (3) at non-singleton dimension 0.  Target sizes: [6, 2, 0].  Tensor sizes: [3, 2, 0]
```

chenyuxyz: i don't quite get why jax and torch does not support it, but maybe that's fine. i would what's more natural from the implementation (meaning don't treat it as a special case)

tinygrad/tensor.py
```diff
@@ -860,6 +859,15 @@ def softplus(self, beta=1): return (1/beta) * (1 + (self*beta).exp()).log()
   def softsign(self): return self / (1 + self.abs())
 
   # ***** broadcasted elementwise mlops *****
+  @staticmethod
+  def _broadcast_tensors(*ts:Tensor, shape:Tuple[sint, ...]=()):
+    # TODO: what if shape is smaller than tensor shapes...
```
chenyuxyz, line 864: `ValueError`?

also fyi i will use this for setitem too to broadcast assign input into the assigned area

geohotstan: hehe I saw the TODO in your PR
I'll try to get this done quick so I don't block you.

geohotstan:
ended up copying https://github.com/google/jax/blob/main/jax/_src/numpy/util.py#L411 their `broadcast_to`
the `_pad_left` and `_broadcast_shape` just expands the shape
the `_broadcast_to` does validation and `reshape` `expand`

chenyuxyz:
cool thanks! nice cleanup for onnx too

PR #4101, opened 2024-04-07
Metal is broken again

uuuvn:
revert commandbuffer broke metal on real devices (but not on CI due to its metal being software emulated)


uuuvn:
before commandbuffer there were LoadOps.SYNC, after commandbuffer it did syncs but revert commandbuffer didn't added syncs back which seems fine for other devices but not for metal where transfers has to be manually synced

PR #4102, opened 2024-04-07
add lerp operation to tensor

davidgonmar:
As the title says. It has the same api as Torch

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -282,6 +282,9 @@ class TestOps(unittest.TestCase):
     helper_test_op([(45,35)], lambda x: x.round(), forward_only=True)
     helper_test_op(None, lambda x: x.round(), vals=[[1.499, 1.5, 1.501, 1.0, 2.1, 0.0, -5.0, -2.499, -2.5, -2.501]], forward_only=True)
     helper_test_op(None, lambda x: x.round(), vals=[[2.5, -1.5]], forward_only=True)
+  def test_lerp(self):
+    helper_test_op([(45,35), (45,35), (45,35)], lambda x,y,z: x.lerp(y,z), forward_only=True)
+    helper_test_op(None, lambda x,y,z: x.lerp(y,z), forward_only=True, vals=[[1.,2.,3.], [4.,5.,6.], 0.5])
 
   def test_tril(self):
     helper_test_op([(3,3)], lambda x: x.tril())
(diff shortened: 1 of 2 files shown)


geohot:
What uses this?

davidgonmar:
> What uses this?

I was working on trying to implement the new optimizer that everyone on twitter seems to be talking about (https://github.com/facebookresearch/schedule_free), and it uses linear interpolation (it is implemented in PyTorch).

tinygrad/tensor.py
```diff
@@ -830,6 +830,8 @@ def ceil(self: Tensor) -> Tensor: return (self > (b := self.trunc())).where(b+1,
   def floor(self: Tensor) -> Tensor: return (self < (b := self.trunc())).where(b-1, b)
   def round(self: Tensor) -> Tensor:
     return ((self > 0) == ((b := self.cast(dtypes.int32) / 2.0).cast(dtypes.int32) == b)).where((self - 0.5).ceil(), (self + 0.5).floor())
+  def lerp(self, end: Tensor, weight: Union[Tensor, float]) -> Tensor:
```
geohot, line 833: Fits on one line?

test/test_ops.py
```diff
@@ -282,6 +282,9 @@ def test_round(self):
     helper_test_op([(45,35)], lambda x: x.round(), forward_only=True)
     helper_test_op(None, lambda x: x.round(), vals=[[1.499, 1.5, 1.501, 1.0, 2.1, 0.0, -5.0, -2.499, -2.5, -2.501]], forward_only=True)
     helper_test_op(None, lambda x: x.round(), vals=[[2.5, -1.5]], forward_only=True)
+  def test_lerp(self):
+    helper_test_op([(45,35), (45,35), (45,35)], lambda x,y,z: x.lerp(y,z), forward_only=True)
```
geohot, line 286: Why forward_only?

davidgonmar: The rest of similar (math) function tests were forward_only, so I assumed it was because grads are checked already with 'primitives'. But I can make it also check the derivatives if you want

geohot: Does it work?

davidgonmar: wdym 'work'? The function itself is just some binary ops, it works fine. The test also passes.

geohot: The backward pass

davidgonmar: Ah, yeah, it works fine. It's also not problematic since it's just some binary ops together.

geohot: Test it?

davidgonmar: activated backward tests for it

PR #4104, opened 2024-04-07
Fix broken test

uuuvn:
This test was broken from its introduction in #4032


uuuvn:
Not sure why it passed in CI

PR #4105, opened 2024-04-07
Fixes for ops_kfd

fxkamd:
Two small fixes as I'm getting tinygrad working with KFD.

### DIFF
--- a/tinygrad/runtime/ops_kfd.py
+++ b/tinygrad/runtime/ops_kfd.py
@@ -374,7 +374,8 @@ class KFDDevice(Compiled):
     with open(f"{KFDDevice.gpus[self.device_id]}/gpu_id", "r") as f: self.gpu_id = int(f.read())
     with open(f"{KFDDevice.gpus[self.device_id]}/properties", "r") as f: self.properties = {line.split()[0]: int(line.split()[1]) for line in f}
     self.drm_fd = os.open(f"/dev/dri/renderD{self.properties['drm_render_minor']}", os.O_RDWR)
-    self.arch = f"gfx{self.properties['gfx_target_version']//100}"
+    target = int(self.properties['gfx_target_version'])
+    self.arch = "gfx%d%x%x" % (target // 10000, (target // 100) % 100, target % 100)
     kio.acquire_vm(KFDDevice.kfd, drm_fd=self.drm_fd, gpu_id=self.gpu_id)
 
     if KFDDevice.event_page is None:
--- a/tinygrad/runtime/ops_kfd.py
+++ b/tinygrad/runtime/ops_kfd.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import Tuple, Any
+from typing import Tuple, List, Any
 import os, fcntl, ctypes, functools, re, pathlib, mmap, struct, errno
 from tinygrad.device import Compiled, LRUAllocator, Compiler, CompilerOptions
 from tinygrad.buffer import BufferOptions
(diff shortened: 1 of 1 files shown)


geohot:
Linter failing on 150 char line limit, we have a pre-commit hook that should catch things like this. Install with `pre-commit install`

tinygrad/runtime/ops_kfd.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from typing import Tuple, Any
+from pathlib import Path
```
nimlgen, line 3: pathlib is imported already

tinygrad/runtime/ops_kfd.py
```diff
@@ -363,10 +371,12 @@ def _wait_signal(self, signal:hsa.amd_signal_t, timeout=60000):
...
     self.device_id = int(device.split(":")[1]) if ":" in device else 0
-    with open(node_sysfs_path(self.device_id+1, "gpu_id"), "r") as f: self.gpu_id = int(f.read())
-    with open(node_sysfs_path(self.device_id+1, "properties"), "r") as f: self.properties = {line.split()[0]: int(line.split()[1]) for line in f}
+    first_gpu = find_first_gpu_node()
+    with open(node_sysfs_path(self.device_id+first_gpu, "gpu_id"), "r") as f: self.gpu_id = int(f.read())
+    with open(node_sysfs_path(self.device_id+first_gpu, "properties"), "r") as f: self.properties = {line.split()[0]: int(line.split()[1]) for line in f}
     self.drm_fd = os.open(f"/dev/dri/renderD{self.properties['drm_render_minor']}", os.O_RDWR)
-    self.arch = f"gfx{self.properties['gfx_target_version']//100}"
+    target = int(self.properties['gfx_target_version'])
+    self.arch = "gfx%d%x%x" % (target // 10000, (target // 100) % 100, target % 100)
```
nimlgen, line 379: Should we change that? We get major and minor which are passed to the compiler (like `gfx1100`), why should we care about revision?

tinygrad/runtime/ops_kfd.py
fxkamd: Thanks, I'll just use pathlib.Path.

fxkamd:
> Looks good.
> 
> Linter failing on 150 char line limit, we have a pre-commit hook that should catch things like this. Install with `pre-commit install`

I've got most pre-commit tests passing now. I get one failure, not sure if pip installed the wrong version of pytest:

'''strip whitespace.........................................................Passed
ruff.....................................................................Passed
mypy.....................................................................Passed
docs2....................................................................Passed
select GPU tests.........................................................Passed
subset of tests..........................................................Failed
- hook id: tests
- exit code: 4

ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: -n=4
  inifile: None
'''

geohot:
`pip -e install .[testing]` should install all you need, I think you are missing `pytest-xdist`

Sorry, we really need to write better docs, it's just in a footnote at the bottom of the readme now.

tinygrad/runtime/ops_kfd.py
```diff
@@ -17,7 +17,14 @@
 libc.munmap.argtypes = [ctypes.c_void_p, ctypes.c_size_t]
 libc.munmap.restype = ctypes.c_int
 
-def node_sysfs_path(node_id, file): return f"/sys/devices/virtual/kfd/kfd/topology/nodes/{node_id}/{file}"
+def is_usable_gpu(gpu_id):
+    try:
```
geohot, line 21: Hmm, why isn't our linter catching spacing here?

tinygrad/runtime/ops_kfd.py
```diff
@@ -17,7 +17,14 @@
...
+      with gpu_id.open() as f:
+        return int(f.read()) != 0
+    except OSError:
+      return False
+gpus = [g.parent for g in pathlib.Path("/sys/devices/virtual/kfd/kfd/topology/nodes").glob("*/gpu_id") if is_usable_gpu(g)]
```
geohot, line 26: This should be in the first call to KFDDevice __init__ (with `if KFDDevice.kfd == -1:`) so it doesn't run on import.

PR #4106, opened 2024-04-08
search: add a BEAM_COMPARE env to optionally not compare to hc/tc

flammit:
setting BEAM_COMPARE=0 will prevents additional memory allocation needed to do the timing tests, until the real buffers can be used


flammit:
ugh... too many lines added.

PR #4107, opened 2024-04-08
search: add a BEAM_COMPARE env to optionally not compare to hc/tc

flammit:
setting BEAM_COMPARE=0 will prevents additional memory allocation needed to do the timing tests, until the real buffers can be used

This is useful for llama2-70B as well as unet3d for doing offline beam-searches when memory utilization is tight to populate the caches using `extra/optimization/search.py` on the output of LOGOPS, then running it live with `BEAM_COMPARE=0 BEAM=N` ensuring that no extra memory will be allocated during the run.

### DIFF
--- a/extra/optimization/search.py
+++ b/extra/optimization/search.py
@@ -23,8 +23,8 @@ if __name__ == '__main__':
     with open(args.file, 'r') as file:
      ast_strs = file.readlines()
 
-  for ast_str in ast_strs:
-    print(f"optimizing ast={ast_str}")
+  for i, ast_str in enumerate(ast_strs):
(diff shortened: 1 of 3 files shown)


chenyuxyz:
i like being able to beam search llama 70B so i am supportive on this direction, but can you make separate line saving refactor and the real search buffer api change? this has too many moving parts and pretty hard to review

chenyuxyz:
`Buffer` now supports an explicit `allocate` method to allocate real memory. I think you can update `bufs_from_lin` to defer buffer allocation. So instead of None, raw_bufs can have unallocated buffers

PR #4112, opened 2024-04-08
refactor: create ASTs with inner dependencies

Qazalin:
Directly connected realizes can fuse,

eg. the 3 `E_128_10` kernels in Adam: https://tiny-tools-client.vercel.app/?id=6bbb1e19fad04b2db5c36448d015e36a
Screenshot 2024-04-08 at 7 17 21 PM

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -22,8 +22,8 @@ class _LBScheduleItem:
   var_vals: Dict[Variable, int]
 
 # recursively create a lazyop
-def _recursive_lazyop(buf:LazyBuffer, membufs:List[LazyBuffer], var_vals:Dict[Variable, int], st:ShapeTracker,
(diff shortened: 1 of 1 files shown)


tinygrad/engine/schedule.py
```diff
@@ -22,8 +22,8 @@ class _LBScheduleItem:
   var_vals: Dict[Variable, int]
 
 # recursively create a lazyop
-def _recursive_lazyop(buf:LazyBuffer, membufs:List[LazyBuffer], var_vals:Dict[Variable, int], st:ShapeTracker,
-                      realizes:Set[LazyBuffer], cache, first=True, assign_to:Optional[LazyBuffer]=None, assign_idx:Optional[int]=None) -> LazyOp:
+def _recursive_lazyop(buf:LazyBuffer, membufs:List[LazyBuffer], outbufs:Tuple[LazyBuffer, ...], var_vals:Dict[Variable, int], st:ShapeTracker,
```
geohot, line 25: What's the difference between buf and outbufs? I don't understand this API

geohot:
I see why this has to be done. I think the API can be cleaner (membufs should be inbufs?), but will merge for now.

PR #4114, opened 2024-04-08
pickle schedules

Qazalin:
Currently we're hitting "need alloc" on Buffer

Screenshot 2024-04-08 at 8 41 35 PM


geohot:
Oh, it's because it's on an EXT device (copy in from numpy), that's the core issue.

geohot:
Should be fixed in #4120

test/test_pickle.py
```diff
@@ -37,5 +39,13 @@ def add(a, b): return a+b+1
     out = add_fxn(x, y)
     np.testing.assert_equal(out.numpy(), 3)
 
+  def test_pickle_schedule(self):
+    a = Tensor([1,2])
+    out = a + 2
+    sched = create_schedule([out.lazydata])
+    pk = pickle.dumps(sched)
+    sched_pk = pickle.loads(pk)
+    assert sched_pk[-1].ast[0].op is BinaryOps.ADD
```
geohot, line 48: Pickle works, but I don't think this assertion is right

PR #4116, opened 2024-04-09
optionally use a copy kernel instead of SDMA

geohot:
~~Broken, needs a sync.~~

Is it stable, does it work? https://wandb.ai/geohot/tinygrad-resnet-tiny5/runs/d7pvn16z

Nope, seems just as broken, and confirmed it isn't using SDMA queue.


geohot:
Works with CUDA P2P

PR #4119, opened 2024-04-09
wmma: clean up the TC alias pattern and remove TC=2 behavior

flammit:
other than the removal of the TC=2 behavior, all other behavior and performance should be the same


chenyuxyz:
I think we keep TC=2 behavior for now, it's a mid step for debugging TC. Also you really want to remove it you will need update the github action too

flammit:
SGTM!

PR #4123, opened 2024-04-10
Changed all instances of fetch_mnist to mnist

aaronstevenson408:
not sure if this is what you were looking for from the discord, but updated examples and such to mnist()


geohot:
You didn't test this at all

PR #4133, opened 2024-04-10
Add copy_from_fd to MallocAllocator

uuuvn:
Just so that it would be tested in CI. I thought about creating `TestAllocator` in test but it seems more complicated

Introducing bug to copy_from_fd (changed ._buf to ._buff) triggers at least test/test_multitensor.py::TestMultiTensor::test_data_parallel_resnet and test/test_multitensor.py::TestMultiTensor::test_data_parallel_resnet_train_step  (didn't test further because CLANG is very slow


uuuvn:
Looks like there are some issues on linux, don't have much time to investigate now so converting into draft

geohot:
This is not where this belongs

PR #4134, opened 2024-04-10
Add driving monitoring model to benchmarks

terafo:


chenyuxyz:
i will check tiny7 issue.

can you update the benchmark script to handle the mac benchmark breakage due to more than 32 buffers in kernel (try except the return early if model failed is fine)?

PR #4136, opened 2024-04-10
Fuzz all permutations of schedule

Qazalin:
`FUZZ_SCHEDULE=1 DEBUG=2 python3 test/test_multitensor.py TestMultiTensor.test_simple_add_X`

need to take more time with:

- [x] What is the right abstraction? fuzzer.py could merge with realize.py -> moved to /tests
- [x] Seedless`.randn` between runs - eg `TestOps.test_multinomial` -> uses the same seed
- [x] Fix assign fuzzing


geohot:
I don't think the fuzzer belongs in tinygrad proper, shouldn't it be in tests?

Qazalin:
status:
This is in a good spot now - time to stress test multi output

Qazalin:
30 of the possible toposorts for `test.models.test_mnist.TestMNIST.test_adam_onestep` - each color is a LazyBuffer
Screenshot 2024-04-13 at 10 47 04 PM
https://tiny-tools-client.vercel.app/?id=c44a8348fff042c5801ff899c4f39913

tinygrad/engine/schedule.py
```diff
@@ -206,6 +205,7 @@ def create_schedule_with_vars(outs:List[LazyBuffer], seen:Optional[Set[LazyBuffe
   graph: DefaultDict[LazyBuffer, List[LazyBuffer]] = defaultdict(list)
   in_degree: DefaultDict[LazyBuffer, int] = defaultdict(int)
   for key, lsi in prescheduled.items():
+    if key not in in_degree: in_degree[key] = 0
```
geohot, line 208: Hmm, so defaultdict should do this, but I see why there's reasons you want to do it explicitly (for iteration over keys later). Do you still need the defaultdict with this line, or can you make in_degree a normal dict?

Qazalin: `assign_targets` needs it, you could change `in_degree[assign]` without having visited the node
```py
    # realize outputs before a parent is assigned to
    parents_assigns = set(schedule_targets[assign_targets[x]].outputs[0] for x in lsi.inputs if x in assign_targets)
    for assign in parents_assigns:
      graph[key].append(assign)
      in_degree[assign] += 1
```

geohot:
Testing on my Mac. 

`DEBUG=2 PYTHONPATH="." FUZZ_SCHEDULE=1 python3 examples/beautiful_mnist.py`

```
...
*** METAL   12346 r_512_2_5_5_2_2_16_4_4                 arg   2 mem 107.76 GB tm    186.83us/  6734.32ms (  140.31 GFLOPS,  237.13 GB/s)
*** METAL   12347 r_32_256_2_20_20                       arg   8 mem 107.76 GB tm    179.71us/  6734.50ms (  386.52 GFLOPS,  255.28 GB/s)
*** METAL   12348 r_32_256_2_20_20n1                     arg   7 mem 107.76 GB tm    175.42us/  6734.68ms (  298.88 GFLOPS,  261.52 GB/s)
*** METAL   12349 r_32_256_2_400n2                       arg   5 mem 107.76 GB tm     58.25us/  6734.74ms (  712.46 GFLOPS,  450.04 GB/s)
```

I think there's a memory leak, it grew to 107 GB

test/external/fuzz_schedule.py
```diff
@@ -0,0 +1,85 @@
...
+          if x.device == "NPY": rawbufs[x] = x.buffer
+          # copy the pre realized input
+          else: rawbufs[x] = Buffer(x.buffer.device, x.buffer.size, x.buffer.dtype, initial_value=x.buffer.as_buffer())
+      schedules[i+1].append(ScheduleItem(ps.ast, tuple(rawbufs[x] for x in ps.outputs if x.size != 0),
+                                         tuple(rawbufs[x] for x in ps.inputs if x.size != 0)))
+
+  # run all schedules with the same seed
+  seed = Tensor._seed
+  for i, schedule in enumerate(schedules):
+    if DEBUG >= 2: print(colored(f"testing premutation {i}", "yellow"))
```
geohot, line 46: Typo. Also maybe this should be DEBUG >= 1, since this is the main thing you care about if you are running this option.

Qazalin:
the sort and kernel runs are generally fast, validation is slow:
```py
  # assert all LazyBuffers realized correctly
  for lb, bufs in outputs.items():
    ground_truth = np.frombuffer(bufs[0].as_buffer(), bufs[0].dtype.np)
    for buf in bufs[1:]:
      try: np.testing.assert_allclose(np.frombuffer(buf.as_buffer(), buf.dtype.np), ground_truth, atol=1e-2, rtol=1e-2)
      except AssertionError as e:
        print(f"COMPARE FAILED FOR {lb}")
        raise e
 ```
 
 LRU=0 helps - but it's doing a copyout for every ScheduleItem.outputs.
 I wonder if we could make validation a GPU kernel

geohot:
I don't think limiting the paths is the solution. Why is the memory growing with each run?

Screenshot 2024-04-16 at 10 25 20 AM

geohot:
Cool, memory issue looks fixed. This good to merge?

Qazalin:
This diff creates new Buffers for each permutation - makes fuzzing multiple toposorts in parallel possible.

geohot:
Yea, totally fine as long as they are being freed, which it looks like they are.

PR #4138, opened 2024-04-10
Rust backend

jaredeh:
Adding Rust backend.

- LOC neutral (assuming can park rust.py in ./extra)
- Tests mostly pass
- Includes examples/export_code_from_onnx.py which creates a zero dependency pure Rust rlib crate
- Performance slightly better than Clang (worse compile times, need to compare 2nd run)
- Core changes required: renderer needs correct buffer size to build functions

The use case is to take a model from TinyGrad, onnx, pytorch, etc and be able to export a pure Rust crate to implement that model.  This is ideal for my use case were I am dealing with small models in userspace apps and I will be very useful for embedded.  Sure there are other ways to do this they all end up with crappier code, more dependencies, and/or worse performance.

You can test this out fairly easily with ./examples/export_code_from_onnx.py --language rust --crate --save

Most tests pass but not all.  There are a few areas where there are still some differing casting assumptions to hammer out.  Which I will commit to work through.  I'm still working on how to support, variables.  If I don't know the range of a variable that guides a loop the rust compiler doesn't know now to ensure you're not doing a buffer overflow.

Performance is a tiny bit little faster than CLANG (on second run, compile times are worse) when compared using ./test/test_speed_v_torch.py, but line by line there are shocking difference where Rust is faster by 20X or more and some where C is faster by 20X.  Honestly being able to look at how they compile out to different machine code to achieve those differences might justify adding Rust by itself.

There are two core changes which Rust support needs, especially to fit in the CDLL usage that is so useful for testing here.  Well really just one.  When I render a function I need to properly size any input/output buffers.  Rust can infer buffer sizes and check for correctness, but only if those sizes are all known at compile time.  The way you compile for CDLL doesn't allow the sizes to be known so you have to specify.  The knowledge of how big the inputs and outputs are is know at the time of the call of get_runner() so I added that.  However, not all buffers are inputs or outputs and so we also need the buffer size added to the UOpsGraph when a global is added, so I stash a size field in the args tuple for globals.  Originally I hoped only to use that but sometimes inputs/outputs sizes disagrees with UOpsGraph size, and in those cases input/outputs agree.

As a side note as we start looking for optimizations it would see this information being available at render time might be useful for other architectures too.


jaredeh:
huh.  So the CI doesn't like it much.  I thought I tested that out better.

Please hold...

geohot:
Won't merge, for the reasons discussed in Discord.

CLANG is nowhere near its performance limit, would much rather spend lines to make CLANG generate more performant C code.

PR #4142, opened 2024-04-11
use BEAM=3 in benchamrk CI

chenyuxyz:
how long does it take, and is the output faster?


chenyuxyz:
very minor improvement, not worth the extra 50% search time

PR #4144, opened 2024-04-11
Resnet fp16 training with fp32 master weight copy

chaosagent:
Need also expand backwards accumulate in fp32.

Be careful with the eval jit -- there is a bug where LAZYCACHE can make the .cast(dtypes.default_float) happen only once and ignore all weight updates!


chenyuxyz:
worked with 2 warmup epochs and 45 total epochs

chenyuxyz:
https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/bz6xk6wo?nw=nwuserchenyuxyz

PR #4150, opened 2024-04-11
Rewritten SVG Logos

MaximilianEmel:
Rewrote the svg logos to use polygons instead of rects, as when the edges of them overlapped, it caused protrusions to appear. Also added an attribute that will cause the logo to render better.


geohot:
Cool, thanks!

PR #4152, opened 2024-04-12
remove CAST_BEFORE_VIEW

chenyuxyz:
testing perf, also this might have issue with assign?

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -53,7 +53,7 @@ jobs:
     - name: Run GPT2 w HALF
       run: JIT=1 HALF=1 python3 examples/gpt2.py --count 10 --temperature 0 --timing | tee gpt2_half.txt
     - name: Run GPT2 w HALF/BEAM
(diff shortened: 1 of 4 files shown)


chenyuxyz:
@chaosagent do you still want this? I don't see perf impact on benchmark so prefer to remove this unless there's big perf gain

chaosagent:
I am OK with removing. I think you will see perf regression on BEAM=2 convolution backwards. Current master doesn't trigger TC that much anyways. I have an alternative approach for that.

chenyuxyz:
okay let me BEAM=2 resent on master too

chaosagent:
oh also, check memory usage. i think this saved some memory too.

chenyuxyz:
`HSA=1 DEFAULT_FLOAT=HALF WARMUP_EPOCHS=2 BS=768 GPUS=6 BENCHMARK=10 MODEL=resnet python3 examples/mlperf/model_train.py` uses ~87.4GB on master and this branch so no memory diff.

chenyuxyz:
both this pr and master have 430ms step time with default changed to HALF. i think it's safe to delete this.

chaosagent:
Sounds good!

PR #4154, opened 2024-04-12
Use `tinygrad.features.dataset` implementation of `mnist()`

c-nnr:
`tests/models/test_mnist.py` currently using the `train` method from `extra/training.py`. This requires the data to be in a numpy array as it selects batches using `np.random.randint` and `Tensor` cannot be indexed using a numpy array.

Happy to open another PR to update `extra/training.py` to use `Tensor.randint` to generate the minibatch and add some speed comparisons

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -602,6 +602,13 @@ class TestAutoCastType(unittest.TestCase):
 
     dtypes.default_float = old_default_float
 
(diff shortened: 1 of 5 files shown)


geohot:
Please remove all changes that aren't strictly required to use the new function, it's hard to review.

PR #4157, opened 2024-04-12
assert if expr_idxs return might be outside of int32

chenyuxyz:


chenyuxyz:
will work on optionally upcast idx to use int64 next

PR #4158, opened 2024-04-12
fix return dtype of getitem Tensor indexing

chenyuxyz:
the use of sum can auto-upcast the result. fixed by using the data dtype as the acc_dtype


chenyuxyz:
@geohotstan fyi this might have contributed to speed diff in beautiful_mnist

PR #4165, opened 2024-04-13
pm4 debug

nimlgen:
add more syncs and tests passed. will debug from this working state.


nimlgen:
tests pass. hlb works, testing resnet

PR #4166, opened 2024-04-13
support grayscale image for resnet example

KuhakuPixel:


geohot:
No test. We should probably just delete a lot of these examples.

Issue #4139, opened 2024-04-10
  AttributeError: 'DiskDevice' object has no attribute 'fd' introduced in #4129


Maroc-OS:
#4129 introduced an AttributeError when trying to close non existing file descriptor

```
tinygrad_latest/test/test_copy_speed.py::TestCopySpeed::testCopySHMtoDefault
  /usr/local/lib/python3.11/site-packages/_pytest/unraisableexception.py:78: PytestUnraisableExceptionWarning: Exception ignored in: <function Buffer.__del__ at 0x108b44f40>
  
  Traceback (most recent call last):
    File "/Users/mac/Desktop/tinygrad/tinygrad_latest/tinygrad/buffer.py", line 43, in __del__
      self.allocator.free(self._buf, self.nbytes, self.options)
    File "/Users/mac/Desktop/tinygrad/tinygrad_latest/tinygrad/device.py", line 107, in free
      self._free(opaque, options if options is not None else BufferOptions())
    File "/Users/mac/Desktop/tinygrad/tinygrad_latest/tinygrad/runtime/ops_disk.py", line 23, in _free
      def _free(self, buf, options): self.device._might_close()
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/mac/Desktop/tinygrad/tinygrad_latest/tinygrad/runtime/ops_disk.py", line 85, in _might_close
      os.close(self.fd)
               ^^^^^^^
  AttributeError: 'DiskDevice' object has no attribute 'fd'
```

PR #4170, opened 2024-04-14
fix AttributeError introduced by #4129 and fix test case

Maroc-OS:
FIX for #4139

This one have no other changes.

The test case problem was the file does not get deleted if we got an exception before we reach `unlink()`.
Then when trying to test again we get a `FileExistError` Exception, and the test fail.


geohot:
Error fixed with only the self.fd check

PR #4171, opened 2024-04-14
GPT2 compile to C

atalw:
Added 2 new things:
- An option to `fread` weights in codegen instead of preloading them (GPT2 weights created a file too big for clang to compile)
- Use model parameter names for buffers in codegen for readability

GPT2 runs but it outputs junk tokens right now. 2 reasons I could think of:
- There is a custom op random function that doesn't have codegen available. Is it possible to compile it to C? (skipped handling [here](https://github.com/tinygrad/tinygrad/pull/4171/files#diff-edd2f864132528e159fd3c7af9bf6fe907a63c9665af1b25366164d0cc7f8929R26))
- Layernorm weights are not included in the JIT cache. Is this expected? (comment [here](https://github.com/tinygrad/tinygrad/pull/4171/files#diff-4f6348e1e7ed106d0b3cdcac30d049769ca3615999cd647d46747d22c4e9561aR44))

Also, would I need to write the token encode and decode functions manually in C? Or is there a way to autogen that I've missed?


atalw:
Stale

PR #4172, opened 2024-04-14
remove limit_dims_to_max

chenyuxyz:
do we still need this? need to clean up if so


chenyuxyz:
eh cuda needs this, will revisit after we have tinybox green on ci

PR #4173, opened 2024-04-14
Multireduce Kernels - prereq refactor

0xtimmy:
A refactor of the linearlizer to help with fusing multiple reduceops into one kernel

It makes the process of rendering a reduction it's own function, so that when there are multiple reducops in one kernel, there is only one block of code responsible 

introduces a new function `render_reduceop()` which does the rendering and moves `render_loop()` out of the `linearize()` so that it can be accessed by `render_reduceop()`

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -170,12 +170,12 @@ class Linearizer(Kernel):
         self.const(x.max+1) if isinstance(x.max, int) else cast(Node, x.max+1).render(self.render_ops, self)), cachable=False) for x in xx if not isinstance(x, NumNode) and x.expr is not None}  # noqa: E501
       self.loop_uops.update(new_loops)
       return tuple(new_loops.values())
-  
-  def render_reduceop(self, reduceop, global_idxs, local_idxs, upcast_idxs, full_upcast_idxs):
+
+  def render_reduceop(self, reduceop: LazyOp, global_idxs, local_idxs, upcast_idxs, full_upcast_idxs):
     # define indecies
     reduce_idxs = [Variable(f"ridx{i}", 0, self.full_shape[i]-1) for i in range(self.first_reduce+self.group_for_reduces, self.shape_len-self.upcasted)]  # noqa: E501
     fake_reduce_idxs = [x*0 for x in reduce_idxs]
-      
+
     # define accumulator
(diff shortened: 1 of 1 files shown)


tinygrad/codegen/linearizer.py
```diff
@@ -163,6 +163,143 @@ def global_store(self, i:int, idxs:List[Node], store:List[UOp]) -> List[UOp]:
       else: stores.append(self.uops.add(UOps.STORE, None, (buf_uop, rendered_idx, var, valid.render(self.render_ops, self))))
     return stores
 
+  # render loop
+  def render_loop(self, xx:List[Variable]) -> Tuple[UOp, ...]:
+      new_loops = {x.expr:self.uops.add(UOps.LOOP, dtypes.int32, (
```
chenyuxyz, line 168: indent size should be 2

tinygrad/codegen/linearizer.py
```diff
@@ -245,150 +374,25 @@ def render_loop(xx:List[Variable]) -> Tuple[UOp, ...]:
       self.loop_uops.update({x.expr:self.uops.add(UOps.SPECIAL, dtypes.int32, (), (len(loop_global_idxs)-1-i, x.expr, x.max+1)) for i,x in enumerate(loop_global_idxs)})  # noqa: E501
       self.loop_uops.update({x.expr:self.uops.add(UOps.SPECIAL, dtypes.int32, (), (len(loop_local_idxs)-1-i, x.expr, x.max+1)) for i,x in enumerate(loop_local_idxs)})  # noqa: E501
     else:
-      render_loop(loop_global_idxs+loop_local_idxs)
+      self.render_loop(loop_global_idxs+loop_local_idxs)
 
     # parse AST
-    loaded_buffers = {}
+    self.loaded_buffers = {}
```
chenyuxyz, line 380: why change this?

0xtimmy: it needed to be accessed while rendering the reduceops and it seemed like just as much a property of the kernel as much as `load_cache`

chenyuxyz: can it be passed into the function directly? ideally we want to keep object state minimal. load_cache is cleared after use too

0xtimmy: bet

PR #4179, opened 2024-04-15
Replicate llm.c in tinygrad

geohot:
https://github.com/karpathy/llm.c


geohot:
Hmm, why is `__len__` breaking onnx?

geohotstan:
> Hmm, why is `__len__` breaking onnx?

when there's an operation like 
`In [4]: np.array(3) * Tensor.ones(3,3)`
`__len__` is called on the Tensor object.

On master when numpy array is a scalar, I think numpy sees the default object `__len__` and just convert the scalar to an int to do the operation.
```
In [4]: np.array(3) * Tensor.ones(3,3)
Out[4]: <Tensor <LB METAL (3, 3) contig:False ShapeTracker(views=(View(shape=(3, 3), strides=(0, 0), offset=0, mask=None, contiguous=False),))> on METAL with grad None>
```
so the tests pass... 

on this branch currently
```
In [4]: np.array(3) * Tensor.ones(3,3)
Out[4]: array([], shape=(3, 3, 0), dtype=float64)
```

Doing operations using numpy arrays with Tensor is not a good idea haha, my bad.
Also there is `tolist` and `item` now. Maybe moving onnx to use that instead of numpy is a good idea.

sorry for blocking. Can skip tests to get it merged. I'll fix them when I wake up tomorrow.

geohot:
No worries, glad it exposed a bug. Want to do the simplest fix?

geohot:
Interesting what torch does:

```
>>> torch.ones(3,3) * np.array(3)
tensor([[3., 3., 3.],
        [3., 3., 3.],
        [3., 3., 3.]])
>>> np.array(3) * torch.ones(3,3)
TypeError: unsupported operand type(s) for *: 'numpy.ndarray' and 'Tensor'
```

geohotstan:
> Interesting what torch does:
> 
> ```
> >>> torch.ones(3,3) * np.array(3)
> tensor([[3., 3., 3.],
>         [3., 3., 3.],
>         [3., 3., 3.]])
> >>> np.array(3) * torch.ones(3,3)
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> TypeError: unsupported operand type(s) for *: 'numpy.ndarray' and 'Tensor'
> ```

I think `__add__` takes precedence over `__radd__` so `np.array(3) * torch.ones(3,3)` goes down numpy codepath instead of Torch's

I looked a bit into why np.array first doesn't work

It appears that numpy does a low level `PySequence_check` that checks `__getitem__` and `__len__` of Tensor a bunch of times to see if it's type compatible. [src](https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects#comment80414178_40380014) (this is old but I tried debugging np * tinygrad_Tensor and it does call `__len__` and `__getitem__` a bunch of times)

for pytorch, `PySequence_check` returns False so the operation is not Type compatible. [src](https://github.com/pytorch/pytorch/issues/99646) (I tried the check in the issues for tinygrad's Tensor and it returned True)

I'm not 100% sure about this 

> (numpy * Tensor should assert? what does PyTorch do?)

I don't think the exception raised for pytorch is at the python level, so the debugger just says the `*` emits the TypeError.

PR #4183, opened 2024-04-16
[DRAFT] wmma: widen TC application by using PADTO on TC axes when necessary

flammit:
I'd be curious to see what the benchmarks say - it might be better to gate this behind BEAM>0 (i.e. keep normal divisibility checks when not searching, but if searching, then widen the applicability).


flammit:
yeah as expected, the non-BEAM benchmarks are way worse... will either gate the PADTO+TC behind beam or close

chenyuxyz:
this needs more tests, i think it might have bugs because i am seeing lots of local after TC and getting way larger than spec flops and mem number

flammit:
closing - does not work

PR #4187, opened 2024-04-16
[WIP] Variables can load tensors with idx offset

patosai:
Before, all bufs were loaded in the graph via recursively loading lazybuffer args, but tensors inside variables have to be loaded in a different way


tinygrad/ops.py
```diff
@@ -50,6 +50,7 @@ class LazyOp:
   op: Op
   src: Tuple[LazyOp, ...] = ()
   arg: Any = None
+  extra_bufs: Any = None
```
geohot, line 53: It's very heavyweight to add something here. Why do we need this?

patosai: oops missed this. still WIP, this was an easy way to get buffer loading working for now. idk if this is the way to go yet, 1 value per var means a large embedding kernel has a ton of vars and may not compile, multiple values per var (either in list or tensor form) means a huge refactor of how variables work. so I've paused this train of thought for now

PR #4192, opened 2024-04-16
tolist to return multidimensional list

geohotstan:
I tried hacking around indexing the memoryview and reconstructing the list with the right shape for awhile, only to find out that `memoryview.tolist()` exists 😄 

is this method valid?

after casting the memoryview, it seems that tolist resolves to a multidimensional list

### DIFF
--- a/test/test_tensor_data.py
+++ b/test/test_tensor_data.py
@@ -31,6 +31,14 @@ class TestTensorData(unittest.TestCase):
     assert dat[0, 0] == 1
     assert dat[1, 1] == 4
 
+  def test_data_const(self):
+    a = Tensor(3, dtype=dtypes.int32)
+    dat = a.data()
+    assert dat.format == "i"
+    assert dat.itemsize == 4
+    assert dat.tolist() == 3
+    assert dat.shape == ()
+
   def test_data_float32(self):
     a = Tensor([[1,2.5],[3,4]], dtype=dtypes.float32)
     dat = a.data()
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -192,7 +192,7 @@ class Tensor:
     assert self.dtype.fmt is not None, f"no fmt dtype for {self.dtype}"
     assert self.numel() == 1, "must have one element for item"
     return self._data().cast(self.dtype.fmt)[0]
-  def tolist(self) -> List[ConstType]: return list(self.data())
+  def tolist(self) -> Sequence[ConstType]: return self.data().tolist()
   def numpy(self) -> np.ndarray:
     if self.dtype == dtypes.bfloat16: return self.float().numpy()
     assert self.dtype.np is not None, f"no np dtype for {self.dtype}"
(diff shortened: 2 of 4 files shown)


chenyuxyz:
why is the return type not `List`?

geohotstan:
> why is the return type not `List`?

mypy failure:
```
tinygrad/tensor.py:195: error: Incompatible return value type (got "list[int]", expected "list[float | int | bool]")  [return-value]
tinygrad/tensor.py:195: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
tinygrad/tensor.py:195: note: Consider using "Sequence" instead, which is covariant
```

for some reason memoryview's tolist typing is `def tolist(self) -> list[int]: ...`
even when memoryview.tolist can return non int values
image

🤔

geohot:
LGTM, any reason this is draft?

tinygrad/tensor.py
```diff
@@ -191,13 +191,13 @@ def _data(self) -> memoryview:
   def data(self) -> memoryview:
     assert self.dtype.fmt is not None, f"no fmt dtype for {self.dtype}"
     assert all_int(self.shape), f"no data if shape is symbolic, {self.shape=}"
-    return self._data().cast(self.dtype.fmt, self.shape if len(self.shape) else (1,))
+    return self._data().cast(self.dtype.fmt, self.shape)
```
geohotstan, line 194: a little unsure of this change.
But an empty shape `()` works and it matches torch.

chenyuxyz: do we already have a test for this?

tinygrad/tensor.py
```diff
@@ -191,13 +191,13 @@ def _data(self) -> memoryview:
...
   def item(self) -> ConstType:
     """Returns the value of this tensor as a standard Python number."""
     assert self.dtype.fmt is not None, f"no fmt dtype for {self.dtype}"
     assert self.numel() == 1, "must have one element for item"
     return self._data().cast(self.dtype.fmt)[0]
-  def tolist(self) -> List[ConstType]: return list(self.data())
+  def tolist(self) -> Sequence[ConstType]: return self.data().tolist()
```
chenyuxyz, line 200: add a TODO note pointing to source if it really needs to be `Sequence` for now. this makes mypy happy but user unhappy

tinygrad/tensor.py
geohotstan: There isn't.
I added a test in `test_tensor_data` but I dunno is it's right :s

test/test_tensor_data.py
```diff
@@ -31,6 +31,14 @@ def test_data_nested(self):
     assert dat[0, 0] == 1
     assert dat[1, 1] == 4
 
+  def test_data_const(self):
+    a = Tensor(3, dtype=dtypes.int32)
+    dat = a.data()
+    assert dat.format == "i"
+    assert dat.itemsize == 4
+    assert dat.tolist() == 3
```
chenyuxyz, line 39: slightly weird that `.tolist()` returns a non-list, will let @geohot decide

geohot: It's fine, it matches PyTorch

PR #4195, opened 2024-04-16
Clang openmp

jumbojets:
OpenMP support for the CLANG runtime. Default is on, but `OMP=0` will turn it off.

On mac, this assumes that `/usr/local` contains OpenMP (see: https://iscinumpy.gitlab.io/post/omp-on-high-sierra/)

TODO:
- [x] pass tests in CI (might be some correctness/omp install/linux issues).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -380,7 +380,7 @@ jobs:
           path: ~/.cache/tinygrad/downloads/
           key: downloads-cache-${{ matrix.backend }}-${{ env.DOWNLOAD_CACHE_VERSION }}
       - name: Set env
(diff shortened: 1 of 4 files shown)


jumbojets:
I changed default to `OMP=0` just so a user that's getting started with TG doesnt need to have libomp installed and configured correctly

I also `noqa: E551`ed a couple of lines that I made longer than 150 chars. Ill try to address those tomorrow, but other than that, this PR should be ready

jumbojets:
On M1 Max

`OMP=0 CLANG=1 DEBUG=0 OPTLOCAL=1 python3 examples/llama.py --prompt "Hello." --size 7B --count 10 --temperature 0 --timing`
Screenshot 2024-04-16 at 11 51 31 PM

`OMP=1 CLANG=1 DEBUG=0 OPTLOCAL=1 python3 examples/llama.py --prompt "Hello." --size 7B --count 10 --temperature 0 --timing`
Screenshot 2024-04-16 at 11 51 36 PM

geohot:
This seems hacked in with "first loop prefix"

It should function more like global/local size I think.

jumbojets:
Good point. I agree that using the global/local size abstraction feels intuitive here, particularly for specifying certain OpenMP parameters. I assume you refer to global_size/local_size parameters in the "ops_*.py" files and not in "kernel.py".

For reference, the OpenCL runtime's usage:
https://github.com/tinygrad/tinygrad/blob/7bc862767c5888e3ebe24cae116abf65fcc7713c/tinygrad/runtime/ops_gpu.py#L50

However, it isn't clear to me how this will address the rendering of the OpenMP pragmas--which needs to be known at "compile" time of the CLANG program.

Passing the raw pragma text from `ops_clang.py` to `cstyle.py` at compile time seems to rely on `CStyleLanguage`. Of course, the renderer could just know the pragma and insert it with an environment variable. However, the latter feels like the wrong approach.

jumbojets:
Got it thanks. Going this route. OpenMP usage will look more like GPU stuff and hopefully will take advantage of the UOps.SPECIAL

Requires some work still

geohot:
Not going to merge. It adds complexity and relies on a third party library, eventually I think we'll support CPU multicore first party

PR #4201, opened 2024-04-17
some docstrings

wozeparrot:
all of creation and data access.

if style is good, I'll go through the rest of the functions in tensor as well.

working on a way to ignore docstrings in line count.


geohot:
Line too long

geohot:
Hmm, I should probably make the doc builder run always, and just only push on mkdocs or master

geohot:
I think we need the line count fix before we can merge, it's +274 now.

wozeparrot:
sound good, working on line count fix

geohot:
I put something in for line count fix, it's not exactly perfect, but it should be fine.

wozeparrot:
Then this should be fine to merge. Can just fix later `sz.py` when we run into weird edge cases if it does break.

PR #4204, opened 2024-04-17
[MLPerf][UNet3D] Add DICE loss + metrics

francislata:
This PR makes the following changes:
- Add `dice_ce_loss` which is used in the [MLPerf training reference implementation](https://github.com/mlcommons/training/blob/master/image_segmentation/pytorch/model/losses.py#L72-L82)
- Update `dice_score` so that it uses tinygrad instead of numpy
- Add tests for both `dice_score` and `dice_ce_loss`
- Slightly update `model_eval.py` for `eval_unet3d` to pass in arguments as `Tensor`


chenyuxyz:
need to add this test to ci, see https://github.com/tinygrad/tinygrad/blob/6eef8ee22adbe354a4ac9bb9db7b9d636f2fcc71/.github/workflows/test.yml#L229

might need to add the package in test setup too

francislata:
it should be good to go @chenyuxyz. let me know if I missed anything.

chenyuxyz:
verified model_eval.py is identical to the master one. not sure why it's 0.87 mean DICE though, maybe we need to update the pretrained model we use

PR #4209, opened 2024-04-18
Apple AMX support in LLVM or CLANG

zadorozhko:
Include aarch64.h where you want to use native AMX commands.
Other stuff is docu, tests and perf meter.


geohot:
Huh? What is this, does any part of it touch tinygrad?

geohot:
Don't include all this README crap, it's just the one header file.

zadorozhko:
No, it's standalone, and just allows you to use AMX commands where you need them.

PR #4211, opened 2024-04-18
linearizer: refactor to define accs with potentially TC-modified idxs

flammit:
This is a pre-req to the PADTO+TC feature branch which cleans up an internal issue with no visible side-effects in the current feature base (all acc's are currently valid when tensor cores are triggered).

The purpose of this PR is to move the calculation/recalculation of all indices before the acc global_load, otherwise acc elements maybe be made with an incorrect idx or as idx=0 if the TC reduce axis is padded.

There should be no behavioral or performance changes from this refactor.


chenyuxyz:
looks good for uop python emulation and real wmma

Issue #4168, opened 2024-04-14
Tinycrad can't find libhsa because hardcoded /lib in the code.


stalkerg:
By this line https://github.com/tinygrad/tinygrad/blob/master/tinygrad/runtime/autogen/hsa.py#L32C92-L32C98 they try to find lib by `/lib/libhsa-runtime64.so` inside ROCM_PATH, but on my system (Gentoo) this file in lib64 - `/usr/lib64/libhsa-runtime64.so` .

I don't think we should use absolute path in that case, it should be follow by ld.so

stalkerg:
We can use `--print-file-name` to find path to lib:
```
$: gcc --print-file-name=libhsa-runtime64.so
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../lib64/libhsa-runtime64.so

$: clang --print-file-name=libhsa-runtime64.so
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../lib64/libhsa-runtime64.so
```

PR #4218, opened 2024-04-19
fix HSA/KFD load for system-wide installation

stalkerg:
It fix for a such issue: #4168
Some distros like Gentoo install HSA and ROCm libs as system-wide libraries instead `/opt` directory. 
ROCM_PATH is more like build-time path, for runtime discovery better to use `find_library` especially if ROCM_PATH not here. 
It's also will work if follow official ROCm instruction https://rocm.docs.amd.com/projects/install-on-linux/en/latest/how-to/native-install/post-install.html


stalkerg:
ohh you actually generate hsa and compgr files. Hmmm I will fix it.

tinygrad/runtime/autogen/comgr.py
```diff
@@ -27,9 +28,12 @@ def char_pointer_cast(string, encoding='utf-8'):
     return ctypes.cast(string, ctypes.POINTER(ctypes.c_char))
 
 
-
 _libraries = {}
-_libraries['libamd_comgr.so'] = ctypes.CDLL(os.getenv('ROCM_PATH', '/opt/rocm/')+'/lib/libamd_comgr.so')
+_libraries['libamd_comgr.so'] = ctypes.CDLL(
```
nimlgen, line 32: Don't change autogenerated files. See `autogen_stubs.sh`

geohot:
We are going to replace HSA with KFD soon anyway.

tinygrad/runtime/autogen/comgr.py
stalkerg: thanks, yeah, I found it as well.

stalkerg:
> We are going to replace HSA with KFD soon anyway.

seems like for KFD we still need libamd_comgr.so

stalkerg:
@nimlgen can you check again? Thank you.

PR #4208, opened 2024-04-18
Multireduce Kernels - Linearizer and Scheduler Changes

0xtimmy:
re: an earlier pr this implements the actual kernel fusion of standard deviation

it:
1) allows `schedule.py` to put two ReduceOps in a kernel if they have the same shapes
2) changes `linearizer.py` to handle the multiple reduceops

an example of a fused kernel:
```
__kernel void r_64_16_2(__global double* data0, const __global double* data1) {
  __attribute__ ((aligned (16))) __local double temp[16];
  int gidx0 = get_group_id(0); /* 64 */
  int lidx1 = get_local_id(0); /* 16 */
  double acc0 = (double)(0.0);
  int alu0 = (gidx0+(lidx1*128));
  bool alu1 = (lidx1<1);
  for (int ridx0 = 0; ridx0 < 2; ridx0++) {
    double val0 = data1[alu0+(ridx0*64)];
    acc0 = (val0+acc0);
  }
  temp[lidx1] = acc0;
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu1) {
    double acc1 = (double)(0.0);
    for (int ridx1 = 0; ridx1 < 16; ridx1++) {
      double val1 = temp[ridx1];
      acc1 = (val1+acc1);
    }
    temp[0] = acc1;
  }
  barrier(CLK_LOCAL_MEM_FENCE);
  double val2 = temp[0];
  double acc2 = (double)(0.0);
  for (int ridx2 = 0; ridx2 < 2; ridx2++) {
    double val3 = data1[alu0+(ridx2*64)];
    double alu2 = (val3-(val2*(double)(0.03125)));
    acc2 = ((alu2*alu2)+acc2);
  }
  temp[lidx1] = acc2;
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu1) {
    double acc3 = (double)(0.0);
    for (int ridx3 = 0; ridx3 < 16; ridx3++) {
      double val4 = temp[ridx3];
      acc3 = (val4+acc3);
    }
    data0[gidx0] = sqrt((acc3*(double)(0.03225806451612903)));
  }
}
```

PR #4220, opened 2024-04-19
multi reduce scheduling spec start

Qazalin:
@0xtimmy I think these tests are a good starting point for the scheduler change #4208.

multi reduce schedule fusion can be more general.

`test_multireduce_multioutput_fancy` failed for one AST with a KeyError - I don't think it should, when I changed the order of the AST, it passed - We need more linearizer tests!
https://gist.github.com/Qazalin/3e0e46a4c99d13353dcd8789d91c477c

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -4,6 +4,7 @@ import unittest
 from tinygrad.codegen.kernel import Opt, OptOps, KernelOptError, tensor_cores
 from tinygrad.codegen.linearizer import Linearizer, UOp, UOps, expand_node, expand_idxs
 from tinygrad.device import Device, Buffer
+from tinygrad.features.graph import print_tree
 from tinygrad.ops import BinaryOps, BufferOps, MemBuffer, ConstBuffer, LazyOp, LoadOps, TernaryOps, ReduceOps, UnaryOps
 from tinygrad.shape.shapetracker import ShapeTracker
 from tinygrad.shape.view import View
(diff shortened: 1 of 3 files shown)


test/test_linearizer.py
```diff
@@ -79,6 +80,42 @@ def test_load_removed(self):
...
+
+    x = Tensor.empty(32, 32)
+    y = Tensor.empty(1, 32)
+    r0 = x.max(axis=0, keepdim=True)
+    r1 = (x - r0).sum(axis=0)
+    out = r0 + r1
+    ast = create_schedule([out.lazydata])[-1].ast
+    for op in ast: print_tree(op)
+    lin = Linearizer(*ast)
+    assert lin.reduceops[0].op is ReduceOps.MAX
```
Qazalin, line 115: @0xtimmy this assert should pass.

0xtimmy: oh good catch, I think this can be solved by just switching around where `[self]` is added in ops.py:
`def lazyops(self) -> List[LazyOp]: return dedup([self] + [item for x in self.src for item in x.lazyops])`
becomes:
`def lazyops(self) -> List[LazyOp]: return dedup([item for x in self.src for item in x.lazyops] + [self])`

so that deeper lazyops always come first

or if we don't want to change the LazyOp code we could put a helper function in `kernel.py` like this:
```
ordered_lazyop_dict = {}
  def ordered_lazyops(op): 
    if op not in ordered_lazyop_dict: ordered_lazyop_dict[op] = dedup([item for x in op.src for item in ordered_lazyops(x)] + [op])
    return ordered_lazyop_dict[op]
  self.reduceops = dedup([x for out in self.ast for x in ordered_lazyops(out) if x.op in ReduceOps])
```

PR #4229, opened 2024-04-19
Wikipedia preprocessing script 

Eliulm:
Mostly copied from [mlcommons](https://github.com/mlcommons/training/blob/master/language_model/tensorflow/bert/cleanup_scripts/create_pretraining_data.py). Will add tests in a separate PR.


Eliulm:
> mostly copied. so what's different and how do we know if it's okay

The changed part is about how we process documents in parallel and batch save them with `process_iterate`

> Can you provide some guidance on how to verify the preprocessed file.

I am working on a test that will first preprocess one document with the reference implementation (200000 resulting samples) and then do the same with our implementation. We can then iterate over the samples and compare them down to the last token. The issue right now is that the reference implementation only works with python3.7, but tinygrad only works with 3.8 and up. Will turn it into a draft for now.

Eliulm:
@chenyuxyz Do we want to keep the reference code in the repo or just have the user clone the mlcommons repository and execute the required commands there?

chenyuxyz:
i think it's fine to keep references in tinygrad repo for now. can you make sure the reference links to mlcommon are permalinks (pin to a commit instead of master)? will try the script later today and merge if all looks good

chenyuxyz (commented):
what was the issue causing data diff on tiny15 and tiny10?

extra/datasets/wikipedia.py
```diff
@@ -0,0 +1,411 @@
...
+      batch = instances[:batch_size]
+      del instances[:batch_size]
+      yield [instance_to_features(instance, tokenizer) for instance in batch]
+
+##################### Load files #####################
+
+def get_val_files(): return sorted(list((BASEDIR / "eval/").glob("*.pkl")))
+
+@diskcache
+def get_train_files(): return sorted(list((BASEDIR / "train/").glob("*/*.pkl")))
```
chenyuxyz, line 389: `@diskcache` searches by name and there's already a `get_train_files` for imagenet.

Eliulm:
> what was the issue causing data diff on tiny15 and tiny10?

There is a special way to handle Chinese tokenization and whitespace. I was basically missing two functions from the tokenizer. However, I believe that these did not affect the convergence behaviour all that much.

chenyuxyz (commented):
tested matched files should pass and unmatched files should fail, works pretty well! some typo and small nits. fix those and this is good to merge

PR #4232, opened 2024-04-20
two stage batchnorm calc_stats!

chaosagent:
```
Let
  x_ij be the jth element of the ith subbatch
  n_i be the number of elements in the ith subbatch
  E_j(x_ij) be the mean of the ith subbatch
  delta_i = E(x) - E_j(x_ij)

  E(x) be the mean of all elements
  Var(x) be the variance of all elements
  n be the total number of elements
(this notation is cursed)

n Var(x)
  = E_ij( (x_ij - E(x))^2 )
  = E_ij( ( (x_ij - E_j(x_ij)) - delta_i )^2 )
  = E_ij( (x_ij - E_j(x_ij))^2 ) - 2E_ij( (x_ij - E_j(x_ij)) delta_i ) + E_ij( delta_i^2 )          (linearity of expectation)
  = E_i( E_j( x_ij ) ) + E_i( delta_i E_j( x_ij - E_j(x_ij) ) ) + E_i( delta_i^2 )
Observe E_j( x_ij - E_j(x_ij) ) = E_j( x_ij ) - E_j(x_ij) = 0. Then
  = E_i( E_j( (x_ij - E_j(x_ij))^2 ) ) + E_i( delta_i^2 )
  = E_i( E_j( (x_ij - E_j(x_ij))^2 ) + delta_i^2 )
  = E(  subbatch variance + (mean - subbatch mean)^2 )

delta_i^2 are rather small though

```

Passes a test!

### DIFF
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -48,15 +48,26 @@ class UnsyncedBatchNorm:
       # This requires two full memory accesses to x
       # https://github.com/pytorch/pytorch/blob/c618dc13d2aa23625cb0d7ada694137532a4fa33/aten/src/ATen/native/cuda/Normalization.cuh
       # There's "online" algorithms that fix this, like https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_Online_algorithm
-      batch_mean = x.mean(axis=(1,3,4))
-      y = (x - batch_mean.reshape(shape=[batch_mean.shape[0], 1, -1, 1, 1]))
-      batch_var = (y*y).mean(axis=(1,3,4))
+      print(x.shape)
+      subbatch_mean = x.mean(axis=(3,4), keepdim=True)
+      subbatch_y = x - subbatch_mean
+      subbatch_ex2 = (subbatch_y * subbatch_y).sum(axis=(3,4), keepdim=True)
+      print(subbatch_mean.shape, subbatch_ex2.shape)
+
+      batch_mean = subbatch_mean.mean(axis=(1,), keepdim=True)
+      delta = batch_mean - subbatch_mean
+      batch_var = (subbatch_ex2 + prod(x.shape[3:]) * delta * delta).sum(axis=(1,), keepdim=True) / prod([x.shape[1], x.shape[3], x.shape[4]])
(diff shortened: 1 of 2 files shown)


chenyuxyz:
tested this with 4216, some kernels are different but no perf difference

chaosagent:
Can't expect any performance difference without fused double reduce... Right now this should perform about the same as split reduce

geohot:
What's the advantage to this?

chaosagent:
> What's the advantage to this?

It's a neat primitive to use for fused batchnorm. The idea is that you can do the batchnorm in two stages, the first being cache local (whatever you want that to mean) and with enough global dims. In the best case, when calculating a convolution, you can do two pass reduce for mean and variance for upcasted elements in your block, and then a second stage reduce later that is >10x smaller (basically free).

Without this, two pass reduce (timmy's thing) is not local, and requires a very large (bs=256*112*112 ~ 16mb) reduce to calculate the full channel mean, which can then be fused with the channel variance by scanning over the data again. Since there are only 64 channels per device in the worst case in resnet, this will have low occupancy and low cache locality. Doing this two stage algorithm allows calculating mean and variance locally, then globally.

chaosagent:
Online softmax is the same basic idea. Calculate local stats in parallel, combine to get global stats. This is a straightforward path to online softmax and flash attention.

chenyuxyz:
are you sure about this leads (or is required) to online softmax and flash attention? iiuc the computation is still sequential, not two stage  
image
https://arxiv.org/abs/1805.02867

chaosagent:
Screenshot_2024-04-22-04-01-57-75_3aea4af51f236e4932235fdada7d1643

chaosagent:
I think a problem with doing this for flashattn is that you need to not only store normalizer information, but also sum e^x*embedding for each subbatch. This expands the memory by a factor of 4096 (vs materializing the QK^T matrix), and a block/subbatch size of 4096 in the V axis is big. So it will not use less memory.

Need to give it some thought. At least this is a way to do fused softmax(QK^T), and the QK^T matrix can be rematerialized for summing Vs if you want to save the memory.

chaosagent:
shelve this until double reduce lands!

cc @0xtimmy

Issue #3572, opened 2024-03-02
Backward pass convs have two reduces


geohot:
See in beautiful_mnist

geohot:
![Screenshot 2024-03-02 at 5 00 44 AM](https://github.com/tinygrad/tinygrad/assets/72895/79572994-26a5-44cf-98b4-1ba71e1b3545)

This is the pattern

davidjanoskyrepo:
I'm not sure how to view the graphs but I would think the double reduce comes from the double max_pool2d. Modifying the model to this:

def __init__(self, use_new:bool=False) -> None:
    input_dim = 32
    expand_dim = 64
    lin_size = 9
    self.layers: List[Callable[[Tensor], Tensor]] = [
      nn.Conv2d(1, input_dim, 5), Tensor.relu,
      # Approximate the effect of the maxpooling's reduction in size by
      # increasing the stride of the convolution
      nn.Conv2d(input_dim, input_dim, 5, stride=1 if not use_new else 2), Tensor.relu,
      nn.BatchNorm2d(input_dim),
      # Remove first maxpooling, which increases the parameter size but only does one reduce
      Tensor.max_pool2d if not use_new else lambda x: x,  # NOTE: this is a hack to make the JIT work
      nn.Conv2d(input_dim, expand_dim, 3), Tensor.relu,
      nn.Conv2d(expand_dim, expand_dim, 3), Tensor.relu,
      nn.BatchNorm2d(expand_dim), Tensor.max_pool2d,
      lambda x: x.flatten(1), nn.Linear(expand_dim*lin_size, 10)]

Results in a speedup without measurable loss and the same number of parameters:

Training old model
Number of parameters: 88044
loss:   0.08 test_accuracy: 98.21%: 100%|██████████████| 70/70 [00:07<00:00,  8.90it/s] 
Training old model
Number of parameters: 88044
loss:   0.06 test_accuracy: 97.79%: 100%|██████████████| 70/70 [00:06<00:00, 10.43it/s] 
Training new model
Number of parameters: 88044
loss:   0.07 test_accuracy: 97.90%: 100%|██████████████| 70/70 [00:04<00:00, 15.64it/s] 
Training new model
Number of parameters: 88044
loss:   0.08 test_accuracy: 98.02%: 100%|██████████████| 70/70 [00:04<00:00, 16.60it/s]

Itssshikhar:
[...]

PR #4233, opened 2024-04-20
merge reduces in conv2d

learossi:
This is a fix for issue #3572. The two sequential reduces come from sequential expand operations within conv2d. When backwards is called on these tensors and the gradient is produced, these two expand operations become reduces. Specifically, the expand within repeat on line 697 of _pool is only separated from the expand in the WINO = 0 branch of conv2d by movement ops. Therefore, the two reduceops show up next to one another in the final graph. 

My solution is to add the ability to forward an expand into _pool. Specifically, _pool takes the integer argument 'new_dim' which is passed into the repeat call with the other _pool arguments. This creates the new dimension the 2nd expand inside conv2d would have created separately with only the one call to repeat.

The problem right now is that the current implementation triggers the idx.max assert in log_lazybuffer -> real_strides -> expr_idxs on my laptop. I'm unsure to what extent this is system specific. As I can tell, the current code for beautiful_mnist  and conv2d generates a shapetracker for the relevant lazybuffer which is about 10x smaller than idx.max. This means adding a new dimension of 32 to the shapetracker early increases idx.max by more than 10x. I don't think this is a bug given that I also trigger the assert on my system by simply increasing one of the image dimensions or the filter number in master by more than 10x or so; conversely, it doesn't feel like intended behavior. I think this could be fixed by adding a new tensor in the middle which inherits only the last view in the problematic shapetracker but I don't understand kernel generation well enough to know whether this breaks some sort of optimization down the line and would like some advice.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -685,7 +685,7 @@ class Tensor:
 
   # ***** processing ops *****
 
-  def _pool(self, k_:Tuple[sint, ...], stride:Union[Tuple[int, ...], int]=1, dilation:Union[Tuple[int, ...], int]=1) -> Tensor:
+  def _pool(self, k_:Tuple[sint, ...], stride:Union[Tuple[int, ...], int]=1, dilation:Union[Tuple[int, ...], int]=1, new_dim: int = None) -> Tensor:
     assert len(self.shape) >= len(k_), f"can't pool {self.shape} with {k_}"
     assert all_int(self.shape) and all_int(k_), f"does not support symbolic {self.shape=}, {k_=}"
     s_, d_ = make_pair(stride, len(k_)), make_pair(dilation, len(k_))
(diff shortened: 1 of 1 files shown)


geohot:
I don't understand this. Can you post the before and after graph of a conv?

learossi:
Yeah sure. Here's the model I'm running which is small enough not to trigger the idx.max assert: 
Screenshot 2024-04-21 124711
Here's an example of the double reduces from the graph that is produced by the main branch:
original_conv_graph_single_node
An here is that same node in the branch: 
revised_conv_graph_single_node
The svgs are way too big to nicely put in here, so here a drive link where you can grab both of the full graphs: 
https://drive.google.com/drive/folders/1Lu2oTNODpXLtIZQ7moG7EE6LaMhPFvlc?usp=sharing. 

You'll notice there still are double reduces even after the fix. There are at least a few double expands in the code in different places that generate this stuff when you call backwards. This just fixes the one specific to conv2d. I tracked down a few of them at one point, but then forgot to save the file where I wrote it down, but I could probably come up with where all the other ones come from in like an hour if you wanted to know.

xXWarMachineRoXx:
> The svgs are way too big to nicely put in here, so here a drive link where you can grab both of the full graphs:

yep, that's not an understatement. Congrats on the progress tho, It has been stale in a while. 
( _I was trying this problem out and couldn't get a hold on where to start_)

geohot:
It should probably be done at the scheduler level like #4441, but if there's other benefits to this I'm open to exploring more.

PR #4236, opened 2024-04-20
Mean of symboilc shape

PlamenHristov:
Hey all, 

First commit. Please let me know if I am on the right track or way off? 

If I'm on the right track let me know what's missing to get the PR in a mergeable state.


chenyuxyz:
it should not have MeanNode, and should include tests that `mean` of symbolic shape tensor with variable shapes are correct

PlamenHristov:
@chenyuxyz ok. Was the mean() implemenation roughly on the right track? Apart from the tests, are there any other places I might need to make a change?

PR #4238, opened 2024-04-20
Relu backward compare to forward input directly

chenyuxyz:
comparing 0 with forward output (max(x, 0)) is the same as comparing to x directly

wip as this fuses differently and increases the kernel


chenyuxyz:
don't think this has any benefit now

PR #4242, opened 2024-04-21
RNNT training

andresgit:
WIP, I tried to reproduce the results of this Nvidia PyTorch 2xA30 implementation: https://github.com/mlcommons/training_results_v2.0/tree/main/NVIDIA/benchmarks/rnnt/implementations/pytorch

The Nvidia implementation trains in fp16 (with fp32 master weights and loss scaling) in around 45 epochs (+- 5 epochs deviation), one epoch had 278528 data samples. To train in 24 hours means around 6.9 ms/sample.

For this tinygrad version, currently one 4090 can train in fp32 (batch size per gpu 32) in around 14 ms/sample, 2x4090 did only around 10 ms/sample and the gpu usage didn't get to 100%. Hopefully the python part can be optimized to get full multigpu usage. Need to test running in fp16.

To try to have an overview and to time and optimize some steps, I wrote out the backward manually. Also at some point I started writing it without using MultiLazyBuffer, so it would need to be rewritten to use MLB, autograd etc and other standard tinygrad features without doing things manually.
The preprocessing I mostly copied from the pytorch implementation and it currently uses Nvidia DALI and pytorch.
I changed some pointer values myself by hand to have a kind of setitem/getitem and to reuse the same kernels without having to run beam again on the same kernels with offset buffers.
Also dropout was slow and currently I just call cuRAND.
Some later stage kernels use a lot of memory, so I prerun some of them in the beginning to get the beam search cached, otherwise it would run out of memory and I would have to decrease the batch size.

I have run it for 6 epochs and it behaves similarly to the pytorch implementation. With pytorch in 6 epochs the loss went from 1500 to 42, word error rate from 100% to 13.6%. For the tinygrad run the loss went from 1700 to 50, wer from 100% to 14.6%.

I run it with just `python model_train.py`

### DIFF
--- a/examples/mlperf/model_train.py
+++ b/examples/mlperf/model_train.py
@@ -249,8 +249,249 @@ def train_unet3d():
   pass
 
 def train_rnnt():
-  # TODO: RNN-T
-  pass
+  traint0 = time.time()
+  import numpy as np
+  import sentencepiece
+  import gc
+  from extra.datasets import librispeech2
+  from extra.models.rnnt2 import RNNT
+  import math
+
+  seed = 0
+  print(f"{seed=}")
+
+  n_classes = 1024
+  in_feats = 256
+  enc_n_hid = 1024
+  enc_pre_rnn_layers = 2
+  enc_post_rnn_layers = 3
+  enc_dropout = 0.1
+  pred_dropout = 0.3
+  joint_dropout = 0.3
+
+  pred_n_hid = 512
+  pred_rnn_layers = 2
+  joint_n_hid = 512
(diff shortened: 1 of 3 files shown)


geohot:
Why is the model reimplemented? Isn't there already an RNN-T for inference?

andresgit:
It is just the current state of various experiments I ran to try to get an overview of everything and to try to improve speed.

In the end I would try to do only minimal changes to the existing implementation.

chenyuxyz:
no progress for a week, closing

PR #4244, opened 2024-04-21
resnet print epoch ops and mem in benchmark

chenyuxyz:
also added a flag to optionally disable reset jitted steps


chenyuxyz:
31_838 T ops, so at 100 TFLOPS it's 31838 / 100 * 37 / 3600 = 3.27 hours
1_525 T mem, so at 5 T/s it's 1525 / 5 * 37 / 3600 = 3.13 hours

PR #4246, opened 2024-04-22
replace CLANG dlopen with mmap+mprotect and remove linker

jumbojets:
Currently, only this minimal example works (because the offset into the shared library is hard-coded)

```py
N = 1024; a, b = Tensor.rand(N, N), Tensor.rand(N, N)
(a@b).realize()
```

A call to `mprotect` in addition `mmap` is necessary (at least for macOS) to remove write permission and add exec permission. Python's `mmap` module is [pending `mprotect`](https://github.com/python/cpython/pull/114674), so in the meantime, it seems we have to load libc--hopefully just once.

As mentioned in discord, certain kernels depend on libm/libc, making skipping the linker problematic. I believe the most straightforward solution here is to just attach our own implementations of the offending functions (`sin`/...), but I'm not sure its the best.

It is not yet clear to me that this will be a win over `CDLL`.

TODO:
- [ ] Dynamically find the function's offset into shared library without linker step
- [ ] Address `libm` dependency (we can't just skip linker)
- [ ] Clean up code (particularly, mmap/mprotect function usage)
- [ ] Benchmark

### DIFF
--- a/tinygrad/runtime/ops_clang.py
+++ b/tinygrad/runtime/ops_clang.py
@@ -36,7 +36,7 @@ class ClangProgram:
     if code_addr == -1: raise OSError('mmap failed to allocate memory')
     ctypes.memmove(code_addr, lib, len(lib))
     if mprotect(code_addr, len(lib), mmap_flags.PROT_READ | mmap_flags.PROT_EXEC, 0) < 0: raise OSError('mprotect failed to make memory executable')
-    self.fxn = ctypes.CFUNCTYPE(None)(code_addr + 0x0000000000003e44) # TODO: just get offset from symbol table?
+    self.fxn = ctypes.CFUNCTYPE(None)(code_addr + 0x0000000000003e44) # TODO: how to best get the function offset?
 
   def __call__(self, *bufs, vals=(), wait=False): return cpu_time_execution(lambda: self.fxn(*bufs, *vals), enable=wait)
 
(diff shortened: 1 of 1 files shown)


tinygrad/runtime/ops_clang.py
```diff
@@ -18,10 +32,11 @@ def compile(self, src:str) -> bytes:
...
   def __init__(self, name:str, lib:bytes):
     self.name, self.lib = name, lib
-    # write to disk so we can load it
-    with tempfile.NamedTemporaryFile(delete=True) as cached_file_path:
-      pathlib.Path(cached_file_path.name).write_bytes(lib)
-      self.fxn = ctypes.CDLL(str(cached_file_path.name))[name]
+    code_addr = mmap(None, len(lib), mmap_flags.PROT_READ | mmap_flags.PROT_WRITE, mmap_flags.MAP_ANON | mmap_flags.MAP_PRIVATE, -1, 0)
+    if code_addr == -1: raise OSError('mmap failed to allocate memory')
+    ctypes.memmove(code_addr, lib, len(lib))
+    if mprotect(code_addr, len(lib), mmap_flags.PROT_READ | mmap_flags.PROT_EXEC, 0) < 0: raise OSError('mprotect failed to make memory executable')
```
geohot, line 38: why do you need mprotect, just mmap +x

chenyuxyz:
close as stale

PR #4249, opened 2024-04-22
BERT language modeling head + trunc normal initializers

Eliulm:


chenyuxyz:
this should be in the same pr as the main training loop. by itself it's not very useful and hard to review without testing the whole thing

PR #4252, opened 2024-04-22
BERT dataloader

Eliulm:


examples/mlperf/dataloader.py
```diff
@@ -139,6 +140,37 @@ def receive_batch():
...
+    "input_mask": Tensor(np.concatenate([s["input_mask"] for s in data], axis=0), dtype=dtypes.float32),
+    "segment_ids": Tensor(np.concatenate([s["segment_ids"] for s in data], axis=0), dtype=dtypes.float32),
+    "masked_lm_positions": Tensor(np.concatenate([s["masked_lm_positions"] for s in data], axis=0), dtype=dtypes.float32),
+    "masked_lm_ids": Tensor(np.concatenate([s["masked_lm_ids"] for s in data], axis=0), dtype=dtypes.float32),
+    "masked_lm_weights": Tensor(np.concatenate([s["masked_lm_weights"] for s in data], axis=0), dtype=dtypes.float32),
+    "next_sentence_labels": Tensor(np.concatenate([s["next_sentence_labels"] for s in data], axis=0), dtype=dtypes.float32),
+  }
+
+# For train: Stop when we run through all data
+# For val: Wrap around val dataset and never stop
```
chenyuxyz, line 159: how do you plan to solve the issue if val is not a multiple of BS? for mlperf it needs to eval on data exactly once

Eliulm: In this implementation, the model processes each data sample exactly once, regardless of the batch size. To my knowledge, in MLPerf, the validation dataset also wraps around after it has been fully processed.

chenyuxyz: i see, can you add the link to this wrapping implementation in reference in comment?

PR #4253, opened 2024-04-22
Optimize bool multiply/indexing/nn.Embedding --- O(N + M) nn.Embedding

alexlitz:
Optimizes bool tensor indexing by fusing cases where a bool tensor constant over a is cast then multiplied with another value before being added to an accumulator. The operating principle is if it is zero over that broadcast dim the computation can be eliminated. Sometimes this requires loop reordering if the dim it is being iterated over is not the lowest loop. So loops are reordered if needed, bools are pulled out it is check if the aforementioned criteria is meet if so the loop(s) are put in an if statement. This helps with the indexing speed as well as many cases with bool broadcast multiplies. The arange case can be optimized further to just more directly do the indexing but would be less general and this seems to get ~2/3 (depending of course on sizes) of the order of magnitude improvements.

Running:
CLANG=1 NOOPT=1 TC=0 DEBUG=4 GRAPH=1 python3 test/external/external_test_mnist_data_select.py
Before:
*** CLANG      4 r_512_784_60000                        arg   4 mem  0.22 GB tm 101742.51ms/101820.88ms (    0.71 GFLOPS,    0.00 GB/s)

After:
*** CLANG      4 r_512_784_60000                        arg   4 mem  0.22 GB tm     10.70ms/    88.36ms ( 6750.10 GFLOPS,   17.75 GB/s)

So a 10,000x speedup.


alexlitz:
> https://github.com/tinygrad/tinygrad/?tab=readme-ov-file#contributing "If your PR looks "complex", is a big diff, or adds lots of lines, it won't be reviewed or merged. Consider breaking it up into smaller PRs that are individually clear wins. A common pattern I see is prerequisite refactors before adding new functionality. If you can (cleanly) refactor to the point that the feature is a 3 line change, this is great, and something easy for us to review."

Sure I can break it up probably into ~3 MRs, I wanted to confirm that at a high level this is something that is wanted (in terms both of the functional improvement and the approach) beforehand though. Breakdown into smaller MRs would be approximately:
misc helpers/checks in uops
add_ends and try_fix_vin: changes that should be helpful for uops flexibility independent of this optimization
Changes specific to this optimization
Does that sound suitable to you?

chenyuxyz:
i don't know, to me it's a 300 line change to core tinygrad that has no tests. why would i want this?

alexlitz:
Well because it works, unborks indexing, helps optimize other kernels and should help make further uops optimizations easier as well. As far as tests I did not add any since it seemed reasonably well covered in existing tests. Up to you though just lmk.

chenyuxyz:
please read https://github.com/tinygrad/tinygrad/?tab=readme-ov-file#contributing

alexlitz:
I have, you mentioned the one part above, I am looking for some clarity here and I don't feel like I have it, do you want changes to this effect or no?

alexlitz:
I see actually the full test suite is not passing that is my bad. I still don't have clarity here on whether at a high level these changes would be wanted.

PR #4259, opened 2024-04-23
Multireduce-Kernels: Linearizer Changes and Tests

0xtimmy:
This is the linearizer stage of the multireduce kernels PR

It contains changes to `linearizer.py`, `kernel.py`, and `uops.py` to allows them to produce kernels with multiple reductions
Test cases have been added to `test_linearizer.py` that check a few things:
- simple fusion of reduceops
- fusion with intermediate calculations (both at the output shape and full shape)
- fusion with multiple outputs (outputs must be the same shape)
- re: @Qazalin , I didn't add your tests in directly because this PR doesn't change the scheduler, but I did put in one case that uses the same ast, there were a few issues 1) was the ordering of reduceops and 2) that intermediate values would sometimes get rendered within a for loop. I fixed both

The main changes to `linearizer.py` are touching up spots that depend on there being only one reduceop, for example storing the result of a local reduce in a temp buffer so it can be loaded back into every thread
I also changed `kernel.py` so the "deeper" reduceops are rendered first
And I had to change `uops.py` so that if statements can be closed before the end of the kernel


tinygrad/codegen/kernel.py
```diff
@@ -71,17 +71,21 @@ def __init__(self, *ast:LazyOp, opts:Optional[CompilerOptions]=None):
     self.ast = ast
     self.lazyops = flatten([op.lazyops for op in self.ast])
 
-    # there's only allowed to be one reduceop
-    reduceops = [x for x in self.lazyops if x.op in ReduceOps]
-    assert len(dedup(reduceops)) <= 1, "max one reduce op in an ast"
-    self.reduceop = reduceops[0] if reduceops else None
+    def get_rchildren(x:LazyOp): return ([x] if x.op in ReduceOps else []) + [r for s in x.src for r in get_rchildren(s)]
```
Qazalin, line 74:
```py
import time
from tinygrad.engine.schedule import create_schedule

st = time.perf_counter()
a = Tensor([1,2,3,4])
for _ in range(24): a = a + a
r = a.sum()
ast = create_schedule([r.lazydata])[-1].ast
lin = Linearizer(*ast)
assert time.perf_counter()-st < 1.0
```

0xtimmy: I futzed around with it and the traversal seems to just be too slow in general so instead of reordering them when the kernel get initialized, I them render in the ast_parse, because ast_parse is recursing through the ast anyways.

I did have to make some changes to `uops.py` so I'm making some more tests to make sure things still linearize properly

Qazalin:
This diff is very hard to review, can you break it down to isolated PRs?

Qazalin:
@0xtimmy can you provide a status update?

0xtimmy:
finals week, I can have a PR up probably tomorrow

0xtimmy:
in regards to @chenyuxyz 's [comment ](https://github.com/tinygrad/tinygrad/pull/4341#issuecomment-2081596422) abt using `insert_before` I came up with two additional solutions: 
1) we can "chunk" each reduceop: so keep rendering them in ast parse and just place the chunk of new uops before any existing loop context
2) we can render all the recudeops out of order with some placeholder for each accumulator, then move them into the proper order and switch out the placeholder during the main ast_parse

just some ideas I had, I can put up PRs once they're tested

chaosagent (commented):
small things

tinygrad/codegen/kernel.py
```diff
@@ -116,8 +113,8 @@ def copy(self):
     ret.opts, ret.ast, ret.lazyops = self.opts, self.ast, self.lazyops
 
     # things downstream of the AST
-    ret.reduceop, ret.outbufs, ret.vars, ret.bufs, ret.earlybufs, ret.full_buf_index = \
-      self.reduceop, self.outbufs, self.vars, [x for x in self.bufs if not isinstance(x, LocalBuffer)], self.earlybufs, self.full_buf_index
+    ret.reduceops, ret.outbufs, ret.vars, ret.bufs, ret.earlybufs, ret.full_buf_index = \
+      self.reduceops, self.outbufs, self.vars, [x for x in self.bufs if not isinstance(x, LocalBuffer)], self.earlybufs, self.full_buf_index
```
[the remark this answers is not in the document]
chaosagent: nevermind, reduceops is not mutated

tinygrad/codegen/linearizer.py
```diff
@@ -293,14 +291,23 @@ def upcast_strides(buf:int):
...
+        for j in self.upcast_in_mid_reduce_axes:
+          self.upcasted -= 1
+          self.group_for_reduces += 1
+        reduce_buf_uop = self.buf_uops[-1]
+        assert reduce_buf_uop is not None, "Local reduce buf must have been uoped at this point"
+        stores = self.global_store(-1, [NumNode(0)]*len(self.sts[-1].shape), acc)
+        endif = self.uops.add(UOps.ENDIF, None, (barrier,), cachable=False)
+        barrier = self.uops.add(UOps.BARRIER, None, (endif, *stores), cachable=False)
+        acc = self.global_load(-1, [NumNode(0)]*len(self.sts[-1].shape), barrier=barrier)
+    return acc
```
0xtimmy, line 310: this will store the result of the local reduce and load it back into each thread

Qazalin: why is the ENDIF still there?

Qazalin (commented):
Is this diff complete? If so CI should be green.

0xtimmy:
do you want to include #4453 in this pr? it has the test for it in an ast form

Qazalin:
yeah let's have the entire linearizer diff here

Qazalin:
can you add the kernel.py and linearizer.py changes here as well? That'd help me verify correctness for #4457

Qazalin:
amd tests pass on real AMD - I'll update the emulator in CI

tinygrad/codegen/linearizer.py
```diff
@@ -101,7 +101,7 @@ def global_load(self, i:int, idxs:List[Node], acc:Optional[LazyOp]=None, barrier
       key = f"{acc is not None}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
       if key not in self.load_cache:
         if acc is not None:
-          self.load_cache[key] = self.uops.add(UOps.DEFINE_ACC, localtype, (), (self.get_reduce_acc(acc), i, acc_count))
+          self.load_cache[key] = self.uops.add(UOps.DEFINE_ACC, localtype, ((barrier,) if barrier else ()), (self.get_reduce_acc(acc), i, acc_count))
```
Qazalin, line 104: so I think a couple things are wrong here:
1. load_cache[key] needs to update. acc is not None isn't sufficient if you have multiple reduces.
I tried doing this - You need to make it unique.
https://github.com/Qazalin/tinygrad/commit/bb677d6b106e29acef712cde02fd426cf940ebba
2. all uops are cachable now - i is the same for both reduces so it'll cache DEFINE_ACC, it shouldn't.

0xtimmy: the barrier is to prevent it from using a previous reduce from `saved_expr` or `load_cache[key]` but I think I found a better way that's more like https://github.com/Qazalin/tinygrad/commit/bb677d6b106e29acef712cde02fd426cf940ebba

w/o barrier:
```
__kernel void r_128_64n1(__global float* data0, const __global float* data1) {
  int gidx0 = get_group_id(0); /* 128 */
  float acc0 = 0.0f;
  int alu0 = (gidx0*64);
  for (int ridx0 = 0; ridx0 < 64; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
  for (int ridx1 = 0; ridx1 < 64; ridx1++) {
    float val1 = data1[alu0+ridx1];
    acc0 = ((val1-acc0)+acc0);
  }
  data0[gidx0] = acc0;
}
```

w/ barrier:
```
__kernel void r_128_64n1(__global float* data0, const __global float* data1) {
  int gidx0 = get_group_id(0); /* 128 */
  float acc0 = 0.0f;
  int alu0 = (gidx0*64);
  for (int ridx0 = 0; ridx0 < 64; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
  float acc1 = 0.0f;
  for (int ridx1 = 0; ridx1 < 64; ridx1++) {
    float val1 = data1[alu0+ridx1];
    acc1 = ((val1-acc0)+acc1);
  }
  data0[gidx0] = acc1;
}
```

tinygrad/codegen/linearizer.py
```diff
@@ -103,10 +105,11 @@ def global_load(self, i:int, idxs:List[Node], acc:Optional[LazyOp]=None, barrier
...
       this_const, idx, valid = (invalid_value, NumNode(0), NumNode(1)) if valid.max == 0 else (const, idx, valid)
       # todo: when multiple reduceops are supported, clearly disambiguate and test acc load keys are unique for each reduceop
-      key = f"{acc is not None}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
+      key = f"{self.reduceops.index(acc) if acc is not None else None}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
       if key not in self.load_cache:
         if acc is not None:
-          self.load_cache[key] = self.uops.add(UOps.DEFINE_ACC, localtype, ((barrier,) if barrier else ()), (self.get_reduce_acc(acc), i, acc_count))
+          id = self.reduceops.index(acc)*2 + (0 if self.bufs[i].__class__ is LocalBuffer else 1)
+          self.load_cache[key] = self.uops.add(UOps.DEFINE_ACC, localtype, (), (self.get_reduce_acc(acc), id, acc_count))
```
0xtimmy, line 110: set "i" in a way that doesn't break with multiple accs

tinygrad/codegen/linearizer.py
0xtimmy: here: https://github.com/tinygrad/tinygrad/pull/4259/commits/c1f4cd539e0d4ece02e396d7a40d70abf30e9ed6#r1602269166

0xtimmy:
> can you add the kernel.py and linearizer.py changes here as well? That'd help me verify correctness for #4457

this includes the kernel + linearizer changes as well as the tests, https://github.com/tinygrad/tinygrad/pull/4259/commits/71db41b26f8f46174f6c77462d8a57bbee39d8e6 has the `assert len(reduceops) < 2` commented out

0xtimmy:
and this is with the assert commented back in

tinygrad/codegen/linearizer.py
Qazalin: I think id should be a combination of {reduceop_idx} + {early acc or late acc}.
eg:
https://github.com/Qazalin/tinygrad/commit/2a7b707d2594491e154195d8e9620e6f652a6698#diff-63d57daaca59e5495bbb108da1c13e782f4a408db8503e015e8094ac07c87ea6R109
you could also something like https://github.com/Qazalin/tinygrad/commit/c44261fb4da9ca1a7ad1483d6fe0afa7c4308b43
I think string is simpler to reason about.

tinygrad/codegen/uops.py
```diff
@@ -225,13 +227,11 @@ def fix_loop_scope(self, get_recursive_parents:Callable[..., Set[UOp]]):
       else:
         parents = get_recursive_parents(u, with_phi=True)
         # don't push any local buffer because there might have STORE and BARRIER (not considered as parent) between DEFINE_LOCAL and here
```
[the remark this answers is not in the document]
0xtimmy: hmmm that's suspicious, i'll take a look when i get back to my computer

test/test_linearizer.py
```diff
@@ -210,10 +210,14 @@ def test_multireduce_loop_scope(self):
...
+    const = ConstBuffer(val=1/64, dtype=dtypes.float, st=ShapeTracker.from_shape((1,1,1)))
     ast = LazyOp(op=BufferOps.STORE, src=(LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BinaryOps.SUB, src=(
-      LazyOp(op=BufferOps.LOAD, src=(), arg=load), LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BufferOps.LOAD, src=(), arg=load),)
-      , arg=((axis,), dtypes.float))), arg=None),), arg=((axis,), dtypes.float)),), arg=store)
+      LazyOp(op=BufferOps.LOAD, src=(), arg=load), LazyOp(op=BinaryOps.MUL, src=(
+        LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BufferOps.LOAD, src=(), arg=load),), arg=((axis,), dtypes.float)),
+        LazyOp(op=BufferOps.CONST, src=(), arg=const),
+        )), ), arg=None),), arg=((axis,), dtypes.float)),), arg=store)
     k = Linearizer(ast)
+    k.hand_coded_optimizations()
```
0xtimmy, line 220: this will create the `DEFINE_LOCAL` that causes problems with loop_scope when rendering multiple reduceops

b/c the `MUL` uses a previous `SUM` as an input it won't get moved around by `fix_loop_scope`, this isn't necessarily problematic in this case it will just do the calculation for every iteration of the loop. BUT if the result of the `MUL` gets used later, outside this loop, it will cause compilation issues because of scope

tinygrad/codegen/uops.py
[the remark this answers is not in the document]
0xtimmy: re: https://github.com/tinygrad/tinygrad/pull/4259/commits/8fba75e2bff67bee9e56d3138dbb3c34cb12e497#r1604165745

tinygrad/codegen/linearizer.py
0xtimmy: I like the string too, the change is in this diff: https://github.com/tinygrad/tinygrad/pull/4259/commits/57125675cebbfc0bc3aa8325e6fab4a3e22e1e56

0xtimmy:
the tests are each for testing some specific part of the process (checking that loop scope is correct for example)
if we're just checking for correctness we could probably cut a few of the tests because `test_mean_std_multireduce` causes p much all the situations we'll run into

0xtimmy:
so I've been testing in my own branch here: https://github.com/0xtimmy/tinygrad/tree/linearizer-testing

and there's some suspicious stuff going on,,, all the kernels look correct but `gpu` and `amd` were not generating the correct output

if we give each reduceop it's own local buffer, that will fix `gpu`; but we should just be able to overwrite values in one local buffer

`amd` simply fails and i'm not sure why, working on that now

Qazalin:
amd is an emulator problem unrelated to this https://github.com/tinygrad/tinygrad/pull/4259#issuecomment-2112848930

0xtimmy:
interesting, is there a way I can test it on real AMD? I'll figure out what's on with `GPU=1` in the meantime

Qazalin:
feel free to join discord and request tinybox AMD access.

Qazalin:
uops.py got a rewrite fyi - I think this diff requires some changes.
The tests are almost there, can you create separate PR for getting the tests upstreamed?
I'd add tests for 2+ back to back reduces as well.

0xtimmy:
yeah i'm in the process of merging the uops rewrite, I'll make a pr for the tests

tinygrad/codegen/uops.py
```diff
@@ -304,6 +304,7 @@ def linearize(self, extra_pm:Optional[PatternMatcher]=None, type_verify=True):
     in_degree: DefaultDict[UOp, int] = defaultdict(int)
     loops = []
     ifs = []
+    stores = []
```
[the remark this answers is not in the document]
0xtimmy: no

test/test_linearizer.py
Qazalin: Is this fixed?

0xtimmy: the test? it ought to be

there is something strange where in the rendered kernel it will move the `MUL` back into the loop if it's only used within the loop scope despite it coming before the `RANGE` in the uops

if the MUL is not used outside the loop
*kernel:*
```
...
  barrier(CLK_LOCAL_MEM_FENCE);
  float val2 = temp[0];
  float acc2 = 0.0f;
    float val3 = data1[alu1+ridx2];
    float alu2 = (val3-(val2*0.03125f));
    acc2 = ((alu2*alu2)+acc2);
  }
  temp[lidx1] = acc2;
  barrier(CLK_LOCAL_MEM_FENCE);
...
```
*uops:*
```
...
  33 UOps.BARRIER        :                           [31]                             None
  34 UOps.LOAD           : dtypes.float              [2, 3, 33]                       None
  35 UOps.ALU            : dtypes.float              [34, 4]                          BinaryOps.MUL
  36 UOps.DEFINE_ACC     : dtypes.float              [37]                             (0.0, '1_-1', 0)
  37 UOps.RANGE          : dtypes.int                [3, 6]                           (6, 0)
  38 UOps.ALU            : dtypes.int                [14, 37]                         BinaryOps.ADD
  39 UOps.LOAD           : dtypes.float              [1, 38]                          None
  40 UOps.ALU            : dtypes.float              [39, 35]                         BinaryOps.SUB
  41 UOps.ALU            : dtypes.float              [40, 40]                         BinaryOps.MUL
  42 UOps.ALU            : dtypes.float              [41, 36]                         BinaryOps.ADD
  43 UOps.PHI            : dtypes.float              [36, 42]                         None
  44 UOps.ENDRANGE       :                           [37]                             None
  45 UOps.STORE          :                           [2, 10, 43]                      None
  46 UOps.BARRIER        :                           [45]                             None
...
```

versus if the MUL is used outside the loop
*kernel*
```
...
  barrier(CLK_LOCAL_MEM_FENCE);
  float val2 = temp[0];
  float alu2 = (val2*0.03125f);
  float acc2 = 0.0f;
    float val3 = data1[alu1+ridx2];
    float alu3 = (val3-alu2);
    acc2 = ((alu3*alu3)+acc2);
  }
  temp[lidx1] = acc2;
  barrier(CLK_LOCAL_MEM_FENCE);
...
```

*uops*
```
...
  33 UOps.BARRIER        :                           [31]                             None
  34 UOps.LOAD           : dtypes.float              [2, 3, 33]                       None
  35 UOps.ALU            : dtypes.float              [34, 4]                          BinaryOps.MUL
  36 UOps.DEFINE_ACC     : dtypes.float              [37]                             (0.0, '1_-1', 0)
  37 UOps.RANGE          : dtypes.int                [3, 6]                           (6, 0)
  38 UOps.ALU            : dtypes.int                [14, 37]                         BinaryOps.ADD
  39 UOps.LOAD           : dtypes.float              [1, 38]                          None
  40 UOps.ALU            : dtypes.float              [39, 35]                         BinaryOps.SUB
  41 UOps.ALU            : dtypes.float              [40, 40]                         BinaryOps.MUL
  42 UOps.ALU            : dtypes.float              [41, 36]                         BinaryOps.ADD
  43 UOps.PHI            : dtypes.float              [36, 42]                         None
  44 UOps.ENDRANGE       :                           [37]                             None
  45 UOps.STORE          :                           [2, 10, 43]                      None
  46 UOps.BARRIER        :                           [45]                             None
...
```

both kernels are correct, so I didn't look into it that much

0xtimmy:
> uops.py got a rewrite fyi - I think this diff requires some changes. The tests are almost there, can you create separate PR for getting the tests upstreamed? I'd add tests for 2+ back to back reduces as well. We need a robust way for testing loop scopes.

I made a separate pr for the tests here: #4665

Qazalin:
wanna delete the skips too?

tinygrad/codegen/linearizer.py
```diff
@@ -390,21 +401,20 @@ def linearize(self):
 
     # parse AST
     loaded_buffers:Dict[Union[MemBuffer, ConstBuffer, LocalBuffer], List[UOp]] = {}
-    acc: List[UOp] = []
+    accs: Dict[LazyOp, List[UOp]] = {}
```
Qazalin, line 404: can you factor out this accs dict refactor to its own PR?

0xtimmy: yes sir

0xtimmy:
> wanna delete the skips too?

but this would mean that the `assert len(reduceops) <= 1` has to go as well, or at least get moved

Qazalin:
yeah you can remove it, the diff supports multiple reduces.

test/test_linearizer.py
Qazalin: hmm - I don't think it should add the MUL to loop even if it's output is correct. In both cases, you should assert that the mul is always outside of the loop.

Do you have a minimal test for when it's uoped in the loop?

0xtimmy: I did, but I expanded it to show the correctness issues. It *IS* uoped in the loop, but it's being rendered in the range, ex the steps `cstyle.py` takes when calculating a standard deviation:

```
=== render step 13 ========================================
UOps.ENDRANGE       :                           [<UOps.RANGE: 20>]               None
  int gidx0 = get_group_id(0); /* 128 */
  int alu0 = (gidx0*64);
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 64; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
============================================================
=== render step 14 ======================================== // hits the MUL here
UOps.ALU            : dtypes.float              [<UOps.PHI: 17>, <UOps.CONST: 6>] BinaryOps.MUL
  int gidx0 = get_group_id(0); /* 128 */
  int alu0 = (gidx0*64);
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 64; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
============================================================
=== render step 15 ========================================
UOps.DEFINE_ACC     : dtypes.float              [<UOps.RANGE: 20>]               (0.0, '1_0', 0)
  int gidx0 = get_group_id(0); /* 128 */
  int alu0 = (gidx0*64);
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 64; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
  float acc1 = 0.0f;
============================================================
... // some other ops for rendering the loop
=== render step 20 ========================================
UOps.ALU            : dtypes.float              [<UOps.ALU: 13>, <UOps.DEFINE_ACC: 5>] BinaryOps.ADD
  int gidx0 = get_group_id(0); /* 128 */
  int alu0 = (gidx0*64);
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 64; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
  float acc1 = 0.0f;
  for (int ridx1 = 0; ridx1 < 64; ridx1++) {
    float val1 = data1[alu0+ridx1];
============================================================
=== render step 21 ========================================
UOps.PHI            : dtypes.float              [<UOps.DEFINE_ACC: 5>, <UOps.ALU: 13>] None // actually renders it here
  int gidx0 = get_group_id(0); /* 128 */
  int alu0 = (gidx0*64);
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 64; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
  float acc1 = 0.0f;
  for (int ridx1 = 0; ridx1 < 64; ridx1++) {
    float val1 = data1[alu0+ridx1];
    acc1 = ((val1-(acc0*0.015625f))+acc1);
============================================================
=== render step 22 ========================================
UOps.ENDRANGE       :                           [<UOps.RANGE: 20>]               None
  int gidx0 = get_group_id(0); /* 128 */
  int alu0 = (gidx0*64);
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 64; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
  float acc1 = 0.0f;
  for (int ridx1 = 0; ridx1 < 64; ridx1++) {
    float val1 = data1[alu0+ridx1];
    acc1 = ((val1-(acc0*0.015625f))+acc1);
  }
============================================================
```

Qazalin: The ALU for `acc0*0.015625f` should uop outside of the RANGE, It's ok if cstyle just inlines it though

0xtimmy: we should be good then 👍

0xtimmy:
the failing tests seem to pass on the tinybox when run with `GPU=1` and `AMD=1` but I will look into it more

Qazalin:
you can skip AMD CI for now - this is how we do it: https://github.com/tinygrad/tinygrad/blob/master/test/test_linearizer.py#L433

0xtimmy:
I messed around with the `GPU=1` a bit and I can't seem to get the issue to reproduce on the tinybox or my machine but I do know how to fix them re:https://github.com/tinygrad/tinygrad/pull/4259/commits/4c96836c036802433aa5877033bef15a80044ed9, which gives each reduceop it's own local buffer

the downside of this is that it might take up more scratch memory, but the upside is that we'll probably have to do it anyways for parallel reduction

tinygrad/codegen/linearizer.py
```diff
@@ -100,10 +100,11 @@ def global_load(self, i:int, idxs:List[Node], acc:Optional[LazyOp]=None, barrier
     for idx, valid, rep_idx in zip(e_idxs, e_valids, iter_idxs(expand_vars)):
       this_const, idx, valid = (invalid_value, NumNode(0), NumNode(1)) if valid.max == 0 else (const, idx, valid)
       # todo: when multiple reduceops are supported, clearly disambiguate and test acc load keys are unique for each reduceop
```
Qazalin, line 102: can remove this todo

test/test_linearizer.py
```diff
@@ -355,9 +360,8 @@ def test_upcast_with_locals(self):
     assert accs[1].dtype == stores[1].vin[-1].dtype == dtypes.float
     assert stores[1].vin[0].uop is UOps.DEFINE_GLOBAL
 
-  @unittest.skip("multireduce isn't supported yet")
   def test_upcast_multireduce_nested_local_upcast(self):
-    x, y, z, w = [Tensor.rand(1,128).realize() for _ in range(4)]
+    x, y, z, w = [Tensor.rand(1,128,128).realize() for _ in range(4)]
```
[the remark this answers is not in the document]
0xtimmy: yeah, this is what worked for me: https://github.com/tinygrad/tinygrad/pull/4682#discussion_r1610870888

tinygrad/codegen/linearizer.py
```diff
@@ -180,13 +180,21 @@ def render_loop(self, xx:List[Variable], depth:int) -> Tuple[UOp, ...]:
...
+  # helper function, the actual render_reduceop function will get initialize once the indexes get initilized
+  def render_reduceop(self, reduceop:LazyOp, seq: int, accs:Dict[LazyOp, List[UOp]],
+                       loaded_buffers:Dict[Union[MemBuffer, ConstBuffer, LocalBuffer], List[UOp]],
+                      global_idxs, local_idxs, reduce_idxs, upcast_idxs, full_upcast_idxs):
     # define indicies
-    full_upcast_idxs = [Variable(f"_uidx{i}", 0, s-1) for i, s in enumerate(self.full_shape[self.shape_len-self.upcasted:])]
-    reduce_idxs = [Variable(f"ridx{i}", 0, self.full_shape[i]-1) for i in range(self.first_reduce+self.group_for_reduces, self.shape_len-self.upcasted)]  # noqa: E501
     fake_reduce_idxs = [x*0 for x in reduce_idxs]
 
+    # add a local buffer for multistage reduce. # TODO: use local alias
```
Qazalin, line 190: I need to take more time with this part - Can you explain why this was moved from https://github.com/tinygrad/tinygrad/pull/4259/files#diff-63d57daaca59e5495bbb108da1c13e782f4a408db8503e015e8094ac07c87ea6L354

tinygrad/codegen/linearizer.py
```diff
@@ -426,10 +436,9 @@ def ast_parse(self, x:LazyOp, accs: Dict[LazyOp, List[UOp]], offs:Optional[List[
     if x.op in BufferOps: return loaded_buffers[x.arg]
     if x.op in [UnaryOps.CAST, UnaryOps.BITCAST]:
       return [self.uops.add(UOps.BITCAST if x.op is UnaryOps.BITCAST else UOps.CAST,
-                            self.get_base_dtype(x.arg), (u,)) for u in self.ast_parse(x.src[0], accs, offs, loaded_buffers)]
+        self.get_base_dtype(x.arg), (u,)) for u in self.ast_parse(x.src[0], accs, offs, loaded_buffers)]
```
Qazalin, line 439: is this diff extra?

tinygrad/codegen/linearizer.py
0xtimmy: yeah so for sequential reduceops you should be able to just reuse the same buffer which is why for a while this block of code was in the `linearize` function

however, it seems to miss behave on GPU: ex the CI for [this commit](https://github.com/tinygrad/tinygrad/pull/4259/commits/c10986b648787bf1c0753882d7152ca97264b1f7) versus the next (which actually makes each local buffer unique)

I'm honestly not sure why, because I can't get the bug to reproduce on my machine or the tinybox but I could trace the issue to the local buffers and giving each reduceop it's own buffer fixed it

I don't think it has anything to do with the `earlybufs`

tinygrad/codegen/linearizer.py
0xtimmy: I think so, I can remove it

tinygrad/codegen/linearizer.py
0xtimmy: so that it gets run for every reduceop

we could keep it in 'linearize' and give it it's own for loop but it seemed more natural to put it in 'render_reduceop' so we can keep using '-1' to reference the local_buffer in stuff like 'self.sts' or 'global_store'

Qazalin:
@0xtimmy Can you write tests that trigger OptOps like UPCAST and UNROLL with multireduce? handcoded_opts don't cover the full search space.

tinygrad/codegen/linearizer.py
Qazalin: I think moving this codeblock to render_reduceop can be its own noop change, makes this diff more readable.

tinygrad/codegen/linearizer.py
```diff
@@ -433,8 +443,7 @@ def ast_parse(self, x:LazyOp, accs: Dict[LazyOp, List[UOp]], offs:Optional[List[
       return [self.uops.add(UOps.BITCAST if x.op is UnaryOps.BITCAST else UOps.CAST,
                             self.get_base_dtype(x.arg), (u,)) for u in self.ast_parse(x.src[0], accs, offs, loaded_buffers)]
     if x.op in ReduceOps and reduce_acc is None:
-      assert offs is None, "not available if we aren't doing reduce"
-      return accs[x]
+      return [accs[x][i] for i in offs] if offs else accs[x]
```
Qazalin, line 446: nit: `offs is not None`

0xtimmy: working on the added tests and `UNROLL` is (I think the only) case where offs and accs don't match:

ex you get a kernel like this:
```
__kernel void r_2_3(__global float* data0, const __global float* data1) {
  int gidx0 = get_group_id(0); /* 2 */
  int alu0 = (gidx0*3);
  float val0 = data1[alu0];
  float val1 = data1[alu0+1];
  float val2 = data1[alu0+2];
  float alu1 = (val2+val1+val0);
  data0[gidx0] = ((val2-alu1)+(val1-alu1)+(val0-alu1));
}
```
where there's one acc, `alu1`, but three offs, resulting in: `(val2-alu1)`, `(val1-alu1)`, and `(val0-alu1)`

to `ast_parse` this looks like:
```
accs[x] =  UOps.PHI            : dtypes.float              [<UOps.DEFINE_ACC: 5>, <UOps.ALU: 13>] None
offs = [0, 0, 0]
```

0xtimmy: but this only happens when there's multiple reduces because it reduces the unrolled shape down to one value then has to use it in the pre-unrolled shape for farther arithmetic

test/test_linearizer.py
```diff
@@ -202,7 +202,7 @@ def test_early_end_local(self):
     k.hand_coded_optimizations()
     k.linearize()
     self.assertEqual(len(endifs:=[x for x in k.uops if x.uop is UOps.ENDIF]), len(ifs:=[x for x in k.uops if x.uop is UOps.IF]))
-    self.assertEqual(len(barriers:=[x for x in k.uops if x.uop is UOps.BARRIER]), 3)
+    self.assertEqual(len(barriers:=[x for x in k.uops if x.uop is UOps.BARRIER]), 4)
```
Qazalin, line 207: why does GPU need 4 barriers but other backends don't?

Qazalin: oh I see how this could happen, you have a warp size of 32 and issues like this can hide.
I think the best way to catch these is have tests with local size of > 32, can you add some with larger lidx?

Qazalin: I'm fine with changing UOps.BARRIER vins in this case. Correctness is more important.

0xtimmy: sweet, I'll play with it and draw something up

tinygrad/codegen/linearizer.py
```diff
@@ -310,9 +310,18 @@ def upcast_strides(buf:int):
       # end the late reduce loop
       self.load_cache.clear()
 
-      # all local indices which were used for group_for_reduce are not valid any more and should be replaced with fake NumNode(0), since they have
-      # been rewritten with fake end_local_idxs.
-    return (accs, loaded_buffers, fake_reduce_idxs, local_idxs[:self.local_dims] + [NumNode(0) for i in range(self.group_for_reduces)], upcast_idxs)
```
Qazalin, line 315: I can't merge this because it breaks `test/test_linearizer_failures.py TestLinearizerFailures.test_failure_12`
you can see the kernel output with DEBUG=4 
fyi the original change was introduced in https://github.com/tinygrad/tinygrad/pull/3180 - Seems like it's not a well-tested edge case but will potentially break gpt2 on BEAM=2

tinygrad/codegen/linearizer.py
```diff
@@ -215,7 +214,7 @@ def calc_tc_idxs(local_sizes: List[int], aliases: List[List[int]]):
       alias_buf_idxs.append((i, localbuf_idx, buf_idxs,))
 
     # reduce loop
-    loop_ctx = self.render_loop(reduce_idxs, 2)
+    loop_ctx = self.render_loop(reduce_idxs, seq*4+2)
```
Qazalin, line 217: what's the reasoning being doing seq*4

tinygrad/codegen/linearizer.py
```diff
@@ -96,11 +96,11 @@ def global_load(self, i:int, idxs:List[Node], acc:Optional[LazyOp]=None, barrier
     acc_count = 0
     for idx, valid, rep_idx in zip(e_idxs, e_valids, iter_idxs(expand_vars)):
       this_const, idx, valid = (invalid_value, NumNode(0), NumNode(1)) if valid.max == 0 else (const, idx, valid)
-      # todo: when multiple reduceops are supported, clearly disambiguate and test acc load keys are unique for each reduceop
-      key = f"{acc is not None}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
+      key = f"{self.reduceops.index(acc) if acc is not None else None}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
       if key not in self.load_cache:
         if acc is not None:
-          self.load_cache[key] = self.uops.add(UOps.DEFINE_ACC, localtype, loop_ctx, (self.get_reduce_acc(acc), i, acc_count))
+          id = f"{self.reduceops.index(acc)}_{i}"
```
Qazalin, line 102: you don't need this anymore, the DEFINE_ACC will always be deduped because it has different UOps.RANGE vins

tinygrad/codegen/linearizer.py
0xtimmy: @chaosagent for parallel reduces

It looked like `depth` was a way to fuse loops from multiple reduceops; we didn't come to a conclusion but it seemed like the thinking around parallel/sequential reduceops is that `self.reduceops` would become a 2D list. so a `seq` number would be shared amongst parallel and incremented when we went to the next group of parallel reduceops

`*4` because it looked like there were more render loops but this could probably be `*2`

Qazalin (commented):
I think there are two merge blockers for this diff:

1.  Locals issues https://github.com/tinygrad/tinygrad/pull/4259/files#r1623394468
2. PADTO failing tests

This diff should also not change any other kernels. You can use `[run_process_replay]` in your commit message to assert this in CI, example: https://github.com/tinygrad/tinygrad/pull/4807

Are there other fixes/refactors you'd wanna include in this diff before transitioning to scheduling?
The scope of the bounty is to schedule mean+std though I think we can get layernorm and softmax fusion for free too.

0xtimmy:
tensor cores? 

with layernorm the problem is that the `output_shape` equals the `full_shape` despite the reduceop; I've made it work in testing but it would require some changes + more tests

for atomic local load/store the main issue is we need to be able to add barriers `STORE`s, I have a refactor in progress for gated stores but the problem remains that we'd need to add a `BARRIER` to the `STORE` vin

this way we can make sure loads/stores are properly `BARRIER`ed . the refactor could better define the `STORE`'s vin which wuld help

0xtimmy:
draft pr for the gated stores refactor: #4868
seems a bit much for just adding barriers to stores but if this is the direction we want to go I think it's a good start

Qazalin (commented):
I see some new changes, is this ready for re-view?
I could repro the PTX failure on a tinybox

tinygrad/codegen/linearizer.py
```diff
@@ -217,10 +217,12 @@ def calc_tc_idxs(local_sizes: List[int], aliases: List[List[int]]):
       if DEBUG >= 3: print(f"store alias: sts={self.sts[0]} idxs={global_idxs+local_idxs+upcast_idxs}")
     return alias_buf_idxs
 
-  def render_reduceop(self, reduceop:LazyOp, accs:Dict[LazyOp, List[UOp]], loaded_buffers:Dict[Union[MemBuffer, ConstBuffer, LocalBuffer], List[UOp]],
-                      global_idxs, local_idxs, upcast_idxs, full_upcast_idxs, reduce_idxs, fake_reduce_idxs, alias_buf_idxs):
+  def render_reduceop(self, reduceop:LazyOp, seq:int, accs:Dict[LazyOp, List[UOp]], \
```
Qazalin, line 220: why do you need to pass seq here?

tinygrad/renderer/cstyle.py
```diff
@@ -118,7 +118,7 @@ def ssa(prefix:str, u:Optional[UOp]=None):
       elif uop is UOps.STORE:
         assert vin[0].dtype is not None and vin[2].dtype is not None
         rendered_store = self.render_store(r[vin[0]], vin[0].dtype, r[vin[2]], vin[2].dtype, strip_parens(r[vin[1]]), vin[0].uop is UOps.DEFINE_LOCAL)
-        kk(f"if ({r[vin[3]]}) {{ {rendered_store} }}" if len(vin) > 3 else rendered_store)
+        kk(f"if ({r[vin[3]]}) {{ {rendered_store} }}" if len(vin) > 3 and vin[3] is not None else rendered_store)
```
Qazalin, line 121: This breaks the UOp dataclass:
```
@dataclass(eq=False)
class UOp:
  uop: UOps
  dtype: Optional[DType] = None
  vin: Tuple[UOp, ...] = tuple()
```

tinygrad/codegen/linearizer.py
0xtimmy: yeah it could, easy change

tinygrad/codegen/linearizer.py
```diff
@@ -360,7 +375,7 @@ def linearize(self):
       self.buf_uops.append(self.uops.add(UOps.DEFINE_LOCAL, PtrDType(temp_dtype), (), ("temp", self.sts[-1].size)))
 
     # kernel name (before late upcast)
-    self.name = ("r" if self.reduceop else ("C" if all(x.op in BufferOps for x in self.lazyops) else "E")) + \
+    self.name = ("r" if len(self.reduceops)>0 else ("C" if all(x.op in BufferOps for x in self.lazyops) else "E")) + \
```
Qazalin, line 378: i dont think this diff is required, they express the same thing

tinygrad/codegen/linearizer.py
```diff
@@ -259,16 +263,22 @@ def upcast_strides(buf:int):
       loaded_buffers.update({b:self.global_load(self.bufs.index(self.local_alias[reduceop][i]) if i in self.local_alias else i,
         global_idxs+local_idxs+reduce_idxs+full_upcast_idxs) for i,b in enumerate(self.bufs) if b in self.earlybufs})
 
+      def gate_acc(r, idxs): return [
+        UOp.alu(TernaryOps.WHERE, valid.render(self.render_ops, self), acc, self.const(0, r.dtype)) if valid.min == 0 and valid.max == 1 else acc
+        for valid, acc in zip(expand_node(self.sts[self.full_buf_index].expr_idxs(idxs)[1], expand_idxs(idxs)), accs[r])]
+      local_accs = {r: gate_acc(r,global_idxs+local_idxs+reduce_idxs+full_upcast_idxs) for r in accs}
```
0xtimmy, line 269: this part should fix the padto issue
basically it treats the result of reductions like a constant using the shapetracker of the full buf (because fused reducops should both go from `full_shape` -> `output_shape`)

0xtimmy:
recaping: https://github.com/tinygrad/tinygrad/pull/4259#pullrequestreview-2098699680
- Local+PADTO valid idxs have been fixed
- PADTO tests have been fixed
- Tensor cores use the right bufs

Blockers / request for review:
- layernorm fusion, the linearizer needs to be able to do reduceop(s) *and* output the full shape
- atomic store, this diff adds a barrier to `UOps.STORE`s `vin[4]`. The renderer produces gated stores: `{ gate(vin[3]) } if len(vin) > 3 else { no gate }` so in the case that a store must be barriered but has no gate, a `CONST` True is generated to take the place of `vin[3]`
- I couldn't figure out what was wrong here: https://github.com/tinygrad/tinygrad/pull/4259#discussion_r1623394468 but I'm working on making more tests today

Qazalin:
@0xtimmy also, does this diff allow for softmax fusion? I didn't find it in test_linearizer

0xtimmy:
it ought to, I just added some tests to double check

softmax will suffer for the same reason as layernorm though, where it can't both do the reduce and output the full shape

Qazalin:
we can limit this diff's scope to softmax, layernorm and mean+std fusion.
have you written tests for the layernorm problem you're describing? Where can I see that

0xtimmy:
I can whip one up p quick, I don't think it needs to have multiple reduceops even: #4919

0xtimmy:
I think that makes sense

tinygrad/codegen/linearizer.py
```diff
@@ -218,13 +218,14 @@ def calc_tc_idxs(local_sizes: List[int], aliases: List[List[int]]):
     return alias_buf_idxs
 
   def render_reduceop(self, reduceop:LazyOp, accs:Dict[LazyOp, List[UOp]], loaded_buffers:Dict[Union[MemBuffer, ConstBuffer, LocalBuffer], List[UOp]],
-                      global_idxs, local_idxs, upcast_idxs, full_upcast_idxs, reduce_idxs, fake_reduce_idxs,
+                      global_idxs, local_idxs, upcast_idxs, full_upcast_idxs, reduce_idxs,
                       alias_buf_idxs:List[Tuple[int, int, List]]) -> Tuple[List[NumNode|Variable], List[NumNode|Variable]]:
+    fake_reduce_idxs = [x*0 for x in reduce_idxs]
```
Qazalin, line 223: @0xtimmy what's the reason for this?

0xtimmy: I originally didn't pass them through `render_reduceop` but I can

tinygrad/codegen/linearizer.py
```diff
@@ -373,6 +387,7 @@ def linearize(self) -> Linearizer:
     # define indexes
     global_idxs, loop_global_idxs = get_grouped_dims("gidx", 0, self.full_shape[:self.global_dims], 3 if self.opts.has_local else 0)
     local_idxs, loop_local_idxs = get_grouped_dims("lidx", self.global_dims, self.full_shape[self.global_dims:self.first_reduce+self.group_for_reduces], 3 if self.opts.has_local else 0)  # noqa: E501
+    reduce_idxs = [Variable(f"ridx{i}", 0, self.full_shape[i]-1) for i in range(self.first_reduce+self.group_for_reduces, self.shape_len-self.upcasted)]  # noqa: E501
```
0xtimmy, line 390: this also seems like it can be removed because it's repeated later:

```
395 410 : reduce_idxs = [Variable(f"ridx{i}", 0, self.full_shape[i]-1) for i in range(self.first_reduce+self.group_for_reduces, self.shape_len-self.upcasted)]  # noqa: E501
396 411 : alias_buf_idxs = self.index_local_aliases(global_idxs,local_idxs,reduce_idxs,upcast_idxs,full_upcast_idxs)
```

Qazalin:
yeah makes sense, can you remove the redundant parts and make it as small as possible?

tinygrad/codegen/uops.py
```diff
@@ -405,7 +405,16 @@ def push(u):
     self._uops = self._uops[:-1]
 
     # TODO: ifs should be removed and just the store should be gated
-    for u in ifs[::-1]: self._uops.append(UOp(UOps.ENDIF, None, (u,)))
+    dfs_until_bar_cache = {}
+    def dfs_until_bar(u: UOp):
```
Qazalin, line 409: could this be simpler now that we don't barrier stores?

0xtimmy: I don't this has to do with barrier-ing stores, but it can be worked into the existing code

PR #4262, opened 2024-04-23
few more KFD -> AMD

chenyuxyz:
benchmark gemm and default_parallel


chenyuxyz:
AMD=1 GEMM stats is incorrect

geohot:
`fuzz_kfd` should just be removed.

chenyuxyz:
deliberately only changed for core and ci.

PR #4263, opened 2024-04-23
Optimize ptx loops

SzymonOzog:
We add new uops in assembly renderer so this gives some speed improvements. Slightly winning over cuda (0.993 on speed compare script) after this one with 32 major kernels left to beat (checking only those with exec times bigger than 500 us - smaller are very noisy)


geohot:
Great to see you back! Let me know when you think we're good with the bounty and the speed is strictly winning, it's really when we switch the default from CUDA to PTX.

PR #4265, opened 2024-04-23
lint for no new noqa E501

Qazalin:
I think changing lines with a noqa is fine, this catches new noqa: E501 in core tinygrad.

fails:
Screenshot 2024-04-23 at 11 44 28 AM

passes:
image


chenyuxyz:
instead making this a more complex check, we could really remove all existing E501 (mostly long debug message now)

Qazalin:
yeah that'd make this a one liner

PR #4266, opened 2024-04-23
setitem no return value

geohotstan:
don't think setitem returns anything ([jax reference](https://github.com/google/jax/blob/main/jax/_src/basearray.pyi#L51))
a little unsure if this affects any previous DISK tensor implementations

also before the assert for contiguous was always returning True while not actually doing anything.
might've been the problem Cameron on discord was running into (I also accidentally ran into this thinking it was scheduler problem)

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -812,7 +812,9 @@ class Tensor:
     if isinstance(self.device, str) and self.device.startswith("DISK"): self.__getitem__(indices).assign(v)
     # TODO: broadcast v to the shape here, refactor for const v and one way broadcast_shape
     if not isinstance(v, Tensor): v = Tensor(v, self.device, self.dtype)
-    assign_to = self.contiguous().realize().__getitem__(indices)
+    assign_to = self.realize().__getitem__(indices)
+    # NOTE: we check that indices is valid first
+    assert cast(LazyBuffer, self.lazydata).st.contiguous, "setitem target needs to be contiguous"
     # NOTE: contiguous to prevent const folding.
     assign_to.assign(v._broadcast_to(broadcast_shape(assign_to.shape, v.shape)).contiguous()).realize()
 
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -809,11 +809,11 @@ class Tensor:
     return ret
 
   def __setitem__(self, indices, v:Union[Tensor, ConstType]) -> None:
-    # NOTE: check that setitem target is valid first
-    assert all(lb.st.contiguous for lb in self.lazydata.lbs), "setitem target needs to be contiguous"
     if isinstance(self.device, str) and self.device.startswith("DISK"):
       self.__getitem__(indices).assign(v)
       return
+    # NOTE: check that setitem target is valid first
+    assert all(lb.st.contiguous for lb in self.lazydata.lbs), "setitem target needs to be contiguous"
     if not isinstance(v, (Tensor, float, int, bool)): raise TypeError(f"can't set a {type(v).__name__} to a Tensor")
     if not isinstance(v, Tensor): v = Tensor(v, device=self.device, dtype=self.dtype)
     assign_to = self.realize().__getitem__(indices)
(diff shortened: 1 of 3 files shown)


tinygrad/tensor.py
```diff
@@ -808,15 +808,15 @@ def calc_dim(tensor_dim:int) -> int:
...
-  def __setitem__(self, indices, v:Union[Tensor, ConstType]):
-    if isinstance(self.device, str) and self.device.startswith("DISK"): return self.__getitem__(indices).assign(v)
+  def __setitem__(self, indices, v:Union[Tensor, ConstType]) -> None:
+    if isinstance(self.device, str) and self.device.startswith("DISK"): self.__getitem__(indices).assign(v)
     # TODO: broadcast v to the shape here, refactor for const v and one way broadcast_shape
     if not isinstance(v, Tensor): v = Tensor(v, self.device, self.dtype)
     assign_to = self.realize().__getitem__(indices)
     # NOTE: we check that indices is valid first
-    assert self.lazydata.contiguous(), "setitem target needs to be contiguous"
+    assert cast(LazyBuffer, self.lazydata).st.contiguous, "setitem target needs to be contiguous"
```
chenyuxyz, line 817: how does this work for multi? should it be `all(lb.st.contiguous for lb in self.lbs)`?

geohotstan: yeah that makes sense.

the other way is to do `self.lazydata = self.lazydata.contiguous()` to force contiguous inside setitem.
positive: don't need to `.contiguous()` in external implementation before setitem
negative: still need to `.contiguous()` before when using TinyJIT (for reasons I'm still trying to figure out) so there's a difference in usage...

tinygrad/tensor.py
```diff
@@ -808,15 +808,16 @@ def calc_dim(tensor_dim:int) -> int:
         ret = ret.permute(*range(first_dim, first_dim+len(big_shape)), *range(0, first_dim), *range(first_dim+len(big_shape), ret.ndim))
     return ret
 
-  def __setitem__(self, indices, v:Union[Tensor, ConstType]):
-    if isinstance(self.device, str) and self.device.startswith("DISK"): return self.__getitem__(indices).assign(v)
+  def __setitem__(self, indices, v:Union[Tensor, ConstType]) -> None:
+    assert self.lazydata.st.contiguous if isinstance(self.lazydata, LazyBuffer) else all(lb.st.contiguous for lb in self.lazydata.lbs), \
```
chenyuxyz, line 812: `.lbs` works for single LazyBuffer too and make this line simpler

tinygrad/tensor.py
```diff
@@ -808,15 +808,16 @@ def calc_dim(tensor_dim:int) -> int:
...
+      "setitem target needs to be contiguous"
+    if isinstance(self.device, str) and self.device.startswith("DISK"): self.__getitem__(indices).assign(v)
     # TODO: broadcast v to the shape here, refactor for const v and one way broadcast_shape
     if not isinstance(v, Tensor): v = Tensor(v, self.device, self.dtype)
     assign_to = self.realize().__getitem__(indices)
     # NOTE: we check that indices is valid first
```
chenyuxyz, line 818: update the NOTE too if you want to switch the order

tinygrad/tensor.py
```diff
@@ -808,15 +808,18 @@ def calc_dim(tensor_dim:int) -> int:
...
     return ret
 
-  def __setitem__(self, indices, v:Union[Tensor, ConstType]):
-    if isinstance(self.device, str) and self.device.startswith("DISK"): return self.__getitem__(indices).assign(v)
-    # TODO: broadcast v to the shape here, refactor for const v and one way broadcast_shape
-    if not isinstance(v, Tensor): v = Tensor(v, self.device, self.dtype)
+  def __setitem__(self, indices, v:Union[Tensor, ConstType]) -> None:
+    if isinstance(self.device, str) and self.device.startswith("DISK"):
+      self.__getitem__(indices).assign(v)
+      return
```
geohotstan, line 814: maybe we just return the Tensor 🤣

chenyuxyz:
thanks!

PR #4270, opened 2024-04-23
prepare mlperf submission

chenyuxyz:


chenyuxyz:
merging this to start, will fill in system details later.
the idea is to run `setup.sh` first (30 minutes untimed limit) and then `run_and_time.sh`, i picked some parameters so that it fits in compile time limit now. with this green has 94 TFLOPS (3h53m) and red has 56 TFLOPS

chenyuxyz:
@chaosagent @flammit fyi

PR #4273, opened 2024-04-24
pad first batch in resnet

chaosagent:
testing it...


chenyuxyz:
this is in master now

PR #4275, opened 2024-04-24
scheduleitem inputs+outputs -> bufs

geohot:


geohot:
Have a better way to do this

PR #4276, opened 2024-04-24
UNet3D dataloader

francislata:
This PR introduces some modifications to the existing `iterate` dataloader for kits19 dataset so that it supports both training and validation sets.

I have added a `preprocess_dataset` function that will be used by the main training loop to preprocess the kits19 dataset that is similar to [this script](https://github.com/mlcommons/training/blob/master/image_segmentation/pytorch/preprocess_dataset.py).


test/external/mlperf_unet3d/kits19.py
```diff
@@ -0,0 +1,165 @@
+# https://github.com/mlcommons/training/blob/master/image_segmentation/pytorch/data_loading/pytorch_loader.py
```
chenyuxyz, line 1: if this is a whole copy, update the link to a permalink, commented so and don't change anything (include indent)

test/external/external_test_datasets.py
```diff
@@ -0,0 +1,64 @@
...
+
+  def test_kits19_training_set(self):
+    sample_pth = self._create_sample()
+    tinygrad_img, tinygrad_lbl = next(iterate([sample_pth], val=False))
+    preprocessed_img_pth, preprocessed_lbl_pth = self._preprocess_sample(sample_pth, False)
+
+    ref_dataset = PytTrain([preprocessed_img_pth], [preprocessed_lbl_pth], patch_size=(128, 128, 128), oversampling=0.4)
+    ref_img, ref_lbl = ref_dataset[0]
+
+    np.testing.assert_allclose(tinygrad_img[:, 0, :, :, :], ref_img, atol=1e-1)
```
chenyuxyz, line 49: image here value is in range [0, 1]? atol=1e-1 is quite big if that's the case

chenyuxyz (commented):
can you confirm that model_eval still runs correctly?

test/external/external_test_datasets.py
francislata: yeah it is quite big. i was thinking of adding an actual file for both the imaging and segmentation but the imaging file from the `kits19` is too big to push. the mock data is too brittle.

do you think it's better to just link those files somewhere and fetch it from the test?

chenyuxyz: why is the mock data brittle? shouldn't preprocessing to be data independent?

francislata:
@chenyuxyz - I was doing some testing with actual data for debugging and I noticed that even with my seeding, tinygrad respects it meanwhile the reference implementation doesn't. I'll continue investigating this inconsistency.

PR #4277, opened 2024-04-24
[DRAFT] search: delay upcasts until after threads have been added

flammit:


tinygrad/features/search.py
```diff
@@ -81,20 +81,24 @@ def bufs_from_lin(lin:Linearizer, allocate:bool=True) -> List[Buffer]:
...
     if a.axis is not None and lin.full_shape[a.axis] == a.amt and Opt(a.op, a.axis, 0) in actions: continue
     lin2 = lin.copy()
     try:
       lin2.apply_opt(a)
-      up, lcl, tc_up = 1, 1, prod(tc.dims)//prod([x[1] for x in tc.threads]) if (tc:=lin2.tensor_core) else 1
-      for s,c in zip(lin2.full_shape, lin2.colors()):
-        if c in {"magenta", "yellow"}: up *= s
-        elif c in {"cyan", "green", "white"}: lcl *= s
-      if up//tc_up > max_up or lcl > max_lcl: continue
+      if _count_upcast_local(lin2) > (max_up, max_lcl): continue
```
chenyuxyz, line 101: this is different - if first return term is <= max_up, the second term is not compared

tinygrad/features/search.py
```diff
@@ -81,20 +81,24 @@ def bufs_from_lin(lin:Linearizer, allocate:bool=True) -> List[Buffer]:
...
+  for s,c in zip(lin.full_shape, lin.colors()):
+    if c in {"magenta", "yellow"}: up *= s
+    elif c in {"cyan", "green", "white"}: lcl *= s
+  return up//tc_up, lcl
+
 # get dictionary of all possible actions
 def get_linearizer_actions(lin:Linearizer, include_0=True) -> Dict[int, Linearizer]:
   acted_lins, max_up, max_lcl = {0:lin} if include_0 else {}, getenv("BEAM_UPCAST_MAX", 256), getenv("BEAM_LOCAL_MAX", 256)
   for i,a in enumerate(actions):
+    if _count_upcast_local(lin)[1] == 1 and a.op in [OptOps.UPCAST, OptOps.UNROLL, OptOps.PADTO]: continue # add threads first to save running time
```
chenyuxyz, line 95: i think there are small kernels that the best / only action is to upcast all or unroll all

flammit: mmm - gotcha.  there's a better way to do this to cover that case...

tinygrad/features/search.py
flammit: hahaha funny I wanted to disable the local max comparison anyway =P

PR #4284, opened 2024-04-24
Ptx store gate cast to bool

SzymonOzog:
Fixes the invalid ptx found by matmul fuzzer


SzymonOzog:
> cool! can you add an test entry below `https://github.com/tinygrad/tinygrad/blob/dbe3e1d548c429fb65bff5e08d9b0228e14384c1/.github/workflows/benchmark.yml#L127`
> 
> these require real devices so we don't have them in regular ci yet

updated

PR #4290, opened 2024-04-25
optim stores lr at least in float and use float in LARS trust region

chenyuxyz:


chenyuxyz:
float32 lr is merged. where rewrite is relatively small and will do later if needed

PR #4302, opened 2024-04-26
make ring allreduce chunks a multiple of 2^n if possible

chenyuxyz:
in resnet, instead of chunking as [43691, 43691, 43691, 43691, 43690, 43690], chunk as [43712, 43712, 43680, 43680, 43680, 43680] and those can have 32 local.

more than 2X faster for the applicable kernels and overall 1% for resnet


unknownusername504:
On the "if possible" part, would padding zeroes make sense as an optimization?

chenyuxyz:
there's already a PADTO optimization, which can achieve similar effect with a better search algorithm at lower level.

the difference is that it's a waste to chunk into arbitrary sizes then pad to 32 if chunking into a multiple of 32 is possible.

PR #4306, opened 2024-04-26
Multireduce Kernels: uops insert cursor

0xtimmy:
Adds a cursor property to the uops graph so that the linearizer can more easily add ops out of order
It will be useful for linearizing kernels with multiple reduceops, because it will let us render each reduceop recursively in ast_parse.

Rendering the reduceops in the ast_parse avoids the need to figure out their order

This somewhat expands on the functionality of `insert_before`, but 1) is based off a uop so we don't have to keep incrementing the index and 2) means we can using all the existing functions without having to add `insert_before` as a parameter

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -8,7 +8,7 @@ from tinygrad.helpers import colored, DEBUG, prod, getenv, to_function_name
 from tinygrad.ops import LazyOp, UnaryOps, BinaryOps, TernaryOps, ReduceOps, ConstBuffer, MemBuffer, BufferOps
 from tinygrad.shape.shapetracker import ShapeTracker
 from tinygrad.shape.symbolic import Variable, NumNode, Node, SumNode, MulNode, DivNode, ModNode, LtNode, AndNode, create_lt_node
-from tinygrad.codegen.kernel import LocalBuffer, Kernel
+from tinygrad.codegen.kernel import LocalBuffer, Kernel, TensorCore
 from tinygrad.features.image import to_image_idx
 
 from tinygrad.codegen.uops import UOps, UOp, UOpGraph
@@ -44,7 +44,7 @@ class Linearizer(Kernel):
     return self.uops.add(UOps.ALU, dtype, (a, render_b), op)
 
   # NOTE: the consts have to be cached for deduping of downstream uops to work
-  def const(self, b:ConstType, dtype:DType=dtypes.int32, insert_before=None) -> UOp:
+  def const(self, b:ConstType, dtype:DType=dtypes.int32, insert_before:Optional[UOp]=None) -> UOp:
     return self.uops.add(UOps.CONST, dtype, tuple(), b, insert_before=insert_before)
 
   def cast(self, val: UOp, dtype) -> UOp: return self.uops.add(UOps.CAST, dtype, (val,)) if val.dtype != dtype else val
@@ -69,7 +69,7 @@ class Linearizer(Kernel):
     AndNode: lambda self,ops,ctx:
       functools.reduce(lambda a,b: ctx.uop_alu_idx(a, b, ops, ctx, BinaryOps.MUL, dtype=dtypes.bool), self.nodes[1:], self.nodes[0].render(ops,ctx)) }
 
-  def global_load(self, i:int, idxs:List[Node], acc=None, barrier:Optional[UOp]=None) -> List[UOp]:
+  def global_load(self, i:int, idxs:List[Node], acc=None, barrier:Optional[UOp]=None, insert_before:Optional[UOp]=None) -> List[UOp]:
     buf = self.bufs[i]
     localtype = self.get_base_dtype(buf.dtype if acc is None else self.reduceop.dtype)
     const = buf.val if isinstance(buf, ConstBuffer) else acc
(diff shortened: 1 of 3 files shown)


chenyuxyz:
can you given an example where this will be useful? it's bad to add a stateful variable for convenience alone

0xtimmy:
of course. it's to set up for rendering multiple reduceops in order, I drafted a PR here: #4307 with the full change

here's an example of code where the reduceops would need to be ordered recursively:
if they're being rendered by ast_parse, it could hit r1 before r0
in this case it would need to go back to before the loop for r1, write the loop for r0, then go back and finish r1

```
 def test_multireduce_multioutput_fancy(self):
    Tensor.manual_seed(0)
    axis = 0
    x = Tensor.randn(32, 32).realize()
    y = Tensor.randn(1, 32).realize()

    r0 = x.sum(axis=axis, keepdim=True)
    out0 = r0 + y
    out1 = r0 * y
    r1 = (x - out0).sum(axis=axis, keepdim=True)
    out2 = r1 + 10
    check_schedule([out2, out0, out1], 1)
    # NOTE: this doesn't compile in Linearizer - KeyError: LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BufferOps.LOAD, src=(), arg=MemBuffer(idx=3, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(32, 32), strides=(32, 1), offset=0, mask=None, contiguous=True),)))),), arg=((0,), dtypes.float)) #L430
    check_schedule([out0, out1, out2], 1)
```

chenyuxyz:
can you pass the insertion index or uop as input param while recursion? pure function is a lot easy to reason about and test even if that means one more input (still better than one more state)

0xtimmy:
For sure, but I'll have to change a number of functions. I'll put up a commit that implements all changes

0xtimmy:
This version uses `insert_before` instead of the cursor

I added another line to the uops.add() to handle a UOp for `insert_before` in addition to an int, that way calling functions don't have to manage an index

tinygrad/codegen/linearizer.py
```diff
@@ -410,28 +411,33 @@ def linearize(self):
...
-                                      for u in self.ast_parse(x.src[0], acc, offs, loaded_buffers)]
+    if x.op is UnaryOps.CAST: return [self.uops.add(UOps.BITCAST if x.arg[1] else UOps.CAST, self.get_base_dtype(x.arg[0]), (u,), x.arg[0], insert_before=insert_before) \
+                                      for u in self.ast_parse(x.src[0], acc, offs, loaded_buffers, insert_before=insert_before)]  # noqa: E501
     if x.op in ReduceOps and not do_reduce:
-      assert offs is None, "not available if we aren't doing reduce"
-      return acc
-
-    values = [self.ast_parse(v, acc, offs, loaded_buffers, loop_ctx=loop_ctx, cache=cache) for v in x.src]
+      if x not in self.reduce_acc:
+        load_cache_copy,self.load_cache = self.load_cache,{}
```
Qazalin, line 423: I don't understand this change

Qazalin:
I think the ast_parse change is a functional diff. we need to separate that with tests.
A PR adding `insert_before` as a standalone diff is good to merge.

tinygrad/codegen/linearizer.py
0xtimmy: yes and no

if control flow reaches here it means `ast_parse` has run into a reduceop that hasn't been rendered, so it needs to go back before any loop or if statements, render the reduceop, then it can continue with the parse

with only one reduceop I don't think it's safe to keep `self.load_cache` because I'm not sure you ever run into a case where you have to leave a context

but with >1 reduceop you will, and so you'll get a UOp graph like below, where the second reduceop will try to reference a future load that was in `self.load_cache`
0: `ast_parse` hits this reduceop first
1: `render_reduceop` loads the value it needs
2: `ast_parse` (from within `render_reduceop`) runs into another reduceop
3: renders it here
4: `ast_parse` tries to reference a future load because it's in `self.load_cache`
```
   0 UOps.DEFINE_GLOBAL  : ptr.dtypes.float          []                               (0, 'data0', True)
   1 UOps.DEFINE_GLOBAL  : ptr.dtypes.float          []                               (1, 'data1', False)
   2 UOps.SPECIAL        : dtypes.int                []                               (0, 'gidx0', 128)
   3 UOps.DEFINE_ACC     : dtypes.float              []                               0.0
   4 UOps.CONST          : dtypes.int                []                               0
   5 UOps.CONST          : dtypes.int                []                               64
   6 UOps.ALU            : dtypes.int                [2, 5]                           BinaryOps.MUL
   7 UOps.LOOP           : dtypes.int                [4, 5]                           None                <- 3
   8 UOps.ALU            : dtypes.float              [13, 3]                          BinaryOps.ADD       <- 4
   9 UOps.PHI            : dtypes.float              [3, 8, 7]                        None
  10 UOps.ENDLOOP        :                           [7]                              None
  11 UOps.LOOP           : dtypes.int                [4, 5]                           None                <- 0
  12 UOps.ALU            : dtypes.int                [6, 11]                          BinaryOps.ADD
  13 UOps.LOAD           : dtypes.float              [1, 12]                          None                <- 1
  14 UOps.ALU            : dtypes.float              [13, 9]                          BinaryOps.SUB       <- 2
  15 UOps.ALU            : dtypes.float              [14, 3]                          BinaryOps.ADD
  16 UOps.PHI            : dtypes.float              [3, 15, 11]                      None
  17 UOps.ENDLOOP        :                           [11]                             None
  18 UOps.STORE          :                           [0, 2, 16]                       None
```

0xtimmy:
Bet, I'll put up a PR for just the `insert_before` change and look for a cleaner way to handle the load_cache

tinygrad/codegen/linearizer.py
Qazalin: I think copying the load cache hides the problem instead of solving the root cause.

PR #4307, opened 2024-04-26
Multireduce Kernels - Move render_reduceop call to ast_parse

0xtimmy:
Having multiple reduceops and multiple outputs means that:
```
reduceops = [x for x in self.lazyops if x.op in ReduceOps]
```

doesn't necessarily produce an in-order list of reduceops, so we can't just render them by running through a loop
this moves the `render_reduceop` call to within `ast_parse` so that they can be properly ordered by recursion, ordering them elsewhere requires finding their dependencies which was pretty slow

it uses the UOpGraph.cursor to write the reduce before any for loops or if statements, so if we run into a reduceop from within a reduceop it can go back and render it in the right place

it's a solution to the problem @Qazalin pointed out in #4220

tbh I don't know if this is the most elegant solution but it seemed like the most robust

### DIFF
--- a/test/test_uop_graph.py
+++ b/test/test_uop_graph.py
@@ -46,5 +46,19 @@ class TestUOpGraph(unittest.TestCase):
     self.assertEqual(out.uop, UOps.CONST)
     self.assertEqual(out.arg, 0)
 
+  def test_cursor(self):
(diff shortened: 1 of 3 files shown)


Qazalin:
@0xtimmy status?

PR #4308, opened 2024-04-26
kernel: fix calculation of smem size to ignore UNROLL

flammit:
The previous calculation included UNROLLs which don't get stored in temp shared_memory, which meant that GROUPs or UPCAST/UNROLLS when alreading GROUP'ing were being incorrectly pruned too early.

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -443,8 +443,8 @@ class Kernel:
     else: amt = -1
 
     if self.reduceop and (opt.op in {OptOps.GROUP, OptOps.GROUPTOP} or (self.group_for_reduces and opt.op not in {OptOps.NOLOCALS, OptOps.PADTO})):
-      acc_sz = dt.base.itemsize if isinstance((dt:=self.reduceop.dtype), ImageDType) else dt.itemsize
-      upcast_sz = prod(self.full_shape[self.shape_len-self.upcasted:])
+      acc_sz, upcast_idx = dt.base.itemsize if isinstance((dt:=self.reduceop.dtype), ImageDType) else dt.itemsize, self.shape_len-self.upcasted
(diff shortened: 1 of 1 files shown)


tinygrad/codegen/kernel.py
```diff
@@ -443,8 +443,8 @@ def apply_opt(self, opt:Opt, append_opt:bool=True):
     else: amt = -1
 
     if self.reduceop and (opt.op in {OptOps.GROUP, OptOps.GROUPTOP} or (self.group_for_reduces and opt.op not in {OptOps.NOLOCALS, OptOps.PADTO})):
-      acc_sz = dt.base.itemsize if isinstance((dt:=self.reduceop.dtype), ImageDType) else dt.itemsize
-      upcast_sz = prod(self.full_shape[self.shape_len-self.upcasted:])
+      acc_sz, upcast_idx = dt.base.itemsize if isinstance((dt:=self.reduceop.dtype), ImageDType) else dt.itemsize, self.shape_len-self.upcasted
+      upcast_sz = prod([1 if a != b else a for a,b in zip(self.full_shape[upcast_idx:], self.sts[0].shape[upcast_idx:])])
```
chenyuxyz, line 447: `prod([a for a,b in ... if a != b])`

chenyuxyz (commented):
is there a perf gain out of this?

flammit:
> is there a perf gain out of this?

I had hoped so, but didn't see any gains on the current best `TRAIN_BEAM=3`.  It should make a bigger difference in areas like GROUP after TC, but still testing it.

flammit:
Looking at the test_linearizer_failures now... sorry.

tinygrad/codegen/kernel.py
chenyuxyz: hmm why does this fail...?

chenyuxyz: oops sorry!

PR #4310, opened 2024-04-27
run clang using stdout

smurfd:
using `subprocess.run` with arguments `check=True, capture_output=True` and returning it with `.stdout` then you can use stdout instead of a file


smurfd:
nevermind, not a fan. ill get back if i fix the errors

PR #4311, opened 2024-04-27
remove an always noop shrink in _pool

chenyuxyz:
leftover from slice -> pad and shrink cleanup


chenyuxyz:
oh nvm that's only for conv...

PR #4326, opened 2024-04-27
run clang and capture output from stdout

smurfd:
trying to have clang use stdout instead of temporary file


smurfd:
nevermind, works localy apparently not as runners...

PR #4330, opened 2024-04-28
wip conv2d weight expand bs then rest

chenyuxyz:
resnet 100 TFLOPS -> 118 TFLOPS. more kernel counts and memory though


chenyuxyz:
no longer faster after #4354

Issue #4109, opened 2024-04-08
Bug in View._reshape_mask


PaulGustafson:
The following code incorrectly evaluates to True:

```View(shape=(2, 3), strides=(0, 0), offset=0, mask=((1, 2), (0, 2)), contiguous=False).reshape((3,2)) == View(shape=(3, 2), strides=(0, 0), offset=0, mask=((1, 3), (0, 2)), contiguous=False)```

This is wrong because the mask on the left has 2 elements, but the mask on the right has 4.

The source of the error is this branch in View._reshape_mask:

```    
else:
      next_mask = next(r_masks, (0, 1))
      # combine if the mask can unfold continuously
      if mask != (0, old_dim) and next_mask[1] - next_mask[0] != 1: return view.mask, True
      mask, old_dim = (next_mask[0] * old_dim + l, (next_mask[1] - 1) * old_dim + r), old_dim * next(r_shape, 1)
```

**Comment by PaulGustafson** [comment]
Another couple examples -- 

```View(shape=(3, 4), strides=(4, 1), offset=0, mask=((2, 3), (0, 3)), contiguous=False).reshape((4,3))```
and
```View(shape=(4, 22), strides=(22, 1), offset=0, mask=((1, 2), (0, 20)), contiguous=False).reshape((22,4))```

**Comment by chenyuxyz** [comment]
it probably needs to explicitly check the end condition for `next_mask`

**Comment by PaulGustafson** [comment]
One fix would be to replace the last return statement with

```  
if prod([m[1] - m[0] for m in r_masks]) != prod([m[1] - m[0] for m in new_mask]): return view.mask, True

return tuple(reversed(new_mask)), False
```

But we/someone should walk through the code to make sure they understand what's going on (I don't -- the nexts + two levels of dims + strides are pretty complicated lol).

PR #4332, opened 2024-04-28
Fix in `_reshape_mask`

obadakhalili:
Addresses the issue in #4109.

The issue appears when `old_dim` has n instances of `next_stride` but with a remainder. This remainder will push the position of the mask and might cross the axis boundary.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -987,23 +987,15 @@ class TestOps(unittest.TestCase):
     helper_test_op([(3,3)], lambda x: torch.nn.functional.pad(x, (0,0,0,0), value=1), lambda x: x.pad((None, None), value=1))
 
   def test_pad_reshape(self):
-    a = Tensor.ones((2,)).reshape((1, 2))
-    b = torch.ones((2,)).reshape((1, 2))
-    helper_test_op([],
-                   lambda: torch.nn.functional.pad(b, (0, 1, 1, 0)).reshape((3, 2)),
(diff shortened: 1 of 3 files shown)


test/test_ops.py
```diff
@@ -986,6 +986,25 @@ def test_pad(self):
     helper_test_op([(3,3)], lambda x: torch.nn.functional.pad(x, (0,0,3,4), value=1), lambda x: x.pad(((3,4), None), value=1))
     helper_test_op([(3,3)], lambda x: torch.nn.functional.pad(x, (0,0,0,0), value=1), lambda x: x.pad((None, None), value=1))
 
+  def test_pad_reshape(self):
+    a = Tensor.ones((2,)).reshape((1, 2))
```
chenyuxyz, line 990: why does it need to only test Tensor.ones? wouldn't random tensor with shape `(1, 2)` work?

chenyuxyz (commented):
Thanks for looking into this! Can you also add the case in test/unit/test_shapetracker? test_ops is quite end2end for this fix and would be nice to include something more local

test/test_ops.py
obadakhalili: Oh, it would work actually. I just had to stick with `Tensor.ones` at the beginning to reproduce the exact faulty view mentioned in the example and kept using it afterward. I will update it to use randomly generated tensors from `helper_test_op`.

chenyuxyz:
thanks!

PR #4333, opened 2024-04-28
use tolist instead of numpy for extracting parameters in onnx

geohotstan:


tinygrad/tensor.py
```diff
@@ -222,8 +222,6 @@ def tolist(self) -> Union[Sequence[ConstType], ConstType]:
     """
     Returns the value of this tensor as a nested list.
 
-    Currently this only works for flattened tensors.
-
```
geohotstan, line 226: does supporting nested lists mean tensors that aren't flat also work?
image

extra/onnx.py
```diff
@@ -13,16 +14,16 @@
   from onnx.mapping import TENSOR_TYPE_TO_NP_TYPE
   tensor_dtype_to_np_dtype = lambda x: TENSOR_TYPE_TO_NP_TYPE[x]
 
-# global numpy cache for parameters
-numpy_cache = {}
-def safe_numpy(t) -> np.ndarray:
+# global Tensor -> list cache for parameters
+to_python_cache = {}
```
chenyuxyz, line 18: can it work with `@functools.lru_cache` and don't need to manually maintain cache?

extra/onnx.py
```diff
@@ -13,16 +14,16 @@
...
+def safe_python(t, tobytes=False) -> ConstType:
```
chenyuxyz, line 19: return type does not seem correct, it should at least union with bytes? also can we rename the function to `to_python_const` or similar? it's not clear to me what's unsafe here

chenyuxyz (commented):
very happy to see this cleanup!

PR #4334, opened 2024-04-28
update setitem tests to test for currently supported cases

geohotstan:
Currently all cases of setitem that involves basic indexing (no copy) works.

For advanced indexing setitem, I think the most correct way of supporting it is the symbolic Tensor stuff. (a wrong way is the hacky setitem I wrote before that doesn't use assign)
I'm still trying to write a PoC of an idea I have. I think supporting advanced indexing setitem and O(n) embedding are basically the same.


test/test_setitem.py
```diff
@@ -4,28 +4,50 @@
...
+    cases = (
+      ((6,6), (slice(2,4), slice(3,5)), Tensor.ones(2,2)),
+      ((6,6), (slice(2,4), slice(3,5)), 1.0),
+      ((6,6), (slice(2,4), slice(3,5)), Tensor([1.,2.])),
+      ((4,4,4,4), (Ellipsis, slice(1,3), slice(None)), Tensor(4)),
+      ((4,4,4,4), (Ellipsis, slice(1,3)), 4),
+      ((4,4,4,4), (slice(1,3), slice(None), slice(0,4,2)), 4),
+      ((4,4,4,4), (slice(1,3), slice(None), slice(None), slice(0,3)), 4),
+      ((6,6), (slice(1,5,2), slice(0,5,3)), 1.0),
+      ((6,6), (slice(5,1,-2), slice(5,0,-3)), 1.0),
```
geohotstan, line 16: added some cases here because `test_indexing` doesn't really test for non-advanced indexing setitem.

chenyuxyz:
we can get these tests in first. is this ready?

geohotstan:
> we can get these tests in first. is this ready?

Yeah sorry, I was away for a few days.

chenyuxyz:
thanks!

PR #4336, opened 2024-04-28
use JIT=2 for gpt2 beam mac benchmark

chenyuxyz:


chenyuxyz:
still fails with jit=2, more like a legit bug

PR #4338, opened 2024-04-28
Symbolic shapes in mean and var calculations for tensors

davidjanoskyrepo:
Adding support with tests.


chenyuxyz:
not this. it needs to create a kernel that uses variable (and verify in jit), not one time compute the underlying variable

PR #4320, opened 2024-04-27
Add `insert_before` to Linearizer Functions

0xtimmy:
This will let the linearizer render chunks of uops at various points in the graph
It will be useful when fusing multiple reduceops because we'll be able to render them recursively (and then we won't have to order them beforehand)

re: an improvement to #4306

PR #4341, opened 2024-04-28
uops.add change insert_before argument order

0xtimmy:
re: #4320, this seemed like the least-cluttering order to put insert_before, lmk what you think


tinygrad/codegen/linearizer.py
```diff
@@ -41,14 +41,14 @@ def expand_node(node:Node, idxs:Optional[Tuple[Union[Variable, NumNode], ...]]=N
 class Linearizer(Kernel):
   def uop_alu_idx(self, a:UOp, b, ops, ctx:Linearizer, op, dtype:DType=dtypes.int32):
     render_b:UOp = cast(UOp, (NumNode(b) if not isinstance(b, Node) else b).render(ops, ctx))
-    return self.uops.add(UOps.ALU, dtype, (a, render_b), op)
+    return self.uops.add(UOps.ALU, dtype, (a, render_b), arg=op)
 
   # NOTE: the consts have to be cached for deduping of downstream uops to work
   def const(self, b:ConstType, dtype:DType=dtypes.int32, insert_before:Optional[UOp|int]=None) -> UOp:
-    return self.uops.add(UOps.CONST, dtype, tuple(), b, insert_before=insert_before)
+    return self.uops.add(UOps.CONST, dtype, tuple(),  insert_before, b)
```
chenyuxyz, line 48: space

chenyuxyz (commented):
why is `insert_before` needed in linearizer `const` and `cast` input?

0xtimmy:
we want to be able to write uops to the graph out-of-order, specifically so that reduceops can be rendered recursively, ex here: https://github.com/tinygrad/tinygrad/pull/4306#discussion_r1581852023

I figure it would be more robust if both const and cast can also be rendered out of order so, ex when we're rendering the loop for the reduceop, we can make sure the consts come before the loop in the uop graph

chenyuxyz:
oh i see. yeah that's fine. `const` and `cast` mean generating a const variant or generating a cast variant out of input in other layers so i was confused

chenyuxyz:
nah it should not be before arg. the test changes are bad. maybe it's fine as is

chenyuxyz:
the direction that might make sense is to separate constructing an uop to adding the uop to graph. `insert_before` is only needed in `add`, not in construction

0xtimmy:
true, I could make `render_loop` or `render_reduction` return a chunk of uops and then have the calling function add them

I'll play with it

PR #4350, opened 2024-04-29
reduceop children chasing can be simpler

Qazalin:
This diff simplifies can_chase to a single question: is the realized child already paired with another reduceop?

Deletes:
1. `reduce_for_op[tr] != r`, this is an unreachable path. reduce_for_op updates once and doesn't revisit the same reduceop (r)
2. can_chase in multi output, It is always true in this code path


Qazalin:
status: will be complete after I add a few more fuse tests

Qazalin:
@geohot process replay tests (openpilot, resnet, hlb_cifar) confirm this diff has no behavior change.

What was the intention behind `and reduce_for_op[tr] != r`? I think it was a part of the lazy rewrite https://github.com/tinygrad/tinygrad/pull/2878/files#diff-566a7254fa321896c55ed04f6d3eb9cf38ac84ee952b4e44a107c006cf886b16R260

Qazalin:
merging this, pretty confident about the change.

PR #4359, opened 2024-04-30
BERT: Checkpoint loading tests

Eliulm:
I think a standalone test to check whether checkpoint loading functionality is correct, is quite helpful to catch regressions caused by incorrect weight mapping or mixed precision changes in the forward pass. MLperf recommends [this](https://github.com/mlcommons/training/blob/master/language_model/tensorflow/bert/README.md#stopping-criteria).

### DIFF
--- a/examples/mlperf/helpers.py
+++ b/examples/mlperf/helpers.py
@@ -230,6 +230,11 @@ def init_bert_from_checkpoint(model, ckpt_dir:str):
         t = t.reshape(*x.shape)
       x.assign(t)
 
+def get_data_bert(GPUS:list[str], it):
+  data: dict[str, Tensor] = next(it)
+  for key in data.keys(): data[key].shard_(GPUS, axis=0)
+  return data
+
 @functools.lru_cache(maxsize=None)
 def load_tf_weights_to_dict(checkpoint_path):
   os.environ['TF_CPP_MIN_LOG_LEVEL'] = '2'
--- a/examples/mlperf/helpers.py
+++ b/examples/mlperf/helpers.py
@@ -4,6 +4,9 @@ import numpy as np
 from tinygrad.nn import state
 from tinygrad.tensor import Tensor
 from tinygrad.dtype import dtypes
+from tinygrad.nn.state import get_state_dict
+
+from extra.models.bert import BertForMLPerf
 
 #
 # checkpointing utils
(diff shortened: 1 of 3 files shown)


test/external/mlperf_bert/external_test_checkpoint_loading.py
```diff
@@ -0,0 +1,82 @@
...
+
+from tinygrad.tensor import Tensor
+from tinygrad.device import Device
+from tinygrad.engine.jit import TinyJit
+from tinygrad.helpers import getenv
+from tinygrad.nn.state import get_state_dict
+from examples.mlperf.helpers import get_mlperf_bert_model, init_bert_from_checkpoint
+from examples.mlperf.dataloader import batch_load_val_bert
+
+# Copied from `examples/mlperf/model_train.py` in `train_bert``
```
chenyuxyz, line 18: don't copy code like this. it should either reuse or be refactored out

examples/mlperf/helpers.py
```diff
@@ -4,6 +4,9 @@
 from tinygrad.nn import state
 from tinygrad.tensor import Tensor
 from tinygrad.dtype import dtypes
+from tinygrad.nn.state import get_state_dict
+
+from extra.models.bert import BertForMLPerf
```
chenyuxyz, line 9: generally the direction should be helpers imports nothing and individual example imports from target. if it's only used in bert, it stays in bert

test/external/mlperf_bert/external_test_checkpoint_loading.py
Eliulm: Ok, but I think we do not want to refactor out the JITs, right? - So, drop the tests?

chenyuxyz: you can refactor a version that does not always use jit, and only apply jit wrapper when needed

Eliulm: Sorry, I meant the eval_step and train_step functions. I think they are fine where they are.

chenyuxyz: my point is that if you want to add test, don't copy code like this. it will get out of sync quickly

think about the best way to make sure the part you want to test is good after 6 months, even if you are no longer working actively on it

PR #4363, opened 2024-04-30
Update links in new docs

smurfd:
Updating links in `docs/` to point towards https://github.com/tinygrad/tinygrad/blob/master instead of https://tinygrad.github.io/, ie making them work :)

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -17,13 +17,13 @@ The `LazyBuffer` graph specifies the compute in terms of low level tinygrad ops.
 
 ## Scheduling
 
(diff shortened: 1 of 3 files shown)


wozeparrot:
I feel like some of these should point to the hosted docs instead, like references to tensor.py and nn

smurfd:
Im more than happy to change if something is not correct :)
But i believe the hosted docs are .md files, like: https://github.com/tinygrad/tinygrad/blob/master/docs/tensor.md and the links touched here are only .py and from context it makes some sense. They also point to subfolders like `tinygrad/tensor.py` for example or `tinygrad/nn/__init__.py`

wozeparrot:
I think it might be better to rewrite them to point to the tensor.md docs since that's the docs that are nicer for new users.

smurfd:
thanks, updated with nn.md and tensor.md. but kept one for tensor where the example copied the `sparse_categorical_crossentropy()` function

PR #4369, opened 2024-05-01
pad resnet training data with training data mean

chenyuxyz:
update model_train resnet to pad both training and eval

started a run in https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/hm5gm8ku?nw=nwuserchenyuxyz


chenyuxyz:
https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/llzhr46j/overview?nw=nwuserchenyuxyz

trains to 0.76042 on epoch 35

PR #4235, opened 2024-04-20
fuse expands

Qazalin:
results so far:

`SAVE_SCHEDULE=1 python3 test/models/test_mnist.py TestMNIST.test_adam_onestep`

Screenshot 2024-04-20 at 5 52 35 PM

`python3 examples/llm.c/export.py`
Screenshot 2024-04-20 at 5 21 00 PM

todo:
- [x] Assign
- [ ] less lines

PR #4370, opened 2024-05-01
[wip] resnet batchnorm backward fusion spec

chaosagent:
small example for easy inspection for now


Qazalin:
thanks - adding this to the scheduler roadmap!

test/test_schedule.py
```diff
@@ -593,5 +593,22 @@ def test_prefer_half_buffer(self):
...
+      bn = nn.BatchNorm2d(16)
+      bn.weight.requires_grad = bn.bias.requires_grad = x.requires_grad = True
+      fw = bn(x).contiguous_backward().relu().contiguous()
+      fw.sum().backward()
+      # we want to minimize number of passes over buffers of same size as x
+      # start: 12 kernels (some extraneous from constructing this test case)
+      # easy case: merge 4 reduces in backward into 1
+      # double reduce case: merge stat calculations from 2 to 1 (be careful of long reduces!)
+      # sum(x - \bar{x}): one kernel just calculates this, can be eliminated
+      # pre-expand fusion: is it fast? -2 kernels possible, 1 fw, 1 bw
```
Qazalin, line 608: does this refer to `E_2_16_64n1 `+`E_2048`
(graph ref: https://tiny-tools-client.vercel.app/?id=f7b72a41bad14974970329924c89b2c0)
?
#4235 could do this, it won't because `<LB METAL (2, 16, 8, 8) float (<UnaryOps.CAST: 3>, None)>` is forced_realize. I think it breaks the API if we fuse a forced_realize parent with its child.

chaosagent: I am referring to E2_16_64n1 (full forward with relu) and E2_16_64 (full backward through batchnorm). The first can be fused with the next conv, and the latter can be fused with the next backward conv. (E_2048 simulates the backward from the next layer, plus relu backward)

This test case does not have the convs to focus on batchnorm, so it cannot happen here. will add more cases.

chaosagent:
the scheduler change is a little tricky, since you need to make sure that each grouping is a contiguous sub-DAG. My solution to this is currently to do the grouping while toposorting, which should work for the specific bn training case, but is it possible to make it clean?

Qazalin:
I need to think about the scheduler change a bit more, but in general we don't wanna do merge schedules, if there is grouping to be done it should be here https://github.com/tinygrad/tinygrad/blob/master/tinygrad/engine/schedule.py#L225-L228

tinygrad/engine/schedule.py
```diff
@@ -92,6 +92,18 @@ def _schedule_group(outs:Tuple[LazyBuffer, ...], realizes:Dict[LazyBuffer, None]
       ast.append(LazyOp(BufferOps.STORE, (op, ), MemBuffer(i, out.dtype, output_view)))
   return _LBScheduleItem(tuple(ast), outs, tuple(inputs), var_vals)
 
+def _replace_bufis(ast: LazyOp, old_lbs: Tuple[LazyBuffer, ...], new_lbs: Tuple[LazyBuffer, ...]):
+  new_arg = MemBuffer(new_lbs.index(old_lbs[ast.arg.idx]), ast.arg.dtype, ast.arg.st) if ast.op in [BufferOps.LOAD, BufferOps.STORE] else ast.arg
+  return LazyOp(ast.op, tuple(_replace_bufis(x, old_lbs, new_lbs) for x in ast.src), new_arg)
+
+def _merge_prescheduled(prescheduled: List[_LBScheduleItem]):
```
Qazalin, line 99: I've gone through this route in multioutput,
https://github.com/tinygrad/tinygrad/blob/6c2cb8ea3fd70208714e67d680c3360f106256ac/tinygrad/engine/schedule.py#L86
I think you need to rebuild the entire AST.

chaosagent:
i implemented `deferring contiguous reduces until you run out of nodes in queue`. it seems to work quite well, and it passed all the tests I had (very surprised that this happened smoothly). huge mess though.

i am prototyping with merge_prescheduled because i need to toposort to find these fusion opportunities (i don't see a way to analyze the graph locally to find them), and I need shapetracker information to match (lazybuffer, st) read pairs, conveniently provided by preschedule. 

the rules as implemented are a little in the style of "performance heuristic" though, which is a little different from the other rules we have. is it possible to move back to pure scheduling land?

Qazalin:
I think all of your fusion targets are children of `E_2048`
Screenshot 2024-05-06 at 4 37 16 PM

https://tiny-tools-client.vercel.app/?id=3ef8c4a72b0c4999acca0dff9288b2fa

chaosagent:
Some of them are also children of the forward pass. How can we tell if there is a path forward -> stuff -> fusion targets so that we don't fuse ?

test/test_schedule.py
```diff
@@ -680,6 +681,140 @@ def test_prefer_half_buffer(self):
...
+    check_schedule([(x + a).sum(axis=0), (x + b).sum()], 2)
+
+    # maybe don't fuse if is 2-axis expand (gemm) and the early asts do not match?
+    # (because there might be too many accumulators)
+    # OK for now since we only fuse for bijective st
+
+    # match by input + ST and two shapes? start with contigouous input only, check shapes (should determine reduces)
+
+    # what if same input + st but one is early and another is late?
+    check_schedule([x.sum(0, keepdim=True) + a, (a + b).sum()], 2)
```
Qazalin, line 755: is this a real-world case?

test/test_schedule.py
```diff
@@ -680,6 +681,140 @@ def test_prefer_half_buffer(self):
...
+    check_schedule([x.sum(), x + y], 2)
+
+    # no fuse f(expand(reduce(x)), x) "diamonds"
+    # or at least it should be fused sequential instead of parallel
+    # fused group needs to be contiguous subDAG -- construct with toposort
+    sum1 = (x + a).sum()
+    check_schedule([sum1, (x + sum1).sum()], 2)
+    del sum1
+
+    # super tricky crossing dag case
```
Qazalin, line 767: do you think fusing this is faster?
Screenshot 2024-05-06 at 7 01 08 PM

Qazalin:
> forward -> BN forward -> stuff -> fusion targets

If you fuse those targets the doesn't the cache fill up with a bunch of the "stuff" bufs?

test/test_schedule.py
chaosagent: The (conservative) heuristic I am using is that this fusion should never add extra loads from bijective shapetrackers. If a shapetracker is bijective, then its size matches the full_shape of the kernel, and all non-bijective loads must be from smaller buffer(region)s. In the normal case, the non-bijective "small" buffers from expands and are very small compared to the bijective ones (here it's 1/16), so adding these won't hurt. Here, fusing the diagonal will save 1 memory pass over a big buffer.

chaosagent: In fact, for simple reduces like these from bijective shapetrackers, it should be fine to fuse many unrelated reduces. Simple reduces don't really need a lot of cache -- the cache really helps when you have expands like (1, a) * (b, 1), since you can do an nm-sized tile with only n + m loads.

chaosagent: i think this may even be a real world case -- consider x and y to be the forward outputs of different layers.

chaosagent:
> If you fuse those targets the doesn't the cache fill up with a bunch of the "stuff" bufs? We wanna fuse if they're sharing parents.

we need to allow small "stuff"s. See the argument for the bijective heuristic above

chaosagent:
hm, i think one of these kernels has a superset of "stuffs" across the rest of the fusion targets. i think that makes it safe to not check the "stuffs"﻿ :thinking:

geohot:
This is cool. Can we get some of these tests merged? (even if they are disabled for now)

PR #4371, opened 2024-05-01
Symbolic arange bounty

reddyn12:


tinygrad/tensor.py
```diff
@@ -423,7 +423,7 @@ def arange(start, stop=None, step=1, **kwargs):
     ```
     """
-    assert all(isinstance(s, (int, float)) for s in (start, stop, step)), f"symbolic arange not supported {start=}, {stop=}, {step=}"
+    start, stop, step = [arg.val if isinstance(arg, Variable) else arg for arg in [start, stop, step]]
```
geohot, line 426: This is *not* making a symbolic kernel, it's just removing the symbolic!

PR #4373, opened 2024-05-01
Fix for MTLLibraryErrorDomain Code=1 "Invalid library file"

mooihi:
[MTLLibraryErrorDomain Code=1 "Invalid library file"](https://github.com/tinygrad/tinygrad/issues/2226#issuecomment-1819771621) is caused by conda.

Conda python binaries are built with a target for older macOS versions. [11.0](https://github.com/conda-forge/python-feedstock/blob/main/.ci_support/osx_arm64_build_typereleasechannel_targetsconda-forge_main.yaml) for M series and [10.9](https://github.com/conda-forge/python-feedstock/blob/main/.ci_support/osx_64_build_typereleasechannel_targetsconda-forge_main.yaml) for Intel Macs.

Using `otool -L` on the conda python binaries show `/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)` while the working python binaries (system and homebrew) show `/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)`.

[Conda team does not seem to want to change the target anytime soon. ](https://github.com/conda-forge/conda-forge.github.io/issues/1844)

I managed to build a python binary via conda forge with only a target change for 13.3. And the metal compilation works fine.
Screenshot 2024-05-02 at 12 11 43 AM

As this is a conda issue, the PR sets the XCODE_METAL flag to 1 when conda environment is detected and uses xcrun for metal compilation. Also include a message for users about the conda incompatibility. I’ve tested the performance and it is actually faster to use xcrun although the cost is to install the Xcode library.

Those who had already ran tinygrad on conda before will need to clear cache by deleting ~/Library/Caches/tinygrad/cache.db for this to work. So the DISABLE_COMPILER_CACHE flag is no longer required.


mooihi:
Can a maintainer take a quick look at this? It can save a new conda user some pain. I dont need the bounty. Thanks.

geohot:
This is not a fix, and relies on the user having XCODE installed (also becomes slower). I think the current error message in there is fine.

PR #4380, opened 2024-05-01
hcq graph

nimlgen:
Started to experiment with HCQ graph for NV and AMD to understand what API we want to unify them

### DIFF
--- a/tinygrad/runtime/graph/hcq.py
+++ b/tinygrad/runtime/graph/hcq.py
@@ -59,13 +59,13 @@ class HCQGraph(Runner):
     for j,ji in enumerate(self.jit_cache):
       if isinstance(ji.prg, CompiledRunner):
         # self.comp_queues_list.append((j,ji))
-        # if j in self.jc_idxs_with_updatable_launch_dims:
-        #   if ji.prg.device in self.comp_queues: self.comp_queues_list.append((ji.prg.device, self.comp_queues.pop(ji.prg.device)))
-        #   self.comp_queues_list.append((j,ji))
-        # else: self.comp_queues[ji.prg.device].exec(ji.prg.clprg, self.kargs_addrs[j], *ji.prg.launch_dims(var_vals))
-        self.comp_queues[ji.prg.device].wait(signal=self.chain_signal, value=j)
-        self.comp_queues[ji.prg.device].exec(ji.prg.clprg, self.kargs_addrs[j], *ji.prg.launch_dims(var_vals))
-        self.comp_queues[ji.prg.device].signal(signal=self.chain_signal, value=j+1)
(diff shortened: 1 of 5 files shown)


tinygrad/runtime/graph/hcq.py
```diff
@@ -0,0 +1,106 @@
+import ctypes, collections, time, itertools
+from typing import List, Any, Dict, cast, Optional, Union, Tuple
+from tinygrad.helpers import GraphException, init_c_var, round_up, colored
+from tinygrad.buffer import Buffer, BufferOptions
+from tinygrad.device import Compiled, CompiledRunner, BufferXfer, MultiDeviceJITGraph, Device, Runner
+from tinygrad.shape.symbolic import Variable
+from tinygrad.engine.realize import ExecItem
+from tinygrad.engine.jit import get_input_replace, get_jit_stats, get_jc_idxs_with_updatable_launch_dims, get_jc_idxs_with_updatable_var_vals
+
+class HCQGraph(Runner):
```
geohot, line 10: Check out the new GraphRunner class

geohot:
How close is this to working? Is it far enough along where you could try ResNet? Ideally I want the MLPerf submissions to use this.

nimlgen:
Quite close already hlb_cifar seems to work fine (will push it). Going to try NV on resnet today. 

Still thinking about the final design synchronization to use for both NV and AMD: use binary semaphores (mostly like HSA) or "timeline" semaphores (where wait passes when `>=`). I think timeline semaphores are better in some ways (like no need to reset on CPUs, less signals to allocate), but it has some logic cost to reset it to 0 at the beginning of each graph run (on the GPU) if I am not missing something obvious and simple solution (will think more on this, how to make it simpler).

geohot:
Oh that's a neat idea, I didn't consider timeline semaphores. I think that's preferred since it's O(1) (I'm fine with O(1) reset). Don't you also have to reset binary semaphores?

nimlgen:
First resnet runs with this. I think it's possible to optimize it more:
AMD https://wandb.ai/nimlgen/tinygrad-examples_mlperf/runs/am7bavet
NV https://wandb.ai/nimlgen/tinygrad-examples_mlperf/runs/w0l52i8h

nimlgen:
Things to do in separate prs:
1) NV sync speed up (sync using qmd). Tested already this makes NV's gpu speed the same as CUDA's.
2) Optimize how we deal with updatable launch dims. Now we build a sep hcq for each of them: we can patch hcq for amd and qmd for nv, should be way faster python time.
3) Amd sdma fast submit() + some python optimization.

nimlgen:
@geohot, want to review? I keep this as minimal hcq only PR, will do speed related changes in the next prs.

PR #4382, opened 2024-05-02
move acc_dt out of lazy

chenyuxyz:
move the logic to tensor.py for forward, and function.py for two places in backward (expand and max)


chenyuxyz:
need to figure out what else needs to be updated. this is accessing 10% more mem in resnet

geohot:
I don't care what this breaks, let's merge it and fix anything later.

In the future, all LazyOp spec changes need a design doc

PR #4391, opened 2024-05-02
resnet eval 4n+3 if epoch < 33

chenyuxyz:
the rule is as thoroughly as 4n+k and we can stop the clock as soon as eval hits target. this can save 24 evals or 12 minutes


chenyuxyz:
example run https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/wz1e8wip?nw=nwuserchenyuxyz

saved 12 minutes to reach epoch 33

PR #4406, opened 2024-05-03
refactor sparse_categorical_crossentropy

chenyuxyz:
factor out the -1 * and / loss_mask.sum() for both smoothing and non-smoothing terms

saved like 1 minutes in resnet estimated for green


chenyuxyz:
ran 15 epochs look all fine, can save 1+ min total https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/h0st0kr9/overview?nw=nwuserchenyuxyz

PR #4407, opened 2024-05-03
Move Reduceops into `ast_parse` - pop ctx method

0xtimmy:
this is the method I settled on for the best way to properly order multiple reduceops

it renders them in ast_parse by "undoing" any uops that belong to a prior loop context, renders the full reduction, then puts those uops back

the "undoing" portion might be bandaid-ey; the main point is to render the entire reduceop and then put in the right place in the uop graph. undoing and then redoing seemed more intuitive because it also removed any context from the rendering of the second reduceop (ex. in `x.std()`, both reduce loops need to load in x. but when the recursion hits the mean reduce, it will already have x in `self.load_cache` from the variance reduce)

`render_reduceop` could also build a sort of subgraph and then ast_parse could handle where to insert it, but that would require a bigger diff


Qazalin (commented):
tests failing https://github.com/tinygrad/tinygrad/actions/runs/8942779865/job/24566055877?pr=4407#step:18:465

I still think pre-sorting self.reduceops with dfs is preferable.
A lot of what you're trying to handle in the linearizer is already taken care of when building the AST.

0xtimmy:
hmmm fair enough, I'll see if I can make something to do the dfs quickly

chaosagent:
i'm working on parallel reduces (two reduceops in one loop). my understanding is you are working on series reduces.

what do you think about self.reduceops = [[r1, r2], [r3], [r4, r5, r6]] where (r1, r2) and (r4, r5, r6) are parallel and [a, b, c] are in series? the self.reduceops list would be toposorted upon linearizer creation. then we can just write for reduceops in self.reduceops: self.render_reduceop(reduceops, ...).

0xtimmy:
yeah I like that better, here's the draft pr for all my linearizer changes https://github.com/tinygrad/tinygrad/pull/4409

PR #4188, opened 2024-04-16
abs is wrong for unsigned ints

Qazalin:
given: (a - b).abs()
- char and unchar don't compile
- uint returns the wrong result

fyi @chenyuxyz

PR #4411, opened 2024-05-03
fix abs of diff of uint

chenyuxyz:
brought up in #4188
abs to unsigned or bool is no-op
use where for int because cstyle does not like relu for int (might have more issue with int in general)
keep double relu for float, would need to check perf before switching

RHIP crashed in CI @Qazalin fyi


chenyuxyz:
fixed by the new abs #4611

PR #4412, opened 2024-05-03
use EVAL_BS=768 for resnet

chenyuxyz:


chenyuxyz:
@flammit when you got a chance can you confirm if this OOM on during setup for you?

while this fits for the whole run, it might OOM during setup. maybe it's fine to have `RESET_STEP=1` for setup

chenyuxyz:
not really worth it

PR #4414, opened 2024-05-04
fix: not using private metal api methods.

Leikoe:
Here is my pr to fix conda. It may seem weird but i'll explain.
I've been trying to debug the MTLInvalidLibrary issue for two months now. I've learned a lot about how metal does it's compilation. Metal has it's own cache for compiled shaders (and even the thing we were putting in lib wasn't really an executable and is getting compiled another time by metal with a different compilation requestType each time we call newFunctionWithName_). Using libraryDataContents was not a good decision as it as a private api and doesn't seem compatible with old libSystem.B versions.
As metal has it's own cache, it's okay to just cache the shader source with tinygrad. I didn't observe any slowdowns using METAL=1 on this commit.
I also removed the dissasembler part because:
- it wasn't compatible with the new way of handling the library
- the apple dissasembler isn't even a part of tinygrad's master anymore
- it was using lines for nothing

If anyone is interested I can make a little writeup on everything I tried and uncovered trying to get the compiled shader to be correct.

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -18,27 +18,19 @@ class MetalCompiler(Compiler):
     super().__init__("compile_metal")
   def render(self, name:str, uops) -> str: return MetalRenderer(name, uops)
   def compile(self, src:str) -> bytes:
-    if self.device is None:
-      # NOTE: if you run llvm-dis on "air" you can see the llvm bytecode
-      air = subprocess.check_output(['xcrun', '-sdk', 'macosx', 'metal', '-x', 'metal', '-c', '-', '-o', '-'], input=src.encode('utf-8'))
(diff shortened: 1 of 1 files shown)


Leikoe:
More about why I think it wasn't working:
It looked like a bug at first. Why would conda make metal's compiler spit out invalid shaders? But the thing is, the shader isn't actually invalid, only it's serialization is. When using older libSystem versions, only dynamic libraries are compiled with a libraryDataContents which is valid newer .metallib format. This makes sense as libraryDataContents isn't public and is probably meant for MTLDynamicLibraries which have a serialize to file method (serializeToURL iirc) which probably uses libraryDataContents internally.

Leikoe:
solves #2226

geohot:
This is not a fix for the issue, it breaks the compiler cache. Would rather people just didn't use conda.

PR #4416, opened 2024-05-04
amd timeline semaphores

nimlgen:


nimlgen:
fyi @geohot thinking of this as final sync idea for both nv and amd. any comments on this?

tinygrad/runtime/ops_amd.py
```diff
@@ -576,28 +578,11 @@ def __init__(self, device:str=""):
...
   def synchronize(self):
-    HWPM4Queue().signal(self.completion_signal).submit(self)
-    self._wait_signal(self.completion_signal)
-    assert (wp:=(self.pm4_write_pointer[0]%(self.pm4_ring.size//4))) == (rp:=self.pm4_read_pointer[0]), \
-      f"didn't run {wp} != {rp} len {self.pm4_ring.size//4}"
+    AMDDevice._wait_signal(self.timeline_signal, self.timeline_value - 1)
 
     # reset kernargs
     self.kernargs_ptr = self.kernargs.va_addr
+    if self.timeline_value > (1 << 31):
```
geohot, line 586: Do you ever test this rollover?

geohot:
Nope, I like the basic idea. One semaphore per hardware queue.

Though if I'm reading this right, the values will have to change on replay? Maybe there's another way to not have to do that.

tinygrad/runtime/ops_amd.py
```diff
@@ -206,15 +206,12 @@ def exec(self, prg:AMDProgram, kernargs, global_size:Tuple[int,int,int]=(1,1,1),
...
-    # have to self wait since flush doesn't work
-    self.signal(sig:=AMDDevice._get_signal())
-    self.wait(sig)
     return self
 
   def wait(self, signal:hsa.amd_signal_t, value=0):
     addr = ctypes.addressof(signal) + SIGNAL_VALUE_OFFSET
     self.q += [amd_gpu.PACKET3(amd_gpu.PACKET3_WAIT_REG_MEM, 5),
-      amd_gpu.WAIT_REG_MEM_MEM_SPACE(1) | amd_gpu.WAIT_REG_MEM_OPERATION(0) | amd_gpu.WAIT_REG_MEM_FUNCTION(3) | amd_gpu.WAIT_REG_MEM_ENGINE(0),
+      amd_gpu.WAIT_REG_MEM_MEM_SPACE(1) | amd_gpu.WAIT_REG_MEM_OPERATION(0) | amd_gpu.WAIT_REG_MEM_FUNCTION(5) | amd_gpu.WAIT_REG_MEM_ENGINE(0),
```
geohot, line 214: What's function 5? Can these be used from defines somewhere?

nimlgen:
Yeah, for graphs we can allocate 2 signals per graph for compute/copy and calculates timeline sync values for this graph, resetting just these 2 values at the beginning back to initial value to start the graph. Like we use this `dev.timeline_signal` as global sync in the case of graph and it's own internal timeline semaphore which restarts at the beginning

geohot:
Yours to merge when you think it's ready

PR #4417, opened 2024-05-04
refactor: pass reduceop into global_load

chaosagent:
prereq refactor for having more than one reduce in Linearizer (parallel, series).

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -74,7 +74,7 @@ class Linearizer(Kernel):
   def global_load(self, i:int, idxs:List[Node], acc:Optional[LazyOp]=None, barrier:Optional[UOp]=None, insert_before:Optional[UOp]=None) -> List[UOp]:
     buf = self.bufs[i]
     localtype = self.get_base_dtype(buf.dtype if acc is None else acc.dtype)
-    const = buf.val if isinstance(buf, ConstBuffer) else self.get_reduce_acc(acc)
+    const = buf.val if isinstance(buf, ConstBuffer) else None
 
     expand_vars = expand_idxs(idxs)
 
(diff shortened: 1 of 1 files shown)


tinygrad/codegen/linearizer.py
```diff
@@ -95,10 +96,11 @@ def global_load(self, i:int, idxs:List[Node], acc=None, barrier:Optional[UOp]=No
     invalid_value = 0
     for idx, valid, rep_idx in zip(e_idxs, e_valids, iter_idxs(expand_vars)):
       this_const, idx, valid = (invalid_value, NumNode(0), NumNode(1)) if valid.max == 0 else (const, idx, valid)
-      key = f"{acc}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
+      key = f"{acc is not None}{localtype}{'CONST'+str(this_const) if this_const is not None and acc is None else (buf.idx if isinstance(buf, MemBuffer) else cast(LocalBuffer, buf).name)}{idx.render()}{valid.render()}"  # noqa: E501
```
Qazalin, line 99: I need to think about this line, changing this key is tricky.
Can you write a test for it? I have one example in `test_load_cache_const_bufs`

tinygrad/codegen/linearizer.py
```diff
@@ -95,10 +96,11 @@ def global_load(self, i:int, idxs:List[Node], acc=None, barrier:Optional[UOp]=No
...
       if key not in self.load_cache:
         if acc is not None:
-          self.load_cache[key] = self.uops.add(UOps.DEFINE_ACC, localtype, (), dtypes.as_const(this_const, localtype), False, insert_before)
+          acc_const = dtypes.as_const(self.get_reduce_acc(acc), localtype)
```
Qazalin, line 102: I think get_reduce_acc is already the correct dtype

tinygrad/codegen/linearizer.py
Qazalin: if you have multiple accs how are you ensuring they uop different DEFINE_ACCs?

tinygrad/codegen/linearizer.py
Qazalin: yeah, CI was ok with removing it

tinygrad/codegen/linearizer.py
chaosagent: each reduceop is associated with a disjoint set of output buffers. self.reduceops.index(acc) is a good idea.

chaosagent: well, we don't have self.reduceops yet, and self.reduceops might end up being 2 dimensional for both parallel and sequence.

0xtimmy: I figured it was fine to use the same local buffer for sequential reduceops as long as they're the same shape, that obviously won't work for parallel reduceops lol

there was some weirdness with the key, I also think we should pass `reduceop` to `global_load` and I was playing with `...{self.reduceops.index(reduceop) if reduceop else ''}...`

but I also start each reduce with an empty `self.load_cache` so it didn't end up needing to be changed

Qazalin:
@chaosagent I'll merge if you either add a TODO or ideally a test for the acc cache key

chaosagent:
added todo.

PR #4409, opened 2024-05-03
Preorder reduces

0xtimmy:
@Qazalin turns out preordering them was p easy

drafting this PR because it includes the linearizer changes (so I could test that it works for asts with multiple reduceops). I can break it down into PRs for the ordering code & linearizer changes if that would be easier

PR #4418, opened 2024-05-04
Multiple Reduce Kernels - kernel properly orders reduceops

0xtimmy:
a precursor pr to the linearizer changes for fused reduceops

this makes the necessary changes to the kernel state for the linearizer function to render multiple reduceops


tinygrad/codegen/kernel.py
```diff
@@ -122,8 +120,8 @@ def copy(self):
     ret.opts, ret.ast, ret.lazyops = self.opts, self.ast, self.lazyops
 
     # things downstream of the AST
-    ret.reduceop, ret.outbufs, ret.vars, ret.bufs, ret.earlybufs, ret.full_buf_index = \
-      self.reduceop, self.outbufs, self.vars, [x for x in self.bufs if not isinstance(x, LocalBuffer)], self.earlybufs, self.full_buf_index
+    ret.reduceops, ret.outbufs, ret.vars, ret.bufs, ret.earlybufs, ret.full_buf_index = \
+      self.reduceops[:], self.outbufs, self.vars, [x for x in self.bufs if not isinstance(x, LocalBuffer)], self.earlybufs, self.full_buf_index
```
chaosagent, line 124: sorry for confusion this [:] isn't actually necessary unless reduceops is mutated XD

0xtimmy: oh my b, I'll change it back

Qazalin:
do you still need insert_before in linearizer functions after this?

0xtimmy:
no i don't think so

tinygrad/codegen/kernel.py
```diff
@@ -78,16 +78,14 @@ def __init__(self, *ast:LazyOp, opts:Optional[CompilerOptions]=None):
     self.lazyops = flatten([op.lazyops for op in self.ast])
 
     # there's only allowed to be one reduceop
-    reduceops = [x for x in self.lazyops if x.op in ReduceOps]
-    assert len(dedup(reduceops)) <= 1, "max one reduce op in an ast"
-    self.reduceop = reduceops[0] if reduceops else None
+    self.reduceops = dedup(flatten([reversed([x for x in out.lazyops if x.op in ReduceOps]) for out in self.ast]))
```
chaosagent, line 81: what ASTs do you plan on passing to the linearizer? how will you handle shapetracker pushing and reduceops?

Qazalin:
reverted insert_before - feel free to add it back in if you needed it.

tinygrad/codegen/kernel.py
0xtimmy: I pass all the asts to the linearizer, b/c we need all the outputs to get stored

tbh I didn't really need to change anything with the shapetracker; it only makes sense to fuse reducops from the same shape

the reduceops will get rendered sequentially this introduces the problem of:
1) loading back local reductions which gets solved [here](https://github.com/tinygrad/tinygrad/pull/4409/files#r1590002200)
2) scope, which with one reduce gets solved by `UOpGraph.fix_loop_scope` *except* that with multiple reduces `get_recursive_parents` will propagate all the way up to the local store of any prior reduceops, so we need to [cut off the recursion at barriers](https://github.com/tinygrad/tinygrad/pull/4409/files#r1590002961) (the parents of barriers aren't dependencies in the same sense as loops or alus)

I think we could handle different shape transitions, the reducops keep a notion of which axis they're reducing so there's enough information to manage everything in the linearizer alone

but same shape seemed to be the common case

Qazalin (commented):
can you keep the max one reduceop assert?

Qazalin:
This is wrong #4220 `test/test_linearizer.py TestLinearizer.test_multireduce`

Please test very carefully before marking a PR as ready for review.

0xtimmy:
I don't know if it should pass, this doesn't include the linearizer changes that make multiple reduceops work

PR #4419, opened 2024-05-04
refactor: linearizer support multiple sets of accumulators

chaosagent:


chaosagent:
@0xtimmy brought up a good point in #4409 about re-broadcasting accs across unroll dim, need to adjust

chaosagent:
I'm not sure if I'm missing some details. @0xtimmy thoughts? or feel free to just do it yourself.

I think loaded_buffers need to be scoped per sequential loop.

0xtimmy:
the `self.earlybufs` get updated (before the `ast_parse`) every time `render_reducop` is called so they are scoped in that sense
I had something similar to what you did with accs, treating it as a `Dict[LazyOp, List[UOp]]`

i think it looks good

chaosagent:
seems like accs and loaded_buffers kind of the same thing

Qazalin:
looks good, can you resolve the conflicts?

tinygrad/codegen/linearizer.py
```diff
@@ -414,21 +416,20 @@ def linearize(self):
...
     if x in cache: return cache[x]
     if x.op in BufferOps: return loaded_buffers[x.arg]
     if x.op is UnaryOps.CAST: return [self.uops.add(UOps.BITCAST if x.arg[1] else UOps.CAST, self.get_base_dtype(x.arg[0]), (u,), x.arg[0]) \
-                                      for u in self.ast_parse(x.src[0], acc, offs, loaded_buffers)]
-    if x.op in ReduceOps and not do_reduce:
-      assert offs is None, "not available if we aren't doing reduce"
-      return acc
+                                      for u in self.ast_parse(x.src[0], accs, offs, loaded_buffers)]
+    if x.op in ReduceOps and do_reduce is None:
+      return [accs[x][i] for i in offs] if offs is not None else accs[x]
```
Qazalin, line 426: can you bring back the assert? we only need this when there are multiple reduceops.
```suggestion
      assert offs is None, "not available if we aren't doing reduce"
      return accs[x]
```

chaosagent:
It's hard to test this because there is no interface behavior change. Thus the behavior change needs to be careful not to use this wrong. Maybe I am not refactoring these things in the correct order.

PR #4420, opened 2024-05-04
batchnorm + conv in test_schedule

Qazalin:
This test covers r3, back to back reduce, parallel reduce and partial fusion:

Screenshot 2024-05-04 at 5 16 38 PM

Screenshot 2024-05-04 at 5 16 38 PM


Qazalin:
https://tiny-tools-client.vercel.app/?id=3ef8c4a72b0c4999acca0dff9288b2fa this is a pretty good target graph for parallel reduce

PR #4423, opened 2024-05-04
Fix llm.c/export.py

Qazalin:
llm.c needed `"#include <tgmath.h>", "#include <stdbool.h>"`.
I think it can share CI path with imagenet to C.


geohot:
I'm working on a new way to do this with graph (#4424), but happy to update it.

Qazalin:
Will rebase master and merge the CI

geohot:
If you want export the fix is fine to merge, I don't think we should put that in CI yet though.

PR #4425, opened 2024-05-04
don't open devices from children

geohot:


chaosagent:
multiprocessing has an API to check if it's a child, maybe we want that?

PR #4428, opened 2024-05-04
resnet dataloader supports padding both ends and try pad last batch

chenyuxyz:
no-op for eval, testing if padding last batch converges better


chenyuxyz:
seems way worse https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/ciuswktw?nw=nwuserchenyuxyz

PR #4436, opened 2024-05-05
helpers.diskcache_clear

chenyuxyz:
drop all tables in diskcache. added a unit test but disabled it by default because it will drop all cache...


chenyuxyz:
existing cursors will raise `sqlite3.OperationalError` once tables are dropped, i think this is fine

PR #4437, opened 2024-05-05
update sops.gz, fix tests and add new linearizer test

flammit:
As a pre-req to requiring op.arg[1] as acc_dtype for ReduceOps.SUM, we need to update the sops.gz file so the ASTs contain the new argument.

The updated sops.gz has a known issue that pops up in the fuzz_linearizer test on METAL, so added that to test_linearizer_failures and also updated the test to pass as expected.

It also has an Opt(OptOps.UNROLL, axis=4, amt=0) in the world space, so increased the unroll range to match (could also do it as a single Op, but it's already a cheap failed check in kernel.py apply_opt).

Had to ignore `test/test_fusion_op.py` because it generates GB-sized AST lines which doesn't work for the sorting/testing frameworks.

Re-ran test_linearizer_failures on all the platforms and updated the behavior of test_failure_22 (not sure when that regressed).


flammit:
Updated to ignore `test/test_linearizer_failures.py` when generated the world as well.  New random doesn't include the error, so reverted the fuzz test and action space changes.

test/test_linearizer_failures.py
```diff
@@ -154,10 +154,11 @@ def test_failure_21(self):
     opts = [Opt(op=OptOps.PADTO, axis=0, amt=32)]
     helper_test_lin(Linearizer(ast), opts, failed_platforms=[])
 
+  @unittest.skipIf(CI and Device.DEFAULT=="METAL", "works fine on intel CI METAL")
```
chenyuxyz, line 157: i think ci is using real metal now

flammit: It fails locally for me, but succeeds on CI

chenyuxyz: I think it's fine to skip this in CI, it's numerical issue and CI might succeed because it emulates half. My point was that CI is not using intel

flammit: Ahhh sorry I'll fix the comment

chenyuxyz:
how many lines are there in the new kernel dataset, and the old one?

flammit:
> how many lines are there in the new kernel dataset, and the old one?

old one is 12330, new one is 11642.  I installed everything needed to run all the models in the `generate_dataset.sh` script.

PR #4440, opened 2024-05-05
feat: allow keeping multiple cookies

wozeparrot:


wozeparrot:
faster on red, can be tuned probably along with how much readahead buffering to do. doesn't look like it does much on green.

PR #4447, opened 2024-05-06
Symbolic arrange attempt

Striletchi-Vlad:
Computed kernel seems symbolic, code is not optimised. Curious if this is the right direction.


geohot:
I'm going to delete this bounty, it's clear it isn't understood

PR #4451, opened 2024-05-06
Add LCG as random generator for BERT dropout. Monkeypatch dropout

Eliulm:
The advantage of the LCG is that it fuses into one kernel, lowering mem requirement for dropout.

The LCG passes all tests in `test_randomness.py` that threefry passes.


examples/mlperf/initializers.py
```diff
@@ -12,6 +12,25 @@ def rand_truncn(*shape, dtype=None, truncstds=2, **kwargs) -> Tensor:
   take = (x.abs() <= truncstds).where(ctr, CNT).min(axis=-1, keepdim=True)  # set to 0 if no good samples
   return (ctr == take).where(x, 0).sum(axis=-1)
 
+# Use combined Linear congruential generator
+def rand_lcg(*shape, device=None, dtype=None, **kwargs):
```
geohot, line 16: Is this a real RNG from somewhere? Does it pass all reasonable randomness tests?

Should we switch from threefry?

Eliulm: LCG random generators used to be the most common ones, but no popular frameworks use them anymore because their quality of randomness is unsuitable for very large-scale simulations. However, they are absolutely fine for dropout, especially since this is a combined LCG, which is a lot better.

>  Does it pass all reasonable randomness tests?

Yes

> Should we switch from threefry?

I would keep threefry, since that is also what other frameworks like JAX use. I think that at some point tinygrad will be able to fuse all threefry kernels into one.

Eliulm:
@chenyuxyz One issue this method has is that it sometimes takes very long to search with beam because some opts are especially terrible. Some opts will take minutes to execute.

PR #4454, opened 2024-05-06
kernel: un-reverse the order of the local indices

flammit:
no change to performance or behavior.  new LOCALS are added to the left side of the LOCALS block (to the left of the first_reduce).


chenyuxyz:
looks good. the NV gemm stats is weird but it's already weird on master, so not unrelated to this

PR #4461, opened 2024-05-06
redo faster sparse_categorical_crossentropy

chenyuxyz:
update LR and DECAY in resnet default that help convergence too


chenyuxyz:
looks good on both red (tiny20) and green (tiny14)
image

PR #4462, opened 2024-05-06
[DRAFT] kernel: prevent TC from choosing and padding small axes

flammit:
as shown on METAL resnet and the original PR testing on K=16 tensor cores, anything "small" might help on the first search iteration, but causes a local minimum which is detrimental.

For performance testing on M2 Max:  running `BEAM=3 python examples/handcode_resnet50_opt.py` approximately 5TF before this and 7TF after this fix.

Will try this on M3 Max shortly.


chenyuxyz:
beam=8 is fine, let's not do this and will fix with a better search

image

flammit:
SGTM!

PR #4466, opened 2024-05-07
filter independent items in jit cache

geohot:
This still isn't ready. Need to consider how assign works in the JIT more.


geohot:
So out of date

PR #4471, opened 2024-05-08
revert the removal of CAST_BEFORE_VIEW

chenyuxyz:
this brings most of the memory gain for resnet back.

effectively revert #4152


chenyuxyz:
@chaosagent fyi

PR #4472, opened 2024-05-08
lazy folding: mul -1 is neg, and neg neg is noop

chenyuxyz:


tinygrad/lazy.py
```diff
@@ -149,14 +149,17 @@ def e(self, op:Union[LoadOps, UnaryOps, BinaryOps, TernaryOps], *in_srcs:LazyBuf
     # const folding
     if op in python_alu and all(s.is_unrealized_unmasked_const() for s in srcs):
       return self.cast(out_dtype).const(exec_alu(op, out_dtype, [s.base.arg for s in srcs]))
+    if op is UnaryOps.NEG and self.base.op is UnaryOps.NEG: return self.base.srcs[0]
```
davidjanoskyrepo, line 152: Is folding/ op fusion done before or after the schedule optimizer is run?

PR #4483, opened 2024-05-09
fix: metal on conda on a13 and higher

Leikoe:
metal not compiling successfully on conda is due to a wrong compile requestType. This can be fixed the following ways:
- compile with a fairly recent libSystem.B.dylib
- compiling a dynamic lib
- patch Metal
- compiling a executable library (as we do right now) + linking to any dynamic libary

This fix is using the last way. It will work on any mac which supports dynamic library and will only trigger then. For older devices it won't change anything.


geohot:
Bad spacing. Don't use globals.

tinygrad/runtime/ops_metal.py
```diff
@@ -11,10 +11,20 @@ def wait_check(cbuf: Any):
   if (error := cbuf.error()) is not None:
     raise RuntimeError(error)
 
+def get_workaround_lib(device):
+    options = Metal.MTLCompileOptions.new()
```
chenyuxyz, line 15: bad indent-width

geohot:
Can we submit this request directly?

Leikoe:
No, I looked for other ways but apple expects you not to use libraryDataContents and therefore store incomplete library data on some versions. The only way to do it properly is either to recompile and pray for system cache or use things with very limited support (BinaryArchive).

Leikoe:
If you were to call private C++ apis from Metal internals you could send the proper requestType but that wouldn't be supported everywhere, this hack is not as bad.

geohot:
Not going to merge a hack like this. From what I gather it really is a conda bug and should be filed as an issue there.

PR #4485, opened 2024-05-09
Split UnaryOps.CAST into UnaryOps.CAST and UnaryOps.BITCAST

Mats-Janssens:
Here is are a couple of small changes to make the split UnaryOps.CAST into two. In this PR I have not yet created a separate Function for bitcast. I did do this and personally do find it cleaner but just to start small.


Mats-Janssens:
I don't understand what needs to happen in the flopcounter so closing this for now

PR #4513, opened 2024-05-10
restore hcq graph

geohot:


tinygrad/runtime/ops_nv.py
```diff
@@ -235,6 +235,9 @@ def __del__(self):
 
   def __call__(self, *args, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tuple[int,int,int]=(1,1,1), vals:Tuple[int, ...]=(), wait=False):
     if prod(local_size) > 1024 or self.max_threads < prod(local_size): raise RuntimeError("Too many resources requsted for launch")
+    if not hasattr(self, "args_struct_t"):
```
geohot, line 238: Huh? What if this isn't called. @nimlgen

nimlgen: yeah, better move this into the graph itself

PR #775, opened 2023-05-15
Add MLPerf UNet3D model

jla524:
Creating this draft PR for visibility.

Changes made:
- Add data loader for the KiTS19 dataset
- Implement functions to compute DICE score
- Implement functions for UNet3D inference

Download dataset:

```
git clone https://github.com/neheller/kits19
cd kits19
pip3 install -r requirements.txt
python3 -m starter_code.get_imaging
```

PR #4518, opened 2024-05-10
Update UNet3D inference to use MLPerf's evaluation set

francislata:
# Inference set changes
I noticed that MLPerf's [inference set](https://github.com/mlcommons/inference/blob/87ba8cb8a6a4f6525f26255fa513d902b17ab060/vision/medical_imaging/3d-unet-kits19/meta/inference_cases.json#L44) includes one extra file, `case_00400`. I did see that the official [kits19 dataset](https://github.com/neheller/kits19/tree/master/data) doesn't have this particular case. To compensate for it, the official MLPerf's inference script  [copies case_00185 to be case_00400](https://github.com/mlcommons/inference/blob/87ba8cb8a6a4f6525f26255fa513d902b17ab060/vision/medical_imaging/3d-unet-kits19/Makefile#L96-L102).

For comparison, here's the result of running `model_eval.py` using `get_val_set` (old):
val_set

Meanwhile, here's the result of running `model_eval.py` when using `get_eval_set` (new):
eval_set

Both of these runs use the pretrained model and is [above the expected MEAN Dice score](https://github.com/mlcommons/inference/blob/87ba8cb8a6a4f6525f26255fa513d902b17ab060/vision/medical_imaging/3d-unet-kits19/README.md#supported-frameworks) of **0.86170**.

# Support for loading model checkpoint inside `model_eval.py`
Lastly, this also adds support of loading a model checkpoint for the UNet3D model when running `model_eval.py`. Here's an example run of a [converged training run on a tinybox green](https://discord.com/channels/1068976834382925865/1227683333798891561/1237806556284715078):
Screenshot 2024-05-10 at 16 09 01


wozeparrot:
The dice score should be the same as the reference no?

francislata:
> The dice score should be the same as the reference no?

this one is an odd one because i was looking at the old PR that introduced this eval function and they got [0.86632](https://github.com/tinygrad/tinygrad/pull/775#issuecomment-1565739063).

i'm pretty sure we're using the same pretrained model from the MLPerf inference README but I can double-check that again.

wozeparrot:
Yea looking at the original pr, they say that the reference target is 0.86330, but I'm not sure where this is from considering that mlcommons upstream https://github.com/mlcommons/inference/blob/master/vision/medical_imaging/3d-unet-kits19/README.md lists it as 0.86170 for all frameworks.

francislata:
yeah I agree. the only one time I found `0.86630` was in [Dell's blog regarding MLPerf](https://infohub.delltechnologies.com/en-us/p/mlperf-tm-inference-4-0-on-dell-poweredge-server-with-intel-r-5th-generation-xeon-r-cpu/):
Screenshot 2024-05-10 at 17 20 36

and that blog was pretty recent (last month).

francislata:
I'll close this for now and come back to it once the training code has been merged to see where the difference is between tinygrad and the reference implementation.

PR #4530, opened 2024-05-11
move copy kernel to out of schedule ordering

Qazalin:
@geohot I think the scheduler should create the correct AST in the DAG transformation part.


Qazalin:
ran this with `USE_COPY_KERNEL=1`
```py
import numpy as np
from tinygrad import Tensor
from tinygrad.engine.realize import run_schedule
from tinygrad.engine.schedule import create_schedule

a = Tensor.randn(4, 4, device="HSA:1").realize()
c = Tensor.ones((4, 4), device="HSA:1").contiguous().realize()
b = a.lazydata._copy("HSA:2")
#b = c.lazydata._copy("HSA:2")
sched = create_schedule([b])
run_schedule(sched.copy())
np.testing.assert_equal(np.frombuffer(a.lazydata.buffer.as_buffer(), a.dtype.np), np.frombuffer(b.buffer.as_buffer(), b.dtype.np))
```
to test.

PR #4535, opened 2024-05-11
amd mockgpu

nimlgen:
AMD works, test_ops passes. Need to clean up a lot.


geohot:
Very smart hooking the syscalls, that's even better than what I was thinking.

PR #4539, opened 2024-05-12
update training script with new models

reddyn12:


reddyn12:
forgot new commits updates pr

PR #4543, opened 2024-05-12
taylor series SIN, LOG2, EXP2

Striletchi-Vlad:


chenyuxyz:
"Taylor approximations for LOG2/EXP2/SIN in function.py passing all tests"

PR #4550, opened 2024-05-12
Taylor series

Striletchi-Vlad:


Striletchi-Vlad:
Could someone confirm if this is the right approach and maybe lock? SIN works, I am still experimenting with log2 and exp2

geohot:
Will lock bounty when tests pass without use of UnaryOps.SIN UnaryOps.EXP2 UnaryOps.LOG2

geohot:
Tests failing.

PR #4555, opened 2024-05-12
[WIP] Taylor approximation functions for LOG2/EXP2/SIN

dimaheve:
Sin seems to be working reasonably well. The coefficients were computed using Chebyshev polynomials (from numpy.polynomial). It is still a work in progress. Quite sure there are better ways to handle the cases in sin.


geohot:
Bounty locked, need to make the tests pass and not use UnaryOps.LOG2/EXP2/SIN anywhere.

geohot:
Actually, not going to lock bounties for this one until tests pass.

PR #4556, opened 2024-05-12
fix error message

ziereis:


geohot:
This is better, but doesn't explain how to fix it.

geohot:
Better. Can you also add a test to confirm this assert happens?

PR #4558, opened 2024-05-12
Test for optim assertion

ziereis:


test/test_optim.py
```diff
@@ -119,5 +119,15 @@ def test_mixed_precision(self):
...
     dtypes.default_float = old_default_float
 
+  def test_assert_tensor_train(self):
+    t = Tensor.ones((1,1), requires_grad=True)
+    optimizer = Adam([t])
+    optimizer.zero_grad()
+    t.sum().backward()
+    Tensor.training = False
+    self.assertRaises(AssertionError, optimizer.step)
+    Tensor.training = True
```
geohot, line 129: Err, this function changes the global state.

PR #4559, opened 2024-05-12
Add simple meshgrid impl

reddyn12:
Used it for retinanet, but need to think of implementation for sequence>2


geohot:
Please read your test and understand why I can't merge this

reddyn12:
@geohot I cleaned up the tests, but now we are testing addition and meshgrid functionality in the same test, is that fine?

geohot:
Weird test, not going to merge

PR #4571, opened 2024-05-13
feat: add acc_dtype to einsum

PhillCli:
Adds acc_dtype to einsum


chenyuxyz:
can you add tests next to `test_sum_acc_dtype`?

PhillCli:
> can you add tests next to `test_sum_acc_dtype`?

@chenyuxyz 
I put it into the existing `test_arg_acc_dtype`, as an explicit `Tensor.einsum` call, as it seems fit naturally with `test_arg_acc_dtype` structure, with other methods like `.sum()`, and `.matmul`.

Is that what you expected?

chenyuxyz:
cool!

PR #4575, opened 2024-05-13
raise error if setitem tensors have requires_grad

chenyuxyz:
working on supporting this, first properly raises error


chenyuxyz:
annoyed that optimizer can flip, but fine for now i guess....

PR #4578, opened 2024-05-14
approximations of Sin/Log2/Exp2

thatdeep:
approximations are tight up to:
* 1e-14 abstol for sin on [-pi/2, pi/2]
* 2e-12 abstol for log2 on [0.5, 1.0]
* 1e-16 abstol for exp2 on [-1/32, 1/32], scales with same reltol outside until clipping occurs

Approximations were created using sollya and remez algorithm closely to what was described in
[https://github.com/ARM-software/optimized-routines/tree/8928e471b062b008717c9c2958240c446317de53/math](https://github.com/ARM-software/optimized-routines/tree/8928e471b062b008717c9c2958240c446317de53/math).

Without immediately obvious analogs of frexp/ldexp for fast and overflow-resistant base-2 integer exponent extraction/injection resorted to manually scale powers of two up to 64 in a loop for Exp2, and manually multiply by powers of two up to 64 in a loop for Exp2. Guess its a matter if discussion what we allowed to do here to make it better.


geohot:
Tests failing.

PR #4593, opened 2024-05-15
docs cleanup and move

geohot:


ibebrett:
This seems to have busted the link to adding a new accelerator:

image

As a fun learning project I wanted to see if I could support https://github.com/adam-maj/tiny-gpu so I am curious if there is still really a cleanly minimal set of ops that need to be supported?

PR #4608, opened 2024-05-15
Revert "simpler abs and sign"

chenyuxyz:
Reverts tinygrad/tinygrad#4606


chenyuxyz:
reverting because it breaks bert, but test is still red

PR #4611, opened 2024-05-15
redo simpler abs and sign

chenyuxyz:
moved Sign logic to function.py, and backward always returns 0 to match torch. rewrite abs as `self * self.sign()`, so its backward also matches torch.


chenyuxyz:
this should also fix #4411

Issue #4612, opened 2024-05-16
llama sharded fails to beam on HSA


wozeparrot:
Reproduces with #4576 and `DEBUG=2 BEAM=4 python3 examples/llama3.py --model /raid/weights/LLaMA-3/8B/ --shard 6`

```
Traceback (most recent call last):
  File "/home/tiny/tinygrad/examples/llama3.py", line 198, in <module>
    model(Tensor([prompt], device=device), 0, TEMPERATURE, top_k=TOP_K, top_p=TOP_P, alpha_f=ALPHA_F, alpha_p=ALPHA_P).realize()
  File "/home/tiny/tinygrad/extra/models/llama.py", line 175, in __call__
    return self.forward(tokens, start_pos, temperature, top_k, top_p, alpha_f, alpha_p)
  File "/home/tiny/tinygrad/extra/models/llama.py", line 168, in forward
    return ret.realize()
  File "/home/tiny/tinygrad/tinygrad/tensor.py", line 166, in realize
    run_schedule(*self.schedule_with_vars(*lst))
  File "/home/tiny/tinygrad/tinygrad/engine/realize.py", line 186, in run_schedule
    for ei in lower_schedule(schedule):
  File "/home/tiny/tinygrad/tinygrad/engine/realize.py", line 179, in lower_schedule
    while len(schedule): yield lower_schedule_item(schedule.pop(0))
  File "/home/tiny/tinygrad/tinygrad/engine/realize.py", line 165, in lower_schedule_item
    runner = get_runner(si.outputs[0].device, si.ast)
  File "/home/tiny/tinygrad/tinygrad/engine/realize.py", line 137, in get_runner
    prg: Program = get_linearizer(Device[dname].renderer, ast).to_program()
  File "/home/tiny/tinygrad/tinygrad/engine/realize.py", line 35, in get_linearizer
    timed = sorted([(nm, tk, time_linearizer(tk, rawbufs, allow_test_size=False, clear_l2=True)) for nm, tk in lins], key=lambda x: x[2])
  File "/home/tiny/tinygrad/tinygrad/engine/realize.py", line 35, in <listcomp>
    timed = sorted([(nm, tk, time_linearizer(tk, rawbufs, allow_test_size=False, clear_l2=True)) for nm, tk in lins], key=lambda x: x[2])
  File "/home/tiny/tinygrad/tinygrad/engine/search.py", line 180, in time_linearizer
[...]

PR #4621, opened 2024-05-16
time with real global buffers in search

chenyuxyz:


chenyuxyz:
fixed #4612, tested with deleting cache on tinybox red and `DEBUG=2 BEAM=4 python3 examples/llama.py --gen 3 --size 8B --shard 6 --prompt "Hello."`

PR #4623, opened 2024-05-17
add back tensor_cores for "AMD" and "NV"

chenyuxyz:
accidentally removed in #4527


chenyuxyz:
hmm it's not needed?

chenyuxyz:
oh because it shared renderer

PR #4625, opened 2024-05-17
Place rewritten uops in same position

SzymonOzog:
Calculating new position sometimes picked a shitty position performancewise - like placing stores between ops that require a sync. This fixes that by putting them where the original op was


geohot:
I'm rewriting UOps in #4560, does that still have this problem?

SzymonOzog:
> I'm rewriting UOps in #4560, does that still have this problem?

Doesn't look like it as it rewrites before linearizing

SzymonOzog:
Closing as #4560 fixed it

PR #4627, opened 2024-05-17
use JITBEAM instead of BEAM in benchmark CI for llama and gpt2

chenyuxyz:
search less kernel, should be faster


chenyuxyz:
i did not expect the final searched kernels to change, but it's 10% slower on all 3 backends

PR #4629, opened 2024-05-17
nv hcq bind api

nimlgen:
LLaMA 3 on NV enqueue time:

```
NV=1 python3 examples/llama.py --gen 3 --size 8B --shard 6 --prompt "Hello." --count 10 --temperature 0 --timing

enqueue in   9.02 ms
total  39.46 ms, 25.34 tok/s, 420.30 GB/s, param 407.35 GB/s
 I
```


nimlgen:
Going to merge NV first. AMD also supports indirect buffers execution on both compute and sdma, so that's the idea. Need to debug AMD's SDMA tho, so just starting with NV

PR #4633, opened 2024-05-17
change sqrt and reciprocal order in rsqrt

chenyuxyz:
this matches rsqrt backward at 0 to torch. also added exact tests for reciprocal, sqrt, and rsqrt

see https://pytorch.org/docs/stable/notes/autograd.html#gradients-for-non-differentiable-functions for how torch defines derivative at these points


geohot:
You still want this?

PR #4638, opened 2024-05-18
root cause fix for UOps.CONST bad args

Qazalin:
resolves the uops.py TODO.


Qazalin:
NOTE: lazy.py's arg is always correct - this is for linearizer internal consts

PR #4640, opened 2024-05-18
Fix flop coutning for mulacc

SzymonOzog:


chenyuxyz:
can you add a regression test?

SzymonOzog:
> can you add a regression test?

This is a prerequisite to removing uops deepcopy from assembly it will be tested by https://github.com/tinygrad/tinygrad/blob/b0cb02f719fe4d5a2bd71f445023d4b36cf7d923/test/test_uops_stats.py#L42C32-L43C1 afterwards - if you want I can merge deepcopy removal into this PR, split so that they are more focused

chenyuxyz:
Splitting pr is good, but behavior change should have its test. It also is not immediately clear to me that the test you pointed to trigger this code path, and not WMMA. Can you add a simple test that triggers this?

SzymonOzog:
> Splitting pr is good, but behavior change should have its test. It also is not immediately clear to me that the test you pointed to trigger this code path, and not WMMA. Can you add a simple test that triggers this?

Oh ptx matmul uses mulacc when in full precision, but you are right - an explicit test is better. Added to PR

chenyuxyz:
are you sure the test fails on master?

SzymonOzog:
> are you sure the test fails on master?

Oh, my bad it seems like it takes info from InterpretedFlopCounter and not flops_mem. MULACC is only used on ptx so it won't fail on master untill deepcopy is removed. Reverted the bad test

chenyuxyz:
you can add a test that calls `flops_mem` directly and verify the output. one day we will remove InterpretedFlopCounter

SzymonOzog:
> you can add a test that calls `flops_mem` directly and verify the output. one day we will remove InterpretedFlopCounter

Updated

chenyuxyz:
thanks!

PR #4642, opened 2024-05-18
Clip by global norm + loss_scaler

Eliulm:
Implements https://www.tensorflow.org/api_docs/python/tf/clip_by_global_norm

### DIFF
--- a/examples/mlperf/model_train.py
+++ b/examples/mlperf/model_train.py
@@ -364,7 +364,7 @@ def train_step_bert(model, optimizer, scheduler, input_ids:Tensor, segment_ids:T
     optimizer.zero_grad()
     (loss * loss_scaler).backward()
 
-    global_norm = Tensor.empty((len(optimizer.params),), dtype=optimizer.params[0].dtype, device=optimizer[0].device).realize()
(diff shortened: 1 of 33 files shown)


examples/mlperf/model_train.py
```diff
@@ -354,15 +354,22 @@ def train_rnnt():
...
     optimizer.zero_grad()
-    loss.backward()
+    (loss * loss_scaler).backward()
+
+    global_norm = Tensor.empty((len(optimizer.params),), device=optimizer[0].device).realize()
+    for i, p in enumerate(optimizer.params): 
+      p.grad = p.grad / loss_scaler
+      global_norm[i] = p.grad.square().sum()
+    global_norm = global_norm.sum().sqrt()
+    for i, p in enumerate(optimizer.params): p.grad = p.grad / Tensor.where(global_norm > 1.0, global_norm, 1.0)
```
wozeparrot, line 372: Isn't `global_norm` just a scalar here? I don't really get the use of `.where` here.

Eliulm: @wozeparrot Yeah, it's a scalar. But what other way is there to evaluate the expression, but not break the graph?

Eliulm:
@chenyuxyz https://wandb.ai/wahl/MLPerf-BERT/runs/6asyog8w?nw=nwuserwahl failed after 34K steps.

PR #4647, opened 2024-05-19
remove per layer realize in llama

chenyuxyz:
previously needed due to kernel buffer count limit, seems fine to remove now


chenyuxyz:
it found a variable scope bug!

geohot:
I think I'm hitting the same one in mac benchmark, might be unrealized and just nondeterminism.

geohot:
Yea, it's non deterministic. https://github.com/tinygrad/tinygrad/actions/runs/9148100711/attempts/1

@Qazalin we need that toposort fuzzer. I will also investigate where non-determinism is coming from.

chenyuxyz:
this was changed to `contiguous`

PR #4651, opened 2024-05-19
reorder DEFINE_GLOBAL in fuzz_uops

Qazalin:
```
FUZZ_UOPS=1 DEBUG=7 python3 test/test_linearizer.py TestLinearizer.test_load_removed
```
works now.


geohot:
The globals should be reordered (and potentially removed) when ScheduleItems become ExecItems. All orders should work, I thought we had this but I guess not.

Qazalin:
here's a smaller example: `FUZZ_UOPS=1 DEBUG=7`
```
from tinygrad import Tensor
out = Tensor([1]) - Tensor([2])
out.realize()
```

geohot:
So are you sure this isn't a bug in the way the searcher runs things?

`   return ExecItem(runner, [si.bufs[x[0]] for x in runner.p.globals])`

Is used in realize.py, and that should deal with this, right?

PR #4653, opened 2024-05-19
correctly insert UOps.END* in fuzz result

Qazalin:
Since the fuzzer uses a different toposort than uops.py we add UOps.END* in fuzz_uops.py


Qazalin:
loops still need some work to generate the correct fuzz results.

PR #4654, opened 2024-05-19
[WIP] Taylor approximation

dimaheve:
SIN Taylor approximation implementation that can handle reducing big angles without losing significant precision.


geohot:
Tests failing.

PR #4658, opened 2024-05-20
run more in amd ci

nimlgen:
to test how slow and what we can add to ci

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -464,7 +464,7 @@ jobs:
         run: python -m pytest -n=auto test/ -k 'not (half or test_efficientnet_safetensors)' --ignore=test/external --ignore=test/models --durations=20
       - name: Run pytest (amd)
         if: matrix.backend=='amd'
-        run: python -m pytest -n=auto test/ test/external/external_test_hcq.py -k 'not (half or test_efficientnet_safetensors)' --ignore=test/external --ignore=test/models --durations=20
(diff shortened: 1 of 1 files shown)


Qazalin:
remu needs new instructions - I think I pretty much implemented all formats so should be easy to add them.

nimlgen:
8m is also pretty slow, gpuocelot is ~3m on the same tests. I will profile remu, but maybe you @Qazalin know what might be slow in remu?

Qazalin:
I've seen line 48: https://github.com/Qazalin/remu/blob/master/src/lib.rs be very slow for large global sizes. Tried multithreading but the overhead cost was greater.

nimlgen:
hmm, you think mt really matters running ci? it runs a worker on each thread, so I think we should try just to optimize single threaded?

Qazalin:
so I'd first root cause where the bottleneck is. The global dims is a guess.

nimlgen:
I have seen `index` to be slow, I think cuz of hashmap. Can we just switch this to array of registers of fixed size?
```
pub struct VGPR {
    values: HashMap<usize, [u32; 256]>,
    pub default_lane: Option<usize>,
}
impl Index<usize> for VGPR {
    type Output = u32;
    fn index(&self, index: usize) -> &Self::Output {
        &self.values.get(&self.default_lane.unwrap()).unwrap()[index]
    }
}
```

Qazalin:
I released this FYI - didn't see a huge perf change though

geohot:
We still want this?

geohot:
Closing for now

PR #4664, opened 2024-05-20
update BEAM_LOCAL_MAX to 1024

chenyuxyz:
we used 1024 for mlperf submission and result steps time is 20% faster. the default should not be worse


chenyuxyz:
@flammit fyi you mentioned you want to remove this limit?

PR #4665, opened 2024-05-20
Multireduce Linearizer Tests

0xtimmy:
This PR includes tests for the multireduce linearizer changes, the `@unittest.skip` will get removed with the actual linearizer changes

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -171,18 +171,13 @@ class TestLinearizer(unittest.TestCase):
     end = max(i for i,u in enumerate(lin.uops) if u.uop is UOps.ENDRANGE)
     assert lin.uops[end+1].uop is UOps.ALU
 
-  @unittest.expectedFailure
+  @unittest.skip
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
   def test_early_end_local(self):
-    shape, output_shape = (32,), (1,)
-    load0 = MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker.from_shape(shape))
-    load1 = MemBuffer(idx=2, dtype=dtypes.float, st=ShapeTracker.from_shape(shape))
-    store = MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker.from_shape(output_shape))
-    ast = LazyOp(op=BufferOps.STORE, src=(LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BinaryOps.ADD, src=(
-      LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BufferOps.LOAD, arg=load0),)),LazyOp(op=BufferOps.LOAD, arg=load1), )),)),), arg=store),
-
-    load0_t = Tensor.randn(shape).realize()
-    load1_t = Tensor.randn(shape).realize()
-    k = helper_linearizer_ast(ast, [load0_t, load1_t], wanna_output=[(load0_t.numpy().sum() + load1_t.numpy()).sum()])[1]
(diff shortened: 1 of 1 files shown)


test/test_linearizer.py
```diff
@@ -191,14 +186,98 @@ def test_early_end_local(self):
     self.assertLess(k.uops.uops.index(barriers[0]), k.uops.uops.index(ifs[0]))
     self.assertLess(k.uops.uops.index(ifs[0]), k.uops.uops.index(endifs[0]))
     self.assertLess(k.uops.uops.index(barriers[1]), k.uops.uops.index(ifs[1]))
-
-  @unittest.expectedFailure
+    x = Tensor.randn(3,27,32).realize()
+    helper_linearizer_ast(ast, [x], wanna_output=[x.numpy().std(axis=2, ddof=0).reshape(-1)])
+
+  @unittest.skip
+  def test_reduceops_order(self):
```
0xtimmy, line 193: this tests if reduceops different ways the linearizer can be given a set of reduceops and makes sure they all end up in the correct order

test/test_linearizer.py
```diff
@@ -191,14 +186,98 @@ def test_early_end_local(self):
...
+      k = Linearizer(*[asts[i] for i in order])
+      def recursive_reduceops(x: LazyOp): return [c for v in x.src for c in recursive_reduceops(v)] + [v for v in list(x.src) if v.op in ReduceOps]
+      for i,r in enumerate(k.reduceops): assert not any([r in recursive_reduceops(x) for x in k.reduceops[:i]]), "reduceops are out of order"
+      x = Tensor.randn(32).realize()
+      outs = [b:=(a:=x.numpy()).sum(), c:=(a - b).sum(), d:=(c - a).sum(), (a-d + a-b).sum()]
+      helper_linearizer_ast([asts[i] for i in order], [x], wanna_output=[outs[i] for i in order])
+
+  @unittest.skip
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
+  def test_multireduce_store_locals(self):
```
0xtimmy, line 215: this makes sure the result of local reductions are stored and loaded back into each thread
ex it's looking for
```
  // first portion of the reduction, populates acc0 
  temp[lidx1] = acc0;
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu0) {
    float acc1 = 0.0f;
      float val1 = temp[ridx1];
      acc1 = (val1+acc1);
    }
    temp[0] = acc1; // making sure this store gets generated 
  }
  barrier(CLK_LOCAL_MEM_FENCE);
  float val2 = temp[0]; // and this load gets generated
  float acc2 = 0.0f;
  // ... later reductions
```

test/test_linearizer.py
```diff
@@ -171,18 +171,13 @@ def test_range_outer_op_after_phi_nested_range(self):
     end = max(i for i,u in enumerate(lin.uops) if u.uop is UOps.ENDRANGE)
     assert lin.uops[end+1].uop is UOps.ALU
 
-  @unittest.expectedFailure
+  @unittest.skip
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
   def test_early_end_local(self):
```
0xtimmy, line 176: this tests that `ENDIF` uops are placed properly when there are multiple local reductions
ex it's looking to make sure we get a kernel that looks like
```
for () { ... }
if () { for() { ... } ... )
for () { ... }
if () { for() { ... } ... )
```
instead of:
```
for () { ... }
if () { 
  for() { ... } ... )
  for () { ... }
  if () { for() { ... } ... )
}

File: test/test_linearizer.py [hunk 4]
```diff
@@ -191,14 +186,96 @@ def test_early_end_local(self):
...
+      f"should have generated 3 BufferOps.STORE to the local buf but got {len(real_local_stores)}")
+    self.assertEqual(len(real_local_loads:=[u for u in k.uops if u.uop is UOps.LOAD and any([lb in u.vin for lb in local_buf])]), 3, \
+      f"should have generated 3 BufferOps.LOAD to the local buf but got {len(real_local_loads)}")
+    self.assertEqual((real_local_stores[1].vin[1].uop, real_local_stores[1].vin[1].arg), (UOps.CONST, 0))
+    self.assertEqual((real_local_loads[1].vin[1].uop, real_local_loads[1].vin[1].arg), (UOps.CONST, 0))
+    x = Tensor.randn(3,27,32).realize()
+    helper_linearizer_ast(ast, [x], wanna_output=[x.numpy().std(axis=2, ddof=0).reshape(-1)])
+
+  @unittest.skip
+  def test_multireduce_upcasting(self):
```
0xtimmy, line 231: this tests if upcasting is preserved in multiple reduceops, for ex:
```
  int alu0 = (gidx0*7);
  float val0 = data1[alu0];
  ...
  float val6 = data1[alu0+6];
  float alu1 = (val6+ ... +val0); // ast_parse will generate the entire sum because the reduceop depends on the load, of which there are 7
  data0[gidx0] = ((val6-alu1)+ ... +(val0-alu1)); // but this reduction depends on the result of the previous reduction, of which there is one, so we need to double check that the entire sum gets generated here as well
```

test/test_linearizer.py
```diff
@@ -191,14 +186,95 @@ def test_early_end_local(self):
...
+    k.upcast()
+    k.linearize()
+    define_globals = [u for u in k.uops.uops if u.uop is UOps.DEFINE_GLOBAL]
+    self.assertEqual(len([u for u in k.uops.uops if u.uop is UOps.LOAD and define_globals[1] in u.vin]), 7)
+    self.assertEqual(len([u for u in k.uops.uops if u.uop is UOps.ALU and u.arg is BinaryOps.SUB]), 7)
+    x = Tensor.randn(2,7).softmax().realize()
+    helper_linearizer_ast(ast, [x], wanna_output=[(x.numpy() - x.numpy().sum(axis=1, keepdims=True)).sum(axis=1)])
+
+  @unittest.skip
+  def test_multireduce_loop_scope(self):
```
0xtimmy, line 243: with multiple reductions, operations on the result of one reduction may get generated while rendering another so the ALU uops would get placed within the loop. This example calculates a layernorm of a tensor then sends it through another `SUM` to illustrate the issue

```
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 32; ridx0++) {
    float val0 = data1[alu0+ridx0];
    acc0 = (val0+acc0);
  }
+ float alu1 = (acc0*0.03125f); // we want the mean calculation to go here, so it's in all scopes
  float acc1 = 0.0f;
  for (int ridx1 = 0; ridx1 < 32; ridx1++) {
    float val1 = data1[alu0+ridx1];
-   float alu2 = (val1-acc0*0.03125f); // the linearizer generates the mean here while rendering the std reduction
+   float alu2 = (val1-alu1); 
    acc1 = ((alu2*alu2)+acc1);
  }
  float acc2 = 0.0f;
  for (int ridx2 = 0; ridx2 < 32; ridx2++) {
    float val2 = data1[alu0+ridx2];
    acc2 = (((val2-alu1)/sqrt((acc1*0.03125f)))+acc2); // BUT the kernel also uses the mean here, and unchanged it will be out-of-scope
  }
  data0[gidx0] = acc2;
```

chenyuxyz:
good tests! you can also comment how you set it up and what's expected to be tested directly in the test file. commenting in code is better than commenting on github. it's easier to update these tests in the future too

PR #4668, opened 2024-05-20
process replay benchmarks

Qazalin:
Triggers ASSERT_COMPILE=1 if your commit has `[run_process_replay]` in the message. Enabling this in benchmarks is a starting point and we wanna support:
1. All CI tests
2. BEAM=2 process replay

example of catching kernel change: https://github.com/tinygrad/tinygrad/actions/runs/9329767427/job/25682587397
example of untriggered process replay https://github.com/tinygrad/tinygrad/actions/runs/9329864104/job/25682791707


geohot:
Very cool idea to enable this with the commit message name!

Qazalin:
converting to draft to try extending `ASSERT_COMPILE=1` (possibly with a wrapper).

Qazalin:
I've tried a couple of options over the past two weeks and ASSERT_COMPILE=1 is the most productive and reliable way to do these tests.
We can make progress on benchmarks and normal CI,
then I can start writing a small wrapper for BEAM=2 replay.

PR #4669, opened 2024-05-20
add (0-x) fold

reddyn12:
already tested in test_uops


chenyuxyz:
prefer no new rules before UPat refactor is done, also these need explicit tests, also please read the pr diff

PR #4671, opened 2024-05-21
Remove uops deepcopy from PTX

SzymonOzog:
No longer needed after UOpsGraph refactor


test/test_linearizer.py
```diff
@@ -108,6 +108,7 @@ def test_end_local(self):
     self.assertEqual(k.uops.uops[-1].uop, UOps.ENDIF)
     self.assertLess(k.uops.uops.index([x for x in k.uops.uops if x.uop is UOps.STORE][-1]), k.uops.uops.index(k.uops.uops[-1]))
 
+  @unittest.skipIf(getenv("PTX"), "Ptx adds additional uops for load indexing")
```
chenyuxyz, line 111: instead of skipping, can you update tests to test PTX behaviors?

SzymonOzog: Sure thing, updated

PR #4675, opened 2024-05-21
Refactors linearizer `acc` to a Dict

0xtimmy:
moving in the direction of supporting multiple reduceops (sequential and parallel), this lets the linearizer track multiple accumulators


tinygrad/codegen/linearizer.py
```diff
@@ -418,22 +420,22 @@ def linearize(self):
     self.applied_opts_cache = self.applied_opts[:]
     return self
 
-  def ast_parse(self, x:LazyOp, acc: List[UOp], offs:Optional[List[int]], loaded_buffers:Dict[Union[MemBuffer, ConstBuffer, LocalBuffer], List[UOp]], do_reduce=False, cache=None) -> List[UOp]: # noqa: E501
+  def ast_parse(self, x:LazyOp, accs: Dict[LazyOp, List[UOp]], offs:Optional[List[int]], loaded_buffers:Dict[Union[MemBuffer, ConstBuffer, LocalBuffer], List[UOp]], do_reduce:Optional[List[UOp]]=None, cache=None) -> List[UOp]: # noqa: E501
```
Qazalin, line 423: can you rename do_reduce to match its new dtype, some ideas: reduce_accs, curr_accs, loop_accs

0xtimmy:
I picked `reduce_acc`

tinygrad/codegen/linearizer.py
```diff
@@ -418,22 +420,22 @@ def linearize(self):
...
+  def ast_parse(self, x:LazyOp, accs: Dict[LazyOp, List[UOp]], offs:Optional[List[int]], loaded_buffers:Dict[Union[MemBuffer, ConstBuffer, LocalBuffer], List[UOp]], reduce_acc:Optional[List[UOp]]=None, cache=None) -> List[UOp]: # noqa: E501
     if cache is None: cache = {}
     if x in cache: return cache[x]
     if x.op in BufferOps: return loaded_buffers[x.arg]
     if x.op in [UnaryOps.CAST, UnaryOps.BITCAST]:
       return [self.uops.add(UOps.BITCAST if x.op is UnaryOps.BITCAST else UOps.CAST,
-                            self.get_base_dtype(x.arg), (u,)) for u in self.ast_parse(x.src[0], acc, offs, loaded_buffers)]
-    if x.op in ReduceOps and not do_reduce:
+                            self.get_base_dtype(x.arg), (u,)) for u in self.ast_parse(x.src[0], accs, offs, loaded_buffers)]
+    if x.op in ReduceOps and not reduce_acc:
```
Qazalin, line 430: not reduce_acc might not be what you want here
None and [] both evaluate as true.
I think you meant `reduce_acc is None`

tinygrad/codegen/linearizer.py
```diff
@@ -418,22 +420,22 @@ def linearize(self):
...
+                            self.get_base_dtype(x.arg), (u,)) for u in self.ast_parse(x.src[0], accs, offs, loaded_buffers)]
+    if x.op in ReduceOps and reduce_acc is None:
       assert offs is None, "not available if we aren't doing reduce"
-      return acc
+      return accs[x]
 
-    values = [self.ast_parse(v, acc, offs, loaded_buffers, cache=cache) for v in x.src]
+    values = [self.ast_parse(v, accs, offs, loaded_buffers, cache=cache) for v in x.src]
     ops = {ReduceOps.SUM:BinaryOps.ADD, ReduceOps.MAX:BinaryOps.MAX}
-    if x.op in ops:
```
Qazalin, line 434: why do you need this change?
I used x.op in ops in the `linearizer` branch and tests passed

0xtimmy: it's not so much that `x.op in ops` needs to be deleted, it's that `reduce_acc` needs to be `not None`; and `ast_parse` won't propagate `reduce_acc` to any recursive calls

both work, because I do think we even call `ast_parse` in a way that would cause problems (ex if `reduce_acc` is not None, `x.op` is a `ReduceOp`)

but we could do `x.op in ops and reduce_acc is not None`

Qazalin: if reduce acc is None you've returned through this code path:
https://github.com/tinygrad/tinygrad/pull/4675/files#diff-63d57daaca59e5495bbb108da1c13e782f4a408db8503e015e8094ac07c87ea6R430-R432

0xtimmy: yeah i see it

PR #4676, opened 2024-05-21
Move from python3.8 to python3.10 and from Union to |

smurfd:
Hey,

Saw that maby a move from Python 3.8 to Python 3.10 was interesting.
Enforced python3.10 version for some of the workflows that did not have a version.
Left the ones with python3.11 or python3.12 untouched.

Switched from typing `Union[type1, type2]` to `type1 | type2`

ran tests locally, 
Did a stupid test for the `match`, `case` also available in python3.10
https://bpa.st/KQKA

found a typo in `tinygrad/shape/symbolic.py` that i will fix right after i create this pr.


smurfd:
3rd time is the charm?

PhillCli:
That could be done with the ruff pyupgrade rule https://docs.astral.sh/ruff/rules/non-pep604-annotation/, and it would also replace `Optional[X]` with `X | None`.

You should probably also update https://github.com/tinygrad/tinygrad/blob/master/setup.py#L26 with `python>=3.10`

smurfd:
Thanks PhillCli, ruff is great :)
added those changes aswell.

smurfd:
Closing this, tinygrad will follow pytorch and EOL of Python3.8 which is 2024-10 https://peps.python.org/pep-0569/

https://github.com/pytorch/pytorch/issues/120718

when its time, you should be able to do:
add to ruff.toml
`"UP007"   # non-pep604-annotation`

update setup.py
`python_requires='>=3.10',`

Then you should be able to run:
```
python3 -m ruff --target-version py310 --unsafe-fixes --diff pyupgrade .
```

EwoutH:
I think a discussion needs to take place if this project should follow [SPEC 0](https://scientific-python.org/specs/spec-0000/) or not.

PR #4678, opened 2024-05-21
example kernel that triggers Memory access fault for resnet on red

chenyuxyz:
@flammit fyi


flammit:
> @flammit fyi

got it - taking a look now

PR #4680, opened 2024-05-22
Int mulacc for ptx 

SzymonOzog:


SzymonOzog:
Hmm, seems to only be a problem on ocelot, compiles and runs fine on my machine and tiny 19

SzymonOzog:
Disabled in ocelot for now and submitted an issue

SzymonOzog:
found a workaround that doesn't anger ocelot so no need to disable it

tinygrad/renderer/assembly.py
```diff
@@ -181,7 +181,7 @@ def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
           assert args[1][0] != "i", "idx not supported"
           kk(f"mov.u32 %{args[1]}, {(self.gid if args[1][0] == 'g' else self.lid)[args[0]]};")
           r[u] = "%" + args[1]
-          kernel = [f".reg .u32 %{args[1]};"] + kernel
+          kernel = [f".reg .b32 %{args[1]};"] + kernel
```
chenyuxyz, line 184: add a comment for this?

PR #4682, opened 2024-05-22
correctness test for multireduce nested locals

Qazalin:
@0xtimmy the assert_allclose fails in the linearizer branch.

You can see the single reduce example in `TestLinearizer.test_upcast_with_locals`

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -355,6 +355,18 @@ class TestLinearizer(unittest.TestCase):
     assert accs[1].dtype == stores[1].vin[-1].dtype == dtypes.float
     assert stores[1].vin[0].uop is UOps.DEFINE_GLOBAL
 
+  @unittest.skip("multireduce isn't supported yet")
+  def test_upcast_multireduce_nested_local_upcast(self):
+    x, y, z, w = [Tensor.rand(1,128).realize() for _ in range(4)]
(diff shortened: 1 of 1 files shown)


test/test_linearizer.py
```diff
@@ -355,6 +355,21 @@ def test_upcast_with_locals(self):
     assert accs[1].dtype == stores[1].vin[-1].dtype == dtypes.float
     assert stores[1].vin[0].uop is UOps.DEFINE_GLOBAL
 
+  @unittest.skip("multireduce isn't supported yet")
+  def test_upcast_multireduce_nested_local_upcast(self):
+    x, y, z, w = [Tensor.rand(1,128).realize() for _ in range(4)]
```
0xtimmy, line 360: in `TestLinearizer.test_upcast_with_locals` the second operand of each matmul is initialized `Tensor.rand(128, 128)` but in this one they're all initialized with `Tensor.rand(1,128)`
I looked at the buffers going into `helper_linearizer_ast` and it seemed like I could leak in values if the shape is just `(1,128)`

the shapetrackers are the same between both tests and changing this line to, for ex:
`x, y, z, w = [Tensor.rand((1,128) if i % 2 == 0 else (1,128,128)).realize()  for i in range(4)]`

seems to make the test pass, with some fixes I made to how the linearizer stores local reduces

PR #4694, opened 2024-05-23
proposal: PHI nodes in TC shouldn't have children inside the loop

Qazalin:
The root cause for llama BEAM=2 failing https://github.com/tinygrad/tinygrad/actions/runs/9145074079/job/25143631108?pr=4647 is that uops.py places children of PHI outside the loop.
In tensor cores, `GRAPHUOPS=1 python3 test/test_linearizer.py TestLinearizer.test_tensor_cores_upcast_unroll_minimal`
image
UOps.PHI is parent to the next WMMA loop child.

The spec in this diff specifies that PHI should be the stopping point of loop children. following the flow: RANGE -> DEFINE_ACC -> PHI -> ENDRANGE

WMMA [doesn't follow this structure](https://github.com/tinygrad/tinygrad/blob/4753283221e9f32ffdc228ce7798d86eecb120f8/tinygrad/codegen/linearizer.py#L251-L254).

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -500,7 +500,6 @@ class TestLinearizer(unittest.TestCase):
       assert len(tc_actions) == 9, f"get_linearizer_actions should contain 9 possible TC actions, only got {len(tc_actions)}"
 
   @unittest.skipIf(Device.DEFAULT != "METAL", "these opts are only valid on METAL")
-  @unittest.skip("UOps toposort is wrong, CAST child should render outside RANGE")
   def test_tensor_cores_upcast_unroll(self):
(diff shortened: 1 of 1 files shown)


Qazalin:
I think a viable option is to use WMMA as parent:
```
float2 wmma8 = __WMMA_8_8_8_half_float(cast2, half2(val2,val6), float2(acc0.x,acc0.y));
acc0.x = wmma8.x;
acc0.y = wmma8.y;
float2 wmma12 = __WMMA_8_8_8_half_float(cast3, half2(val3,val7), float2(acc0.x,acc0.y));
acc0.x = wmma12.x;
acc0.y = wmma12.y;

// after:

float2 wmma8 = __WMMA_8_8_8_half_float(cast2, half2(val2,val6), float2(acc0.x,acc0.y));
acc0.x = wmma8.x;
acc0.y = wmma8.y;
float2 wmma12 = __WMMA_8_8_8_half_float(cast3, half2(val3,val7), wmma8);
acc0.x = wmma12.x;
acc0.y = wmma12.y;
```

chenyuxyz:
@flammit what do you think? It sounds fine to me that instead of using acc to chain unrolled wmma, we pass previous wmma directly. Other than of our spec requirement, is there a reason to prefer one or another?

Also is there a proposed change to this too?

flammit:
> @flammit what do you think? It sounds fine to me that instead of using acc to chain unrolled wmma, we pass previous wmma directly. Other than of our spec requirement, is there a reason to prefer one or another?
> 
> Also is there a proposed change to this too?

Seems fine to chain the unrolled WMMAs and I don't see a performance impact.  It should also look a lot nicer from the standpoint of the generated code (way fewer lines):

```
// after:

float2 wmma8 = __WMMA_8_8_8_half_float(cast2, half2(val2,val6), float2(acc0.x,acc0.y));
float2 wmma12 = __WMMA_8_8_8_half_float(cast3, half2(val3,val7), wmma8);
acc0.x = wmma12.x;
acc0.y = wmma12.y;
```

@Qazalin do you want a proposed diff/PR for this?

PR #4698, opened 2024-05-23
Fix tensor cores in PTX

SzymonOzog:
Fixes,  `‎TestKernelOpts.test_tensor_core_opts`, previously assumed that all wmmas take the output of previous wmma as input but that's not always the case if we unroll


chenyuxyz:
test?

SzymonOzog:
> test?

Already tested by `TestKernelOpts.test_tensor_core_opts` but gpuocelot doesn't support wmma so it didn't get caught in CI

chenyuxyz:
Cool! verified `PTX=1 python3 -m pytest test/test_linearizer.py -k test_tensor_core_opts` is fixed by this. Added some instructions for reviewer to test your change is always welcome

PR #4700, opened 2024-05-23
fix UOps.CAST noop for vectorized dtypes

Qazalin:
dtypes.float.vec(2) is not dtypes.float.vec(2) but == works

before:
```
kernel void test(device float* data0, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  float2 val0 = (float2)(*((device float2*)(data0+0)));
  *(data0+0) = sqrt((float2)(val0).x);
}
```

after:
```
kernel void test(device float* data0, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  float2 val0 = (float2)(*((device float2*)(data0+0)));
  *(data0+0) = sqrt(val0.x);
}
```

### DIFF
--- a/test/test_uop_graph.py
+++ b/test/test_uop_graph.py
@@ -1,7 +1,7 @@
 import unittest
 from tinygrad import dtypes, Variable
 from tinygrad.dtype import PtrDType
-from tinygrad.ops import BinaryOps, MemBuffer, TernaryOps, UnaryOps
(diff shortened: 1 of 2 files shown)


chenyuxyz:
fixed this in #4704

PR #4708, opened 2024-05-24
Context-based uops queues instead of relying on heapq

uuuvn:
was doing rdna3 assembly, to make it fast without a lot of code duplication (i've ended up with a mini uops but rdna3ops) i need to have more fine-grained control over UOps ordering to avoid creating another graph layer and linearizer but layer down and with rdna-specific quirks like avoiding consequtive dependent alu instructions to prevent pipeline stalls, etc

i've heard the heapq problem mentioned in voice channel so i decided to pr this independently of my rdna3 adventures


uuuvn:
it works by first adding dependencies to graph to force external range inputs to be linearized before entering loop and then having multiple queues for each context (None is root, then RANGE UOp for inner contexts)

Qazalin:
what is the blocker to make this use virtual edges instead of another special toposort?
we wanna make the `graph: DefaultDict[UOp, List[UOp]]` such that any toposort is valid.

uuuvn:
> what is the blocker to make this use virtual edges instead of another special toposort? we wanna make the `graph: DefaultDict[UOp, List[UOp]]` such that any toposort is valid.

i don't think it's possible to do it such that all toposorts are valid and all valid uops sequences can be represented with just virtual edges. For example if we have const node, something that uses that const node and two ranges we want to make sure that const is defined outside of those ranges. We can do this either by having a notion of context in toposort (with current heapq being a weak hacky impl of that) or by making everything that's not part of the range depend on range end phi or range depend on it. Either way we have to decide if const will be declared before or after the range 1/2 before we make virtual edges.

PR #4718, opened 2024-05-24
nv/amd profiler

nimlgen:


nimlgen:
@geohot, this works, but I really don't like the number of lines. There is a lot of duplication between NV and AMD when it comes to code parts that use HWQueues (like AMD/NVAllocator). Should we create base classes for them (which can also serve as a specification for the API HCQ needs), such as HCQCompatibleCompiled and HCQCompatibleAllocator? This is just off the top of my head; I will think more about how to remove this duplication.

geohot:
Do what you think is good. Keep in mind that any abstractions should work for Qualcomm and Intel too.

geohot:
In the context of making LLaMA fast I'm still quite interested in this. Even if it's not perfect, we should probably get it merged so it doesn't go stale like it did.

PR #4728, opened 2024-05-25
Bitshift

SzymonOzog:
Multiplies and divides by a power of 2 can be executed with a simple bitshift, for cstyle backends it's done by the compiler but assembly needs this explicitly


geohot:
I think we should block more patterns on UPat refactor, and perhaps block that on process replay to confirm it doesn't change anything.

SzymonOzog:
> I think we should block more patterns on UPat refactor, and perhaps block that on process replay to confirm it doesn't change anything.
 Updated with UPat

PR #4733, opened 2024-05-26
multireduce OptOps tests (start)

Qazalin:
@0xtimmy This is what I mean by using OptOps. First just use the existing tests with a multireduce AST and ensure they match the numpy version.

`test_tensor_core_opts_multireduce` fails for AST:
```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(128, 128, 1), strides=(128, 1, 0), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ ADD 
  2    ┣━┳ SUM (2,)
  3    ┃ ┗━┳ MUL 
  4    ┃   ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(128, 128, 128), strides=(128, 0, 1), offset=0, mask=None, contiguous=False),)))
  5    ┃   ┗━━ LOAD MemBuffer(idx=2, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(128, 128, 128), strides=(0, 1, 128), offset=0, mask=None, contiguous=False),)))
  6    ┗━┳ SUM (2,)
  7      ┗━┳ MUL 
  8        ┣━━ LOAD MemBuffer(idx=3, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(128, 128, 128), strides=(128, 0, 1), offset=0, mask=None, contiguous=False),)))
  9        ┗━━ LOAD MemBuffer(idx=4, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(128, 128, 128), strides=(0, 1, 128), offset=0, mask=None, contiguous=False),)))
```
because it's indexing the wrong buffer.

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -988,7 +988,8 @@ def _temp_create_multireduce_ast(r0:Tensor, r1:Tensor, merge=lambda r0,r1: LazyO
     else: arg = op.arg
     return LazyOp(op.op, tuple(_deep_replace(x, offset) for x in op.src), arg)
   op0_loads = len([x for x in op0.lazyops if x.op is BufferOps.LOAD])
-  op = LazyOp(BufferOps.STORE, (merge(_deep_replace(op0), _deep_replace(op1, op0_loads)), ), MemBuffer(0, s0[-1].ast[-1].arg.dtype, s0[-1].ast[-1].arg.st))
+  out = merge(_deep_replace(op0), _deep_replace(op1, op0_loads))
+  op = LazyOp(BufferOps.STORE, (out, ), MemBuffer(0, s0[-1].ast[-1].arg.dtype, s0[-1].ast[-1].arg.st))
   print_tree(op)
   return op,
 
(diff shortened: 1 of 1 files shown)


Qazalin:
TC issue happened both on AMD and METAL.
CUDA failed on tinybox because WMMA is using too many registers. I think the uops should still be correct though.

Qazalin:
@0xtimmy  I'm fixing the TC stuff (in progress) - it needs some deep refactors.
Can you try looking deeper into the GPU CI issue with temp buffers:
https://github.com/tinygrad/tinygrad/pull/4259#discussion_r1612224675
Why is doing temp0, temp1 any different from just sharing temp. Are you sure we're not missing a BARRIER?

0xtimmy:
sweet, will do

0xtimmy:
an update: I got the issue to reproduce on an aws instance and it seems like the issue is in the local reduce (ex below) and it is that the top eight threads are reading the wrong value from `temp[0]`. 
adding more barriers doesn't seem to change anything; however, if `acc1` gets stored at an index >7 all threads will read the correct value

looking into it more

```
if (alu0) {
          float val1 = temp[ridx1];
          data0[4+gidx0*16+ridx1] = val1;
          acc1 = (val1+acc1);
        }
        data0[gidx0] = acc1;
        temp[0] = acc1; <-
      }
      barrier(CLK_LOCAL_MEM_FENCE);
      float val2 = temp[0]; <- ex: val2 = 16 for lidx1=[0:7] (the correct value) but -9.5367431640625e-07 for lidx1=[8:15]
```

PR #4736, opened 2024-05-26
docs: fix nn rendering

chenyuxyz:
need to import nn at the start of the page. maybe there's a better way to move it on top?


wozeparrot:
weird just having in tensor.py worked locally for me

chenyuxyz:
weird, for me locally behaves the same as the deployed one. maybe you can fix this better?

PR #4738, opened 2024-05-27
rename Tensor.replace -> Tensor._replace

ridoy:
Tensor.replace shouldn't be used by users ([source](https://youtu.be/-iH5wvFnsKs?t=20290)).

- Renamed symbol `Tensor.replace` -> `Tensor._replace`
- Auto updated replace references in examples/
- Replaced with `Tensor.assign` in nn.state.load_state_dict
- Removed replace from docs


chenyuxyz:
see discussion on https://github.com/tinygrad/tinygrad/pull/3968#issuecomment-2024330548

It's still not clear if this is the final API we want, but we will keep this as a public version API for now

PR #4739, opened 2024-05-27
Taylor Approximation (at about 0) for exponential proof of concept

mesozoic-egg:
Simple POC, just want to know if I'm on the right track


geohot:
Tests failing.

PR #4745, opened 2024-05-27
tweak docs style

wozeparrot:
I do like this more, not sure about you @chenyuxyz

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -13,9 +13,8 @@ nav:
     - Function: function.md
     - dtypes: dtypes.md
     - nn (Neural Networks): nn.md
-  - Notes:
-    - Developer: developer.md
-    - Environment: env_vars.md
+    - Environment Variables: env_vars.md
+  - Developer: developer.md
 #- tinygrad: reference/
 
 #extra_css:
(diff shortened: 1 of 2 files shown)


chenyuxyz:
sure was testing something similar

btw do you know how to change favicon color based on dark / light mode? white favicon is bad with light mode broswer

mkdocs.yml
```diff
@@ -3,17 +3,19 @@ site_name: tinygrad docs
...
+  - index.md
+  - Quickstart: quickstart.md
+  - Showcase: showcase.md
+  - MNIST Tutorial: mnist.md
+  - API Reference:
+    - Tensor: tensor.md
+    - Function: function.md
+    - dtypes: dtypes.md
+    - nn (Neural Networks): nn.md
+  - Notes:
```
chenyuxyz, line 16: Notes is meh... env var should be in API and developer can have its own section?

wozeparrot: yea i didn't really know where to put these two

wozeparrot:
don't think its possible to change favicon based on light or dark mode, also the browser tab bar color isn't always determined by the site theme color or even the system color preference.

wozeparrot:
oh it is possible if we use a svg favicon

PR #4750, opened 2024-05-27
docs: polish tensor.py

chenyuxyz:


tinygrad/tensor.py
```diff
@@ -1962,7 +1968,7 @@ def square(self):
     return self*self
   def clip(self, min_, max_):
     """
-    Clips (limits) the values in the tensor between `min_` and `max_` element-wise.
+    Clips (clamps) the values in the tensor between `min_` and `max_` element-wise.
```
wozeparrot, line 1971: is limits not better here? tho i guess both work

chenyuxyz: in torch, `clip` is an alias of `clamp`

PR #4742, opened 2024-05-27
TC accessing wrong global bufs in multireduce AST

Qazalin:
latest update: Simplest multi reduce tensor core cases work now, for both indexing and buffer aliases.

(old) Update: now the buffer aliases in TC are correct but idxs into those buffers are wrong.

a fused wmma should look like:
https://gist.github.com/Qazalin/392c710da95083fda2aab35324400412

@flammit sorry for the huge diff - I isolated the failure in `python3 test/test_linearizer.py TestKernelOpts.test_tensor_core_opts_multireduce`. The idea is that when fusing two back-to-back matmuls with TC=1 the global bufs are wrong:

METAL=1
normal:

```c
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 128; ridx0++) {
    float val0 = *(data1+alu0+ridx0);
    float val1 = *(data2+gidx1+(ridx0*128));
    acc0 = ((val0*val1)+acc0);
  }
  float acc1 = 0.0f;
  for (int ridx1 = 0; ridx1 < 128; ridx1++) {
    float val2 = *(data3+alu0+ridx1);
    float val3 = *(data4+gidx1+(ridx1*128));
    acc1 = ((val2*val3)+acc1);
  }
```
tc:

```c
  float2 acc0 = float2(0.0,0.0);
  for (int ridx0 = 0; ridx0 < 16; ridx0++) {
    float2 val0 = (float2)(*((device float2*)(data1+alu7+(ridx0*8)+alu5+alu2)));
    float2 val1 = (float2)(*((device float2*)(data2+alu6+(ridx0*1024)+alu4+alu3)));
    float2 wmma0 = __WMMA_8_8_8_float_float(float2(val0.x,val0.y), float2(val1.x,val1.y), float2(acc0.x,acc0.y));
    acc0.x = wmma0.x;
    acc0.y = wmma0.y;
  }
  float2 acc1 = float2(0.0,0.0);
  for (int ridx1 = 0; ridx1 < 16; ridx1++) {
    float2 val2 = (float2)(*((device float2*)(data1+alu7+(ridx1*8)+alu5+alu2))); // this should be data3
    float2 val3 = (float2)(*((device float2*)(data2+alu6+(ridx1*1024)+alu4+alu3)));
    float2 wmma1 = __WMMA_8_8_8_float_float(float2(val2.x,val2.y), float2(val3.x,val3.y), float2(acc1.x,acc1.y));
    acc1.x = wmma1.x;
[...]

PR #4763, opened 2024-05-28
map tensor core bufs to reduceop

Qazalin:
In multireduce tensor core kernels, we need to:
1. Map tensor core bufs to reduceop
2. Map the alias of those bufs
Allowing reduce_reduceop access to the local aliases of its A and B.

verified correctness in the full linearizer diff: #4742

The assumption for now is to fuse reduces with exactly the same shape and strides.

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -35,7 +35,6 @@ class Opt:
     return self.axis
 
 class TensorCoreOptions(NamedTuple):
-  bufs: Tuple[int, int] # the local aliased buffers for A and B
   axes: List[int] # the location of the original N and M axes if still in the shape
   axes_exist: List[bool] # true if the original N and M axes are still in the shape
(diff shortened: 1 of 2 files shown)


Qazalin:
This is the fully diff for TC: https://github.com/tinygrad/tinygrad/pull/4742/commits/0d90737bd6fda04b7dff4d14f4ec50c33ea97ac5 (excl BARRIER addition) https://github.com/tinygrad/tinygrad/pull/4742#issuecomment-2135900720

Qazalin:
This is the full diff for TC: https://github.com/tinygrad/tinygrad/pull/4742/commits/0d90737bd6fda04b7dff4d14f4ec50c33ea97ac5 (excl BARRIER addition) https://github.com/tinygrad/tinygrad/pull/4742#issuecomment-2135900720

Qazalin:
The assumption for now is to fuse reduces with exactly the same shape and strides.

PR #4769, opened 2024-05-29
nv in benchmarks

nimlgen:
@chenyuxyz, can we add nv into benchmarks, any objections?


chenyuxyz:
12 minutes is quite long. maybe you can keep this one as draft. and replace the ones that NV is faster than CUDA with NV when it's ready, then it can share the same txt file.

geohot:
I think NV should replace CUDA in benchmarks. Maybe leave one CUDA, but we definitely don't need to run both.

PR #4780, opened 2024-05-30
Implement a gradient scaling operation

nikvaessen:
As a side-project, I'm implementing wav2vec 2.0 pre-training and fine-tuning in TinyGrad. During this process I expect to come across some bugs or unimplemented features. Two features I know are not available are the gumbel-softmax operation and gradient scaling. 
 
Hereby a PR which adds the gradient scaling operation. I think this is useful not only for my use-case (in wav2vec 2.0, the gradient of the first 7 layers should be 10x smaller than other layers), but also for implementing adversarial losses (e.g., for GANs you can maximize the loss by doing `loss.gradscale(-1).backward()`).

### DIFF
--- a/test/test_gradscale.py
+++ b/test/test_gradscale.py
@@ -1,0 +1,116 @@
+import unittest
+import numpy as np
+
+from tinygrad import Tensor
+from tinygrad.nn.optim import Adam
+from tinygrad.nn import Linear
+
+
+def _get_test_scales():
+    ints = [i for i in range(-10, 11)]
+    floats = [float(i) for i in ints]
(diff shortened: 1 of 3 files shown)


chenyuxyz:
i think the proper way to implement this is to change the forward to `t * scale + t.detach() * (1 - scale)` and let it handled by autograd

chenyuxyz:
I don't think this needs to be a public API in `tensor.py`. How does torch / jax do this? Will consider this if there's a matching API.

nikvaessen:
There is no matching API for this, so if you think it's not worth to support this, this can be closed.

PR #4786, opened 2024-05-30
Multireduce atomic local load/store test

0xtimmy:
With multiple reduceops the result of a warp-wise reduce must be loaded back into every thread. So we need to add two barriers, one to make the threads wait for `temp[0]` to get populated before loading it into each thread, and another to make sure each thread has done the load before allowing others to attempt the next reduceop, overwriting the contents of `temp`.

This test has a large `lidx` which aims to cause the thread scheduler to let synchronization issues arise if both barriers are not in place
```
...
  barrier(CLK_LOCAL_MEM_FENCE);
  float val2 = temp[0]; <---
  // barrier(CLK_LOCAL_MEM_FENCE); <--- make sure a barrier ends up
  float acc2 = 0.0f;
   ...
  }
  temp[lidx1] = acc2; <---
  barrier(CLK_LOCAL_MEM_FENCE);
...
```


0xtimmy:
re: #4259

Qazalin (commented):
can you also explore a nested local reduce? I've added some in `test_double_reduce_multireduce` but they didn't test this case

test/test_linearizer.py
```diff
@@ -1075,6 +1075,19 @@ def test_local_and_grouped_reduce_multireduce(self):
...
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_shared, "test requires shared")
+  def test_atomic_local_multireduce(self):
+    # reducops will need to use the local buffer to load the result of a local reduce into every thread, barriers are needed on both sides
+    # of the load to ensure 1) the correct value is in the local buffer and 2) the value isn't overwritten by the next reduceop
+    N = 512
+    Tensor.manual_seed(1882)
+    a,b = Tensor.rand(4,4,N).realize(), Tensor.rand(4,4,N).realize()
+    r0,r1 = a.sum(-1), b.sum(-1)
+    ast = _temp_create_multireduce_ast(r0, r1)
```
Qazalin, line 1088: I think `_temp_create_multireduce_ast` should be able to do back to back reduces with minimal to no changes
Just return the leaf reduceop `merge=lambda r0,r1: r1`

test/test_linearizer.py
```diff
@@ -1075,6 +1081,120 @@ def test_local_and_grouped_reduce_multireduce(self):
...
+    a,b = Tensor.rand(4,4,N).realize(), Tensor.rand(4,4,N).realize()
+    dummy = Tensor.rand(4,4,1).realize()
+    r0,r1 = (a-dummy).sum(-1), b.sum(-1)
+    ast = _temp_create_multireduce_ast(r0, r1, replace_idxs={2:r1}, merge=lambda r0,_: r0)
+    helper_linearizer_ast(ast, [a,b], [[Opt(OptOps.GROUP, 0, 2)]])
+
+  @unittest.skip("multireduce isn't supported yet")
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_shared, "test requires shared")
+  def test_atomic_store_unrolled_multireduce(self):
```
0xtimmy, line 1162: this tests an issue with kernels that only have the thread-wise reduce like below. We want to make sure that the first reduce finishes before the data gets loaded and stored in the local buffer for the next reduce (even if they don't depend on each other)

```
  float2 val0 = (float2)(*((__global float2*)(data1+alu1)));
  // float2 val3 = (float2)(*((__global float2*)(data2+alu1))); <--- data for the next reduce
  temp[lidx0] = (val0.y+val0.x);
  // temp[lidx0] = (val3.y+val3.x); <--- we want to make sure this store happens after the first reduce
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu0) {
    float acc0 = 0.0f;
    for (int ridx0 = 0; ridx0 < 2; ridx0++) {
      float val1 = temp[ridx0];
      acc0 = (val1+acc0);
    }
    temp[0] = acc0;
  }
  barrier(CLK_LOCAL_MEM_FENCE);
  float val2 = temp[0];
```

test/test_linearizer.py
```diff
@@ -1075,6 +1081,120 @@ def test_local_and_grouped_reduce_multireduce(self):
       [Opt(OptOps.LOCAL, 0, 2), Opt(OptOps.GROUPTOP, 0, 2), Opt(OptOps.UPCAST, 0, 8), Opt(OptOps.UNROLL, 1, 4)],
     ])
 
+  @unittest.skip("multireduce isn't supported yet")
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
+  def test_just_enough_barriers(self):
```
0xtimmy, line 1086: a quick check to make sure there are barriers between local buffer reads/writes and there aren't duplicate buffers

test/test_linearizer.py
```diff
@@ -1075,6 +1081,120 @@ def test_local_and_grouped_reduce_multireduce(self):
...
+    r0,r1 = a.sum(), b.sum()
+    ast = _temp_create_multireduce_ast(r0, r1)
+    helper_linearizer_ast(ast, [a,b], [
+      [Opt(OptOps.UNROLL, 0, 2), Opt(OptOps.GROUP, 0, 2)]
+    ])
+
+  @unittest.skip("multireduce isn't supported yet")
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_shared, "test requires shared")
+  def test_atomic_store_nested_range_multireduce(self):
```
0xtimmy, line 1174: check that nested ranges are fine as well

test/test_linearizer.py
Qazalin: can these very specific tests be replaced with correctness tests?
Ideally we want np.testing.assert_allclose to catch everything.

0xtimmy: the asts/kernels they generate are tested for correctness in the other tests. this is to catch if there are duplicate barriers, which doesn't break correctness but we don't want all the same

I'm happy to move this check into the other tests though

Qazalin: If correctness doesn't catch this it's ok.

Qazalin (commented):
is this ready for review?

Qazalin:
UOps.STORE's vin[3] is rendered as the gate.

If you move to gated stores it should also nest the compute for the store inside the gate (incl ALUs for indexing and the loop or whatever computation is required for the store value)
Because the vin[3] gate doesn't do this we have UOps.IF but we wanna generalize all of these to the gate and delete IF.

test/test_linearizer.py
Qazalin: oh yeah another test failing in the `linearizer` branch. Thanks for this!

test/test_linearizer.py
```diff
@@ -982,16 +982,22 @@ def check_opt(opts, create_k, expected_color_size):
...
+                                 merge=lambda r0,r1: LazyOp(BinaryOps.ADD, (r0, r1))) -> Tuple[LazyOp, ...]:
   assert len(s0:=r0.schedule()) == 1 and len(s1:=r1.schedule()) == 1, "inputs should be realized"
+  assert all({idx:replace_idxs[idx] is r0 or replace_idxs[idx] is r1 for idx in replace_idxs}.values()), "replace idxs should be in {{r0, r1}}"
   op0, op1 = s0[0].ast[0].src[0], s1[0].ast[0].src[0]
+  _replace_idxs = {idx:(op0 if replace_idxs[idx] is r0 else op1) for idx in replace_idxs}
   def _deep_replace(op:LazyOp, offset=0):
-    if op.op is BufferOps.LOAD: arg = MemBuffer(op.arg.idx+offset, op.arg.dtype, op.arg.st)
+    if op.op is BufferOps.LOAD:
+      if op.arg.idx+offset in _replace_idxs: return _replace_idxs[op.arg.idx+offset]
+      else: arg = MemBuffer(op.arg.idx+offset, op.arg.dtype, op.arg.st)
```
Qazalin, line 994: fyi this will break if you try use the same MemBuffer twice in the AST - fine for now, this is a temp function anyways

PR #4788, opened 2024-05-30
compute aliased buffer idxs pre reduce

Qazalin:
Tensor cores change the idxs in the middle of rendering a reduceop.
However, this change should happen only once per kernel no matter how many reduceops we wanna render.

With this diff, indexes into the local aliases are computed before rendering any reduceops.

The expected diff for two matmuls of exactly the same TensorCoreOptions on different `a` and `b` in one kernel is:
```diff
- extern "C" __global__ void test(half* data0, const half* data1, const half* data2) {
+ extern "C" __global__ void test(half* data0, const half* data1, const half* data2, const half* data3, const half* data4) {
  int alu9 = (alu4+alu5+alu8+alu7+alu3);
+   // first reduce
    half val0 = data1[alu9];
  half val1 = data1[alu9+1];
  half val2 = data1[alu9+8];
  // ...
  half val11 = data2[alu10+144];
  float4 wmma0 = __WMMA_8_16_16_half_float(make_half8(val0,val1,val4,val5,val2,val3,val6,val7), make_half4(val8,val9,val10,val11), make_float4(0.0,0.0,0.0,0.0));

+  // second reduce
  half v0 = data3[alu9];
  half v1 = data3[alu9+1];
  // ...
  half v11 = data4[alu10+144];
  float4 wmma1 = __WMMA_8_16_16_half_float(make_half8(v0,v1,v4,v5,v2,v3,v6,v7), make_half4(v8,v9,v10,v11), make_float4(0.0,0.0,0.0,0.0));

  data0[alu11] = (half)(wmma0.x)+(half)(wmma1.x);
  data0[alu11+1] = (half)(wmma0.y)+(half)(wmma1.y);
  data0[alu11+128] = (half)(wmma0.z)+(half)(wmma1.z);
  data0[alu11+129] = (half)(wmma0.w)+(half)(wmma1.w);
 }
 ```
 where both reduces use the same idxs.

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -177,8 +177,7 @@ class Linearizer(Kernel):
     self.loop_uops.update(new_loops)
     return tuple(new_loops.values())
 
-  def render_reduceop(self, reduceop:LazyOp, accs:Dict[LazyOp, List[UOp]], loaded_buffers:Dict[Union[MemBuffer, ConstBuffer, LocalBuffer], List[UOp]],
-                      global_idxs, local_idxs, upcast_idxs, full_upcast_idxs, reduce_idxs, fake_reduce_idxs):
+  def index_local_aliases(self, global_idxs, local_idxs, reduce_idxs, upcast_idxs, full_upcast_idxs):
     def calc_tc_idxs(local_sizes: List[int], aliases: List[List[int]]):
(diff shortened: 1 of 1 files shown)


Qazalin:
this unblocks progress on the simplest fused TC cases.
Next I'm exploring a spec for:
1. Back to back fused matmul - `TODO: apply tc even if the sources are not from LOAD` is a prereq.
2. Support applying TC on a subset of self.reduceops. eg, if because of dtypes some reduces don't have a valid TC.

PR #4790, opened 2024-05-30
CMPLT is safe to pad

chenyuxyz:
0 < 0 evals to False


chenyuxyz:
training resnet on red looks good. will merge after training is done
image

PR #4791, opened 2024-05-31
Refactor UOps match logic to UPat instead of dicts

alecchendev:
Refactors the `_match` function to use `UPat`s. I currently added a `from_dict` method on `UPat` so that this PR can do the main refactor without also including all the changes to rewrite the patterns from dicts to `UPat`s. If this looks good I'm happy to add those here or in a followup PR (and remove the `from_dict` method).

Speed benchmark (M1 macbook air):

```
# master (236390aa)
***** model forward in  31.49 ms
***** model schedule in  10.47 ms
***** model lower in 2608.46 ms
# this pr
***** model forward in  31.40 ms
***** model schedule in  10.41 ms
***** model lower in 1556.09 ms
```

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
-from typing import Iterator, Optional, Tuple, Any, Dict, List, DefaultDict, Set, Callable
+from typing import Iterator, Optional, Tuple, Any, Dict, List, DefaultDict, Set, Callable, Union
 import functools, itertools, heapq
 from collections import defaultdict
 from enum import Enum, auto
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from tinygrad.dtype import dtypes, DType
 from tinygrad.shape.symbolic import sint, Variable
 from tinygrad.ops import UnaryOps, BinaryOps, TernaryOps, exec_alu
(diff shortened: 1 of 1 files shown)


Qazalin:
@alecchendev What backend did you benchmark against?

I ran `PROFILE=0 python3 test/external/external_benchmark_schedule.py` on both an M1 mac and tinybox green - This diff causes a regression.
eg. tinybox green
master:
```
***** model forward in  51.64 ms
***** model schedule in  16.26 ms
***** model lower in 2097.72 ms
```
this branch:
```
***** model forward in  52.52 ms
***** model schedule in  16.24 ms
***** model lower in 2141.94 ms
```

alecchendev:
> What backend did you benchmark against?
> 
> I ran PROFILE=0 python3 test/external/external_benchmark_schedule.py on both an M1 mac and tinybox green - This diff causes a regression.

Huh, interesting, I ran the same command on my M1 macbook (air) to get my results.

alecchendev:
Oh but actually I'm noticing if I run the test many times I get results closer to yours. Not sure why the first time I had run the test yesterday and today it showed the opposite

geohot:
Oh this is a much better way to do the bounty!

geohot:
Merged. @alecchendev the bounty is yours once you merge the second PR.

tinygrad/codegen/uops.py
```diff
@@ -67,45 +67,62 @@ def uop_alu_resolve(u:UOp) -> sint:
...
+    else: assert vin is None
+    arg = pat.get("arg")
+    allow_len = pat.get("__allow_len__", set())
+    return cls(uop, arg, vin, name, dtype, allow_len)
+
+def _match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> bool:
+  if pat.name in store and store[pat.name] != uop: return False
+  if pat.name is not None: store[pat.name] = uop
+  if pat.arg is not None and uop.arg != pat.arg: return False
+  if isinstance(pat.dtype, set) and uop.dtype not in pat.dtype: return False
```
geohot, line 98: You can nest these ifs and only use one isinstance.

alecchendev:
Awesome! should get around to the next PR within a few days

PR #4799, opened 2024-06-01
process replay for test/*

Qazalin:
I used a .sh version of this to verify the UPat from_dict diff, now we have the infra to run it on github actions.
- Starts automatically if the commit message contains `[run_process_replay]`.
- Starting manually from github's UI is possible as well: https://github.com/tinygrad/tinygrad/actions/workflows/test.yml


Qazalin:
@chenyuxyz fyi I need to bump timeout to 20m because process replay runs tests on both master and the PR.
https://github.com/tinygrad/tinygrad/pull/4645

PR #4803, opened 2024-06-01
faster submit on compute queue

nimlgen:


nimlgen:
done with the new indirect `bind` api

PR #4804, opened 2024-06-01
Refactor to class style

Eliulm:
Mostly copied from @wozeparrot. Testing new impl at https://wandb.ai/wahl/MLPerf-BERT/runs/5an2hioj?nw=nwuserwahl


Eliulm:
This [run](https://wandb.ai/wahl/MLPerf-BERT/runs/8awkco8a?nw=nwuserwahl) trains with full float and half linear. Seems like its past the 72k mark

PR #4805, opened 2024-06-02
perf: default LAZYCACHE to 0

chenyuxyz:
wip testing. only CONST is cached this way. 10% faster for unjitted llama


geohot:
I'm pretty sure we want the LAZYCACHE, if defaulting to 0 is really preferred, we should just remove it. It does add complexity.

PR #4806, opened 2024-06-02
layernorm backward is indepedent of its mean

chenyuxyz:
10% less kernel for `test/models/test_train.py::TestTrain::test_transformer`, should make llm.c faster


geohot:
Any blockers on this?

chenyuxyz:
was testing on bert. i think these norms don't have good enough test coverage especially for training, but I can merge this first

PR #4812, opened 2024-06-03
support view in nv/amd

nimlgen:


geohot:
length and size should use the passed in size, not length - offset. it can truncate from the end

PR #4813, opened 2024-06-03
improve augmented assign error message

Qazalin:
The "must be contiguous for assign" error message can be incomprehensible for a new user - I think this gives more context and an actionable way to resolve it. https://discord.com/channels/1068976834382925865/1070745817025106080/1246946660790173726

image


geohot:
LGTM, I think Rust does an amazing job with error messages like this.

Qazalin:
yeah it was the my inspo

PR #4820, opened 2024-06-03
WIP: Remove BinaryOps.SUB

wpmed92:
This PR removes `BinaryOps.SUB`, and implements the sub functionality using a combination of `BinaryOps.ADD` and `UnaryOps.NEG`


geohot:
So we want to leave SUB in the backends, it should just be removed from Tensor to the Linearizer. It should be readded in a final processing pass before UOp linearization but after simplification.

PR #4821, opened 2024-06-03
Construct UOps patterns using UPat

alecchendev:
Followup on #4791. While doing this, I took the chance to use positional arguments wherever it wasn't too hard, let me know if I should reverse these.

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -77,21 +77,6 @@ class UPat:
   dtype: Optional[Union[DType, Set[DType]]] = None
   allow_len: Set[int] = field(default_factory=set)
 
-  @classmethod
-  def from_dict(cls, pat:Dict[str, Any]) -> UPat:
-    name, uop, dtype = pat.get("__name__"), pat.get("uop"), pat.get("dtype")
-    assert isinstance(name, str) or name is None
-    assert isinstance(uop, (UOps, set)) or uop is None
-    assert isinstance(dtype, (DType, set)) or dtype is None
-    vin = pat.get("vin")
-    if isinstance(vin, list): vin = [UPat.from_dict(x) for x in vin]
-    elif isinstance(vin, tuple): vin = tuple(UPat.from_dict(x) for x in vin)
-    elif isinstance(vin, dict): vin = UPat.from_dict(vin)
-    else: assert vin is None
-    arg = pat.get("arg")
-    allow_len = pat.get("__allow_len__", set())
-    return cls(uop, arg, vin, name, dtype, allow_len)
-
 def _match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> bool:
   if pat.name in store and store[pat.name] != uop: return False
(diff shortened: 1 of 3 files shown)


geohot:
Excellent use of process replay. @Qazalin I'm assuming by the long time CI took that it passed? Why is it so slow, shouldn't it just compare a hash?

tinygrad/codegen/uops.py
```diff
@@ -77,30 +77,19 @@ class UPat:
...
-  if isinstance(pat.arg, set) and uop.arg not in pat.arg: return False
-  elif pat.arg is not None and uop.arg != pat.arg: return False
-  if isinstance(pat.dtype, set) and uop.dtype not in pat.dtype: return False
-  if isinstance(pat.dtype, DType) and uop.dtype != pat.dtype: return False
-  if isinstance(pat.uop, set) and uop.uop not in pat.uop: return False
-  if isinstance(pat.uop, UOps) and uop.uop != pat.uop: return False
+  if pat.name is not None:
+    if pat.name in store and store[pat.name] != uop: return False
+    store[pat.name] = uop
+  if pat.arg is not None and uop.arg != pat.arg: return False
```
geohot, line 84: Err, arg can be a set now after the PTX shift addition? Why was this not caught by process replay?

Qazalin:
@geohot yes they passed! https://github.com/tinygrad/tinygrad/actions/runs/9357955542/job/25758850270
I'm pushing to update_benchmark to do the same.

geohot:
I don't understand how the PTX one passed...from what I read it no longer correctly handles "arg" as a set.

Qazalin:
process replay runs ASSERT_COMPILE=1 against a cache built up from master, maybe no kernels changed for the limited tests we run on ptx CI? https://github.com/tinygrad/tinygrad/blob/master/.github/workflows/test.yml#L487

tinygrad/codegen/uops.py
alecchendev: Oops, must've accidentally skipped over this when rebasing, will also add a regression test for this

alecchendev: fwiw the previous implementation did not pass the regression test i just added

alecchendev:
IIUC it looks like the logic added to match an arg set in #4728 would never match the positive case where the arg is in the set, so the patterns added in that PR never got used. With the latest commit here this is fixed, and now the ptx test is failing

Qazalin:
@alecchendev hmm, can you prove this with a separate PR for a regression test?

geohot:
@SzymonOzog Can you add a regression test confirming shift is actually selected on master?

alecchendev:
> @alecchendev hmm, can you prove this with a separate PR for a regression test? maybe by adapting test_arg_set to the master PatternMatcher

#4827

geohot:
Congrats, assuming this passes master CI, bounty is yours!

Can pay via PayPal or USDC, e-mail george@tinygrad.org to claim

Issue #4811, opened 2024-06-03
Too much shared data in ptxas compile


satin-spirit:
I am not sure what exactly triggers this, but when using `BEAM=2` I get an error like this after ~15-30 minutes of beam search.
```
tinygrad.device.CompileError: compile failed: ptxas error   : Entry function 'r_64_64_2_4_32_48_4_2_4_2' uses too much shared data (0x10000 bytes, 0xc000 max)
```
I'm not incredibly familiar with the guts of tinygrad. Should this error just be caught and ignored? If so, I can probably make a PR for that.

geohot:
Yea, BEAM should already be ignoring that...not sure why it crashed on this

chenyuxyz:
I made BEAM not ignoring our `CompileError`, as this means tinygrad is generating invalid codes. Ignoring compile error hid some legit bugs before.

satin-spirit:
It does run just fine without BEAM. Though, it's a little concerning to me that tinygrad can output invalid kernels in the first place. For example, what if this error happens when BEAM is disabled, just by chance? Then the user wouldn't have any real way to continue. As far as I can tell, kernels are generated deterministically w/out BEAM. Seems like a bug worth fixing.

satin-spirit:
Closing this as the original issue was fixed and works great now. Thank you!

PR #4822, opened 2024-06-03
default skip all exception in beam

chenyuxyz:
added a flag `BEAM_STRICT_MODE` to catch compile error or other exceptions on demand


chenyuxyz:
addressed #4811

PR #4829, opened 2024-06-04
Regression tests for bitshift

SzymonOzog:
Disabled right now because it will be fixed by  #4821


Qazalin (commented):
Can you also add a test for when bitshift isn't triggered? (arg not in set)

test/test_uops.py
```diff
@@ -245,5 +245,29 @@ def test_local_indirect(self):
...
+    uops.add(UOps.SINK, None, (a1,))
+    Device[Device.DEFAULT].renderer.render("test", uops)
+    self.assertEqual(uops.uops[-1].arg, BinaryOps.SHL)
+
+  @unittest.skip("Broken")
+  def test_bitshift_right(self):
+    uops = UOpGraph()
+    g1 = uops.add(UOps.DEFINE_GLOBAL, PtrDType(dtypes.int32), (), (0, True))
+    c1 = uops.add(UOps.CONST, dtypes.int, (), 2)
+    l1 = uops.add(UOps.LOAD, dtypes.int, (g1, c1), 2)
```
Qazalin, line 266: err, LOAD doesn't have arg

SzymonOzog: Updated, my bad - a leftover from copying

SzymonOzog:
> Can you also add a test for when bitshift isn't triggered? (arg not in set)

Added

geohot:
#4821 was merged

SzymonOzog:
> #4821 was merged

Enabled tests

geohot:
I'm fine with merging this as is, but I'm very supportive of end to end tests that apply the pattern matcher rules and confirm the output is the same. This could be fuzzed as well.

SzymonOzog:
@Qazalin There are tests that specifically check the correctness of the calulations https://github.com/tinygrad/tinygrad/blob/5ac30c29d88fe4ae2240a9f1ac4b7b9c642a8427/test/test_uops.py#L125 

In terms of test_ops introducing an error in shift left makes 107 tests fail and an error in shift right makes 22 tests fail. 

Do you feel like there are some more explicit test needed?

Qazalin:
Those tests don't rely on the pattern matcher, though I could verify that if there are errors in the pattern matcher, some test_ops fail. I think this is sufficient, thanks!

PR #4827, opened 2024-06-04
Add arg set regression test for pattern matcher

alecchendev:
Logic to recognize arg sets in the pattern matcher was added in #4728. It would correctly reject uops if the arg was not in the set, but it would also fail to match on uops with args in the set. This regression test demonstrates this.

https://github.com/tinygrad/tinygrad/pull/4821#issuecomment-2147112516

PR #4830, opened 2024-06-04
Fix arg set in pattern matcher

alecchendev:
Follow up on #4827. Not sure if this will actually pass CI all the way if there were things that need to change now that the new patterns will be recognized (edit: NVM!)


Qazalin (commented):
cool, can you nest these ifs for dtype and uop as well?

PR #4831, opened 2024-06-04
Approximations for EXP2, LOG2, SIN and POW

samayala22:
WIP

3 ULPs max error approximations (adapted from SLEEF library)

Also Added 4 more bitwise ops (SHL, SHR, AND, OR) for fast and correct bit manipulations.

Since pow(x,y) = exp(y * log(x) is not a valid identity per the C99 standard, I figured I should also implement a proper POW approximation. This also lets us remove the sketchy NaN injections currently used and should also fix the backward pass for abnormal numbers. 

Implemented:

- [X] F32 EXP2
- [X] F64 EXP2
- [X] F32 POW
- [ ] F64 POW
- [ ] F32 LOG2
- [ ] F64 LOG2
- [ ] F32 SIN
- [ ] F64 SIN


samayala22:
I'm busy with other stuff.

PR #4834, opened 2024-06-04
Nest ifs for dtype and uop in pattern matcher

alecchendev:
Makes things consistent with arg, also this addresses most of the performance regression in #4791.

Average over 10 runs on an m1 macbook air:

master
```
forward 29.906 ms
schedule 9.563 ms
lower 1571.959 ms ​
```

this branch:
```
forward 29.568 ms
schedule 9.413 ms
lower 1479.284 ms ​
```


Qazalin (commented):
can you enable process replay?

Qazalin:
confirmed this is faster than master.

PR #4835, opened 2024-06-04
Optional half matmul

Eliulm:


examples/mlperf/initializers.py
```diff
@@ -41,7 +41,8 @@ def __init__(self, in_features, out_features, bias=True, std=0.02):
     self.bias = Tensor.zeros(out_features, dtype=dtypes.float32) if bias else None
   
   def __call__(self, x:Tensor):
-    return x.linear(self.weight.cast(dtypes.default_float).transpose(), self.bias.cast(dtypes.default_float) if self.bias is not None else None)
+    x = x.contiguous().cast(dtypes.half if getenv("HALF_LINEAR", 0) else dtypes.default_float)
```
chenyuxyz, line 44: can you make a variable for `dtypes.half if getenv("HALF_LINEAR", 0) else dtypes.default_float`? i think that reads better and less repeated code

examples/mlperf/model_train.py
```diff
@@ -430,6 +430,7 @@ def train_bert():
   target, achieved                                      = getenv("TARGET", 0.72), False
 
   config["DEFAULT_FLOAT"] = dtypes.default_float.name
+  config["HALF_LINEAR"]   = getenv("HALF_LINEAR", 0)
```
chenyuxyz, line 433: would be nice to also log dropout so it's easy to tell if dropout is used or not for now, not related to this pr though

examples/mlperf/initializers.py
```diff
@@ -41,7 +41,8 @@ def __init__(self, in_features, out_features, bias=True, std=0.02):
     self.bias = Tensor.zeros(out_features, dtype=dtypes.float32) if bias else None
   
   def __call__(self, x:Tensor):
-    return x.linear(self.weight.cast(dtypes.default_float).transpose(), self.bias.cast(dtypes.default_float) if self.bias is not None else None)
+    matmul_dtype = dtypes.half if getenv("HALF_LINEAR", 0) else dtypes.default_float
+    return x.contiguous().cast(matmul_dtype).linear(self.weight.contiguous().cast(matmul_dtype).transpose(), self.bias.cast(dtypes.default_float) if self.bias is not None else None)
```
chenyuxyz, line 45: do you need these `contiguous`? weight should be contiguous and is there a benefit to do this with `x`?

chenyuxyz (commented):
have you benchmarked default float32 with / without this cast? curious to know the speed difference

examples/mlperf/initializers.py
Eliulm: currently, uop count explodes when I don't do that. Seems like a bug. I think Qazalin is on that already

chenyuxyz: if it's a temporary issue, can you add a TODO note for some context

Qazalin: Yeah I'm looking into the root cause for this

Eliulm:
> have you benchmarked default float32 with / without this cast? curious to know the speed difference

So on my machine, full float is 10.8 TFLOPS, float with half linear is 19TFLOPS and full half is 20.8TFLOPS

chenyuxyz:
Cool merging! Interesting that changing attention matmul caused the nan

Issue #4752, opened 2024-05-28
[DRAFT PROPOSAL] Linearizer Refactor


flammit:
This is preliminary and will be updated or discarded as comments come in.

## Current Issues

1) To maximize performance, we need to be able to utilize shared memory buffers more generically as temporary inputs and outputs.

    * Serial reduce fusion: Needed to store the intermediate outputs in the fused kernel (as in 0xTimmy/qazalin's work on mean and std).

    * GEMM: Instead of loading directly to the register file, pre-fetch blocks of A and B inputs into shared memory to hide device memory latency.  Necessary to get AMD tensor cores over 100TF and to rocblas speed, likely needed to get NV to pytorch speed as well.

    * Flash attention: Load blocks of K, V into shared memory, loop to block reduce Q, etc.

    * Fused winograd conv: winograd-transformed input and filter blocks are calculated and stored shared memory, transformed outputs are stored to shared memory and un-winograded to device memory

2) Linearizer.render_reduceop has extra complexity to support OptOps.GROUP.

    * Having OptOps.GROUP applied this way means it only makes sense in single reduce kernels.

    * Custom UOps-generating logic exists in Linearizer to make the LocalBuffer, store the results, barrier, and an additional compute block to finish the reduce.

3) Both Kernel and Linearizer contains extra complexity to support OptOps.TC.

    * The thread structure imposed by using TCs may conflict with other fused operations.

    * The local_alias/alias_buffer framework is being abused to generate the element allocations to each tensor core thread.

    * The logic for the threading structure and applicability of tensor cores is split over `kernel.py` and `linearizer.py` making it hard to reason about.

    * Tensor cores should be easily testable by swapping out calls to the hardware UOps.WMMA with a 32 thread implementation that does the same M, N, K GEMM (even a naive single-threaded version).

## Proposal Overview: Add two additional graph hierarchies

Currently the hierarchy looks like:

`Lazy Graph -> Kernel Graph -> UOps Graph`

We propose to add two additional layers, `Block` and `Thread`, with `Block` taking the place of the current `Kernel` layer and `Kernel` now representing a generalized fused op.

`Lazy Graph -> Kernel Graph -> Block Graph -> Thread Graph -> UOps Graph`

This structure is analogous to the one proposed in the Mirage paper: [Wu, et al.: A Multi-Level Superoptimizer for Tensor Programs. arXiv cs.LG. 9 May 2024](https://arxiv.org/abs/2405.05751).

[...]

PR #4839, opened 2024-06-05
refactor `linearize` to render_block, P1

Qazalin:
To start the Linearizer refactor, we can unify uoping elementwise ops and reduceops into a single `render_block` function,  (partially) following the design described in #4752. Partially because render_block uops early and late reduce.
a follow-up diff will separate these into 2 blocks.


geohot:
LGTM, this good to merge?

PR #4840, opened 2024-06-05
Replace tqdm

VicKnightDR:
progress bar behaving same in terms of output data. 

still need to change variables to make it visually look identical, and change the import statements inside most of the files.

Note: this can replace the progress bar and the trange function. 
Don't think we can do much more (e.g tqdm.write) within 5 lines, but replacing those other commands shouldn't be an issue.

Test script:

```from tqdm import tqdm
import random
import time

def progress_bar(iterable, length=50, fill='█'):
    start = time.time(); print('{0}%|{1} 0/{2} 00:00<?,?it/s'.format(0, ' ' * length, len(iterable)), end='\r')
    for i, item in enumerate(iterable):
        yield item; elapsed, percent, bar = time.time()-start, round(100*((i+1)/float(len(iterable))),1), fill * int(length*(i+1)//len(iterable)) + ' ' * (length-int(length*(i+1)//len(iterable)))
        print("{0}%|{1}|{2}/{3}[{4}<{5},{6}]".format(percent,bar,i+1,len(iterable), time.strftime("%H:%M:%S", time.gmtime(elapsed)),time.strftime("%H:%M:%S", time.gmtime(elapsed*(1-1/(percent/100)))),str(round((i+1)/elapsed,2)) + 'it/s' if (i+1)/elapsed >= 0.1 else str(round(elapsed/(i+1),2))+'s/it'), end='\n' if i+1==len(iterable) else '\r')

def test():
   random_sleep = [random.random() for i in range(10)]
   for i in progress_bar(range(10)):
      time.sleep(random_sleep[i])
   for i in tqdm(range(10)):
      time.sleep(random_sleep[i])
      
test()

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -15,10 +15,10 @@ def prod(x:Iterable[T]) -> Union[T,int]: return functools.reduce(operator.mul, x
 OSX = platform.system() == "Darwin"
 CI = os.getenv("CI", "") != ""
 def progress_bar(iterable, length=50, fill='█'):
-    start = time(); print(f'{0}% | {'-'*length} 0/{len(iterable)} 00:00<?,?it/s', end='\r')
+    start = time(); print(f'{0}%|{' '*length} 0/{len(iterable)} 00:00<?,?it/s', end='\r')
     for i, item in enumerate(iterable):
-        yield item; elapsed, percent, bar = time()-start, round(100*((i+1)/float(len(iterable))),1),fill*int(length * (i+1) // len(iterable)) + '-' * (length - int(length * (i+1) // len(iterable)))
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tinygrad/helpers.py [hunk 1]
```diff
@@ -15,7 +15,11 @@ def prod(x:Iterable[T]) -> Union[T,int]: return functools.reduce(operator.mul, x
 # NOTE: helpers is not allowed to import from anything else in tinygrad
 OSX = platform.system() == "Darwin"
 CI = os.getenv("CI", "") != ""
-
+def progress_bar(iterable, length=50, fill='█'):
+    start = time.time(); print('{0}%|{1} 0/{2} 00:00<?,?it/s'.format(0, ' ' * length, len(iterable)), end='\r')
```
chenyuxyz, line 19: don't use `;` to combine lines, and indent 2

also don't update files in extra (those are not tested, and having tqdm there is fine) and add tests

geohot:
This is not close to enough to lock a bounty. Don't bother changing the stuff in extra, it's only the core library change that's required.

geohot:
Locked the bounty for #4846. Most notably, this PR didn't change the two places tqdm was actually used in the main library, just a few meaningless things in extra.

PR #4841, opened 2024-06-05
add mlperf train subset of openimages

nikvaessen:
This PR adds the train subset of openimages for mlperf, required to train RetinateNet.


chenyuxyz:
looks good. tested model_eval still runs. eval is very slow though (> 1 sec per step of bs=8 on 4090), so figure out how to make eval faster would help training if you are interested

PR #4844, opened 2024-06-05
support symbolic reshape with non-contiguous

chenyuxyz:
pre-requisite for symbolic arange (make symbolic ones that can be folded).


chenyuxyz:
will follow up with easier way to get the int shape from symbolic shape, might reduce some `isinstance(s, int)`

PR #4846, opened 2024-06-05
tqdm replacement

SnakeOnex:
Here is my attempt for the tqdm replacement.
My implementation is **11 lines of code**, which is > 5.
I think 5 lines of code is possible  by sacrificing functionality. (or with 350 chars per line)
Below I give overview of which features are missing compared to the og tqdm and which features make this be ~12 lines instead of 5. 

I would appreciate feedback:
  1. How many lines is okay, is it really just 5?
  2. Can i do ~20 lines, make it a class iterator, with set desctiption and unit support?

Features:
- [x] size adjustment based on terminal width
- [x] high iters/sec speed (naive implementation can slow down the looping coz stdout write is slow, here the bar update gets unjusted based on the average iter length)
- [x] ~matching tqdm char by char (this one is tricky, since tqdm has fancy stuff like half bar, 1/10th bar characters to increase precision, this would add to the complexity / lines, but for the final bar look and for most of the iterations this implementation exactly matches tqdm as per the tests)
- [x] set description support
- [x] replaced in state.py (works, but had to remove the description about used RAM)
- [x] replaced in helpers.py (works, but had to remove the units="B", unit_scale=True)

Performance benchmark:
```
    iters = 100000
    st = time.perf_counter()
    for i in tinytqdm(range(iters)): pass
    tinytqdm_time = time.perf_counter()-st

    st = time.perf_counter()
    for i in tqdm(range(iters)): pass
    tqdm_time = time.perf_counter()-st

    st = time.perf_counter()
    for i in range(iters): pass
    for_time = time.perf_counter()-st

    print(f'TinyTQDM: {tinytqdm_time:.4f}s')
    print(f'TQDM: {tqdm_time:.4f}s')
    print(f'For loop: {for_time:.4f}s')
```
Output:
```
100%|█████████████████████████████████████████████████████████████████████████████| 100000/100000 [00:00<00:00, 6562468.66it/s]
100%|█████████████████████████████████████████████████████████████████████████████| 100000/100000 [00:00<00:00, 5337622.80it/s]
TinyTQDM: 0.0153s
TQDM: 0.0210s
For loop: 0.0021s
```
When updating bar every iter:
```
TinyTQDM: 4.0481s
TQDM: 0.0202s
For loop: 0.0021s
```


tinygrad/nn/state.py
```diff
@@ -119,8 +118,7 @@ def __init__(self):
     model_state_dict = get_state_dict(model)
     if DEBUG >= 1 and len(state_dict) > len(model_state_dict):
       print("WARNING: unused weights in state_dict", sorted(list(state_dict.keys() - model_state_dict.keys())))
-    for k,v in (t := tqdm(model_state_dict.items(), disable=CI or not verbose)):
-      t.set_description(f"ram used: {GlobalCounters.mem_used/1e9:5.2f} GB, {k:50s}")
```
geohot, line 123: This changes behavior.

tinygrad/nn/state.py
```diff
@@ -1,9 +1,8 @@
 import os, json, pathlib, zipfile, pickle, tarfile, struct
-from tqdm import tqdm
 from typing import Dict, Union, List, Optional, Any, Tuple
 from tinygrad.tensor import Tensor
 from tinygrad.dtype import dtypes
-from tinygrad.helpers import prod, argsort, DEBUG, Timing, CI, unwrap, GlobalCounters
+from tinygrad.helpers import prod, argsort, DEBUG, Timing, CI, unwrap, GlobalCounters, tinytqdm as tqdm
```
geohot, line 5: Don't import as

geohot:
A few more lines than 5 is okay if it matches output perfectly.

tinygrad/helpers.py
```diff
@@ -239,3 +238,15 @@ class CStruct(ctypes.Structure):
   return CStruct
 def init_c_var(ctypes_var, creat_cb): return (creat_cb(ctypes_var), ctypes_var)[1]
 def flat_mv(mv:memoryview): return mv if len(mv) == 0 else mv.cast("B", shape=(mv.nbytes,))
+
+def tinytqdm(iterable, desc='', disable=False, total=None, update_rate=100, char='█'):
```
geohot, line 242: Can you add a type signature for this?

obadakhalili:
I like the task, so I tried to help by providing a solution to support `.set_description` in #4851. In the PR, I also suggest other solutions, like using `generator.send` and implementing `__iter__`/`__next__` protocol.

I tried to open the PR to your branch @SnakeOnex, but it doesn't seem to be possible because it comes from a fork. If the PR is welcome, I can open one in your fork.

SnakeOnex:
Yes, i thought of that, but it goes quite a bit against the minimal line requirement. I tried implementing exactly earlier today, a class with __iter__, that has .update() and .set_description() methods, i got it down to ~20 lines (it wasn't respecting the 150 characters per line limit tho). I will commit it soon. Feel free to msg me on discord if you want to discuss / colaborate more.

SnakeOnex:
I tried to make the solution match tqdm as much as possible, particularly:
- support for both passing iterable & making separate bar instance
- handling `break`  (making sure when user breaks, the bar is updated to the last state & newline is printed)
    - there is also an edge case, when if not done properly, two progress bars right will overwrite each other
- exactly matching the character outputs to tqdm (checked in "e2e" tests, where bars are compared in each iteration, character per character)
    - bar can differ by 1 character due to tqdm's special chars (half blocks, etc)
- comparative in performance for high iter/s (due to dynamic update skips)

After trying many different iterations (using ANSI special characters to move cursor to not have to handle the `break` edge case etc) I arrived at this implementation which has **20 lines**, which is way over the limit of 10 lines, I would argue it is actually a usable & reliable tqdm replacement, not just a hack.

I tried to make the replacement as general as possible, but it is used only **twice in tinygrad**.  It would be possible to cut the lines down by removing features / edge case handling (assuming loop is never broken, no two bars in a row, no high it/s performance, etc). It would be a stepdown from tqdm tho.

tinygrad/helpers.py
```diff
@@ -196,15 +195,14 @@ def wrapper(*args, **kwargs) -> bytes:
...
-
 def fetch(url:str, name:Optional[Union[pathlib.Path, str]]=None, allow_caching=not getenv("DISABLE_HTTP_CACHE")) -> pathlib.Path:
   if url.startswith(("/", ".")): return pathlib.Path(url)
   fp = pathlib.Path(name) if name is not None and (isinstance(name, pathlib.Path) or '/' in name) else pathlib.Path(_cache_dir) / "tinygrad" / "downloads" / (name if name else hashlib.md5(url.encode('utf-8')).hexdigest())  # noqa: E501
   if not fp.is_file() or not allow_caching:
     with urllib.request.urlopen(url, timeout=10) as r:
       assert r.status == 200
       total_length = int(r.headers.get('content-length', 0))
-      progress_bar = tqdm(total=total_length, unit='B', unit_scale=True, desc=url)
+      progress_bar = tinytqdm(total=total_length, unit='B', desc=f"{url}: ", disable=False)
```
geohot, line 205: Does this support unit_scale?

SnakeOnex: it doesn't, unit_scale is the only thing missing from being an exact replacement with pretty much same API (for the tinygrad usecases)
I didn't add it since my solution was already quite a bit over the bounty line count

SnakeOnex:
Added unit_scale, this matches tqdm exactly char to char. Adding unit_scale upped the lines to 24 which is 22 net, given the two import removals.

Here is an example of fetch (tqdm vs tinytqdm):
demo

SnakeOnex:
did some extensive testing of the unit_scale bar char to char, it matches basically always, but tqdm sometimes does a bit strange rounding (f"9.995:.2f" is 10.00 in python, but tqdm makes it 9.99), but other than that, we are char to char matching. i removed the unit_scale test, it was flaky due to these rounding diffs

geohot:
Congrats, bounty is yours! george@tinygrad.org to claim, can pay via PayPal or USDC

PR #4501, opened 2024-05-10
Should this symbolic test fail?

reddyn12:

PR #4848, opened 2024-06-05
fix reshape size check from symbolic shape to int shape

chenyuxyz:
Added `.val` for all sint Nodes and use that to get int_size. Also reshape_and_permute shares the same code path, but the Variables are no longer bound so passed the check in that case

fix #4501 too as we now support reshape into MulNode and SumNode provided that the int_size are the same


chenyuxyz:
hmm in jitted function, if the inputs are symbolized before passing into the function, the function output shape can use the old variable value.

geohot:
Stale, closing

PR #4854, opened 2024-06-07
[1/n] Begin refactoring multi_merge_dim into shapetracker

kvkenyon:
Decoupling the multi merge dims from kernel. Should be able to use _merge_dims instead of having two different expressions for when dimensions can merge.


chenyuxyz:
closing this because you removed the TODO to add test, without adding the test. In general we don't want the pr that only moves things around without a good reason.

If you want to refactor this, I'd refactor the reshape and _merge_dim first, and reuse that in multi merge

kvkenyon:
oops meant to be a draft still working on it

PR #4858, opened 2024-06-07
faster codegen process replay

Qazalin:
This diff rewrites process replay to:

1. Save every instance of a Linearizer created in HEAD
2. Recreate that Linearizer in master with ast, applied_opts, opts
3. Assert Program.src is the same.

This allows process replay to take max 2 minutes instead of 16-19 minutes per backend.

The drawback of ASSERT_COMPILE=1 is that even if only the Linearizer.name changes it can fail, making it fragile and slow (tests can't run with -n=auto).

### DIFF
--- a/test/external/replay_codegen.py
+++ b/test/external/replay_codegen.py
@@ -18,11 +18,12 @@ for offset in tqdm(range(0, row_count, page_size)):
     good_uops = k.linearize().uops
     good_src = k.opts.render("test", good_uops)
     try: assert compare_src == good_src
-    except AssertionError as e:
+    except AssertionError:
       print("PROCESS REPLAY FAILED")
       print(compare_k.ast)
       print(compare_k.applied_opts)
       diff = list(difflib.unified_diff(good_src.splitlines(), compare_src.splitlines()))
       for line in diff:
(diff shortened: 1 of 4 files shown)


test/external/replay_codegen.py
```diff
@@ -0,0 +1,27 @@
...
+    good_uops = k.linearize().uops
+    good_src = k.opts.render("test", good_uops)
+    try: assert compare_src == good_src
+    except AssertionError:
+      print("PROCESS REPLAY FAILED")
+      print(compare_k.ast)
+      print(compare_k.applied_opts)
+      diff = list(difflib.unified_diff(good_src.splitlines(), compare_src.splitlines()))
+      for line in diff:
+        print(colored(line, "red" if line.startswith("-") else "green" if line.startswith("+") else None))
```
Qazalin, line 27: not raising yet because this issue still stands https://github.com/tinygrad/tinygrad/issues/4860

Qazalin:
some symbolic ASTs fail with [run_process_replay] (4860)
https://github.com/tinygrad/tinygrad/actions/runs/9416967155/job/25941191234#step:23:82
Merging because this tool has signal value for the Linearizer refactor diffs.

PR #4867, opened 2024-06-07
refactor: simplify _match

roelofvandijk:
The checks are the same for `arg`, `dtype` and `uop`.

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -80,22 +80,13 @@ class UPat:
 def _match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> bool:
   if pat.name in store and store[pat.name] != uop: return False
   if pat.name is not None: store[pat.name] = uop
-  if pat.arg is not None:
-    if isinstance(pat.arg, set):
-      if uop.arg not in pat.arg: return False
-    elif uop.arg != pat.arg: return False
-  if pat.dtype is not None:
-    if isinstance(pat.dtype, set):
-      if uop.dtype not in pat.dtype: return False
-    elif uop.dtype != pat.dtype: return False
-  if pat.uop is not None:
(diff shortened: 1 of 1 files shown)


roelofvandijk:
This class would be simpler if these three would have signatures like `Optional[Set[DType]]`, set with 1 element if only one.

Qazalin:
can you do a [run_process_replay] in your head commit to check kernels haven't changed?

tinygrad/codegen/uops.py
```diff
@@ -80,22 +80,13 @@ class UPat:
...
-  if pat.dtype is not None:
-    if isinstance(pat.dtype, set):
-      if uop.dtype not in pat.dtype: return False
-    elif uop.dtype != pat.dtype: return False
-  if pat.uop is not None:
-    if isinstance(pat.uop, set):
-      if uop.uop not in pat.uop: return False
-    elif uop.uop != pat.uop: return False
+  for p, u in [(pat.arg,uop.arg), (pat.dtype, uop.dtype), (pat.uop, uop.uop)]:
+      if p is not None:
```
geohot, line 84: Wrong spacing. Also, have you benchmarked this?

roelofvandijk: 1) How did neither `pre-commit` nor the CI linter error on this?
2) Yes, I benchmarked it with ` JIT=0 python3 -O examples/llama.py --gen 1 --prompt "Hello." --count 10 --temperature 0 --timing`. No discernible speed differences. There are some performance optimizations possible here, but I did not apply them to keep the logic identical.

roelofvandijk: Ah, `pylint` is not checking this folder because there is no `__init__.py`.

roelofvandijk:
Replaced by https://github.com/tinygrad/tinygrad/pull/4876

PR #4868, opened 2024-06-07
Gated Stores Refactor

0xtimmy:
This refactor removes UOps.IF and shifts that responsibility to the `UOp.vin[3]` of `STORE`s.
in `uops.py` it goes over the graph to link uops with a `STORE` (and the `STORE`'s gate). Then it tries to group arithmetic with the store in a way that's safe in regards to scope.

draft; just wanted to show the direction


Qazalin:
oh I see what you're trying to do,

You can think about the UOps.IF -> gated store refactor more like generalizing https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/linearizer.py#L277 to all ALUs (including indexing) for gated stores https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/linearizer.py#L168.

linearizer.py manually inserts UOps.IF,
Instead, we wanna make the linearizer only care about gating the store, then in the UOpGraph, this gated store is rewritten to IF -> indexing ALUs + whatever computation for the store val -> STORE -> ENDIF

Issue #4857, opened 2024-06-07
[PROPOSAL] Replace Symboilc with UOps


geohot:
There's no difference between the matching rules in symbolic and the UPat rules. These need to be unified. Everywhere `Node` is used, it should be `UOp` instead. They are the same thing.

It's very nice to have the UOps as a graph so we don't have to worry about creating extras. I believe if deduping is a first pass step of UOp Linearization, we don't have to worry about tracking any global context.

The `Linearizer` calls `expr_idxs` in the `ShapeTracker` with a list of Nodes. This should be a list of `UOps`. But it goes deeper, as the nodes shouldn't be created in the first place. A loop is represented as a RANGE UOp only.

Task list:
- [x] Convert DIV to UnaryOps.RECIP and BinaryOps.IDIV
- [ ] Convert linearizer to directly use UOps
- [x] Add min/max tracking to UOps
- [ ] Support >2 inputs to ADD, MUL, MAX, etc... (optional, but like current symbolic and easier to pattern match on)
- [ ] Move float4 folding to a UPat and out of the Linearizer. Will be a fancy pattern match rule, but I think doable.
- [ ] Add an integer type with unlimited precision to be specialized. This fixes the 64-bit indexing issue.
- [ ] Move the symbolic rules to UPat and get the symbolic tests passing on UOps.

In a more abstract sense, the `Linearizer` is only the UOp linearizer. The lowering from `LazyOp` to `UOps` is now completely a graph operation, perhaps to be called lowering.

Questions:
- UOps have a SINK for multiple stores and LazyOp has a tuple. The tuple for LazyOp is nice since it matches the buffer count, but why is this different.
- Are UOps and LazyOps the same thing? Should the scheduler just output UOps?

geohot:
A simpler checklist to approach this:

- [ ] Move `render_ops` out of Linearizer to `st_to_uops`
- [ ] Remove `InterpretedFlopCounter`

geohot:
This is mostly done!

PR #4870, opened 2024-06-08
test uop as symbolic

geohot:


geohot:
Working toward #4857, should be good to merge except for line count.

PR #4874, opened 2024-06-08
bug: this should be a comparison to None

roelofvandijk:
I'm pretty sure this is a (minor) bug. This `.get` functions as an existence check here (default return is `None`). By checking its falseness instead of comparing to `None`, this will not return on values existing in `self.nodes` that are `False` such as `0`.

And equivalent and probably clearer (but potentially slower) statement here would be:

`if (key:=up.tuple()) in self.nodes: return self.nodes[key]`


Qazalin:
`self.nodes: Dict[Tuple, UOp]` so it's always either `UOp|None`

roelofvandijk:
`dict.get` default return on missing key is `None`
https://docs.python.org/3/library/stdtypes.html#dict.get

PR #4875, opened 2024-06-08
fix: pylint will not lint folders without __init__.py

roelofvandijk:
This folder was missing an `__init__.py` file, so the contents were not linted. Same problem for the folders in `runtime`.

This PR adds the file and fixes the warnings. I am not happy with the ignore on the `g_valid` issue.

Here are the warnings after adding the `__init__.py` file:

```
pylint...................................................................Failed
- hook id: pylint
- exit code: 14

************* Module tinygrad.codegen.uops
tinygrad/codegen/uops.py:277:8: R1705: Unnecessary "else" after "return", remove the "else" and de-indent the code inside it (no-else-return)
************* Module tinygrad.codegen.linearizer
tinygrad/codegen/linearizer.py:92:68: E0606: Possibly using variable 'g_valid' before assignment (possibly-used-before-assignment)
tinygrad/codegen/linearizer.py:245:10: W0622: Redefining built-in 'iter' (redefined-builtin)
tinygrad/codegen/linearizer.py:236:17: W0622: Redefining built-in 'next' (redefined-builtin)
tinygrad/codegen/linearizer.py:237:25: W0612: Unused variable 'reduce' (unused-variable)
```

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -233,17 +233,17 @@ class Linearizer(Kernel):
       # run tensor cores AST
       wmma_sz = [prod(l) for l in tc.thread_local_sizes]
       def upcast_strides(buf:int):
-        strides, next = [], 1
+        strides, next_ = [], 1
         for (sz, stride, reduce) in self.upcasted_axis(buf)[tc.num_upcasts():]:
-          strides.append((0 if stride == 0 else next, sz))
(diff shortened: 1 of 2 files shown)


geohot:
Good find

roelofvandijk:
Related: warnings for `tinygrad/runtime`: https://github.com/tinygrad/tinygrad/issues/4877

PR #4873, opened 2024-06-08
unbind Variable pre LazyOp

Qazalin:
Tensor variable break process replay because they create ASTs with bound Variables. https://github.com/tinygrad/tinygrad/issues/4860

This diff unbinds the Variable before LazyOping.

before: `test/test_symbolic_jit.py TestSymbolicJit.test_var`
```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(1, 1), strides=(0, 0), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ DIV 
  2    ┣━┳ SUM (0, 1)
  3    ┃ ┗━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(Variable('i', 1, 10), 3), strides=(3, 1), offset=0, mask=None, contiguous=True),)))
  4    ┗━┳ CAST dtypes.float
  5      ┗━┳ MUL 
  6        ┣━━ CONST ConstBuffer(val=Variable('i', 1, 10).bind(1), dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(1, 1), strides=(0, 0), offset=0, mask=None, contiguous=True),)))
  7        ┗━━ CONST ConstBuffer(val=3, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(1, 1), strides=(0, 0), offset=0, mask=None, contiguous=True),)))
```
after: 

```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(1, 1), strides=(0, 0), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ DIV 
  2    ┣━┳ SUM (0, 1)
  3    ┃ ┗━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(Variable('i', 1, 10), 3), strides=(3, 1), offset=0, mask=None, contiguous=True),)))
  4    ┗━┳ CAST dtypes.float
  5      ┗━┳ MUL 
  6        ┣━━ CONST ConstBuffer(val=Variable('i', 1, 10), dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(1, 1), strides=(0, 0), offset=0, mask=None, contiguous=True),)))
[...]

PR #4876, opened 2024-06-08
[run_process_replay] faster and simpler match function

geohot:
Similar to #4867 but improves speed.


geohot:
@Qazalin Hmm, I got the checkmark, but it looks like process replay failed.

Qazalin:
yeah because of this https://github.com/tinygrad/tinygrad/pull/4873

roelofvandijk:
Move the `dtype` check to the back, it's statistically uncommon/

geohot:
Didn't see a speedup moving dtype

PR #4882, opened 2024-06-08
perf: faster element deletion during matching

roelofvandijk:
`hasattr` is slow.

M3 Macbook
`PROFILE=0 python test/external/external_benchmark_schedule.py`
Before:
```
***** model forward in  16.69 ms
***** model schedule in   5.56 ms
***** model lower in 892.72 ms
```

After:
```
***** model forward in  16.70 ms
***** model schedule in   5.57 ms
***** model lower in 745.73 ms
```

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -268,9 +268,11 @@ class UOpGraph:
           recurse_cnt += 1
         changed += recurse_cnt
         # NOTE: this changes UOp, so we have to delete caches
-        up.vin = tuple(rewrite(x) for x in up.vin)
-        if hasattr(up, "parents"): del up.parents
-        if hasattr(up, "cmp_tuple"): del up.cmp_tuple
+        up.vin = tuple([rewrite(x) for x in up.vin])
+        try: del up.parents
+        except AttributeError: pass
+        try: del up.cmp_tuple
+        except AttributeError: pass
         # replace with cached nodes
         if found:=self.nodes.get(key:=up.tuple()): return found
         self.nodes[key] = up


geohot:
What am I going to complain about about this PR?

tinygrad/codegen/uops.py
```diff
@@ -268,9 +268,11 @@ def rewrite(u:UOp) -> UOp:
           recurse_cnt += 1
         changed += recurse_cnt
         # NOTE: this changes UOp, so we have to delete caches
-        up.vin = tuple(rewrite(x) for x in up.vin)
-        if hasattr(up, "parents"): del up.parents
-        if hasattr(up, "cmp_tuple"): del up.cmp_tuple
+        up.vin = tuple([rewrite(x) for x in up.vin])
```
geohot, line 271: Answer btw is this line. Why is this changed? Is this faster? In what versions of Python is it faster.

roelofvandijk: In all 3.X versions AFAIK.

* because the list is fully consumed here, there is no benefit of early exit that a generator would offer
* generators can be more memory efficient in such cases, but incur a performance overhead

This matters at scale, let me make a different PR.
Anyway: should not have included that change here - apologies.

roelofvandijk:
Took out the LC into tuple.

tinygrad/codegen/uops.py
roelofvandijk: For builtins like `sum`, `max`, `min` a generator comprehension can be more memory efficient because a list comprehension would construct the whole output, while the builtins can process a generator comprehension one-by-one.

Those do comsume the generator completely as well, however, so LC can be faster in those cases as well.

In reality, for the small objects that are used here, I doubt it matters.

PR #4884, opened 2024-06-08
process replay in all of CI

Qazalin:
enables BEAM=2, benchmarks, CI, openpilot process replay with the same infra.


Qazalin:
status: openpilot is the last one left with non deterministic process replay.

PR #4887, opened 2024-06-08
Convert BinaryOps.DIV to UnaryOps.RECIP and BinaryOps.IDIV

JhennerTigreros:
This PR introduces `UnaryOps.RECIP` and `BinaryOps.DIV`, and made the needed changes of functions to support both of them.

`a = Tensor([1,2,3], dtype=dtypes.float).tanh().realize()`

before

```
 0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ SUB 
  2    ┣━┳ MUL 
  3    ┃ ┣━━ CONST ConstBuffer(val=2.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
  4    ┃ ┗━┳ DIV 
  5    ┃   ┣━━ CONST ConstBuffer(val=1.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
  6    ┃   ┗━┳ ADD 
  7    ┃     ┣━━ CONST ConstBuffer(val=1.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
  8    ┃     ┗━┳ EXP2 
  9    ┃       ┗━┳ MUL 
 10    ┃         ┣━┳ MUL 
 11    ┃         ┃ ┣━━ CONST ConstBuffer(val=2.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
 12    ┃         ┃ ┗━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True),)))
 13    ┃         ┗━━ CONST ConstBuffer(val=-1.4426950408889634, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
 14    ┗━━ CONST ConstBuffer(val=1.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
```

after

```
*** PYTHON     1 copy       12,  PYTHON <- NPY          arg   2 mem  0.00 GB tm     13.23us/     0.01ms (    0.00 GFLOPS,    0.00 GB/s)
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ SUB 
  2    ┣━┳ MUL 
  3    ┃ ┣━━ CONST ConstBuffer(val=2.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
  4    ┃ ┗━┳ RECIP 
  5    ┃   ┗━┳ ADD 
  6    ┃     ┣━━ CONST ConstBuffer(val=1.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
  7    ┃     ┗━┳ EXP2 
  8    ┃       ┗━┳ MUL 
  9    ┃         ┣━┳ MUL 
 10    ┃         ┃ ┣━━ CONST ConstBuffer(val=2.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
 11    ┃         ┃ ┗━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(1,), offset=0, mask=None, contiguous=True),)))
 12    ┃         ┗━━ CONST ConstBuffer(val=-1.4426950408889634, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
 13    ┗━━ CONST ConstBuffer(val=1.0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(3,), strides=(0,), offset=0, mask=None, contiguous=False),)))
```

I'm working on test to pass, I find out two blockers right now:

1. The `test_symbolic_test::test_var*`
 are failing due to the function `F.Div.forward` now use `UnaryOps.RECIP` and `BinaryOps.MUL` to made the multiplication and when `BinaryOps.MUL` received 0 and INF value (RECIP returns this when the denominator is 0) return nan and the expected value is 0. I try to return 0 instead of INF but more tests start to fail, so I'm reading the higher logic to find a solution.

2. For CLANG and PTX, some errors are due to my current implementation, but now I'm focusing on fixing the symbolic tests, later I will fix this. I'm reading PTX ISA to understand how to implement the IDIV (Maybe casting the floats to integers before using the `div.{dtype}` instruction)


geohot:
If BinaryOps.DIV sill exists, this is wrong.

JhennerTigreros:
I forget to remove the BinaryOps.DIV, I'm working on rshift test, due to use UnaryOps.RECIP return a float its raise a dtype error because its expect uint not float.

JhennerTigreros:
I'm confident of my solution. I have two stoppers that I described on the description of the PR. I open the PR to received feedback. I need to study for an exam now, so I will be back tomorrow morning. Thanks for the time!

JhennerTigreros:
Update:

After some read I understood on the basics how PatternMatcher works and find the bug, as now I use `BinaryOps.MUL` and `UnaryOps.RECIP` when se have a 0 / 0 the expected result is `nan`. but this rule
`(UPat(UOps.ALU, BinaryOps.MUL, [UPat(), UPat(UOps.CONST, 0, name="c")]), lambda c: c)` 
unfold the division to 0, so I modify the rule to support the case when `UnaryOps.RECIP` returns `nan` to render as `nan` the value.
 
I have the following tests to fix:
* I have some tests (METAL, CLANG, GPU, NV) failing due to my C implementation of `BinaryOps.IDIV` returns `-2.147484e+09` or `2.147484e+09` when divide by 0, so I need to improve that.
* The LLVM and PTX test are failing due to bad type when both numbers are float. I will change the condition to cast both to integers.
* The PTX need to round the value to nearest integer I read the PTX ISA and try to use `rnd` modifier for `div`, but doesn't work 🤔 
* For openpilot and onnx I don't know yet what is failing, I suppose is due to the other errors described above.

Thanks for the time.

JhennerTigreros:
Update:

Finally almost all test pass, I have an error with Metal test, but when I test that on my local machine the fuzzer pass 

Screenshot 2024-06-10 at 10 35 19 PM

I use the same command as in the test workflow

```
PYTHONPATH="." METAL=1 FUZZ_ALL_ACTIONS=1 DEPTH=2 FUZZ_N=48 FUZZ_MAX_SIZE=10000000 python test/external/fuzz_linearizer.py
```

I'm working know on the ONNX+Optimization tests, it fails due to rshift implementing self.div

tinygrad/renderer/cstyle.py
```diff
@@ -25,10 +25,12 @@ class CStyleLanguage(Renderer):
   type_map: Dict[DType, str] = {}
   code_for_op: Dict = {
     UnaryOps.NEG: lambda x,dtype: f"(!{x})" if dtype == dtypes.bool else f"(-{x})", UnaryOps.SQRT: lambda x,dtype: f"sqrt({x})",
+    UnaryOps.RECIP: lambda x,dtype: f"(1/{x})",
     UnaryOps.EXP2: lambda x,dtype: f"exp2({x})", UnaryOps.LOG2: lambda x,dtype: f"log2({x})", UnaryOps.SIN: lambda x,dtype: f"sin({x})",
-    BinaryOps.ADD: lambda a,b,dtype: f"({a}+{b})", BinaryOps.SUB: lambda a,b,dtype: f"({a}-{b})", BinaryOps.MUL: lambda a,b,dtype: f"({a}*{b})",
-    BinaryOps.DIV: lambda a,b,dtype: f"({a}/{b})", BinaryOps.MAX: lambda a,b,dtype: f"max({a},{b})", BinaryOps.MOD: lambda a,b,dtype: f"({a}%{b})",
-    BinaryOps.CMPLT: lambda a,b,dtype: f"({a}<{b})", BinaryOps.CMPNE: lambda a,b,dtype: f"({a}!={b})", BinaryOps.XOR: lambda a,b,dtype: f"({a}^{b})",
+    BinaryOps.ADD: lambda a,b,dtype: f"({a}+{b})", BinaryOps.SUB: lambda a,b,dtype: f"({a}-{b})", BinaryOps.MAX: lambda a,b,dtype: f"max({a},{b})",
+    BinaryOps.IDIV: lambda a,b,dtype: f"(int)({a}/{b})",
```
chenyuxyz, line 31: don't think you want to hard code `(int)`, why does not this preserve the dtype?

tinygrad/renderer/cstyle.py
```diff
@@ -239,7 +241,8 @@ def render_kernel(self, function_name, kernel, bufs, uops, prefix=None):
   return {arg[3].name}2(c.thread_elements()[0], c.thread_elements()[1]);\n}}""")
     return super().render_kernel(function_name, kernel, bufs, uops, prefix)
 
-code_for_op_half = {BinaryOps.MAX: lambda a,b,dtype: f"__hmax({a},{b})" if dtype in (dtypes.half, dtypes.bfloat16) else f"max({a},{b})",
+code_for_op_half = {UnaryOps.RECIP: lambda x,dtype: f"({'(half)'if dtype == dtypes.half else ''}1/{x})",
```
chenyuxyz, line 244: why override this?

chenyuxyz (commented):
the dice error looks like noise, feel free to add a very small atol / rtol

tinygrad/renderer/cstyle.py
JhennerTigreros: I remember to overrided this because some test_uops and test_ops where failing, I tests again without this override and all works (maybe I had some errors on my first try), submit the commit without this line now, thank you.

JhennerTigreros:
Almost all tests passes. I have the metal test errors on CI, but when I run them on my local machine it passes.

My current specs are

```
Device Name: Apple M1 Pro
Max Threads per Threadgroup: MTLSize(width: 1024, height: 1024, depth: 1024)
Are raster order groups supported: true
```

Which are the Metal runner config? or maybe there is something broken with the dataset that I upload? 🤔

chenyuxyz:
It's possible that CI metal does not support bfloat16 and we did not hit it before. Fine to add `expected-failures=1` to skip it for now

how did you generated the dataset? how many ast lines are there v.s. the master. I saw a big file size drop and would like to check it's generated correctly

JhennerTigreros:
> It's possible that CI metal does not support bfloat16 and we did not hit it before. Fine to add `expected-failures=1` to skip it for now

I just check the MSL and it supports bfloat for Metal 3.1 or later, [page 25](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf)

Screenshot 2024-06-11 at 2 30 47 PM

> how did you generated the dataset? how many ast lines are there v.s. the master. I saw a big file size drop and would like to check it's generated correctly

I just ran `./extra/optimization/generate_dataset.sh` in a CUDA machine (RTX 4070 Super) and let finish it (Does is It need to be generated on Metal machine?), I'm running that again in the same CUDA machine just to validate.

chenyuxyz:
CI metal is quite different from real metal for some unknown reason, like it does not support ICB. If it works on a real mac machine I think it's fine. We probably want to update that test to make it more robust, but it does not need to be this pr.

Try adding `CI=1` if you want to use cuda machine, I think it won't run the bfloat16 kernels. Last time I used a mac to generate the dataset and that was fine, but it might just be lucky.

Still want the new kernel count v.s. master kernel count of the dataset to make sure it's not off too much

JhennerTigreros:
It start throwing these errors. 

Screenshot 2024-06-11 at 2 36 34 PM

Adding this line again `UnaryOps.RECIP: lambda x,dtype: f"({'(half)'if dtype == dtypes.half else ''}1/{x})",`` let pass the tests, is needed to support those `half` operations.

Line count (is the same as kernel count?) of the datasets 

master: 11644

my branch: 11888

test/external/external_test_metrics.py
```diff
@@ -10,7 +10,7 @@ class ExternalTestMetrics(unittest.TestCase):
   def _test_metrics(self, tinygrad_metrics, orig_metrics, pred, label):
     tinygrad_metrics_res = tinygrad_metrics(Tensor(pred), Tensor(label)).squeeze().numpy()
     orig_metrics_res = orig_metrics(torch.from_numpy(pred), torch.from_numpy(label)).numpy()
-    np.testing.assert_equal(tinygrad_metrics_res, orig_metrics_res)
+    np.testing.assert_allclose(tinygrad_metrics_res, orig_metrics_res, atol=1e-6, rtol=1e-3)
```
chenyuxyz, line 13: that's very big of an margin, do you need this big?

JhennerTigreros: Nop, reduced to atol=1e-8, rtol=1e-7

JhennerTigreros:
I'm don't understand why this test fails, I'm running locally and don't say memory error 🤔 

Screenshot 2024-06-11 at 6 02 15 PM

chenyuxyz:
I can reproduce the error locally with python 3.8.18, what's the python version your are using?

JhennerTigreros:
I'm using 3.12.3

chenyuxyz:
looks like you have it fixed? also there are two data set files in the pr

JhennerTigreros:
Yep, weird behavior. Literally I just restart my computer a free caches of my machine and run the generated script again 🤔 .

chenyuxyz:
one last thing from me: remove the code change in `/extra`, those are not tested so better to leave it as is

will trigger the benchmark run after

extra/assembly/assembly.py
```diff
@@ -146,7 +146,7 @@ def uops_to_asmstyle(lang, function_name:str, uops:List[UOp]):
         pred_reg = lang.newreg((u, 'pred'), dtype=dtypes.bool)
         lang.ins.append(AssemblyInstruction(UOps.ALU, pred_reg, [lang.tor[x] for x in vin], args))
         lang.ins.append(AssemblyInstruction(UOps.CAST, out, [pred_reg], args))
-      elif args == BinaryOps.DIV and lang.no_div:
+      elif args == BinaryOps.IDIV and lang.no_div:
```
chenyuxyz, line 149: still this and one below, please also read the whole pr diff again

test/test_custom_function.py
```diff
@@ -43,8 +43,8 @@ def forward(self, a:LazyBuffer, b:LazyBuffer) -> LazyBuffer:
                              arg={"GPU": atan2_gpu, "CPU": atan2_cpu}[a.device], srcs=(a.contiguous(), b.contiguous()))
   def backward(self, grad_output:LazyBuffer) -> Tuple[Optional[LazyBuffer], Optional[LazyBuffer]]:
     denom = (self.a.e(BinaryOps.MUL, self.a)).e(BinaryOps.ADD, self.b.e(BinaryOps.MUL, self.b))
-    return grad_output.e(BinaryOps.MUL, self.b.e(BinaryOps.DIV, denom)) if self.needs_input_grad[0] else None, \
-           grad_output.e(BinaryOps.MUL, self.a.const(0).e(BinaryOps.SUB, self.a).e(BinaryOps.DIV, denom)) if self.needs_input_grad[1] else None
+    return grad_output.e(BinaryOps.MUL, self.b.e(BinaryOps.MUL, denom.e(UnaryOps.RECIP))) if self.needs_input_grad[0] else None, \
+           grad_output.e(BinaryOps.MUL, self.a.const(0).e(BinaryOps.SUB, self.a).e(BinaryOps.MUL, denom.e(UnaryOps.RECIP))) if self.needs_input_grad[1] else None # noqa: E501
```
chenyuxyz, line 47: missed this earlier, we don't want new `# noqa: E501`

tinygrad/codegen/uops.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from typing import Iterator, Optional, Tuple, Any, Dict, List, DefaultDict, Set, Callable, Union, cast, TypeVar
+import math
```
chenyuxyz, line 3: this should merge with the next line

test/test_custom_function.py
JhennerTigreros: I have this modification on pattern matcher.

`(UPat(UOps.ALU, BinaryOps.MUL, [UPat(name="x"), UPat(UOps.CONST, 0, name="c")]), lambda x,c: c if not isinstance(x.arg, (int, float)) else x if math.isnan(x.arg) else c)`

I need that two how the x*0 unfolds work, I try to create a new rule the specific x=nan but don't work. 🤔

JhennerTigreros: I try this rule

```
(UPat(UOps.ALU, BinaryOps.MUL, [UPat(UOps.CONST, math. nan, name="x"), UPat()]), lambda x: x),
```

Without modifying the original, and throws errors, Can I have two rules for the same Ops but different `CONST`? 🤔 

Screenshot 2024-06-11 at 7 10 31 PM

JhennerTigreros: Ignored, just rewrite the whole line. Testing now

JhennerTigreros: I change the `math` import to just import isnan function, is the unique use here. and pass the check. Is that valid? Also I rewrite the whole line to

`(UPat(UOps.ALU, BinaryOps.MUL, [UPat(name="x"), UPat(UOps.CONST, 0, name="c")]), lambda x,c: x if isinstance(x.arg, float) and isnan(x.arg) else c),`

chenyuxyz: you don't need to change the algorithm, the point is that we don't a very long line as it's less readable

tinygrad/codegen/uops.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 from typing import Iterator, Optional, Tuple, Any, Dict, List, DefaultDict, Set, Callable, Union, cast, TypeVar
 import functools, itertools, heapq
+from math import isnan
```
chenyuxyz, line 4: just `import functools, itertools, heapq, math`. The point is to have `math` be in the same line as other std imports

test/test_custom_function.py
JhennerTigreros: If I leave the math import merged with the next one, the line need a E501, so in this case is valid to left a E501 in this line, if understand correctly?

chenyuxyz: no, just break the line somehow, either parentheses or `\`

tinygrad/codegen/uops.py
```diff
@@ -194,10 +195,12 @@ def loop_collapse(loop_start, loop_end, compval, idx, mval, multconst):
...
   (UPat(UOps.ALU, BinaryOps.SUB, (UPat(name="x"), UPat(UOps.CONST, 0))), lambda x: x),   # x-0 -> x
-  (UPat(UOps.ALU, BinaryOps.DIV, (UPat(name="x"), UPat(UOps.CONST, 1))), lambda x: x),   # x/1 -> x
-  (UPat(UOps.ALU, BinaryOps.DIV, (UPat(name="x"), UPat(UOps.CONST, -1))), lambda x: -x), # x/-1 -> -x
+  (UPat(UOps.ALU, BinaryOps.IDIV, (UPat(name="x"), UPat(UOps.CONST, 1))), lambda x: x),   # x/1 -> x
+  (UPat(UOps.ALU, BinaryOps.IDIV, (UPat(name="x"), UPat(UOps.CONST, -1))), lambda x: -x), # x/-1 -> -x
   # ** zero folding **
-  (UPat(UOps.ALU, BinaryOps.MUL, [UPat(), UPat(UOps.CONST, 0, name="c")]), lambda c: c), # x*0 -> 0 or 0*x -> 0
+  #x*0 -> 0 or 0*x -> 0
+  #if x is nan it should render the nan value.
+  (UPat(UOps.ALU, BinaryOps.MUL, [UPat(name="x"), UPat(UOps.CONST, 0, name="c")]), lambda x,c: c if not isinstance(x.arg, (int, float)) else x if math.isnan(x.arg) else c), # noqa: E501
```
chenyuxyz: does `lambda x,c: x if isnan(x.arg) else c),` work?

JhennerTigreros: Nop, in some tests `x.arg` is None and `isnan` throws error because expect a real number.

Screenshot 2024-06-11 at 7 55 07 PM

test/test_custom_function.py
JhennerTigreros: Ohh got the point, thx for the explanation.

tinygrad/codegen/uops.py
chenyuxyz: hmm that's weird, but probably has nothing to do with this pr

JhennerTigreros:
I forgot to add `dtypes.bfloat16` in the override for `code_for_op_half`

JhennerTigreros:
I will let finish the benchmark and solve the errors raised.

chenyuxyz:
there's `hrcp` that should work for both half and bfloat16 https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH____HALF__FUNCTIONS.html

JhennerTigreros:
Thanks, pushed the fix.

chenyuxyz:
don't think `hrcp` requires the `1`, it itself is reciprocal

chenyuxyz:
@geohot please take a final look

test/test_uops.py
```diff
@@ -170,14 +170,24 @@ def test_sqrt(self):
...
-    self.assertEqual(exec_alu(BinaryOps.DIV, dtypes.int8, (-50, 6)), -8)
+    self.assertEqual(exec_alu(BinaryOps.IDIV, dtypes.int8, (8, 2)), 4)
+    self.assertEqual(exec_alu(BinaryOps.IDIV, dtypes.int8, (7, 3)), 2)
+    self.assertEqual(exec_alu(BinaryOps.IDIV, dtypes.int8, (7, -3)), -2)
+    self.assertEqual(exec_alu(BinaryOps.IDIV, dtypes.int8, (-50, 6)), -8)
 
-    np.testing.assert_allclose(exec_alu(BinaryOps.DIV, dtypes.float32, (8.0, 2.0)), 4.0)
-    np.testing.assert_allclose(exec_alu(BinaryOps.DIV, dtypes.float32, (7.0, 3.0)), 2+(1.0/3.0))
-    np.testing.assert_allclose(exec_alu(BinaryOps.DIV, dtypes.float32, (7.0, -3.0)), -2-(1.0/3.0))
+    np.testing.assert_allclose(exec_alu(BinaryOps.IDIV, dtypes.float32, (8.0, 2.0)), 4.0)
```
geohot, line 178: This should be asserting in the type checker.

JhennerTigreros: Thanks, just added a condition for validate if `vin[0].dtype` is int in IDIV.

tinygrad/codegen/uops.py
```diff
@@ -453,6 +456,8 @@ def type_verify(self):
         elif arg in (BinaryOps.CMPLT, BinaryOps.CMPNE):
           assert dtype == dtypes.bool, f"{arg} output dtype mismatch {dtype=} != {dtypes.bool}"
           assert vin[0].dtype == vin[1].dtype, f"{arg} dtype mismatch {dtype=} != {vin[0].dtype=} != {vin[1].dtype=}"
+        elif arg is BinaryOps.IDIV:
+          assert dtypes.is_int(vin[0].dtype), f"{arg} output dtype mismatch {dtype=} != {dtypes.int}"
```
chenyuxyz, line 460: need to check two inputs and output are all int

geohot:
Wait does integer Tensor divison still work after this PR? Are we testing that?

JhennerTigreros:
Hi, I see some in `test_ops.py` that check integer Tensor division 

Screenshot 2024-06-12 at 3 54 29 PM

I'm checking new error on `test_failure_32`, after updating my branch with master, taking a look before submitting a commit with the fixes.

`RuntimeError: CUDA Error 1, invalid argument`

Screenshot 2024-06-12 at 3 56 45 PM

JhennerTigreros:
Weird, I move to a previous commit (91c58048987710fc23f94c1898aeef5b4643a4d8) were tests passed on CI and on my local machine doesn't pass, pushing new commits to verify with CI. (maybe my machine has some problem with drivers or compiler)

JhennerTigreros:
Also need to regenerate the dataset due to changes in some `LazyOp` inside tests, running.

`PYTHONPATH=. ./extra/optimization/generate_dataset.sh`

JhennerTigreros:
I tested Fuzz Linearizer Metal on my local machine with diff python env and passed. My platform is `macOS-14.5-arm64-arm-64bit`

Python 3.12.3

Screenshot 2024-06-12 at 5 17 55 PM

Python 3.11.9

Screenshot 2024-06-12 at 5 24 51 PM


Python 3.8.19
Screenshot 2024-06-12 at 5 19 53 PM

JhennerTigreros:
I have some problems running the `generate_dataset.sh` due to memory constrains on my machine, so I just split some tests on different lines to prune the memory, also add JIT=2 and fix some issues related with the following error (read a message from geo about this)

```
RuntimeError: Error Domain=MTLCommandBufferErrorDomain Code=1 "Discarded (victim of GPU error/recovery) (00000005:kIOGPUCommandBufferCallbackErrorInnocentVictim)" UserInfo={NSLocalizedDescription=Discarded (victim of GPU error/recovery) (00000005:kIOGPUCommandBufferCallbackErrorInnocentVictim), NSUnderlyingError=0x13383b930 {Error Domain=IOGPUCommandQueueErrorDomain Code=5 "(null)"}}
```

Kernel count. My branch `11602` vs master `11644`

tinygrad/renderer/llvmir.py
```diff
@@ -15,12 +15,13 @@ def is_bool_or_unsigned(dtype: DType): return dtype == dtypes.bool or dtypes.is_
...
   UnaryOps.SQRT: lambda builder, x, dtype: builder.call(builder.module.declare_intrinsic('llvm.sqrt', [x.type]), [x], fastmath=MFLAGS),
   BinaryOps.ADD: lambda builder, x, y, dtype: builder.or_(x, y) if dtype == dtypes.bool else builder.add(x, y) if dtypes.is_int(dtype) else builder.fadd(x, y, flags=MFLAGS),  # noqa: E501
   BinaryOps.SUB: lambda builder, x, y, dtype: builder.sub(x, y) if dtypes.is_int(dtype) else builder.fsub(x, y, flags=MFLAGS),
   BinaryOps.MUL: lambda builder, x, y, dtype: builder.mul(x, y) if is_bool_or_unsigned(dtype) or dtypes.is_int(dtype) else builder.fmul(x, y, flags=MFLAGS),  # noqa: E501
-  BinaryOps.DIV: lambda builder, x, y, dtype: builder.udiv(x, y) if is_bool_or_unsigned(dtype) else builder.sdiv(x, y) if dtypes.is_int(dtype) else builder.fdiv(x, y, flags=MFLAGS),  # noqa: E501
+  BinaryOps.IDIV: lambda builder, x, y, dtype: builder.udiv(x, y) if is_bool_or_unsigned(dtype) else builder.sdiv(x, y) if dtypes.is_int(dtype) else builder.call(builder.module.declare_intrinsic('llvm.floor', [x.type]), [builder.fdiv(x, y, flags=MFLAGS)], fastmath=MFLAGS),  # noqa: E501
```
geohot, line 24: Why does this need llvm.floor when it didn't before? IDIV should have the same behavior as DIV. Why is there float support in here?

JhennerTigreros: Yep, IDIV only supports integers and the type checker should raise if try with floats, changing this line.

geohot:
Congrats, bounty is yours!

PR #4889, opened 2024-06-08
10% faster lower_schedule

croumegous:
The rewrite function is now defined outside the while loop to avoid redefining it every iteration.

On my laptop (average of 150 iterations)
baseline: 1760ms
this PR: 1580ms

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -254,29 +254,29 @@ class UOpGraph:
     # recursive rewrite
     changed = getenv("UOPS_REWRITE", 1)
     run_cnt = 0
+    @functools.lru_cache
+    def rewrite(u:UOp) -> UOp:
+      nonlocal changed
+      recurse_cnt = 0
+      up = u
+      # locally recursively rewrite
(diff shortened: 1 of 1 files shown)


tinygrad/codegen/uops.py
```diff
@@ -254,32 +254,32 @@ def graph_rewrite(self, sink, pm):
...
-        try: del up.cmp_tuple
-        except AttributeError: pass
-        # replace with cached nodes
-        if found:=self.nodes.get(key:=up.tuple()): return found
-        self.nodes[key] = up
-        return up
-      sink = rewrite(sink)
-      run_cnt += 1
-      assert run_cnt < 100, "exceeded 100 rewrite loops!"
+        changed = 0
```
chenyuxyz, line 279: bad indent size

croumegous:
Well, tests are failing for UOps.UNMUL, and I have no idea how to fix that, so I'll just close the PR, sorry for the incovenience.  
(But I still think the idea of moving the function out of the loop is good (if you can make it work) so ofc feel free to use it)

PR #4893, opened 2024-06-09
refactor: using a cached function enables direct cache clearing

roelofvandijk:
For cached functions, an explicit `cache_clear()` method exists. This is cleaner and slightly faster.

`PROFILE=0 python test/external/external_benchmark_schedule.py`

Master
```
***** model forward in  16.24 ms
***** model schedule in   5.76 ms
***** model lower in 834.88 ms

```

Branch
```
***** model forward in  16.13 ms
***** model schedule in   5.65 ms
***** model lower in 795.30 ms
```


geohot:
Does this pass process replay? What's your lru_cache size?

PR #4895, opened 2024-06-09
Convert BinaryOps.DIV to UnaryOps.RECIP and BinaryOps.IDIV.

sytandas:
Make it clean as possible,
try to find the logical error behind llvm unaryop.


geohot:
If BinaryOps.DIV sill exists, this is wrong.

sytandas:
Got it, work on it.

PR #4911, opened 2024-06-11
add missing dir level for amd mockgpu

nimlgen:


geohot:
Sweet, looks like this fixes it!

PR #4913, opened 2024-06-11
don't use uops.add while constructing

geohot:
This does make things slower due to the extra BFS on the UOps graph, but it isolates UOp construction from the graph and will allow speedups there.

The key victory is removing state from the Linearizer.

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -49,7 +49,7 @@ class Linearizer(Kernel):
 
   # NOTE: the consts have to be cached for deduping of downstream uops to work
   def const(self, b:ConstType|Variable, dtype:DType=dtypes.int32) -> UOp:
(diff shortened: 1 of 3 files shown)


tinygrad/codegen/linearizer.py
```diff
@@ -428,7 +429,7 @@ def render_block(self, outputs:Tuple[LazyOp, ...], global_idxs, local_idxs, upca
       # TODO: delete render_reduceop and move the logic for group_for_reduces to Block
       local_idxs[:], upcast_idxs[:] = self.render_reduceop((r:=reduceops[0]),accs,loaded_buffers,global_idxs,local_idxs,upcast_idxs, full_upcast_idxs,
                                                      reduce_idxs,fake_reduce_idxs,alias_buf_idxs[r])
-      return accs[r]
+      return [accs[r]]
```
geohot, line 432: @Qazalin can you add types to this function? types should have caught this

PR #4914, opened 2024-06-11
revert hsa autogen removal

Qazalin:
I think only deleting rocm-llvm should work.


geohot:
Open question why CI worked here.

PR #4922, opened 2024-06-12
llama3 `--download_model`

wozeparrot:
adds `--download_model` to automatically download a recommended model
also makes `--model` nicer to use by detecting the load format if its just pointed to a directory


geohot:
```
jesse@x1:~/tinygrad$ python3 examples/llama3.py --download_model
usage: llama3.py [-h] [--download_model] --model MODEL [--size {8B,70B}] [--shard SHARD]
                 [--quantize {int8,nf4}] [--api] [--host HOST] [--port PORT] [--debug]
                 [--seed SEED] [--timing] [--profile]
llama3.py: error: the following arguments are required: --model
```

wozeparrot:
I guess the main problem with putting it in the normal fetch cache is that the index file points to very specific filenames.

Issue #4923, opened 2024-06-12
[WIP] Proposal: Shared memory allocation for multi reduce


Qazalin:
llm.c has two ways of allocating memory for a multi reduce kernel with GROUP:

1. split a single block between reduceops

https://github.com/karpathy/llm.c/blob/master/llmc/layernorm.cuh#L165-L175

[reuse a split](https://github.com/karpathy/llm.c/blob/master/llmc/layernorm.cuh#L295) when we stored it to global.

```c 
extern __shared__ float shared[];
// the first half of shared memory is bias, second is weight
size_t rounded_C = CEIL_DIV(C, (32 * x128::size)) * (32 * x128::size);
float* dbias_shared = shared;
float* dweight_shared = shared + rounded_C;
```

2. one block per RANGE

https://github.com/karpathy/llm.c/blob/master/train_gpt2_fp32.cu#L526

```c
 // two reductions of up to 1024 threads:
 // 1) inside warp (shuffle), 2) cross-warp (shared memory), 3) inside warp (shuffle)
 // this results in much cleaner assembly than a multi-warp cg::reduce
 __shared__ float shared_maxval[32];
 __shared__ float shared_sumval[32];
```

write once per reduce

```c
 // reduce maxval within each warp
 float warp_maxval = cg::reduce(warp, thread_maxval, cg::greater<float>{});
 // thread 0 in each warp writes to shared memory
 if (lane_id == 0) { shared_maxval[warp_id] = warp_maxval; }

 // (warp-level) reduce sumval, thread 0 in each warp saves result in shared memory
 float warp_sumval = cg::reduce(warp, thread_sumval, cg::plus<float>{});
 if (lane_id == 0) { shared_sumval[warp_id] = warp_sumval; }
 __syncthreads();
```

--- 

3. @0xtimmy's: use the same memory twice at the expense of an extra sync

```c
  // 1st early reduce stores to smem
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 2; ridx0++) {
    float val0 = *(data1+alu1+ridx0);
    acc0 = (val0+acc0);
  }
  *(temp+lidx1) = acc0;
  // 1st late reduce stores the final result to smem
  threadgroup_barrier(mem_flags::mem_threadgroup);
  if (alu0) {
    float acc1 = 0.0f;
    for (int ridx1 = 0; ridx1 < 16; ridx1++) {
      float val1 = *(temp+ridx1);
[...]

PR #4924, opened 2024-06-12
allocate shared memory per block

Qazalin:
This diff allocates one temp buffer per block and unblocks multi reduce.

LocalBuffer allocation needs to be more thought through https://github.com/tinygrad/tinygrad/issues/4923


Qazalin:
I think this is the right complexity/speed trade-off for now. @0xtimmy the atomic store diff adds complexity to global_store, BARRIER in uops and changes the STORE spec, do you think the extra SYNC is worth the memory trade-off? have you benchmarked it?

We need to think through what the trade-offs for speed / memory banking are for these options https://github.com/tinygrad/tinygrad/issues/49232

PR #4925, opened 2024-06-12
amd hdp flush regs are on seg2

nimlgen:


geohot:
Ahh, is this the cause of the page faults?!

nimlgen:
Hard to say if this fixes 100% of all page faults, will test for a longer time.

PR #4929, opened 2024-06-12
more typing in linearizer uoping utils

Qazalin:
there was a bug in render_block because of accs not being typed.
This diff also showed a lot of places in the code where idx are wrongly/incompletely typed. These should all be just a List[UOp]


Qazalin:
merging pre multireduce

PR #4930, opened 2024-06-12
fix typing for test to run in py38

smurfd:
Fix for test `test/test_fuzz_shape_ops.py` to be able to run by python3.8
`python3 -m pytest test/test_fuzz_shape_ops.py` would others fail.


chenyuxyz:
`from __future__ import annotations` should work?

smurfd:
yes, that works :)

PR #4935, opened 2024-06-12
Residual in MLM loss + Change default steps

Eliulm:
Resolves small inconsistency. [Reference](https://github.com/mlcommons/training/blob/master/language_model/tensorflow/bert/run_pretraining.py#L315)


chenyuxyz:
cool! can you add a comment that links to the reference to justify the reimplementation?

PR #4936, opened 2024-06-12
failed test case for getitem with leading Nones

chenyuxyz:
torch matched numpy so tinygrad is incorrect.
another repro
```
t = np.arange(12).reshape((3, 4))
print(t[None, None, np.array([1, 2])])

t = torch.arange(12).reshape((3, 4))
print(t[None, None, torch.tensor([1, 2])].numpy())

t = Tensor.arange(12).reshape(3, 4)
print(t[None, None, Tensor([1, 2])].numpy())
```


chenyuxyz:
@geohotstan fyi, I was trying to understand how None was handled and want to remove the dim re-calculation in advance indexing. It looks like it's not handled correctly in general

geohotstan:
> @geohotstan fyi, I was trying to understand how None was handled and want to remove the dim re-calculation in advance indexing. It looks like it's not handled correctly in general

huh, that's a weird one...
my bad for not finding it.
from first glances, maybe `calc_dim()` is wrong. I'll investigate. 

I had an idea for an object oriented way of keeping track of dims and their values, which would probably be less bug prone since it doesn't require any dim calculation.
but I had to rush some other project recently :c

PR #4939, opened 2024-06-12
tqdm total=0 fix

SnakeOnex:


tinygrad/helpers.py
```diff
@@ -208,6 +208,7 @@ def fetch(url:str, name:Optional[Union[pathlib.Path, str]]=None, allow_caching=n
       with tempfile.NamedTemporaryFile(dir=path, delete=False) as f:
         while chunk := r.read(16384): progress_bar.update(f.write(chunk))
         f.close()
+        progress_bar.update(close=True)
```
wozeparrot, line 211: is this required to be called?

SnakeOnex: good question, the original `tqdm` implementation didn't have this and it worked fine. 
However, in their [documentation](https://tqdm.github.io/docs/tqdm/#update) they actually say: 
"The last line is highly recommended, but possibly not necessary if `t.update()` will be called in such a way that filesize will be exactly reached and printed."

The point of the `close=True` line is to output a '\n' that ends the progress bar line. (so that when for example fetch is called twice in a row, the second bar doesn't overwrite the first one)
so the `tinytqdm` implementation as it is right now needs the `close=True` to be called when using in the manual `.update()` mode.

the question is how does `tqdm` do it. 
this code overwrites the test string
```
print('test string: ', end='')
bar = tqdm()
```
this however prints two empty bars (they don't overwrite each other)
```
bar = tqdm()
bar = tqdm()
```

https://stackoverflow.com/a/78127892
I found this, but I think it hints that they do some complicated multi bar position handling, which wouldn't be tinyTM fit to implement for us.
But if you have any ideas how to do this in a tinyTM and smart manner I am interested. (I already tried stuff like outputting a `\n` in each iter and then going a line up with ANSI movement chars, but this would nuke my terminal at high iters/s)

wozeparrot: can we just print a single `\n` at the beginning to separate the bars?

SnakeOnex: then if you have only one bar, there would be an empty line before it

wozeparrot: that might be a okay tradeoff, could also have a classvar or global var that stores the number of bars drawn and only print `\n` if its greater than 0 or something like that.

SnakeOnex: the way it is now actually matches `tqdm` (`tqdm` also has a .close() method and it is recommended to be used) and it only adds 1 line while matching `tqdm` behaviour everywhere

the solution of having global counter -> if its not zero: print an empty line infront might work, i think that will add atleast 2 lines tho (arguably also might be a bit hacky), 
i can try it tho

PR #4947, opened 2024-06-13
construct Tensor from python list/tuple directly

chenyuxyz:
no numpy. annoying that half memoryview is 3.12 feature...

### DIFF
--- a/test/test_schedule.py
+++ b/test/test_schedule.py
@@ -445,7 +445,7 @@ class TestSchedule(unittest.TestCase):
 
   def test_double_from(self):
     x = Tensor([1,2,3,4])
-    out = x.to('npy')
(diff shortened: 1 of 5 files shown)


chenyuxyz:
@geohot take a look?

chenyuxyz:
about 50% slower compared to numpy with a 200x200x200 array of floats. About half of the time was to infer the dtype, so explicitly specifying dtype is a workaround, which can make it slightly faster than numpy

given that, I am inclined to merge this first and continue other numpy removal

chenyuxyz:
another practical speed thing is that `dtypes.frompy` can only return `bool, default_float, default_int`, so we can return early once seeing a float. This does not help if the array is all int though

PR #4949, opened 2024-06-14
autogen _try_dlopen

wozeparrot:
tries a variety of paths including ctypes find_library before failing with a nicer error


wozeparrot:
@nimlgen can you check that the rocm bump to 6.1.2 is OK?

nimlgen:
I think should be good, do not see any bad changes in the api

PR #4952, opened 2024-06-14
move uops add logic to linearize

geohot:


Qazalin:
yeah fuzz_uops should create an empty graph

PR #4953, opened 2024-06-14
IF ends at STORE, RANGE ends at PHI [run_process_replay]

Qazalin:
STORE and PHI define the end of the IF and RANGE graph. Whereas BARRIER is a side-effect of multiple IFs+GROUP.


Qazalin:
Process replay shows this reorders some ALUs, looks ok for correctness.

PR #4958, opened 2024-06-14
perf: graph_rewrite line reduction and make it a little bit faster

croumegous:
Tried to make rewrite fast but in the end it's not significantly faster than before.  
However I've achieved some line reduction that are nice to have, I guess.

In my local bench `PROFILE=0 python test/external/external_benchmark_schedule.py` lower was like ~30ms faster so has i said not significantly faster. But the new way to refresh cached property is [indeed faster](https://pastebin.com/xNKgyRg8).

The setdefault is exactly the same speed but it's nice for line reduction

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -294,14 +294,10 @@ class UOpGraph:
         changed += recurse_cnt
         # NOTE: this changes UOp, so we have to delete caches
         up.vin = tuple(rewrite(x) for x in up.vin)
-        try: del up.parents
-        except AttributeError: pass
-        try: del up.cmp_tuple
-        except AttributeError: pass
+        if 'parents' in up.__dict__: delattr(up, 'parents')
+        if 'cmp_tuple' in up.__dict__: delattr(up, 'cmp_tuple')
         # replace with cached nodes
-        if found:=self.nodes.get(key:=up.tuple()): return found
-        self.nodes[key] = up
-        return up
+        return self.nodes.setdefault(up.tuple(), up)
       sink = rewrite(sink)
       run_cnt += 1
       assert run_cnt < 100, "exceeded 100 rewrite loops!"


geohot:
This might be a bit too "whitespace" to merge. Can you at least run it with process replay?

tinygrad/codegen/uops.py
```diff
@@ -294,14 +294,10 @@ def rewrite(u:UOp) -> UOp:
...
-        except AttributeError: pass
-        try: del up.cmp_tuple
-        except AttributeError: pass
+        if 'parents' in up.__dict__: delattr(up, 'parents')
+        if 'cmp_tuple' in up.__dict__: delattr(up, 'cmp_tuple')
         # replace with cached nodes
-        if found:=self.nodes.get(key:=up.tuple()): return found
-        self.nodes[key] = up
-        return up
+        return self.nodes.setdefault(up.tuple(), up)
```
geohot, line 300: Interesting. Do we have this pattern elsewhere in the code?

croumegous: Yes but might not be usefull everywhere (for perf) as the second argument is evaluated :
```python
>>> a = {"a": 1}
>>> a.setdefault("a", print("b"))
b
1
```
So it some case it's better to keep additional `if` to avoid extra compute

tinygrad/codegen/uops.py
```diff
@@ -294,14 +294,10 @@ def rewrite(u:UOp) -> UOp:
         changed += recurse_cnt
         # NOTE: this changes UOp, so we have to delete caches
         up.vin = tuple(rewrite(x) for x in up.vin)
-        try: del up.parents
-        except AttributeError: pass
-        try: del up.cmp_tuple
-        except AttributeError: pass
+        if 'parents' in up.__dict__: delattr(up, 'parents')
+        if 'cmp_tuple' in up.__dict__: delattr(up, 'cmp_tuple')
```
Qazalin, line 298: hmm so we had an if here but it got removed because of being "too slow"
- is this faster if you isolate it?
- is there a perf diff between this and hasattr?

croumegous: I've updated my [benchmark with time result at the end ](https://paste.fo/50de9641ee2a).
`del up.parents ` is as fast as `delattr(up, 'parents')`  
but `hasattr` is indeed very slow as well as `getattr` (they are both used a lot in tinygrad, but we can't use `.__dict__` on every object)

PR #4960, opened 2024-06-14
Getitem pin None dimension

geohotstan:
The bug before was when there were consecutive Nones paired with other advanced indexing...

This PR fetches the dimensions before taking out the Nones so that it's pinned (no longer need to calc it). 
Also added a few more test cases.

Still the code isn't pretty... at least it's correct now I think.


chenyuxyz:
thanks!

PR #4962, opened 2024-06-14
RNNT

DKormann:


DKormann:
first epoch is gonna take around 100 minutes on single gpu
[00:12:33.4 batch 59/557 Loss :1.40793]
mainly bottlenecked by insane memory requirements for lstm backward pass

PR #4322, opened 2024-04-27
search children in fusion

Qazalin:
This expands the scope of fusion - eg.
```
r3_50257_768_256(adam_m_lm_head_weight, adam_v_lm_head_weight, lm_head_weight, adam_b1, b166, b167, b432, b446, b1118, adam_b2, b177, b15, adam_lr, b181, b182)
```
vs mater:
```
r_50257_768_256n1(grad_lm_head_weight, b420, b436, b951)
E_38597376(adam_m_lm_head_weight, adam_b1, b166, b167, grad_lm_head_weight)
E_38597376n1(adam_v_lm_head_weight, adam_b2, b166, b177, grad_lm_head_weight)
E_38597376n2(lm_head_weight, b15, adam_lr, adam_m_lm_head_weight, b181, adam_v_lm_head_weight, b182)
```

PR #4972, opened 2024-06-14
Multireduce Scheduler Tests

0xtimmy:
Beginning of the scheduler tests for multireduce. This covers what I think are the obvious cases to look out for, I'll add some as I got to reflect whatever changes to the scheduler


test/test_schedule.py
```diff
@@ -518,13 +528,28 @@ def test_reduce_shrink(self):
     out = x + y
     check_schedule(out, 2)  # TODO: this should be 1
 
+  # multireduce spec
+  @unittest.skip
+  def test_multireduce_shrink(self):
+    a = Tensor.empty(32, 32)
```
Qazalin, line 534: can you replace empty with randn so that we can test correctness?

Qazalin (commented):
The best way to create scheduler specs is if you check the schedule kernel count with the current counts and then modify.

So you wouldn't skip tests. Once the fusion is complete we'd have a diff like this:
https://github.com/tinygrad/tinygrad/pull/4322/files

Qazalin:
two more development experience points that'd help:
- Always seed the randn kernel
- Realize the Tensor.randn before check_schedule to cleanup DEBUG=4

0xtimmy:
bet

test/test_schedule.py
```diff
@@ -602,10 +647,65 @@ def test_reduce_shrink_output(self):
     out1 = out0[0] + Tensor.empty(1, )
     check_schedule([r, out0, out1], 3)
 
+  # multireduce spec
+  @unittest.skip
+  def test_multireduce_fusion(self):
```
Qazalin, line 652: can you break this into multiple defs, ideally each having one check_schedule

test/test_schedule.py
```diff
@@ -734,6 +834,19 @@ def test_reduce_simple_chase(self):
...
+  @unittest.skip
+  def test_multireduce_simple_chase(self):
+    a = Tensor.rand(4, 4, 4)
+    r = (a + (a.sum(0, keepdim=True) + 6)).sum(0) * 2
+    b = r.sum(0) + 8
+    c = r.sum(1) + 12
+    np_r = (a.numpy() + (a.numpy().sum(0) + 6)).sum(0) * 2
+    np.testing.assert_allclose(b.numpy(), np_r.sum(0) + 8)
+    np.testing.assert_allclose(c.numpy(), np_r.sum(1) + 12)
+    schedule = check_schedule([b, c], 3)
```
Qazalin, line 847: you can't reschedule b and c.
these should be `s = create_schedule()` then `run_schedule(s)`

PR #4975, opened 2024-06-15
DEFINE_ACC takes UOps.CONST in vin instead of arg

JhennerTigreros:
I tried to make the change straightforward. The unique thing that I am thinking about is the Python Emulator, instead of clear the `idp` I overwrite to just have the last position, that due the `DEFINE_ACC` is in first in `self.uops`, so `RANGE` data isn't append to `ul` and raise KeyError when trying to get from `ul` list.

Also I take a look of some kernels generated from `TestOps.test_gemm` and doesn't seem any change, but I changed Pattern Matcher, to be sure that nothing break, should this PR have run_process_replay ?


JhennerTigreros:
I tried to make the change straightforward. The unique thing that I am thinking about is the Python Emulator, instead of clear the `idp` I overwrite to just have the last position, that due the `DEFINE_ACC` is in first in `self.uops`, so `RANGE` data isn't append to `ul` and raise KeyError when trying to get from `ul` list.

Also I take a look of some kernels generated from `TestOps.test_gemm` and doesn't seem any change, but I changed Pattern Matcher, to be sure that nothing break, should this PR have `run_process_replay `?

JhennerTigreros:
Also I'm working to change this line

`ul[i] = [[inp[0][0][0]] * warp_size for _ in range(dtype.count)] if dtype.count > 1 else [inp[0][0]] * warp_size``

For a reason that I don't know yet, pass `inp` doesn't work

geohot:
Can you put the const first instead of last? Also, add [run_process_replay] to the title of your PR, this shouldn't change the generated source.

JhennerTigreros:
Some errors on process replay, taking a look. I come back when I have a fix, may be due to my changes in Pattern Matcher.

JhennerTigreros:
Update

After some reading of the linearizer I notice up that the problem is with the priority queue, for some weird reason the `IDIV` pop first than `MOD`, the only change in the queue is that the new `UOps.CONST` value is pushed early so maybe this broke the whole priority queue. I will be reading the oficial docs to understand the implementation

JhennerTigreros:
Just thinking, I added a new UOps to the graph maybe this change the complete behavior and the process replay should fail due to minor changes on the underlying graph? I supposed this because the diff is mainly in the order of the `ALU` and subsequent references for that `ALU`. Also the BFS could change due that `DEFINE_ACC` now had one more children.

master

```
Parent -> UOps.DEFINE_ACC     : dtypes.float              [<UOps.RANGE: 21>]               (0.0, 0, 0)
0 -> UOps.RANGE          : dtypes.int                [<UOps.CONST: 7>, <UOps.CONST: 7>] (2, 0)
```

my branch

```
Parent -> UOps.DEFINE_ACC     : dtypes.float              [<UOps.CONST: 7>, <UOps.RANGE: 21>] (0, 0)
0 -> UOps.CONST          : dtypes.float              []                               0.0
1 -> UOps.RANGE          : dtypes.int                [<UOps.CONST: 7>, <UOps.CONST: 7>] (2, 0)
```

JhennerTigreros:
Update: 

I'm more sure that the current changes affect the BFS order, that because now the const UOps is not render in the pattern matcher, instead is renderer in the graph generation in `graph_dedup`. That's why we don't need this rule now in PM `(UPat(UOps.DEFINE_ACC, name="root", src=(UPat(UOps.CONST),)), lambda root: root.src[0]),` because now exists a children with that value. 

For example in the PTX diff current vs. master can see

```
{
 	.reg		.u64 %dat_u64_<3>;
+	.reg		.f32 %const_f32_<5>;
 	.reg		.s32 %const_s32_<97>;
 	.reg		.s64 %const_s64_<2>;
-	.reg		.f32 %const_f32_<1>;
 	.reg		.s32 %alu_s32_<2>;
 	.reg		.s64 %cast_s64_<2>;
 	.reg		.s64 %alu_s64_<3>;
```

This is because the `UOps.CONST` is later in the BFS reverse order, so the `.reg` is renderer with high suffix.

To check indeep this, how can I run whole process replay in my local env and see the whole changes that the PR produce to debug deeply? 🤔 I'm trying but need to have some SQLite stuff.

Qazalin:
docs for running process replay locally: #5083

JhennerTigreros:
Thanks @Qazalin .

After reading the output of the failed ones. I think this PR produce changes on the codegen due to new children in the graph for `DEFINE_ACC`. In all cases the change is moving some ALUs order of generation, so I will remove [run_process_replay] in the title, is this ok?

tinygrad/runtime/ops_python.py
```diff
@@ -90,7 +90,7 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
         elif uop is UOps.CONST:
           ul[i] = [[arg] * warp_size for _ in range(dtype.count)] if dtype.count > 1 else [arg] * warp_size
         elif uop is UOps.DEFINE_ACC:
-          ul[i] = [[arg[0]] * warp_size for _ in range(dtype.count)] if dtype.count > 1 else [arg[0]] * warp_size
+          ul[i] = [x[:] if isinstance(x, list) else x for x in inp[0]]
```
geohot, line 93: This doesn't look right

geohot:
Hmm yea, I guess I shouldn't expect process replay to pass. It makes this harder to review, but I think it's worth doing. The Python backend stuff doesn't look right at first glance, is it the minimal change you can make?

Qazalin:
process replay only compares the Program.src.

Qazalin (commented):
I don't like that float4 DEFINE_ACC has a float4 const.
Can you make sure CONST is always the scalar of localtype.

tinygrad/runtime/ops_python.py
JhennerTigreros: Changed. Thx.

JhennerTigreros:
> I don't like that float4 DEFINE_ACC has a float4 const. Can you make sure CONST is always the scalar of localtype and in uops type_verify it asserts it. Example of float4 acc: `TestLinearizer.test_grouped_store_phis`

Done. Just use the `scalar()` value of the dtype and move the type check of `DEFINE_ACC` to its own if, this change made IMAGE=2 test start failing, reading the `image_conv2d` to understand what can be failing due to this change.

Qazalin:
From the process replay run I only see ALU reordering which is ok since it's adding a node to the graph.
https://github.com/tinygrad/tinygrad/actions/runs/9636874680/job/26575334937#step:24:452

tinygrad/codegen/uops.py
```diff
@@ -460,8 +460,9 @@ def flops_mem(self, ignore_indexing=False) -> Tuple[sint, sint]:
   def type_verify(self):
     for u in self.uops:
       uop, arg, src, dtype = u.op, u.arg, u.src, u.dtype
-      if uop in {UOps.CONST, UOps.DEFINE_ACC}:
-        if uop is UOps.DEFINE_ACC: arg = arg[0]
+      if uop is UOps.DEFINE_ACC:
+        assert dtype is not None and src[0].dtype is dtype.scalar(), f"type of {src[0].dtype=} must be a scalar {dtype.scalar()}"
+      if uop is UOps.CONST:
         assert dtype is not None and type(arg) is type(dtypes.as_const(arg, dtype)), f"type of {arg=} does not match {dtype}"
```
Qazalin, line 466: The dtypes.as_const assert should still work?

JhennerTigreros: Yes, changed.

geohot:
Good work!

PR #4976, opened 2024-06-15
Nicer UPat [run_process_replay]

uuuvn:

PR #4978, opened 2024-06-15
Incomplete/buggy rule breaks process replay on #4976

uuuvn:
x+-y -> x-y rule didn't cover -y+x which is the same because a + b = b + a and my nicer uops always does that, but because this rule didn't cover that case that it was probably supposed to cover process replay broke

Here's diff:
image
Old rule:  21)))+(-(((val
New rule: val21)))-(((val
etc


Qazalin:
process replay diff: https://github.com/tinygrad/tinygrad/actions/runs/9527939440/job/26265125669#step:23:43

PR #4982, opened 2024-06-15
only fold 0*x if x is not float

chenyuxyz:
not correct if x is inf, which breaks pow backward, and return 0 is wrong. wip something something UNMUL failed now...


geohot:
I don't like all the compute we waste because of inf...I wonder where it's actually used and what people want to express with it, it's not a real value.

PR #4984, opened 2024-06-15
hcq update waits and signals in place

nimlgen:
Do not rebuild queues for syncing, update signals/waits in place.

```
enqueue in   3.81 ms
total   9.16 ms, 109.13 tok/s, 1525.43 GB/s, param 1471.83 GB/s
 male
```


nimlgen:
Thinking if we want to support/issue commands for mailboxes. We currently limited usage of them due to high cost of calling a syscall, so maybe we want to remove them?

geohot:
What are mailboxes and why do we need them? Are they on both AMD and NVIDIA or just one? Is normal CUDA runtime using them?

PR #4985, opened 2024-06-15
Float in scaled dot product attention

Eliulm:
It seems that the `q @ k` in `scaled_dot_product_attention` produces `-inf` after some time. Casting to `fp32` up to the softmax seems to fix the issue. It [trained](https://wandb.ai/wahl/MLPerf-BERT/runs/47z8vcpm?nw=nwuserwahl) past the critical step range, but is not done yet


examples/mlperf/initializers.py
```diff
@@ -67,4 +65,12 @@ def __call__(self, x:Tensor):
...
-    return (xn * self.weight.cast(dtypes.default_float) + self.bias.cast(dtypes.default_float))
\ No newline at end of file
+    return (xn * self.weight.cast(dtypes.default_float) + self.bias.cast(dtypes.default_float))
+
+def scaled_dot_product_attention_bert(query:Tensor, key:Tensor, value:Tensor, attn_mask:Tensor=None, dropout_p:float=0.0, is_causal:bool=False):
+  assert all_int(query.shape), f"does not support symbolic shape {query.shape}"
+  if is_causal: attn_mask = Tensor.ones(query.shape[-2], key.shape[-2], requires_grad=False, device=query.device).tril(0).cast(dtypes.bool)
+  if attn_mask is not None and attn_mask.dtype == dtypes.bool: attn_mask = (attn_mask == 0).where(-float("inf"), 0)
+  # NOTE: During training we get -inf from key @ query, so we cast to float
+  qk = query.cast(dtypes.float32) @ key.cast(dtypes.float32).transpose(-2,-1) / math.sqrt(query.shape[-1])
```
chenyuxyz, line 75: you can matmul `query` and `key` with `acc_dtype=dtypes.float32` to get float32 from matmul, I think we want to patch the core version instead of adding a separate one.

chenyuxyz (commented):
cool! can you also run a run that has 200k steps?

examples/mlperf/initializers.py
chenyuxyz: also not sure what triggered the nan, even if there are `-inf` into softmax, as long as there's one finite number it should be fine.

Eliulm: @chenyuxyz Yes, but the infs are one full row that we softmax over.

Eliulm:
> cool! can you also run a run that has 200k steps?

https://wandb.ai/wahl/MLPerf-BERT/runs/wdro2fpi?nw=nwuserwahl

examples/mlperf/initializers.py
Eliulm: Ok, so I have figured out what is going on. The `inf`s are a result of under and overflow which itself is not an issue. However, if there are enough `inf` to fill a full row over the softmax dimension after that, we get `NaN`. That means that `acc_dtype=dtypes.float32` wont fix it unless we add some flag that allows us to do the matmul without casting to lower precision right after the accumulation.

chenyuxyz: oh i can make dot return in acc_dtype if it's specified

chenyuxyz: added in #4994, now if acc_dtype is explicitly set, it does not downcast

tinygrad/tensor.py
```diff
@@ -2700,8 +2700,8 @@ def scaled_dot_product_attention(self, key:Tensor, value:Tensor, attn_mask:Optio
     assert all_int(self.shape), f"does not support symbolic shape {self.shape}"
     if is_causal: attn_mask = Tensor.ones(self.shape[-2], key.shape[-2], requires_grad=False, device=self.device).tril(0).cast(dtypes.bool)
     if attn_mask is not None and attn_mask.dtype == dtypes.bool: attn_mask = (attn_mask == 0).where(-float("inf"), 0)
-    qk = self @ key.transpose(-2,-1) / math.sqrt(self.shape[-1])
-    return ((qk+attn_mask) if attn_mask is not None else qk).softmax(-1).dropout(dropout_p) @ value
+    qk = self.matmul(key.transpose(-2,-1), acc_dtype=dtypes.float32) / math.sqrt(self.shape[-1])
```
chenyuxyz, line 2703: `acc_dtype=least_upper_dtype(self.dtype, key.dtype, dtypes.float32)` so that float64 is not downcasted.

chenyuxyz (commented):
cool, can you start a run with this change? want to confirm it trains

Eliulm:
new run at https://wandb.ai/wahl/MLPerf-BERT?nw=nwuserwahl

PR #4998, opened 2024-06-16
fix einsum output str

raymond-van:
Found this small bug on the einsum output string while implementing ellipsis support for einsum.

Bug occurs when there's no "->" in the formula (so empty rhs) and multiple input operands

In this case output string should be sorted and repeated characters should be summed over (removed)


chenyuxyz:
Thanks!

PR #5000, opened 2024-06-16
add GPT2 half/half+beam to AMD

geohot:
Also all three should test winograd CIFAR


geohot:
@wozeparrot GPT2 HALF and BEAM should be on all now, and so should CIFAR WINO

PR #5004, opened 2024-06-17
Fix/opencl Better error Messages

GabrielZCode:
Better openCL error messages!! 

Using the same strategy as generate_nv() function in generate_stubs.sh , I've extracted the error messages from https://github.com/KhronosGroup/OpenCL-Headers/tree/main opencl Headers . --> generate_opencl_error_codes() at the same bash script with a new option   ./autogen_stubs.sh opencl-error-code   

Now by calling the check function on tinygrad/runtime/ops_gpu.py we get the original Error message from openCL , getting it by the statusCode .

The statusCode dict is located at tinygrad/runtime/autogen/opencl.py


tinygrad/runtime/ops_gpu.py
```diff
@@ -5,12 +5,14 @@
...
 from tinygrad.device import BufferOptions, LRUAllocator, Compiled, Compiler, CompileError
+from tinygrad.runtime.autogen.opencl import opencl_status_codes
+
 
 # see test/external/external_osx_profiling.py to determine this ratio. it's in like GPU clocks or something
 OSX_TIMING_RATIO = (125/3) if OSX else 1.0
 
 def check(status):
-  if status != 0: raise RuntimeError(f"OpenCL Error {status}")
+    if status != 0: raise RuntimeError(f"OpenCL Error {opencl_status_codes.get(status, 'Unkown Error')}")
```
nimlgen, line 15: spacing

.gitignore
```diff
@@ -53,3 +53,9 @@ weights
 *.lprof
 *.pkl
 site/
+.clang-format
+.tool-versions
+setup_extra.py
```
nimlgen, line 58: unrelated change, why it's here?

autogen_stubs.sh
```diff
@@ -22,12 +22,43 @@ fixup() {
   grep FIXME_STUB $1 || true
 }
 
+generate_opencl_error_codes(){
```
nimlgen, line 25: no need to git clone, do the same way as it's done in `generate_opencl`. I think you can add just `cl_ext.h` there and parse errors. Also errors are negative values, can we just build a dict Dict[error int -> str] in ops_gpu with a one-liner?

.gitignore
GabrielZCode: it's just some local files I don't want to push, if not in .gitignore , how should i make git ignore it

autogen_stubs.sh
GabrielZCode: are the other errors not important so i can just get what is present on cl_ext.h.
There's errors on cl.h, cl_egl.h, cl_dx9_media_sharing.h, cl_d3d11.h and cl_d3d10.h files.

About the one-liner, it's possible , it'll just not be genned by the bash script

GabrielZCode:
I've tried other ways (looking for what symbols are exported from the openCL.so file , but the errors wasn't there), but using sed for extracting the errors was the simplest solution i've found, if anyone gets a better idea, i have no problems to implement it, but right now i don't have any.

autogen_stubs.sh
GabrielZCode: "do the same way as it's done in generate_opencl", you mean, just assume the header file it's already available on the local directory?, I don't get how that's better than git clone if the dir doesn't exist. OpenCL headers are not default to be present on all linux distros.

GabrielZCode: cl_ext only have the -1141 error code, cl.h have 72 error codes

GabrielZCode: just a note, there's two error codes that are not negative values, they are hexadecimal [0x40D4] [0x40D5]

tinygrad/runtime/ops_gpu.py
```diff
@@ -5,12 +5,14 @@
...
 from tinygrad.device import BufferOptions, LRUAllocator, Compiled, Compiler, CompileError
+from tinygrad.runtime.autogen.opencl import opencl_status_codes
+
 
 # see test/external/external_osx_profiling.py to determine this ratio. it's in like GPU clocks or something
 OSX_TIMING_RATIO = (125/3) if OSX else 1.0
 
 def check(status):
-  if status != 0: raise RuntimeError(f"OpenCL Error {status}")
+  if status != 0: raise RuntimeError(f"OpenCL Error {opencl_status_codes.get(status, 'Unkown Error')}")
```
nimlgen, line 15: typo in `Unknown`

GabrielZCode: fixed

autogen_stubs.sh
GabrielZCode: @nimlgen can you help me on the questions above? Any other thing missing for the PR merge?

.gitignore
geohot: This is not resolved

GabrielZCode: ok. fixed

tinygrad/runtime/ops_gpu.py
```diff
@@ -5,12 +5,14 @@
...
 from tinygrad.device import BufferOptions, LRUAllocator, Compiled, Compiler, CompileError
+from tinygrad.runtime.autogen.opencl import opencl_status_codes
+
 
 # see test/external/external_osx_profiling.py to determine this ratio. it's in like GPU clocks or something
 OSX_TIMING_RATIO = (125/3) if OSX else 1.0
 
 def check(status):
-  if status != 0: raise RuntimeError(f"OpenCL Error {status}")
+  if status != 0: raise RuntimeError(f"OpenCL Error {opencl_status_codes.get(status, 'Unknown Error')}")
```
chenyuxyz, line 15: i think it at least should print the status code in the unknown case.

GabrielZCode: ok. done

GabrielZCode: it'll print in all cases, it's a good information to have

GabrielZCode:
Simplified a lot the gen_stubs function, now it's just two seds and two cats commands, the rest is the git clone check (only run if the directory does not exist), the 100 lines is the error dict in opencl.py

autogen_stubs.sh
GabrielZCode: @nimlgen i think i got your idea, did a simplification of the seds

tinygrad/runtime/autogen/opencl.py
```diff
@@ -1793,3 +1793,107 @@ class struct__cl_name_version(Structure):
     'struct__cl_kernel', 'struct__cl_mem', 'struct__cl_name_version',
     'struct__cl_platform_id', 'struct__cl_program',
     'struct__cl_sampler', 'union__cl_image_desc_0']
+opencl_status_codes = {}
+opencl_status_codes[-1002] = "CL_INVALID_D3D10_DEVICE_KHR                         "
```
nimlgen, line 1797: all strings have a lot of spaces at the end

autogen_stubs.sh
```diff
@@ -168,6 +195,7 @@ generate_hsa() {
 }
 
 if [ "$1" == "opencl" ]; then generate_opencl
+elif [ "$1" == "opencl-error-code" ]; then generate_opencl_error_codes
```
nimlgen, line 198: move into opencl

autogen_stubs.sh
```diff
@@ -22,12 +22,38 @@ fixup() {
...
+  if [ ! -d "$OPENCL_HEADERS_SRC" ]; then
+    git clone https://github.com/KhronosGroup/OpenCL-Headers.git $OPENCL_HEADERS_SRC
+    pushd .
+    cd $OPENCL_HEADERS_SRC
+    git reset --hard $OPENCL_COMMIT_HASH
+    popd
+  fi
+  
+  file=$OPENCL_HEADERS_SRC/opencl_headers_coalesce.h
+  cat $OPENCL_HEADERS_SRC/CL/*.h > $file # Form a single file from all headers, to extract all errors present
```
nimlgen, line 38: in `opencl` we take headers from `/usr/include/CL`, so let's not change this *partially* in this pr, can you take all needed headers from `/usr/include/CL/`?

GabrielZCode: So, I think that's a distro compatibility problem, for example, I'm running openSuse and /usr/include doesn't even have a CL folder, of course I can fix it and put it myself, but other developers might run on the same issue.

If that's acceptable , i can comment it (for reference) and set the OPENCL_HEADERS_SRC=/usr/include/CL .

Should i keep it as is or force the /usr/include/CL usage?

autogen_stubs.sh
GabrielZCode: done

tinygrad/runtime/autogen/opencl.py
GabrielZCode: fixed, but it's in the fork now that the PR is closed

PR #5007, opened 2024-06-17
Convert a bunch more rules [run_process_replay]

uuuvn:


tinygrad/codegen/uops.py
```diff
@@ -206,7 +209,7 @@ def loop_collapse(loop_start, loop_end, compval, idx, mval, multconst):
   (UOp.const(dtypes.bool, True).lt(UOp.var()), lambda: UOp.const(dtypes.bool, False)),
   # a conditional with the same results either way is a noop, also fold const conditionals
   (UOp.alu(TernaryOps.WHERE, UOp.var(), UOp.var("val"), UOp.var("val")), lambda val: val),
-  (UOp.alu(TernaryOps.WHERE, UOp.cvar('gate'), UOp.var('true'), UOp.var('false')), lambda gate, true, false: true if gate.arg else false),
+  (UOp.alu(TernaryOps.WHERE, UOp.cvar('gate'), UOp.var('c0'), UOp.var('c1')), lambda gate, c0, c1: c0 if gate.arg else c1),
```
Qazalin, line 212: i think true/false is more readable - why change it?

uuuvn: i was comparing generated UPats with compiled ones to be extra sure everything is one to one correct, forgot to revert this thing thought

uuuvn: old ones had c0 c1 so i just made it temporarily back for my local for each rule compile and see if it's the same module permutations of arrays into less clear c0 c1 and forgot to change back

PR #5009, opened 2024-06-17
[run_process_replay] for opencl/openpilot

Qazalin:
An unknown setting in codegen/* is generating a different kernel from the expected openpilot kernel src

```py
import pickle, os
from tinygrad.device import Device

if "FLOAT16" not in os.environ: os.environ["FLOAT16"] = "0"
if "IMAGE" not in os.environ: os.environ["IMAGE"] = "2"
if "NOLOCALS" not in os.environ: os.environ["NOLOCALS"] = "1"
Device.DEFAULT = "GPU"
k, src = pickle.load(open(fetch("https://github.com/Qazalin/containers/raw/test/k"), "rb"))
k2 = k.copy()
prg = k2.to_program()
assert k2.ast == k.ast
assert k2.applied_opts == k.applied_opts
assert k2.opts == k.opts
assert k2.name == k.name
assert len(prg.uops._uops) == len(k.uops._uops), f"expected {len(k.uops._uops)} uops, got {len(prg.uops._uops)}"
assert prg.src == k.to_program().src

```


Qazalin:
yay, Lowerer fixed this!

PR #5012, opened 2024-06-17
Tensor.sum returns in acc_dtype if specified

chenyuxyz:


chenyuxyz:
need to think about how to do overflow half to inf on PYTHON, but it is not supported on master either. so marked as TODO for now to get this in

PR #5019, opened 2024-06-17
tiny toposort library [run_process_replay]

Qazalin:


tinygrad/codegen/uops.py
```diff
@@ -328,23 +328,15 @@ def graph_dedup(self, sink):
...
+      all_nodes[n], early_in_degree[n] = None, 0
       for x in n.vin:
         early_in_degree[n] += 1
         children[x].append(n)
       unprocessed_nodes += list(n.vin)
-    early_queue = [x for x in all_nodes if early_in_degree[x] == 0]
-    replace_nodes: Dict[UOp, UOp] = {}
-    while len(early_queue):
-      n = early_queue.pop(0)
-      if n in replace_nodes: continue
```
Qazalin, line 340: huh, I think this does nothing -
green https://github.com/tinygrad/tinygrad/commit/64eecd27e7f65ac3ff91e0ee0981e10592c3d8b4 in CI https://github.com/tinygrad/tinygrad/actions/runs/9554614732

tinygrad/helpers.py
```diff
@@ -76,6 +77,19 @@ def temp(x:str) -> str: return (pathlib.Path(tempfile.gettempdir()) / x).as_posi
 
 class GraphException(Exception): pass
 
+V = TypeVar("V")
+def bfs(graph:DefaultDict[T, List[T]], in_degree:DefaultDict[T, int], insert:Callable[[T, Dict[T, V]], None]) -> Dict[T, V]:
```
chenyuxyz, line 81: isn't `in_degree` a function of `graph`, what if the two inputs disagree?

also feel this is a lot less flexible, like it's more difficult to experiment with a different traversing order

Qazalin: I think it's worth the perf cost to not have to loop through the entire graph - Though I can make in_degree optional.
Will include cycle assert to make this safer.

Yeah the goal I started with was to library BFS specifically, rather than flexibility focusing on removing repetitiveness.

PR #5024, opened 2024-06-18
 RECIP(-0.0) should be -inf

chenyuxyz:
added test_dtype_alu for PYTHON backend, everything passed except a inf to int cast. Also skipped some uint tests that numpy changed behavior and threw errors in 2.0.


chenyuxyz:
deleted all cache in my fork and ran the test https://github.com/chenyuxyz/tinygrad/pull/291/checks

PR #5029, opened 2024-06-18
Rename UOp.uop and vin to op and src

DKormann:


geohot:
Don't change code in extra

DKormann:
okay sure how is this supposed to work?

DKormann:
what do you think about renaming UPat.uop and vin also for clarity?

PR #5030, opened 2024-06-18
Better default hparams for large BS

Eliulm:
For large BS, the current LR and total steps won't work. LR scaling is derived from the [reference](https://github.com/mlcommons/logging/blob/master/mlperf_logging/rcp_checker/training_4.0.0/rcps_bert.json):

### DIFF
--- a/examples/mlperf/model_train.py
+++ b/examples/mlperf/model_train.py
@@ -398,7 +398,7 @@ def train_bert():
   seed = config["seed"] = getenv("SEED", 12345)
 
   # ** hyperparameters **
-  BS                 = config["GLOBAL_BATCH_SIZE"]      = getenv("BS", 16 * len(GPUS) if dtypes.default_float in {dtypes.float16, dtypes.bfloat16} else 8 * len(GPUS))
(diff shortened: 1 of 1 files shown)


examples/mlperf/model_train.py
```diff
@@ -398,11 +398,11 @@ def train_bert():
   seed = config["seed"] = getenv("SEED", 12345)
 
   # ** hyperparameters **
-  BS                 = config["GLOBAL_BATCH_SIZE"]      = getenv("BS", 4 * len(GPUS)) # FP32 4090: 6 GPUS -> BS24
-  EVAL_BS            = config["EVAL_BS"]                = getenv("EVAL_BS", 4 * len(GPUS))
-  max_lr             = config["OPT_BASE_LEARNING_RATE"] = getenv("OPT_BASE_LEARNING_RATE", 0.000004166 * BS)
+  BS                 = config["GLOBAL_BATCH_SIZE"]      = getenv("BS", 16 * len(GPUS) if dtypes.default_float in {dtypes.float16, dtypes.bfloat16} else 8 * len(GPUS))
```
chenyuxyz, line 401: i did a microbenchmark the other day and constructing small sets like this is slower. so i have slower converting stuff to just use tuple https://github.com/tinygrad/tinygrad/pull/5012/files for example

PR #5031, opened 2024-06-18
rename uop [run_process_replay] 

DKormann:
not changing extra.

renaming UPat as well otherwise there are collisions

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,9 +43,9 @@ repos:
         language: system
         always_run: true
         pass_filenames: false
-      - id: pylint
-        name: pylint
-        entry: env PYTHONPATH="." python3 -m pylint tinygrad/
-        language: system
-        always_run: true
-        pass_filenames: false
+      # - id: pylint
+      #   name: pylint
+      #   entry: env PYTHONPATH="." python3 -m pylint tinygrad/
+      #   language: system
+      #   always_run: true
+      #   pass_filenames: false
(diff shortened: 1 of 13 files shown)


geohot:
btw, process replay should work on this

.pre-commit-config.yaml
```diff
@@ -43,9 +43,9 @@ repos:
         language: system
         always_run: true
         pass_filenames: false
-      - id: pylint
-        name: pylint
-        entry: env PYTHONPATH="." python3 -m pylint tinygrad/
-        language: system
-        always_run: true
-        pass_filenames: false
+      # - id: pylint
```
Qazalin, line 46: ?

tinygrad/renderer/cstyle.py
```diff
@@ -104,10 +104,10 @@ def ssa(prefix:str, u:Optional[UOp]=None):
       c[prefix] += 1
       return ret
 
-    child_count = Counter(v for ru in uops for v in ru.vin)
+    child_count = Counter(v for ru in uops for v in ru.src)
 
     for u in uops:
-      uop,dtype,vin,args = u.uop,u.dtype,u.vin,u.arg
+      uop,dtype,vin,args = u.op,u.dtype,u.src,u.arg
```
Qazalin, line 110: why is cstyle sitll using uop and vin

geohot: I think renaming that is a bigger change than it needs to be. The smaller you can make PRs the better.

DKormann:
would make new pr for vin-src

PR #5032, opened 2024-06-18
contiguous instead of realize in threefry counts1

chenyuxyz:


chenyuxyz:
this is incorrect because _rng_counter has realize anyway

PR #5033, opened 2024-06-18
linearizer: change globals to merge into left axis/gridDims.x first

flammit:
also fixes Variable max size to be correct and add docs for the off parameter

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -627,12 +627,13 @@ class TestLinearizer(unittest.TestCase):
     _assert_grouped_dims("gidx", (2,3,4,5,), (32,16,16,), True, [20,3,2])
     _assert_grouped_dims("gidx", (Variable("start_pos",1,2),3,4,5,), (32,16,16,), True, [20,3,Variable("start_pos",1,2)])
 
-    # TODO: support sint collapse
-    # _assert_grouped_dims("gidx", (Variable("start_pos",1,2),3,4,5,), (16,16,16,), False, [Variable("start_pos",1,2)*3,4,5])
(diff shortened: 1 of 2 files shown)


test/test_linearizer.py
```diff
@@ -628,22 +628,26 @@ def _assert_grouped_dims(prefix, dims, max_sizes, reverse_dims, expected_sizes):
...
 
+    with self.assertRaises(AssertionError): # dim too large and not factorable
+      get_grouped_dims("gidx", 0, (128,3,4), (16,4,23,), False,)
+
     with self.assertRaises(AssertionError): # too large for sizes
       get_grouped_dims("gidx", 0, (2,3,4,5,6), (16,16,16,), False,)
 
     with self.assertRaises(AssertionError): # variable too large
-      get_grouped_dims("gidx", 0, (Variable("start_pos", 0, 17),3,4), (16,16,16,), False,)
+      get_grouped_dims("gidx", 0, (Variable("start_pos", 0, 16),3,4), (16,16,16,), False,)
```
chenyuxyz, line 650: can you add some cases that has Variable and collapsed?

chenyuxyz: oh i see `TODO: support sint properly`, maybe it's not supported yet?

flammit: Oh right it's not supported yet, but I added a test case that passes and one that should when it is.

chenyuxyz:
i think it's fine to merge this, generally i feel this portion can be made simpler

also interested in benchmarking some kernels that hit this to know which way of folding is more performant

PR #5035, opened 2024-06-18
io_uring for copies from disk

nimlgen:
Some early io_uring experiments

disk_read_speed.py:
| | import io | iouring |
|---|---|---|
amd | 9.93 GB/s | 14.73 GB/s |
nv | 9.54 GB/s | 15.36 GB/s |

llama 70b ~9.5gb/s for now


nimlgen:
Do not if we want to use io_uring, but raw performance (disk->gpu) using it is about 18-20gb/s. Will test any alternatives, like aio, maybe it will be a simpler api or just syscalls with the same performance.

geohot:
I want it to be async, but if normal aio async is the same speed that's fine.

nimlgen:
aio is only ~11-12gb/s, so uring is the fastest. Will cleanup and merge uring.

nimlgen:
@geohot any comments/suggestions?

I have removed calling liburing and just doing some mmap+syscalls now to not have this dependency. Maybe this adds some lines, but I think it's worth it.

tinygrad/runtime/ops_disk.py
```diff
@@ -58,3 +100,25 @@ def _might_close(self):
     if self.count == 0:
       if hasattr(self, 'fd'): os.close(self.fd)
       self.size = None
+  def _iouring_setup(self):
+    DiskDevice._tried_io_uring_init = True
+
+    if platform.system() != 'Linux': return
+
+    fd = libc.syscall(425, 4096, ctypes.byref(p:=io_uring.struct_io_uring_params()))
```
geohot, line 108: What's 425?

geohot:
Very readable!

`copy_from_disk` is clearly copy paste between NV and AMD, but maybe this is part of a bigger refactor to share?

test/unit/test_disk_tensor.py
```diff
@@ -302,5 +302,43 @@ def test_bf16_disk_write_read(self):
     ct = t.llvm_bf16_cast(dtypes.float)
     assert ct.numpy().tolist() == [9984., -1, -1000, -9984, 20]
 
+  @unittest.skipUnless(Device.DEFAULT in {"NV", "AMD"}, "only AMD and NV supports ")
```
geohot, line 305: These tests should work everywhere, right? Aside from checking if io_uring is there it's just a normal copy.

tinygrad/runtime/ops_disk.py
```diff
@@ -29,10 +34,47 @@ def copyout(self, dest:memoryview, src:DiskBuffer):
...
+        # Prepare sqe
+        sqe_index = (tail:=DiskDevice.io_uring.sq.ktail[0]) & DiskDevice.io_uring.sq.kring_mask[0]
+        sqe = DiskDevice.io_uring.sq.sqes[sqe_index]
+        sqe.opcode, sqe.fd, sqe.off = io_uring.IORING_OP_READ, self.device.fd, fd_offset + next_read_offset
+        sqe.addr, sqe.len, sqe.user_data = copy_batch[0], min(seg_len, total_copy_size - next_read_offset), len(reqs)
+
+        # Send sqe
+        DiskDevice.io_uring.sq.array[sqe_index] = sqe_index
+        DiskDevice.io_uring.sq.ktail[0] = tail + 1
+        libc.syscall(426, DiskDevice.io_uring.ring_fd, 1, 1, (1 << 0)) # IORING_ENTER_GETEVENTS
```
geohot, line 56: I see a comment, but a const is always better.

PR #5039, opened 2024-06-18
Resnet load

reddyn12:
Fixed test name


reddyn12:
Was the problem with the test name or the fix itself?

PR #5047, opened 2024-06-19
Speed up _match function [run_process_replay]

abgita:
- Prevent extra checks when `pat.name` is `None`
- Exit early on mismatch

Before:
```bash
UOPS_REWRITE=1 PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  42.78 ms
***** model schedule in  13.86 ms
***** model lower in 1664.71 ms
```

After:
```bash
UOPS_REWRITE=1 PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  42.40 ms
***** model schedule in  13.99 ms
***** model lower in 1523.17 ms
```


geohot:
Minor speedup for more lines/complexity isn't worth it. The whole algorithm is wrong.

PR #5050, opened 2024-06-19
potential bug in hcq graph for multi gpu

chenyuxyz:
was testing this scheduler enqueue order because it made single gpu llm.c faster. I would expect any topological order works.

But training resnet with JIT=1, the loss is not going down. It has faster step time, maybe some kernels are not executed properly.
JIT=2 seems fine, a lot slower but loss is going down.

JIT=1 https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/iosm77yr?nw=nwuserchenyuxyz
JIT=2 https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/tfgrgsc9?nw=nwuserchenyuxyz

@nimlgen fyi. I can test with CUDA / NV tomorrow to isolate the issue further.


chenyuxyz:
cifar on 6 gpus trained fine though...

nimlgen:
Will take a look

chenyuxyz:
more likely to be bug in our graph
`CUDA=1` trained https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/auk9ob8e?nw=nwuserchenyuxyz
`NV=1` had the same loss issue https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/d7rp83d0/workspace?nw=nwuserchenyuxyz

chenyuxyz:
i think it's fixed. also on master now NV trains resnet at similar speed to CUDA

PR #5055, opened 2024-06-19
add annotations [run_process_replay]

DKormann:


tinygrad/codegen/uops.py
```diff
@@ -289,7 +289,7 @@ def print(self):
     for i,u in enumerate(self):
       print(f"{i:4d} {str(u.op):20s}: {str(u.dtype) if u.dtype is not None else '':25s} " f"{str([self.uops.index(x) for x in u.src]):32s} {u.arg}")
 
-  def graph_rewrite(self, sink, pm):
+  def graph_rewrite(self, sink: UOp, pm: PatternMatcher):
```
chenyuxyz, line 292: we don't have a strict style guide, but i'd remove the spaces in between like other functions in this file

geohot: Is there a way to have the linter check this?

DKormann: ok

DKormann: is this  a rule for only type annotations or also eg. lambda function

geohot:
I don't think we should change any until there's a linter rule.

PR #5056, opened 2024-06-19
Revert "UPat for has_valid in load/store [run_process_replay] (#5052)"

Qazalin:
I think as a prereq for this need to first work on a fast pattern matcher.

before pattern matching valids:
```
***** model lower in 2136.13 ms
```
after:
```
***** model lower in 2359.20 ms
```


Qazalin:
will fix the process replay checker and get this merged

geohot:
You can leave the matching rules but also have the Linearizer do it. Matching rule will hit if there's a const folding or something.

Issue #5051, opened 2024-06-19
bf16 cifar on AMD benchmark has nan loss


chenyuxyz:
bisected to this https://github.com/tinygrad/tinygrad/actions/runs/9435797854/job/25995095374

this one is fine https://github.com/tinygrad/tinygrad/actions/runs/9435244478/job/25988401090

chenyuxyz:
`UOPS_REWRITE=0` seems fine, so some rules caused this
```
time UOPS_REWRITE=0 STEPS=10 DEFAULT_FLOAT=BFLOAT16 python3 examples/hlb_cifar10.py
shuffling training dataset in 4422.45 ms (epoch=0)
  0 51481.65 ms run, 51481.25 ms python,    0.39 ms AMD, 1190.91 loss, 0.001495 LR, 0.65 GB used,     13.15 GFLOPS,    676.80 GOPS
  1 28914.66 ms run, 28914.27 ms python,    0.39 ms AMD, 1190.34 loss, 0.002991 LR, 2.63 GB used,     23.35 GFLOPS,    675.19 GOPS
  2  634.88 ms run,    2.27 ms python,  632.61 ms AMD, 1696.11 loss, 0.003128 LR, 2.63 GB used,   1063.49 GFLOPS,    675.19 GOPS
  3  629.06 ms run,    1.95 ms python,  627.11 ms AMD, 2270.43 loss, 0.002686 LR, 2.63 GB used,   1073.33 GFLOPS,    675.19 GOPS
  4  625.82 ms run,    1.91 ms python,  623.91 ms AMD, 1525.46 loss, 0.002258 LR, 2.63 GB used,   1078.88 GFLOPS,    675.19 GOPS
  5  627.64 ms run,    1.90 ms python,  625.74 ms AMD, 1539.48 loss, 0.001823 LR, 2.63 GB used,   1075.76 GFLOPS,    675.19 GOPS
  6  629.40 ms run,    1.91 ms python,  627.49 ms AMD, 1382.03 loss, 0.001389 LR, 2.63 GB used,   1072.75 GFLOPS,    675.19 GOPS
  7  627.81 ms run,    1.91 ms python,  625.91 ms AMD, 1321.19 loss, 0.000961 LR, 2.63 GB used,   1075.47 GFLOPS,    675.19 GOPS
[...]

PR #5057, opened 2024-06-19
commenting out this rewrite rule fixed bf16 cifar on red

chenyuxyz:
#5051, but why?


nimlgen:
Let's hold this, since it might be backend related. Just noticed that with `UOPS_REWRITE=0` loss is a bit different from run to run.

chenyuxyz:
seems to be fixed

PR #5059, opened 2024-06-19
type annos in uops [run_process_replay]

DKormann:


tinygrad/codegen/uops.py
```diff
@@ -363,8 +363,8 @@ def linearize(self, extra_pm:Optional[PatternMatcher]=None, type_verify=True):
     # BFS toposort
     graph: DefaultDict[UOp, List[UOp]] = defaultdict(list)
     in_degree: DefaultDict[UOp, int] = defaultdict(int)
-    loops = []
-    ifs = []
+    loops:List[UOp] = []
```
Qazalin, line 366: did you check the line just above this to see how we annotate types?
That rule only applies to function params.

DKormann: you are right
would be nice to have a definitive ruleset though
personally would advocate that every colon is followed by space or newline that would also be easiest to add to linter right

PR #5065, opened 2024-06-20
minor cleanup getting output shape of _pool

chenyuxyz:
math.ceil makes intent clear, also same formula works in both cases [run_process_replay]


chenyuxyz:
intersting non-deterministic replay

PR #5066, opened 2024-06-20
remove set in function.Flip getting arg

chenyuxyz:
contructing set for every i is expensive and unnecessary.

```
print(timeit.timeit("a = tuple([-1 if i in set((0, 2, 3)) else 1 for i in range(7)])"))
print(timeit.timeit("a = tuple([-1 if i in (0, 2, 3) else 1 for i in range(7)])"))
```
```
0.8064187910640612
0.34930695802904665
```


chenyuxyz:
#5068 solved the root cause

PR #5072, opened 2024-06-20
fix check_process_replay for special characters (test)

Qazalin:


chenyuxyz:
works now, thanks for fixing this quickly!

Issue #5074, opened 2024-06-20
hash and eq for UOp: proposal / clarification request


DKormann:
to make _match and rewrite graph simpler would like to make UOp have a hash function based on its components.
probably would need to make it frozen

is this a bad idea? was reading related issues but maybe not enough

DKormann:
if I understand correctly UOp.cmp_tuple is only existing to make > < work deterministically which is needed for deterministic ordering?
could replace that
might save a chunk of lines

DKormann:
could make hash of switched src be the same if op is commutative

Qazalin:
so in general hash is very slow - Have you benchmarked how hash would work?
If you freeze the UOp you should also create a new class from scratch every time, how does that impact rewrite perf?

DKormann:
as long as you cache your hash should be fine?
creating UOp from scratch also in my experience super cheap.

however i am also seeing more clearly the advantages now of tinygrad used uops until now with hash and eq being tied essentially to the object.

PR #5075, opened 2024-06-20
add UOp.__hash__ [run_process_replay]

DKormann:
trying to implement #5074


DKormann:
this makes graph rewrite slower for now but will save lines.
i think might lead to speedup eventually

DKormann:
more easy savings to come!

DKormann:
moved to new pr. dont need hash here for now

PR #5076, opened 2024-06-21
Add bottom up graph rewrite 

mesozoic-egg:
Depth first search with post order rewrite; when a node is rewritten, its parent src is replaced, this rewritten node is then traversed again before processing its parent node. 

This reduces the node visit count by almost half, got an overall speed up of by 10%, and 50% speed up for the graph_rewrite function.

There's still some correctness issue to fix, because some patterns become unmatched if the children get rewritten. For example

```
(UPat(UOps.ALU, TernaryOps.WHERE, src=(
  UPat(UOps.ALU, BinaryOps.CMPLT, src=(
    UPat(UOps.ALU, BinaryOps.ADD, src=[
      UPat(name="idx"), 
      UPat(UOps.ALU, BinaryOps.MUL, src=[
        UPat(UOps.CONST, name="mval"), 
        UPat(UOps.RANGE, src=(
          UPat(name="loop_start"), 
          UPat(name="loop_end")
        ))
      ])
    ]),
    UPat(UOps.CONST, name="compval"))), 
  UPat(UOps.CONST, name="multconst"), UPat(UOps.CONST, 0)
)), loop_collapse),
``` 
will not match if the CONST * RANGE got rewritten because of the `(-1*UOp.var('x'), lambda x: -x)` rule. Will go through these carefully.

Here's the node visit count comparison for an arange UOP tree
arange_uops

And the speed up for running external_benchmark_schedule.py

speed_overall
speed_graphrewrite

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -320,6 +320,31 @@ class UOpGraph:
       assert run_cnt < 100, "exceeded 100 rewrite loops!"
     return sink
 
+  def graph_rewrite_bottomup(self, sink: UOp, pm: PatternMatcher):
+    @functools.lru_cache
+    def rewrite(_uop: UOp):
+      rewritten = None
+      _rewritten = pm.rewrite(_uop)
+      while _rewritten:
+        rewritten = _rewritten
+        _rewritten = pm.rewrite(rewritten)
+      return rewritten
+
+    @functools.lru_cache
+    def traverse(_uop: UOp, parent: UOp, i):
+      for _i, src in enumerate(_uop.src):
+        traverse(src, _uop, _i)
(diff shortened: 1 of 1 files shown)


geohot:
I think it's simpler than this. What if you just toposort the nodes then walk down the list?

DKormann:
> I think it's simpler than this. What if you just toposort the nodes then walk down the list?

after applying a pattern you also wanna go back to check if children became able to change right?

mesozoic-egg:
Wrote the toposort list version in #5079 but i think it suffers from the backtracking issue, unless I modify the list upon rewrite and resume loop aftewards

mesozoic-egg:
work is continued in PR 5079, closing this

PR #5078, opened 2024-06-21
simplify graph rewrite + speedup [run_process_replay]

DKormann:


DKormann:
external_benchmark_schedule:
in master we call _match 582762 times now reduced to 350357

tinygrad/codegen/uops.py
```diff
@@ -294,31 +294,19 @@ def print(self):
...
-      sink = rewrite(sink)
-      run_cnt += 1
-      assert run_cnt < 100, "exceeded 100 rewrite loops!"
-    return sink
+    if not getenv("UOPS_REWRITE", 1): return sink
+    def rewrite(u:UOp) -> UOp:
+      if (up := u) in cache: return cache[up]
+      for _ in range(100):
+        if (rewritten := pm.rewrite(up)) is None:
+          up = UOp(up.op, up.dtype, tuple(rewrite(x) for x in up.src), up.arg)
```
Qazalin, line 302: is recreating UOp a required change for your cache to work?
this diff is rewriting the retry loops, removing lru, recreating UOp instead of overriding,

DKormann: on the one side i think this is simpler faster and correct

but i still need to think about it deeper

PR #5079, opened 2024-06-21
Graph rewrite bottomup without backtracking

mesozoic-egg:
This is the toposort and walk down the list approach, without backtracking - constrasting #5076 . The example arange loop is not fully rewritten because the new UNMUL being inserted wasn't in the list after the first traversal. 

image

It is possible to insert element to the list when a node is rewritten, then resume the for loop from the newly inserted position though (same as ##5076 but with a loop)


mesozoic-egg:
This will work if we constrain the pattern such that a rewritten node cannot be further simplified though (currently the loop collapse rewrite return a node that can further be re-written

geohot:
So there's two caveats to a single pass:
* You can run in a loop on each node, it's not just 1 rule application per node.
* All newly inserted nodes need to have the simplify run on them.

I would approach this by first writing tests with simple rules to cover all the cases. In general, I'm very okay with constraining the rules or expressing them differently if we can run it faster.

mesozoic-egg:
Added some test and helpers to compare bottom-up-without-backtracking and topdown rewrite for correctness. The existing patterns in `test_uop_graph.py` pass. 

The arange pattern fails because the pattern returns a node that can further be simplified:
image

Will add more tests coverage before I make changes to the pattern and algorithms

mesozoic-egg:
I think I have got it to be correct now, although not fast yet. 

This is the arange uop rewrite comparison, bottom up vs top down, with inserted node traversal:
Screenshot 2024-06-23 at 15 53 14

Note that it actually has more visit count to each node, it is due to the inserted node traversal to work with existing patterns:
```
      if rewritten:
        # Non-top level nodes may satisfy further patterns
        return self.graph_rewrite_bottomup_no_backtrack(rewritten, pm, counter)
```

If I can constrain the pattern to output clean nodes, then this if block can be removed and result in much fewer visit count:
Screenshot 2024-06-23 at 15 51 51

mesozoic-egg:
I think this is ready for review before I make more changes 

1. Node rewrite now occurs bottom up, with tests
2. Complex pattern (sum_collapse, arange) returns a flag asking for another round of bottom up processing. This essentialy works like the unbounded while loop, but only applicable to specified pattern; I have not come up with a way to make the two output node that cannot be further simplified yet


Speed comparison for the rewrite part:
Screenshot 2024-06-24 at 20 09 59
Screenshot 2024-06-24 at 20 09 38

Overall speed comparison:
Screenshot 2024-06-24 at 20 14 49
Screenshot 2024-06-24 at 20 14 10

Node visit count comparison:

Screenshot 2024-06-24 at 19 57 21
Screenshot 2024-06-24 at 19 57 15

geohot:
Why is CI failing? In general, we don't review until CI is passing and you think the code is final.

DKormann:
if i understand correctly you added a new rewrite rule for loop collapse because the problem was that the bottom up rewrite rewrites parents so that the original loop collapse pattern doesnt apply any more to the WHERE UOp since the rule was designed to be used on the root first and than on the sources is that right?

mesozoic-egg:
@DKormann, the original loop collapse is looking for this pattern:
```
┗UOps.ALU dtypes.float BinaryOps.MUL
  ┣UOps.RANGE dtypes.float (2, 0)
  ┃ ┣UOps.CONST dtypes.float 0.0
  ┃ ┗UOps.CONST dtypes.float 2.0
  ┗UOps.CONST dtypes.float -1.0
```

However with the bottom up rewrite, the ALU would be rewritten to
```
┗UOps.ALU dtypes.float UnaryOps.NEG
  ┗UOps.RANGE dtypes.float (2, 0)
    ┣UOps.CONST dtypes.float 0.0
    ┗UOps.CONST dtypes.float 2.0
```

and as a result not match, I added the pattern to look for this NEGate node and replace the -1 with a hard coded value when the `negate` flag arg is set to true, check out the `test_loop_collapse_match` test case in test_uop_graph.py

geohot:
What do you think about #5159?

mesozoic-egg:
@geohot , IIUC #5159 seems to accomplish more or less the same effect (dfs post order queue but with inserted elements midway, while mine does a recursion when handling this insertion), I'm fine with either approach though.

I actually spent more efforts on identifying & fixing the failing pattern match. Patterns may no longer match when child node are rewritten to something else first. This seems to be the major cause of failing tests.

I will have to get the tests all fixed before knowing if this will work out or not (I think I'm getting close....) Meanwhile if there are other better directions I'm okay with dropping this lol

geohot:
Oh I just wanted to try and write it, idk if mine is better. I think @DKormann might know which rules we have to change.

DKormann:
I think I got them might be just two

mesozoic-egg:
Only PTX tests failing now, tackling that next

PR #5081, opened 2024-06-21
hcq spec

nimlgen:
Thinking of making something like HCQ Spec (set of commands in queues and features to be implemented to make it compatible with HCQGraph, copy_from_disk). Qcom should support this too, but some little tweaks for UnifiedMemory (like as_buffer implementation).

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -196,16 +196,16 @@ class HCQCompatCompiled(Compiled):
     super().__init__(device, allocator, renderer, compiler, runtime, HCQGraph)
 
   @classmethod
-  def _read_signal(self, sig): pass # reads a value for a signal
+  def _read_signal(self, sig): raise NotImplementedError("need _read_signal") # reads a value for a signal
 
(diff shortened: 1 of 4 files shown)


tinygrad/device.py
```diff
@@ -183,3 +183,70 @@ def __init__(self, device:str, allocator:Allocator, renderer:Optional[Renderer],
     self.dname, self.allocator, self.compiler, self.runtime, self.graph = device, allocator, compiler if compiler else Compiler(), runtime, graph
     self.renderer = renderer if renderer else Renderer()
   def synchronize(self): pass  # override this in your device
+
+# **************** for HCQ Compatible Devices ****************
+
+class HCQCompatComiled(Compiled):
```
geohot, line 189: typo

geohot:
Looking good. Hopefully this refactor is pretty easy to confirm it's correct with the tests. If there's anywhere the tests pass that doesn't work on the real device, we should figure out how to improve the tests.

PR #5085, opened 2024-06-21
[NOT FOR MERGE] Approximations for SIN/EXP2/LOG2 in function.py (almost) passing all tests.

hikettei:
# Description

(Moved to https://github.com/tinygrad/tinygrad/pull/5187)
An approximation of mathematical functions(sin, exp2, and log2) based on [SLEEF](https://arxiv.org/pdf/2001.09258).

## Changes

- Added two bitwise operations, `BinaryOps.AND` and `BinaryOps.OR` to quickly compute numbers that cannot be fully covered by FP64 when reducing the argument.
    - [x] [TO FIX] As of now, some tests of PTX fail using `assembly.py` due to this IR.
 
- Added an approximation implementation of `Sin`, `Log2`, `Exp2`. (`_xsin(x)`, `_xlog2(x)`, `_xexp2(x)`)

## Implementation

- Sin
    - Depending on the scale of the input value, different approaches are used for approximation.
    - For example, when executing in FP32, Cody-Waite Reduction is used for `x < 39000.0`, and Payne-Hanek Reduction (this is why I had to implement new uops) is used for larger values to reduce the input value to the range of `[-π/4, +π/4]`. After this reduction, the sine function is approximated using a polynomial.
    - This ensures that even very large inputs, such as those greater than 1e22, can be correctly approximated.
    - In deep learning, it is rare to deal with very large numbers such as `x >= 39000.0`, so a `fast (default: False)` option has been implemented. When this option is set to True, the bottleneck Payne-Hanek Reduction is skipped. (See the benchmark below)
    - In terms of speed, there is room for optimization. Specifically, looking at the generated code, it might not be necessary to separate the three reduction stages.

- Log2
    - The implementation is completed based on [SLEEF](https://arxiv.org/pdf/2001.09258) (Section 5.5, Implementation of log).
    - 1.0 ULP for FP32, 3.5 ULP for FP64.

- Exp2
    - First, it reduces the input to `[−(log 2)/2,(log 2)/2]` by finding `r` and integer `k` such that `x = k log 2 + r, |r| ≤ (log 2)/2`.
    - then uses a polynomial approximation to directly approximate the exponential function of this domain.
    - 3.5 ULP for FP64, 1.0 ULP for FP32

## Workload

- Resolve the out-of-bounty issues.
    - [x] Fix the scheduler tests by doing either of the following:
        - [x] Investigate why `.cast(dtype, True)` does not produce a single kernel (in another issue).
        - [x] Replace EXP2 with SQRT in `test_scheduler.py` (in another PR).
    - [x] Address the `cycle detected in graph` issue with EXP2.
    - [x] Investigate why AMD may fail due to REMU (this is out of my knowledge).

- In this PR, I have to complete the following tasks:
    - [x] Implement FP32/64 for SIN/EXP2/LOG2, which passes all tests in `uops test`.
        - `SIN` passes all tests regardless of backends except for AMD.
        - The accuracy of `EXP2` and `LOG2` is currently sufficient, but they fail the scheduler-related tests.
    - [x] Implement FP16 for SIN/EXP2/LOG2.
    - [x] Implement BFloat16 for SIN/EXP2/LOG2
    - [ ] (Optional) remove `UnaryOps.SIN, UnaryOps.EXP2, UnaryOps.LOG2`.
    - [x] Fix: Nan location mismatch at fp16/fp32, log2/exp2 in AMD, LLVM.
    - [x] ~~Try to replace bitwise ops `(SHR, SHL, AND, OR)` with `bitcast` and `ADD/MUL/DIV`.~~
        - [x] If impossible, create another PR to implement `BinaryOps.AND`, `BinaryOps.OR`.
    - [x] (TO FIX) the current implementation of `Log2` fails due to insufficient precision in Pow's backward and Log(Exp(x)) (in the `ONNX+Optimization Test`).
        - [x] There is still room to increase the precision of Log2 from 3.5 ULP to 1.0 ULP. It needs to be stable around 1e-7 (eps) as well.
        - [x] Pow is unstable in certain backends? (fails on only CLANG) -> Fixed by using partially fp64.
    - [ ] (Optional) I think the current implementation of `_xsin` still has room for optimization.
        - Currently, all levels `lv1, lv2, and lv3` are being calculated regardless of the range of `x`, which makes it slow.
    - [x] Refactor the code to make it more readable, rewrite verbose lines, and move this out of `function.py`.
    - [ ] Make it `< 8000 lines`
    - [ ] Passing all tests

## Current status

~~All the tests related to the accuracy of the approximation calculations for `sin/log2/exp2` are passing, so I believe the implementation approach is correct. However, when the approximation is enabled, the number of kernels increases to two (i.e.: cannot be fused), causing the Scheduler-related tests to fail.~~

EDIT: All tests have passed except for AMD. I am now running the integrated test at https://github.com/hikettei/tinygrad/pull/2.


geohot:
Nice, I think this is the first one to make tests pass!

For BinaryOps.AND and BinaryOps.OR, you can bitcast to bool and use MUL/ADD. If this isn't fast, we should make it fast. I consider this outside the scope of the bounty though, so if you get the tests passing with AND/OR the bounty is yours.

hikettei:
Thank you for the advice! Exp and Log is not as difficult as sin, so I guess it can be done quickly. Nevertheless, the AMD tests are not passing though :(.

hikettei:
I was able to implement Exp2 which passes the `uops test` locally.

I think one of the challenging parts of this bounty is that the tests (e.g.: `test_schedule.py`) assume that Exp2 is fused into a single kernel. However, with my understanding and approximation implementation, using `bitcast=True` causes Exp2 to be split into two kernels, resulting in test failures.

What would be the ideal solution to pass the tests?
- Find an approximation implementation that is fused into a single kernel. (UnaryOps.EXP2/UnaryOps.SIN/UnaryOps.LOG2 can be removed.)
- Modify `test_schedule.py` to adapt to the new results
- Keep both the approximation implementation and `UnaryOps.EXP2`, and disable the approximation path with `Tensor.exp2(fast=False)`.

geohot:
Hmm, bitcast shouldn't be splitting the kernel, but also out of scope of the bounty. Want to post an issue with a minimal repro that causes a kernel to be split? It's probably a bug, I would have expected it to be in a single kernel.

You can ignore the scheduler tests if they are the only failures.

geohot:
The AMD may be failing because REMU, our AMD emulator, doesn't support all the ops. @Qazalin, can you look into this and add an error message instead of a crash?

geohot:
This looks pretty close actually. Bounty locked. What's the issue with FP16 support?

For BinaryOps.AND and OR, have you tried doing it with bitcast to bool and MUL/ADD? If there's a reason that doesn't work, we'll have to get those merged in a separate PR with tests.

tinygrad/function.py
```diff
@@ -36,13 +36,371 @@ def forward(self, x:LazyBuffer) -> LazyBuffer:
...
+      -2.50521083763502045810755e-08,
+      2.75573192239198747630416e-06,
+      -0.000198412698412696162806809,
+      0.00833333333333332974823815,
+    )
+    u = _mla(u, s, d.const(-0.166666666666666657414808))
+    u = _mla(s, u.e(BinaryOps.MUL, d), d)
+  return u
+
+def _xsin(x: LazyBuffer, is_metal: bool=False, fast: bool=False) -> LazyBuffer:
```
geohot, line 303: Hmm, why do you need an is_metal here? If metal is behaving differently we should probably fix it in the METAL backend if possible.

geohot:
Also, want to move this out of function.py and into a new file? Just import the three functions in function.py

hikettei:
Thank you very much for your detailed response!!

> Want to post an issue with a minimal repro that causes a kernel to be split? It's probably a bug

Yes, I will do that. (It's 7:00 in Japan now and I need to go to school so it might take a bit longer)

`> What's the issue with FP16 support?`

There is no issue related to tinygrad. Actually, I haven't started working on it because I haven't found any documentation on implementing FP16 support in SLEEF or PyTorch. (I plan to implement it tonight along with a simulation using Julia)

`> For BinaryOps.AND and OR, have you tried doing it with bitcast to bool and MUL/ADD?`

Actually, I don't quite understand how to use the method with bitcast and MUL/ADD. When I do `.cast(dtypes.bool, True)`, I encounter the following error:

```python
def cast(self, dtype:DType, bitcast:bool=False):
      if self.dtype == dtype: return self
      if self.device.startswith("DISK") and not bitcast: raise RuntimeError("attempted to cast disk buffer (bitcast only)")
      if self.is_unrealized_unmasked_const() and not bitcast:
        return create_lazybuffer(self.device, self.st, dtype, LoadOps.CONST, dtypes.as_const(self.base.arg, dtype))
      new_shape = self.shape
      if bitcast and self.dtype.itemsize != dtype.itemsize:
>       if not self.device.startswith("DISK"): raise RuntimeError("shape changing bitcast only supported on DISK right now")
E       RuntimeError: shape changing bitcast only supported on DISK right now

tinygrad/lazy.py:102: RuntimeError
```

tinygrad/function.py
hikettei: When running in FP32, it was necessary to calculate the value of 2/pi in FP64 for the range where `x >= 39000.0`. However, since FP64 is not supported on METAL, I set is_metal=True to ignore fp64 computation.

Upon further investigation, I found that FP32 was sufficient, so I will remove `is_metal` later.

hikettei:
> Also, want to move this out of function.py and into a new file?

Yes, I totally agree. What name should it have? (`tinygrad/fastmath.py`?)

hikettei:
I think the current implementation in `function.py` is too complicated to read, making the review and discussion harder. So, I will share a code based on the experimental code for reference. (Apologies if there are any inconsistencies). 

<details>
  <summary>_xsin(x)</summary>

Corresponding code in SLEEF:
- https://github.com/shibatch/sleef/blob/4e08851f59fc2b545f9c393c6a23dfd311a26308/src/libm/sleefsp.c#L464
- https://github.com/shibatch/sleef/blob/4e08851f59fc2b545f9c393c6a23dfd311a26308/src/libm/sleefdp.c#L789

FP16 is not working.

```julia
import Base: TwicePrecision, significand_bits, significand_mask, exponent_mask, exponent_bias
const two_over_pi_f = [
    0x00000000,
    0x28be60db,
    0x9391054a,
    0x7f09d5f4,
    0x7d4d3770,
    0x36d8a566,
    0x4f10e410
]
function float_to_bits(x::Float32)
    return reinterpret(UInt32, x)
end

function bits_to_float(x::UInt32)
    return reinterpret(Float32, x)
end

function float_to_bits(x::Float64)
    return reinterpret(UInt64, x)
end

function bits_to_float(x::UInt64)
    return reinterpret(Float64, x)
end

function float_to_bits(x::Float16)
    return reinterpret(UInt16, x)
end

function bits_to_float(x::UInt16)
    return reinterpret(Float16, x)
end

function my_frexp(value::Float16)
    if value == Float16(0.0)
        return Float16(0.0), 0
    end
    
    bits = float_to_bits(value)
    exponent = (bits >> 10) & 0x1F
    
    if exponent == 0
        return value, 0
    end
    
    exp = exponent - 14
    return bits_to_float((bits & 0x83FF) | 0x3C00), exp
end

function my_frexp(value::Float32)
    if value == 0.0f0
        return 0.0f0, 0
    end
    
    bits = float_to_bits(value)
    exponent = (bits >> 23) & 0xFF
    
    if exponent == 0
        return value, 0
    end
    
    exp = exponent - 126
    return bits_to_float((bits & 0x807FFFFF) | 0x3F000000), exp
end

function my_frexp(value::Float64)
    if value == 0.0
        return 0.0, 0
    end
    
    bits = float_to_bits(value)
    exponent = (bits >> 52) & 0x7FF
    
    if exponent == 0
        return value, 0
    end
    
    exp = exponent - 1022
    return bits_to_float((bits & 0x800FFFFFFFFFFFFF) | 0x3FE0000000000000), exp
end

function reduce_large(a::Float16)
    f, e = my_frexp(a)
    ia = UInt64(abs(Float16(f)) * 0x1.0p32)
    i = UInt64(e) >> 5
    e = UInt64(e) & 31

    if e != 0
        hi = (two_over_pi_f[Int(i) + 1] << e) | (two_over_pi_f[Int(i) + 2] >> (32 - e))
        mid = (two_over_pi_f[Int(i) + 2] << e) | (two_over_pi_f[Int(i) + 3] >> (32 - e))
        lo = (two_over_pi_f[Int(i) + 3] << e) | (two_over_pi_f[Int(i) + 4] >> (32 - e))
    else
        hi = two_over_pi_f[Int(i) + 1]
        mid = two_over_pi_f[Int(i) + 2]
        lo = two_over_pi_f[Int(i) + 3]
    end
    
    p = UInt64(ia) * UInt64(lo)
    p = UInt64(ia) * UInt64(mid) + (p >> 32)
    p = (UInt64(ia) * UInt64(hi) << 32) + p

    q = Int32(p >> 62)
    p = p & 0x3fffffffffffffff
    if p & 0x2000000000000000 != 0
        p -= 0x4000000000000000
        q += 1
    end
    
    d = Float64(p)
    d *= 0x1.921fb54442d18p-62
    r = Float16(d)
    if a < Float16(0.0)
        r = -r
        q = -q
    end
    
    return r, q
end


function reduce_large(a::Float64)
    f, e = my_frexp(a)
    ia = UInt64(abs(f) * 0x1.0p32)
    i = UInt64(e) >> 5
    e = UInt64(e) & 31
    if e != 0
        hi = (two_over_pi_f[Int(i) + 1] << e) | (two_over_pi_f[Int(i) + 2] >> (32 - e))
        mid = (two_over_pi_f[Int(i) + 2] << e) | (two_over_pi_f[Int(i) + 3] >> (32 - e))
        lo = (two_over_pi_f[Int(i) + 3] << e) | (two_over_pi_f[Int(i) + 4] >> (32 - e))
    else
        hi = two_over_pi_f[Int(i) + 1]
        mid = two_over_pi_f[Int(i) + 2]
        lo = two_over_pi_f[Int(i) + 3]
    end
    
    p = UInt64(ia) * UInt64(lo)
    p = UInt64(ia) * UInt64(mid) + (p >> 32)
    p = (UInt64(ia) * UInt64(hi) << 32) + p

    q = Int32(p >> 62)
    p = p & 0x3fffffffffffffff
    if p & 0x2000000000000000 != 0
        p -= 0x4000000000000000
        q += 1
    end
    
    d = Float64(p)
    d *= 0x1.921fb54442d18p-62
    r = Float64(d)
    if a < 0.0
        r = -r
        q = -q
    end
    
    return r, q
end

function reduce_large(a::Float32)
    f, e = my_frexp(a)
    # f, e = OK
    ia = UInt64(abs(float(f)) * 0x1.0p32)
    i = UInt64(e) >> 5
    e = UInt64(e) & 31
    if e != 0
        hi = (two_over_pi_f[Int(i) + 1] << e) | (two_over_pi_f[Int(i) + 2] >> (32 - e))
        mid = (two_over_pi_f[Int(i) + 2] << e) | (two_over_pi_f[Int(i) + 3] >> (32 - e))
        lo = (two_over_pi_f[Int(i) + 3] << e) | (two_over_pi_f[Int(i) + 4] >> (32 - e))
    else
        hi = two_over_pi_f[Int(i)+ 1]
        mid = two_over_pi_f[Int(i) + 2]
        lo = two_over_pi_f[Int(i) + 3]
    end

    # hi, ok -> ok
    # mid, ok -> ok
    # lo, ok -> ok

    
    p = UInt64(ia) * UInt64(lo)
    p = UInt64(ia) * UInt64(mid) + (p >> 32)
    p = ((UInt64(ia) * UInt64(hi)) << 32) + p
    
    q = Int32(p >> 62)
    p = p & 0x3fffffffffffffff
    if p & 0x2000000000000000 != 0
        p -= 0x4000000000000000
        q += 1
    end
    
    d = Float64(p) # maybe no need to use Float64, Float32 is enough?
    d *= 0x1.921fb54442d18p-62
    r = Float32(d)
    if a < 0.0
        r = -r
        q = -q
    end
    
    return r, q
end

function xsin(x)
    r, q = reduce_large(x)
    q = q % 4
    if q == 0
        return sin(r)
    elseif q == 1
        return sin(r + π/2)
    elseif q == 2
        return sin(-r)
    elseif q == 3
        return -sin(r + π/2)
    end
end
```
</details>

<details>
  <summary>_xlog2(x)</summary>

- https://github.com/shibatch/sleef/blob/4e08851f59fc2b545f9c393c6a23dfd311a26308/src/libm/sleefsp.c#L1671
- https://github.com/shibatch/sleef/blob/4e08851f59fc2b545f9c393c6a23dfd311a26308/src/libm/sleefdp.c#L2072

```julia
import Base: TwicePrecision, significand_bits, significand_mask, exponent_mask, exponent_bias

function float_to_bits(x::Float32)
    return reinterpret(Int32, x)
end

function bits_to_float(x::Int32)
    return reinterpret(Float32, x)
end

function float_to_bits(x::Float64)
    return reinterpret(Int64, x)
end

function bits_to_float(x::UInt64)
    return reinterpret(Float64, x)
end

function bits_to_float(x::Int64)
    return reinterpret(Float64, x)
end

const FLT_MIN = 1.0#1.4012984643248170709e-4

function ilogb2kf(d::Float32)
    return ((float_to_bits(d) >> 23) & 0xff) - 0x7f
end

function ldexp3kf(d::Float32, e::Int32)
    return bits_to_float(float_to_bits(d) + (e << 23))
end

function ilogb2kf(d::Float64)
    return ((float_to_bits(d) >> 52) & 0x7ff) - 0x3ff
end

function ldexp3kf(d::Float64, e::Int64)
    return bits_to_float(float_to_bits(d) + (e << 52))
end

function mlaf(x, y, z)
    return x * y + z
end

function dfmul_f2_f_f(x, y)
    xh = upper(x)
    yh = upper(y)
    xl = x - xh
    yl = y - yh
    
    rx = x * y;
    ry = (xh * yh) - rx + (xl * yh) + (xh * yl) + (xl * yl);
    return rx, ry
end

function dfadd_f2_f_f2(x, yx, yy)
    rx = x + yx
    ry = x - rx + yx + yy
    return rx, ry
end

function xlog2(d::Float32)
    o = d < FLT_MIN
    if o
        d *= 2^32
        d *= 2^32
    end
    
    e = ilogb2kf(Float32(d * (1.0/0.75)))
    m = ldexp3kf(d, -e)

    if o
        e -= 64
    end

    x = (m - 1) / (m + 1)
    x2 = x * x

    t = +0.4374088347e+0
    t = mlaf(t, x2, +0.5764843822e+0)
    
    t = mlaf(t, x2, +0.9618024230e+0)
    r = mlaf(x2 * x, t, mlaf(x, +0.2885390043e+1, e))

    if d == Inf
        r = Inf
    end

    if d < 0 || d == NaN
        r = NaN
    end

    if d == 0
        r = -Inf
    end

    return Float32(r)
end

const DFLT_MIN = 1.0#1.4012984643248170709e-4

function xlog2(d::Float64)
    o = d < DFLT_MIN
    if o
        d *= 2^32
        d *= 2^32
    end
    
    e = ilogb2kf(Float64(d * (1.0/0.75)))
    m = ldexp3kf(d, -e)

    if o
        e -= 64
    end

    x = (m - 1) / (m + 1)
    x2 = x * x

    t = +0.2211941750456081490e+0
    t = mlaf(t, x2, +0.2200768693152277689e+0)
    t = mlaf(t, x2, +0.2623708057488514656e+0)
    t = mlaf(t, x2, +0.3205977477944495502e+0)
    t = mlaf(t, x2, +0.4121985945485324709e+0)
    t = mlaf(t, x2, +0.5770780162997058982e+0)
    t = mlaf(t, x2, +0.96179669392608091449)

    a, b = dfmul_f2_f_f(2.885390081777926774, x)
    sx, sy = dfadd_f2_f_f2(e, a, b)
 
    r = mlaf(t, x * x2, sx + sy)
    if d == Inf
        r = Inf
    end

    if d < 0 || d == NaN
        r = NaN
    end

    if d == 0
        r = -Inf
    end

    return Float64(r)
end
```
</details>

<details>
  <summary>_xexp2(x)</summary>

- https://github.com/shibatch/sleef/blob/4e08851f59fc2b545f9c393c6a23dfd311a26308/src/libm/sleefsp.c#L1538

- https://github.com/shibatch/sleef/blob/4e08851f59fc2b545f9c393c6a23dfd311a26308/src/libm/sleefdp.c#L1910C21-L1910C26
</details>

tinygrad/function.py
hikettei: I have deleted `is_metal` option.

hikettei:
It gets really, really close!

Passed all tests with Metal/OpenCL. LLVM/CLANG should pass all if NaN handling is tweaked. NV will pass after refactoring the last two tests, but AMD and PTX need a few more commits.

https://github.com/hikettei/tinygrad/pull/1

I've split the PR, so I'm currently running integration tests on my fork. I plan to create a PR once the foundation is a bit more solid.

hikettei:
The remaining task is to implement (u)int64 AND/OR on AMD, which will allow all tests to pass. After that, we can move on to the review stage.

hikettei:
I have confirmed that all kernels can be fused into one. Here's the latest version of the simple benchmark.

## Benchmark

### Conditions

- Device: Apple M3 Pro Max (11 Core CPU, 14 Core GPU)
- N: 10000x10000

| Runtime | Op | Dtype | Base(m) | Approx(ms) | Approx_fast(ms) |
|---|---|---|---|---|---|
|CLANG| sin | fp16 | 363.41 | 1041.16 | 348.48 |
|CLANG| sin | fp32 | 438.88 | 934.55 | 251.88 |
|CLANG| sin | fp64 | 783.28 | 1724.33 | 872.14 |
|CLANG| exp2 | fp16 | 419.21 | 148.58 | x |
|CLANG| exp2 | fp32 | 251.58 | 187.16 | x |
|CLANG| exp2 | fp64 | 526.56 | 360.89 | x |
|CLANG| log2 | fp16 | 691.15 | 280.76 | x |
|CLANG| log2 | fp32 | 318.36 | 263.66 | x |
|CLANG| log2 | fp64 | 629.59 | 547.40 | x |
|METAL| sin | fp16 | 84.03 | 76.18  | 82.46 |
|METAL| sin | fp32 | 163.88 | 184.78 | 167.57 |
|METAL| sin | fp64 | x | x | x |
|METAL| exp2 | fp16 | 80.90 | 81.96  | x |
|METAL| exp2 | fp32 | 163.35 | 160.55 | x |
|METAL| exp2 | fp64 | x | x | x |
|METAL| log2 | fp16 | 78.44 | 80.41 | x |
|METAL| log2 | fp32 | 167.19 | 166.89 | x |
|METAL| log2 | fp64 | x | x | x |

(x = not implemented, approx_fast = an option to ignore payne_hanek reduction for sin)

FYI: FP16 implementation can be further optimized by using [table-lookup-based algorithm](https://github.com/shibatch/sleef/issues/320)

hikettei:
Moved to https://github.com/tinygrad/tinygrad/pull/5187

chenyuxyz:
moved bounty lock and updated spreadsheet to point to #5187

hikettei:
thanks, so I will close the old one.

PR #5087, opened 2024-06-21
explicitly check getitem indices can have at most one ellipsis

chenyuxyz:
previous error message with multiple `...`:
```
if index_type not in [None, int, slice, Tensor]: raise IndexError(f"{index_type=} not supported")
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: index_type=<class 'ellipsis'> not supported
```

this pr:
```
if len(ellipsis_idx) > 1: raise IndexError("an index can only have a single ellipsis ('...')")
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: an index can only have a single ellipsis ('...')
```

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -1039,9 +1039,9 @@ class TestOps(unittest.TestCase):
   def test_slice_errors(self):
     a = Tensor.ones(4, 3)
     b = Tensor(2)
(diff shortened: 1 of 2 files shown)


chenyuxyz:
oh we have that already it's just never triggered due to the order

PR #5093, opened 2024-06-21
recursive uopgraph dedup [RUN_PROCESS_REPLAY]

DKormann:
is this a bad idea?


geohot:
We need a framework for measuring speed across a diverse set of kernels in order to answer that question. Want to write one?

DKormann:
can it be included in process replay testing?

PR #5095, opened 2024-06-22
update llama model RMSNorm casting

chenyuxyz:
following the original implementation, cast back to input dtype before multiplying weight. slightly faster https://github.com/meta-llama/llama/blob/main/llama/model.py


chenyuxyz:
about 1% faster

PR #5096, opened 2024-06-22
update llama attention casting

chenyuxyz:
updated scaled_dot_product_attention middle cast and removed hard-coded half in llama attention.


chenyuxyz:
changed mid cast behavior in #5100 with its own tests

examples/llama3.py
```diff
@@ -93,7 +93,7 @@ def NF4Linear(block_size):
     -1.0, -0.6961928009986877, -0.5250730514526367, -0.39491748809814453, -0.28444138169288635, -0.18477343022823334, -0.09105003625154495, 0.0,
     0.07958029955625534, 0.16093020141124725, 0.24611230194568634, 0.33791524171829224, 0.44070982933044434, 0.5626170039176941, 0.7229568362236023, 1.0,
   ]
-  CODE = Tensor.stack(*[Tensor(c) for c in _CODE])
+  CODE = Tensor.stack(*[Tensor(c, dtype=dtypes.float16) for c in _CODE])
```
chenyuxyz, line 96: @wozeparrot fyi, I don't think it was intended to use float32 for these? nf4 30% faster but still very slow

PR #5100, opened 2024-06-22
fix attention to always return in the same dtype as input

chenyuxyz:
mid cast to default_float does not work as intended when default is float32 and qkv is in half


chenyuxyz:
@Eliulm fyi, this should still work with default half training bert

PR #5101, opened 2024-06-22
tinytqdm.set_description and tinytrange

chenyuxyz:


geohot:
Should we drop the "tiny" from the function name. Just from helpers import `tqdm, trange`

chenyuxyz:
sure, done

PR #5112, opened 2024-06-23
Proof or disproof of the mergeability of two arbitrary ShapeTrackers in Lean (generic version of https://github.com/tinygrad/tinygrad/pull/2218/files)

Humbulani1234:
Please find attached my proposal for the proof. Should the proposal be viable I intend to complete it, Lean it, and write some few tests.
[tinyproof_final.pdf](https://github.com/user-attachments/files/15944195/tinyproof_final.pdf)

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -2,7 +2,7 @@
 <head>
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <style>
-#result {  font-size: 48px; }
+# result {  font-size: 48px; }
 #time {  font-size: 16px;  color: grey; }
 #mybox {  padding: 20px; }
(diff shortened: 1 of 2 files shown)


chenyuxyz:
saying you intend to prove by induction and contradiction is very far from a proof proposal

Humbulani1234:
Good day,The response doesn’t necessarily mean I can drop this line of proof? My understanding is that you just want more - at least one proposition demonstrated.RegardsHumbulani Ndou081 385 5918On 23 Jun 2024, at 20:27, chenyu ***@***.***> wrote:﻿
saying you intend to prove by induction and contradiction is very far from a proof proposal

Issue #5097, opened 2024-06-22
t.shard(axis=0) does not free unused parts of buffer in origin device


chaosagent:
```
>>> devices = [f"AMD:{i}" for i in range(6)]
>>> t = Tensor.empty(4096, 4096)
>>> t.shard_(devices, axis=0).realize()
>>> print(t.lazydata.lbs[0].buffer.base.size)
16777216
>>> print(t.lazydata.lbs[1].buffer.base.size)
2797568
```

Proposed solution: add a flag `allow_buffer_view=True` to Tensor.contiguous, plumb it through function.py, and make LoadOps.VIEW optional in lazy.py somehow. We want to enable subbuffer creation for ephemeral tensors but not long-living ones or input data; set this flag to False in MultiLazyBuffer.from_sharded.

Alternate solution (hack): In LazyBuffer.copy_to_device, do a copy (somehow) even if src.device = dst.device as long as the src op is LoadOps.VIEW

Other idea: Don't allow user to have Tensors with viewed Buffers at all (ie view the buffers internally for copies and memory scheduling)

Workaround: As the user, don't shard tensor that is already on one of the devices.

Don't know what is good.

chenyuxyz:
fixed by #5122

PR #5120, opened 2024-06-24
make buffer view optional with a flag

chaosagent:
toward https://github.com/tinygrad/tinygrad/issues/5097


geohot:
Can process replay run on this?

chaosagent:
yeah, but process replay just takes asts and puts them into Linearizer. i tried it but ran into #5127

Qazalin:
Yeah process replay is for codegen/*, though the infra can support lazy/schedule with some work to get master's ASTs for the comparison.

I ran process replay in CI right now and #5127 didn't come up - seems to be a non-deterministic Variable deduping failure.

PR #5124, opened 2024-06-24
verify_lazyop

Qazalin:
- [x] Write spec and pass all of CI, red for wrong ASTs
- [x] Fix wrong multi reduce ASTs - Skipped for now, multi reduce ASTs are all wrong and need to be rewritten 
- [x] Benchmark speed.


tinygrad/ops.py
```diff
@@ -144,3 +145,31 @@ def truncate_fp16(x):
...
+    assert out.arg.idx == i, f"unexpected output buffer idx {out.arg.idx} != {i}"
+    assert out.op is BufferOps.STORE, f"kernels must have stores as the output, got {out.op}"
+    assert out.arg.st.size == ast[-1].arg.st.size, f"outputs must have the same size, got {out.arg.st.size}"
+    for op in out.lazyops:
+      for s in op.src:
+        children[s][op] = None
+        in_degree[op] += 1
+  q = deque((op, op.arg.st) for out in ast for op in out.lazyops if in_degree[op] == 0)
+  sts: Dict[LazyOp, ShapeTracker] = {}
+  while q:
```
geohot, line 163: It's probably simpler to write this as a recursive function with functools.lru_cache, maybe simpler to reason about?

geohot:
Merging to make progress on lowerer, but see above comment on recursive. BFS might be better, but we really need to have this in a library.

Qazalin:
I'll iterate on it with DFS, so far the DFS versions I tried were harder to reason about or wrong.

geohot:
Maybe we stick with BFS and just make this a library.

PR #5125, opened 2024-06-24
Add MLLogger

Eliulm:
Changes: 

1. Adds MLLogger
2. No eval at the start. According to the [rules](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#94-quality-measure), we are allowed to skip this one.
3. Removes eval log file
4. Step numbering starts with 1
5. Checkpointing is on by default


chenyuxyz:
can you have a run with this and test the logging is correct with compliance checker? https://github.com/mlcommons/logging/tree/master/mlperf_logging/compliance_checker

Eliulm:
Should be compliant with 4.0.0 now:

`
INFO - Running compliance on file: bert.log
INFO -  Compliance checks: training_4.0.0/common.yaml
INFO -  Compliance checks: training_4.0.0/closed_common.yaml
INFO -  Compliance checks: training_4.0.0/closed_bert.yaml
INFO - SUCCESS
`

examples/mlperf/model_train.py
```diff
@@ -546,7 +610,11 @@ def train_bert():
         et = time.time()
         eval_times.append(et - st)
 
-        if BENCHMARK and j == BENCHMARK: break
+        if BENCHMARK and j == BENCHMARK:
+          # assume INITMLPERF has BENCHMARK set
```
chenyuxyz, line 614: this is fine. can you assert if it's not set?

examples/mlperf/model_train.py
```diff
@@ -577,10 +648,15 @@ def train_bert():
         seconds = total_seconds % 60
         print(f"Reference Convergence point reached after {i * BS} datasamples and {hours}h{minutes}m{seconds:.2f}s.")
         achieved = True
-        # stop once hitting the target
```
chenyuxyz, line 580: keep the comment?

examples/mlperf/helpers.py
```diff
@@ -225,3 +225,14 @@ def get_data_bert(GPUS:list[str], it):
   data: dict[str, Tensor] = next(it)
   for key in data.keys(): data[key].shard_(GPUS, axis=0)
   return data
+
+def get_fake_data_bert(GPUS:list[str], BS:int):
+  return {
+    "input_ids": Tensor.zeros((BS, 512), dtype=dtypes.float32).shard_(GPUS, axis=0),
```
chenyuxyz, line 231: add `contiguous` to these `zeros` to make these real buffers. for now we call `contiguous` when sharding but this might change in the future

chenyuxyz (commented):
nice! some minor comments

also you can copy the log 5 times with different file names and test as if it's the whole submission

Eliulm:
I get
 
```
(env) eliulm@tiny10:~/tinygrad$ python3 -m mlperf_logging.package_checker . training 4.0.0
INFO -  Checking System Description Files
INFO - ------------------------------
INFO - PACKAGE CHECKER FOUND NO ERRORS, SUCCESS !
INFO - ------------------------------

** Detailed log output is also at package_checker.log
```

from the package checker. Logs have to be moved to the results folder though.

PR #5129, opened 2024-06-24
Mergeability proof of two ShapeTrackers

Humbulani1234:
Please find attached my revised proof proposal, here are the changes 

- Implemented an Induction for a particular proposition - The last section of the document. There other propositions to implement, however I have done only one so I can get an assessment if my approach is viable. The others will follow the exact same logic

- The proof only uses Induction 

- There are still some nuances to be ironed, like variables ranges, indexes, making the proof rigorous and I feel like somethings are better represented as sequences of vectors. However, the logic and structure of the proof is intact

- I will Lean it if the arguments I put prove to be sound and then write tests.

[tiny_final_draft.pdf](https://github.com/user-attachments/files/15960083/tiny_final_draft.pdf)


Humbulani1234:
Will resubmit with a Lean file and tests for one proposition.

PR #5136, opened 2024-06-24
llama concat_weights always send to device[0] first

chenyuxyz:
removing that fixed the llama shard weight double copy. it was from #3966 but i think we fixed replace after that


wozeparrot:
can you fix in llama3 as well?

chenyuxyz:
pending the buffer view contiguous change, otherwise it's copying the full weight on each gpu

chenyuxyz:
i think it's not always guaranteed that the sharding we want among devices are the same as the view on disk, and we cannot directly copy shards from disk to individual gpus

PR #4957, opened 2024-06-14
Linearizer -> Lowerer

geohot:
- [x] Unroll
- [x] Upcast
- [x] Group
- [x] Tensor Cores
  - [x] HIP Tensor Cores
  - [x] CUDA Tensor Cores 
- [x] float4
  - [x] Image

PR #5140, opened 2024-06-25
second iteration on verify_lazyop

Qazalin:
- DFS!
- Fixed bug with it not catching the wrong softmax AST https://github.com/tinygrad/tinygrad/pull/4957#pullrequestreview-2137021025


Qazalin:
This diff finds some ASTs from the ancient times when reduceop arg was new_shape:)
https://github.com/flammit/tinygrad/blob/d216994816986d583fa1ff081080294a00ce638e/tinygrad/lazy.py#L116

open question why despite that the correctness tests continued to pass.

Issue #4927, opened 2024-06-12
Consecutive reduceop fusion


Qazalin:
in mean+stddev, softmax and layernorm, one reduceop builds up on its parent reduceop.
tinygrad is making progress towards fusing these into a single kernel.

### Milestones

1. mean+stddev fusion
- linearizer changes: https://github.com/tinygrad/tinygrad/pull/4259 
- scheduler changes: pending

2. softmax/layernorm fusion

0xtimmy:
Beginning of scheduler tests: #4972

Qazalin:
this is very old pre rangeify stuff, and mostly solved by PCONTIG.

PR #5141, opened 2024-06-25
More multireduce scheduler tests

0xtimmy:
Some more tests that look at specific changes to the scheduler
- std, argmin/max, attention are important cases that are affected by multireduce
- reduce-expand patterns need to be realized properly (reduce->expand->reduce is OK, reduce->expand->reduce->expand needs to be broken up)
- if there are two fuseable reduceops, it only matters that the last one can be cleanly paired
- sequential matmuls cannot be fused


0xtimmy:
re: #4927

test/test_schedule.py
```diff
@@ -684,6 +684,81 @@ def test_reduce_shrink_output(self):
...
+    out = (a+a.sum(-1, keepdim=True)).sum(-1)
+    # run_schedule(check_schedule(out, 1))
+    run_schedule(check_schedule(out, 2))
+    np.testing.assert_allclose(out.numpy(), (a.numpy()+a.numpy().sum(axis=-1,keepdims=True)).sum(axis=-1), atol=1e-4, rtol=1e-4)
+
+  # multireduce spec
+  def test_reduce_expand_reduce_expand_fusion(self):
+    Tensor.manual_seed(0)
+    a = Tensor.randn(4, 32).realize()
+    out = a+(a+a.sum(-1,keepdim=True)).sum(-1, keepdim=True)
```
Qazalin, line 745: Is keepdim=True required to make fusion work? Why can't the scheduler change ShapeTrackers regardless of keepdim user args?
We wanna make fusion be the default path.

0xtimmy:
> Yeah you need to push the expands to before the first reduce in a reduce -> expand -> reduce. These ASTs in multi reduce test_linearizer tests were wrong because they had implicit expands. I'd merge the diff if you can make those ASTs pass verify_lazyop.

what do you mean by push the expand? for std for example the expand is in the middle of the two reduces. I can play games with the shape tracker but in principle there's a reduce -> expand -> reduce chain of dependency
```
x = Tensor.randn( ... )
mean = x.sum() - x.size
diff = x - mean // <- expand
std = ((diff*diff).sum() / x.size).sqrt()
```

test/test_schedule.py
0xtimmy: the `keepdim=True` is to avoid throwing a broadcasting error, `Tensor.var` uses `keepdim=True` as well. unrelated to the fusion (as far as I know)

this code, for example,
```
x = Tensor.randn(4, 32).realize()
y = x+x.sum(-1)
```
will throw:
`ValueError: cannot broadcast from shape=(4,) to shape=(4, 32)`

Qazalin:
Yeah - we wanna do the mean in such a way that after the reduce it's the correct shape and doesn't need the extra expand.

Here's a discord thread related to this:  https://discord.com/channels/1068976834382925865/1068982781490757652/1255910222933917786

The requirements for multi reduce are that the AST passes `verify_lazyop` and generates the correct kernel.

0xtimmy:
I have a four hour flight coming up so I'll see if I can put something together:

one of the issues is these lines in `verify_lazyop`:
```assert st.shape == expected_shape, f"unexpected reduceop shape {st.shape} != {expected_shape}"```
```assert sts[x].shape == st.shape, f"found implicit movement op {x.op} {sts[x].shape} != {op.op} {st.shape}"```
if we have two reduceops the first has to either:
1. reduce the shape to the expected output shape: `(st.shape == expected_shape) = True` but at the expand `(sts[x].shape == st.shape) = False`
2. don't reduce the shape but do something with the shapetracker in which case: `(st.shape == expected_shape) = False` but at the expand `(sts[x].shape == st.shape) = True`

**Comment by Qazalin** [comment]
hmm can you explain further why one would be invalid in all cases, to me, both should be valid:
1. st.shape == expected_shape checks that the axis is correct
2. sts[x].shape == st.shape checks that operands of elementwise ops have exactly the same shape.

If you correctly expand the input buffer for the first reduce and adjust the axis.

**Comment by 0xtimmy** [comment]
True but I think `verify_lazyop` keeps track of it's own `ShapeTracker` and I don't think expands go into the ast.
so if we have an ast like below

-> the deepest `ReduceOps.SUM`, if the axis is correct, will bring the shapetracker to (4,1)
-> the `BinaryOps.ADD` will fail the assert because it's getting `(4,1)` and `(4,32)`

the only shapetrackers the ast knows are those of the `BufferOps`
and `verify_lazyop` changes the st on every `ReduceOp`

maybe I'm missing something? I've played with changing the the shapetrackers on the BufferOps but regardless of what we do with the strides, if they have different axes they'll fail the `assert st.shape == expected_shape` and if they have the same axes, they'll cause a failure with `sts[x].shape == st.shape`. The expand would fix that but I don't think the ast sees expands

for example:
`x = Tensor.randn(4,32)`
`out = (x+x.sum(-1).expand(4,32)).sum(-1)`
```
ast = 
LazyOp(op=BufferOps.STORE, src=(
  LazyOp(op=ReduceOps.SUM, src=(
    LazyOp(op=BinaryOps.ADD, src=(
      LazyOp(op=BufferOps.LOAD, src=(), arg=MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4,32),strides=(32,1),offset=0,mask=None,contiguous=True),)))),
      LazyOp(op=ReduceOps.SUM, src=(
        LazyOp(op=BufferOps.LOAD, src=(), arg=MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4,32),strides=(32,1),offset=0,mask=None,contiguous=True),)))),
      ), arg=(1,)),
    )),
  ), arg=(1)),
), arg=MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker.from_shape((4,1))))
```

**Comment by Qazalin** [comment]
Yeah in your AST
```
  0 ━┳ BufferOps.STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4, 1), strides=(1, 0), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ ReduceOps.SUM 1
  2    ┗━┳ BinaryOps.ADD None
  3      ┣━━ BufferOps.LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4, 32), strides=(32, 1), offset=0, mask=None, contiguous=True),)))
  4      ┗━┳ ReduceOps.SUM (1,)
  5        ┗━━ BufferOps.LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4, 32), strides=(32, 1), offset=0, mask=None, contiguous=True),)))
  ```
becuase line 4 reduces a (4, 32) ShapeTracker to (4, 1), you can't add a LazyOp of shape (4, 1) again to a (4, 32). The trick is to make the reduce output a shape (4, 32) instead.
Have you tried expanding the LOAD on line 5 before reducing? Maybe that will work. The idea is to make your expanded strides 0 so that the Linearizer doesn't wastefully recompute them.

0xtimmy:
Expanding the load on line 5 causes issues with:
```
expected_shape = tuple(1 if i in op.arg else s for i,s in enumerate(sts[op.src[0]].shape))
assert st.shape == expected_shape, f"unexpected reduceop shape {st.shape} != {expected_shape}"
```

a reduceop will end up reducing the shape, because of this line in `verify_lazyop` ; `LazyOp`s don't keep their own `ShapeTracker` like `LazyBuffer`s
```
st = ShapeTracker.from_shape(expected_shape)
```
so if we take any ast, a `ReduceOp` will be expected to produce the same shape as the output shape; but at the same time be expected that shape to later match the full shape if there are any sequential `ReduceOp`s

so, for example, if the LOAD on line 5 is expanded to (4,32,32) and the output is (4,1,1) we'll get
```
line 4:  (4,32,32) => (4,32,1)  // fails expected shape check b/c it expects the reduce to produce the output shape
line 2:  (4,32,1)   => (4,32,1)  // no expand! all good here
line 1:  (4,32,1)   => (4,1,1)    // reduce produces the output shape, all good!
```
to remedy this we could do:
```
line 4:  (4,1,32) => (4,1,1)    // reduce produces the output shape! all good here
line 2:  (4,1,1)   => (4,32,1)  // fails implicit movement op
line 1:  (4,32,1) => (4,1,1)    // reduce again produces the output shape, all good
```
but that causes an implicit movement op

this could be solved by having `ReduceOp`s modify the strides of the `ShapeTracker` instead of the shape, but that's not how `verify_lazyop` works at present. The only way to change the st in `verify_lazyop` is with a `BufferOp` or a `ReduceOp`

Or it could be solved if `verify_lazyop` tracked a pair of sts, corresponding to what will be the `output_shape` and the `full_shape` in `Kernel`

Qazalin:
yea - I moved that assert out https://github.com/tinygrad/tinygrad/pull/5276, we definitely want the first option.

0xtimmy:
update in this pr: #5323

the big remaining issue is reuse of reduceops, if we have something like layernorm:
```
(x - mu) / sigma
```
if x is shape (4,32,32)
mu will have shape (4,32,1) because it is one reduce down the ast
std will have shape (4,32,32) because it is two reduces down the ast
so scheduler will need to generate two mean calculations, 
- one that reduces to (4,32,1) to get reduced into the standard deviation
- one that reduces to (4,1,1) to get used in the final layernorm calculation

have a simple example in the tests that come with the PR

Issue #4878, opened 2024-06-08
Matching engine is slow ($500 bounty)


geohot:
Rewrite the matching engine to get a 2x speedup on my machine (intel core ultra 7 155h) in "model lower"

```
jesse@x1:~/tinygrad$ PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  67.59 ms
***** model schedule in  16.62 ms
***** model lower in 1507.84 ms
```

Target is 754 ms. Must pass process replay, and will likely be achieved through a rewrite of the matcher and graph rewrite engine.

roelofvandijk:
https://github.com/tinygrad/tinygrad/pull/4882

chenyuxyz:
the script has a lot more than rewriting, it's slower for me with `UOPS_REWRITE=0 PROFILE=0 python3 test/external/external_benchmark_schedule.py`

croumegous:
another ~10% easy gain: https://github.com/tinygrad/tinygrad/pull/4889

GabrielZCode:
Funny , it's slower for you with UOPS_WRITE=0 , on my machine (i9 1900H) on this command :  PROFILE=0 python3 test/external/external_benchmark_schedule.py  is `
***** model forward in  35.23 ms
***** model schedule in  12.32 ms
***** model lower in 18468.31 ms
`

and with UOPS_REWRITE=0  PROFILE=0 python3 test/external/external_benchmark_schedule.py, it's way faster --> `
***** model forward in  41.56 ms
***** model schedule in  13.65 ms
***** model lower in 1410.08 ms
`

abgita:
Another small gain https://github.com/tinygrad/tinygrad/pull/5047

DKormann:
not sure if this is possible just by rewriting matcher and graph rewrite.
printing the time spend in graph_rewrite i get only 353ms out of 1068ms total for model lower

chenyuxyz:
this was considered done and bounty awarded. there's a new $400 bounty for 2X the speed

PR #5152, opened 2024-06-26
[NOMERGE] 49% faster pattern matcher (passes process replay)

roelofvandijk:
Some work towards: https://github.com/tinygrad/tinygrad/issues/4878. Should now be at~60% of the baseline.

* Caches on UOp methods
* No unhashable members of UPat (set, list -> tuple), create permutations on pat construction
* Replace `__unmatch` with `in` check
* Check op and src first in `_match`
* pull cached rewrite function out of double while loop

Once all usages of UPat are changed, most of the `__post_int__` can go and the typing can be fixed again to simpler tyes (`Optional[Union[DType, set(DType]]` -> `Tuple[DType,...]`. I held off for now because this would make the diff explode.

## Timing 
759.37/509.61 = 1.49

`PROFILE=0 python -O test/external/external_benchmark_schedule.py` best of 5

master
```
***** model forward in  15.84 ms
***** model schedule in   5.33 ms
***** model lower in 759.37 ms
```
branch
```
***** model forward in  15.84 ms
***** model schedule in   5.35 ms
***** model lower in 509.61 ms
```


tinygrad/codegen/uops.py
```diff
@@ -307,30 +315,22 @@ def print(self):
...
-    # recursive rewrite
-    changed = getenv("UOPS_REWRITE", 1)
+  @functools.lru_cache(None)  # pylint: disable=method-cache-max-size-none
+  def recursive_rewrite(self, pm: PatternMatcher, up:UOp) -> UOp:
+    recurse_cnt = 0
+    while (rewritten:=pm.rewrite(up)) is not None:
+      assert recurse_cnt < 100, f"recursive_rewrite looped {up} <--> {rewritten}"
+      up = rewritten
+      recurse_cnt += 1
+    up = UOp(up.op, up.dtype, tuple(self.recursive_rewrite(pm, x) for x in up.src), up.arg)
```
roelofvandijk, line 325: Create a new one, no need to clear the caches. This change also allows UOp to become a frozen dataclass (which is a tad slower though).

tinygrad/codegen/uops.py
```diff
@@ -60,10 +60,12 @@ def ge(self, x): return -self.lt(x)
   def max(self, x): return UOp.alu(BinaryOps.MAX, self, x)
   def min(self, x): return -UOp.alu(BinaryOps.MAX, -self, -x)
   @staticmethod
+  @functools.lru_cache(None)
```
roelofvandijk, line 63: These two caches give a big speedup.

tinygrad/codegen/uops.py
```diff
@@ -60,10 +60,12 @@ def ge(self, x): return -self.lt(x)
...
   def const(dtype:Optional[DType], b:ConstType|Variable):
     if isinstance(b, Variable): return UOp(UOps.DEFINE_VAR, dtype, (), b)
     return UOp(UOps.CONST, dtype, arg=dtypes.as_const(b, dtype) if dtype is not None else b)
   @staticmethod
+  @functools.lru_cache(None)
```
roelofvandijk, line 68: These two caches give a big speedup.

tinygrad/codegen/uops.py
```diff
@@ -307,30 +315,22 @@ def print(self):
...
+    up = UOp(up.op, up.dtype, tuple(self.recursive_rewrite(pm, x) for x in up.src), up.arg)
+    return self.nodes.setdefault(up.tuple(), up)
+
+  def graph_rewrite(self, sink, pm):
+    changed = getenv("UOPS_REWRITE", True)
     run_cnt = 0
     while changed:
-      changed = 0
-      @functools.lru_cache
-      def rewrite(u:UOp) -> UOp:
```
roelofvandijk, line 317: Pulling this out of here allows for better caching and makes the code simpler

DKormann:
caching the staticmethods of UOp ist interesting. will that create a permanent persistent caching?

roelofvandijk:
`Permanent persistent` for as long as the python execution runs. On staticmethods it won't store copies of the class instances.

chenyuxyz:
i think the goal of the bounty is to improve the algorithm, not putting the current algorithm behind caches and make immutable objects mutable

roelofvandijk:
> i think the goal of the bounty is to improve the algorithm, not putting the current algorithm behind caches and make immutable objects mutable

I agree, and I don't care about the bounty.

geohot:
Caches like that on ALU are dangerous. They will keep all parent UOps around forever.

tinygrad/codegen/uops.py
```diff
@@ -307,30 +315,22 @@ def print(self):
     for i,u in enumerate(self):
       print(f"{i:4d} {str(u.op):20s}: {str(u.dtype) if u.dtype is not None else '':25s} " f"{str([self.uops.index(x) for x in u.src]):32s} {u.arg}")
 
-  def graph_rewrite(self, sink:UOp, pm:PatternMatcher):
-    # recursive rewrite
-    changed = getenv("UOPS_REWRITE", 1)
+  @functools.lru_cache(None)  # pylint: disable=method-cache-max-size-none
```
geohot, line 318: Again, things now live forever.

geohot:
In general, I'm opposed to caches on any UOp with parents. They will make them live the whole lifetime of the Python process.

tinygrad/codegen/uops.py
geohot: Would merge a PR with just this one.

roelofvandijk: https://github.com/tinygrad/tinygrad/pull/5156

tinygrad/codegen/uops.py
roelofvandijkO: True. I see the risk - they can, however, be explicitly cleared after the graph rewrite (or after `linearize`). I added it [here](https://github.com/tinygrad/tinygrad/pull/5152/commits/4a998e665282e6dc556311cec84b0444c2b1aec5).

geohot: Complexity!

roelofvandijkO:
> Caches like that on ALU are dangerous. They will keep all parent UOps around forever. Caching const is probably okay.

I added a cache clear after the rewrite. Is that still too dangerous (in case the cache clear is removed accidentally)?

> I see some stuff in here that's wins and other stuff that's trading complexity for speed. 

As is tradition. I'll try to stick to more of the first.

> I don't like the loss of type specificity on UPat. 

That is not the end state of this PR, if I modify all the instantiations to have the correct types right away, the types come back (and will be more specific e.g. `Optional[Union[DType, set(DType]]` becomes `Tuple[DType,...]`). Diff will explode.

> Is `in` with a tuple of size 1 fast?

Faster (and simpler) than an `isinstance` check in `__unmatch`. My idea here was to define `UPat.ops` and `UPat.dtypes` which are tuples of matching `Uops` and `Dtypes`. Did not do the rename to keep the diff small. Moreover, tuples are immutable and hashable (and sets are not).

> And why are the other checks moved below the recursive src check?

Statistical optimization based on the benchmark. Ops unmatch more often, then src (with early stopping!), etc. I also found this way simpler: it's a match, unless somewhere it gets unmatched.

```
Total time: 0.584024 s
File: /Users/rvd/src/rvd/tinygrad/tinygrad/codegen/uops.py
Function: _match at line 119

   119                                           @profile
   120                                           def _match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> bool:
   121    473152      68088.0      0.1     11.7    if pat.op and uop.op not in pat.op: 
   122    179127      21690.0      0.1      3.7      return False
   123    294025      35267.0      0.1      6.0    if pat.src:
   124    425878      56474.0      0.1      9.7      for vp in pat.src if isinstance(pat.src, tuple) else [(pat.src,)*len(uop.src)]:
   125    286733      44834.0      0.2      7.7        if len(uop.src) != len(vp) and (len(uop.src) not in pat.allow_len): 
   126     37310       8040.0      0.2      1.4          return False
   127    249423      25025.0      0.1      4.3        new_store = store.copy()
   128    349887      63706.0      0.2     10.9        for uu, vv in zip(uop.src, vp):
   129    346318     134695.0      0.4     23.1          if not _match(uu, vv, new_store): break
   130                                                 else:
   131      3569        616.0      0.2      0.1          store.update(new_store)
   132      3569        283.0      0.1      0.0          break
   133    139145      33606.0      0.2      5.8      else: return False  # None of the src matches
   134    117570      17435.0      0.1      3.0    if pat.dtype and uop.dtype and uop.dtype not in pat.dtype: 
   135      2289        643.0      0.3      0.1      return False
   136    115281      20182.0      0.2      3.5    if pat.name is not None and store.setdefault(pat.name,uop) is not uop: 
   137      4860       1445.0      0.3      0.2      return False
   138    110421      14737.0      0.1      2.5    if pat.arg is not None and (uop.arg not in pat.arg if isinstance(pat.arg, (set,tuple)) else pat.arg != uop.arg): 
   139      9248       3155.0      0.3      0.5      return False
   140    101173      34103.0      0.3      5.8    return True
```

roelofvandijk:
> Caches like that on ALU are dangerous. They will keep all parent UOps around forever. Caching const is probably okay.

I added a cache clear after the rewrite. Is that still too dangerous (in case the cache clear is removed accidentally)?

> I see some stuff in here that's wins and other stuff that's trading complexity for speed. 

As is tradition. I'll try to stick to more of the first.

> I don't like the loss of type specificity on UPat. 

That is not the end state of this PR, if I modify all the instantiations to have the correct types right away, the types come back (and will be more specific e.g. `Optional[Union[DType, set(DType]]` becomes `Tuple[DType,...]`). Diff will explode.

> Is `in` with a tuple of size 1 fast?

Faster (and simpler) than an `isinstance` check in `__unmatch`. My idea here was to define `UPat.ops` and `UPat.dtypes` which are tuples of matching `Uops` and `Dtypes`. Did not do the rename to keep the diff small. Moreover, tuples are immutable and hashable (and sets are not).

> And why are the other checks moved below the recursive src check?

Statistical optimization based on the benchmark. Ops unmatch more often, then src (with early stopping!), etc. I also found this way simpler: it's a match, unless somewhere it gets unmatched.

```
Total time: 0.584024 s
File: /Users/rvd/src/rvd/tinygrad/tinygrad/codegen/uops.py
Function: _match at line 119

   119                                           @profile
   120                                           def _match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> bool:
   121    473152      68088.0      0.1     11.7    if pat.op and uop.op not in pat.op: 
   122    179127      21690.0      0.1      3.7      return False
   123    294025      35267.0      0.1      6.0    if pat.src:
   124    425878      56474.0      0.1      9.7      for vp in pat.src if isinstance(pat.src, tuple) else [(pat.src,)*len(uop.src)]:
   125    286733      44834.0      0.2      7.7        if len(uop.src) != len(vp) and (len(uop.src) not in pat.allow_len): 
   126     37310       8040.0      0.2      1.4          return False
   127    249423      25025.0      0.1      4.3        new_store = store.copy()
   128    349887      63706.0      0.2     10.9        for uu, vv in zip(uop.src, vp):
   129    346318     134695.0      0.4     23.1          if not _match(uu, vv, new_store): break
   130                                                 else:
   131      3569        616.0      0.2      0.1          store.update(new_store)
   132      3569        283.0      0.1      0.0          break
   133    139145      33606.0      0.2      5.8      else: return False  # None of the src matches
   134    117570      17435.0      0.1      3.0    if pat.dtype and uop.dtype and uop.dtype not in pat.dtype: 
   135      2289        643.0      0.3      0.1      return False
   136    115281      20182.0      0.2      3.5    if pat.name is not None and store.setdefault(pat.name,uop) is not uop: 
   137      4860       1445.0      0.3      0.2      return False
   138    110421      14737.0      0.1      2.5    if pat.arg is not None and (uop.arg not in pat.arg if isinstance(pat.arg, (set,tuple)) else pat.arg != uop.arg): 
   139      9248       3155.0      0.3      0.5      return False
   140    101173      34103.0      0.3      5.8    return True
```

roelofvandijk:
I'm leaving this up, but don't expect it to go in. I still believe the all-tuples UPat idea has merit, but it does make the UPat definitions uglier.

geohot:
Want to break this into smaller PRs with each feature? I'm supportive of the all tuples UPat if it's faster.

geohot:
How much is the reordering in match vs the UPat change? I do think the syntax looks worse now.

Before:
```
diane@blak:~/tinygrad$ PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  19.66 ms
***** model schedule in   7.47 ms
***** model lower in 798.28 ms
```

After:
```
diane@blak:~/tinygrad$ PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  20.15 ms
***** model schedule in   7.10 ms
***** model lower in 740.92 ms
```

PR #5162, opened 2024-06-26
remove upcasts heuristic in hcopt

chenyuxyz:
made hcopt llama 30%~100% faster on NV/AMD/METAL, will test all benchmarks


chenyuxyz:
meh it made conv slower, probably want better default in a different way

PR #5167, opened 2024-06-26
Fix batchnorm shapes for resnet.load_pretrained

reddyn12:
Update shapes to work with unsynced batchnorm

Need to use unsynced instead of synced to get fp16 training working for retinanet, This is also a bug which I think has to do with lack of casting in synced batchnorm. I tried fixing, but it increased step times.

Repro:
```
import functools
from typing import List
from tinygrad import dtypes, Tensor, GlobalCounters
from extra.models import resnet
from examples.mlperf.initializers import Conv2dHeNormal, Linear
from examples.hlb_cifar10 import UnsyncedBatchNorm

def _sum(x:List[Tensor]):
  ans = x[0].mean()
  for xx in x[1:]:
    ans+=xx.mean()
  return ans 

if getenv('BN', 1): resnet.BatchNorm = functools.partial(UnsyncedBatchNorm, num_devices=1)
resnet.Conv2d = Conv2dHeNormal
resnet.Linear = Linear

res = resnet.ResNeXt50_32X4D()
res.load_from_pretrained()
res.fc = None

X = Tensor.zeros((10, 3, 800, 800))

out = res.forward(X)
print(_sum(out).numpy(), dtypes.default_float)
print(f"{GlobalCounters.mem_used / 1e9:.2f} GB used")
```

```DISABLE_COMPILER_CACHE=1 DEFAULT_FLOAT=float16 BN=1 python3 fptest.py```
Change BN and def_float vals

fp16
- BN=0    inf
- BN=1    0.6377
fp32
- BN=0    0.6375836
- BN=1    0.6375836

### DIFF
--- a/extra/models/resnet.py
+++ b/extra/models/resnet.py
@@ -126,6 +126,7 @@ class ResNet:
     return self.forward(x)
 
   def load_from_pretrained(self):
+    import numpy as np
     # TODO replace with fake torch load
 
     model_urls = {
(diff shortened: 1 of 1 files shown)


chenyuxyz:
if it's just a few unqueeze, i think it's fine to just change it to `obj.assign(dat.reshape(obj.shape))`, and can remove the case for `()` since it works with reshape too.

the weight you are loading contains number that's out of range of half, which will be case into inf. what you want is to have model weight in float32, and only cast to default_float in `__call__`

PR #5170, opened 2024-06-26
Mimic unsyncedBN for fp16

reddyn12:


chenyuxyz:
don't think we want to hard code dtype in core tinygrad like that, it does not work with float64 for example. you can make your own layers like how resnet / bert did it

PR #5172, opened 2024-06-27
 Handling Multiple UnaryOps.BITCAST in Function for Proper Kernel Fusion [run_process_replay]

hikettei:
# Description

This is a prereq for the approximation bounty https://github.com/tinygrad/tinygrad/pull/5187.

I am implementing `sin/log2/exp2` using a combination of UOps. I noticed that the kernel does not fuse properly when bitcast is performed more than twice within the `forward` method of the `Function` class.

```python
# in test/test_schedule.py
class CycleBitcast(Function):
  def bitwise_cast(self, x: LazyBuffer):
    return x.cast(dtypes.int32, True)

  def forward(self, x: LazyBuffer):
    x = x.cast(dtypes.float32)
    a = self.bitwise_cast(x)
    b = self.bitwise_cast(x.e(UnaryOps.NEG))
    return a.e(BinaryOps.ADD, b)

check_schedule(b, 1) # => test.test_schedule.KernelCountException: len(sched)=2 != 1
```

Additionally, using `TernaryOps.WHERE` after an unfused bitcast likely fails due to sorting issues, resulting in `RuntimeError: cycle detected in graph, prescheduled 4 but only scheduled 0`. (see [code reference](https://github.com/hikettei/tinygrad/blob/c28fd4dd78cff572325662942bf7bb8e01a33111/tinygrad/fastmath.py#L85-L91)).

To address this, I added the `bitcast_no_fuse` option. When working within  `Function`, it disables the optimization that rewrites `UnaryOps.BITCAST` to `LoadOps.VIEW`. ([code in master](https://github.com/hikettei/tinygrad/blob/c28fd4dd78cff572325662942bf7bb8e01a33111/tinygrad/lazy.py#L110)

This ensures proper scheduling and enables kernel fusion for the bounty.


geohot:
Hmm, is there an obvious time to use and not use this flag? Is this fusion selectable automatically when it makes sense?

hikettei:
> is there an obvious time to use and not use this flag?

I am using it here, for example: https://github.com/tinygrad/tinygrad/pull/5187/files#diff-31374348b3f46c9071e557ee494a0a497a7dfbfde4fc6190bacc0ece3118f23cR82

I was trying to solve the issue where using multiple Bitcast operations within `def forward` causes a `RuntimeError: cycle detected in graph, prescheduled 4 but only scheduled 0`. During this process, I accidentally discovered that disabling the conversion of BITCAST to VIEW operations allows scheduling to proceed successfully.

(w/o this option, in `JIT=2 LLVM=1 python -m pytest test/test_scheduler.py`)
image

I use this option when performing bitcast within `def forward`.

Qazalin:
I don't think this is the right fix,
@hikettei can you share a commit hash that actually produces these errors. I'm running tests against 24a56f4cfd6844612a4b82b05aa43837a66deca6 and getting several api-level errors.

hikettei:
@Qazalin Thank you for taking a look! Please refer to https://github.com/tinygrad/tinygrad/pull/5187/commits/3210764e0c3b0479161786958de5bfc543c42a33. I have removed all API changes in the latest commit.

geohot:
@Qazalin can you repro scheduler cycles?

Qazalin:
Ran test_schedule against https://github.com/tinygrad/tinygrad/commit/3210764e0c3b0479161786958de5bfc543c42a33 - no failing tests. @hikettei what is required in the diff to show the cycles?

test/test_schedule.py
```diff
@@ -40,6 +42,22 @@ def check_schedule(t:Union[Tensor, List[Tensor]], allowed:int, to_prerealize:Opt
     l.linearize()
   return sched
 
+class CycleBitcast(Function):
```
Qazalin, line 45: I don't think this is a good test. If I run:
```
  def bitwise_cast(self, x: LazyBuffer):
    return x.cast(dtypes.int32, True)
```
instead it also passes with only 1 ScheduleItem.

hikettei:
sorry I forgot to mention that: this is only reproduced on VIEW_SUPPORTED_DEVICES
image

Qazalin:
meh, I think LoadOps.VIEW should go in the scheduler. This isn't related to your bounty though, fine with this.

PR #5173, opened 2024-06-27
Introduce BinaryOps.AND/OR/SHR/SHL to all runtime.

hikettei:
# Description

This is a prereq for the approximation bounty https://github.com/tinygrad/tinygrad/pull/5187.

# Changes

- Added `BinaryOps.AND` and `BinaryOps.OR` for fast and safe bitwise calculations.
- Added `Tensor.bitwise_and` and `Tensor.bitwise_or`.
- Added support for `BinaryOps.SHL` and `BinaryOps.SHR` in tinygrad/renderer/cstyle.py.
- (EDIT) Modified `Tensor.lshift/Tensor.rshift` to use bit shift operations instead of pow.

# TODO

- [x] Add int64/uint64 support for AMD (something related to REMU?)


Qazalin (commented):
[released](https://github.com/Qazalin/remu/commit/ccea0b1a2e8a64e282f34ab3199738c27ecc0de0), thanks for the patience!

tinygrad/tensor.py
```diff
@@ -2447,6 +2447,34 @@ def div(self, x:Union[Tensor, ConstType], reverse=False, upcast=True) -> Tensor:
     if upcast: numerator, denominator = numerator.cast(least_upper_float(numerator.dtype)), denominator.cast(least_upper_float(denominator.dtype))
     return F.Div.apply(numerator, denominator)
 
+  def bitwise_and(self, x:Union[Tensor, ConstType], reverse=False) -> Tensor:
```
Qazalin, line 2450: do you wanna add an assert for self.dtype.is_int?

hikettei: yes. i have added that one.

hikettei:
Thank you Qazalin for the patch! and I think it is ready for the merge if all tests passed

geohot:
1. Give this PR a better name
2. You add SHL/SHR, but they aren't tested
3. Pay attention to spacing and match the surrounding lines

geohot:
So I thought about it more. Why do you need to expose SHL and SHR? Isn't MUL and IDIV exactly the same thing?

hikettei:
I just experimented. If we don't consider speed, SHL/SHR were completely unnecessary. It was completely my misunderstanding. (And probably the compiler will optimize the multiplication)

I was under the impression that there was no operator for exact calculations with pow when the right-hand side is a LazyBuffer, but I had forgotten that I implemented `pow2if` myself. I have removed all instances of SHL/SHR.

Adding BinaryOps.AND/BinaryOps.OR is the only change needed. I will create a new PR once the tests are completed.

Thank you very much for pointing this out.

PR #5176, opened 2024-06-27
ruff: remove unnecessary getattr or setattr

roelofvandijk:
https://docs.astral.sh/ruff/rules/get-attr-with-constant/
https://docs.astral.sh/ruff/rules/set-attr-with-constant/

~Rules not enabled.~


chenyuxyz:
why is the rule not enabled?

roelofvandijk:
Closing this - the rule needs exceptions.

PR #5178, opened 2024-06-27
ruff rule if-exp-instead-of-or-operator (FURB110)

roelofvandijk:
⚠️ This is a preview rule https://docs.astral.sh/ruff/rules/if-exp-instead-of-or-operator

I can take out the rule again.


geohot:
I should use the other style, merged.

PR #5180, opened 2024-06-27
ruff: close file handle

roelofvandijk:
These file handles are not closed.

SIM115 tripped on them.
https://docs.astral.sh/ruff/rules/open-file-with-context-handler/

### DIFF
--- a/tinygrad/engine/realize.py
+++ b/tinygrad/engine/realize.py
@@ -37,7 +37,9 @@ def get_linearizer(renderer:Renderer, ast:Tuple[LazyOp, ...]) -> Linearizer:
         k = timed[0][1]
         if logkerns is not None and logkerns_level > 1: logkerns.writelines([f"{(lin.ast, lin.applied_opts)}\n" for (_,lin,_) in timed[1:]])
   # TODO: check the correctness inline once compare_linearizer is in core
-  if logkerns is not None: logkerns.writelines([f"{(k.ast, k.applied_opts)}\n"])
+  if logkerns is not None:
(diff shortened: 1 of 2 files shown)


tinygrad/engine/realize.py
```diff
@@ -37,7 +37,9 @@ def get_linearizer(renderer:Renderer, ast:Tuple[LazyOp, ...]) -> Linearizer:
         k = timed[0][1]
         if logkerns is not None and logkerns_level > 1: logkerns.writelines([f"{(lin.ast, lin.applied_opts)}\n" for (_,lin,_) in timed[1:]])
   # TODO: check the correctness inline once compare_linearizer is in core
-  if logkerns is not None: logkerns.writelines([f"{(k.ast, k.applied_opts)}\n"])
+  if logkerns is not None:
+    logkerns.writelines([f"{(k.ast, k.applied_opts)}\n"])
+    logkerns.close()
```
chenyuxyz, line 42: this should stay opened after writelines, there can be multiple calls to this function

tinygrad/engine/schedule.py
```diff
@@ -312,11 +312,12 @@ def create_schedule_with_vars(outs:List[LazyBuffer], seen:Optional[Set[LazyBuffe
     for x in graph[ps.outputs[0]]:
       in_degree[x] -= 1
       if in_degree[x] == 0: queue.append(prescheduled[x])
+  if logops: logops.close()
```
geohot, line 315: I don't think this is right.

roelofvandijk: Removed it.

PR #5181, opened 2024-06-27
ruff: in-dict-keys

roelofvandijk:
https://docs.astral.sh/ruff/rules/in-dict-keys/


chenyuxyz:
i wanted to do this before but sometimes it's less readable, especially when it's not immediately clear that the iterator is a dict

geohot:
Why are there exceptions to the rule?

roelofvandijk:
`safetensors.safe_open` does not support direct iteration over the keys like python `dict`s

geohot:
I'm not dealing with exceptions. keys is fine

PR #5182, opened 2024-06-27
ruff: suppressible-exception

roelofvandijk:
This is ruff SIM105
https://docs.astral.sh/ruff/rules/suppressible-exception/

Did not enable the rule.


chenyuxyz:
cool! rules not enabled due to potential perf reason?

PR #5189, opened 2024-06-27
benchmark use JITBEAM for llama and gpt2

chenyuxyz:
try this again


chenyuxyz:
oh those are the `clear_l2` kernels

chenyuxyz:
speed is similar now

PR #5190, opened 2024-06-27
CACHECOLLECTING -> CAPTURING and don't capture clear_l2

chenyuxyz:
fixed first time BEAM slowness


chenyuxyz:
llama3-8B was using JITBEAM, the tok/s improved from 40 -> 50

PR #5192, opened 2024-06-27
use NV instead of CUDA in benchmark

chenyuxyz:
also reenabled mixtral on green


chenyuxyz:
@nimlgen fyi. it's almost faster for all, except some cifar that's < 10% slower. NV also makes benchmark time 20% faster

PR #5194, opened 2024-06-27
ruff: superfluous-else-return

roelofvandijk:
https://docs.astral.sh/ruff/rules/superfluous-else-return/


geohot:
All in tests.

roelofvandijk:
Enabling the linter rule would enforce that they stay gone.

PR #5195, opened 2024-06-27
ruff: unnecessary-assign

roelofvandijk:
https://docs.astral.sh/ruff/rules/unnecessary-assign/


geohot:
This is becoming golf

PR #5199, opened 2024-06-27
remove uopgraph dedup + slight speedup

DKormann:


DKormann:
the new graph_rewrite looks a lot like my recursive PR #5093 and at the same time only minimal changes to the current graph_rewrite

roelofvandijk:
```
def __inner_rewrite(n: UOp) -> UOp:
  if n not in replace:
    key = (n.op, n.dtype, tuple(__inner_rewrite(s) for s in n.src), n.arg)
    new_n = UOp(*key)
    replace[n] = nodes.setdefault(key, pm.rewrite(new_n) or new_n)
  return replace[n]
```

DKormann:
dont think it works here

geohot:
Nice!

PR #5203, opened 2024-06-27
experiment: in search, don't apply actions that caused slowdown

chenyuxyz:
testing... cut search time by half, no perf diff in llama and gpt2


geohot:
Ideally we move to a new kind of search that can use probability, then downweight actions that slowed things down. We can also add a prior to that search based on which actions are generally good.

chenyuxyz:
yea this was a quick way to test that direction, equivalent to 0/1 probability and downweight 100%. It works surprising well so  will think more about the full change (no plan to merge this). And it shows the issue with TC or multi-step actions in general

also thinking about an offline search decision tree like structure that can replace hcopt

PR #5206, opened 2024-06-28
Add SDXL Inference to Examples

tobias17:
## Overview

Brings in SDXL 1.0 Base inference code for image generation.

The code currently still has all of my references, can take them out if that's wanted.

## Performance

On my 4060ti 16 GB it takes ~8 seconds for the first step and ~4 seconds for each additional step at 1024x1024 resolution.

## Example Generations
base_floating_islands

Custom image sizes is also supported:
base_mountain_king

Custom model weights can easily be used (Juggernaut XL in this case):
jug_tree

jug_flame_villain

### DIFF
--- a/examples/sdxl.py
+++ b/examples/sdxl.py
@@ -1,22 +1,19 @@
 # This file incorporates code from the following:
 # Github Name                    | License | Link
-# tinygrad/tinygrad              | MIT     | https://github.com/tinygrad/tinygrad/blob/64cda3c481613f4ca98eeb40ad2bce7a9d0749a3/LICENSE
 # Stability-AI/generative-models | MIT     | https://github.com/Stability-AI/generative-models/blob/fbdc58cab9f4ee2be7a5e1f2e2787ecd9311942f/LICENSE-CODE
 # mlfoundations/open_clip        | MIT     | https://github.com/mlfoundations/open_clip/blob/58e4e39aaabc6040839b0d2a7e8bf20979e4558a/LICENSE
 
-from tinygrad.tensor import Tensor, dtypes
+from tinygrad import Tensor, TinyJit, dtypes
 from tinygrad.nn import Linear, Conv2d, GroupNorm, LayerNorm, Embedding
 from tinygrad.nn.state import safe_load, load_state_dict
-from tinygrad.helpers import fetch
-from tinygrad import TinyJit
+from tinygrad.helpers import fetch, trange
+from examples.stable_diffusion import ClipTokenizer, ResnetBlock, Mid, Downsample, Upsample
 import numpy as np
 
 from typing import Dict, List, Union, Callable, Optional, Any, Set, Tuple
-import os, math, re, gzip, argparse, tempfile
+import math, argparse, tempfile
 from abc import ABC, abstractmethod
-from functools import lru_cache
 from pathlib import Path
-from tqdm import trange
 from PIL import Image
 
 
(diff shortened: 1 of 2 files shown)


chenyuxyz:
cool! can you import the components from stable_diffusion examples directly to reuse as many as possible?

examples/sdxl.py
```diff
@@ -0,0 +1,1107 @@
+# This file incorporates code from the following:
+# Github Name                    | License | Link
+# tinygrad/tinygrad              | MIT     | https://github.com/tinygrad/tinygrad/blob/64cda3c481613f4ca98eeb40ad2bce7a9d0749a3/LICENSE
+# Stability-AI/generative-models | MIT     | https://github.com/Stability-AI/generative-models/blob/fbdc58cab9f4ee2be7a5e1f2e2787ecd9311942f/LICENSE-CODE
+# mlfoundations/open_clip        | MIT     | https://github.com/mlfoundations/open_clip/blob/58e4e39aaabc6040839b0d2a7e8bf20979e4558a/LICENSE
+
+from tinygrad.tensor import Tensor, dtypes
```
chenyuxyz, line 7: from tinygrad import Tensor, TinyJit, dtypes

also from tinygrad.helpers import trange should work

examples/sdxl.py
```diff
@@ -0,0 +1,1107 @@
...
+    return model.conditioner(batch_c), model.conditioner(batch_uc, force_zero_embeddings=["txt"])
+
+  # https://github.com/Stability-AI/generative-models/blob/fbdc58cab9f4ee2be7a5e1f2e2787ecd9311942f/sgm/modules/diffusionmodules/denoiser.py#L42
+  def denoise(self, x:Tensor, sigma:Tensor, cond:Dict) -> Tensor:
+
+    def sigma_to_idx(s:Tensor) -> Tensor:
+      dists = s - Tensor(self.sigmas).unsqueeze(1)
+      return dists.abs().argmin(axis=0).view(*s.shape)
+
+    sigma = Tensor(self.sigmas[sigma_to_idx(sigma).numpy()])
```
chenyuxyz, line 963: is there a reason this is convert from Tensor to numpy then Tensor?

examples/sdxl.py
```diff
@@ -0,0 +1,1107 @@
...
+      x, penultimate = r(x, attn_mask=attn_mask), x
+    return x.permute(1,0,2), penultimate.permute(1,0,2)
+
+  def __call__(self, text:Tensor) -> Tensor:
+    tokens = Tensor(self.tokenizer.encode(text, pad_with_zeros=True), dtype=dtypes.int64).reshape(1,-1)
+
+    x = self.model.token_embedding(tokens).add(self.model.positional_embedding).permute(1,0,2)
+    x, penultimate = self.text_transformer_forward(x, attn_mask=self.model.attn_mask)
+    x = self.model.ln_final(x)
+    pooled = x[Tensor.arange(x.shape[0]), tokens.argmax(axis=-1).numpy().item()] @ self.model.text_projection
```
chenyuxyz, line 672: is this `numpy()` needed?

chenyuxyz (commented):
overall have you encountered anything that's missing from the core tinygrad?

tobias17:
> cool! can you import the components from stable_diffusion examples directly to reuse as many as possible?

would you prefer I directly import from one to the other? or make a stable_diffusion folder in examples and pull out some of the common components that both import?

I also started from scratch since SD 1.5 was created in a completely different repo than all of the models after it, so some of the things are _slightly_ off one another, but will take a crack and finding the overlap.

tobias17:
> overall have you encountered anything that's missing from the core tinygrad?

Everything worked pretty smoothly, most of my confusion during the porting process came from the source having spaghetti logic and the general pytorch complexity (the MultiheadAttention took me a while since they store KVQ weights and biases as single blocks).

The one thing that was used a lot in the source and I did not see in tinygrad was the [einops rearrange function](https://github.com/Stability-AI/generative-models/blob/fbdc58cab9f4ee2be7a5e1f2e2787ecd9311942f/sgm/modules/attention.py#L711) which seems like a really useful and readable way to perform permutes and reshapes. This is something I am considering investigating next to see if I can make a clean version for tinygrad.

chenyuxyz:
i think if you import einops it would work with tinygrad Tensor, if that helps with readability we can start from there and write a tinygrad version later

probably fine to just import from stable_diffusion if it's exactly the same, if it's slightly off then don't overdo it

tobias17:
> i think if you import einops it would work with tinygrad Tensor, if that helps with readability we can start from there and write a tinygrad version later

I wrote the permutes and reshapes by hand in my port and it's good enough for now, just thought it might be a nice QoL feature to have in the framework.

tobias17:
@chenyuxyz I was able to directly import a few things, the original is just hard coded in many places with sizes and other various assumptions, still got some good lift though.

examples/sdxl.py
tobias17: 4aef32d

examples/sdxl.py
tobias17: i was originally suspicious of it causing problems with the Jit along with it being a Tensor not allowing me to load the state dict in strict mode.

I have removed it in 4aef32d along with turning the state dict load strict to False. Is there an easy way to not have the state dict pick up certain tensors?

examples/sdxl.py
tobias17: Gave me problems originally, but apparently works without now, removed in 4aef32d

chenyuxyz:
works on my mac! default prompt step 10.

merging this, and there's a $1000 bounty to train mlperf stable diffusion if you are interested

rendered

PR #5210, opened 2024-06-28
reuse more sd in sdxl

chenyuxyz:
@tobias17 I think there are more that can be reused. I replaced the identical ones but UNet and openai Clip might be tweaked a bit to reuse more


chenyuxyz:
i think it's better to rewrite the stable_diffusion using the unet in sdxl

PR #5061, opened 2024-06-19
generic gate folding

Qazalin:
This diff folds symbolic NumNode 0 and 1 and CONST literal True and False gates.
After the symbolic removal we should always render a literal boolean for valid and remove the `None` in UPat.

PR #5214, opened 2024-06-28
test: valid is always bool [run_process_replay]

Qazalin:
Prevents the Linearizer from rendering NumNode(0) and NumNode(1) for valids. Now all non-bool gates are asserted in the UOps type checker.

### DIFF
--- a/tinygrad/codegen/linearizer.py
+++ b/tinygrad/codegen/linearizer.py
@@ -215,7 +215,7 @@ class Linearizer(Kernel):
       if self.late_gate is not None: valid *= self.late_gate
       # TODO: let UPat check this once it's fast
       if valid.min == 1: stores.append(UOp(UOps.STORE, None, (buf_uop, rendered_idx, var)))
-      else: stores.append(UOp(UOps.STORE, None, (buf_uop, rendered_idx, var, valid.render(render_ops, self.loop_uops).cast(dtypes.bool))))
+      else: stores.append(UOp(UOps.STORE, None, (buf_uop, rendered_idx, var, valid.render(render_ops, self.loop_uops))))
     return stores
 
   # render loop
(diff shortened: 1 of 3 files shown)


geohot:
Want to add an assert to the UOps type checker that confirms they are always bool?

Qazalin:
yea this added it https://github.com/tinygrad/tinygrad/pull/5061

if the uops rewriter is working correctly it should never hit this but looks like it does in multithreading retinanet.

geohot:
Yea this should definitely be fixed in the Linearizer, but also...why isn't the rewriter working?

reddyn12:
> but looks like it does in multithreading retinanet.

I didn't root cause, but multitreading on single gpu worked. Multithreading on multi gpu hit that assert. Try running train with TEST=1 on multigpu to see if multithreading is a problem

Qazalin:
@reddyn12 the dataset is still downloading for me, wanna try running it on this branch?

reddyn12:
I don't have access to a multigpu setup rn, but TEST=1 doesn't need any dataset, uses fake data

tinygrad/codegen/linearizer.py
```diff
@@ -141,7 +141,8 @@ def global_load(self, i:int, idxs:List[Node], acc:Optional[LazyOp]=None, barrier
     invalid_value = 0
     acc_count = 0
     for idx, valid, rep_idx in zip(e_idxs, e_valids, iter_idxs(expand_vars)):
-      this_const, idx, valid = (invalid_value, NumNode(0), NumNode(1)) if valid.max == 0 else (const, idx, valid)
+      this_const, idx = (invalid_value, NumNode(0)) if valid.max == 0 else (const, idx)
+      valid_uop = UOp.const(dtypes.bool, valid.b) if valid.min == valid.max else valid.render(render_ops, self.loop_uops)
```
Qazalin, line 145: This felt like a better check, if it renders a non-bool ALU as valid it should be fixed in symbolic.

PR #5219, opened 2024-06-28
simpler TinyJit inside TinyJit detection

chenyuxyz:
suggested in https://github.com/tinygrad/tinygrad/commit/73395b998b6bb5c7a5c4cabde72b29a81ab51ead#commitcomment-143660402

### DIFF
--- a/tinygrad/engine/jit.py
+++ b/tinygrad/engine/jit.py
@@ -3,7 +3,7 @@ from typing import TypeVar, Generic, Callable, List, Tuple, Union, Dict, cast, O
 import functools, itertools, collections
 from tinygrad.tensor import Tensor
 from tinygrad.lazy import LazyBuffer
-from tinygrad.helpers import flatten, merge_dicts, DEBUG, Context, ContextVar, GRAPH, BEAM, getenv, all_int, GraphException, colored, JIT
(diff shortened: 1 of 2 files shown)


geohot:
Oh interesting. We use capturing in some of the tests, maybe we forget to clear it.

chenyuxyz:
The failure was due to capturing not clearing if an exception was raised (as expected) during capture step. I think capturing is closer to a Context than a list, and can be refactored this way

tinygrad/engine/jit.py
```diff
@@ -146,20 +145,24 @@ def __call__(self, *args, **kwargs) -> ReturnType:
...
-        self.ret = self.fxn(*args, **kwargs)
-        if len(params:=get_parameters(self.ret)): Tensor.realize(params[0], *params[1:])
-        capturing.clear()
+        try:
+          self.ret = self.fxn(*args, **kwargs)
+          if len(params:=get_parameters(self.ret)): Tensor.realize(params[0], *params[1:])
+          capturing.clear()
+        except Exception as e:
+          capturing.clear()
+          raise e
```
geohot, line 165:
```
finally:
  capturing.clear()
```

geohot:
Can we merge this? (and any reason to not switch to finally?)

chenyuxyz:
yea let me update and merge, i missed your reply

PR #5228, opened 2024-06-29
tqdm add back unit scale e2e test

chenyuxyz:
no longer flaky after unit scale fix. it was removed https://github.com/tinygrad/tinygrad/pull/4846/commits/d392cc931cdf811f47894cb571b42cfe117ccd15


chenyuxyz:
it looks like something else is calling `time.perf_counter`, even if i can get this to work, mocking `time.perf_counter` seems too brittle

PR #5235, opened 2024-06-30
Compare State Dicts Helper Function

tobias17:
## Overview

Proposed up-streaming of helper function that takes in 2 state dicts and prints which keys are missing from another.

Very useful when implementing an existing model (like StableDiffusion) for inference to check for weight key mismatching.

## Testing

Note: only the keys are used by the function, accepts full state dict for easier calling. This function was also properly tested with a full model and weights file.

```python
>>> compare_state_dicts({"proj_in.weight":0,"proj_in.bias":0}, {"proj_in.weight":0,"proj_in.bias":0})
Both state dicts contain the same keys
```

```python
>>> compare_state_dicts({"proj_in.weight":0,"proj_in.bias":0,"sigmas":0}, {"proj_in.weight":0,"proj_in.bias":0})

Model Only:
sigmas

```

```python
>>> compare_state_dicts({"proj_in.weight":0,"proj_in.bias":0}, {"proj_in.weight":0,"proj_in.bias":0,"pos_ids":0,"tok_ids":0})

Weights Only:
pos_ids
tok_ids

```

```python
>>> compare_state_dicts({"proj_in.weight":0,"proj_in.bias":0,"sigmas":0}, {"proj_in.weight":0,"proj_in.bias":0,"pos_ids":0,"tok_ids":0})

Model Only:
sigmas

Weights Only:
pos_ids
tok_ids

```

### DIFF
--- a/tinygrad/nn/state.py
+++ b/tinygrad/nn/state.py
@@ -129,6 +129,15 @@ def load_state_dict(model, state_dict:Dict[str, Tensor], strict=True, verbose=Tr
       else: v.replace(state_dict[k].to(v.device)).realize()
       if consume: del state_dict[k]
 
+def compare_state_dicts(left_state_dict:Dict, right_state_dict:Dict, left_name:str="Model", right_name:str="Weights") -> None:
+  """
+  Compares 2 state dicts, printing which keys are missing from another for debugging purposes.
+  """
+  left_keys, right_keys = set(left_state_dict.keys()), set(right_state_dict.keys())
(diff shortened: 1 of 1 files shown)


chenyuxyz:
Generally we only add API to nn after it's used several times in examples, or if there's a matched API in torch that people use a lot. There's a cost to maintain these API

tobias17:
> Generally we only add API to nn after it's used several times in examples, or if there's a matched API in torch that people use a lot. There's a cost to maintain these API

Fair enough, thought I would offer in case it was wanted, will decline PR.

PR #5236, opened 2024-06-30
fix: message when applegpu tools missing

gip:
Show error message when applegpu tools is missing.

Happy to add a section in the docs on how to install applegpu tools in another PR. Although if we do that it would make sense to also add how to install tools for other arch.


tinygrad/runtime/ops_metal.py
```diff
@@ -30,10 +30,14 @@ class MetalProgram:
...
-        shader.flush()
-        os.system(f"cd {pathlib.Path(__file__).parents[2]}/extra/disassemblers/applegpu && python3 compiler_explorer.py {shader.name}")
+      applegpu_path = pathlib.Path(__file__).parents[2] / "extra" / "disassemblers" / "applegpu"
+      if not applegpu_path.exists():
+        print("Apple GPU tools not installed: skipping Metal assembly output")
+      else:
+        with tempfile.NamedTemporaryFile(delete=True) as shader:
+          shader.write(lib)
+          shader.flush()
+          os.system(f"cd {applegpu_path} && python3 compiler_explorer.py {shader.name}")
```
nimlgen, line 40: Can we just check return code of os.system instead of `.exists()`? And just print a nicer error message like: "Error running disarm: <err test from os.system>. Make sure you have https://github.com/dougallj/applegpu cloned to tinygrad/extra/disassemblers or env var APPLEGPU_DISASM_PATH set to correct path` (this is just an example :))

gip: Updated. Note that I can't test on Windows (and os behavior is different there I think).

PR #5237, opened 2024-06-30
Add x!=0 -> (bool)x pattern [run_process_replay] [no_assert]

cpwrs:
Replaces x != 0 pattern with x.cast(bool)


geohot:
I see `(bool)(val0)!=1` in process replay now

cpwrs:
I think.
`x != 0 != 1` is `not(bool(x))`,
I could add a bool comparison pattern to cast subsequent CMPNE down

PR #5239, opened 2024-06-30
fix stable_diffusion with fp16

chenyuxyz:


chenyuxyz:
TODO: --fp16 uses more memory now

PR #5245, opened 2024-07-01
rename opts to renderer where it refers to a Renderer instance

DKormann:
this was super confusing to me trying to understand kernel.py any opts variable mostly refers to list of OptOps or similar but sometimes to a renderer instead


DKormann:
wont quite pass process replay since the unpickled Linearizer expects arg 'opts'

Qazalin:
opts is more like linearization options for specifying features but i get where you're coming from

DKormann:
yeah i tried to not rename it where that is the case. But Kernel.opts is a Renderer. or do i understand it wrong?

PR #5248, opened 2024-07-01
revert stable diffusion validation with threefry

chenyuxyz:
need to fix threefry before enabling this again.


chenyuxyz:
@tobias17 fyi, should fix your repro issue

PR #5250, opened 2024-07-01
Move RMSNorm to nn.RMSNorm.

JhennerTigreros:
I found only one official implementation of RMSNorm for PyTorch here https://pytorch.org/torchtune/stable/_modules/torchtune/modules/rms_norm.html#RMSNorm, but there is a problem installing torchtune in CI process due to an issue with setuptools, I just copy-paste the implementation inside the test function. it is ok?


tinygrad/nn/__init__.py
```diff
@@ -209,6 +209,33 @@ def __call__(self, x:Tensor):
...
+  - Described: https://paperswithcode.com/method/rmsnorm
+  - Paper: https://arxiv.org/pdf/1910.07467
+
+  ```python exec="true" source="above" session="tensor" result="python"
+  t = Tensor.rand(2, 5) * 2 + 1
+  rms_norm = nn.RMSNorm(t.size(-1), 1e-6)
+  print(t.std().item())
+  ```
+  ```python exec="true" source="above" session="tensor" result="python"
+  t = rms_norm(t)
```
geohot, line 225: ??

tinygrad/tensor.py
```diff
@@ -2693,6 +2693,25 @@ def sequential(self, ll:List[Callable[[Tensor], Tensor]]):
     """
     return functools.reduce(lambda x,f: f(x), ll, self)
 
+  def rmsnorm(self, axis=-1, eps:float=1e-6) -> Tensor:
+    """
+    Applies Root Mean Square Layer Normalization over a summed inputs to a neuron in one layer.
```
geohot, line 2698: Typo

JhennerTigreros:
Sorry, I just want to check the NV and GPU tests to check if the PyTorch class works as expected, can't run locally. Will reopen when made all the needed changes.

tinygrad/nn/__init__.py
JhennerTigreros: I don't understand what's the problem here. Should rms_norm be renamed to norm?

PR #5251, opened 2024-07-01
Move RMSNorm to nn.RMSNorm

JhennerTigreros:
Sorry for the other PRs, this should be fine. 

I found only one official implementation of RMSNorm for PyTorch here https://pytorch.org/torchtune/stable/_modules/torchtune/modules/rms_norm.html#RMSNorm, but there is a problem installing torchtune in CI process due to an issue with setuptools, I just copy-paste the implementation inside the test function. it is ok?


tinygrad/tensor.py
```diff
@@ -2693,6 +2693,25 @@ def sequential(self, ll:List[Callable[[Tensor], Tensor]]):
...
 
+  def rmsnorm(self, axis=-1, eps:float=1e-6) -> Tensor:
+    """
+    Applies Root Mean Square Layer Normalization over the summed inputs to a neuron in one layer.
+
+    - Described: https://paperswithcode.com/method/rmsnorm
+    - Paper: https://arxiv.org/pdf/1910.07467
+
+
+    ```python exec="true" source="above" session="tensor" result="python"
```
geohot, line 2704: This is testing randn?

geohot:
Sorry, I'm not giving more feedback on this.

PR #5253, opened 2024-07-02
Pulled CLIP and UNet into Seperate Files

tobias17:
## Overview

The abstracted CLIP and UNet models got pulled into their own respective files in the `extra/models` folder.

`examples/stable_diffusion.py` and `examples/sdxl.py` both had their implementations of these models removed and reference the common model.

## Testing
```sh
$ NV=1 python3 examples/stable_diffusion.py --seed 0 --noshow
output validated with distance=7.163018835854018e-07
```

```sh
$ NV=1 python3 examples/sdxl.py --seed 0 --noshow
output validated with distance=0.0012854303931817412
```

### DIFF
--- a/extra/models/clip.py
+++ b/extra/models/clip.py
@@ -316,8 +316,7 @@ class Open:
       x = self.transformer(x, attn_mask=self.attn_mask)
       x = self.ln_final(x)
 
-      pooled = x[Tensor.arange(x.shape[0]), text.argmax(dim=-1)]
-      pooled = pooled @ self.text_projection
+      pooled = x[:, text.argmax(dim=-1)] @ self.text_projection
       return pooled
 
 
--- a/extra/models/clip.py
+++ b/extra/models/clip.py
@@ -1,0 +1,366 @@
+from tinygrad import Tensor, dtypes
+from tinygrad.helpers import fetch
+from tinygrad.nn import Linear, LayerNorm, Embedding
+
+from typing import List, Optional, Union, Tuple
+from abc import ABC, abstractmethod
+from functools import lru_cache
+import re, gzip
+
+class Tokenizer:
+  """
+  Namespace for CLIP Text Tokenizer components.
+  """
+
+  @staticmethod
+  def get_pairs(word):
+    """
+    Return set of symbol pairs in a word.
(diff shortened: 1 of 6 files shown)


extra/models/clip.py
```diff
@@ -0,0 +1,366 @@
...
+  def __call__(self, text:str) -> Union[Tensor,Tuple[Tensor,...]]:
+    pass
+
+
+class Closed:
+  """
+  Namespace for OpenAI CLIP model components.
+  """
+
+  # https://github.com/tinygrad/tinygrad/blob/64cda3c481613f4ca98eeb40ad2bce7a9d0749a3/examples/stable_diffusion.py#L329
```
chenyuxyz, line 133: i think these comments that pin to stable diffusion become obsolete after the refactor

extra/models/clip.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def __call__(self, text:Tensor) -> Tensor:
+      seq_len = text.shape[1]
+
+      x = self.token_embedding(text)
+      x = x + self.positional_embedding[:seq_len]
+      x = self.transformer(x, attn_mask=self.attn_mask)
+      x = self.ln_final(x)
+
+      pooled = x[Tensor.arange(x.shape[0]), text.argmax(dim=-1)]
```
chenyuxyz, line 336: is the arange needed? compare to `x[:, text.argmax(dim=-1)]`

chenyuxyz: hmm maybe not, it just looks weird to me

chenyuxyz: `list(range(x.shape[0]))` works

extra/models/clip.py
```diff
@@ -0,0 +1,366 @@
...
+    cs = bs[:]
+    n = 0
+    for b in range(2**8):
+      if b not in bs:
+        bs.append(b)
+        cs.append(2**8+n)
+        n += 1
+    cs = [chr(n) for n in cs]
+    return dict(zip(bs, cs))
+  @lru_cache()
```
chenyuxyz, line 48: i think this is broken for python < 3.10

PR #5257, opened 2024-07-02
perf: specialize match function at UPat init [run_process_replay]

roelofvandijk:
⚠️ This is probably too complex for the 50% speed increase it offers on the `_match` function, but I wanted to share the idea. 

At `UPat` initialization, the `_match` function can be specialized based on which elements of the `UPat` are not None.
This cuts down on unnecessary checks and attribute access.

```
PROFILE=0 python -O test/external/external_benchmark_schedule.py
master
***** model forward in  15.68 ms
***** model schedule in   5.22 ms
***** model lower in 575.02 ms
n:  331816  tm: 210.41ms  tot: 371.24ms uops.py:111:_match                                 <-  66% uops.py:141:rewrite

branch
***** model forward in  15.61 ms
***** model schedule in   5.02 ms
***** model lower in 525.86 ms
n:  331816  tm:  72.74ms  tot: 254.60ms uops.py:134:_match                                 <-  57% uops.py:122:src_unmatch
```

Summary of changes:
  * reduce `_match` to a tuple of `UPat`-specific match functions that only includes the relevant functions. This way, all member lookups are also removed from the matching calls.
  * move permutations from the `src` match (each match call) to the `UPat` init (once)
  * early stopping for the src children match
  * `op`, `dtype` and `arg` matches are simplified to `in tuple`

I tested real functions instead of lambda functions but that does not seem to matter.

Example:

`UPat(UOps.CONST, name="mval")` has `unmatch_funcs=(lambda uop,_: uop.op not in(UOps.CONST,), lambda uop,store: store.setdefault("mval", uop) is not uop)`

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -107,33 +107,34 @@ class UPat:
 
   def __post_init__(self):
     self.f: tuple[Callable[[Any, Any], Any], ...] = ()
-    if (o:=self.op) is not None: self.f += (lambda uop,_: uop.op not in o,) if isinstance(o,set) else (lambda uop,_: o != uop.op,)
-    if (a:=self.arg) is not None: self.f += (lambda uop,_: uop.arg not in a,) if isinstance(a,set) else (lambda uop,_: a != uop.arg,)
+    op = () if self.op is None else tuple(self.op) if isinstance(self.op, set) else (self.op,)
+    arg = () if self.arg is None else tuple(self.arg) if isinstance(self.arg, set) else (self.arg,)
+    dtype = () if self.dtype is None else tuple(self.dtype)if isinstance(self.dtype, set) else (self.dtype,)
+    if op: self.f += (lambda uop,_: uop.op not in op,)
+    if arg: self.f += (lambda uop,_: uop.arg not in arg,)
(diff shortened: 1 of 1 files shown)


DKormann:
why is it faster? because the lambdas or because you have the different loop for src_unmatch

roelofvandijk:
Both

DKormann:
oh i see the any( ... ) goes through all children that makes sense

DKormann:
this seems like you could get it passed easy when split into independent prs no? also i dont see a reason why this cant be done in almost no additional lines

DKormann:
well than try to pr at least the idea with lambda functions

geohot:
How are you getting 575 ms baseline? On an M3 Pro

```
diane@blak:~/tinygrad$ PROFILE=0 python -O test/external/external_benchmark_schedule.py
***** model forward in  18.74 ms
***** model schedule in   6.11 ms
***** model lower in 872.02 ms
```

I'm okay with the tuple refactor, I just think it should be like 3 lines and not make the syntax worse.

roelofvandijkO:
Python 3.12

roelofvandijk:
Python 3.12, M3 Pro

roelofvandijk:
Closing this, might pull out only the all-tuple init

PR #5263, opened 2024-07-02
use None in UOp.cmp_tuple for UOps.ALU [run_process_replay]

chenyuxyz:
if `self.op is UOps.ALU`, `(type(self.op), self.op.value)` is always the same. is there a reason to construct it this way?


geohot:
Only issue I see is now there no order between an ADD and MUL on the same two inputs

chenyuxyz:
fix in #5280

PR #5267, opened 2024-07-02
add return type for HCQCompatAllocator _alloc

vpachkov:
While working on qcom allocator, spent some time to dive into the copyin, copyout, transfer and offset of HCQCompatAllocator to find out that they require allocated memory object to have "va_addr" and other stuff
So this change make allocator abstraction more transparent

_It adds 3 new lines of code btw, so if it is considered to be superfluous, I'm ok :)_


geohot:
Types are always worth the lines! @nimlgen can you review?

nimlgen:
Thanks!

PR #5273, opened 2024-07-03
add timing to SDXL

chenyuxyz:


chenyuxyz:
3370ms for M2 Max, 840ms for green, 1080ms for red

PR #5277, opened 2024-07-03
nv allocate local memory dynamically

nimlgen:


nimlgen:
CUDA uses `0x2c1a000` there, but changing it doesn't give more memory. It's a virtual address and I will take a look what kernel doing with it (I am not sure if it's gpu physical memory).

PR #4203, opened 2024-04-17
POC new memory schedule algo

nimlgen:
new:
```
STEPS=700 BS=512 GPUS=1 TARGET_EVAL_ACC_PCT=93.5 DEBUG=0 python3 examples/hlb_cifar10.py                 
memory reduced from 201.00 MB to 198.60 MB
shuffling training dataset in 1101.67 ms (epoch=0)
memory reduced from 1974.03 MB to 742.41 MB
memory reduced from 2187.59 MB to 1946.16 MB
  0 3424.75 ms run, 3424.23 ms python,    0.51 ms METAL, 1198.37 loss, 0.000021 LR, 0.84 GB used,    196.68 GFLOPS,    673.57 GOPS
memory reduced from 1956.68 MB to 742.41 MB
memory reduced from 2187.59 MB to 1946.16 MB
JIT memory reduced from 2688.57 MB to 2185.25 MB
  1  707.03 ms run,  649.41 ms python,   57.62 ms METAL, 1197.87 loss, 0.000043 LR, 3.81 GB used,    950.39 GFLOPS,    671.95 GOPS
  2  450.06 ms run,    2.87 ms python,  447.19 ms METAL, 1183.85 loss, 0.000064 LR, 3.81 GB used,   1493.02 GFLOPS,    671.95 GOPS
  3  444.86 ms run,    2.09 ms python,  442.77 ms METAL, 1168.12 loss, 0.000085 LR, 3.81 GB used,   1510.49 GFLOPS,    671.95 GOPS
```

old:
```
 NEW_MEMPLANNER=0 STEPS=700 BS=512 GPUS=1 TARGET_EVAL_ACC_PCT=93.5 DEBUG=0 python3 examples/hlb_cifar10.py
shuffling training dataset in 649.45 ms (epoch=0)
memory reduced from 1974.03 MB to 1024.01 MB
memory reduced from 2187.59 MB to 2185.50 MB
  0 2578.58 ms run, 2578.06 ms python,    0.52 ms METAL, 1198.26 loss, 0.000021 LR, 0.84 GB used,    261.22 GFLOPS,    673.57 GOPS
memory reduced from 1956.68 MB to 1006.66 MB
memory reduced from 2187.59 MB to 2185.50 MB
JIT memory reduced from 3192.16 MB to 2919.52 MB
[...]

PR #5278, opened 2024-07-03
new memory scheduler

nimlgen:
Redo of #4203


tinygrad/engine/jit.py
```diff
@@ -115,7 +115,7 @@ def add_buffer(self, b:Buffer) -> Buffer:
     if found:=self.buffer_replace.get(b, None): return found
     if b.is_allocated() or b.lb_refcount > 0: return b
     if b._base is not None:
-      self.buffer_replace[b] = ret = Buffer(b.device, b.size, b.dtype, base=self.buffer_replace.get(b._base, b._base), offset=b.offset)
+      self.buffer_replace[b] = ret = Buffer(b.device, b.size, b.dtype, base=self.add_buffer(b._base), offset=b.offset)
```
nimlgen, line 118: With new memory scheduler caught this bug, since a subbuffer can appear earlier than it's base buffer (so this makes the subbuf to point to the wrong old buffer).

PR #5279, opened 2024-07-03
uops work from lowerer [run_process_replay] [no_assert]

geohot:


geohot:
Oh, actually fixes a bug!

DKormann:
whats the use of divides? shouldn't sums of const be evaluated by rewrite?

PR #5280, opened 2024-07-03
fix UOp.cmp_tuple for ALU

chenyuxyz:
for ALU, use self.arg instead of self.op to compare

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -777,8 +777,9 @@ class TestLinearizer(unittest.TestCase):
     # check that the float4 cast collapses for all stores
     for store in local_stores+global_stores:
       assert store.src[2].dtype == dtypes.float.vec(2) and store.src[2].op is not UOps.CAST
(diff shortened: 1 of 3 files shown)


chenyuxyz:
`[run_process_replay] [no_assert]` timed out on metal test due to diffs

test/test_linearizer.py
```diff
@@ -777,8 +777,9 @@ def get_recursive(uop): return set.union(set(uop.src), [uop], *[get_recursive(v)
     # check that the float4 cast collapses for all stores
     for store in local_stores+global_stores:
       assert store.src[2].dtype == dtypes.float.vec(2) and store.src[2].op is not UOps.CAST
-    # check the children's vins
-    assert barrier.src == tuple(local_stores)
```
chenyuxyz, line 781: @Qazalin fyi, this was added 6 months ago. the comparison worked because we were not comparing ALU.

PR #5282, opened 2024-07-03
remove "no-nans-fp-math"="true" for LLVM

chenyuxyz:
fixed isnan for llvm (still have issue with < nan)


chenyuxyz:
@hikettei fyi this might help some of the llvm nan issue you have seen

hikettei:
Thank you!!!

PR #5284, opened 2024-07-04
fix README example

SnakeOnex:
the main README example wasn't setting Tensor.training = True nor using the context manager and was giving the following error:

```
python tinygrad_example.py                                                
  File "/home/snakeonex/fun/tiny-shakespear/tinygrad_example.py", line 18, in <module>                    
    optim.step()                                                                                          
  File "/home/snakeonex/tinygrad/tinygrad/nn/optim.py", line 34, in step                                  
    Tensor.realize(*self.schedule_step())                                                                 
                    ^^^^^^^^^^^^^^^^^^^^                                                                  
  File "/home/snakeonex/tinygrad/tinygrad/nn/optim.py", line 39, in schedule_step                         
    assert Tensor.training, (         
           ^^^^^^^^^^^^^^^                                                                                
AssertionError: Tensor.training=False, Tensor.training must be enabled to use the optimizer.              
                - help: Consider setting Tensor.training=True before calling Optimizer.step().
```

- [x] fix README
- [x] add test
- [x] changed \```py -> \```python (for consistency docs/*.md all use \```python while README uses \```py

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -135,6 +135,8 @@ jobs:
         python docs/abstractions2.py
     - name: Test Quickstart
       run: awk '/```python/{flag=1;next}/```/{flag=0}flag' docs/quickstart.md > quickstart.py &&  PYTHONPATH=. python quickstart.py
+    - name: Test README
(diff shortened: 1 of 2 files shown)


chenyuxyz:
thanks!

PR #5288, opened 2024-07-04
rewrite default arg

DKormann:
slight simplification and only run once


Qazalin (commented):
hmm does this make ptx faster?

DKormann:
i dont know. Also its not passing ptx and i have no idea why it shouldn't really make a difference

Qazalin:
it's applying a bad rewrite rule?

DKormann:
i have no why its failing it should apply any different rules at all

PR #5291, opened 2024-07-05
sort src of commutative uop during rewrite

DKormann:
this will sort src of UOp when they are commutative. very helpful for symbolic comparison and should lead to better reusability during rewrite

### DIFF
--- a/tinygrad/runtime/ops_llvm.py
+++ b/tinygrad/runtime/ops_llvm.py
@@ -12,6 +12,7 @@ class LLVMCompiler(Compiler):
     super().__init__("compile_llvm")
   def compile(self, src:str) -> bytes:
     mod = llvm.parse_assembly(src)
+    # print(str(mod))
     mod.verify()
     self.device.optimizer.run(mod)
     if DEBUG >= 5: print(self.device.target_machine.emit_assembly(mod))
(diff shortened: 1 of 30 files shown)


DKormann:
this leads to a bug in llvm

DKormann:
solves #5304

tinygrad/runtime/ops_llvm.py
```diff
@@ -12,6 +12,7 @@ def __init__(self, device:LLVMDevice):
     super().__init__("compile_llvm")
   def compile(self, src:str) -> bytes:
     mod = llvm.parse_assembly(src)
+    # print(str(mod))
```
chenyuxyz, line 15: have you read the pr diff?

chenyuxyz (commented):
can you separate the fix (with a regression test) from the main change? it's hard to review a pr with multiple moving parts

tinygrad/runtime/ops_llvm.py
DKormann: sorry

DKormann:
> can you separate the fix (with a regression test) from the main change? it's hard to review a pr with multiple moving parts

yes will do

PR #5294, opened 2024-07-05
hotfix: fix metal with vars

nimlgen:


chenyuxyz:
is this why metal graph crashed sometimes?

nimlgen:
No, that's the recent patch. Forgot that we don't test graphs in our ci.

PR #5296, opened 2024-07-05
LoadOps.VIEW in the scheduler spec

Qazalin:
lazy.py has `allow_buffer_view` for cast and contiguous because it doesn't have enough information to select fusion vs a VIEW. The scheduler can decide subbuffer vs fusion instead.

fyi @hikettei


hikettei:
Thank you for letting me know! So cycle graph dependency issues are resolved ☺️

PR #5298, opened 2024-07-05
enable sin on METAL in test_dtype_alu

chenyuxyz:


Qazalin:
This is still flaky, have seen 2 distinct failures so far for this, eg:
https://github.com/tinygrad/tinygrad/actions/runs/9818240159/job/27110533344#step:8:278

PR #5307, opened 2024-07-06
use precise::sin in metal

chenyuxyz:


chenyuxyz:
@Qazalin fyi i saw this one was causing ci failure

PR #5315, opened 2024-07-07
refactor: expr_view from ShapeTracker to View

roelofvandijk:
In the spirit of https://github.com/tinygrad/tinygrad/issues/3662 (still relevant?), this should be a method on a view.

Can be cached with a lru_cache like the other view methods, miniscule speed improvement (like 1% on llama). `idxs` needs to become a tuple for that. Did not do that here to keep the refactor clean.


geohot:
Cool, agree it's cleaner.

PR #5322, opened 2024-07-08
Make vectorization of CONST explicit

### LINKED ISSUE #5313: Use UOps.VECTORIZE for consts
const renderers have an extra call for rendering a vectorized const. This can simplify to:
```
  # returns a str expression of the const with the given type
  def render_const(self, x:ConstType, dtype:DType) -> str:
    if math.isnan(x): return self.render_cast("NAN", dtype)
    elif math.isinf(x): return self.render_cast(("-" if x < 0 else "") + "INFINITY", dtype)
    elif dtype == dtypes.bool: return "1" if x else "0"
    elif dtype == dtypes.float: return f"{x}f"
    return str(x)
```

Current UOp:
`UOp(UOps.CONST, dtypes.float.vec(4), (), 0.0)`
Desired:
`UOp(UOps.VECTORIZE, dtypes.float.vec(4), (UOp(UOps.CONST, dtypes.float, (), 0.0), ))`

Should also fix DEFINE_ACC's extra type_verify assert.

greg-niemeyer:
Resolves #5313


greg-niemeyer:
@Qazalin regarding

> Should also fix DEFINE_ACC's extra type_verify assert.

I was looking at [you convo with geohot regarding this](https://discord.com/channels/1068976834382925865/1068982781490757652/1254860202566488165 )but I wasn't able to come to a firm conclusion whether you were saying that fixing #5313 requires an update to the assert, or will be implicitly fixed by the requested change. If the assert needs updating, please let me know.

greg-niemeyer:
I'm diff'ing process replay runs + ASSERT_PROCESS_REPLAY=0 with and without commit 61f9453 and I'm seeing different kernels getting generated. So looks like I've got more work to do.

Qazalin:
directionally correct.

greg-niemeyer:
Note: I added the GEP-VECTORIZE-CONST rule because I was getting some strange UOps sequences without it after updating UOps.const(). Attaching GRAPHUOPS=1 images with (gg.uops.svg) and without the rule (bg.uops.svg) -- note the extra PHI-GEP-VECTORIZE-CONST on the bottom of bg.uops.svg.

Adding `assert dtype.count == 1` to render_const causes a bunch of tests to fail, so I'll investigate this and see where that takes me. I'm guessing the reason I'm failing the assert is connected to the CI failures I'm seeing.

gg uops
bg uops

davidgonmar:
@greg-niemeyer when I was playing around with the code, when rendering DEFINE_ACC, it calls render_const with vectorized dtypes sometimes. That might be part of the assert fail.

greg-niemeyer:
> @greg-niemeyer when I was playing around with the code, when rendering DEFINE_ACC, it calls render_const with vectorized dtypes sometimes. That might be part of the assert fail.

right, I'm digging into the DEFINE_ACC stuff as we speak

greg-niemeyer:
Had to add [the render_cast for !(int, bool, float)](https://github.com/tinygrad/tinygrad/pull/5322/commits/499f7e85ab764e90dd7a000ce5ed0b7d14393c52) back -- there are longs, shorts, etc. that need their long-form cast rendered.

greg-niemeyer:
I'm not super happy with [DEFINE_ACC implicitly vectorizing CONST in cstyle](https://github.com/tinygrad/tinygrad/pull/5322/files#diff-9694d15708a36a57a7e1485db60b1b4b1d2fc1e7278bc87127a73efd0bfa7dc0R171-R173). Should I update DEFINE_ACC in UOps such that its src is VECTORIZE or CONST, consistent with its dtype?

Qazalin:
> Had to add [the render_cast for !(int, bool, float)](https://github.com/tinygrad/tinygrad/pull/5322/commits/499f7e85ab764e90dd7a000ce5ed0b7d14393c52) back -- there are longs, shorts, etc. that need their long-form cast rendered.

Yeah I see why you need it.

> I'm not super happy with [DEFINE_ACC implicitly vectorizing CONST in cstyle](https://github.com/tinygrad/tinygrad/pull/5322/files#diff-9694d15708a36a57a7e1485db60b1b4b1d2fc1e7278bc87127a73efd0bfa7dc0R171-R173). Should I update DEFINE_ACC in UOps such that its src is VECTORIZE or CONST, consistent with its dtype?

Correct, DEFINE_ACC should vectorize the const in src[0].

tinygrad/codegen/uops.py
```diff
@@ -66,6 +66,8 @@ def min(self, x): return -UOp.alu(BinaryOps.MAX, -self, -x)
   @functools.lru_cache(maxsize=None)
   def const(dtype:Optional[DType], b:ConstType|Variable):
     if isinstance(b, Variable): return UOp(UOps.DEFINE_VAR, dtype, (), b)
+    if dtype is not None and dtype != dtype.scalar():
+      return UOp(UOps.VECTORIZE, dtype, src=tuple(UOp(UOps.CONST, dtype.scalar(), arg=dtypes.as_const(b, dtype)) for i in range(dtype.count)))
```
Qazalin, line 70: hmm i think it should call as_const with dtype.scalar()

tinygrad/renderer/cstyle.py
```diff
@@ -170,7 +168,9 @@ def ssa(prefix:str, u:Optional[UOp]=None):
           bufs.append((nm:=f"data{args[0]}", (dtype,args[1])))
           r[u] = nm
         elif uop is UOps.WMMA: kk(f"{self.render_dtype(dtype)} {ssa('wmma',u)} = __{args[0]}({r[src[0]]}, {r[src[1]]}, {r[src[2]]});")
-        elif uop is UOps.DEFINE_ACC: kk(f"{self.render_dtype(dtype)} {ssa('acc',u)} = {self.render_const(src[0].arg, dtype)};")
+        elif uop is UOps.DEFINE_ACC:
+          val = r[src[0]] if dtype.count > 1 else self.render_const(src[0].arg, dtype)
+          kk(f"{self.render_dtype(dtype)} {ssa('acc',u)} = {val};")
```
Qazalin, line 173: actually this can a simple one-liner now.

tinygrad/runtime/ops_python.py
```diff
@@ -89,8 +89,7 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
             ul[i] = [x[2-arg[0]] for x in warp]
         elif uop is UOps.CONST:
           ul[i] = [[arg] * warp_size for _ in range(dtype.count)] if dtype.count > 1 else [arg] * warp_size
-        elif uop is UOps.DEFINE_ACC:
-          ul[i] = [[inp[0][0]] * warp_size for _ in range(dtype.count)] if dtype.count > 1 else [inp[0][0]] * warp_size
+        elif uop is UOps.DEFINE_ACC: ul[i] = inp[0] if dtype.count > 1 else [inp[0][0]] * warp_size
```
Qazalin, line 92: are you sure this is correct?
uops tests failing:
https://github.com/tinygrad/tinygrad/actions/runs/9868214822/job/27282699306?pr=5322

`IMAGE=2 PYTHON=1 python3 test/test_ops.py TestOps.test_gemm`

Qazalin:
should also pass run_process_replay.
Seeing extra consts: https://github.com/tinygrad/tinygrad/actions/runs/9878626439/job/27282996159#step:24:52

tinygrad/runtime/ops_python.py
greg-niemeyer: I'm not sure it's correct, but I think it is. I'm investigating the failure, but I haven't found many clues yet as to what's going on.

greg-niemeyer:
Also, I'm pretty sure the UPats I wrote are nonsensical and need to be redone. I will be busy the next couple of days but I'll work on this when I can.

tinygrad/runtime/ops_python.py
greg-niemeyer: still haven't cracked this yet, but i noticed that this test only fails with IMAGE=2. IMAGE=1 & no IMAGE envvar passes. printing and diffing the ul[i] values in this branch and master show that the values are the same. puzzling, but I'm getting closer.

greg-niemeyer:
> Also, I'm pretty sure the UPats I wrote are nonsensical and need to be redone. I will be busy the next couple of days but I'll work on this when I can.

i rewrote the UPats, but they're very verbose. i plan to rework them after i figure out this uop issue.

tinygrad/runtime/ops_python.py
greg-niemeyer: You were right about this 😄 I fixed it, but i hope i can make my fix less ugly in a cleanup patch. [0][0][0] doesn't seem that nice.

The new failures are unrelated to my latest changes. Bisecting through my git history, they started appearing after I rebased on the new lowerer. Investigating.

greg-niemeyer:
I'm able to reproduce the GPU/IMAGE and openpilot test failures locally on my M1 macbook with only these two commits:
0152a86c66ff7d7432d5c6fc8ecdf4063ff03155: remove test_const_vectorize_fold 
0ff3cb92c13e291b216d949c5a48015f213873a1: remove const folding UPat for VECTORIZE

As I understand it, these patches shouldn't result in test failures, only process_replay failures due to extra casts. Probing further, I found that these patches pass the same tests prior to migrating to the new lowerer, but fail afterward.
[This branch](https://github.com/greg-niemeyer/tinygrad/tree/pre-lowerer) was synced to the commit immediately preceding the lowerer change, while [this one](https://github.com/greg-niemeyer/tinygrad/tree/post-lowerer) was synced to the lowerer change itself prior to cherry-picking the aforementioned commits.

The test commands I'm using are:
- `GPU=1 IMAGE=2 python -m pytest -n=auto test/test_ops.py`
- `PYTHONPATH="." FLOAT16=0 DEBUGCL=1 GPU=1 IMAGE=2 python examples/openpilot/compile2.py`
- `JIT=2 METAL=1 python -m pytest -n=auto test/ --ignore=test/external --ignore=test/models --durations=20`
  - The METAL tests should pass regardless of whether my two patches are present.

`FORWARD_ONLY=1 GPU=1 IMAGE=2 python -m pytest test/test_ops.py::TestOps::test_simple_padding_conv2d -s` is what I use for most of my debugging, since it seems to be the simplest repro case I have.

Two data points I think are interesting:
- I was able to get the tests to pass again by updating the fix_image_idx() function using CONSTs instead of VECTORIZE in the LOAD. Demo branch [here](https://github.com/greg-niemeyer/tinygrad/tree/post-lowerer-fix-image-idx-hack). Perhaps it's obvious that this would "fix" the issue given the error signature, but the confirmation is nice.
- I was able to get the GPU/IMAGE tests by commenting out a UPat that folds CONSTs via exec_alu(). Demo branch [here](https://github.com/greg-niemeyer/tinygrad/tree/post-lowerer-with-ALU-CONST-hack).
  - This one was a bit of a surprise to me, though perhaps it should be taken with a grain of salt since it also caused a bunch of the METAL tests to fail.

I tried a bunch of other experiments printing and stepping through various parts of `cstyle`, and `uopgraph` with the debugger, but I'm having trouble narrowing the issue down furtherer. Any suggestions would be appreciated.

greg-niemeyer:
I instrumented the PatternMatcher rules and identified [this UPat](https://github.com/tinygrad/tinygrad/pull/5322/files#diff-00bd44b667ec90ae1d3e984e699bc6b498c84ca1b1bd15a025437ded227457bfR500) as the source of the divergent behavior when using VECTORIZE instead of CONST in [the post-lowerer branch](https://github.com/greg-niemeyer/tinygrad/tree/post-lowerer). Changing the rule to use UOp.var() instead of UOp.cvar() results in VECTORIZE being matched properly and fixes the GPU/IMAGE and openpilot tests. The AMD test is still failing, so that's the next thing to work on.

tinygrad/codegen/uopgraph.py
```diff
@@ -477,7 +497,7 @@ def loop_collapse(loop_start, loop_end, compval, idx, mval, multconst, rng):
   (UOp.load(UOp.var("buf"), UOp.var("idx"), UOp.const(dtypes.bool, True), UOp.cvar("var")), lambda buf,idx,var: UOp.load(buf, idx, dtype=var.dtype)),
   (UOp.load(UOp.var("buf"), UOp.var("idx"), UOp.const(dtypes.bool, True), UOp.cvar("var"), UOp.var("barrier")),
    lambda buf,idx,var,barrier: UOp.load(buf, idx, barrier, dtype=var.dtype)),
-  (UOp.load(UOp.var(), UOp.var(), UOp.const(dtypes.bool, False), UOp.cvar("var")), lambda var: var),
+  (UOp.load(UOp.var(), UOp.var(), UOp.const(dtypes.bool, False), UOp.var("var")), lambda var: var),
   (UOp.load(UOp.var(), UOp.var(), UOp.const(dtypes.bool, False), UOp.cvar("var"), UOp.var()), lambda var: var),
```
Qazalin, line 501: wanna also change this to var?

tinygrad/codegen/uopgraph.py
```diff
@@ -362,7 +362,9 @@ def loop_collapse(loop_start, loop_end, compval, idx, mval, multconst, rng):
   # sum collapse to mul (with possible GEP)
   (UPat(UOps.PHI, src=(UPat(UOps.DEFINE_ACC, name="phi_input", src=[UPat(UOps.CONST), UPat(UOps.RANGE, name="loop")]),
                        UPat(UOps.ALU, BinaryOps.ADD, src=(UPat(name="val1"), UPat(name="val2"))))), sum_collapse),
-  (UPat(UOps.PHI, src=(UPat(UOps.GEP, name="phi_input", src=(UPat(UOps.DEFINE_ACC, src=[UPat(UOps.CONST), UPat(UOps.RANGE, name="loop")]),)),
+  (UPat(UOps.PHI, src=(UPat(UOps.DEFINE_ACC, name="phi_input", src=[UPat(UOps.VECTORIZE), UPat(UOps.RANGE, name="loop")]),
```
Qazalin, line 365: I think UPat takes a Set[UOp] as well, do you need to copy the entire UPat?

tinygrad/codegen/uopgraph.py
```diff
@@ -385,15 +387,33 @@ def loop_collapse(loop_start, loop_end, compval, idx, mval, multconst, rng):
...
   (UOp(UOps.PHI, src=(UOp(UOps.DEFINE_ACC).name("acc"), UOp.var("acc"))), lambda acc: UOp.cast(acc.src[0], acc.dtype)),
   (UOp(UOps.PHI, src=(UOp(UOps.DEFINE_ACC, src=(UOp.cvar(),)), UOp.var("x"))), lambda x: x),
+  (UPat(UOps.PHI, src=(UPat(UOps.DEFINE_ACC, src=(UPat(UOps.VECTORIZE, src=[UPat(UOps.CONST) for _ in range(2)]))), UPat(name="x"))), lambda x: x),
+  (UPat(UOps.PHI, src=(UPat(UOps.DEFINE_ACC, src=(UPat(UOps.VECTORIZE, src=[UPat(UOps.CONST) for _ in range(4)]))), UPat(name="x"))), lambda x: x),
+  (UPat(UOps.PHI, src=(UPat(UOps.DEFINE_ACC, src=(UPat(UOps.VECTORIZE, src=[UPat(UOps.CONST) for _ in range(8)]))), UPat(name="x"))), lambda x: x),
   (UOp(UOps.PHI, src=(UOp.cvar(), UOp.var("x"))), lambda x: x),
+  (UPat(UOps.PHI, src=(UPat(UOps.VECTORIZE, src=[UPat(UOps.CONST) for _ in range(2)]), UPat(name="x"))), lambda x: x),
+  (UPat(UOps.PHI, src=(UPat(UOps.VECTORIZE, src=[UPat(UOps.CONST) for _ in range(4)]), UPat(name="x"))), lambda x: x),
+  (UPat(UOps.PHI, src=(UPat(UOps.VECTORIZE, src=[UPat(UOps.CONST) for _ in range(8)]), UPat(name="x"))), lambda x: x),
```
Qazalin, line 408: I'm supportive of a well-tested PR for generic vectorize fold, we repeat this here too https://github.com/tinygrad/tinygrad/blob/0b3a34e3b1a7b8f73432d3ee4d555911c5cb80ab/tinygrad/codegen/uopgraph.py#L339-L341

tinygrad/renderer/cstyle.py
```diff
@@ -171,7 +169,8 @@ def ssa(prefix:str, u:Optional[UOp]=None):
           bufs.append((nm:=f"data{args[0]}", (dtype,args[1])))
           r[u] = nm
         elif uop is UOps.WMMA: kk(f"{self.render_dtype(dtype)} {ssa('wmma',u)} = __{args[0]}({r[src[0]]}, {r[src[1]]}, {r[src[2]]});")
-        elif uop is UOps.DEFINE_ACC: kk(f"{self.render_dtype(dtype)} {ssa('acc',u)} = {self.render_const(src[0].arg, dtype)};")
+        elif uop is UOps.DEFINE_ACC:
+          kk(f"{self.render_dtype(dtype)} {ssa('acc',u)} = {r[src[0]] if dtype.count > 1 else self.render_const(src[0].arg, dtype)};")
```
Qazalin, line 173: hmm, r[src[0]] should work for both?

Qazalin:
can you use vectorize for WMMA too?

tinygrad/codegen/uopgraph.py
greg-niemeyer: Thanks, I was thinking about this but unsure if I should since I'm unfamiliar with these UPats. Made the suggested change. If I may ask, why are you referring to the uop as `invalid`? I don't see any references to `invalid` in the code.

greg-niemeyer:
Didn't have too much time to work on this today, but I managed to fix the AMD test failure and address some of the review feedback. Tomorrow I'll have more time, so I should be able to make decent progress on cleaning up the UPats, writing tests, and WMMA.

tinygrad/codegen/uopgraph.py
Qazalin: "invalid" refers to a CONST 0 for a padded LOAD:
https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/lowerer.py#L119-L124

before, it was CONST(0, dtypes.float.vec(4))
Now it's VECTORIZE(CONST(0), dtypes.float.vec(4)) so your UOp for "var" changed.

tinygrad/codegen/uopgraph.py
greg-niemeyer: By generic, did you mean creating a function that does this processing? I spent some time thinking about how I could do this with a function, but struggled to come up with a good design that covers all required cases succinctly. I chose to generate these rules with a list comprehension and unpack them inline instead. I feel that it's sufficiently concise, but more flexible.

greg-niemeyer: Added some tests for these rules, too.

greg-niemeyer:
> can you use vectorize for WMMA too?

I dumped the UOps.WMMAs produced when running test_linearizer.py and they look like they're getting vectorized already -- they all look like:
UOps.WMMA           : dtypes._half2             [<UOps.VECTORIZE: 16>, <UOps.VECTORIZE: 16>, <UOps.VECTORIZE: 16>] ('WMMA_8_8_8_half_half', (8, 8, 8), dtypes.half, dtypes.half, (2, 2, 2), 'METAL', (9, 9, 9), (8,))
UOps.WMMA           : dtypes._float2            [<UOps.VECTORIZE: 16>, <UOps.VECTORIZE: 16>, <UOps.VECTORIZE: 16>] ('WMMA_8_8_8_float_float', (8, 8, 8), dtypes.float, dtypes.float, (2, 2, 2), 'METAL', (9, 9, 9), (8,))
UOps.WMMA           : dtypes._float2            [<UOps.VECTORIZE: 16>, <UOps.VECTORIZE: 16>, <UOps.VECTORIZE: 16>] ('WMMA_8_8_8_half_float', (8, 8, 8), dtypes.half, dtypes.float, (2, 2, 2), 'METAL', (10, 10, 10), (9,))

Did you have something else in mind?

Also, I'll look into what I can do to fix process_replay

greg-niemeyer:
> Also, I'll look into what I can do to fix process_replay

I was able to remove the extra cast, but undoing the diff to the way parentheses and some dtype casts (e.g. half) are applied in edge cases proved difficult. I tried:

- Reverting the conditional logic and `val` assignment in `render_const`
- adding `half` [here](https://github.com/tinygrad/tinygrad/pull/5322/files#diff-9694d15708a36a57a7e1485db60b1b4b1d2fc1e7278bc87127a73efd0bfa7dc0R53)
- adding `isinf()` and `isnan()` checks [here](https://github.com/tinygrad/tinygrad/pull/5322/files#diff-9694d15708a36a57a7e1485db60b1b4b1d2fc1e7278bc87127a73efd0bfa7dc0R173)

But these changes caused diffs in other places.

Qazalin:
yea that extra render_cast in cstyle is fine for now, it's not your fault. WMMA looks ok too.

tinygrad/codegen/uopgraph.py
```diff
@@ -336,9 +336,8 @@ def loop_collapse(loop_start, loop_end, compval, idx, mval, multconst, rng):
    lambda x: UOp(x.op, dtypes.int32, x.src, x.arg)),
   # VECTORIZE/GEP
   (UOp(UOps.GEP, src=(UOp(UOps.VECTORIZE).name("cast"),)).name("gep"), lambda gep, cast: cast.src[gep.arg]),
-  (UOp(UOps.VECTORIZE, dtypes.float.vec(2), tuple(UOp(UOps.GEP, dtypes.float, src=(UOp.var('x'),), arg=i) for i in range(2))), lambda x: x),
-  (UOp(UOps.VECTORIZE, dtypes.float.vec(4), tuple(UOp(UOps.GEP, dtypes.float, src=(UOp.var('x'),), arg=i) for i in range(4))), lambda x: x),
-  (UOp(UOps.VECTORIZE, dtypes.float.vec(8), tuple(UOp(UOps.GEP, dtypes.float, src=(UOp.var('x'),), arg=i) for i in range(8))), lambda x: x),
+  *[(UOp(UOps.VECTORIZE, dtypes.float.vec(i), tuple(UOp(UOps.GEP, dtypes.float, src=(UOp.var('x'),), arg=j)
+      for j in range(i))), lambda x: x) for i in [2, 4, 8]],
```
Qazalin, line 340: wanna put this loop refactor in a separate PR?

greg-niemeyer: Done. See #5507

Qazalin:
I'd keep the cstyle render_cast for now and add [run_process_replay] back. We'll remove that in another diff.
Let's focus this diff on just the CONST -> VECTORIZE change.

greg-niemeyer:
> I'd keep the cstyle render_cast for now and add [run_process_replay] back. We'll remove that in another diff. Let's focus this diff on just the CONST -> VECTORIZE change.

I made the requested changes and added [run_process_replay] to the PR title, but even with just the minimal update to cstyle required for CONST -> VECTORIZE to work, we have that extra cast in process_replay. 

Also, I'm getting on a plane in a couple hours and will be traveling the next two weeks, so my availability will be limited.

greg-niemeyer:
I was able to fix the process_replay for METAL tests, but its ugly. AFAICT, ugliness is tough to get rid of due to the differences in VECTORIZE vs implicitly vectorized CONST renders cascading into higher level uops with their own rendering idiosyncracies.

Qazalin:
We'll cleanup the rendering diff, because this diff changes UPats too it's better to assert process replay to avoid regressions.

greg-niemeyer:
Update -- been short on time due to travel, but making progress. I've got the issue repro'ing locally with gpuocelot in a docker. The particular assertion we're seeing in CI is related to the diff between STORE rendering VECTORIZE vs implicitly vectorized CONST. I tried [inlining the const render in the STORE](https://github.com/greg-niemeyer/tinygrad/commit/c42a73cb5e0f52579e1ea48b68e41fcb6385a971) but it results in another diff, so I need to think a little more on how to maintain parity between the renders.

tinygrad/renderer/assembly.py
```diff
@@ -165,16 +165,17 @@ def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
...
           assert args[1][0] != "i", "idx not supported"
           kk(f"mov.u32 %{args[1]}, {(self.gid if args[1][0] == 'g' else self.lid)[args[0]]};")
           r[u] = "%" + args[1]
           kernel = [f".reg .u32 %{args[1]};"] + kernel
         elif uop is UOps.CONST:
-          if dtype.count > 1: r[u] = [const(args, dtype.scalar(), mov=True) for _ in range(dtype.count)]
-          else: r[u] = const(args, dtype, mov=True)
+          assert dtype.count == 1
+          n = next((x.dtype.count for x in uops.uops if x.op == UOps.VECTORIZE and all(s is u for s in x.src)), None) # type: ignore
+          r[u] = [const(args, dtype, mov=True) for _ in range(n)] if n else const(args, dtype, mov=True)
```
greg-niemeyer, line 178: This probably deserves some explanation -- I found that the process_replay for CONST is sensitive to the topo sort order of the uopgraph because the CONST render kk's a mov instruction. When VECTORIZE'd CONST replaces CONST, the sorting changes in a way that makes it difficult to maintain process_replay parity. There were two main options I considered:
1. remove the type_verify assert for CONST and add a pat in this renderer that rewrites VECTORIZE'd CONST back to the original implicitly vectorized CONST, then push another PR that adds the assert back, changes the renderer, and introduces a process_replay diff.
2. in the CONST render, search the graph for the child VECTORIZE'd CONST and render it.

I chose the second because the first option seems against the spirit of the PR, and I figure the second can be reverted in a subsequent PR that introduces a process_replay diff.

greg-niemeyer:
I'm working on a WMMA-related process_replay regression introduced after merging upstream master in my dev branch. Should have it figured out tomorrow.

tinygrad/renderer/assembly.py
greg-niemeyer: This actually doesn't work because there are now situations where CONST is shared by VECTORIZE and another UOp that expects a scalar CONST, so this generates bogus instructions, hence the CI failures. I'm now thinking it may not be possible to maintain process_replay parity in the PTX backend without the WAR I mentioned above, or pushing a different PR that makes the CONST renders insensitive to the sort order.

greg-niemeyer:
@Qazalin in light of geohot's comment in #5729, I reverted my process replay hacks in the PTX renderer and kept the change needed to functionally pass the tests. I can provide a detailed write-up of all the classes of PTX process_replay diffs if that's what it takes to get this merged without [run_process_replay]. There are 3 distinct classes of diff and they can all be coherently explained. Additionally, I can revert my hacks in cstyle and do the same for all the process replay diffs there if desired.

I could also just shelve this if it's better to just keep implicitly vectorized CONST for now and not worth the effort to review the diffs.

tinygrad/renderer/cstyle.py
Qazalin, line 1: I think these two are hacks too, or do we need them for correctness?

greg-niemeyer: Yes, these are hacks. We only need a small change in cstyle for correctness, which I just pushed. If you check the CI failures, they're all due to process_replay.

Qazalin: are these from render_cast?
https://github.com/tinygrad/tinygrad/actions/runs/10116690164/job/27980057908?pr=5322#step:24:389

tinygrad/codegen/uops.py
```diff
@@ -99,21 +102,23 @@ def divides(self, v):
...
+  def vmax(self) -> UOp: return x if (x:=self._min_max[1]) is not None else self.const(dtypes.max(cast(DType, self.dtype))).vmax
   @functools.cached_property
   def _min_max(self) -> Tuple[Optional[UOp], Optional[UOp]]:
     # NOTE: returned UOp is assumed to be CONST
     if self.op in (UOps.DEFINE_VAR, UOps.RANGE): return self.src[0], self.src[1] if isinstance(self.src[1].arg, int) else None
     # TODO: UOps.SPECIAL is UOps.DEFINE_VAR
     if self.op is UOps.SPECIAL: return self.const(0), self.const(self.arg[1]-1) if isinstance(self.arg[1], int) else None
     if self.op is UOps.CONST: return self, self
+    if self.op is UOps.VECTORIZE and all(x.op == UOps.CONST for x in self.src):
+      return min(self.src, key=lambda x: x.arg), max(self.src, key=lambda x: x.arg)
```
Qazalin, line 116: hmm, I think this is extra complexity. you always have the same const across a vectorize.
Can it just be self.const(dtypes.max(self.dtype.scalar())) in the above?

Qazalin:
renderers look much cleaner now! I don't think we need the extra logic for vmin and vmax though

tinygrad/renderer/cstyle.py
greg-niemeyer: They're from the CONST render, since the DEFINE_ACC is using r[src[0]] instead of calling render_const() directly.

tinygrad/codegen/uops.py
greg-niemeyer: > you always have the same const across a vectorize.
Not true. I've seen different value consts being vectorized in test_real_world.py. I was surprised by this and I recorded it in my notes while I was hacking at the PTX renderer (see the unhighlighted white text above the diff):
Pasted Graphic 3

I agree that it is extra complexity though -- I definitely need to cleanup the vmin & vmax logic. Will try to have a fix out shortly, but I need to understand the upstream behavior a little better. chenyu's generic max folding was a sore point for my rebasing haha

Qazalin: ohh right you can have a const folded valid mixed with invalid.

greg-niemeyer: I think I've got the vmin/vmax stuff reasonably clean, if the extra `sconst` def is acceptable. The `sconst` def's purpose is to save lines and satisfy mypy, otherwise we'd have casts everywhere.

I checked the process_replay and it's consistent with what we were seeing before the cleanup.

tinygrad/codegen/uopgraph.py
```diff
@@ -205,14 +207,18 @@ def index_collapse(idx,rng,buf,add,mul,ld,reduce_allow_any_len):
   (NOp.max(NOp.cvar('c'), -(NOp(UOps.RANGE, name="s")+NOp.cvar('c2'))), lambda c,s,c2: -(s+c2) if -(s.src[1].arg-1+c2.arg) >= c.arg else None),
   # const rules
   (NOp(UOps.GEP, src=(NOp.cvar("c"),), name="root"), lambda root, c: root.const(c.arg)),
+  *[(UPat(UOps.GEP, name="root", src=(UPat(UOps.VECTORIZE, name="v", src=[UPat(UOps.CONST) for _ in range(i)]))),
+     lambda root, v: root.const(v.src[root.arg].arg)) for i in [2, 4, 8]],
```
Qazalin, line 211: isn't v.src[root.arg] const already?

tinygrad/codegen/uopgraph.py
```diff
@@ -205,14 +207,18 @@ def index_collapse(idx,rng,buf,add,mul,ld,reduce_allow_any_len):
...
-  (UPat(UOps.VECTORIZE, name="root", src=UPat(UOps.CONST, name="c")), lambda root, c: root.const(c.arg)),
-  # a phi on a DEFINE_ACC without loops or a CONST is a noop. this is for correctness, not just speed
+  # a phi on a DEFINE_ACC without loops, CONST, or a vectorized CONST is a noop. this is for correctness, not just speed
   (NOp(UOps.PHI, src=(NOp(UOps.DEFINE_ACC, name="acc"), NOp.var("acc"))), lambda acc: UOp.cast(acc.src[0], acc.dtype)),
   (NOp(UOps.PHI, src=(NOp(UOps.DEFINE_ACC, src=(NOp.cvar(),)), NOp.var("x"))), lambda x: x),
   (NOp(UOps.PHI, src=(NOp.cvar(), NOp.var("x"))), lambda x: x),
+  *[(UPat(UOps.PHI, src=(UPat(UOps.VECTORIZE, src=[UPat(UOps.CONST) for _ in range(i)]), UPat(name="x"))), lambda x: x) for i in [2, 4, 8]],
   # a DEFINE_ACC without inputs is a const + GEP on a const is the const
-  (NOp(UOps.DEFINE_ACC, src=(NOp.cvar(),), name="root"), lambda root: UOp.cast(root.src[0], root.dtype)),
+  (NOp(UOps.DEFINE_ACC, src=(NOp.var(),), name="root"), lambda root: UOp.cast(root.src[0], root.dtype)),
```
Qazalin, line 219: don't need the cast.

Qazalin:
spent some time with this code, I like sconst, you can remove [run_process_replay] from the title, this is gonna simplify kernels.

test/test_uop_graph.py
```diff
@@ -171,6 +162,180 @@ def _test_vec(geps):
     xy2 = tuple(UOp(UOps.GEP, dtypes.float, (val2, ), i) for i in range(2))
     self.assertIs(_test_vec(xy1+xy2).op, UOps.VECTORIZE)
 
+  def test_gep_vec_const_fold(self):
```
Qazalin, line 165: can you write this with a loop? That'd be easier to read.

tinygrad/codegen/uops.py
```diff
@@ -69,11 +69,16 @@ def min(self, x): return -(-self).max(-x)
   def where(self, x, y): return self.alu(TernaryOps.WHERE, x, y)
   def recip(self): return self.alu(UnaryOps.RECIP)
   def const(self:Union[UOp, DType, None], b:ConstType|Variable): return UOp._const(self.dtype if isinstance(self, UOp) else self, b)
+  def sconst(self:Union[UOp, DType, None], b:ConstType|Variable):
+    return UOp._const(cast(DType, self.dtype).scalar() if isinstance(self, UOp) else self, b)
```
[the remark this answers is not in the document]
greg-niemeyer: added logic to address this and confirmed locally it behaves as expected

test/test_uop_graph.py
greg-niemeyer: Done here and with `test_phi_vec_const_fold` and `test_gep_acc_vec_fold` as well

tinygrad/codegen/uopgraph.py
greg-niemeyer: removed the cast

tinygrad/codegen/uopgraph.py
greg-niemeyer: yes. I simplified the logic. I saw this in two other unrelated pats as well. Will push a separate PR for that.

greg-niemeyer:
Woohoo checks passed! Should I mark this PR ready for review?

tinygrad/codegen/uopgraph.py
Qazalin: cool, those PRs need [run_process_replay] as well.

greg-niemeyer: will do. that change is actually dependent on this one, so I'll push and tag later.

Qazalin:
holding on merging this because AMD got a big performance regression. I think their HIP compiler is missing optimizations it would've done otherwise.

This is a good simplification for tinygrad's internals but we can't do it at the cost of almost 2 orders of magnitude regression in AMD=1 sdxl, even if it's a compiler weirdness.

Are you interested in seeing if we can work around this in a clean way in cstyle? I think we can give you tinybox access.

greg-niemeyer:
> Are you interested in seeing if we can work around this in a clean way in cstyle? I think we can give you tinybox access.
yeah let's do it

Qazalin:
fyi either @wozeparrot or me will give you the tinybox access by EOD.

greg-niemeyer:
> holding on merging this because AMD got a big performance regression. I think their HIP compiler is missing optimizations it would've done otherwise.

this perf regression was completely my fault. I wrote some pathological UPats that used lists instead of tuple for VECTORIZE.src. since we are using vec(8)'s on AMD, we were doing 8! (+40,000) comparisons when we should have been doing one, worst-case.

still seeing a minor perf regression. with my change, sdxl takes roughly 800ms longer on the first step, then is 300ms faster on the second step, so ~500ms slower in aggregate on AMD=1 sdxl.
this branch:
```
(tinygrad) greg-niemeyer@tiny10:~/tinygrad$ PYTHONPATH="." AMD=1 python3 examples/sdxl.py --seed 0 --noshow --timing --weights=./weights/sd_xl_base_1.0.safetensors --out=./rendered.png | tee sdxl_good_enuff.txt
ram used:  6.93 GB, sigmas                                            : 100%|█|
100%|███████████████████████████████████████████| 10/10 [00:26<00:00,  0.38it/s]
loaded weights in 17849.89 ms, 6.93 GB loaded at 0.39 GB/s
created batch
step in 13785.81 ms, using 5.30 GB
step in 3052.05 ms, using 7.59 GB
step in 1291.28 ms, using 7.59 GB
step in 1161.24 ms, using 7.59 GB
step in 1162.83 ms, using 7.59 GB
step in 1155.86 ms, using 7.59 GB
step in 1162.48 ms, using 7.59 GB
step in 1157.88 ms, using 7.59 GB
step in 1166.16 ms, using 7.59 GB
step in 1151.83 ms, using 7.59 GB
```

master:
```
(tinygrad) greg-niemeyer@tiny10:~/tinygrad$ PYTHONPATH="." AMD=1 python3 examples/sdxl.py --seed 0 --noshow --timing --weights=./weights/sd_xl_base_1.0.safetensors --out=./rendered.png
| tee sdxl_master.txt
ram used:  6.93 GB, sigmas                                            : 100%|█|
100%|███████████████████████████████████████████| 10/10 [00:25<00:00,  0.39it/s]
loaded weights in 17865.05 ms, 6.93 GB loaded at 0.39 GB/s
created batch
step in 13015.36 ms, using 5.30 GB
step in 3394.70 ms, using 7.59 GB
step in 1295.29 ms, using 7.59 GB
step in 1163.01 ms, using 7.59 GB
step in 1163.04 ms, using 7.59 GB
step in 1165.24 ms, using 7.59 GB
step in 1166.17 ms, using 7.59 GB
step in 1165.33 ms, using 7.59 GB
step in 1167.06 ms, using 7.59 GB
step in 1158.80 ms, using 7.59 GB
```

Qazalin:
yea sdxl is still too slow, https://github.com/tinygrad/tinygrad/actions/runs/10216851155/job/28269312677#step:9:1

Can you diff the kernels it's creating by running process replay locally? https://github.com/tinygrad/tinygrad/blob/master/test/external/process_replay/README.md#running-locally

I'd expect this to be a noop change for performance.

greg-niemeyer:
> Can you diff the kernels it's creating by running process replay locally? https://github.com/tinygrad/tinygrad/blob/master/test/external/process_replay/README.md#running-locally
> 
> I'd expect this to be a noop change for performance.

It's not an issue with the generated kernels. I applied my old [cstyle hacks for process_replay parity](https://github.com/tinygrad/tinygrad/pull/5322/files/3b8e47b9deeca3ce8c1b09ebda675c8def6cfd93#diff-9694d15708a36a57a7e1485db60b1b4b1d2fc1e7278bc87127a73efd0bfa7dc0R164-R182) and confirmed that the perf issue is still present, even with identical kernels.

I've been diffing sdxl runs instrumented with cProfile to see if there are any noticeable anomalies in the functions where sdxl is spending the most time, and it looks like the match is still adding significant overhead. I was able to remove two of the GEP-related rules with no discernible impact on local process_replay, and that shaved off some of it, but I still need to figure out the other sources. Will keep digging.

greg-niemeyer:
On the tip of my dev branch (after removing the unneeded GEP rules) I'm consitently seeing ~300-400ms overhead when compared to master at the same point in time. The overhead persists regardless of process_replay differences, and I conducted perf analysis on both this dev branch, and [one branched from this dev branch + my cstyle hacks for process_replay parity](https://github.com/greg-niemeyer/tinygrad/tree/vec-perf). The results are the same.

Since the kernels aren't the problem, I collected time spent in each function call with `cProfile`:
`PYTHONPATH="." AMD=1 python3 -m cProfile -o sdxl_dev.prof examples/sdxl.py --seed 0 --noshow --timing --weights=./weights/sd_xl_base_1.0.safetensors --out=./rendered.png`

and output the top-100 functions sorted by `cumtime` (see [the docs](https://docs.python.org/3/library/profile.html) for more details). I diffed the results from my dev branch vs master, and what I see is significantly more calls to `PatternMatcher` machinery that appear to simply be a function of the increase in UOpGraph edges due to use of more vec(8) UOps:

dev branch:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
996688/999    1.363    0.000   22.602    0.023 uopgraph.py:450(__inner_rewrite)
1394669/11943    0.395    0.000   22.389    0.002 uopgraph.py:452(<genexpr>)
   299555    1.899    0.000   19.471    0.000 uops.py:198(rewrite)
5491152/2380667    7.640    0.000   10.755    0.000 uops.py:171(_match)
```
vs

master:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
966380/999    1.592    0.000   22.226    0.022 uopgraph.py:449(__inner_rewrite)
1352103/11943    0.376    0.000   22.014    0.002 uopgraph.py:451(<genexpr>)
   291156    1.870    0.000   18.937    0.000 uops.py:194(rewrite)
5395541/2346966    7.456    0.000   10.168    0.000 uops.py:167(_match)
```

I think its worth pushing to `update_benchmark` again to confirm the overhead is not as extreme as what's seen in
>yea sdxl is still too slow, https://github.com/tinygrad/tinygrad/actions/runs/10216851155/job/28269312677#step:9:1

But I'm not sure I can do much to eliminate the remaining overhead in the context of this PR. I can keep digging and thinking about it though.

Qazalin:
> due to use of more vec(8) UOps
Are you sure? Which of these patterns are compounding to 18% slower sdxl? Can we make them faster?

greg-niemeyer:
> Are you sure? Which of these patterns are compounding to 18% slower sdxl? Can we make them faster?

I'm actively working on this and trying to answer your question definitively, but the going is slow. I am strongly suspecting WMMA as the problematic UOp, but the precise reason why is currently eluding me. I'll have a better write-up for you soon, hopefully.

One thing that concerns me -- I'm unable to reproduce a performance regression of the magnitude seen in the latest [update_benchmark run you pushed](https://github.com/tinygrad/tinygrad/actions/runs/10226381342/job/28296551849). If you look at the SDXL run, we see that first step running at `step in 23420.09 ms, using 5.30 GB`, which is [~2x the normal time](https://github.com/tinygrad/tinygrad/actions/runs/10217286448/job/28270706438).

Running manually on tiny10, i see step1 of this dev branch running around ~13400-13700ms, which is roughly 400ms or so slower than what I see master running at, which is somewhere around 12900-13200ms. That's like a 3% slower run as opposed to a 100% slower run. Is there any environment difference that I should be aware of between tiny10 and the machine CI is running on? Perhaps a different python version or something?

Qazalin:
Thanks, really looking forward to merging this.
I just pushed to update_benchmark again, if the issue still remains I'll try to repro on both tinyboxes and give you an update.
Our CI doesn't run on tiny10.

greg-niemeyer:
> I am strongly suspecting WMMA as the problematic UOp, but the precise reason why is currently eluding me. I'll have a better write-up for you soon, hopefully.

After adding a bunch of instrumentation to the pattern matcher, I see that on master when we have an implicitly vectorized CONST, WMMAs get created, then they hit this rule:
` (NOp.var("add") + NOp(UOps.WMMA, name="wmma"),
    lambda add, wmma: UOp(wmma.op, wmma.dtype, (wmma.src[0], wmma.src[1], wmma.src[2]+add), wmma.arg)),`
and then, in a lot of cases, the output of that rule gets rewritten by:
`(NOp.var('x') + 0, lambda x: x),`

When explicit VECTORIZE + CONST is sub'd for the implicitly vectorized CONST, the +0 rule doesn't get hit because the +0 creates a CONST NOp with dtype=None. We still get the correct UOp output in the end, but it goes through a longer, multi-step series of rewrites.

So basically, I'm missing VECTORIZE'd versions of the generic ALU const folding rules. The +0 one is the one hit most by sdxl, and this is responsible for the overwhelming majority of the extra rewrite calls I was seeing.

My latest change includes a hack that I found minimizes the difference of the (relatively minor) perf regression I was able to observe on tiny10, which I was unable to generalize because doing so added back overhead (weirdly enough). My hope was that if the regression observed in CI was correlated with what I was seeing on tiny10, this would prove it and help and I would be able to work on a cleaner solution after the fact.

> Thanks, really looking forward to merging this.
I just pushed to update_benchmark again, if the issue still remains I'll try to repro on both tinyboxes and give you an update.
Our CI doesn't run on tiny10.

Unfortunately, by the looks of the latest update_benchmark my hack didn't seem to help, so I'm a bit stuck now. I'm also looking forward to getting this merged, but I'm a little depressed by the results lol. Looking forward to your investigation, thanks for helping.

greg-niemeyer:
So all I did was remove the hack and rebase on master, and the benchmark in CI is now showing much improved numbers, though still not exactly the same as what I'm seeing on tiny10. Testing locally on tiny10, I see similar behavior as before where this branch performs step 1 a couple hundreds of a millisecond slower, consistently. If I reapply the hack, that slowdown goes away. Regardless, if I just wrap the sdxl.py command with `time`, there's basically no diff between the perf of this branch (with or without the hack) and master (synced to the same state that's getting merged back into the dev branch) -- and looking at how process_replay works and the headers from the update_benchmark job, wrapping the command in `time` is not too far off from how CI is benchmarking the run.

Qazalin:
timings look much better! Let's move forward with this once CI passes.

Qazalin:
I wonder if render_const and some more cstyle/ptx stuff can be simpler now.

greg-niemeyer:
> I wonder if render_const and some more cstyle/ptx stuff can be simpler now.

I think so -- the original purpose of #5313 can probably be addressed. I'm down to push more changes toward this end.

greg-niemeyer:
#5991 simplifies cstyle render_const. I can look for more simplifications and push them in separate PRs, or pile more into that one if there's other low hanging fruit.

greg-niemeyer:
oof -- i forgot there's process replay stuff that happens when trying to simplify that way. i closed it and will open another when i have something decent.

greg-niemeyer:
#6024 should be good for render_const

PR #5323, opened 2024-07-08
fix linearizer to adjust for expand-less multireduces

0xtimmy:
This changes the linearizer to handle sequential, fused reduceops in different dimensions to resolve the reduce-expand-reduce pattern in LazyOps

basically, if we have a variance calculation:
```
x = randn(4,32)       // x.shape = (4,32)
mu = x.sum()          // mu.shape = (4,1) ; there is a reduction
diff = x - mu         // diff.shape = (4,32) ; and expand back to the full shape
var = (diff^2).mean() // var.shape = (4,1) ; and then a reduction again
```
this is how multiple reduceops are fused currently.
what we want, and what this diff implements is:
```
mu = x.sum()          // mu.shape = (4,32,1) ; x is loaded with shape=(4,32,32), strides=(32, 0, 1) so there is a reduction
diff = x - mu         // diff.shape = (4,32,1) ; x is loaded with shape=(4,32,1), strides=(32,1,0) so there's no expand
var = (diff^2).mean() // var.shape = (4,1,1) ; and then a reduction again
```

draft, remaining bugs are:
- reuse of reduceops, bug demonstrated by `test_verify_lazyop_reduceop_reuse` in `test_linearizer.py`
- clang/lllvm for sequential reduceops


Qazalin:
I think the first step should be to fix all the existing multi reduce tests in test_linearizer to pass verify_lazyop. Then we can iterate on codegen.

tinygrad/codegen/uops.py
```diff
@@ -386,11 +386,14 @@ def linearize(self, extra_pm:Optional[PatternMatcher]=None, do_type_verify=True)
 
     # fixup gated stores with an IF block to save extra local loads
     @functools.lru_cache(None)
-    def _dfs(u:UOp, gate:UOp) -> UOp:
-      if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER:
+    def _dfs(u:UOp, gate:Union[UOp|None]=None, phi:Union[UOp|None]=None) -> UOp:
```
Qazalin, line 389: is late_gate reset enough for correctness in your tests? Was fine for me https://github.com/tinygrad/tinygrad/pull/5346

We have to write a simpler IF rewrite for multireduce at some point, but let's focus on correctness before speed.

tinygrad/codegen/linearizer.py
Qazalin, line 1: I wouldn't spend too much time on Linearizer fixups since it'll be replaced by Lowerer soon. (PR 4957)

test/test_linearizer.py
```diff
@@ -203,6 +203,91 @@ def test_early_end_local(self):
...
+          ), arg=(2,)),
+        )),
+      ), arg=(1,)),
+    ), arg=MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4,1,1), strides=(1,0,0), offset=0, mask=None, contiguous=False),))))
+    verify_lazyop(ast)
+    Tensor.manual_seed(0)
+    x = Tensor.randn(4,32).realize()
+    helper_linearizer_ast((ast,), [x], wanna_output=[(x.numpy()+x.numpy().sum(axis=-1, keepdims=True)).sum(axis=-1)])
+
+  def test_verify_lazyop_parallel(self):
```
Qazalin, line 222: let's focus on sequential for now.

Qazalin:
To really make progress here we should narrow down on mean+std fusion correctness.
Then generalize to other sequential (softmax, layernorm).
Parallel is worth doing as a standalone project.

tinygrad/codegen/uops.py
0xtimmy: reseting the late gate is enough for correctness, this is so that the dfs will propagate through reduction to visit potential preceding reduceops and give them their own if blocks

0xtimmy:
Hmm sounds good, I took some time to work on the scheduler while the lowerer was getting merged. I can push that if you want to take a look.
I'll revisit the codegen and see if I can get all opts to work for std this weekend

Qazalin:
yea we really need to first pass `TestLinearizer.test_var_multireduce` before scheduling work.

0xtimmy:
#5515 should pass `test_var_multireduce` with the lowerer, working on factoring in the old tests

Qazalin:
fyi closing this in favor of your lowerer diff.

PR #5324, opened 2024-07-08
Support invert dunder method

obadakhalili:
Supports the following:
```py
a = Tensor([True, False])
b = ~a # a.logical_not()
```


chenyuxyz:
i think it needs to invert bits and works beyond boolean

also need tests, preferably in `test_ops.py`

chenyuxyz:
no update in 2 weeks, feel free to reopen with updates

PR #5332, opened 2024-07-08
add dataset download file for retinanet

reddyn12:


chenyuxyz:
doesn't the existing `datasets/openimages.py` work? why not? can you make it work instead of adding a separate file?

also generally if you don't spend time read and clean up your pr and make it good, it won't get reviewed/merged

reddyn12:
My plan was to replace the current open images with my new implementation and make sure it also works on model_eval. Factoring out the download part was the first step. I'll use the download method in the current implementation as it doesn't require the fifty one dependency, which is bloated. Ill download it and see what the difference is, if any.

reddyn12:
Master download works fine

PR #5335, opened 2024-07-08
fix Tensor.all and Tensor.any for PTX

chenyuxyz:
supported boolean acc and boolean phi. and rewrite boolean max to uint8 max


chenyuxyz:
i think we only need to change max. sum in acc_dtype=bool seems to work fine

PR #5336, opened 2024-07-08
minor PTX matcher cleanup [run_process_replay]

chenyuxyz:
uop.cast syntatic sugar and some newline/space cleanup

### DIFF
--- a/tinygrad/renderer/assembly.py
+++ b/tinygrad/renderer/assembly.py
@@ -243,7 +243,7 @@ ptx_matcher = PatternMatcher([
     for op in PTXRenderer.asm_for_op.keys() if op not in PTXRenderer.supports_half],
   (UPat(UOps.ALU, name="x", dtype=dtypes.bool, arg=BinaryOps.MAX),
     lambda x: UOp(UOps.ALU, dtypes.uint8, tuple(s.cast(dtypes.uint8) for s in x.src), x.arg).cast(dtypes.bool)),
-  # TODO: this one looks sketchy
+  # TODO: this one looks sketchy, root.arg is applied to outer cast?
   (UPat(UOps.LOAD, name="root", dtype=dtypes.bool, src=(UPat(name="x"),UPat(name="y"),UPat(name="z"),UPat(name="k"))),
     lambda root,x,y,z,k: UOp(UOps.CAST, dtypes.bool, (UOp(root.op, dtypes.int8, (x,y,z,UOp(UOps.CAST, dtypes.uint8, (k,)))),), root.arg)),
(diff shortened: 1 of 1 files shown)


tinygrad/renderer/assembly.py
```diff
@@ -238,24 +236,24 @@ def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
...
+  (UPat(UOps.ALU, BinaryOps.ADD, [UPat(name="non_muls"), UPat(UOps.ALU, BinaryOps.MUL, name="muls")], "root"),
     lambda root, muls, non_muls: UOp(UOps.ALU, root.dtype, muls.src + (non_muls,), TernaryOps.MULACC)),
   *[(UPat(UOps.ALU, op, dtype=dtypes.half, name="x"),
-    lambda x: UOp(UOps.CAST, dtypes.half, (UOp(x.op, dtypes.float32, tuple([UOp(UOps.CAST, dtypes.float32, (vv,)) for vv in x.src]), x.arg),)))
+    lambda x: (UOp(x.op, dtypes.float32, tuple([vv.cast(dtypes.float32) for vv in x.src]), x.arg).cast(dtypes.half)))
     for op in PTXRenderer.asm_for_op.keys() if op not in PTXRenderer.supports_half],
   (UPat(UOps.ALU, name="x", dtype=dtypes.bool, arg=BinaryOps.MAX),
-    lambda x: UOp(UOps.CAST, dtypes.bool, (UOp(UOps.ALU, dtypes.uint8, tuple(UOp(UOps.CAST, dtypes.uint8, (s,)) for s in x.src), x.arg),))),
+    lambda x: UOp(UOps.ALU, dtypes.uint8, tuple(s.cast(dtypes.uint8) for s in x.src), x.arg).cast(dtypes.bool)),
+  # TODO: this one looks sketchy, root.arg is applied to outer cast?
```
chenyuxyz, line 246: @SzymonOzog fyi. this was first introduced in #4043. it's not clear to me why an arg in added to cast, also the middle dtype is in int8, not uint8. does this rule ever hit?

SzymonOzog: root.arg was probably an oversight, the middle type was int8 just because we needed to load it into a regular register and not a predicate register - should also be fine with uint8

chenyuxyz: i fixed it in #5338, int8 does not work if the this rule hits because we have dtype checks. this rule did not hit before

SzymonOzog: My bad then - apologies for misleading

PR #5340, opened 2024-07-09
[bug fix] nested commutative pattern _match [run_process_replay] [no_assert]

DKormann:
since the test uses list for src the UPat should match in commutative fashion.

### DIFF
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -155,22 +155,20 @@ T = TypeVar("T")
 def __unmatch(m1:Union[T, Set[T]], m2:T) -> bool: return m2 not in m1 if isinstance(m1, set) else m2 != m1
 
 def __match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> List[Dict[str, UOp]]:
-  stores = [store.copy()]
+  stores, res = [store.copy()], []
   if pat.name is not None and stores[0].setdefault(pat.name, uop) is not uop: return []
   if pat.arg is not None and __unmatch(pat.arg, uop.arg): return []
   if pat.dtype is not None and uop.dtype is not None and __unmatch(pat.dtype, uop.dtype): return []
   if pat.op is not None and __unmatch(pat.op, uop.op): return []
   if pat.src is None: return stores
 
-  res = []
   # only one if it's a tuple
   # try all permutations if it's a list
   # repeat if it's a UPat
   for vp in itertools.permutations(pat.src) if isinstance(pat.src,list) else ([pat.src] if isinstance(pat.src,tuple) else [(pat.src,)*len(uop.src)]):
     if len(uop.src) != len(vp) and (len(uop.src) not in pat.allow_len) and not pat.allow_any_len: return []
     new_stores = [stores[0].copy()]
-    for uu, vv in zip(uop.src, vp):
-      new_stores = [rstore for nstore in new_stores for rstore in __match(uu, vv, nstore)]
+    for uu, vv in zip(uop.src, vp): new_stores = [rstore for nstore in new_stores for rstore in __match(uu, vv, nstore)]
     res.extend(new_stores)
   return res
 
--- a/tinygrad/codegen/uops.py
+++ b/tinygrad/codegen/uops.py
@@ -163,6 +163,9 @@ def __match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> List[Dict[str, UOp]]:
   if pat.src is None: return stores
 
   res = []
+  # only one if it's a tuple
+  # try all permutations if it's a list
+  # repeat if it's a UPat
   for vp in itertools.permutations(pat.src) if isinstance(pat.src,list) else ([pat.src] if isinstance(pat.src,tuple) else [(pat.src,)*len(uop.src)]):
     if len(uop.src) != len(vp) and (len(uop.src) not in pat.allow_len) and not pat.allow_any_len: return []
     new_stores = [stores[0].copy()]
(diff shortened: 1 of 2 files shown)


DKormann:
the changes to _match may lead to somewhat of a slower rewrite. but it will match more completely and generate better uops in the end

DKormann:
speed change for me for `PROFILE=0 python test/external/external_benchmark_schedule.py`

Qazalin:
can you run it with `[run_process_replay] [no_assert]`? Wanna see what this changes.

tinygrad/codegen/uops.py
```diff
@@ -154,22 +154,27 @@ def compile(u: UOp, name:Optional[str]=None) -> UPat:
...
-  if pat.arg is not None and __unmatch(pat.arg, uop.arg): return False
-  if pat.dtype is not None and uop.dtype is not None and __unmatch(pat.dtype, uop.dtype): return False
-  if pat.op is not None and __unmatch(pat.op, uop.op): return False
-  if pat.src is None: return True
+def __match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> List[Dict[str, UOp]]:
+  stores, res = [store.copy()], []
+  if pat.name is not None and stores[0].setdefault(pat.name, uop) is not uop: return []
+  if pat.arg is not None and __unmatch(pat.arg, uop.arg): return []
+  if pat.dtype is not None and uop.dtype is not None and __unmatch(pat.dtype, uop.dtype): return []
+  if pat.op is not None and __unmatch(pat.op, uop.op): return []
```
Qazalin, line 162: hmm, I'd return False here,
because there are some cases where we match and stores are empty, like NOOP rules:
```
(UOp.store(UOp.var(), UOp.var(), UOp.var(), UOp.const(dtypes.bool, False)), lambda: UOp(UOps.NOOP)),
```

DKormann: on a successful match we should never return and empty list
when there are no var names we return [{}] when we match and [] on no match

tinygrad/codegen/uops.py
```diff
@@ -154,22 +154,26 @@ def compile(u: UOp, name:Optional[str]=None) -> UPat:
...
 def __unmatch(m1:Union[T, Set[T]], m2:T) -> bool: return m2 not in m1 if isinstance(m1, set) else m2 != m1
 
-def _match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> bool:
-  if pat.name is not None and store.setdefault(pat.name, uop) is not uop: return False
-  if pat.arg is not None and __unmatch(pat.arg, uop.arg): return False
-  if pat.dtype is not None and uop.dtype is not None and __unmatch(pat.dtype, uop.dtype): return False
-  if pat.op is not None and __unmatch(pat.op, uop.op): return False
-  if pat.src is None: return True
+def __match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> List[Dict[str, UOp]]:
+  stores, res = [store.copy()], []
```
Qazalin, line 158: why do you need to copy stores regardless of the commutative branch?
https://github.com/tinygrad/tinygrad/commit/b8a994ab91815b606e284a46c58b3ccea6fdfc0a

tinygrad/codegen/uops.py
```diff
@@ -154,22 +154,26 @@ def compile(u: UOp, name:Optional[str]=None) -> UPat:
...
-      store.update(new_store)
-      return True
-  return False
+    if len(uop.src) != len(vp) and (len(uop.src) not in pat.allow_len) and not pat.allow_any_len: return []
+    new_stores = [stores[0].copy()]
+    for uu, vv in zip(uop.src, vp): new_stores = [rstore for nstore in new_stores for rstore in __match(uu, vv, nstore)]
+    res.extend(new_stores)
+  return res
+
+def _match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> bool:
```
roelofvandijk, line 174: I wonder if this logic can just move here https://github.com/tinygrad/tinygrad/blob/d5a68ae6b31a9c4edba7e38bf29f25fb9b7c66a8/tinygrad/codegen/uops.py#L190

tinygrad/codegen/uops.py
DKormann: oh thats good. I think res is needed however

tinygrad/codegen/uops.py
DKormann: i dont see it

tinygrad/codegen/uops.py
```diff
@@ -154,22 +154,26 @@ def compile(u: UOp, name:Optional[str]=None) -> UPat:
...
+def __match(uop:UOp, pat:UPat, store:Dict[str, UOp]) -> List[Dict[str, UOp]]:
+  if pat.name is not None and store.setdefault(pat.name, uop) is not uop: return []
+  if pat.arg is not None and __unmatch(pat.arg, uop.arg): return []
+  if pat.dtype is not None and uop.dtype is not None and __unmatch(pat.dtype, uop.dtype): return []
+  if pat.op is not None and __unmatch(pat.op, uop.op): return []
+  if pat.src is None: return [store]
   # only one if it's a tuple
   # try all permutations if it's a list
   # repeat if it's a UPat
+  res = []
```
Qazalin, line 166: i'd appreciate a type annot

tinygrad/codegen/uops.py
DKormann: ah wait i see it

DKormann: nice actually negative line diff

PR #5353, opened 2024-07-09
graph_rewrite sort src

DKormann:
improves matching and almost solves symbolic equality

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -35,7 +35,7 @@ def _assert_eq(tensor:Tensor, target_dtype:DType, target):
   if DEBUG >= 2: print(tensor.numpy())
   try:
     assert tensor.dtype == target_dtype
-    np.testing.assert_allclose(tensor.numpy(), target, rtol={dtypes.float16:1e-3, dtypes.bfloat16:1e-2}.get(target_dtype, 1e-7))
+    np.testing.assert_allclose(tensor.numpy(), target, rtol={dtypes.float16:1e-3, dtypes.bfloat16:1e-2}.get(target_dtype, 1e-6))
   except AssertionError as e:
     raise AssertionError(f"\ntensor {tensor.numpy()} dtype {tensor.dtype} does not match target {target} with dtype {target_dtype}") from e
 
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -35,7 +35,7 @@ def _assert_eq(tensor:Tensor, target_dtype:DType, target):
   if DEBUG >= 2: print(tensor.numpy())
   try:
     assert tensor.dtype == target_dtype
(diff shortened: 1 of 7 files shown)


test/test_dtype.py
```diff
@@ -35,7 +35,7 @@ def _assert_eq(tensor:Tensor, target_dtype:DType, target):
   if DEBUG >= 2: print(tensor.numpy())
   try:
     assert tensor.dtype == target_dtype
-    np.testing.assert_allclose(tensor.numpy(), target, rtol={dtypes.float16:1e-3, dtypes.bfloat16:1e-2}.get(target_dtype, 1e-7))
+    np.testing.assert_allclose(tensor.numpy(), target, rtol={dtypes.float16:1e-3, dtypes.bfloat16:1e-2}.get(target_dtype, 1e-6))
```
chenyuxyz, line 38: why is this needed?

test/test_uop_graph.py
```diff
@@ -15,8 +15,8 @@
 
 class TestGraphRewrite(unittest.TestCase):
   def test_dedup(self):
-    v1 = UOp(UOps.DEFINE_VAR, dtypes.float)
-    v2 = UOp(UOps.DEFINE_VAR, dtypes.float)
+    v1 = UOp(UOps.DEFINE_VAR, dtypes.float, arg=Variable('v', 0, 1))
```
chenyuxyz, line 18: also change this to `dtypes.int`?

tinygrad/codegen/uops.py
```diff
@@ -350,7 +350,7 @@ def graph_rewrite(sink:UOp, pm:PatternMatcher) -> UOp:
   replace: Dict[UOp, UOp] = {}
   def __inner_rewrite(n:UOp) -> UOp:
     if n in replace: return replace[n]
-    replace_source = (n.op, n.dtype, tuple(__inner_rewrite(y) for y in n.src), n.arg)
+    replace_source = (n.op, n.dtype, (lambda x: tuple(sorted(x) if n.commutative() else x))(__inner_rewrite(y) for y in n.src), n.arg)
```
chenyuxyz, line 353: it might be less sorts and separate the rewrite with sort if we sort in the constructor? so `n.src` here is always sorted if n is commutative, instead of only sort in rewrite

test/test_dtype.py
DKormann: there are slight rounding errors due to different order of calculations 
test_dtype.py :: test_arange will produce 5.7999997 instead of 5.8

tinygrad/codegen/uops.py
DKormann: seemed like a big change but id  love to do that

test/test_dtype.py
DKormann: can put it down to 3e-7

PR #5356, opened 2024-07-10
mini refactor: simpler view.unbind

roelofvandijk:
2 changes @chenyuxyz:
* v.unbind() is atomic, can be called twice to remove the `var_unbound_var` variable
* a function `substitute` (can also be a lambda) reduces some complexity here

One could argue that the 4  `new_...` variables don't have to be created now, but it is more readable to keep them.


chenyuxyz:
`var_unboundvar_val` was there so that you don't need to call `unbind` twice

`tuple(map)` is probably fine, i don't find `tuple(tuple(map))` for mask more readable

roelofvandijk:
Addressed your comments.

PR #5357, opened 2024-07-10
optimize memplanner

nimlgen:
This approach allocates one big buffer and `offsets` it for other buffers. So that's basically an offline dynamic storage allocation problem. I will play with this more, but I think ~10% overhead (from the theoretical best) should be reachable and not that complicated. Also, one buffer benefits us, since small allocations are not rounded to physical 4k.

On ResNet, it goes from 79GB to 71GB with a simple best-fit algo:
`1  350.39 ms run,   14.72 ms python,  85.29 ms fetch data,  250.38 ms AMD * 6,  7.02 loss, 0.00 acc, 0.030180 LR, 70.81 GB used, 108726.35 GFLOPS`
vs
`2  351.82 ms run,   29.50 ms python, 116.64 ms fetch data,  205.68 ms AMD * 6,  7.02 loss, 0.00 acc, 0.012934 LR, 79.25 GB used, 108286.10 GFLOPS`

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -337,28 +337,78 @@ def _internal_memory_planner(buffers:List[Union[List[Buffer], Tuple[Buffer, ...]
       if buf.base not in first_appearance: first_appearance[buf.base] = i
       last_appearance[buf.base] = i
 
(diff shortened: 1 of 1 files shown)


nimlgen:
Idea is not bad, will reopen when it'll be ready to merge

PR #5365, opened 2024-07-11
Retinanet initializers

reddyn12:


extra/models/retinanet.py
```diff
@@ -1,9 +1,13 @@
 import math
 from tinygrad.helpers import flatten, get_child
-import tinygrad.nn as nn
+from tinygrad.nn import Conv2d
 from extra.models.resnet import ResNet
 import numpy as np
 
+Conv2dNormal = Conv2d
+Conv2dNormal_priorprob = Conv2d
```
wozeparrot, line 8: how does this differ from `Conv2dNormal`?

reddyn12: Not in the base case, but when used in training script there is a diff argument:
```retinanet.Conv2dNormal_priorprob = functools.partial(Conv2dNormal, b=-math.log(99))```

PR #5367, opened 2024-07-11
Move transcendental to UOp level

geohot:
@hikettei The bounty is yours, but if you are interested, I thought about it more and I'm moving it to UOps. It's the same code.


chenyuxyz:
i added more transcend math tests to master (we did not really test large input sin, but it works so very nice)

geohot:
Crazy how similar the UOp syntax is to the LazyBuffer. We should refactor UOp to not have Optional dtype.

PR #5369, opened 2024-07-11
more transcend cleanup [run_process_replay]

chenyuxyz:
fix test name, less # noqa: E501 and removed the cast


geohot:
btw, since they are UOps you can use all the syntactic sugar now

PR #5370, opened 2024-07-11
remove getattr [run_process_replay]

geohot:


chenyuxyz:
some issue with process replay since it only captured ast and not conditional on env variable

geohot:
Yea, gonna merge this and fix

PR #5379, opened 2024-07-11
use TC options as ContextVars [run_process_replay]

Qazalin:
prereq for https://github.com/tinygrad/tinygrad/pull/5378


Qazalin:
ok if process replay fails, I need to upstream these ContextVars first.

PR #5383, opened 2024-07-11
fuzz test transcend math

chenyuxyz:
found something wrong with float64 sin reduction. @hikettei can you take a look?

```
import numpy as np
print(Tensor([39800.0], dtype=dtypes.float64).sin().numpy())
print(Tensor([39800.0], dtype=dtypes.float32).sin().numpy())
print(Tensor([39800.0], dtype=dtypes.float16).sin().numpy())
print(np.sin(np.array([39800.0], dtype=np.float64)))
print(np.sin(np.array([39800.0], dtype=np.float32)))
print(np.sin(np.array([39800.0], dtype=np.float16)))
```

```
CLANG=1 python test.py
[0.92785633]
[0.7428573]
[-0.7705]
[0.74285722]
[0.7428572]
[-0.7705]
```

### DIFF
--- a/test/test_transcendental.py
+++ b/test/test_transcendental.py
@@ -13,21 +13,19 @@ settings.load_profile("my_profile")
 
 class TestTranscendentalMath(unittest.TestCase):
   @unittest.skipUnless(is_dtype_supported(dtypes.float64, Device.DEFAULT), f"no float64 on {Device.DEFAULT}")
+  @unittest.skipIf(getenv("CUDACPU") or (getenv("MOCKGPU") and Device.DEFAULT == "NV"), "crashed")
   @given(ht.float64, strat.sampled_from([(Tensor.exp, np.exp), (Tensor.log, np.log), (Tensor.sin, np.sin)]))
   def test_float64(self, x, op):
     if op[0] == Tensor.sin:
       # TODO: reduction does not work
       if abs(x) > 2 ** 32: return
-      # crashes in CUDACPU
-      if (getenv("CUDACPU") or (getenv("MOCKGPU") and Device.DEFAULT == "NV")): return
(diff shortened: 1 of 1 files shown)


hikettei:
Of course.

- One immediate action was to set the `switch_over=1e14` for fp64 execution, (just followed the original SLEEF https://github.com/shibatch/sleef/blob/master/src/common/misc.h#L81).

- This will allow calculations with an error margin of up to 1e-5 abs within the range of x <= `203474075648.0`

- Please allow me to think about values larger than that until tomorrow.

chenyuxyz:
yea if i do switch over 1e14, then it failed at 1e14. take your time and you can verify with this fuzz test

hikettei:
i am making corrections here https://github.com/tinygrad/tinygrad/pull/5385.

chenyuxyz:
@hikettei fyi i have included your latest change, it looks like log of a number very close to zero also produce different output from numpy

i want to have a version of this merged and i can start to clean up using uop syntactic sugar. It's fine to say skip if sin input is too large or log input is too close to 0. let me know if this is known limitation or we need to investigate more

hikettei:
@chenyuxyz Thank you for the question!

1. fp64 xlog2(x) where x is very close to zero. (nearly FLT_MIN)

I have confirmed that it can compute down to the value of DBL_MIN. I should pass the log2 test. I will create a small PR. please wait a moment.

2. fp64 xsin(x) where x is very large (nearly FLT_MAX)

This is a limitation due to approximations in tinygrad. I searched for other implementations to calculate fp64 sin, but it overflows without at least uint128 for reduction.
Implementing this in tinygrad requires significant modifications. So, please skip inputs that exceed the maximum value of fp32 for now when testing fp64 sin.

PR #5384, opened 2024-07-11
merge get_children_dfs [run_process_replay]

DKormann:


geohot:
Why is this better? It looks less readable and less generic to me

Qazalin:
oh this was my [suggestion](https://discord.com/channels/1068976834382925865/1068982781490757652/1260913954759376979) after profiling codegen/*
The slowest calls in the uop toposort are these two, if we can merge them I think it'd be a speedup.

DKormann:
yeah it felt like there would be a simplification maybe even possible to fold into the toposort or graph_rewrite not sure anymore

DKormann:
we are also only doing one function call now but i dont think its faster

PR #5388, opened 2024-07-11
db in wal mode

wozeparrot:
switch the database to WAL mode: <https://www.sqlite.org/wal.html>

the largest benefit this provides is that running tests with `xdist` with a large number of processes can run without database locking issues.

that issue is seen downstream: https://github.com/NixOS/nixpkgs/pull/318001


geohot:
Hmm that disk tensor issue might not be the DB. I don't actually know if those tests are threadsafe, they write to the disk.

wozeparrot:
`test_disk_tensor` should be thread safe now, db change is still needed as that's unrelated https://github.com/NixOS/nixpkgs/pull/316931#issuecomment-2153439239

PR #5389, opened 2024-07-11
indexing getting better

geohot:


geohot:
There's three kernels in indexing now. @Qazalin we need the scheduler to fuse them.

`NOOPT=1 DEBUG=5 python3 test/test_arange.py TestIndexing`

geohot:
Why is process replay asserting?

Qazalin:
oh you don't need the explicit no_assert anymore, it always defaults to no_assert.

Qazalin:
hmm process replay is still asserting in diff-only.
test_var_multireduce passes verify_lazyop now! thanks.

geohot:
Yea, want to add another `verify_lazyop` in Lowerer? You'll have to support ReduceOps.WMMA too

Qazalin:
yea WMMA needs work, also need to figure out how to handle the global dim in the local reduce. 
like:
```
r_16_16
  0 ━┳ BufferOps.STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(16, 1), strides=(1, 0), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ BinaryOps.ADD None
  2    ┣━┳ UnaryOps.CAST dtypes.float
  3    ┃ ┗━┳ UnaryOps.CAST dtypes.half
  4    ┃   ┗━┳ ReduceOps.SUM (1,)
  5    ┃     ┗━┳ BufferOps.LOAD MemBuffer(idx=-1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(1, 16), strides=(0, 1), offset=0, mask=None, contiguous=True),)))
  6    ┃       ┗━┳ BufferOps.STORE MemBuffer(idx=-1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(1, 16), strides=(0, 1), offset=0, mask=None, contiguous=True),)))
  7    ┃         ┗━┳ ReduceOps.SUM ()
  8    ┃           ┗━━ BufferOps.LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(16, 16), strides=(1, 16), offset=0, mask=None, contiguous=False),)))
  9    ┗━━ BufferOps.LOAD MemBuffer(idx=2, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(16, 1), strides=(1, 0), offset=0, mask=None, contiguous=True),)))

AssertionError: found implicit movement op ReduceOps.SUM (16, 16) != BufferOps.STORE (1, 16)
```
we have to keep the state of that global dim, ignore in local store then add it back in the global store.

geohot:
Can I fix the process replay still asserting?

Qazalin:
weird, no_assert works fine https://github.com/tinygrad/tinygrad/pull/5393

Qazalin:
ohh it's not NO_ASSERT that's failing is a rendering error for UNMUL

geohot:
Ahh, okay I'll look

geohot:
Maybe always process replay should never actually fail? Can other changes cause it to assert like this?

Qazalin:
Yea, here's the AST that fails to lower in master:
```
FAILED TO RECREATE KERNEL
(LazyOp(op=BufferOps.STORE, src=(LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BinaryOps.MUL, src=(LazyOp(op=BinaryOps.ADD, src=(LazyOp(op=ReduceOps.SUM, src=(LazyOp(op=BufferOps.CONST, src=(), arg=ConstBuffer(val=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(16385, 32767), strides=(0, 0), offset=0, mask=((0, 16385), (16383, 32767)), contiguous=False), View(shape=(16384, 16384), strides=(1, 32768), offset=0, mask=None, contiguous=False))))),), arg=(1,)), LazyOp(op=BufferOps.CONST, src=(), arg=ConstBuffer(val=-1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(16384, 1), strides=(0, 0), offset=0, mask=None, contiguous=False),))))), arg=None), LazyOp(op=BufferOps.LOAD, src=(), arg=MemBuffer(idx=1, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(16384, 1), strides=(1, 0), offset=0, mask=None, contiguous=True),))))), arg=None),), arg=(0,)),), arg=MemBuffer(idx=0, dtype=dtypes.int, st=ShapeTracker(views=(View(shape=(1, 1), strides=(0, 0), offset=0, mask=None, contiguous=True),)))),)
[]
```

geohot:
Hmm, so if we add new AST support in a branch, this will fail CI? We should just have this not assert.

geohot:
okay, so test_var_multireduce doesn't work... I got single kernel indexing working though!

Qazalin:
Yea lowerer just creates
```
__kernel void r_81_32_32(__global float* data0) {
  int gidx0 = get_group_id(0); /* 81 */
  data0[gidx0] = 0.0f;
}
```
for it, I'm not sure if self.first_reduce and self.full_shape generalize to multi reduce now.

geohot:
Screenshot 2024-07-11 at 4 34 52 PM

geohot:
first_reduce should, but full_shape doesn't. full_shape should be the axis-wise-max of all the shapes.

Qazalin:
yea first_reduce might still work fine, but the Lowerer shouldn't really use it for self.ridxs https://github.com/tinygrad/tinygrad/blob/c2da4454cda15580b4bad6f3950f09d88ade6699/tinygrad/codegen/lowerer.py#L145

PR #5392, opened 2024-07-11
speedup _match

DKormann:
matching was way worse than exponential on certain cases

```python
u = UOp.const(dtypes.int, 1)
p = UPat(name = "a")

for i in range(11):
  u = u+u
  p = UPat(UOps.ALU, BinaryOps.ADD, [p, p])

len(_match(u, p, {}))
```


DKormann:
still the runtime is exponential i will look into caching for the _match function itself

DKormann:
not sure how relevant this is because it might be exponential runtime only in regard to UPat depth which maybe shouldnt be too big in the first place

DKormann:
will look into faster caches maybe store could be a tuple instead of a dict since it should only hold a small number of keys anyway

geohot:
Can you benchmark this generally on real world workloads? We should never do speedups that only improve a contrived test

PR #5397, opened 2024-07-11
generalize FUSE_AS_ONE_KERNEL in the scheduler

Qazalin:
first starting with passing ReduceOps on const.


geohot:
Right now it hits issues with the ReduceOps not being contiguous. This is what we need to make generic.

Qazalin:
hmm, so if the st buildup can't be contig it should either break that kernel or try fixup the ShapeTrackers pre reduce.

PR #5403, opened 2024-07-12
refactor: consolidate replace [run_process_replay]

roelofvandijk:
Check: Are PRs like this that condense the logic welcome?


Qazalin:
hmm, I guess the tcc vs cc distinction has some semantic value.

roelofvandijk:
Agreed - is there a better name for it here? `replace` is taken.

geohot:
PRs like this are welcome, it's shorter, faster, and more readable.

PR #5412, opened 2024-07-12
reverse global dim order

chenyuxyz:
this is the old linearizer bahavior and a lot faster for resnet kernels with > 3 global dim. eventually we want to search this


chenyuxyz:
need to fix the limit first

PR #5432, opened 2024-07-12
Tensor method support dim <-> axis and keepdim <-> keepdims

chenyuxyz:
default is dim and keepdim like torch, and axis / keepdims work


chenyuxyz:
auto-generated one is quite slow, hard-coded one is faster but ugly

chenyuxyz:
still ~10% slower unjitted gpt2 on tinyboxes

PR #5444, opened 2024-07-13
new UOp style patterns [run_process_replay]

cpwrs:
~~adds UOp ability to express "all permutations of src",~~ rewrites more patterns in the new UOp style

First PR towards hopefully a bigger refactor. Making UOp equally expressive means they could pattern match without UPat at all. At the very least UPat could be refactored to a dataclass with cleaner types and one constructor (compile).

Patterns UOp can't express yet:
- [ ] All permutations of src
- [ ] For any dtype in set
- [ ] For any op in set
- [ ] Allowed len (less hacky)

### DIFF
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -377,8 +377,8 @@ constant_folder = PatternMatcher([
   (UOp(UOps.PHI, src=(UOp(UOps.DEFINE_ACC, src=(UOp.cvar(),)), UOp.var("x"))), lambda x: x),
   (UOp(UOps.PHI, src=(UOp.cvar(), UOp.var("x"))), lambda x: x),
   # a DEFINE_ACC without inputs is a const + GEP on a const is the const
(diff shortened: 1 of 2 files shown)


DKormann:
> pattern match without UPat at all

DKormann:
you can just make PHI and DEFINE_ACC (passes tests for me)

geohot:
PHI and DEFINE_ACC aren't commutative though.

A new UOp shouldn't be created for this. Maybe UPat should superclass UOp.

cpwrs:
> A new UOp shouldn't be created for this.

Got it. Any reason UOps.VAR is an exception? The UOp syntax is nice for patterns, especially with all the dunders, which is why I thought adding expression would be worth a few non-rendered UOps. 

It's hard to make a clean refactor if some pattern stuff is expressed in UOp and some is not. I think it should be all or nothing.

There's still a lot of rewrites to the new style without permute() in here, I'll revert the permute stuff.

geohot:
I don't like UOp.VAR, it really should just be a NamedUOp superclass.

geohot:
Merged this, easier to read. I'd support a refactor with the `NOp` class

cpwrs:
Sounds good I'll work on that

PR #5446, opened 2024-07-13
Add narrow type for UPat.arg

roelofvandijk:


Qazalin:
just using int is wrong, what if your const is float or bool?

roelofvandijk:
This is what is currently used in the codebase according to mypy. Should I add `float` and `bool` now, or should we do that once it they are used?

geohot:
What about WMMA? This is just wrong, I don't trust mypy. Look at the types at runtime.

PR #5457, opened 2024-07-13
use python 3.12 in CI [run_process_replay]

chenyuxyz:
might be faster? keep 3.8 for linter


chenyuxyz:
tensorflow-addons is annoying

PR #5461, opened 2024-07-13
process replay add linearizing timer

DKormann:
best of 5 linearizer timer

with NOOPT=1 i'm getting consistent results

will error process replay because added helper function that it cant import


Qazalin:
first thoughts:

1. can this give a reliable estimate?

like how are these times better when there's a clear regression: https://github.com/tinygrad/tinygrad/actions/runs/9939433853/job/27453996953?pr=5461#step:24:293
vs a noop
https://github.com/tinygrad/tinygrad/actions/runs/9939194908/job/27453265611#step:18:627

2. makes some tests flaky - is there a fix?

DKormann:
I believe in the process replay CI we check out master. So a better time would mean better time on master compared to the regression?

Qazalin:
oh it should definitely be the other way...

DKormann:
locally timings are pretty stable for me when run with RUN_PROCESS_REPLAY=2 NOOPT=1

DKormann:
> oh it should definitely be the other way... like "better" means this diff is good

i get your point, for local development i thought its nice you can do a change , run process replay and it tells you if your change was good or not

DKormann:
might wanna compare lin times only on a few real world examples anyway since improvement on arbitrary unit test might not be relevant https://github.com/tinygrad/tinygrad/pull/5392#issuecomment-2224130966

Qazalin:
wanna make RUN_PROCESS_REPLAY a contextvar and set it in some real world tests?

DKormann:
on metal seems stable : https://github.com/tinygrad/tinygrad/actions/runs/9947323023/job/27479754738?pr=5461#step:18:101

test/models/test_real_world.py
```diff
@@ -63,37 +63,41 @@ def test(t, t2): return model(t, Tensor([801]), t2).realize()
...
+      @TinyJit
+      def test(t, t2):
+        for l in model: t = l(t, t2)
+        return t.realize()
+      helper_test("test_unet_resblock", lambda: (Tensor.empty(4, 16, 8, 8), Tensor.empty(1, 24)), test, 0.01, 43)
 
   @unittest.skipUnless(is_dtype_supported(dtypes.float16), "need dtypes.float16")
   def test_llama(self):
-    dtypes.default_float = dtypes.float16
+    with Context(RUN_PROCESS_REPLAY = 2 if RUN_PROCESS_REPLAY else 0):
```
Qazalin, line 77: wanna add this to `def setUp` and `def tearDown`?
example: extra/optimization/test_beam_search.py

DKormann:
seems stable on everything but llvm clang and opencl

DKormann:
continue in #5560

PR #5464, opened 2024-07-13
Retinanet add load from checkpoint

reddyn12:


wozeparrot:
this feels unnecessary

reddyn12:
> this feels unnecessary

I load from checkpoint for eval. The goal is to have 4 model instances running at the same time.

PR #5468, opened 2024-07-14
fix "shape changing bitcast only supported on DISK right now"

davidgonmar:


davidgonmar:
will continue testing things tomorrow

PR #5482, opened 2024-07-14
Add Interpolate Function

tobias17:
# Overview

Brings in the interpolate function [from pytorch](https://pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html) with the linear (and bilinear trilinear) interpolation modes.

## Current Concerns

1. I tried my best to clean it up but the function still feels a little icky, if you have suggestions on how to do it more elegantly please forward them.
2. The pytorch version defaults the mode to `nearest` which is currently not implemented, meaning that the default is `linear`. If we want to add `nearest` in the future it and make it the default then that would break any code using the old default. If this is something we care about then either `nearest` would need to be added before this merges or `mode` should have it's default value stripped so people have to explicitly call for `linear`.


tinygrad/tensor.py
```diff
@@ -1041,8 +1041,8 @@ def gather(self:Tensor, dim:int, index:Tensor) -> Tensor:
     ```
     """
     assert index.ndim == self.ndim, f"self.ndim must equal index.ndim, {self.ndim=}, {index.ndim=}"
-    assert all(s >= i for d,(s,i) in enumerate(zip(self.shape, index.shape)) if d != dim), "requires self.shape[d] >= index.shape[d] for all d != dim"
     dim = self._resolve_dim(dim)
+    assert all(s >= i for d,(s,i) in enumerate(zip(self.shape, index.shape)) if d != dim), "requires self.shape[d] >= index.shape[d] for all d != dim"
```
chenyuxyz, line 1045: can you have this in a standalone pr with a regression test?

chenyuxyz (commented):
there's also https://github.com/tinygrad/tinygrad/blob/3c378efcb60506ad733f90430df7054faf4c123f/extra/onnx_ops.py#L393
that eventually should go into tinygrad proper. does not have to be this pr though

PR #5484, opened 2024-07-14
Add passing tests for shape changing bitcasts for non-disk tensors

kvkenyon:
After investigating, shape changing bitcasts work for non-disk tensors unless the data is non-contiguous. Non-contiguous shape changing bitcasts are not allowed in torch. Is this something we want? The only difference between disk and non-disk is when using assign as disk uses contiguous to force_realize where as non disk will not force realize.


kvkenyon:
Let me try and fix the two tests for gpu and py before closing

kvkenyon:
I tested with clang and metal locally

kvkenyon:
Non view supported devices will fail, on a shape changing bitcast need to merge multiple elements i.e, need multiple loads to be scheduled, then those need to be joined via or and shifts and then casted.

PR #5489, opened 2024-07-15
some limit_dims to limit global merging

chenyuxyz:
only supports merging dims in a way that does not surpass limit, no splitting yet


chenyuxyz:
no process replay diff for the things we run in CI, i think we only hit it in unet3d. @francislata fyi let me know if you still have issue after this

PR #5490, opened 2024-07-15
move colorize_float to helpers.py

wang-edward:
Moved `colorize_float` (used in speed tests) to `helpers.py`.

as per
https://github.com/tinygrad/tinygrad/blob/e21910367768fe034e7b9876d10e694b4cc743f6/test/external/speed_compare_cuda_ptx.py#L9-L10
Let me know if I should make a seperate PR to update all references to it, or keep working in this one.


geohot:
Cool!

Issue #5496, opened 2024-07-15
 Neural Networks documentation examples fail


Benetti-Hub:
The documentation on https://docs.tinygrad.org/nn/ currently fails to generate the output for the examples, instead, it gives the following errors:

```python
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/site-packages/markdown_exec/formatters/python.py", line 71, in _run_python
    exec_python(code, code_block_id, exec_globals)
  File "/opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/site-packages/markdown_exec/formatters/_exec_python.py", line 8, in exec_python
    exec(compiled, exec_globals)  # noqa: S102
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<code block: session tensor; n1>", line 1, in <module>
    conv = nn.Conv1d(1, 1, 3)
            ^^
NameError: name 'nn' is not defined
```

This is due to the fact that [`nn.md`](https://github.com/tinygrad/tinygrad/blob/master/docs/nn.md?plain=1#L3) uses `BatchNorm2d` instead of `BatchNorm` (changing it from tinygrad.nn.BatchNorm2d to tinygrad.nn.BatchNorm fixes the issue).

Additionally, adding `mkdocs build --strict` in https://github.com/tinygrad/tinygrad/blob/master/.github/workflows/docs.yml would ensure that these kinds of errors are reflected in the outcome of the github action.

geohot:
Fixed, would merge PR to add strict and catch these errors

Benetti-Hub:
Thank you! PR added

PR #5497, opened 2024-07-15
add strict mkdocs check

Benetti-Hub:
Adds a mkdocs check on https://github.com/tinygrad/tinygrad/blob/master/.github/workflows/docs.yml to avoid errors like the one reported in https://github.com/tinygrad/tinygrad/issues/5496

I have added the check after `mkdocs gh-deploy --force` to make sure that the documentation is deployed anyway. If needed I can switch them of course.

Let me know if you would like me to proceed with this adjustment or if there are any other changes required.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,5 +26,5 @@ jobs:
           restore-keys: |
             mkdocs-material-
       - run: pip install -e .[docs]
-      - run: mkdocs gh-deploy --force
       - run: mkdocs build --strict
+      - run: mkdocs gh-deploy --force
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,3 +27,4 @@ jobs:
             mkdocs-material-
       - run: pip install -e .[docs]
       - run: mkdocs gh-deploy --force
+      - run: mkdocs build --strict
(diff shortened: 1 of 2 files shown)


wozeparrot:
Could you add a docs build check as part of the tests?

.github/workflows/test.yml
```diff
@@ -119,6 +119,17 @@ jobs:
...
+    name: build-docs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+      - run: pip install -e .[docs]
+      - run: mkdocs build --strict
+
```
wozeparrot, line 132: should just be part of the linter job

Benetti-Hub: Yeah that made more sense, I have moved them

.github/workflows/docs.yml
```diff
@@ -26,4 +26,5 @@ jobs:
           restore-keys: |
             mkdocs-material-
       - run: pip install -e .[docs]
-      - run: mkdocs gh-deploy --force
\ No newline at end of file
+      - run: mkdocs gh-deploy --force
+      - run: mkdocs build --strict
```
wozeparrot, line 30: probably just swap the order, so that we don't get failed docs being deployed.

PR #5516, opened 2024-07-16
FID Inception Model

tobias17:
## Overview

Added the base Inception model and the FID variant required for Stable Diffusion mlperf.

## Testing

Ran identical inputs through tinygrad and pytorch versions of FidInceptionV3 and compared outputs in the following ways:
| Computation | Value |
| :-: | :-: |
| `mean(abs(tiny - torch))` | 0.0004 |
| `mean(abs(tiny))` | 0.3080 |
| `mean(abs(torch))` | 0.3080

### DIFF
--- a/extra/models/inception.py
+++ b/extra/models/inception.py
@@ -239,52 +239,70 @@ class FidInceptionE2(InceptionE):
 
 @lru_cache()
 def default_fid():
-  return torch_load(fetch("https://github.com/mseitzer/pytorch-fid/releases/download/fid_weights/pt_inception-2015-12-05-6726825d.pth", "pt_inception-2015-12-05-6726825d.pth"))
+  return torch_load(str(fetch("https://github.com/mseitzer/pytorch-fid/releases/download/fid_weights/pt_inception-2015-12-05-6726825d.pth", "pt_inception-2015-12-05-6726825d.pth")))
 
 class FidInceptionV3:
-  def __init__(self, inception:Inception3):
-
-    self.blocks = [
-      inception.Conv2d_1a_3x3,
-      inception.Conv2d_2a_3x3,
-      inception.Conv2d_2b_3x3,
-      lambda x: Tensor.max_pool2d(x, kernel_size=(3,3), stride=2, dilation=1),
-
-      inception.Conv2d_3b_1x1,
-      inception.Conv2d_4a_3x3,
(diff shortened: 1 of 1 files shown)


extra/models/inception.py
```diff
@@ -0,0 +1,286 @@
...
+
+class FidInceptionV3:
+  def __init__(self):
+    inception = Inception3(cls_map={
+      "A":  FidInceptionA,
+      "C":  FidInceptionC,
+      "E1": FidInceptionE1,
+      "E2": FidInceptionE2,
+    })
+    state_dict = default_fid()
```
chenyuxyz, line 252: can these be inside a `load from_pretrained` method? preferably nothing is realized / loaded in init

tobias17: 9de2e70c37e8ad5304b13505d60be5c27f2caac2

chenyuxyz: it should be an instance method that loads in the weight once the model has inited, like resent
`https://github.com/tinygrad/tinygrad/blob/4ad83d032eabe7e1b0933c16367ffe9eb71123d1/extra/models/resnet.py#L128`, or search other `load_from_pretrained` examples

tobias17: ok, will take some reorganization since the reference model is not architected for this

tobias17: da0cf2fe0b6b8a9eb2eab99c67f30fb025a3d887

ended up better since we are realizing less weights now

extra/models/inception.py
```diff
@@ -0,0 +1,308 @@
...
+    branch3x3dbl = x.sequential([self.branch3x3dbl_1, self.branch3x3dbl_2])
+    outputs = [
+      self.branch1x1(x),
+      Tensor.cat(self.branch3x3_2a(branch3x3), self.branch3x3_2b(branch3x3), dim=1),
+      Tensor.cat(self.branch3x3dbl_3a(branch3x3dbl), self.branch3x3dbl_3b(branch3x3dbl), dim=1),
+      self.branch_pool(x.max_pool2d(kernel_size=(3,3), stride=1, padding=1)),
+    ]
+    return Tensor.cat(*outputs, dim=1)
+
+@lru_cache()
```
chenyuxyz, line 240: i would not lru_cache this, it won't be free after it's loaded. also this can just inline in the load function

otherwise looks good, will merge after this is fixed

PR #5525, opened 2024-07-17
touchup Tensor.interpolate and Tensor.lerp

chenyuxyz:
rewrite lerp to save one sub and thus flops.
use Tensor.lerp for interpolate and some minor cleanups


chenyuxyz:
nvm the lerp does not help, it's less ops because we count a-b as (a+(-b)) and it doubles the flop now

PR #5546, opened 2024-07-18
intel xmx tensor core

ianpaul10:
Initial shot at introducing the intel tc. Currently only supports float16. Also need to optimize/improve perf.


nimlgen:
I don't think we want to introduce ops_intel for this. Can you just init ops_gpu based on device name or vendor (`CL_DEVICE_VENDOR`)?

ianpaul10:
Right, I guess that makes sense, sorry about that. Was just following this [tutorial](https://mesozoic-egg.github.io/tinygrad-notes/addingaccelerator.html) which I guess was overkill.

Updated to just be based on `CL_DEVICE_VENDOR`= "Intel" or `INTEL=1` (the 2nd one might mostly just for local testing/emulation)

flammit:
I don't understand how this works at all... 

Doesn't K=16 for the Intel XMX?  Is the layout really the same as METAL?

geohot:
Can you paste a working generated kernel? I'll lock bounty when I see that.

ianpaul10:
Yes, will get a working kernel up as soon as I can. Will close this in the meantime since it's not fully baked.

PR #5564, opened 2024-07-18
rm bitcast only on disk

reddyn12:


reddyn12:
nvmd, I did the wrong thing

reddyn12:
Assuming I'm allowed to change the test as:
```
import torch
import numpy as np

data=[100000]
th = torch.tensor(data, dtype=torch.float32)
thb = th.view(torch.uint8)

tg = Tensor(data, dtype=dtypes.float32)
tgb = tg.bitcast(dtypes.uint8)

np.testing.assert_equal(thb.numpy(), tgb.numpy())
```

reddyn12:
This should fix all the failing tests

PR #5566, opened 2024-07-18
swap first and last global in hcopt / hc tc path

chenyuxyz:


chenyuxyz:
improved hlb cifar / stable diffusion / resnet default but not quite to pre-lowerer level yet

PR #5571, opened 2024-07-19
Bitcast on non disk

reddyn12:
Should pass all tests


geohot:
Please test on your branch first. If you submit again and it's not correct, you will not be eligible for this bounty.

reddyn12:
> Please test on your branch first. If you submit again and it's not correct, you will not be eligible for this bounty.

I have tested my branch with LLVM, CLANG, GPU, and CUDA. I wanted to see what failing tests I had left. Is it fine if I open it as a draft? I tested everything but PYTHON and METAL. I can do PYTHON on my machine but I don't have an m series Mac to hit the fp64 error

reddyn12:
> That GPU offset doesn't work btw. It might work on the sim, but you haven't read the OpenCL spec.

Will read it

PR #5574, opened 2024-07-19
full reverse global instead of just swap ends

chenyuxyz:
seems to be slightly faster


geohot:
Want to just restore the old behavior to indexing?

PR #5584, opened 2024-07-19
beamgraph support

geohot:


geohot:
Better version in the MCTS stuff

PR #5588, opened 2024-07-19
Fix typo in Runtime Overview docs

francislata:
I was reading through the Runtime Overview docs section and came across some small typos.


nimlgen:
Thanks!

Issue #5244, opened 2024-07-01
llama speed tracking issue


chaosagent:
- master: 130 tok/s green, 100 tok/s red on 7B, 4 gpus
- master: roughly 15 tok/s green/red 70B

memory layouts (2 all reduce for 4gpu, 2 allreduce + 2 "smear" for 6gpu): x(axis=None) -> w1/w3(axis=0) -> x(axis=-1) -> w2(axis=-1) -> allreduce -> x(axis=None) -> (wq, wk, wv)(axis=0) -> x(axis=kv head dimension) -> attention(axis=kv head dimension) -> x(axis=-1) -> wo(axis=-1) -> allreduce -> next layer

llama 3 OOM: llama 3 has a very big embedding matrix (1GB+). currently code loads weights onto 1 gpu then the rest of them, and embeddings are loaded last, so it will OOM on that loading gpu. solve by loading large tensors first. also can use MLB(weight shards, axis=weight shard axis) to concat weights without realizing on one gpu.

llama 3 OOM with large context: solve with kv shard.

kv shard layout: we have 8 kv heads but 6 gpus. we can shard (2, 2, 1, 1, 1, 1), but we need a way to move to/from this layout from the evenly sharded weights. propose "reshard" api: allow shard on Tensor(mlb), and def shard(devices, axis, splits).

slow kernels: round shards to 32. hard to do generally because you don't want to do this for BS, but you do want it for llm tensor parallel. or we could get the search to learn how to padto+local.

PoC + api proposal: https://github.com/tinygrad/tinygrad/compare/master...chaosagent:tinygrad:llama3_70b 19 tok/s 70b, 140 tok/s 7b on green, fits llama 3 70b w/ 8192 context on amd, but not NV (memory fragmentation?). don't know how to support mlb.pad yet.

jit tune -> 21 tok/s. cpu time spent mostly updating var vals and launch bounds.

currently the api looks like this (notice the splits argument):
```
      # shard
      R = 64
      n_kv_heads = params["args"].get("n_kv_heads", params["args"]["n_heads"])
      if isinstance(device, tuple):
        for k,v in nn.state.get_state_dict(model).items():
          if 'scale' in k: v.shard_(device, axis=None)  # from quantized
          elif '.attention.wo' in k: v.shard_(device, axis=-1, splits=v.shape[-1] // n_kv_heads if "70" not in model_size else R)
[...]

PR #5599, opened 2024-07-20
allow specify splits in shard, handle multiple different splits in MLB.e

chaosagent:
toward #5244 ... 

This PR allows user to specify how to split along sharded axis. Prerequisite for arbitrary shrink on axis and further sharded reshape.

I spent too much time bikeshedding details of this, so here it is, I guess.


geohot:
https://pytorch.org/docs/stable/generated/torch.split.html

geohot:
As discussed on the call, does this match torch behavior?

PR #5600, opened 2024-07-20
MLB support reshape for uneven shards

chaosagent:
depends #5599

allow reshapes for uneven split in cases where no data needs to be moved around

### DIFF
--- a/test/test_multitensor.py
+++ b/test/test_multitensor.py
@@ -446,6 +446,18 @@ class TestMultiTensor(unittest.TestCase):
         X = ((Tensor(data).shard(devices, axis=0) + 1).realize() - 1).realize()
         np.testing.assert_equal(X.numpy(), data)
 
+  def test_multiple_uneven_shard(self):
+    N = 4
+    X = Tensor.rand(4, 1, 257).contiguous().realize()
+    Y = Tensor.rand(4, 1, 257).contiguous().realize()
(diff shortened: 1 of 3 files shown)


geohot:
Why do you want uneven shards?

chaosagent:
good question. there are essentially two new things I want to be able to do. i'm not happy with my approach but i posted because i was feeling stuck. 

when sharing wq/wk/wv in attention, we want to shard the projection weights evenly over 6 devices, but then shard along the head dim when doing attention. both of these are uneven shards, but they have a different uneven distribution, and there needs to be a little redistribution in between.

the other is support arbitrary shrinks on MLB access in order to: MLB([8 shards from huggingface weights], axis=hfaxis).shard(six devices, axis=whatever), and have this do the minimal amount of weight movement. my idea is to just have from_sharded shrink the input MLB to the desired weight ranges, and do copy_to_device on that.

both of these cases can also be handled somewhat by having Tensor.cat([lb.copy_to_device(dev) for lb in self.lbs]).shrink(some range) not copy the lbs that are completely shrunk away. i gave this some thought today but it needs more exploration. maybe this stuff is not necessary with that, but involves some messing with the scheduler, especially if we want to avoid copying unnecessary bytes between devices.

chaosagent:
can write this better!

PR #5601, opened 2024-07-20
Bitcast

reddyn12:
Final attempt. Is PYTHON supposed to simulate bitcast?


reddyn12:
I thought I had it :(. Ill put up a pr tmr with only tests, gl all

reddyn12:
error: -13
CL_MISALIGNED_SUB_BUFFER_OFFSET  
   is returned in errcode_ret if there are no devices in
   context associated with buffer for which the origin value is
   aligned to the CL_DEVICE_MEM_BASE_ADDR_ALIGN value.

PR #5603, opened 2024-07-20
default try padding in apply_tensor_cores

chenyuxyz:
TC_OPT default to 2 (try padding)


chenyuxyz:
overall win for green box but don't like device specific default...

PR #5612, opened 2024-07-20
Failing bitcast test

reddyn12:


geohot:
This test is unreadable. Can you just write a bunch of simple tests that are easy to reason about the expected behavior of

reddyn12:
@geohot I can make a bunch of smaller tests, but won't that be trading 'complexity' with thoroughness? Unless I make 110 tests(11 dtypes * 10 possible casts)? Would this test have been better if I commented it, or a test shouldn't have any logic in it?

PR #5616, opened 2024-07-21
MCTS tweaks

chenyuxyz:
MCTS 500 is competitive with BEAM=8 on resnet on M1 Max.
- increment trial times even with compiled error and runtime error.
- use best time of children as the node value.


geohot:
Good to merge?

PR #5624, opened 2024-07-21
Pass main function name to Compiler.compile() [skip_process_replay]

uuuvn:
to allow for true shellcode backends

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -174,12 +174,12 @@ class CompileError(Exception): pass
 
 class Compiler:
   def __init__(self, cachekey:Optional[str]=None): self.cachekey = None if getenv("DISABLE_COMPILER_CACHE") else cachekey
-  def compile(self, src:str) -> bytes: raise NotImplementedError("need a compile function")
-  def compile_cached(self, src:str) -> bytes:
-    if self.cachekey is None or (lib := diskcache_get(self.cachekey, src)) is None:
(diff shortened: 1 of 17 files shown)


tinygrad/device.py
```diff
@@ -174,12 +174,12 @@ class CompileError(Exception): pass
 
 class Compiler:
   def __init__(self, cachekey:Optional[str]=None): self.cachekey = None if getenv("DISABLE_COMPILER_CACHE") else cachekey
-  def compile(self, src:str) -> bytes: raise NotImplementedError("need a compile function")
-  def compile_cached(self, src:str) -> bytes:
-    if self.cachekey is None or (lib := diskcache_get(self.cachekey, src)) is None:
+  def compile(self, src:str, fname:Optional[str]) -> bytes: raise NotImplementedError("need a compile function")
+  def compile_cached(self, src:str, fname:Optional[str]) -> bytes:
+    if self.cachekey is None or (lib := diskcache_get(self.cachekey, {'src': src, 'fname': fname if fname is not None else ''})) is None:
```
uuuvn, line 179: this `None => ''` thing is required because diskcache_put types sqlite tables based on types of first key

uuuvn:
[skip_process_replay] because of version bump

uuuvn:
One alternative is to pass this information it in source code. For example in clang we can create void _start that calls main function and call it in shellcode ignoring program name

uuuvn:
closed, found a better way to do this

PR #5630, opened 2024-07-22
hcq cache invalidation for beam

nimlgen:


nimlgen:
Looks more stable now

PR #5678, opened 2024-07-24
fold DEFINE_VAR to const if vmin = vmax

chenyuxyz:


chenyuxyz:
we only need this if we are adding init rewriter for UOps

PR #5681, opened 2024-07-24
shorter BufferOps.CONST creation

Qazalin:


tinygrad/engine/schedule.py
```diff
@@ -49,13 +49,9 @@ def _recursive_lazyop(buf:LazyBuffer, inputs:List[LazyBuffer], outputs:Tuple[Laz
...
-    if isinstance(buf.arg, Variable):
-      val, var_val = buf.arg.unbind()
-      var_vals.__setitem__(val, var_val)
-    else:
-      assert isinstance(buf.arg, get_args(ConstType)), f"cannot create ConstBuffer with value {buf.arg}"
-      val = buf.arg
-    return LazyOp(BufferOps.CONST, (), ConstBuffer(val, buf.dtype, unbound_st))
+    if isinstance(arg, Variable): var_vals.update([arg:=arg.unbind()])
+    else: assert isinstance(arg, get_args(ConstType)), f"cannot create ConstBuffer with value {arg}"
+    return LazyOp(BufferOps.CONST, (), ConstBuffer(arg[0] if isinstance(arg, tuple) else arg, buf.dtype, unbound_st))
```
chenyuxyz, line 54: maybe there's a better way to do this? abusing dynamic type and effectively adding one if does not feel good

Qazalin: oh I can revert if you are strongly against it. I think the previous implementation was less readable with that __setitem__ and the normal path duplicating buf.arg

chenyuxyz: probably fine, i don't like that `__setitem__` either, will update if i think of something better

Qazalin: I actually like this much better https://github.com/tinygrad/tinygrad/pull/5687
thanks for the feedback!

PR #5687, opened 2024-07-24
bring unbind back in Varaible const

Qazalin:
https://github.com/tinygrad/tinygrad/pull/5681/files#r1690087335


Qazalin:
@chenyuxyz do you think we can skip this https://github.com/tinygrad/tinygrad/actions/runs/10080557458/job/27870570213?pr=5687 - I think it's a bug, It's blocked me a couple times.

chenyuxyz:
i think it's fine to lower the fuzz upper bound to a smaller number each time we see a failure. @hikettei might know better

PR #5694, opened 2024-07-24
sort UOp src if commutative

chenyuxyz:
redo #5353


chenyuxyz:
quite slower, likely broke something else

chenyuxyz:
this would not work for a while because it sorts the expanded UOps pre UPat.compile

PR #5695, opened 2024-07-24
start triton backend

geohot:


geohot:
I run into the same problem I always do with triton. It's not good about handling axis reshapes, even NOOPs completely nerf performance.

PR #5701, opened 2024-07-25
Multiple gradients for force-matching problems

scdunne:
In order to implement neural network potentials in tinygrad, we need to have a consistent way to backpropagate multiple times. This test shows a minimal example with a PyTorch implementation of this calculation, and my guess at how we would write the equivalent operation in tinygrad. The attached image shows the computations we would need in a graph. With this current implementation, the weight update is incorrect. Please let me know if anyone has an idea of how I can modify the intended operation, or if this is really an inherent issue with the framework. Thank you!

Screenshot 2024-07-23 131856

### DIFF
--- a/test/test_multiple_gradient.py
+++ b/test/test_multiple_gradient.py
@@ -5,11 +5,19 @@ import torch
 
 
 class TestMultipleGradient(unittest.TestCase):
+    '''
+    Set up an idealized system where the weights are initialized to the correct value
+    of an all-ones tensor. Check that the gradient is zero.
+    '''
+    def total_loss(true_energy, model_energy, true_force, model_force):
+        # Set up the energy-matching and force-matching objectives
+        fm_loss = ((true_force - model_force)**2).mean() 
+        em_loss = ((true_energy - model_energy)**2).mean()
+        loss = fm_loss + em_loss
+        # Predicted energy is exact, so d(loss)/d(weights) == 0
+        return loss
+
     def test_pytorch_force_matching_loss_gradient(self):
-        '''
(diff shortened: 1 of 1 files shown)


geohot:
Hmm, interesting, thanks for the test. backward deletes the graph, so I'm not sure calling it twice is okay. But it definitely shouldn't silently return the wrong answer.

scdunne:
Ahh, I see. I think I missed that backward deletes the graph, which would explain a lot. Do you think there's a way to get around this, or would something akin to torch.autograd.grad(..., retain_graph=True) need to be written?

davidgonmar:
The main issue is that Tinygrad does not form a Tensor level computational graph for the derivatives. As can be seen in the image, the Force is a gradient wrt positions. Then, the loss takes as parameter the Force. In this case, the loss will backpropgate the Energy branch correctly, but it will not backpropagate the Force branch at all, since it does not know 'where the Force came from wrt the inputs (positions). The solution would be to also keep a compute graph for gradients (the Function class should take as arguments Tensors for the .backward instead of LazyBuffers). This change would also enable higher order derivatives, and is what PyTorch does (the backward computation is done at Tensor level).

geohot:
Can you write a more minimal test of exactly this issue, and mark it with unitttest.expectedFailure? Mark this one with expectedFailure too

davidgonmar:
> Can you write a more minimal test of exactly this issue, and mark it with unitttest.expectedFailure? Mark this one with expectedFailure too

Will do. I also think it's straightforward to fix it unless I'm missing something, if it's fine I will give it a chance tomorrow.

PR #5717, opened 2024-07-25
skip hashing unrealized children

Qazalin:
image

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -198,8 +198,8 @@ def _recursive_group(tr:LazyBuffer, st:ShapeTracker, r:LazyBuffer, children:Defa
   for tr_next in children[tr]:
     # max one reduceop per kernel
     # can only fuse contiguous
-    if tr_next.op in ReduceOps or len(st_childs:=dedup(s for s in tr_next.srcs if s.base == tr)) > 1:
-      group[tr_next] = False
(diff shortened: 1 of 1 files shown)


chenyuxyz:
cool! can you add some comments about what group/ungroup is supposed to track and a regression test?

Qazalin:
oh if you remove ungroup everything in test_schedule breaks.
I think for the timing regression we need some sort of process replay for performance.

chenyuxyz:
some proxy tests that shows the root cause of this issue would be nice, i don't quite understand where the slow down was coming from (like is it kernel orders, or some redundant src or something else?)

Qazalin:
It's because I made 3 O(n) operations:
https://github.com/tinygrad/tinygrad/pull/5717/files#diff-120c8f826d67db123ecd7a1de1d0af40b894be9c3c3a27655819b4cd2689c581L255-L257
into 1 O(1) and 1 O(n) on a smaller n. Both contributed to this.

Qazalin:
could isolate this down to the hashing for a ReduceOp with a large graph.

chenyuxyz:
interesting, so we still have python scheduling in resnet steps

Qazalin:
That extra hash isn't even required with the new abstractions in _get_isolated_children because we only care about a realized child.
An unrealized ReduceOp child doesn't need to get hashed directly.

Qazalin:
closing for staleness, I'll revisit this once indexing is complete.

PR #5729, opened 2024-07-26
PTX render vec CONST

greg-niemeyer:
master currently creates duplicate `mov` instructions for vectorized CONSTs. These duplicate instructions create process_replay diffs when using VECTORIZE+scalar CONST instead of an implicitly vectorized CONST UOp. This PR makes the generated PTX code consistent between the two scenarios. The process_replay diff created by this change looks like:
image

This change is motivated by #5322, whose aim is to make all vectorization of CONSTs explicit and keep CONST UOps scalar.


Qazalin:
I think the lambda is unreadable. In scalars dtype.count is 1, does that make the implementation simpler?

greg-niemeyer:
> I think the lambda is unreadable. In scalars dtype.count is 1, does that make the implementation simpler?

I changed the lambda to something closer to the original code. Unfortunately, the higher-order UOps expect r[u] to be a list if src is vectorized, or a str if src is a scalar. I plan to simplify the expression further in #5322 once CONSTs are guaranteed to be scalar, if I can eventually get that merged.

geohot:
Hmm this probably shouldn't be dealt with in the renderer, but rather VECTORIZE the CONST

Qazalin:
I think this PTX is cleaner, the VECTORIZE refactor will also auto dedup this in UOps. https://github.com/tinygrad/tinygrad/pull/5322/files#diff-b4326fe5564554a8bdc311a3a0eb1cc070b9ce10a94c266303b41b580c26d68cR173

PR #5741, opened 2024-07-26
UOp._min_max for MUL and MOD

chenyuxyz:


chenyuxyz:
read through the process replay diff and it removed 10%~20% of early alu

PR #5742, opened 2024-07-26
NOp field name [run_process_replay]

DKormann:


geohot:
Always be in the habit of "[run_process_replay]" in the PR title on refactors like this

DKormann:
run process replay is running per default https://github.com/tinygrad/tinygrad/actions/runs/10118219467/job/27984684885?pr=5742

chenyuxyz:
it runs by default but does not assert if all are the same, and refactoring like this should keep kernels the same

DKormann:
ahh thanks

PR #5755, opened 2024-07-27
remove NOp.var() with _nop constructor [run_process_replay]

DKormann:


DKormann:
with _nop constructor we can
- [x] remove .var() 
- [ ] remove NOp.cvar NOp.const NOp.name ...
- [ ] completely remove Upat from
- [ ] fold class NOp back into UOp (if this is desirable)
- [ ] unify UPat.__init__ with UOp.compile()

DKormann:
will continue on monday

geohot:
I like this less than `NOp.var`, you have to import two things

DKormann:
okay but will be able to remove UPat from import in exchange

geohot:
I'm pretty happy with where it is now. If there's some simple syntax to use less `UPat` I'm open to it, but this is a big change

PR #5758, opened 2024-07-27
[PTX] Cleanup wmma casting

SzymonOzog:


tinygrad/renderer/assembly.py
```diff
@@ -120,11 +120,11 @@ def const(x:ConstType, dtype:DType, mov=False):
         return out
       return self.render_const(x, dtype)
 
-    def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
+    def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False, binary_reg=False):
```
geohot, line 123: Is this better expressed as a type?

SzymonOzog: Tried adding bin32 as a new dtype but didn't manage to get all tests to pass, maybe doing something wrong but I think it would require some ugly hacks to get working

geohot:
Closing, feel free to reopen with a cleaner way to do the binary_reg thing. Is a uint32 a binary reg?

PR #5759, opened 2024-07-27
nv/cuda compilers touchup

nimlgen:


geohot:
`check` is needed by `ops_cuda`

PR #5760, opened 2024-07-27
more unmatching vectorize/gep asserts [run_process_replay]

Qazalin:


tinygrad/codegen/uopgraph.py
```diff
@@ -150,10 +150,7 @@ def index_collapse(idx,rng,buf,add,mul,ld,reduce):
    lambda x: UOp(x.op, dtypes.int32, x.src, x.arg)),
   # VECTORIZE/GEP
   (NOp(UOps.GEP, src=(NOp(UOps.VECTORIZE, name="cast"),), name="gep"), lambda gep, cast: cast.src[gep.arg]),
-  *[(NOp(UOps.VECTORIZE, dtypes.float.vec(i), tuple(NOp(UOps.GEP, dtypes.float,
-                         src=(NOp.var('x', dtype=dtypes.float.vec(i)),), arg=j) for j in range(i))), lambda x: x) for i in [2, 4, 8, 16]],
-  *[(NOp(UOps.VECTORIZE, dtypes.half.vec(i), tuple(NOp(UOps.GEP, dtypes.half,
-                         src=(NOp.var('x', dtype=dtypes.half.vec(i)),), arg=j) for j in range(i))), lambda x: x) for i in [2, 4, 8, 16]],
+  *[(NOp(UOps.VECTORIZE, src=tuple(NOp(UOps.GEP, src=(NOp.var('x'),), arg=j) for j in range(i))), lambda x: x) for i in [2, 4, 8, 16]],
```
geohot, line 153: This is wrong I think.

What if you have a float4 and extract 2 elements from it (into a float2)? The type of the VECTORIZE and 'x' need to match. Can you add a test for this?

PR #5771, opened 2024-07-28
feat: allow passing gradient to .backward() to compute vjp

davidgonmar:
This allows the user to pass a `gradient` parameter to Tensor.backward. The API is similar to the one in PyTorch: if the tensor is not a scalar, then gradient must be passed. If it is scalar and it is not passed, it is assumed to be 1.0.
This means that backward can generalize to vector-jacobian products (can be seen as a 'generalization' of the gradient).

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -727,7 +727,7 @@ class Tensor:
   def backward(self, gradient:Optional[Tensor]=None) -> Tensor:
     """
     Propagates the gradient of a tensor backwards through the computation graph.
-    Must be called on a scalar tensor or the `gradient` argument must be provided.
+    If the 'gradient' argument is not provided, the tensor must be a scalar, and the gradient is implicitly set to 1.0.
 
     ```python exec="true" source="above" session="tensor" result="python"
     t = Tensor([1.0, 2.0, 3.0, 4.0], requires_grad=True)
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -724,22 +724,22 @@ def _walk(node, visited):
...
     ```
     """
-    assert self.shape == tuple(), f"backward can only be called for scalar tensors, but it has shape {self.shape})"
-
+    assert self.shape == tuple() or gradient is not None, "backward must be called on a scalar tensor or with the gradient argument"
+    assert gradient is None or gradient.shape == self.shape, f"gradient shape must match tensor shape, {gradient.shape!r} != {self.shape!r}"
     # fill in the first grad with one. don't use Tensor.ones because we don't need contiguous
     # this is "implicit gradient creation"
-    self.grad = Tensor(1.0, dtype=self.dtype, device=self.device, requires_grad=False)
+    self.grad = Tensor(1.0, dtype=self.dtype, device=self.device, requires_grad=False) if gradient is None else gradient
```
geohot, line 742: Create the implicit gradient before the asserts

davidgonmar: you mean something like
```python
self.grad = Tensor(1.0, dtype=self.dtype, device=self.device, requires_grad=False)
assert self.shape == tuple() or gradient is not None, "backward must be called on a scalar tensor or with the gradient argument"
assert gradient is None or gradient.shape == self.shape, f"gradient shape must match tensor shape, {gradient.shape!r} != {self.shape!r}"
self.grad = self.grad if gradient is None else gradient
```
?

davidgonmar: but this could be incorrect. 'gradient' (which is really a vector jacobian product) must be of the same shape as the tensor (self.shape == self.grad.shape) always. So creating an 'implicit' (scalar with value 1) grad (vector-jacobian product) for a tensor of different shape is incorrect. Did you mean something else?

geohot: No. The same code.

```
if gradient is None:
  assert self.shape == tuple()
  gradient = ...```

davidgonmar: > No. The same function, don't change self.grad
> 
> ```
> if gradient is None:
>   assert self.shape == tuple()
>   gradient = ...```
> ```

Corrected in the last commit

PR #5772, opened 2024-07-28
add failing test for second order derivatives

davidgonmar:
As commented in https://github.com/tinygrad/tinygrad/pull/5701, the 'minimal' failing test case is a second order derivative. The main issue is that tinygrad does not keep computational graph for gradients, so gradients cannot be 'backpropagated' again. This means that, for Tinygrad, the gradient of W.grad wrt inputs(for example) is 'None'. Common uses for this includes directly computing second order derivatives or even what is mentioned in https://github.com/tinygrad/tinygrad/pull/5701.

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -59,45 +59,32 @@ class TestTinygrad(unittest.TestCase):
   @unittest.expectedFailure
   def test_second_order_backward_pass(self):
     def test_pytorch():
-      x = torch.tensor(x_init, requires_grad=True)
-      W = torch.tensor(W_init, requires_grad=True)
-      m = torch.tensor(m_init)
-      # forward pass
-      out = x.matmul(W).relu()
-      out = torch.nn.functional.log_softmax(out, dim=1)
-      out = out.mul(m).add(m).sum()
+      x = torch.tensor(x_init)
+      m = torch.tensor(m_init, requires_grad=True)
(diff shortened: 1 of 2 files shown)


geohot:
Merge conflict. Is it easy to add support for retain_graph?

davidgonmar:
> Merge conflict. Is it easy to add support for retain_graph? Will that fix it?

Nope. That will not fix it (it is necessary to add support for retain_graph (and easy to support it at first glance) to fix it, but it wont). In order to fix it, it is necessary to somehow compute gradients (do the backward pass) on the Tensor level. Why? Because in order to backpropagate gradients, you need to know where gradients came from (same way as when you do regular backpropagation, the graph tells you 'how to reach the inputs from the output of the graph', you need to know how to reach the inputs from the gradients. Since the Function class works at a LazyBuffer level for the backward, it does not know how a gradient was constructed (the gradient has no 'children' nor tensor operation that 'made' it. I think simply changing the autograd Function to use Tensor instead of LazyBuffer (and some minor changes) should enable support for it. It should also not take more lines and speed for existing scripts should be similar (all of this is speculation).
I am planning on writing a detailed issue that explains how it should work and this solution with more detail, and if it looks fine, try to get it working.

geohot:
I think this example can be more minimal.

PR #5777, opened 2024-07-28
remove float division from idiv in python_alu

samm393:
Currently python_alu uses int(x/y) for integer division. This does not work for very large integers as the float casting is not precise enough. This pr replaces that with x//y, along with the required logic to make it truncate towards 0.

```python
print(Tensor([18446744073709551612,], dtype=dtypes.uint64, device='PYTHON').div(Tensor([1,], dtype=dtypes.uint64, device='PYTHON'), upcast=False).numpy())
```

```sh
[0] #main
[18446744073709551612] #this pr


chenyuxyz:
i think it logic can be simpler, also can you add a test for it?

samm393:
Now with simpler logic. Test added for the failing case. 
I can't find any tests for tensor integer division though - test_div_int() only seems to test float division with integer input.
Am I missing something or do randomized tests for integer division need to be added?

chenyuxyz:
yea something like `test_idiv` next to `test_div_int` would be nice, does not need to be this pr

PR #5779, opened 2024-07-28
[WIP] Shape change bitcast

samm393:
Adds support for shape changing bitcasts using MetaOps.View on all supported devices, and an alternative tensor level function _view_dtype() is used on devices where buffer view is not available.

Conditions on strides are enforced as in torch.Tensor.View(dtype). This is not strictly necessary for the _view_dtype() operation, but have been kept as I assume the use case is to emulate a view bitcast.

Tests are added for both contiguous and non-contiguous tensors, as well as tests for the exceptions that should be raised.

PR #5777  is a prereq

Working on support for multitensor

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -183,11 +183,11 @@ jobs:
     - name: Run LLaMA 7B on 6 GPUs
       run: NV=1 python3 examples/llama.py --gen 1 --size 7B --shard 6 --prompt "Hello." --count 10 --temperature 0  --timing | tee llama_six_gpu.txt
     - name: Run LLaMA-3 8B BEAM
-      run: NV=1 JITBEAM=2 IGNORE_BEAM_CACHE=1 python3 examples/llama3.py --model weights/LLaMA-3/8B-SF-DPO/ --benchmark | tee llama3_beam.txt
+      run: NV=1 JITBEAM=2 IGNORE_BEAM_CACHE=1 python3 examples/llama3.py --model weights/LLaMA-3/8B-SF-DPO/ --benchmark --temperature 0 | tee llama3_beam.txt
     - name: Run LLaMA-3 8B on 4 GPUs
-      run: NV=1 python3 examples/llama3.py --shard 4 --model weights/LLaMA-3/8B-SF-DPO/ --benchmark | tee llama3_four_gpu.txt
+      run: NV=1 python3 examples/llama3.py --shard 4 --model weights/LLaMA-3/8B-SF-DPO/ --benchmark --temperature 0 | tee llama3_four_gpu.txt
     - name: Run LLaMA-3 8B on 6 GPUs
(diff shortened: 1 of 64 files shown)


geohot:
Gave you the bounty lock, but I think this code can be made a lot simpler

geohot:
Unlocking bounty. When I see code like
```python
if dtype.itemsize > self.dtype.itemsize:
    shift_factors = (Tensor.arange(n, dtype=new_int, device=self.device)*8*self.dtype.itemsize).exp2().cast(new_int)
    a = self.bitcast(old_int).cast(new_int).reshape(new_shape + (-1,))
    return (a*shift_factors).sum(-1, acc_dtype=new_int).bitcast(dtype)
  shift_factors = (Tensor.arange(n, dtype=old_int, device=self.device)*8*dtype.itemsize).exp2().cast(old_int)
  a = self.bitcast(old_int).unsqueeze(-1).expand(self.shape + (n,))
  return a.div(shift_factors, upcast=False).cast(new_int).reshape(new_shape).bitcast(dtype)
```
it's not close

samm393:
> Unlocking bounty. When I see code like
> 
> ```python
> if dtype.itemsize > self.dtype.itemsize:
>   shift_factors = (Tensor.arange(n, dtype=new_int, device=self.device)*8*self.dtype.itemsize).exp2().cast(new_int)
>   a = self.bitcast(old_int).cast(new_int).reshape(new_shape + (-1,))
>   return (a*shift_factors).sum(-1, acc_dtype=new_int).bitcast(dtype)
> shift_factors = (Tensor.arange(n, dtype=old_int, device=self.device)*8*dtype.itemsize).exp2().cast(old_int)
> a = self.bitcast(old_int).unsqueeze(-1).expand(self.shape + (n,))
> return a.div(shift_factors, upcast=False).cast(new_int).reshape(new_shape).bitcast(dtype)
> ```
> 
> it's not close

Is code like
```python
  def _view_dtype(self, dtype:DType) -> Tensor:
    # https://pytorch.org/docs/stable/generated/torch.Tensor.view.html
    # only called when we cannot use MetaOps.VIEW
    ns, os = dtype.itemsize, self.dtype.itemsize
    new_int, old_int = dtypes.fields()[f"uint{8*ns}"], dtypes.fields()[f"uint{8*os}"]
    new_shape = self.shape[:-1] + (self.shape[-1]*os//ns,)
    shift_factors = Tensor.arange(8*max(ns, os), step=8*min(ns, os), device=self.device).exp2().cast(new_int if ns > os else old_int)
    ret = self.bitcast(old_int).unsqueeze(-1)
    if ns < os: return ret.div(shift_factors, upcast=False).cast(new_int).reshape(new_shape).bitcast(dtype)
    ret = ret.cast(new_int).reshape(new_shape + (-1,))
    return (ret*shift_factors).sum(-1, acc_dtype=new_int).bitcast(dtype)
```
any closer? Or is my whole approach is wrong?

geohot:
I have no idea why there's `exp2`

PR #5788, opened 2024-07-29
rename upcast_axis into plural [run_process_replay]

ignaciosica:


chenyuxyz:
see https://github.com/tinygrad/tinygrad/blob/2c94316bd2ed9facaef069ec199f7151a49b1bae/README.md?plain=1#L178 to run reprocess replay for refactors

Issue #5774, opened 2024-07-28
better error message for Tensor creation methods with negative dims


chenyuxyz:
currently `Tensor.full((-2,-2), 2)` and `Tensor.full((-2,-2), 2).realize()` both have no error.

torch error: `RuntimeError: Trying to create tensor with negative dimension -2: [-2, -2]`
numpy error: `ValueError: negative dimensions are not allowed`

fix this generically for all Tensor creation methods

PR #5790, opened 2024-07-29
Add check for negative dimension in view

ignaciosica:
fixes: #5774

Is it okay to check at a view level?

~~Regression tests pending~~

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -421,7 +421,7 @@ class Tensor:
     counts1, counts2 = (x & 0xffffffff).cast(dtypes.uint32), ((x >> 32) & 0xffffffff).cast(dtypes.uint32)
 
     out = counts1.cat(counts2).rshift(8).cast(dtypes.float32).div(2 ** 24)[:num]
-    if not all(s >= 0 for s in shape): raise RuntimeError(f"Trying to create view with negative dimension: {shape=}")
+    if not all(s >= 0 for s in shape): raise RuntimeError(f"Trying to create Tensor with negative dimension: {shape=}")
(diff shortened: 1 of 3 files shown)


chenyuxyz:
i think it should check in Tensor level, if the arg to a function is invalid, we want to check before passing the arg to tinygrad internal

ignaciosica:
Sure, moved check to Tensor level and leaved an assert in view create

tinygrad/tensor.py
```diff
@@ -441,6 +443,7 @@ def full(shape:Tuple[sint, ...], fill_value:ConstType, **kwargs):
     print(Tensor.full((2, 3), False).numpy())
     ```
     """
+    if not all(s >= 0 for s in argfix(shape)): raise RuntimeError(f"Trying to create view with negative dimension: {shape=}")
```
chenyuxyz, line 446: argfix is called twice, also don't just copy paste error messages. the message does not make sense in Tensor

geohot:
This shouldn't be in there three times. I'm fine with having the assert internally only, and I like the tests.

tinygrad/tensor.py
ignaciosica: Also, python won't let me use walrus operator in that context:

```
if not all(s >= 0 for s in (new_shape := argfix(shape))): raise RuntimeError(f"Trying to create Tensor with negative dimension: {shape=}")
```
highlights `:=` as having wrong

ignaciosica:
> This shouldn't be in there three times. I'm fine with having the assert internally only, and I like the tests. They pass with only the internal assert?

If the assert is changed for RuntimeError they do, or by changing tests accordingly

geohot:
`RuntimeError` is preferred if it can happen if the user does something wrong, so in this case.

PR #5792, opened 2024-07-29
retrieve defined opencl error codes

tyoc213:


geohot:
Do this at init. Also, add tests

tyoc213:
Does this work @geohot ?

tinygrad/runtime/ops_gpu.py
```diff
@@ -9,8 +9,16 @@
 # see test/external/external_osx_profiling.py to determine this ratio. it's in like GPU clocks or something
 OSX_TIMING_RATIO = (125/3) if OSX else 1.0
 
+@functools.lru_cache(maxsize=1)
+def _get_errors():
+  ed = {getattr(cl, k): k for k in filter(lambda e: e.startswith("CL_"), dir(cl)) if getattr(cl, k) <=0}
+  print(ed)
+  return ed
+_ = _get_errors()
```
nimlgen, line 17: something like `cl_errors = {getattr(cl, k): k for k in filter(lambda e: e.startswith("CL_"), dir(cl)) if getattr(cl, k) <= 0}` is enough.

tinygrad/runtime/ops_gpu.py
```diff
@@ -9,8 +9,16 @@
...
 def check(status):
-  if status != 0: raise RuntimeError(f"OpenCL Error {status}")
+  def report(s):
+    raise RuntimeError(f"OpenCL Error {s}: {_get_errors().get(s, 'Unknown error')}")
+  if status != 0: report(status)
```
nimlgen, line 21: simpler. just `raise RuntimeError`, why `def report` then calling it?

tinygrad/runtime/ops_gpu.py
tyoc213: The print is extra debug I missed.

For the call of get errors, that is how I interpreted @geohot comment "do at init", as if I moved the method to a any of the `CL*` classes it will become a messy PR... so the only "init" left is the one done while loading the file so that the dictionary can be created when loaded.

tinygrad/runtime/ops_gpu.py
tyoc213: Some code of that method was striped to make it cacheable, will throw the error directly.

tinygrad/runtime/ops_gpu.py
```diff
@@ -9,8 +9,11 @@
 # see test/external/external_osx_profiling.py to determine this ratio. it's in like GPU clocks or something
 OSX_TIMING_RATIO = (125/3) if OSX else 1.0
 
+@functools.lru_cache(maxsize=1)
+def _get_errors(): return {getattr(cl, k): k for k in filter(lambda e: e.startswith("CL_"), dir(cl)) if getattr(cl, k) <=0}
+_get_errors()
```
wozeparrot, line 14: why not just `error_map = {getattr(cl, k): k for k in filter(lambda e: e.startswith("CL_"), dir(cl)) if getattr(cl, k) <= 0}`?

nimlgen: @tyoc213 it's not resolved. I wrote you the same, no need to write anything that complex (with caches, functions, etc).

Tests failed. Also, I don't think that new file is the best place to put the test

tyoc213: Got it now.

Fixed the test.

Moved the test to top dir and renamed it as it makes clear
 it is specific to ocl.

tyoc213:
I'm trying to solve the gpu test error

PR #5793, opened 2024-07-29
add `Tensor.__floordiv__`

chenyuxyz:


chenyuxyz:
not correct for float input

PR #5797, opened 2024-07-29
install mkdocs?

nimlgen:


chenyuxyz:
fixed in https://github.com/tinygrad/tinygrad/pull/5798

PR #5808, opened 2024-07-30
UOp bound for div negative number

chenyuxyz:
assumes rounding towards zero like cstyles


chenyuxyz:
add somehow is not folded, but cleaner kernel code

image

PR #5812, opened 2024-07-30
UOps.IF with rewrite rules

Qazalin:
This is [more correct](https://github.com/tinygrad/tinygrad/actions/runs/10166804663/job/28117790767?pr=5812#step:24:94), no material performance diff:
master:
```
~/code/tinygrad> PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  58.11 ms
***** model schedule in  20.10 ms
***** model uops in 118.36 ms
***** model linearize in 1565.30 ms
```
this branch:
```
~/code/tinygrad> PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  58.09 ms
***** model schedule in  19.80 ms
***** model uops in 119.52 ms
***** model linearize in 1547.35 ms
```

### DIFF
--- a/test/test_linearizer_failures.py
+++ b/test/test_linearizer_failures.py
@@ -353,7 +353,6 @@ class TestLinearizerFailures(unittest.TestCase):
     helper_test_lin(Kernel(ast), opts=opts, failed_platforms=["AMD", "HIP"])
 
   # llama3 8B failure with BEAM=2 https://github.com/tinygrad/tinygrad/actions/runs/10150118124/job/28066519425#step:14:1, these don't compile
-  @unittest.expectedFailure
   @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test needs local")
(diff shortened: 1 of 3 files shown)


chenyuxyz:
saw this in metal diff
image

const True or false can be removed, you can use `vmax.arg == False` or `vmin.arg == True` to fold gate

Qazalin:
Yea we have a UPat for that, it didn't do it before because I rewrote IF way too early. Now should be good.
https://github.com/tinygrad/tinygrad/blob/master/tinygrad/codegen/uopgraph.py#L299-L306

chenyuxyz:
don't think that match though, the process replay diff are from this pr. i think that only matches if the whole gate clause is const

it's probably a good idea to fold `or False` and `and True` anyway even outside of load store

Qazalin:
ahhh, the expander is creating ALUs outside of the graph that never get rewritten by `constant_folder` again.

PR #5813, opened 2024-07-30
UNet3D KiTS19 preprocessing dir cleanups

francislata:
This PR cleans up how the preprocessed directory is passed in. Before, the root preprocessing directory is passed in and the function themselves determines if it is a training or validation dataset. Now, two preprocessed constants are now defined: `TRAIN_PREPROCESSED_DIR` & `VAL_PREPROCESSED_DIR` and callers of these functions will have to pass in one of these constants.

### DIFF
--- a/extra/datasets/kits19.py
+++ b/extra/datasets/kits19.py
@@ -12,7 +12,8 @@ from tinygrad.tensor import Tensor
 from tinygrad.helpers import fetch
 
 BASEDIR = Path(__file__).parent / "kits19" / "data"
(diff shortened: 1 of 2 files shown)


chenyuxyz:
can you confirm the model_eval still runs with this change?

francislata:
> can you confirm the model_eval still runs with this change?
Screenshot 2024-07-30 at 12 29 50

I can confirm `model_eval.py` for UNet3D still runs with this change.

PR #5814, opened 2024-07-30
fix opencl_ioctl on comma

vpachkov:
- remove unused code
- add CP_REG_TO_MEM opcode
- fixed parse_cmd_buf for more than 1 command object by correcting an offset
- fixed memory mappings for cases when memory was allocated with KGSL_MEMFLAGS_USE_CPU_MAP.
KGSL_MEMFLAGS_USE_CPU_MAP: If set on call and return, the returned GPU address will be 0. Calling mmap() will set the GPU address. So there are no IOCTL_KGSL_GPUOBJ_INFO ioctls for that type of memory and it resulted to crash right after get_mem.

### DIFF
--- a/extra/qcom_gpu_driver/opencl_ioctl.py
+++ b/extra/qcom_gpu_driver/opencl_ioctl.py
@@ -1,7 +1,7 @@
 import ctypes, ctypes.util, struct, fcntl, re
 from hexdump import hexdump
-from tinygrad.runtime.ops_gpu import CLDevice, CLAllocator
 import pathlib, sys
+from tinygrad.helpers import to_mv
 sys.path.append(pathlib.Path(__file__).parent.parent.parent.as_posix())
 
 ops = {}
@@ -44,6 +44,8 @@ def get_mem(addr, vlen):
   for k,v in mmaped.items():
     if k <= addr and addr < k+len(v):
       return v[addr-k:addr-k+vlen]
+  # hope it was mmapped by someone else
+  return bytes(to_mv(addr, vlen))
 
 def hprint(vals):
   ret = []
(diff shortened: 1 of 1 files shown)


extra/qcom_gpu_driver/opencl_ioctl.py
```diff
@@ -100,7 +106,7 @@ def ioctl(fd, request, argp):
       mmaped[s.gpuaddr] = mmap.mmap(fd, s.size, offset=s.id*0x1000)
     if name == "IOCTL_KGSL_GPU_COMMAND":
       for i in range(s.numcmds):
-        cmd = get_struct(s.cmdlist+s.cmdsize*i, msm_kgsl.struct_kgsl_command_object)
+        cmd = get_struct(s.cmdlist+ctypes.sizeof(msm_kgsl.struct_kgsl_command_object)*i, msm_kgsl.struct_kgsl_command_object)
```
nimlgen, line 109: hmm, they are not equal? do we know why?

vpachkov: they are not... both on my device and on comma. s.cmdsize shows 64, but the real offset should be the size of the struct = 32. without this, we basically skip to garbage and parse some garbage instead of real command objects

nimlgen: okay, that's strange. in the kernel they also use `ptr += sizeof(obj);` and not cmd_size to move it (`kgsl_drawobj_cmd_add_cmdlist`). should `s.objsize` also be changed?

vpachkov: yeah, I know... well, just checked it, s.objsize is correct :) 32 just like it is supposed to be on both devices I have access to

PR #5828, opened 2024-07-31
UOp if min==max folds to CONST

chenyuxyz:


mrdaybird:
This slows down the pattern matcher a lot. Are we getting any significant speedup from this? We should analyze if this is beneficial or not.
Before:
```
git:(4e89d455) PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  43.46 ms
***** model schedule in  15.36 ms
***** model uops in 139.69 ms
***** model linearize in 1646.36 ms
```
After:
```
git:(c3da458b) PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model forward in  43.78 ms
***** model schedule in  15.28 ms
***** model uops in 124.78 ms
***** model linearize in 1904.89 ms
```

PR #5839, opened 2024-07-31
indexing fusion final

Qazalin:
single arange kernels work beautifully,
- [x] kernels with multiple (distinct) aranges should pad
- [ ] push through different views of the same LazyBuffer


Qazalin:
Tomorrow:
- TestIndexing.test_advancedindex and add more like it in test_schedule
- Merge the inline const fix #5840 for argmax
- Test fused arange children TestOps.test_interpolate_bilinear

PR #5840, opened 2024-07-31
Fix compile error for max with inline const

ignaciosica:
Fixes #5838

I think that for correctness this should be generalized for other Ops as well as the same problem may arise in those cases to, but if so many symbolic tests fail as it compares the generated output with fixed ones

Or by directly casting every const, but the same problem arises


Qazalin:
can you add tests for this? the best file for this is test/test_uops.py.
I wanna see tests for every ALU in code_for_uop, max isn't the only one with this problem.

ignaciosica:
Sure, adding tests today

ignaciosica:
Sorry I couldn't finish it yet, I going to work on it this weekend

ignaciosica:
Sorry I don't have the tests yet, they weren't as straightforward as I expected as const ops are folded before the kernel renders, still working on them though. Maybe there is a straightforward way but I am not seeing it right now

Qazalin:
added `test_renderer_failures.py` - should be enough infra for this diff.

ignaciosica:
Ok, the testing for unary ops was rather tricky as it was constantly being fold before rendering. I could not manage to come up with failing case, sorry

Qazalin:
you don't need to test unary we fold it anyways.

ignaciosica:
> can you add tests for this? the best file for this is test/test_uops.py. I wanna see tests for every ALU in code_for_uop, max isn't the only one with this problem.

MAX is the only BinaryOp in METAL `code_for_uop`, or you meant default `code_for_uop`
image

Do you see ops like `BinaryOps.MUL: lambda a,b,dtype: f"({a}*{b})"` failing because of this issue?

Qazalin:
yea max should be the only one

ignaciosica:
> added `test_renderer_failures.py` - should be enough infra for this diff.

So is this test enough?

geohot:
Tests failing

ignaciosica:
> Tests failing

fixed, no longer expects compiler error

tinygrad/renderer/cstyle.py
```diff
@@ -137,6 +137,7 @@ def ssa(prefix:str, u:Optional[UOp]=None):
         elif uop is UOps.ALU:
           # remove parens if ALU types are the same. TODO: can do more here
           if args in {BinaryOps.ADD,BinaryOps.MUL,BinaryOps.XOR}: operands = [strip_parens(r[v]) if v.arg == args else r[v]for v in src]
+          elif args is BinaryOps.MAX: operands = [self.render_cast(r[v],v.dtype) if v.op is UOps.CONST and v.dtype is not None else r[v] for v in src]
```
Qazalin, line 140: you can cast the dtype here with cast(DType
we assert CONST always has a dtype.

ignaciosica: Sure, changed

PR #5844, opened 2024-07-31
fix UOp ALU bound

chenyuxyz:
root cause of resnet bug, the ALU bound is only correct for scalar, not vectorized


chenyuxyz:
fixed resnet training https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/fl1x3i0k/logs

PR #5849, opened 2024-07-31
UOp folding SumNode div and mod

chenyuxyz:


chenyuxyz:
NOTE: not correct because `(a+b+c)//d != (a//d+b//d+c//d)`

PR #5850, opened 2024-07-31
Working Boolean Index Example

reddyn12:
What are the thoughts on doing boolean indexing this way? If we can't access the ```__device__``` count var, can do (y>=0).sum()


geohot:
Of course you can do this manually, but this isn't tinygrad style.

PR #5873, opened 2024-08-02
nn.Embedding handles out of bounds error

eitanturok:
Currently, out of bounds indexing in `nn.Embedding` returns a tensor of zeros. In [discord](https://discord.com/channels/1068976834382925865/1069001075828469790/1268617507149971548), @geohot suggested we return `nan` when we index out of bounds. This PR does that.

Previously, the code
```python
from tinygrad import Tensor
from tinygrad.nn import Embedding

vocab_size, embd_size = 8, 2
layer = Embedding(vocab_size, embd_size)
idx = Tensor([3, 100])
out = layer(idx)

print(out.numpy())
```
would output
```python
[[[ 0.6379649  -0.24740033]
  [ 0.          0.        ]]]
```
Notice how the bottom row, which is indexing the `100`th element of the vocab, is all zeros even though it is out of bounds (the vocab is of size 8).

With the changes we've made, now the code snippet above outputs nans
```python
[[[ 0.6379649  -0.24740033]
  [nan nan]]]
```

I also added a test for this.


geohot:
Thinking about it more, zeros are fine if it has to add a where clause like this

PR #5876, opened 2024-08-02
Link to tinygrad-tensor-puzzles in docs

obadakhalili:
I implemented [Sasha's Tensor Puzzles](https://github.com/srush/Tensor-Puzzles) using tinygrad, and saw that it could be useful to new users of tensor libraries. Thus, I thought I could link my solutions in tinygrad docs. I'm unsure how desired this is though, but here is a PR ..


geohot:
I don't think solutions are the right thing to link, but rather perhaps a very clean fork of the puzzles themselves

PR #5879, opened 2024-08-02
Example for fast numpy() conversion

reddyn12:
Same shape of ```(32, 12087, 268)``` takes ```6 sec``` on master using ```tens.numpy()``` compared to ```0.46 sec``` here


geohot:
Sure, it can be faster. The trick is to do it in tinygrad in a clean maintainable way

PR #5881, opened 2024-08-02
New BERT dataloader

Eliulm:
This PR enhances the dataloader by adding the equivalent of `tf.data.experimental.parallel_interleave`, improving convergence to [2.3M - 2.8M steps](https://wandb.ai/wahl/MLPerf-BERT/runs/9qqq1oa4?nw=nwuserwahl)(matching reference) and speeding up the dataloader from 20ms to ~3ms. Note, we now batch all samples of one topic to one file, so deleting the diskcache and regenerating the data is necessary. :/

#### Testing Limitations:
Direct testing against the reference is not possible because:
1. The reference only sets the graph seed, not the operation seed for random transformations.
2. `sloppy=True` in parallel interleave causes nondeterministic behavior.

A test confirming that removing the `sloppy` flag and operation-based shuffle matches the reference was written but not pushed, as it required changes to both the reference code and our dataloader. Would this test be of interest still?

As far as MLPerf submission is concerned, the [rules](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#66-training-data-order) state: `The training and test data must be traversed in the same conceptual order as the reference implementation.` Although we are not matching the reference exactly down to the last sample, we are going through the same topics, in the same order, introducing the same degree of randomness.

### DIFF
--- a/examples/mlperf/dataloader.py
+++ b/examples/mlperf/dataloader.py
@@ -227,8 +227,9 @@ def batch_load_train_bert(BS:int):
   while True:
     batch = []
     for _ in range(BS):
-      random.shuffle(buffer)
-      batch.append(buffer.pop(0))
+      index = random.randint(0, 999)
+      batch.append(buffer[index])
+      buffer[index] = dataset.get()
     yield process_batch_bert(batch)
 
 # Reference: https://github.com/mlcommons/training/blob/1c8a098ae3e70962a4f7422c0b0bd35ae639e357/language_model/tensorflow/bert/run_pretraining.py, Line 416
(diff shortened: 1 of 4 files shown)


chenyuxyz:
not testing the exact orders match is fine, i will merge this

can you add scripts for BERT to
`https://github.com/tinygrad/tinygrad/tree/master/examples/mlperf/training_submission_v4.1/tinycorp/benchmarks`
so we can start to test end to end.

also add a README in benchmark folder to describe the latest steps to acquire datasets (similar to reference, and it's required by mlperf), we will run that and sync bert data to all tinyboxes

Eliulm:
> can you add scripts for BERT to `https://github.com/tinygrad/tinygrad/tree/master/examples/mlperf/training_submission_v4.1/tinycorp/benchmarks` so we can start to test end to end.
> 
> also add a README in benchmark folder to describe the latest steps to acquire datasets (similar to reference, and it's required by mlperf), we will run that and sync bert data to all tinyboxes

will do

PR #5896, opened 2024-08-04
UOp generic div folding

chenyuxyz:


chenyuxyz:
merging now, plan to combine this with UOp.divides later

chenyuxyz:
few % faster on red covs

PR #5915, opened 2024-08-05
Float4 support for CLANG

ignaciosica:


tinygrad/renderer/cstyle.py
```diff
@@ -180,15 +180,18 @@ def ssa(prefix:str, u:Optional[UOp]=None):
...
           r[u] = (r[src[0]] if from_ssa else f"{(r[src[0]])}") + \
-            (f"[{args}]" if src[0].dtype.count > (8 if self.device in {"CUDA", "NV"} else 4) else f".{'xyzwabcd'[args]}")
+            (f"[{args}]" if src[0].dtype.count > (8 if self.device in {"CUDA", "NV"} else 4) or self.device == 'CLANG' else f".{'xyzwabcd'[args]}")
         else: raise RuntimeError(f"failed to render {u}")
 
     # NOTE: this relies on bufs dict preserving order
     return self.render_kernel(name, kernel, list(bufs.values()), uops)
 
+def _make_clang_dtype(self, dtype):
+  return f"typedef {self.render_dtype(dtype.scalar())} {self.render_dtype(dtype)} __attribute__((vector_size({dtype.itemsize})));"
```
geohot, line 190: The pointer should probably be aligned

tinygrad/renderer/cstyle.py
```diff
@@ -41,7 +41,7 @@ def render_cast(self, x:str, var_dtype:DType, bitcast=False) -> str:
   def render_vectorize(self, x:List[str], var_dtype:DType) -> str:
     assert len(x) == var_dtype.count, f"cast is wrong size {len(x)} != {var_dtype.count}"
     assert self.float4 is not None, "vectorized cast is not supported on this platform"
-    return f"{self.float4.replace('float4', self.render_dtype(var_dtype))}({','.join(x)})"
+    return f"{self.float4.replace('float4', self.render_dtype(var_dtype))}" + (f"{{{','.join(x)}}}" if self.device == "CLANG" else f"({','.join(x)})")
```
geohot, line 44: Why is clang different here?

ignaciosica: The custom vector type support the following direct initializations that don't work if used with parenthesis.

```
typedef float float4 __attribute__((aligned(16),vector_size(16)));
float4 x = (float4){0.0f,0.0f,0.0f,0.0f};
float4 y = (float4){0.0f};
```

For parenthesis to work, an initialization function is required, much like make_float4, but with no other benefit than that.

```
float4 make_float4(float a, float b, float c, float d) {
    return (float4){a, b, c, d};
}
float4 m = make_float4(1.0f, 2.0f, 3.0f, 4.0f);
```

Other backends like CUDA and AMD do use the initialization function, that's the reason for the clause there

tinygrad/renderer/cstyle.py
ignaciosica: Sure, added it

PR #5922, opened 2024-08-05
whisper

DKormann:
- [x] fix jit error for long input
- [x] stream input output
- [x] timestamps
- [x] long batch inference
- [x] (speed up if necessary)


DKormann:
when the context gets full during recording i clear create new context with only text from current frame. I wanna make sure to only do this once to avoid loop.
this doubles effective token limit to around 400 tokens / 30 seconds of recording

DKormann:
will try couple refactors tomorrow and than long batch inference

DKormann:
working on speed optimization rn now at about 8.5x realtime on single GPU

chenyuxyz:
stale. we have a $200 bounty to make whisper work for tinygrad meeting recording if you are interested

PR #5942, opened 2024-08-06
Add docs

eitanturok:
Add docs for ndim, numel, element_size, nbytes, is_floating_point, size. @geohot mentioned this in discord.

1. Better name than "Convenience Ops"? Maybe "Size ops"?
2. Maybe move is_floating_point to casting ops?

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -3015,15 +3015,15 @@ class Tensor:
 
     ```python exec="true" source="above" session="tensor" result="python"
     t = Tensor([8, 9], dtype=dtypes.int8)
-    print(t.numel)
+    print(t.nbytes())
     ```
     ```python exec="true" source="above" session="tensor" result="python"
     t = Tensor([8, 9, 10, 11], dtype=dtypes.int8)
-    print(t.numel)
+    print(t.nbytes())
     ```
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -2952,12 +2952,117 @@ def sparse_categorical_crossentropy(self, Y:Tensor, ignore_index=-1, label_smoot
...
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([8, 9], dtype=dtypes.float64)
+    print(t.is_floating_point())
+    ```
+    """
+    return dtypes.is_float(self.dtype)
+
+  def size(self, dim=None) -> Union[sint, Tuple[sint, ...]]:
+    """
+    If `fim` is not specified, return the shape of the tensor. Otherwise return the length along dimension `dim`.
```
geohot, line 3052: typo

eitanturok: Fixed.

geohot:
"Convenience Ops" -> "Tensor Properties"?

eitanturok:
I changed the name to "Tensor Properties". But I think "Element Size Ops" or "Counting Ops" might be more descriptive.

If we keep the "Tensor Properties" name, lmk if I keep this documentation in the "ops" page of the docs.

PR #5943, opened 2024-08-06
Improve error message if clang is not installed

KucicM:
Current error message is `FileNotFoundError: [Errno 2] No such file or directory: 'clang'`

### DIFF
--- a/tinygrad/runtime/ops_clang.py
+++ b/tinygrad/runtime/ops_clang.py
@@ -12,7 +12,7 @@ class ClangCompiler(Compiler):
                                '-o', str(output_file.name)], input=src.encode('utf-8'))
         return pathlib.Path(output_file.name).read_bytes()
     except FileNotFoundError as e:
-      if shutil.which('clang'): raise Exception('clang not installed')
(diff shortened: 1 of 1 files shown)


geohot:
This is slow, I'd rather catch the error and provide a way to fix it

KucicM:
Added try/except around compile, but I'm not sure what do you mean by `provide a way to fix it.` 
Should message be something like `clang not installed. run: sudo apt install clang to fix it`?

geohot:
Looking at it, I'm not sure we need this complexity. File not found 'clang' is a pretty clear error

wmuron:
>  File not found 'clang' is a pretty clear error

I wanted to try out the library, since I am following this project for some time now, and still didn't have a chance to give it a try.

The first experience when trying to run a simplest hello world in tinygrad is getting a wall of exceptions, and vague message `FileNotFoundError [Errno 2] No such file or directory: 'clang'`. I can understand that it's self-explanatory, but from a user perspective it's a bit harsh starting point, IMO.

PR #5962, opened 2024-08-07
simplify ALU vmin==vmax fold [run_process_replay]

greg-niemeyer:
x.op is always UOps.ALU, so removed that check. From #5828, it looks like this is the intent of the rule.


chenyuxyz:
thanks!

PR #5967, opened 2024-08-07
Only add vectorization struct and initialization method if used

ignaciosica:


Qazalin:
yea this will work right now because we always have both scalar and vec dtype UOps in a single kernel.
But you should consider the prefix import in both paths (without duplicates, maybe `if any` could help)

ignaciosica:
Didn't have that case in mind

Do you like this better?

``` python
typed_uops = [uop for uop in uops if uop.dtype is not None]
if any(uop.dtype.scalar() == dtypes.half for uop in typed_uops): prefix += ["#include <cuda_fp16.h>"]
if any(uop.dtype.scalar() == dtypes.bfloat16 for uop in typed_uops): prefix += ["#include <cuda_bf16.h>"]

for dtype in dedup(uop.dtype for uop in typed_uops if uop.dtype.scalar() in (dtypes.half, dtypes.bfloat16)):
  if dtype.count == 1: prefix += [_make_cuda_dtype(self, dtype)]
```

Qazalin:
> Do you like this better?
structurally looks good but do you see the few redundancies?

ignaciosica:
I do, removed redundancies

PR #5975, opened 2024-08-08
Bitcast

reddyn12:


geohot:
Please don't submit incomplete PRs like this

reddyn12:
> Please don't submit incomplete PRs like this

I noticed that a view of a buffer isn't the same as the bit cast view of a buffer as it views the whole buffer. I thought it would be simpler to satisfy the assert for the offset attribute, than specify two different view types. Also because all the gpu/onnx tests passed with just returning the whole buffer in the offset function, I thought I didn't violate anything. The TODO statement was more for anyone who needed the offset func to work. I tried to implement it but got misaligned errors and when I realigned the offset, the right part of the buffer isn't read because I think the memory is planned beforehand. How do I go about this?

PR #5976, opened 2024-08-08
gated store rewrite to UOps.IF

ianpaul10:
PR for bounty: "gated store rewrite to UOps.IF, well tested. spec: https://github.com/tinygrad/tinygrad/pull/5652"

Changes include: 

- [x] Wrapping UOps.STOREs that were previously gated in-line with UOps.IFs. I'm not sure if the UOps.STORE's src should include the IF or not. I've included it for now. Unsure on best practices. 
- [x] Grouping UOps.STOREs that are reliant on the same gate / UOps.IF into a single if block. I'm not sure if there's a cleaner way to go about this. I can certainly reduce the line count in this area but also not sure if there's a smarter way to go about keeping them together.
- [x] cstyle renderer 
- [x] Enable previously failing tests
- [x] llvmir renderer
- [x] Fix llvmir & ptx tests

### DIFF
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -413,7 +413,7 @@ def no_vectorized_alu(alu):
 def create_gate(root:UOp) -> Optional[UOp]:
   @functools.lru_cache(None)
   def _gate_srcs(u:UOp, gate:UOp) -> UOp:
-    if u.op in {UOps.BARRIER, UOps.EXPAND}: return u
+    if u.op in { UOps.BARRIER, UOps.EXPAND }: return u
     if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER and gate.op is not UOps.IF:
       return UOp(u.op, u.dtype, u.src[:-1] + (UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)
     if u.op is UOps.STORE and len(u.src) == 4 and u.src[-1].op in {UOps.ALU, UOps.CAST}:
@@ -446,26 +446,8 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
   def find_gate(x:UOp) -> Optional[UOp]:
     if x.op is UOps.IF: return x
     return next((ret for s in x.src if (ret:=find_gate(s)) is not None), None)
-  if len(root.src) == 4 and (sub_gate:=find_gate(root.src[2])) and sub_gate.src[0] is root.src[-1].src[0]:
-    return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
-  return None
-
-def merge_gates(sink:UOp) -> Optional[UOp]:
-  @functools.lru_cache(None)
-  def has_range(x:UOp) -> bool: return any(s2p.op is UOps.RANGE for s2p in x.sparents)
-  if len([x.op is UOps.STORE and len(x.src) == 4 for x in sink.src]) <= 1: return None
-  if_src_to_if_op: Dict[UOp, UOp] = {}
-  for x in sink.src:
-    if x.op is UOps.STORE and len(x.src) == 4 and has_range(x.src[2]) and x.src[2] not in x.src[-1].src:
-      if_to_update = if_src_to_if_op.get(x.src[-1].src[0], x.src[-1])
-      if_src_to_if_op[x.src[-1].src[0]] = UOp(UOps.IF, None, (if_to_update.src[0],) + if_to_update.src[1:] + (x.src[2],), None)
-  if len(if_src_to_if_op) == 0: return None
-  new_sink_srcs = []
-  for x in sink.src:
-    if x.op is UOps.STORE and len(x.src) == 4:
-      new_sink_srcs.append(UOp(UOps.STORE, x.dtype, x.src[:-1] + (if_src_to_if_op[x.src[-1].src[0]],), x.arg))
-    else: new_sink_srcs.append(x)
-  return UOp(UOps.SINK, sink.dtype, tuple(new_sink_srcs), sink.arg)
+  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3]: return None
+  return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
 
 reducer = PatternMatcher([
   (NOp(UOps.REDUCE, name="root"), do_reduce),
@@ -475,7 +457,6 @@ reducer = PatternMatcher([
   (NOp(UOps.STORE, name="root"), delete_redundant_gates),
   # late fixup of unfoldable image loads
   (UPat(UOps.LOAD, src=(UPat(name="buf"), UPat()), allow_any_len=True, name="load"), fix_unfoldable_image_load),
-  # (NOp(UOps.SINK, name="sink"), merge_gates),
 ])
 
 no_pyint = PatternMatcher([(UPat({UOps.CONST, UOps.ALU, UOps.SPECIAL, UOps.RANGE, UOps.EXPAND}, dtype=dtypes.pyint, name="x"),
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -452,12 +452,13 @@ def merge_gates(sink:UOp) -> Optional[UOp]:
   def find_gate(x:UOp) -> Optional[UOp]:
     if x.op is UOps.RANGE: return x
     return next((ret for s in x.src if (ret:=find_gate(s)) is not None), None)
-  if_src_to_if_op, new_sink_srcs = {}, []
+  if_src_to_if_op: Dict[UOp, UOp] = {}
   for x in sink.src:
     if x.op is UOps.STORE and len(x.src) == 4 and x.src[-1].op is UOps.IF and find_gate(x.src[2]) and x.src[2] not in x.src[-1].src:
       if_to_update = if_src_to_if_op.get(x.src[-1].src[0], x.src[-1])
       if_src_to_if_op[x.src[-1].src[0]] = UOp(UOps.IF, None, (if_to_update.src[0],) + if_to_update.src[1:] + (x.src[2],), if_to_update.arg)
   if len(if_src_to_if_op) == 0: return None
+  new_sink_srcs = []
   for x in sink.src:
     if x.op is UOps.STORE and len(x.src) == 4 and x.src[-1].op is UOps.IF:
       new_sink_srcs.append(UOp(UOps.STORE, x.dtype, x.src[:-1] + (if_src_to_if_op[x.src[-1].src[0]],), x.arg))
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -447,9 +447,8 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
   def find_gate(x:UOp) -> Optional[UOp]:
     if x.op is UOps.IF: return x
     return next((ret for s in x.src if (ret:=find_gate(s)) is not None), None)
-  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3]:
-    if (sub_gate := find_gate(root.src[2])) is not None and sub_gate.src[0] is root.src[-1].src[0]:
-      return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
+  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3] \
+    and not ((sub_gate := find_gate(root.src[2])) is not None and sub_gate.src[0] is root.src[-1].src[0]):
     return None
   return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
 
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -414,7 +414,19 @@ def create_gate(root:UOp) -> Optional[UOp]:
   @functools.lru_cache(None)
   def _gate_srcs(u:UOp, gate:UOp) -> UOp:
     if u.op is UOps.BARRIER: return u
-    if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER and gate.op is not UOps.IF: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)  # noqa: E501
+
+    # NOTE: For my dumbass tomorrow. We want to have only a single if around a barrier and a store. TODO on how to distinguish them within this loop.
+    # Possibly undouble them within the dedupe method or within the final merge ifs method
+
+    # main of this branch
+    # if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER and gate.op is not UOps.IF: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)  # noqa: E501
+
+    # other options of this branch
+    # if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER and gate.op is not UOps.IF: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)  # noqa: E501
+    # if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)  # noqa: E501
+    if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate if gate.op is not UOps.IF else gate.src[0], u.src[-1])),), u.arg)  # noqa: E501
+    # if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate if gate.op is not UOps.IF else gate.src[0],)),), u.arg)  # noqa: E501
+
     if u.op is UOps.STORE and len(u.src) == 4 and u.src[-1].op in { UOps.ALU, UOps.CAST } and u.src[-1].dtype == dtypes.bool: return UOp(u.op, u.dtype, u.src[:-1] + (UOp(UOps.IF, None, (u.src[-1],)),), u.arg)  # noqa: E501
     return u if (replace_source:=tuple(_gate_srcs(x, gate) for x in u.src)) == u.src else UOp(u.op, u.dtype, replace_source, u.arg)
   return None if len(root.src) == 3 or (ret:=_gate_srcs(root, root.src[3])) is root else ret
@@ -444,17 +456,24 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
   def find_gate(x:UOp) -> Optional[UOp]:
     if x.op is UOps.IF: return x
     return next((ret for s in x.src if (ret:=find_gate(s)) is not None), None)
-  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3]: return None
+  # if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3]: return None
+  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate is not root.src[3]:
+    if (sub_gate := find_gate(root.src[2])) and sub_gate is not None and sub_gate.src[0] is root.src[-1].src[0]:
+      abcdef = 1
+      return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
+  # if len(root.src) == 3 or (gate:=find_gate(root.src[2])) is None or gate is not root.src[3]:
+    return None
   return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
 
 def merge_gates(sink:UOp) -> Optional[UOp]:
   @functools.lru_cache(None)
-  def find_gate(x:UOp) -> Optional[UOp]:
+  def find_range(x:UOp) -> Optional[UOp]:
     if x.op is UOps.RANGE: return x
-    return next((ret for s in x.src if (ret:=find_gate(s)) is not None), None)
+    return next((ret for s in x.src if (ret:=find_range(s)) is not None), None)
   if_src_to_if_op: Dict[UOp, UOp] = {}
   for x in sink.src:
-    if x.op is UOps.STORE and len(x.src) == 4 and x.src[-1].op is UOps.IF and find_gate(x.src[2]) and x.src[2] not in x.src[-1].src:
+    if x.op is UOps.STORE and len(x.src) == 4 and x.src[-1].op is UOps.IF and find_range(x.src[2]) and x.src[2] not in x.src[-1].src:
+    # if x.op is UOps.STORE and len(x.src) == 4 and x.src[-1].op is UOps.IF and x.src[2] not in x.src[-1].src:
       if_to_update = if_src_to_if_op.get(x.src[-1].src[0], x.src[-1])
       if_src_to_if_op[x.src[-1].src[0]] = UOp(UOps.IF, None, (if_to_update.src[0],) + if_to_update.src[1:] + (x.src[2],), if_to_update.arg)
   if len(if_src_to_if_op) == 0: return None
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -415,18 +415,16 @@ def create_gate(root:UOp) -> Optional[UOp]:
   def _gate_srcs(u:UOp, gate:UOp) -> UOp:
     if u.op is UOps.BARRIER: return u
     if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER and gate.op is not UOps.IF:
-      # print(f"Adding IF with src {u.src[-1]} to {u.op}")
       return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)
-    if u.op is UOps.STORE and len(u.src) == 4 and u.src[-1].op in { UOps.ALU, UOps.CAST } and u.src[-1].dtype == dtypes.bool:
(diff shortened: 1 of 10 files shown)


Qazalin:
bounty locked, good luck!
Your test output is correct but the implementation can be simpler,
Can you rewrite all gated stores with IF? There shouldn't be two paths for gated stores in renderers.

Also, can it do the grouping with a rewrite rule, like a merge_gates?
image

tinygrad/codegen/uopgraph.py
```diff
@@ -562,6 +565,17 @@ def push(u:UOp):
       if x.op is UOps.DEFINE_ACC:
         idx = min([self._uops.index(l) for l in x.src if l.op is UOps.RANGE])
         self._uops.insert(idx, x)
+      elif x.op is UOps.STORE and len(x.src) == 4 and x.src[-1].op is UOps.IF:
```
geohot, line 568: If possible, it's much better to do things as graph rewrite rules than here. Does this have to be done in the linearizer?

ianpaul10:
> bounty locked, good luck!

Thanks!

> Your test output is correct but the implementation can be simpler, Can you rewrite all gated stores with IF? There shouldn't be two paths for gated stores in renderers.

If I understood you correctly, I think all gated stores should now be handled in the renderer [now](https://github.com/tinygrad/tinygrad/pull/5976/files#diff-9694d15708a36a57a7e1485db60b1b4b1d2fc1e7278bc87127a73efd0bfa7dc0R130) (I think I just had that in as a safety precaution but things seem to work without it).

> Also, can it do the grouping with a rewrite rule, like a merge_gates?

Sorry, I'm probably missing something obvious, but I seem to only be getting it to work within the`linearize` function because of the way the prioritization is modified for any `RANGE` uops within the `push` function for adding items to the queue.

I tried to simplify things to only be in the `push` function but it doesn't work for all cases (specifically with nested ifs, TODO for me). Am I off-base at all with that observation? I don't want to take up too much of your time, but what sort of change to the prioritization/ordering of things could happen within a rewrite rule?

tinygrad/codegen/uops.py
```diff
@@ -222,7 +222,7 @@ def type_verify(uops):
     if uop is UOps.GEP: assert dtype == src[0].dtype.scalar(), f"GEP of {src[0].dtype=} should be {src[0].dtype.scalar()} != {dtype}"
     if uop is UOps.STORE:
       assert dtype is None, f"{uop} dtype must be None, got {dtype}"
-      if len(src) == 4: assert src[3].dtype == dtypes.bool, f"gate dtype mismatch {src[3].dtype} != {dtypes.bool}"
+      if len(src) == 4: assert src[3].dtype == dtypes.bool or src[3].op is UOps.IF, f"gate dtype mismatch {src[3].dtype} != {dtypes.bool}"
```
Qazalin, line 225: this shouldn't exist then:
```
src[3].dtype == dtypes.bool or
```

tinygrad/renderer/cstyle.py
```diff
@@ -115,7 +115,8 @@ def ssa(prefix:str, u:Optional[UOp]=None):
       uop,dtype,src,args = u.op,u.dtype,u.src,u.arg
       # these four uops don't have output dtypes
       if uop is UOps.IF:
-        kk(f"if ({r[src[0]]}) {{")
+        try: kk(f"if ({r[src[0]]}) {{")
+        except KeyError: kk(f"if ({r[src[0].src[0]]}) {{")
```
Qazalin, line 119: what is this?

tinygrad/codegen/uops.py
ianpaul10: Right, missed that sorry

Qazalin:
> the prioritization/ordering of things could happen within a rewrite rule?

I think if you do the graph rewriting right you don't need to worry about the queue.

tinygrad/renderer/cstyle.py
ianpaul10: Unnecessary now, some key errors were occurring with the nested refs when I was making bigger changes

ianpaul10:
> I think if you do the graph rewriting right you don't need to worry about the queue.

Ok, I'll give it a closer look and figure it out. Thanks!

tinygrad/renderer/cstyle.py
```diff
@@ -167,7 +167,7 @@ def ssa(prefix:str, u:Optional[UOp]=None):
             val = self.render_cast(precast, dtype, bitcast=True)
           elif uop is UOps.CAST: val = self.render_cast(r[src[0]], dtype, bitcast=False)
           else: val = self.render_vectorize([r[x] for x in src], dtype)
-          if child_count[u] <= 1: r[u] = val
+          if child_count[u] - len([x.op is UOps.IF and len(x.src) > 1 and u in x.src[1:] for x in u.parents]) <= 1: r[u] = val
```
ianpaul10, line 170: This was needed to stop unnecessary variables from being created (sometimes out of scope/in a nested loop) when it was only referenced by the STORE inside the code, and by the IF only for correct graph structure/ordering.

Qazalin: OK, can you move this to a separate diff with a test?
Your diffs should have the essential lines to get **correctness**. Net noop changes are always separate PRs.

Qazalin: is this the thing that's doing
image
?

ianpaul10: > OK, can you move this to a separate diff with a test?

Sounds good, I'll open up a PR specifically for this today.

> is this the thing that's doing

Ya that change would be due to this line modification.

The only reason I needed to include it is due to `test_padto_where_multireduce` failing with my new additions. It feels like a more subtle preexisting  bug that I might've ran into where a STORE src's could be out of scope if another UOp also references it, causing it to become a variable.

This patch only addresses it for the case of when an IF UOp refers to it, but I think there's the possibility of it bubbling up elsewhere.

Qazalin:
The diff looks very close!

Some things I noticed:

1. This isn't wrapping the second reduceop in the IF for multi reduce. It should, LOAD is expensive. `TestLinearizer.test_double_reduce_multireduce`
image

2. Can you comment this part better?
https://github.com/tinygrad/tinygrad/pull/5976/files#diff-00bd44b667ec90ae1d3e984e699bc6b498c84ca1b1bd15a025437ded227457bfR456-R457 - Why did gate creation needed changes?

ianpaul10:
W.r.t. point 1: Yes I saw that too after rebasing and having those tests now uncommented. I think it should be fixed now by joining the IFs in `delete_redundant_gates`. LMK if that isn't the appropriate place for that.

I'm going to work on doing it in within the first IF to reduce redundancy. I broke it out for now to be a bit more explicit about what it's doing.

W.r.t. point 2: Done. LMK if that's the sort of detail you were looking for.

And any advice/tips on how to debug the PTX tests? I saw similar failures on METAL previously (I'm doing dev work on a Mac) but struggling to recreate/debug the errors now even through emulation.

Qazalin:
@ianpaul10 you can access a green box for that, sent info on discord.

Qazalin:
yay! green CI, is the diff good for review?

tinygrad/renderer/assembly.py
```diff
@@ -126,8 +126,8 @@ def render_load(self, loc, dest, dtype, gate=None, alt=None, ss="", offset=0) ->
     if gate: return [f"@{gate} ld{ss}.{self.mem_types[dtype]} {dest}, [{loc}+{offset}];", f"@!{gate} mov.b{self.types[dtype][1:]} {dest}, {alt};"]
     return [f"ld{ss}.{self.mem_types[dtype]} {dest}, [{loc}+{offset}];"]
 
-  def render_store(self, loc, val, dtype, gate=None, ss="", offset=0) -> List[str]:
-    return [(f"@{gate} " if gate else "") + f"st{ss}.{self.mem_types[dtype]} [{loc}+{offset}], {val};"]
```
Qazalin, line 130: this is what I wanted to see, thanks.

tinygrad/codegen/uopgraph.py
```diff
@@ -477,9 +479,30 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
...
     return next((ret for s in x.src if (ret:=find_gate(s)) is not None), None)
-  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3]: return None
+  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3]:
+    if (sub_gate := find_gate(root.src[2])) and sub_gate is not None and sub_gate.src[0] is root.src[-1].src[0]:
+      return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
+    return None
   return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
 
+def merge_gates(sink:UOp) -> Optional[UOp]:
+  @functools.lru_cache(None)
```
Qazalin, line 489: I'd early return here if there are no gated stores in the sink.

tinygrad/renderer/assembly.py
```diff
@@ -175,26 +175,22 @@ def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
 
     for u in uops:
       uop,dtype,src,args = u.op,u.dtype,u.src,u.arg
-      if uop is UOps.IF:
-        assert src[0].dtype is not None
-        kk(*self.render_bra(f"IF_{r[src[0]][1:]}_{uops.index(u)}", _cast(r[src[0]], dtypes.bool, src[0].dtype, u=u, pred=True)))
```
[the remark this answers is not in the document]
ianpaul10: Right ok, my bad. What about the current changes where we set and enter the conditional branch at the start of the IF and exit/return at the ENDIF?

I'm still getting up to speed on PTX (& cuda) so not sure about the best approach. What was in place didn't seem correct which is why I removed it all together.

tinygrad/renderer/assembly.py
```diff
@@ -178,23 +178,24 @@ def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
...
+        kk(f"ENDIF_{r[src[0].src[0]][1:]}_{uops.index(src[0])}:")
       elif uop is UOps.STORE:
         assert src[0].dtype is not None and src[2].dtype is not None
         assert src[0].dtype == dtypes.int64, "store isn't int64"
         assert src[1].op is UOps.CONST, f"store isn't const {u}"
         mem_type = '.shared' if src[0].op is UOps.DEFINE_LOCAL or any(x.op is UOps.DEFINE_LOCAL for x in src[0].parents) else '.global'
         if src[2].dtype.count > 1:
-          kk((f"@{r[src[3]]} " if len(src)>3 else "") + \
-              f"st{mem_type}.v{src[2].dtype.count}.{self.mem_types[src[2].dtype.scalar()]} [{r[src[0]]}+{src[1].arg}], {{{', '.join(r[src[2]])}}};")
+          kk(f"st{mem_type}.v{src[2].dtype.count}.{self.mem_types[src[2].dtype.scalar()]} [{r[src[0]]}+{src[1].arg}], {{{', '.join(r[src[2]])}}};")
```
Qazalin, line 196: what changed here?

tinygrad/renderer/assembly.py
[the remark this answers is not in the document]
Qazalin: your new diff looks more correct, I'll look into why you had challenges with the old stuff.

tinygrad/codegen/uopgraph.py
```diff
@@ -477,9 +479,30 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
...
+  def find_range(x:UOp) -> Optional[UOp]:
+    if x.op is UOps.RANGE: return x
+    return next((ret for s in x.src if (ret:=find_range(s)) is not None), None)
```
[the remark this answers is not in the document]
ianpaul10: Ah good idea. Will switch to that. Should be able to save a few lines too.

tinygrad/renderer/assembly.py
ianpaul10: Simply removed the `(f"@{r[src[3]]} " if len(src)>3 else "") + \`, since there should never be in-line gated stores here anymore (they should be handled as IFs now)

tinygrad/renderer/assembly.py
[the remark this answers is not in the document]
ianpaul10: I'm not sure if it's helpful or not but it was putting the branch "entrypoint" at the ENDIF step, which excluded the actual store. Example snippet from `test_padto_where` from this branch without the rewrite of IF/ENDIF and master below.

master
```
        @%alu_pred_0	st.global.s32 [%alu_s64_1+0], %alu_s32_2;
```

this branch (without changes to IF/ENDIF:
```
        @%alu_pred_0	bra IF_alu_pred_0_31;
	st.global.s32	[%alu_s64_1+0], %alu_s32_2;
	IF_alu_pred_0_31:
```

[test_padto_where__local_failing_ptx.txt](https://github.com/user-attachments/files/16696555/test_padto_where__local_failing_ptx.txt)
[test_padto_where__master_passing_ptx.txt](https://github.com/user-attachments/files/16696557/test_padto_where__master_passing_ptx.txt)

Qazalin:
your diff shouldn't include any noqa for long lines, can you try to break the code down?

tinygrad/codegen/uopgraph.py
```diff
@@ -477,9 +481,29 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
...
+  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3]:
+    if (sub_gate := find_gate(root.src[2])) is not None and sub_gate.src[0] is root.src[-1].src[0]:
+      return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
+    return None
   return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
 
+def merge_gates(sink:UOp) -> Optional[UOp]:
+  @functools.lru_cache(None)
+  def has_range(x:UOp) -> bool: return any(s2p.op is UOps.RANGE for s2p in x.sparents)
+  if not any(x.op is UOps.STORE and len(x.src) == 4 and x.src[-1].op is UOps.IF for x in sink.src): return None
```
Qazalin, line 493: will STORE ever have len(x.src) == 4 and not be IF here? I think all gates should have been rewritten already, aren't they?

ianpaul10: Oh ya good catch, I guess I was just being extra cautious. Will update now

tinygrad/codegen/uopgraph.py
```diff
@@ -477,9 +481,28 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
   def find_gate(x:UOp) -> Optional[UOp]:
     if x.op is UOps.IF: return x
     return next((ret for s in x.src if (ret:=find_gate(s)) is not None), None)
-  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3]: return None
+  if len(root.src) == 3 or (gate:=find_gate(root)) is None or gate.src[0] is not root.src[3] \
+    and not ((sub_gate := find_gate(root.src[2])) is not None and sub_gate.src[0] is root.src[-1].src[0]):
```
ianpaul10, line 485: Unsure about the best way to do this. I had it as a nested if statement before, but then `return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)` was getting repeated again 🤷 This way only adds 2 lines (could be only 1 if we want to bump the `return None` to this line) but is maybe a bit harder to read.

ianpaul10:
I'm not sure about the norms in this repo, but would you like me to clean up the commit history into only a few isolated commits, or do you normally keep the history as-is?

tinygrad/codegen/uopgraph.py
Qazalin: "extra caution" is the wrong way to think about the uopgraph codebase.
The pattern matcher must be deterministic and have clear, localized behavior.

Qazalin:
how much did you benchmark this against:
PROFILE=0 python3 test/external/external_benchmark_schedule.py
is it fast?

tinygrad/codegen/uopgraph.py
```diff
@@ -477,9 +481,28 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
...
+    return None
   return UOp(UOps.STORE, root.dtype, root.src[:3], root.arg)
 
+def merge_gates(sink:UOp) -> Optional[UOp]:
+  @functools.lru_cache(None)
+  def has_range(x:UOp) -> bool: return any(s2p.op is UOps.RANGE for s2p in x.sparents)
+  if not any(x.op is UOps.STORE and len(x.src) == 4 for x in sink.src): return None
```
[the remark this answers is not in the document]
ianpaul10: Do you mean something like `if not any(x.op is UOps.IF in sink.parents): return None`? Or do you mean checking the `x.src[-1].op is UOps.IF`?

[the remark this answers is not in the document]
ianpaul10: Sorry, I completely misunderstood your question last night.

Yes, that's clever. It should work since we shouldn't care about positioning of the IF when there's only one.

tinygrad/codegen/uopgraph.py
```diff
@@ -448,7 +448,11 @@ def create_gate(root:UOp) -> Optional[UOp]:
   @functools.lru_cache(None)
   def _gate_srcs(u:UOp, gate:UOp) -> UOp:
     if u.op is UOps.BARRIER: return u
-    if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)
+    if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER:
+      # NOTE: gate could already be wrapped in an IF, so only take IF's src[0] in that case. Will join IFs in delete_redundant_gates
+      return UOp(u.op, u.dtype, u.src[:-1] + (UOp(UOps.IF, None, (gate if gate.op is not UOps.IF else gate.src[0], u.src[-1])),), u.arg)
+    if u.op is UOps.STORE and len(u.src) == 4 and u.src[-1].op in { UOps.ALU, UOps.CAST } and u.src[-1].dtype == dtypes.bool:
```
Qazalin, line 454: do you actually need all these checks?

tinygrad/codegen/uopgraph.py
Qazalin: what is the most readable and simplest you can make this?
I can't read this tbh.

tinygrad/codegen/uopgraph.py
```diff
@@ -477,9 +481,28 @@ def delete_redundant_gates(root:UOp) -> Optional[UOp]:
...
 
+def merge_gates(sink:UOp) -> Optional[UOp]:
+  @functools.lru_cache(None)
+  def has_range(x:UOp) -> bool: return any(s2p.op is UOps.RANGE for s2p in x.sparents)
+  if len([x.op is UOps.STORE and len(x.src) == 4 for x in sink.src]) <= 1: return None
+  if_src_to_if_op: Dict[UOp, UOp] = {}
+  for x in sink.src:
+    if x.op is UOps.STORE and len(x.src) == 4 and has_range(x.src[2]) and x.src[2] not in x.src[-1].src:
+      if_to_update = if_src_to_if_op.get(x.src[-1].src[0], x.src[-1])
+      if_src_to_if_op[x.src[-1].src[0]] = UOp(UOps.IF, None, (if_to_update.src[0],) + if_to_update.src[1:] + (x.src[2],), if_to_update.arg)
```
Qazalin, line 497: i dont remember if having an arg...

Qazalin:
so I think you can break this into two PRs:
1. Rewrite all gated stores to UOps.IF
2. Add merge_gates

tinygrad/codegen/uopgraph.py
ianpaul10: I was just following suit with when the IF was getting constructed in `create_gate`, but `None` probably makes more sense, you're right. Will swap out

tinygrad/codegen/uopgraph.py
```diff
@@ -451,7 +451,7 @@ def _gate_srcs(u:UOp, gate:UOp) -> UOp:
     if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER:
       # NOTE: gate could already be wrapped in an IF, so only take IF's src[0] in that case. Will join IFs in delete_redundant_gates
       return UOp(u.op, u.dtype, u.src[:-1] + (UOp(UOps.IF, None, (gate if gate.op is not UOps.IF else gate.src[0], u.src[-1])),), u.arg)
-    if u.op is UOps.STORE and len(u.src) == 4 and u.src[-1].op in { UOps.ALU, UOps.CAST } and u.src[-1].dtype == dtypes.bool:
+    if u.op is UOps.STORE and len(u.src) == 4 and u.src[-1].op in { UOps.ALU, UOps.CAST }:
```
Qazalin, line 454: is not IF a bette idea?

ianpaul10: 🤦

ianpaul10:
More than happy to split into two separate PRs, but `Rewrite all gated stores to UOps.IF` PR won't pass tests until `Add merge_gates` is included. Are you ok with me swapping the order of your proposal?

Qazalin:
hmm, interesting. Why would it not pass?
If it doesn't take much of your time I'd really wanna see why just rewriting to UOps.IF doesn't pass the tests.

ianpaul10:
It doesn't pass because of the issue when a STORE is dependent on a variable within a RANGE but the IF isn't, so the IF gets written first but the variable we want to write is out of scope. (the same issue I brought up here https://github.com/tinygrad/tinygrad/pull/5976#issuecomment-2279784567)

It's probably easier to look at an example: `DEBUG=7 METAL=1 PYTHONPATH="." python3 -m pytest test/test_linearizer.py::TestKernelOpts::test_padto_where` when `merge_gates` isn't included in the rewrite rules results in the following invalid output: 
[test_padto_where__fails.txt](https://github.com/user-attachments/files/16737331/test_padto_where__fails.txt)

and error msg: 
Screenshot 2024-08-25 at 12 11 22 AM

tinygrad/codegen/uopgraph.py
ianpaul10: Actually that seems to cause issues by including store's where `UOps.EXPAND` is the last src uop. I think limiting it to only actual bool ops is cleaner. What do you think?

Qazalin:
you're having that problem because those IF srcs are wrong.
My expected output is:
Screenshot 2024-08-25 at 6 31 35 PM
(this is `python3 test/test_linearizer.py TestLinearizer.test_padto_max_multireduce`)

[This](https://github.com/tinygrad/tinygrad/pull/5976#discussion_r1730038001) is also because of a bug in this diff.

Qazalin:
better, can you cleanup the diff (removing merge_gates and stuff)?
I'll take ownership of that process replay red and merge if all it good.

tinygrad/ops.py
```diff
@@ -341,7 +341,7 @@ def type_verify(uops):
     if uop is UOps.GEP: assert dtype == src[0].dtype.scalar(), f"GEP of {src[0].dtype=} should be {src[0].dtype.scalar()} != {dtype}"
     if uop is UOps.STORE:
       assert dtype is None, f"{uop} dtype must be None, got {dtype}"
-      if len(src) == 4: assert src[3].dtype == dtypes.bool, f"gate dtype mismatch {src[3].dtype} != {dtypes.bool}"
+      if len(src) == 4: assert src[3].op is UOps.IF, f"{uop} gate op must be {UOps.IF}, got {src[3].op}"
```
Qazalin, line 344: I'd just hadcode UOps.IF instead of format.

Qazalin:
Process replay failed because this is somehow making things non deterministic.
printed a full log in:
https://github.com/tinygrad/tinygrad/actions/runs/10558375543/job/29247702829#step:24:61

Qazalin:
hmm, also broke local reduce gating:
image
TestLinearizerFailures.test_failure_43

tinygrad/ops.py
```diff
@@ -380,7 +380,6 @@ def flops_mem(uops:List[UOp], ignore_indexing=False) -> Tuple[sint, sint]:
         if len(u.src) > 3: dont_count = dont_count.union(u.src[2].sparents)
       elif u.op is UOps.STORE:
         dont_count = dont_count.union(u.src[1].sparents)
-        if len(u.src) > 3: dont_count = dont_count.union(u.src[3].sparents)
```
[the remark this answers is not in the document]
ianpaul10: Sure, will add tests now

Qazalin:
test_failure_43 is just wrong, see where the IF is?
The tests don't just fail on AMD, just copying the AST and applied opts in the logs will show you how the program code changed.
eg do something like:
````
ast = // from logs
opts = // from logs
k = Kernel(ast)
for opt in opts: k.apply_opt(opt)
print(k.to_program().src)
```
do you see a change similar to what's in CI? I'd root cause why the alu ordering is different.

ianpaul10:
I think the local reduce gating should be fixed and process replay should now be working. Thanks for the pointers on debugging. It took me a while to wrap my head around things but it should be in a working state now.

Qazalin:
cool! I'll take some time to review this today.

Qazalin:
This doesn't look right:
https://github.com/tinygrad/tinygrad/actions/runs/10621574069/job/29444432870?pr=5976#step:24:480
In general, it's good practice to keep your branch up-to-date and make sure the process replay diff matches what the reviewer would expect.

Like this is great:
image
I'm not even sure how this is green
image

Qazalin:
feel free to mark as ready for review when the upcast issues are fixed.

ianpaul10:
I found that the cleanest way to have the IF to correctly reference the STORE's 3rd src was to have it happen after the expand.

When I tried to add the reference to the STORE's 3rd src when it was still an EXPAND op, I would end up with issues like we saw in the process replay. That seemed to require the expand logic needing to change to account for the additional references within IF's src. This all seemed a bit messier vs just adding the final STORE 3rd src.

Am I understanding that correctly or am I missing something?

Qazalin:
benchmarks red 
https://github.com/tinygrad/tinygrad/actions/runs/10714523120/job/29708335366

Qazalin:
PTX looks wrong https://github.com/tinygrad/tinygrad/actions/runs/10717084318/job/29716009654?pr=5976#step:6:1

ianpaul10:
Yes, I think I fixed the issue in the latest commit. 

However, I'm unable to verify on tinybox green, I think I lost access to be able to ssh into the server.

What's the easiest way for me to trigger the benchmarks?

ianpaul10:
Thank you, I really appreciate your patience. I'll take a look at your changes and get the merge_gates submitted.

Qazalin:
fyi had to revert because of a BEAM failure, can you add a regression test for it and put up the diff with the fix?

PR #5983, opened 2024-08-08
sort UOp folded mod/div terms

chenyuxyz:


chenyuxyz:
we might just want a more generic way to sort adds

PR #6005, opened 2024-08-09
ELF loader improvements and tests

uuuvn:


wozeparrot:
don't feel like `extra/datasets` is a good place to put the test files, somewhere under `test/` would be better

uuuvn:
> don't feel like `extra/datasets` is a good place to put the test files, somewhere under `test/` would be better

Yeah, every other place seems worse though. Created test/blobs and moved it there

uuuvn:
Maybe move extra/datasets/sops.gz there too?

geohot:
Is this the highest quality you can make this code? Before I point anything out, please try to preempt any feedback and fix things. This is how you become a better software engineer.

geohot:
How about this, the current `_elf_parse_names` contains a bug that `_strtab` fixes.

Can you submit a minimal PR that fixes that bug and contains a regression test for it? (aka it would currently fail)

geohot:
Closing for #6011

PR #5515, opened 2024-07-16
Multireduce Lowerer

0xtimmy:
Codegen changes to handle a basic multireduce kernel

PR #6007, opened 2024-08-09
Lowerer Multireduce Uopgraph

0xtimmy:
This is another portion of #5515 that makes minimal changes to the uopgraph to support reductions.
- it tracks the dependencies if `UOps.PHI`s
- uses that to prioritize `UOps.RANGE`s


test/test_linearizer.py
```diff
@@ -114,25 +113,29 @@ def test_multireduce(self):
...
+      # [Opt(OptOps.UNROLL, 0, 4), Opt(OptOps.UNROLL, 1, 4), Opt(OptOps.GROUPTOP, 0, 8), Opt(OptOps.GROUPTOP, 0, 8)],
     ]
     wanna_output = (x.numpy()-x.numpy().sum(-1, keepdims=True)).sum(-1).reshape(1,1)
-    helper_linearizer_ast((store, ), [x], wanna_output=[wanna_output], opts=opts)
+    lins = helper_linearizer_ast((store, ), [x], wanna_output=[wanna_output], opts=opts)
+    for l in lins:
+      ranges = [u.op for u in l.uops if (u.op is UOps.RANGE and u.arg[1]) or (u.op is UOps.ENDRANGE and u.src[0].arg[1])]
+      for i,u in enumerate(ranges):
+        if i == 0: continue
+        assert ranges[i-1] != u, f"multireduce nested the ranges! {ranges[i-1], {u}}"
```
0xtimmy, line 134: this check tests for whether the reduces are nested [O(n^2)]:
```
__kernel void r_32_32(__global float* data0, const __global float* data1) {
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 32; ridx0++) {
    float val0 = data1[ridx0];
    float acc1 = 0.0f;
    for (int ridx1 = 0; ridx1 < 32; ridx1++) {
      float val1 = data1[ridx1];
      acc1 = (acc1+val1);
    }
    acc0 = (acc0+val0+(-acc1));
  }
  data0[0] = acc0;
}
```

versus sequential [O(n)]:
```
__kernel void r_32_32(__global float* data0, const __global float* data1) {
  float acc0 = 0.0f;
  for (int ridx0 = 0; ridx0 < 32; ridx0++) {
    float val0 = data1[ridx0];
    acc0 = (acc0+val0);
  }
  float acc1 = 0.0f;
  for (int ridx1 = 0; ridx1 < 32; ridx1++) {
    float val1 = data1[ridx1];
    acc1 = (acc1+val1+(-acc0));
  }
  data0[0] = acc1;
}
```

tinygrad/codegen/uopgraph.py
```diff
@@ -475,13 +475,16 @@ def find_gate(x:UOp) -> Optional[UOp]:
 
 # *** uop graph ***
 
-def get_children_dfs(u:UOp, children:Dict[UOp, List[UOp]], in_degree:Dict[UOp, int]):
-  if u in children: return
+def get_children_dfs(u:UOp, children:Dict[UOp, List[UOp]], srcs:Dict[UOp, Set[UOp]], in_degree:Dict[UOp, int]):
+  if u in children: return srcs[u]
+  srcs[u] = set()
```
Qazalin, line 480: set is nondeterministic, can you use Dict?

tinygrad/codegen/uopgraph.py
```diff
@@ -543,8 +546,9 @@ def linearize(self, extra_pm:Optional[PatternMatcher]=None, skip_check=False) ->
     # filter nodes that don't link to a sink
     # BFS toposort
     children: Dict[UOp, List[UOp]] = {}
+    range_srcs: Dict[UOp, Set[UOp]] = {}
```
Qazalin, line 549: so arg[1] refers to if the loop is implicit (gidx) or explicit (for) - since we don't care about the implicit ones already I think this can be faster if we just track `arg[1] is True`

0xtimmy: what do you mean by just tracking `arg[1] is True`? I think it already does check `arg[1]` before adding to `range_srcs`

I need to keep some unique identifier so that we can prioritize ranges by how deep they are in the ast, if we want to be extra tight on space I could probably track `arg[0]`

Qazalin: I see a few checks for truthy `arg[1]` in the lines after this, why do you need to track the global loops?

0xtimmy: I think they're leftover from debugging, I should be able to get rid of them

tinygrad/codegen/uopgraph.py
```diff
@@ -553,13 +557,16 @@ def get_recursive_children(x:UOp, end:UOps, include_self=False) -> Set[UOp]:
...
+    range_phi = {r:[p for p in scope_children[r] if p.op is UOps.PHI] for r in scope_children if r.op is UOps.RANGE}
+    range_srcs = {p:range_srcs[p] for p in range_srcs if p.op is UOps.PHI}
 
     queue:List[Tuple[int, UOp]] = []
     def push(u:UOp):
       priority = 0
       # prefer uops that are loop children
-      for l, ss in scope_children.items():
-        if l.op is UOps.RANGE and u in ss: priority -= l.arg[0]*1000 + l.arg[1]
+      if u.op is UOps.RANGE and u.arg[1]: priority += u.arg[0] + 10000 * \
```
geohot, line 567: Feel free to use more lines for this and add comments. I don't understand what this is doing

0xtimmy: I changed it to look like this:
```
if u.op is UOps.RANGE and u.arg[1]: 
        priority += u.arg[0]
        for p in range_phi[u]: # for every phi in this ranges scope
          # add up all the ranges that precede it
          # except those that are also part of this reduction
          priority += 10000*len([r for r in range_srcs[p] if not set(range_phi[u]).intersection(set(range_phi[r]))])
```

is that more clear?

0xtimmy: there are other ways to do the intersection if we don't want to use sets, I just used it here because it more intuitive to what's actually being calculated

tinygrad/codegen/uopgraph.py
```diff
@@ -553,13 +557,22 @@ def get_recursive_children(x:UOp, end:UOps, include_self=False) -> Set[UOp]:
...
     queue:List[Tuple[int, UOp]] = []
     def push(u:UOp):
       priority = 0
-      # prefer uops that are loop children
-      for l, ss in scope_children.items():
-        if l.op is UOps.RANGE and u in ss: priority -= l.arg[0]*1000 + l.arg[1]
+      # ensure ranges fall in the proper order
+      if u.op is UOps.RANGE and u.arg[1]:
+        priority += u.arg[0]
+        # for each phi in this range's scope
```
Qazalin, line 569: does this comment add value/context?

0xtimmy: not really, just writing out the thought process. I'll remove it in the next commit

0xtimmy: do you like the set().intersection change?

Qazalin: Would using `any` be enough since range_phi is already deduped? Though .intersection is readable.
Fine with whichever one you like.

0xtimmy: yeah I can switch it back to any

PR #6023, opened 2024-08-10
docs: fix broken links and update is_floating_point

chenyuxyz:
broken links would only show as INFO and not an error.


chenyuxyz:
added a validation so broken anchor warns now and should throw errors in CI

PR #6024, opened 2024-08-10
Cleanup cstlye render_const [run_process_replay]

greg-niemeyer:
Simplifies cstyle `render_const` in the spirit of #5313. This change takes advantage of the fact that CONSTs are always scalar, and localizes the complexity of deciding when to apply casts and extra parentheses to CONSTs in ALUs, where CONSTs are used to construct expressions as opposed to being used for simple assignments (where type promotion is easy to reason about).

process_replay impact: less excessive casting and parens.

e.g.
image

### DIFF
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -47,11 +47,10 @@ class CStyleLanguage(Renderer):
   def render_const(self, x:ConstType, dtype:DType) -> str:
     if math.isnan(x): val = "NAN"
     elif math.isinf(x): val = ("-" if x < 0 else "") + "INFINITY"
-    elif dtype.scalar() == dtypes.bool: val = "1" if x else "0"
-    elif dtype.scalar() == dtypes.float: val = f"{x}f"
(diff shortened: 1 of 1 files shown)


Qazalin:
because NAN is always float32 this should cast in render_const.
#define NAN (__int_as_float(0x7fffffff))"
In general, I'm fine with doing all render_casts in the const renderer.

geohot:
Can rendering vectorized just be removed from render const as a separate PR?

geohot:
Maybe assert its count is 1

tinygrad/renderer/cstyle.py
```diff
@@ -45,13 +45,13 @@ def render_vectorize(self, x:List[str], var_dtype:DType) -> str:
 
   # returns a str expression of the const with the given type
   def render_const(self, x:ConstType, dtype:DType) -> str:
+    assert dtype.count == 1, "consts should be scalar, got {dtype}"
```
Qazalin, line 48: err

PR #6030, opened 2024-08-10
AST is UOp

Qazalin:
Complete indexing fusion needs a great graph rewriting foundation, the brute force approach with _recurse_reduceops has been challenging to get perfectly right.

---
edit:

passes e2e tests,
scheduler creates UOps.ST_IDX, UOps.ST_VALID, Kernel swaps these after optops, Lowerer replaces them with idxs.
deletes LazyOp, MemBuffer etc and Lowerer is less lines.

TODO:
- [x] It's slow because UOps doesn't work with method_cache.
- [x] refactor search.py
- [x] image dtype
- [x] make model linearize fast again
- [x] Update hardcoded LazyOp tests with the new style AST

--- 
additional improvements:

1. how much simpler can Lowerer be? like a lot of these can just be a `replace`
2. do we need ReduceOps.WMMA
3. similarly, do we need dtypes in the wmma_arg?
4. lazy.py's CAST doesn't need arg


Qazalin:
master:
```
***** model forward in  57.86 ms
***** model schedule in  22.14 ms
***** model uops in 264.75 ms
***** model linearize in 3772.35 ms
```
d902c831fe02a2da4ec8787a658faff26806ff71
```
***** model forward in  57.35 ms
***** model schedule in  24.27 ms
***** model uops in 553.93 ms
***** model linearize in 6946.44 ms
```

PR #6049, opened 2024-08-12
fix openpilot benchmark

nimlgen:


nimlgen:
@wozeparrot looks like a typo, right?

wozeparrot:
yup

PR #6053, opened 2024-08-12
test: uop and lazyop have the same compare

Qazalin:
This is important for the method_cache in ast as UOp.
it's the LazyOp `__eq__`+ dtype compare.


geohot:
I like this, the LazyOp one was always much better

PR #6054, opened 2024-08-12
proposal: MetaOps.EXT

Qazalin:
`MetaOps.CUSTOM, MetaOps.COPY, MetaOps.EMPTY, MetaOps.VIEW` don't fit into any of our existing UOps.

MetaOps.KERNEL and MetaOps.EXT can be the two paths in realize.py.

after AST is UOp:
MetaOps.KERNEL -> UOps.SINK
MetaOps.EXT -> UOps.EXT


geohot:
Yea, that's fine, and remove KERNEL/EXT from MetaOps when you switch to UOps only

PR #6058, opened 2024-08-12
touchup cl_errors

chenyuxyz:
also enabled the test in CI


geohot:
This is a dangerous test for CI. It actually allocates tons of RAM

PR #6061, opened 2024-08-13
Apply transcendental rewrite rules only if required by backend

andredaprato:
Applies transcendental rewrite rules if a `Renderer` doesn't have code for the ops. I've rewritten it so that `TRANSCENDENTAL=1` has no effect anymore so I think `TRANSCENDENTAL` should only handle one value, but I'd like to be sure this is the right approach first.

### DIFF
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -133,8 +133,9 @@ def div_folding(x:UOp, c:int) -> Optional[UOp]:
 
 # ***** transcendental *****
 
-transcendental_folding = PatternMatcher([(UPat(UOps.ALU, dtype=TRANSCENDENTAL_SUPPORTED_DTYPES, src=(UPat(name="d"),), arg=k), cast(Callable, v))
-                                         for k,v in ((UnaryOps.EXP2, xexp2), (UnaryOps.LOG2, xlog2), (UnaryOps.SIN, xsin))])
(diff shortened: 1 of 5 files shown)


tinygrad/renderer/llvmir.py
```diff
@@ -71,6 +53,20 @@ class LLVMRenderer(Renderer):
   has_local = False
   has_shared = False
   global_max = None
+  code_for_op: Dict[Op, Callable] = {
```
geohot, line 56: Why did this move? Can you do the minimal change required?

tinygrad/codegen/uopgraph.py
```diff
@@ -503,9 +504,7 @@ def __init__(self, sink:Union[UOp, List[UOp]], opts:Optional[Renderer]=None):
     # used by linearizer
     self._uops: Optional[List[UOp]] = None
     self.opts = opts
-    self.folder = constant_folder
-    if TRANSCENDENTAL >= 2 or (opts is not None and TRANSCENDENTAL >= 1 and opts.device in {"CLANG", "LLVM"}):
-      self.folder = self.folder + transcendental_folding
+    self.folder = constant_folder + transcendental_folding({} if TRANSCENDENTAL >= 2 or opts is None else opts.code_for_op.keys())
```
geohot, line 507: So if opts is None you default to folding all

geohot:
Cool, good work

PR #6079, opened 2024-08-14
fix qcom process_replay for kernel diff

chenyuxyz:
skipping the wrong exception?


chenyuxyz:
@Qazalin fyi, there was a time parsing error and the whole kernel diff was incorrectly skipped. I skipped the time diff part so the kernel diff part can run

PR #6084, opened 2024-08-15
rewrite bool ADD to OR and MUL to AND

chenyuxyz:
fixed running `tinyphysics.onnx`, which contains a getitem from a boolean tensor.

only can repro through BEAM_COMPARE, which i think is a different bug in test_linearizer_failure

### DIFF
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -261,6 +261,8 @@ constant_folder = PatternMatcher([
   (NOp.var('x') // 1, lambda x: x),   # x//1 -> x
   (NOp.var('x') // -1, lambda x: -x), # x//-1 -> -x
   (NOp.var('x') / NOp.var('x'), lambda x: x.const(1)), # x/x -> 1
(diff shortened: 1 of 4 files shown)


chenyuxyz:
@Qazalin can you help with remu unimplemented instruction?

tinygrad/codegen/uopgraph.py
```diff
@@ -258,6 +261,8 @@ def index_collapse(idx,rng,buf,add,mul,ld,reduce):
   (NOp.var('x') // 1, lambda x: x),   # x//1 -> x
   (NOp.var('x') // -1, lambda x: -x), # x//-1 -> -x
   (NOp.var('x') / NOp.var('x'), lambda x: x.const(1)), # x/x -> 1
+  (NOp.var('x') & NOp.cvar('c'), lambda x,c: (x if c.arg else c) if x.dtype==dtypes.bool else None),
+  (NOp.var('x') | NOp.cvar('c'), lambda x,c: (c if c.arg else x) if x.dtype==dtypes.bool else None),
```
geohot, line 265: `NOp.var('x', dtype=dtypes.bool)` is same, right?

PR #6087, opened 2024-08-15
Tensor interpolate nearest mode

geohotstan:
did this before in onnx heh


geohot:
Can you remove the logic from onnx and reuse the new stuff in Tensor?

tinygrad/tensor.py
```diff
@@ -1961,15 +1962,21 @@ def interpolate(self, size:Tuple[int, ...], mode:str="linear", align_corners:boo
...
     for i in range(-len(size), 0):
       scale = (self.shape[i] - int(align_corners)) / (size[i] - int(align_corners))
       arr, reshape = Tensor.arange(size[i], dtype=dtypes.float32, device=self.device), [1] * self.ndim
-      index = (scale*arr if align_corners else (scale*(arr+0.5))-0.5).clip(0, self.shape[i]-1)
-      reshape[i] = expand[i] = size[i]
-      low, high, perc = [y.reshape(reshape).expand(expand) for y in (index.floor(), index.ceil(), index - index.floor())]
-      x = x.gather(i, low).lerp(x.gather(i, high), perc)
+      if mode in {"nearest", "nearest-exact"}:
+        index = (scale * arr).floor() if mode == "nearest" else (scale*(arr+0.5)-0.5).round()
+        # TODO: not sure if fastest idx selection method looping like dis
```
geohot, line 1972: Yea, this isn't ready

tinygrad/tensor.py
```diff
@@ -1961,8 +1962,23 @@ def interpolate(self, size:Tuple[int, ...], mode:str="linear", align_corners:boo
...
+      arrs = (Tensor.arange(s, dtype=dtypes.int32, device=self.device) for s in size)
+      scales = (self.shape[i] / size[i] for i in range(-len(size), 0))
+      # NOTE: for nearest-exact...
+      # https://github.com/pytorch/pytorch/commit/6adbe044e39c8e8db158d91e151aa6dead6e9aa4#diff-bf10b066c2c62c193fa16ea880093284cbb3e53a3406b702759cb31041e6ae91
+      #   // index_f32 = (output_index + 0.5) * scale - 0.5
+      # // input_index = round(index_f32)
+      # // Same as Pillow and Scikit-Image/Scipy ndi.zoom
+      # const int64_t src_index =
+      # std::min(static_cast<int64_t>(floorf((dst_index + 0.5) * scale)), input_size - 1);
+      idxs = [(arr*scale).floor().cast(dtypes.int) if mode == "nearest" else (((arr+0.5)*scale)).floor().cast(dtypes.int)
```
geohotstan, line 1977: for some reason the psuedo-code they gave for nearest exact is 
```
      // index_f32 = (output_index + 0.5) * scale - 0.5
      // input_index = round(index_f32)
```

but the actual cpp implementation is `((output_index + 0.5) * scale).floor()` ...
I don't get it... 

but tests pass with the actual implementation
results also match `PIL.NEAREST` I think.
Gonna spam a few more tests and then try to clean up.

chenyuxyz (commented):
@Qazalin can you check remu for this too?

geohotstan:
I rewrote the resize op.
The implementation before was just horribly hacked. I don't even know how tests passed looking back...
The resize op does not directly call `Tensor.interpolate` because the `_apply_coordinate_function` has a bunch of transformations that is different from the modes for `Tensor.interpolate`. (Tensor.interpolate assumes a specific mode in `apply_coordinate_function`)
I just copied the impl of `Tensor.interpolate` over to onnx resize.

for `test/external/external_test_image_resize.py` I wrote some random tests for fun cuz I was curious how bad the actual discrepancies were
currently our `nearest` and `nearest-exact` modes perfectly match `torch.interpolate`. 
torch.interpolate's `nearest-exact` does not exactly match PIL's `NEAREST`. (around 1% to 5% of mismatched elements)
torch.interpolate's `nearest` does not match PIL's `NEAREST` at all. (over 50% mismatched elements)
I'm not too sure what to think of this. Should we try to match PIL's `NEAREST` exactly?

tinygrad/tensor.py
```diff
@@ -1961,15 +1962,21 @@ def interpolate(self, size:Tuple[int, ...], mode:str="linear", align_corners:boo
...
-      x = x.gather(i, low).lerp(x.gather(i, high), perc)
+      if mode in {"nearest", "nearest-exact"}:
+        index = ((arr + (0.5 if mode == "nearest-exact" else 0)) * scale).floor().cast(dtypes.int)
+        indexes.append(index.reshape(reshape[-len(size):]).expand(size))
+      elif mode == "linear":
+        index = (scale*arr if align_corners else (scale*(arr+0.5))-0.5).clip(0, self.shape[i]-1)
+        low, high, perc = [y.reshape(reshape).expand(expand) for y in (index.floor(), index.ceil(), index - index.floor())]
+        x = x.gather(i, low).lerp(x.gather(i, high), perc)
+      else: raise ValueError(f"invalid {mode=}")
+    if mode in {"nearest", "nearest-exact"}: x = x[tuple(slice(None) for _ in range(x.ndim - len(indexes))) + tuple(indexes)]
```
geohotstan, line 1979: also, this line of code looks weird because pylint won't allow `x = x[..., *indexes]`

test/test_ops.py
```diff
@@ -2039,6 +2067,18 @@ def test_interpolate_bilinear(self):
       lambda x: torch.nn.functional.interpolate((10*x).relu().type(torch.uint8), size=out_sz, mode="bilinear"),
       lambda x: Tensor.interpolate((10*x).relu().cast('uint8'), size=out_sz, mode="linear"), forward_only=True)
 
+  def test_interpolate_nearest(self):
```
geohot, line 2070: Does the bilinear test above work? It's still skipped

tinygrad/tensor.py
geohot: Hmm, why won't pylint allow that?

test/test_ops.py
geohotstan: it doesn't work yet. 
I'll try to fix this too.

geohotstan: it seems that there are uint8 specific interpolation kernels for pytorch.
I think I've pinned down the difference that's causing numerical issues to this
https://github.com/pytorch/pytorch/blob/2bd02e0c82c6d76157244655256062bade5070ae/aten/src/ATen/native/cpu/UpSampleKernel.cpp#L943
I don't 100% get what's going on, so I'm gonna stare at it for a bit longer until I figure it out

also apparently there is also some lerp numerical instability as well.
https://github.com/pytorch/pytorch/pull/18871
Though I copied their impl and it didn't fix the bilinear test.

geohotstan:
ok #6209 had an effect on linear interpolation
I'm prettttyyy sure it's cuz of this line
before: `data0[alu0] = ((float)(acc0)+((float)((val2+(-acc0)))*val1));`
now: `data0[alu0] = ((float)(acc0)+((float)((val2+(acc0*((unsigned char)(-1)))))*val1));`

this is cuz the acc0 is neg-ed, and that `(unsigned char)(-1)` wraps around to `255` so now it becomes `val2 + (acc0 * 255)` instead of `val2 - acc0`

gonna see if there's a principled way of fixing this.

PR #6088, opened 2024-08-15
add nearest support to interpolate

abeleinin:
Follows torch implementation


geohot:
Hmm, should it still be using lerp? How is this test passing?

geohot:
Can you update the onnx implementation to use this?

abeleinin:
Yeah, lerp doesn't do anything for "nearest". I'll fix that to not waste compute and clean up the onnx implementation

abeleinin:
Looking into uint8 now. 

Do you want the `nearest-exact` mode added?

abeleinin:
torch implements a native interpolate for [uint8 on cpu](https://github.com/pytorch/pytorch/pull/90771), which means it doesn't convert to and from float. I found that in the tests for Resize V2 in torch/vision they used atol=1 when [comparing between the two implementations](https://github.com/pytorch/vision/blob/be7cdf16de5f1d2eee52c73dc99298cc5daeb805/test/test_transforms_v2.py#L711). However, there still seems to be a large discrepancy between the outputs in the CI tests, so I'll see if I can resolve this.

geohot:
Does torch perfectly match PIL? If so, I want ours to match too

abeleinin:
No, [torch and PIL produce different results](https://github.com/pytorch/vision/issues/2950#issuecomment-1163720072), but torch implementation has the `antialias` parameter to produce similar results to PIL.

abeleinin:
`nearest` and `nearest-exact` match torch exactly now, and I cleaned up the onnx implementation to use the same gather-based interpolation from `tensor.py`. As mentioned above, bilinear uint8 support isn't the same between torch and PIL, I've been getting around 12 MAE on average. Outside of NV and LLVM our difference between torch is less than 1 for bilinear uint8. I've been trying to debug the LLVM kernels, but haven't solved the issue yet.

abeleinin:
bilinear uint8 now works on all devices when `atol=1`

geohot:
Why is there an atol? What's causing the difference?

abeleinin:
The difference is caused by casting the input to a float and then recasting to uint8 on return. It isn't an off-by-one error, but instead, it's a difference in precision between using floats to do the linear interpolation vs using ints.

geohot:
Can it be made to match torch? I'm asking the same stuff on #6087

chenyuxyz:
#6308

PR #6097, opened 2024-08-15
Lowerer Multireduce Grouping

0xtimmy:
This PR is another part #5515 which allows grouping for multireduce kernels.
- kernel changes to allow grouping with multiple reduces 
- changes `fixup_ast `so that the shape of each temp buffer is based on dims for it's reduce
- changes `local_loads `and `group_for_reduces `calculation in lowerer to account for multiple local reductions
- changes `_to_uop` in lowerer to handle the new "store_back" and "load_back" cases where the value from a local reduction is stored & loaded back into each thread
- changes `create_gate` in uopgraph to stop recursion when it encounters a barrier, this is to account for the following case
-  changes sanity checks to recurse through the buffer calculation on PTX; since it will add the index to the local buffer pointer it's not always stored to a `DEFINE_LOCAL` sometimes is an `ALU` (with a `DEFINE_LOCAL` in `src`)
```
  for (int ridx0 = 0; ridx0 < 2; ridx0++) { /* first reduction */ }
  temp1[lidx1] = acc0;
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu0) {
    for (int ridx1 = 0; ridx1 < 16; ridx1++) { /* first reduction pt 2*/ }
    temp1[0] = acc1;
  }
  barrier(CLK_LOCAL_MEM_FENCE);
  float val2 = temp1[0]; // <- recurses! if we don't cut of recursion at the barrier it will gate this load
  for (int ridx2 = 0; ridx2 < 2; ridx2++) { /* second reduction */ } // <-recurses
  temp2[lidx0] = acc2;  // <- recurses
  barrier(CLK_LOCAL_MEM_FENCE); // <- recurses
  if (alu0) { // <- recurses
    for (int ridx3 = 0; ridx3 < 16; ridx3++) { /* second reduction pt 2 */ } // <- recurses
    data0[0] = acc3; // global store (gated)
  }
``` 

this has a bug! because the two local reduces operate on different dimensions (so that there are no expands / implicit shape changes) they use two different lidxs. This means we're running double the thread groups.

I have a proposed fix in PR which will try to reuse lidxs from other local reduces


tinygrad/codegen/kernel.py
```diff
@@ -704,12 +704,19 @@ def fix_st(warp_dims, tcd_dims, tcd_expand, pattern_1, pattern_2, st1):
...
-            (1,) * (self.first_upcast - self.group_for_reduces - self.first_reduce) + tuple([x[0] for x in self.upcasted_axis(0)])
-          local_buffer = MemBuffer(-1, start.dtype, ShapeTracker.from_shape(local_shape))
+          arg = tuple(i for i in range(self.first_reduce, self.first_reduce+self.group_for_reduces) \
+                      if self.sts[reduce_idx].shape[i] != self.sts[reduce_idx+1].shape[i])
+          if len(arg) == 0: return start
+          local_shape = (1,) * self.global_dims + self.full_shape[self.global_dims:self.global_dims+self.local_dims] + \
+            tuple([self.full_shape[i] if self.sts[reduce_idx].shape[i] != self.sts[reduce_idx+1].shape[i] else 1 \
+              for i in range(self.first_reduce, self.first_reduce+self.group_for_reduces)]) + \
+            (1,) * (self.shape_len - self.upcasted - self.group_for_reduces - self.first_reduce) + tuple([x[0] for x in self.upcasted_axis(0)])
+          local_buffer = MemBuffer(-self.reduceops.index(op)-1, start.dtype, ShapeTracker.from_shape(local_shape))
```
Qazalin, line 714: MemBuffer should definitely be pointer dtype

tinygrad/codegen/kernel.py
```diff
@@ -704,12 +704,19 @@ def fix_st(warp_dims, tcd_dims, tcd_expand, pattern_1, pattern_2, st1):
           return LazyOp(op.op, (ret,), new_reduce_axes) if (new_reduce_axes:=tuple(i for i in arg if i-self.first_upcast not in reduce_axes)) else ret
         if self.group_for_reduces:
           start = LazyOp(op.op, tuple(fixup_ast(x, apply_to_st) for x in op.src), arg)
-          local_shape = (1,) * self.global_dims + self.full_shape[self.global_dims:self.global_dims+self.local_dims+self.group_for_reduces] + \
-            (1,) * (self.first_upcast - self.group_for_reduces - self.first_reduce) + tuple([x[0] for x in self.upcasted_axis(0)])
-          local_buffer = MemBuffer(-1, start.dtype, ShapeTracker.from_shape(local_shape))
+          arg = tuple(i for i in range(self.first_reduce, self.first_reduce+self.group_for_reduces) \
+                      if self.sts[reduce_idx].shape[i] != self.sts[reduce_idx+1].shape[i])
+          if len(arg) == 0: return start
```
Qazalin, line 709: can you add a comment for this? what does 0 axis mean?

tinygrad/codegen/kernel.py
Qazalin, line 714: why do you need multiple temps for a back to back reduce?

tinygrad/codegen/lowerer.py
```diff
@@ -115,10 +115,11 @@ def lower(self, ast:LazyOp, opts:Renderer) -> UOp:
     first_upcasted = len(full_shape)-ki.upcasted
     # if there's no reduce, this is first_upcasted
     first_reduce = [x!=y for x,y in zip(ast.src[0].arg.st.shape[:first_upcasted]+(0,), full_shape[:first_upcasted]+(1,))].index(True)
-    local_loads = [x for x in ast.parents if x.op is BufferOps.LOAD and x.arg.idx == -1]
+    local_loads = [x for x in ast.parents if x.op is BufferOps.LOAD and x.arg.idx < 0]
```
Qazalin, line 118: Feel free to put this whole refactor of == -1 or < 0 in a different diff!

tinygrad/codegen/lowerer.py
```diff
@@ -115,10 +115,11 @@ def lower(self, ast:LazyOp, opts:Renderer) -> UOp:
...
     # NOTE: this is taking the first one...there may be subtlelies here with multireduces
```
Qazalin, line 119: can you change this comment if it no longer applies and describe what it's doing now?

tinygrad/codegen/lowerer.py
```diff
@@ -180,9 +181,24 @@ def _to_uop(self, x:LazyOp) -> UOp:
...
+            x.src[0].src[0].src[0].op is BufferOps.LOAD and x.src[0].src[0].src[0].arg.idx < 0
+        if load_back:
+          idx, _ = st_to_uops(x.arg.st, [UOp.const(u.dtype, 0) if i in x.src[0].src[0].arg else u for i,u in enumerate(self.ridxs)], x.arg.dtype)
+          return UOp(UOps.LOAD, load_dtype, (buf, idx) + ((UOp.const(load_dtype, 0), valid) if has_valid else ()) + barrier)
         return UOp(UOps.LOAD, load_dtype, (buf, idx) + ((UOp.const(load_dtype, 0), valid) if has_valid else ()) + barrier)
       # NOTE: only store the local reduceop in the first thread (this is wrong for non group for reduces!)
-      if x.arg.idx >= 0:
+      store_back = \
+        x.arg.idx < 0 and \
+        x.src[0].op in [ReduceOps.SUM, ReduceOps.MAX] and \
```
Qazalin, line 196:
```suggestion
        x.src[0].op in ReduceOps and \
```

tinygrad/codegen/lowerer.py
```diff
@@ -180,9 +181,24 @@ def _to_uop(self, x:LazyOp) -> UOp:
...
+        x.src[0].src[0].op is BufferOps.LOAD and \
+        x.src[0].src[0].arg.idx < 0
+      zero = UOp(op=UOps.CONST, dtype=dtypes.int32, src=(), arg=0)
+      if store_back: idx, _ = st_to_uops(x.arg.st, [zero if i in x.src[0].arg else u for i,u in enumerate(self.idxs)], x.arg.dtype)
```
Qazalin, line 200: This index override could use a comment too.

tinygrad/codegen/kernel.py
0xtimmy: yes I can, though I want your opinion a bit: this happens if one reduce is grouped and the other is not, thus `group_for_reduces` is true but there's nothing to group for this reduce.

should this be treated as a mistake? the highlighted code can handle it but it should be discouraged

tinygrad/codegen/kernel.py
0xtimmy: We could put it in one temp; that would just require an extra barrier. Would you prefer that?

tinygrad/codegen/lowerer.py
0xtimmy: This is intentional, we don't want `Reduceops.WMMA`.
I can play with it though, this might be redundant with the other parts of the expression

Qazalin: but we don't have double WMMA do we? Are you asserting this?

tinygrad/codegen/kernel.py
Qazalin: I've seen llm.c do the one buffer per reduceop, I think it's fine yea

0xtimmy: we'd need a way to do:
```
UOps.LOAD
UOps.BARRIER
UOps.STORE
```
whereas right now we only have a way to do
```
UOps.STORE
UOps.BARRIER
UOps.LOAD
```

tinygrad/codegen/kernel.py
Qazalin: hmm, I see this only happens in indexing kernels.
The reduceop in indexing is special, it's an arange we fold into the other reduce (kind of like merging the two reduceops/folding the arange inside the reduceop). Can we just skip group for reduce in these?

0xtimmy: this can happen if you group one reduce axis and not the other. 

`test_partial_opt_multireduce` in` test_linearizer.py:267` makes sure these cases are correct, but there's no good reason someone would want to do that

0xtimmy:
these commits update the PR with: new changes from master & multi-buffer strategy for multireduce

0xtimmy:
It's missing changes to `hand_coded_optimizations`, I can include those here or leave them in a separate PR

Qazalin:
Looks good! let's move forward with this diff after benchmarks are green.

Qazalin:
This is making llama somewhat slower:
image
and mixtral too:
image

I suspect it's the compiler because you change from temp1 to temp0.

Qazalin:
yea it was the compiler
image

image

0xtimmy:
I may have forgotten to mention it but this PR will generate slower kernels because it uses separate `lidx`s for each reduce, making it really inefficient. #6098 is meant to fix this

Qazalin:
cool, I'll review that. @0xtimmy what are the other (correctness) blockers?

PR #6098, opened 2024-08-15
Lowerer lidx reuse

0xtimmy:
Another part of #5515. This PR extends #6097 by adding "lidx reuse" basically it searches the reduce lidxs for duplicates across reduces and then opts to use those.

Ex:
```
__kernel void r_16_16_2_2(__global float* data0, const __global float* data1) {
  __attribute__ ((aligned (16))) __local float temp1[16];
  __attribute__ ((aligned (16))) __local float temp2[16];
  // two lidxs
  int lidx0 = get_local_id(0); /* 16 */
  int lidx1 = get_local_id(1); /* 16 */
  bool alu0 = ((!(bool)(lidx0))&(!(bool)(lidx1)));
  for (int ridx0 = 0; ridx0 < 2; ridx0++) { /* reduction one */ }
  temp1[lidx1] = acc0; // <--- uses lidx1
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu0) {
    for (int ridx1 = 0; ridx1 < 16; ridx1++) { /* reduction one part two */ }
    temp1[0] = acc1;
  }
  barrier(CLK_LOCAL_MEM_FENCE);
  float val2 = temp1[0];
  float acc2 = 0.0f;
  for (int ridx2 = 0; ridx2 < 2; ridx2++) { /* reduction two */ }
  temp2[lidx0] = acc2; // <--- uses lidx0
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu0) {
    for (int ridx3 = 0; ridx3 < 16; ridx3++) { /* reduction two part two */ }
    data0[0] = acc3;
  }
}
```
Becomes:
```
__kernel void r_16_16_2_2(__global float* data0, const __global float* data1) {
  __attribute__ ((aligned (16))) __local float temp1[16];
  __attribute__ ((aligned (16))) __local float temp2[16];
  // one lidx
  int lidx0 = get_local_id(0); /* 16 */
  int alu0 = (lidx0*2);
  bool alu1 = (!(bool)(lidx0));
  bool alu2 = (alu1&alu1);
  for (int ridx0 = 0; ridx0 < 2; ridx0++) { /* recuduction one */ }
  temp1[lidx0] = acc0; // <--- both use lidx0
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu2) {
    for (int ridx1 = 0; ridx1 < 16; ridx1++) { /* reduction one part two */ }
    temp1[0] = acc1;
  }
  barrier(CLK_LOCAL_MEM_FENCE);
  float val2 = temp1[0];
  for (int ridx2 = 0; ridx2 < 2; ridx2++) { /* reduction two */ }
  temp2[lidx0] = acc2; // <--- both use lidx0
  barrier(CLK_LOCAL_MEM_FENCE);
  if (alu2) {
    for (int ridx3 = 0; ridx3 < 16; ridx3++) { /* reduction two part two */ }
    data0[0] = acc3;
  }
}
```


0xtimmy:
this has the grouping changes included to show that the tests pass, but can be merged independently as well

Qazalin:
hmm, I'm getting:
```
test/test_linearizer.py::TestLinearizer::test_double_reduce_multireduce - RuntimeError: OpenCL Error -54: CL_INVALID_WORK_GROUP_SIZE
```
running on M1 with GPU=1

Qazalin:
so to me this speedup doesn't look worth the complexity. If you really want this, can you think of the simplest way to express it?
IIRC this doesn't impact correctness.

0xtimmy:
I think the best way would be to force GROUP and GROUPTOP to be the same across reductions; I really see no reason why someone would want otherwise

that would reduce some of the logic required to reuse indexes

the issue is that have two reduces creates two sets of local indexes, so if we group a reduce by 16, it would be 16^2 threads for two reduces (when we only need 16)

Qazalin:
I'd continue iterating on a simpler fix.
Maybe this can be a graph rewrite rule, like `fold_reduce_dims` - or even an OptOp if that level of abstractions makes sense.
(ofc, if Lowerer is the best place to put this that's cool too)

I think the current diff is complicated, it reminds me of our old Linearizer class. We have abstractions in uopgraph to handle UOp replacement, can we use any of those?

0xtimmy:
I'll keep working on making it simpler. I put it in the lowerer because that's after verifying the ast and before generating the `lidxs`
We want it after the verify because we just want to reuse the indices, we don't want to do anything funny to the shape.
And we want it before generating the `lidxs` because we don't want to run the rest of the flow with the final `lidxs`

0xtimmy:
Some more refactoring to try and make it more readable. The changes can be compressed for line length if need be
++ fixing PTX

test/test_linearizer.py
```diff
@@ -270,31 +274,39 @@ def test_double_reduce_multireduce(self):
...
+      opts=[[Opt(OptOps.GROUPTOP, 0, 4), Opt(OptOps.GROUPTOP, 1, 2), Opt(OptOps.GROUPTOP, 2, 4), Opt(OptOps.GROUPTOP, 3, 8)]])
 
   @unittest.skipIf(CI and Device.DEFAULT in {"AMD"}, "AMD CI doesn't support multiple sync threads yet")
   @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
   @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_shared, "test requires shared")
   def test_partial_opt_multireduce(self):
     # check how it works with one reduce optimized and one unoptimized
     Tensor.manual_seed(0)
-    x = Tensor.randn(27, 15, 5, dtype=dtypes.float).softmax(1).realize()
+    x = Tensor.randn(27, 12, 5, dtype=dtypes.float).softmax(1).realize()
```
Qazalin, line 293: why did this change?

tinygrad/codegen/lowerer.py
```diff
@@ -58,8 +68,15 @@ def lower(self, ast:UOp, opts:Renderer) -> UOp:
         self.idxs = get_grouped_dims("idx", full_shape[:global_dims], opts.global_max, reverse=True)
       else:
         # define indexes for GPU-like execution
-        self.idxs = get_grouped_dims("gidx", full_shape[:global_dims], opts.global_max, reverse=True) + \
-                    get_grouped_dims("lidx", full_shape[global_dims:first_reduce+group_for_reduces], opts.local_max)
+        reduceops = [x for x in ast.parents if x.op is UOps.REDUCE_AXIS and all(i < first_reduce+group_for_reduces for i in x.arg[1])]
+        if len(reduceops) > 0 and not any(r.op is UOps.WMMA for r in reduceops):
```
Qazalin, line 72: we should have an assert there's no double WMMA.

tinygrad/codegen/kernel.py
```diff
@@ -511,7 +511,8 @@ def has_expanded_axis(shape, strides): return any(s > 1 and st == 0 for s,st in
         for sz in ([256, 16] if prod(self.sts[0].shape[:self.first_reduce]) <= 32 else [16]):
           if all(st.shape[self.first_reduce] % sz == 0 or st.shape[self.first_reduce] == 1 for st in self.sts):
             try: # may fail due to excessive smem usage
-              self.apply_opt(Opt(OptOps.GROUPTOP, 0, sz))
+              if len(self.reduceops) < 2:
```
Qazalin, line 514: but nothing breaks?
image

tinygrad/codegen/lowerer.py
```diff
@@ -34,6 +34,16 @@ def get_grouped_dims(prefix, dims:Tuple[sint, ...], max_sizes:Optional[Tuple[int
           idx //= dims[c]
   return ret[::-1] if reverse else ret
 
+# gets the max of the full shape for each
+def get_reduce_dims(reduces: List[List], full_shape: Tuple) -> Tuple[Tuple[int, ...], List[int]]:
```
Qazalin, line 38: bad spacing in param:type

test/test_linearizer.py
0xtimmy: I was playing with different combinations of divisors, I can change it back

tinygrad/codegen/kernel.py
Qazalin: I don't understand, you pushed something that removed it, why did it exist in the first place?

0xtimmy: I started working on this earlier; it was meant to prevent anything from sneaking through `hand_coded_optimizations` while I was still testing.

I meant to delete it, the cases that were causing trouble are covered by `test_partial_opt_multireduce`

tinygrad/codegen/uopgraph.py
```diff
@@ -312,6 +312,8 @@ def index_collapse(idx,rng,buf,add,mul,ld,reduce):
   ((NOp.var("x") / NOp.var("x2")) / NOp.var("x3"), lambda x,x2,x3: x/(x2*x3)), # (x/x2)/x3 -> x/(x2*x3)
   (-(NOp.var("x") + NOp.var("y")), lambda x,y: (-x)+(-y)),  # -(x+y) -> -x + -y
   ((NOp.cvar("c0") + NOp.var("x")).lt(NOp.cvar("c1")), lambda x,c0,c1: UOp.lt(x, c1-c0)),  # c0 + x < c1 -> x < c1 - c0
+  ((NOp.var("x") & NOp.var("x")), lambda x: x),
+  ((NOp.var("x") | NOp.var("x")), lambda x: x),
```
Qazalin, line 316: can be its own diff.

test/test_linearizer.py
```diff
@@ -195,6 +197,8 @@ def test_mid_dim_multireduce(self):
     wanna_output = (x.numpy()-x.numpy().sum(axis=1, keepdims=True)).sum(axis=1).reshape(27,1,1,5)
     lins = helper_linearizer_ast(sink, [x], wanna_output=[wanna_output], opts=opts)
     for l in lins:
+      if Device.DEFAULT not in {"METAL"}:
```
Qazalin, line 200: why is METAL=1 different? is this the global_max thing?

0xtimmy: it was because I had reuse disabled if `opts` had `tensor_cores`. I'll check if we still need it

0xtimmy: seems to work, pushing a diff

tinygrad/codegen/lowerer.py
```diff
@@ -58,8 +70,15 @@ def lower(self, ast:UOp, opts:Renderer) -> UOp:
         self.idxs = get_grouped_dims("idx", full_shape[:global_dims], opts.global_max, reverse=True)
       else:
         # define indexes for GPU-like execution
-        self.idxs = get_grouped_dims("gidx", full_shape[:global_dims], opts.global_max, reverse=True) + \
-                    get_grouped_dims("lidx", full_shape[global_dims:first_reduce+group_for_reduces], opts.local_max)
+        reduceops = [x for x in ast.parents if x.op is UOps.REDUCE_AXIS and all(i < first_reduce+group_for_reduces for i in x.arg[1])]
+        if len(reduceops) > 0:
```
Qazalin, line 74: why is this counting local reduces?

test/test_linearizer.py
Qazalin: it's not the job of the reviewer to tell you these, when you open a diff for review you must read it 3 times before and question every line yourself.
I can't really trust this diff to be complete enough to merge it.

tinygrad/codegen/lowerer.py
0xtimmy: this is to determine whether we try to merge the `lidx`s used for local reduces, so it should only apply if there are local reduces and it shouldn't apply to reduces that before `ridx`s

Qazalin: even if there's only 1 local reduceop?
ref: test_arange_big

0xtimmy: I figured `len(reduceops) > 1` is tighter but `len(reduceops) > 0` is simpler (are there local reducops or not?)
both are correct and I'm happy to change it to `len(reduceops) > 1`

tinygrad/codegen/uopgraph.py
0xtimmy: removing these lines, redundant re: #6412

tinygrad/codegen/lowerer.py
```diff
@@ -34,6 +34,18 @@ def get_grouped_dims(prefix, dims:Tuple[sint, ...], max_sizes:Optional[Tuple[int
           idx //= dims[c]
   return ret[::-1] if reverse else ret
 
+# gets the max of the full shape for each
+# produces a new shape with duplicate reduce dimensions removed
+def get_reduce_dims(reduces:List[List], full_shape:Tuple) -> Tuple[Tuple[int, ...], List[int]]:
+  permute = list(range(len(full_shape)))
+  for ax in range(max(len(r) for r in reduces)):
+    for idx in (idxs:=[r[ax] for r in reduces if ax < len(r)]):
+      permute[idx] = next(i for i in idxs if full_shape[i] == full_shape[idx])
```
0xtimmy, line 43: this seemed like a more readable way to find reusable local dims

tinygrad/codegen/lowerer.py
```diff
@@ -34,6 +34,18 @@ def get_grouped_dims(prefix, dims:Tuple[sint, ...], max_sizes:Optional[Tuple[int
...
+
+  return functools.reduce(
+    lambda acc, x: ((*acc[0], x[1]), [acc[2] if x[0] == j else j for j in acc[1]], acc[2]+1) if x[0] in permute else acc,
+    enumerate(full_shape), (cast(Tuple[int, ...], tuple()), permute[:], 0))[:2]
```
0xtimmy, line 47: this removes the reused dims from the `full_shape` tuple and the adjusts the permute

it seemed like the best way to do it considering number of lines and code style

tinygrad/codegen/lowerer.py
```diff
@@ -58,8 +70,15 @@ def lower(self, ast:UOp, opts:Renderer) -> UOp:
         self.idxs = get_grouped_dims("idx", full_shape[:global_dims], opts.global_max, reverse=True)
       else:
         # define indexes for GPU-like execution
-        self.idxs = get_grouped_dims("gidx", full_shape[:global_dims], opts.global_max, reverse=True) + \
-                    get_grouped_dims("lidx", full_shape[global_dims:first_reduce+group_for_reduces], opts.local_max)
+        reduceops = [x for x in ast.parents if x.op is UOps.REDUCE_AXIS and all(i < first_reduce+group_for_reduces for i in x.arg[1])]
```
0xtimmy, line 73: this is `i < first_reduce+group_for_reduces` instead of `i >= first reduce and i < first_reduce+group_for_reduces` because there shouldn't be a `UOps.REDUCE_AXIS` on a dim `< first_reduce `

tinygrad/codegen/lowerer.py
```diff
@@ -58,8 +70,15 @@ def lower(self, ast:UOp, opts:Renderer) -> UOp:
...
+        if len(reduceops) > 1:
+          # try to reuse grouped dims across sequential reduceops
+          rdims, p = get_reduce_dims([[i-global_dims for i in r.arg[1]] for r in reduceops], full_shape[global_dims:first_reduce+group_for_reduces])
+          lidxs = get_grouped_dims("lidx", rdims, opts.local_max)
+          self.idxs = get_grouped_dims("gidx", full_shape[:global_dims], opts.global_max, reverse=True) + [lidxs[i] for i in p]
```
0xtimmy, line 78: here `get_grouped_dims` is called without the reused dims, then `p` assigns the `lidx`s to a place in the shape
if a dim is reused, the `lidx` will end up duplicated in `self.idxs`

that's how this PR uses the same `lidx` for sequential reduces

PR #6103, opened 2024-08-16
bilinear interp uint8 fails

geohot:


geohot:
#6088 #6087 interpolate should also work for uint8

PR #6104, opened 2024-08-16
move UOp add_chain, mod_folding, div_folding into cached methods

chenyuxyz:
is this faster?
[run_process_replay]


geohot:
Err, this is going to keep many uops around forever, and I don't even think the cache is gonna hit where you want

chenyuxyz:
yea no it's not helpful

PR #6105, opened 2024-08-16
einops rearrange

andredaprato:
Passes tests, but I'm sure it can be simplified much more, and the error handling/parsing can definitely be more rigorous.


andredaprato:
Closed in favor of #6106 which is much better.

PR #6106, opened 2024-08-16
Rearrange

samm393:


geohot:
Can you copy in all the tests from einops? How much code reuse can we get with our einsum?

samm393:
> Can you copy in all the tests from einops? (to a new file) How much code reuse can we get with our einsum?

All the rearrange tests from https://github.com/arogozhnikov/einops/blob/master/tests/test_ops.py that can be run as is have already been copied in.

There are a few more that test correctness of other backends against numpy - they could  be run too if I can import einops?

tinygrad/tensor.py
```diff
@@ -1631,6 +1631,41 @@ def argmin(self, axis=None, keepdim=False):
     """
     return (-self).argmax(axis=axis, keepdim=keepdim)
 
+  @staticmethod
```
geohot, line 1634: This should not be staticmethod. It shouldn't support a list of Tensor, only one tensor called `self`

geohot:
If that's all of them it's probably good on tests. Is there any place you think bugs might be?

samm393:
> If that's all of them it's probably good on tests. Is there any place you think bugs might be?

It's broken for things like "a b () -> b a" and "a -> () a ()". Looking into how to fix
Going to add the tests from https://github.com/arogozhnikov/einops/blob/master/tests/test_examples.py as well

tinygrad/tensor.py
```diff
@@ -1296,7 +1296,7 @@ def unflatten(self, dim:int, sizes:Tuple[int,...]):
     print(Tensor.ones(5, 12, 3).unflatten(-2, (2, 2, 3, 1, 1)).shape)
     ```
     """
-    dim = self._resolve_dim(dim)
+    dim = self._resolve_dim(dim) if resolve else dim
```
geohot, line 1299: This shouldn't be a flag

tinygrad/tensor.py
```diff
@@ -1266,7 +1266,7 @@ def transpose(self, dim0=1, dim1=0) -> Tensor:
     order[dim0], order[dim1] = order[dim1], order[dim0]
     return self.permute(order)
 
-  def flatten(self, start_dim=0, end_dim=-1):
+  def flatten(self, start_dim=0, end_dim=-1, resolve=True):
```
geohot, line 1269: Same thing here. Why would I ever use this? This is a public API

samm393:
I think it's ready to be looked at now
Have added:

-  Support for "()"
-   Same name convention as einops (has to be a valid python literal which also has no leading or trailing underscore)
-   Errors out if an incorrect dimension size is passed (even if it is not used)
-   Supports unicode ellipsis

mattismegevand:
Hey, I was working on this a bit earlier before I saw that you had a PR for this bounty.
I think we should cache necessary reshape and permutation given the input shape and pattern that way we could skip a bunch of redundant computation.

This is what they do in einops and PyTorch.
https://github.com/arogozhnikov/einops/blob/309337ddc9e9ae90e463dcba2dc4905b347eda37/einops/einops.py#L288
https://github.com/pytorch/pytorch/blob/e72e924eb5afccbd55dc6a9ecc5eeb2cbab47cb3/functorch/einops/rearrange.py#L23

geohot:
Congrats! Bounty is yours, can pay via PayPal or USDC. Also owe you $125 from the last one, IIRC

geohot:
Err, actually the first one I tried didn't work.

`img_0.rearrange("(bw bh) 1 c w h -> c (bw w) (bh h)", bw=4)`

```
  File "/Users/diane/tinygrad/tinygrad/tensor.py", line 1669, in rearrange
    assert sorted(lhs) == sorted(rhs) and len(lhs) == len(set(lhs)), f"name mismatch in {formula}"
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: name mismatch in (bw bh) 1 c w h -> c (bw w) (bh h)
```

samm393:
> Err, actually the first one I tried didn't work.
> 
> `img_0.rearrange("(bw bh) 1 c w h -> c (bw w) (bh h)", bw=4)`
> 
> ```
>   File "/Users/diane/tinygrad/tinygrad/tensor.py", line 1669, in rearrange
>     assert sorted(lhs) == sorted(rhs) and len(lhs) == len(set(lhs)), f"name mismatch in {formula}"
>                                           ^^^^^^^^^^^^^^^^^^^^^^^^^
> AssertionError: name mismatch in (bw bh) 1 c w h -> c (bw w) (bh h)
> ```
> 
> `rearrange(img_0, "(bw bh) 1 c w h -> c (bw w) (bh h)", bw=4)` is fine

Thats embarrassing. Didn't realize that was valid syntax. Fix proposed in #6184

PR #6112, opened 2024-08-16
Dsp progress

nimlgen:


nimlgen:
Some progress on lib removal:

currently ioctl hangs when it comes to invokes (which should setup the shell I think?). the lib calls invokes in 2 threads for some reason, but I hope we should be able to do that using just one

geohot:
Yea the threads are dumb. It should all be doable in the main thread, just don't use adsprpc lib

nimlgen:
Got it working without lib. The way it works is the following: when app asks to open a skel lib (`remote_handle64_open`) is just an invoke ioctl with some parameters, but it blocks it until all data is loaded. And that's why the lib uses the second listener thread to server dsp request to load ELFs (`listner_worker` in `run_3.py`).

Will look if we can just use only the main thread, but at least removing the lib we can not write elfs to disk and pass the, from memory.

tinygrad/runtime/support/elf.py
```diff
@@ -1,23 +1,31 @@
...
 
 @dataclass(frozen=True)
-class ElfSection: name:str; header:libc.Elf64_Shdr; content:bytes # noqa: E702
+class Elf64Section: name:str; header:libc.Elf64_Shdr; content:bytes # noqa: E702
 
-def elf_loader(blob:bytes, force_section_align:int=1) -> Tuple[memoryview, List[ElfSection], Any]:
+@dataclass(frozen=True)
+class Elf32Section: name:str; header:libc.Elf32_Shdr; content:bytes # noqa: E702
+
+def elf_loader(blob:bytes, force_section_align:int=1, elf32:bool=False) -> Tuple[memoryview, Union[List[Elf32Section], List[Elf64Section]], Any]:
```
geohot, line 12: elf loader needs some tests in general

tinygrad/runtime/ops_dsp.py
```diff
@@ -0,0 +1,93 @@
...
+
+  def _close_lib(self, libname):
+    if (ret:=adsp.remote_handle64_close(self.loaded_libs[libname])) != 0: raise RuntimeError(f"Failed to close library {libname}: {ret}")
+    self.loaded_libs.pop(libname)
+    self.lru_lib_order.remove(libname)
+
+  def _unregister_lib(self, libname):
+    if (h:=self.loaded_libs.get(libname)) is not None: self._close_lib(libname)
+
+  def _ensure_lib_opened(self, libname):
```
nimlgen, line 83: maybe just keep it simple? don't know if this LRU is practical at all, since we can have up to 128 libs opened

geohot:
Oh, there's a lib limit? Why is this still using adsprpc?

nimlgen:
Added `extra/dsp/invoke_bug.py` to repro the bug. Strange but it passes with `clang-12`, but fails with `clang-18` (but IIRC some tests pass with clang-18 but fail with clang-12)

nimlgen:
> Oh, there's a lib limit? Imposed by fastrpc?

Yeah opening too many libs returns -1. I have benchmarked around 119 libs can be opened at the same time. Yeah, looks like imposed by fastrpc.

> Why is this still using adsprpc?

Was debugging, revert it back.

nimlgen:
@geohot, what's you opinion: we still want to merge this into the core with the known bug? I am still looking into mprotect, but no luck yet.

geohot:
Let's get QCOM GPU merged first then revisit

PR #6121, opened 2024-08-17
real strides with uops

geohot:


chenyuxyz:
done in https://github.com/tinygrad/tinygrad/pull/6365

PR #6125, opened 2024-08-17
switch real_size to use uops [run_process_replay]

geohot:


geohot:
@chenyuxyz can you add support for vmin/vmax on div and mod that aren't by a const

PR #6138, opened 2024-08-17
re:6125 switch real_size to use uops [run_process_replay]

chenyuxyz:
change MOD range to take any positive UOp


geohot:
For axis_is_masked I had luck with 
```
+    _, ovalid = self.expr_idxs()
+    ret = f'idx{axis}' in [v.expr for v in ovalid.vars()]       
+    _, valid = self.to_indexed_uops()
+    test = axis in [x.arg for x in graph_rewrite(valid).sparents if x.op is UOps.RANGE]
+    if test != ret:                                                                           
+      print("MISMATCH", self, axis, test, ret)                                                
+      print(ovalid)                       
+      print(graph_rewrite(valid))
+      assert False       
+    return ret         
```
keeping both around and diffing them while I run examples

PR #6140, opened 2024-08-17
type verify intermediate UOps [run_process_replay]

Qazalin:
CONST, ALU, BITCAST and DEFINE_GLOBAL/DEFINE_LOCAL share specs, the rest are special.


Qazalin:
I don't like that is is calling type_verify again, it should be all in __post_init__

PR #6142, opened 2024-08-17
Barrier consecutive load-store

0xtimmy:
Pre-req for multireduce

This adds a few lines to uopgraph that will insert barriers in the event of consecutive load-stores to the same local buffer

This is needed for multireduce to prevent writing to a local buffer, before every thread loads the result of the previous group reduce. This lets us use only one local buffer.
Ex:
```
  ...
  temp1[lidx1] = acc0; // store result of first reduce
  barrier(CLK_LOCAL_MEM_FENCE); // barrier prevents the grouped reduce from starting before every thread stores their reduce
  if (alu0) {
    float acc1 = 0.0f;
    for (int ridx1 = 0; ridx1 < 16; ridx1++) { /* first grouped reduce */ }
    temp1[0] = acc1;
  }
  barrier(CLK_LOCAL_MEM_FENCE); // barrier prevents threads from loading the result of the grouped reduce before it's finished
  float val2 = temp1[0];
  float acc2 = 0.0f;
  for (int ridx2 = 0; ridx2 < 2; ridx2++) { /* second reduce */ }
  barrier(CLK_LOCAL_MEM_FENCE); // PR adds a barrier here, prevents threads from storing their second reduce
                                // before every thread has loaded the result of the grouped reduce
  temp1[lidx0] = acc2;
  barrier(CLK_LOCAL_MEM_FENCE); // barrier before second grouped reduce
  if (alu0) {
```

* doesn't apply within `UOps.IF` because we can't gate barriers

### DIFF
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -458,7 +458,7 @@ def create_gate(root:UOp) -> Optional[UOp]:
 
 def bar_load_store(root:UOp, buf:UOp, idx, val) -> Optional[UOp]:
   @functools.lru_cache(None)
-  def _bar_load(u:UOp) -> UOp:
+  def _bar_load(u:UOp) -> List[UOp]:
     if u.op is UOps.LOAD and u.src[0] is buf and u.src[2].op not in {UOps.IF}: return [u]
     if u.op is UOps.BARRIER or u.op is UOps.LOAD: return []
     return [l for x in u.src for l in _bar_load(x)]
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -454,11 +454,21 @@ def create_gate(root:UOp) -> Optional[UOp]:
   def _gate_srcs(u:UOp, gate:UOp) -> UOp:
     if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)
     return u if (replace_source:=tuple(_gate_srcs(x, gate) for x in u.src)) == u.src else UOp(u.op, u.dtype, replace_source, u.arg)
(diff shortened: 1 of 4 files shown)


Qazalin:
we really don't wanna do this in a toposort, can you try doing it with graph rewrite rules?

0xtimmy:
I agree, the trouble with the rewrite rules is that we'd have to put `Uops.BARRIER` is a `UOp.src` ; there didn't seem to be any good candidates

I do have a branch that will make a dependency chain LOAD -> BARRIER -> STORE, but it modifies how UOPs.STORE interprets it's src

0xtimmy:
this commit accomplished the barrier through rewrite, left the topsort changes commented out for reference

tinygrad/codegen/uopgraph.py
```diff
@@ -454,11 +454,21 @@ def create_gate(root:UOp) -> Optional[UOp]:
   def _gate_srcs(u:UOp, gate:UOp) -> UOp:
     if u.op is UOps.LOAD and u.src[-1].op is UOps.BARRIER: return UOp(u.op, u.dtype, u.src[:-1]+(UOp(UOps.IF, None, (gate, u.src[-1])),), u.arg)
     return u if (replace_source:=tuple(_gate_srcs(x, gate) for x in u.src)) == u.src else UOp(u.op, u.dtype, replace_source, u.arg)
-  return None if len(root.src) == 3 or (ret:=_gate_srcs(root, root.src[3])) is root else ret
+  return None if len(root.src) == 3 or root.src[3].op is UOps.BARRIER or (ret:=_gate_srcs(root, root.src[3])) is root else ret
+
+def bar_load_store(root:UOp, buf:UOp, idx, val) -> Optional[UOp]:
```
Qazalin, line 459: idx and val are unused, is this intentional?

Qazalin:
hmm so now src[3] can mean two different things... This does change the spec yea.
If this is the only way to do it, the multiple temps option makes more sense #6097.

0xtimmy:
sounds good, I'll update the main PR

Issue #5858, opened 2024-08-01
Proposal: supporting second derivatives


davidgonmar:
Second derivatives are useful on some cases (see https://github.com/tinygrad/tinygrad/pull/5701), but not possible by the way autodiff works in Tinygrad. Here is my take on how they could be supported with no expected extra cost:

Let's take scalar derivatives as an example (higher dimensional derivatives are pretty much the same in the implementation context).
A simple example is for torch is:
```python
def f(x): return x**3
t = Tensor(3.0, requires_grad=True)
fout = f(t) # Tensor(27.0)
fout.backward(retain_graph=True)
df_dt = t.grad # the derivative function is 3x ** 2, so derivative will be 81.0. This means 'how does the output of the function change when I change t
```
What happens here is that fout 'remembers' where it came from w.r.t its inputs (the leaf tensors marked as requires_grad). Now, if we wanted to do a second derivative in PyTorch, it would be like this
```python
t.grad = None # the graph will be 'backpropagated again', but we don't want it to affect the new calculation.
df_dt.backward() # second derivative is 6x, so derivative will be 18.0. This means 'how does the first derivative change when I change t'. This is why we are doing .backward() on the first derivative.
d2f_dt2 = t.grad
```
Now, what is happening here? df_dt knows where it came from w.r.t its inputs (t in this case)! (at least that is what happens in torch). However, this is how a derivative is created in tinygrad:
```python
class Neg(Function):
  def forward(self, x:LazyBuffer) -> LazyBuffer: return x.e(UnaryOps.NEG)
  def backward(self, grad_output:LazyBuffer) -> LazyBuffer: return grad_output.e(UnaryOps.NEG)
```
As you can see, the .backward() works with LazyBuffers, but they do not handle autodiff. So if we run the previous example in tinygrad, the second derivative will come out as 'None', because the first derivative does not know where it came from w.r.t its inputs (the _ctx of the grad is None). The solution here is, at first glance:
1 - Support retain_graph : easy, just do not delete the self._ctx if retain_graph = True. That way the graph information remains.
2 - The Function class needs to work at Tensor level for the backward function. Example:
```python
class Neg(Function):
  def forward(self, x:LazyBuffer) -> LazyBuffer: return x.e(UnaryOps.NEG)
  def backward(self, grad_output:Tensor) -> Tensor: return -grad_output
```
[...]

PR #6145, opened 2024-08-17
add support for retain_graph in backward

davidgonmar:
Prerequisite for https://github.com/tinygrad/tinygrad/issues/5858.
Simply allows not deleting the graph information when running backward (defaults to False). Most common usage is gradient accumulation or second order derivatives. Defaults to False as PyTorch.

Another important usage as I mentioned is gradient accumulation (you can look it up, it is a simple concept). See the test for the expected behaviour (to match pytorch). In order for it to be correct, the gradient accumulation needs to only be done on leaf tensors (hence the t0.grad = None in .backward(). In eager frameworks like PyTorch, this reduces memory usage significantly. It could also reduce it on Tinygrad for training (but I'm not 100% sure it will). It simply frees grads that will not be used).

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -773,7 +773,7 @@ class Tensor:
         if g is not None and t.requires_grad:
           assert g.shape == t.shape, f"grad shape must match tensor shape, {g.shape!r} != {t.shape!r}"
           t.grad = g if t.grad is None else (t.grad + g)
-      if len(t0._ctx.parents) is not 0: t0.grad = None # matches PyTorch behaviour. Do not accum grad if t0 is not a leaf
+      if len(t0._ctx.parents) != 0: t0.grad = None # matches PyTorch behaviour. Do not accum grad if t0 is not a leaf
       if not retain_graph: del t0._ctx
     return self
 
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -762,11 +762,11 @@ class Tensor:
 
     assert self.shape == gradient.shape, f"grad shape must match tensor shape, {gradient.shape!r} != {self.shape!r}"
     self.grad = gradient
-
     for t0 in reversed(self._deepwalk()):
       if t0.grad is None: raise RuntimeError(f"tensor {t0} has no grad")
       token = _METADATA.set(dataclasses.replace(md, backward=True) if (md := t0._ctx.metadata) is not None else None)
       grads = t0._ctx.backward(t0.grad.lazydata)
+      t0.grad = None # we do not need the grad anymore. This matches PyTorch behaviour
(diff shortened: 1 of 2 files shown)


davidgonmar:
Update:
when you do this in torch
```
import torch
import numpy as np

# Define the size N
N = 10  # You can change N to the desired value

# Create a tensor with shape (N*N,) filled with 60000, half-precision, and requires gradient
t = torch.tensor([60000] * (N * N), dtype=torch.float16, requires_grad=True).reshape(N, N)

# Check if the mean of the tensor is 60000
np.testing.assert_allclose(t.mean().item(), 60000)

# Compute the mean of the square of the tensor and perform backpropagation
t.square().mean().backward()

# Check the gradients
np.testing.assert_allclose(t.grad.numpy().flatten(), [60000 * 2 / (N * N)] * (N * N))
```
it will throw since t.grad is None. Why? because t is not a leaf. In tinygrad, a lot of tests rely on the fact that intermediate gradients are kept. This will probably increase memory usage in those cases since intermediate tensors that are present in the Python scope will 'maintain' their grads, even though they are useless. What PyTorch does is t.grad = None just after calling backward. The only grads that are kept are the ones for leaf tensors. I think what PyTorch does is more 'scalable'. In order to emulate it, what I did is, after calling .backward a second time (assuming the tensor graph is kept), before running the actual .backward, I set all non-leaf graphs to None. This is necessary to allow gradient accumulation and second order derivatives (from the user perspective). Ideally, I think this should be changed to match the pytorch behaviour (and some tests changed not to expect grads on non-leaf tensors). If this seems good, I can refactor everything later after merging this.

tinygrad/tensor.py
```diff
@@ -737,23 +737,26 @@ def multinomial(self:Tensor, num_samples:int = 1, replacement:bool = False) -> T
...
     If the 'gradient' argument is not provided, the tensor must be a scalar, and the gradient is implicitly set to 1.0.
-
+    If 'retain_graph' is false, the graph used to compute the grads will be freed. Otherwise, it will be kept. Keeping it can increase memory usage.
     ```python exec="true" source="above" session="tensor" result="python"
     t = Tensor([1.0, 2.0, 3.0, 4.0], requires_grad=True)
     t.sum().backward()
     print(t.grad.numpy())
     ```
     """
+    toposorted = self._deepwalk()
```
geohot, line 759: Why does this have to move?

davidgonmar: > Why does this have to move?

On the 'Update comment' I mentioned the details of what I want to achieve. I need to call the deepwalk before assigning the gradient to the root tensor, since the deepwalk resets the grads. If I reset it after setting it, then the root grad is 'None' and that's not what we want.

PR #6160, opened 2024-08-18
single arange fusion with graph rewrite

Qazalin:
trains beautiful_mnist with graph rewrite fusion:

```
FUSE_ARANGE=1 AST_REWRITE=1 python3 examples/beautiful_mnist.py
```


Qazalin:
so this is still slow
image
maybe I can cache the UOp's shapetracker with full_shape.

PR #6161, opened 2024-08-18
always use `(({a}>{b})?{a}:{b})` for cstyle max

chenyuxyz:
fix max not defined for dtypes like uint8.


Qazalin:
fyi you can also remove this diff's changes #5840

chenyuxyz:
it's kinda dumb if `a` and `b` are computed twice, maybe we just rewrite max to lt and where

geohot:
Really the transcendental thing should be made more generic, and if there's no BinaryOps.MAX it uses the rewrite rule

tinygrad/renderer/cstyle.py
ignaciosica, line 1: won't let me comment in the right location because is far away from changes but max can also be removed from AMDRenderer kernel_prefix

chenyuxyz:
will at least make is a rewrite rule like #6162

PR #6163, opened 2024-08-18
helpers: Add gzip support

m-xyz:
Add gunzip support to the fetch function.


geohot:
Bad spacing, no tests

PR #6166, opened 2024-08-19
(tensors) feat: add pathlib support to Tensors

stefpi:
cherry picked pathlib support from https://github.com/tinygrad/tinygrad/pull/6146 and added regression test for it in tests/test_tensor_data.py.

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -1,7 +1,7 @@
 import subprocess
 import numpy as np
 import torch
-import unittest, copy, mmap, random, math
+import unittest, copy, mmap, random, math, pathlib, tempfile
 from tinygrad import Tensor, Device, dtypes
 from tinygrad.engine.schedule import create_schedule
 from tinygrad.helpers import getenv, temp, CI, _METADATA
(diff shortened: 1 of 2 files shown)


test/test_tensor_data.py
```diff
@@ -64,5 +64,12 @@ def test_data_float16(self):
     assert dat.shape == (2,2)
     # NOTE: python can't deref float16
 
+  def test_data_pathlib(self):
```
geohot, line 67: This is not a good test for various reasons. It also doesn't test all the functionality

stefpi: :+1: I'll try again. What would be the ideal spot to put it in your opinion?

geohot: http://www.catb.org/~esr/faqs/smart-questions.html

geohot: Have you read all the other tests?

stefpi: Apparently didn't read enough of them. I'll go find a better spot.

stefpi: Moved to /test/test_tensor.py. Also made some changes to the test to improve cleanup. Based on the README of the CIFAR dataset, I moved the test to be more focused on numpy arrays that are stored in bin files. Hope this is more suitable. Hope this is an improvement, if not, please let me know. I would like to keep trying.

PR #6167, opened 2024-08-19
gunzip support

cameronslee:
adds gunzip support to fetch from #6146.

regression test in tests/unit/test_helpers.py checks that the fetched file is no longer a compressed .gz file.
gzip module compares bytes(magic number) of the file and raises OSError exception if they do not match.

with fetch tests enabled:
Screenshot 2024-08-18 at 6 22 49 PM

### DIFF
--- a/test/unit/test_helpers.py
+++ b/test/unit/test_helpers.py
@@ -1,7 +1,7 @@
 import unittest
 from PIL import Image
 from tinygrad.helpers import Context, ContextVar
-from tinygrad.helpers import merge_dicts, strip_parens, prod, round_up, fetch, fully_flatten, from_mv, to_mv, get_contraction, get_shape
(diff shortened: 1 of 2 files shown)


test/unit/test_helpers.py
```diff
@@ -161,6 +161,12 @@ def test_fetch_subdir(self):
       assert pimg.size == (77, 77), pimg.size
     assert img.parent.name == "images"
 
+  def test_fetch_gunzip(self):
+    with self.assertRaises(OSError):
+      with open(fetch("https://github.com/tinygrad/tinygrad/archive/refs/tags/v0.9.2.tar.gz", allow_caching=False, gunzip=True), 'rb') as f:
+        with gzip.GzipFile(fileobj=f) as gz:
```
geohot, line 167: huh? what do you think gunzip does?

PR #6168, opened 2024-08-19
bugfix for print at DEBUG>=3 with regression test

Zaffer:


Zaffer:
first bugfix/improvement for #6146

PR #6170, opened 2024-08-19
Bugfix dont print get_kernal timings when DEBUG < 3

Zaffer:
This implements the bugfix that @geohot did on pull request #6146 
Test is a regression test for the DEBUG level not being 1.


Zaffer:
@chenyuxyz all test passing now.

chenyuxyz:
it's not about test failing, the test does not make sense

PR #6171, opened 2024-08-19
add cross entropy and std_mean

baolong281:
Adds cross entropy and std_mean from #6146 with regression tests


geohot:
tests are way too verbose

PR #6174, opened 2024-08-19
Gzip in fetch

reddyn12:
Use hash() to save a line
Scale progress bar length by 4 when ```gunzip=True``` for more acurate times


geohot:
Ugh, this introduces a serious bug. https://stackoverflow.com/questions/34330284/what-is-the-risk-of-collisions-when-relying-on-python-hash-function

You are no longer eligible for bounties.

PR #6175, opened 2024-08-19
Add gunzip to fetch with assorted tests

nextgens:
Add gunzip to fetch with assorted tests (not a gzip file, default behaviour of fetch, correct size of output for a valid file)


geohot:
Close in favor of #6176, that PR has the right test

PR #6176, opened 2024-08-19
Support `gunzip` in `fetch`

eitanturok:


nextgens:
IMHO the goal is to test that gzip is enabled/working and that the API is stable rather than test gzip itself

That's the approach I went for in #6175

tinygrad/helpers.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import annotations
-import os, functools, platform, time, re, contextlib, operator, hashlib, pickle, sqlite3, tempfile, pathlib, string, ctypes, sys
+import contextlib, ctypes, functools, gzip, hashlib, operator, os, pathlib, pickle, platform, re, sqlite3, string, sys, tempfile, time
```
geohot, line 2: Why did this change?

test/unit/test_helpers.py
```diff
@@ -161,6 +161,24 @@ def test_fetch_subdir(self):
       assert pimg.size == (77, 77), pimg.size
     assert img.parent.name == "images"
 
+  def test_fetch_gunzip_valid(self):
```
geohot, line 164: Thank you, finally the correct test

tinygrad/helpers.py
eitanturok: I sorted the imports alphabetically and added `gzip`.

chenyuxyz: but why?

eitanturok: I usually sort my imports.

LMK if you want me to do undo it.

geohot:
email george@tinygrad.org with a PayPal or USDC (on ethereum) to claim $50

PR #6179, opened 2024-08-19
added pathlib support for tensors with tests

kaushal07wick:


kaushal07wick:
#6146

geohot:
Tests should be in test_disk_tensor, can you work with #6180 and I'll give you each a microbounty if tar and pathlib get in?

PR #6182, opened 2024-08-19
default num_classes value for one_hot

Onlinerocker:
If `num_classes` is set to -1, the number of classes will be inferred as one greater than the largest class value in the input tensor.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -1658,7 +1658,8 @@ class Tensor:
     """
     def parse_formula(formula: str):
       lparens, rparens = map(lambda x: [i for i, ch in enumerate(formula.split()) if ch == x], ("(", ")"))
-      assert len(lparens) == len(rparens) and sorted(flatten(pairs := list(zip(lparens, rparens)))) == flatten(pairs), "bracket mismatch"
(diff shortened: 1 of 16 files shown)


tinygrad/tensor.py
```diff
@@ -2925,15 +2925,18 @@ def dropout(self, p=0.5) -> Tensor:
     if not Tensor.training or p == 0: return self
     return self * (Tensor.rand(*self.shape, requires_grad=False, dtype=dtypes.default_float, device=self.device) >= p) * (1/(1.0 - p))
 
-  def one_hot(self, num_classes:int) -> Tensor:
+  def one_hot(self, num_classes:int=-1) -> Tensor:
     """
     Converts `self` to a one-hot tensor.
 
+    'num_classes' defaults to -1, which means num_classes will be inferred as max(self) + 1.
```
geohot, line 2932: Why ' and not `

Onlinerocker: typo. updated to `

geohot:
Merged, actually the easiest $50! email george@tinygrad.org with a PayPal or USDC (on ethereum) to claim

PR #6183, opened 2024-08-19
Support and tests for nearest mode in interpolate

Sir-Goose:


geohot:
Does it support uint8 and match torch?

Sir-Goose:
Matches torch but checked now and uint8 is off by a max absolute difference of 1. Will look into that.

Sir-Goose:
> I think your code looks good, how are you getting a max absolute difference of 1?
> 
> I just added a uint8 test and it passes.
> 
> ```
> class TestOpsUint8(unittest.TestCase):
>   def test_interpolate_nearest(self):
>     out_sz = (10, 10)
>     helper_test_op([(2,3,64,64)],
>       lambda x: torch.nn.functional.interpolate((10*x).relu().type(torch.uint8), size=out_sz, mode="nearest-exact"),
>       lambda x: Tensor.interpolate((10*x).relu().cast('uint8'), size=out_sz, mode="nearest"), forward_only=True)
> ```

Was just a bug with my test. @braxtonowens Thank you for yours I've added it to the PR. At least now we have confirmed uint8 works. 

@geohot looks good to merge if you are happy.

geohot:
uint8 linear. see discussion on #6088 and #6087

chenyuxyz:
#6308

PR #5939, opened 2024-08-06
nvidia fp8 support

davidgonmar:
Supports fp8 arithmetic and m16n8k32 tensor cores, with both e4m3 and e5m2 variants.

Arithmetic is supported as a graph rewrite rule that casts fp8 arithmetic to float, and stores the result in fp8 or bool depending on the op.
The approach for tensor cores is similar to existing implementations.

IMPORTANT: FP8 TC support only works on Cuda Toolkit 12.4+ (https://github.com/NVIDIA/cutlass/blob/f93a69134ec8259fd235f220209d6f8734a5cb06/include/cutlass/arch/mma_sm89.h#L53). I tried with a lower version and it did not compile, 12.4 works fine, and CUTLASS also compiles sm_89 tensor core instructions on cuda 12.4+ only.

Any feedback is welcome.

PR #6191, opened 2024-08-19
Add version field to CudaCompiler

davidgonmar:
For https://github.com/tinygrad/tinygrad/pull/5939


geohot:
Why do I want this?

davidgonmar:
> Why do I want this?
fp8 tensor cores only supported by compiler with toolkit version >= 12.4 (mentioned in the fp8 pr). In the fp8 pr I pass an additional parameter with the compiler version to the CUDARenderer class to decide if fp8 tensor cores should be enabled.

geohot:
This misses my point entirely about breaking up the PRs. This isn't the feature, this is a new unused param

PR #6192, opened 2024-08-19
fix intel wmma flop counting, add flop counting tests for different tensor cores

CaltropHungerton:
The intel tensor cores in the A770 have a warp size of 8 instead of 32. The modification in my XMX pr accounting for this got reverted when the flop counting function was moved into ops.py, and there weren't any tests to catch this. I have added tests and reinstated the change to count the flops correctly.


test/test_uops_stats.py
```diff
@@ -95,6 +95,16 @@ def test_simple_matmul(self):
     # NOTE: it's hard to assert on the memory here, all depends on caching
     assert required_mem <= mem
 
+  def test_simple_matmul_half(self):
+    if getenv("PYTHON"):
```
geohot, line 99: skip if

test/test_uops_stats.py
```diff
@@ -95,6 +95,16 @@ def test_simple_matmul(self):
...
+      GlobalCounters.reset()
+      N = 128
```
geohot, line 101: does it have to be so big? PYTHON=1 is very slow

CaltropHungerton: Changed to 16x16, should be fine for the TCs we have. Would only need to be bumped up if someone added a crazy exotic large TC like the one they have in the H100s (64x16x16).

PR #6199, opened 2024-08-20
track running stats bug (BatchNorm)

rootdrew27:
This PR builds upon #6146

## Changes to Tests (`test/`)
- Altered `test_batchnorm2d` to use valid variance values (i.e. non-negative values).
- Altered `test_batchnorm2d` to only assign and test the running stats when `track_running_stats=True`
- The PyTorch Batch now properly sets `track_running_stats`
- Also possible combinations of parameters for `test_batchnorm2d` are tried (though probs not needed).
- To mimic pytorch, `test_fold_conv_batchnorm_notrain()` is updated to allow 4 kernels.

## Changes to BatchNorm (`tinygrad.nn.__init__.py__`)
- The tracked stats are used if in eval mode (i.e. Tensor.training == False).
- `calc_stats()` is added.
- Variable names are changed to reflect their usage more accurately.

### DIFF
--- a/test/test_schedule.py
+++ b/test/test_schedule.py
@@ -195,7 +195,7 @@ class TestSchedule(unittest.TestCase):
       c1 = nn.Conv2d(3,32,3)
       bn = nn.BatchNorm2d(32, track_running_stats=False)
       out = bn(c1(img)).relu()
-      check_schedule(out, 1, [c1.weight, c1.bias])
+      check_schedule(out, 4, [c1.weight, c1.bias])
 
   def test_fold_conv_batchnorm(self):
     with Tensor.train():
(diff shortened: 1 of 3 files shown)


geohot:
If you break it, do the tests break?

test/test_schedule.py
```diff
@@ -195,7 +195,7 @@ def test_fold_conv_batchnorm_notrain(self):
       c1 = nn.Conv2d(3,32,3)
       bn = nn.BatchNorm2d(32, track_running_stats=False)
       out = bn(c1(img)).relu()
-      check_schedule(out, 1, [c1.weight, c1.bias])
+      check_schedule(out, 4, [c1.weight, c1.bias])
```
geohot, line 198: Can you add the test that keeps this 1?

PR #6202, opened 2024-08-20
delete arg from (LazyBuffer) cast [run_process_replay]

Qazalin:


geohot:
Cool, I think that was there for legacy reasons

PR #6206, opened 2024-08-20
nv turing support

nimlgen:


nimlgen:
Will revisit later. After testing saw some problems with old version of qmd

PR #6211, opened 2024-08-20
batchnorm bugfix, using online estimate

Itssshikhar:


geohot:
I don't understand what bug this fixes or why there's more params in batchnorm now

PR #5572, opened 2024-07-19
draft: move SPLIT_REDUCEOP into kernel.py

cpwrs:
draft PR for bounty #5542

```
kernel.apply_opt(OptOps.SPLIT, axis, amt)
k1, k2 = kernel.split()
```
Now working on kernels splitting during search.

To do
- [X] Write a test with a hand coded LazyOp with a KERNEL in the middle. Get the lowerer generating the right thing.
- [X] Add support for 0-n ExecItems for each ScheduleItem.
- [X] lazy.py reduce split -> OptOps.SPLIT
- [ ] Add support for varying kernel count in search.

Issue #5542, opened 2024-07-18
Move SPLIT_REDUCEOP into kernel.py


geohot:
One LazyOp, two Programs.

Include a KERNEL in the middle of the LazyOp to get a second kernel.

1) Write a test with a hand coded LazyOp with a KERNEL in the middle. Get the lowerer generating the right thing.
2) Add support for 0-n ExecItems for each ScheduleItem.
3) Move SPLIT logic from `lazy.py` to `kernel.py`

Once it's in kernel.py, we can search over it.

Emmanueltech:
Hi I would like to work on this bounty.

Qazalin:
Completed in #9214.

PR #6226, opened 2024-08-21
One ScheduleItem, multiple ExecItems

cpwrs:
I rewrote #5572 with the new UOp AST, and the diff got big. Splitting bounty #5542 into a few PRs.

This one is essentially #5880. It adds support for 0-n `ExecItem`s for each `ScheduleItem`, and prepares `get_kernel` for one `Kernel` splitting into n `Kernel`s during search.


tinygrad/engine/realize.py
```diff
@@ -186,26 +187,28 @@ def run(self, var_vals:Optional[Dict[Variable, int]]=None, wait=False, jit=False
...
     return et
 
-def lower_schedule_item(si:ScheduleItem) -> ExecItem:
+def lower_schedule_item(si:ScheduleItem) -> List[ExecItem]:
   assert len(set(x.device for x in si.bufs)) == 1 or (si.ast.op is UOps.EXT and si.ast.arg[0] is MetaOps.COPY)
   if si.ast.op is UOps.SINK:
-    runner = get_runner(si.outputs[0].device, si.ast)
-    return ExecItem(runner, [si.bufs[x] for x in runner.p.globals], si.metadata)
+    dname = si.outputs[0].device
+    runners = [get_runner(dname, k) for k in get_kernels(Device[dname].renderer, si.ast)]
```
Qazalin, line 194: are you recreating Kernel for si.ast every time?

cpwrs: As much as before and the same cache. Just had to move `get_kernel` out of `get_runner` because `get_kernels` could return multiple Kernels after `OptOps.SPLIT`.

cpwrs: I see let me fix that..

cpwrs: Should be caching again

Qazalin: Can you add a test for it in test_method_cache?

tinygrad/engine/realize.py
```diff
@@ -61,7 +61,8 @@ def get_kernel(renderer:Renderer, ast:UOp) -> Kernel:
                   raise RuntimeError(f"mismatch of {diff_count}/{b.numel()} items with type {b.dtype}, max {(b-bufs[0]).abs().max().item()}")
   if logkerns is not None: logkerns.writelines([f"{(k.ast, k.applied_opts)}\n"])
   if DEBUG >= 5: print((k.ast, k.applied_opts)) # print here to show final applied_opts for all kernels instead of just in beam_search
-  return k
+  # NOTE: this can return multiple kernels for one ast in the future (for things like split reduce op)
+  return [k]
```
geohot, line 65: I wouldn't put this in this PR. It's not a used feature, and it might actually be one kernel with split

cpwrs:
I don't know how to hit any of these realize features without split in search. Should I allow multiple SINKs in an AST coming from the scheduler?

geohot:
So it's still just one Kernel, the split optimization needs to happen in Kernel.

PR #6233, opened 2024-08-22
Add leave argument to tqdm

L1aoXingyu:
Add leave argument to tqdm which can make progress bar disappear after iteration.


L1aoXingyu:
Anyone can give some comments?

chenyuxyz:
if you add a feature, you need to add tests too.

but i don't really like this since now many functions can update the bar, which adds complexity. and we don't really use the feature

L1aoXingyu:
ok, I will close it.

PR #6234, opened 2024-08-22
remove NEG from handwritten ast in tests

chenyuxyz:
skipped two tests in `test_argmax_multireduce_axis0` that i don't know how to fix. @Qazalin do you have an instruction?

other tests either use different Ops to have similar coverage, and for test_linearizer_failures i just removed the NEG


Qazalin:
hmm, these look green?
image

chenyuxyz:
i will remove NEG in next pr that regenerates the kernels (some are from tests), so i am removing all NEG now

PR #6253, opened 2024-08-23
[wip] add turing tc support

ignaciosica:
# Turing Tensor Core support

[reference](https://docs.nvidia.com/cuda/parallel-thread-execution/#warp-level-matrix-fragment-mma-1688)

Turing tc dimensions are `m16n8k8`

Getting ~ 10500 GFLOPS in local device with BEAM=2

TODO:
- generalize wmma rendering for different sizes (Probably better in a separate PR)
- investigate why `bfloat16` was failing CI and add support

PROBLEMS:
- No real tensor core test in CI
- Hack in ops_python for checking cuda arch


geohot:
Before we do this tensor cores need to be refactored. Those if statements shouldn't exist, that should be in the TensorCoreOptions class

ignaciosica:
Sure, thanks for your feedback.

ignaciosica:
closed in favor of #6720

PR #6258, opened 2024-08-23
Einsum ellipsis support

### LINKED ISSUE #4156: Tensor.einsum does not correctly handle unspecified dimensions
I've been converting a model from PyTorch to tinygrad, and ran into an issue with einsums (defined in Tensor.py). 

The einsum "...id, ...jd -> ...ij" executes fine in PyTorch but throws an error (below) when executed in tinygrad on the exact same input data. I've included a minimal example below the error for debugging. 
image

```python
import tinygrad
import torch
import numpy as np

# Define the input tensor
inp1_np = np.random.randn(16, 29, 256).astype(np.float32)
inp2_np = np.random.randn(16, 29, 256).astype(np.float32)

# Convert the input tensors to torch tensors
inp1 = torch.tensor(inp1_np)
inp2 = torch.tensor(inp2_np)

out = torch.einsum("...id, ...jd -> ...ij", inp1, inp2)
# Evaluate the output tensor
out_np = out.numpy()

# Convert the input tensors to tinygrad tensors
inp1_tg = tinygrad.Tensor(inp1_np)
inp2_tg = tinygrad.Tensor(inp2_np)

# Perform the einsum operation using tinygrad - will throw error!
out_tg = tinygrad.Tensor.einsum("...id, ...jd -> ...ij", inp1_tg, inp2_tg)

# Evaluate the output tensor
out_tg_np = out_tg.numpy()
```

I looked at the locals when the error appeared, and it looks like the "..." part of the einsum is being interpreted as an index for each of the dots. I might be wrong, but I think this is supposed to represent an arbitrary number of indices, not specific indices. 
image

Let me know if you need any more information!

Max-We:
This implementation follows the ellipsis behavior described by the PyTorch `einsum` documentation (https://pytorch.org/docs/stable/generated/torch.einsum.html#torch-einsum). To confirm whether the implementation works, tests have been added which verify each behavior.

The approach is to "expand" the ellipsis in the operands on the lhs and rhs of the formula to actual subscripts. Then, this new formula is processed as usual.

This PR resolves #4156.


Max-We:
Changing to draft, some tests are failing

geohot:
This is a big line addition. Can it share code with `rearrange`?

Max-We:
> This is a big line addition. Can it share code with `rearrange`?

It seems like the `einsum` and `rearrange` formulas behave differently even though they look similar. `einsum` supports 1) multiple operands, 2) subscripts having the same value and 3) omitting the rhs (inclusive arrow), while `rearrange` does not. On the other hand, `rearrange` supports 4) braces in the formula, while `einsum` does not. Together, this results in individual differences in the parsing of the formula string and the process of expanding ellipsis. Overall, I don't see a way of saving a significant number of lines (>5) via refactoring while maintaining readability, but I'd be glad if someone else does.

chenyuxyz:
definitely not 32 lines, i think 5 max

Max-We:
> definitely not 32 lines, i think 5 max. like there are multiple broadcast related functions so i don't think `is_broadcastable` is needed just for input validation

I changed the approach down to 25 lines, if that's still too long, feel free to close the pr. I'm curious how an implementation that handles all tests would look like in 5 lines.

chenyuxyz:
i think the tests are good, if you want you can have a pr for these tests and make as expectedFailure for TODOs

Max-We:
> i think the tests are good, if you want you can have a pr for these tests and make as expectedFailure for TODOs

I created a new PR #6286 and will close this PR.

PR #6262, opened 2024-08-24
advanced setitem

ttomsa:
Need to handle repeated indices, fix the new_shape atrocity and clean up


ttomsa:
It's too messy, I'm just going to refactor __getitem__.

geohot:
@chenyuxyz can you review this when you are back? I'm not familiar at all with the getitem code

tinygrad/tensor.py
```diff
@@ -942,7 +942,7 @@ def pad(self, arg:Tuple[Optional[Tuple[sint, sint]], ...], value:float=0.0) -> T
   #   2. Bool indexing is not supported
   #   3. Out of bounds Tensor indexing results in 0
   #     - e.g: Tensor([1, 2, 3])[Tensor([4, 3, 2])] -> [0, 0, 3] index 4 and 3 are out of bounds
-  def __getitem__(self, indices) -> Tensor:
+  def __getitem__(self, indices, v: Union[Tensor, None] = None) -> Tensor:
```
chenyuxyz, line 945: don't overload the `__getitem__` API, if you need a different function with additional Tensor as input, create a function for it.

chenyuxyz (commented):
mostly looks fine. can you run some benchmark against torch? it's fine if it's slow just want to know where we are

tinygrad/tensor.py
ttomsa: Will benchmark tomorrow. I can't think of a cleaner way than passing v as an argument as advanced setitem needs access to getitem's internals and v. I tried breaking down getitem but found it looked worse.

ttomsa:
> mostly looks fine. can you run some benchmark against torch? it's fine if it's slow just want to know where we are

advanced setitem is ~10x slower than getitem vs pytorch ~3x slower.

Code_rSOeCUqL5h

tinygrad/tensor.py
geohot: Can you write a function they both depend on?

tinygrad/tensor.py
```diff
@@ -1045,29 +1044,47 @@ def calc_dim(tensor_dim:int) -> int:
...
-      raise NotImplementedError("Advanced indexing setitem is not currently supported")
 
-    assign_to = self.realize().__getitem__(indices)
-    # NOTE: contiguous to prevent const folding.
-    v = v.cast(assign_to.dtype)._broadcast_to(_broadcast_shape(assign_to.shape, v.shape)).contiguous()
-    assign_to.assign(v).realize()
+    res = self.realize()._getitem(indices, v)
+    # if shapes match and data is not shared it's a copy and we assign to self
+    if res.shape == self.shape and res.lazydata is not self.lazydata:
+      self.assign(res.cast(self.dtype).contiguous()).realize()
```
chenyuxyz, line 1084: do you need this cast? do you need this contiguous?

ttomsa: The contiguous was a leftover from trying to assign to self when self was non contiguous, I removed it. The cast is the same as before, can't assign with different dtypes. I moved it to getitem so it casts early, it's a bit cleaner.

chenyuxyz:
bounty is yours, congrats! email george@tinygrad.org to claim the bounty (paypal or USDC).

PR #6270, opened 2024-08-24
simpler & faster UOp key

roelofvandijk:
This potentially could just use the builtin `hash` function. Does not need to be a cached property anymore either.

`python -O test/external/external_test_speed_llama.py`

master
```
codegen(0)      mean runtime:  169.99ms, runs:   231.78,  143.90,  157.94,  146.22,  170.13
codegen(1)      mean runtime:  160.19ms, runs:   212.01,  142.28,  144.54,  153.92,  148.23
methodcache     mean runtime:  122.26ms, runs:   130.98,  117.75,  118.34,  126.58,  117.65
profile         mean runtime:  358.61ms, runs:   362.39,  356.50,  364.70,  353.77,  355.69
```

branch
```
codegen(0)      mean runtime:  150.76ms, runs:   208.69,  128.98,  138.77,  135.32,  142.07
codegen(1)      mean runtime:  144.40ms, runs:   191.80,  128.79,  138.18,  130.26,  132.98
methodcache     mean runtime:  107.44ms, runs:   118.65,  102.11,  102.56,  111.41,  102.46
profile         mean runtime:  334.07ms, runs:   335.63,  327.87,  327.13,  342.77,  336.98
```

### DIFF
--- a/tinygrad/engine/realize.py
+++ b/tinygrad/engine/realize.py
@@ -146,7 +146,7 @@ class BufferXfer(BufferCopy):
 
 # **************** method cache ****************
 
-method_cache: Dict[Tuple[str, bytes, int, int, bool], CompiledRunner] = {}
+method_cache: Dict[Tuple[str, int, int, int, bool], CompiledRunner] = {}
 def get_runner(dname:str, ast:UOp) -> CompiledRunner:
   ckey = (dname, ast.key, BEAM.value, NOOPT.value, False)
   if cret:=method_cache.get(ckey): return cret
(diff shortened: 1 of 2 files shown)


Qazalin:
https://stackoverflow.com/questions/34330284/what-is-the-risk-of-collisions-when-relying-on-python-hash-function

roelofvandijkO:
@Qazalin If the risk for collisions is so high here, should the hash use [here](https://github.com/tinygrad/tinygrad/blob/85591bd1ae6ca9ff1e78c999468ee9902ca0093b/tinygrad/codegen/kernel.py#L770) also become a `.key` use?

It might be useful to remove the hash function of the class to avoid calling it inadvertently.

```
  def __hash__(self):
    raise NotImplementedError
```

roelofvandijk:
@Qazalin If the risk for collisions is so high here, should the hash use [here](https://github.com/tinygrad/tinygrad/blob/85591bd1ae6ca9ff1e78c999468ee9902ca0093b/tinygrad/codegen/kernel.py#L770) also become a `.key` use?

It might be useful to remove the hash function of the class to avoid calling it inadvertently.

```
  def __hash__(self):
    raise NotImplementedError
```

Qazalin:
key and hash mean different things for Uop, so I don't think the risk is comparable.

PR #6284, opened 2024-08-26
process replay rewrite

Qazalin:
This file was getting out of control. This should also be much easier to unit test.

It's 3 sections:
1. differs: These are single-threaded, independent functions and return a bool (`changed`)
2. diff runners: multi threaded, running differs across the entire process replay capture
3. main loop: Runs both diff runners, **only** raises their exceptions if `[run_process_replay]` is set.

### DIFF
--- a/test/external/process_replay/process_replay.py
+++ b/test/external/process_replay/process_replay.py
@@ -58,6 +58,9 @@ def diff_kernel(offset:int) -> bool:
         # NOTE: replay with the captured renderer, not the one in master
         good_src = k.opts.render(name, cast(List,k.to_program().uops))
     except Exception as e:
+      k = Kernel(ast, opts=opts)
+      for opt in applied_opts: k.apply_opt(opt)
+      print(k.to_program())
       logging.warning(f"FAILED TO RECREATE KERNEL {e}")
       logging.info(ast)
       logging.info(applied_opts)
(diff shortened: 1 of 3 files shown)


test/external/process_replay/process_replay.py
```diff
@@ -1,28 +1,41 @@
...
 # *** process replay settings
+
+# internal
 PAGE_SIZE = 100
 REF = os.getenv("GITHUB_REF_NAME", "")
 MAX_DIFF_PCT = getenv("PROCESS_REPLAY_MAX_DIFF_PCT", 20)
 RUN_ID = os.getenv("GITHUB_RUN_ID", "HEAD")
 TABLE_NAME = f"process_replay_{RUN_ID}_{getenv('GITHUB_RUN_ATTEMPT')}_{VERSION}"
+early_stop = multiprocessing.Event()
+logging.basicConfig(level=logging.INFO, format='%(message)s')
```
chenyuxyz, line 16: consider setting proper levels instead of changing behavior based on `ASSERT_DIFF`. it's easier to maintain if the log flows are the same, and the only difference is that warning does throw

Qazalin: hmm, I wonder why changing behavior based on ASSERT_DIFF is bad, I think it's intentional.
but yea logging is exactly the same, whether CI is red or not is different.

PR #6290, opened 2024-08-26
update LazyOp to UOp AST in extra/ helpers and to_movement_ops

greg-niemeyer:


greg-niemeyer:
I looked into cleaning up extract_sa_pairs.py, extract_policynet.py, and pretrain_valuenet.py as well, but I don't know how to run these scripts (I seem to be missing some required metadata to run these examples).

Qazalin:
don't worry about those, as we say in the README:
> We don't care about the code in extra

Qazalin:
load_worlds should also be fixed up, those filters are wrong.

greg-niemeyer:
the filters should be good now -- tested with
- any("REDUCE" in x for x in load_worlds(True, False, False)) -> True
- any("image" in x for x in load_worlds(False, True, False)) -> False
- any("Variable" in x for x in load_worlds(False, False, True)) -> False
but lin_to_feats also has some issue I'm observing after running extra/optimizations/rl.py. looking into that now.

greg-niemeyer:
When I ran rl.py, lin_to_feats had the following failure modes:

1. symbolic shapes were causing math functions to fail
2. the Kernel shapetrackers were exceeding MAX_DIMS or MAX_BUFS

I addressed (1) in this diff, since that seems like something that should be updated in helpers. IMO (2) is user error and should be fixed by the caller. We could add an additional helper function that the caller can use to determine if the kernel can be used with lin_to_feats(), but that seems beyond the scope of this PR and more appropriate in a PR that fixes up the examples in extra/optimizations (if we decide we care about this code).

greg-niemeyer:
eh, on second though, I think it makes sense to update rl.py in this patch since it's the only thing that I can run that tests lin_to_feat and the newly added is_lin_valid_feat. pretrain_valuenet.py is a dependency of rl.py, so it requires updating too. And finally, I'm also including the removal of LazyOp in extract_sa_pairs.py and extract_policynet.py because otherwise attempting to run them just throws an import error. I still can't get them to work cos I don't have their required files, but this at least gets them farther.

Qazalin:
meh, we really don't care about functions in extra/ that aren't used in CI. So I'd rather keep them as-is. Isn't this diff all we need?
image

greg-niemeyer:
to_movement_ops.py runs in CI

PR #6293, opened 2024-08-27
SDXL Batched Inference Fixes

tobias17:
## Overview

1. Fixes the current issues stopping SDXL from running batched inference
2. Fixes issues related to methods using global variables that caused issues when running those classes in another file
3. Fixes the type hinting of components touched in this PR

## Testing

```
$ PYTHONPATH=. NV=1 python examples/sdxl.py --seed 0
saving /tmp/rendered.png
output validated with distance=0.00289628142490983
```

```
$ PYTHONPATH=. NV=1 python examples/sdv2.py --seed 0
saving /tmp/rendered.png
```

```
$ PYTHONPATH=. NV=1 python examples/stable_diffusion.py --seed 0
saving /tmp/rendered.png
output validated with distance=9.23173956834944e-07
```

I also made a custom delta of `examples/sdxl.py` with hard-coded batching code to validate these changes worked
```
$ PYTHONPATH=. NV=1 python examples/sdxl.py --seed 0
saving /tmp/rendered_0.png
saving /tmp/rendered_1.png
```

### DIFF
--- a/extra/models/clip.py
+++ b/extra/models/clip.py
@@ -108,7 +108,7 @@ class Tokenizer:
       self.cache[token] = word
       return word
 
-    def encode(self, text:str, pad_with_zeros:bool=False):
+    def encode(self, text:str, pad_with_zeros:bool=False) -> List[int]:
       bpe_tokens: List[int] = []
       text = Tokenizer.whitespace_clean(text.strip()).lower()
       for token in re.findall(self.pat, text):
@@ -123,7 +123,7 @@ class Tokenizer:
 class Embedder(ABC):
   input_key: str
   @abstractmethod
-  def __call__(self, text:str) -> Union[Tensor,Tuple[Tensor,...]]:
+  def __call__(self, x:Union[str,List[str],Tensor]) -> Union[Tensor,Tuple[Tensor,...]]:
     pass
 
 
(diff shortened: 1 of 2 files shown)


extra/models/clip.py
```diff
@@ -222,9 +222,11 @@ def __init__(self, ret_layer_idx:Optional[int]=None):
     self.transformer = Closed.ClipTextModel(ret_layer_idx)
     self.input_key   = "txt"
 
-  def __call__(self, text:str) -> Union[Tensor,Tuple[Tensor,...]]:
-    tokens = Tensor(self.tokenizer.encode(text))
-    return self.transformer.text_model(tokens.reshape(1,-1))
+  def __call__(self, texts:Union[str,List[str],Tensor]) -> Union[Tensor,Tuple[Tensor,...]]:
```
chenyuxyz, line 225: can `texts` be `Tensor` here?

same for other functions. I find mixing str and Tensor weird because we don't support string tensors

tobias17: This ends up being weird since it's the input to the `Embedder` which for CLIP is the raw text to get tokenized and then embedded, the CLIP tokenizer expects a string input to return a Tensor

The reason there is a `Union` between `str` and `Tensor` is that SDXL has multiple types of embedders, some acting on the text prompt and some acting on values like target image size, the typing was to make the system happy with both kinds of inputs, while each embedder starts with an assert now to make sure it got the right kind of input.

I understand your point, and in theory this could get converted, but it would require porting the entire tokenizer logic to tinygrad, which seems like a lot of effort for little gain.

On another note, the whole Embedder system was due to porting the structure in the reference implementation. The whole thing is causing issues in relationship to sharding, and I am currently looking into how I want this to get refactored. Everything is currently abstracted away and built around dependency injection, which is beneficial for researchers trying out multiple architecture configurations, but terrible for a direct implementation that is understandable. Making things more explicit should make it both more understandable and removed this kind of "dual input types".

chenyuxyz: i see, asserting for now should be fine

PR #6298, opened 2024-08-27
more scheduler rewrite benchmark, it's o(n^2)...

Qazalin:
newplot


Qazalin:
better,
newplot (4)
It's all get_output_st.

PR #6299, opened 2024-08-27
Replace dlopen with mmap in CLANG

christopherm99:
## Performance
Tested on an M1 MacBook Pro.
```py
from tinygrad.runtime.ops_clang import ClangProgram

with open("test.o", "rb") as f: lib = f.read()
for _ in range(1000): ClangProgram("test", lib)
```
image

## Notes

Technically, forcing the `.text` section to start at address 0x0 is out of the ELF spec, but this means that we don't need to `mmap` extra zeros or account for the offset to the function. 

No new autogen files are required (`libc.py` now includes `libkern/OSCacheControl.h`, downloaded from [opensource.apple.com](https://opensource.apple.com/source/Libc/Libc-583/include/libkern/OSCacheControl.h.auto.html)).


geohot:
So I wouldn't parse macho, way too mac specific. If this goes in, it removes the linker.

https://github.com/tinygrad/tinygrad/pull/4492 had many of the right ideas, but it was never polished to the point that it could be merged.

christopherm99:
Got it, I saw what #4492 did but I felt like targeting `...-none-unknown-elf` seemed a bit hacky. I'll take a look to see if I can find a clean way to remove the linker, but if I don't find anything I'll close this.

geohot:
I don't think `none-unknown-elf` was the sketchy part of that PR, afaik all clang should support it.

tinygrad/runtime/graph/clang.py
```diff
@@ -14,7 +14,8 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
     super().__init__(jit_cache, input_rawbuffers, var_vals)
     if not all(isinstance(ji.prg, CompiledRunner) for ji in jit_cache): raise GraphException
 
-    prgs = '\n'.join(dedup([cast(CompiledRunner, ji.prg).p.src for ji in jit_cache]))
+    new_decl = "static inline __attribute__((always_inline)) void"
```
geohot, line 17: I don't think this is the right way to deal with graph. Use a linker section to ensure the function runner is always first

tinygrad/runtime/ops_clang.py
```diff
@@ -1,24 +1,37 @@
...
-from tinygrad.helpers import cpu_time_execution, DEBUG, cpu_objdump
+from tinygrad.helpers import OSX, cpu_time_execution, DEBUG, cpu_objdump, mv_address
 from tinygrad.renderer.cstyle import ClangRenderer
+from tinygrad.runtime.support.elf import elf_loader
+from tinygrad.runtime.autogen import libc
 
 class ClangCompiler(Compiler):
   def compile(self, src:str) -> bytes:
+    triple = f"{platform.machine()}-none-unknown-elf"
+    asm = subprocess.check_output(['clang', '-march=native', '-O2', '-Wall', '-Werror', '-x', 'c', '-S', '-ffreestanding', '-nostdlib', '-target',
```
geohot, line 11: There has to be a better way than this. subprocess is really expensive

tinygrad/runtime/support/elf.py
```diff
@@ -23,7 +23,7 @@ def _to_carray(sh, ctype): return (ctype * (sh.header.sh_size // sh.header.sh_en
   # Prealloc image for all fixed addresses.
   image = bytearray(max([sh.header.sh_addr + sh.header.sh_size for sh in progbits if sh.header.sh_addr != 0] + [0]))
   for sh in progbits:
-    if sh.header.sh_addr != 0: image[sh.header.sh_addr:sh.header.sh_addr+sh.header.sh_size] = sh.content
+    if sh.header.sh_addr != 0 or sh.name == '.text': image[sh.header.sh_addr:sh.header.sh_addr+sh.header.sh_size] = sh.content
```
geohot, line 26: Why does 0 need a special rule in the first place?

christopherm99: If the section does not appear in the process' memory then `sh_addr` is set to 0 (see `man 5 elf`). I think this is relevant for the NV backend?

tinygrad/runtime/ops_clang.py
christopherm99: Will look at using `libclang` instead, with autogenerated bindings.

tinygrad/runtime/graph/clang.py
christopherm99: Ok, sounds good. I think having multiple functions was creating relocations, but I'll see if I can work around this.

tinygrad/runtime/ops_clang.py
christopherm99: So from what I can tell, `libclang` doesn't actually support compilation, and is really just made for AST browsing. `clang -cc1` (the compiler frontend) is actually implemented using `libclang-cpp`, which makes no guarantees about a stable API, and, as the name suggests, is written in C++, which would make calling out to it from Python much harder (not to mention we'd likely have to require a specific version of Clang and LLVM to be installed because of the unstable API). There are a couple projects I found that also use this API, such as [DragonFFI](https://github.com/aguinet/dragonffi), [Cling](https://github.com/root-project/cling), and [this blog post](https://blog.memzero.de/libclang-c-to-llvm-ir/), but all these are written in C++.

christopherm99:
So my reading on this suggests that if we don't want to use `subprocess` to invoke clang, we're looking at interfacing with clang's unstable C++ API, since `libclang` doesn't support compilation. Essentially, this means creating a new library `clanglite`, that is, an `llvmlite` equivalent for clang.

Since there are no guarantees about the stability of this API, we'd either have to fix a version of clang to be compatible with (like how `llvmlite` requires LLVM 14.x.x). Additionally, since it's a C++ API, we need an FFI wrapper similar to [`llvmlite`'s wrapper](https://github.com/numba/llvmlite/tree/main/ffi). I think it should be possible to autogenerate a lot of this with a tool like SWIG, but the major downside is a new dependency for tinygrad.

I can start working on such a library, but I just want to make sure that this would be the sort of solution that could be merged before I spend too much time struggling with SWIG.

Issue #6045, opened 2024-08-12
Add dtypes.void for UOps


Qazalin:
You need to update codegen/* and remove the `cast(DType,` stuff in renderer/*.

jackhwalters:
I can take this up

jackhwalters:
Could I have a bit more detail on what needs to be done?

Lazarus42:
Also interested in this. Would be down to pair.

mikeest1972:
I would like to work with @jackhwalters  and @Lazarus42  on this issue. Its my first one also!

chenyuxyz:
#6471

PR #6300, opened 2024-08-27
added dtypes.void removed option from Uops

HD787:
adds void type described in #6045 

Some tests still need to be altered to handle void and not instantiate Uops to None


tinygrad/ops.py
```diff
@@ -62,7 +62,7 @@ class UOps(Enum):
 @dataclass(frozen=True, eq=False)
 class UOp:
   op: UOps
-  dtype: Optional[DType] = None
+  dtype: DType = dtypes.void
```
Qazalin, line 65: does NOp work correctly?

HD787: Appears to, had to handle a case where UPats are instantiated with NOps, but that's it

test/test_dtype.py
```diff
@@ -65,7 +66,7 @@ def setUpClass(cls):
     if not cls.DTYPE or not is_dtype_supported(cls.DTYPE): raise unittest.SkipTest("dtype not supported")
     cls.DATA = rand_for_dtype(cls.DTYPE, 10)
   def setUp(self):
-    if self.DTYPE is None: raise unittest.SkipTest("base class")
+    if self.DTYPE is None or dtypes.void: raise unittest.SkipTest("base class")
```
roelofvandijk, line 69: if self.DTYPE in (None, dtypes.void)

tinygrad/ops.py
```diff
@@ -229,7 +229,7 @@ def __init__(self, op:Optional[Union[UOps, Set[UOps]]]=None, arg:Any=None, src:O
                name:Optional[str]=None, dtype:Optional[Union[DType, Set[DType]]]=None, allow_any_len:bool=False, location=None,
                custom_early_reject:Optional[Set[Tuple[UOps, Any]]]=None):
     self.op: Optional[Tuple[UOps, ...]] = None if op is None else (tuple(op) if isinstance(op, set) else (op,))
-    self.dtype: Optional[Tuple[DType, ...]] = None if dtype is None else (tuple(dtype) if isinstance(dtype, set) else (dtype,))
+    self.dtype: Optional[Tuple[DType, ...]] = None if dtype is None or dtype is dtypes.void else (tuple(dtype) if isinstance(dtype, set) else (dtype,)) # noqa: E501
```
roelofvandijk, line 232: `if dtype in (None, dtypes.void)` - but can dtype still be None?

HD787: I like these suggestions, I'll change them. It appears that it can be None here, although it is possible this is the the result of tests instantiating certain values to None, I think I've found most of those occurrences though.

geohot:
Not thought through, just a blind find/replace

PR #6302, opened 2024-08-27
Support `reduction` parameter in all built-in loss functions

### LINKED ISSUE #6247: Support `reduction` param in other loss functions
torch lets you control how the loss is reduced in most (all?) loss functions, which is convenient.

Currently tinygrad only supports this in the newly added `Tensor.cross_entropy`.

Note: I wanted this for `Tensor.sparse_categorical_crossentropy`, but was able to work around it by using `Tensor.cross_entropy` instead (with an appropriate reshaping).

Still worth tracking this issue in general though, IMO.

tim-becker:
Relatively simple. `sparse_categorical_cross_entropy` is the only non-trivial case because
* it supports `ignore_index`, so computing the mean is not as simple as `Tensor.mean`
* it was already optimized assuming `reduction="mean"`.

The implementation of `sparse_categorical_cross_entropy` in this PR was written to:
1. Reduce amount of new code
2. Maintain performance of default (`"mean"`) case

A side effect of 1. is that `sparse_categorical_cross_entropy` now typically produces an extra kernel. However, my measurements show identical perf (on metal) to the previous implementation. I can push the perf testing code I added to test/test_speed_v_torch.py if you want.

If preserving the number of kernels is more important than the code being concise and clean, we could easily have an optimized version per reduction type instead. But, it feels more like tinygrad style to remove the extra kernel through smarter kernel fusion rule. In this case, it's as simple as fusing `x.sum(-1).sum()` into a single kernel.

edit: @Qazalin informed me that `FUSE_CONV_BW=1` does this fusing, so this extra kernel will disappear as soon as that becomes default.

Fixes #6247

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import dataclasses
 import time, math, itertools, functools, struct, sys, inspect, pathlib
 from contextlib import ContextDecorator
-from typing import List, Tuple, Callable, Optional, ClassVar, Type, Union, Sequence, Dict, DefaultDict, cast, get_args, Set
+from typing import List, Tuple, Callable, Optional, ClassVar, Type, Union, Sequence, Dict, DefaultDict, cast, get_args, Set, Literal
 from collections import defaultdict
 import numpy as np
 
@@ -88,6 +88,8 @@ def _pad_left(*shapes:Tuple[sint, ...]) -> Tuple[Tuple[sint, ...], ...]:
   return tuple((1,) * (max_dim - len(shape)) + shape for shape in shapes)
 def _broadcast_shape(*shapes:Tuple[sint, ...]) -> Tuple[sint, ...]:
   return tuple(0 if 0 in nth_dim_sizes else max(nth_dim_sizes) for nth_dim_sizes in zip(*_pad_left(*shapes)))
+
+ReductionStr = Literal["mean", "sum", "none"]
 
 class Tensor:
   """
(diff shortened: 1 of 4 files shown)


tinygrad/tensor.py
```diff
@@ -3040,13 +3045,14 @@ def binary_crossentropy_logits(self, y:Tensor) -> Tensor:
     print(t.binary_crossentropy_logits(y).item())
     ```
     """
-    return (self.maximum(0) - y * self + (1 + self.abs().neg().exp()).log()).mean()
+    return (self.maximum(0) - y * self + (1 + self.abs().neg().exp()).log())._do_reduction(reduction)
 
-  def sparse_categorical_crossentropy(self, Y:Tensor, ignore_index=-1, label_smoothing=0.0) -> Tensor:
+  def sparse_categorical_crossentropy(self, Y:Tensor, ignore_index=-1, label_smoothing=0.0, reduction:Literal["mean", "sum", "none"]="mean") -> Tensor: # noqa: E501
```
wozeparrot, line 3050: no new `noqa: E501`

tinygrad/helpers.py
```diff
@@ -353,3 +353,5 @@ def dfs(x:Any, cache:dict):
   if (cx:=cache.setdefault(x, [0,0,False]))[2]: return f"{' '*d} x{cx[0]}"
   cx[2], srcs = True, ('None' if srcfn(x) is None else ''.join(f'\n{pretty_print(s, rep, srcfn, cache, d+2)},' for s in srcfn(x)))
   return f"{' '*d}{f'x{cx[0]}:=' * (cx[1]>1)}{rep(x)}" % srcs
+
+ReductionStr = Literal["mean", "sum", "none"]
```
geohot, line 357: Why is this in helpers? Is it used anywhere besides Tensor?

tim-becker: nah, and probably never will be... I moved it to tensor.py

tim-becker: Also rebased and squashed

geohot:
Cool, merged!

PR #6305, opened 2024-08-28
Fix uint8 cast underflow

geohotstan:
some of my findings:
- apparently float -> unsigned casting has always been problematic [[1]](https://embeddeduse.com/2013/08/25/casting-a-negative-float-to-an-unsigned-int)
- `TestOpsUint8.test_cast` works with on my local `LLVM=1` and `CLANG=1` but fails on `METAL=1`, `GPU=1` and `CLANG=1 NOOPT=1` (for CLANG and NOOPT, it fails for loop sizes larger than 448)
- the `self.prg()` in `ExecItem.run()` fails to cast correctly for the ExecItem that is responsible for casting. (I checked the buf values)

Also numpy has similar float -> unsigned casting [issues](https://github.com/numpy/numpy/issues/22951) though it's to unsigned int. Their float -> unsigned char works.

I'm not sure this 2 step casting is the right way to do it but it's the solution proposed in [1] and it pins it to the right value for floats inside int64 range. This seems a little hacky so I only limited to float32 -> uint8

The fact that `CLANG=1` and `NOOPT=1` fails for loop sizes over 448 is just bewildering...
```
Device.DEFAULT = "CLANG"
with Context(NOOPT=1):
  for i in range(400, 500):
    ts =torch.tensor(np.random.uniform(low=-2, high=2, size=i).astype(np.float32), requires_grad=(False))
    t = Tensor(ts.detach().numpy())
    try:
      np.testing.assert_allclose(ts.type(torch.uint8).numpy(), t.cast("uint8").numpy())
      print(i, "success")
    except: print(i, "fail")
```
```
...
446 success
447 success
448 fail
449 fail
450 fail
451 fail
452 fail
...
```
and some loop sizes above 448 has some values with 255 underflow... (buffer allocation problem?)
Should the fix be at a deeper level?


tinygrad/tensor.py
```diff
@@ -3183,7 +3183,8 @@ def cast(self, dtype:DTypeLike) -> Tensor:
     print(t.dtype, t.numpy())
     ```
     """
-    return self if self.dtype == (dt:=to_dtype(dtype)) else F.Cast.apply(self, dtype=dt)
+    if (dt := to_dtype(dtype)) == dtypes.uint8 and self.dtype == dtypes.float: return F.Cast.apply(F.Cast.apply(self, dtype=dtypes.int64), dtype=dt)
```
chenyuxyz, line 3186: i think casting twice to make it correct is fine. but only this pair is too specific. can you check if all float to all uint works, in particular some uint64 should fail because int64 cannot fully represent it

test/test_dtype_alu.py
```diff
@@ -160,6 +165,35 @@ def test_float_cast(self, a, dtype): universal_test_cast(a, dtypes.float32, dtyp
...
+  @given(strat.data(), strat.sampled_from(dtypes_float), strat.sampled_from((dtypes.uint8, dtypes.uint16, dtypes.uint32)))
+  def test_float_cast_to_unsigned_overflow_underflow(self, a, float_dtype, unsigned_dtype):
+    max_value = {dtypes.uint8: 255, dtypes.uint16: 65535, dtypes.uint32: 2**32-1}[unsigned_dtype]
+    float_strat = {dtypes.float16: ht.float16, dtypes.float32: ht.float32, dtypes.float64: ht.float64}[float_dtype]
+    underflow_strat = float_strat.filter(lambda x: x < 0)
+    overflow_strat = float_strat.filter(lambda x: x > max_value)
+    universal_test_cast(a.draw(underflow_strat), float_dtype, unsigned_dtype)
+    universal_test_cast(a.draw(overflow_strat), float_dtype, unsigned_dtype)
+
+  @unittest.skipIf(Device.DEFAULT == "METAL", "uint64 casting is unstable")
```
chenyuxyz, line 188: what does unstable mean here?

tinygrad/tensor.py
```diff
@@ -3350,7 +3350,9 @@ def cast(self, dtype:DTypeLike) -> Tensor:
     print(t.dtype, t.numpy())
     ```
     """
-    return self if self.dtype == (dt:=to_dtype(dtype)) else F.Cast.apply(self, dtype=dt)
+    if dtypes.is_unsigned(dt:=to_dtype(dtype)) and dtypes.is_float(self.dtype):
```
chenyuxyz, line 3353: can you add some comments to this line?

test/test_dtype_alu.py
```diff
@@ -160,6 +165,35 @@ def test_float_cast(self, a, dtype): universal_test_cast(a, dtypes.float32, dtyp
   @given(ht.int32, strat.sampled_from(dtypes_float+dtypes_int+dtypes_bool))
   def test_int32_cast(self, a, dtype): universal_test_cast(a, dtypes.int32, dtype)
 
+  @unittest.skipUnless(is_dtype_supported(dtypes.float64, Device.DEFAULT) and is_dtype_supported(dtypes.float16, Device.DEFAULT),
+                       f"no float64 or float16 on {Device.DEFAULT}")
+  @given(strat.data(), strat.sampled_from(dtypes_float), strat.sampled_from((dtypes.uint8, dtypes.uint16, dtypes.uint32)))
+  def test_float_cast_to_unsigned(self, a, float_dtype, unsigned_dtype):
+    max_value = {dtypes.uint8: 255, dtypes.uint16: 65535, dtypes.uint32: 2**32-1}[unsigned_dtype]
```
chenyuxyz, line 172: dtypes.max and below

test/test_dtype_alu.py
```diff
@@ -160,6 +165,35 @@ def test_float_cast(self, a, dtype): universal_test_cast(a, dtypes.float32, dtyp
   @given(ht.int32, strat.sampled_from(dtypes_float+dtypes_int+dtypes_bool))
   def test_int32_cast(self, a, dtype): universal_test_cast(a, dtypes.int32, dtype)
 
+  @unittest.skipUnless(is_dtype_supported(dtypes.float64, Device.DEFAULT) and is_dtype_supported(dtypes.float16, Device.DEFAULT),
```
chenyuxyz, line 168: instead of skip the whole test, just skip the cases when the casting is not possible (metal can do float16 for example but not float64)

chenyuxyz (commented):
want to get this in?

geohotstan:
oh whoops, I had a long log of local CI failures so I didn't check my recent messages.
Yeah I'll try to get this is, sorry.

geohotstan:
@chenyuxyz 
do you know what version of `numpy` is cached in CI?
my local numpy is producing different numbers when doing overflow/underflow.
even `@reproduce_failure('6.119.3', b'AXicY2JgYGBkQAUNQAwAAlwAhA==')` passes locally

chenyuxyz:
idk what version ci is using, likely < 2, and it's possible numpy changed behavior

geohotstan:
ok this was frustrating :c
There are a bunch of caveats around overflow/underflow casting.
- numpy rounds large float values to nearest express-able float value before even casting (no wrap if float value exceeds largest express-able float) 
- numpy casting to uint8 and uint16 clamps values to min/max of int32
- numpy casting to uint32 and uint64 also first clamps before casting

it's just really hard to deal with all of that and still write decent tests
so I ended up only adding the special case where
`float -> uint8/uint16 and float values in between min/max of int32 are properly wrapped'
this special case should perfectly match with numpy

it is what it is
I hope this is ok :/

chenyuxyz:
what does jax do?

geohotstan:
> what does jax do?

looks like they just clamp to avoid cursed behavior LOL
no wrap around at all
```
In [13]: jnp.array([-1.,-266., 256, 259999., 3., 4.], dtype=jnp.float32).astype('uint8')
Out[13]: Array([  0,   0, 255, 255,   3,   4], dtype=uint8)

In [15]: jnp.array([-1.,-266., 99999999, 259999., 3., 4.], dtype=jnp.float16).astype('uint16')
/opt/homebrew/lib/python3.11/site-packages/jax/_src/numpy/lax_numpy.py:2152: RuntimeWarning: overflow encountered in cast
  out = np.array(object, dtype=dtype, ndmin=ndmin, copy=False)  # type: ignore[arg-type]
Out[15]: Array([    0,     0, 65535, 65535,     3,     4], dtype=uint16)

In [17]: jnp.array(2**32, dtype=jnp.float64).astype('uint32')
<ipython-input-17-d4c819086274>:1: UserWarning: Explicitly requested dtype <class 'jax.numpy.float64'> requested in array is not available, and will be truncated to dtype float32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
  jnp.array(2**32, dtype=jnp.float64).astype('uint32')
Out[17]: Array(4294967295, dtype=uint32)

In [23]: jnp.array(2**64, dtype=jnp.float32).astype('uint64')
/opt/homebrew/lib/python3.11/site-packages/jax/_src/numpy/array_methods.py:66: UserWarning: Explicitly requested dtype uint64 requested in astype is not available, and will be truncated to dtype uint32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
  return lax_numpy.astype(arr, dtype)
Out[23]: Array(4294967295, dtype=uint32)
```

chenyuxyz:
and what if it's on a gpu (similarly for torch uint8)? i think on gpu you don't get data dependent RuntimeWarning?

tinygrad/tensor.py
```diff
@@ -3559,7 +3559,10 @@ def cast(self, dtype:DTypeLike) -> Tensor:
     print(t.dtype, t.numpy())
     ```
     """
-    return self if self.dtype == (dt:=to_dtype(dtype)) else F.Cast.apply(self, dtype=dt)
+    # NOTE: casting from float to uchar and ushort with underflow or overflow values will cause values to wrap around if it's within int32 range
+    if (dt:=to_dtype(dtype)) in {dtypes.uint8, dtypes.uint16} and dtypes.is_float(self.dtype):
```
chenyuxyz, line 3563: swap the order to match the NOTE and the line below

geohotstan:
it has the same clamping behavior on CUDA but if I do it this way it doesn't have eager RuntimeError
Did I do this correctly? lol...
image

jax does mention this difference in behavior from numpy here https://jax.readthedocs.io/en/latest/_autosummary/jax.Array.astype.html#jax.Array.astype
and here
https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#miscellaneous-divergences-from-numpy

Issue #6294, opened 2024-08-27
Safetensor save, load, re-save does not work


PaulGustafson:
Minimal example on iMac with AMD Radeon Pro 5500 XT GPU --

```
from tinygrad import Tensor
from tinygrad.nn.state import safe_save, safe_load, get_state_dict
import os
os.environ['GPU'] = '1'

t1 = Tensor([1, 2, 3, 4, 5])
state_dict = get_state_dict(t1)
safe_save(state_dict, "model.safetensors")
state_dict2 = safe_load("model.safetensors")
safe_save(state_dict2, "model.safetensors")   # Deletes the existing file and does not replace it!
state_dict3 = safe_load("model.safetensors")   # Error: No such file
```

vladov3000:
Here is what is happening:
> safe_save(state_dict, "model.safetensors")

This will create and open(2) a read-write file descriptor `model.safetensors`. 

> safe_save(state_dict2, "model.safetensors")

This second call will begin by unlink(2)ing `model.safetensors`. This will disassociate the name of the file from the data stored.  However, the memory mapping remains cached with the device object. The disk device sees it is already allocated (`if self.size is not None: return`) and so it does not create a new file.

There are several solutions. I think DiskDevice should be robust against files being unlinked underneath it. It should detect if the file is still linked through a `stat` system call, close the old file descriptor, and open a new one. Making a PR.

References: The age old wisdom of RFTM "read the f-ing man pages" applies. `man 2 unlink`.

chenyuxyz:
seems to be fixed on master. feel free to open a new issue

PR #6307, opened 2024-08-28
assert before saving on disk to avoid deleting the file

Isalia20:
Fixes the issue in #6294  by asserting that no tensor should have disk device open before saving. Otherwise it causes the file to delete as described in the issue. Including the code to reproduce from the issue for convenience:
```
from tinygrad import Tensor
import os
os.environ['GPU'] = '1'

t1 = Tensor([1, 2, 3, 4, 5])
```

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -184,8 +184,8 @@ class Compiler:
     return lib
 
 class Compiled:
-  def __init__(self, device:str, allocator:Allocator, renderer:Optional[Renderer], compiler:Optional[Compiler], runtime, graph=None, count=None):
-    self.dname, self.allocator, self.compiler, self.runtime, self.graph, self.count = device, allocator, compiler or Compiler(), runtime, graph, count
+  def __init__(self, device:str, allocator:Allocator, renderer:Optional[Renderer], compiler:Optional[Compiler], runtime, graph=None, count:Optional[int]=None):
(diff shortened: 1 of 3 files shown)


Isalia20:
Get an error on the linter but I think it's a false one:
`tinygrad/nn/state.py:55: error: "Allocator" has no attribute "device"  [attr-defined]`

Should I add type ignore?

geohot:
No. That's real.

Isalia20:
I've changed it to device.count directly. This should work

Isalia20:
should be ready now

tinygrad/device.py
```diff
@@ -184,8 +184,9 @@ def compile_cached(self, src:str) -> bytes:
     return lib
 
 class Compiled:
-  def __init__(self, device:str, allocator:Allocator, renderer:Optional[Renderer], compiler:Optional[Compiler], runtime, graph=None):
-    self.dname, self.allocator, self.compiler, self.runtime, self.graph = device, allocator, compiler or Compiler(), runtime, graph
+  def __init__(self, device:str, allocator:Allocator, renderer:Optional[Renderer],
+               compiler:Optional[Compiler], runtime, graph=None, count=0):
+    self.dname, self.allocator, self.compiler, self.runtime, self.graph, self.count = device, allocator, compiler or Compiler(), runtime, graph, count
```
geohot, line 189: How is count not always 0?

Isalia20: I added it so mypy wouldn't give the error on Device not having the count, but I agree it's not the cleanest of solutions. I'll close this PR now and come up with something better

PR #6308, opened 2024-08-28
add support and tests for nearest modes in interpolate, adapt uint8 bilinear to torch implementation

pedrorohde:
- add support and tests for `nearest` mode in interpolate, equivalent to torch
- add support and tests for `nearest-exact` mode in interpolate, equivalent to torch
- adapt uint8 linear interpolate, making it equivalent to torch
  - adapting torch's [multiplication trick](https://github.com/pytorch/pytorch/blob/2bd02e0c82c6d76157244655256062bade5070ae/aten/src/ATen/native/cpu/UpSampleKernel.cpp#L944) for uint8 interpolation
  - implemented in `lerp` with no cast to float (5757458f5b3b5f5feef5fa87f76cbabf95ec7096)

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -2119,7 +2119,6 @@ class TestOpsUint8(unittest.TestCase):
   def test_cast_relu(self):
     helper_test_op([(2,3,64,64)], lambda x: x.relu().type(torch.uint8), lambda x: x.relu().cast('uint8'), forward_only=True)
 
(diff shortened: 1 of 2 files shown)


geohot:
Cool, this is the first one I have seen to match torch!

geohot:
Congrats! I love when I can just click merge.

george@tinygrad.org to claim the $100 bounty, can pay via PayPal or USDC (on eth)

PR #6309, opened 2024-08-28
Proposal: Better UOps.SWIZZLE

Qazalin:
Docs: http://localhost:8000/developer/uop/#tinygrad.ops.UOps.SWIZZLE.


geohot:
We need some way to get the docs on a staging branch, obviously that localhost link doesn't work unless you run the branch

geohot:
This is how I intended SWIZZLE. Up to you if it makes sense.

PR #6316, opened 2024-08-29
real strides with uops

chenyuxyz:
fixed #6121. not sure what's the status of these, feel free to merge this if it's helpful


geohot:
Do we have enough tests to be confident this is correct? I see a whole lot of process replay changes, which I wouldn't expect if real_strides was outputting the same thing.

chenyuxyz:
will separate adding rules from update real_strides

PR #6328, opened 2024-08-30
Sharded SDXL Inference

tobias17:
## Overview

Brings in changes required to make SDXL inference work across multiple GPUs.

The "normal" inference method takes 2 separate inputs, both the size of your batch_size, and:
1. `Tensor.cat` them together
2. Run the combined input through the model
3. `Tensor.chunk` the output to split it apart again

While this is easy and fast for normal inference, this `cat` and `chunk` causes issues with sharded tensors. To get around this, I added a new `SplitVanillaCFG` class that calls the 2 inputs independently, removing this issue. If there is a way to actually cat and chunk the sharded tensors that is cleaner than this approach, please do explain.

The sampler takes in the guider class as an optional parameter, so normal inference (via `python examples/sdxl.py`) should not be affected by this change, only custom scripts that set the sampler can use the new guider.

## Testing
```
$ PYTHONPATH=. NV=1 python examples/sdxl.py --seed 0
saving /tmp/rendered.png
output validated with distance=0.00289628142490983
```

I also ran SDXL through a custom script that sharded the model & input, and fed the `SplitVanillaCFG` to the denoiser. It was able to generate 24 images in a single batch across 6 GPUs


geohot:
Why are cat and chunk causing issues with sharded Tensors?

A key philsophy of tinygrad is not to work around things like this, but instead to make sure they are fixed. It's not just for you, it's for future you.

tobias17:
> Why are cat and chunk causing issues with sharded Tensors?
> 
> A key philsophy of tinygrad is not to work around things like this, but instead to make sure they are fixed. It's not just for you, it's for future you.

I investigated this further and have reached the conclusion that the cat and chunk is going against how sharded Tensors are handled. While I agree with the philosophy, this is a case of "the framework is not built for this" rather than a "a bug is stopping this".

To not go with the approach in this PR would require either:
1. A rethinking of how multi-GPU tensors are handled in tinygrad in one or more fundamental ways, most likely adding more complexity for what seems to only be a niche issue right now
2. A rewriting of the SDXL code to be "number of batch dimensions" agnostic

For these reasons I will be looking to push this current PR forward.

chenyuxyz:
i think it's fine for now if it unblocks multi gpu stable diffusion. can you open an issue about the fundamental example that is not working for tracking?

PR #6329, opened 2024-08-30
Tests for `_time_program`

jovsa:
No major changes to `_time_program`, but added initial tests.


geohot:
Hmm, I really don't like mocks, they are brittle and make things annoying to change. How can you test for behavior?

jovsa:
> Hmm, I really don't like mocks, they are brittle and make things annoying to change. How can you test for behavior?

@geohot: done - added some tests for behavior.

PR #6330, opened 2024-08-30
conv bw in one kernel with graph_rewrite

Qazalin:
first_reduce swizzles, then AST becomes a `(REDUCE_AXIS, src=(REDUCE_AXIS, )`, so we just need to merge the axes.


geohot:
Cool, I like how understandable this is.

Qazalin:
next: fixup_st should build on the strides of the second LOAD.
image
right now I'm just overriding the st:

PR #6334, opened 2024-09-01
Flux.1

samm393:
This PR adds Flux.1-schnell and T5 (encoder only)

```code
python examples/flux1.py  --prompt="bananas and a can of coke" --seed=0
```
Runs in ~ 45s on a 4090 (weight loading included)
bananas1

```code
python examples/flux1.py  --prompt="bananas and a can of coke" --seed=0 --name="flux-dev"
```
Runs in ~100s on a 4090 (weight loading included)
bananas2


reddyn12:
Got 25.6 sec on M3 Max

geohot:
Solid bananas! Will look over when the PR is out of draft.

examples/flux1.py
```diff
@@ -0,0 +1,962 @@
...
+  q, k = apply_rope(q, k, pe)
+
+  x = Tensor.scaled_dot_product_attention(q, k, v)
+  x = x.rearrange("B H L D -> B L (H D)")
+
+  return x
+
+def rope(pos: Tensor, dim: int, theta: int) -> Tensor:
+  assert dim % 2 == 0
+  scale = Tensor.arange(0, dim, 2, dtype=dtypes.float64, device=pos.device) / dim
```
mattkanwisher, line 38: So for this to work on OSX you need to do this. Would be nice to platform gate this for OSX hardware


  scale = Tensor.arange(0, dim, 2, dtype=dtypes.float32, device=pos.device) / dim

samm393: Thanks for spotting. Reference implementation just uses float64, diffusers implementation downcasts only for metal (as you suggest). Will take a look at how much of a difference it makes (might cause problems for long prompts?)

bghira: it doesn't make much - fp32 looks better sometimes tbh

anthony0727:
May I ask why depend on `sdxl` rather than using the one in tinygrad? https://github.com/tinygrad/tinygrad/blob/44c18a39a51dcdfdf1e53431d7b1f4edbb82a4c8/examples/sdxl.py

geohot:
Testing now!

bghira:
can i open a follow-up pull request to add attention masking on the transformer & real classifier-free guidance? both of which are great for improving the quality of the result.

geohot:
you don't need permission

bghira:
just wasn't sure how involved the examples should be, or if it's considered out-of-scope, but that sounds like it's all good

geohot:
Congrats, bounty is yours! george@itnygrad.org to claim, can pay via PayPal or USDC

PR #6339, opened 2024-09-02
FSDP in tinygrad

CaltropHungerton:
Initial attempt at FSDP. Still thinking about how I want the API to look, currently you put the FSDP decorator in front of the neural network layer definition you want to shard (ex. in front of your transformer block class). I want it to work such that you can shard your batch across devices in a totally seamless manner and have the gradient all-reduction just work alongside FSDP. Currently far from finished.


chenyuxyz:
stale

PR #6365, opened 2024-09-05
real strides with uops [run_process_replay]

chenyuxyz:


chenyuxyz:
merging this to move forward. For symbolic shapes, it can sometimes return None instead of the stride. The corresponding tests are isolated and marked as expectedFailure for TODO. The reason is that UOps div and mod folding expects CONST to work and for symbolic shape it would involve DEFINE_VAR.

None of the llama / gpt2 kernel changed in process replay. And one image process replay gets better because it fixed a case that the old real_strides returned None

Issue #6274, opened 2024-08-24
`total_length` value in `fetch` function is wrong if `gunzip` is set to True


Benetti-Hub:
The total size of the dowloaded content is larger than the one displayed by `total_length` due to compression. This causes the progress bar to overflow:

```terminal
https://storage.googleapis.com/cvdf-datasets/mnist/train-images-idx3-ubyte.gz: 475%|█████████████████████████████████████████████████████████████████████████████████
```

Also it follows that this assert does not work for gzip related dowloads:

```python
if (file_size:=os.stat(f.name).st_size) < total_length: raise RuntimeError(f"fetch size incomplete, {file_size} < {total_length}")
```

Steps to reproduce:
- Delete the cache for tinygrad (or disable it)
- Execute the beautiful_mnist example
```python
python examples/beautiful_mnist.py
```

geohot:
Yea...how do you fix this? Maybe our tqdm can better handle the display?

chenyuxyz:
i set length to None so tqdm does not print the bar in https://github.com/tinygrad/tinygrad/pull/6377. don't know if there's a way to get the correct size directly

PR #6372, opened 2024-09-05
Fix tqdm overflow

Benetti-Hub:
Fixes https://github.com/tinygrad/tinygrad/issues/6274.

I am not sure this solution is the most elegant, but in order to maintain the check on `total_length`, it seems necessary to temporarily retain the file (to my knowledge, the total uncompressed cannot be known a priori for a gzip file).


geohot:
Not remotely worth it. Just fix the bar

PR #6375, opened 2024-09-05
tensor roll

Isalia20:


tinygrad/tensor.py
```diff
@@ -1317,6 +1317,29 @@ def unflatten(self, dim:int, sizes:Tuple[int,...]):
...
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor.rand(3, 4, 1).roll(shifts=-1, dims=0))
+    ```
+    """
+    dims, shifts = (dims, ) if isinstance(dims, int) else dims, (shifts, ) if isinstance(shifts, int) else shifts
+    dims = tuple(i % len(self.shape) for i in dims)
+    all_dims = range(len(self.shape))
+    all_shifts = [shifts[dims.index(i)] % self.shape[i] if i in dims else 0 for i in all_dims]
+    rolled = self
+    for shift, dim in zip(all_shifts, all_dims):
```
geohot, line 1337: enumerate

Isalia20:
Reduced the line count a lot, still using range len for the shape though. I can replace it with enumerate but then I'd just do:
`for i, _ in enumerate()`
for which I don't really see point in doing

tinygrad/tensor.py
Isalia20: Reduced the line count a lot, still using range len for the shape though. I can replace it with enumerate but then I'd just do:
`for i, _ in enumerate()`
for which I don't really see point in doing

tinygrad/tensor.py
```diff
@@ -1317,6 +1317,24 @@ def unflatten(self, dim:int, sizes:Tuple[int,...]):
...
+    print(Tensor.rand(3, 4, 1).roll(shifts=1, dims=0))
+    ```
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor.rand(3, 4, 1).roll(shifts=-1, dims=0))
+    ```
+    """
+    dims, shifts = (dims,) if isinstance(dims, int) else dims, (shifts,) if isinstance(shifts, int) else shifts
+    dims = tuple(i % len(self.shape) for i in dims)
+    all_shifts = [shifts[dims.index(i)] % self.shape[i] if i in dims else 0 for i in range(len(self.shape))]
+    return functools.reduce(lambda t, args: t[tuple(slice(None) if j != args[0] else slice(-args[1], None) for j in range(len(t.shape)))].cat(
```
geohot, line 1335: Unreadable

Isalia20: Is more readable now and tried to keep the lines minimal

geohot:
Merged!

PR #6387, opened 2024-09-06
add UOps.VALID

geohot:


geohot:
@Qazalin I'll hand this off to you?

PR #6388, opened 2024-09-06
move DEFINE_VAR min/max from src to arg

chenyuxyz:
new arg is (Variable, min as CONST, max as CONST). not exactly sure why some PTX kernels became simpler...


chenyuxyz:
something to do with `cmp_tuple` and the use of DEFINE_VAR there. merging given that other CI + benchmarks kernels remained unchanged and PTX became simpler

PR #6391, opened 2024-09-06
new working branch for vectorize folding

geohot:
replaces #6280


geohot:
`VERBOSE=1 GRAPHUOPS=1 RESTRICT_KERNEL=49 PYTHONPATH="." python3 -O test/external/external_benchmark_schedule.py`

Leikoe:
when saying a single value can be broadcasted for the folder rules to still work, do you mean modifying dtypes.as_const() to create a tuple of the val when the val isn't a tuple & the dtype.count > 1 ?

PR #6395, opened 2024-09-06
remove _slice [run_process_replay]

geohotstan:


tinygrad/tensor.py
```diff
@@ -1259,8 +1254,10 @@ def pad2d(self, padding:Sequence[int], value:float=0.0) -> Tensor:
     print(t.pad2d((1, 1, 2, 0), value=-float("inf")).numpy())
     ```
     """
```
geohotstan, line 1256: Just a fyi, currently `pad2d` does a lot of the heavy lifting `_slice` used to do. 
the documentation says `(padding_left, padding_right, padding_top, padding_bottom)` but we use `pad2d` for `len(padding) == 2 or 4 or 6` throughout the codebase.

geohot: Why does pad2d have to shrink?

tinygrad/tensor.py
```diff
@@ -3346,8 +3343,9 @@ def image_conv2d(self, weight:Tensor, bias:Optional[Tensor]=None, groups=1, stri
     else: w = w.reshape(cout//4, H, rcin_hi, W, rcin_lo, 4).permute(0,1,2,3,5,4)
 
     # padding
-    padding_ = [padding]*4 if isinstance(padding, int) else (padding if len(padding) == 4 else [padding[1], padding[1], padding[0], padding[0]])
-    x = x._slice((None, (-padding_[2], x.shape[1]+padding_[3]), (-padding_[0], x.shape[2]+padding_[1]), None, None, None))
+    padding = self._padding2d(padding, 2)
```
geohot, line 3346: What's this? This is worse than _slice I think

tinygrad/tensor.py
geohotstan: negative padding
```
  def test_pad2d(self):
    helper_test_op([(3,3,3,3)], lambda x: torch.nn.functional.pad(x, (1,2,3,4)), lambda x: x.pad2d(padding=(1,2,3,4)))
    helper_test_op([(3,3,3,3)], lambda x: torch.nn.functional.pad(x, (-1,2,-3,4)), lambda x: x.pad2d(padding=(-1,2,-3,4)))
    helper_test_op([(3,3,3,3)], lambda x: torch.nn.functional.pad(x, (1,2,3,4), value=5), lambda x: x.pad2d(padding=(1,2,3,4),value=5))
    helper_test_op([(3,3,3,3)], lambda x: torch.nn.functional.pad(x, (-1,2,-3,4), value=5), lambda x: x.pad2d(padding=(-1,2,-3,4),value=5))
```

geohot:
Merged!

PR #6401, opened 2024-09-07
Move SPLIT_REDUCEOP into kernel.py

ttomsa:
Need to clean up. This way more than 1 extra kernel can be created, for example Tensor.rand(1, 5, 4, 255, 256, 256).sum() creates 2 extra kernels although it currently raises an error.


cpwrs:
My `split` stuff shouldn't be here, the same Kernel should deal with the added `SINK`. 

I've got a branch with this working, gonna make a PR this week if you want to build on that.

PR #6405, opened 2024-09-08
Gated stores as IFs rewrite with beam fix

ianpaul10:
ref https://github.com/tinygrad/tinygrad/pull/5976 & https://github.com/tinygrad/tinygrad/pull/6376

Running `BEAM=4 NV=1 HALF=1 IGNORE_BEAM_CACHE=1 DEBUG=2 python3 extra/gemm/simple_matmul.py`:

this branch: Screenshot 2024-09-08 at 11 05 00 AM

master: 
Screenshot 2024-09-08 at 11 04 37 AM

PR #6413, opened 2024-09-09
test new style gated store rendering

Qazalin:
So PTX fails for these gated stores.

diff if we use an alu gate vs a UOps.IF
image

@ianpaul10 you're fixing this in #6405, can you open a PR with just the ptx diff including this test?


ianpaul10:
Will do. Thank you for writing the tests for it!

PR #6439, opened 2024-09-09
PTX IF render patch

ianpaul10:
Ref https://github.com/tinygrad/tinygrad/pull/6413 & https://github.com/tinygrad/tinygrad/pull/6405

### DIFF
--- a/tinygrad/renderer/assembly.py
+++ b/tinygrad/renderer/assembly.py
@@ -176,13 +176,15 @@ class PTXRenderer(Renderer):
       uop,dtype,src,args = u.op,u.dtype,u.src,u.arg
       if uop is UOps.IF:
         kk(*self.render_bra(f"IF_{r[src[0]][1:]}_{uops.index(u)}", _cast(r[src[0]], dtypes.bool, src[0].dtype, u=u, pred=True)))
+        kk(f"bra ENDIF_{r[src[0]][1:]}_{uops.index(u)};")
+        kk(f"IF_{r[src[0]][1:]}_{uops.index(u)}:")
       elif uop is UOps.BARRIER and self.barrier: kk(self.barrier)
       elif uop is UOps.ENDRANGE:
         kk(self.code_for_op[BinaryOps.ADD](r[src[0]], r[src[0]], "1", dtypes.int, self.types[dtypes.int]),
             self.code_for_op[BinaryOps.CMPLT](pred:=ssa("pred", dtype="pred"), r[src[0]], r[src[0].src[1]], dtypes.int, self.types[dtypes.int]))
         kk(*self.render_bra(f"LOOP_{r[src[0]][1:]}", pred))
       elif uop is UOps.ENDIF:
-        kk(f"IF_{r[src[0].src[0]][1:]}_{uops.index(src[0])}:")
+        kk(f"ENDIF_{r[src[0].src[0]][1:]}_{uops.index(src[0])}:")
       elif uop is UOps.STORE:
         assert src[0].dtype == dtypes.int64, "store isn't int64"
         assert src[1].op is UOps.CONST, f"store isn't const {u}"
--- a/tinygrad/renderer/assembly.py
+++ b/tinygrad/renderer/assembly.py
@@ -63,6 +63,10 @@ ptx_matcher = PatternMatcher([
     lambda root,z: UOp(root.op, root.dtype, root.src[:2] + (z.cast(dtypes.uint8),), root.arg)),
   (UPat(UOps.STORE, name="root", src=(UPat(),UPat(),UPat(),UPat.var("g", dtypes.int))),
     lambda root,g: UOp(root.op, root.dtype, root.src[:3] + (g.cast(dtypes.uint8),), root.arg)),
+  (UPat(UOps.STORE, name="root", src=(UPat(),UPat(),UPat(),UPat(UOps.IF, name="g"))),
+    lambda root,g: UOp(root.op, root.dtype, root.src[:3] +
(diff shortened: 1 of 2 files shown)


tinygrad/renderer/assembly.py
```diff
@@ -177,13 +177,15 @@ def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
       if uop is UOps.IF:
         assert src[0].dtype is not None
         kk(*self.render_bra(f"IF_{r[src[0]][1:]}_{uops.index(u)}", _cast(r[src[0]], dtypes.bool, src[0].dtype, u=u, pred=True)))
+        kk(f"bra ENDIF_{r[src[0]][1:]}_{uops.index(u)};")
```
Qazalin, line 180: why doesn't this use render_bra like above?

Qazalin:
so you en up creating two branches for one IF/ENDIF?

tinygrad/renderer/assembly.py
ianpaul10: Will swap that over

ianpaul10:
Yes, that's correct. The alternative is to have the start of the uop.IF check the inverse of the src[0], and if it's true (i.e. the IF clause isn't met) skip to the uop.ENDIF, but that seemed a bit more complex to me. I'm happy to switch it over to that though if you think it's better to only have the one branch.

Qazalin:
I think a one branch approach is possible, can you try to make it as simple as possible?

I'd also look into what other compilers doo, you can disassemble what CUDA does with DEBUG=6.

Qazalin:
why is it different from `CUDA=1 DEBUG=5 python3 test/test_renderer_failures.py TestPTXFailures.test_gated_store_with_if`

ianpaul10:
It looks like CUDA is doing the inverting when creating the bool variable itself. And then if true, skip over the below block and skip to the `$L__BB0_2`.

PTX is keeping the bool the same as the METAL generated code, and then if it's not true, skips over the below block. Do we want it to instead invert the bool when generating the variable and then just check if that's true to determine whether to proceed?

CUDA:
```
 15 .visible .entry test(
 16 	.param .u64 test_param_0
 17 )
 18 .maxntid 4, 1, 1
 19 {
 20 	.reg .pred 	%p<2>;
 21 	.reg .b32 	%r<3>;
 22 	.reg .b64 	%rd<5>;
 23
 24
 25 	ld.param.u64 	%rd1, [test_param_0];
 26 	mov.u32 	%r1, %tid.x;
 27 	setp.eq.s32 	%p1, %r1, 0;                           // <-- %p1 gets set to true if %r1 == 0, whereas in the METAL gen code and PTX generated code, the if is checking if it's not equal to 0. 
 28 	@%p1 bra 	$L__BB0_2;
 29
 30 	cvta.to.global.u64 	%rd2, %rd1;
 31 	mul.wide.s32 	%rd3, %r1, 4;
 32 	add.s64 	%rd4, %rd2, %rd3;
 33 	mov.u32 	%r2, 1;
 34 	st.global.u32 	[%rd4], %r2;
 35
 36 $L__BB0_2:
 37 	ret;
 38
 39 }
```

PTX:
```
	.reg		.u64 %dat_u64_<1>;
	.reg		.s64 %const_s64_<1>;
	.reg		.s32 %const_s32_<2>;
	.reg		.pred %cast_pred_<1>;
	.reg		.s64 %cast_s64_<1>;
	.reg		.s64 %alu_s64_<2>;
	.reg		.u32 %lidx0;
	ld.param.u64	%dat_u64_0, [data0+0];
	mov.b64		%const_s64_0, 0;
	mov.b32		%const_s32_0, 1;
	mov.b32		%const_s32_1, 2;
	mov.u32		%lidx0, %tid.x;
	setp.ne.b32	%cast_pred_0, %lidx0, 0;                               // <-- %cast_pred_0 get set to true if %lidx0 != 0
	cvt.s64.s32	%cast_s64_0, %lidx0;
	shl.b64		%alu_s64_0, %cast_s64_0, %const_s32_1;
	add.s64		%alu_s64_1, %alu_s64_0, %dat_u64_0;
	@!%cast_pred_0	bra ENDIF_cast_pred_0_10;
	st.global.s32	[%alu_s64_1+0], %const_s32_0;
	ENDIF_cast_pred_0_10:
	ret;
```

Qazalin:
if I'm not merging this it's because this isn't the right place to put the fix.
This must be in graph_rewrite, renderer can't change compute,

I'd also question why `test_arange_big` is correct in master.

ianpaul10:
Ok, sorry I didn't pick up on that sooner. That explanation makes sense.

I set up the alteration as a part of the ptx_matcher which will be called during graph_rewrite. I also made the change in the draft if gated stores rewrite PR to ensure it works for all cases and not just this simplified one https://github.com/tinygrad/tinygrad/pull/6405.

Qazalin:
I'd merge a generic version of the rule for TestOps.test_arange_big.

ianpaul10:
@Qazalin by "a generic version of the rule for TestOps.test_arange_big", do you mean the rewrite rule that converts the CMPNE ALU to an XOR ALU (https://github.com/tinygrad/tinygrad/blob/master/tinygrad/renderer/assembly.py#L46)? I'm a bit confused as to how a generic version of that would work since the ALU could be a parent of other UOps, so I thought we'd only want to invert the ALU that is inside the IF UOp.

Qazalin:
Would that replace the need for the other rule?

ianpaul10:
No, but I don't understand what other rule you could be referring to.

PR #6440, opened 2024-09-09
UOps stats patch

ianpaul10:
ref https://github.com/tinygrad/tinygrad/pull/6405

### DIFF
--- a/test/test_uops_stats.py
+++ b/test/test_uops_stats.py
@@ -228,7 +228,7 @@ class TestStatsOptimized(unittest.TestCase):
     for opt in opts: k.apply_opt(opt)
     p = k.to_program()
     print(p.name, p.op_estimate, p.mem_estimate, p.lds_estimate)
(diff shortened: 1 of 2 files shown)


geohot:
Why do you think this is right?

ianpaul10:
It seems right when the gate is an IF (in this PR https://github.com/tinygrad/tinygrad/pull/6405) but I guess it's now not removing enough in this case when the gates are just the isolated uops.

Should I just add the test in this PR and then have the line removed when switching over to gated stores? I may have misunderstood Qazalin's ask.

Qazalin:
Do you need to change the flopcounter for your bounty?
Isolate the failure here and put up the diff with tests. We wanna see regression tests that fails in master.

ianpaul10:
Ok, sorry I might've misunderstood the original ask.

I isolated the failure on master when measuring the op_estimate that came up during the BEAM search. Is this more of what you had in mind?

PR #6446, opened 2024-09-09
Check disk device before saving

Isalia20:
Reopening #6307 . Compiled has a count added which is usually a parent class for different devices. count will be used by DiskDevice(and incremented where needed by `_might_open` function) and is added to avoid mypy errors(since class _Device's `__getitem__` has a type annotation of returning Compiled).

Open to comments/suggestions if there is a better way to deal with this.

### DIFF
--- a/test/unit/test_disk_tensor.py
+++ b/test/unit/test_disk_tensor.py
@@ -202,6 +202,14 @@ class TestSafetensors(unittest.TestCase):
       assert v.numpy().dtype == tensors[k].dtype
       np.testing.assert_allclose(v.numpy(), tensors[k])
 
+  @unittest.expectedFailure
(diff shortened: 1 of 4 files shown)


geohot:
What increments count?

Isalia20:
`Device[f"disk:{fn}"]` line calls a `__getitem__` which is annotated to return a `Compiled` object. Since `Compiled` originally doesn't have count attribute, mypy thinks that it tries to access attribute which doesn't exist(which really does but only for DiskDevice). I added the count to "silence" mypy. (We can also consider adding noqa which should be cleaner, since then all other devices won't have the `count` attribute)

geohot:
Yea, don't add an unused attribute to silence a typechecker

PR #6447, opened 2024-09-09
Faster UOp hashing

tim-becker:
1. Add `HashEnum` which caches the `__hash__` result for enums
2. Remove the custom `__hash__` implementation on `UOp`: it is effectively equivalent to the builtin `object` one but is a lot slower
3. Use `eq=False` for `NOp` (like `UOp` does)

Combined, PR makes both rewrite and linearize about 10% faster on my machine.

Before (min of 5 samples):
```
PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model tensor in     24.80 ms
***** model schedule in   11.01 ms
***** model opts(53) in  119.97 ms
***** model lower in       8.42 ms
***** model rewrite in   654.21 ms
***** model linearize in 208.02 ms
14967
all 1031.90 ms
```

After (min of 5 samples):
```
PROFILE=0 python3 test/external/external_benchmark_schedule.py
***** model tensor in     24.38 ms
***** model schedule in   10.69 ms
***** model opts(53) in  111.51 ms
***** model lower in       8.23 ms
***** model rewrite in   581.43 ms
***** model linearize in 172.87 ms
14967
all 914.40 ms
```

edit note: Originally I was also using `slots=True` on the `UOp` and `NOp` dataclasses for another 5% gains, but sadly that requires python>=3.10, so I removed it

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 import os, functools, platform, time, re, contextlib, operator, hashlib, pickle, sqlite3, tempfile, pathlib, string, ctypes, sys, gzip
 import itertools, urllib.request, subprocess, shutil, math, json, contextvars
+from enum import Enum
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


geohot:
Cool, yea that custom hash function shouldn't have gotten pushed

PR #6452, opened 2024-09-10
simplify image load with valid

chenyuxyz:
start with a very simple case


chenyuxyz:
maybe this case is too simple, not really seeing a perf diff
image

chenyuxyz:
no longer needed

PR #6454, opened 2024-09-10
_min_max of MUL of 2 non-positive inputs

chenyuxyz:


chenyuxyz:
half-way to fold unrolled arange, red resnet 1% faster
image

PR #6456, opened 2024-09-10
Refactor TC [run_process_replay]

ignaciosica:
This PR is a refactor on tensor cores:
- Unifies tensor core code in kernel
- ~Fixes AMX tensor core into real shape, so hack for counting is no longer needed.~ moved to #6524 
- Enables much cleaner Turing TC (ignaciosica/tinygrad#15) to live in renderer (only +10 LOC with ops_python, cuda and ptx support)


tinygrad/codegen/kernel.py
```diff
@@ -318,26 +318,10 @@ def _apply_tc_opt(self, use_tensor_cores:int, axis:int, opt_level:int) -> bool:
...
-          for (i, sz) in tc.threads: self.apply_opt(Opt(OptOps.UPCAST, tc_opts.axes[i], sz), append_opt=False)
-        elif self.opts.device in {"CUDA", "NV"}:
-          self.apply_opt(Opt(OptOps.UNROLL, tc_opts.axes[2]-self.first_reduce, 8), append_opt=False)
-          self.apply_opt(Opt(OptOps.UNROLL, tc_opts.axes[2]-self.first_reduce, 2), append_opt=False)
-          # NOTE: LOCALS and UPCAST can be swapped here. it doesn't seem faster
-          self.apply_opt(Opt(OptOps.UPCAST, tc_opts.axes[1], 2), append_opt=False)
-          self.apply_opt(Opt(OptOps.UPCAST, tc_opts.axes[0], 2), append_opt=False)
-          for (tc_dim, tc_amt) in tc.threads: self.apply_opt(Opt(OptOps.LOCAL, tc_opts.axes[tc_dim], tc_amt), append_opt=False)
+        for r in tc.reduce_axes: self.apply_opt(Opt(OptOps.UNROLL, tc_opts.axes[2]-self.first_reduce, r), append_opt=False)
+        for (tc_dim,tc_amt) in tc.threads: self.apply_opt(Opt(OptOps.UPCAST if AMX else OptOps.LOCAL, tc_opts.axes[tc_dim], tc_amt), append_opt=False)
```
ignaciosica, line 322: I would much rather check for local here but the problem is that when emulating with python, it "has" local so AMX fails

ignaciosica:
I can't benchmark on all devices, but based on the tests I ran locally (METAL and CUDA with Turing), TC's performance doesn't seem to be slowed down.

ignaciosica:
I know it could be better to separate AMX changes from this PR but they are kind of mutually dependent on each other. If preferred I will separate them into different PR's though.

ignaciosica:
The reduce_axes is the shape of the last dim on the tensor core.
I tried using that shape directly on the TC and calculating dims dynamically but didn't like how it turned out. 

``` python
axes: Tuple[int,int,Tuple[int,...]]
@property
def dims(self) -> Tuple[int,...]: # N, M, K
  return tuple(prod(ax) if isinstance(ax, tuple) else ax for ax in self.axes)
```

ignaciosica:
Currently, every TC unrolls the reduce dim first, which is why it is not included in `opts_seq` and performed first. Nonetheless, it could be the case for a TC in the future where this is not true, the question here is whether to include it or not.

PR #6459, opened 2024-09-10
fix UOp const_factor for ADD [run_process_replay]

chenyuxyz:


chenyuxyz:
it's not really used now because we break add chain first, but simple enough to keep it for completeness

negatives are wrong too will fix separately

Issue #6417, opened 2024-09-09
collapse IDIV/ADD from unrolled arange


geohot:
![image](https://github.com/user-attachments/assets/5a8a03b4-bbf3-4fec-903f-e0c51cc9d0e1)

chenyuxyz:
`DEBUG=4 python3 test/test_arange.py TestArange.test_complexity_w_unroll` to repro

PR #6465, opened 2024-09-11
fold unrolled arange div pattern

chenyuxyz:


chenyuxyz:
#6417, pretty specific matcher though

PR #6471, opened 2024-09-11
add void type to uop

geohot:


tinygrad/ops.py
```diff
@@ -337,9 +337,13 @@ class UOps(HashEnum):
 @dataclass(frozen=True, eq=False)
 class UOp(MathTrait):
   op: UOps
-  dtype: Optional[DType] = None
+  dtype: DType = dtypes.void
   src: Tuple[UOp, ...] = tuple()
   arg: Any = None
+  @classmethod
+  def unwrap_dtype(cls, dtype:Optional[DType]) -> Any:
```
Qazalin, line 344: I need this to let NOp still (safely) accept Optional[DType].
We really need a better abstraction for UOp <-> NOp, I don't think pure inheritance in the answer.

Qazalin: https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype

Qazalin: https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype

Qazalin:
green run_process_replay with -O to skip the type checker:
https://github.com/tinygrad/tinygrad/actions/runs/10808934262/job/29982971486

PR #6483, opened 2024-09-11
Even faster enums

tim-becker:
Python's `IntEnums` are even faster to hash and compare than our `HashEnum`, but are gross because they inherit `int.__str__` and `int.__eq__`. The latter causes some annoying issues, e.g. `EnumA.FOO == EnumB.BAR == some_int` can be true.

Turns out this is fairly easy to work around by overriding `auto()` to be globally unique. Still have to be careful about comparing enums to ints, but tinygrad code doesn't do that.

This makes stuff another ~5% faster.


tim-becker:
Sadly can't do [run_process_replay] because this changes how the enums are pickled

geohot:
Nice! Crazy how much gain there was to be had here

PR #6486, opened 2024-09-12
simple validhack

chenyuxyz:


chenyuxyz:
this idea is good, #6452 is more correct

PR #6500, opened 2024-09-12
Misc rewrite perf improvements

tim-becker:
Each commit is independent and small. If any are deemed not worth the style or line count, they're easy to revert.

Combined, this is about a 20% speedup on my machine.

cc @roelofvandijk for the tip on the listcomp overhead

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 import itertools, functools
-from dataclasses import dataclass, replace
+from dataclasses import dataclass
 from collections import defaultdict
 from typing import Optional, List, Tuple, cast, Dict, Final, DefaultDict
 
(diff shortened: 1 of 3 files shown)


geohot:
All looks good! Overall I'd say readability is a wash. Surprised map is faster than comprehension and `+` is faster than itertools.chain

tim-becker:
FYI `itertools.chain` is faster for large sequences, but these lists are so small that the optimized list code wins out

roelofvandijk:
Great stuff, but the `map` is slower for me

tim-becker:
Interesting. I'm on python 3.11.10. I originally had it as a list comp but figured I'd try a map too, and it was faster. Here are the stats on my machine (in % time spent on that line):
`tuple(self.rewrite(s) for s in n.src)` : 8.2%
`tuple([self.rewrite(s) for s in n.src])`: 6.0%
`tuple(map(self.rewrite, n.src))`: 5.4%

tim-becker:
Oh, that seems to only be true in the profiler... With profiling off, listcomp is 2-3% faster than `map` for me. Reproducible on 3.11 and 3.12.

I'll open a new PR that switches that back to a listcomp

Issue #6378, opened 2024-09-06
image_dot of 2 half inputs returns in half instead of float


chenyuxyz:
`PYTHONPATH="." GPU=1 IMAGE=0 python -m pytest test/test_ops.py -k test_gemm_fp16` passed

`PYTHONPATH="." GPU=1 IMAGE=2 python -m pytest test/test_ops.py -k test_gemm_fp16` failed with `Exception: forward pass failed shape (64, 64): dtype mismatch: tinygrad=float32 | torch=float16`

need to either confirm the desired behavior is to return in float32 and update the test, or correctly cast the return type

geohot:
It should be float16 I think

PR #6506, opened 2024-09-13
cast images to float16 if FLOAT16 is used or image is already float16

ryanlueds:
Correctly cast return type from image_conv2d to float16 if needed. 

Root cause:
`def imageh(shp): return ImageDType(100, 2, "imageh", 'e', 1, shape=shp, base=dtypes.float32)`
will cast value to float32

response to issue #6378


geohot:
Needs a test

PR #6520, opened 2024-09-15
fix typing for test_ops

chenyuxyz:
mostly passed `TYPED=1 python3 -m pytest -n=auto test/test_ops.py`.

one last test specifically set an invalid value to test the exception, and to ignore that we need to import typeguard. And to get a working version of typeguard, we would need to get rid of dependency on tensorflow_addons because it requires a very old version of typeguard

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -118,7 +118,7 @@ class TestOps(unittest.TestCase):
       with self.assertRaises(ValueError): method((2, -3, 0))
 
   def test_negative_dims_full(self):
-    with self.assertRaises(ValueError): Tensor.full(-3, 2)
+    with self.assertRaises(ValueError): Tensor.full((-3,), 2)
     with self.assertRaises(ValueError): Tensor.full((2, -3), 4)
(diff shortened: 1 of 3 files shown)


roelofvandijkO:
Maybe tensorflow_addons can be replaced with tensorflow-models: 
https://github.com/tensorflow/models (tf-models-official)

And then use [this version of LAMB](https://www.tensorflow.org/api_docs/python/tfm/optimization/lamb ) here: https://github.com/tinygrad/tinygrad/blob/6be0cc387ca0c29c019fc27b47126873ae417a41/test/external/external_test_optim.py#L115C1-L115C17

roelofvandijk:
Maybe tensorflow_addons can be replaced with tensorflow-models: 
https://github.com/tensorflow/models (tf-models-official)

And then use [this version of LAMB](https://www.tensorflow.org/api_docs/python/tfm/optimization/lamb ) here: https://github.com/tinygrad/tinygrad/blob/6be0cc387ca0c29c019fc27b47126873ae417a41/test/external/external_test_optim.py#L115C1-L115C17

PR #6522, opened 2024-09-15
Fix typing on CUDA

terafo:
`init_c_struct_t` currently expects to receive `ctypes._SimpleCData`, not it's subclass, like `ctypes.c_ulong`. Therefore, backends that use this function crash with `TYPED=1`.

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 import os, functools, platform, time, re, contextlib, operator, hashlib, pickle, sqlite3, tempfile, pathlib, string, ctypes, sys, gzip
 import itertools, urllib.request, subprocess, shutil, math, json, contextvars
 from dataclasses import dataclass
-from typing import Dict, Tuple, Union, List, ClassVar, Optional, Iterable, Any, TypeVar, TYPE_CHECKING, Callable, Sequence
+from typing import Dict, Tuple, Union, List, ClassVar, Optional, Iterable, Any, TypeVar, TYPE_CHECKING, Callable, Sequence, Type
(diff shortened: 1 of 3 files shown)


nimlgen:
I see it still failing for NV, want to fix?

terafo:
@nimlgen so I fixed that specific issue and figured out what's wrong with typeguard and NV otherwise, but my [fix](https://github.com/tinygrad/tinygrad/pull/6522/commits/d21b47a4e20721429b2c557fa2fc9c4ea0ed5790) is too messy and I think it should be resolved in another PR. This PR still fixes CUDA, which is valuable.

tinygrad/helpers.py
```diff
@@ -305,7 +305,7 @@ def to_mv(ptr, sz) -> memoryview: return memoryview(ctypes.cast(ptr, ctypes.POIN
 def mv_address(mv:memoryview): return ctypes.addressof(ctypes.c_char.from_buffer(mv))
 def to_char_p_p(options: List[bytes], to_type=ctypes.c_char): return (ctypes.POINTER(to_type) * len(options))(*[ctypes.cast(ctypes.create_string_buffer(o), ctypes.POINTER(to_type)) for o in options])  # noqa: E501
 @functools.lru_cache(maxsize=None)
-def init_c_struct_t(fields: Tuple[Tuple[str, ctypes._SimpleCData], ...]):
+def init_c_struct_t(fields: Tuple[Union[Tuple[str, Type[ctypes._SimpleCData]], Tuple[str, Type[ctypes._SimpleCData], int]], ...]):
```
geohot, line 308: Why do you need `Type[`?

terafo: Because what ends up being passed into this function aren't instances of `_SimpleCData` itself, it's instances of it's subclasses, thus typeguard crashes test_ops.

PR #6524, opened 2024-09-15
Fix amx shape [run_process_replay]

ignaciosica:
Fix AMX shape: (sz,sz,sz) -> (sz,sz,1)
AMX no longer has threads

I was wrong when I thought the abstraction only went so far, refactoring TC I realized AMX could be correctly represented with current TC abstraction. Flops counting is fixed now without needing any hack.

This fix was originally on #6456 but I moved all the AMX shape changes here.


geohot:
Yea, I thought 32,32,1 made more sense as the shape

PR #6525, opened 2024-09-15
Don't re-check patterns when uop.arg is None

tim-becker:
If `uop.arg is None`, we're currently checking all relevant patterns twice.

This change is only a 2-3% net speedup on the benchmark because this isn't a super common situation, but I still think this change is justified because it's more "correct". It could also be a bigger speedup with different kernels.


geohot:
Agreed it's more correct

PR #6529, opened 2024-09-16
UOp rewrite const to the end of MUL

chenyuxyz:
similar to ADD


chenyuxyz:
slightly annoying and not really needed for now

PR #6530, opened 2024-09-16
simplex-lt fold

chenyuxyz:
`a0*x0 + a1*x1 + ... < 0 is equivalent to -x0-x1- ... < 0 if xi >= 0 and ai < 0.`


chenyuxyz:
this is not universally faster, similar speed overall

PR #6534, opened 2024-09-16
minimum new style expand

geohot:


tinygrad/codegen/uopgraph.py
```diff
@@ -474,6 +490,29 @@ def _gate_srcs(u:UOp, gate:UOp) -> UOp:
     lambda ex,x,y: UOp(UOps.EXPAND, ex.dtype, tuple((x+y).gep(i) for i in range(256 if AMX else 8)), ex.arg)),
 ])
 
+def no_vectorized_load_store(ls:UOp):
+  idx = ls.src[1]
+  if idx.dtype.count == 1: return None
+  tv = [UOp(ls.op, ls.dtype.scalar(), (ls.src[0],) + tuple(j.gep(i) for j in ls.src[1:])) for i in range(idx.dtype.count)]
+  return UOp(UOps.VECTORIZE if ls.dtype != dtypes.void else UOps.SINK, ls.dtype, tuple(tv))
```
Qazalin, line 497: this should have arg:
```suggestion
  return UOp(UOps.VECTORIZE if ls.dtype != dtypes.void else UOps.SINK, ls.dtype, tuple(tv), ls.arg)
```

geohot: I don't think that's right...that's never the base SINK

PR #6536, opened 2024-09-16
remove logic to vectorize reduces

geohot:
Does this make any kernels faster? It's a reg alloc issue


geohot:
Looks like a wash. Slight gain for comma being removed.

PR #6545, opened 2024-09-16
Metal with CDLL instead of py-objc

mesozoic-egg:
~The `cdll.py` loads the obj-c and metal runtime, and calls objc_msgSend to interact with objects. `api.py` imitates the py-objc API, so only the import needs to be changed when this is ready~

~`support/metal.py` interfaces with DLL libraries~

~Consumer (ops_metal.py, graph/metal.py) calls the send_message to interact with objc instance~

Use ctypes and objc runtime to interact with Mac's GPU. Essentially a simplified implementation of [objc_util](https://omz-software.com/pythonista/docs/ios/objc_util.html#module-objc_util) and [rubicon](https://github.com/beeware/rubicon-objc) but without the method translation step, so the consumer part will call the objective C method explicitly (aka "sending message") and handle the ctypes pointers.


mesozoic-egg:
There are still cases not covered, but you can try out with `METAL=1 PYTHONPATH='.' python script.py`
```python
from tinygrad import Tensor
Tensor.manual_seed(10)

a = Tensor.rand(6, )
b = Tensor.rand(6, )
c = a + b
print(c.numpy())

```

geohot:
Is that `api.py` auotgenerated? I don't want something handcoded, then it counts toward lines

mesozoic-egg:
This approach may not work.. I just realized that some functions are defined inline or as macro, like the
```c++
#  define MTL_INLINE static inline
MTL_INLINE MTLIndirectCommandBufferExecutionRange MTLIndirectCommandBufferExecutionRangeMake(uint32_t location, uint32_t length) API_AVAILABLE(macos(10.14), macCatalyst(13.0), ios(13.0))
{
    MTLIndirectCommandBufferExecutionRange icbRange = {location, length};
    return icbRange;
}
```

so they won't be accessible via ctypes, closing this until I come up with a better solution

tinygrad/runtime/graph/metal.py
```diff
@@ -14,76 +15,80 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
...
-    icb_descriptor = Metal.MTLIndirectCommandBufferDescriptor.new()
-    icb_descriptor.setCommandTypes_(Metal.MTLIndirectCommandType(Metal.MTLIndirectCommandTypeConcurrentDispatch))
-    icb_descriptor.setInheritBuffers_(False)
-    icb_descriptor.setInheritPipelineState_(False)
-    icb_descriptor.setMaxKernelBufferBindCount_(31)
-    self.icb = self.device.device.newIndirectCommandBufferWithDescriptor_maxCommandCount_options_(icb_descriptor, len(self.jit_cache),
-                                                                                                  Metal.MTLResourceOptions(0))
-    if self.icb is None: raise GraphException("create indirect command buffer failed, does your system support this?")
+    icb_descriptor = send_message(libobjc.objc_getClass(b"MTLIndirectCommandBufferDescriptor"), "new")
+    send_message(icb_descriptor, "setCommandTypes:", 32)
```
geohot, line 19: Way less readable

geohot:
This increases line count and reduces readability. Like I have said several times, you are going to need to autogen from the header

tinygrad/runtime/graph/metal.py
```diff
@@ -14,76 +15,80 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
...
+
+    self.icb = send_message(self.device.device, "newIndirectCommandBufferWithDescriptor:maxCommandCount:options:",
+      icb_descriptor, len(self.jit_cache),0)
+    if self.icb.value is None: raise GraphException("create indirect command buffer failed, does your system support this?")
     self.needs_icb_fix = int(type(self.icb).__name__ != "AGXG15XFamilyIndirectCommandBuffer")    # not required on M3
 
     if len(self.vars): self.int_buf = self.device.allocator.alloc(len(self.vars)*dtypes.int32.itemsize)
-    all_resources = [self.int_buf.buf] if len(self.vars) else []
-    all_pipelines = []
+    all_resources, all_pipelines = [self.int_buf.buf] if len(self.vars) else [], []
```
geohot, line 30: Don't mix changes like this

tinygrad/runtime/support/metal.py
```diff
@@ -0,0 +1,39 @@
...
+libobjc.objc_msgSend.restype = objc_id
+libobjc.objc_getClass.restype = objc_id
+libobjc.sel_registerName.restype = objc_id
+libmetal.MTLCreateSystemDefaultDevice.restype = objc_id
+libdispatch.dispatch_data_create.restype = objc_id
+
+def send_message(ptr: objc_id, selector: str, /, *args: Any, restype: type = objc_id) -> Any:
+  sender = libobjc["objc_msgSend"]
+  sender.restype = restype
+  return sender(ptr, libobjc.sel_registerName(selector.encode()), *args)
```
geohot, line 22: Don't call this over and over

tinygrad/runtime/support/metal.py
```diff
@@ -0,0 +1,39 @@
...
+def to_ns_str(s: str) -> objc_id: return send_message(libobjc.objc_getClass(b"NSString"), "stringWithUTF8String:", s.encode())
+
+def to_ns_array(items: List[Any]): return (objc_id * len(items))(*items)
+
+def int_tuple_to_struct(t: Tuple[int, ...], _type: type = c_ulong):
+  class Struct(Structure): pass
+  Struct._fields_ = [(f"field{i}", _type) for i in range(len(t))]
+  return Struct(*t)
+
+def dedup(items: List[objc_id]) -> List[objc_id]:
```
geohot, line 33: Why does this need to be custom?

tinygrad/runtime/support/metal.py
```diff
@@ -0,0 +1,39 @@
+from ctypes import CDLL, c_void_p, Structure, c_ulong
+from ctypes.macholib.dyld import dyld_find  # pyright: ignore
```
geohot, line 2: junk

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,58 @@
...
+libdispatch.dispatch_data_create.restype = objc_instance
+
+def msg(ptr: objc_instance, selector: str, /, *args: Any, restype: type = objc_id) -> Any:
+  sender = libobjc["objc_msgSend"] # Using attribute access returns a new reference so setting restype is safe
+  sender.restype = restype
+  return sender(ptr, libobjc.sel_registerName(selector.encode()), *args)
+
+def to_ns_str(s: str) -> objc_instance: return msg(libobjc.objc_getClass(b"NSString"), "stringWithUTF8String:", s.encode())
+
+def to_ns_array(items: List[Any]): return (objc_instance * len(items))(*items)
```
geohot, line 35: I don't think this function does what it's name says, and maybe you don't need it at all

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,58 @@
...
-import Metal, libdispatch
-from typing import List, Any, Tuple, Optional
-from tinygrad.helpers import prod, getenv, DEBUG, unwrap2
-from tinygrad.device import Compiled, Compiler, CompileError, LRUAllocator
+import os, subprocess, pathlib, tempfile, functools
+from typing import List, Any, Tuple, Optional, cast
+from tinygrad.helpers import prod, getenv, DEBUG
+from tinygrad.device import Compiled, Compiler, LRUAllocator
 from tinygrad.renderer.cstyle import MetalRenderer
+import ctypes
```
geohot, line 7: Move to line 2

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,58 @@
...
+
+class objc_id(ctypes.c_void_p): # This prevents ctypes from converting response to plain int, and dict.fromkeys() can use it to dedup
```
geohot, line 9: These don't have to be released?

tinygrad/runtime/ops_metal.py
```diff
@@ -38,27 +81,28 @@ def __init__(self, device:MetalDevice, name:str, lib:bytes):
...
-    command_buffer = self.device.mtl_queue.commandBuffer()
-    encoder = command_buffer.computeCommandEncoder()
-    encoder.setComputePipelineState_(self.pipeline_state)
-    for i,a in enumerate(bufs): encoder.setBuffer_offset_atIndex_(a.buf, a.offset, i)
-    for i,a in enumerate(vals,start=len(bufs)): encoder.setBytes_length_atIndex_(ctypes.c_int32(a), 4, i)
-    encoder.dispatchThreadgroups_threadsPerThreadgroup_(Metal.MTLSize(*global_size), Metal.MTLSize(*local_size))
-    encoder.endEncoding()
-    command_buffer.commit()
+    if prod(local_size) > msg(self.pipeline_state, "maxTotalThreadsPerThreadgroup", restype=ctypes.c_ulong):
+      raise RuntimeError("local size too big")
```
geohot, line 94: Don't change functionality

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,58 @@
...
+
+def to_ns_str(s: str) -> objc_instance: return msg(libobjc.objc_getClass(b"NSString"), "stringWithUTF8String:", s.encode())
+
+def to_ns_array(items: List[Any]): return (objc_instance * len(items))(*items)
+
+def int_tuple_to_struct(t: Tuple[int, ...], _type: type = ctypes.c_ulong):
+  class Struct(ctypes.Structure): pass
+  Struct._fields_ = [(f"field{i}", _type) for i in range(len(t))]
+  return Struct(*t)
+
```
geohot, line 41: I'm not sure these need classes, also line 44 is wrong

tinygrad/runtime/ops_metal.py
```diff
@@ -20,11 +63,11 @@ def compile(self, src:str) -> bytes:
       # NOTE: if you run llvm-dis on "air" you can see the llvm bytecode
       air = subprocess.check_output(['xcrun', '-sdk', 'macosx', 'metal', '-x', 'metal', '-c', '-', '-o', '-'], input=src.encode('utf-8'))
       return subprocess.check_output(['xcrun', '-sdk', 'macosx', 'metallib', '-', '-o', '-'], input=air)
-    options = Metal.MTLCompileOptions.new()
-    options.setFastMathEnabled_(getenv("METAL_FAST_MATH"))
-    try: library = unwrap2(self.device.device.newLibraryWithSource_options_error_(src, options, None))
-    except AssertionError as e: raise CompileError(e) from e
```
geohot, line 26: Where'd this exception go? Can you make sure there's tests for all exceptions?

tinygrad/runtime/ops_metal.py
```diff
@@ -69,46 +113,47 @@ def __init__(self, device:MetalDevice):
     self.device:MetalDevice = device
     super().__init__()
   def _alloc(self, size:int, options) -> MetalBuffer:
-    ret = self.device.device.newBufferWithLength_options_(size, Metal.MTLResourceStorageModeShared)
-    if ret is None: raise MemoryError(f"Metal OOM while allocating {size=}")
+    ret = msg(self.device.device, "newBufferWithLength:options:", size, 0, restype=objc_instance)
+    if ret.value is None: raise RuntimeError("Metal failed to allocate buffer")
```
geohot, line 117: Why did this exception change?

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,58 @@
...
+libmetal = load_library("/Library/Frameworks/Metal.framework/Metal")
+# Must be loaded for default Metal Device: https://developer.apple.com/documentation/metal/1433401-mtlcreatesystemdefaultdevice?language=objc
+load_library("/Library/Frameworks/CoreGraphics.framework/CoreGraphics")
+libdispatch = load_library("/usr/lib/libSystem.dylib") # libdispatch is part of libSystem on mac
+libobjc.objc_getClass.restype = objc_id
+libobjc.sel_registerName.restype = objc_id
+libmetal.MTLCreateSystemDefaultDevice.restype = objc_instance
+libdispatch.dispatch_data_create.restype = objc_instance
+
+def msg(ptr: objc_instance, selector: str, /, *args: Any, restype: type = objc_id) -> Any:
```
geohot, line 28: use a typevar, this returns restype, right?

tinygrad/runtime/graph/metal.py
```diff
@@ -1,89 +1,97 @@
...
-    if len(self.vars): self.int_buf_view = self.int_buf.buf.contents().as_buffer(self.int_buf.buf.length()).cast('i')
-    self.range = Metal.MTLIndirectCommandBufferExecutionRangeMake(0, len(self.jit_cache))
+    if len(self.vars): self.int_buf_view = self.device.allocator.as_buffer(self.int_buf).cast('i')
+    self.range = int_tuple_to_struct((0, len(self.jit_cache)), ctypes.c_ulong)
 
   def __call__(self, input_rawbuffers: List[Buffer], var_vals: Dict[Variable, int], wait=False) -> Optional[float]:
+
     if self.command_buffer is not None and self.command_buffer in self.device.mtl_buffers_in_flight: wait_check(self.command_buffer)
-    all_resources = dedup(self.all_resources + [x._buf.buf for x in input_rawbuffers])
+    all_resources = self.all_resources + [x._buf.buf for x in input_rawbuffers]
```
geohot, line 58: Why did this change?

tinygrad/runtime/ops_metal.py
mesozoic-egg: The intended use case for objc_id is for functions that return something that cannot be released in the runtime, for example objc_getClass returns a class object that's loaded, sel_registerName registers a name in the runtime (repeated call return the same reference); would a different name make it clearer?

tinygrad/runtime/ops_metal.py
mesozoic-egg: I tried to replace this with just `(_type * len(t))(*t)`, despite having the exact same bytes it caused the program to hang when passed to `dispatchThreadgroups:`, haven't figured out how the reason though

tinygrad/runtime/graph/metal.py
```diff
@@ -1,71 +1,85 @@
...
-    self.needs_icb_fix = int(type(self.icb).__name__ != "AGXG15XFamilyIndirectCommandBuffer")    # not required on M3
+    icb_descriptor = msg(libobjc.objc_getClass(b"MTLIndirectCommandBufferDescriptor"), "new", restype=objc_instance)
+    msg(icb_descriptor, "setCommandTypes:", MTLIndirectCommandType.MTLIndirectCommandTypeConcurrentDispatch)
+    msg(icb_descriptor, "setInheritBuffers:", False)
+    msg(icb_descriptor, "setInheritPipelineState:", False)
+    msg(icb_descriptor, "setMaxKernelBufferBindCount:", 31)
+
+    self.icb = msg(self.device.device, "newIndirectCommandBufferWithDescriptor:maxCommandCount:options:",
+      icb_descriptor, len(self.jit_cache), MTLResourceOptions.MTLResourceCPUCacheModeDefaultCache, restype=objc_instance)
+    icb_label = bytes(msg(msg(self.icb, "description", restype=objc_id), "UTF8String", restype=ctypes.c_char_p)).decode()
```
geohot, line 33: swap lines 33 and 34?

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,53 @@
...
 from tinygrad.renderer.cstyle import MetalRenderer
 
+class objc_id(ctypes.c_void_p): # This prevents ctypes from converting response to plain int, and dict.fromkeys() can use it to dedup
+  def __hash__(self): return hash(self.value)
+  def __eq__(self, other): return self.value == other.value
+
+class objc_instance(objc_id): # method with name "new", "alloc" should be freed after use
+  def __del__(self): msg(self, "release")
+
+def load_library(path: str): return ctypes.CDLL(path)
```
geohot, line 15: Why is this a function, just call ctypes.CDLL?

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,53 @@
...
+T = TypeVar("T")
+# Ignore mypy error reporting incompatible default, because typevar default only works on python 3.12
+def msg(ptr: objc_id, selector: str, /, *args: Any, restype: type[T] = objc_id) -> T: # type: ignore [assignment]
+  sender = libobjc["objc_msgSend"] # Using attribute access returns a new reference so setting restype is safe
+  sender.restype = restype
+  return sender(ptr, libobjc.sel_registerName(selector.encode()), *args)
+
+def to_ns_str(s: str): return msg(libobjc.objc_getClass(b"NSString"), "stringWithUTF8String:", s.encode())
+
+def int_tuple_to_struct(t: Tuple[int, ...], _type: type = ctypes.c_ulong):
```
geohot, line 40: `to_struct` with an `*args` param would be cleaner I think? there's no reason this is int it has a type arg

geohot:
Can bump line count to 9800 and remove pyobjc from setup.py

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,53 @@
...
+libobjc.sel_registerName.restype = objc_id
+libmetal.MTLCreateSystemDefaultDevice.restype = objc_instance
+libdispatch.dispatch_data_create.restype = objc_instance
+
+T = TypeVar("T")
+# Ignore mypy error reporting incompatible default, because typevar default only works on python 3.12
+def msg(ptr: objc_id, selector: str, /, *args: Any, restype: type[T] = objc_id) -> T: # type: ignore [assignment]
+  sender = libobjc["objc_msgSend"] # Using attribute access returns a new reference so setting restype is safe
+  sender.restype = restype
+  return sender(ptr, libobjc.sel_registerName(selector.encode()), *args)
```
geohot, line 36: you need to cache sel_registerName

geohot:
`OSError: dlopen(/Library/Frameworks/Metal.framework/Metal, 0x0006): tried: '/Library/Frameworks/Metal.framework/Metal' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Library/Frameworks/Metal.framework/Metal' (no such file), '/Library/Frameworks/Metal.framework/Metal' (no such file)`

geohot:
Timing on second run of test/test_ops.py:

This branch `Ran 307 tests in 16.648s`
master `Ran 307 tests in 16.635s`

mesozoic-egg:
Regarding OS error dyld not found, maybe the `dyld_find` is required after all... [source](https://github.com/python/cpython/blob/ad7c7785461fffba04f5a36cd6d062e92b0fda16/Lib/ctypes/macholib/dyld.py#L121)
```python
def dyld_find(name, executable_path=None, env=None):
    """
    Find a library or framework using dyld semantics
    """
    for path in dyld_image_suffix_search(chain(
                dyld_override_search(name, env),
                dyld_executable_path_search(name, executable_path),
                dyld_default_search(name, env),
            ), env):

        if os.path.isfile(path):
            return path
        try:
            if _dyld_shared_cache_contains_path(path):
                return path
        except NotImplementedError:
            pass

    raise ValueError("dylib %s could not be found" % (name,))
```

The path I have were tested on M2 and Intel mac (both mac os 14), it's possible the path differs on different system and the dyld_find should compensate that

mesozoic-egg:
I don't think performance would be gained much, because pyobjc calls the same dyld files under the hood. 

This PR is more like a weight shred.

All the python objc bridge projects call the same underlying OS dyld files, so speed should be roughly the same. They all add a lot of code to make the syntax feel more pythonic, and even more code to make the objc reference counting work alongside python's garbage collector. This PR didn't do that which is why it's less readable and some syntax appear weird.

pyobjc differs in that it has compiled code that make memory access easier. One notable example is the `as_buffer` used before. pyobjc writes that as an C extension to fill a block memory directly [Source here](https://github.com/ronaldoussoren/pyobjc/blob/79eb3f96548ec9e4b466166b9fb1101f89fd4869/pyobjc-core/Modules/objc/objc-class.m#L93).  In this PR I had to do some roundable way to ctypes `memoryview((ctypes.c_char * (src.offset + src.size)).from_address(ptr.value)).cast("B")` to make it work.

But the fact pyobjc is compiled means most people are just downloading a precompiled version of it, so it may not work with all the package manager, so things might break even when the underlying dyld is performing correctly

Leikoe:
why is this bounty locked ? I closed my previous pr because @geohot said he wanted autogenerated bindings & I'm doing that right now and just saw this was locked

Leikoe:
> I'm not sure performance would be gained much, because pyobjc calls the same dyld files under the hood. This PR is more like a weight shred.
> 
> All the python objc bridge projects I have seen (objc-util, rubicon-objcc, pyobjc) all the same underlying OS dyld files, so speed should be roughly the same. They all add a lot of code to make the syntax feel more pythonic, and even more code to make the objc reference counting work alongside python's garbage collector. This PR didn't do that which is why it's less readable, some syntax appear weird and I also had to do manual memory management.
> 
> pyobjc differs in that it has compiled code that make memory access easier. One notable example is the `as_buffer` used before. pyobjc writes that as an C extension to fill a block memory directly ([source here](https://github.com/ronaldoussoren/pyobjc/blob/79eb3f96548ec9e4b466166b9fb1101f89fd4869/pyobjc-core/Modules/objc/objc-class.m#L93)). In this PR I had to do some roundable way to ctypes `memoryview((ctypes.c_char * (src.offset + src.size)).from_address(ptr.value)).cast("B")` to make it work.
> 
> But the fact pyobjc is compiled means most people are just downloading a precompiled version of it, so the release may not catch up with all the OS versions. Things might break even when the underlying objc-c runtime and python runtime are performing correctly.

are you sure you are releasing the created objects ? pyobjc & rubicon do it automatically and you only do it on the MTLDevice & NSStrings

tinygrad/runtime/ops_metal.py
```diff
@@ -38,27 +88,31 @@ def __init__(self, device:MetalDevice, name:str, lib:bytes):
           print("Error running disassembler: Make sure you have https://github.com/dougallj/applegpu cloned to tinygrad/extra/disassemblers/applegpu")
     assert lib[:4] == b"MTLB", "Invalid Metal library. Could be due to using conda. Try system python or METAL_XCODE=1 DISABLE_COMPILER_CACHE=1."
     data = libdispatch.dispatch_data_create(lib, len(lib), None, None)
-    self.library = unwrap2(self.device.device.newLibraryWithData_error_(data, None))
```
mesozoic-egg, line 41: all the places with that take error should probably be passed an explicit ctypes.byref(objc_instance(0)). And when it signals an error, do 
```
bytes(msg(msg(compileError, "localizedDescription", restype=objc_instance),
                                                              "UTF8String", restype=ctypes.c_char_p)).decode()
``` to convert it to something readable and then raise

**Comment by mesozoic-egg** [comment]
@Leikoe , `objc_instance` has a `__del__` that sends "release" message to the pointer, so when the ptr variable gets garbage collected it will also get released in objc. Are there specific places you find that's not handling this properly?

File: tinygrad/runtime/ops_metal.py [hunk 20]
```diff
@@ -1,15 +1,64 @@
 from __future__ import annotations
 import os, subprocess, pathlib, ctypes, tempfile, functools
-import Metal, libdispatch
-from typing import List, Any, Tuple, Optional
-from tinygrad.helpers import prod, getenv, DEBUG, unwrap2
+from typing import List, Any, Tuple, Optional, cast, TypeVar
+from tinygrad.helpers import prod, getenv, DEBUG
 from tinygrad.device import Compiled, Compiler, CompileError, LRUAllocator
 from tinygrad.renderer.cstyle import MetalRenderer
+from ctypes.macholib.dyld import dyld_find
```
geohot, line 7: Why do you need this? Just `/System` isn't universal?

mesozoic-egg: i thought dyld_find is safer, but just read that /System is the core part of the OS so it should be universal

geohot:
Is this ready?

mesozoic-egg:
Yes @geohot

tinygrad/runtime/ops_metal.py
```diff
@@ -1,15 +1,61 @@
...
+
+class objc_instance(objc_id): # method with name "new", "alloc" should be freed after use
+  def __del__(self): msg(self, "release")
+
+@functools.lru_cache(None)
+def sel(name: str): return libobjc.sel_registerName(name.encode())
+
+class MTLResourceOptions:
+  MTLResourceCPUCacheModeDefaultCache = 0
+  MTLResourceStorageModeShared = 0 << 4
```
geohot, line 20: This `0 << 4` is not correct, and subtle bugs are super dangerous. I'm now blocking this on parsing the constants from the headers.

geohot: Wait, Apple might just be dumb

geohot:
Congrats, bounty is yours! george@tinygrad.org to claim, can pay via PayPal or USDC (on eth)

Leikoe:
> I think it will make more sense with all the options listed out below where the expanded form is `MTLResourceStorageModeShared = MTLStorageModeShared << MTLResourceStorageModeShift`
> 
> 
> 
> ```c++
> 
> #define MTLResourceCPUCacheModeShift            0
> 
> #define MTLResourceCPUCacheModeMask             (0xfUL << MTLResourceCPUCacheModeShift)
> 
> 
> 
> #define MTLResourceStorageModeShift             4
> 
> #define MTLResourceStorageModeMask              (0xfUL << MTLResourceStorageModeShift)
> 
> 
> 
> #define MTLResourceHazardTrackingModeShift      8
> 
> #define MTLResourceHazardTrackingModeMask       (0x3UL << MTLResourceHazardTrackingModeShift)
> 
> 
> 
> typedef NS_OPTIONS(NSUInteger, MTLResourceOptions)
> 
> {
> 
>     MTLResourceCPUCacheModeDefaultCache  = MTLCPUCacheModeDefaultCache  << MTLResourceCPUCacheModeShift,
> 
>     MTLResourceCPUCacheModeWriteCombined = MTLCPUCacheModeWriteCombined << MTLResourceCPUCacheModeShift,
> 
> 
> 
>     MTLResourceStorageModeShared  = MTLStorageModeShared << MTLResourceStorageModeShift,
> 
>     MTLResourceStorageModeManaged  = MTLStorageModeManaged << MTLResourceStorageModeShift,
> 
>     MTLResourceStorageModePrivate  = MTLStorageModePrivate << MTLResourceStorageModeShift,
> 
>     MTLResourceStorageModeMemoryless  = MTLStorageModeMemoryless << MTLResourceStorageModeShift,
> 
>     
> 
>     MTLResourceHazardTrackingModeDefault  = MTLHazardTrackingModeDefault << MTLResourceHazardTrackingModeShift,
> 
>     MTLResourceHazardTrackingModeUntracked = MTLHazardTrackingModeUntracked << MTLResourceHazardTrackingModeShift,
> 
>     MTLResourceHazardTrackingModeTracked  = MTLHazardTrackingModeTracked << MTLResourceHazardTrackingModeShift,
> 
>     
> 
>     // Deprecated spellings
> 
>     MTLResourceOptionCPUCacheModeDefault        = MTLResourceCPUCacheModeDefaultCache,
> 
>     MTLResourceOptionCPUCacheModeWriteCombined  = MTLResourceCPUCacheModeWriteCombined,
> 
> } ;
> 
> ```
> 
> 
> 
> I agree copying just the `0 << 4` looks quite dangerous... I will make a follow up to populate the header constants 

Since I still care about tinygrad as a project, here is how:
- parse the bridge support file, its just xml. You will find both enum values & structs.

PR #6547, opened 2024-09-16
Move locals before upcasts METAL

ignaciosica:
This is a pre-req to tensor cores refactor (#6456) to have green process-replay, where to unify kernel TCs code, locals needs to go before upcasts as AMD requires local first.

Based on local benchmarking, performance doesn't seem to be compromised.


geohot:
Why can't the refactor support both?

ignaciosica:
It definitively could. Based on local benchmarking I ran for METAL and CUDA I saw no impact on performance so I saw an opportunity for unifying (AMD already has local first and AMX doesn't have local (the only platform I didn't benchmark was Intel)). Kernel code ended pretty clean. First unrolls, then local and finally upcast.

geohot:
I don't want to change this, I would rather this be an option in a refactor that passes process replay

PR #6551, opened 2024-09-17
bring back webgpu

andredaprato:
- CI tests succeed locally, and so does test_search (which can't run in CI). I
don't have access to a machine with METAL, so I don't know if any errors are
specific to CI, but I'll try to get access to one.
- Implemented bool tensors as `i32` using rewrite rules. Allocation is not
  great, and it makes some tests slow, but we don't use the `Allocator` for
  exported models. Suggestions here would be appreciated.
- Tweaked compile_efficient to make it work again.

We can add f16 support when wgpu supports it. It should come soon https://github.com/gfx-rs/wgpu/pull/5701.


tinygrad/dtype.py
```diff
@@ -19,6 +21,9 @@ def vec(self, sz:int):
     if sz == 1 or self.name == 'void': return self  # void doesn't vectorize, and sz=1 is scalar
     return DType(self.priority, self.itemsize*sz, f"{INVERSE_DTYPES_DICT[self.name]}{sz}", None, sz)
   def scalar(self) -> DType: return DTYPES_DICT[self.name[:-len(str(self.count))]] if self.count > 1 else self
+  def __eq__(self, x):
+    return x and (self.priority, self.itemsize, self.name, self.fmt, self.count) == (x.priority, x.itemsize, x.name, x.fmt, x.count)
+  def __lt__(self, x): return (self.priority, self.itemsize, self.name, self.fmt, self.count) < (x.priority, x.itemsize, x.name, x.fmt, x.count)
```
andredaprato, line 26: I think this should be a separate PR.

When applying the rewrite rules we end up comparing a `DType` to a `PtrDType`.
This isn't supported with the current implementation because
`dataclass(order=True)` enforces that arguments have the same type. What is the
desired behavior here? Should comparing `PtrDType` to `DType` be the same as
comparing `DType` and `DType`? This is how `__eq__` was implemented.

geohot:
Bounty locked. A whole bunch of stuff in here should be standalone prereqs.

geohot:
Still working on this? I closed that standalone PR which changed types because it's definitely possible to do better

andredaprato:
Yes I'm going to keep working on it this week. I appreciate your feedback on that PR, and I agree it can be done better.

geohot:
Still working?

geohot:
I see webgpu test failing

andredaprato:
I haven't been working on this much, but I pushed changes to merge this branch with the new cstyle (which is really nice).

The merge also made threefry the default RNG.  This broke randomness functions since threefry uses uint64 in `_threefry_random_bits`, so a lot of things are broken right now.

Something also broke fancy indexing tests, but it works with `NOOPT=1`. I'm going to look at recent commits to see what's causing this, but it might just be a rendering issue.

I haven't been able to fix the errors on Metal; they are all related to inf/nan. Webgpu allows handling of runtime infs and nans to be implementation specific (https://www.w3.org/TR/WGSL/#differences-from-ieee754), and they can yield indeterminate values. Testing locally on Vulkan these tests pass fine, but I'm not sure if it's possible to work around this on Metal. It's probably sensible to skip these cases since we can't guarantee behavior anyways. We skipped a couple inf/nan tests with webgpu before; there are just more tests with inf/nan cases now.

I don't know what to do about threefry yet; maybe we could create our own int64 type in wgsl. If someone else wants to work on these issues and polish things, they're welcome to, as I'm not sure how much I'm going to keep working on this.

PR #6552, opened 2024-09-17
other side of simple out of bound valid case

chenyuxyz:
462 -> 455


chenyuxyz:
would need to refactor SPECIAL with DEFINE_VAR next

PR #6576, opened 2024-09-18
use int32 in tests where possible

andredaprato:
I don't like changing int8 to int32 but we won't be able to run these tests on webgpu without changing it. The rest of the changes avoid implicitly creating long tensors due to numpy/torch initialization which we do in most places already.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -66,7 +66,7 @@ def helper_test_op(shps, torch_fxn, tinygrad_fxn=None, atol=1e-6, rtol=1e-3, gra
 
 def prepare_test_op(low, high, shps, vals, forward_only=False):
   if shps is None:
-    ts = [torch.tensor(x, requires_grad=(not forward_only)) for x in vals]
+    ts = [(t:=torch.tensor(x, requires_grad=(not forward_only))).to(torch.int32 if not torch.is_floating_point(t) else None) for x in vals]
   else:
     np.random.seed(0)
     np_data = [np.random.uniform(low=low, high=high, size=size).astype(_to_np_dtype(dtypes.default_float)) for size in shps]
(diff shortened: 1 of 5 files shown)


geohot:
webgpu doesn't support int8?

andredaprato:
No, only 32 bit scalars and f16 are supported in the spec, https://www.w3.org/TR/WGSL/#host-shareable-types.

test/test_ops.py
```diff
@@ -66,7 +66,7 @@ def compare(s, tinygrad_output, torch_output, atol, rtol):
 
 def prepare_test_op(low, high, shps, vals, forward_only=False):
   if shps is None:
-    ts = [torch.tensor(x, requires_grad=(not forward_only)) for x in vals]
+    ts = [(t:=torch.tensor(x, requires_grad=(not forward_only))).to(torch.int32 if not torch.is_floating_point(t) else None) for x in vals]
```
geohot, line 69: I really don't like this change, I can't predict what it does

test/test_uops_stats.py
```diff
@@ -17,48 +17,48 @@ def get_stats(x:Tensor):
 
 class TestMemoryCount(unittest.TestCase):
```
geohot, line 18: This should be a unit test anyway

geohot:
I don't like PRs like this. This is not an improvement to tinygrad, this is a weird prereq that's changing tests.

How can you do this stuff while improving tinygrad? If a test doesn't test anything accelerator specific, it should be in test/unit then you don't have to worry about it.

test/test_rearrange_einops.py
```diff
@@ -164,7 +164,7 @@ def test_rearrange_ellipsis_ops(self):
       ("a b c d e -> b (a c d) e", "a b ... e -> b (a ...) e"),
     ]
 
-    xnp = np.arange(2 * 3 * 4 * 5 * 6).reshape([2, 3, 4, 5, 6])
+    xnp = np.arange(2 * 3 * 4 * 5 * 6, dtype=np.int32).reshape([2, 3, 4, 5, 6])
```
geohot, line 167: This test should be refactored, not have uint32 spammed everywhere

Issue #5985, opened 2024-08-08
vits example download blocked by google drive virus scan


AlexCheema:
In the vits example (also used in other examples like conversation), the download is blocked for `ljs` and `vctk` because Google redirects to a page warning about virus scan. https://github.com/tinygrad/tinygrad/blob/df44a4e861ff25d9552f82d1164aaa36c3d4abd4/examples/vits.py#L662-L663

I guess easy fix is to upload somewhere else and update the links.

This is what the links return:

<img width="525" alt="Screenshot 2024-08-08 at 19 28 00" src="https://github.com/user-attachments/assets/00dea84d-80d2-4d76-b8bf-5d9d0791c66f">

<img width="507" alt="Screenshot 2024-08-08 at 19 28 08" src="https://github.com/user-attachments/assets/5eaa555d-2be2-4f5b-bb57-1307a691f773">

chenyuxyz:
#10688 updated for `vctk`

PR #6616, opened 2024-09-20
add gdown in helper to solve issue #5985

domijin:
Fixing [issue](https://github.com/tinygrad/tinygrad/issues/5985) by using gdown for model file hosted in google drive. 

image


geohot:
Tons of lines for a specific solution. Answer: don't fetch from google drive

domijin:
How about
1. reusing existing function: https://github.com/domijin/tinygrad/blob/036c2f5b2650b3b4a67966971b4c2234114faa24/extra/datasets/wikipedia_download.py#L10
2. consolidate the two functions to reduce the change and make it more readable
3. someone host the file elsewhere

Do nothing will keep the issue open and make the showcase example not working, which really discourage me at the beginning.

PR #6335, opened 2024-09-02
whisper long batch

DKormann:
couple conceptual cleanups for whisper inference + support for long batching

working towards completely merging #5922

PR #6617, opened 2024-09-20
remove unimplemented streaming to fix whisper example

domijin:
Fixing the TypeError in decoder method, as streaming param is not implemented. 

image


domijin:
> Why didn't the test catch this? 

There is no test. It is simplily missed out from this [PR](https://github.com/tinygrad/tinygrad/pull/6335/files#diff-44a76d30f5fe1f3da5b76b95d8e27755b73db342128938562e4165cdf4cda5f7L340).

What would you do given this context?

PR #6619, opened 2024-09-20
removed valid in openpilot conv

chenyuxyz:


chenyuxyz:
16.86ms for 0.9.4 and 19.62ms for 0.9.7

will check the 45 valids left to see find a more generic way to write this

chenyuxyz:
16.87ms for 0.9.4 and 19.77ms for 0.9.7. conv is slightly faster but still quite slow for non-valid reasons

PR #6641, opened 2024-09-21
const_like type fix

chenyuxyz:
`Tuple[ConstType, ...]` instead of `Tuple[ConstType]`


geohot:
Hmm, how did this pass mypy?

PR #6649, opened 2024-09-22
fix wrong directory access for bert vocab in mlperf/model_eval.py

anuraglamsal:
The Bert model in extra/models downloads the vocab in extra/weights, but model_eval.py searches for the vocab in the top tinygrad directory. Trivial but noticed it, so.


geohot:
I assume you tested this?

anuraglamsal:
Yes I did.

PR #6651, opened 2024-09-22
test: add UOp.alu_arg [run_process_replay]

Qazalin:
to replace:
```
alu_op: BinaryOps = op.arg[0]
```


Qazalin:
meh, not worth it

PR #6655, opened 2024-09-23
Use Tensor.randint in test_embedding

andredaprato:
A little simpler and respects tinygrad defaults using int32 instead of int64.


chenyuxyz:
this is not simpler, it just changed the test for no good reason

andredaprato:
It allows the test to run when using webgpu. Should I just include this in the webgpu pr instead of considering it a prereq?

PR #6669, opened 2024-09-23
relax idx simplification given valid

chenyuxyz:
apply to kernels in op 0.9.7.
if a valid has a complicated expr, we cannot drop valid but it's possible to simplify idx given valid


chenyuxyz:
image

PR #6690, opened 2024-09-23
FastKAN example

mdaiter:
This implements a FastKAN, detailed here: https://arxiv.org/abs/2405.06721

Super quick to train! Trains on MNIST in here. Also, I've tested the Attention transformer module included in here as well. Not sure if I should break this out into a separate file, or I should keep it here.


Kinvert:
Worked for me, 3070ti.

loss:   0.51 test_accuracy: 85.03%: 100%|████████████████████████████████| 70/70 [00:23<00:00,  3.04it/s]

PR #6695, opened 2024-09-24
viz fold const nodes and UOp/float4 syntax highlight

Qazalin:
image


Qazalin:
VALID looks like this
image

PR #6699, opened 2024-09-24
do not always update float dims

nimlgen:


tinygrad/engine/jit.py
```diff
@@ -80,9 +80,11 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
     mem_estimate: sint = 0
     lds_estimate: sint = 0
 
+    def is_sym_dim(dim) -> bool: return not all(isinstance(d, int) or isinstance(d, float) for d in dim)
```
chenyuxyz, line 83: you can do `isinstance(d, (int, float))`

PR #6721, opened 2024-09-25
Global buffers in kernel

ttomsa:
To be able to move split_reduceop to kernel.py we need kernel to create a new global buffer to split the reduce op. This let's kernel add to the global buffers. It also removes bufs_from_lin() in get_kernel.


ttomsa:
Is there a consensus on how kernel should add a global buffer? This way doesn't work because it only adds to one schedule item.

PR #6725, opened 2024-09-25
schedule with buffer count limit (LazyBuffer version)

Qazalin:


Qazalin:
big graph will do this instead.

Issue #6689, opened 2024-09-23
Environment variables for back-end/accelerators are confusing.


haraschax:
https://docs.tinygrad.org/env_vars/ describes env variables as enabling a certain backends. But it just changes the default. That makes a command like this nonsense:
```
CUDA=1 NV=1 CLANG=1 python -c "from tinygrad import Device; print(Device.DEFAULT)"
```
This API seems a lot better:
```
DEFAULT_BACKEND=GPU python -c "from tinygrad import Device; print(Device.DEFAULT)"
```

Would also be cool for a command like this to work:
```
python -c "from tinygrad import Device; print(Device.available_device())"
```

geohot:
`DEFAULT_BACKEND` is so long!

I'd support a PR for `get_available_backends`, it should be one line

chenyuxyz:
`DEV=GPU python -c "from tinygrad import Device; print(Device.DEFAULT)"` works now

haraschax:
nice!

PR #6733, opened 2024-09-25
add get_available_backends()

pixqc:
`python -c "from tinygrad import Device; print(Device.get_available_backends())"`
['DSP', 'LLVM', 'PYTHON', 'CUDA', 'NV', 'QCOM', 'CLANG', 'METAL', 'NPY', 'AMD', 'DISK', 'HIP', 'GPU']

ref https://github.com/tinygrad/tinygrad/issues/6689#issuecomment-2370459624

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -37,6 +37,7 @@ class _Device:
           return device
       except Exception: pass
     raise RuntimeError("no usable devices")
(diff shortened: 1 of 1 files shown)


geohot:
What do you think this does?

pixqc:
Thought this oneliner would suffice, should've paid more attention, sorry for having wasted your time 🙏

PR #6740, opened 2024-09-25
use validhack logic to rewrite buffer idx

chenyuxyz:
saved a whopping one mod in the conv backward kernel...


chenyuxyz:
no seeing huge perf gain but the kernel diff looks simpler

PR #6744, opened 2024-09-25
use shl everywhere

geohot:


tinygrad/renderer/assembly.py
```diff
@@ -35,14 +35,6 @@ def render_val(x, dtype):
 supports_half: List[Op] = [UnaryOps.EXP2, BinaryOps.ADD, BinaryOps.MUL, BinaryOps.MAX, BinaryOps.CMPLT, TernaryOps.WHERE]
```
hmihaylov-sc, line 35: unused var

tinygrad/renderer/assembly.py
```diff
@@ -35,14 +35,6 @@ def render_val(x, dtype):
...
 shiftable_consts = set([2**i for i in range(64)])
```
hmihaylov-sc, line 36: unused var

PR #6747, opened 2024-09-26
[MLPerf] Prepare openimages dataset script

francislata:
**Overview**
This PR adds support for an optional `BASE_DIR` env. variable to place the openimages dataset at a specific location. In addition, it performs basic checks to see if a specific dataset type is already available from the `BASE_DIR` location.

The download script usage is as follows:
```python
BASE_DIR=<path> python3 extra/datasets/openimages.py
```

By default, it will download __both__ training and validation sets.

**Remaining task(s)**
- [x] Verify `MODEL=retinanet python3 examples/mlperf/model_eval.py` produces the same result as before:
Screenshot 2024-09-26 at 18 00 54

### DIFF
--- a/examples/mlperf/model_eval.py
+++ b/examples/mlperf/model_eval.py
@@ -94,11 +94,11 @@ def eval_retinanet():
     x /= input_std
     return x
 
-  from extra.datasets.openimages import openimages, iterate
+  from extra.datasets.openimages import download_dataset, iterate, BASEDIR
   from pycocotools.coco import COCO
   from pycocotools.cocoeval import COCOeval
   from contextlib import redirect_stdout
-  coco = COCO(openimages('validation'))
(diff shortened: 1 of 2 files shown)


chenyuxyz:
adding `BASE_DIR` is fine. can we remove the pandas dependency? ideally we want to move more things into tinygrad and worry less about what makes pandas dataframe fast

francislata:
> adding `BASE_DIR` is fine. can we remove the pandas dependency? ideally we want to move more things into tinygrad and worry less about what makes pandas dataframe fast
> 
> i suggest don't do that in the same PR and don't change pandas stuff in this pr either

sounds good to me @chenyuxyz. in the meantime, i've reverted the pandas related changes i made initially to unblock this PR.

PR #6752, opened 2024-09-26
even simpler PatternMatcher gc test

Qazalin:
This is because the gc doesn't cleanup pdict after rewrite.

When DEFINE_GLBOAL arg is Buffer, it'll hold on to that Buffer forever.


geohot:
Why delete the others? Merge this?

Qazalin:
meh, I like the old one, it's more comprehensive.

PR #6758, opened 2024-09-26
use where in dropout

chenyuxyz:
should save memory since we only store mask in bool instead of the upcasted used in mul


chenyuxyz:
run in <https://wandb.ai/chenyuxyz/MLPerf-BERT/runs/h3v1brer?nw=nwuserchenyuxyz>, a few % faster. i don't have a systematic way to think about this, but it's faster and bert is our only serious example that uses dropout so i'd like to build on top of this

PR #6771, opened 2024-09-26
add get_available_backends

geohotstan:
I couldn't get it down to a 1-liner...

```
python -c "from tinygrad import Device; print(Device.get_available_backends())"
['METAL', 'GPU', 'CLANG', 'LLVM']
```

since the order is implicit in `["METAL", "AMD", "NV", "CUDA", "QCOM", "GPU", "CLANG", "LLVM"]` I think `device = available_backends[0]` is ok? :s


geohot:
Opens every device

tinygrad/device.py
```diff
@@ -28,16 +28,18 @@ def __get_canonicalized_item(self, ix:str) -> Compiled:
     return ret
   @property
   def default(self) -> Compiled: return self[self.DEFAULT]
+  def get_available_backends(self) -> List[str]: return list(self._get_available_backends())
+  def _get_available_backends(self) -> Iterator[str]:
```
geohot, line 32: why? just iterator

geohotstan:
```
python -c "from tinygrad import Device; print(Device.get_available_backends())"
['METAL', 'GPU', 'CLANG', 'LLVM']
```

can't print it
or am I dumb... 😢

ryandorrington:
#6689

PR #6788, opened 2024-09-27
[MLPerf] Add focal loss for RetinaNet

francislata:
**Overview**
This PR adds the `sigmoid_focal_loss` that will be used as part of RetinaNet training.

To test correctness, I brought over the implementation from [MLPerf's reference implementation](https://github.com/mlcommons/training/blob/cdd928d4596c142c15a7d86b2eeadbac718c8da2/single_stage_detector/ssd/model/focal_loss.py) and tested a few cases to ensure correctness.


chenyuxyz:
given that this adds something new that's only used by retinanet, let's review these after you have a retinanet training run (the openimage one is an improvement to the existing pipeline). easier for reviewer to review and i think it's more flexible for you too

francislata:
sounds good to me @chenyuxyz. i'll close this for now and will reopen once the training portion is successful.

PR #6793, opened 2024-09-28
Construct c_ulong in blitCommandEncoder copy method

mesozoic-egg:
This method isn't run on CI so it wasn't caught, but running `METAL=1 pytest test/test_jit.py::TestJit::test_jitted_transfers`, or the following two examples, would cause the program to hang, because `copyFromBuffer:sourceOffset:toBuffer:destinationOffset:size:` expects NSUInteger (c_ulong) rather than NSInteger (the default c_int/c_long). Interestingly though other methods that expect NSUInteger didn't experience an error like this. 

- [A low level transfer example](https://gist.github.com/mesozoic-egg/530649deca9ef490eb7b0fbe3066d118)

Sharding example (intra device, just for demo):
```python

from tinygrad import Tensor, Device
import numpy as np

np.random.seed(0)
data = np.random.rand(10).astype(np.float32)

a = Tensor(data)
a = a.shard(['METAL:0', 'METAL:1'], axis=0)
a.realize()
```

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -135,7 +135,8 @@ class MetalAllocator(LRUAllocator):
     dest_dev.synchronize()
     src_command_buffer = msg(src_dev.mtl_queue, "commandBuffer", restype=objc_instance)
     encoder = msg(src_command_buffer, "blitCommandEncoder", restype=objc_instance)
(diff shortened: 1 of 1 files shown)


mesozoic-egg:
parsing the headers and let `msg` do the appropriate casting would be a good idea..

geohot:
Can you add a regression test to CI?

geohot:
@mesozoic-egg still waiting for regression test

mesozoic-egg:
See #6919 for the (failed) attempt on adding CI regression test. Runner behavior deviates from real device. Currently stuck..

PR #6799, opened 2024-09-29
match torch rmsnorm implementation

ryandorrington:


geohot:
Does the old test still pass? Can you leave them both?

tinygrad/nn/__init__.py
```diff
@@ -291,11 +291,16 @@ class RMSNorm:
...
+  def __init__(self, normalized_shape: Union[int, Tuple[int, ...]], eps: float = 1e-5, elementwise_affine: bool = True):
+    self.normalized_shape = (normalized_shape,) if isinstance(normalized_shape, int) else tuple(normalized_shape)
+    self.axis, self.eps, self.elementwise_affine = tuple(-1-i for i in range(len(self.normalized_shape))), eps, elementwise_affine
+    self.weight = Tensor.ones(*self.normalized_shape) if elementwise_affine else None
 
-  def __call__(self, x:Tensor) -> Tensor: return self._norm(x.float()).cast(x.dtype) * self.weight
+  def __call__(self, x: Tensor) -> Tensor:
+    assert self.normalized_shape == x.shape[-len(self.normalized_shape):], f"Last dimensions of {x.shape} must match {self.normalized_shape}"
+    x = x * (x.square().mean(axis=self.axis, keepdim=True) + self.eps).rsqrt()
+    if not self.elementwise_affine: return x
```
geohot, line 302: `return x * self.weight if self.weight is not None else x`

tinygrad/nn/__init__.py
```diff
@@ -291,11 +291,16 @@ class RMSNorm:
   print(norm(t).numpy())
   ```
   """
-  def __init__(self, dim, eps=1e-6): self.eps, self.weight = eps, Tensor.ones(dim)
-
-  def _norm(self, x:Tensor): return x * (x.square().mean(-1, keepdim=True) + self.eps).rsqrt()
```
geohot, line 296: Was this function overridden anywhere? Why change this?

ryandorrington: _norm specifically? The _norm method was only used within __call__ in both tinygrad and the referenced llama implementation. Integrating the normalization code directly into __call__ stylistically aligns better with the other nn layers.

ryandorrington:
Should RMSNorm be given a functional implementation on Tensor? GroupNorm and InstanceNorm don't have functional implementations but they are sort of layernorm wrappers. Could be argued that RMSNorm is sufficiently different.

ryandorrington:
@geohot can you please review this PR

geohot:
Going to merge this if CI passes, sorry I missed it

geohot:
Hmm, tests failing.

geohot:
Merged! Think it might be able to be refactored to share code with LayerNorm though

PR #6802, opened 2024-09-29
Compute FID Score

tobias17:
## Overview

Implements the code for computing a FID score from a batch of Inception activations.

### DIFF
--- a/extra/models/inception.py
+++ b/extra/models/inception.py
@@ -308,10 +308,10 @@ class FidInceptionV3:
     return x
 
   def compute_score(self, inception_activations:Tensor, val_stats_path:str) -> float:
-    if self.m1 is None or self.s1 is None:
+    if self.m1 is None and self.s1 is None:
       with np.load(val_stats_path) as f:
         self.m1, self.s1 = f['mu'][:], f['sigma'][:]
-      assert self.m1 is not None and self.s1 is not None
+    assert self.m1 is not None and self.s1 is not None
 
     m2 = inception_activations.mean(axis=0).numpy()
     s2 = np.cov(inception_activations.numpy(), rowvar=False)
--- a/extra/models/inception.py
+++ b/extra/models/inception.py
@@ -4,6 +4,8 @@ from tinygrad.nn.state import load_state_dict, torch_load
 from tinygrad.helpers import fetch
 
 from typing import Optional, Dict
+import numpy as np
+from scipy import linalg
 
 # Base Inception Model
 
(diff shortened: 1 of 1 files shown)


extra/models/inception.py
```diff
@@ -301,3 +306,37 @@ def __call__(self, x:Tensor) -> Tensor:
       lambda x: Tensor.avg_pool2d(x, kernel_size=(8,8)),
     ])
     return x
+
+  def compute_score(self, inception_activations:Tensor, val_stats_path:str) -> float:
+    if self.m1 is None or self.s1 is None:
```
chenyuxyz, line 311: it overwrites both m1, s1 even if one is None and another is not, this might cause silent issue. checking both are None to enter this path and check both are not None after this block

chenyuxyz (commented):
have you verified with a reference or how did you test this?

tobias17:
> have you verified with a reference or how did you test this?

yes, I ran this code against the saved images from an inference generation run. as increased the number of images being used the FID score went down until I used all the images and got numbers in the ballpark of what the mlperf inference documentation stated.

extra/models/inception.py
tobias17: I do not understand what your suggestion is. These are designed to be loaded in pairs, meaning if you load 1 you should be loading the other associated with it. Setting just 1 is invalid when you do not know where the other is coming from.

chenyuxyz: `meaning if you load 1 you should be loading the other associated with it`. the implementation does not enforce this

tobias17: I am not following, the logic as it's written is that if either is None then set both of them and make sure they are both non-None afterwards. The code always loads both at the same time and makes sure we do not get into a weird state were halves are not properly loaded.

chenyuxyz: If m1 and s1 will always both be None or both be non-None, then `if self.m1 is None and self.s1 is None:` is equivalent. It has the benefit to express what's the state expected here. We expect both to be None, not any of those to be None. So it reads weird to me that the starting condition of this block is different from the end condition

Also having the assert check outside of if block separate the setup from the validation.

tobias17: ok this makes sense, updated the code as this is much more clean

extra/models/inception.py
```diff
@@ -4,6 +4,8 @@
 from tinygrad.helpers import fetch
 
 from typing import Optional, Dict
+import numpy as np
+from scipy import linalg
```
chenyuxyz, line 8: i think this is fine for now, but as tinygrad itself does not have these dependency, think more about if an import needs to be in the top level

PR #6804, opened 2024-09-29
Multireduce Check smem usage

0xtimmy:
This PR changes the smem calculation in `kernel.py` used to check if the kernel in within the accelerators `shared_max`.

Kernels produce one temp buffer per reduce, sized to hold the data for that reduce. (re: the `local_shape` calculation in `fixup_ast` in `kernel.py`

So instead of a product of all local dims, this pr calculates the product of dims used in each reduce then sums them up and multiplies them with the `local_sz`, `acc_sz`, and `upcast_sz`


tinygrad/codegen/kernel.py
```diff
@@ -389,8 +389,21 @@ def apply_opt(self, opt:Opt, append_opt:bool=True):
...
+      grouped_sz = 0
+      for sh0,sh1 in [(self.sts[r*2+len(self.bufs)].shape, self.sts[r*2+len(self.bufs)+1].shape) for r in range(len(self.reduceops))]:
+        reduce_sz = prod(s1 for s0,s1 in list(zip(sh0, sh1))[self.first_reduce:self.first_reduce+self.group_for_reduces] if s0 != s1)
+        if sh0[axis] != sh1[axis]: reduce_sz *= amt
+        grouped_sz += reduce_sz
+
+      # another way to do it in fewer lines but a bit less clear
+      # grouped_sz = sum(prod(s1 for s0,s1 in list(zip(sh0, sh1))[self.first_reduce:self.first_reduce+self.group_for_reduces] if s0 != s1) \
+      #   * (amt if sh0[axis] != sh1[axis] else 1) \
+      #   for sh0,sh1 in [(self.sts[r*2+len(self.bufs)].shape, self.sts[r*2+len(self.bufs)+1].shape) for r in range(len(self.reduceops))])
```
0xtimmy, line 404: this is another way to calculate the smem usage that saves lines but is a little less clear

tinygrad/codegen/kernel.py
```diff
@@ -389,8 +389,21 @@ def apply_opt(self, opt:Opt, append_opt:bool=True):
...
       upcast_sz = prod([a for a,b in zip(self.full_shape[self.first_upcast:], self.sts[0].shape[self.first_upcast:]) if a == b])
-      local_sz = prod(self.full_shape[self.first_reduce-self.local_dims:self.first_reduce+self.group_for_reduces])
-      smem_sz = amt*acc_sz*upcast_sz*local_sz
+      local_sz = prod(self.full_shape[self.first_reduce-self.local_dims:self.first_reduce])
+
+      # simplest way to do the sum imo
+      grouped_sz = 0
+      for sh0,sh1 in [(self.sts[r*2+len(self.bufs)].shape, self.sts[r*2+len(self.bufs)+1].shape) for r in range(len(self.reduceops))]:
+        reduce_sz = prod(s1 for s0,s1 in list(zip(sh0, sh1))[self.first_reduce:self.first_reduce+self.group_for_reduces] if s0 != s1)
+        if sh0[axis] != sh1[axis]: reduce_sz *= amt
```
0xtimmy, line 398: account for amt if this opt is a group

test/test_linearizer.py
```diff
@@ -307,6 +307,48 @@ def test_partial_opt_multireduce(self):
         if i == 0: continue
         assert ranges[i-1] != u, f"multireduce nested the ranges! {ranges[i-1], {u}}"
 
+  @unittest.skipIf(CI and Device.DEFAULT in {"AMD"}, "AMD CI doesn't support multiple sync threads yet")
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_shared, "test requires shared")
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "test requires locals")
+  def test_multireduce_smem_usage(self):
```
0xtimmy, line 313: this test makes sure that:
1) the new smem checker correctly calculates smem usage
2) that it still throws a `KernelOptError` if max_shared is exceeded

0xtimmy: It is SUPER SLOW because it uses so many upcasts to drive smem usage up to the boundary without pushing the thread group boundaries. I'm not sure if it makes sense to include in tests for that reason but I do think it's at least relevant to test the PR

test/test_linearizer.py
```diff
@@ -307,6 +307,48 @@ def test_partial_opt_multireduce(self):
...
+    first_reduce = UOp(UOps.REDUCE_AXIS, dtypes.float, (first_x,), (BinaryOps.ADD, (2,)))
+    second_x = UOp(UOps.LOAD, dtypes.float, (g1, x.lazydata.st.reshape((U, N, 1)).to_uop()))
+    second_reduce = UOp(UOps.REDUCE_AXIS, dtypes.float, (first_reduce+second_x,), (BinaryOps.ADD, (1,)))
+    store = UOp(UOps.STORE, src=(g0, ShapeTracker.from_shape((U, 1, 1,)).to_uop(), second_reduce))
+    sink = UOp(UOps.SINK, src=(store,))
+    wanna_output = (x.numpy()+x.numpy().sum(axis=(1,), keepdims=True)).sum(axis=(1,)).reshape(U,1,1)
+
+    # before the PR this will raise an error when it doesn't need to
+    # num bytes stored in temp buffers = 2 ^ floor(log2(max_smem / 4 / (8 + 8))) * 4 * (8 + 8) <= max_smem
+    safe_opts = [
```
0xtimmy, line 332: make sure the checker works regardless of what order the opts come in

tinygrad/codegen/kernel.py
```diff
@@ -389,8 +389,21 @@ def apply_opt(self, opt:Opt, append_opt:bool=True):
                                       (self.group_for_reduces and opt.op not in {OptOps.NOLOCALS, OptOps.PADTO})):
       acc_sz = self.reduceop.dtype.itemsize
       upcast_sz = prod([a for a,b in zip(self.full_shape[self.first_upcast:], self.sts[0].shape[self.first_upcast:]) if a == b])
-      local_sz = prod(self.full_shape[self.first_reduce-self.local_dims:self.first_reduce+self.group_for_reduces])
```
0xtimmy, line 392: when there are multiple grouped reduceops, a local buffer of the same size if made for each to the size of the reduce
(local buffers are made for reduceops in [kernel.py](https://github.com/tinygrad/tinygrad/blob/6a73ad89a2e51b5a14cfe480a2f6d9d9f3f81ad5/tinygrad/codegen/kernel.py#L691) )

so this calculation is no longer correct. We do still need to get the product of `self.full_shape[self.first_reduce-self.local_dims:self.first_reduce`. Buf for `self.full_shape[self.first_reduce:self.first_reduce+self.group_for_reduces]` we need to sum up the products of each reduce's axis

ex. for  `self.full_shape[self.first_reduce:self.first_reduce+self.group_for_reduces] = (4, 4)`  we might have a kernel:
```
extern "C" __global__ void __launch_bounds__(16) r_8_8_2_2_2_2_16_16_8_8(float* data0, const float* data1) {
  __shared__ float temp0[4];
  __shared__ float temp1[4];
...
```
in which `(4 + 4) * sizeof(float)` local memory is used instead of `4 * 4 * sizeof(float)`

0xtimmy: tensor core optimizations are mutually exclusive with grouping (re [kernel.py](https://github.com/tinygrad/tinygrad/blob/6a73ad89a2e51b5a14cfe480a2f6d9d9f3f81ad5/tinygrad/codegen/kernel.py#L404)) and tensor core optimization also do not affect `self.group_for_reduces`, so this PR shouldn't change anything about how tensor cores are scheduled

PR #6815, opened 2024-09-30
Fix tensor saving and loading twice.

vladov3000:
Solves #6294. See [this comment](https://github.com/tinygrad/tinygrad/issues/6294#issuecomment-2382370590) for the details. TLDR; disk devices keep around unlinked files and don't create new files.

This is my first contribution, so there may be a few rough edges. Namely, mypy has poor data flow analysis for `Optional`, so I used sentinel values instead. And no, it doesn't make sense for `self.fd` to be 0 since `mmap`ing stdin will give an error. Also, don't know if multiple assignment is ok here.

I reproduced the given script and got it working. I also added a unit test for this edge case.

Only 1 new line of code + 3 lines of test code.

### DIFF
--- a/tinygrad/runtime/ops_disk.py
+++ b/tinygrad/runtime/ops_disk.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 import os, sys, mmap, _posixshmem, io, ctypes, ctypes.util, platform, contextlib
-from typing import Optional, Generator, Tuple, Callable, List
+from typing import Generator, Tuple, Callable, List
 from tinygrad.helpers import OSX, round_up
 from tinygrad.device import Compiled, Allocator
 from tinygrad.runtime.autogen import io_uring, libc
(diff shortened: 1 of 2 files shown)


tinygrad/runtime/ops_disk.py
```diff
@@ -71,21 +71,22 @@ class DiskDevice(Compiled):
   def __init__(self, device:str):
     if not DiskDevice._tried_io_uring_init: self._iouring_setup()
 
-    self.size: Optional[int] = None
-    self.count = 0
+    self.size = self.fd = self.count = 0
```
geohot, line 74: Why did you remove a type? Don't save lines like this!!

vladov3000: I removed the type because I don’t think annotating the type for an int assignment is necessary. The line above, ‘self.count = 0’ does not have a type annotation either. 

For more context, I changed the type of self.size from Optional[int] to an int where size = 0 instead of size = None represents that the file has not been memory mapped yet. This is because extra asserts/casts/ifs are required to work around mypy and there is a lack of helper functions like map or apply from Haskell. So I used an int and cut down on the code size. If you aren’t ok with this, let me know and I will use Nones instead of 0s in future PRs.

PR #6822, opened 2024-10-01
Wolpertinger agents added to examples

mdaiter:
Hey all!

Added an example about Wolpertinger agents to the repo. I wanted to grow out our fledgling RL example base. Wolpertinger agents seemed complex enough to demonstrate hierarchical RL while still being able to navigate reasonably. This runs as a standalone repo (I've got a copy in my own repo).

Getting great performance on `CartPole-v1`. It takes about 5 seeds, but usually it'll get to the maximum reward in about 500 steps. Pendulum and others also run reasonably.

I've got a few questions:
1. Where should I put the pyflann repo? There are custom fixes that needed to be landed in there, and pyflann (from what I can tell) isn't being maintained anymore.
2. Should these examples be as standalone as this? There isn't a good way of condensing this algorithm down to the bare essentials past this.
3. Is there any way to squeeze more performance out of this? I'm running the training step through `@TinyJit`.

Thanks all


chenyuxyz:
This should be in its own repo, not the main tinygrad repo.

For kernel speed you can start by checking if all kernels are jitted/method cached and not recompiled in each step (`DEBUG=2` shows that), then use kernel search by adding `BEAM` env var. I recommend start with `BEAM=2`

mdaiter:
🫡 thanks @chenyuxyz ! I'll check it out.

mdaiter:
(Also, while I've got you here @chenyuxyz , I was gonna go after porting world models and CW-VAEs next. Any recs for what should be its own repo and when that distinction happens?)

Issue #4206, opened 2024-04-18
cl_khr_fp16 is assumed to work on all hardware, even when this is not the case


softcookiepp:
When using certain older GPUs via the Rusticl backend (I am using a RX 570 for this,) the following error shows up when trying to perform any operations on float16 tensors:

`RuntimeError: OpenCL Compile Error

input.cl:1:26: warning: unsupported OpenCL extension 'cl_khr_fp16' - ignoring [-Wignored-pragmas]
input.cl:3:3: error: use of undeclared identifier 'half4'
input.cl:4:15: error: unknown type name 'half4'
input.cl:4:36: error: use of undeclared identifier 'half4'
input.cl:4:44: error: use of undeclared identifier 'val0'
input.cl:4:51: error: use of undeclared identifier 'val0'
input.cl:4:60: error: use of undeclared identifier 'val0'
input.cl:4:67: error: use of undeclared identifier 'val0'
input.cl:4:76: error: use of undeclared identifier 'val0'
input.cl:4:83: error: use of undeclared identifier 'val0'
input.cl:4:92: error: use of undeclared identifier 'val0'
input.cl:4:99: error: use of undeclared identifier 'val0'
Error executing LLVM compilation action.`

Here is the code that can be used to reproduce this bug provided you have the hardware:

`from tinygrad import Tensor, dtypes
t = Tensor([1, 2, 3, 4], dtype = dtypes.float16)
(t + t).realize()`

Given that cl_khr_fp16 is an optional OpenCL extension, it may be better to not assume that it exists on all backends, and to use type promotion or similar means of eliminating the issue.

chenyuxyz:
implicitly upcasting to float32 if half is not available is bad, but maybe we can support a global flag to say all half should be interpret as float

sirhcm:
fixed in #16220

PR #6846, opened 2024-10-02
correctly cast images to float16 + check if float16 is supported on GPU device

ryandorrington:
#6506
#6378
#4206


geohot:
I support this direction, but tests are failing

ryandorrington:
Thoughts on adding DISABLE_FLOAT16, as mentioned by @chenyuxyz in #4206, to upcast all float16 to float32?

Would something like STRICT_FLOAT=1, which forces dtype to be DEFAULT_FLOAT and makes it immutable, be a better approach?

chenyuxyz:
close as stale

PR #6855, opened 2024-10-02
Make self.fd Optional.

vladov3000:
Right now, `self.fd` is sometimes in `DiskDevice`. This messes up mypy/pylint, because those tools work poorly with attributes that are added and removed dynamically. Hence, we make it an `Optional` field instead.

Necessary for my fix for #6294.

### DIFF
--- a/tinygrad/engine/realize.py
+++ b/tinygrad/engine/realize.py
@@ -125,10 +125,11 @@ class BufferCopy(Runner):
     else: name = f"{type(self).__name__[6:].lower()} {total_sz:8d}, {dest_device[:7]:>7s} <- {src_device[:7]:7s}"
     super().__init__(colored(name, "yellow"), dest_device, 0, total_sz)
   def copy(self, dest, src):
-    disk_supports_fast_copyout = src.device.startswith("DISK") and hasattr(src.allocator.device, 'io_uring') and hasattr(src.allocator.device, 'fd')
-    if src.device.startswith("DISK") and hasattr(dest.allocator, 'copy_from_disk') and disk_supports_fast_copyout and src.nbytes >= 4096:
+    is_disk = src.device.startswith("DISK")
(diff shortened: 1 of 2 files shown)


geohot:
So hasattr for this is also checked in realize, and that will change that behavior. Is this okay?

tinygrad/engine/realize.py
```diff
@@ -125,10 +125,11 @@ def __init__(self, total_sz, dest_device, src_device):
...
     super().__init__(colored(name, "yellow"), dest_device, 0, total_sz)
   def copy(self, dest, src):
-    disk_supports_fast_copyout = src.device.startswith("DISK") and hasattr(src.allocator.device, 'io_uring') and hasattr(src.allocator.device, 'fd')
-    if src.device.startswith("DISK") and hasattr(dest.allocator, 'copy_from_disk') and disk_supports_fast_copyout and src.nbytes >= 4096:
+    is_disk = src.device.startswith("DISK")
+    disk_supports_fast_copyout = is_disk and hasattr(src.allocator.device, 'io_uring') and getattr(src.allocator.device, 'fd', None) is not None
+    if is_disk and hasattr(dest.allocator, 'copy_from_disk') and disk_supports_fast_copyout and src.nbytes >= 4096:
       dest.allocator.copy_from_disk(dest._buf, src._buf, src.nbytes)
-    elif src.device.startswith("DISK") and hasattr(dest.allocator, 'as_buffer'):
+    elif is_disk and hasattr(dest.allocator, 'as_buffer'):
```
vladov3000, line 132: I factored out `is_disk = src.device.startswith("DISK")`, because otherwise the line for `disk_supports_fast_copyout` would have breached our 150 column limit.

geohot: just use \ i can't read this diff!

minimal PRs only

vladov3000: Collapsed the diff in this file to 1 line of code.

PR #6857, opened 2024-10-02
Add pthread_jit_write_protect_np and sys_icache_invalidate with no-op fallbacks.

vladov3000:
These 2 functions are necessary if we want to mmap JIT code on OSX to replace dlopen for the clang backend.

### DIFF
--- a/tinygrad/runtime/autogen/libc.py
+++ b/tinygrad/runtime/autogen/libc.py
@@ -166,6 +166,21 @@ off_t = ctypes.c_int64
 mode_t = ctypes.c_uint32
 size_t = ctypes.c_uint64
 __off_t = ctypes.c_int64
+
+try:
+    pthread_jit_write_protect_np = _libraries["libc"].pthread_jit_write_protect_np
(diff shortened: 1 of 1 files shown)


tinygrad/runtime/autogen/libc.py
```diff
@@ -166,6 +166,21 @@ class Union(ctypes.Union, AsDictMixin):
...
+    pthread_jit_write_protect_np.argtypes = [ctypes.c_int32]
+except AttributeError:
+    pthread_jit_write_protect_np = lambda _: None
+
+try:
+    sys_icache_invalidate = _libraries["libc"].sys_icache_invalidate
+    sys_icache_invalidate.restype = None
+    sys_icache_invalidate.argtypes = [ctypes.POINTER(None), size_t]
+except AttributeError:
+    sys_icache_invalidate = lambda a, b: None
```
vladov3000, line 182: Can't do `lambda _, _` because arguments cannot have the same name.

PR #6866, opened 2024-10-03
CLANG fixed ops python [run_process_replay]

ignaciosica:
As requested by George


geohot:
Yea you want the tests to be independent of the implementation, like the real hardware

PR #6116, opened 2024-08-17
default threefry

wozeparrot:

PR #6867, opened 2024-10-03
test conv schedules are identical (they aren't!) [pr]

Qazalin:
This was true before https://github.com/tinygrad/tinygrad/pull/6116
Now it fails because of rand's implicit Tensor copy cache:
image


Qazalin:
will leave to woze to decide if we should do something about this.

PR #6868, opened 2024-10-03
Fix gpt2 example

roryclear:
gpt2 example gives different outputs with JIT since [c100f3d](https://github.com/tinygrad/tinygrad/commit/c100f3d40618ff7f19ded78eee89d8a0dc253135).

Outputs on master:
master output

Outputs with change:
output after change

### DIFF
--- a/examples/gpt2.py
+++ b/examples/gpt2.py
@@ -133,7 +133,6 @@ class GPT2:
         weights[k] = weights[k].T
     # lm head and wte are tied
     weights['lm_head.weight'] = weights['wte.weight']
-    model.lm_head.weight.realize()
 
     load_state_dict(model, weights)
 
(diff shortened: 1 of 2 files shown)


geohot:
Hmm, why?

roryclear:
I think because rand doesn't work in JIT after rand has been called and not realized. Not sure why that is yet but I have a repro here:
`from tinygrad import Tensor, TinyJit
x = Tensor.rand(1)
#x.realize()
r = TinyJit(lambda: Tensor.rand(1))
for _ in range(5):
    print(r().numpy())`

roryclear:
and that behaviour with threefry starts when realize was removed from it [7703e7b](https://github.com/tinygrad/tinygrad/commit/97d7082).

geohot:
Hmm, that repro is interesting. Can you add a failing test?

roryclear:
> Hmm, that repro is interesting. Can you add a failing test?

Sure, I just added one

wozeparrot:
Should be fixed with #6869

PR #6892, opened 2024-10-04
dropout contiguous after >= p

chenyuxyz:
the final contiguous in rand is annoying because dropout saves the rand tensor, not the bool tensor for backward. Make contiguous optional and call contiguous in dropout saves 6G mem in bert training. @wozeparrot thought on this direction?


wozeparrot:
Yea, I think having something like this flag should be fine for `rand`, rand is a bit special of a creation method anyways.

PR #6906, opened 2024-10-05
added padding='same' support for conv1d

sofinvalery:


geohot:
No tests, why just conv1d?

PR #6916, opened 2024-10-06
dtype=float in bert log_softmax for loss and accuracy

chenyuxyz:


chenyuxyz:
no magic improvement but fine to include

PR #6917, opened 2024-10-06
no type-tracker getitem refactor

geohotstan:
I mentioned months ago I had an idea for instead of using types and crazy list comprehension to keep track of dimensions, we use objects.
Yeah this is it.
I have no idea how I managed to add more lines... yikes

Not sure if this is more or less readable/comprehensible than the last getitem but maybe it's more maintainable?

UPDATE: scratch the object stuff lmaooooooo
we doing match-case boiiiiii


geohotstan:
The tensor ops in getitem remain the same. 
The only things that changed are how to arrive at the args for the tensor ops.
So there shouldn't be any changes.

Just to be sure I diffed `DEBUG=4` metal code for `TestIndexing.test_index` and `TestIndexing.test_advancedindex` and it was the same.

chenyuxyz:
nice cleanup!

PR #6919, opened 2024-10-06
Ci test jit transfer metal (blitCommandEncoder)

mesozoic-egg:
Follow up on #6793


mesozoic-egg:
It seems like the test never fails on the CI's Mac environment. See [the run](https://github.com/tinygrad/tinygrad/actions/runs/11201869602) where I reverted the  #6793 patch, in commit b0a58add128171dae3611ba88e4cab9bb026666a

but running on a real device does fail (the test_jitted_transfer is set up to run on a real device already)

This might just be one of the instances where mac's github runner differs from a real device. I raised another [issue](https://github.com/actions/runner-images/issues/10663) on their project with regards to `newBufferWithBytesNoCopy:length:options:deallocator:`

kind of stuck, not sure how to catch this in CI

Issue #6911, opened 2024-10-06
Add support for logcumsumexp


geohot:
Matching https://pytorch.org/docs/stable/generated/torch.logcumsumexp.html

KunalSin9h:
working on it @geohot

chenyuxyz:
also needs `cummax` for stability

KunalSin9h:
ok @chenyuxyz

the-praxs:
Working on it in #6915

jeffzh4ng:
@geohot @chenyuxyz can you take a look at https://github.com/tinygrad/tinygrad/pull/6921?

chenyuxyz:
merged. might no be the most stable way to write it but we can revisit it later, it's on par with other numerically questionable methods

PR #6921, opened 2024-10-06
implement logcumsumexp

jeffzh4ng:
#6911 

- using the same numerical stability method as `logsumexp` since the only thing that differs is the cumsum, not the exps: exp(x1) + exp(x2) + ... = exp(m) (exp(x1-m) + exp(x2-m) + ...).
- pytorch requires `dim` by caller. i set the `axis=0` as default which matches some of the behavior i see from other ops

### DIFF
--- a/docs/tensor/ops.md
+++ b/docs/tensor/ops.md
@@ -13,6 +13,7 @@
 ::: tinygrad.Tensor.softmax
 ::: tinygrad.Tensor.log_softmax
 ::: tinygrad.Tensor.logsumexp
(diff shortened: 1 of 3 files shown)


chenyuxyz:
cool yea i was thinking of something else

chenyuxyz:
`exp(x1) + exp(x2) + ... = exp(m) (exp(x1-m) + exp(x2-m) + ...)`, but isn't this `m` different for different N, thus need cummax?

PR #6923, opened 2024-10-06
AM driver

nimlgen:
Current "plan":
- [x] map gpu
- [x] boot gpu (load fw + power management)
- [x] (not planned anymore) interrupts (was exploring this, but seems a minimal kernel module is needed expose as poll on fd)
- [x] ring setup + post init
- [x] vm + explore cpu/gpu shared mappings (large bar?).
- [x] finally send some compute
- [x] anything to support p2p?


geohot:
Do you need interrupts, or can polling work to start? I really want to avoid a kernel module

nimlgen:
> Do you need interrupts, or can polling work to start? I really want to avoid a kernel module

Maybe yeah, do not see any reasons why we cannot avoid irqs.

nimlgen:
Updated timeline:
1) If everything is fine should be able to setup ring by this Monday/Tuesday

nimlgen:
Not sure we need gart, but not sure I can creatine gpu mapping from userland...

We can read/write vram from cpu using bars, this should be enough

geohot:
yea GART can be added later as an optimization

geohot:
yay, no MES!

nimlgen:
`sudo rmmod amdgpu`
run with: `sudo -E DEBUG=2 AM=1 python3 test/test_ops.py TestOps.test_tiny_add`

tinygrad/runtime/support/am/gmc.py
```diff
@@ -0,0 +1,135 @@
...
+    self.dummy_page_pm = self.adev.mm.palloc(0x1000)
+
+  def init(self, root_pt):
+    print("GMC init")
+    self.root_pt = root_pt
+    self.vm_config = 0x1fffe05 # 2 level, 1 gb huge pages
+    self.init_mmhub()
+
+  def init_aperture_regs(self, block:Union["MM", "GC"]):
+    getattr(self.adev, f"reg{block}VM_CONTEXT0_PAGE_TABLE_START_ADDR_LO32").write((self.vm_base >> 12) & 0xffffffff)
```
geohot, line 30: Is `getattr` needed here? I see it's different in the MES IP block

tinygrad/runtime/support/am/amring.py
```diff
@@ -0,0 +1,93 @@
...
+
+    self.rptr[0] = 0x0
+    self.wptr[0] = 0x0
+
+    self.next_ptr = 0
+
+    self.fill_mqd()
+
+  def fill_mqd(self):
+    self.mqd.header = 0xC0310800
```
geohot, line 29: Does the `ops_amd` driver use the same MQD struct? I see similar stuff on lines 109-122

tinygrad/runtime/ops_am.py
```diff
@@ -0,0 +1,271 @@
...
+    self._gpu_addr = x.va_addr + off
+    self._mc_addr = x.mc_addr + off
+    self._signal = to_mv(x.cpu_addr + off, 16).cast("Q")
+    super().__init__(value)
+  def __del__(self): AMDevice.signals_pool.append(self._o)
+  def _get_value(self) -> int: return self._signal[0]
+  def _get_timestamp(self) -> decimal.Decimal: return decimal.Decimal(self._signal[1]) / decimal.Decimal(100)
+  def _set_value(self, new_value:int): self._signal[0] = new_value
+
+class AMComputeQueue(HWComputeQueue):
```
geohot, line 44: This should be the same class as AMDComputeQueue, right?

geohot:
Two takeaways from reading this:

1) It's worth the investment to get the autogen to be as good as possible. Not sure why one of the IP blocks needs getattr, but readability is the most valuable thing.

2) This should share as much code as possible with ops_amd, to the point that it almost isn't a different device. There should be an abstraction layer which abstracts the "runtime" from the "driver" so that the same command queues can be submitted to either.

tinygrad/runtime/support/am/amdev.py
```diff
@@ -0,0 +1,151 @@
...
+  def wdoorbell64(self, index, val): self.doorbell64[index//2] = val
+
+  def ip_base(self, ip, inst, seg):
+    ipid = amdgpu_discovery.__dict__.get(f"{ip}_HWIP")
+    try: x = self.regs_offset[ipid][inst][seg]
+    except (KeyError, IndexError): x = 0
+    return x
+
+  def start_discovery(self):
+    mmIP_DISCOVERY_VERSION = 0x16A00
```
geohot, line 105: Where can we pull these from?

tinygrad/runtime/support/am/amring.py
nimlgen: mqd is a ring's context. compute_* are shadowed registers and looks like they are "shadowed" to mqd.

Executing the following in the ring updates its field in the mqd (without any additional syncs/cmds at all). And it inits these registers from mqd as well.
```python
self.q += [amd_gpu.PACKET3(amd_gpu.PACKET3_SET_SH_REG, 1), gfxreg(amd_gpu.regCOMPUTE_STATIC_THREAD_MGMT_SE0), 0xffffffff]
```

MQD also contains HQD regs. It writes them back on a dequeue request into mqd as well. And in `hqd_load` we can load them directly into registers (without any kiq queues now)

tinygrad/runtime/ops_amd.py
```diff
@@ -284,22 +282,43 @@ class AMDQueueDesc:
   doorbell: memoryview
   put_value: int = 0
 
-class AMDDevice(HCQCompiled):
+class KFDIface:
```
geohot, line 285: Want to merge this as a separate PR? How many lines does it add/subtract?

PR #6925, opened 2024-10-07
return inf in time_linearizer if RuntimeError

chenyuxyz:


chenyuxyz:
doesn't really help, it looks like NV hangs after this crash

Issue #6910, opened 2024-10-06
nn.Embedding output shape is wrong


geohot:
```python
from tinygrad import Tensor, nn

>>> nn.Embedding(10, 16)(Tensor([3]).reshape(()))
<Tensor <LB NV (1, 1, 16) float ShapeTracker(views=(View(shape=(1, 1, 16), strides=(0, 0, 1), offset=0, mask=None, contiguous=True),))> on NV with grad None>
>>> nn.Embedding(10, 16)(Tensor([3]).reshape((1,)))
<Tensor <LB NV (1, 1, 16) float ShapeTracker(views=(View(shape=(1, 1, 16), strides=(0, 0, 1), offset=0, mask=None, contiguous=True),))> on NV with grad None>
>>> nn.Embedding(10, 16)(Tensor([3]).reshape((1,1)))
<Tensor <LB NV (1, 1, 16) float ShapeTracker(views=(View(shape=(1, 1, 16), strides=(0, 0, 1), offset=0, mask=None, contiguous=True),))> on NV with grad None>
>>> nn.Embedding(10, 16)(Tensor([3]).reshape((1,1,1)))
<Tensor <LB NV (1, 1, 10, 16) float ShapeTracker(views=(View(shape=(1, 1, 10, 16), strides=(0, 0, 16, 1), offset=0, mask=None, contiguous=True),))> on NV with grad None>
>>> nn.Embedding(10, 16)(Tensor([3]).reshape((1,1,1,1)))
<Tensor <LB NV (1, 1, 1, 10, 16) float ShapeTracker(views=(View(shape=(1, 1, 1, 10, 16), strides=(0, 0, 0, 16, 1), offset=0, mask=None, contiguous=True),))> on NV with grad None>
```

Compared to
```python
>>> import torch
>>> torch.nn.Embedding(16, 10)(torch.Tensor([3]).long().reshape(())).shape
torch.Size([10])
>>> torch.nn.Embedding(16, 10)(torch.Tensor([3]).long().reshape((1,))).shape
torch.Size([1, 10])
>>> torch.nn.Embedding(16, 10)(torch.Tensor([3]).long().reshape((1,1))).shape
torch.Size([1, 1, 10])
>>> torch.nn.Embedding(16, 10)(torch.Tensor([3]).long().reshape((1,1,1))).shape
torch.Size([1, 1, 1, 10])
[...]

PR #6930, opened 2024-10-07
fix embedding shape

conwayz:
https://github.com/tinygrad/tinygrad/issues/6910
output shape should be input_shape + (embed_dim,) and the sum reduction should be in the vocab dimension


geohot:
Thanks!

PR #6943, opened 2024-10-07
added padding='same' support for conv1d

sofinvalery:


geohot:
Test is very incomplete. You shouldn't need that if. Is the stride restriction also true on torch?

sofinvalery:
yeah

Issue #6909, opened 2024-10-06
If BF16 isn't supported, it should autocast to float


geohot:
Should fix #6905 with CLANG. Add a rewrite rule to do this for the clang backend and confirm llama works

AlexCheema:
Watching this. I think some users might have run into this issue using exo: https://github.com/exo-explore/exo/issues/152

kayo09:
How is this a "good first issue"? :/

FlorianHeigl:
@kayo09 "good first issue" means it's good a good choice for a developer who wants to start contributing to the project, like one where they can learn about the structure, there's no highly complex interdependencies that would need months of constant experience with the code base to fix. that doesnt' mean it'll be easy for someone who can't code and solve problems where they need to understand what's behind what the code does. (not me)

kayo09:
I love open source. Thanks @FlorianHeigl for reinstating that fact

Ra5hidIslam:
I was able to reproduce the __bf16 support error on an Intel Broadwell VM (Xeon Model 79) using Clang. The compiler fails because it doesn't support the __bf16 type on this specific target.

On Apple Silicon and newer x86 hardware, the issue doesn't trigger as native support exists.

Proposed fix in tinygrad/renderer/cstyle.py:
Mapping dtypes.bfloat16 to unsigned short in the type_map and ensuring the renderer handles the cast to float for math operations resolves the backend crash on older targets.

# cstyle.py  line number: 232
```
type_map = {dtypes.bool:"_Bool", dtypes.half:"__fp16", dtypes.bfloat16:"unsigned short"}

```
Does this align with the desired path for soft-promotion in the Clang backend?

PR #6945, opened 2024-10-07
WIP: autocast bf16 to float32 for clang

1ntEgr8:
I hooked the rewriter using the `extra_matcher` field of the renderer (mimicking PTX). The rewrite rules are not correct (does not perform the shift), will fix soon. I was able to compile and run the llama example.

#6909 

PS: just noticed bfloat autocasting for metal after rebase, will refactor my change

### DIFF
--- a/tinygrad/runtime/ops_clang.py
+++ b/tinygrad/runtime/ops_clang.py
@@ -12,7 +12,7 @@ class ClangCompiler(Compiler):
   def compile(self, src:str) -> bytes:
     # TODO: remove file write. sadly clang doesn't like the use of /dev/stdout here
     with tempfile.NamedTemporaryFile(delete=True) as output_file:
-      subprocess.check_output(['clang', '-shared', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib',
+      subprocess.check_output(['clang', '-shared', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib', '-lsystem',
                                '-', '-o', str(output_file.name)], input=src.encode('utf-8'))
       return pathlib.Path(output_file.name).read_bytes()
 
--- a/tinygrad/codegen/uopgraph.py
+++ b/tinygrad/codegen/uopgraph.py
@@ -575,7 +575,7 @@ def full_graph_rewrite(sink:UOp, opts:Optional[Renderer]=None) -> UOp:
       sink = graph_rewrite(sink, sym+reducer)
       sink = graph_rewrite(sink, sym+get_extra_patterns(tuple(opts.code_for_op.keys()) if opts is not None else (), TRANSCENDENTAL>=2))
 
-  if opts is not None and opts.extra_matcher is not None: sink = graph_rewrite(sink, folder+opts.extra_matcher)
+  if opts is not None and opts.extra_matcher is not None: sink = graph_rewrite(sink, opts.extra_matcher)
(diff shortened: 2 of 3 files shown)


geohot:
Tests failing

tinygrad/runtime/ops_clang.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, cachekey="compile_clang", args:Optional[List[str]]=None):
   def compile(self, src:str) -> bytes:
     # TODO: remove file write. sadly clang doesn't like the use of /dev/stdout here
     with tempfile.NamedTemporaryFile(delete=True) as output_file:
-      subprocess.check_output(['clang', '-shared', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib',
+      subprocess.check_output(['clang', '-shared', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib', '-lsystem',
```
geohot, line 15: Why do you need this?

1ntEgr8: I ran this command on a M2 Mac 

```
DEBUG=5 DISABLE_COMPILER_CACHE=1 CLANG=1 python3 examples/llama3.py --download_model --size 8B
```

It fails with:

```
ld: dynamic executables or dylibs must link with libSystem.dylib for architecture arm64
```

The clang command that gets executed:

```
['clang', '-shared', '-march=native', '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib', '-', '-o', '/var/folders/0k/nj410z392tsg9484w4f976m80000gn/T/tmp8be37epu']
```

So, we are building a shared library (`-shared`) with no stdlib (`-nostdlib`). The error message says for arm64 this is not allowed, so I had to add the `-lsystem` flag (could've also gotten away with removing `-nostdlib`).

Some of the tests are failing because `-lsystem` is a MacOS thing, so let me guard the inclusion of the flag.

1ntEgr8:
So, did a little more digging.

In examples/llama3.py, `fix_bf16` is called. If `SUPPORT_BF16` is set to zero, then the `bfloat16` type won't materialize in the AST. However, compilation still fails with this error:
```
<stdin>:12:338: error: cannot take the address of an rvalue of type 'int'
    data0[ridx0] = (1/(alu2?(alu3?__builtin_nanf(""):(alu1?((alu4<__builtin_inff())?((alu4<-150.0f)?0.0f:((alu4<128.0f)?((alu4!=128.0f)?(((((((((((((alu6*0.0001535920892f)+0.001339262701f)*alu6)+0.009618384764f)*alu6)+0.05550347269f)*alu6)+0.2402264476f)*alu6)+0.6931471825f)*alu6)+1.0f)*(*((float*)&((alu5<<23)+1065353216)))*(*((float*)&(((cast0-alu5)<<23)+1065353216)))):__builtin_inff()):__builtin_inff())):__builtin_nanf("")):0.0f)):__builtin_inff()));
```

Looking at the AST with DEBUG=6, it indeed is trying to take the address of an rvalue.

Say the previous error is fixed, there are now two things we need to handle.
1. If the input tensor is bfloat16, then we must first convert that to float32. This eliminates any bfloat16 variables from being introduced from source tensors
2. If a kernel internally uses bfloat16, we must rewrite those operations to use float32.

We cannot only do (2) without (1); it breaks correctness.
Also, we cannot only do (1) without (2); there is no guarantee that a kernel cannot internally use bfloat16.

PR #6946, opened 2024-10-08
FSDP

mesozoic-egg:
FSDP works out of the box (mostly) with the MultiLazyBuffer code, ~~with a minor tweak to make optimizer shard on the correct axis.~~

~~See `beautiful_mnist_multigpu.py` for the example. It shards the model instead of data across devices.~~
See latest comment on llama3 (4B) training run

**Approach**: Model parameters remain sharded for all the elementwise operation (data will be sharded to each device). All_reduce is used for the reduce operation. 

**I think** this is effectively equivalent to Pytorch's implementation where they do all_gather + op + reduce_scatter for all the operation (paper[https://arxiv.org/abs/2304.11277]), would love to get a second opinion, though. Pytorch's implementation gets quite complicated, especially considering they need to register hooks onto the backward pass.

~~Doing more memory profiling and will come up with some examples that train bigger models~~


geohot:
Cool! For the bounty, you have to demonstrate an LLM training that doesn't fit on a single GPU. Do you have tinybox access? They are a bit tied up now for MLPerf, but we can def get you access.

mesozoic-egg:
Access to tinybox would be great!

mesozoic-egg:
Not yet done but I got some prelim results:

Training a GPT2 model (I arbitrarily changed the size so it causes OOM without the shard but can fit FSDP)
Model total size (param + optimizer): 20G
Per device memory: 17G
gpt2

Ran a smaller model to see how much memory is saved:
Model total size: 9G (parameters + optimizer state)
GPU memory without shards: 12G
Per GPU memory when sharded to two: 9G

I thought the memory save up would be greater though. 
With shards:
shards

Without shards:
without_shards

mesozoic-egg:
Screenshot 2024-10-11 at 09 59 23

mesozoic-egg:
The memory savings seem to be around 30%, probably should compare against the pytorch implementation...

mesozoic-egg:
Once in a while there seems to be an error regarding axis not matching depending on model implementation, I think it boils down to this example:

```python
x = Tensor.empty(4, 8, requires_grad=False).shard_(GPUS, axis=1)
w = Tensor.empty(4, 8, requires_grad=True).shard_(GPUS, axis=0)
c = x * w
c.backward(Tensor.empty(4, 8).shard_(GPUS, axis=None))
print(w.grad.lazydata.axis) # --> 1, Tensor.assign will then complain
```

In my earlier attempt I worked around it by just always doing a gather and reshard, but that seem unnecessary, might be a better idea to solve this properly

chenyuxyz:
if you want to do minimal comm, check ring allreduce and i think you can do something similar to convert from axis=0 to axis=1. chunk underlying lazybuffers on the axis=0 tensor by axis=1, reshuffle the buffers, and reassemble to remove the axis=0 shard

mesozoic-egg:
One solution I arrived at is checking the gradient's axis before assigning:
```python
  def step(self):
    for x in self.params:
      if isinstance(x.grad.lazydata, MultiLazyBuffer) and (axis:=x.grad.lazydata.axis) is not None and axis != x.lazydata.axis:
        print("ALL GATHER")
        x.grad.all_gather_()
      x.assign(x.detach() - x.grad)
    Tensor.realize(*self.params)
```

Also removed the unncessary `shard_axis` parameters by calling shard on the optimizer instead of model. And remove the extra buffer allocation in `assign`, by realizing them at the start:
```python
model = Model()
opt = Optimizer(nn.state.get_parameters(model))
  for p in nn.state.get_parameters(opt):
    p.shard_(GPUS, axis=0)
    p.realize()
```

However when benchmarking a (two layer MLP)[https://github.com/tinygrad/tinygrad/blob/7de6e4e91abc7310fda3a5a037b279481a69e001/extra/fsdp/linear.py] I saw that sharding consumes more memory.

```
Total tensor size: 864 bytes (plain SGD)
Peak mem without sharding: 1120 bytes
Peak mem per device with sharding: 1696 bytes
```

Other than the gradient all gather I added, another part that contribute to the extra mem use is the all_gather inside alu (in the else clause). Both are essentially solving the axis not matching problems. But this implementation consumes too much memory.
```python
    for mlb in msrcs:
      if (mlb.axis == axis and (mlb.axis is None or mlb.bounds == bounds)) or not_all_real: srcs.append(mlb.lbs)
      elif mlb.axis is None and axis is not None: srcs.append(to_sharded(mlb.lbs, axis, bounds))
      else: srcs.append(to_sharded([mlb.copy_to_device(lb.device) for lb in mlb.lbs], axis, bounds))
```

I think I should do something like ring all reduce to minimize memory use when assembling the axes, like chenyu suggested

rggs:
Where does the `all_gather` happen for each of the shard units during the forward pass?

mesozoic-egg:
@rggs: pytorch's all_gather refers to two things: in the context of reduce operation (sum), this is handled by all_reduce inside multi.py; in the context of elementwise operation (e.g. multiplication), this is handled by the last else branch inside `alu` of `multi.py`. The distinction between forward and backward is not necessary here, because tensors are only materialized upon gradient update which combines both forward and backward pass.

rggs:
@mesozoic-egg Ah very cool, thanks for the clarification!

mesozoic-egg:
Sharding across four devices for MNIST with AdamW:
Parameter size (with optimizer): 14.4 mB
Device 0 peak mem: 65 mB
Device 1 - 3 peak: 17 mB

Without sharding the peak mem is 63 mB

need to implement a ring all gather

mesozoic-egg:
I have come to realize that naively sharding the models, then calling all gather to align the axis during training, is the same as just replicating the model across all devices with existing implementation. FSDP doesn't work out of the box unfortunately : / will try implementing some sort of redistribute algorithm like (this)[https://github.com/pytorch/pytorch/blob/15722debfb95863d666ca88ff6ac98a57f89c4eb/torch/distributed/tensor/_redistribute.py#L156]

rggs:
Ok, this seems more in line with what I was seeing. E.g. if you have multiple FSDP units in sequence, you have to do an `all_gather` for each one and then release the sibling weights for that unit for each device/shard during the forward pass.
Is the idea of implementing the redistribute algorithm to bake FSDP capability into the tensor itself rather than using a wrapper class that calculates the shards and handles the communications?

mesozoic-egg:
I think the reshard is working well, `extra/fsdp/gpt2.py` example (model size with adam optimizer: 13.3mB) memory usage went from 28mb (no shard) to 17mb per device (2 shards).

Right now another factor that contributes to memory is upon allocation, all the content is first saved on device 0, then distributed. Ideally allocation can occur directly on device.

mesozoic-egg:
The trick I found was to set device to CPU during initialization, then upon the start of training shard them to each GPU device. 

Also found out that sharding strategy may not model specific. For example, for convolution you should set the shard on dimension 1. And for a two layer MLP, you should shard on axis0 for the first layer, and axis 1 on the second. The reason is that doing a reduce on a sharded axis does not improve memory. Suppose you do a reduce on a Tensor of shape (4, 2) on axis 1, the result is (4,1). If the shard occurs on axis 1, then each device would need to hold the same amount of data, on top of the communication overhead. Although for GPT model, this did not occur.

I think I can start trying out a few bigger training.. 🤞

mesozoic-egg:
Training llama 4B (8B model with half the layers)
Screenshot 2024-10-28 at 04 09 31

mesozoic-egg:
Was able to train a 4B llama3 model on five 4090s (llama-3 8B with half the `n_layers`). Device 0 is only used for data initialization and the model was sharded on the other four GPUS.
The model with adamW is 55 GB, each device is handling 20G peak memory usage
Screenshot 2024-10-29 at 04 48 49

chenyuxyz:
it's fine to train for a bit longer on tinybox (less than 24 hours), can you show some metrics that it's indeed training?

also now you are closer to finish, some of the multi api changes could be their own prs and we can get those merged separately

mesozoic-egg:
I trained for 30min on some rented GPUs `CUDA=1 LRU=0 SHARD=4 RING=2 PYTHONPATH='.' python extra/fsdp/llama3.py` and here's the result, the test loss is going down consistently
```
Model 18.17 GB Optimizer: 54.49 GB
Model params: 4.54 B Optimizer params: 13.62 B
Epoch   50 loss:   7.24 test_loss:   7.94 elapsed time:  1.8 min
Epoch  100 loss:   6.22 test_loss:   7.68 elapsed time:  2.7 min
Epoch  150 loss:   7.32 test_loss:   7.71 elapsed time:  3.5 min
Epoch  200 loss:   7.24 test_loss:   7.41 elapsed time:  4.4 min
Epoch  250 loss:   6.53 test_loss:   7.27 elapsed time:  5.2 min
Epoch  300 loss:   5.61 test_loss:   7.33 elapsed time:  6.1 min
Epoch  350 loss:   6.16 test_loss:   7.08 elapsed time:  6.9 min
Epoch  400 loss:   6.09 test_loss:   6.88 elapsed time:  7.7 min
Epoch  450 loss:   4.89 test_loss:   6.95 elapsed time:  8.6 min
Epoch  500 loss:   6.02 test_loss:   6.84 elapsed time:  9.4 min
Epoch  550 loss:   5.53 test_loss:   6.75 elapsed time:  10.2 min
Epoch  600 loss:   5.29 test_loss:   6.73 elapsed time:  11.1 min
Epoch  650 loss:   5.02 test_loss:   6.85 elapsed time:  11.9 min
Epoch  700 loss:   5.57 test_loss:   6.67 elapsed time:  12.8 min
Epoch  750 loss:   6.69 test_loss:   6.61 elapsed time:  13.6 min
Epoch  800 loss:   5.98 test_loss:   6.35 elapsed time:  14.4 min
Epoch  850 loss:   7.10 test_loss:   6.45 elapsed time:  15.3 min
Epoch  900 loss:   5.78 test_loss:   6.23 elapsed time:  16.1 min
Epoch  950 loss:   5.88 test_loss:   5.92 elapsed time:  16.9 min
Epoch 1000 loss:   4.55 test_loss:   6.18 elapsed time:  17.8 min
Epoch 1050 loss:   5.33 test_loss:   5.99 elapsed time:  18.6 min
Epoch 1100 loss:   4.54 test_loss:   6.02 elapsed time:  19.5 min
Epoch 1150 loss:   5.05 test_loss:   6.13 elapsed time:  20.3 min
Epoch 1200 loss:   5.16 test_loss:   6.15 elapsed time:  21.1 min
Epoch 1250 loss:   4.75 test_loss:   6.18 elapsed time:  22.0 min
Epoch 1300 loss:   5.66 test_loss:   6.05 elapsed time:  22.8 min
Epoch 1350 loss:   5.21 test_loss:   6.10 elapsed time:  23.6 min
Epoch 1400 loss:   5.27 test_loss:   5.95 elapsed time:  24.5 min
Epoch 1450 loss:   4.67 test_loss:   6.00 elapsed time:  25.3 min
Epoch 1500 loss:   4.89 test_loss:   5.93 elapsed time:  26.2 min
Epoch 1550 loss:   4.90 test_loss:   6.01 elapsed time:  27.0 min
Epoch 1600 loss:   5.20 test_loss:   6.07 elapsed time:  27.8 min
Epoch 1650 loss:   4.79 test_loss:   6.06 elapsed time:  28.7 min
Epoch 1700 loss:   5.14 test_loss:   5.90 elapsed time:  29.5 min
Epoch 1750 loss:   5.60 test_loss:   5.76 elapsed time:  30.3 min
Epoch 1800 loss:   5.47 test_loss:   5.87 elapsed time:  31.2 min
Epoch 1850 loss:   4.56 test_loss:   5.42 elapsed time:  32.0 min
Epoch 1900 loss:   4.99 test_loss:   5.83 elapsed time:  32.9 min
Epoch 1950 loss:   4.04 test_loss:   5.88 elapsed time:  33.7 min
Epoch 2000 loss:   3.97 test_loss:   5.91 elapsed time:  34.5 min
Saved weights to weights/llama3_fsdp.safetensors
Inferenced tokens 20: 100%|█████████████████████| 20/20 [01:38<00:00,  0.20it/s]
Inference:  , stand and-do'd
to devour do first men, without ' throat, and
cannot to
```

How do I get access to a tinybox to test there?

chenyuxyz:
start by asking for tinybox permission in discord

mesozoic-egg:
After 4 hr of training on tiny18:
```
Epoch  100 loss:   6.42 test_loss:   7.54 elapsed time:  2.4 min
Epoch  200 loss:   7.06 test_loss:   7.23 elapsed time:  3.6 min
...
Epoch 22200 loss:   0.21 test_loss:   8.45 elapsed time:  4.1 hr
Epoch 22300 loss:   0.18 test_loss:   8.03 elapsed time:  4.1 hr
Epoch 22400 loss:   0.18 test_loss:   8.25 elapsed time:  4.2 hr
Epoch 22500 loss:   0.15 test_loss:   8.37 elapsed time:  4.2 hr
Epoch 22600 loss:   0.21 test_loss:   8.28 elapsed time:  4.2 hr
Epoch 22700 loss:   0.22 test_loss:   8.44 elapsed time:  4.2 hr
Epoch 22800 loss:   0.19 test_loss:   8.40 elapsed time:  4.2 hr
```

I think i'm overfitting, but it the training does occur. Might be a good idea to split the PR now

mesozoic-egg:
Now I actually think the pattern matcher is a better approach to remove unnecessary chunks of transfer, instead of coding up the transfer logic directly. I'll work on that at a later time. Feel free to unlock this bounty for now.

PR #6926, opened 2024-10-07
extra/gemm/max_matmul: start of custom kernels for GEMM

flammit:
As a prerequisite to implementing full speed GEMM for NV, here's the handwritten versions of GEMM that shows the incremental progress needed to get there and associated speed improvements.

| Acc | Variation | Performance |
| --- | --- | --- |
| FP32 | hcopt | 1301.50 us, would be 105600.10 GFLOPS matmul, 77.34 GB/s  |
| FP32 | flat_smem_input | 1309.41 us, would be 104962.67 GFLOPS matmul, 76.88 GB/s  |
| FP32 | swizzled_smem_input |  882.69 us, would be 155705.02 GFLOPS matmul, 114.04 GB/s |
| FP32 | 2_stage_swizzled_smem_input |  831.49 us, would be 165292.77 GFLOPS matmul, 121.06 GB/s |
| FP32 | max |  826.37 us, would be 166316.89 GFLOPS matmul, 121.81 GB/s |
| FP16 | 3_stage_swizzled | 505.66 us, would be 271798.97 GFLOPS matmul, 199.07 GB/s | 
| FP16 | max | 404.48 us, would be 339791.71 GFLOPS matmul, 248.87 GB/s | 

The command for hcopt is: `PYTHONPATH=. CUDA=1 GEMM_VARIATION="hcopt" DTYPE_IN=half DTYPE_OUT=half DTYPE_ACC=float CNT=1024 INPUT=RAND python3 ./extra/gemm/max_matmul.py `.

The command for the rest of the FP32 acc is: `PYTHONPATH=. CUDA=1 GEMM_VARIATION="$VARIATION" DTYPE_IN=half DTYPE_OUT=float DTYPE_ACC=float CNT=1024 INPUT=RAND python3 ./extra/gemm/max_matmul.py`.

The command for the rest of the FP16 acc is: `PYTHONPATH=. CUDA=1 GEMM_VARIATION="$VARIATION" DTYPE_IN=half DTYPE_OUT=half DTYPE_ACC=half CNT=1024 INPUT=ONES python3 ./extra/gemm/max_matmul.py`.

PR #6956, opened 2024-10-08
ops_cuda: add optional dynamic smem parameter

flammit:
This is required to enable larger than 48kb shared memory usage on a per-kernel basis.

Initially to be used for a 3-stage FP16/FP16 GEMM example in #6926 (needs 72KB of SMEM which is doable in RTX4090), but eventually can do a deeper integration to allow architecture-specific increases in maximum SMEM allocation (when needed by the kernel opts).

### DIFF
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -31,8 +31,8 @@ def cu_time_execution(cb, enable=False) -> Optional[float]:
   return ret.value * 1e-3
 
 class CUDAProgram:
-  def __init__(self, device:CUDADevice, name:str, lib:bytes):
-    self.device, self.name, self.lib = device, name, lib
+  def __init__(self, device:CUDADevice, name:str, lib:bytes, smem:int=0):
+    self.device, self.name, self.lib, self.smem = device, name, lib, smem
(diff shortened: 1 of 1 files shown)


tinygrad/runtime/ops_cuda.py
```diff
@@ -56,7 +56,8 @@ def __call__(self, *args, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
     else:
       for i in range(len(args)): self.c_args.__setattr__(f'f{i}', args[i])
       for i in range(len(vals)): self.c_args.__setattr__(f'v{i}', vals[i])
-    return cu_time_execution(lambda: check(cuda.cuLaunchKernel(self.prg, *global_size, *local_size, 0, None, None, self.vargs)), enable=wait)
+    if self.smem > 0: check(cuda.cuFuncSetAttribute(self.prg, cuda.CU_FUNC_ATTRIBUTE_MAX_DYNAMIC_SHARED_SIZE_BYTES, self.smem))
```
nimlgen, line 59: looks good, can we move this check to __init__ or we have to reset it before launch?

flammit: Good call - wilco

PR #6962, opened 2024-10-09
Fix save and loading tensors from the same file.

### LINKED ISSUE #6294: Safetensor save, load, re-save does not work

vladov3000:
Fixes #6294 and adds a test to ensure there is no regression later on.

### DIFF
--- a/test/unit/test_disk_tensor.py
+++ b/test/unit/test_disk_tensor.py
@@ -202,6 +202,16 @@ class TestSafetensors(unittest.TestCase):
       assert v.numpy().dtype == tensors[k].dtype
       np.testing.assert_allclose(v.numpy(), tensors[k])
 
+  def test_save_load_same_file(self):
+    path = temp("model.safetensors")
+    state_dict1 = get_state_dict(Tensor([1, 2, 3, 4, 5]))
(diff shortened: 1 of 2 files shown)


geohot:
This logic isn't readable. Why does might_open close it sometimes? Why you checking link count?

vladov3000:
`tinygrad.nn.state.safe_save` wants to overwrite the old weights stored in disk. It does this by unlinking the old weights file and writing to a new disk tensor with the same file name.

However, it reuses the same disk device `disk:/some/path/here`, and therefore uses the same file descriptor that points to a now "unlinked" (i.e. stat.st_nlink == 0) file. So nothing actually gets written to the filesystem at that path but rather to the fd pointing to nothing. This does not apply to shared memory objects (`shm:` devices), because we are not persisting them to the file system.

The solution is to reopen the file when we detect that the backing fd is unlinked. To ensure we do not leak fds, we close the old fd before opening the new one.

Issue #6900, opened 2024-10-05
fancy indexing, incomprehensible error


geohot:
```python
batch_size, seq_length = 64, 5000
sel = Tensor.randint(batch_size, high=X_train.shape[0]-seq_length)
X = X_train[sel:sel+seq_length]
```
Doesn't work

```python
batch_size, seq_length = 64, 5000
sel = Tensor.randint(batch_size, high=X_train.shape[0]-seq_length)
X = X_train[sel.reshape(-1, 1)+Tensor.arange(seq_length).reshape(1, -1)]
```
Does

Should the top example work?

geohotstan:
I don't think the top example works...
that passes in a `slice(Tensor, Tensor, None)`
I think our error is `TypeError: slice indices must be integers or None or have an __index__ method` 


torch:
```
Cell In[6], line 4
      2 batch_size, seq_length = 64, 5000
      3 sel = torch.randint(size=(batch_size,), high=X_train.shape[0]-seq_length)
----> 4 X = X_train[sel:sel+seq_length]

TypeError: only integer tensors of a single element can be converted to an index
```

jax:
```
Cell In[10], line 13
     11 # Select a random start index
     12 sel = random.randint(key, (batch_size,), minval=0, maxval=X_train.shape[0] - seq_length)
---> 13 X_train[sel:sel+seq_length]

File /opt/homebrew/lib/python3.11/site-packages/jax/_src/array.py:336, in ArrayImpl.__getitem__(self, idx)
    334   return lax_numpy._rewriting_take(self, idx)
    335 else:
--> 336   return lax_numpy._rewriting_take(self, idx)
...
...
...
File /opt/homebrew/lib/python3.11/site-packages/jax/_src/core.py:684, in check_bool_conversion(arr, warn_on_empty)
    681     raise ValueError("The truth value of an empty array is ambiguous. Use "
    682                      "`array.size > 0` to check that an array is not empty.")
    683 if arr.size > 1:
--> 684   raise ValueError("The truth value of an array with more than one element is "
    685                     "ambiguous. Use a.any() or a.all()")

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

geohot:
Does it work in torch?

chenyuxyz:
the error says "TypeError: only int slicing is supported" now

PR #6967, opened 2024-10-09
Add slice parameter type checking to disallow Tensor usage for slices 

mnovosad1095:
Added behaviour matching PyTorch when using tensors for slicing 
Pls correct me if we are not supposed to support single element tensors when slicing
#6900


test/imported/test_indexing.py
```diff
@@ -98,6 +98,9 @@ def test_index(self):
     numpy_testing_assert_equal_helper(reference[2, 2, 2], 27)
     numpy_testing_assert_equal_helper(reference[:], consec((3, 3, 3)))
 
+    # indexing with single element integer tensor
```
chenyuxyz, line 101: don't add test directly to this file, it's a copy of torch's test.

add these tests to test_ops close to other fancy getitem tests

tinygrad/tensor.py
```diff
@@ -1023,11 +1023,28 @@ def pad(self, arg:Tuple[Optional[Tuple[sint, sint]], ...], value:float=0.0) -> T
   #   3. Out of bounds Tensor indexing results in 0
   #     - e.g: Tensor([1, 2, 3])[Tensor([4, 3, 2])] -> [0, 0, 3] index 4 and 3 are out of bounds
   def _getitem(self, indices, v: Optional[Tensor] = None) -> Tensor:
+    def _fix_slice(x: slice) -> slice:
+      # convert slice scalar Tensor slice parameters to int
+      def process_component(component):
+        if isinstance(component, Tensor):
+          if component.numel() > 1 or not dtypes.is_int(component.dtype): 
+            raise TypeError("only integer tensors of single element can be converted to a slice index")
+          return component.item()
```
chenyuxyz, line 1032: it should not have a `.item` here, as it becomes data dependent

mnovosad1095: Got it
But then the question is, should tinygrad support single-element tensors in slices, which will involve some more clever refactoring, or not and just throw an type err that this operation is not supported?

test/imported/test_indexing.py
mnovosad1095: Got it, will fix

chenyuxyz:
too many lines. please read the existing code and put the check in an appropriate place, not just adding plenty of code just for your specific usage. it should not check specifically for tinygrad.Tensor and should allow only the supported slice.

mnovosad1095:
Thanks for your feedback! 
I've moved the check under the slice 0-step check and made it more concise and general, as you asked. 
Now the error message would look something like this:

`TypeError: Unsupported slice for dimension 0. Expected integers or None, got slice(float, int, NoneType).`

chenyuxyz:
i think this is fine as we clearly don't support it now, would like to know how other framework handles it and we should look into matching those

PR #6970, opened 2024-10-09
bring viz to core

Qazalin:
if you pip install tinygrad you can use viz.


tinygrad/ops.py
```diff
@@ -662,6 +662,7 @@ def rewrite(self, n:UOp) -> UOp:
     return ret
 
 def graph_rewrite(sink:UOp, pm:PatternMatcher, ctx=None) -> UOp:
+  if TRACK_MATCH_STATS and not isinstance(pm, TrackedPatternMatcher): pm = TrackedPatternMatcher(pm.patterns)
```
geohot, line 665: Why is this needed?

Issue #6901, opened 2024-10-05
nn.conv should support padding='same'


geohot:
Needs extensive tests to confirm behavior matches torch

KunalKumar-1:
hey @geohot  What types of extensive tests would you recommend running to validate that the new padding='same' functionality behaves identically to existing PyTorch convolution layers? Should we test different kernel sizes, stride values, input sizes, and padding modes?

wilhelmagren:
Hi, I've been working on this but don't have too much time rn to complete it. If someone else is close to done, could you @ me in this thread to let me know its already been resolved? 

Thanks,

PR #6975, opened 2024-10-10
add support for padding='same' in nn.conv

bhavyagada:
#6901 
Implemented by referencing the [PyTorch conv source](https://pytorch.org/docs/stable/_modules/torch/nn/modules/conv.html)


geohot:
Tons of lines! How can you express concisely?

bhavyagada:
@geohot I reduced the line count by half. 

Is this fine, or do you want me to simplify the loop using list comprehension, which may reduce readability but would bring the line count down to 5?

geohot:
Looking much better!

bhavyagada:
Do I need to do anything else?

geohot:
Nope, merged!

Issue #6935, opened 2024-10-07
fold_unrolled_divs doesn't work on small expressions


S-Lykles:
`fold_unrolled_divs`  in `ops.py` is meant to rewrite expression like:
x//4+(x+1)//4+(x+2)//4+(x+3)//4 -> x

When the max of x is smaller than one of the denominators, some of the individual terms are already folded. 
For example `x[0-3]//4` becomes `x` and then this pattern is not detected anymore.

I wrote a quick test, currently it will fail. 
```python
  def test_arange_unrolled4_small(self):
    gidx = Variable("gidx", 0, 3)
    unrolled_div = (gidx)//4+(gidx+2)//4+(gidx+3)//4+(gidx+1)//4
    self.helper_test_variable(unrolled_div, 0, 3, "gidx")
```

chenyuxyz:
addressed in https://github.com/tinygrad/tinygrad/pull/6977
did not change if there's only term left (e.g., only `(x+3)//4`), as i think it might belong to `div_folding` instead of `fold_unrolled_divs`

S-Lykles:
(x[0-1]+3)//4 fails for the same reason as (3*x[0-3]+3)//4
I found this issue because I was working on an algorithm that would fold things like 
(3*x[0-1]+3) // 4 and (3*x[0-3]+3) % 4
They are foldable because
(3*x[0-3]+3) % 4 = ((3-4)*x[0-3]+3) % 4
and we already know that:
(-x[0-3] + 3) % 4 = (-x[0-3] + 3)
because the min and the max are between 0 and 4

Basically if some mod expression is congruent to an expression who's min and max are between 0 and the mod then it can be folded.

PR #6977, opened 2024-10-10
improve fold_unrolled_divs

chenyuxyz:
addressed #6935
the first few terms in fold_unrolled_divs might have been folded already, so the check should first try to add those terms back. there is a case that every but one term is folded which is not an add chain anymore, so just added as a failed test case for now


chenyuxyz:
nice win
image

PR #6998, opened 2024-10-11
early UOp const + maskless view

Qazalin:
I don't think we can fold this VALID though:
```
UOp(UOps.SINK, dtypes.void, arg=KernelInfo(local_dims=0, upcasted=0, dont_use_locals=False), src=(
  UOp(UOps.STORE, dtypes.void, arg=None, src=(
    UOp(UOps.DEFINE_GLOBAL, PtrDType(dtypes.float), arg=0, src=()),
    UOp(UOps.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(1,), strides=(0,), offset=0, mask=None, contiguous=True),)), src=()),
    UOp(UOps.REDUCE_AXIS, dtypes.float, arg=(BinaryOps.ADD, (0,)), src=(
      UOp(UOps.ALU, dtypes.float, arg=TernaryOps.WHERE, src=(
        UOp(UOps.VALID, dtypes.bool, arg=None, src=(
          UOp(UOps.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(Variable(UOps.DEFINE_VAR, dtypes.int, arg=('i', 1, 10), src=()),), strides=(0,), offset=0, mask=None, contiguous=False),)), src=()),)),
        UOp(UOps.CONST, dtypes.float, arg=1.0, src=()),
        UOp(UOps.CONST, dtypes.float, arg=0.0, src=()),)),)),)),))
```

### DIFF
--- a/test/test_multitensor.py
+++ b/test/test_multitensor.py
@@ -619,13 +619,13 @@ class TestMultiTensor(unittest.TestCase):
         assert ast.op is UOps.STORE
         assert ast.src[2].arg is BinaryOps.ADD
         assert ast.src[2].src[0].op is UOps.LOAD
-        assert ast.src[2].src[1].src[1].op is UOps.CONST and ast.src[2].src[1].src[1].arg == 1
+        assert ast.src[2].src[1].op is UOps.CONST and ast.src[2].src[1].arg == 1
       t = 2 * t
       for si in t.schedule():
         ast = si.ast.src[0]
         assert ast.op is UOps.STORE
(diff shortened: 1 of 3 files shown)


geohot:
Without a mask? Isn't that just True?

Qazalin:
I think the challenge is that REDUCE_AXIS needs an src with shape.

Qazalin:
OK maybe we could do this early
image

PR #7001, opened 2024-10-11
refactor _make_hip_code_for_op into pm rules

ignaciosica:
process_replay has diff as casting is no longer fixed and is now a uop


geohot:
Does process replay match?

ignaciosica:
It does not... but because casting is done at uop level instead of render level.

ignaciosica:
image

This is an example of how it changes

geohot:
Cool, and yea I see what you mean about the precision loss. For bfloat16 I'm fine with removing the middle cast.

PR #7002, opened 2024-10-11
add types in all nn/init.py classes

bhavyagada:
- Adds type annotations to all classes and functions in the nn/init.py file.
- Additionally, type annotations have been included for variables that do not require new lines.


bhavyagada:
Do all variables also need types? e.g.

self.eps: float = eps
self.track_running_stats: bool = track_running_stats
self.momentum: float = momentum

versus

self.eps, self.track_running_stats, self.momentum = eps, track_running_stats, momentum

tinygrad/nn/__init__.py
```diff
@@ -142,13 +143,13 @@ class ConvTranspose2d(Conv2d):
   print(t.numpy())
   ```
   """
-  def __init__(self, in_channels, out_channels, kernel_size, stride=1, padding=0, output_padding=0, dilation=1, groups=1, bias=True):
+  def __init__(self, in_channels:int, out_channels:int, kernel_size:Union[int, Tuple[int, ...]], stride:int=1, padding:int=0, output_padding:int=0, dilation:int=1, groups:int=1, bias:bool=True) -> None: #noqa:E501
```
geohot, line 146: Can't add these

tinygrad/nn/__init__.py
```diff
@@ -142,13 +143,13 @@ class ConvTranspose2d(Conv2d):
...
     super().__init__(in_channels, out_channels, kernel_size, stride, padding, dilation, groups, bias)
-    scale = 1 / math.sqrt(in_channels * prod(self.kernel_size))
-    self.weight = Tensor.uniform(in_channels, out_channels//groups, *self.kernel_size, low=-scale, high=scale)
-    self.output_padding = output_padding
+    scale: float = 1 / math.sqrt(in_channels * prod(self.kernel_size))
+    self.weight: Tensor = Tensor.uniform(in_channels, out_channels//groups, *self.kernel_size, low=-scale, high=scale)
```
geohot, line 149: Is this needed?

bhavyagada: I think the compiler can figure it out, but I added them for consistency since some other classes already had variable types. I'll remove them if they're not needed.

bhavyagada:
I removed all the obvious types. Can you review again?

geohot:
Cool, all look good!

PR #7005, opened 2024-10-11
Add type annotations to classes in `nn/__init__.py`

mszep:
With this change, `mypy --strict tinygrad/nn/__init__.py` only complains about return values of functions in other modules.


geohot:
Closing for #7002

PR #7006, opened 2024-10-11
Implement Graph Rewrite for "Gated Store Rewrite to UOps.IF" Bounty

nicktalati:
This is for the bounty "gated store rewrite to UOps.IF." As far as I can tell, adding a graph rewrite to the end of full_graph_rewrite seemed to be the simplest way to accomplish this. I did see other people who attempted to solve this and they seemed to take a more complex approach that involved modifying the linearizer; I didn't think that was necessary, however this is my first bounty attempt (and actually my first open source contribution), so it's possible I'm missing something.

It may also be necessary to modify the renderers to not look for stores with boolean gates -- if all else looks to be on the right track, I'll do that asap.


nicktalati:
Quick update -- I can see that the llvm renderer is not handling UOps.IFs, and that I need to adjust the PTX renderer to properly handle the new stores (with IF gates). Will investigate further this weekend.

nicktalati:
All checks are now passing. 

**Key Changes:**

- Updated spec to use `to_uops_list` instead of `linearize_uop` (supports multiple `UOp`s)
- Added graph rewrite pass to `full_graph_rewrite` to handle gated stores and replace gate with `UOps.IF`
- Fixed bug in PTX renderer where correct jump condition was inverted
- Added support in LLVM renderer for `UOps.IF` and `UOps.ENDIF`

geohot:
Can you add a rule to type verify to confirm there's no more gated stores, and remove support from the renderers for them?

nicktalati:
Yes I can square that away tomorrow if that works!

nicktalati:
Removed the rule in `type_verify` that accepts stores with a boolean gate and removed support for these from the renderers. Note that we still accept gated stores as long as the gate is an IF (this is necessary for the linearizer to construct the IF..ENDIF blocks). I could modify each renderer to strip the IF source from each store if desired (with another rewrite), though I assume this is unnecessary especially with the `type_verify` check.

tinygrad/renderer/assembly.py
```diff
@@ -153,7 +154,8 @@ def _cast(a, dtype:DType, atype:DType, bitcast=False, u=None, pred=False):
     for u in uops:
       uop,dtype,src,args = u.op,u.dtype,u.src,u.arg
       if uop is UOps.IF:
-        kk(*self.render_bra(f"IF_{r[src[0]][1:]}_{uops.index(u)}", _cast(r[src[0]], dtypes.bool, src[0].dtype, u=u, pred=True)))
+        pred_reg = _cast(r[src[0]], dtypes.bool, src[0].dtype, u=u, pred=True)
```
Qazalin, line 157: can invert be a rule in extra_mather?
Also I think that pred=True doesn't do anything.

Qazalin:
more left to delete in ptx.
why does llvm look so complex?

Qazalin:
also IF in src is fine.

tinygrad/renderer/assembly.py
nicktalati: Thanks for the feedback! Quick update -- I'm not yet sure why, but when I try to add a rule to the ptx_matcher to invert the IF predicate with:

```
  # invert predicate for UOps.IF
  (UPat((UOps.IF,), name="x", allow_any_len=True, src=(UPat(dtype=dtypes.bool),)),
   lambda x: UOp(x.op, dtype=x.dtype, src=(x.src[0].logical_not(),) + x.src[1:], arg=x.arg)),
```

the ptx unit tests hang. I'll investigate this further tonight when i can use a gpu. As for the pred=True, I've removed that.

And in terms of the llvm renderer, it previously wasn't handling IF/ENDIFs at all (they are only ever created for the new gated stores, i believe), and I think the stack is necessary to handle nested IF/ENDIFs -- if you do think it's too complex, I can work on that more today and try to simplify it further.

nicktalati:
Okay, everything should be cleaned up now. 

As far as I can tell, all unnecessary gate logic has been removed from the PTX Renderer. I should note that I was unable to move the gate invert into the pattern matcher, as after inverting the gate condition, the IF still matches the pattern and so we get an infinite recursion. 

I couldn't think of a clean way to robustly set a flag on the UOp so that it wouldn't be processed multiple times, but if that's desired and anyone has ideas I'm all ears!

Lastly, if there is more to do, would there be any way to get this bounty locked?

tinygrad/renderer/llvmir.py
```diff
@@ -90,13 +91,23 @@ def render(self, name:str, uops:List[UOp]) -> str:
 
     for u in uops:
       uop,dtype,src,args = u.op,u.dtype,u.src,u.arg
-      if uop is UOps.STORE:
+      if uop is UOps.IF:
+        condition_value = lvars[src[0]]
+        if condition_value.type != ir.IntType(1):
```
Qazalin, line 96: how could this be true?

Qazalin:
Is this still being actively worked on?

You don't need that extra is int check for LLVM. If you wanna get this merged feel free to split into 3 PRs.
One for the LLVM if/else
The other for PTX inversion (without graph_rewrite is fine)
Then lastly putting that gated store to IF rewrite in uopgraph.py

We merge small diffs a lot faster than something that adds 3 features at once.
You can consider the bounty locked as long as there's active work being done.

nicktalati:
Hi, sorry for the delay -- been busy with work.

Yes, I'll get that int check removed and split this into three PRs -- will definitely be done by Saturday. 

Thanks!

chenyuxyz:
still active?

nick-talati:
Yes -- Since the UOps (and specifically the STORE) have been restructured I have some changes to make; should be done within a week.

nicktalati:
Yes -- Since the UOps (and specifically the STORE) have been restructured I have some changes to make; should be done within a week.

nicktalati:
Sorry, been very busy with work -- closing for now.

PR #7008, opened 2024-10-11
Fix exponential complexity in _is_padding_okay [pr]

louisnovy:
Highly connected deep DAGs can make compile times run into a brick wall without this.

I noticed the problem when evaluating the gradient of implicit functions where it's common to share lots of terms.
A simple example:

```py
from typing import Callable
from tinygrad import Tensor


def length(v: Tensor):
    return (v*v).sum(axis=1).sqrt()

class Implicit:
    def __init__(self, f: Callable[[Tensor], Tensor]): self.f = f
    def __call__(self, x: Tensor): return self.f(x)

    @classmethod
    def sphere(cls, radius: float): return cls(lambda p: length(p) - radius)
    @classmethod
    def gyroid(cls, scale: float):
        def f(p: Tensor):
            x, y, z = (p / scale).T
            return scale * (Tensor.sin(x)*Tensor.cos(y) + Tensor.sin(y)*Tensor.cos(z) + Tensor.sin(z)*Tensor.cos(x))
        return cls(f)

    def translate(self, t: Tensor): return Implicit(lambda p: self.f(p - t))
    def scale(self, s: float | Tensor): return Implicit(lambda p: self.f(p / s) / s)
    def offset(self, o: float | Tensor): return Implicit(lambda p: self.f(p) - o)
    def shell(self, thickness: float): return Implicit(lambda p: Tensor.abs(self.f(p)) - thickness)

    @property
    def complement(self): return Implicit(lambda p: -self.f(p))
    def union(self, other: Implicit): return Implicit(lambda p: Tensor.minimum(self.f(p), other.f(p)))
    def intersection(self, other: Implicit): return Implicit(lambda p: Tensor.maximum(self.f(p), other.f(p)))
    def difference(self, other: Implicit): return Implicit(lambda p: Tensor.maximum(self.f(p), -other.f(p)))
    def symmetric_difference(self, other: Implicit): return self.union(other).difference(self.intersection(other))

    def __invert__(self): return self.complement
    def __and__(self, other): return self.intersection(other)
    def __or__(self, other): return self.union(other)
    def __xor__(self, other): return self.symmetric_difference(other)
    
    def smooth_union(self, other: Implicit, k: float):
        def smooth_min(a: Tensor, b: Tensor, k: float):
            k = k * 1.0 / (1.0 - 0.5**0.5)
            return Tensor.maximum(Tensor.minimum(a, b), k) - length(Tensor.maximum(k - Tensor.stack(a, b, dim=1), 0.0))
        return Implicit(lambda p: smooth_min(self.f(p), other.f(p), k))


f = Implicit.sphere(1.0)

# This is the problem
# f &= Implicit.gyroid(0.1)

for i in range(3): f = f.smooth_union(Implicit.sphere(1.0).translate(Tensor([0.0, 0.0, float(i)])), 0.1)

q = Tensor([[0.0, 0.0, 0.0]], requires_grad=True)
d = f(q)
d.backward(Tensor.ones_like(d))
q.grad.realize()

```
Reproduce by uncommenting the gyroid line (which encourages creation of three slightly different subgraphs after rewrites).
The rest is simply there to create a non-trivial function.

What's the best way to regression test this kind of thing?


Qazalin:
good catch! linters failing.

can you create a minimal recursive graph that reproduces this behavior?
some examples in the codebase: TestFusionOp.test_recursive_add, TestIndexing.test_recursive_st_fixup,

louisnovy:
Alright, here's the pathological test case (N=10):  
Screenshot 2024-10-13 at 13 39 21

I've generated this with the high level api for brevity. Is it better to create it directly?

PR #7009, opened 2024-10-12
derivative of softmax is indepedent of max

chenyuxyz:
two less kernels per softmax backward. testing on bert seems to be 3% faster

209 -> 213 tflops
<https://wandb.ai/chenyuxyz/MLPerf-BERT/runs/ed4m45ph/overview>


geohot:
How can things like this be autodiscovered?

chenyuxyz:
don't know, maybe when we do math on lazybuffer level it would end up being a no-op somewhere and be cancelled?

PR #7017, opened 2024-10-12
generic process replay decorator [pr]

Qazalin:
given args and Context state, `@process_replay` can diff the return value of any function.


Qazalin:
stale, big graph is higher priority right now.

Issue #6918, opened 2024-10-06
TRACEMETA=0 isn't respected if set in a Context


geohot:
I'd expect no metadata to be saved in a `with Context(TRACEMETA=0):`

Also, if TRACEMETA=0 is set on start, the wrappers will never be added.

saurabhaloneai:
working on it @geohot

wozeparrot:
hmm, `TRACEMETA=0` not applying the wrapper was intentional, because you could set it to prevent the slight wrapper overhead if needed.

TheTacoMerchant:
It seems I don't understand the problem. Would I be correct in saying that this issue can be closed when:

- [ ] No metadata is saved in a with Context(TRACEMETA=0) block
- [ ] Even if TRACEMETA=0 is set on start, we should save metadata in a Context(TRACEMETA=2) block

PR #7021, opened 2024-10-12
respect TRACEMETA=0

TheTacoMerchant:
### What this is
Makes it so that TRACEMETA=0 context var is respected in a Context() environment.

### Why it's good
fixes https://github.com/tinygrad/tinygrad/issues/6918

### Testing
Added unit test


geohot:
This isn't right. Call it again with a different tracemeta

geohot:
I don't think the problem is understood here

PR #7024, opened 2024-10-12
fix all test warnings

bhavyagada:
Attempting to fix all test warnings with `PYTHONPATH="." pytest -n=auto test`

This fixes most warnings, but not all. The remaining warnings can be ignored using the -W flag:
`PYTHONPATH="." pytest -n=auto test -Wi`

TODO: Look into the strange behavior from running `test_gc.py` as explained in the comments!


test/test_nn.py
```diff
@@ -211,7 +211,7 @@ def _run_conv2d_same_padding_test(self, BS, C1, C2, H, W, K, S, padding='same',
 
   def test_conv2d_same_padding_odd_input(self):
     BS, C1, H, W = 16, 16, 29, 31
-    C2, K, S, P = 32, 4, 1, 'same'
+    C2, K, S, P = 32, 5, 1, 'same'
```
geohot, line 214: Changing the test is not a good way to fix things

bhavyagada: The warning was in the test itself. PyTorch throws a warning if same padding is used with even kernel size and odd dilation.

bhavyagada:
@geohot Most warnings are due to floating point overflows, invalid values, or divisions in the dtype_alu and dtype tests.

https://github.com/numpy/numpy/issues/23606#issuecomment-1512752172
This seems to be the only solution to get rid of them. 

Before I make any changes, is it okay to suppress warnings like this?

bhavyagada:
The behavior of `test_gc.py` is strange.

- When I run the test suite using PYTHONPATH="." pytest -n=auto test, I get the following warnings, but no errors:

```
test/test_gc.py::TestGC::test_gc
test/test_gc.py::TestGC::test_gc_complex
  /Users/bhavya/Desktop/projects/tinygrad/test/test_gc.py:10: FutureWarning: `torch.distributed.reduce_op` is deprecated, please use `torch.distributed.ReduceOp` instead
    return sum([isinstance(x, Tensor) for x in gc.get_objects()])

test/test_gc.py::TestGC::test_schedule_gc
test/test_gc.py::TestGC::test_schedule_gc_with_inputs
  /Users/bhavya/Desktop/projects/tinygrad/test/test_gc.py:13: FutureWarning: `torch.distributed.reduce_op` is deprecated, please use `torch.distributed.ReduceOp` instead
    return sum([isinstance(x, Buffer) for x in gc.get_objects()])
```

- When I run the test suite sequentially using PYTHONPATH="." pytest test, I get the following errors along with the above warnings:

```
FAILED test/test_gc.py::TestGC::test_gc - ReferenceError: weakly-referenced object no longer exists
FAILED test/test_gc.py::TestGC::test_gc_complex - ReferenceError: weakly-referenced object no longer exists
FAILED test/test_gc.py::TestGC::test_schedule_gc - ReferenceError: weakly-referenced object no longer exists
FAILED test/test_gc.py::TestGC::test_schedule_gc_with_inputs - ReferenceError: weakly-referenced object no longer exists
```

- And, when I run it as a standalone test using PYTHONPATH="." pytest test/test_gc.py, there are no warnings or errors!

The warnings are strange because `test_gc.py` doesn't even use PyTorch. I'm not sure, but it seems that PyTorch somehow uses `torch.distributed` internally because there is no instance of `torch.distributed.reduce_op` in the codebase.

I don't understand the errors so I need to investigate further into what's going on, but please let me know if you could help me with anything to speed up this process.

bhavyagada:
Lastly, `test_ops.py` throws user warnings because some tests for functions std_mean, std, and var have `degrees of freedom <= 0`. From what I understand, these tests are designed to test such behavior (Correct me if I'm wrong!).

Since we can't change the tests, we can ignore the remaining warnings using the -W ignore flag:
`PYTHONPATH="." pytest -n=auto test -Wi`.

This will ignore all warnings, but as only a few from `test_ops` and `test_gc` remain, so it should be fine for now.

chenyuxyz:
how does this save test running time?

bhavyagada:
I'm not sure. I just reported my observation. I'm assuming it eliminates the time it takes to capture the warnings and produce the warning outputs. There were over 200 warnings, so it kind of makes sense to me.

geohot:
Can you just put that decorator in one place? Don't copy paste comment like that

bhavyagada:
I have no idea why the tests suddenly fail.

test/test_dtype_alu.py
```diff
@@ -93,6 +93,8 @@ def universal_test_midcast(a, b, c, op1, op2, d1:DType, d2:DType):
   np.testing.assert_allclose(tensor_value, numpy_value, rtol=1e-6 if getenv("PTX") else 1e-7)
 
 class TestDTypeALU(unittest.TestCase):
+  def setUp(self): np.errstate(all='ignore').__enter__() # context manager for floating-point error handling
```
geohot, line 96: Are you sure this is cleaned up? Can you just put the decorator on the class?

bhavyagada: Wow, I didn't know classes could be wrapped! I was so focused on wrapping functions that I didn't consider that possibility. I will do better.

geohot:
Merged! I still see a bunch of errors in CI though, they seem torch related

bhavyagada:
Yes, in `test_ops`. Couldn't find any solution for them. They can be ignored locally using the `-Wi` flag.

ignaciosica:
Don't know if it's because of how I'm running, but the `@np.errstate(all='ignore')` decorator makes pytest skip the tests entirely

with:
image

without:
image

(ignore failing tests)

ignaciosica:
CI AMD on this pr
image

CI AMD on older pr #7142
image

bhavyagada:
It wasn't intended, just poor foresight on my part. I opened a new PR to fix this.

chenyuxyz:
so it's "faster" because it skips tests! if you don't root cause it ends up causing more issues

PR #7025, opened 2024-10-13
fix BatchNorm error on input dtype that is not float32

hbhungg:
Using `BatchNorm` in training mode, if input is of `dtypes.double` will cause `AssertionError: assign dtype mismatch dtypes.float != dtypes.double`. This is due to the `self.running_mean` and `self.running_var` was create with default `dtypes.float`.


geohot:
Just cast it. This is imcomplete and adds complexity

hbhungg:
Done, I cast `self.running_mean` and `self.running_var` to the input dtype, removing the needs to have dtype params.

PR #7026, opened 2024-10-13
int64 indexing

ttomsa:


geohot:
Needs tests

ttomsa:
This is what it looks like without pyint. If UOps.CAST isn't in _min_max it needs to be its own pm so I chose to include it without breaking the tests. Maybe it's a bit hacky?

PR #7027, opened 2024-10-13
add a diagram of tinygrad stack to the home page(docs site)

hurrrsh:
- Adds a redesigned diagram screenshot from the tinygrad_intro.pdf to the docs landing page under a 'tinygrad stack' header to help new users/contributors get a brief idea about tinygrad.


geohot:
Merged, but diagram could def be improved and explained

PR #7028, opened 2024-10-13
Replace CLANG dlopen with mmap + remove linker step

detolly:
not really working yet but doesn't segfault at least
 
Bounty: Replace CLANG dlopen with mmap + remove linker step

### DIFF
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -99,6 +99,7 @@ class CStyleLanguage(Renderer):
 
   string_rewrite = base_rewrite
   extra_matcher = extra_pm
+  tinygrad_section = False
 
   def get_kernel_modifier(self, uops:List[UOp]) -> str: return ""
   def render_kernel(self, function_name:str, kernel:List[str], bufs:List[Tuple[str,Tuple[DType,bool]]], uops:List[UOp], prefix=None) -> str:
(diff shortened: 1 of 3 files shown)


detolly:
I have another branch at https://github.com/detolly/tinygrad/tree/clang_ops_no_linker_batch_call where everything is compiled in the batch instead of calling the mmap'd executable memory in the batch. performance seems to be quite similar.

There seem to be some problems with clang and segments in mach-o/whatever on mac os. I don't have a mac at the moment so I'm not sure how to debug that. 

@geohot any comments?

detolly:
Note: Compiling directly without ld (-c) is difficult because of math functions such as sqrt. Previously they were being linked using the dynamic linker. Writing a quick static linker for rodata would probably not be that hard, but writing a dynamic linker for math functions is honestly getting a bit ridiculous, and would probably not be better than using dlopen in the first place. What I have implemented is a middle-way approach that links libm statically.

This approach comes with its downsides also, though. Without the rewriting of the log2 op, I'm pretty sure this doesn't work at all, because it's trying to link cpu-test symbols from libc. There may be other issues.

Static linking of the system library on macOS is apparently not a thing (let alone the mach-o format and the apple linker), and after some experimentation in a x86 VM of macOS 13, I think I can safely say that this is not an approach that is suitable for mac. Therefore I have left the old dlopen approach for mac.

Regardless, I'm halting development on this until further review.

Issue #6905, opened 2024-10-05
clang backend fails using version 18.1.3


lorenzoliusso:
using CLANG backend
Model already downloaded
using LLaMA-7B model
  0%|                                                                                                                                                                               | 0/292 [00:00<?, ?it/s]
fatal error: error in backend: Do not know how to soft promote this operator's operand!
clang: error: clang frontend command failed with exit code 70 (use -v to see invocation)
Ubuntu clang version 18.1.3 (1ubuntu1)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
clang: note: diagnostic msg: Error generating preprocessed source(s) - ignoring input from stdin.
clang: note: diagnostic msg: Error generating preprocessed source(s) - no preprocessable inputs.
loaded weights in  79.66 ms, 0.03 GB loaded at 0.42 GB/s

...
File "/home/lorenzo/projects/tinygrad/tinygrad/runtime/ops_clang.py", line 15, in compile
    subprocess.check_output(['clang', '-shared', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib',
  File "/usr/lib/python3.12/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
File "/usr/lib/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['clang', '-shared', '-march=native', '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib', '-', '-o', '/tmp/tmp6v5p3a9d']' returned non-zero exit status 1.

(noob)
I was just runnig the llama examples.
I don't understand what clang is complainig about.

geohot:
Can you run with DEBUG=4 and paste the broken source?

geohot:
Oh, it fails for me too, but with a different error.

```
tiny@tiny19:~/tinygrad$ CLANG=1 python3 examples/llama3.py --model /raid/weights/LLaMA-3/8B/
seed = 1728209391
  0%|                                                                                                                                                                                                                                                  | 0/292 [00:00<?, ?it/s]
<stdin>:2:48: error: __bf16 is not supported on this target
void E_4194304_4(__fp16* restrict data0, const __bf16* restrict data1) {
                                               ^
[...]

PR #7032, opened 2024-10-13
autocast bfloat16 to float32 for llama on clang

lorenzoliusso:
#6909 and #6905 
The predefined function fix_bf16 was not working correctly.
I was able to load the weights and run examples/llama3.py.

### DIFF
--- a/extra/models/llama.py
+++ b/extra/models/llama.py
@@ -206,8 +206,11 @@ def convert_from_huggingface(weights:Dict[str, Tensor], model: Transformer, n_he
   return sd
 
 def fix_bf16(weights:Dict[Any, Tensor]):
-  if getenv("SUPPORT_BF16", 1):
+  if Device.DEFAULT == "CLANG":
     # TODO: without casting to float16, 70B llama OOM on tinybox.
-    return {k:v.cast(dtypes.float16) if v.dtype == dtypes.bfloat16 else v for k,v in weights.items()}
+    return {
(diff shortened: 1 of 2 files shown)


AlexCheema:
Watching this. I think some users have ran into this issue on exo e.g. https://github.com/exo-explore/exo/issues/152

blindcrone:
It's possible this is related to https://github.com/llvm/llvm-project/issues/63017

Which has a fix in upstream llvm-project if I'm reading this right. I'm currently checking if upgrading to LLVM 19 automagically fixes this (This version is current tip of tree on github, but not packaged in major distros because they tend to be slow to update high-dependency-surface tools like compilers for good reason. However, the AUR has a -git package). If it works, I'll report back here

extra/models/llama.py
tunguz, line 1: Great fix, someone please review this and accept this PR ASAP.

lorenzoliusso: Not sure if this is what they wanted, probably they want to keep using the SUPPORT_BF16 environment variable

lorenzoliusso:
It's not an LLVM issue; fix_bf16() was not casting correctly. However, I believe this is just a workaround, and a more robust solution should be proposed.

geohot:
llvm_bf16_cast should be moved to rewrite rules in the renderer

lorenzoliusso:
Do we want to keep `fix_bf16`?

geohot:
Should be fixed in master

PR #7033, opened 2024-10-13
remove arbitrary multiplication case

chenyuxyz:
adds the wrongly simplified kernel in test_linearizer_failures #7019


chenyuxyz:
test training with this in <https://wandb.ai/chenyuxyz/MLPerf-BERT/runs/ygl1c94c/overview>

geohot:
Hmm, why is it wrong? The NaN thing?

PR #7042, opened 2024-10-14
fix BatchNorm error on input dtype that is not float32

hbhungg:
Reopen of #7025 since I accidentally closed it.

Using `BatchNorm` in training mode, if input is of `dtypes.double` will cause `AssertionError: assign dtype mismatch dtypes.float != dtypes.double`. This is due to the `elf.running_mean` and `self.running_var` was create with default `dtypes.float`.

Should I fix it by:
- Cast `self.running_mean` and `self.running_var` to input dtype if floats. (Current changes)
- Cast `x` to float32 in `self.calc_stats(x)`.


tinygrad/nn/__init__.py
```diff
@@ -54,6 +55,7 @@ def __call__(self, x:Tensor) -> Tensor:
     batch_mean, batch_var = self.calc_stats(x)
     # NOTE: wow, this is done all throughout training in most PyTorch models
     if self.track_running_stats and Tensor.training:
+      if x.dtype in dtypes.floats: self.running_mean, self.running_var = self.running_mean.cast(x.dtype), self.running_var.cast(x.dtype)
```
geohot, line 58: cast in the assign!

PR #7046, opened 2024-10-14
GGUF support

leopf:
Supports loading Q4_0 models, which use both Q4_0 and Q6_K quantization.

## Example
```python
import os
from extra.gguf import load_gguf
from tinygrad import Tensor, dtypes, Device

fn = "Meta-Llama-3-8B-Instruct.Q4_0.gguf"
gguf_tensor = Tensor.empty(os.stat(fn).st_size, dtype=dtypes.uint8, device=f"disk:{fn}").to(Device.DEFAULT)
kv_data, state_dict = load_gguf(gguf_tensor)
```

## TODO
- [x] Q8_0 support
- [x] end to end tests/testing data alignment
- [x] move to nn/state.py
- [x] more cleanup
- [?] testing assumptions in `ggml_tensor_to_numpy`


geohot:
Q8_0 should work also

tinygrad/nn/state.py
```diff
@@ -226,3 +226,69 @@ def persistent_load(self, pid): return deserialized_objects.get(pid, pid)
         base_offset += 8 + lens[i]
       f.seek(rwd)
       return TorchPickle(f).load()
+
+class GGUFConv:
```
geohot, line 230: Converter is a better name I think, Conv is convolution

leopf: I agree! I was unhappy with this class definition for a couple of other reasons, which is why I have moved the conversion logic to a function `ggml_data_to_tensor`, which also saved 7 sz.py lines.

leopf:
Now supporting Q4_0, Q4_1, Q6_K and Q8_0 + tests.

I am testing everything besides the assumptions made about the tensor layout in `ggml_tensor_to_numpy` (`test_disk_tensor.py`).
The only way I was able to test these assumptions is with the official gguf python package, which is a reimplementation of the libggml logic with numpy. Unfortunately this reimplementation is not feature complete and does not implement quantization for Q6_K, it also only supports numpy < 2 and does not free memory after closing gguf files, which is why I think its better to use unofficial low level bindings to the actual libggml, which allows us to test _almost_ everything.

I have tested everything but Q6_K with the official high level python package, by reimplementing `_test_dequantization` and `_test_load_gguf` with it (tests pass). But I think dealing with the `gguf` package as a dev dependency may induce headaches.

If we want both libraries for testing, I would be happy to include this.

geohot:
Looking good!

Can you get this llama running? https://huggingface.co/bartowski/Llama-3.2-1B-Instruct-GGUF

That'll be my end to end test, then the bounty is yours

leopf:
I have tried getting LLama3.2 1B to work, but the preexisting Transformer in llama3.py is not compatible with llama3.2. The major part that seems to be missing is RoPE scaling. I have implemented this but still only got gibberish (with the safetensor version). Due to my lack of memory, it is very time consuming to debug the llama examples on my machine.

I have gotten gpt2 running and included it as an example. You can try different quantization methods and compare against the unquantized version.

I hope this is sufficient to validate the change.
If getting llama3.2 to run is a requirement I will have to invest more time into this.

```bash
$ PYTHONPATH="." python examples/gpt2-gguf.py --temperature 0 --quant Q4_0 
ram used:  0.60 GB, lm_head.weight                                    : 100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 149/149 [00:01<00:00, 75.43it/s]
loaded weights in 1975.72 ms, 0.65 GB loaded at 0.33 GB/s
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [00:01<00:00, 71.77it/s]
What is the answer to life, the universe, and everything?

The answer to life is that we are all connected.

The answer to life is that we are all connected.

The answer to life is that we are all connected.

The answer to life is that we are all connected.

The answer to life is that we are all connected.

The answer to life is that we are all connected.

The answer to life is that we are all connected.

The answer to life is that we
```

```bash
$ PYTHONPATH="." python examples/gpt2.py --model_size "gpt2" --temperature 0
using NV backend
using gpt2
ram used:  0.50 GB, lm_head.weight                                    : 100%|██████████████████████████████████████████████████████████████████████████████████████████████████| 149/149 [00:00<00:00, 318.59it/s]
loaded weights in 467.99 ms, 0.65 GB loaded at 1.39 GB/s
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [00:01<00:00, 76.95it/s]
Generating text...
What is the answer to life, the universe, and everything?

The answer to life, the universe, and everything is that we are all connected.

The answer to life, the universe, and everything is that we are all connected.

The answer to life, the universe, and everything is that we are all connected.

The answer to life, the universe, and everything is that we are all connected.

The answer to life, the universe, and everything is that we are all connected.

The answer to
```

geohot:
GPT2 is fine, testing locally now

geohot:
Congrats, bounty is yours! george@tinygrad.org to claim, can pay either via PayPal or USDC on ethereum

geohot:
So it's merged, but I have a bunch of nitpits.

* It should be `gguf_load`, not `load_gguf` to match torch and safe
* It should raise if a GGUF format type is not supported. Right now it returns None
* The supported types should be documented

leopf:
Agree with naming and docs.  I should have seen the naming inconsistency.
"Right now it returns None" - it should raise a KeyError, since there is no block size defined for it.

PR #7051, opened 2024-10-14
browser as a device for webgpu (proof of concept)

mesozoic-egg:
1. run webgpu on an actual browser
2. single TCP connection
3. can scale to multiple remote browser connections

https://github.com/user-attachments/assets/c6f71094-30a8-45f4-9839-f900df3e920e

Essentially turning the browser into a server that can take request from the python program, thought this would be fun to explore, Let me know what y'all think


geohot:
Hmm, interesting. Distributing training?

mesozoic-egg:
Browser is more ubiquitous than any other webgpu library that require additional installation, you can train on almost any internet connected devices if they have chrome installed 

Also chrome will likely be the most performant because all the serious graphics web app will target it? (this is just speculation) 

If I or someone else get webgpu compiler done I can better evaluate this approach, for now it's just an idea lol

PR #7060, opened 2024-10-15
refactor: improved the consistency of the frexp in transcendental

hikettei:
- Removed the weird line `{dtypes.float64: 1022, dtypes.float32: 126, dtypes.float16: 15}[v.dtype]` from `frexp`.


chenyuxyz:
so this is just shifting both by 1 and resulting compute stays the same? is there a reference to these hard coded m1, m2 constants?

PR #7063, opened 2024-10-15
Bring back WebGPU

wpmed92:
Continuing the work started in https://github.com/tinygrad/tinygrad/pull/6551

This PR is currently mainly the changes from that PR, with the addition of fixing `inf` issues by passing in `inf` as a uniform.
I didn't include the test skips from the PR as I'd like to skip as few as possible.

With infs fixed test ops are improving somewhat:

Before

Screenshot 2024-10-15 at 12 49 27

After
Screenshot 2024-10-15 at 12 53 35

TODO:
- [x] Fix broken threefry (no long in webgpu): solved with `ulong` shift emulation with `vec2<u32>`
- [x] Fix inf tests  
RE: _"Webgpu allows handling of runtime infs and nans to be implementation specific (https://www.w3.org/TR/WGSL/#differences-from-ieee754), and they can yield indeterminate values."_

Operations involving inf might flush to zero, I reported a bug in `wgpu` https://github.com/pygfx/wgpu-py/issues/620, but reading the specs it might be indeed spec compliant, it's just very weird that multiplying a positive number by positive infinity results in zero.
I found that `someVar * inf` only produces correct `inf` if the `inf` is passed in as a uniform, so I'm doing that. Otherwise, if `inf` is a constant, the result will be 0. This is consistent between `chrome` and `wgpu`.

- [x] Fix cast, cmp tests
- [x] Export and run EfficientNet
- [x] Fix nan tests
- [x] Fix broken fancy indexing (issue was wrong local size order)


wpmed92:
Currently working on solving threefry without ulong (webgpu doesn't have it).
I'll introduce ulong as vec2<u32>, and support the oprations (<<, |) needed for threefry. The other option would be to rework threefry itself to avoid ulong.

wpmed92:
Down to `19` from initial `28` errors.

Screenshot 2024-10-22 at 23 47 31

`test_randomness`:

Screenshot 2024-10-22 at 23 48 59

wpmed92:
Randomness tests are passing, and EfficientNet compiles and runs again:

Screenshot 2024-10-23 at 10 07 53

Screenshot 2024-10-23 at 13 43 01

wpmed92:
- `test_ops.py` passing with fancy indexing skips
- `test_randomness.py` passing
- EfficientNet model export and run test passing
- Linting passing -> max line count reached so the job itself fails

I'll increase test coverage and do refactoring/fixing.

wpmed92:
Fancy indexing tests passing now. Local size wasn't passed in in the correct order (was `lidx2`, `lidx0`, `lidx1`)

test/test_compile_failures.py
```diff
@@ -1,11 +1,12 @@
 import unittest
-from tinygrad import Tensor
+from tinygrad import Tensor, Device
 from tinygrad.engine.realize import lower_schedule
 
 class TestCompileFailures(unittest.TestCase):
   def compile(self, out:Tensor):
     for _ in lower_schedule(out.schedule()): pass
 
+  @unittest.skipIf(Device.DEFAULT == "WEBGPU", "No uint8 on WEBGPU")
```
chenyuxyz, line 9: use the `is_dtype_supported` helper

tinygrad/device.py
```diff
@@ -48,13 +48,15 @@ class BufferOptions:
   cpu_access: bool = False
   host: bool = False
   nolru: bool = False
+  wgpu_bool: bool = False
```
chenyuxyz, line 51: final solution cannot have these, could be a rewrite rule on store/load dtype

tinygrad/renderer/wgsl.py
```diff
@@ -0,0 +1,92 @@
...
+    (UPat(UOps.ALU, name="x", dtype=dtypes.ulong, arg=BinaryOps.SHR), lambda r,x: f"ushr({r[x.src[0]]},u32({gep_ulong(r, x.src[1])}))"),
+  ])
+
+  def render_dtype(self, dt:DType, mutable=True) -> str: return "var"
+  def render_kernel(self, function_name:str, kernel:List[str], bufs:List[Tuple[str,Tuple[DType,bool]]], uops:List[UOp], prefix=None) -> str:
+    local_size = [num for _, num in sorted([u.arg for u in uops if u.op is UOps.SPECIAL and u.arg[0][0] == 'l'], key=lambda x: x[0])]
+    if not local_size: local_size = [1]
+    bind_it = iter(range(len(bufs)))
+    prg = "fn nan() -> f32 { let bits = 0xffffffffu; return bitcast<f32>(bits); }\n"
+    prg += "fn ushl(v: vec2<u32>, shift: u32) -> vec2<u32> {\n"
```
geohot, line 85: This sort of stuff should be done with rewrite rules

tinygrad/device.py
wpmed92: Already have rewrite rules for bool. This is for a workaround that there are no bool buffers in webgpu, you have to create them as uint32. This hack is from the original pr. I have another idea for it.

wpmed92:
Thinking about that `uchar` in wgsl is doable, even though it doesn't have that datatype. We just have to rewrite the load/stores like so:
`index/4` selects the `dword`, than`(dword >> (8*(index%4)))&0xFF` gives us the byte. Will give it a try. It would also solve the `wgpu_bool` hack.

wpmed92:
Current state:
- It's currently `81` lines above the limit
- ulong shift rendering is cleaner now
- the pattern to rewrite ulong mul to ulong shift is done (in random there are kernels that shift by multiplying)
- tests now use `is_dtype_supported`, so much less `Device.DEFAULT == "WEBGPU"`
- the new style index doesn't work with webgpu because you can't do c-style pointer arithmetic, so I'm splitting the new style index to buffer and index, and rendering array-style buffer accesses

test/test_ops.py
```diff
@@ -484,9 +487,12 @@ def test_div_naninf(self):
     helper_test_op([(45,65)], lambda x: (-math.inf)/x)
     helper_test_op([(45,65)], lambda x: math.nan/x)
 
+  @unittest.skipIf(Device.DEFAULT == "WEBGPU", "nan location mismatch")
   def test_pow_full(self):
     helper_test_op([(45,65), (45,65)], lambda x,y: x**y)
     helper_test_op([(45,65), (45,65)], lambda x,y: x.pow(y))
+
+  @unittest.skipIf(Device.DEFAULT == "WEBGPU", "nan location mismatch")
```
geohot, line 495: This needs to be fixed

tinygrad/device.py
wpmed92: As discussed during the meeting this can stay for now.

wpmed92:
- `nan` issues in `pow` tests have been solved by rewriting `a * select(1, nan, gate)` to `select(a, nan, gate)`. This bypasses the missing nan propagation in webgpu.
- `test_isnan` now passing by rewriting `a != a` to a custom `is_nan()` function, taken from here: https://github.com/pygfx/wgpu-py/blob/main/tests/test_not_finite.py  (obfuscates the compiler so that the nan check is correct)

wpmed92:
With supporting `char`, `uchar`, `short`, `ushort`, test coverage has been increased

Screenshot 2024-11-07 at 14 57 10

to

Screenshot 2024-11-07 at 15 33 32

I'll continue with cleanup, and separate out some stuff to standalone PRs, like the `num_batches_tracked` "long" vs "int" with `is_dtype_supported` in `device` can now be done.

tinygrad/nn/__init__.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, sz:int, eps=1e-5, affine=True, track_running_stats=True, mome
     self.weight: Optional[Tensor] = Tensor.ones(sz) if affine else None
     self.bias: Optional[Tensor] = Tensor.zeros(sz) if affine else None
 
-    self.num_batches_tracked = Tensor.zeros(1, dtype='long', requires_grad=False)
+    self.num_batches_tracked = Tensor.zeros(1, dtype='long' if Device.DEFAULT != "WEBGPU" else 'int', requires_grad=False)
```
wpmed92, line 39: This can be done cleanly now with `is_dtype_supported`

tinygrad/renderer/wgsl.py
```diff
@@ -0,0 +1,93 @@
...
+  code_for_op = {**CStyleLanguage.code_for_op, TernaryOps.WHERE: lambda a,b,c,dtype: f"select({c},{b},{a})"}
+  nan = "nan()"
+  type_map = type_map
+
+  string_rewrite = PatternMatcher([
+    (UPat(Ops.CONST, dtype=dtypes.bool, name="x"), lambda ctx,x: "true" if x.arg else "false"),
+    (UPat(Ops.CONST, dtype=(dtypes.char, dtypes.short), name="x"), lambda ctx,x: f"i32({x.arg})"),
+    (UPat(Ops.CONST, dtype=(dtypes.uchar, dtypes.ushort, dtypes.uint32), name="x"), lambda ctx,x: f"bitcast<u32>({x.arg}i)" \
+     if x.arg < 0 else f"{x.arg&0xFFFFFFFF}u"),
+    (UPat(Ops.CONST, arg=math.inf, name="x"), lambda ctx, x: f"{type_map[x.dtype]}({ctx.infinity})"),
```
geohot, line 62: This shouldn't be duplicated because of the type thing

tinygrad/renderer/wgsl.py
```diff
@@ -0,0 +1,93 @@
+from typing import List, Tuple
+from tinygrad.dtype import DType, PtrDType, dtypes
+from tinygrad.ops import UOp, Ops, UnaryOps, TernaryOps, PatternMatcher, UPat, GroupOp
+from tinygrad.renderer.cstyle import CStyleLanguage, base_rewrite
+from tinygrad.helpers import strip_parens
+import math
+
+def fixup_binops(c,a,b):
```
geohot, line 8: Why isn't this two rules instead of a function?

tinygrad/runtime/ops_webgpu.py
```diff
@@ -0,0 +1,68 @@
...
+    return ((timestamps:=self.device.queue.read_buffer(query_buf).cast("Q").tolist())[1] - timestamps[0]) / 1e9 if wait else None
+
+class WebGpuAllocator(Allocator):
+  def __init__(self, device): self.device = device
+  def _alloc(self, size: int, options):
+    scaled_size = size*options.scale_size
+    return self.device.create_buffer(size=scaled_size, usage=wgpu.BufferUsage.STORAGE | wgpu.BufferUsage.COPY_DST | wgpu.BufferUsage.COPY_SRC)
+  def copyin(self, dest, src: memoryview):
+    scale = dest.size // len(src)
+    if scale == 4: self.device.queue.write_buffer(dest, 0, bytearray([byte for b in src for byte in [b, 0, 0, 0]]))
```
geohot, line 52: There has to be a simpler way to do this.

tinygrad/renderer/wgsl.py
wpmed92: I propose a `render_cast` in cstyle: use it in the `CONST` string rewrites where we cast, and override `render_cast` in wgsl. This adds 1 line in cstyle, 1 line in wgsl, and removes 4 from wgsl, so -2 overall.

wpmed92: https://github.com/tinygrad/tinygrad/pull/7687

wpmed92:
Stable diffusion running in cli with WebGPU. Exporting next.
Screenshot 2024-11-14 at 00 03 38

geohot:
Packed stuff looking good! I'm getting you the lines right now

geohot:
You have lines now, failure is a (valid) linter error

wpmed92:
@geohot new SD deployment is ready: https://wpmed92.github.io/stable-diffusion-wgpu-tinygrad/

tinygrad/runtime/ops_webgpu.py
```diff
@@ -0,0 +1,66 @@
...
+from tinygrad.device import  Compiled, Allocator, Compiler
+from tinygrad.renderer.wgsl import WGSLRenderer
+import wgpu
+import struct
+
+def create_uniform(wgpu_device, val) -> wgpu.GPUBuffer:
+  buf = wgpu_device.create_buffer(size=4, usage=wgpu.BufferUsage.UNIFORM | wgpu.BufferUsage.COPY_DST)
+  if isinstance(val, int): wgpu_device.queue.write_buffer(buf, 0, val.to_bytes(4, "little"))
+  else: wgpu_device.queue.write_buffer(buf, 0, struct.pack('<f', val))
+
```
geohot, line 11: why is there an empty line here?

tinygrad/runtime/ops_webgpu.py
```diff
@@ -0,0 +1,66 @@
...
+    if wait:
+      command_encoder.resolve_query_set(query_set=query_set, first_query=0, query_count=2, destination=query_buf, destination_offset=0)
+    self.dev.queue.submit([command_encoder.finish()])
+    return ((timestamps:=self.dev.queue.read_buffer(query_buf).cast("Q").tolist())[1] - timestamps[0]) / 1e9 if wait else None
+
+# WebGPU buffers have to be 4-byte aligned
+class WebGpuAllocator(Allocator):
+  def __init__(self, dev): self.dev = dev
+  def _alloc(self, size: int, options):
+    aligned_size = (size + 3) & ~3
```
geohot, line 49: Use `round_up` in helpers

geohot:
Getting:
```
WGPU: request_adapter() is deprecated, use request_adapter_sync() instead.
WGPU: request_device() is deprecated, use request_device_sync() instead.
```
when running local on Mac. This okay?

tinygrad/renderer/wgsl.py
```diff
@@ -0,0 +1,99 @@
...
+    (UPat(Ops.BITCAST, dtype=(dtypes.short, dtypes.ushort), name="x"), lambda ctx,x: f"bitcast<{type_map[x.dtype]}>({ctx[x.src[0]]}&0xFFFF)"),
+    (UPat(Ops.BITCAST, name="x"), lambda ctx,x: f"bitcast<{type_map[x.dtype]}>({ctx[x.src[0]]})"),
+    (UPat(Ops.LOAD, src=(UPat.var("b"), UPat.var('v'), UPat.var("g"))), lambda ctx,b,v,g: f"select({ctx[v]}, {ctx[b]}, {ctx[g]})"),
+    (UPat(Ops.LOAD, src=(UPat.var('b'),), allow_any_len=True), lambda ctx, b: ctx[b]),
+    (UPat(Ops.INDEX, src=(UPat.var("buf"), UPat.var('idx'))),
+    lambda ctx,buf,idx: f"{ctx[buf]}[{strip_parens(ctx[idx]) if idx.arg == Ops.ADD else ctx[idx]}]"),
+    (UPat(Ops.STORE, src=(UPat.var('b'), UPat.var("v")), allow_any_len=True),\
+     lambda ctx,b,v: f"atomicAdd(&{ctx[b]}, {ctx[v]});" if b.src[0].dtype.itemsize < 4 else f"{ctx[b]} = {ctx[v]};"),
+    # fix nan check: 'a != a -> is_nan()'
+    (UPat(Ops.CMPNE, src=(UPat.var("a"), UPat.var("b"))), lambda ctx,a,b: f"is_nan({ctx[a]})" if a == b else None),
```
geohot, line 79: You don't need that equals check, just call the vars the same thing

`(UPat(Ops.CMPNE, src=(UPat.var("a"), UPat.var("a"))), lambda ctx,a: f"is_nan({ctx[a]})"`

Vipitis:
> Getting:
> 
> ```
> WGPU: request_adapter() is deprecated, use request_adapter_sync() instead.
> WGPU: request_device() is deprecated, use request_device_sync() instead.
> ```
> 
> when running local on Mac. This okay? Easy to fix or any reason not to?

That's a change in [v0.19](https://github.com/pygfx/wgpu-py/blob/main/CHANGELOG.md#v0190---04-11-2024), so using `_sync()` will need to require that as minimum.

tinygrad/renderer/wgsl.py
```diff
@@ -0,0 +1,99 @@
...
+    (UPat(Ops.DEFINE_LOCAL, name="x"), lambda ctx,x: f"var<workgroup> {ctx[x]}: array<{type_map[x.dtype.base]}, {x.arg[1]}>;"),
+    (UPat(Ops.BITCAST, dtype=(dtypes.char, dtypes.uchar), name="x"), lambda ctx,x: f"bitcast<{type_map[x.dtype]}>({ctx[x.src[0]]}&0xFF)"),
+    (UPat(Ops.BITCAST, dtype=(dtypes.short, dtypes.ushort), name="x"), lambda ctx,x: f"bitcast<{type_map[x.dtype]}>({ctx[x.src[0]]}&0xFFFF)"),
+    (UPat(Ops.BITCAST, name="x"), lambda ctx,x: f"bitcast<{type_map[x.dtype]}>({ctx[x.src[0]]})"),
+    (UPat(Ops.LOAD, src=(UPat.var("b"), UPat.var('v'), UPat.var("g"))), lambda ctx,b,v,g: f"select({ctx[v]}, {ctx[b]}, {ctx[g]})"),
+    (UPat(Ops.LOAD, src=(UPat.var('b'),), allow_any_len=True), lambda ctx, b: ctx[b]),
+    (UPat(Ops.INDEX, src=(UPat.var("buf"), UPat.var('idx'))),
+    lambda ctx,buf,idx: f"{ctx[buf]}[{strip_parens(ctx[idx]) if idx.arg == Ops.ADD else ctx[idx]}]"),
+    (UPat(Ops.STORE, src=(UPat.var('b'), UPat.var("v")), allow_any_len=True),\
+     lambda ctx,b,v: f"atomicAdd(&{ctx[b]}, {ctx[v]});" if b.src[0].dtype.itemsize < 4 else f"{ctx[b]} = {ctx[v]};"),
```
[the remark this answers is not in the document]
wpmed92: This `atomicAdd` approach relies on zero-initialized buffers. With `wgpu` on Metal it is the case, otherwise `test_dtype`/`test_dtype_alu` would have failing tests. I tried to rewrite with `atomicCompareExchange`, which wouldn't care if the buffer has garbage or all zeros, but that's not yet supported in `wgpu`.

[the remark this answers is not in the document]
geohot: Hmm, I see you aren't using the LRUAllocator. Let me see if I can write a failing test, if not it's fine

geohot:
Merged! Bounty is yours, can pay via PayPal or USDC on Ethereum

PR #7064, opened 2024-10-15
auto detect light/dark mode

mesozoic-egg:
https://github.com/user-attachments/assets/0e0af927-dcb0-4faa-bc1d-ae7eb2576a8c

(have to refresh page for styles that were applied by JS)


geohot:
Why should we spend lines to support light mode? Let's focus on making a single experience really really good

PR #7065, opened 2024-10-15
new scheduler [pr]

geohot:
No reason to work on this in toonygrad, I think it's mature enough to work on here.

`python3 test/test_tiny.py` and `python3 test/test_ops.py TestOps.test_tiny_add` passes.

Still missing a way to map the UOps to buffers, so everything reruns.


geohot:
@Qazalin is this enough for you to go off of to do the LazyBuffer -> UOp replacement? the very basic tests pass, if you get the scheduler tests working on this I can deal with making sure buffers get freed

(top priority still VIZ bugs)

PR #7078, opened 2024-10-16
generic elementwise view rewrite rule + merge_views

Qazalin:


Qazalin:
yea these VALIDs are correct now, what was the stride...
image

geohot:
That VALID can just be rewritten to True since it doesn't have a mask.

PR #7082, opened 2024-10-16
do not serialize MTL library

mesozoic-egg:
not the best fix. I think with JIT and IndirectCommandBuffer, serializing the IR of MTLLibrary may not be necessary?


mesozoic-egg:
Running `METAL=1 DISABLE_COMPILER_CACHE=1 PYTHONPATH='.' python examples/beautiful_mnist.py` seem to reveal we are using unsigned long for metal? might be a bug because it's not supported
image
image
```c++
```

geohot:
I really don't want to do this. It's been proposed many times, the bug just needs to be fixed, not hacked around.

mesozoic-egg:
yeah i was dumb, it's `unsigned long long` not being supported

PR #7086, opened 2024-10-16
no None in ei [pr]

geohot:


geohot:
Ugh I never merged this? We want this.

PR #7089, opened 2024-10-16
viz ui fixes

Ry4nW:


Qazalin:
the scroll is kinda laggy, I think notion does this correctly:

https://github.com/user-attachments/assets/8b3149de-7064-4738-b20b-d580abba50d0

geohot:
Yea, scrolling shouldn't use javascript.

Qazalin:
For the resize/collapse I think figma does it right.

right now when i resize the left sidebar the right one changes too
there should be one collapse button that just focuses the screen on the center.

https://github.com/user-attachments/assets/87d3fcff-92e2-4f20-9360-df62baafe994

Qazalin:
scroll is better but it stopped working when I hit this kernel
`VIZ=1 python3 test/test_schedule.py`
image

Qazalin:
I just tried to go down with arrow keys:
image

VIZ=1 test/test_ops.py TestOps.test_tiny_add

Qazalin:
collapse functions well! 
can you give this a max width for resize:
image

Qazalin:
autoscroll doesn't work and the scroll is laggy, I don't like this
```
    // reduce scroll lag
    if (Date.now() - lastPress > 300) lastPress = Date.now();
```
why do you need any javascript for the default scroller?

Ry4nW:
there's too many rerenderings otherwise. not laggy anymore, just slow. maybe some select and enter for faster scroll, but there's already mouse scroll for that

Qazalin:
after many feedback iterations, when I see bugs in the simplest possible test I lose trust in this diff.
image

That lag is unacceptable, I don't see why the speed would be any different than master.

Qazalin:
I'd merge two simple, small PRs for that vscode opener and collapse/resize.

PR #7095, opened 2024-10-16
viz touchups [pr]

Qazalin:


tinygrad/viz/eslint.config.mjs
Qazalin, line 1: The root dir one is enough.

PR #7106, opened 2024-10-16
some info on test_failure_52

nimlgen:


geohot:
This fails locally on Mac too...what's wrong with 52? Could it be an actual out of bounds memory access?

`PYTHONPATH="." python3 test/test_linearizer_failures.py TestLinearizerFailures.test_failure_52`

nimlgen:
Yeah, It's an oob access

PR #7107, opened 2024-10-16
MSE in tensors.py and tests implemented

littlemountainman:
MSE with testing implemented


ifxlh:
Would love this!

geohot:
Shorter to write out

PR #7111, opened 2024-10-16
fetch tiny llama if nothing passed

dmx2:
`examples/llama.py` fetches [TinyLlama](https://huggingface.co/TinyLlama/TinyLlama-1.1B-Chat-v1.0) weights if nothing is passed.


examples/llama.py
```diff
@@ -344,6 +349,9 @@ def main(string: str):
   parser.add_argument("--shard", type=int, default=1, help="number of devices to load the weights to")
 
   args = parser.parse_args()
+  if not args.model:
```
chenyuxyz, line 352: i think it's better to only do this only if `gen==tiny`, otherwise this silently overwrites the default `gen==1`

examples/llama.py
```diff
@@ -190,6 +190,11 @@ def load(fn:str):
   else:
     return torch_load(fn)
 
+def fetch_tiny_llama():
+  print("No model specified. Fetching tiny LLaMA...")
```
chenyuxyz, line 194: does not make sense to have this print inside the function

dmx2: Ok.

examples/llama.py
dmx2: Ok, was going off geohot comment from Discord:
"can someone make llama autodownload small llama weights if you don't pass in anything?"

Set to =="tiny" only now.

chenyuxyz:
@geohot do you want to test this since you asked for it?

examples/llama.py
```diff
@@ -190,6 +190,10 @@ def load(fn:str):
   else:
     return torch_load(fn)
 
+def fetch_tiny_llama():
+  fetch("https://huggingface.co/TinyLlama/TinyLlama-1.1B-Chat-v1.0/resolve/main/tokenizer.model", name="tokenizer.model")
```
geohot, line 194: Don't use this name, what if you have other llamas?

dmx2: Oh true. Model and tokenizer need names to be recognized, but `subdir` should be sufficient. It mimics the `llama3.py` fetch as well.

geohot:
Do you need a subdir? Again, same problem. The default hash naming is a lot better

dmx2:
The hash naming is definitely better. The issue was that the model load function used path suffix checks and the tokenizer was found by file name.

The load function should be redone. If this looks ok, I will do that with the data hashing in a separate PR.

examples/llama.py
```diff
@@ -180,15 +180,21 @@ def convert(name) -> Tensor:
...
+def load(fn: str):
   if fn.endswith('.index.json'):
     with open(fn) as fp: weight_map = json.load(fp)['weight_map']
     parts = {n: load(str(Path(fn).parent / Path(n).name)) for n in set(weight_map.values())}
     return {k: parts[n][k] for k, n in weight_map.items()}
   elif fn.endswith(".safetensors"):
     return safe_load(fn)
   else:
-    return torch_load(fn)
+    try: return torch_load(fn)
```
geohot, line 191: Hmm, I don't love this try except. Can you show it always works? If it does, you def don't need the `elif` clause

examples/llama.py
```diff
@@ -439,7 +449,7 @@ def main(string: str):
 
   LLAMA_SUFFIX = {"1": "", "2": "-2", "3": "-3", "code": "-code", "tiny": "-tiny"}[args.gen]
   MODEL_PATH = args.model or Path(__file__).parents[1] / f"weights/LLaMA{LLAMA_SUFFIX}/{args.size}"
```
geohot, line 451: This should be the defaults to argparse, not this

examples/llama.py
```diff
@@ -341,9 +347,13 @@ def main(string: str):
   parser.add_argument("--size", type=str, default=None, help=f"""Size of model to use {", ".join([f"{list(v.keys())} for gen '{k}'" for k, v in MODEL_PARAMS.items()])}""")
   parser.add_argument("--quantize", type=str, default=None, help="Quantize the weights to int8 or nf4 in memory")
   parser.add_argument("--model", type=Path, default=None, help="Folder with the original weights to load, or single .index.json, .safetensors or .bin file")
+  parser.add_argument("--tokenizer", type=Path, default=None, help="Path to tokenizer.model if not in model directory.")
   parser.add_argument("--shard", type=int, default=1, help="number of devices to load the weights to")
 
   args = parser.parse_args()
+  if args.model is None and args.gen == "tiny":
```
geohot, line 354: Huh, that's not the default?!?

geohot:
Closing. Too much feedback and there's a random whitespace change in here

PR #7115, opened 2024-10-17
uop index

geohot:


geohot:
We probably don't want this. What we want is just the casting to float4* to be in the load/store

PR #7120, opened 2024-10-17
test for viz with exception in rewrite rule [pr]

Qazalin:
TestVIz.test_track_rewrites_with_exception -> exception in the fxn decorated with @track_rewrites
TestViz.test_track_rewrites_with_exception_alt -> exception in the rewrite rule rhs
both viz up to the point of exception:
image


Qazalin:
This was for demo that VIZ tracks graphs up to exceptions as long as there is at least one call to graph_rewrite.

PR #7121, opened 2024-10-17
check libc path

Ry4nW:


geohot:
tests failing, not tested

PR #7135, opened 2024-10-17
early rewrite global ASSIGN

Qazalin:


Qazalin:
meh, it's better the way it is.

PR #7143, opened 2024-10-18
Refactor hip_bfloat16 cast into uop

ignaciosica:
~Strange issue where CI fails but whole test suit passes in red box.~
CI
image
tiny10
image

Fixed! Thanks Qazalin


ignaciosica:
New rule to add float as intermediary cast is now needed as before, with the float() operator implemented, the compiler automatically casted to float when casting hip_bfloat16 to other types.

``` c
inline __attribute__((device)) hip_bfloat16(float val) {
  union { float fp32; unsigned int u32; } u = {val};
  if (~u.u32 & 0x7f800000) { u.u32 += 0x7fff + ((u.u32 >> 16) & 1); } 
  else if (u.u32 & 0xffff) { u.u32 |= 0x10000; }
  data = (u.u32 >> 16);
}
--------
inline __attribute__((device)) operator float() const {
  unsigned int uval = data << 16;
  return *reinterpret_cast<float*>(&uval);
}
```
``` python
def cast_float_bf16(x: UOp) -> UOp:
  x = x.bitcast(dtypes.uint)

  is_not_inf_nan = -x & 0x7f800000
  has_mantissa = x & 0xffff
  x = is_not_inf_nan.where(x + ((x >> 16) & 1) + 0x7fff, has_mantissa.where((x | 0x10000), x))

  return (x >> 16).cast(dtypes.ushort).bitcast(dtypes.bfloat16)
--------
lambda x: (x.bitcast(dtypes.ushort).cast(dtypes.uint)<<16).bitcast(dtypes.float)
```

geohot:
Is the issue on AMD fixed? Looks good to merge

geohot:
Going to merge, but two nitpicks.

* Is there an _rlshift_ in Python? We should add those ones too
* This uop cast float -> bfloat16 should probably be somewhere more generic, and also used on platforms (like M1) that don't support blfloat16

PR #7145, opened 2024-10-18
Fix libc where find_library returns None

jla524:


nimlgen:
We can do in a more general way pathing clang2py, but I think it's good for now. Thanks!

PR #7149, opened 2024-10-18
schedule_cache

Qazalin:
todo:
- [x] multi output SINK
- [x] ASSIGN

misc:
- [x] tensor metadata
- [x] add back assign asserts
- [x] how do we assert cycles.


Qazalin:
Up to https://github.com/tinygrad/tinygrad/pull/7149/commits/e78861f0037ebcdb9c36094536cc0e1f99589d1a - it will look at LOAD src to see if a LOAD is to the assigned version of the Buffer or it's pre realized version. This is very slow to compute for big big graphs like resnet. I think I need to stick a bool arg in LOAD.

Qazalin:
found out I need to do this last, after all of schedule.py is rewritten.

PR #7152, opened 2024-10-18
Windows cuda

nimlgen:


geohot:
Cool, I support this. Can you add Windows (non CUDA okay) to CI?

Issue #7020, opened 2024-10-12
TinyJit returns wrong values


chenyuxyz:
from #6803 
```
from tinygrad import Tensor, TinyJit

@TinyJit
def f(x) -> Tensor: return (x + 1).realize()

a1, a2 = Tensor([1.1]), Tensor([1.2])
print(f"{f(a1).numpy()=}")
print(f"{f(a2).numpy()=}")
diff = f(a1) == f(a2)
print(f"{diff.numpy()=}")
```
returns
```
f(a1).numpy()=array([2.1], dtype=float32)
f(a2).numpy()=array([2.2], dtype=float32)
diff.numpy()=array([ True])
```

if commenting out TinyJit
```
f(a1).numpy()=array([2.1], dtype=float32)
f(a2).numpy()=array([2.2], dtype=float32)
diff.numpy()=array([False])
```

chenyuxyz:
generated code without JIT
```
kernel void E_n1(device bool* data0, const device float* data1, const device float* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  float val0 = *(data1+0);
  float val1 = *(data2+0);
  *(data0+0) = ((val0!=val1)!=1);
}
```
and with JIT
```
kernel void E_n1(device bool* data0, const device float* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  float val0 = *(data1+0);
  *(data0+0) = ((val0!=val0)!=1);
}
```

nimlgen:
I think we're just comparing the same tensor twice, since jit returns the same tensor. This works:

```
@TinyJit
def f(x) -> Tensor: return (x + 1).realize()

@TinyJit
def f2(x) -> Tensor: return (x + 1).realize()

a1, a2 = Tensor([1.1]), Tensor([1.2])
print(f"{f(a1).numpy()=}")
print(f"{f2(a2).numpy()=}")
diff = f(a1) == f2(a2)
print(f"{diff.numpy()=}")
```

And If we're writing to the same underlying buffer on each jit run, even if the kernel is correct, we end up comparing the same results, so...

tobias17:
The buffer reuse seems to be the problem:
```python
@TinyJit
def f(x) -> Tensor: return (x + 1).realize()

def p(x) -> Tensor: return (f(x) + 1.0).realize()

if __name__ == "__main__":
[...]

PR #7154, opened 2024-10-18
Tensor Clone Function

tobias17:
## Overview

Implements a `Tensor.clone()` function that creates a copy of the data in a separate buffer.

## Why

This is some follow-up work from #7020 where JIT reuses buffers to store the output. While this is normally not an issue, calling the same JIT function without using the output buffer overwrites the output data.

The clone function is a proposed solution to this problem, allowing the user to clone out the data before it gets overwritten by the next call.

## Testing

The following is the existing example to show the JIT reuse issue
```python
from tinygrad import Tensor, TinyJit

@TinyJit
def f(x:Tensor) -> Tensor: return (x + 1).realize()

for _ in range(2):
  a1, a2 = Tensor([1.1]), Tensor([1.2])
  print(f(a1).numpy(), f(a2).numpy())
  diff = f(a1) - f(a2)
  print(diff.numpy())
  print()
```
Which produces the following result:
```
[2.1] [2.2]
[0.]

[2.1] [2.2]
[0.]
```

Adding the clone function fixes this
```python
from tinygrad import Tensor, TinyJit

@TinyJit
def f(x:Tensor) -> Tensor: return (x + 1).realize()

for _ in range(2):
  a1, a2 = Tensor([1.1]), Tensor([1.2])
  print(f(a1).numpy(), f(a2).numpy())
  diff = f(a1).clone().realize() - f(a2).clone().realize()
  print(diff.numpy())
  print()
```
(note that realize() must be called after to fix this JIT issue, there was consideration to build this realize into clone() itself)
```
[2.1] [2.2]
[-0.10000014]

[2.1] [2.2]
[-0.10000014]
```

This also works using sharded tensors
```python
from tinygrad import Tensor, TinyJit

@TinyJit
def f(x:Tensor) -> Tensor: return (x + 1).realize()

for _ in range(2):
  a1, a2 = Tensor([1.1, 2.1]).shard(('NV','NV'), axis=0), Tensor([1.2, 2.2]).shard(('NV','NV'), axis=0)
  print(f(a1).numpy(), f(a2).numpy())
  diff = f(a1).clone().realize() - f(a2).clone().realize()
  print(diff.numpy())
  print()
```
```
[2.1 3.1] [2.2 3.2]
[-0.10000014 -0.10000014]

[2.1 3.1] [2.2 3.2]
[-0.10000014 -0.10000014]
```

## Next Steps

I'm throwing this PR out to gain feelers on this approach. It does feel a little hacky, working around the problem rather than having JIT use new buffers for each return, but at the same time JIT using new buffers each time would have performance implications while not gaining anything for 99% of cases.

If JIT using new buffers each time is the approach we want to go then I am glad to close this PR and implement that.

If this is the approach we want to go then I will add some tests.


geohot:
Hmm, so I guess I'm fine with this API to handle assign, and the rest is just details.

tinygrad/tensor.py
```diff
@@ -322,6 +322,15 @@ def numpy(self) -> 'np.ndarray':  # type: ignore [name-defined] # noqa: F821
     assert all_int(self.shape), f"no data if shape is symbolic, {self.shape=}"
     return np.frombuffer(self._data(), dtype=_to_np_dtype(self.dtype)).reshape(self.shape)
 
+  def clone(self) -> Tensor:
+    """
+    Creates a clone of this tensor allocating a seperate buffer for the data.
+    """
+    ret = Tensor(self.lazydata.clone(), self.device, requires_grad=self.requires_grad)
+    if self.grad is not None: ret.grad = self.grad.to(self.device)
+    if hasattr(self, '_ctx'): ret._ctx = self._ctx
```
geohot, line 331: Careful, why do you want to do this? Clone should be at the Function level

geohot:
Needs tests

tinygrad/tensor.py
tobias17: I switched it to `self.grad.clone()` instead, but don't actually need it in my case just thought it would be a good addition.

Why would I need this at the function level when clone is only doing a data copy? It is not doing anything more than the `Tensor.to` method does and uses the same boundaries.

chenyuxyz:
the decision in the meeting was that this is probably fine to merge, @geohot agree?

PR #7164, opened 2024-10-20
fold true branch of where with uop_given_valid

chenyuxyz:
gated load and where can use the same logic to simplify


chenyuxyz:
issues with arange

PR #7167, opened 2024-10-20
remove True/False arg in UOps.RANGE

spectralsouls:


spectralsouls:
will fix the failling tests

tinygrad/codegen/linearize.py
```diff
@@ -37,13 +37,13 @@ def get_recursive_children(x:UOp, end:UOps, include_self=False) -> Set[UOp]:
...
-    if u.op is UOps.RANGE and u.arg[1]:
-      priority += u.arg[0]
+    if u.op is UOps.RANGE and u.arg:
+      priority += u.arg
       for p in range_phi[u]:
         priority += 10000*len([r for r in range_srcs[p] if not any(i in range_phi[u] for i in range_phi[r])])
     # prefer uops that are loop children
     else:
-      priority -= sum([(l.arg[0]+1) + 1000*l.arg[1] for l,ss in scope_children.items() if l.op is UOps.RANGE and u in ss])
+      priority -= sum([(2 if l.arg else 1) + 1000*l.arg for l,ss in scope_children.items() if l.op is UOps.RANGE and u in ss])
```
geohot, line 46: Not right!

geohot:
Please fix tests and make process replay pass if possible

geohot:
This misses the point! The point is to remove the need for the True/False, not encode it in a different way

PR #7170, opened 2024-10-20
post-schedule lazybuf from Buffer [pr]

Qazalin:


Qazalin:
resnet training hangs
image

PR #7185, opened 2024-10-21
Add __rlshift__ and __rrshift__ special methods

ignaciosica:


chenyuxyz:
these should come with some usage (tests or other part of the code that uses it) to justify

ignaciosica:
> these should come with some usage (tests or other part of the code that uses it) to justify

I know, that's why I didn't added in the first place, but George requested them (https://github.com/tinygrad/tinygrad/pull/7143#issuecomment-2425818519). Right now the are not used anywhere in the codebase and I don't think it's a very common case where you want to shift a const with a uop, but maybe I'm missing something and that's why he requested them

Qazalin:
tbh I'm supportive of adding this, I wanted something like this in the UOp API when I was writing some tests.

PR #7186, opened 2024-10-21
Tensor.keccak("sha3_256")

leopf:
Calculates a keccak hash over the last dimension. Right now only sha3 (224, 256, 384, 512) is supported, but it is easy to add more types of keccak hashes.

Tested, see `test/test_keccak.py`.

Its batched, making tree hashing trivial to implement:
```python
def tree_hash(data: Tensor, cfg: str | tuple[int, int] = "sha3_256"):
  rate = { "sha3_224": 144, "sha3_256": 136, "sha3_384": 104, "sha3_512": 72 }.get(cfg, cfg[0])
  b_out, blk_size, is_lvl_1 = (200 - rate) // 2,  rate - 1, True # rate - 1 is max per permutation

  while data.shape[-1] != b_out or is_lvl_1:
    pad_size = 0 if data.shape[-1] % blk_size == 0 else blk_size - data.shape[-1] % blk_size
    data = data.pad(tuple(None for _ in range(len(data.shape) - 1)) + ((0, pad_size),))
    data = data.reshape(*data.shape[:-1], -1, blk_size).keccak(cfg).flatten(-2)
    is_lvl_1 = False

  return data
```

I have added this to the tensor. Does it belong there?

It's very slow.


leopf:
It crashes with NV MOCKGPU. Seems to be something in the mmap emulation.

geohot:
Revived https://github.com/tinygrad/tinygrad/pull/10545

leopf:
I added a simple benchmark.

Running `NDATA=4000 DEBUG=2 python test/external/external_benchmark_keccak.py` dispatches 2180 kernels and takes 48s on my machine. I have not been able to run it with NDATA=4_000_000 yet.

I am happy to improve the benchmark and address you comments from #10545 but I think this is a long way from high performance.

wozeparrot:
for performance, we really only care about 4kb blocks, everything above that will be tree'd.

leopf:
Added shake, improved bm.

getting:
```bash
$ python test/external/external_benchmark_keccak.py
--- benchmarking (hash: shake_128, data size: 64.0 MiB, block size: 4096 B, batch size: 16384)
run: 1, elapsed time: 51706.50 ms, throughput: 1.24 MiB/s
run: 2, elapsed time: 14529.62 ms, throughput: 4.40 MiB/s
run: 3, elapsed time: 216.88 ms, throughput: 295.09 MiB/s
run: 4, elapsed time: 118.63 ms, throughput: 539.50 MiB/s
run: 5, elapsed time: 118.79 ms, throughput: 538.75 MiB/s
```

```bash
$ NRUNS=20 BS=1 python test/external/external_benchmark_keccak.py
--- benchmarking (hash: shake_128, data size: 0.00390625 MiB, block size: 4096 B, batch size: 1)
run: 1, elapsed time: 26833.39 ms, throughput: 0.00 MiB/s
run: 2, elapsed time: 14348.02 ms, throughput: 0.00 MiB/s
run: 3, elapsed time: 190.87 ms, throughput: 0.02 MiB/s
run: 4, elapsed time:   2.74 ms, throughput: 1.42 MiB/s
run: 5, elapsed time:   1.94 ms, throughput: 2.01 MiB/s
run: 6, elapsed time:   2.02 ms, throughput: 1.94 MiB/s
run: 7, elapsed time:   2.74 ms, throughput: 1.42 MiB/s
run: 8, elapsed time:   2.07 ms, throughput: 1.89 MiB/s
run: 9, elapsed time:   2.15 ms, throughput: 1.82 MiB/s
run: 10, elapsed time:   2.60 ms, throughput: 1.50 MiB/s
run: 11, elapsed time:   3.04 ms, throughput: 1.29 MiB/s
run: 12, elapsed time:   2.17 ms, throughput: 1.80 MiB/s
run: 13, elapsed time:   2.40 ms, throughput: 1.63 MiB/s
run: 14, elapsed time:   2.02 ms, throughput: 1.94 MiB/s
run: 15, elapsed time:   2.02 ms, throughput: 1.93 MiB/s
run: 16, elapsed time:   2.77 ms, throughput: 1.41 MiB/s
run: 17, elapsed time:   3.26 ms, throughput: 1.20 MiB/s
run: 18, elapsed time:   1.99 ms, throughput: 1.96 MiB/s
run: 19, elapsed time:   2.48 ms, throughput: 1.57 MiB/s
run: 20, elapsed time:   2.32 ms, throughput: 1.68 MiB/s
```

tinygrad/tensor.py
```diff
@@ -1909,6 +1909,56 @@ def std_mean(self, axis:int|Sequence[int]|None=None, keepdim=False, correction=1
...
+
+    # https://keccak.team/keccak_specs_summary.html
+
+    def ctensor(l: Sequence[ConstType], dtype: DType = dtypes.uint64): return Tensor.stack(*(Tensor(v, dtype=dtype, device=self.device) for v in l))
+    rot_offsets = [44, 43, 21, 14, 28, 20, 3, 45, 61, 1, 6, 25, 8, 18, 27, 36, 10, 15, 56, 62, 55, 39, 41, 2]
+    rot_offsets_v0, rot_offsets_v1 =  ctensor([0] + [1 << v for v in rot_offsets]), ctensor([1] + [1 << (64 - v) for v in rot_offsets])
+
+    # calculated from π step
+    reorder_indexes = [0,6,12,18,24,3,9,10,16,22,1,7,13,19,20,4,5,11,17,23,2,8,14,15,21]
+    rnd_const_masks = [ctensor([v]).pad((0, 24)) for v in (1, 0x8082, 0x800000000000808a, 0x8000000080008000, 0x808b, 0x80000001, 0x8000000080008081,
```
wozeparrot, line 1952: these shouldn't need to be a tensor

leopf: The round const must only be applied to the first element of the state (https://github.com/wozeparrot/tinygrad/blob/hashing/extra/hashing.py#L54).

The idea is to basically XOR the entire state when applying the round const, by padding the round const with 0 (neutral element for XOR). So it would make sense to create these tensors only once for all iterations of f1600, which is why I have made it a tensor there.

The alternative would be to do this in a new line in the f1600 loop:
```python
state = (state[:,:1] ^ rc[i]).cat(state[:,1:], dim=1)
```

This makes the graph more complex and compilation slower. Under the hood this does something very similar with a few extra steps.

Am I missing or misunderstanding something?

wozeparrot: oh, ic what this is doing, yea this is probably the best way to do that

leopf:
What is left to be done to merge this? Do you have some speed target in mind?

I am not sure how much faster this can be, by optimizing on the frontend. performance is memory constrained.

Ways to speed it up:

execution:
- faster list indexing (used for permuting), like `tensor[:,[1, 3, 3, 7]]`. The current behavior also creates copies.
- as both loops are sequential, it probably would make sense to fuse the entire hash (or one f1600) and only parallelize over the batch dim.
- other general stuff like scheduling, loop splitting

compilation:
- XOR self folding (x ^ 0 -> x) - reduces scheduling time and kernel size (happy to submit this as a followup)
- allowing Ops.XOR for REDUCE_AXIS and using it to reduce `t1 = (p[:,:,0] ^ p[:,:,1] ^ p[:,:,2] ^ p[:,:,3] ^ p[:,:,4]).roll(-1, 1)`. I tested it by adding XOR in spec.py. This is probably the right way of describing it, but it causes the scheduler to create extra kernels atm, making execution slower.

wozeparrot:
Can you add a external test for all the test vectors, something like the one in my pr. 

We can merge this first since stuff is blocked on this, then start looking into speed.

leopf:
Added external tests. There are recursion errors if messages get too big, which is why there is a <8192.

I ran them. They all passed.

tinygrad/tensor.py
wozeparrot: this should be able to just be a normal tensor construction then without using `ctensor`?

leopf: Yes. `ctensor` is used for the rotation vectors anyway, so the fn will still be required. Using ctensor here is just a more concise way of creating this const tensor (without having to specify dtype and device again). While the stack is unnecessary for this single element, it doesnt actually make the tensor graph more complex than if I created and padded a tensor here. I think it will end up costing an extra line, if dtype device and unsqueeze must be written out again.

Can change it to not using ctensor, if you prefer it that way.

wozeparrot: yea was mainly just wondering if it made the tensor graph more complex, as long as it doesn't, I agree that this is cleaner

wozeparrot:
cool! george@tinygrad.org to claim bounty.

if you want to submit followup prs for speed, feel free.

PR #7192, opened 2024-10-21
GGUF cleanup

leopf:
In reference to the last comments on #7046.

`ggml_data_to_tensor` does not return `None` if a type is not supported, it raises a `KeyError`. Mypy does not recognize this, which means that explicitly setting the return type to `Tensor` creates a mypy error.
Do we want to spend one line on an explicit error?


leopf:
Ill make a separate PR for the error.

Issue #7161, opened 2024-10-19
Overlapped setitem failed on some devices


MaxWolf-01:
The following piece of code creates different behavior, depending on which environment variable (CLANG, NV, GPU, ...) is set.
With CLANG=1 the test fails.
With GPU=1 or CUDA=1 the test passes (which it should - the code seems correct).
Tested on latest tinygrad from source (https://github.com/tinygrad/tinygrad/commit/30989fb459f38f0443abed3343f37e475622f069) ; Ubuntu 22.04; 3070 Ti Laptop GPU

CASE `GPU=1`, `CUDA=1`: Test passes

CASE `CLANG=1`: Test fails
```
Traceback (most recent call last):
  File "...", line 68, in <module>
    test_buffered_tensor_queue()
  File "...", line 54, in test_buffered_tensor_queue
    assert res == [[4.0], [3.0], [2.0], [1.0]], res
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: [[4.0], [3.0], [3.0], [3.0]]
```

Code to reproduce:
```python
from collections import deque

from tinygrad import Tensor
from tinygrad.dtype import DType, dtypes


class BufferedTensorQueue:
    __slots__ = "shape", "_buffered", "_queue"

    def __init__(self, size: int, shape: tuple[int, ...], dtype: DType | None = None) -> None:
        self.shape = shape
        self._buffered: deque[Tensor] = deque(maxlen=size)
        self._queue = Tensor.zeros((size, *shape), dtype=dtype).contiguous()

    def __len__(self) -> int:
        """Returns the number of buffered tensors."""
        return len(self._buffered)

    def get(self) -> Tensor:
        return self._queue

    def buffer(self, item: Tensor) -> None:
        assert item.shape == self.shape, f"Expected shape {self.shape}, got {item.shape}."
        self._buffered.appendleft(item)

    def enqueue_buffered(self) -> None:
        """Enqueues the buffered elements and clears the buffer."""
        if len(self._buffered) == 0:
            return
        buffered = Tensor.cat(*self._buffered).unsqueeze(-1)
        replace = len(self._buffered)
        self._queue[replace:] = self._queue[:-replace]  # todo sth is going wrong here - test fails, might be tiny issue
        self._queue[:replace] = buffered
        self._buffered.clear()


def test_buffered_tensor_queue(this_fails: bool = False) -> None:
    # import os
[...]

PR #7193, opened 2024-10-21
Failing test: different behavior on different devices

MaxWolf-01:
A simple test for #7161 
~~Interestingly it fails LLVM and not CLANG in CI (in constrast to running it locally).~~ It also fails on LLVM locally.

### DIFF
--- a/test/test_device_behavior.py
+++ b/test/test_device_behavior.py
@@ -1,0 +1,28 @@
+import os
+import unittest
+
+from tinygrad import Tensor
+
+
+class TestDeviceBehavior(unittest.TestCase):
+    def test_device_behavior_gpu(self):
+        os.environ["GPU"] = "1"
+        self._should_pass()
+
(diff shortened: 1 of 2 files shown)


geohot:
Any way to make it simpler? Is this the minimum failing case?

MaxWolf-01:
Three lines now. This is the simplest I could come up with.

geohot:
What file does it belong in? Can you add a comment and mark expected failure?

MaxWolf-01:
Not sure why CLANG passes in CI but fails locally for me.

```bash
CLANG=1 python -c """
from tinygrad import Tensor
t = Tensor([[3.0], [2.0], [1.0]]).contiguous()
t[1:] = t[:-1]
assert t.tolist() == [[3.0], [3.0], [2.0]]  # assertion error
"""
```

geohot:
Interesting, merged! Thanks for the clean test

PR #7196, opened 2024-10-21
ggml_data_to_tensor touchups

chenyuxyz:
tiny reordering and variable name changes


leopf:
Forgot to add an explicit return type here earlier. But I believe mypy wont let you, since it interprets the return type as `Tensor | None`.

chenyuxyz:
fixed the return type, i think indent the whole block is better than `blocks = t if blk_info is None else ...` just to work around type checker

leopf:
Agreed. Looks great!

PR #7203, opened 2024-10-22
Fix bf16 tests using torch as reference value

ignaciosica:
Enables much stricter tolerance (same as for other floats) now in binary operations.
Makes it easier adding other types only supported in torch in the future.

`.float().numpy()` float casting before numpy call is required.

### DIFF
--- a/test/test_dtype_alu.py
+++ b/test/test_dtype_alu.py
@@ -3,6 +3,7 @@ import unittest
 from tinygrad import Tensor, dtypes, Device
 import operator
 import numpy as np
+from ml_dtypes import bfloat16
 from hypothesis import given, strategies as strat, settings
 from tinygrad.dtype import DType
 from tinygrad.helpers import CI, getenv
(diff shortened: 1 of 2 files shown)


test/test_dtype_alu.py
```diff
@@ -60,24 +64,28 @@ class ht:
...
 def universal_test(a, b, dtype, op):
-  if not isinstance(op, tuple): op = (op, op)
+  if not isinstance(op, tuple): op = (op, op, op)
   tensor_value = (op[0](Tensor([a], dtype=dtype), Tensor([b], dtype=dtype))).numpy()
-  numpy_value = op[1](np.array([a]).astype(_to_np_dtype(dtype)), np.array([b]).astype(_to_np_dtype(dtype)))
-  if dtype is dtypes.bfloat16: np.testing.assert_allclose(tensor_value, numpy_value, atol=1e-3, rtol=1e-2)
-  elif dtype in dtypes_float: np.testing.assert_allclose(tensor_value, numpy_value, atol=1e-10)
-  else: np.testing.assert_equal(tensor_value, numpy_value)
+  if dtype in torch_dtypes_map.keys():
+    reference_value = op[2](torch.tensor([a]).to(torch_dtypes_map[dtype]), torch.tensor([b]).to(torch_dtypes_map[dtype])).float().numpy()
```
chenyuxyz, line 70: use `.item()` to avoid the `float()` conversion should work on torch tensor

test/test_dtype_alu.py
```diff
@@ -60,24 +64,28 @@ class ht:
   bool = strat.booleans()
 
 def universal_test(a, b, dtype, op):
-  if not isinstance(op, tuple): op = (op, op)
+  if not isinstance(op, tuple): op = (op, op, op)
```
chenyuxyz, line 67: don't really like this change, i think just using torch for bfloat16 is simpler

ignaciosica: Ok, will refactor

test/test_dtype_alu.py
ignaciosica: Thanks

ignaciosica:
Making draft as I don't like the copy paste feel of `universal_test_unary_torch`. Will work on it tommorrow. Also, one run still failed in [CI](https://github.com/tinygrad/tinygrad/actions/runs/11451286748/job/31860245877#step:22:183) in a previous run.

PR #7204, opened 2024-10-22
mypy fix

Ry4nW:
added check for get
errors1

fixing
image


geohot:
Please write readable code

PR #7205, opened 2024-10-22
remove mask logic in View add [pr]

chenyuxyz:
real_stride can simplify


chenyuxyz:
yea was mostly interested to see what are being merged now, the goal was to clean up the merging logic

btw process replay would pass if all merged vews are still merged, which is not the case without those mask logic

PR #7206, opened 2024-10-22
no need to DEFINE_VAR arg sort [pr]

geohot:


geohot:
See, I saved a line :)

I knew some of them were useless. Btw, I wonder about doing this at scale. What lines can you remove and your tests still pass?

PR #7213, opened 2024-10-22
qcom faster copies

nimlgen:
sync:   13.08 ms @ 5.13 GB/s
vs
sync:    8.98 ms @ 7.47 GB/s


nimlgen:
With gdb OpenCL just uses libc's memcpy as we do (and they set WC flags on allocations). Will see what we're missing rn

nimlgen:
Will revisit. This cache is broken, but kernel uses this one.

PR #7218, opened 2024-10-22
does this vm2 mask case do anything [pr]

chenyuxyz:


chenyuxyz:
it's doing something and it does not unit test coverage

PR #7231, opened 2024-10-23
uop_given_valid in real_strides

chenyuxyz:
simplified idx allows deriving more strides


chenyuxyz:
no process replay diff now. 2 more TODO cases in TestRealStrides after this

PR #7251, opened 2024-10-24
Revert "move metal tc check to renderer [pr]"

geohot:
Reverts tinygrad/tinygrad#7248

Breaks process replay!


Qazalin:
I think it failed because of:
```
def __reduce__(self): return self.__class__, ()
```
in Renderer.

PR #7269, opened 2024-10-24
Upgrade mypy to 1.13.0. [pr]

### LINKED ISSUE #7176: Bump mypy to 1.12.1
And fix the type issues

cesar-guibo:
Upgrades mypy to 1.12.1 #7176

\# type: ignore was added to Tensor.data because mypy is inferring
memoryview[int] as the return type but the exact type of a memoryview
can't be declared in python to fix this inference

### DIFF
--- a/tinygrad/multi.py
+++ b/tinygrad/multi.py
@@ -109,9 +109,11 @@ class MultiLazyBuffer(MathTrait):
     new_real = [all(transposed) for transposed in zip(*[mlb.real for mlb in msrcs])] if not_all_real else self.real
     assert any(new_real), "output contains no real lb"
     for mlb in msrcs:
-      if (mlb.axis == axis and (mlb.axis is None or mlb.bounds == bounds)) or not_all_real: srcs.append(mlb.lbs)
-      elif mlb.axis is None and axis is not None: srcs.append(to_sharded(mlb.lbs, axis, bounds))
-      else: srcs.append(to_sharded([mlb.copy_to_device(lb.device) for lb in mlb.lbs], axis, bounds))
+      if axis is None or ((mlb.axis == axis and mlb.bounds == bounds) or not_all_real): srcs.append(mlb.lbs)
+      else:
+        assert bounds is not None
+        if mlb.axis is None: srcs.append(to_sharded(mlb.lbs, axis, bounds))
+        else: srcs.append(to_sharded([mlb.copy_to_device(lb.device) for lb in mlb.lbs], axis, bounds))
     new_real_lbs:Dict[int,LazyBuffer] = {i:lsrcs[0].alu(op, *lsrcs[1:]) for i,(lsrcs,r) in enumerate(zip(zip(*srcs), new_real)) if r}
     # NOTE: const dtype should match real
     real_dtype = next(iter(new_real_lbs.values())).dtype
--- a/tinygrad/runtime/graph/hcq.py
+++ b/tinygrad/runtime/graph/hcq.py
@@ -1,5 +1,5 @@
 import collections, time
-from typing import List, Any, Dict, cast, Optional, Tuple, Set, Union
+from typing import List, Any, Dict, cast, Optional, Tuple, Set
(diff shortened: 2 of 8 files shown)


geohot:
Between this and #7280 I think you can make one great PR. I prefer the asserts to the elifs in that PR

tinygrad/runtime/ops_python.py
```diff
@@ -74,9 +74,11 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
         dl[i] = dtype
         if uop is UOps.DEFINE_GLOBAL:
           assert dtype.fmt is not None
+          if sys.version_info[1] < 12: assert dtype.fmt != 'e' and dtype.fmt != '@e'
```
geohot, line 77: Copy pasted with below. Why isn't this needed in the other PR?

tinygrad/runtime/graph/hcq.py
```diff
@@ -51,7 +51,13 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
...
 
     for j,ji in enumerate(self.jit_cache):
-      enqueue_dev = ji.prg.device if (is_exec_prg:=isinstance(ji.prg, CompiledRunner)) else Device[ji.bufs[1].device] #type:ignore
+      if not (is_exec_prg:=isinstance(ji.prg, CompiledRunner)):
+        assert ji.bufs[1] is not None, "can't select device"
+        enqueue_dev = Device[ji.bufs[1].device]
+      else: enqueue_dev = ji.prg.device
+      assert isinstance(enqueue_dev, HCQCompiled), "not instance of HCQCompiled"
+      assert enqueue_dev.hw_copy_queue_t is not None, 'no hw copy queue'
+      enqueue_queue:Union[HWComputeQueue, HWCopyQueue, HWCommandQueue]
```
geohot, line 60: Is HwCommandQueue needed? Pay attention to inheritance

tinygrad/runtime/ops_python.py
cesar-guibo: The way that #7280 was defining DTypeFmtType as a Generic instead of a Literal was probably not allowing mypy to check if memoryview was accepting the half format

tinygrad/runtime/graph/hcq.py
cesar-guibo: Later in the code, enqueue_queue gets reassigned to a value of type HWCommandQueue. But inheritance allows switching this Union for just a HWCommandQueue type since the others inherit from it. I will make this change

tinygrad/runtime/ops_python.py
cesar-guibo: I removed these asserts because even though python supports half for memoryview in 3.12, the mypy signatures for memoryview still don't support it in version 1.13.0

tinygrad/multi.py
```diff
@@ -109,9 +109,11 @@ def alu(self, op:Union[MetaOps, UnaryOps, BinaryOps, TernaryOps], *in_srcs:Multi
     new_real = [all(transposed) for transposed in zip(*[mlb.real for mlb in msrcs])] if not_all_real else self.real
     assert any(new_real), "output contains no real lb"
     for mlb in msrcs:
-      if (mlb.axis == axis and (mlb.axis is None or mlb.bounds == bounds)) or not_all_real: srcs.append(mlb.lbs)
-      elif mlb.axis is None and axis is not None: srcs.append(to_sharded(mlb.lbs, axis, bounds))
-      else: srcs.append(to_sharded([mlb.copy_to_device(lb.device) for lb in mlb.lbs], axis, bounds))
+      if axis is None or ((mlb.axis == axis and mlb.bounds == bounds) or not_all_real): srcs.append(mlb.lbs)
```
geohot, line 112: Why does this logic look different? I can't review this

geohot:
Want to start with non controversial change?. Don't change logic and don't add type ignore

PR #7278, opened 2024-10-24
generic vmin/vmax of ints mul

chenyuxyz:


chenyuxyz:
don't think this hits in real use case though

PR #7287, opened 2024-10-25
remove shr/shl in transcendental

chenyuxyz:
`>>` and `<<` work on LazyBuffer


chenyuxyz:
hmm i guess SHR and SHL are meant to use in late rewrite

PR #7291, opened 2024-10-25
use a metaclass instead of overriding new

geohot:


geohot:
Will do this in ordering PR

PR #7293, opened 2024-10-25
finally big graph

Qazalin:
TODO:
- [ ] schedule_cache!!
- [x] find a way to preserve metadata as we're rewriting uops.


Qazalin:
back to draft til I fix metadata.

FeepingCreature:
Haven't looked into it in depth but according to `git bisect` since this commit the sdxl.py demo fails with `RecursionError: maximum recursion depth exceeded` in 

```
  File "tinygrad/ops.py", line 728, in rewrite
    new_src = tuple(map(self.rewrite, n.src))
```

Issue #7180, opened 2024-10-21
Finish #7179 adding 1B llama 3.2 support in GGUF


geohot:
See #7179 for a start

PR #7295, opened 2024-10-25
Llama 3.2 1B load from GGUF

eliotgolding:
#7180 

Some kv_data from the file isn't used, namely the model params which are hardcoded and the tokens which are copied from 8B. (Maybe you want that changed for this example?)


geohot:
Good work!

PR #7298, opened 2024-10-25
order flipper as instant rewrite rule [pr]

geohot:
Hmm, it does have to be an instant rule it seems.

It's slower as a rewrite rule because it has to construct the UOp.

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -1058,9 +1058,9 @@ symbolic = PatternMatcher([
   # ** mod **
   # mod folding
   (UPat.var("x") % UPat.cvar("c", vec=False), lambda x,c: newx if 0 < c.arg and (newx:=mod_folding(x,c.arg)) is not None else None),
-  # flip order of ADD, MUL to sort
-  (UPat(UOps.ALU, name='x'), lambda x:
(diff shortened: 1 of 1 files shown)


tinygrad/ops.py
```diff
@@ -192,10 +192,11 @@ def dfs(x:Any, cache:dict):
 class UOpMetaClass(type):
   ucache:WeakValueDictionary[Tuple, UOp] = WeakValueDictionary()
   def __call__(cls, op:UOps, dtype:DType=dtypes.void, src:Tuple[UOp,...]=tuple(), arg:Any=None):
-    if op is UOps.ALU and arg in COMMUTATIVE and (src[0] is not src[1] and src[1].tuplize < src[0].tuplize and src[0].op is not UOps.NOOP):
-      src = src[::-1]
     if (ret:=UOpMetaClass.ucache.get(key:=(op, dtype, src, arg), None)) is not None: return ret
-    UOpMetaClass.ucache[key] = ret = super().__call__(op, dtype, src, arg)
```
unknownusername504, line 198: What if the key was not unique per communicative case instead of rewriting the result? Something like age order would allow lookup of (a+b) vs (b+a) if from left to right, args ordered oldest to newest. Age is just one example I think would be natural to encode in the data structure.

PR #6959, opened 2024-10-08
implement cummax

jeffzh4ng:
https://github.com/tinygrad/tinygrad/issues/6937

this pr implements cummax . the next prs i will
1. implement logcumsumexp using cummax
2. make two-stage optimization for cummax like cumsum

**cummax**
matches the [signature](https://pytorch.org/docs/stable/generated/torch.cummax.html) of `torch.cummax: Tensor, int -> Tuple[Tensor,Tensor`
computing indices cannot use `.pad().pool().argmax(-1)` because `.pad().pool()` loses original tensor's ordering
instead the proposed solution computes indices by:
- boolean mask, logically encoding if the current index is the cummax
- boolean mask => idx_mask, replacing T with the index, and F with 0
- idx_mask => max_idxs by filling in the zeros with the running index via `.pad().pool().max(-1)`

PR #7305, opened 2024-10-25
Implementation of cummax

cesar-guibo:
Fixed the compile warning that was causing the test failures in #6959


chenyuxyz:
i think `cummax` should share most of the code with `cumsum`, with `max` after pool instead of `sum`

cesar-guibo:
I plan on moving the current cumsum code to an associative scan implementation which can then be used to implement cumsum and cummax. I will create a separate standalone PR for that.

PR #7306, opened 2024-10-25
Refactor code_for_op into pm rules [run_process_replay]

ignaciosica:
Alternative `alu_rewrite` with functools.partial to avoid closure won't work due to how `deconstruct_function` in ops works. And wrapper function wouldn't avoid closure. I know it's kind of a hacky solution but is the best way I could find to create a separate rule for every op and dtype.

render_rewrite patterns for CLANG
image


ignaciosica:
Converting to draft as I am not completely sure about the closure hack

ignaciosica:
closing, I see no need of introducing a hack

PR #7308, opened 2024-10-25
CI speed: Invalidate gpuocelot cache monthly

roelofvandijk:
Automatically invalidate the `gpuocelot` build cache at the beginning of each month.

The gpuocelot build cache went out of sync, resulting in a recompilation on each CI job (see the ninja warnings [here](https://github.com/tinygrad/tinygrad/actions/runs/11521503810/job/32075303745?pr=7305)).

```
ninja explain: CMakeFiles/gpuocelot_analysis.dir/src/analysis/DivergenceGraph.cpp.o is dirty
ninja explain: output CMakeFiles/gpuocelot_analysis.dir/src/analysis/DominatorTree.cpp.o older than most recent input /usr/include/linux/errno.h (1724981335987500422 vs 1727457702000000000)
```

The cache was not updated because the bad cache was hit every time (had a static cache key).

This PR removes ~1m30s of unnecessary work from the nv and ptx jobs. The nv job is the slowest unit test job.

Main:
image

PR:
image


geohot:
This is a hack, this is not the right fix for this issue

PR #7313, opened 2024-10-26
fix simplify_valid

chenyuxyz:
the simplex should compare with valid bound, not its vmin


chenyuxyz:
beam resnet works on green too

PR #7314, opened 2024-10-26
adding if/endif support to llvm renderer

nicktalati:
This adds IF/ENDIF support to the LLVM renderer as part of the "Gated Store Rewrite to UOps.IF" bounty. This is the first of three pull requests as discussed here: https://github.com/tinygrad/tinygrad/pull/7006.

### DIFF
--- a/test/test_renderer_failures.py
+++ b/test/test_renderer_failures.py
@@ -42,8 +42,8 @@ class TestCStyleFailures(unittest.TestCase):
     ret = _test_uop_result([Tensor([1])], uops)[0]
     self.assertEqual(ret[0], 1)
 
-@unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "need local")
 class TestPTXFailures(unittest.TestCase):
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_local, "need local")
   def test_gated_store_with_alu(self):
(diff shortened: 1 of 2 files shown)


geohot:
Can you write a test for this?

Qazalin:
Passes a simple test, mergable once the is_terminated one is justified with a test.

nicktalati:
Hi Qazalin,

Currently, the LLVM renderer is not adding any terminator instructions to blocks (other than branches handled by the if/endif logic). So I do believe the is_terminated check could technically be removed without altering functionality -- my motivation for including it was to guard against a potentially hard-to-find bug if any other branches or returns were added in the future. If that isn't the case or if it should be removed for the moment I'm happy to; otherwise I could use a bit more guidance on what a reasonable test would look like in this case.

Qazalin:
Any green diff must be justified with a unit test.
The reviewers in this project will look at every line you add and ask: If I remove that line, will any test show red?

nicktalati:
Thanks @Qazalin, I appreciate the explanation and think I am better understanding the standards of tinygrad. Given that, and since existing UOps don't support other terminator instructions, I think it best to remove that check; I'll take care of that now.

Qazalin:
done in https://github.com/tinygrad/tinygrad/pull/7345

PR #7315, opened 2024-10-26
fixing branch condition on UOps.IF in the ptx renderer

nicktalati:
This fixes a bug in the PTX renderer where the correct branch condition was inverted. This is the second of three PRs for the "Gated Store Rewrite to UOps.IF" bounty as discussed here: https://github.com/tinygrad/tinygrad/pull/7006. 

If this (and the previous PR) are successfully merged to master, I'll quickly create the final PR which will perform the actual graph rewrite, as well as remove all logic for gated stores from renderers.


geohot:
Can you add a test that would have failed without this?

PR #7333, opened 2024-10-28
Fix calculation of vmin and vmax in multiplication

S-Lykles:
Fix calculation of vmin and vmax in multiplication when one src is negative and the other src has negative min and positive max.

Using render from `test_uop_symbolic`
```python
print(render(-2 * Variable("a", -2, 4)))
```
Outputs:
```
('(a*-2)', -2147483648, 2147483647)
```

Added a test as well

### DIFF
--- a/test/unit/test_uop_vmin_vmax.py
+++ b/test/unit/test_uop_vmin_vmax.py
@@ -38,6 +38,13 @@ class TestVminVmaxProperties(unittest.TestCase):
     self.assertEqual(uop.vmin, -15)
     self.assertEqual(uop.vmax, -6)
 
+  def test_vmin_vmax_with_negative_multiplication2(self):
+    # vmin and vmax when multiplying by a negative number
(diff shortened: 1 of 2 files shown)


chenyuxyz:
cool, this used to have an issue but maybe it's due to unrelated bugs. let me test with resnet and will merge if it looks good

S-Lykles:
Wait this is kind of the same as #7278 ?

S-Lykles:
I was working on improving div_folding and I use the min/max of the sub-expression for that.

S-Lykles:
So like `a[-2, inf]*b[0, 4]` should be `[-8, inf]` and not `[nan, nan]`?

chenyuxyz:
yea bounds should not have nan as comparison with nan is weird

i am testing with #7336 to limit these to non-float, i think float bounds are not properly handled anyway and this approach is cleaner

PR #7336, opened 2024-10-28
limit UOps.ALU min/max to non-float only

chenyuxyz:
does this impact anything? some inf is incorrect now


chenyuxyz:
there's no kernel diff in benchmark, merging this as currently floats are not handled correctly. if we want to reintroduce float bounds (we will as we start to do math on lazy buffers) it should come with more tests

chenyuxyz:
clearly it's a lot slower, investigating...

PR #7360, opened 2024-10-29
use assertEqual with new style uops [pr]

geohot:


Qazalin:
meh, i had kept it because of print_diff. It was more useful than the default python repr.

PR #7365, opened 2024-10-29
Clang JIT v2

uuuvn:


tinygrad/renderer/cstyle.py
```diff
@@ -172,9 +172,11 @@ class ClangRenderer(CStyleLanguage):
 
   # language options
   buffer_suffix = " restrict"
-  type_map = {dtypes.bool:"_Bool", dtypes.half:"__fp16"}
+  type_map = {dtypes.bool:"_Bool", dtypes.half:"__fp16", dtypes.char:"signed char"}
```
uuuvn, line 175: I am not sure weather i should do it here or in dtype.py because despite being implementation defined almost everything has char signed by default

tinygrad/runtime/ops_clang.py
```diff
@@ -1,21 +1,46 @@
...
-import ctypes, subprocess, pathlib, tempfile
+import platform, struct, ctypes, subprocess, pathlib, tempfile, tinygrad.runtime.autogen.libc as libc
+from mmap import PROT_READ, PROT_WRITE, PROT_EXEC, MAP_ANON, MAP_PRIVATE
+from tinygrad.runtime.support.elf import elf_loader
 from tinygrad.device import Compiled, Compiler, MallocAllocator
-from tinygrad.helpers import cpu_time_execution, DEBUG, cpu_objdump
+from tinygrad.helpers import cpu_time_execution, DEBUG, OSX, getenv, cpu_objdump
 from tinygrad.renderer.cstyle import ClangRenderer
 
+MAP_JIT, global_handle = 0x0800 if OSX else 0, ctypes.CDLL(None)
```
uuuvn, line 9: https://man7.org/linux/man-pages/man3/dlopen.3.html

> If filename is NULL, then the returned handle is for the main
> program.  When given to [dlsym(3)](https://man7.org/linux/man-pages/man3/dlsym.3.html), this handle causes > a search for
> a symbol in the main program, followed by all shared objects
> loaded at program startup, and then all shared objects loaded by
> dlopen() with the flag RTLD_GLOBAL.

tinygrad/runtime/ops_clang.py
```diff
@@ -1,21 +1,46 @@
...
     # TODO: remove file write. sadly clang doesn't like the use of /dev/stdout here
     with tempfile.NamedTemporaryFile(delete=True) as output_file:
-      subprocess.check_output(['clang', '-shared', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib',
+      subprocess.check_output(['clang', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib',
                                '-', '-o', str(output_file.name)], input=src.encode('utf-8'))
       return pathlib.Path(output_file.name).read_bytes()
 
+class ClangJITCompiler(ClangCompiler):
+  def __init__(self, key="compile_clang_jit"):
+    # A bug in LLVM replaces two-iteration MAX reduce loop into a single **call** into fmaxf from libm when loop is unrolled
```
uuuvn, line 29: This is caused by some -O2 optimization (fine in -O1), narrowing it down to a single optimization pass (or confirming that it is a bug in loop unrolling and not just enabled by loop unrolling) and i'm going to file a bug to llvm and find a better way to work around this in tinygrad for affected clang versions

tinygrad/runtime/ops_clang.py
uuuvn: I thought about using autogen_stubs on macos in separate runner to generate bindings for mac-only libc functions but macos headers use a lot of syntax and features that aren't supported by clang2py

tinygrad/runtime/ops_clang.py
uuuvn: https://reviews.llvm.org/D122610

uuuvn: apparently this only happens with __fp16 and only after loop unroll.......

https://godbolt.org/z/z33cz5h8a

uuuvn: Tracking it here now - https://github.com/llvm/llvm-project/issues/114520

tinygrad/runtime/ops_clang.py
```diff
@@ -1,21 +1,45 @@
...
     with tempfile.NamedTemporaryFile(delete=True) as output_file:
-      subprocess.check_output(['clang', '-shared', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib',
+      subprocess.check_output(['clang', *self.args, '-O2', '-Wall', '-Werror', '-x', 'c', '-fPIC', '-ffreestanding', '-nostdlib',
                                '-', '-o', str(output_file.name)], input=src.encode('utf-8'))
       return pathlib.Path(output_file.name).read_bytes()
 
+class ClangJITCompiler(ClangCompiler):
+  def __init__(self, key="compile_clang_jit"):
+    # x18 is reserved register on at least macOS and Windows and is clobbered on context switch
+    super().__init__(key, ['-c', '-march=native', f'--target={platform.machine()}-none-unknown-elf', '-fno-math-errno',
```
geohot, line 30: Does this need an output file?

tinygrad/renderer/cstyle.py
geohot: This doesn't make sense

tinygrad/renderer/cstyle.py
```diff
@@ -175,9 +175,11 @@ class ClangRenderer(CStyleLanguage):
...
   code_for_op = {**({k:v for k,v in CStyleLanguage.code_for_op.items() if k not in [UnaryOps.EXP2, UnaryOps.SIN, UnaryOps.LOG2]}),
                  UnaryOps.SQRT: lambda x,dtype: f"__builtin_sqrt({x})" if dtype == dtypes.float64 else f"__builtin_sqrtf({x})"}
+  extra_matcher = PatternMatcher([(UPat.var('x', dtypes.float64).cast(dtypes.float16), lambda x: x.cast(dtypes.float32).cast(dtypes.float16))]) + \
```
geohot, line 181: If you want this it should be in a separate PR

tinygrad/runtime/graph/clang.py
```diff
@@ -14,7 +13,7 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
     super().__init__(jit_cache, input_rawbuffers, var_vals)
     if not all(isinstance(ji.prg, CompiledRunner) for ji in jit_cache): raise GraphException
 
-    prgs = '\n'.join(dedup([cast(CompiledRunner, ji.prg).p.src for ji in jit_cache]))
+    prgs = '\n'.join(dedup([cast(CompiledRunner, ji.prg).p.src.replace('void', 'static inline void') for ji in jit_cache]))
```
geohot, line 16: Do not write hacks

tinygrad/runtime/ops_clang.py
uuuvn: The ClangJITCompiler doesn't need it (actually the thing that doesn't like output to stdout is a linker if i recall correctly). I reverted my don't use file thing in v2 because now ClangCompiler is used by some other parts of tinygrad that might require it

PR #7368, opened 2024-10-29
global metadata 3

Qazalin:
follow-up on https://github.com/tinygrad/tinygrad/pull/7367
to unblock https://github.com/tinygrad/tinygrad/pull/7293


Qazalin:
found out this shouldn't exist in ctx, schedule_cache can definitely cache the same bufs with different metadata.

PR #7378, opened 2024-10-29
Fix save and loading tensors from the same file + change disk device logic for clarity

### LINKED ISSUE #6294: Safetensor save, load, re-save does not work

ryandorrington:
Fix  #6294


geohot:
Better, but I still don't understand the link check

ryandorrington:
> Better, but I still don't understand the link check

The disk device is reused, which means initialisation doesn't run and the file mapping logic is skipped:
```Python
if self.size is not None: return
```

I would guess that it would be better to delete the disk device when the file is unlinked:
```Python
pathlib.Path(fn).unlink(missing_ok=True)
```
but I can't figure that out.

geohot:
I don't understand why this is needed

PR #7381, opened 2024-10-29
Mypy upgrade

mku11:
Couple of notes for the code review:

1. multi.py: the reason for setting default values (0, []) is because downstream flow has type constraints on axis and bounds which believe are there for a good reason.

2. memoryview.cast: The problem that previous PRs faced is due to mypy looking at the stub instead of the impl. 
def cast(self, *args, **kwargs). The only way I could come up with to pacify mypy (not very readable) is to cast the params to Any. Let me know if you prefer type ignore it.

4. hcq.py: I can't think of any other way to do this other than unwrapping the if-statement and adding some extra lines, anyway this assists mypy to do type inference correctly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(name='tinygrad',
         'triton': ["triton-nightly>=2.1.0.dev20231014192330"],
         'linting': [
             "pylint",
-            "mypy==1.11.2",
+            "mypy==1.13.0",
             "typing-extensions",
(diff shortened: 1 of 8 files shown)


mku11:
fixes #7176

chenyuxyz:
you can submit pr to fix stuff that's less controversial, but adding Any defeats the purpose

PR #7396, opened 2024-10-30
skip rendering big graphs in viz

Qazalin:
Pagination needs to be its own project.
image

at least it shouldn't hang.


geohot:
I don't think this is the right fix. You should have a timeout. Number of nodes doesn't determine how hard a graph is to render.

PR #7404, opened 2024-10-30
Move the cumsum logic to _associative_scan

cesar-guibo:
Moves the code used by cumsum to _associative_scan and changes the sum executed after pool to a general _reduce. This allows the implementation of methods like cummax (#7305, #6959) without the need to rewrite the cumsum logic.


chenyuxyz:
stale, testing failing

PR #7410, opened 2024-10-30
Use `mmap` instead of `CDLL` in clang

donno2048:


donno2048:
@geohot why was it closed?

donno2048:
@geohot

PR #7432, opened 2024-10-31
pad2d to feature parity with torch pad

geohotstan:
not too sure how symbolic works or if it would work for these other modes.
The original pad2d symbolic for `constant` mode should be the same though.

also for `@unittest.skipIf(CI, "non-constant modes with negative pads have odd numerical errors in torch CI")`
these tests fail for CI but pass locally.


geohot:
We should probably rename it from `pad2d` to `pad` and confirm it matches torch

test/test_ops.py
```diff
@@ -1267,10 +1267,73 @@ def test_double_slice(self):
...
+
+  @unittest.skipIf(CI, "non-constant modes with negative pads have odd numerical errors in torch CI")
+  def test_pad2d_modes_negative_pads(self):
+    # negative pads
+    for shp, pad in (((3,3,3,3), (-1,2,2,-1)), ((1,1,5,5), (3,-3,0,-3)), ((1,1,5,5), (3,-5,1,-5)), ((1,1,5,5), (0,0,0,-5))):
+      for mode in ("reflect", "replicate", "circular"):
+        with self.subTest(shp=shp, pad=pad, mode=mode):
+          helper_test_op([shp], lambda x: torch.nn.functional.pad(x, pad, mode=mode), lambda x: x.pad2d(pad, mode=mode))
+
+    # TODO: fix backward for circular negative pads, problems occur when more-than-once wrap around happens when negative pads come before
```
geohotstan, line 1331: these negative pads with modes is cursed
for example for forwards:
image
I have no idea why torch does this.
I was very close to slapping a `if any(pB<0 or pA<0 for pB,pA in pX): raise NotImplementedError('negative padding with circular mode not implemented')` on it and calling it a day lol

but for the rest I think our pad2d matches torch's pad.

chenyuxyz:
i trust these are probably correct, but it's very hard to review. how about
(1) split the test and have different test names for different modes, make it as expected failure to start is fine
(2) having one mode per pr, or have similar things in one pr

it's slightly more work but i believe the result would be better

PR #7433, opened 2024-10-31
Add Tensor.hardsigmoid

geohotstan:


tinygrad/tensor.py
```diff
@@ -2333,6 +2333,19 @@ def sigmoid(self):
     ```
     """
     return F.Sigmoid.apply(self.cast(least_upper_float(self.dtype)))
+  def hardsigmoid(self, alpha:float=1/6, beta:float=0.5):
+    """
+    Applies the Hardsigmoid function element-wise.
+    NOTE: default `alpha` and `beta` values is taken from torch and differs from other common implementations
```
chenyuxyz, line 2339: i don't know what other common implementations are and find this odd, i think either say what the common one is, or just say it's the same as torch?

chenyuxyz (commented):
can you add a link to the doc as well. see https://github.com/tinygrad/tinygrad/pull/7419

geohotstan:
> can you add a link to the doc as well. see #7419

my bad forgot about this :(

> i think hard sigmoid is a shift version of relu6, can you reuse that?

if my math is right...
it would be `return (6 * alpha * self + 6 * beta).relu6() / 6` 
or we could just use relu? `return (alpha * self + beta).relu() - (alpha * self + beta - 1).relu()`

not sure which is nicer

PR #7449, opened 2024-10-31
Add shl and shr to llvmir

ignaciosica:
Saw chenyu pr on shl and shr refactor of transcendental and most of llvm tests failed due to the lack of support of those op in the backend. I don't know if it was removed or never added to the backend for some specific reason but I believe it is worth having support for them in llvm backend.

### DIFF
--- a/tinygrad/renderer/llvmir.py
+++ b/tinygrad/renderer/llvmir.py
@@ -1,7 +1,7 @@
 from typing import Dict, Callable, List, Optional
 from llvmlite import ir
 from tinygrad.dtype import DType, PtrDType, dtypes
-from tinygrad.ops import Op, UnaryOps, BinaryOps, TernaryOps, UOps, UOp
+from tinygrad.ops import Op, UnaryOps, BinaryOps, TernaryOps, UOps, UOp, PatternMatcher, UPat
 from tinygrad.renderer import Renderer
 
 MFLAGS = ('nsz', 'arcp', 'contract', 'afn', 'reassoc') # All from fast math, but nnan and ninf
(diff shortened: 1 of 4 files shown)


chenyuxyz:
i ended up not doing that because for the middle uop phrase we want to keep a minimal ops, so prefer `*2` over `<<1` because we can keep the simplification rules minimum. don't want to double rules to have one for mul and one for shl for example.

ignaciosica:
I see, okay, closing then! Thanks

ignaciosica:
But having support in cstyle and ptx don't force that? Why having support in llvm different?

ignaciosica:
Reopening, I believe I incorrectly assessed your comment

tinygrad/renderer/llvmir.py
```diff
@@ -63,7 +63,12 @@ class LLVMRenderer(Renderer):
     BinaryOps.MAX: lambda builder, x, y, dtype: builder.select(builder.icmp_unsigned(">", x, y) if is_bool_or_unsigned(dtype) else builder.icmp_signed(">", x, y) if dtypes.is_int(dtype) else builder.fcmp_unordered(">", x, y, flags=MFLAGS), x, y),  # noqa: E501
     BinaryOps.MOD: lambda builder, x, y, dtype: builder.urem(x, y) if is_bool_or_unsigned(dtype) else builder.srem(x, y) if dtypes.is_int(dtype) else builder.frem(x, y),  # noqa: E501
     BinaryOps.XOR: lambda builder, x, y, dtype: builder.xor(x, y), BinaryOps.AND: lambda builder, x, y, dtype: builder.and_(x, y), BinaryOps.OR: lambda builder, x, y, dtype: builder.or_(x, y), # noqa: E501
+    BinaryOps.SHL: lambda builder, x, y, dtype: builder.shl(x,y), BinaryOps.SHR: lambda builder, x, y, dtype: builder.lshr(x,y) if dtypes.is_unsigned(dtype) else builder.ashr(x,y), # noqa: E501
     TernaryOps.WHERE: lambda builder, x, y, z, dtype: builder.select(x, y, z)}
+  extra_matcher = PatternMatcher([
```
chenyuxyz, line 68: either enforce this dtype alignment in all backends, or just cast in the `code_for_op` similar to cstyle.

ignaciosica:
something broke fro ptx, making it a draft, reviewing in 1h

chenyuxyz:
i think it's easier to cast in llvm renderer, but if you are changing SHR/SHL spec you also need to update this https://github.com/tinygrad/tinygrad/blob/53db3478fef19d2c74d10a7b99e49a266e3abe16/tinygrad/ops.py#L786

ignaciosica:
> i think it's easier to cast in llvm renderer, but if you are changing SHR/SHL spec you also need to update this
> 
> https://github.com/tinygrad/tinygrad/blob/53db3478fef19d2c74d10a7b99e49a266e3abe16/tinygrad/ops.py#L786

Can't change shl and shr spec as llvmir and ptx have different requirements. How is this solved? I guess the spec just covers the tighter case that covers both?

Given x << y or x >> y: 
- llvmir require x.dtype == y.dtype
- ptx requires y.dtype == uint

what about the following spec?
``` python
*[(UPat(UOps.ALU, src=(UPat(name="x"), UPat(name="y")), name="alu", arg=shift),
 lambda alu,x,y: alu.dtype == x.dtype and (x.dtype == y.dtype or y.dtype == dtypes.uint)) for shift in (BinaryOps.SHL, BinaryOps.SHR)],
```

geohot:
That spec is fine

ignaciosica:
> That spec is fine

current one or the one in the comment?

Issue #7181, opened 2024-10-21
No CI testing of the float sym_infer pathway


geohot:
On QCOM devices, the global_sizes can be fractional. This should be tested in CI

https://github.com/tinygrad/tinygrad/actions/runs/11437282016/job/31816482620

sky0walker99:
i would like to work on this issue

sky0walker99:
.take

PR #7456, opened 2024-10-31
test float uop in sym_infer

gordbegli:
Affects https://github.com/tinygrad/tinygrad/issues/7181, whose linked attribute error was already fixed by https://github.com/tinygrad/tinygrad/commit/31fcccc77939b8d0d698c17a8c447f711c06efd1. 
This ~~updates types and~~ adds a test with `uop` as a `float`.

Edit: I removed all the type updates because I don't think the complexity is worth it for a single backend.


gordbegli:
Unsure if I took the right approach here. An alternative to updating GlobalCounter's types is to cast assignment to global_mem and global_ops.

chenyuxyz:
casting ops/mem to int is fine

kayo09:
Everything seems to be okay to fix #7181, is it just the builds that are failing? @gordbegli

tinygrad/runtime/support/hcq.py
```diff
@@ -143,7 +143,7 @@ def bind_sints_to_ptr(self, *vals:sint, ptr:int, fmt, mask:Optional[int]=None):
       else: self.mv_sints.append((mv, i, self._new_sym(val), mask))
 
   def _apply_var_vals(self, var_vals:Dict[Variable, int]):
-    resolved_syms = [sym_infer(sym, var_vals) for sym in self.syms]
+    resolved_syms = [cast(int,sym_infer(sym, var_vals)) for sym in self.syms]
```
gordbegli, line 146: Casted here since `var_vals` is defined as `Dict[Variable, int]`.

tinygrad/shape/view.py
```diff
@@ -317,7 +317,9 @@ def reshape(self, new_shape: Tuple[sint, ...]) -> Optional[View]:
     if self_all_int and not all_int(new_shape):
       if len(self.shape) != len(new_shape): raise ValueError(f"cannot symbolic reshape non-contiguous {self} -> {new_shape}")
       for si, so in zip(self.shape, new_shape):
-        if not isinstance(so, int): so = sym_infer(so, dict([v.unbind() for v in so.vars()]))
+        if not isinstance(so, (int, float)):
+          so = cast(int, sym_infer(so, dict([v.unbind() for v in so.vars()])))
+          if isinstance(so, float): raise ValueError(f"{new_shape=} contains Variable {so} that infers to float")
```
gordbegli, line 322: Raised error if `isinstance(so,float)` because `so` has type `SupportsIndex` and float doesn't.

tinygrad/shape/view.py
```diff
@@ -317,7 +317,9 @@ def reshape(self, new_shape: Tuple[sint, ...]) -> Optional[View]:
     if self_all_int and not all_int(new_shape):
       if len(self.shape) != len(new_shape): raise ValueError(f"cannot symbolic reshape non-contiguous {self} -> {new_shape}")
       for si, so in zip(self.shape, new_shape):
-        if not isinstance(so, int): so = sym_infer(so, dict([v.unbind() for v in so.vars()]))
+        if not isinstance(so, (int, float)):
+          so = cast(int, sym_infer(so, dict([v.unbind() for v in so.vars()])))
+          if isinstance(so, float): raise ValueError(f"cannot shape to new shape {new_shape=}, contains Variable that infers to float {so}")
```
gordbegli, line 322: Raised error when `isinstance(so,float)` because `so` has type `SupportsIndex` and float doesn't.

geohot:
This still wanted?

gordbegli:
Yes. I'll fix the conflicts.

gordbegli:
Passes CI now. Would appreciate feedback if you're still online @geohot.

test/unit/test_uop_symbolic.py
```diff
@@ -617,6 +617,7 @@ def test_sym_infer(self):
     c = Variable("c", 0, 10)
     var_vals = {a: 2, b: 3, c: 4}
     assert sym_infer(5, var_vals) == 5
+    assert sym_infer(4.2, var_vals) == 4.2
```
geohot, line 620: This should fail the type checker

geohot:
I merged since I know qualcomm uses this, but the type should be rethought here. Maybe the whole way qualcomm does it should be rethought.

PR #7459, opened 2024-11-01
Split SDXL Gen Fix

tobias17:
## Overview

Implements the fix when calling a JIT function 2 separate times before using the result.

This PR relies on the merging of #7154 and will be converted from draft to PR once that is done.

## Testing

Ran SDXL batched inference code and computed the CLIP score (higher is better).

Before
```
clip_score: 0.2597
```

After
```
clip_score: 0.3468
```

### DIFF
--- a/examples/sdxl.py
+++ b/examples/sdxl.py
@@ -303,7 +303,7 @@ class VanillaCFG(Guider):
 
 class SplitVanillaCFG(Guider):
   def __call__(self, denoiser, x:Tensor, s:Tensor, c:Dict, uc:Dict) -> Tensor:
-    x_u = denoiser(x, s, uc)
+    x_u = denoiser(x, s, uc).clone().realize()
     x_c = denoiser(x, s, c)
     x_pred = x_u + self.scale*(x_c - x_u)
     return x_pred


chenyuxyz:
for future reference, can you add the script you used in `Testing` section?

examples/sdxl.py
```diff
@@ -303,7 +303,7 @@ def __call__(self, denoiser, x:Tensor, s:Tensor, c:Dict, uc:Dict) -> Tensor:
 
 class SplitVanillaCFG(Guider):
   def __call__(self, denoiser, x:Tensor, s:Tensor, c:Dict, uc:Dict) -> Tensor:
-    x_u = denoiser(x, s, uc)
+    x_u = denoiser(x, s, uc).clone().realize()
```
chenyuxyz, line 306: do you need the realize or just the clone?

tobias17: you do have to realize, since if you don't then the graph references the original buffer for the JIT and it gets overwritten, the realize is what actually copies the data into a new buffer

tobias17:
> for future reference, can you add the script you used in `Testing` section?

it is a bit of a chonker and was being worked on, but you are right and I'll try and do this moving forward

here is the script as is current for reference: https://github.com/tobias17/tinygrad-ports/blob/a348bda789544041a79332ff0c199a0da28ab8c8/eval_sd/model_eval.py

chenyuxyz:
it's fine to merge this since it's better, but the whole clone realize thing is not great. it feels like assign in disguise and we really want to fix it in a more explicit way

PR #7463, opened 2024-11-01
sink with graph rewrite [pr]

Qazalin:


Qazalin:
makes openpilot benchmark slower, I need to solve this in a different way.

PR #7468, opened 2024-11-01
update shifts spec

ignaciosica:
Update spec to cover llvm and ptx requirements

given `x >> y` or `x << y`:

[LLVM](https://llvm.org/docs/LangRef.html#shl-instruction): `x.dtype == y.dtype`

[PTX](https://docs.nvidia.com/cuda/parallel-thread-execution/#logic-and-shift-instructions-shl): `y.dtype == uint32`

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -788,8 +788,10 @@ spec = PatternMatcher([
   (UPat(UOps.ALU, dtype=dtypes.bool, src=(UPat(name="x"), UPat(name="y")), arg=BinaryOps.CMPLT), lambda x,y: x.dtype == y.dtype),
   (UPat(UOps.ALU, dtype=dtypes.bool, src=(UPat(name="x"), UPat(name="y")), arg=BinaryOps.CMPNE), lambda x,y: x.dtype == y.dtype),
   # and SHL/SHR, the shift distance is an int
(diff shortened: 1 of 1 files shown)


tinygrad/ops.py
```diff
@@ -788,8 +788,8 @@ def graph_rewrite(sink:UOp, pm:PatternMatcher, ctx=None) -> UOp:
   (UPat(UOps.ALU, dtype=dtypes.bool, src=(UPat(name="x"), UPat(name="y")), arg=BinaryOps.CMPLT), lambda x,y: x.dtype == y.dtype),
   (UPat(UOps.ALU, dtype=dtypes.bool, src=(UPat(name="x"), UPat(name="y")), arg=BinaryOps.CMPNE), lambda x,y: x.dtype == y.dtype),
   # and SHL/SHR, the shift distance is an int
-  (UPat(UOps.ALU, src=(UPat(name="x"), UPat()), name="alu", arg=BinaryOps.SHL), lambda alu,x: alu.dtype == x.dtype),
-  (UPat(UOps.ALU, src=(UPat(name="x"), UPat()), name="alu", arg=BinaryOps.SHR), lambda alu,x: alu.dtype == x.dtype),
+  *[(UPat(UOps.ALU, src=(UPat(name="x"), UPat(name="y")), name="alu", arg=shift),
```
chenyuxyz, line 791: i think it's more readable to keep it as two lines (like CMPLT and CMPNE above)

PR #7476, opened 2024-11-02
realize before copy rule [pr]

Qazalin:


Qazalin:
ahh BUFFER_VIEW can't share this
image

Qazalin:
works! This diff needs UPat.any now.

PR #7487, opened 2024-11-02
Transcendental exp with large input

### LINKED ISSUE #7421: TRANSCENDENTAL=2 Tensor([11], dtype='half').exp() returns inf
TRANSCENDENTAL=2 METAL=1 python -c "from tinygrad import Tensor; print(Tensor([11], dtype='half').exp().float().item())"

same inf for METAL/AMD/NV. CLANG and LLVM gives the correct ~59770.

eliotgolding:
Resolves #7421, see comment there.

I was only able to test this on CLANG and CUDA by the way.

### DIFF
--- a/test/test_transcendental.py
+++ b/test/test_transcendental.py
@@ -12,6 +12,15 @@ settings.register_profile("my_profile", max_examples=200, deadline=None, derando
 settings.load_profile("my_profile")
 
 class TestTranscendentalMath(unittest.TestCase):
+  @given(strat.sampled_from([(dtypes.float64, 709.5), (dtypes.float32, 88.7), (dtypes.float16, 11)]))
+  def test_exp_near_inf(self, dtype_x):
+    dtype, x = dtype_x
+    if not is_dtype_supported(dtype, Device.DEFAULT): raise unittest.SkipTest(f"no {dtype.name} on {Device.DEFAULT}")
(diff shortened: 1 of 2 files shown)


test/test_transcendental.py
```diff
@@ -12,6 +12,15 @@
 settings.load_profile("my_profile")
 
 class TestTranscendentalMath(unittest.TestCase):
+  @given(strat.sampled_from([(dtypes.float64, 709.5), (dtypes.float32, 88.7), (dtypes.float16, 11)]))
+  def test_exp_near_inf(self, dtype_x):
+    dtype, x = dtype_x
+    if not is_dtype_supported(dtype, Device.DEFAULT): raise unittest.SkipTest(f"no {dtype.name} on {Device.DEFAULT}")
+    with Context(TRANSCENDENTAL=2):
+      y = Tensor([x], dtype=dtype).exp()
+      if dtype == dtypes.float16: y = y.float()  # can't read out half
```
chenyuxyz, line 21: just compare with numpy output to avoid this special case

chenyuxyz:
see #7507, we ended up moving transcendental to the end of the rewrite and no rewrite to its order. Thanks for root causing this and I used your test cases

eliotgolding:
Ok thank you for the feedback

PR #7497, opened 2024-11-03
move transcendental to last rewrite witout any symbolic

### LINKED ISSUE #7421: TRANSCENDENTAL=2 Tensor([11], dtype='half').exp() returns inf

chenyuxyz:
fix #7421


geohot:
see #7498

chenyuxyz:
#7507

PR #7515, opened 2024-11-03
safe softmax trick in MCTS ucb_explored_children

chenyuxyz:
fixed
```
  File "numpy/random/mtrand.pyx", line 971, in numpy.random.mtrand.RandomState.choice
ValueError: probabilities contain NaN
```
when all ucb_explored_children are big negative numbers result in all NaN probabilities


chenyuxyz:
tested with `PYTHONPATH=. DEBUG=2 BEAM=100 HALF=1 python examples/gpt2.py --count 10 --temperature 0`. fixed the error on mac for me

PR #4357, opened 2024-04-30
fusion tests from test_opt

Qazalin:
`test_llama` remains in external_test_opt because `model.forward` calls realize internally.

PR #7549, opened 2024-11-05
fix test_schedule conv2d bug

geohotstan:
wrote a fix for the pool edge case but hit problems in `test_schedule`

it turns out that the 3 tests in `test_schedule` was doing conv with input shape `[2,2]` and weight shape `[3,3]` and was passing. 
Torch raises error for this `RuntimeError: Calculated padded input size per channel: (2 x 2). Kernel size: (3 x 3). Kernel size can't be greater than actual input size`
so I changed the weight shape, and kernel count increased....
I'm not sure whats suppose to fuse and what's not suppose to fuse.
also not sure if decreasing the weight shape is even the right test...


Qazalin:
hmm so these are pretty old tests originally coming from external_test_opt. #4357
@chenyuxyz should tensor assert this?

don't worry about kernel count, things will fuse differently when your shapes change.

chenyuxyz:
thanks and yea we should check that in tensor. Ideally these should be runtime error but can be a separate cleanup for all

PR #7553, opened 2024-11-05
Refactor tests to not use Ops.ALU

wpmed92:
- Remove op, arg hack from `ops.py`
- Update tests to not use `Ops.ALU`, but the actual op (which used to be the arg)
- regenerate dataset


wpmed92:
I regenerated the dataset with `extra/optimization/generate_dataset.sh`, just saw that the current dataset is much smaller than the previous on master, will look into why.

chenyuxyz:
dataset seems too small
`https://github.com/tinygrad/tinygrad/actions/runs/11685310452/job/32538457601?pr=7553` showed 259 rows only. it should have a few thousands

wpmed92:
@chenyuxyz yeah I'm looking into why it's so small

chenyuxyz:
thanks for the help to improve dataset gen, will merge #7554 if that works.

PR #7562, opened 2024-11-05
[bounty] WIP mmap clang

cold0dev:
Works on M1 mac, but it needs macholib to parse Mach-O.
Linux is quite unstable and requires more work, as the clang produces weird instructions like bl 0 on ARM.

TODO:
- Fix linux
- Write our own Mach-O parser


geohot:
Less clean than #7365, don't import `extra` from core tinygrad

PR #7566, opened 2024-11-06
remove a type: ignore that shouldn't exist [pr]

Qazalin:


chenyuxyz:
there are a few more, these can be found with `warn_unused_ignores` but it seems we cannot enable it by default because linux/osx discrepancy

PR #7581, opened 2024-11-07
fix Tensor._pool edge case

geohotstan:
fixed failure test case from #7483 (did it by handling stride before dilation)

the issue before was when `o*s > i+d` the shrink for strides doesn't work.
I used the same shrinks-shift -> reshape method like before except I anchored it with `o` instead of `k`. 
It was easier for me to reason about using `o` as anchor since each `o` layer maps to an "effective kernel" so we just have to shift until the `input` gets shaved down to the `kernel`.
Only problem is the `repeat` if often much larger then if we used `k`...

time to check process replay!
almost forgot about process replay LOL


chenyuxyz:
don't think process replay shows since these are tensor level change

are you saying convs would get slower in some case? is there a test case that shows the diff?

geohotstan:
> don't think process replay shows since these are tensor level change
> 
> are you saying convs would get slower in some case? is there a test case that shows the diff?

oh..
I was just scared that the larger repeat might have caused slowdowns in either more views or extra python time in simplify views
but it looks like the time is relatively similar with master.

geohotstan:
image
I remember hitting similar issues in #3335 where either tensor core or float4 was selecting an inefficient axis of equivalent shapetrackers because of "same shape different stride" shenanigans

e.g.
image

preeeetty sure this is whats causing the slowdown since after `simplify` the different for large repeat should be reduced.

geohotstan:
@chenyuxyz could you help me run the benchmark again?

I still haven't figured out the exact reason for the slowdown 😔
I just want to see if this hacked version from the original implementation also has slowdown too.

geohotstan:
@chenyuxyz one more time 😅

geohotstan:
ok resnet training speed looks fine now.

The generated `DEBUG=4` code is all the same throughout the slowness so that means the slowness is in python time?
For the previous slowness, the UOp graph for this branch and master had some different multi-view shapetrackers (which actually was equal) so I'm guessing the extra time was spent on it's symbolic pattern matching...?
This is not confirmed but just a guess.

but I think this current fix is fine for now on my end

PR #7584, opened 2024-11-07
simplify valid in real_strides

chenyuxyz:
also valid order matters. it changed because AND is commutative.


chenyuxyz:
pretty big kernel improvements, will do valid order first

chenyuxyz:
will do this in a different pr

PR #7585, opened 2024-11-07
clear disk cache for benchmark machines

chenyuxyz:


chenyuxyz:
it's about 2~5 minutes slower depends on the benchmark. once these are fast enough we should always clear cache

PR #7589, opened 2024-11-08
different valid order

chenyuxyz:
in simplify_valid, we start with valids that are in others' parent so the others is more likely to be simplified


chenyuxyz:
helps conv backward mostly. so beautiful mnist, cifar and resnet training https://wandb.ai/chenyuxyz/tinygrad-examples_mlperf/runs/g3cxd6qr?nw=nwuserchenyuxyz

PR #7593, opened 2024-11-08
openpilot benchmark: add cast from numpy to benchmark

haraschax:
Initializing the Tensors to input into the models takes a considerable amount of time (5ms on latest master?). This makes it much slower than the thneed version of the openpilot runner. The input casts need to be added to the benchmark for the comparison to be fair and this not to regress.


geohot:
@nimlgen is there a better way to do this?

PR #7595, opened 2024-11-08
Remove Ops.ALU

wpmed92:
`Ops.ALU` is not used anymore, which used to be arg, is now op.
There were two occurrences in tests that still referenced `Ops.ALU`, I replaced them with `GroupOp.ALU`


geohot:
Woo thanks!

PR #7601, opened 2024-11-08
better int64 indexing

ttomsa:


test/test_int64_idx.py
```diff
@@ -0,0 +1,11 @@
+import unittest
+from tinygrad import Tensor, dtypes, Device
+
+class TestInt64Indexing(unittest.TestCase):
+  @unittest.skipIf(Device.DEFAULT in ("METAL", "NV"), "memory error in METAL, NV takes too long")
+  def test_int64_indexing(self):
+    t = Tensor.ones((1 << 31) + 4, dtype=dtypes.int8).contiguous()
```
geohot, line 7: This is gonna use tons and tons of ram.

geohot:
There's definitely ways to test this without huge tensors or UOps tests (which are pretty useless with how quickly they go stale, we want to test behavior).

Think about it, why do we want this, then test explicitly for that behavior? It's not for >4GB tensors.

ttomsa:
> There's definitely ways to test this without huge tensors or UOps tests (which are pretty useless with how quickly they go stale, we want to test behavior).
> 
> Think about it, why do we want this, then test explicitly for that behavior? It's not for >4GB tensors.

I took this to mean there was a way to test for int64 at the tensor level without using a ton of memory and I couldn't see how. My bad.

PR #7602, opened 2024-11-08
idx valid simplification in to_indexed_uop

chenyuxyz:


chenyuxyz:
need to flatten AND first

PR #7605, opened 2024-11-08
objdump intel syntax

ttomsa:
more readable and the first operand is the destination like arm64/ptx


geohot:
Needs a test

test/test_objdump.py
```diff
@@ -0,0 +1,17 @@
+from tinygrad import Device
+from tinygrad.helpers import cpu_objdump
```
geohot, line 2: Should be a `disassemble` method on the Compiler, see CUDA/AMD/METAL/QCOM

ttomsa: This is a bit outside the intended scope of this pr. Regardless I decided to just use llvm-objdump as it works on windows and macos.

geohot:
I shouldn't have merged this with the try except. I don't have llvm-objdump on my computer.

geohot:
That test also passes if you don't have llvm-objdump.

ttomsa:
I didn't change the test cause the CI has llvm-objdump. Llvm-objdump comes with llvm not clang. I think it's better cause it works on other platforms. If you're on windows you just need llvm whereas with objdump you need wsl/mingw.

PR #7609, opened 2024-11-09
add Tensor.linspace

geohotstan:
was debating whether or not to add this since there is no actual linspace op in onnx but maybe onnx.AffineGrid can use it?
but then I saw https://x.com/spikedoanz/status/1855017414909739073/photo/1


test/test_ops.py
```diff
@@ -233,6 +233,22 @@ def test_arange_big(self):
   def test_arange_4096(self):
     helper_test_op([], lambda: torch.arange(4096, dtype=torch.int32), lambda: Tensor.arange(4096), forward_only=True)
 
+  def test_linspace(self):
```
geohot, line 236: How many of these tests actually test unique behaviors? More isn't always better

PR #7612, opened 2024-11-09
more general operands for exec_alu

sohaiberrabii:
allow vector operands with any Iterable


geohot:
When do you need this? Also, there's 0 tests

PR #7615, opened 2024-11-09
Do not link to compiler-rt

uuuvn:


uuuvn:
Hm, i'm not sure a stacked pr thing quite worked

geohot:
Is the shared change related?

uuuvn:
> Is the shared change related? Is this minimal?

The alternative to shared change is to copy paste subprocess call from ClangCompiler, i thought the current thing was cleaner.

> This also isn't testing this generically. How do you know that's the only pattern linking to compiler-rt?

I don't think there is a way to test this generically other than just verify for every compiled object file that it doesn't have GOT/PLT references. To do that with current clang compiler you would have to either run subprocess twice for -c and link, or twice for -shared and for -c just to test it or write some hacks to try to search for clues in shared file (like just strings for __trunc, but that's not general as there is way more functions in compiler-rt that might get linked). Anyway we get this in an actually generic way for free with #7365 as it just doesn't implement linking to anything so i think it's fine to test the only known (and fixed) case right now

PR #7625, opened 2024-11-10
swizzle parents with graph rewrite

Qazalin:
I think this is a few UPats away from just relying on merge_views. st_fixup is terrible.


Qazalin:
plan
image

Qazalin:
again graph_rewrite being slower than manual dfs
image

Qazalin:
image

PR #7628, opened 2024-11-10
correct _get_test_global_size for odd globals

chenyuxyz:
always double factor might underestimate for odd global


chenyuxyz:
questionable... beam output seems slightly better without this but probably just all noise

PR #7632, opened 2024-11-11
calling view on graph edges is fine [pr]

Qazalin:
it's kind of like the `gep` simplify stuff.
UOp.view will continue to borrow logic from LazyBuffers's ._view as big graph progresses.


Qazalin:
I think UOp will eventually have self.base too.

PR #7633, opened 2024-11-11
swizzle tc [pr]

ignaciosica:
Draft Swizzle TC. 
For validation only, I'll split the refactor into smaller, separate PRs once it's ready and approved.

Ideally, I'd prefer to implement this entirely as a graph_rewrite, as that's the general direction we're moving towards. However, when I tried this approach, it didn't align well with the specifics of the shapetracker fix. The recursive nature of the pattern-matching rules contrasts with the one-time fixes each buffer operation requires in this context. As a result, the rules ended up being 80% checks to ensure shapetracker wasn’t already fixed, and only 20% actual modifications to it. The same issue applies to reduce_axis, at least with tensor cores, where only one rewrite is needed.

I'm yet to refine the rules though.

This might have a significant overlap with qazalin work also


ignaciosica:
``` python
(UPat({*GroupOp.ALU,Ops.CAST,Ops.BITCAST,Ops.ASSIGN}, name="e").view(name="v"), 
 lambda e,v: e.replace(src=tuple(s.view(v.st) for s in e.src)))

(UPat(GroupOp.Buffer, name="b").view(name="v"), 
 lambda b,v: b.replace(src=tuple((v.arg).to_uop() if s.op is Ops.VIEW else s for s in b.src)))]))
```

Both rules are adapted from `view_left` in the schedule. The first rule, which propagates views, is identical to the one in the schedule. The second rule, however, differs slightly: it replaces the view in the buffers instead of appending the view argument.

I would have preferred to use `apply_swizzle` from the schedule, but here it’s necessary to replace the view for a proper fix.

ignaciosica:
I think all the remaining changes are quite tightly coupled to be moved into separate prs. Though I can try if preferred.

ignaciosica:
NV job failed due to timeout, can't rerun job

tinygrad/codegen/kernel.py
```diff
@@ -607,97 +607,82 @@ def name(self) -> str:
...
-        arg = KernelInfo(self.local_dims, self.upcasted, self.dont_use_locals)
-      return op.replace(src=tuple(fixup_ast(x, apply_to_st) for x in op.src), arg=arg)
-    # NOTE: rewrite with an empty PatternMatcher to dedup UOps
-    return graph_rewrite(fixup_ast(self.ast), PatternMatcher([]))
+
+      return ret
+
+    return graph_rewrite(fixup_ast(self.ast), PatternMatcher([
+      (UPat({*GroupOp.ALU,Ops.CAST,Ops.BITCAST,Ops.ASSIGN}, name="e").view(name="v"), lambda e,v: e.replace(src=tuple(s.view(v.st) for s in e.src))),
+      (UPat(GroupOp.Buffer, name="b").view(name="v"), lambda b,v: b.replace(src=tuple((v.arg).to_uop() if s.op is Ops.VIEW else s for s in b.src)))]))
```
Qazalin, line 685: hmm it's not merging? also careful with STORE shapetracker.

Qazalin: this is also my fault that pattern should only really apply to LOAD/VALID ops.

Qazalin:
triggered a retry for NV. I'd look into why merging is wrong.
The swizzle spec is that LOAD -> VIEW becomes LOAD (v2+v1). If you can't find why it's not working I can take a look. 

good work on lines!

geohot:
Triggered a retry? It looks like it's passing process replay to me

ignaciosica:
> Ready to merge? 
Just now taking a look at qazalin feedback

tinygrad/codegen/kernel.py
```diff
@@ -607,97 +607,82 @@ def name(self) -> str:
...
-      if op.op in GroupOp.Buffer:
-        # for locals, we use the ShapeTracker that's in the srcs
-        st = op.st_arg if op.src[0].op is Ops.DEFINE_LOCAL else self.sts[self.bufs.index(op)]
-        st_uop = (st if apply_to_st is None else apply_to_st(st)).to_uop()
-        if op.op is Ops.VALID: return op.replace(src=(st_uop,))
-        if op.op is Ops.STORE: return op.replace(src=(op.src[0], st_uop, fixup_ast(op.src[2], apply_to_st)))
-        return op.replace(src=(op.src[0], st_uop, *[fixup_ast(x, apply_to_st) for x in op.src[2:]]))
+    def fixup_ast(op:UOp) -> UOp:
+      ret = op.replace(src=tuple(fixup_ast(x) for x in op.src))
+      if op.op in GroupOp.Buffer and op in self.bufs: return ret.view(self.sts[self.bufs.index(op)])
```
Qazalin, line 613: what's "and op in self.bufs"

tinygrad/codegen/kernel.py
```diff
@@ -607,97 +607,82 @@ def name(self) -> str:
...
+          srcs = list((ret.src[0] if ret.src[0].op is not Ops.CAST else ret.src[0].src[0]).src)
+          for i, tc_pattern in enumerate([tc.st1_pattern, tc.st2_pattern]):
+            if tc_pattern: srcs[i] = srcs[i].view(fix_st(srcs[i].st, *tc_pattern))
+
+            if self.use_tensor_cores == 3:  # for TC=3, emulate the warp addressing with locals
+              local_shape = tuple(1 if i >= self.first_reduce and i < self.first_upcast else s for i, s in enumerate(self.full_shape))
+              st_uop = ShapeTracker.from_shape(local_shape).to_uop()
+              local_buffer = UOp(Ops.DEFINE_LOCAL, tc.dtype_in.ptr(local=True), (), (f"temp{i + 1}", st_uop.arg.real_size()))
+              local_store = UOp.store(local_buffer, st_uop, srcs[i])
+              if tc_pattern: local_store = local_store.view(fix_st(local_store.st, *tc_pattern))
```
Qazalin, line 646: ohh you're viewing a STORE, you should probably split the STORE upat from LOAD.

Qazalin: feel free to modify view_left. but you understand why VIEW of STORE in schedule is invalid right? STORE is the kernel boundary so it's always base.

tinygrad/codegen/kernel.py
ignaciosica: that's for bufs who's st have been modified within kernel (through opt opts). Before it checked the source was not DEFINE_LOCAL but found it cleaner this way.

Qazalin: can you narrow it down to that specific local case? this can hide a legit bug.

ignaciosica: before: `st = op.st_arg if op.src[0].op is Ops.DEFINE_LOCAL else self.sts[self.bufs.index(op)]`
after: 
if the buf is in op then there is a st associated to it on self.sts

ignaciosica: > can you narrow it down to that specific local case? this can hide a legit bug.

For DEFINE_LOCAL no fixing the st is needed. There is no special case. At least in this stage of st fixing. Maybe I'm missing something though

tinygrad/codegen/kernel.py
Qazalin: in this case it does you're actually swizzling the store's st

Qazalin:
meh, I see why it's not easy to make fixup_ast transforms use merge. You're gonna need a replace.

tinygrad/codegen/kernel.py
ignaciosica: Now I'm confused haha

ignaciosica:
> Like all those manual .simplifys in fix_st can go away too.

I tried removing those earlier but process replay has diff if I removed the last one

tinygrad/codegen/kernel.py
ignaciosica: I guess it's not much of a swizzle for the stores but a shapetracker replacement. The swizzle is only for the pattern fix for the tensor cores. I think I ended up messing the terms and believing the same thing should apply for both cases when in reality it should (tc pattern fixing and buffers shapetracker replacement).

ignaciosica:
> It'll be weird that load.view() in schedule means v2+v1 and in kernel it means replace(v1, v2)

Totally agree.

Still looking into it. Now it "should" work as the swizzle is only for the tc_pattern fixup but when lowering the LOAD it produces the wrong indexed_uops. I don't understand yet the to_indexed_uops code nor at the st level or view so it might take me a while to figure out whats happening.

geohot:
@Qazalin I'll leave this to you to merge when you think it's ready. Remember, done is better than perfect

Qazalin:
> I don't understand yet the to_indexed_uops code so it might take me a while to figure out whats happening.

@ignaciosica I think it's worth taking some time on how this can be done with merge_views.

Fundamentally any shapetracker transformation _is_ a merge view. It's just that kernel is doing these way too early, the st you're swizzling is already v2+v1.

ignaciosica:
Yes, I'm working on it. I think the problem might be that the patterns weren't thought as a swizzle but as a st replacement so the swizzling with them is wrong. I'm trying to fix the metal tc just by modifying the pattern to try to prove this point.

Qazalin:
should this diff be converted to draft?

geohot:
Good to merge if tests pass

ignaciosica:
Okay, I'll continue working on actually swizzling the shapetrackers.

PR #7634, opened 2024-11-11
cache tc early_upcast_axes [pr]

ignaciosica:
With the new swizzle tc, this property might be called up to 5 times per kernel with tc. It's not much but it might as well be cached


ignaciosica:
Don't think so really, sorry I didn't test it. Will test now

Qazalin:
status?

ignaciosica:
On the first tests I made I couldn't get any measurable difference. Today I'll round up the tests.

ignaciosica:
Closing, no measurable speed up. Sorry for opening without testing for speed improvement first.

PR #7635, opened 2024-11-11
[pr] Have PTX share code with LLVM

mesozoic-egg:
Converted the rendering to use pattern matcher. Currently I'm just checking if the two versions output the same code for common use cases in the test_ptx.py script


geohot:
Bounty locked, good luck!

geohot:
You can use [pr] to check if it's identical, if that's what you are targeting (not required, but free test)

geohot:
I'm going to merge this since it's cleaner and passes process replay.

But I think for the bounty there's at least 20 lines to lose here that will change the output, but not affect functionality. Ex: why is DEFINE_ACC 13 lines, why is bool casting 4 lines?

PR #7644, opened 2024-11-11
UNet: cast to half only if supported

wpmed92:
These half casts in UNet should only happen if they are supported. This is a blocker for Stable Diffusion WebGPU. `f16` support in `wgpu` is [coming together](https://github.com/gfx-rs/wgpu/pull/5701), once it's in, we can do f16 SD.


extra/models/unet.py
```diff
@@ -222,16 +223,17 @@ def get_d_and_n_heads(dims:int) -> Tuple[int,int]:
     ]
 
   def __call__(self, x:Tensor, tms:Tensor, ctx:Tensor, y:Optional[Tensor]=None) -> Tensor:
-    t_emb = timestep_embedding(tms, self.model_ch).cast(dtypes.float16)
```
wpmed92, line 225: This was a redunant cast.

PR #7650, opened 2024-11-12
Fix tqdm MLPerf dataloader script

francislata:
# Overview
This fixes an issue where a tqdm context manager style isn't supported when this file got migrated over to use tinygrad's `tqdm` and throws the following error:
Screenshot 2024-11-12 at 05 50 14

For now, this PR uses the iterative approach to update the progress bar.

Here is the result for UNet3D:
Screenshot 2024-11-12 at 05 54 29

And here is the result for ResNet:
Screenshot 2024-11-12 at 05 58 45

# Notes
- I've also updated the `total` to be divided by the default batch sizes of these dataloaders so that it shows the proper progress bar length.

### DIFF
--- a/examples/mlperf/dataloader.py
+++ b/examples/mlperf/dataloader.py
@@ -363,16 +363,12 @@ if __name__ == "__main__":
     files = get_val_files() if val else get_train_files()
 
     if not preprocessed_dir.exists(): preprocess_dataset(files, preprocessed_dir, val)
-    with tqdm(total=len(files)) as pbar:
-      for x, _, _ in batch_load_unet3d(preprocessed_dir, val=val):
-        pbar.update(x.shape[0])
+    for x, _, _ in tqdm(batch_load_unet3d(preprocessed_dir, val=val), total=len(files) // 6): pass
 
   def load_resnet(val):
     from extra.datasets.imagenet import get_train_files, get_val_files
     files = get_val_files() if val else get_train_files()
-    with tqdm(total=len(files)) as pbar:
-      for x,y,c in batch_load_resnet(val=val):
-        pbar.update(x.shape[0])
+    for x, _, _ in tqdm(batch_load_resnet(val=val), total=len(files) // 64): pass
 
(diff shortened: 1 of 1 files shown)


examples/mlperf/dataloader.py
```diff
@@ -363,16 +363,12 @@ def load_unet3d(val):
     files = get_val_files() if val else get_train_files()
 
     if not preprocessed_dir.exists(): preprocess_dataset(files, preprocessed_dir, val)
-    with tqdm(total=len(files)) as pbar:
-      for x, _, _ in batch_load_unet3d(preprocessed_dir, val=val):
-        pbar.update(x.shape[0])
+    for x, _, _ in tqdm(batch_load_unet3d(preprocessed_dir, val=val), total=len(files) // 6): pass
```
geohot, line 366: `    for x, _, _ in (pbar:=tqdm(batch_load_unet3d(preprocessed_dir, val=val), total=len(files))): pbar.update(x.shape[0])`

geohot:
Was anything wrong with the code I proposed. `tqdm(range(` is frowned upon

francislata:
> Was anything wrong with the code I proposed? `tqdm(range(` is frowned upon

@geohot - i tried your changes and it's actually exceeding the range of the progress bar since the `__iter__` implementation inside `tqdm` updates the progress bar by 1 unit already because it's used inside of the `for` loop. so the call to `pbar.update(x.shape[0])` is an extra update.

> `tqdm(range(` is frowned upon
i can switch to `trange` instead but do let me know if i missed something here from my reply above.

geohot:
Oh yea, that's not great. I mean you could subtract one...

What would it take to make our tqdm support this style and match real tqdm?

francislata:
Looking at the current code, probably a few lines to support that context manager version to match real tqdm. I think that one can be a separate PR and this one can be closed since it should just work as is.

francislata:
Closing this since #7770 adds context manager support to `tqdm`.

PR #7670, opened 2024-11-13
fix: instead of negation trick, flip max flip

rezaarezvan:
instead of negating the max, we can first flip all the values take the max, and then flip again.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -1578,7 +1578,7 @@ class Tensor(SimpleMathTrait):  # pylint: disable=abstract-method
     print(t.min(axis=1, keepdim=True).numpy())
     ```
     """
-    return self.max() - (self.max() - self).max(axis=axis, keepdim=keepdim)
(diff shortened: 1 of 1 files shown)


rezaarezvan:
put this as draft since i need to test it (using CI), my machine fails on some unrelated metal stuff

tinygrad/tensor.py
```diff
@@ -1578,7 +1578,7 @@ def min(self, axis:Optional[Union[int, Sequence[int]]]=None, keepdim=False):
     print(t.min(axis=1, keepdim=True).numpy())
     ```
     """
-    return -((-self).max(axis=axis, keepdim=keepdim))
+    return self.max(axis=axis, keepdim=keepdim) - (self.max(axis=axis, keepdim=keepdim) - self).max(axis=axis, keepdim=keepdim)
```
geohot, line 1581: min def should not require two maxes

rezaarezvan:
closing for #7675

PR #7672, opened 2024-11-13
refactor: modernize typing

roelofvandijk:
Now we are on 3.10, we can us the builtins as types.

Some files as example. I can do the rest, but wanted to know if this is a desired change first.

Will ultimately save 10-20 lines or so because some of the typing imports can go.

Can easily be done using [`pyupgrade`](https://github.com/asottile/pyupgrade)
`pyupgrade --py310-plus tinygrad/function.py` for per-file or as pre-commit hook. 
However, `pyupgrade` applies some other syntax changes (which I have removed here).

### DIFF
--- a/tinygrad/function.py
+++ b/tinygrad/function.py
@@ -1,6 +1,5 @@
 """This is where the forwards and backwards passes live."""
 import math
-from typing import Tuple, Optional
 from tinygrad.helpers import argsort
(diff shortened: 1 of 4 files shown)


geohot:
I'll have to think more about this. Is this a subset of this that's a clear readability win? I don't care about line count

chenyuxyz:
mostly in master now

PR #7674, opened 2024-11-13
refactor: double if check

roelofvandijk:
`is_load` was checked twice, this seems simpler.


geohot:
Things like this need "[pr]" in the title to confirm process replay runs and it produces the same outputs

roelofvandijk:
pr passed

PR #7675, opened 2024-11-13
fix: Tensor min function for unsigned ints

bjsi:
Uses the flip idea from discord to fix the min function for unsigned ints


Qazalin:
AMD should be green now. [The CI emulator needed one new instruction.](https://github.com/Qazalin/remu/commit/b873f2d254ebfdc81fc113e311c80efaedf061f8)

chenyuxyz:
it can be done with bitwise not and fix both uint and int, still this is better than master

PR #7677, opened 2024-11-13
combine pad2d with pad

geohotstan:
did it in a way where we still have backward compatibility with previous `pad`
it's a little bit confusing though.

we could just have flat padding, but there are quite a few places using original pad with `None` as dimension.


geohot:
Nice! Yea, we can eventually unify on a syntax, but i still prefer this to the terrible `pad2d` function which wasnt't 2d at all

PR #7679, opened 2024-11-13
assert op is not store in view

ignaciosica:
- assert op is not store in view
- update view spec


ignaciosica:
Qazalin, are the failing test due to the thing you mentioned about image being a little hacky yet?

Qazalin:
yea `lazy` eventually folds that STORE - but it's more correct to fold it early. good catch!

PR #7683, opened 2024-11-13
add Tensor.nll_loss

geohotstan:


chenyuxyz:
i think this can be simpler, will change it later

PR #7687, opened 2024-11-13
Add render_cast

wpmed92:
This PR is related to [WebGPU](https://github.com/tinygrad/tinygrad/pull/7063) simplification. It adds `render_cast` to be used in string rewrites that use cast (like `CONST` and `CAST`). Adds +1 line, but removes [4](https://github.com/tinygrad/tinygrad/blob/60e7ac6685916b86cebd837d1ab6b98fc2123703/tinygrad/renderer/wgsl.py#L58) from WebGPU, where the +/-inf, nan, and cast string rewrites have to be duplicated just because of different cast style. This approach makes it possible to not duplicate.

### DIFF
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -19,25 +19,25 @@ base_rewrite = PatternMatcher([
   (UPat(Ops.VECTORIZE, name="x"),
    lambda ctx,x: f"{ctx.float4.replace('float4', ctx.render_dtype(x.dtype))}" + \
     (f"{{{','.join([ctx[y] for y in x.src])}}}" if ctx.device == "CLANG" else f"({','.join([ctx[y] for y in x.src])})")),
-  (UPat(Ops.CAST, name="x"), lambda ctx,x: f"({ctx.render_dtype(x.dtype)})({ctx[x.src[0]]})"),
+  (UPat(Ops.CAST, name="x"), lambda ctx,x: f"({ctx.render_cast(x.dtype, ctx[x.src[0]])})"),
   (UPat(Ops.BITCAST, name="x"), lambda ctx,x: f"(*(({ctx.buffer_prefix}{ctx.render_dtype(x.dtype)}*)&{ctx[x.src[0]]}))"),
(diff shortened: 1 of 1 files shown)


geohot:
Does this pass process replay? (add [pr] to title if nothing should change and this is only a refactor)

tinygrad/renderer/cstyle.py
```diff
@@ -19,25 +19,25 @@
...
   (UPat(Ops.BITCAST, name="x"), lambda ctx,x: f"(*(({ctx.buffer_prefix}{ctx.render_dtype(x.dtype)}*)&{ctx[x.src[0]]}))"),
   (UPat(Ops.DEFINE_LOCAL, name="x"), lambda ctx,x: f"{ctx.smem_align}{ctx.smem_prefix}{ctx.render_dtype(x.dtype.base)} {ctx[x]}[{x.arg[1]}];"),
   (UPat(Ops.BARRIER), lambda ctx: ctx.barrier),
   (UPat(Ops.NOOP, name="x"), lambda ctx,x: ctx[x.src[0]]),
   (UPat(Ops.SPECIAL, name="x"), lambda ctx,x: f"{ctx.code_for_workitem[x.arg[0][0]](x.arg[0][-1])}; /* {x.arg[1]} */"),
   # const
-  (UPat(Ops.CONST, arg=math.inf, name="x"), lambda ctx, x: f"(({ctx.render_dtype(x.dtype)}){ctx.infinity})"),
-  (UPat(Ops.CONST, arg=-math.inf, name="x"), lambda ctx, x: f"(({ctx.render_dtype(x.dtype)})-{ctx.infinity})"),
-  (UPat(Ops.CONST, dtype=dtypes.floats, name="x"), lambda ctx,x: f"(({ctx.render_dtype(x.dtype)}){ctx.nan})" if math.isnan(x.arg) else None),
+  (UPat(Ops.CONST, arg=math.inf, name="x"), lambda ctx, x: f"({ctx.render_cast(x.dtype, ctx.infinity)})"),
```
geohot, line 29: It doesn't pass because of needless parens here

wpmed92: The pr doesn’t pass because how consts were casted and how Ops.CAST was rendered were different, now they’re the same, since both use render_cast

geohot:
Fine, merged

wpmed92:
@geohot Yeah for the consts they are fine but for Ops.CAST rendering you are right that the extra parens is not needed, it wasn't there before. I'll put up another PR that removes just that

PR #7694, opened 2024-11-14
simplify bitwise_not

mszep:
Trying to understand the recently added `bitwise_not`, I found that this simpler version also passes the tests.

However I don't know if there is another reason to implement by explictly bit-shifting a dtype-dependent number of times.


chenyuxyz:
i prefer to explicitly normalize input early because otherwise it's hard to tell which layer does the int wrapping, and it's much harder to match arbitrary modulus if we want to rewrite and render it back to `~`

PR #7709, opened 2024-11-15
Buffer transfer on CLOUD devices

mdaiter:
From some feedback, re-opening this PR.

My last PR focused on reading data synchronously out of a web buffer, onto a host device, and then dumping it in another web buffer.

Seems silly in hindsight.

This one lets users input a destination and source buffer, a size, and a source and destination GPU. I wasn't too sure if I should just append a number and keep the buffers on the same device.

Should `cloudDevice.allocator.transfer(2, 1, 10, 'METAL', 'METAL')` be `cloudDevice.allocator.transfer(2, 1, 10, 'METAL:0', 'METAL:1')`?

Also, where should I put a test?


mdaiter:
Sorry, just went off of the suggestion from Discord. My bad

PR #7714, opened 2024-11-15
add Tensor.meshgrid

geohotstan:
I implemented `Tensor.meshgrid`, `Tensor.affine_grid` and `Tensor.grid_sample`
do we want to support these 3 in Tensor? They seem quite obscure.....

There isn't a `meshgrid` onnx OP but this can be used in `affine_grid` onnx OP to make it pretty.
I don't know how often these 3 ops are used but `affine_grid` is already tested in onnx and `grid_sample` tests are currently excluded but I implemented `grid_sample` (only 2D tho, and kinda hacked) since the pytorch docs says `affine_grid` and `grid_sample` are usually used together.

here are the implementations of the other two ops just for an idea of line-count

```
  def affine_grid(self, size:Tuple[int, ...], align_corners:bool=False):
    """
    Generates a 2D or 3D sampling grid.

    The input tensor is a batch of affine matrices (theta) that can have shape (N, 2, 3) for 2D or (N, 3, 4) for 3D.
    Input `size` is the output image size (N, C, H, W) for 2D or (N, C, D, H, W) for 3D.

    This function is often used in conjunction with `grid_sample` to build Spatial Transformer Networks.

    ```python exec="true" source="above" session="tensor" result="python"
    theta = Tensor([[2, 0, 0.5], [0, 1, 0.3]]).expand(1, 2, 3)
    size = (1, 1, 3, 3)
    print(grid.numpy())
    ```
    ```python exec="true" source="above" session="tensor" result="python"
    grid = theta.affine_grid(size=size)
    print(grid.numpy())
    ```
    """
    N, _, *spatial_dims = size
    def _generate_grid(steps):
      return Tensor.linspace(-1, 1, steps, device=self.device) if align_corners else Tensor.linspace(-1+1/steps, 1-1/steps, steps, device=self.device)
    grids = Tensor.meshgrid(*(_generate_grid(d) for d in spatial_dims))
    base_grid = Tensor.stack(*reversed(grids), Tensor.ones_like(grids[0], device=self.device), dim=-1)
    base_grid = base_grid.reshape(1, prod(spatial_dims), len(grids)+1).expand(N, -1, -1)
    return (base_grid @ self.transpose(1, 2)).reshape(N, *spatial_dims, -1)

  def grid_sample(self, grid:Tensor, align_corners:bool=False, mode:Literal["linear"] | Literal["nearest"]="linear"):
    """
    Compute grid sample.
    Only supports 2D spatial dimensions.
    
    The grid specifies the sampling pixel locations normalized to [-1, 1].
    The interpolation algorithm is selected with `mode` which currently only supports `linear` and `nearest`.

    This function is often used in conjunction with `affine_grid` to build Spatial Transformer Networks.

    ```python exec="true" source="above" session="tensor" result="python"
    t = Tensor.arange(9).expand(1, 1, 3, 3)
    grid = Tensor([[[[-1, -1], [1, -1]], [[-0.5, 0], [1, 0.5]]]])
    print(t.numpy())
    ```
    ```python exec="true" source="above" session="tensor" result="python"
    print(t.grid_sample(grid, align_corners=True, mode="linear").numpy())
    ```
    ```python exec="true" source="above" session="tensor" result="python"
    print(t.grid_sample(grid, align_corners=False, mode="nearest").numpy())
    ```
    """
    flat_spatial, (N, C, H, W), (H_out, W_out) = self.flatten(2), self.shape, grid.shape[1:-1]
    # Scale grid coordinates from [-1, 1] to [0, dim-1] if align corners else from [-1, 1] to [-0.5, dim-0.5]
    x,y = ((grid[..., i] + 1) * (d - 1) / 2  if align_corners else ((grid[..., i] + 1) * d) / 2 - 0.5 for i,d in enumerate((W,H)))
    if mode == "nearest": return flat_spatial.gather(2, (y.round() * W + x.round()).reshape(N, 1, -1).expand(-1, C, -1)).reshape(N, C, H_out, W_out)
    # linear
    (xs, ys), ws = tuple((c.floor(), c.ceil()) for c in (x,y)), tuple((c - c.floor()).unsqueeze(1) for c in (x,y))
    idxs = ((y*W + x).reshape(N, -1) for y in ys for x in xs)
    pixel_values = tuple(flat_spatial.gather(2, i.unsqueeze(1).expand(-1, C, -1)).reshape(N, C, H_out, W_out) for i in idxs)
    # apply mask to corners
    if not align_corners: pixel_values = tuple(p*m for p,m in zip(pixel_values, ((x>=0) * (x<=W-1) * (y>=0) * (y<=H-1) for y in ys for x in xs)))
    for w in ws: pixel_values = tuple(pair[0].lerp(pair[1], w) for pair in list(zip(pixel_values[::2], pixel_values[1::2])))
    return pixel_values[0]
```


tinygrad/tensor.py
```diff
@@ -1343,6 +1343,28 @@ def chunk(self, chunks:int, dim:int=0) -> List[Tensor]:
...
+    """
+    Generates coordinate matrices from coordinate vectors.
+
+    When `indexing` is `"ij"` the dimensions are in the same order as the cardinality of the inputs.
+    When `indexing` is `"xy"` the cardinality of first and second dimensions are swapped.
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    x, y = Tensor([1, 2, 3]), Tensor([4, 5, 6])
+    grid_x, grid_y = x.meshgrid(y)
+    print("\\n".join([repr(x.numpy()) for x in (grid_x, grid_y)]))
```
chenyuxyz, line 1356: just print in two lines, that's what the doc reader cares about and str join repr numpy is confusing

chenyuxyz (commented):
numpy has it too (so jax has it) https://numpy.org/doc/stable/reference/generated/numpy.meshgrid.html

test/test_ops.py
```diff
@@ -216,6 +216,21 @@ def test_chunk(self):
     for i in range(len(tor)):
       helper_test_op([], lambda: tor[i], lambda: ten[i], forward_only=True)
 
+  def test_meshgrid(self):
```
chenyuxyz, line 219: also test more than 2-D

PR #7730, opened 2024-11-16
remove add in pad when padding with non-zero

chenyuxyz:
this is simpler because shapetracker is shared

```
from tinygrad import Tensor
Tensor([1, 2, 3]).pad(((1, 1),), 4).realize()
```
with NOOPT=1

this
```
  int gidx0 = gid.x; /* 5 */
  bool alu0 = ((gidx0<4)&((gidx0<1)!=1));
  int val0 = (alu0?*(data1+(gidx0+-1)):0);
  *(data0+gidx0) = ((alu0?1:0)?val0:4);
```

master
```
  int gidx0 = gid.x; /* 5 */
  bool alu0 = ((gidx0<4)&((gidx0<1)!=1));
  int val0 = (alu0?*(data1+(gidx0+-1)):0);
  *(data0+gidx0) = (val0+((alu0?1:0)?0:4));
```

will find a way to rewrite with uop simplification, this should be
```
  *(data0+gidx0) = (alu0?*(data1+(gidx0+-1)):4);
```

### DIFF
--- a/test/external/external_test_opt.py
+++ b/test/external/external_test_opt.py
@@ -68,7 +68,7 @@ class TestInferenceMinKernels(unittest.TestCase):
     model = ResNet18()
     for p in get_parameters(model): p.assign(np.zeros(p.shape, dtype=_to_np_dtype(p.dtype)))
     img = Tensor.randn(1, 3, 224, 224)
-    with CLCache(23):
+    with CLCache(24):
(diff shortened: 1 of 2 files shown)


chenyuxyz:
`cond?1:0` removed in #7733. i think this might work with cat too

PR #7737, opened 2024-11-16
add Tensor.scatter

geohotstan:
saw someone mention `scatter` in discord
I made it almost work like a year ago but got stuck on overlapping indices
the setitem idea is pretty neat!!!!!

Just very satisfying to have this implemented once and for all. :D
I'll clean this up when I wake up tomorrow morning if we want to support this.
tf and jax's scatter is a lil different from torch `scatter` but still very similar. This just matches torch.


test/external/external_test_onnx_backend.py
```diff
@@ -90,7 +90,6 @@ def supports_device(cls, device: str) -> bool:
 
 # no scatternd gathernd
 backend_test.exclude('test_gathernd_*')
```
geohotstan, line 92: there are scatter tests in onnx but voodoo magic is making those tests disappear
I checked that scatter tests are indeed in `test_cases` for 
`globals().update(backend_test.enable_report().test_cases)` (this is after all the excludes)
but somehow it's just not being tested...

gonna investigate this further.

geohotstan:
@ttomsa btw I think you were right about the error being overlapping 0's
After I copied your new implementation, the error went away 😅

tinygrad/tensor.py
```diff
@@ -2326,6 +2326,35 @@ def interpolate(self, size:Tuple[int, ...], mode:str="linear", align_corners:boo
...
+    print(t.scatter(dim=1, index=Tensor([[0, 0], [1, 0]]), src=9).numpy())
+    ```
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(t.scatter(dim=1, index=Tensor([[0, 0], [1, 0]]), src=Tensor([[3, 3], [9, 9]]), reduce="add").numpy())
+    ```
+    """
+    index, dim  = index.to(self.device), self._resolve_dim(dim)
+    src = src.cast(self.dtype) if isinstance(src, Tensor) else Tensor(src, device=self.device, dtype=self.dtype)._broadcast_to(index.shape)
+    assert index.ndim == self.ndim == src.ndim, f"self.ndim, index.ndim and src.dim must all equal, {self.ndim=} {index.ndim=} {src.ndim=}"
+    assert all((True if d == dim else se >= ind) and sr >= ind for d,(se,ind,sr) in enumerate(zip(self.shape, index.shape, src.shape))), \
```
chenyuxyz, line 2348: just `(d == dim or se >= ind)`?

chenyuxyz (commented):
is it possible to share some logic with setitem?

tinygrad/tensor.py
```diff
@@ -97,6 +97,15 @@ def _pad_left(*shapes:Tuple[sint, ...]) -> Tuple[Tuple[sint, ...], ...]:
 def _broadcast_shape(*shapes:Tuple[sint, ...]) -> Tuple[sint, ...]:
   return tuple(0 if 0 in nth_dim_sizes else smax(nth_dim_sizes) for nth_dim_sizes in zip(*_pad_left(*shapes)))
 
+def _masked_setitem(target:Tensor, values:Tensor, mask:Tensor, axes:Tuple[int, ...]):
```
geohotstan, line 100: not really sure what to name this function lol

PR #7738, opened 2024-11-16
Fix None arg and Tuple arg comparison for vectorize

cesar-guibo:
In tinygrad/ops.py, the following pattern is currently making the comparison between tuplized operands for commutative ops.
```
symbolic = symbolic_simple+PatternMatcher([
  # ** COMMUTATIVE flipping **
  (UPat(GroupOp.Commutative, name='x'), lambda x: x.replace(src=x.src[::-1]) if x.src[1].tuplize < x.src[0].tuplize else None),
  # group like
```
This is currently creating an error in the cases where one operand is a Ops.VECTORIZE with a None arg and the other is a Ops.VECTORIZE with a Tuple arg. So, I created a method to standardize the creation of vectorize UOps with a method that sets the default arg to an empty tuple


geohot:
Why does VECTORIZE have a tuple arg?

PR #7744, opened 2024-11-17
failing tests for tensor modulo

eitanturok:
There is an error when computing the modulo of one tensor with another tensor. I added some tests to track this failure, though it looks like we didn't have any tests checking that Tensor.mod works in the first place.

For example, the code
```py
from tinygrad import Tensor
Tensor([6]) % Tensor([8])
```
returns the error
```sh
TypeError: unsupported operand type(s) for %: 'Tensor' and 'Tensor'
```

In [ops.py](https://github.com/tinygrad/tinygrad/blob/eeb222f98b32db1fbba93d7ec4d3b43298d20bf1/tinygrad/ops.py#L84), a comment says that mod is "not in tensor". I was wondering why mod has not been moved to tensor. I searched previous issues/PRs but couldn't find anything.

Cheers!


chenyuxyz:
i think test for new feature should be added together when the feature is added, and the current error message is clear that mod is not supported

this is different from a supported feature that has bug behavior which merging a failing test is valuable

Issue #7699, opened 2024-11-14
alu of tensors on different devices give incomprehensible errors


chenyuxyz:
```
from tinygrad import Tensor, Device
a = Tensor([0, 1, 2, 3], device=f"{Device.DEFAULT}")
b = Tensor([0, 1, 2, 3], device=f"{Device.DEFAULT}:1")
print(f"{(a+b).tolist()}")
```

```
Traceback (most recent call last):
  File "/Users/chenyu/code/tinygrad/test.py", line 5, in <module>
    print(f"{(a+b).tolist()}")
             ^^^^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/tensor.py", line 3663, in _wrapper
    ret = fn(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/tensor.py", line 306, in tolist
    return self.data().tolist()
           ^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/tensor.py", line 3638, in _wrapper
    if _METADATA.get() is not None: return fn(*args, **kwargs)
                                           ^^^^^^^^^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/tensor.py", line 280, in data
    return self._data().cast(self.dtype.fmt, self.shape)
           ^^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/tensor.py", line 3638, in _wrapper
    if _METADATA.get() is not None: return fn(*args, **kwargs)
                                           ^^^^^^^^^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/tensor.py", line 264, in _data
    cpu = self.cast(self.dtype.base).contiguous().to("CLANG").realize()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/tensor.py", line 3638, in _wrapper
    if _METADATA.get() is not None: return fn(*args, **kwargs)
                                           ^^^^^^^^^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/tensor.py", line 223, in realize
    run_schedule(*self.schedule_with_vars(*lst), do_update_stats=do_update_stats)
[...]

PR #7756, opened 2024-11-17
Added error message for when buffers aren’t on same device

Leorasz:
I added an error message to make it easier to debug, should solve #7699


chenyuxyz:
i don't think this helps the end library user

PR #7759, opened 2024-11-18
Add xavier alias for glorot

mdaiter:
Just ran across this while converting some PyTorch.

Seems like PyTorch centered around Xavier for naming, while TensorFlow centered around Glorot.

Just an alias contribution!


geohot:
No test, and there's no `xavier_uniform` in either torch or TF that I can find

mdaiter:
https://pytorch.org/docs/stable/_modules/torch/nn/init.html#xavier_uniform_
^torch

Do aliases need tests?

PR #7761, opened 2024-11-18
feat: implement ADOPT optimizer

wilhelmagren:
Recently published paper fixing convergence issues with Adam and guarantees optimal convergence rate.

"ADOPT addresses the non-convergence issue of Adam by removing the current gradient from the second moment estimate and changing the order of the momentum update and the normalization by the second moment estimate." [1]

[1] Shohei Taniguchi et al., "ADOPT: Modified Adam Can Converge with Any β2 with the Optimal Rate", https://arxiv.org/abs/2411.02853


geohot:
This doesn't belong in main tinygrad, only if three projects are using it

PR #7767, opened 2024-11-18
global metadata to uop buffer tracking [pr]

Qazalin:


Qazalin:
blocking merge to add tests asserting uops are freed. All benchmarks are flat.

PR #7769, opened 2024-11-18
alu(c?t0:f0, c?t1:f1) -> c?alu(t0,t1):alu(f0,f1)

chenyuxyz:


chenyuxyz:
does not seem to be universally better. need to think more

chenyuxyz:
continue in #7900

PR #7772, opened 2024-11-18
Raise error when model architecture does not match state dict

eitanturok:
Raise an error if the size of your tensors in the loaded model don't match the input model.

Mentioned in [discord](https://discord.com/channels/1068976834382925865/1069001075828469790/1307730523820658810).


geohot:
Thanks!

PR #7773, opened 2024-11-18
bounty: parallel BLAKE3 in tinygrad

bjsi:
Here's a parallel BLAKE3 implementation in tinygrad.

- Tested against the full set of test vectors [from the official BLAKE3 repository](https://github.com/BLAKE3-team/BLAKE3/tree/master/test_vectors)


test/testextra/test_blake3.py
```diff
@@ -0,0 +1,172 @@
...
+  def test_official_vectors(self):
+    """Test against the official test vectors from: https://github.com/BLAKE3-team/BLAKE"""
+    for vector in self.vectors:
+      input_len = vector["input_len"]
+      expected = vector["hash"]
+      text = self.generate_input(input_len)
+      actual = blake3(text)
+      self.assertEqual(actual, expected)
+
+  def test_file_input(self):
```
geohot, line 159: Huh, why are files involved?

geohot:
Two questions:

1) How fast is this on GPU in MB/s?
2) It's long. Can it be shorter?

test/testextra/test_blake3.py
bjsi: Gotcha, updated, removed a bunch of lines.

bjsi:
1. Performance

I spent the last couple days trying to make it faster. I reordered the data, added JIT to the `mix` function and played around with the mix loop to try and make it more efficient. 

### NV NVIDIA RTX4090

|    Size  | Total Time | MB/s |
|----------|------------|------|
|   100MB  |   12.64s   |    8 |
|   500MB  |   13.09s   |   38 |
|  1GB  |    8.88s   |  113 |
|  2GB  |    9.05s   |  221 |
|  4GB  |   12.53s   |  319 |

### METAL Mac M2

|    Size  | Total Time | MB/s |
|----------|------------|------|
|   100MB  |   6.43s    |   16 |
|   500MB  |   7.01s    |   71 |
|  1GB  |   6.74s    |  148 |
|  2GB  |   8.63s    |  232 |

I tried testing larger inputs, but ran into OOM errors. Also I wasn't able to find any service online that would rent me an AMD GPU so I didn't test that yet.

2. I stripped it down to < 100 lines

bjsi:
M2 GPU backend, 500mb input tensor:
```sh
Benchmarking 508.6 MB...
Time: 0.79s
Throughput: 676.3 MB/s
```
Throughput should get much better as the input size increases because the number of tree hashing steps only increases logarithmically with the input size, but currently it's either lagging or crashing on larger inputs. 500mb + GPU backend was the best performer so far, the metal backend does fine at 256mb but starts to lag/crash at 500mb.

M2 GPU backend 1000mb input tensor:
```sh
Benchmarking 988.1 MB...
Time: 26.31s
Throughput: 39.4 MB/s
```
Adding JIT around smaller chunks instead fixes this, but comes with a performance hit of 5-10x at least :thinking: please let me know if there's something I can improve. Also tests in CI are failing, but pass just fine locally, maybe because of the long JIT warmup?

geohot:
CI needs to work before I'll review this. CI checks correctness, not performance.

bjsi:
NV is failing due to a bug in [GPUOcelot](https://github.com/gpuocelot/gpuocelot) where it doesn't seem to support the shift instruction `shf`. There's no details about this error in CI, I was only able to work this out by running the CI tests in a docker container using [act](https://github.com/nektos/act).

```sh
| terminate called after throwing an instance of 'parser::PTXParser::Exception'
|   what():
| Failed to parse file '':
| 23
| 24
| 25 	ld.param.u64 	%rd1, [E_3_param_0];
| 26 	ld.param.u64 	%rd2, [E_3_param_1];
| 27 	cvta.to.global.u64 	%rd3, %rd1;
| 28 	cvta.to.global.u64 	%rd4, %rd2;
| 29 	mov.u32 	%r1, %tid.x;
| 30 	mul.wide.s32 	%rd5, %r1, 4;
| 31 	add.s64 	%rd6, %rd4, %rd5;
| 32 	ld.global.u32 	%r2, [%rd6];
| 33 	shf.l.wrap.b32 	%r3, %r2, %r2, 16; <---- errors here
| 34 	add.s64 	%rd7, %rd3, %rd5;
| 35 	st.global.u32 	[%rd7], %r3;
| 36 	ret;
| 37
| 38 }
|  (33, 5):  syntax error
| /var/run/act/workflow/15: line 2:  3079 Aborted                 (core dumped) python -m test.testextra.test_blake3
```

dmikushin:
These hashers can be often seen in miners code. The key challenge is to access the lookup tables quickly. The obvious solution is to map lookup tables to the constant memory. But a jedi solution is to map lookup tables to the shared memory, multiple copies of them in different banks to avoid bank conflicts (yes, this sounds crazy, but for those who do not hesitate testing crazy ideas the win is 2x-3x). I'm not sure how useful is this for tinygrad scope of control, JFWIW.

geohot:
Is there a perf test I can check MB/s?

bjsi:
@dmikushin Thanks Dmitry, I did some research and found this [implementation of BLAKE3 in cuda](https://github.com/Blaze-3/BLAKE3-gpu) which seems interesting. I'm sure it would be faster than what I'm doing here. I was hoping to get things working using TinyGrad's abstractions though.

bjsi:
> Is there a perf test I can check MB/s?

Added perf check code. Change the number to change the input tensor size. It crashes over 512mb for me: `GPU=1 python3 -m extra.utilities.blake3 512`.

image

extra/utilities/blake3.py
```diff
@@ -0,0 +1,108 @@
...
+        iv = self.IV.reshape(8, 1).expand(8, paired.shape[-1])
+        counts = Tensor.zeros((2, paired.shape[-1]), dtype=dtypes.uint32)
+        lengths = Tensor.full((1, paired.shape[-1]), 64, dtype=dtypes.uint32)
+        states = iv.cat(iv[:4], counts, lengths, flags, dim=0)
+        chain_vals = ((self.compress_blocks(states, paired, iv) * pair_mask)[:8] + remainder).realize()
+        chain_vals = chain_vals.pad((None, (0, chain_vals.shape[1])))
+      return chain_vals.realize()
+    return TinyJit(tree_hash)
+
+  def tensor_to_blake_input(self, tensor: Tensor, padded_input_size: int) -> Tuple[Tensor, Tensor, Variable]:
```
khpark43, line 62:
```suggestion
  def tensor_to_blake_input(self, tensor: Tensor, padded_input_size: int) -> Tuple[Tensor, Tensor, UOp]:
```
Looks like you tried using `Variable`(`UOp`) from tinygrad.ops, but here it’s actually `UOp.variable`.

bjsi: Ah, thanks for pointing that out!

geohot:
Still waiting on a perf test

bjsi:
Apologies, I've run out of time to work on this. Please feel free to unlock and let others try.

| Input Size | Best Throughput | Average Time |
|------------|----------------|---------------|
| 256.0 MB | 439.9 MB/s | 0.62s |
| 500.0 MB | 629.0 MB/s | 0.86s |
Average of the last 3 runs (stabilized performance)

Note: Input sizes of 700MB and larger produce invalid hash values (all zeros). I didn't manage to figure out why.

PR #7782, opened 2024-11-19
capture the schedule context in process replay [pr]

Qazalin:


geohot:
This is a big slowdown for GPU Image CI, went from 6 -> 8 mins, and now 11 with my added compile3 test

geohot:
And too slow for benchmark

Qazalin:
my bad, merging a diff to delete this ctx from process replay. It shouldn't impact the scheduler's output asts.
https://github.com/tinygrad/tinygrad/pull/7785

Issue #7739, opened 2024-11-16
suttle bug with duplicate indices in setitem


ttomsa:
a = Tensor([1,2,3,4]).reshape(2,2)
a[[0,0], [0,0]] = Tensor([2,0])

a is:
[[2 2] [3 4]] instead of [[0 2] [3 4]]

essentially it doesn't override if it's a 0.
I will look to fix it in a couple of days.

geohot:
I don't understand that example, I don't think that's supposed to work. If you want, add a failing test in a PR you think should pass

chenyuxyz:
isn't that just setitem? it works in numpy and torch and i think we should fix it

geohot:
Oh...I see, sorry. It's using the RHS Tensor as like a queue to pop off of

ttomsa:
I think it's why #7737 is failing

geohotstan:
#7737 is another problem (cuda only weirdness).

One hack for this is to turn the mask into a `nan` mask and do 
`for dim in axis: vb = functools.reduce(lambda x,y: y.isnan().where(x, y), vb.split(1, dim))` 
but this just shifts the "unable to override 0" problem to "unable to override nan"
e.g.
```
a = Tensor([1,2,3,4]).reshape(2,2).float()
a[[0,0], [0,0]] = Tensor([2,float("nan")])
```
would be wrong.

is nan better than 0? haha
(maybe we could now do a Ops.View with arg=Tensors and pattern match that to fancy indexing? :^) )

ttomsa:
Oh yeah my bad, I was looking at the ptx errors and saw when an index was wrong a 0 was involved and realized overriding with 0 doesn't work but I mixed up the actual tensor and desired tensor in my head.

PR #7786, opened 2024-11-19
add failure test for setitem bug

geohotstan:
minimal failure test case for 
#7739

### DIFF
--- a/test/test_setitem.py
+++ b/test/test_setitem.py
@@ -79,6 +79,15 @@ class TestSetitem(unittest.TestCase):
     t[1] -= 1
     np.testing.assert_allclose(t.numpy(), [[0, 1], [3, 4]])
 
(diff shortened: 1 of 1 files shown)


test/test_setitem.py
```diff
@@ -79,6 +79,15 @@ def test_setitem_consecutive_inplace_operator(self):
     t[1] -= 1
     np.testing.assert_allclose(t.numpy(), [[0, 1], [3, 4]])
 
+  # TODO: #7739 fix when setting value 0 to overlapping indices
+  @unittest.expectedFailure
+  def test_setitem_overlapping_indices(self):
+    t = Tensor([1,2,3,4])
+    n = np.array([1,2,3,4])
+    t[[1,1]] = Tensor([1,0])
```
chenyuxyz, line 87: can you also add one that works like Tensor([5,6])?

PR #7789, opened 2024-11-19
update quickstart doc to not dependent on numpy

chenyuxyz:
should remove fetch_mnist from other places too. also indexing is slow!


chenyuxyz:
yea index is too slow...

PR #7793, opened 2024-11-19
fix advanced setitem overlap with 0

ttomsa:
So it might look a bit complicated but essentially the mask reduce is done manually inside the loop instead of outside which requires the squeeze(). Before in y.where(y, x) 0 is treated as False and so it takes the value from x. Instead we use the mask of y so that in the case where the mask is True and the value is 0 it still takes from y.

Edit: fix for #7739

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -1192,11 +1192,11 @@ class Tensor(SimpleMathTrait):  # pylint: disable=abstract-method
         for dim in sum_axis: vb = vb.unsqueeze(dim)
         # axis to be reduced to match self.shape
         axis = tuple(range(first_dim, first_dim + len(big_shape)))
-        # apply mask to v(broadcasted) and reduce such that if v contains repeated indices the last one remains
+        # apply mask to vb(broadcasted) and reduce such that if mask contains repeated indices the last one remains
         vb = vb * mask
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -1192,11 +1192,11 @@ def calc_dim(tensor_dim:int) -> int:
...
         axis = tuple(range(first_dim, first_dim + len(big_shape)))
-        # apply mask to v(broadcasted) and reduce such that if v contains repeated indices the last one remains
+        # apply mask to vb(broadcasted) and reduce such that if mask contains repeated indices the last one remains
         vb = vb * mask
-        for dim in axis: vb = functools.reduce(lambda x,y: y.where(y, x), vb.split(1, dim))
-        # reduce mask and select from v(get rid of extra dims from reduce) for each True element in mask else select from self
-        ret = mask.any(axis).where(vb.squeeze(), self)
+        for dim in axis: mask, vb = functools.reduce(lambda x,y: (x[0]|y[0], y[0].where(y[1], x[1])), zip(mask.split(1, dim), vb.split(1, dim)))
+        # select from vb for each True element in mask else select from self, squeeze to remove extra dims from reduce
+        ret = mask.squeeze().where(vb.squeeze(), self)
```
chenyuxyz, line 1199: cool yea i agree something along this line is needed.

not related to this fix, but something feels wrong about always squeeze here, like what if the the output should have a legit 1 in shape?

PR #7797, opened 2024-11-20
fix advanced setitem with 1 in shape

ttomsa:
chenyu pointed out the issue where if shape contained a 1. This fixes that. This way it only removes the dims from the reduce axis instead of removing all dims with 1.


chenyuxyz:
Thanks!

PR #7800, opened 2024-11-20
early rewrite preload

Qazalin:


Qazalin:
the lazy deletion diff solves this in a cleaner way.

PR #7801, opened 2024-11-20
delete lazy [pr]

Qazalin:
aiming to delete lazy.py while:

1. no change to lazy.py users (tensor, multi, etc).
2. pass scheduler process replay. **Big graph** is identical.

### DIFF
--- a/examples/openpilot/compile3.py
+++ b/examples/openpilot/compile3.py
@@ -50,16 +50,20 @@ def compile(onnx_file):
 
   # checks from compile2
   kernel_count = 0
+  read_image_count = 0
   gated_read_image_count = 0
   for ei in run_onnx_jit.captured.jit_cache:
     if isinstance(ei.prg, CompiledRunner):
       kernel_count += 1
+      read_image_count += ei.prg.p.src.count("read_image")
       gated_read_image_count += ei.prg.p.src.count("?read_image")
-  print(f"kernel_count: {kernel_count}  gated_read_image_count: {gated_read_image_count}")
-  assert kernel_count <= getenv("ALLOWED_KERNEL_COUNT", 0) or getenv("ALLOWED_KERNEL_COUNT", 0) == 0, "too many kernels!"
+  print(f"{kernel_count=},  {read_image_count=}, {gated_read_image_count=}")
+  if (allowed_kernel_count:=getenv("ALLOWED_KERNEL_COUNT", -1)) != -1:
+    assert kernel_count <= allowed_kernel_count, f"too many kernels! {kernel_count=}, {allowed_kernel_count=}"
+  if (allowed_read_image:=getenv("ALLOWED_READ_IMAGE", -1)) != -1:
+    assert read_image_count == allowed_read_image, f"different read_image! {read_image_count=}, {allowed_read_image=}"
   if (allowed_gated_read_image:=getenv("ALLOWED_GATED_READ_IMAGE", -1)) != -1:
-    assert gated_read_image_count <= allowed_gated_read_image, \
-      f"too many gated read_image! {gated_read_image_count=}, {allowed_gated_read_image=}"
+    assert gated_read_image_count <= allowed_gated_read_image, f"too many gated read_image! {gated_read_image_count=}, {allowed_gated_read_image=}"
 
   with open(OUTPUT, "wb") as f:
     pickle.dump(run_onnx_jit, f)
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -1,4 +1,4 @@
-import sys, atexit, functools
+import sys, functools
 from collections import defaultdict, deque
 from dataclasses import dataclass, field
 from typing import FrozenSet, Set, Tuple, List, Dict, Optional, DefaultDict
@@ -216,14 +216,8 @@ def full_ast_rewrite(pre:UOp, ctx:ScheduleContext) -> Tuple[UOp, ScheduleItemCon
         and ShapeTracker.from_shape(s.shape).shrink(m) == s.shrink(m)) for x in ops):
       raise RuntimeError("self operand of augmented assign must be contiguous.\nhelp: consider using .contiguous():\n"
                          +colored("   - a += a.T\n", "red")+colored("   + a += a.T.contiguous()", "green"))
-  if getenv("RUN_PROCESS_REPLAY"): PROCESS_REPLAY_CAPTURE.append(((pre, ctx.assigns), sink))
+  if getenv("RUN_PROCESS_REPLAY"): diskcache_put("schedule_process_replay", str(pre.key), (pre, si_ctx.assigns, {}, sink))
   return sink, si_ctx
-
-PROCESS_REPLAY_CAPTURE: List[Tuple[Tuple[UOp, Set[UOp]], UOp]] = []
-if getenv("RUN_PROCESS_REPLAY"):
-  @atexit.register
-  def save_process_replay() -> None:
-    for x,ret in PROCESS_REPLAY_CAPTURE: diskcache_put("schedule_process_replay", str(x[0].key), (*x, {}, ret))
 
 # **** Schedule grouping
 
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -449,13 +449,7 @@ def create_schedule_with_vars(outs:List[LazyBuffer]) -> Tuple[List[ScheduleItem]
       ast, ast_ctx = full_ast_rewrite(UOp.sink(*stores), ctx)
       prescheduled.append(ScheduleItem(ast, tuple(u.buffer for u in ast_ctx.bufs if u.size != 0), tuple(ast_ctx.metadata),
                                        frozenset(ubuf for ubuf,ops in ast_ctx.assign_adj.items() if any(x.op is Ops.PRELOAD for x in ops))))
-      for u in ast_ctx.sinked:
-       # can only schedule once
-       lazy = ast_ctx.lazybufs[u]
-       st = unwrap(lazy.st)
-       lazy.op = Ops.VIEW
-       lazy.arg = st
-       lazy.src = (u,)
+      for buf_uop in ast_ctx.sinked: (luop:=ast_ctx.lazybufs[buf_uop]).become(buf_uop.view(unwrap(luop.st))) # can only schedule once
   # do BFS
   schedule_targets = {out:si for si in prescheduled for out in si.outputs}
   graph: DefaultDict[ScheduleItem, List[ScheduleItem]] = defaultdict(list)
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -40,7 +40,7 @@ class ScheduleItem:
 
 @dataclass(frozen=True)
 class ScheduleContext:
-  lazybufs: Dict[UOp, LazyBuffer] = field(default_factory=dict)      # this maps BUFFER uops of this schedule to the underlying lazybuffer
+  lazybufs: Dict[UOp, List[LazyBuffer]] = field(default_factory=dict) # this maps BUFFER uops of this schedule to the underlying lazybuffer
   var_vals: Dict[Variable, int] = field(default_factory=dict)        # this maps a BIND's DEFINE_VAR to its value
   assigns: Set[UOp] = field(default_factory=set)                     # this holds all the BUFFER uops we ASSIGN to in this schedule
   realizes: Dict[UOp, UOp] = field(default_factory=dict)             # this holds all the BUFFER uops we mutate in this schedule
@@ -72,8 +72,6 @@ def to_uop(buf:LazyBuffer, ctx:ScheduleContext, cache:Dict[LazyBuffer, UOp]) -> 
   elif is_scheduled(buf):
     buf_uop = buf.buf_uop
     op = buf.src[1].replace(src=tuple(to_uop(x, ctx, cache) for x in buf.src[1].src))
-    # BUFFER_VIEW overrides the underlying buffer
-    if op.op is Ops.BUFFER_VIEW: buffers[buf_uop] = (x:=op.src[0]).base.buffer.view(buf.st.size, dtype, unwrap(x.st).views[0].offset*x.dtype.itemsize)
   # ASSIGN uses the target buffer, otherwise we create a new buffer
   else:
     src = tuple(to_uop(x, ctx, cache) for x in buf.srcs)
@@ -81,7 +79,7 @@ def to_uop(buf:LazyBuffer, ctx:ScheduleContext, cache:Dict[LazyBuffer, UOp]) -> 
     op = UOp(buf.op, dtype.base, src, buf.arg)
   ret = UOp(Ops.VIEW, dtype.base, (buf_uop,) if op is None else (buf_uop, op.alu(Ops.CONTIGUOUS) if buf.forced_realize else op), buf.st)
   # track the underlying lazydata for this op
-  if op is not None: ctx.lazybufs[buf_uop] = buf
+  if op is not None: ctx.lazybufs[buf_uop] = [buf]
   cache[buf] = ret
   return ret
 
@@ -142,7 +140,7 @@ view_right = merge_views+PatternMatcher([
 
 @dataclass(frozen=True)
 class ScheduleItemContext:
-  lazybufs: Dict[UOp, LazyBuffer]
+  lazybufs: Dict[UOp, List[LazyBuffer]]
   ops_metadata: Dict[UOp, Metadata]
   assigns: Set[UOp]
   var_vals: Dict[Variable, int]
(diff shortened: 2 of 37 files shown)


Qazalin:
status: test_tiny_add passes!
I'm still exploring the options for keeping track of realized uops. I think understanding these uop lifetimes is th linee key to getting this green and merged.

The direction is to make the green diff in ops.py as minimal as possible, methods coming from lazy.py will be one liners, with the actual folding / if else stuff becoming UPats in init_big_graph.

Qazalin:
status: test_tiny_add passes!
I'm still exploring the options for keeping track of realized uops. I think understanding these uop lifetimes is the key to getting this green and merged.

The direction is to make the green diff in ops.py as minimal as possible, methods coming from lazy.py will be one liners, with the actual folding / if else stuff becoming UPats in init_big_graph.

Qazalin:
schedule is playing nice so far, I need to find a way to .substitude in BFS order:

The ADD is real:
image
image

geohot:
I've been looking into exactly that. See https://github.com/tinygrad/tinygrad/commit/144e9f00dfa2c4016f82422570781263c59fbd7e#diff-88cfb30e18dbaf87a4f5cd9031cd6d72d7e332ac9dcc1e527059b042106bcd1fR261

Qazalin:
uop VIEW must match lazy's version
image

Qazalin:
The direction is more clear, I'll let the scheduler (re)create the big graph.
`big_graph = UOp.sink(*outs)` is making it harder to make progress on the primary goal.

Qazalin:
yay! This diff is much easier to work with.

Now we have real failing tests showing bugs in UOp.view, contiguous and missing ALU folding stuff. I'll work toward making this match lazy.

image

Qazalin:
yay so many test_tinys passing! The swizzle work today really helped with this diff. more problems:
- jit broken - don't know why yet
- test_abs broken trying to access device on CONST
- test_image broken, "make things that can't be images not images" trying to access buffer of unrealized mul
- contiguous forced_realize needs work

Qazalin:
test_ops: 66 failed, 269 passed
all of them related to CONST.

Qazalin:
fixed all the const stuff. no DEVICECONST, just early VIEW normal CONST. late rewrite to VALID.
10 failing test_ops left, all related to forced_realize i think.

Qazalin:
9 test_ops left.

Qazalin:
7, left,
CONST ended up being much simpler, it's consistent with how we already think about CONST in master.
Very happy with these folding rules. `big_graph = UOp(Ops.NOOP)` folding in #7940 makes BUFFER stuff easy.
image

Qazalin:
the memory planner fix is terrible, but it hints to a future direction.
Buffer on ScheduleItem will be gone. ScheduleItem can be just a UOp container, we'll get the buffers later in realize.

geohot:
So the JIT is fine?

tinygrad/ops.py
```diff
@@ -378,19 +413,34 @@ def stride(self, arg:Tuple[int, ...]): return self.view(unwrap(self.st).stride(a
 
   # *** uop Buffer stuff ***
 
+  buffer_num = itertools.count(0)
```
geohot, line 416: cute

Qazalin:
no JIT is still weird, see test_tiny.

I added some always_on tests to show the new behavior.

geohot:
Btw, the test_ops thing is weird. I think it's (method) cache related, the tests don't fail standalone

tinygrad/ops.py
```diff
@@ -278,6 +281,8 @@ def full_shape(self) -> Tuple[sint, ...]:
   def shape(self) -> Tuple[sint, ...]: return unwrap(self.st).shape
   @property
   def size(self) -> int: return self.arg[1][1] if self.op is Ops.BUFFER else unwrap(self.st).size
+  @property
+  def nbytes(self) -> int: return self.size*self.dtype.itemsize
```
geohot, line 285: Things like this can be merged separately.

geohot:
Also, I would split this into required vs optimizations

Qazalin:
planning to merge
1. patterns in ops_folding (diff will look like #7940, red lazy green schedule upat)
2. buffer_num = itertools.count(0)
ofc every ops method I use in those diffs will be merged with that too.

Qazalin:
I'm targeting correctness first (test_ops). Kernel asserts if I don't do some zero folding on REDUCE_AXIS early, maybe it shouldn't.

geohot:
The JIT doesn't look broken, it was just sort of a bad test

geohot:
Moving const reduce simplify here https://github.com/tinygrad/tinygrad/pull/7979

tinygrad/engine/schedule.py
```diff
@@ -338,9 +341,25 @@ def _as_const(u:UOp, val:ConstType) -> UOp:
...
 
+def simplify_const_reduce(ctx, src:UOp, to_store:UOp, **kwargs) -> Optional[UOp]:
+  # must be unmasked
+  if any(v.mask is not None for v in unwrap(src.st).views): return None
+  prshape = prod(unwrap(to_store.src[0].st).shape[i] for i in to_store.arg[1])
+  val = src.src[1].arg
+  match to_store.arg[0]:
+    case Ops.ADD: ret = val*prshape
+    case Ops.MUL: ret = val**prshape
+    case _: ret = val
```
geohot, line 352: Careful, I think you should add Ops.MAX here and assert (or return None) if it's something unhandled

Qazalin:
207 down to 166 with weakrefs
image
image

JIT looks fixed.
Last blocker on correctness is #8000.
working on const folding now!

Qazalin:
getting greener,
left:
- [ ] permuted assignment
- [ ] const folding
- [ ] contiguous folding
- [ ] missing fxns, eg .clone?

Qazalin:
only the OOM stuff left, I know we said GC is optimization not correctness.
But looks like it's blocking the merge.

Qazalin:
so the rest of the correctness failures come down to figuring out what datastrcture / reference makes sense for these 3:
image.

geohot:
There's failures in WebGPU that don't look like OOM

Qazalin:
> There's failures in WebGPU that don't look like OOM.
> padto failures in AMD?

It was because of the weakref change to raw dict, the JIT expects ASTs to rerun when it changes realized input bufs, so using Dict to keep track of lazydata -> realized uop isn't correct.

compile3 is pickling realized buffers. In LazyBuffer = UOp there's no .buffer in UOp, if you pickle a UOp it won't pickle the realized Buffer.
Do you wanna keep the support for pickling realized buffers? I'd try something like check the realized dict and on __reduce__ an on unpickle, append again.

Qazalin:
The WeakKeyDict solves some of the GC problems but there are weird cases where the BUFFER uop is preventing Buffer gc.

geohot:
Yes, you have to be able to pickle realized buffers, and you can put that logic in Buffer __reduce__

Why are all the tests non even finishing? If it's GC, is there a small set you can disable to make them pass? I don't want tests that use tons of memory anyway

examples/openpilot/compile3.py
Qazalin, line 1: This diff is untasteful, but captured JIT pickle doesn't work yet. working on adding it back.

Qazalin:
re: tests, I'm running them in a sequence to see where they fail. It's around here
https://github.com/tinygrad/tinygrad/actions/runs/12223389081/job/34094864928#step:15:93. finding that set.

Qazalin:
skip set: https://github.com/tinygrad/tinygrad/pull/7801/commits/f01e3592330ea5f9f0efa3be101fc9386426f4d3

Qazalin:
Correctness is green.

For feature completeness, It needs captured JIT pickling back. The compile3.py diff won't exist after it's fixed.

For speed, it needs Buffer freeing and GC, benchmarks getting OOM: https://github.com/tinygrad/tinygrad/actions/runs/12224022775/job/34096257776.

Qazalin:
>  and you can put that logic in Buffer `__reduce__`

I don't think so, Buffer `__reduce__` never gets called when pickling lazydata, it doesn't live on UOp.

This diff solves pickle feature completeness:
image

Moving to weakref experiments now.

Qazalin:
>  and you can put that logic in Buffer `__reduce__`

I don't think so, pickle.dumps lazydata doesn't call Buffer `__reduce__`. Buffer doesn't live on the uop.

This diff solves pickle feature completeness:
image

Qazalin:
The GC problems are fundamentally because in master, we `del srcs` after scheduling.
This diff doesn't do it yet.
See how EMPTY is gone early, vs kept around until exit.
image

Otherwise the maps from UOp to Buffer are working as expected, Buffer frees as soon as UOp deletes.

Qazalin:
So the solution is to override the tensor's lazydata in the scheduler. quick prototype:
```
a = Tensor([1, 2, 3, 4]).realize()
a.lazydata = UOp(Ops.VIEW, a.lazydata.dtype, (buf_uop:=realized[a.lazydata],), unwrap(a.lazydata.st))
realized[a.lazydata] = buf_uop
print(a)
assert tuple(a.tolist()) == (1, 2, 3, 4)
```
image

examples/openpilot/compile2.py
```diff
@@ -149,7 +149,8 @@ def to_ref(b:Buffer): return struct.pack("Q", id(b)).decode("latin_1")
         })
         if needs_load:
           t = Tensor.empty(b.dtype.shape, dtype=b.dtype)
-          t.lazydata.buffer = b
+          # TODO: what does this do? uop is immutable
+          #t.lazydata.buffer = b
```
geohot, line 153: It's used in the export. You want to replace this with setting the buffer in the dicts, but really we should just delete compile2

tinygrad/engine/schedule.py
```diff
@@ -338,9 +338,28 @@ def _as_const(u:UOp, val:ConstType) -> UOp:
...
+    prshape = prod(unwrap(x.st).shape[i] for i in root.arg[1])
+    ret = x.base.src[1].arg
+    match root.arg[0]:
+      case Ops.ADD: ret *= prshape
+      case Ops.MUL: ret **= prshape
+      case Ops.MAX: pass # NOTE: Ops.MAX is passthrough
+      case _: return None
+    return UOp.const(root.dtype, ret)
+  return None
+
```
Qazalin, line 355: moving this part to its own diff.

tinygrad/ops.py
```diff
@@ -203,20 +204,35 @@ def dfs(x:Any, cache:dict):
 
 class UOpMetaClass(type):
   ucache:Dict[Tuple, weakref.ReferenceType[UOp]] = {}
-  def __call__(cls, op:Ops, dtype:DType=dtypes.void, src:Tuple[UOp,...]=tuple(), arg:Any=None):
+  def __call__(cls, op:Ops, dtype:DType=dtypes.void, src:Tuple[UOp,...]=tuple(), arg:Any=None,
+               # NOTE: these don't live on the UOp class, they're internal state a pickled uop can map to
+               _device_buffer:Optional[Buffer]=None, _buf_uop:Optional[UOp]=None):
```
geohot, line 209: I don't like this. If the UOp cache hits, this won't set. Why does this have to be done here?

tinygrad/ops.py
```diff
@@ -299,13 +315,27 @@ def axis_arg(self) -> Tuple[int, ...]:
...
+    return UOp(Ops.CAST, dtype, (self,))
+  def bitcast(self, dtype:DType, allow_buffer_view=True):
+    if self.can_view() and allow_buffer_view:
+      if self.dtype.itemsize == dtype.itemsize: output_shape = self.shape
+      else:
+        if not self.device.startswith("DISK") or not all_int(self.shape): raise RuntimeError(f"shape changing bitcast not supported on {self}")
+        # https://pytorch.org/docs/stable/generated/torch.Tensor.view.html
+        if (self.shape[-1]*self.dtype.itemsize) % dtype.itemsize != 0: raise RuntimeError("unsupported size in bitcast")
+        output_shape = self.shape[:-1]+((self.shape[-1]*self.dtype.itemsize) // dtype.itemsize,)
+      return UOp.metaop(Ops.BUFFER_VIEW, output_shape, dtype, self.device, None, (self,))
```
geohot, line 337: I don't think this logic belongs here. It should be a rewrite rule of BITCAST

tinygrad/ops.py
```diff
@@ -332,16 +362,52 @@ def r(self, op:Ops, axis:Tuple[int, ...]):
     axis = tuple(sorted([x for x in axis if resolve(self.shape[x] != 1)]))
     return self if len(axis) == 0 else UOp(Ops.REDUCE_AXIS, self.dtype, (self,), (op, axis))
   def assign(self, x:UOp): return UOp(Ops.ASSIGN, self.dtype, (self,x), None if self.st is None or self.st.contiguous else self.st)
-  def contiguous(self): return UOp(Ops.CONTIGUOUS, self.dtype, (self,))
+  def contiguous(self, allow_buffer_view=True):
+    if not unwrap(self.st).contiguous or self.size != self.base.size or self.is_unrealized_const():
+      if allow_buffer_view and self.can_view(): return self.metaop(Ops.BUFFER_VIEW, self.shape, self.dtype, self.device, None, (self,))
```
geohot, line 367: Same as above with this BUFFER_VIEW.

tinygrad/ops.py
```diff
@@ -332,16 +362,52 @@ def r(self, op:Ops, axis:Tuple[int, ...]):
...
+      return self.alu(Ops.CONTIGUOUS)
+    forced_realize.add(self.base)
```
geohot, line 369: This should be in the scheduler, not here

tinygrad/tensor.py
```diff
@@ -2995,7 +2997,7 @@ def _broadcasted(self, y:Union[Tensor, UOp, ConstType], reverse:bool=False, matc
     return x._broadcast_to(out_shape:=_broadcast_shape(x.shape, y.shape)), y._broadcast_to(out_shape)
 
   def _to_const_val(self, x:Union[Tensor, ConstType]) -> Union[Tensor, ConstType]:
```
geohot, line 2999: This should be removed from Tensor

geohot:
test skips:
* optimization:
** `ALLOWED_KERNEL_COUNT=256 ALLOWED_GATED_READ_IMAGE=47`
** `with CLCache(24):` in external_test_opt
** `Test*OpsConstFolding`
** `@unittest.skip("split_reduceop isn't supported")`
* required features
** `TestTensorMetadata`
* bugs?
** `with self.assertRaises(AssertionError, msg="indexing folding no longer supported"):`
** `@unittest.skipIf(Device.DEFAULT=="CLANG" and CI, "fails on clang CI")`
** `@unittest.skipIf(CI, "fails in CI after delete_lazy")`
** all of test_schedule
* GC
** `TestRealWorld`, `TestTrain`, `TestSpeed`, `TestWinograd`, `TestLrScheduler`
** `test_schedule_gc`, `test_schedule_gc_with_inputs`
** test_lru_alloc, `@unittest.skip("TODO: why does this fail in CI")`
* fine
** `@unittest.skip("forced_realize isn't supported")`
* unknown
** `test_sharded_memory`
** `TestMoveTensor`
** `test_multi_input_model_export`

tinygrad/ops.py
```diff
@@ -332,16 +362,52 @@ def r(self, op:Ops, axis:Tuple[int, ...]):
...
+    return UOp.metaop(Ops.COPY, src.shape, src.dtype, device, src.nbytes, (src,)).view(unwrap(self.st))
+  def clone(self) -> UOp: return self.copy_to_device(self.device, clone=True)
+  def is_unrealized_const(self): return (s:=self.base).op is Ops.VIEW and len(s.src) == 2 and s.realized is None and s.src[1].op is Ops.CONST
   def is_unrealized_unmasked_const(self): return self.is_unrealized_const() and all(v.mask is None for v in unwrap(self.st).views)
+  def can_view(self):
+    return (self.st is not None and self._device is not None and self.st.consecutive and not self.is_unrealized_const() and
+            not isinstance(self.dtype, ImageDType) and self.device.split(":")[0] in view_supported_devices)
+  @property
+  def srcs(self): return self.src
+  @srcs.deleter
```
geohot, line 403: Looks like this isn't used

Qazalin: tensor uses it. for EMPTY fake realize hack.

tinygrad/tensor.py
Qazalin: yea I hate that tensor is accessing args to begin with, I'll make this an actual wrapper that asserts.

tinygrad/ops.py
Qazalin: yea I need it for pickle. If I return the BUFFER in`__reduce__` I need to somehow add it back in.
I'll try to see if this can be handled in the Buffer init maybe

Qazalin:
re: bugs

- sorry that test_indexing expectedFailure was untasteful, #8133

These pass locally and on tinybox, I'll try a github actions docker next
- `@unittest.skipIf(Device.DEFAULT=="CLANG" and CI, "fails on clang CI")`
- `@unittest.skipIf(CI, "fails in CI after delete_lazy")`

Qazalin:
it's red now because before some buf_uops that ended up folding didn't get deleted from the ctx.
#8155 revealed this.
https://github.com/tinygrad/tinygrad/pull/8151 is finalizing the buffer merging changes, that should make this green.

PR #7805, opened 2024-11-20
add inverse trig functions to Tensor

geohotstan:
copied from onnx to tensor with some code cosmetics applied

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -2490,10 +2490,10 @@ class Tensor(SimpleMathTrait):  # pylint: disable=abstract-method
     print(Tensor([-3., -2., -1., 0., 1., 2., 3.]).atan().numpy())
     ```
     """
-    x = (1 > self.abs()).detach().where(self.abs(), self.abs().reciprocal())
+    magnitude = self.abs()
+    x = (1 > magnitude).detach().where(magnitude, magnitude.reciprocal())
     x = x * polyN(x*x, [-0.013480470, 0.057477314, -0.121239071, 0.195635925, -0.332994597, 0.999995630])
-    x = (self.abs() > 1).detach().where(1.570796327 - x, x)
-    return self.sign() * x
+    return self.sign() * (magnitude > 1).detach().where(1.570796327 - x, x)
 
   # ***** math functions *****
 
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -2461,6 +2461,40 @@ class Tensor(SimpleMathTrait):  # pylint: disable=abstract-method
     """
     return self.sin() / self.cos()
 
+  def asin(self):
+    """
+    Computes the inverse sine (arcsine) of the tensor element-wise.
+
+    ```python exec="true" source="above" session="tensor" result="python"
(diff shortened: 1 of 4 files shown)


tinygrad/tensor.py
```diff
@@ -2461,6 +2461,40 @@ def tan(self):
...
+    Computes the inverse tangent (arctan) of the tensor element-wise.
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor([-3., -2., -1., 0., 1., 2., 3.]).atan().numpy())
+    ```
+    """
+    magnitude = self.abs()
+    x = (1 > magnitude).detach().where(magnitude, magnitude.reciprocal())
+    x = x * polyN(x*x, [-0.013480470, 0.057477314, -0.121239071, 0.195635925, -0.332994597, 0.999995630])
+    return self.sign() * (magnitude > 1).detach().where(1.570796327 - x, x)
```
chenyuxyz, line 2496: math.pi / 2?

tinygrad/tensor.py
```diff
@@ -2461,6 +2461,40 @@ def tan(self):
...
+  def atan(self: Tensor):
+    """
+    Computes the inverse tangent (arctan) of the tensor element-wise.
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor([-3., -2., -1., 0., 1., 2., 3.]).atan().numpy())
+    ```
+    """
+    magnitude = self.abs()
+    x = (1 > magnitude).detach().where(magnitude, magnitude.reciprocal())
```
chenyuxyz, line 2494: keep the conditions the same reads better, and it's faster too due to caching

tinygrad/tensor.py
```diff
@@ -2461,6 +2461,40 @@ def tan(self):
...
+  def acos(self: Tensor):
+    """
+    Computes the inverse cosine (arccosine) of the tensor element-wise.
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor([-0.9, -0.6, -0.3, 0., 0.3, 0.6, 0.9]).acos().numpy())
+    ```
+    """
+    x = (1.0 - self.abs()).sqrt() * polyN(self.abs(), [-0.0187293, 0.0742610, -0.2121144, 1.5707288])
+    return (self < 0) * math.pi + x * (1 - 2 * (self < 0))
```
chenyuxyz, line 2483: write with where like atan?

tinygrad/tensor.py
```diff
@@ -2461,6 +2461,40 @@ def tan(self):
...
+    """
+    Computes the inverse sine (arcsine) of the tensor element-wise.
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor([-0.9, -0.6, -0.3, 0., 0.3, 0.6, 0.9]).asin().numpy())
+    ```
+    """
+    return (self / (1 - self * self).sqrt()).atan()
+
+  def acos(self: Tensor):
```
geohot, line 2474: No type signature required here (asin doesn't have)

geohotstan:
Thanks for the review
Yeah that was careless of me 😔

I think I made it even better now.

PR #7810, opened 2024-11-21
test SD with WINO and JIT=1/2

chenyuxyz:
on M1 Max JIT=2 is 10% faster than JIT=1...


chenyuxyz:
cannot repro on CI...

PR #7826, opened 2024-11-21
lazy view folding is all UPats

Qazalin:


Qazalin:
I need to first build some tooling to process replay from LazyBuffer, the tests I expected to fail didn't do anything.
There's 3x more kernels in openpilot than what it should be because of this, need a way to diff big graphs first.

PR #7830, opened 2024-11-21
move attention upcast

chenyuxyz:
still upcast before softmax, but faster because intermediate buffer can be stored in half (as long as qk is within half range).


chenyuxyz:
trained two berts on red and green and no nan, so i think we are good

PR #7842, opened 2024-11-22
Use MTLCompiler directly

uuuvn:
MTLCompilerService is just a wrapper around MTLCompiler dylib that shares llvm instance between processes. All it does is dlopen on appropriate MTLCompiler version and forwards the rest of the request to it

I've also commited a bunch of tools and two scripts to compile via xpc and via MTLCompiler directly that i wrote for exploring all of this to extra in case they will be needed for something else. (i've looked at other accepted PRs and they commit extra/ stuff together with tinygrad/ even when tinygrad/ stuff doesn't depend on extra/ so i did the same here)

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
-import os, subprocess, pathlib, ctypes, tempfile, functools
-from typing import List, Any, Tuple, Optional, cast
-from tinygrad.helpers import prod, getenv, T
+import os, subprocess, pathlib, struct, ctypes, tempfile, functools
+from typing import List, Any, Tuple, cast
+from tinygrad.helpers import prod, to_mv, getenv, _cache_dir, T
 from tinygrad.device import Compiled, Compiler, CompileError, LRUAllocator
 from tinygrad.renderer.cstyle import MetalRenderer
 
@@ -24,12 +24,15 @@ class MTLPipelineOption:
 
 libobjc = ctypes.CDLL("/usr/lib/libobjc.dylib")
 libmetal = ctypes.CDLL("/System/Library/Frameworks/Metal.framework/Metal")
+compiler = ctypes.CDLL("/System/Library/PrivateFrameworks/MTLCompiler.framework/MTLCompiler")
 # Must be loaded for default Metal Device: https://developer.apple.com/documentation/metal/1433401-mtlcreatesystemdefaultdevice?language=objc
 ctypes.CDLL("/System/Library/Frameworks/CoreGraphics.framework/CoreGraphics")
 libdispatch = ctypes.CDLL("/usr/lib/libSystem.dylib") # libdispatch is part of libSystem on mac
 libobjc.objc_getClass.restype = objc_id
 libobjc.sel_registerName.restype = objc_id
 libmetal.MTLCreateSystemDefaultDevice.restype = objc_instance
+compiler.MTLCodeGenServiceCreate.restype = ctypes.c_void_p
+compiler.MTLCodeGenServiceBuildRequest.restype = ctypes.c_void_p
 libdispatch.dispatch_data_create.restype = objc_instance
 
 # Ignore mypy error reporting incompatible default, because typevar default only works on python 3.12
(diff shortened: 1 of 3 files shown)


uuuvn:
One other interesting thing i've found while doing this is that the MTLB files that are produced even when the compilation is done with metal device are compiled second time into MTLPSBIN somewhere inside program with the help of actually platform-dependent library (AGXCompilerCore in my case) and targetData, maybe we can compile down to that ourselves to avoid unnecessary XPC calls and stuff

tinygrad/runtime/ops_metal.py
```diff
@@ -65,19 +68,33 @@ def metal_src_to_library(device:MetalDevice, src:str) -> objc_instance:
...
+      @ctypes.CFUNCTYPE(None, ctypes.c_void_p, ctypes.c_int32, ctypes.c_void_p, ctypes.c_size_t, ctypes.c_char_p)
+      def callback(blockptr, error, dataPtr, dataLen, errorMessage):
+        nonlocal lib, err
+        if error != 0:
+          err = CompileError(errorMessage.decode())
+          return
+        reply = bytes(to_mv(dataPtr, dataLen))
+        offset = struct.unpack('<Q', reply[8:16])[0]
+        lib, err = reply[offset:], None
+      params = "-fmetal-math-mode=safe -fmetal-math-fp32-functions=precise -std=metal3.0 --driver-mode=metal -x metal"
```
uuuvn, line 92: It looks like "-fmetal-math-mode=safe -fmetal-math-fp32-functions=precise" aren't available on old versions of macos, i'll find a way to fix this and then mark it ready to review again

uuuvn:
```
# compile_apple.py
Compilation took 30.4 ms
# compile_xpc.py
Opening xpc connection took 2.6 ms
Compilation took 26.9 ms
# compile_direct.py
Initialization took 0.9 ms
Compilation took 7.7 ms
```

geohot:
Can you make this PR minimal and as readable as possible?

uuuvn:
> Can you make this PR minimal and as readable as possible? It's +531

Removed everything from extra/ . Should i submit it afterwards or just keep to myself? I think those scripts and helpers will be useful for other people who might want to do anything macos/metal related in tinygrad.

tinygrad/runtime/ops_metal.py
```diff
@@ -65,19 +68,28 @@ def metal_src_to_library(device:MetalDevice, src:str) -> objc_instance:
...
-      library_contents = msg(library, "libraryDataContents", restype=objc_instance)
-      lib = ctypes.string_at(msg(library_contents, "bytes"), cast(int, msg(library_contents, "length", restype=ctypes.c_ulong)))
-    assert lib[:4] == b"MTLB", "Invalid Metal library. Using conda? Corrupt XCode?"
+      lib, err = None, CompileError("MTLCodeGenServiceBuildRequest returned without calling the callback")
+      @ctypes.CFUNCTYPE(None, ctypes.c_void_p, ctypes.c_int32, ctypes.c_void_p, ctypes.c_size_t, ctypes.c_char_p)
+      def callback(blockptr, error, dataPtr, dataLen, errorMessage):
+        nonlocal lib, err
+        if error == 0: lib = (reply:=bytes(to_mv(dataPtr, dataLen)))[sum(struct.unpack('<LL', reply[8:16])):]
+        else: err = CompileError(errorMessage.decode())
+      params = f'-fno-fast-math -std=metal3.1 --driver-mode=metal -x metal -fmodules-cache-path=\"{os.path.join(_cache_dir, "tinygrad")}\"'
```
geohot, line 87: Do you need a cache?

uuuvn: Yes, otherwise it's 250 ms instead of 8 ms. As i understand it caches stdlib compilation

geohot:
This can be way more polished.

tinygrad/runtime/ops_metal.py
uuuvn: No cache:
Compilation took 254.2 ms
Cache:
Compilation took 7.8 ms

This was in compile_direct.py

geohot: There should be a comment explaining this, right? It's not obvious why.

Saying "this was in compile_direct.py" does nothing for me.

geohot:
I'm going to close this.

If you want PRs merged into tinygrad, please ask, is this understandable to a smart person reading it? If you have a random 13 and -0x10 in there, it doesn't matter how smart you are, the context just isn't there. Same thing re: the caching. Someone should expect that tinygrad can do the caching, if the stdlib thing is true I shouldn't have to read 500 lines to learn about it.

I can tell you put a lot of effort into this reverse engineering. Similar with the clang PRs. But being a good software engineer is all about writing code that's readable and understandable to others. It's fine to say, look, I'm smarter than George and if he can't understand it it's on him. But if you want that, create your own fork of tinygrad.

We spoke on Discord VC once about if you wanted to be a software engineer. If you are half way about it, there's no reason to submit PRs here. But if you are serious about software engineering, you need to write code such that others can read, and I promise I won't be the only person giving you this feedback.

PR #7843, opened 2024-11-22
Add Raspberry Pi (aarch64) support for Clang

hornos:
-march=native does not work on Raspberry Pi


geohot:
No test. Failing linter

PR #7846, opened 2024-11-22
mypy==1.13.0

JakobSpahn:


tinygrad/tensor.py
```diff
@@ -262,7 +262,7 @@ def _data(self) -> memoryview:
     if self.device != "CLANG": buf.options = BufferSpec(nolru=True)
     return buf.as_buffer(allow_zero_copy=True if self.device != "CLANG" else False)
 
-  def data(self) -> memoryview:
+  def data(self) -> memoryview[Any]:
```
JakobSpahn, line 265: Using `memoryview[Any]` instead of the proper narrow `Union[memoryview[float], memoryview[int], memoryview[bool]`. This saves a bunch of lines without much cheating imo.

tinygrad/runtime/graph/hcq.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
 
     for j,ji in enumerate(jit_cache):
       enqueue_dev = ji.prg.dev if (is_exec_prg:=isinstance(ji.prg, CompiledRunner)) else Device[ji.bufs[1].device] #type:ignore
+      assert isinstance(enqueue_dev, HCQCompiled) and (isinstance(enqueue_dev.hw_copy_queue_t, type) and enqueue_dev.hw_copy_queue_t is HWQueue)
```
JakobSpahn, line 55: The second expression of this assert could also be replaced with `enqueue_dev.hw_copy_queue_t is not None`, which is however less narrow.

tinygrad/runtime/ops_python.py
```diff
@@ -67,12 +67,12 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
         assert dtype is not None, f"{uop} is missing a dtype"
         dl[i] = dtype
         if uop is Ops.DEFINE_GLOBAL:
-          assert dtype.fmt is not None
-          ul[i] = [pbufs.pop(0).cast(dtype.fmt)] * warp_size
+          assert (dtype.fmt is not None)
+          ul[i] = [pbufs.pop(0).cast(cast(LegacyFmtStrType, dtype.fmt))] * warp_size
```
JakobSpahn, line 71: The stubs for memoryview cast have not yet been updated for half https://github.com/python/typeshed/blob/e1c74f08f12b8bec6afe266951f0756dc1b43ebe/stdlib/builtins.pyi#L872

tinygrad/runtime/ops_python.py
```diff
@@ -68,11 +68,11 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
         dl[i] = dtype
         if uop is Ops.DEFINE_GLOBAL:
           assert dtype.fmt is not None
-          ul[i] = [pbufs.pop(0).cast(dtype.fmt)] * warp_size
+          ul[i] = [pbufs.pop(0).cast(cast(LegacyFmtStrType, dtype.fmt))] * warp_size  # TODO typeshed missing stub for cast
```
JakobSpahn, line 71: The typeshed stubs for `memoryview.cast` have not yet been updated for half  ([mypy/typeshed/stdlib/builtins.pyi](https://github.com/python/mypy/blob/v1.13.0/mypy/typeshed/stdlib/builtins.pyi#L769)), so the Literal type is narrowed for < python 3.12 support.  (https://docs.python.org/3/library/typing.html#typing.cast).

The alternative is to perform runtime checks on `sys.version_info < (3,12)` and then throwing asserts if `dtype.fmt` corresponds to half ('e', '@e').

tinygrad/tensor.py
geohot: Why is it more lines? You can use `|` for union

tinygrad/runtime/ops_python.py
geohot: If you are going to cast it anyway, can it just be a string?

tinygrad/tensor.py
```diff
@@ -2,10 +2,10 @@
 from __future__ import annotations
 import time, math, itertools, functools, struct, sys, inspect, pathlib, string, dataclasses, hashlib
 from contextlib import ContextDecorator
-from typing import List, Tuple, Callable, Optional, ClassVar, Type, Union, Sequence, Dict, DefaultDict, cast, get_args, Literal
+from typing import List, Tuple, Callable, Optional, ClassVar, Type, Union, Sequence, Dict, DefaultDict, cast, get_args, Literal, Any
 from collections import defaultdict
 
-from tinygrad.dtype import DType, DTypeLike, dtypes, ImageDType, ConstType, least_upper_float, least_upper_dtype, sum_acc_dtype, to_dtype, truncate
+from tinygrad.dtype import DType, DTypeLike, dtypes, ImageDType, ConstType, least_upper_float, least_upper_dtype, sum_acc_dtype, to_dtype, truncate, LegacyFmtStrType  # noqa E501
```
geohot, line 8: Never add new ones of these!

geohot:
tinygrad/multi.py
tinygrad/nn/__init__.py
tinygrad/nn/state.py

These three seem fine, I'd merge a PR with just them. The others have tradeoffs

JakobSpahn:
I am going to close this and split it up into multiple PRs for better readability.

PR #7853, opened 2024-11-22
tar_extract with Tensors

leopf:
Supports filename for backwards compatibility.

Limitations:
- ~~max file entry size is 8GiB~~
- ~~max filename length is 100 chars~~

~~These limitations are solved by extensions like PAX and USTAR. These extensions will probably add 5-10 lines each. Are they wanted?~~

Right now it passes the tests in `test_disk_tensor.py`, more tests to come.


leopf:
Added support for the USTAR, PAX and GNU extensions to support long filenames, and large files (greater than 8GiB).

I have also rewritten the tests to test those formats specifically. There are also dirs and links inside the tar now to make sure they are handled properly.

Missing from the tests are the large file tests, since I have not found a way to test this without actually creating a file greater than 8GiB. I have done this locally though, with:
```bash
truncate -s 9G largefile

# GNU Base-256 tar
tar --format=gnu -cf test_gnu.tar largefile

# PAX tar
tar --format=pax -cf test_pax.tar largefile
```

tinygrad/nn/state.py
```diff
@@ -131,17 +131,41 @@ def __init__(self):
       else: v.replace(state_dict[k].to(v.device)).realize()
       if consume: del state_dict[k]
 
-def tar_extract(fn:os.PathLike) -> Dict[str, Tensor]:
+def tar_extract(fn_or_data: Union[os.PathLike, Tensor]) -> Dict[str, Tensor]:
```
geohot, line 134: Don't support both, it all should be Tensor only

tinygrad/nn/state.py
```diff
@@ -131,17 +131,41 @@ def __init__(self):
...
-  t = Tensor(pathlib.Path(fn))
-  with tarfile.open(fn, "r") as tar:
-    return {member.name:t[member.offset_data:member.offset_data+member.size] for member in tar if member.type == tarfile.REGTYPE}
+
+  t = fn_or_data if isinstance(fn_or_data, Tensor) else Tensor(pathlib.Path(fn_or_data))
+  pos, tensor_size, result, fn_ow, fsz_ow = 0, prod(t.shape), {}, None, None
+  if tensor_size < 512*2: raise tarfile.ReadError("Invalid tar file!") # tar files end with at least 2 empty sections
+  while pos + 512 < tensor_size:
+    header, pos = t[pos:pos + 512].data(), pos + 512
+    if header[124] & 0x80: fsz_ow = int.from_bytes(header[125:136], "big")
```
geohot, line 150: There's a method in tarfile to do this, frombuf

geohot:
A lot of lines, can it be simpler and shorter?

leopf:
I removed filename usage support. Had to also change  `cifar` for `mypy`. I guess these are two things in one PR. Would you like me to temporarily bring back filename usage and split this up?

Using `frombuf` I simplified further and have removed 6 lines. I do not really see a way to make it shorter or simpler without removing some extensions, reintroducing many of the limitations from my initial comment.

test/unit/test_disk_tensor.py
```diff
@@ -333,63 +333,79 @@ def test_copy_from_disk_huge(self):
...
-    self.test_dir = tempfile.mkdtemp()
+    self.tar_path = pathlib.Path(tempfile.mktemp(".tar"))
     self.test_files = {
-      'file1.txt': b'Hello, World!',
-      'file2.bin': b'\x00\x01\x02\x03\x04',
-      'empty_file.txt': b''
+      'a/file1.txt': b'Hello, World!',
+      'a/b/file2.bin': b'\x00\x01\x02\x03\x04',
+      'empty_file.txt': b'',
+      'very'*15 + "/" + "very"*15 + '_long_filename.txt': b'Hello, World!'
```
chenyuxyz, line 346: also test one that has more than 512 bytes

leopf: I have added tests for filenames and data larger than 512 bytes.

leopf:
I have moved the tests since they do not rely on disk tensors anymore.

leopf:
Added a test with data length 512 to test ln. 163.

Listened to the meeting.

Tars end with at least 2 empty 512 long blocks. When `frombuf` encounters `b"\x00"*512` it will raise `EOFHeaderError`. Initially I wanted to just suppress this outside the loop, but `tarfile.EOFHeaderError` is not accessible (mypy complains) because it is not in `__all__`. As a workaround I rewrote this logic there. Alternatively I could suppress the `HeaderError`, but this would also suppress checksum errors etc.

tinygrad/nn/datasets.py
```diff
@@ -9,7 +9,7 @@ def _mnist(file): return Tensor.from_url(base_url+file, gunzip=True)
          _mnist("t10k-images-idx3-ubyte.gz")[0x10:].reshape(-1,1,28,28).to(device), _mnist("t10k-labels-idx1-ubyte.gz")[8:].to(device)
 
 def cifar(device=None):
-  tt = tar_extract(fetch('https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz', gunzip=True))
+  tt = tar_extract(Tensor(fetch('https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz', gunzip=True)))
```
geohot, line 12: `Tensor.from_url`

tinygrad/nn/state.py
```diff
@@ -132,17 +132,37 @@ def __init__(self):
...
+    # handle GNU @LongLink
+    if info.type == b"L": next_ow = { "path": bytes(t[pos:pos+info.size].data()).decode().rstrip("\x00") }
+
+    # handle @PaxHeader (https://www.mkssoftware.com/docs/man4/pax.4.asp)
+    if info.type == b"x":
+      pax_text, entry_len = bytes(t[pos:pos+info.size].data()).decode().rstrip("\x00"), 0
+      while len(pax_text:=pax_text[entry_len:].lstrip()) > 0:
+        next_ow.update(dict([ pax_text[pax_text.index(" ")+1:(entry_len:=int(pax_text[:pax_text.index(" ")]))-1].split("=", 1) ]))
+
+    pos += info.size if info.size % 512 == 0 else info.size + (512 - info.size % 512)
```
geohot, line 163: Should it assert on unhandled type? Is that possible?

geohot:
A few nits, it's close

tinygrad/nn/state.py
```diff
@@ -132,17 +132,37 @@ def __init__(self):
...
-  with tarfile.open(fn, "r") as tar:
-    return {member.name:t[member.offset_data:member.offset_data+member.size] for member in tar if member.type == tarfile.REGTYPE}
+  # https://en.wikipedia.org/wiki/Tar_(computing)
+
+  pos, result, next_ow = 0, {}, cast(Dict[str, str], {})
+  if len(t.shape) != 1 or len(t) < 1024 or t.dtype != dtypes.uint8: raise tarfile.ReadError("Invalid tar file!")
+
+  while pos + 512 < len(t) and (header:=bytes(t[pos:pos+512].data())).count(b"\x00") != 512:
+    info, pos, ow, next_ow = tarfile.TarInfo.frombuf(header, "ascii", "strict"), pos + 512, next_ow, {}
+    for n, v in ((n, cfn(ow[n])) for n, cfn in [("path", str), ("size", int)] if n in ow): setattr(info, n, v)
```
geohot, line 150: I can't read this

geohot:
Please make this as readable as possible. Readability > line count. But also, don't use excessive lines

tinygrad/nn/state.py
```diff
@@ -132,17 +132,37 @@ def __init__(self):
...
+
+    if info.type == tarfile.REGTYPE: result[info.name] = t[pos:pos+info.size]
+
+    # handle GNU @LongLink
+    if info.type == b"L": next_ow = { "path": bytes(t[pos:pos+info.size].data()).decode().rstrip("\x00") }
```
geohot, line 155: Duplicated code with line 159

leopf:
So I basically discarded everything I wrote. Turns out you can just pass a `fileobj` to `tarfile.open`. To do this I have implemented `RawTensorIO`. The typing for the IO stuff is super messy, which is why there are ignores (see [https://github.com/python/typeshed/blob/main/stdlib/_io.pyi#L65](https://github.com/python/typeshed/blob/main/stdlib/_io.pyi#L65)). Sorry for not seeing this earlier.

I think this solution is very flexible. There are a 4 of lines that can be saved with this in `gguf_load` as well. As far as I can tell, it should also allow making `torch_load` Tensor only. The latency problems can be solved with `io.BufferedReader`.

geohot:
Clever :)

tinygrad/nn/state.py
```diff
@@ -1,11 +1,25 @@
...
 from tinygrad.multi import MultiLazyBuffer
 
+class RawTensorIO(io.RawIOBase, BinaryIO): # type: ignore[misc]  # incompatible definitions of methods in the base classes
+  def __init__(self, t: Tensor):
+    if len(t.shape) != 1 or t.dtype != dtypes.uint8: raise ValueError("Invalid Tensor, expected 1 dimensional tensor with dtype=uint8")
+    self._position, self._tensor = 0, t
+  def readable(self) -> bool: return True
+  def readinto(self, buffer: Union[bytearray, memoryview]) -> int:  # type: ignore[override]  # expects Buffer, which is not always writable
+    data = self._tensor[self._position:self._position+len(buffer)].data()
+    buffer[:len(data)], self._position = data, self._position + len(data)
```
geohot, line 16: don't put this on one line, it's harder to read

leopf: Will split into two lines.

leopf:
The typing issue is tricky. I was able to remove the `ignore[misc]` and `ignore[override]` by switching over to `io.IOBase`, but it comes at the cost of 3 useless function definitions to fully implement BinaryIO (in case of `write`) and fix the typing (for `writelines` and `__enter__`). Let me know what you prefer.

leopf:
The typing issue is tricky. I was able to remove the `ignore[misc]` and `ignore[override]` by switching over to `io.IOBase`, but it comes at the cost of 3 useless function definitions to fully implement BinaryIO (in case of `write`) and fix the typing (for `writelines` and `__enter__`). Must use `Any` because `io` expects a `Buffer`, which is not in `collections.abc` until python 3.12. Let me know what you prefer.

And line count...

leopf:
The typing issue is tricky. I was able to remove the `ignore[misc]` and `ignore[override]`, but it comes at the cost of 6 lines: 
- 3 useless function definitions to fix the typing (`write`, `writelines`, `__enter__`)
- `None` check on read to make it match `BinaryIO`

Must use `Any` for `readinto`, `write`, `writelines` because `io` expects a `Buffer`, which is not in `collections.abc` until python 3.12.

If using just `IOBase` (instead of `RawIOBase`) it is a little (3 lines) shorter but then `BufferedReader` does not work.

Let me know what you prefer.

And line count...

tinygrad/nn/state.py
```diff
@@ -1,11 +1,36 @@
...
+import os, json, pathlib, zipfile, pickle, tarfile, struct, functools, io
+from typing import Dict, Union, List, Optional, Any, Tuple, Callable, BinaryIO, Iterable
 from tinygrad.tensor import Tensor
 from tinygrad.dtype import dtypes
 from tinygrad.helpers import prod, argsort, DEBUG, Timing, CI, unwrap, GlobalCounters, tqdm
 from tinygrad.shape.view import strides_for_shape
 from tinygrad.multi import MultiLazyBuffer
 
+class TensorIO(io.RawIOBase, BinaryIO):
+  readable, seekable = (lambda _: True,) * 2
```
geohot, line 10: Write functions!

PR #7858, opened 2024-11-23
remove vectorized define_acc in ptx render [pr]

mesozoic-egg:
Ops.DEFINE_ACC was created in `lower_reduce_axis`, which sets the dtype for DEFINE_ACC with that of Ops.REDUCE_AXIS. REDUCE_AXIS was created in `_reduce_op` in lazy.py, which sets its dtype as with that of the lazybuffer. Vectorized dtype in lazy buffer is not valid. 

Some additional testing I did was raising exception inside the vectorized define_acc render (not in this PR), and ran beautiful_mnist and handcoded_opt, saw no errors


mesozoic-egg:
Let me know if there's more I should do to verify this is actually dead code

geohot:
If [pr] passes, it's almost definitely okay, merged!

PR #7861, opened 2024-11-23
move swizzle upats to ops, prereq for swizzle tc [pr]

Qazalin:


Qazalin:
@ignaciosica what's the status on the bounty?

ignaciosica:
I’m working on removing `expanded_shape` and the two reshapes in `fix_st` (ignaciosica/tinygrad#83). To be honest, expanded_shape always felt like a bit of a hack. I’ve managed to remove it from Metal, CUDA, and Intel backends, but I’m currently stuck with AMD Tensor Cores (TC). I don’t fully understand AMD’s TC shape and swizzle yet, which is complicating progress.

I believe this is the first step toward properly handling TC shapes with swizzling. Though, While trying to fix Metal TC after removing expanded_shape and the reshapes, I’m still facing challenges with merging views. However, its much closer to a solution than before.

Qazalin:
sounds good, feel free to draft a PR when you're ready.

PR #7865, opened 2024-11-23
track buffer numbers on UOp [pr]

Qazalin:
makes diffing the big graph with #7801 much easier. That -1 doesn't make sense anyways.


Qazalin:
huh
image

PR #7866, opened 2024-11-23
prepare mypy==1.13.0: legacy cast

JakobSpahn:
Use a helper for better readability compared to #7846. Two different calls to a `cast` in one line is confusing.


JakobSpahn:
Other than throwing `# type: ignore` I don't see how we can avoid narrowing the str to the literal types of the [format parameters in the stub](https://github.com/python/mypy/blob/v1.13.0/mypy/typeshed/stdlib/builtins.pyi#L769).

tinygrad/helpers.py
```diff
@@ -280,6 +280,12 @@ class CStruct(ctypes.Structure):
 def init_c_var(ctypes_var, creat_cb): return (creat_cb(ctypes_var), ctypes_var)[1]
 def flat_mv(mv:memoryview): return mv if len(mv) == 0 else mv.cast("B", shape=(mv.nbytes,))
 
+# TODO typeshed stub is missing the stub for memoryview float16 cast (format: Literal['e', '@e']). Remove once support is added.
+# src: https://github.com/python/mypy/blob/v1.13.0/mypy/typeshed/stdlib/builtins.pyi#L769
+def to_legacy_fmt(fmt:str):
```
geohot, line 285: I don't think this is the right way to handle this. Think about what you want to do with this, not a noop helper

JakobSpahn:
Switched to narrowing the type via asserts instead of `typing.cast`. Is a few more lines than the other, but more readable imo.

tinygrad/runtime/ops_python.py
```diff
@@ -68,9 +69,11 @@ def __call__(self, *bufs, global_size:Tuple[int,int,int]=(1,1,1), local_size:Tup
         dl[i] = dtype
         if uop is Ops.DEFINE_GLOBAL:
           assert dtype.fmt is not None
+          if TYPE_CHECKING or sys.version_info < (3, 12): assert dtype.fmt != "e"
```
geohot, line 72: Don't copy paste, refactor to be clean

tinygrad/tensor.py
```diff
@@ -286,6 +287,7 @@ def item(self) -> ConstType:
     """
     assert self.dtype.fmt is not None, f"no fmt dtype for {self.dtype}"
     assert self.numel() == 1, "must have one element for item"
+    if TYPE_CHECKING or sys.version_info < (3, 12): assert self.dtype.fmt != "e"
```
geohot, line 290: Don't copy paste. This function should probably just call `data()`

geohot:
Close

geohot:
Why is this draft? Looks good

PR #7876, opened 2024-11-24
Ptx simplify render load [pr]

mesozoic-egg:
Pre-req is #7877 (diff should be clean once that one is merged)
~~This only changes the output because there's no leading space for the vectorized form, which expands into a tab for readability. I thought it's not worth the extra if statement~~
kept the extra space


mesozoic-egg:
i actually think the before version is more readable...

PR #7882, opened 2024-11-24
Improved mod folding

### LINKED ISSUE #6935: fold_unrolled_divs doesn't work on small expressions

S-Lykles:
This allows `mod_folding` to detect cases like:
`(3+3x[0-3])%4 -> 3-x[0-3]`
and 
`(9+9x[0-3]+9y[0-3])%10 -> 9-x[0-3]-y[0-3]`

The next step is to add/combine this with div folding to detect the same pattern with floor division, there is a lot of similarity between the two functions now. That would also resolve #6935.

### DIFF
--- a/test/unit/test_uop_symbolic.py
+++ b/test/unit/test_uop_symbolic.py
@@ -192,6 +192,17 @@ class TestSymbolic(unittest.TestCase):
   def test_mod_to_sub(self):
     self.helper_test_variable((1+Variable("a",1,2))%2, 0, 1, "(a+-1)")
 
+  def test_mod_congruence(self):
+    self.helper_test_variable((3+3*Variable("a",0,3))%4, 0, 3, "((a*-1)+3)")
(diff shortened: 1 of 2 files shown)


S-Lykles:
Ill reopen when I have fixed the fuzz failures

PR #7883, opened 2024-11-24
Fix missing final rewrite in viz

nib9888:
In `_replace_uop` in `viz/serve.py`, master assumes rewriting `base` requires only rewriting its `src`s, however this is not true with matches from `(UPat(Ops.SINK, name="root")` and similar.

Fixes visible EXPAND nodes after the final expander rewrite in `uopgraph.py`.

### DIFF
--- a/tinygrad/viz/serve.py
+++ b/tinygrad/viz/serve.py
@@ -68,7 +68,10 @@ def uop_to_json(x:UOp) -> Dict[int, Tuple[str, str, List[int], str, str]]:
   return graph
 def _replace_uop(base:UOp, replaces:Dict[UOp, UOp]) -> UOp:
   if (found:=replaces.get(base)) is not None: return found
-  replaces[base] = ret = base.replace(src=tuple(_replace_uop(x, replaces) for x in base.src))
+  ret = base.replace(src=tuple(_replace_uop(x, replaces) for x in base.src))
+  if (final := replaces.get(ret)) is not None:
+      return final
+  replaces[base] = ret
   return ret
 @functools.lru_cache(None)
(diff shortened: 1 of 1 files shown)


Qazalin:
looks fixed, thanks!
image

image

wanna add a test for this case in `test/test_viz.py`? I'd try to pick a simple pattern matcher with this kind of rhs, doesn't have to be EXPAND specifically.

nib9888:
@Qazalin Did some more digging, and while this PR does fix an issue, it was a bit deeper than I thought and still don't fully understand what was wrong (so I don't know if it's fully fixed or if this is just a band-aid).

Repro: Checkout a commit before this PR, and look at rewrite 75 in `uopgraph.py:499` for kernel `r_64_2_49_8_16_16_4_4_4` in `DEBUG=2 VIZ=1 KERNEL=2 python3 examples/handcode_opt.py` -- the graph doesn't match the diff of the rewrite that should have occured.

I looked through what was going on and for some reason, while the displayed graph for rewrite 74 matched its diff, in rewrite 75 the `UOp(Ops.CAST, src=...)` (found by recursing into `src`s from `UOp(Ops.SINK, src=...)` in `_rewrite_uop`) that should have been found in `replaces` was different to what was recorded and so `found` was `None`. `_rewrite_uop` then continues to recurse and somehow catches up on some missed rewrites or something, meaning rewrite 76 and onwards just lag behind by one I think.

Basically `sink` in `rewrite_details` somehow ends up out of sync with the rewrites it should have been updated with, leading to the observed missing final rewrite.

Don't really have motivation to continue on this (limited understanding of tinygrad as whole makes things very slow going) but I couldn't leave this as it was

PR #7884, opened 2024-11-24
[BUGFIX] Tensor([]).data()

leopf:
Right now `memoryview` raises an error if `data` is called on a Tensor with `0 in Tensor.shape`. This is because `memoryview` does not allow casting to a shape that includes a 0.

```
TypeError: memoryview: cannot cast view with zeros in shape or strides
```

I have added a test for this and a fix for `Tensor.shape == (0,)`.

There still is a problem with `Tensor([]).reshape(0, 2).data()`. Since casting is not allowed I do not see how to make the shapes match in that case.

### DIFF
--- a/test/test_tensor_data.py
+++ b/test/test_tensor_data.py
@@ -24,6 +24,13 @@ class TestTensorData(unittest.TestCase):
     assert dat[0] == 1
     assert dat[1] == 2
 
+  def test_data_empty(self):
+    a = Tensor([], dtype=dtypes.int32)
+    dat = a.data()
+    assert dat.itemsize == 4
+    assert list(dat) == []
+    assert dat.shape == (0,)
+
   def test_data_uint8(self):
     a = Tensor([1,2,3,4], dtype=dtypes.uint8)
     dat = a.data()
--- a/test/test_tensor_data.py
+++ b/test/test_tensor_data.py
@@ -31,6 +31,13 @@ class TestTensorData(unittest.TestCase):
     assert list(dat) == []
     assert dat.shape == (0,)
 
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -273,7 +273,7 @@ def data(self) -> memoryview:
     """
     assert self.dtype.fmt is not None, f"no fmt dtype for {self.dtype}"
     assert all_int(self.shape), f"no data if shape is symbolic, {self.shape=}"
-    return self._data().cast(self.dtype.fmt, self.shape)
+    return self._data().cast(self.dtype.fmt) if self.shape == (0,) else self._data().cast(self.dtype.fmt, self.shape)
```
chenyuxyz, line 276: i think it's fine to say size 0 always return a shape (0,) memoryview, because memoryview cannot have other shape that has 0 in it. And this if should combined with `if 0 in self.shape` in `_data()`

leopf: I have reverted the previous commit. Not sure what exactly you mean by "this if should combined with `if 0 in self.shape` in `_data()`".

test/test_tensor_data.py
```diff
@@ -24,6 +24,13 @@ def test_data(self):
     assert dat[0] == 1
     assert dat[1] == 2
 
+  def test_data_empty(self):
+    a = Tensor([], dtype=dtypes.int32)
+    dat = a.data()
+    assert dat.itemsize == 4
+    assert list(dat) == []
+    assert dat.shape == (0,)
```
chenyuxyz, line 32: add a case for reshape into (0, 2) that you described

PR #7890, opened 2024-11-25
PTX simplify: use a dict matcher for prefix [pr]

mesozoic-egg:


tinygrad/renderer/ptx.py
```diff
@@ -176,35 +176,27 @@ def ssa(prefix:str, u:Optional[UOp]=None, dtype:Optional[str]=None) -> str:
...
-        bufs.append((f"data{u.arg}", u.dtype))
-        r[u] = ssa('dat', u, self.types[dtypes.ulong if u.dtype.__class__ == PtrDType else u.dtype])
+      elif u.op is Ops.DEFINE_GLOBAL: bufs.append((f"data{u.arg}", u.dtype))
       elif u.op is Ops.WMMA:
         self.wmma_r = [ssa("wmma", dtype="b32") for vv in u.src[:2] for i in range(0, len(r[vv]), 2)]
         r[u] = [ssa("wmma", dtype=self.types[u.dtype.scalar()]) for _ in range(u.dtype.count)]
+      prefix, *_dtype = {Ops.CAST: ("cast",), Ops.BITCAST: ("cast",), Ops.ENDRANGE: ("pred", "pred"), Ops.RANGE: ("ridx",), Ops.DEFINE_ACC: ("acc",),
+        Ops.DEFINE_VAR: ("dat",), Ops.CONST: ("const",), Ops.DEFINE_LOCAL: ("local", "u64"),
+        Ops.DEFINE_GLOBAL: ("dat", lambda u: self.types[dtypes.ulong if u.dtype.__class__ == PtrDType else u.dtype])}.get(u.op, (None,))
+      if prefix is None: prefix = "alu" if u.op in GroupOp.ALU else None
```
geohot, line 195: You don't need this, add it to the dict with *GroupOp.ALU

tinygrad/renderer/ptx.py
```diff
@@ -176,35 +176,27 @@ def ssa(prefix:str, u:Optional[UOp]=None, dtype:Optional[str]=None) -> str:
...
+      if prefix: r[u] = ssa(cast(str, prefix), u, cast(str, _dtype[0](u) if _dtype and callable(_dtype[0]) else _dtype[0] if _dtype else None))
```
geohot, line 196: I really don't like the use of callable here

mesozoic-egg:
So I removed the condiitonal since dtype is always ptr for DEFINE_GLOBAL and DEFINE_LOCAL. i think dtype itself is still needed though. The dtype for a register is either 1) the dtype of the uop, 2) "pred" for boolean or 3) "u64" for pointer. I can either put this in `def ssa` or keep them in the dictionary. Keeping in the dict seems cleaner to me

mesozoic-egg:
I think the dtype variable itself is still needed though. The special case to handle is predicate for "ENDRANGE" and u64 for the pointer. If I the dictionary returns only the prefix, then I need something like:
```python
      dtype = "pred" if prefix == "pred" else self.types[dtypes.ulong] if isinstance(u.dtype, PtrDType) else None
```
I think having the dictionary return a tuple: `('dat', self.types[dtypes.ulong])`, `('const', None)` etc. is  more readable

geohot:
For DEFINE_GLOBAL and DEFINE_LOCAL, why is the type not just the PtrType, and all PtrDTypes are rendered as ulongs?

geohot:
I'll merge this and consider the bounty complete though! The ENDRANGE refactor is probably beyond the scope of this, and the code as is is now readable. Can pay via PayPal or USDC on ETH

mesozoic-egg:
Thx! It's not immediately clear to me how removal of dtype would be done, but maybe after I make more contributions it would become more obvious. Will open a PR if some ideas come along

PR #7892, opened 2024-11-25
second try at block linearize

geohot:


geohot:
Just lines left I think. Curious how perf of generated kernels compares

geohot:
Going to merge if tests pass. This add lines sadly, but it's a lot cleaner

PR #7897, opened 2024-11-25
Remove tensor core's expanded_shape and reshapes

ignaciosica:
This is a draft pr to show progress. Changes in CI are just to be able to test the other backends while amd is still broken. Will remove those before marking as ready for review.

> I’m working on removing expanded_shape and the two reshapes in fix_st. To be honest, expanded_shape always felt like a bit of a hack. I’ve managed to remove it from Metal, CUDA, and Intel backends, but I’m currently stuck with AMD Tensor Cores (TC). I don’t fully understand AMD’s TC shape and swizzle yet to be honest.
>
> I believe this is the first step toward properly handling TC shapes with swizzling. Though, While trying to fix Metal TC after removing expanded_shape and the reshapes, I’m still facing challenges with merging views. However, it's much closer to a solution than before.

(this is quoted as its the progress I reported to Qazalin on saturday)

I'm currently having trouble with amd tc as they are the only ones whose expanded_shape reshaped the upcasted axes. In the other cases the expanded_shaped was used to reshape the reduce axes so what I did was use the expanded_shape as the true shape for the reduce avoiding also both reshapes required in `fix_st`. 

Although I managed some progress and now a true swizzle TC is more viable without those hidden reshapes, there is still one point that does not convince me about a true swizzle in this particular case. In the following paragraph I'll try to explain it as clearly as possible.

All shapetracker transformations performed in kernel.py (opts) are basically reshapes and permutes. This transformation are performed over the shapetrackers themselves and are accumulated over time (over different opts) resulting in a final view for the buffer. I believe the swizzle for the tc is from the same nature of the other transformations, nonetheless, is performed later and, if done through a true swizzle (merging views) the shapetracker not only does the permute but also adds the previous views, kind of messing the resulting view. (All other transformation in opts are performed directly)

I'm yet to understand view addition and indexed_uop. Maybe once I better grasp those the true swizzle for tc becomes more trivial.


ignaciosica:
Closing in favor of #8292

PR #7899, opened 2024-11-25
hcq replace update with sint

nimlgen:


nimlgen:
even faster enqueue, need to cleanup

llama 8b green
old:
```
enqueue in   2.66 ms
total  26.42 ms, 37.85 tok/s, 1126.18 GB/s, param 609.21 GB/s
```

new:
```
enqueue in   2.41 ms
total  26.59 ms, 37.61 tok/s, 1118.96 GB/s, param 605.30 GB/s
```

nimlgen:
This pr removes update function from hcq queues. It's a good start, still some things like args state and to graph's symbolic logic to cleanup

PR #7900, opened 2024-11-25
alu(c?t0:f0, c?t1:f1) -> c?alu(t0,t1):alu(f0,f1)

chenyuxyz:
only do if at least one branch is const, so total alu won't increase


chenyuxyz:
can do more with gated load too

PR #7903, opened 2024-11-25
Revert "move attention upcast"

chenyuxyz:
Reverts tinygrad/tinygrad#7830


chenyuxyz:
revert because this broke `python examples/sdv2.py --fp16`

PR #7904, opened 2024-11-26
test metal alloc ulong

mesozoic-egg:
Just adding a test for the fix in #7902 , without which the failure looks like: 
image


geohot:
I'm very reluctant to have tests that allocate large amounts of RAM

mesozoic-egg:
Right, let me try coming up with something better

PR #7905, opened 2024-11-26
match lazy view in uop try 2

Qazalin:
scheduler process replay:
image

how is _view movement ops different from swizzle upats? I think it shouldn't be.


Qazalin:
weird I can start CI with workflow_dispatch but doesn't trigger on PR
https://github.com/tinygrad/tinygrad/actions/runs/12027450863/job/33528518304

PR #7912, opened 2024-11-26
uop global buf number tracking try 2 [pr]

Qazalin:


Qazalin:
really don't like that llama unjitted is 30% slower
image

Qazalin:
so if I just keep track of state, it's a little bit slower, where is the 30% coming from?
image

Qazalin:
meh, global state is expensive
image
image

PR #7916, opened 2024-11-26
Factoring gcd out of mod

S-Lykles:
Curious if this will be faster/better, I just want to see the process replay on this
In theory this should lead to smaller intermediate calculations and it makes it more likely for the mod to be a power of 2


chenyuxyz:
we used to have this, see https://github.com/tinygrad/tinygrad/pull/6566. but i don't quite remember why i undo it, maybe it's fine now

S-Lykles:
One advantage is that this gives a lower vmax

chenyuxyz:
yea i think the generated code would be simpler too, this seems to be different from the old approach, as `test_div_into_mod` did not change. i can push this to benchmark and merge if all looks good

S-Lykles:
Cool, Ill mark it ready then:)

PR #7923, opened 2024-11-27
amd pkt3 refactor

nimlgen:


nimlgen:
IMO pkt3 now more readable, it's a bit slower queue to generate, so speed_v_theoretical becomes flaky (since no hcq graph there is used and queue is regenerated each time). So draft for now

geohot:
So I debate if this is more readable. It adds an intermediate function, and now you have to read more in order to understand the full thing.

Definitely support adding the constants from headers though instead of `mem_data_sel=3, mem_int_sel=0`

PR #7925, opened 2024-11-27
add TODO case to test_mod_congruence

chenyuxyz:
same alu count but better bounds


chenyuxyz:
@S-Lykles fyi in case you are interested

PR #7926, opened 2024-11-27
Don't take const in gcd and change the "nothing_changed" condition

S-Lykles:
Biggest difference is probably actually that I forgot to check if gcd changed if nothing else changed
The #7925 todo was fixed by not using the const in the gcd, and then taking it out.


S-Lykles:
Sorry I thought I updated all tests...

chenyuxyz:
not sure if the few ms resnet speed slowdown is real, will test more after merging but i think the pr itself is good. thanks!

chenyuxyz:
nvm resnet is also slightly faster, i was looking at the outdated stats page

PR #7934, opened 2024-11-28
unused block ops

geohot:


Qazalin:
oh sorry for the force push, I think PRs shouldn't base on update_benchmark though.

PR #7940, opened 2024-11-28
late folding size 0 ops

Qazalin:
this but upats https://github.com/tinygrad/tinygrad/pull/2556.
eg. test_zero_size rewrites to Ops.NOOP
image
image


Qazalin:
green but regresses, is this python time?
image
image

Qazalin:
extra VALIDs realized in llama because of this. This diff silently increases llama and gpt2 kernel count.
image
time to add the tests for it.

Qazalin:
image

Qazalin:
kk there's an untested case in llama gen 1 JITBEAM=1, 2460 vs 2457.

PR #7945, opened 2024-11-28
BLOCK ops in Ops

chenyuxyz:
did this break conv speed?


chenyuxyz:
so this is fine, will do if flipping next

PR #7951, opened 2024-11-29
true tc swizzle [pr]

ignaciosica:
Unifying view merge

I can't believe how small the diff ended up. The st needs to start contiguous so that the view addition doesn't get messed up. 

I believe it is still worth removing expanded_shape and cleaning up other things from tc as well (going to keep working on those).


Qazalin:
cool! I'll merge if benchmarks pass.
btw, I think you can rewrite def fix_st to be simpler, it should take a shape, and be something along the lines of get_tc_swizzle_st.

Qazalin:
looks alright
image

PR #7954, opened 2024-11-29
Fix atomic store in WebGPU

wpmed92:
Fixes the previously failing `test_jit_assign_int8`
Instead of just `atomicAdd`-ing the packed data into position, we first clear the position by `atomicAnd`-ing with a mask, then perform the `atomicAdd`

Screenshot 2024-11-29 at 10 28 42


tinygrad/renderer/wgsl.py
```diff
@@ -73,12 +76,14 @@ class WGSLRenderer(CStyleLanguage):
...
-    (UPat(Ops.STORE, src=(UPat.var('b'), UPat.var("v")), allow_any_len=True),\
-     lambda ctx,b,v: f"atomicAdd(&{ctx[b]}, {ctx[v]});" if b.src[0].dtype.itemsize < 4 else f"{ctx[b]} = {ctx[v]};"),
+    (UPat(Ops.STORE, src=(UPat.var('b'), UPat.var("v"))),lambda ctx,b,v:\
+     # (load & mask) | var -> mask = v.src[0].src[1], var = v.src[1]
+     ctx.render_atomic_store(ctx[b], ctx[v.src[0].src[1]], ctx[v.src[1]]) if b.src[0].dtype.itemsize < 4 else f"{ctx[b]} = {ctx[v]};"),
     # fix nan check: 'a != a -> is_nan()'
     (UPat.var("a") != UPat.var("a"), lambda ctx,a: f"is_nan({ctx[a]})"),
   ]) + base_rewrite
 
+  def render_atomic_store(self, buf, mask, var) -> str: return f"atomicAnd(&{buf},{mask});\natomicAdd(&{buf},{var});"
```
geohot, line 86: Why is this a function if it's only called in one place?

geohot:
Nice!

PR #7956, opened 2024-11-29
conceptual uop st cleanup [pr]

Qazalin:
prereq for attaching CONST to a shape and device. The Tensor API expects CONST to have a device. It's still a "meta op". It really should be a flat CONST.reshape(shape)._copy(device), on milestone 4's cleanup list.


Qazalin:
setitem ASSIGN isn't considered a Buffer op, maybe it should be. Just replace STORE with a swizzled ASSIGN, why parent a STORE with ASSIGN and give ASSIGN an arg?

PR #7957, opened 2024-11-29
Div mod recombine test for issue

S-Lykles:
Added some test cases for div mod folding which are currently not simplifying


S-Lykles:
For #7958

PR #7962, opened 2024-11-30
QwQ-32B-Preview support

francislata:
## Overview
This PR adds support for [QwQ-32B-Preview](https://huggingface.co/Qwen/QwQ-32B-Preview). Currently, it works with a basic prompt starting like **Hello.**:
Screenshot 2024-11-29 at 19 13 59

Command used to run script: `PYTHONPATH=. python3 examples/qwq.py`.

**Note about `convert_from_huggingface` changes**
I noticed that there was no need to permute `q_proj` and `k_proj` as this was only applicable for llama weights (#2228). In my testing, if this was applied, it will produce wrong values.

## Tasks
I've listed the tasks below that will need to be done before it is ready:
- [x] Dig into why `convert_from_huggingface`'s `permute` calls are not necessary for this model
- [x] Add support for custom prompt and being able to define the weights location


francislata:
I'll compare the outputs against Huggingface with `temperature=0`.

examples/qwq.py
```diff
@@ -0,0 +1,106 @@
...
+    "total_num_weights": 17,
+    "tokenizer": "Qwen/QwQ-32B-Preview"
+  }
+}
+
+def download_weights(total_num_weights):
+  model = fetch("https://huggingface.co/Qwen/QwQ-32B-Preview/resolve/main/model.safetensors.index.json?download=true", "model.safetensors.index.json", subdir="qwq_32b_preview")
+
+  for i in range(1, total_num_weights + 1):
+    filename = f"model-{str(i).zfill(5)}-of-{str(total_num_weights).zfill(5)}.safetensors"
```
chenyuxyz, line 26: just `{i:05d}`

examples/qwq.py
```diff
@@ -0,0 +1,106 @@
...
+
+if __name__ == "__main__":
+  Tensor.no_grad = True
+
+  parser = argparse.ArgumentParser(description="Run QwQ in tinygrad", formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+  parser.add_argument("--size", choices=["32B"], default="32B", help="Model size")
+  parser.add_argument("--count", type=int, default=30, help="Max number of tokens to generate")
+  parser.add_argument("--temperature", type=float, default=0.7, help="Temperature in the softmax")
+  parser.add_argument("--prompt", type=str, default="Hello.", help="Phrase to start with")
+  parser.add_argument("--weights", type=str, default=None, help="Path to the downloaded weights")
```
chenyuxyz, line 61: can you add a timing option similar to llama too?

chenyuxyz (commented):
bounty locked

can you add a validation case that has a slightly longer prompt and longer output?

francislata:
some updates from my last set of change:
1) added `--timing` option.

here's an example with `DEBUG=2`:
Screenshot 2024-12-03 at 05 47 38

and here's an example without `DEBUG=2`:
Screenshot 2024-12-03 at 05 49 37

2) added another longer validation case:
prompt is: "Can you tell me more about machine learning?"
with answer (number of tokens set to `50`): "Sure, I'd be happy to help! Machine learning is a subset of artificial intelligence that focuses on building systems that can learn from data and make predictions or decisions without being explicitly programmed to do so. It's a fascinating field with a lot of real"

here's a run of the new validation case:
Screenshot 2024-12-03 at 06 04 51

chenyuxyz:
cool, i will test later today. does the validation outputs match Huggingface?

francislata:
oh i haven’t done that yet. HF seems to complain when i tried with temperature is set to 0. but i’ll give it another look later today.

chenyuxyz:
it might be accumulated precision errors. if the first few tokens are always the same (with and without JIT), i think it's fine. the output is coherent enough that it cannot be very wrong

francislata:
i see. i tested it with and without JIT and i get the same result.

regarding the output for the longer prompt: **Can you tell me more about machine learning?**, i get similar tokens in the beginning and somewhere in the middle too:

from tinygrad:
`Can you tell me more about machine learning? Sure, I'd be happy to help! Machine learning is a subset of artificial intelligence that focuses on building systems that can learn from data and make predictions or decisions without being explicitly programmed to do so. It's a fascinating field with a lot of real`

from huggingface:
`Can you tell me more about machine learning? Sure! Machine learning is a subset of artificial intelligence that involves training algorithms to make predictions or decisions based on data. It's all about building models that can learn patterns from data and then use those patterns to make predictions on new, unseen data.\nThat`

also, the results produced are still coherent.

here's what i exactly ran to test the huggingface code:
Screenshot 2024-12-04 at 21 26 27

chenyuxyz:
cool, i got about 7 tok/s with JITBEAM=2 on M4 Max and the same validation output
image

chenyuxyz:
bounty is yours! Can pay via paypal or USDC, email [george@tinygrad.org](mailto:george@tinygrad.org) to claim

PR #7964, opened 2024-11-30
swizzle store [pr]

Qazalin:
fine to swizzle store within kernel boundaries, you're right @ignaciosica ASSIGN is already kinda doing it.


ignaciosica:
Great, thanks! By kernel boundaries we basically mean shape right?

Qazalin:
yea for example no expands,
be careful with reshape not creating implicit movement ops (like the ASSIGN but in this diff).
changing things like offset is allowed too.

PR #7973, opened 2024-11-30
fp16 conv2d numerical inaccuracy

geohotstan:
This is a blocker for true fp16 usage in onnx.
This inaccuracy is also exacerbated when the output is used as input for other ops

for instance this conv->elu->conv->... at the start of openpilot (output of elu is used as input for the next conv)
```
# 0: op "Conv" input shapes [(1, 12, 128, 256), (32, 12, 3, 3), (32,)] opt {'dilations': (1, 1), 'group': 1, 'kernel_shape': (3, 3), 'pads': (1, 1, 1, 1), 'strides': (2, 2)}
# 1: op "Elu" input shapes [(1, 32, 64, 128)] opt {'alpha': 1.0}
# 2: op "Conv" input shapes [(1, 32, 64, 128), (32, 1, 3, 3), (32,)] opt {'dilations': (1, 1), 'group': 32, 'kernel_shape': (3, 3), 'pads': (1, 1, 1, 1), 'strides': (1, 1)}
# 3: op "Elu" input shapes [(1, 32, 64, 128)] opt {'alpha': 1.0}
# 4: op "Conv" input shapes [(1, 32, 64, 128), (16, 32, 1, 1), (16,)] opt {'dilations': (1, 1), 'group': 1, 'kernel_shape': (1, 1), 'pads': (0, 0, 0, 0), 'strides': (1, 1)}
# 5: op "Conv" input shapes [(1, 16, 64, 128), (16, 1, 3, 3), (16,)] opt {'dilations': (1, 1), 'group': 16, 'kernel_shape': (3, 3), 'pads': (1, 1, 1, 1), 'strides': (1, 1)}
# 6: op "Elu" input shapes [(1, 16, 64, 128)] opt {'alpha': 1.0}
# 7: op "Conv" input shapes [(1, 16, 64, 128), (16, 16, 1, 1), (16,)] opt {'dilations': (1, 1), 'group': 1, 'kernel_shape': (1, 1), 'pads': (0, 0, 0, 0), 'strides': (1, 1)}
...
```

caught this using a little verifier I wrote where I compare outputs to torch onnx for each op
for openpilot in particular, after the 7th conv op, `rtol=2e-3, atol=2e-3` (rtol atol for `external_model_benchmark.py`) is already violated.

not too sure whats the right thing to do here.

### DIFF
--- a/test/test_numerical_accuracy.py
+++ b/test/test_numerical_accuracy.py
@@ -6,9 +6,9 @@ from tinygrad.device import Device, is_dtype_supported
 
 class TestNumericalAccuracy(unittest.TestCase):
   # TODO: blocker for true onnx fp16
-  @unittest.skipIf(not is_dtype_supported(dtypes.half, Device.DEFAULT) or Device.DEFAULT == "CLANG", "non-cpu backends fail")
+  @unittest.skipIf(not is_dtype_supported(dtypes.half) or Device.DEFAULT == "CLANG", "non-clang backends fail")
   @unittest.expectedFailure
-  def test_conv2d_fp16(self):
+  def test_conv2d_fp16_default(self):
     x,w,b = np.random.randn(1, 12, 128, 128), np.random.randn(32, 12, 3, 3), np.random.randn(32)
     tiny_x, tiny_w, tiny_b = (Tensor(t, dtype=dtypes.half) for t in (x,w,b))
     torch_x, torch_w, torch_b = (torch.tensor(t, dtype=torch.half) for t in (x,w,b))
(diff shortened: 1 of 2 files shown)


test/test_ops.py
```diff
@@ -1956,6 +1956,14 @@ def test_padded_conv2d_bs1(self):
       lambda x,w: torch.nn.functional.conv2d(x,w,padding=padding).relu(),
       lambda x,w: Tensor.conv2d(x,w,padding=padding).relu())
 
+  # TODO: fix
+  @unittest.skipIf(CI and Device.DEFAULT in {"LLVM", "PTX", "NV"}, "crashes in CI")
```
chenyuxyz, line 1960: is_dtype_supported

chenyuxyz:
it might also be affected by hand_coded_optimization, i think these are better put into their own test file with more focus on numerical issues

test/test_numerical_accuracy.py
```diff
@@ -0,0 +1,27 @@
+import unittest
+import numpy as np
+import torch
+from tinygrad import Tensor, Context, dtypes
+from tinygrad.device import Device, is_dtype_supported
+
+class TestNumericalAccuracy(unittest.TestCase):
+  # TODO: blocker for true onnx fp16
+  @unittest.skipIf(not is_dtype_supported(dtypes.half, Device.DEFAULT) or Device.DEFAULT == "CLANG", "non-cpu backends fail")
```
chenyuxyz, line 9: `Device.DEFAULT` is default for is_dtype_supported, can drop to make it cleaner

chenyuxyz (commented):
how much of this is torch half being less accurate? also is torch on gpu different from torch on cpu (given that it pass for CLANG?)

i think it's less of an issue if tinygrad in float matches tinygrad in half with small tol, than tinygrad in half on gpu vs torch in half on cpu

geohotstan:
> is torch on gpu different from torch on cpu

Yeah it's about a `rtol=1e-3 atol=1e-3` difference (tested on colab)

> i think it's less of an issue if tinygrad in float matches tinygrad in half with small tol

it's `rtol=2e-2, atol=2e-2` big :(

> it might also be affected by hand_coded_optimization

Yeah there's also a smaller `rtol=2e-3 atol=2e-3` difference 


-------

inaccuracies will probably accumulate as sizes get bigger also.

geohotstan:
new way to test this coming soooooooon

PR #7974, opened 2024-11-30
update test for gated store

mesozoic-egg:
I think Ops.STORE in general is now handled by Ops.INDEX, so this should be the correct gated store uop test right?

Specifically,

STORE uop is from `lower_load_store`, which define it as `  return UOp(Ops.STORE, dtypes.void, (buf.index(idx, valid if has_valid else None), x.src[2]))` 

One place gate is added in the 2nd position of the src of store is in `move_mask`, where it does `  return UOp.load(nidx, x.const_like(0), mask, *x.src[1:], dtype=x.dtype) if x.op is Ops.LOAD else UOp.store(nidx, x.src[1], mask, *x.src[2:])`

Another place gate is added is in `create_gate`, where it does `      return UOp(u.op, u.dtype, u.src[:-1]+(UOp(Ops.IF, dtypes.void, (gate, u.src[-1])),), u.arg)`

So the overall STORE op is
```
UOp(Ops.STORE, src=(
  UOp(Ops.INDEX, src=(
    UOp(Ops.DEFINE_GLOBAL)
    UOp(Ops.SPECIAL)
  UOp()  <--- value to store
  UOp()  <--- optional gate
)
```

### DIFF
--- a/test/test_uops.py
+++ b/test/test_uops.py
@@ -251,16 +251,16 @@ class TestGatedStoreRewrite(unittest.TestCase):
   def test_tiny_gate_store(self):
     gmem = UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(), (), 0)
     gidx0 = UOp(Ops.SPECIAL, dtypes.int, (), ('gidx0', 4))
-    idx = gidx0 * UOp.const(dtypes.int, 2)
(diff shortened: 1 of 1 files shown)


Qazalin:
why is it still marked as expected failure?

mesozoic-egg:
I thought I would get some tests merged in first, but maybe i should put the changes in as well...

PR #7975, opened 2024-12-01
Remove gated store, put rewrite to uopgraph [pr]

mesozoic-egg:
work in progress, It seems like most of the work are already done, maybe just require some cleaning up and more tests


mesozoic-egg:
Process replay fails for PTX because of the intended removal of gated store, the rest passes
Screenshot 2024-12-01 at 19 12 29

I literally just moved the rewrite from cstyle/llvm into uopgraph, and put it after the two rewrite that generates UOps.Store. 

I would like to know if I should make the two rewrites (move_mask and create_gate) output the IF node directly so the gated store rewrite itself can be removed (as well as the uop test)

geohot:
Cool! It actually makes the PTX code worse...but I think that's okay

geohot:
Can pay via paypal or USDC, email george@tinygrad.org to claim

PR #7987, opened 2024-12-02
fix: private segment sgpr on gfx103x

wozeparrot:
was incorrect before. rdna 2, from my testing, fully works with this


nimlgen:
yep, this looks right now.

PR #7990, opened 2024-12-02
mypy==1.13.0

JakobSpahn:
Depends on https://github.com/tinygrad/tinygrad/pull/7863, https://github.com/tinygrad/tinygrad/pull/7866.
Finalizes the attempt of https://github.com/tinygrad/tinygrad/pull/7846.


tinygrad/tensor.py
```diff
@@ -281,7 +281,7 @@ def data(self) -> memoryview:
     assert self.dtype.fmt is not None, f"no fmt dtype for {self.dtype}"
     assert all_int(self.shape), f"no data if shape is symbolic, {self.shape=}"
     if TYPE_CHECKING or sys.version_info < (3, 12): assert self.dtype.fmt != "e"
-    return self._data().cast(self.dtype.fmt) if 0 in self.shape else self._data().cast(self.dtype.fmt, self.shape)
+    return cast(memoryview, self._data().cast(self.dtype.fmt) if 0 in self.shape else self._data().cast(self.dtype.fmt, self.shape))
```
JakobSpahn, line 284: Cast `memoryview[int] | memoryview[float] | memoryview[bool]` to `memoryview`.

geohot: Any reason to not make return type `memoryview[int] | memoryview[float] | memoryview[bool]`?

JakobSpahn: In the [typeshed stub](https://github.com/python/mypy/blob/v1.13.0/mypy/typeshed/stdlib/builtins.pyi#L739), `memoryview` is generically defined as:

```python
_I = TypeVar("_I", default=int)  
class memoryview(Sequence[_I]):
  ...

```
Which leads to the typechecker substituting every `memoryview` type decl with `memoryview[int]`.

Changing the return type will require to change the types of a lot of variables/function args from `memoryview` to the broader `memoryview[int] | memoryview[float] | memoryview[bool]`. For example all the `Allocator._copyin` implementations (and so on).

I can use a type alias such as `MemoryviewType = memoryview[int] | memoryview[float] | memoryview[bool]` but that will be very confusing with both `MemoryviewType` and `memoryview` scattered across the codebase.

Imo the cleanest solution is to return a `memoryview` or a `memoryview[Any]`

geohot: kk, cast is fine

geohot:
Is there a reason this is draft?

tinygrad/runtime/graph/hcq.py
```diff
@@ -54,6 +54,8 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
 
     for j,ji in enumerate(jit_cache):
       enqueue_dev = ji.prg.dev if (is_exec_prg:=isinstance(ji.prg, CompiledRunner)) else Device[ji.bufs[1].device] #type:ignore
+      assert isinstance(enqueue_dev, HCQCompiled)
+      assert (enqueue_dev.hw_copy_queue_t is not None) and issubclass(enqueue_dev.hw_compute_queue_t, HWQueue)
```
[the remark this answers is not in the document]
geohot: Still should be one line

JakobSpahn:
Changed to one line assert, should be ready now

tinygrad/runtime/graph/hcq.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
 
     for j,ji in enumerate(jit_cache):
       enqueue_dev = ji.prg.dev if (is_exec_prg:=isinstance(ji.prg, CompiledRunner)) else Device[ji.bufs[1].device] #type:ignore
+      assert isinstance(enqueue_dev, HCQCompiled) and issubclass(enqueue_dev.hw_compute_queue_t, HWQueue) and enqueue_dev.hw_copy_queue_t is not None
```
geohot, line 57: Err, how sure are you that this is right? What if it doesn't have a copy queue?

tinygrad/runtime/graph/hcq.py
```diff
@@ -54,7 +54,11 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
 
     for j,ji in enumerate(jit_cache):
       enqueue_dev = ji.prg.dev if (is_exec_prg:=isinstance(ji.prg, CompiledRunner)) else Device[ji.bufs[1].device] #type:ignore
```
JakobSpahn, line 56: Kinda hacky, but could save one line by removing the following assert and declaring `enqueue_dev: HCQCompiled`.

tinygrad/runtime/graph/hcq.py
JakobSpahn: Sorry I missed that QCOM had a graph

tinygrad/runtime/graph/hcq.py
```diff
@@ -54,7 +54,11 @@ def __init__(self, jit_cache: List[ExecItem], input_rawbuffers: List[Buffer], va
...
-      enqueue_queue = self.comp_queues[enqueue_dev] if is_exec_prg else self.copy_queues.setdefault(enqueue_dev, enqueue_dev.hw_copy_queue_t())
+      assert isinstance(enqueue_dev, HCQCompiled)
+      if is_exec_prg: enqueue_queue = self.comp_queues[enqueue_dev]
+      else:
+        assert issubclass(enqueue_dev.hw_compute_queue_t, HWQueue) and enqueue_dev.hw_copy_queue_t is not None
+        enqueue_queue = self.copy_queues.setdefault(enqueue_dev, enqueue_dev.hw_copy_queue_t())
```
geohot, line 61: This is unreadable and needs refactoring

JakobSpahn: Going to open up a new PR with the other changes and convert this one to draft.

JakobSpahn: Let me know if it is readable enough now.
Unfortunately, mypy doesn't have the path sensitivity for something with as few lines as possible like:
```python
if not is_exec_prg: assert (enqueue_dev.hw_copy_queue_t is not None)
enqueue_queue = self.comp_queues[enqueue_dev] if is_exec_prg else self.copy_queues.setdefault(enqueue_dev, enqueue_dev.hw_copy_queue_t())
```

geohot:
Assuming benchmark tests pass, the bounty is yours! E-mail george@tinygrad.org to claim, can pay via PayPal or USDC on ETH

PR #7992, opened 2024-12-02
only mutate the image Buffer [pr]

Qazalin:
`buf` is UOp in delete_lazy, UOp is immutable. This shouldn't impact correctness of the schedule.


Qazalin:
ew it's mutable?

PR #7993, opened 2024-12-02
add SELU to Tensor

geohotstan:
was prepping an onnx_ops clean up and found a missed op from onnx_ops

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -2688,7 +2688,7 @@ class Tensor(SimpleMathTrait):
     """
     return self.maximum(0) + (alpha * ((self / alpha).exp() - 1)).minimum(0)
 
-  def selu(self, alpha=1.67326, scale=1.0507):
+  def selu(self, alpha=1.67326, gamma=1.0507):
     """
     Applies the Scaled Exponential Linear Unit (SELU) function element-wise.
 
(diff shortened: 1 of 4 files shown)


tinygrad/tensor.py
```diff
@@ -2688,6 +2688,19 @@ def celu(self, alpha=1.0):
...
+    Applies the Scaled Exponential Linear Unit (SELU) function element-wise.
+
+    - Described: https://paperswithcode.com/method/selu
+    - Paper: https://arxiv.org/abs/1706.02515v5
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor([-3., -2., -1., 0., 1., 2., 3.]).softsign().numpy())
+    ```
+    """
+    return scale * (self.relu() - (-alpha * self.exp() + alpha).relu())
```
geohotstan, line 2702: `return scale * (self > 0).detach().where(self, alpha * (self.exp() - 1))`
which implementation is cleaner? ^ This one is slightly faster and looks more like the implementation in `https://paperswithcode.com/method/selu`

tinygrad/tensor.py
```diff
@@ -2688,6 +2688,19 @@ def celu(self, alpha=1.0):
     """
     return self.maximum(0) + (alpha * ((self / alpha).exp() - 1)).minimum(0)
 
+  def selu(self, alpha=1.67326, scale=1.0507):
```
chenyuxyz, line 2691: just use alpha and gamma? it matches onnx naming and one less indirection

tinygrad/tensor.py
chenyuxyz: yea i think `scale * (self >= 0).detach().where(self, alpha * (self.exp() - 1))` is more readable

tinygrad/tensor.py
```diff
@@ -2688,6 +2688,19 @@ def celu(self, alpha=1.0):
...
+    """
+    Applies the Scaled Exponential Linear Unit (SELU) function element-wise.
+
+    - Described: https://paperswithcode.com/method/selu
+    - Paper: https://arxiv.org/abs/1706.02515v5
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor([-3., -2., -1., 0., 1., 2., 3.]).softsign().numpy())
```
chenyuxyz, line 2699: softsign?

tinygrad/tensor.py
geohotstan: image

pytorch called it `scale` but I guess people just call `.selu()` anyways 
will change to gamma

PR #7996, opened 2024-12-03
Improved div folding

### LINKED ISSUE #6935: fold_unrolled_divs doesn't work on small expressions

### LINKED ISSUE #7958: Div mod recombine pattern not always matching
Some div mod recombine patterns are not always being folded properly at the moment. See #7957 for examples. 
For some reason `div_folding` and `mod_folding` are changing the `x` in the following pattern before this pattern can match.
```python
  (UPat.var("x")%UPat.cvar("c")+(UPat.var("x")//UPat.cvar("c"))*UPat.cvar("c"), lambda x,c: x), # (x%c)+(x//c)*c = x
```

I don't quite understand why because this pattern is higher in the list and I thought the matching was happening from the root.

S-Lykles:
Combined div and mod folding, this now also performs the same optimizations as #7887 for div folding.
This fixes  #6935 and also some cases of #7958
With `split_rem=True` this would fix #7958 fully as it makes the treatment of the remainder more consitent between mod and idiv. I added it as a flag to make it easier to play around with and see the difference. 

Im just fuzzing this a little a little longer before Ill mark as ready

### DIFF
--- a/test/unit/test_uop_symbolic.py
+++ b/test/unit/test_uop_symbolic.py
@@ -433,8 +433,7 @@ class TestSymbolic(unittest.TestCase):
   def test_div_mod_recombine_folded_mod(self):
     a = Variable("a", 0, 2)
     b = Variable("b", 0, 100)
-    with self.assertRaises(AssertionError):
-      self.helper_test_variable((31 * a + 1) % 30 + ((31 * a + 1) // 30) * 30, 1, 63, "((a*31)+1)")
+    self.helper_test_variable((31 * a + 1) % 30 + ((31 * a + 1) // 30) * 30, 1, 63, "((a*31)+1)")
     with self.assertRaises(AssertionError):
       self.helper_test_variable((31 * b + 1) % 18 + ((31 * b + 1) // 18) * 18, 1, 3101, "((b*31)+1)")
 
--- a/test/unit/test_uop_symbolic.py
+++ b/test/unit/test_uop_symbolic.py
@@ -205,6 +205,13 @@ class TestSymbolic(unittest.TestCase):
     self.helper_test_variable((7+9*Variable("x",0,2)+9*Variable("y",0,2)+Variable("z",0,2))%10, 3, 9, "(((z+(x*-1))+(y*-1))+7)")
     self.helper_test_variable((10+12*Variable("x",0,2)+Variable("y", 0, 4)%3)%13, 8, 12, "(((x*-1)+(y%3))+10)")
 
+  def test_div_congruence(self):
+    self.helper_test_variable((3+3*Variable("a",0,3))//4, 0, 3, "a")
+    self.helper_test_variable((18+17*Variable("a",0,2)+17)//18, 1, 3, "(a+1)")
+
+  def test_div_congruence_multiple_vars(self):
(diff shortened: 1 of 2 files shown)


S-Lykles:
Have fuzzed this to 50_000 cases

chenyuxyz:
really nice! yea combing div and mod is a good idea

i saw `((lidx0+2)>>2)` became `(((lidx0>>1)+1)>>1)` now, which might not be ideal but i think fine to include for now. can you add a test case for it?

tinygrad/ops.py
```diff
@@ -892,80 +892,63 @@ def split_uop(x:UOp, sep:Ops):
     for s in x.src: yield from split_uop(s, sep)
   else: yield x
 
-def mod_folding(x:UOp, c:int) -> Optional[UOp]:
-  # simplify x % c, None means no change
+def div_and_mod_folding(x: UOp, c: int, which: Literal[Ops.MOD, Ops.IDIV], split_rem: bool=False) -> Optional[UOp]:
+  # simplify x // c or x % c, None means no change, c must be > 0
+  assert c > 0
```
chenyuxyz, line 897: if we are asserting this, i think it's good to also assert which is either `Ops.MOD` or `Ops.IDIV`

test/unit/test_uop_symbolic.py
```diff
@@ -426,8 +433,7 @@ def test_div_mod_recombine(self):
   def test_div_mod_recombine_folded_mod(self):
     a = Variable("a", 0, 2)
     b = Variable("b", 0, 100)
-    with self.assertRaises(AssertionError):
-      self.helper_test_variable((31 * a + 1) % 30 + ((31 * a + 1) // 30) * 30, 1, 63, "((a*31)+1)")
+    self.helper_test_variable((31 * a + 1) % 30 + ((31 * a + 1) // 30) * 30, 1, 63, "((a*31)+1)")
     with self.assertRaises(AssertionError):
```
chenyuxyz, line 437: if you are interested, i can add a $200 bounty and lock to you to fully fix this divmod case.

tinygrad/ops.py
S-Lykles: I asserted `c>0` because it would not be a type error to input a negative `c`,  `which` is fully typed though so that should be caught by mypy

S-Lykles:
> i saw `((lidx0+2)>>2)` became `(((lidx0>>1)+1)>>1)` now, which might not be ideal but i think fine to include for now. can you add a test case for it?

So `((lidx0+2)//2)` does not directly get rewritten to `((lidx0//2)+1)//2`. I think what is happening is because some intermediate expression now get rewritten in a different order this somehow ends up like this. I might make a seperate pattern for `(x//c+a)//d -> (x+a*c)//(c*d)`

PR #7997, opened 2024-12-03
merge where alu when one branch is identity element const

chenyuxyz:
this is another case that does not increase alu count


chenyuxyz:
did not really hit, will keep this for later

PR #8003, opened 2024-12-03
simple onnx_ops cleanups

geohotstan:
These are some simple clean ups
There'll probably be a few rounds of these
Next round will be deleting all the `_format_padding`,  and `_padded` mess, but those will also get removed once we merge pool/ceil_mode.


test/external/external_test_onnx_backend.py
```diff
@@ -40,9 +40,9 @@ def supports_device(cls, device: str) -> bool:
 
 backend_test = onnx.backend.test.BackendTest(TinygradBackend, __name__)
 
-# TODO figure out why it's returning wrong values, geohotstan's uneducated guess is it's due to imprecision from float64 (double) -> float32
-# see Type Constraints: https://onnx.ai/onnx/operators/onnx_aionnxpreviewtraining_Adam.html#type-constraints
-backend_test.exclude('test_adam_multiple_cpu')
+backend_test.exclude('test_adam_*')
```
geohot, line 43: Can we use the optim.py optimizers for these ops?

PR #8004, opened 2024-12-03
test: late buffer ref (only if it's scheduled)

Qazalin:
Nothing above schedule, (eg. tensor/multi) should use/mutate these refs, right?


Qazalin:
stats are fine with it too
image.

This is only used in the memory planner and the JIT for buffer replaces.

It's required for delete_lazy JIT correctness.

PR #8005, opened 2024-12-03
const_like can return a valid [pr]

Qazalin:
prereq for delete_lazy, there won't be any DEVICECONST. 

CONST starts as a VIEW of a BUFFER. (lazy/uop creates it, tensor uses it)
VIEW(BUFFER, CONST) gets rewritten to VALID (schedule creates it, codegen uses it)
once it's indexed it collapses to a CONST. (codegen unchanged.)


Qazalin:
VIEW(BUFFER, CONST) always exists because tensor.py expects CONST to have device.
VALID (even unmasked) always exists because kernel.py expects CONST to have shape.
both can go away with a global, conceptual cleanup of CONST.

PR #8010, opened 2024-12-03
Fold nested div with const

S-Lykles:
Fold (x//c+a)//d -> (x+a*c)//(c*d)
Fix the point from #7996


S-Lykles:
So this is failing because of vectorized ops and ranges, have to look more at it. Its lines like `if u.op is Ops.CONST: const += f` that are not written properly for this

chenyuxyz:
`vec=False` to avoid vectorized const

S-Lykles:
Yeah I thought that but Upat.var doesn't have a vec keyword, and the cvar in the divisor is already a scalar

PR #8016, opened 2024-12-03
silent bug in op.st [pr]

ignaciosica:
WMMA also reduces axes

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -275,7 +275,7 @@ class UOp(MathTrait, metaclass=UOpMetaClass):
     assert all_same([x.shape for x in src_sts]), f"UOp parents must have the same shape {self} {[x.shape for x in src_sts]}"
     # all other ops have a contiguous shapetracker
     from tinygrad.shape.shapetracker import ShapeTracker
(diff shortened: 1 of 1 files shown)


chenyuxyz:
can you add regression test that would fail on master? otherwise it's hard to see why this is a bug
don't hard code check internal structures, but have behavior check that will prevent this from being wrong again

ignaciosica:
It's difficult right now to trigger a fail from the frontend. I'm working on swizzling the output of the tensor core and in that context many tests fail without this change.

When pushing the swizzled view to the store, it recursively swizzles the rest of the tree along the way, and without this changes it pushes the wrong view as the st for the WMMA won't be reduced.

[progress on swizzle tc store](https://github.com/ignaciosica/tinygrad/pull/91/files/155baf3ff639f767c3c475127da66f55fc099440..de4eea5c60ac4830c6e5629f380a2e6df7e392fa)

ignaciosica:
I agree it always better to add regression tests, let me think if I come up with one

chenyuxyz:
nah this is more correct and i can see why it's annoying to trigger. it's fine given that you are working on followups

PR #8028, opened 2024-12-04
x86 assembly backend

ttomsa:
It's 90% there, I think it's far enough I can take some feedback. There's a lot of possible optimizations but I went for close to the least amount of lines.


ttomsa:
@geohot why isn't this locked? Makes it seem like there's some massive issue with the pr. It's basically done just need to fix/remove float16 (accuracy issues in tests), cleanup idiv/mod, add some simple coalescing and that's it.

geohot:
I locked it for you, but want the tests to pass before review.

ttomsa:
@geohot this is ready for review

geohot:
Looking pretty good!
* Are all tests passing?
* How does the speed compare to CLANG and LLVM?

CatsAreFluffy:
uint64 to float conversion is incorrect for large inputs, eg:
```python

x = Tensor([1<<63],dtype=dtypes.uint64)
print(x.float().tolist())
```
I think replacing lines 119 and 120 with this would work: `lambda c: (c.src[0] >> 1).cast(dtypes.int64).cast(c.dtype) * 2 + (c.src[0] & 1).cast(dtypes.int32).cast(c.dtype)),`
Also, uint8/int8 mod is broken since it reads it from `rdx` instead of `ah`, and vec2 moves generate `movlps` between registers which is illegal.

geohot:
How does perf compare to O0? That's the blocker on this

ttomsa:
@geohot currently 1.2x slower than O0. This was just by adding shl/shr and simple register reuse. Still 3x slower than O1 though.

geohot:
How is speed looking on this?

saliba06:
> It's 90% there, I think it's far enough I can take some feedback. There's a lot of possible optimizations but I went for close to the least amount of lines.

kozuedoingregression (commented):
explaining compiler flags used

PR #8037, opened 2024-12-04
torch_load with Tensors

leopf:
`pickle`, `zipfile`, `tarfile` all support usage with IO instead of filename.

Tested by `test/unit/test_disk_tensor.py:TestTorchLoad`.

### DIFF
--- a/tinygrad/nn/state.py
+++ b/tinygrad/nn/state.py
@@ -1,4 +1,4 @@
-import os, json, pathlib, zipfile, pickle, tarfile, struct, functools, io
+import json, pathlib, zipfile, pickle, tarfile, struct, functools, io
 from typing import Dict, Union, List, Optional, Any, Tuple, Callable, BinaryIO, Iterable, TypeVar
 from tinygrad.tensor import Tensor
 from tinygrad.dtype import dtypes
@@ -177,7 +177,8 @@ def tar_extract(t: Tensor) -> Dict[str, Tensor]:
 
 # torch support!
 
-def torch_load(fn:Union[Tensor,str]) -> Dict[str, Tensor]:
+@accept_filename
+def torch_load(t:Tensor) -> Dict[str, Tensor]:
   """
   Loads a torch .pth file from disk.
 
@@ -185,10 +186,7 @@ def torch_load(fn:Union[Tensor,str]) -> Dict[str, Tensor]:
   state_dict = nn.state.torch_load("test.pth")
   ```
   """
-  t = fn if isinstance(fn, Tensor) else Tensor(pathlib.Path(fn))
   fobj = io.BufferedReader(TensorIO(t))
-
-  def passthrough_reset(v: bool): return fobj.seek(0, 0) or v
 
   offsets: Dict[Union[str, int], int] = {}
   lens: Dict[Union[str, int], int] = {}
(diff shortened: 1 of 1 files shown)


tinygrad/nn/state.py
```diff
@@ -170,15 +170,18 @@ def tar_extract(t: Tensor) -> Dict[str, Tensor]:
 
 # torch support!
 
-def torch_load(fn:str) -> Dict[str, Tensor]:
+def torch_load(fn:Union[Tensor,str]) -> Dict[str, Tensor]:
```
geohot, line 173: We shouldn't union, we should only support Tensor.

tinygrad/nn/state.py
```diff
@@ -220,17 +223,17 @@ def find_class(self, module, name):
...
     base_name = myzip.namelist()[0].split('/', 1)[0]
     for n in myzip.namelist():
       if n.startswith(f'{base_name}/data/'):
         with myzip.open(n) as myfile:
           offsets[n.split("/")[-1]] = myfile._orig_compress_start # type: ignore
     with myzip.open(f'{base_name}/data.pkl') as myfile:
       return TorchPickle(myfile).load()
-  elif tarfile.is_tarfile(fn):
-    with tarfile.open(fn, "r") as tar:
+  elif passthrough_reset(tarfile.is_tarfile(fobj)):
```
geohot, line 235: Err, do you want this reset to be on the open?

tinygrad/nn/state.py
geohot: Wrote decorator here https://github.com/tinygrad/tinygrad/pull/8049

tinygrad/nn/state.py
leopf: `passthrough_reset` is not required `tarfile.is_tarfile`, it resets correctly.

I made a mistake when testing this, I had failing tests without `passthrough_reset` `tarfile.is_tarfile` for some reason.

leopf:
I have removed `passthrough_reset` entirely and split up the `if` blocks, which have unconditional returns anyway. I have now inserted two `fobj.seek(0, 0)` at the required lines.

leopf:
I did some digging, I propose we keep `passthrough_reset` as it was.

Position keeping for `is_tarfile` has been introduced with python 3.11 ([https://github.com/python/cpython/pull/26488](https://github.com/python/cpython/pull/26488)).

Position keeping for `is_zipfile` will be introduced soon ([https://github.com/python/cpython/pull/122397](https://github.com/python/cpython/pull/122397)).

geohot:
You can keep passthrough_reset, but it shouldn't be on the if, it should be on the open, right? Also, scoping.

leopf:
Not sure I understand correctly. I have created `passthrough_reset` to reset after the file check to make sure it is reset for both cases `True` and `False`.

```python
def passthrough_reset(v: bool): return fobj.seek(0, 0) or v

# can be rewritten as

def passthrough_reset(v: bool):
  fobj.seek(0, 0) # returns new position (0)
  return v

```

Moving `passthrough_reset` to `tarfile.open` would break this for python < 3.11, because line 257 depends on `fobj` (to be at the start position) as well; An additional `passthrough_reset` is not required for `tarfile.open` because it is already reset after the condition.

---

Moved `passthrough_reset` and `fobj` down.

geohot:
Oh, I see. The is changes the seek ptr...okay LGTM

PR #8039, opened 2024-12-04
add _one_hot_along_dim helper for Tensor.arange masking

geohotstan:
my scatter got heat from twitter
😔😔😔😔😔😔😔😔😔😔😔😔😔😔😔

does adding a `_arange_select` function help readability?

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -2343,7 +2343,7 @@ class Tensor(SimpleMathTrait):
     src = src.unsqueeze(-1).expand(*src.shape, self.shape[dim]).transpose(-1, dim)
     mask = index.unsqueeze(-1)._one_hot_along_dim(self.shape[dim]).transpose(-1, dim)
     # pad src and mask to self.shape so that reduce can be done with padded values as no-ops
-    src, mask = (x.pad(tuple((0, self.shape[i] - x.shape[i]) if i != dim else None for i in range(self.ndim)) + (None,)) for x in (src, mask))
+    src, mask = (x.pad(tuple(None if i == dim or i == x.ndim - 1 else (0, self.shape[i] - x.shape[i]) for i in range(x.ndim))) for x in (src, mask))
     if reduce == "add": return mask.where(src, 0).sum(-1, acc_dtype=self.dtype) + self
     if reduce == "multiply": return mask.where(src, 1).prod(-1, acc_dtype=self.dtype) * self
     return _masked_setitem(self, src, mask, (-1,))
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -2340,7 +2340,7 @@ class Tensor(SimpleMathTrait):
     # shrink src to index shape to shrink away the unused values
     src = src.shrink(tuple((0,s) for s in index.shape))
     # prepare src and mask for reduce with respect to dim
-    src = src.unsqueeze(-1).expand(src.shape + (self.shape[dim],)).transpose(-1, dim)
+    src = src.unsqueeze(-1).expand(*src.shape, self.shape[dim]).transpose(-1, dim)
     mask = index.unsqueeze(-1)._one_hot_along_dim(self.shape[dim]).transpose(-1, dim)
     # pad src and mask to self.shape so that reduce can be done with padded values as no-ops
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -104,6 +104,9 @@ def _masked_setitem(target:Tensor, values:Tensor, mask:Tensor, axes:Tuple[int, .
   # select from values for each True element in mask else select from self
   return mask.where(values, target)
 
+def _arange_select(target:Tensor, arange_size:sint, target_dim:sint=0):
```
chenyuxyz, line 107: factoring out this is nice. also pick a function name that describes what it does, not how it's done. use of arange is implementation details and it would be good to separate function name from implementation (onehot_dim or similar, see if there are similar things in onnx or torch).

geohotstan:
There is also the example from https://docs.tinygrad.org/quickstart/#training

```python
def sparse_categorical_crossentropy(self, Y, ignore_index=-1) -> Tensor:
    loss_mask = Y != ignore_index
    y_counter = Tensor.arange(self.shape[-1], dtype=dtypes.int32, requires_grad=False, device=self.device).unsqueeze(0).expand(Y.numel(), self.shape[-1])
    y = ((y_counter == Y.flatten().reshape(-1, 1)).where(-1.0, 0) * loss_mask.reshape(-1, 1)).reshape(*Y.shape, self.shape[-1])
    return self.log_softmax().mul(y).sum() / loss_mask.sum()
```

I changed it to use _one_hot_along_dim in Tensor.py but I don't think having a helper function in quickstart is a good idea... or maybe not? Should I change this too?

tinygrad/tensor.py
```diff
@@ -2336,11 +2335,15 @@ def scatter(self, dim:int, index:Tensor, src:Union[Tensor, ConstType], reduce:Un
...
-    mask = (index.unsqueeze(-1) == Tensor.arange(self.shape[dim], requires_grad=False, device=self.device)).transpose(-1, dim)
-    src = src.unsqueeze(-1).expand((None,)*src.ndim + (self.shape[dim],)).transpose(-1, dim).shrink(tuple((0,s) for s in mask.shape))
-    src, mask = (x.pad(tuple((0, self.shape[i] - x.shape[i]) if i != dim else None for i in range(self.ndim)) + (None,)) for x in (src, mask))
+    # shrink src to index shape to shrink away the unused values
+    src = src.shrink(tuple((0,s) for s in index.shape))
+    # prepare src and mask for reduce with respect to dim
+    src = src.unsqueeze(-1).expand(*src.shape, self.shape[dim]).transpose(-1, dim)
+    mask = index.unsqueeze(-1)._one_hot_along_dim(self.shape[dim]).transpose(-1, dim)
+    # pad src and mask to self.shape so that reduce can be done with padded values as no-ops
+    src, mask = (x.pad(tuple(None if i == dim or i == x.ndim - 1 else (0, self.shape[i] - x.shape[i]) for i in range(x.ndim))) for x in (src, mask))
```
chenyuxyz, line 2346: i found the old one better, because it has one less check in if condition and self.ndim is easier to reason because x would change (yes ndim is the same, but that's one layer of thinking) but self remained the same

tinygrad/tensor.py
```diff
@@ -3354,6 +3357,10 @@ def dropout(self, p=0.5) -> Tensor:
     if not Tensor.training or p == 0: return self
     return (Tensor.rand_like(self, requires_grad=False, dtype=dtypes.default_float, contiguous=False) >= p).contiguous().where(self, 0) / (1.0 - p)
 
+  # helper function commonly used for indexing
+  def _one_hot_along_dim(self:Tensor, num_classes:sint, dim:int=0):
```
chenyuxyz, line 3361: it is weird that `dim` here is not the `dim` we used in other place. given that dim here is almost always 0, maybe we have the final reshape outside of this function

geohotstan: Yeah that dim is weird. Kinda hacked just for getitem.

I don't think having reshape outside this function works since some use-cases requires the arange to target a specific dimension for masking

I tried to write it in a way where `dim` is still the `dim` concept used in other places. So instead of `0` as default, it's `-1` to indicate we're masking the `self` Tensor on the last dim.

chenyuxyz: okay, will merge this in first. it's definitely a good thing to factor this idea into a function

PR #8061, opened 2024-12-05
bounty: Add pattern for div mod recombine with gcd

S-Lykles:
pattern needed for "fix divmod folding" bounty
The other test case is still failing because a uop like `(y + (x//c)*c) + x%c` doesn't match. Maybe needs some thinking about how add chains should be matched.


chenyuxyz (commented):
seems fine, it's a generalization of `(x%c)+(x//c)*c = x` when `c2==c3==1`

chenyuxyz:
some wins
image

PR #8066, opened 2024-12-06
improve idiv _min_max

chenyuxyz:
for the cases that the we don't know the exact bounds, we might still know the sign. with this, can remove some resolve for symbolic shapetracker


chenyuxyz:
no kernel changed, it needs to be more complicated than symbolic arange (which we can save a few resolve with this) for the kernel to change

PR #8068, opened 2024-12-06
handle stride 0 reduce [pr]

geohot:


Qazalin:
```
VIZ=1 python test/test_schedule.py TestSchedule.test_image_conv_fusion_minimal
```
before:
image

after:
image


more wins in the process replay diff:
image

PR #8069, opened 2024-12-06
combine get inputs and type_parse function in onnx

geohotstan:
we use `prepare_input` to handle both parsing `inputs` and `inputs` validation
also the additional skipped tests should've failed before :D


chenyuxyz:
reverted, seems to have issue on benchmark

PR #8070, opened 2024-12-06
move more onnx ops into tensor_methods mapping

geohotstan:
The idea I had initially was to extend the `tensor_methods` mapping as much as I could to replace ops in onnx_ops.
This only maps the "same function, different name" case.
There are other cases like "same function, different argument name" cases, and many more...

If this idea is ok, I'll go hard on the ops deletion using this.


geohotstan:
okok got it

PR #7937, opened 2024-11-28
move copy into the JIT for openpilot compile3

geohot:
@nimlgen Fails on device with

```
comma@tiny24:/data/tinygrad$ PYTHONPATH="." QCOM=1 taskset -c 4-7 python3 examples/openpilot/compile3.py https://github.com/commaai/openpilot/raw/v0.9.7/selfdrive/modeld/models/supercombo.onnx
loaded model
created tensors
run 0
Traceback (most recent call last):
  File "/data/tinygrad/examples/openpilot/compile3.py", line 82, in <module>
    test_val = compile() if not getenv("RUN") else None
               ^^^^^^^^^
  File "/data/tinygrad/examples/openpilot/compile3.py", line 44, in compile
    ret = run_onnx_jit(**inputs).numpy()
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/data/tinygrad/tinygrad/engine/jit.py", line 229, in __call__
    input_buffers, var_vals, names, st_vars_dtype_device = _prepare_jit_inputs(args, kwargs)
                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/tinygrad/tinygrad/engine/jit.py", line 182, in _prepare_jit_inputs
    if tensors: Tensor.realize(*tensors)
                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/tinygrad/tinygrad/tensor.py", line 3795, in _wrapper
    ret = fn(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^
  File "/data/tinygrad/tinygrad/tensor.py", line 227, in realize
    run_schedule(*self.schedule_with_vars(*lst), do_update_stats=do_update_stats)
  File "/data/tinygrad/tinygrad/engine/realize.py", line 217, in run_schedule
    ei.run(var_vals, do_update_stats=do_update_stats)
  File "/data/tinygrad/tinygrad/engine/realize.py", line 168, in run
    et = self.prg(bufs, var_vals, wait=wait or DEBUG >= 2)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/tinygrad/tinygrad/engine/realize.py", line 133, in __call__
    self.copy(dest, src)
  File "/data/tinygrad/tinygrad/engine/realize.py", line 139, in copy
[...]

PR #8072, opened 2024-12-06
test copy inside jit [pr]

geohot:


geohot:
Ugh why does #7937 fail?

PR #8081, opened 2024-12-06
combine get inputs and type_parse function in onnx [fixed]

geohotstan:
my bad I did run model benchmark locally and it passed but made a small adjustment after thinking it would be fine.

problem was this:
image
it looks like dim_value despite being Optional can be set to `0` when it's a variable dim.


chenyuxyz:
cool, we also need to fix variable dim at some point

geohotstan:
> cool, we also need to fix variable dim at some point

yeah to do that we'll need to do more parsing at the beginning of `get_run_onnx`. I had plans for this already :D
will post PRs to share ideas to see if it's acceptable.

PR #8082, opened 2024-12-06
Fix add chain

S-Lykles:
If the root of the add chain is split, i.e. its sources are both also add, then the `fold_unrolled__divs` pattern won't match at the root of the add chain.
```
  (UPat(Ops.ADD, name="divs", src=[UPat(), UPat(Ops.IDIV)]), fold_unrolled_divs),
```

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -1081,7 +1081,8 @@ symbolic_simple = PatternMatcher([
 symbolic = symbolic_simple+PatternMatcher([
   # ** COMMUTATIVE flipping **
   (UPat(GroupOp.Commutative, name='x'), lambda x: x.replace(src=x.src[::-1]) if x.src[1].tuplize < x.src[0].tuplize else None),
-  ((UPat.var("x1") + (UPat.var("x2")) + UPat.var("x3")), lambda x1,x2,x3: x1+x3+x2 if x3.tuplize < x2.tuplize else None),
+  ((UPat.var("x1", dtype=dtypes.ints) + (UPat.var("x2", dtype=dtypes.ints)) + UPat.var("x3", dtype=dtypes.ints)),
+    lambda x1,x2,x3: x1+x3+x2 if x3.tuplize < x2.tuplize else None),
   # group like
   ((UPat.var("x") + UPat.var("y")) + UPat.var("x") * UPat.cvar("c"), lambda x,y,c: (x+x*c)+y),
   # ** boolean algebra **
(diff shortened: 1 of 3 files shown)


chenyuxyz:
added another $200 bounty for you to fix this cleanly. it should work with all COMMUTATIVE ops (basically linearize all chains) and it need a lot more tests for different kinds of tree, and it cannot regress rewrite speed too much

S-Lykles:
@chenyuxyz so this does linearize all commutative+associative chains. This is not reordering any of the terms in the chains though. So things like merging `x` and `x*10` is still broken if they are far apart in the chain, the same for the div arange folding which depends on the terms `x//a` `(x+1)//a` etc all being next to each other in the chain. With the way sorting on `UOp.tuplize` works they would end up next to each other but right now there is just no sorting. I dont know if you meant for the bounty to include the sorting or just the flattening.

chenyuxyz:
i was expecting sorting because your old commit that has tuplize swap + flattening should give a bubble sort

S-Lykles:
Ill work on the sorting in a different commit though, Ill keep this one to just the flattening

tinygrad/ops.py
```diff
@@ -1081,7 +1083,10 @@ def sint_to_uop(x:sint, dtype:DType=dtypes.int) -> UOp: return UOp.const(dtype,
 symbolic = symbolic_simple+PatternMatcher([
   # ** COMMUTATIVE flipping **
   (UPat(GroupOp.Commutative, name='x'), lambda x: x.replace(src=x.src[::-1]) if x.src[1].tuplize < x.src[0].tuplize else None),
-  # group like
+  # ** group stuff **
+  # flattening of ADD, MULL chains etc, generalization of (x1+x2) + (x3+x4) -> x1+x2+x3+x4
```
chenyuxyz, line 1087: typo `MULL`, also you just need `Associative` for this right? i know you are working on followups, but always good to keep individual ones standalone and minimum

chenyuxyz (commented):
benchmark looks fine

can you update the pr title and description?

tinygrad/ops.py
S-Lykles: It does flip the ordering so it needs to be commutative too `(x1+x2)+x3` becomes `(x1+x3)+x2`

S-Lykles:
I changed my mind actually and I'm going to close this. This is not a very efficient way of doing it, the ordering and flattening should just be done in one go together

PR #8093, opened 2024-12-06
assign early folding

Qazalin:
prereq for CONST -> VALID rewrite.

image

image

if it's a flat CONST it must be the src of a BUFFER op. I found correctness problems with UPCAST otherwise.


Qazalin:
couldn't repro the generate_dataset stuff despite many tries ended up just graph_rewriting the ASTs. I propose we make this a benchmark and download artifact.

PR #8099, opened 2024-12-07
don't allow explicit nan and inf

geohot:
Opening this PR for debate.


TiborGY:
Saw the poll. 
If I am ever to take the plunge and invest time into rewriting a bunch of stuff to use tinygrad instead of legacy BLAS/LAPACK linalg, I am going to need FP64 + full IEEE math support. Including denormals and NaN/Inf as far as the HW supports them.
If there was a toggle for the speed/IEEE-compliance tradeoff at runtime, that would be best.

geohot:
Closing this, we're keeping them

PR #8100, opened 2024-12-07
VALID early folding

Qazalin:


Qazalin:
This could've folded earlier:
image

```
VIZ=1 python3 test/test_symbolic_ops.py TestSymbolicOps.test_ones_sum
```

it's a larger problem with reduceop folding checking all_int. I think the scheduler upats can handle this though it's not in the #8068 diff.

image

PR #8104, opened 2024-12-07
hcq remove update from args state

nimlgen:


nimlgen:
Even faster enqueue now

```
enqueue in   3.10 ms
total 158.56 ms, 6.31 tok/s, 888.81 GB/s, param 870.15 GB/s
```

```
enqueue in   4.03 ms
total 158.47 ms, 6.31 tok/s, 889.31 GB/s, param 870.64 GB/s
```

PR #8109, opened 2024-12-08
Support asymmetrical pads for all pooling functions

geohotstan:
TODO:
- [x] add docs and figure out how to test this in test_ops
- [lol] ~fix onnx maxpool return_indices for repeated values~ (gonna figure this out later)
- [x] maybe move maxpool return_indices and unmaxpool into tensor.py
- [x] figure out why `test_averagepool_3d_dilations_large_count_include_pad_is_1_ceil_mode_is_True_cpu` is failing

looks like this is ballooning up into a huge PR again -_-
will split up stuff to individual PRs when I get to it.


tinygrad/tensor.py
```diff
@@ -2000,20 +2000,21 @@ def _pool(self, k_:Tuple[sint, ...], stride:Union[Tuple[int, ...], int]=1, dilat
     return x.permute(*range(len(noop)), *[len(noop)+i*2 for i in range(len(i_))], *[len(noop)+i*2+1 for i in range(len(i_))])
 
   def _padding2d(self, padding:Union[int, Sequence[int]], dims:int) -> Sequence[int]:
+    if not isinstance(padding, int) and not (len(padding) == 2*dims or len(padding) == dims):
+      raise ValueError(f"Expected padding of length {2*dims} or {dims}, but got {len(padding)} for tensor of shape {self.shape}")
```
geohotstan, line 2004: I moved [this](https://github.com/tinygrad/tinygrad/pull/8109/files#diff-7703e7b63c9844f019ae3609629fd9c9bab290d11ae6608ed2d09edd49bab2dbL2096) from `conv2d` into `_padding2d`

should we rename `_padding2d` to something like `_resolve_pool_pad`? And run it in the beginning of every pool function to resolve the pad like we do with `_resolve_dim`?

test/external/external_test_onnx_backend.py
```diff
@@ -164,6 +164,12 @@ def supports_device(cls, device: str) -> bool:
 backend_test.exclude('test_scatter_elements_with_reduction_max_cpu') # max not yet supported
 backend_test.exclude('test_scatternd_max_cpu') # max not yet supported
 
+# TODO: unsure why this is failing
+# 0: op "AveragePool" input shapes [(1, 1, 32, 32, 32)] opt {'ceil_mode': 1, 'count_include_pad': 1, 'dilations': (2, 2, 2),
+# 'kernel_shape': (5, 5, 5), 'strides': (3, 3, 3)}
+# torch doesn't support dilation but test vs torch passes when artificially expanding the kernel to (9, 9, 9)
```
geohotstan, line 170: for the modified case that can be ran from torch 
```
x_shape = (32, 32, 32)
dilations = (1, 1, 1)
kernel_shape = (9, 9, 9)
strides = (3, 3, 3)
count_include_pad = 1
ceil_mode = True
```

using onnx's ground truth runner from https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/averagepool.py#L560

it appears the onnx result differs from torch

what.

geohotstan:
onnx tests covering asymmetric pads

```
test_averagepool_2d_same_lower_cpu
test_averagepool_2d_same_upper_cpu
test_convtranspose_autopad_same_cpu
test_convtranspose_output_shape_cpu
test_maxpool_2d_same_lower_cpu
test_maxpool_2d_same_upper_cpu
```

chenyuxyz:
is there a way to split this into standalone smaller changes? it's pretty hard to follow

geohotstan:
> is there a way to split this into standalone smaller changes? it's pretty hard to follow

my bad. Will do.

also there's only 1 model in the top 100 models that uses asymmetrical pads
`51: briaai/RMBG-1.4 (708372 downloads)`
`Node 5: op "MaxPool" input shapes [(1, 32, 512, 512)] opt {'ceil_mode': 0, 'kernel_shape': (2, 2), 'pads': (0, 0, 1, 1), 'strides': (2, 2)}`
looooooooool

geohotstan:
no test coverage for asymmetric padding:
AVG_POOL: `ceil_mode=True` and `count_include_pad=False` (because ceil_mode and count_include_pad are both implicit with `padding` parameter)
MAX_POOL: `ceil_mode=True` (same reason)
CONV_TRANSPOSE: only 2 tests covering it from ONNX. (padding is applied implicitly so cannot pad -> convtranspose)

Unfortunately TensorFlow and Jax both do not provide these behaviors (despite Jax allowing asymmetric pads, Jax's conv_transpose applies pad before, so pad -> convtranspose, so the behavior is different.)

Main asymmetric padding usage is conv2d, which we have already. I'm having second thoughts of having this, but idk. :s

chenyuxyz:
seems fine, i like the removal of function arg and it's in onnx spec anyway

PR #8113, opened 2024-12-08
Int64 indexing implemented.

7undSiebziger:
# Int64 Index Conversion for Large Memory Operations

## What
Adds automatic conversion to int64 indices when operations would exceed int32 bounds.

## Why
Enables support for:
- Arrays larger than 2GB
- Index calculations that exceed int32 range
- Prevents integer overflow issues

## How
- Added pattern matcher rule for int64 conversion
- Detects when values exceed int32 range using _min_max tracking
- Converts operations and their inputs to int64 when needed
- Preserves existing int64 operations

## Testing
Added unit tests that verify:
- Basic int32 overflow handling
- Complex operations with large values
- Value preservation after conversion
Tests are efficient (no large memory allocation) while still covering the functionality.

### DIFF
--- a/test/unit/test_uop_symbolic.py
+++ b/test/unit/test_uop_symbolic.py
@@ -359,6 +359,7 @@ class TestSymbolic(unittest.TestCase):
     self.helper_test_variable((-Variable("idx", 0, 100)+199)//-4 + 50, 1, 26, "((((idx*-1)+199)//-4)+50)")
     self.helper_test_variable((-Variable("idx", 0, 100)+200)//-4 + 50, 0, 25, "((((idx*-1)+200)//-4)+50)")
     self.helper_test_variable((-Variable("idx", 0, 100)+201)//-4 + 50, 0, 25, "((((idx*-1)+201)//-4)+50)")
+    self.helper_test_variable((-Variable("idx", 0, 100)+202)//-4 + 50, 0, 25, "((((idx*-1)+202)//-4)+50)")
 
   def test_sum_div_big_const(self):
     gidx0 = Variable("gidx0", 0, 24)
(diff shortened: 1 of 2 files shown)


7undSiebziger:
@geohot Could you review this PR when you have time? Let me know if there's anything else I should add or improve.

tinygrad/codegen/uopgraph.py
```diff
@@ -446,6 +446,17 @@ def delete_redundant_gates(buf:UOp, idx:UOp, val:UOp, store_gate:UOp, cast:Optio
   # remove the gate from the index
   return UOp.store(buf.index(idx).cast(cast.dtype) if cast is not None else buf.index(idx), val)
 
+# Implementation of int64 indexing
+def int64_indexing(buf:UOp, idx:UOp):
+  def needs_int64(u:UOp) -> bool:
+    return hasattr(u, '_min_max') and max(abs(x) for x in u._min_max) > dtypes.max(u.dtype)
```
davidjanoskyrepo, line 452: Comparing abs of min to the dtype max seems odd rather than min to min. Since this is int index as opposed to uints, couldn't this be reduced to numel/2, with a focus on converting large positive indices to smaller negative indices? Or avoiding all that just the numel. Kind of bakes in offset based indexing.

test/unit/test_uop_symbolic.py
```diff
@@ -754,5 +755,109 @@ def test_unrolled_arange(self):
...
+        # Test ALU operations
+        alu_idx = (idx + make_int64(1)) * make_int64(2)
+        load9 = UOp.load(buf.index(alu_idx))
+        simplified9 = load9.simplify()
+        assert simplified9.src[0].src[1].dtype == dtypes.int64
+
+    def test_int64_conversion_from_int32(self):
+        # Create int32 indices that will overflow
+        idx = UOp.const(dtypes.int32, 0x7fffffff)  # Max int32
+        idx._min_max = (0x7fffffff, 0x7fffffff)  # Set min/max values
```
davidjanoskyrepo, line 830: For values like these, why not use dtype min/max function? Also the testing structure doesn't seem to systematically test all edge points. 0 indexing, -1 and mixing pos and neg tend to be some common ones.

tinygrad/codegen/uopgraph.py
7undSiebziger: Thanks for the feedback!

We now directly compare the min and max indices against the dtype’s own representable range rather than relying on abs. This respects signed indexing properly (checking min against min, not abs(min) against max).

For numel it is not used as the current implementation focuses purely on whether the indices fit within the representational limits of the chosen data type. So the decision to switch to int64 is driven by the integer’s fundamental representability, not how large or meaningful the indices are relative to the array’s length.

Let me know if that makes sense or I am missing something here.

test/unit/test_uop_symbolic.py
7undSiebziger: Good point. Now a lot of additional tests should be added.

7undSiebziger:
Any updates @davidjanoskyrepo

davidjanoskyrepo:
The tests look better. Not sure if gh will go for the implementation though.

davidjanoskyrepo (approved):

7undSiebziger:
@geohot 
Can you give feedback on that?

PR #8120, opened 2024-12-09
[POC] add test_onnx_ops.py tests

geohotstan:
I think there are 2 ways to make sure onnx ops is well tested
1. somehow make onnx directly call tensor.py apis with tensor.py apis well tested vs torch in `test_ops.py`
2. just straight up test vs ORT? We write additional tests on top of `external_test_onnx_backend.py` to covers everything, making sure onnx_ops is well tested

I'm having some trouble with the 1st way. There are a lot of things onnx expects/supports that torch doesn't, so tests would have to be hacked. 
e.g. writing tests vs torch for asymmetrical pads for pool ops is just cursed.

so maybe we go for 2nd way? This way we can also add tests where `external_test_onnx_backend.py` doesn't cover (onnx's test suite coverage is uhhhh, not great.)


chenyuxyz:
using ORT as the ground truth sounds fine, and that would be a straight super set to existing onnx tests (i imagine ORT should pass those tests too)

geohotstan:
> using ORT as the ground truth sounds fine, and that would be a straight super set to existing onnx tests (i imagine ORT should pass those tests too)

yeah speaking of super set, I looked into directly injecting these tests into the `external_test_onnx_backend.py` test suite, but that test suite validates via provided output. 
for example: https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/conv.py#L66
so it wouldn't really work for testing vs ort.

geohotstan:
https://github.com/apache/tvm/blob/main/tests/python/frontend/onnx/test_forward.py
hehehehehehhehehehehehehehehhe 🤭

geohotstan:
got a nice new test_onnx coming adapting tvm's onnx test suite :D
I think using their structure, we can really test the numerical stuff too.

I can visualize a fully tested onnx, just gotta cooooode.

PR #8123, opened 2024-12-09
bounty: flattening and sorting of all commutative+associative trees

S-Lykles:
Passing symbolic unit test and fuzzing, still failing a lot of other tests because some other optimizations are broken


S-Lykles:
Again about 5000 fewer lines in GPU+IMAGE 

```
sieds@SL-ZBook:../Downloads/logs_31836430919/ONNX+Optimization Tests $ tail -n 2 27_Run\ process\ replay\ tests.txt
2024-12-09T14:57:58.7090622Z 5649 insertions(+)
2024-12-09T14:57:58.7091269Z 5683 deletions(-)
sieds@SL-ZBook:../Downloads/logs_31836430919/ONNX+Optimization Tests $ tail -n 2 ../GPU\ IMAGE+compile\ Tests/27_Run\ process\ replay\ tests.txt
2024-12-09T15:07:30.1006751Z 24480 insertions(+)
2024-12-09T15:07:30.1007350Z 25068 deletions(-)
```

S-Lykles:
So the last broken metal test is from moving DEFINE_ACC, haven't figured out why though...

S-Lykles:
So basically what happened is that after I implemented sorting of these chains, these patterns broke because they expect `DEFINE_ACC` to be at the top of the chain.
```
  (acc_pat.assign(UPat.any(arange_m, arange_m+UPat.var("extra"))+acc_pat), loop_collapse),
  # indexing, with cast or where
  (acc_pat.assign(UPat.var("idx").eq(UPat(Ops.RANGE, name="rng")).cast()*index_load+acc_pat), index_collapse),
  (acc_pat.assign(UPat.var("idx").eq(UPat(Ops.RANGE, name="rng")).where(index_load, UPat.const(None, 0.0))+acc_pat), index_collapse),
  # parentless reduce  # TODO: add MUL
  (acc_pat.assign(UPat((Ops.ADD, Ops.MAX), src=[acc_pat, UPat.var("ret")], name="alu")), reduce_collapse),
 ```

I fixed this by moving the `DEFINE_ACC` op to after all other ops so it would stay at the top back to where it was originally and that fixed the pattern. This did also cause commutative flipping to change `acc = acc + something` to be `acc = something + acc`. I figured this wasnt a problem since floats and ints etc are commutative. The problem is that with the way cstyle renders `ALU`  it strips the parens. This means that the uop graph that represents `acc = acc + (a + b)` gets rendered as `acc = acc + a + b` which is not the same as `acc = a + b + acc`. In particular `acc = acc + a*b + b*c` is able to use two mulacc instructions to accumulate a*b and b*c into acc.  `acc = a*b + b*c + acc` has one fewer mulacc instruction. Mulacc has higher precision than a mul and an add, so on one particular metal test this ends up failing because of the slight loss of precision... (see #8156). 

I don't like relying on this behaviour of the cstyle renderer so `DEFINE_ACC` should go back and be at the bottom of the chain. I will have to fix those uopgraph patterns instead

**Comment by S-Lykles** [comment]
So regarding speed regression I ran `external_benchmark_schedule.py` 100 times in a loop to get average timing for different patterns. This is only looking at "model rewrite" time
On master I get 1308 ms

With master plus only the following three patterns added I get 1403 ms
```
 ((UPat.var("y") + UPat.var("x") * UPat.cvar("c0")) + UPat.var("x") * UPat.cvar("c1"), lambda x,y,c0,c1: y+x*(c0+c1)),
 ((UPat.var("y") + UPat.var("x")) + UPat.var("x") * UPat.cvar("c"), lambda x,y,c: y+x*(c+1)),
 ((UPat.var("y")+UPat.var("x")) + UPat.var("x"), lambda y,x: y+x*2),
```

Without chain insert I get 1440 ms
With all patterns I get 1596 ms, so in total approx 22% regression.

The slowest pattern by far is chain insert, the problem is that the rewriting is happening from the sources so its hard to sort efficiently.

**Comment by chenyuxyz** [comment]
yea unjitted llama is also 25% slower. speed v theoretical looks fine, but the cifar nan issue seems real and need to be fixed (and unit tested)

feel free to submit small prereq prs that make this one smaller

**Comment by S-Lykles** [comment]
so the failing test on amd seems to be just a flaky blfoat16 test
```
Max absolute difference among violations: 0.0012207
Max relative difference among violations: 0.18518519
 ACTUAL: array([-0.007812], dtype=float32)
 DESIRED: array([-0.006592])
Falsifying example: test_bfloat16(
    self=<test.test_dtype_alu.TestDTypeALU testMethod=test_bfloat16>,
    a=0.32666015625,
    b=0.333251953125,
    op=sub,
)
```

```python
a = 0.32666015625
b = 0.333251953125
Tensor([a], dtype=dtypes.bfloat16) - Tensor([b], dtype=dtypes.bfloat16)
torch.tensor([a], dtype=torch.bfloat16) - torch.tensor([b], dtype=torch.bfloat16)
```

Torch and tinygrad master also give a result of `-0.0078125` which just has a high relative error with `-0.0065918`

S-Lykles:
The cifar nan issue was because I broke stable sigmoid.  With #8806 I can make it so that `x` is sorted next to `1/(x+1)` previously it was quite hard to sort to `x` close to `1/((x+1)*(x+1))` which made it hard to make this rule reliable.

S-Lykles:
Added a test that would have caught the cifar nan thing

S-Lykles:
The onnx test is failing because two shapes are not comparing equal in verify ast

```
s0 = ShapeTracker(views=(View(shape=(UOp(Ops.ADD, dtypes.int, arg=None, src=(
  UOp(Ops.MUL, dtypes.int, arg=None, src=(
    UOp(Ops.DEFINE_VAR, dtypes.int, arg=('b', 1, 5), src=()),
    UOp(Ops.CONST, dtypes.int, arg=-1, src=()),)),
  UOp(Ops.DEFINE_VAR, dtypes.int, arg=('e', 6, 10), src=()),)), 1), strides=(1, 0), offset=0, mask=None, contiguous=True),))

s1 = ShapeTracker(views=(View(shape=(UOp(Ops.ADD, dtypes.int, arg=None, src=(
  UOp(Ops.DEFINE_VAR, dtypes.int, arg=('e', 6, 10), src=()),
  UOp(Ops.MUL, dtypes.int, arg=None, src=(
    UOp(Ops.DEFINE_VAR, dtypes.int, arg=('b', 1, 5), src=()),
    UOp(Ops.CONST, dtypes.int, arg=-1, src=()),)),)), 1), strides=(0, 0), offset=0, mask=None, contiguous=False),))
```

```
print(graph_rewrite(s0.shape[0], sym) is graph_rewrite(s1.shape[0], sym))
```

The test is reading a text file of ast's and one of the views in a previously generated kernel now has a different order of uops so it needs to be canonicalized again.

S-Lykles:
> The onnx test is failing because two shapes are not comparing equal in verify ast
> 
> ```
> s0 = ShapeTracker(views=(View(shape=(UOp(Ops.ADD, dtypes.int, arg=None, src=(
>   UOp(Ops.MUL, dtypes.int, arg=None, src=(
>     UOp(Ops.DEFINE_VAR, dtypes.int, arg=('b', 1, 5), src=()),
>     UOp(Ops.CONST, dtypes.int, arg=-1, src=()),)),
>   UOp(Ops.DEFINE_VAR, dtypes.int, arg=('e', 6, 10), src=()),)), 1), strides=(1, 0), offset=0, mask=None, contiguous=True),))
> 
> s1 = ShapeTracker(views=(View(shape=(UOp(Ops.ADD, dtypes.int, arg=None, src=(
>   UOp(Ops.DEFINE_VAR, dtypes.int, arg=('e', 6, 10), src=()),
>   UOp(Ops.MUL, dtypes.int, arg=None, src=(
>     UOp(Ops.DEFINE_VAR, dtypes.int, arg=('b', 1, 5), src=()),
>     UOp(Ops.CONST, dtypes.int, arg=-1, src=()),)),)), 1), strides=(0, 0), offset=0, mask=None, contiguous=False),))
> ```
> 
> ```
> >>> print(graph_rewrite(s0.shape[0], sym) is graph_rewrite(s1.shape[0], sym))
> True
> ```
> 
> The test is reading a text file of ast's and one of the views in a previously generated kernel now has a different order of uops so it needs to be canonicalized again.

I fixed this now with a pattern over views which call `View.create` to create the canonical version of each view. Maybe `View` should have metaclass just like UOp or some other way to prevent this?

S-Lykles:
Also not sure why `test/test_randomness.py::TestRandomness::test_multinomial` was broken on amd a few commits back but it is working now

S-Lykles:
`model rewrite` speed in ms on master out of 10 runs
```
Statistics:
Mean: 1802,9
Min:  1684.43
Max:  2003.43
```
On this branch
```
Mean: 1898,7
Min:  1776.43
Max:  1998.08
```

tinygrad/ops.py
```diff
@@ -138,6 +138,9 @@ class Ops(FastEnum):
   # TernaryOps
   WHERE = auto(); MULACC = auto() # noqa: E702
 
+  # this is here becasue it needs to be at the top of a reduce chain
```
chenyuxyz, line 141: this and the `tuplize` change below needs to have a dedicated spec test to say what case needs to be sorted in what way. otherwise this is very hard to maintain

chenyuxyz (commented):
is there pieces in this pr that's less related and can be separated (like stable sigmoid)?

S-Lykles:
> is there pieces in this pr that's less related and can be separated (like stable sigmoid)?

I made  #8816 a different pr.
I had to make quite some changes to fold_unrolled_divs to get that pattern to work again and I found it easier to redo it and simplify it using the guarantees of the orderering. The new fold_unrolled_divs is also more extensive as it allows for folding any range instead of only ranges starting at 0. I could add this functionality seperately maybe, but its just that combining it with the new guaranteed ordering makes the function quite elegant and simple.

S-Lykles:
I'm not sure its a good idea to make the ordering a part of the spec. I feel the spec should be that symbolic guarantees certain simplifications like:
```
a + b + c + a  -> a*2 + b + c         (currently not simplified on master)
```

Or loop collapse in the case of `DEFINE_ACC`. The order can then be any order which achieves those simplifications. 
I think its better to extensively test the optimizations.

Its actually quite annoying how brittle a lot of the test in `test_uop_symbolic` are. If I change `a+(b*2)` to be `(b*2)+a` some tests will break, most of this pr is actually just fixing those tests. 

I have a half finished writeup of how the new rules together with new tuplize achieves these simplifications, in particular the reasoning behind the tuplize change. Ill finish and share that

S-Lykles:
# Simplification of Commutative and Associative Chains

Associative+Commutative chains are compostitions of multiple operations that are associative and commutative.
For example:

```
a+b+a
(a+b)+(c+d)+e
((a*b)*c)*(d*e)
```

By flattening and ordering such chains we achieve two goals, the first is a canonical form, the second is to make patterns involving these ops more reliable.
Since all pattern matching rules are local, we must guarantee that the chains are sorted in such away that the terms of the patterns will appear close together.
Consider the pattern `a+a -> 2a`, if we try to match this pattern on `a+b+a` (this is actually `(a+b)+a`) we will not match it.
We could add another pattern `a+b+a -> 2a+b`, but this would not scale very well to `a+b+c+a` etc.
Instead if we can sort the chain to `(a+a)+b` then the pattern will work again.

The goal is then to define an ordering such that all terms which are required to be close together are close together.
This may not be possible for all patterns. Hypothetically, if we had the pattern

```
a*b + a*c -> a*(b+c)
```

It is not possible to define an ordering which guarantees this for both of the following instances,

```
a*b + d + a*c
a*d + b*e + c*d
```

Should we sort on the first term? The second term? Should we bring all terms involving multiplication together?
Whenever we have a term that contains two or more variables, these patterns become impossible to guarantee.
Tinygrad has various irreducible terms, these are the leaf nodes of the computation graph, variables loaded from memory, range indices, global indices, constants etc.
The "irreducible" terms when it comes to sorting are these leaf nodes _or_ binary or ternary ops themselves involving atomic terms.
We add one aditional condition, that is that binary or ternary terms involving constants do not themselves become atomic terms.
Examples:
`a, a+1, 5*a, a//10, a%42, log2(a), sin(a), a.where(12, 32)` all have `a` as their "atomic" term.
`b*c, b*c+1, 5*(b*c), (b*c)//10, (b*c)%42, log2(b*c), sin(b*c), (b*c).where(12, 32)` all have `b*c` as their "atomic" term.

When we sort a chain, we ensure all of these atomic terms are clustered. The only thing left to do then is determine how we should sort
`a+1, a//10, log2(a), etc` with respect to each other.

The following patterns need thinking
```
a+a -> 2a
a+c*a -> (c+1)*a            # c is a constant
c1*a + c2*a -> (c1+c2)*a    # c1 and c2 are constants
x*(1/(x+1)) -> 1-1/(x+1)    # stable sigmoid
c*(x//c) + x%c -> x
x//c + (x+1)//c + ... + (x+c-1)//c -> x  # unrolled arange
```

For the first three we need to make sure that multiplication or addition is lower in order than all other unary or binary ops.
Next, if there are multiple ops applied to a term we compare them from the inside out.
```
a+1 < (a+22)//10 < a*10 < (a*11)+4 < (a*11)+5 etc
```
This achieves the desired ordering for all but the unrolled arange pattern.
Now we have a problem, because `(x+1)//c < x//c`. However if we change the pattern to be:
`(x+1)//c + ... + (x+c-1)//c -> x - x//c`
it works again.

## Rules to perform ordering and flattening

We use 2 rules to perfrom the ordering and 1 rule to perform the flattening.

The first rule applies to all commutative oparations. The other 2 only apply to operatations that are both commutative and associative.

1. Within a node, if both nodes are leafs they are sorted with the smallest first on the left (b∘a) -> (a∘b)
   additionally if one of the terms is of the same op, then that term is on the left. Additionally If one of the nodes is a op node of the same type, this node is alway on the left (b∘(a∘c)) -> ((a∘b)∘c)

```
    ∘                 ∘
   / \       ->      / \
  b   a             a   b

    ∘                 ∘
   / \       ->      / \
  a   ∘             ∘   a
     / \           / \
    _   _         _   _
```

2. At each level we sort consecutive nodes. This is a bubble sort if we apply the rule minimally but it can be implemented slightly more efficiently as an insertion sort because the left branch will have already been sorted.
   It is not possible to use better sorting algorithms because the rewriting is happening from the sources first, we don't have information about the coming nodes.

```
      ∘               ∘
     / \     ->      / \
    ∘   a           ∘   b
   / \             / \
  _   b           _   a
```

3. If both nodes are op nodes of the same type, the trees are merged. Since each of the source have already been sorted, this can use a merge sort. (a∘c)∘(b∘d) -> (((a∘b)∘c)∘d)

```
     ∘                ∘
    / \        ->    / \
   ∘   ∘            ∘   d
  / \ / \          / \
  a d b d         ∘   c
                 / \
                a   b
```

S-Lykles:
So I can get a working order if I change unrolled arange from 
`x//c + (x+1)//c + ... + (x+c-1)//c -> x`
to 
`(x+1)//c + ... + (x+c-1)//c -> x - x//c`
Problem with that is that for `c=2` it rewrites 
`(x+1)//2` to `x-x//2` which isnt neccesarilly a better representation.
Then what happens is `uop_given_valid` breaks

PR #8130, opened 2024-12-09
delete lazy gc, realized uop is used in Function.apply 

geohot:


Qazalin:
yes, buffer.view() should absolutely work https://github.com/tinygrad/tinygrad/pull/8136.

PR #8135, opened 2024-12-10
iteration 3 on big_graph = UOp.sink(*outs)

Qazalin:
exploring what it'd take to make the big graph small and flexible for `merge_views+symbolic_flat`
Like a lot of the stuff that exists is filler, by the time full_ast_rewrite returns the AST is pretty good, how can we move those transformations to as early as possible?


Qazalin:
obscure diff after "delete lazy".

PR #8141, opened 2024-12-10
correct pow to int

chenyuxyz:
need more work on making sure it const folds and consolidate with existing stuff.

opening now since CI gives wrong answer on remu and it's correct on tinybox red


chenyuxyz:
llvm failed on ubuntu
remu gives wrong answer
PTX gives wrong answer on CI and real 4090

PR #8142, opened 2024-12-10
Run more WebGPU tests

wpmed92:
Runs 2x as many tests as before. Changed to blacklist style test declarations. Skipped some tests that were failing due to dtype things. (There is one skip that's because of a valid error)
Passing locally on Metal, testing if they also pass on Vulkan.


chenyuxyz:
better to have the bugfix as a standalone pr since it changed behavior

PR #8151, opened 2024-12-10
delete_lazy pruning

Qazalin:


Qazalin:
ahh if there's a VIEW left in the middle it just ends up going away?
image


this is master:
image

Qazalin:
merged all merge worthy stuff from here to real delete_lazy.

PR #8154, opened 2024-12-11
speed vs theoretical on mac benchmark

chenyuxyz:


chenyuxyz:
3 minute is quite slow though... maybe not now

PR #8158, opened 2024-12-11
move to_python_const from onnx_ops to onnx

geohotstan:
experimental idea

`required_input_python_consts` uses positions instead of names since the names can be anything but positionally they stay the same.
only `to_python_const` left is in gather.

if this idea is fine, I'll later do a full typing PR and update some debug messages to pair with this.


chenyuxyz:
i like the idea in general, can you annotate the function with the correct python types too?

geohotstan:
I went through and added typing to all the ops as well
the ones that are without typing are from `OnnxBackendPyTorchConvertedModelTest`. Some weird alternative args.

extra/onnx.py
```diff
@@ -85,6 +85,15 @@ def get_run_onnx(onnx_model: ModelProto):
     "Softsign", "Asinh", "Acosh", "Atanh",  "Elu", "Celu", "Selu", "Xor", "Round", "Erf")
   }
 
+  # these values are expected to be python consts
+  required_input_python_consts: Dict[str, tuple[int, ...]] = {
```
chenyuxyz, line 89: i like that these Tensor functions have correct types, but this mapping is weird and hard to maintain. will need a better way to organize these proto conversion stuff in the future

PR #8160, opened 2024-12-11
stable diffusion UI - convenient send via Enter

maxeozakh:
support sending prompt while focused on the input via Enter
also showing visually that model is running via rendering different title text

<details>
  <summary>before/after demo</summary>
  

https://github.com/user-attachments/assets/5c9b12ee-2b88-4c51-b4ab-fe6613e908c8
</details>

### DIFF
--- a/examples/webgpu/stable_diffusion/index.html
+++ b/examples/webgpu/stable_diffusion/index.html
@@ -177,19 +177,21 @@
     </a>
     
     <div id="mybox">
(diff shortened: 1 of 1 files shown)


chenyuxyz:
@wpmed92 do you like it? these are fine as long as it keeps things simple and adds value to UX

wpmed92:
@chenyuxyz Just tested it, it's working. Starts diffusion upon enter, and writes out "Running model" which makes it clearer that the model is running until the first step time is shown. I like these changes.

PR #8162, opened 2024-12-11
test: remove contiguous_child

Qazalin:


Qazalin:
goehot added this to delete_lazy.

PR #8183, opened 2024-12-12
special tol when f16 and bf16 are tc input dtypes

ignaciosica:
fixes flaky 
image


geohot:
Can you fix it so the test actually makes CI fail?

ignaciosica:
Sure, I'll look into it, it don't know why it didn't failed in CI, I remember that specific test failing to me on GitHub CI, not in benchmark though.

chenyuxyz:
still want this? is CI flaky (seems fine to me)? is it flaky locally?

ignaciosica:
> is it flaky locally?

yes

image

ignaciosica:
I believe the run failed on benchmark rather than on GitHub ci, I couldn't find George's run though

chenyuxyz:
seems fine to me. after tensor refactor and cleanups are done i am interested to better understand these errors. like float16 and bfloat16 are larger, but how much. also important as we will have float8

PR #8186, opened 2024-12-12
uopgraph touchup [pr]

wpmed92:
We always put the lambdas in the `UPat`s on one line if it fits the `150` limit, there were some cases where they were on 2 although they fit.


geohot:
So I'm in general against PRs like this.

It's fine if you really need lines for some specific thing, but it's not worth doing preemptively. Line savings is only valuable if it reduces complexity and makes it easier to read.

PR #8187, opened 2024-12-12
proposal: preserve src size of REDUCE_AXIS while const folding

Qazalin:
This is a prereq for early CONST folding with symbolic correctness.
The CONST src folds first and the maskless VIEW goes away:
image
image

when we match REDUCE_AXIS, it needs a way to know if the source was a size 0 VIEW.


geohot:
I don't think this is the right way to approach this. Why does the input arg not have a shape?

Qazalin:
I must fold CONST to something that matches UPat.cvar.
otherwise symbolic+ won't match it. CONST when it's in `ops_folding` must have the exact same spec as UOp.const.
the shape and device aren't ever required anymore in the scheduler, besides REDUCE_AXIS folding, everything else is fine with a shapeless source.

PR #8188, opened 2024-12-12
add "tinygrad" to be part of cache_dir [pr]

chenyuxyz:
instead of having sqlite / http download / metal compile to add "tinygrad" separately. also make it non-private since it's used in metal


chenyuxyz:
also tested by deleting cache, run `PYTHONPATH=. JITBEAM=2 python3 examples/llama3.py --download_model --size 1B --quantize float16 --benchmark` on this branch, then checkout master and run it again and all cache hit

PR #8214, opened 2024-12-13
Cleaner patterns [pr]

S-Lykles:


S-Lykles:
tldr: 
define a few variables x,y, etc as UPat.var("x"),  UPat.var("y") and then you can use use those variables in the pattern instead of writing UPat.var or UPat.cvar all the time

chenyuxyz:
discussed with geohot, don't really like this because it adds stuff in the global scope, and it makes rules less local (so you need to scroll more to read the whole rules)

PR #8216, opened 2024-12-13
add real AMX tests to benchmark

ignaciosica:
Can someone trigger benchmark job?

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -62,6 +62,8 @@ jobs:
       run: BIG=2 MPS=1 python3.11 test/test_speed_v_torch.py | tee torch_speed.txt
     - name: Test tensor cores
       run: METAL=1 python3.11 test/test_linearizer.py TestLinearizer.test_tensor_cores TestLinearizer.test_tensor_cores_padded
+    - name: Test AMX tensor cores
+      run: CLANG=1 AMX=1 python3.11 test/test_linearizer.py TestLinearizer.test_tensor_cores TestLinearizer.test_tensor_cores_padded
     - name: Run Tensor Core GEMM (float)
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -62,6 +62,8 @@ jobs:
       run: BIG=2 MPS=1 python3.11 test/test_speed_v_torch.py | tee torch_speed.txt
     - name: Test tensor cores
       run: METAL=1 python3.11 test/test_linearizer.py TestLinearizer.test_tensor_cores TestLinearizer.test_tensor_cores_padded
+    - name: Test AMX tensor cores
+      run: CLANG=1 AMX=1 python3.11 test/test_linearizer.py TestLinearizer.test_tensor_cores TestLinearizer.test_tensor_cores_padded
```
chenyuxyz, line 66: you would need `DEBUG=2` or some other ways to tell if AMX is really being used

geohot:
Oh, do we have to use benchmark? What about the Apple runner in GitHub CI?

ignaciosica:
> Oh, do we have to use benchmark? What about the Apple runner in GitHub CI?

I already tested that and its not supported in the apple runner

.github/workflows/benchmark.yml
ignaciosica: Sorry, I thought the tests asserted that they were triggered

chenyuxyz: oh yea it asserts, i was thinking running `simple_matmul` with AMX like other gemm

chenyuxyz:
we can benchmark this like other gemm later. 2 seconds is fine to start with, and i do see test asserts if tc is not used

PR #8217, opened 2024-12-13
add unit tests for to_dtype

pkotzbach:


test/test_dtype.py
```diff
@@ -854,5 +854,18 @@ def test_max_w_alu(self):
         t = Tensor([[1, 2], [3, 4]], dtype=d)
         (t*t).max().item()
 
+class TestToDtype(unittest.TestCase):
+  def test_dtype_to_dtype(self):
+    dtype = dtypes.void
+    res = to_dtype(dtype)
+    assert isinstance(res, DType)
```
chenyuxyz, line 861: use `self.assertIsInstance` and `self.assertEqual` for better error message.

also `dtypes.void` is an internal dtype, test with with common dtype like `dtypes.int32`

chenyuxyz:
thanks!

PR #8221, opened 2024-12-13
add detach to graph [pr]

geohot:


geohot:
@Qazalin I expect things like this to work. we need to move folding rules before bufferization.

PR #8227, opened 2024-12-13
cpu objdump in intel syntax

ttomsa:
More readable and x86 backend is in intel syntax.


geohot:
What does this do on ARM?

ttomsa:
doesn't do anything. I tried "-m arm64" on my x86 machine and the flag is ignored.

geohot:
Hmm, that warning is ugly

PR #8230, opened 2024-12-14
type check jit add method in realize

mesozoic-egg:
I think it was the circular import that prevented this from being added in the first place? 

Combining `TYPE_CHECKING` conditonal import and quotation mark for forward reference seems to solve the problem


geohot:
Other things can be used there, they are in tests IIRC

mesozoic-egg:
Oh right, I just saw the usage in test_search.py and fuzz schedule, probably fine to leave as is

PR #8235, opened 2024-12-14
switch backward to use gradient [pr]

geohot:
CAST_BEFORE_VIEW needs to go in the scheduler

Is realize always realizing the Tensors?


geohot:
new gradient uses 11 less kernels on hlb_cifar. gonna just update the scheduler

geohot:
Will merge if this passes benchmark!

PR #8238, opened 2024-12-14
fix some sigmoid extreme

chenyuxyz:
quite brittle... the problem is it has 3 terms and mul might have bad order


chenyuxyz:
merging this one, will fix other extreme backward stuff and see how small we can make the rules to be

PR #8239, opened 2024-12-14
[pr] GroupOp.Associative

leopf:
The replaced rules can be applied to all associative ops including `Ops.XOR` and `Ops.MAX` as well. I have left them out to make sure there are no changes to the graph.

Can submit this separately or include in this PR, if you want this at all.

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -164,6 +164,9 @@ class GroupOp:
   # BinaryOps that can be flipped
   Commutative = {Ops.ADD, Ops.MUL, Ops.MAX, Ops.CMPNE, Ops.XOR, Ops.AND, Ops.OR}
 
+  # BinaryOps where f(f(a,b),c) = f(a,f(b,c))
+  Associative = {Ops.ADD, Ops.MUL, Ops.AND, Ops.OR}
+
   # BinaryOps that satisfy f(x,x)=x see https://en.wikipedia.org/wiki/Idempotence
   Idempotent = {Ops.OR, Ops.AND, Ops.MAX}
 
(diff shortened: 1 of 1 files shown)


tinygrad/ops.py
```diff
@@ -1091,6 +1094,8 @@ def lt_folding(x:UOp, c:int) -> Optional[UOp]:
     return cast(UOp, functools.reduce(operator.add, np).divides(d))<(c//d)
   return None
 
+def two_stage_associative_folding(f: UOp, x: UOp, c1: UOp, c2: UOp): return x.alu(f.op,c1.alu(f.op,c2))
```
geohot, line 1097: Why is this not a lambda?

leopf: Would have to be written in two lines:
```python
  *((UPat.var("x").alu(op, UPat.cvar("c1")).alu(op, UPat.cvar("c2")).named("f"), 
     lambda f,x,c1,c2: x.alu(f.op,c1.alu(f.op,c2))) for op in GroupOp.Associative),
```
Is this preferred? 
Usage of op directly is not possible because it would require closure support (see `deconstruct_function`).

geohot:
Why do you need closures for this to be a lambda?

leopf:
I do not _need_ closures, but thought it would be cleaner to write:
```python
*((UPat.var("x").alu(op, UPat.cvar("c1")).alu(op, UPat.cvar("c2")), lambda x,c1,c2: x.alu(op,c1.alu(op,c2))) for op in GroupOp.Associative),
```

Instead of:
```python
  *((UPat.var("x").alu(op, UPat.cvar("c1")).alu(op, UPat.cvar("c2")).named("f"),
     lambda f,x,c1,c2: x.alu(f.op,c1.alu(f.op,c2))) for op in GroupOp.Associative),
```

I have pushed the solution without closures.

PR #8241, opened 2024-12-14
add GatherND and ScatterND to onnx ops

geohotstan:
need to still ~verify correctness~ (wait still wrong) and check if there's a better implementation.


geohotstan:
There's probably a simpler implementation but the implementation currently is correct.

These two ops are another case of onnx test suite not testing enough cases 🫠
Will add the tests in once I figure out how to superset onnx tests.

chenyuxyz:
i think it's fine to add test cases that's lacking, maybe the ones you had bug that was not caught, by calling the onnx function like other Tensor functions to start with.

PR #8244, opened 2024-12-14
PatternMatcher closure support

leopf:
This will allow a cleaner rewrite of #8239.

I am working on making the symbolic stuff more general. Having closures will definitely help.


geohot:
We probably don't want this. No closures are by design, it makes things harder to reason about.

PR #8247, opened 2024-12-14
type_verify(big_graph.toposort) [pr]

Qazalin:
1. shape change bitcast can have "implicit" movement ops, this is probably fine?
2. the root cause of GC asserts is toposort's lru_cache:
image


geohot:
That doesn't make sense, re: the toposort. That cache should only be in scope for the function.

Qazalin:
image
looks like I need to manually clear a local_cache?

geohot:
Do you really need the del?

Qazalin:
GC root cause is closures - https://github.com/tinygrad/tinygrad/pull/8261.

I'm gonna keep the BUFFER_VIEW weirdness as-is for now, it can change dtype, it can change shape. It's an untasteful op.

Qazalin:
Closing in favour of 8311

PR #8250, opened 2024-12-14
merge buffer uops in view collapse [pr]

Qazalin:
all these things trace back to https://github.com/tinygrad/tinygrad/issues/8253 - I think the right fix is to debufferize everything. This diff gives an idea of what it'd take to make merge_bufs pass [RealizeMeansRealize](https://github.com/tinygrad/tinygrad/pull/8235/files#diff-fe6095a1b53c4e9685483ed630615a990ff0ba751a4b743359e1447e6720a55cR96).

The red is because BUF1 -> VIEW -> BUF2
where the VIEW is an expand, neither buf2 nor buf1 are the valid merged buffer. It needs to create a new buffer. `test_unrealized_expanded_merge`.
All of this work doesn't make sense, debufferization makes this problem not exist.

### DIFF
--- a/test/test_schedule.py
+++ b/test/test_schedule.py
@@ -1991,5 +1991,18 @@ class TestBigGraph(unittest.TestCase):
     merged = merge_rewrite(folded, ScheduleContext({c:[unrealized_op],b:[folded]}))
     self.assertIs(merged, unrealized_op)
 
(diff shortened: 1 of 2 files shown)


Qazalin:
This issue was tracked back to required_grad in randn

PR #8258, opened 2024-12-15
remove unused code

pkotzbach:


pkotzbach:
Hello @Qazalin,
could you tell me why you closed it? I would appreciate feedback :)
I am running static analysis of tinygrad code and i have found that this is not used anywhere nor tested, so I think its good to remove that to decrease number of lines

PR #8261, opened 2024-12-15
minimal failing test for UOp.toposort gc [pr]

Qazalin:
> That [UOp.toposort] cache should only be in scope for the function.
https://github.com/tinygrad/tinygrad/pull/8247#issuecomment-2543353016

However, the closure in toposort keeps a reference to the BUFFER uop [around](https://docs.python.org/3/c-api/cell.html) forever.

### DIFF
--- a/test/test_gc.py
+++ b/test/test_gc.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python
-import gc
+import gc, inspect
 import unittest
 import numpy as np
 from tinygrad.device import Buffer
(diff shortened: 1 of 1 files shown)


geohot:
Huh, does it matter that it's a property?

Qazalin:
1. does it matter that it's a property? No.

Screenshot 2024-12-15 at 7 25 59 PM

2. try gc.collect. fixes, though I think this should be automatic.

PR #8262, opened 2024-12-15
Isolate ClangPrograms from global symbols when possible

uuuvn:
`GNU_SOURCE` is required in autogen because dlmopen is a GNU extension. `ISOLATE_DLOPEN` is disabled on OSX for the very same reason.
test_isolated_dlopen checks that `ISOLATE_DLOPEN=1` prevents ClangProgram from linking to `sqrtf` which is chosen as a symbol that should always exist in libc/libm
`extra_matcher` rewrites float64 => float16 cast (implemented in clang as a call into compiler-rt) as float64 => float32 => float16 which is just a bunch of instructions
`-fno-math-errno` is required because otherwise math functions like `__builtin_sqrt` have to set errno and won't be replaced by an instruction because of that


geohot:
Why do I want these 10 lines? What value does this add?

uuuvn:
> Why do I want these 10 lines? What value does this add?

In https://github.com/tinygrad/tinygrad/pull/7615#issuecomment-2469311501 you said that you want a generic way to ensure that nothing links to compiler-rt. I've spent quite a bit of time trying out different ways to achieve that and this is the simplest one that doesn't rely on testing for specific patterns. The second to simplest was modifying elf loader and using a part of it in compiler to check that no external symbols are linked with but it was less clean (to my eye at least) and more likely to break if llvm changed the way it makes elfs. I can rewrite this PR to do that instead if you think this is the more correct way to do that

PR #8264, opened 2024-12-15
free toposort cache after it goes out of scope [pr]

Qazalin:


geohot:
Interesting. I still don't understand why the other one doesn't work, I suspect it's an antipattern we use in other places. My expectation is that when the function goes out of scope the lru_cache is cleared.

Qazalin:
Seems to be something related to how python deals with closures.
gpt wrote this to explain it to me:
```
import inspect

def outer_function(x):
  def inner_function(y):
    return x + y
  return inner_function

closure = outer_function(10)
print(closure(5)) # 15
print(inspect.getclosurevars(closure)) # ClosureVars(nonlocals={'x': 10}, globals={}, builtins={}, unbound=set())
```

Qazalin:
Having Buffers on uop is such a good stress test for gc we've found like 3-4 bugs in ops just from this xd.

PR #8265, opened 2024-12-15
Fix double => half cast on clang

uuuvn:
This is a less generic version of #8262 with only one of two cases of external linkage fixed and tests assuming that there is no `__truncdfhf2` in global link-map list (might not be true if something else loads an object with RTLD_GLOBAL that has it). 80% of the value for 20% of the lines
The second one (math functions like sqrtf) are harder to test for on a tight line budged because they are present in libc/libm. See the aforementioned #8262  for the simplest way i've found to do that

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -52,9 +52,6 @@ def _test_cast(a:Tensor, target_dtype:DType):
   if target_dtype == dtypes.half and Device.DEFAULT == "PYTHON":
     # TODO: struct.pack cannot pack value > 65504 (max of half) into e format
     a = (a > 65504).where(65504, a)
-  if CI and Device.DEFAULT == "CLANG" and (target_dtype, a.dtype) in [(dtypes.double, dtypes.half), (dtypes.half, dtypes.double)]:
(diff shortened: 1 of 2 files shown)


geohot:
Can you add a currently failing test that this fixes.

uuuvn:
> Can you add a currently failing test that this fixes.

```
FAILED test/test_dtype.py::TestDoubleDType::test_casts_from - OSError: /tmp/tmpthoz099r: undefined symbol: __truncdfhf2
FAILED test/test_dtype.py::TestHalfDType::test_casts_to - OSError: /tmp/tmpq7k9h3wv: undefined symbol: __truncdfhf2
```
There are already multiple failing tests, they were just disabled and i've already reenabled them with [this](https://github.com/tinygrad/tinygrad/pull/8265/files#diff-cec88baf0c5b9632b4ee303d2553727a09624b421da4282ab8888e2cc3b34a1aL55) change.

uuuvn:
https://github.com/uuuvn/tinygrad/actions/runs/12341013526/job/34439139886?pr=2#step:15:356

CI run with those tests failing when i remove PatternMatcher

geohot:
Cool, merged

Issue #3961, opened 2024-03-27
kernel index can overflow int32


chenyuxyz:
#3271 and beam searching resnet for example

- [x] assert if index > int32  #4157
- [ ] fix linearizer and check index max, use int64 if needed. assert if index > int64

Issue #3271, opened 2024-01-29
Matrix multiplication error


andresgit:
I'm on the latest master running this code:

```python
from tinygrad import Tensor
m1 = Tensor.rand((4096,599*64), device="cuda").realize()
m2 = Tensor.rand((599*64,1024), device="cuda").realize()
res = m1@m2
res.realize()
```
It gives me this error:
```python
Traceback (most recent call last):
  File "/workspace/temp.py", line 5, in <module>
    res.realize()
  File "/workspace/tinygrad/tinygrad/tensor.py", line 116, in realize
    run_schedule(self.lazydata.schedule())
  File "/workspace/tinygrad/tinygrad/realize.py", line 48, in run_schedule
    prg = lower_schedule_item(si)
  File "/workspace/tinygrad/tinygrad/realize.py", line 39, in lower_schedule_item
    return Device[si.out.device].get_runner(si.ast)
  File "/workspace/tinygrad/tinygrad/device.py", line 349, in get_runner
    def get_runner(self, ast:LazyOp) -> CompiledASTRunner: return self.to_program(self.get_linearizer(ast))
  File "/workspace/tinygrad/tinygrad/device.py", line 322, in to_program
    return CompiledASTRunner(k.ast, k.name, self.compiler.render(to_function_name(k.name), k.uops), self, k.global_size, k.local_size)
  File "/workspace/tinygrad/tinygrad/device.py", line 286, in __init__
    self.lib, self.clprg = lib, self.device.runtime(self.name, lib)
  File "/workspace/tinygrad/tinygrad/runtime/ops_cuda.py", line 57, in __init__
    self.module = init_c_var(cuda.CUmodule(), lambda x: check(cuda.cuModuleLoadData(ctypes.byref(x), lib)))
  File "/workspace/tinygrad/tinygrad/helpers.py", line 209, in init_c_var
    def init_c_var(ctypes_var, creat_cb): return (creat_cb(ctypes_var), ctypes_var)[1]
  File "/workspace/tinygrad/tinygrad/runtime/ops_cuda.py", line 57, in <lambda>
    self.module = init_c_var(cuda.CUmodule(), lambda x: check(cuda.cuModuleLoadData(ctypes.byref(x), lib)))
  File "/workspace/tinygrad/tinygrad/runtime/ops_cuda.py", line 28, in check
[...]

PR #8268, opened 2024-12-16
upcast overflowed idx to int64 [pr]

mesozoic-egg:
The approach is to check for overflow in `idx` and `valid` generated by `to_indexed_uops`

If a node overflow, its srcs need to be checked to see if this overflows is a result of an ALU operation, or that the node simply inherits the dtype from srcs. 

That means the search has to be bottom up, and upcasting is either `Ops.CAST + replace` or just `replace`. Also need to make sure `Ops.CAST` does not get in the way of `vmin` and `vmax` calculation.

Pattern matcher is tricky to use because 1. matching require calculating `vmin` `vmax`, 2. tree should not be mutated directly to handle Ops.CAST in vmin, vmax calculation 

Casting back to int is required because if a node is within int32 (e.g. MOD node), its siblings will also be int32 and those siblings will not be upcasted when checking for vmin and vmax

Example kernel:
```c++
extern "C" __global__ void __launch_bounds__(1) test(float* data0, float* data1) {
  int gidx0 = blockIdx.x; /* 129 */
  int gidx1 = blockIdx.y; /* 4096 */
  int gidx2 = blockIdx.z; /* 4096 */
  bool alu0 = (((gidx0<1)!=1)&(((gidx1<1)!=1)&((gidx2<1)!=1)));
  long alu1 = (((long)(gidx0))+((long)((gidx1*129)))+(((long)(gidx2))*528384ll));
  float val0 = (alu0?*(data1+alu1):0.0f);
  if (alu0) {
    *(data0+alu1) = (val0+(alu0?1.0f:0.0f));
  }
}
```

example kernel for the cast back case:
```c++
extern "C" __global__ void __launch_bounds__(1) test(float* data0, float* data1) {
  int gidx0 = blockIdx.x; /* 129 */
  int gidx1 = blockIdx.y; /* 4096 */
  int gidx2 = blockIdx.z; /* 4096 */
  bool alu0 = (((gidx0<1)!=1)&(((gidx1<1)!=1)&((gidx2<1)!=1)));
  int cast0 = ((int)((((long)(gidx0))+((long)((gidx1*129)))+(((long)(gidx2))*528384ll)+-1073741824ll)));
  float val0 = (alu0?*(data1+cast0):0.0f);
  if (alu0) {
    *(data0+cast0) = (val0+(alu0?1.0f:0.0f));
  }
}
```


tinygrad/renderer/wgsl.py
```diff
@@ -49,7 +49,8 @@ class WGSLRenderer(CStyleLanguage):
   code_for_op = {**CStyleLanguage.code_for_op, Ops.WHERE: lambda a,b,c,dtype: f"select({c},{b},{a})"}
   nan = "nan()"
   type_map = { dtypes.float: "f32", dtypes.uchar: "u32", dtypes.ushort: "u32", dtypes.short: "i32",
-              dtypes.char: "i32", dtypes.int32: "i32", dtypes.uint32: "u32", dtypes.bool: "bool" }
+              dtypes.char: "i32", dtypes.int32: "i32", dtypes.uint32: "u32", dtypes.bool: "bool",
+              dtypes.int64: "i32" } # wgsl has no int64, shapetracker upcast becomes a no-op
```
chenyuxyz, line 53: i think we want to assert if it needs int64 and device does not support it. it's an illegal memory access that either fails silently or crashes

tinygrad/shape/shapetracker.py
```diff
@@ -74,7 +74,10 @@ def reduce(self, axis:Tuple[int, ...]) -> Tuple[sint, ...]: return tuple(1 if i
 
   def to_uop(self) -> UOp: return UOp(Ops.VIEW, dtypes.void, (), self)
   def to_indexed_uops(self, _idxs:Optional[List[UOp]|Tuple[UOp, ...]]=None) -> Tuple[UOp, UOp]:
-    return views_to_indexed_uops(self.views, tuple(_idxs) if _idxs is not None else None)
+    idx, valid = views_to_indexed_uops(self.views, tuple(_idxs) if _idxs is not None else None)
+    def upcast(u: UOp): return UOp(u.op, dtypes.int64 if u.dtype is not dtypes.bool else u.dtype, arg=u.arg, src=tuple(upcast(_u) for _u in u.src))
```
chenyuxyz, line 78: how about vmin?

chenyuxyz:
this looks very complex, can it be simpler? what's wrong with the old approach

also don't test based on custom `r`, `e` functions, it's hard to maintain

mesozoic-egg:
the old approach (upcast in to_index_op) could result in partial upcasting when only 1 Ops.VIEW overflows. But I'll revisit and try to simplify things

mesozoic-egg:
Turns out I shouldn't be replacing all idx with long, but should instead do an explicit cast when it actually exceeds limit. This should be on the right track, will look for a better way to test this instead of hard coding UOps

tinygrad/renderer/cstyle.py
```diff
@@ -75,6 +75,7 @@ class CStyleLanguage(Renderer):
   extra_args: list[str] = []
   float4: Optional[str] = None
   type_map: dict[DType, str] = {}
+  invalid_dtype: dict[DType, None] = {}
```
davidjanoskyrepo, line 78: Why not just a list?

mesozoic-egg: I think it could be a list as well, other parts of the code base use a dict for such lookup purpose so I just followed the style

khpark43: Other parts of the code base use `dict[sth, None]` to preserve order, I guess set is correct.

chenyuxyz, line 78: use `is_dtype_supported`

test/test_uop_graph.py
```diff
@@ -712,6 +712,90 @@ def test_expand_ifs_dumb(self):
     for st in sink.src:
       self.assertEqual(len(st.src), 2)
 
+@unittest.skipIf(Device.DEFAULT == "WEBGPU", "WEBGPU does not support int64 upcasted indexing")
```
chenyuxyz, line 715: skipUnless `is_dtype_supported`

test/test_uop_graph.py
```diff
@@ -712,6 +712,90 @@ def test_expand_ifs_dumb(self):
...
+  def test_symbolic_fold(self):
+    st = ShapeTracker(views=(View(shape=(2097153, 4194303), strides=(0, 0), offset=0, mask=((0, 2097153), (2097151, 4194303)), contiguous=False),
+                             View(shape=(2097152, 2097152), strides=(1, 4194304), offset=0, mask=None, contiguous=False)))
+    _, valid = st.to_indexed_uops()
+    cmplt = self.find_ops_in_ast(valid, Ops.CMPLT)
+    assert all((_src.dtype is dtypes.int for _src in cmplt.src))
+
+
+@unittest.skipUnless(Device.DEFAULT == "WEBGPU", "Upcasted indexing fail on webgpu because of no int64 support")
+class TestIndexingOverflowWEBGPU(unittest.TestCase):
```
chenyuxyz, line 787: test name should not contain WEBGPU?

chenyuxyz (commented):
should also tested from higher level, like in test_tensor, test the one in https://github.com/tinygrad/tinygrad/issues/3271 (mentioned in https://github.com/tinygrad/tinygrad/issues/3961)

think about how you can write less custom functions in the test. test is part of the code and need to be maintained. How likely do we need (and how easy) to update these tests if we update uop spec etc

mesozoic-egg:
Removed [pr] in title because the currently this isn't upcasted:
```python
Tensor.empty(2**11, 2**11, 1).permute((2, 1, 0)).expand(2**11, -1, -1).contiguous())
```
Screenshot 2025-01-04 at 02 54 45 1

i think i have the correct implementation now, need to convert the hand written uop tests into higher level tensor tests

test/test_tensor.py
```diff
@@ -764,5 +767,71 @@ def test_complex_backward(self):
     self.assertEqual(len(bw), 1)
     self.assertEqual(bw[0].name, "sigmoid")
 
+dtype_supported = (is_dtype_supported(dtypes.long), "int64 is supported")
```
chenyuxyz, line 770: this is bad for maintainability. it's fine to copy these in tests to make the scope local and easier to update

tinygrad/codegen/lowerer.py
```diff
@@ -6,6 +6,30 @@
 from tinygrad.ops import KernelInfo, UOp, Ops, graph_rewrite, PatternMatcher, UPat, sint, identity_element, sint_to_uop
 from tinygrad.renderer import Renderer
 from tinygrad.helpers import all_int, prod, partition, flatten
+from tinygrad.codegen.uopgraph import sym
+from tinygrad.device import is_dtype_supported
+
+def overflow(u, dtype): return u.vmax > dtypes.max(dtype) or u.vmin < dtypes.min(dtype)
```
chenyuxyz, line 12: type annotation, also if it's only used for int32 and int64, start with just that pair. other combination is not tested

tinygrad/codegen/lowerer.py
```diff
@@ -6,6 +6,30 @@
...
+
+# If a node overflow, its srcs need to be checked to see if this overflow is the result of an ALU operation,
+# or that the node simply inherits the dtype from srcs. Upcast is either `Ops.CAST`+`replace` or just `replace`.
+def upcast(u: UOp):
+  srcs = [upcast(_src) for _src in u.src]
+  if u.dtype.scalar() is dtypes.int:
+    dtype = dtypes.int64.vec(u.dtype.count) if u.dtype.count > 1 else dtypes.int64
+    upcasted = u.replace(dtype=dtype, src=tuple([_src.cast(dtype) for _src in srcs]))
+    if overflow(u, u.dtype):
+      assert is_dtype_supported(dtypes.long), f"UOp overflow without int64 support, {u.op=} {u.dtype=}"
```
chenyuxyz, line 22: don't think this is correct because `is_dtype_supported` tests against Device.default, not the device of UOp. but generally i am not sure if you want to assert here.

tinygrad/codegen/lowerer.py
```diff
@@ -106,6 +130,7 @@ def lower_reduce_axis(ctx: IndexContext, x: UOp):
 
 def lower_load_store(ctx: IndexContext, x: UOp):
   idx, valid = x.st_arg.to_indexed_uops(ctx.ridxs if x.op is Ops.LOAD and x.src[0].op is Ops.DEFINE_LOCAL else ctx.idxs)
```
chenyuxyz, line 132: any reason `folded_upcast` is not done in `to_indexed_uops`?

mesozoic-egg: `def real_size` uses `to_indexed_uops` and would break vmax limit if there's an Ops.CAST in the UOp tree. `to_program` uses this vmax for mem calculation.
image

tinygrad/codegen/lowerer.py
mesozoic-egg: I tried a few approach to assert and all of them seem a bit overcomplicated (type_verify, passing renderer.device to is_dtype_supported, etc.). I ended up putting in the wgsl render method to assert nothing is int64 because it feels like a trivial case that doesn't come up often. Would like to know what you think

mesozoic-egg:
@chenyuxyz real_size is now fixed, it uses the non-upcasted version from views_to_indexed_uops. I put dtype assert in render, instead of linearizer because we have emulated CI test for float16 that require lowering the datatype, but not rendering them. Putting a dtype checker there would break those test, and adding an emulate flag for the checker does not seem worth the complexity.

test/test_tensor.py
```diff
@@ -764,5 +767,70 @@ def test_complex_backward(self):
...
+    # Assert the dtype of the INDEX value, This will need be updated if UOp spec changes
+    store = next(uop for uop in prg.uops if uop.op is Ops.STORE)
+    assert store.op is Ops.STORE
+    idx = self._find_op(store, Ops.INDEX)
+    if idx is not None: # PTX turns Ops.INDEX into pointer arithmetic earlier than cstyle, plus it's already cast to int64
+      assert idx.op is Ops.INDEX
+      idx_val = idx.src[1]
+      assert idx_val.dtype is dtype
+
+  # This prevents kernel.py from combining the dims into 1
```
chenyuxyz, line 799: can these test in a way that does not dependent on kernel.py implementation details?

tinygrad/device.py
```diff
@@ -255,7 +255,7 @@ def is_dtype_supported(dtype:DType, device:Optional[str]=None) -> bool:
   if dtype == dtypes.bfloat16:
     # NOTE: this requires bf16 buffer support
     return device in {"AMD"} or (device in {"CUDA", "NV"} and not CI and not getenv("PTX"))
-  if device == "WEBGPU": return dtype in [dtypes.bool, dtypes.char, dtypes.uchar, dtypes.short,
+  if device == "WEBGPU": return dtype in [dtypes.void, dtypes.bool, dtypes.char, dtypes.uchar, dtypes.short,
```
chenyuxyz, line 258: is this needed? it's an internal dtype so either it's supported for all or not supported for all

chenyuxyz (commented):
why is process replay not passing? is any test realizing kernels?

mesozoic-egg:
ready for a final review now : ) @chenyuxyz

chenyuxyz (commented):
cool it's closer! minor comments

test/test_tensor.py
```diff
@@ -764,5 +769,66 @@ def test_complex_backward(self):
...
+  @unittest.skipIf(is_dtype_supported(dtypes.long), "int64 is supported")
+  @unittest.expectedFailure
+  def test_int64_unsupported_overflow_sym(self):
+    self.do_op_then_assert(dtypes.long, 2048, 2048, UOp.variable("dim3", 0, 2048).bind(32))
+
+  @unittest.skipIf(is_dtype_supported(dtypes.long), "int64 is supported")
+  @unittest.expectedFailure
+  def test_int64_unsupported_overflow(self):
+    self.do_op_then_assert(dtypes.long, 2048, 2048, 2048)
+
```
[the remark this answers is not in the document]
chenyuxyz: also a test that shows the error if backend does not support int64

mesozoic-egg:
Thx. Updated `DType` for typehint, `src` uses a tuple, and to show the error on int64 unsupported backend in tests, I'm asserting on the actual `KeyError` rasied by the renderer.

I'm not sure how summing a cross product would generate a kernel that overflows int64 though. Isn't cross product (3d vector) just some multiplication addition between six variables? ((𝑢2𝑣3−𝑢3𝑣2)𝐢−(𝑢1𝑣3−𝑢3𝑣1)𝐣+(𝑢1𝑣2−𝑢2𝑣1)𝐤). Or are you referring to something else? 

@chenyuxyz

chenyuxyz:
oh sorry i mean outer product

mesozoic-egg:
Doesn't index overflow by definition require the number of element to exceed int32? It would at least require having 2**31 number of int8 to achieve that, which is 2GB of memory usage. There might be something obvious I'm missing here, appreciate it if you could point it out. 

When summing outer product, the intermediate outer product never need to be realized, so indexing stay within the size of the two input vectors. Unless those tensors are big or the intermediate buffer is realized, overflow never occurs. For example:

```python
dim = 4096
a = Tensor.arange(dim).reshape((dim, 1)).realize()
b = Tensor.arange(dim).reshape((1, dim)).realize()
(a * b).sum().realize()
```

would produce a kernel similar to this
```c++
extern "C" __global__ void __launch_bounds__(256) r_256_16_4096(int* data0, int* data1, int* data2) {
  __shared__ int temp0[256];
  int lidx0 = threadIdx.x; /* 256 */
  int acc0 = 0;
    int val0 = *(data1+((lidx0<<4)+ridx0));
    for (int ridx1 = 0; ridx1 < 4096; ridx1++) {
      int val1 = *(data2+ridx1);
      acc0 = (acc0+(val0*val1));
    }
  }
  *(temp0+lidx0) = acc0;
  __syncthreads();
  if ((((bool)(lidx0))!=1)) {
    int acc1 = 0;
    for (int ridx2 = 0; ridx2 < 256; ridx2++) {
      int val2 = *(temp0+ridx2);
      acc1 = (acc1+val2);
    }
    *(data0+0) = acc1;
  }
}
```
The max index into each tensor is limited to the two vectors' length, and the index can never overflow, unless the two vectors themselves are very big.

Also, the issue mentioned in [3271](https://github.com/tinygrad/tinygrad/issues/3271#issuecomment-1923515761) occurred in the 9GB intermediate buffer.

@chenyuxyz

chenyuxyz:
cool i think this is ready. for future reference, can you give an example script that failed on master and works with this change? the one in 3271 somehow works now

mesozoic-egg:
Sure! I added a skipped `test_tensor.py::TestIdxUpcast::test_overfow_kernel_run` that reproduce the error if you run it on master, requires 2GB of memory to run.

```
...
RuntimeError: CUDA Error 700, an illegal memory access was encountered
```

The failed kernel is:

```c++
extern "C" __global__ void __launch_bounds__(48) E_58_262144_3_16_3(signed char* data0, signed char* data1) {
  int gidx0 = blockIdx.x; /* 262144 */
  int gidx1 = blockIdx.y; /* 58 */
  int lidx0 = threadIdx.x; /* 3 */
  int lidx1 = threadIdx.y; /* 16 */
  int alu0 = (gidx0<<4);
  signed char val0 = *(data1+(lidx1+alu0));
  int alu1 = (lidx1+alu0+(gidx1*37748736)+(lidx0*12582912));
  *(data0+alu1) = val0;
  *(data0+(alu1+4194304)) = val0;
  *(data0+(alu1+8388608)) = val0;
}
```

This is the expected kernel running on the PR's branch:

```c++
extern "C" __global__ void __launch_bounds__(48) E_58_262144_3_16_3(signed char* data0, signed char* data1) {
  int gidx0 = blockIdx.x; /* 262144 */
  int gidx1 = blockIdx.y; /* 58 */
  int lidx0 = threadIdx.x; /* 3 */
  int lidx1 = threadIdx.y; /* 16 */
  int alu0 = (gidx0<<4);
  signed char val0 = *(data1+(lidx1+alu0));
  long alu1 = (((long)(lidx1))+((long)((lidx0*12582912)))+((long)(alu0))+(((long)(gidx1))*37748736ll));
  *(data0+alu1) = val0;
  *(data0+(alu1+4194304ll)) = val0;
  *(data0+(alu1+8388608ll)) = val0;
}
```


The reason 3271 failed was because matmul tried to store an intermediate tensor of size `(4096, 1024, 599)` which is 
a total of 2512388096 element (bigger than 2**31) when doing `((4096,599*64) @ (599*64,1024)`:

```
  0 ━┳ STORE MemBuffer(idx=0, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4096, 1024, 599, 1), strides=(613376, 599, 1, 0), offset=0, mask=None, contiguous=True),)))
  1  ┗━┳ SUM (4096, 1024, 599, 1)
  2    ┗━┳ MUL 
  3      ┣━━ LOAD MemBuffer(idx=1, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4096, 1024, 599, 64), strides=(38336, 0, 64, 1), offset=0, mask=None, contiguous=False),)))
  4      ┗━━ LOAD MemBuffer(idx=2, dtype=dtypes.float, st=ShapeTracker(views=(View(shape=(4096, 1024, 599, 64), strides=(0, 1, 65536, 1024), offset=0, mask=None, contiguous=False),)))
```

But it seems like tinygrad no longer handles matmul the same way? I'm not able to get this same UOp and not sure how things were handled back then (the issue was from almost a year ago).

tinygrad/shape/shapetracker.py
```diff
@@ -70,12 +89,16 @@ def reduce(self, axis:tuple[int, ...]) -> tuple[sint, ...]: return tuple(1 if i
...
-    return views_to_indexed_uops(self.views, tuple(_idxs) if _idxs is not None else None)
+    idx, valid = views_to_indexed_uops(self.views, tuple(_idxs) if _idxs is not None else None)
+    return folded_upcast(idx), folded_upcast(valid)
 
   # upper bound on buffer size required to fit this shapetracker
   def real_size(self) -> int:
     if 0 in self.shape: return 0
-    return int((v.shrink(v.mask) if (v:=self.views[0]).mask else v).to_indexed_uops()[0].vmax + 1)
+    view = (v.shrink(v.mask) if (v:=self.views[0]).mask else v)
+    idx, _ = views_to_indexed_uops((view,))
```
mesozoic-egg, line 99: This oneliner would be barely readable if I wrap `views_to_indexed_uops` around it, so I just expanded it to three lines. I think the assertion should be kept though.

chenyuxyz:
cool! once ci passes, bounty is yours. george@tinygrad.org to claim

PR #8280, opened 2024-12-17
assert to prepare for grad uop [pr]

geohot:


geohot:
what's still missing?

Qazalin:
yeah CAST_BEFORE_VIEW=1 in ops isn't gonna be correct for this one.
This really needs the scheduler tensor_map refactor. I'm working on finishing that by EOW. #8495

PR #8287, opened 2024-12-17
viz profiler

nimlgen:
hcq part


nimlgen:
perfetto is a sep html for now. maybe this a good way to clean up. Apart from that profiling works in VIZ now!

PR #8292, opened 2024-12-17
Tensor core refactor

ignaciosica:
This PR showcases progress on the tensor core refactor. The main question revolves around the swizzling of the view, particularly whether pushing the `st` itself when the shape matches is the right approach. While I'm not entirely sure about the new spec, this method was the cleanest way to fix the `ShapeTracker` for WMMA outputs.

### Overview

The refactor focuses primarily on the implementation side of tensor cores (TC) rather than their abstraction, as George mentioned. While an abstraction refactor could align with the kernel OptOps refactor into pattern-matching (PM) rules, the current work keeps tensor cores clean. This approach addresses not only the intricate thread patterns required for specific backends but also the simplicity of supporting new TC shapes.

### Implementation details

**Kernel** `OptOps`

Each tensor core now takes the number of cooperating threads for WMMA as input, instead of relying on specific thread patterns. Using this thread count, the kernel localizes as much as possible from the TC dimensions (N, M), upcasting the remainder. The reduced dimension (K) continues to be unrolled as before.

All kernel OptOps use a fixed value of 2, enabling a more straightforward permutation of shapes.

**New Features**

1. `upcast_size`
A new property, `upcast_size`, specifies the input size for WMMA instructions. Although this could theoretically be inferred from the st of the resulting source for the WMMA uop, padded tensor cores make this logic overly complex. Explicitly defining upcast_size simplifies this.
2. `swizzle`
The `st1_pattern` and `st2_pattern` properties have been replaced with `swizzle`, a tuple of three optional permutations describing the thread patterns for each backend. This update enables swizzling for outputs as well, allowing us to use documented thread patterns instead of relying on trial and error.

### Advantages of the refactor

The new implementation reduces the complexity of adding support for new tensor core shapes. Previously, the process required defining thread patterns, reduce axes, upcast axes, expanded shapes, and local/upcast order, all being specific to Tinygrad. Now, only three elements are necessary:

1. Tensor core dimensions (N, M, K)
2. Number of collaborating threads
3. upcast_size for the WMMA instruction

These values are readily available in backend documentation and will get the tc running. While swizzling still needs to be implemented manually to get the correct result, this approach eliminates concerns about Tinygrad-specific details.

### Making sense of the swizzle

In order to explain the swizzle, I'm going to make a walkthrough in adding `tf32` tc support for CUDA. As a sidenote; I'm tempted to add the support for the other TC shapes in order to claim the bounties (Turing tc, int8 tc, ...) but it would be a good sign if somebody else could figure them out somehow simply.

Form the docs, we learn that the tensor core shape for `sm_80` with `tf32` as input dtype is:

image

From the documentation, we obtain all the necessary information to add the tensor core (TC) to the backend (dims, threads and upcast_size). The tensor core can then be added to the backend as follows:
```python
tensor_cores += [TensorCore(dims=(8,16,8), threads=32, upcast_size=(4,2,4), dtype_in=dtypes.float, dtype_out=dtypes.float)]
```

To determine the swizzle, the simplest approach is to perform a matrix multiplication with the same shape as the tensor core being configured. Running the following command: `DEBUG=3 N=8 M=16 K=8 python extra/gemm/simple_matmul.py` yields:

``` bash
TENSOR CORES [(1, 8, 1)] [(0, 16, 8)] WMMA_8_16_8_float_float
r_2_2_2_2_2_2_2_2_2_2
    UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(), arg=0, src=()),
    UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 1, 1, 1, 2, 2), strides=(1, 2, 4, 32, 64, 0, 0, 0, 8, 16), offset=0, mask=None, contiguous=False),)), src=()),
      UOp(Ops.WMMA, dtypes.float.vec(4), arg=('WMMA_8_16_8_float_float', (8, 16, 8), dtypes.float, dtypes.float, 'CUDA', 32, (((9, 2), (8, 2)), ((9, 2),), ((9, 2), (8, 2))), (5, 6, 7)), src=(
        UOp(Ops.CONTRACT, dtypes.float.vec(4), arg=((9, 2), (8, 2)), src=(
          UOp(Ops.LOAD, dtypes.float, arg=None, src=(
            UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(), arg=1, src=()),
            UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(0, 0, 0, 32, 64, 1, 2, 4, 8, 16), offset=0, mask=None, contiguous=False),)), src=()),)),)),
        UOp(Ops.CONTRACT, dtypes.float.vec(2), arg=((9, 2),), src=(
          UOp(Ops.LOAD, dtypes.float, arg=None, src=(
            UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(), arg=2, src=()),
            UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(1, 2, 4, 0, 0, 8, 16, 32, 0, 0), offset=0, mask=None, contiguous=False),)), src=()),)),)),
        UOp(Ops.CONST, dtypes.float.vec(4), arg=0.0, src=()),)),)),)),))
*** CUDA      14 r_2_2_2_2_2_2_2_2_2_2                     arg  3 mem  0.00 GB tm     12.42us/     2.26ms (     0.16 GFLOPS    0.1|0.2     GB/s) ['matmul']
...
AssertionError: 
Not equal to tolerance rtol=0.03, atol=0.0001

Mismatched elements: 124 / 128 (96.9%)
Max absolute difference among violations: 2.8756514
Max relative difference among violations: 3.9030735
 ACTUAL: array([[2.556512, 2.556512, 2.556512, 2.556512, 2.556512, 2.556512,
        2.556512, 2.556512],
       [3.623818, 3.623818, 3.623818, 3.623818, 3.623818, 3.623818,...
 DESIRED: array([[1.895115, 1.126101, 2.415895, 1.189833, 1.251278, 1.979367,
        2.423452, 1.761977],
       [2.318467, 1.628582, 3.357059, 2.569354, 2.071491, 2.852631,...
```

While the tensor core runs successfully, the assertion error indicates that swizzling has not yet been performed. To resolve this, we need to configure the strides in the ShapeTracker such that each thread is responsible for loading and storing the elements as specified in the documentation. In order to figure out the required swizzle, we are going to focus on the current ShapeTrackers for the buffers: 

``` python
A   => ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(0, 0, 0, 32, 64, 1, 2, 4, 8, 16), offset=0, mask=None, contiguous=False)
B   => ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(1, 2, 4, 0, 0, 8, 16, 32, 0, 0), offset=0, mask=None, contiguous=False),
C/D => ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 1, 1, 1, 2, 2), strides=(1, 2, 4, 32, 64, 0, 0, 0, 8, 16), offset=0, mask=None, contiguous=False)
```

Since the matrix multiplication matches the shape of the tensor core in use, all dimensions are either localized, reduced, or upcasted by TC OptOps. This means there are no global dimensions or extra reduce dimensions.

With a thread count of 32 and each thread dimension fixed to 2, the resulting shape for local dimensions is `(2, 2, 2, 2, 2)`, as `32 = 2 ** 5`. The product of `M * N = 8 * 16 = 4 * 32` leaves a remainder of 4 that cannot be localized, so it is upcasted, resulting in an upcasted shape of `(2, 2)`.

The reduce dimension `K = 8` is split into `(2, 2, 2)`. this results in all st having 10 dimensions of size 2 (or size 1 for reduced dimensions in the WMMA output).

The shape tracker configuration is as follows: `(locals)(reduce)(upcast) = (2, 2, 2, 2, 2) (2, 2, 2) (2, 2)`

From the docs we get the desired strides for the threads and upcasts.

image

In the figure, color shading represent the threads. Threads are continuous from darker to lighter and then they follow the dotted line. That represent the elements that each thread is responsible for loading into the tensor core. Thread0 (T0) is responsible for loading a0-a3, b0-b1 and stores c0-c3. Thread1 (T1) is responsible for the following, lighter orange, elements.

From the figure we can get the desired strides for locals for each buffer and also for the upcasting.

desired local // upcast strides for A = 1, 2, 8, 16, 32 // 64, 4
desired local // upcast strides for B = 8, 16, 1, 2, 4 // 32
desired local // upcast strides for C = 2, 4, 8, 16, 32 // 1, 64

In order to get the swizzle permutations, we just need to start from the real strides of the shapetracker and permute to the desired strides for fixing the buffer.

image

The new implementation automatically upcasts along the non 0 dimensions from the reduced+upcasted axes.

Now, if we introduce the permutations we get from the figures as the swizzle and run the same test as before, we get the correct result. We can see the shapetrackers have been fixed, getting the desired strides from the figures above.
``` python
tensor_cores += [TensorCore(dims=(8,16,8), threads=32, upcast_size=(4,2,4), dtype_in=dtypes.float, dtype_out=dtypes.float,
			    swizzle=(((5,6,8,9,3),(0,1,2,4,7)),((5,6,0,1,2),(3,4,8,9,7)),((1,2,8,9,3),(5,6,7,0,4))))]
```

``` bash
TENSOR CORES [(1, 8, 1)] [(0, 16, 8)] WMMA_8_16_8_float_float
r_2_2_2_2_2_2_2_2_2_2
    UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(), arg=0, src=()),
      UOp(Ops.WMMA, dtypes.float.vec(4), arg=('WMMA_8_16_8_float_float', (8, 16, 8), dtypes.float, dtypes.float, 'CUDA', 32, (((9, 2), (8, 2)), ((9, 2),), ((9, 2), (8, 2))), (5, 6, 7)), src=(
        UOp(Ops.CONTRACT, dtypes.float.vec(4), arg=((9, 2), (8, 2)), src=(
          UOp(Ops.LOAD, dtypes.float, arg=None, src=(
            UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(), arg=1, src=()),
            UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(1, 2, 8, 16, 32, 0, 0, 0, 64, 4), offset=0, mask=None, contiguous=False),)), src=()),)),)),
        UOp(Ops.CONTRACT, dtypes.float.vec(2), arg=((9, 2),), src=(
          UOp(Ops.LOAD, dtypes.float, arg=None, src=(
            UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(), arg=2, src=()),
            UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(8, 16, 1, 2, 4, 0, 0, 0, 0, 32), offset=0, mask=None, contiguous=False),)), src=()),)),)),
        UOp(Ops.CONST, dtypes.float.vec(4), arg=0.0, src=()),)),)),)),))
*** CUDA      14 r_2_2_2_2_2_2_2_2_2_2                     arg  3 mem  0.00 GB tm     26.40us/     2.21ms (     0.08 GFLOPS    0.0|0.0     GB/s) ['matmul']
opened device CLANG from pid:727791
*** CLANG     15 copy      512,   CLANG <- CUDA            arg  2 mem  0.00 GB tm    101.26us/     2.31ms (     0.00 GFLOPS    0.0|0.0     GB/s) 
*** CLANG     16 copy      256,   CLANG <- CUDA            arg  2 mem  0.00 GB tm     64.78us/     2.37ms (     0.00 GFLOPS    0.0|0.0     GB/s) 
*** CLANG     17 copy      512,   CLANG <- CUDA            arg  2 mem  0.00 GB tm     38.89us/     2.41ms (     0.00 GFLOPS    0.0|0.0     GB/s) 
```

### Generated kernel for tf32 wmma
``` c
__device__ float4 __WMMA_8_16_8_float_float(float4 a, float2 b, float4 c){
  asm("mma.sync.aligned.m16n8k8.row.col.f32.tf32.tf32.f32"
      "{%0, %1, %2, %3}, {%4, %5, %6, %7},"
      "{%8, %9}, {%0, %1, %2, %3};"
    : "r"(a_pk[0]), "r"(a_pk[1]), "r"(a_pk[2]), "r"(a_pk[3]), "r"(b_pk[0]), "r"(b_pk[1]));
  return c;
}
extern "C" __global__ void __launch_bounds__(32) r_2_2_2_2_2_2_2_2_2_2(float* data0, float* data1, float* data2) {
  int alu0 = (lidx1<<4);
  int alu1 = (lidx2<<5);
  int alu2 = (lidx0>>2);
  int alu3 = (alu2<<3);
  int alu4 = (lidx0&1);
  int alu5 = ((lidx0>>1)&1);
  int alu6 = ((alu5<<1)+alu4+alu3+alu0+alu1);
  float val0 = *(data1+alu6);
  float val1 = *(data1+(alu6+4));
  float val2 = *(data1+(alu6+64));
  float val3 = *(data1+(alu6+68));
  int alu7 = ((alu4<<3)+(alu5<<4)+alu2+(lidx1<<1)+(lidx2<<2));
  float val4 = *(data2+alu7);
  float val5 = *(data2+(alu7+32));
  float4 wmma0 = __WMMA_8_16_8_float_float(make_float4(val0,val2,val1,val3), make_float2(val4,val5), make_float4(0.0f,0.0f,0.0f,0.0f));
  int alu8 = ((alu4<<1)+(alu5<<2)+alu3+alu0+alu1);
  *((float2*)((data0+alu8))) = make_float2(wmma0.x,wmma0.y);
  *((float2*)((data0+(alu8+64)))) = make_float2(wmma0.z,wmma0.w);
}
```

Feel free to ask any questions. My goal is to make this guide as clear, readable, and easy to follow as possible. Questions about implementation are more than welcome also.


ignaciosica:
image

ignaciosica:
image

ignaciosica:
Based on the following docs ([intel subgroup gemm](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroup_matrix_multiply_accumulate.html)) I refactored intel xmx tensor cores. Couldn't test on a real device yet though.

image

ignaciosica:
I just realized I pushed as nimlgen. Sorry about that, reverting

ignaciosica:
image

ignaciosica:
I removed Intel bf16 Tensor Cores since they fail the tc_padded test (all other tests pass). The issue seems related to casting; previously, it was completely broken due to a bug in the Intel string_rewrite pattern matching. I'm marking this as ready for review because I believe it's independent of the work in this PR and can be addressed in a subsequent, standalone PR.

geohot:
I'm trying to understand this. I see this is swizzling the output of the tensor cores, then pushing a permute if it's the same shape (is there a reason this is the right heuristic, it seems fragile to things like 1s in shape).

What changed that the output needs to be swizzled? The old code doesn't require this.

tinygrad/codegen/kernel.py
```diff
@@ -596,48 +594,43 @@ def reduced_axes(start, stop):
...
+              UOp.const(tc.dtype_out.vec(tc.upcast_size[2]), 0.0)), arg=wmma_arg)
+            tc_uop = UOp(Ops.UNROLL, tc.dtype_out, (wmma,), arg=tc_upcast_axes[2])
 
           else: # for TC=3 MUL/SUM instead of WMMA
             tc_uop = UOp(Ops.REDUCE_AXIS, tc.dtype_out, ((srcs[0] * srcs[1]).cast(tc.dtype_out),), (Ops.ADD, tc_reduce_axes))
 
-          new_reduce_axes = tuple(i for i in axes if i not in tc_reduce_axes)
-          return ret.replace(src=(tc_uop,), arg=(Ops.ADD, new_reduce_axes)) if new_reduce_axes else tc_uop
+          ret = ret.replace(src=(tc_uop,), arg=(Ops.ADD, new_axes)) if (new_axes := tuple(i for i in axes if i not in tc_reduce_axes)) else tc_uop
+          return ret.view(get_tc_swizzle_st(unwrap(ret.st).shape, *tc.swizzle[2])) if self.use_tensor_cores == 1 and tc.swizzle[2] else ret
```
geohot, line 633: This isn't right. use_tensor_cores 2 and 3 should use the *exact* same shapetracker patterns

tinygrad/codegen/kernel.py
```diff
@@ -656,7 +649,8 @@ def fix_st(st: ShapeTracker, wd_pattern, tcd_pattern):
 
       return ret
 
-    return graph_rewrite(fixup_ast(self.ast), view_left)
+    return graph_rewrite(fixup_ast(self.ast),  merge_views + PatternMatcher([
```
geohot, line 652: I would not expect a tensor core change to include this, if you really want this it should be a standalone PR.

tinygrad/codegen/kernel.py
ignaciosica: I agree that `TC=3` should match the st swizzle for output as well. For `TC=2`, it never did the st `fix_up`, it only applies TC shape (before my refactor there was an explicit comment stating no st fix_up for TC=2).

> image

I'll make it a draft until I fix the swizzle for `TC=3`, but my guess is that the problem is in the swizzling of the emulated local rather than on the swizzling of the output.

ignaciosica:
> What changed that the output needs to be swizzled? The old code doesn't require this. 

The thread access pattern for the tensor cores did not align with the official documentation for either inputs or outputs. This misalignment was to blame in the trial and error for the swizzle, as the correct result of the tensor cores relied on a new, valid, undocumented thread access pattern (I'm not sure they always exist for the fixed output, that was the reason I was having trouble with `tf32` TC). 

As `thread access pattern` I mean the elements of the tensor core each collaborating thread is in charge of loading (from `A` and `B`) and storing (from `C`).

For larger or more complex shapes, such as WGMMA, this approach isn't practical. The WGMMA swizzle is highly intricate, and if its documentation cannot be easily integrated into Tinygrad's tensor core implementation, the complexity for its introduction will increase significantly.

(I'll try to draw the old thread access pattern for cuda TC to see the discrepancies with the documentation, post it here once I'm done) 

> Is there a way to push the permute up through the tensor core such that it's still only swizzles on the input?

In CUDA, for example, multiple tensor cores share the same shape but may have different swizzle patterns depending on the data type for input and output. That is why having the ability to swizzle the output is important.

tinygrad/codegen/kernel.py
ignaciosica: Did you mean the change to the schedule's view_right spec? I'll move that to a standalone PR

ignaciosica:
closed in favor of #8426

PR #8303, opened 2024-12-18
remove the fake buffer, try 2 [pr]

geohot:


Qazalin:
I would just start by replacing the BUFFER uop in src[0] with a DEVICE. You can use the existing generate_valid infra for it.
That src replace also doesn't work for symbolic BIND.

geohot:
The bigger problem I ran into is that this isn't recursive. I talked about it in TIP 1, basically we need to do const folding before bufferization

PR #8309, opened 2024-12-18
cuda transfer

nimlgen:


geohot:
We need a way to test this in CI

Issue #8294, opened 2024-12-17
TIP 4 - The shape and device of UOps


geohot:
## Abstract

All UOps that are valid in Tensor must have a shape (from a ShapeTracker and a device). This TIP explains the derivation rules for these properties.

## Shape Proposal

Ops.BUFFER is a piece of memory on the GPU. It has shape (<size>,) with stride (1,)

Ops.CONST is a virtual "Buffer" containing one element. It has 0 dimensions and has shape ()

Ops.VIEW can apply on top of the existing ShapeTracker to change it in an additive way

All other UOps pass through the ShapeTracker.

## Device Proposal

Ops.DEVICE is the source of a device

Ops.COPY can change the device

## Result

The `device`, `st`, and `shape` properties on UOps will reflect these derivation rules.

geohot:
I don't understand why Ops.EMPTY exists, shouldn't it just be Ops.BUFFER?

chenyuxyz:
I think Ops.EMPTY predates Ops.BUFFER so maybe we can remove it now. Also there are empty with 0 elements `Tensor([])` that we currently support

geohot:
This spec has an issue.

Ops.BUFFER has a pointer dtype, which the view somehow removes. That makes the VIEW required, which I'd prefer if it weren't.

Qazalin:
> Ops.BUFFER is a piece of memory on the GPU. It has shape (,) with stride (1,)


I propose we change this, add a VIEW source to BUFFER to preserve the output shapetracker of a realized BUFFER.

Currently, a realized BUFFER preserves its output st with VIEW(BUFFER), this causes correctness issues in #8412, we can't merge views while also changing the base shapetracker.

If the VIEW lives in the BUFFER source, we can safely apply movement ops and merge_views in the end:

```
BUFFER(VIEW(ShapeTracker.from_shape(32, 32))).view(...)
```
in the end, becomes:
```
LOAD(BUFFER, base+view)
```

Qazalin:
interesting case to think about if CONST has a shape=():
```
  File "/Users/qazal/code/tinygrad/tinygrad/ops.py", line 289, in st
    assert all_same([x.shape for x in src_sts]), f"UOp parents must have the same shape {self} {[x.shape for x in src_sts]}"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #8311, opened 2024-12-18
Tensor uop spec

Qazalin:
This PR adds asserts for the complete set of UOps allowed to exist in `tensor.lazydata` as of now in master.

`tensor_uop_spec` is divided to 3 sections:

1. [Section 1](https://github.com/tinygrad/tinygrad/pull/8311/files#diff-120c8f826d67db123ecd7a1de1d0af40b894be9c3c3a27655819b4cd2689c581R23-R37): Well-defined UOps that already align close to the natural representation. I think we can keep these as they are.

2. [Section 2](https://github.com/tinygrad/tinygrad/pull/8311/files#diff-120c8f826d67db123ecd7a1de1d0af40b894be9c3c3a27655819b4cd2689c581R41-R57): These UOps have room for refactoring to a better representation, eg. COPY args

3. [Section 3](https://github.com/tinygrad/tinygrad/pull/8311/files#diff-120c8f826d67db123ecd7a1de1d0af40b894be9c3c3a27655819b4cd2689c581R61-R74): These UOps aren't well thought through and need new specs. TIP 1 and TIP 4 propose changes related to CONST, shape/device on tensors that improve these UOps. This section can guide future TIPs as well.


Qazalin:
I think the failure on image is a real bug in tensor. Isolated in:
https://github.com/tinygrad/tinygrad/compare/master...failing_dtype_assert_in_gradient?expand=1

tinygrad/engine/schedule.py
```diff
@@ -16,6 +16,64 @@
...
+   # arg: (number, size)
+   isinstance(buf.arg, tuple) and len(buf.arg) == 2 and all(isinstance(x, int) for x in buf.arg) and \
+   # dtype
+   isinstance(buf.dtype, (DType, ImageDType))),
+  # movement ops
+  (UPat(GroupOp.Movement, name="mv", src=(UPat.var("x"),)), lambda mv,x: isinstance(mv.arg, tuple) and mv.dtype == x.dtype),
+  # tensor variable bindings
+  (UPat(Ops.BIND, dtype=dtypes.int, src=(UPat(Ops.DEFINE_VAR), UPat.cvar(dtype=dtypes.int)), arg=None), lambda: True),
+  # DETACH and CONTIGUOUS change how we interpret the source UOp
+  (UPat(Ops.DETACH, name="detach", src=(UPat.var("x"),), arg=None), lambda detach,x: detach.dtype == x.dtype),
```
geohot, line 35: UPat((Ops.DETACH, Ops.CONTIGUOUS),

tinygrad/engine/schedule.py
```diff
@@ -16,6 +16,64 @@
...
+  # Tensor const has a ShapeTracker of shape=() and fake buffer of size 1
+  (UPat(Ops.VIEW, name="view", arg=ShapeTracker.from_shape(()), src=(UPat(Ops.BUFFER, name="fake", arg=(-1, 1)),
+                                                                     UPat({Ops.CONST, Ops.BIND}, name="const_uop"))),
+   lambda view,fake,const_uop: view.dtype == fake.dtype == const_uop.dtype),
+
+  # TODO: is EMPTY required? Can it be represented with an existing spec?
+  (UPat(Ops.EMPTY, src=(), arg=None), lambda: True),
+
+  # TODO: BUFFER_VIEW is overloaded, can we break it into multiple well defined UOps?
+  (UPat(Ops.BUFFER_VIEW, src=(UPat(),)), lambda: True),
```
geohot, line 74: Can you describe what this is now?

tinygrad/engine/schedule.py
```diff
@@ -16,6 +16,73 @@
...
+   # arg (device, clone?)
+   # TODO: use Ops.DEVICE src
+   isinstance(copy.arg, tuple) and len(copy.arg) == 2 and isinstance(copy.arg[0], str) and isinstance(copy.arg[1], bool) and \
+   # dtype
+   copy.dtype == copyin.dtype),
+
+  # VIEW(BUFFER) applies a ShapeTracker on top of the underlying device buffer
+  # NOTE: VIEW size exactly matches the underlying BUFFER, tensor doesn't apply movement ops to the VIEW
+  (UPat(Ops.VIEW, name="view", src=(UPat(Ops.BUFFER, name="buf"),)),
+   lambda view,buf: view.dtype == buf.dtype and view.size == buf.size and view.st.contiguous),
```
Qazalin, line 51: Why does this VIEW exist?

Is it purely there to represent the shape of the on-device Buffer (aka. preserve STORE st), because of `tensor.shape`?


I added some asserts to restrict the ShapeTracker of this VIEW on BUFFER, maybe TIP 4 is a good place to define this
https://github.com/tinygrad/tinygrad/issues/8294.

movement ops on BUFFER look like EXPAND, RESHAPE, etc at the tensor level.
The scheduler is free to add up these sts, but the resulting VIEW(BUFFER) has a different meaning, `LOAD`.

PR #8323, opened 2024-12-18
[tests wip] VIZ tracks becomes 

Qazalin:
Manually fuzz tested on a bunch of kernels and looks like this works!

```
VIZ=1 python3 test/test_ops.py TestOps.test_tiny_add
```

with `becomes` tracking:
image

vs without it, the tensor uop gets mutated later:
image

I think we still need to early pickle the pre-become UOp though. Is there a better way to preserve the UOp pre mutation?

I wanna add tests that don't need to call tinygrad APIs (eg. create_schedule) to test this, ideally `test_post_mutate_uops_*` tests should have their own isolated `@track_rewrites` decorated graph_rewrites, making test_viz completely isolated from the rest of tinygrad.

test_viz should also have some overhead tests, run effcientnet without viz compare with with viz, assert it's a reasonable range that doesn't regress.

```
~/code/tinygrad (viz_pickle) > VIZ=1 python3 examples/efficientnet.py 
281 8.961818 tabby, tabby cat
did inference in 1980.28 ms
rewrote 100 graphs and matched 198350 times, saved to /var/folders/6s/09rjw89s6dngdw20gml2sr0c0000gn/T/rewrites.pkl
*** viz is starting
preparing kernels: 100it [00:00, 5999.61it/s]
*** started viz on http://127.0.0.1:8000
*** ready in 329.81ms
```

vs master:

```
~/code/tinygrad (master) > VIZ=1 python3 examples/efficientnet.py 
281 8.961818 tabby, tabby cat
did inference in 29834.69 ms
rewrote 100 graphs and matched 198350 times, saved to /var/folders/6s/09rjw89s6dngdw20gml2sr0c0000gn/T/rewrites.pkl
*** viz is starting
preparing kernels: 100it [00:00, 366.61it/s]
*** started viz on http://127.0.0.1:8000
*** ready in 652.31ms
```


Qazalin:
```
VIZ=1 python3 test/test_ops.py TestOps.test_tiny_add
```

with `becomes` tracking:
image


vs without it, the tensor uop gets mutated later:
image

Issue #8316, opened 2024-12-18
Docs broken for torch_load


leopf:
![image](https://github.com/user-attachments/assets/2eddb711-ffeb-4705-808d-d5b3f6ea07ee)

```python
@accept_filename
def torch_load(t:Tensor) -> Dict[str, Tensor]: ...
```

Probably broken because of `@accept_filename`. 

Should be:
`torch_load(t: Union[Tensor, str, pathlib.Path]) -> Dict[str, Tensor]`

`tar_extract` is missing from docs as well.
This will probably be broken for `gguf_load` too.

chenyuxyz:
broken as it should document that the function can accept both file name and Tensor? if so can just add a note in docstring

chenyuxyz:
yea best would be document through api, then through docstring. you can run `serve_docs.sh` to test locally (need to install doc dependepncy)

PR #8332, opened 2024-12-19
nn.state docs cleanup

leopf:
Related to #8316.

I have added a small extension to solve the decorator issue (https://mkdocstrings.github.io/griffe/guide/users/how-to/support-decorators/).
I tried other solutions, but none of them solved the problem completely.

Griffe supports dynamic analysis but I think the option is hidden in the mkdocstrings api (https://mkdocstrings.github.io/griffe/guide/users/loading/#forcing-dynamic-analysis).

I really dont like having an extension for just this one thing. I will look into this more, maybe there is something more maintainable.

---

- added `tar_extract`
- corrected `safe_load_metadata` (do we want this in docs?)
- `gguf_load` modernization
- minor stuff

https://github.com/user-attachments/assets/683a4c86-d39a-4aa7-a1f4-5c31ae6265c8


docs/mkdocs_exts.py
```diff
@@ -0,0 +1,8 @@
+import griffe
+
+class StateAcceptFilenameDecorator(griffe.Extension):
+  def on_function_instance(self, *, func: griffe.Function, **_) -> None:
+    if any(d.callable_path == "tinygrad.nn.state.accept_filename" for d in func.decorators):
```
geohot, line 5: Way too specific I think. Nobody will remember to update this

leopf:
Sorry for not getting to this earlier.

- No extension, a more general one would have been too complex.
- Signatures added as md code blocks, automatic signatures are disabled in `nn.md`.

Downside: text color is slightly different, types dont have hyperlinks.

image

tinygrad/nn/state.py
```diff
@@ -290,16 +298,16 @@ def q_to_uint8(t: Tensor, b: int) -> Tensor:
       return d * (xl.bitwise_or(xh).bitcast(dtypes.int8) - 32).flatten(-2) * scales
   raise ValueError(f"GGML type '{ggml_type}' is not supported!")
 
-@accept_filename
```
geohot, line 293: Why is this removed?

leopf: Providing a `Path` or `str` (disk tensor) does not work as the tensor device must support ALU ops for dequantization (passing a `str` or `Path` will fail right now "RuntimeError: CAST isn't supported on DISK").

See: `NOTE: The provided tensor must be on a device that supports execution.`

Issue #8275, opened 2024-12-16
TIP 1 - CONST in big graph


geohot:
## Abstract

CONST is a well understood UOp, having a dtype and an arg of the value. However, in Tensor, a device and shape are also required, and there's not a straightforward way to add them.

## Motivation

Find a Tensor CONST representation that's amenable to CONST folding using the existing folding rules. This will require

`ADD(CONST(???), CONST(???))`

## Current State

Consts are represented as `VIEW(BUFFER(-1), CONST)`, getting the device from BUFFER and shape from VIEW. An ADD would be `ADD(VIEW(BUFFER(-1), CONST), VIEW(BUFFER(-1), CONST)` and doesn't easily fold.

## Proposal

Firstly, BUFFER(-1) isn't a BUFFER, and should be replaced with a DEVICE(arg=<NAME>) Op.

Then `CONST(VIEW(DEVICE()))` can be a CONST with a Device and shape. This modifies the CONST Op to allow a src. This src is ignored, and CONST always has the value of CONST.

This can't express a masked VIEW, but that's okay. Just use `VIEW(CONST(DEVICE())))`, this is how they are represented in Tensor. The VIEW and CONST will be swapped in rewrite rules.

## Result

`ADD(CONST(VIEW(DEVICE())), CONST(VIEW(DEVICE())))` will fold as expected to `CONST(VIEW(DEVICE()))`

## Alternatives

Potentially, we could just use `CONST` and assume the shape and device come from elsewhere. But this has several issues.

1) What is the shape and device of the Tensor? All Tensors must have shape and device.
2) These UOps won't have ShapeTrackers, so `.st` will return None.
3) If all become CONST, the output won't have a shape.

This seems a lot harder to reason about as things are not local. A key principle of UOps is that you want things to have their full meaning locally. The `CONST(VIEW(DEVICE()))` has very little downside too, all CONST srcs can be rewritten to nothing once we are in a kernel.

## Work Items

- [x] Replace BUFFER(-1) with DEVICE
- [x] Replace `VIEW(DEVICE, CONST)` with `VIEW(CONST(DEVICE()))` in Tensor const creation.
- [x] Add rewrite rule to rewrite `VIEW(CONST(DEVICE()))` to `CONST(VIEW(DEVICE()))` if the view is unmasked
- [x] Add in normal constant folding, observe that it works.

Qazalin:
1. does CONST look like:
```
[...]

PR #8333, opened 2024-12-19
remove the fake buffer 3

Qazalin:
The scheduler still relies on the VIEW(BUFFER, CONST) pattern to match CONST UOps. Or it hardcodes Ops.CONST directly.

This diff should refactor the scheduler such that all the asserts in TestConst pass while on the VIEW(BUFFER, CONST) pattern is gone.

Once merged, it completes work item 1 of "CONST in big graph" https://github.com/tinygrad/tinygrad/issues/8275.


geohot:
I don't understand the point of this. I still can't use the normal logic for CONST folding if I change to this.

PR #7567, opened 2024-11-06
failing test for View.__add__ RecursionError

Qazalin:
I hit this while writing fusion UPats in big graph #7556.
No asserts pre symbolic deletion `518c022c29104d79c7a50ec41af5b7e6404da317g`.

PR #8336, opened 2024-12-19
failing test for View.__add__ RecursionError 2

Qazalin:
similar to #7567,
I hit this while trying early merge folding of tensor const, If I'm doing something wrong, can `__add__` give a more comprehensible assert?


Qazalin:
resolved #8337.

PR #8337, opened 2024-12-19
temp fix for symbolic shape view add [pr]

chenyuxyz:
something is still wrong with symbolic shape shrink, but it should not recurse forever


Qazalin:
ah thanks xd, I'll rebase to see the new error
is it a wrong add?

chenyuxyz:
not sure what's considered a wrong add, but add should give either a merged view or None and never an error. feel free to add more tests and tag me if you see more

chenyuxyz:
@Qazalin you might want to check how you got this View in the first place. It broke because vm1 has contiguous=False, even though it's contiguous, and also the mask is no-op. Creating the View with `View.create` when possible in the tests, which would canonicalize these.

i still think it's good to not have these function fail randomly so will keep this fix and the test

Qazalin:
hmm - I'm 90% sure I just used ShapeTracker.from_shape() and `__add__` of (some_st)+(some_st)
The scheduler doesn't ever construct Views manually. I'll trace back to how that View got created.

PR #8338, opened 2024-12-19
Openpilot compile: fix for openpilot use

haraschax:


examples/openpilot/compile3.py
```diff
@@ -27,7 +27,8 @@ def compile(onnx_file):
 
   input_shapes = {inp.name:tuple(x.dim_value for x in inp.type.tensor_type.shape.dim) for inp in onnx_model.graph.input}
   input_types = {inp.name: tensor_dtype_to_np_dtype(inp.type.tensor_type.elem_type) for inp in onnx_model.graph.input}
-  if getenv("FLOAT16", 0) == 0: input_types = {k:(np.float32 if v==np.float16 else v) for k,v in input_types.items()}
+  # Float inputs are always float32
```
chenyuxyz, line 30: FLOAT16 was hardcoded default to 1 in line 4 and there are other mentions of `FLOAT16` in this file, so not sure about this change

haraschax: After some more checks I think this change is correct. We want the openpilot tinyjits to get float32 floats in and out, regardless if the tinyjit operates on float16 internally.

The only other mentions of FLOAT16 in this file disables the onnx ref test, which is correct. Since there is small diff with FLOAT16=1.

PR #8349, opened 2024-12-20
replace tiktoken

leopf:
Tested for gpt2 and llama3.

The main issue is, that a lot of the tokenizer regex patterns (like the gpt2 one) are not `re` compatible and rely on the third party `regex`. 
So I am not sure this improves anything in terms of experience. It is definitely slower (there is room for improvement).

Happy to do some more cleanup, if you want this regardless.

### DIFF
--- a/examples/gpt2.py
+++ b/examples/gpt2.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import os, argparse
 from typing import Optional, Union
-import tiktoken
+import extra.bpetokenizer as tiktoken
 from tinygrad import Tensor, TinyJit, Device, GlobalCounters, Variable, dtypes
 from tinygrad.ops import UOp
 from tinygrad.helpers import Timing, DEBUG, JIT, getenv, fetch, colored, trange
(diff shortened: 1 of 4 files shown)


geohot:
Why are there regexes?

leopf:
The regexes that are incompatible with `re` are the ones defined by the models.
Like this one defined for gpt2:
`r"""'(?:[sdmt]|ll|ve|re)| ?\p{L}++| ?\p{N}++| ?[^\s\p{L}\p{N}]++|\s++$|\s+(?!\S)|\s"""`

Problem is: `\p{L}` (all unicode letters) and `\p{N}` (all unicode numbers).

The regexes are required to split the text into chunks that will then be tokenized one by one. This is chosen before tokenizer training and AFAICT unavoidable during inference. See the [tiktoken implementation](https://github.com/openai/tiktoken/blob/main/src/lib.rs#L242) (`regex.find_iter` is splitting into chunks).

There are some possible workarounds. Ill see if there is a nicer solution for this.

leopf:
Found a way to do this without regex. I do not notice any performance problems, but startup is slower. 

This is not ready yet. A lot of cleanup to be done. I also want to generalize this more.

geohot:
Looks stale, feel free to reopen when it's ready

PR #8350, opened 2024-12-20
CAST_BEFORE_VIEW in the scheduler

Qazalin:
we need to move "make things that can't be images not images" to graph rewrite and run after all the movement ops are applied.
This diff is what it takes to do CAST_BEFORE_VIEW=1 in the scheduler with the current stuff.

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -363,19 +363,6 @@ def replace_contiguous(ctx:ScheduleContext, alu:UOp):
     if (replace_src:=ctx.contiguous.get(s, None)) is not None: new_src[i] = replace_src
   if tuple(new_src) != alu.src: return alu.replace(src=tuple(new_src))
 
-def cast_before_view(ctx:ScheduleContext, b:UOp, base:UOp, root:UOp, src:UOp, view:UOp):
-  if root.dtype.itemsize > src.dtype.itemsize: return None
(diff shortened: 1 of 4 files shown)


tinygrad/engine/schedule.py
```diff
@@ -505,7 +532,8 @@ def create_schedule_with_vars(outs:list[UOp]) -> tuple[list[ScheduleItem], dict[
   cache: dict[UOp, UOp] = {}
   # to_uop is removing (many) of the movement ops
   for u in (big_graph:=UOp.sink(*(to_uop(x, ctx, cache) for x in outs))).src: ctx.realizes[u.buf_uop] = u
-  big_graph = graph_rewrite(big_graph, remove_movement_ops+ops_folding+do_realize, ctx)
+  big_graph = graph_rewrite(big_graph, rewrite_views, ctx)
```
Qazalin, line 535: The spec will go over the reasoning behind this, but the basic idea is that when we have:
```
CAST(EXPAND(src))
```
in dfs order:
1. src gets added to realizes
2. CAST(EXPAND(src)) -> EXPAND(CAST(src))

how do we judge when not to realize src?
What if src has two children, one is an expand and one is COPY(src).

Qazalin:
Agree this adds a lot of complexity, there might be a more natural way to do this.

PR #8353, opened 2024-12-20
remove some VIEWs we don't need [pr]

geohot:


geohot:
Nice to see this passing! Do we want it?

PR #8364, opened 2024-12-21
Add PReLu op to Tensor

geohotstan:
I kept this PR private for awhile cuz I found it hacky...
The `channel_dim` was created just to hack compatibility for pytorch and onnx tests.
pytorch expects the `channel_dim == 1` case with unconventional broadcasting while onnx expects conventional broadcasting.

onnx tests however includes pytorch prelu tests through their `OnnxBackendPyTorchConvertedModelTest` so I can't do `channel_dim == 1`. This is why the `channel_dim=None` option is there
and onnx follows strictly conventional broadcasting even when `weight` is a non-scalar 1-D Tensor

🫠

The documentation is there to explain the behavior though.


tinygrad/tensor.py
```diff
@@ -2753,6 +2753,31 @@ def celu(self, alpha=1.0):
     """
     return self.maximum(0) + (alpha * ((self / alpha).exp() - 1)).minimum(0)
 
+  def prelu(self, weight:Tensor, channel_dim:int|None=None):
```
chenyuxyz, line 2756: don't quite understand the constraint in onnx spec and tests, but i'd prefer this api to match torch without indirection, then somehow call this in onnx prelu (we have other functions that works that way because api is close but not identical anyway). let me know what you think

extra/onnx_ops.py
```diff
@@ -47,10 +47,7 @@ def Constant(sparse_value:Tensor|None=None, value:Tensor|None=None, value_float:
 
 def HardSigmoid(x:Tensor, alpha:float=0.2, beta:float=0.5): return (alpha*x + beta).clip(0, 1)
 def Gelu(x:Tensor, approximate:str|None=None): return x.gelu() if approximate == "tanh" else 0.5 * x * (1 + (x/math.sqrt(2)).erf())
-# TODO: fix this
-def PRelu(X:Tensor, slope:Tensor):
-  slope = slope[0] if slope.shape[-1] != X.shape[-1] else slope # HACK OnnxBackendPyTorchConvertedModelTest HAS WEIRD SLOPE WHERE IT'S [0.25, 0.25, 0.25] FOR ANY X.SHAPE
-  return (X > 0).where(X, X * slope)
+def PRelu(X:Tensor, slope:Tensor): return (X > 0).where(X, X * slope)
```
geohot, line 50: Why add prelu to Tensor if it can't be used here?

geohotstan: I'm still debating on whether or not to add this to tensor...
My attempt to add it with shared behavior was way too hacky.

geohotstan:
Problem is this:
There are 2 set of tests onnx runs, `OnnxBackendNodeModelTest` and `OnnxBackendPyTorchConvertedModelTest`

the test for these two that covers prelu expects two different behaviors...
`OnnxBackendPyTorchConvertedModelTest` matches torch's behavior (which is the one in tensor.py currently)
`OnnxBackendNodeModelTest` is the current one in onnx_ops.py

There is no good way of reconciling these two.
Usually pytorch converted models have `onnx_model.producer_name` set as `"pytorch"` but `OnnxBackendPyTorchConvertedModelTest` doesn't even have that, so I can't even inject a special `opt` for the test.

I haven't found an onnx model that uses `prelu` yet.
I think we keep this op in limbo until I find an onnx model that uses prelu and see which behavior it expects.

geohotstan:
image
ort fails these lmfao

PR #8368, opened 2024-12-21
start on the big_graph_spec

Qazalin:
two things blocking this:
- [x] ASSIGN setitem must use swizzle
- [x] image


Qazalin:
some good ideas here but kinda meh, there's nothing fundamentally wrong with merging the VIEWs, there's [a better way to write this spec](https://github.com/tinygrad/tinygrad/pull/8380/files#diff-120c8f826d67db123ecd7a1de1d0af40b894be9c3c3a27655819b4cd2689c581R604-R612).

PR #8380, opened 2024-12-22
give BUFFER a contiguous ShapeTracker shape=(N,) and stride=(1,)

Qazalin:
BUFFER is a contiguous piece of memory, it must have a ShapeTracker that is its size.
This will unblock #8412 and make merge_views in the scheduler always correct.


Qazalin:
back to this now that const folding is done. I wanna also remove the concept of VIEW with two srcs after this.

Qazalin:
moved to 8495.

Qazalin:
finally merged in 8811.

PR #8383, opened 2024-12-23
Update __getitem__ docs and misc tests clean ups

geohotstan:


tinygrad/tensor.py
```diff
@@ -1033,16 +1033,17 @@ def shrink(self, arg:tuple[Optional[tuple[sint, sint]], ...]) -> Tensor:
...
+        - `padding` length must be even.
 
-    2. Group padding: (..., (padding_top, padding_bottom), (padding_left, padding_right))
-       - This structure matches pad for jax, numpy, tensorflow and others.
-       - For each axis, padding can be `None`, meaning no padding, or a tuple `(start, end)`.
-       - `padding` must have the same length as `self.ndim`.
+    2. Group padding: `(..., (padding_top, padding_bottom), (padding_left, padding_right))`
+        - This structure matches pad for JAX, NumPy, TensorFlow, and others.
+        - For each axis, padding can be `None`, meaning no padding, or a tuple `(start, end)`.
+        - `padding` must have the same length as `self.ndim`.
```
geohotstan, line 1046: This wasn't rendering properly for mkdocs (was enumerating to 7...)
need 4 spaces for sub bullet points
need to remind myself this is markdown

also btw there are some idiosyncrasies to markdown I don't think the other documentation are aware of
e.g. the + at line 1036 breaks the line into two. Two adjacent lines without two whitespaces at end of first line will not line break

image
was this in tensor.py

image

tinygrad/tensor.py
```diff
@@ -1210,6 +1182,43 @@ def _getitem(self, indices, v: Optional[Tensor] = None) -> Tensor:
...
+      ```python exec="true" source="above" session="tensor" result="python"
+      print(t[:, None].shape)
+      ```
+
+    NOTE: Out-of-bounds indexing results in a value of `0`.
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t[Tensor([4, 3, 2])].numpy())
+    ```
+    """
```
geohotstan, line 1221: should we have it like this?
image
image

chenyuxyz:
yay doc is great

PR #8387, opened 2024-12-23
Fix FC layer ResNet load_from_pretrained error

francislata:
# Overview
In this PR, I am fixing an issue that I noticed when the fully connected layer is set to `None` and `load_from_pretrained` is called.

In some cases, like the `RetinaNet` MLPerf training script, there's no need for this layer and can be done by passing in `num_classes = None` as implemented [here](https://github.com/tinygrad/tinygrad/blob/c587b3b08c6471b2a3daadc1554b7593ed58936e/extra/models/resnet.py#L93).

# Demo
To show this issue, here's what happens in `master`:
```python
>>> model = ResNeXt50_32X4D(num_classes=None)
>>> model.load_from_pretrained()
  File "/home/paperspace/tinygrad/extra/models/resnet.py", line 140, in load_from_pretrained
    obj: Tensor = get_child(self, k)
                  ^^^^^^^^^^^^^^^^^^
  File "/home/paperspace/tinygrad/tinygrad/helpers.py", line 66, in get_child
    else: obj = getattr(obj, k)
                ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'weight'
```

With this PR, it fixes the issue:
```python
>>> from extra.models.resnet import ResNeXt50_32X4D
>>> model = ResNeXt50_32X4D(num_classes=None)
>>> model.load_from_pretrained()
>>> 
```


chenyuxyz:
can you update https://github.com/tinygrad/tinygrad/blob/master/test/models/test_resnet.py too?

not sure if it's in CI, but passing locally is good enough. changes like this need some comments about context and some tests / script for future maintainability

francislata:
added another test case for loading the model without an FC layer.

here's the result **without the fix**:
Screenshot 2024-12-26 at 16 11 12

here's the result **with the fix**:
Screenshot 2024-12-26 at 16 11 25

extra/models/resnet.py
```diff
@@ -137,7 +137,11 @@ def load_from_pretrained(self):
 
     self.url = model_urls[(self.num, self.groups, self.base_width)]
     for k, dat in torch_load(fetch(self.url)).items():
-      obj: Tensor = get_child(self, k)
+      try:
+        obj: Tensor = get_child(self, k)
+      except AttributeError:
+        if 'fc.' in k and self.fc is None:
+          continue
```
chenyuxyz, line 144: else re-raise the issue

PR #8391, opened 2024-12-24
a few more UPat.var -> UPat.cvar in the scheduler [pr]

Qazalin:
`simplify_reduceop` stays even after `symbolic+ops_folding` since it's folding shape changing ops.
`const_arg` is simple, self is either GroupOp.Movement(CONST) or CONST(). No more VIEW with len(src) == 2 having a different meaning.


Qazalin:
This mixtral regression was weird
image

i think it's because base isn't cached.

Qazalin:
it was .base, we're back.
image

It can't be cached because of uop mutablity, we need to be very careful with accessing base before the delete_movementops pass.

PR #8393, opened 2024-12-24
Add variations for some ADD patterns

S-Lykles:

PR #8394, opened 2024-12-24
New ordering

S-Lykles:
This creates a different ordering for uops, specifically all binary ops by a constant get sorted on the non constant part first.  So `x+1`, `x*5` etc get sorted next to `x` before `y`. This also moves all the binary ops before the irreducible ops so that in an op chain the op is always on the left. When you do `a+b+c` in python this becomes `(a+b)+c` previously this would then get flipped to `c+(a+b)` because the `ADD` op is above `DEFINE_LOCAL`, `DEFINE_ACC` etc. Moving `DEFINE_ACC` to after math ops causes some problems on metal, see #8156. To fix that I added another pass inbetween symbolic rewriting and late rewriting which also makes all the assembly backend accumulate more efficiently.


S-Lykles:
This also allows some add patterns like in #8393 to be written more efficiently since you always know on which side the second add will be.

PR #8401, opened 2024-12-25
linearize input UOps while doing type_verify [pr]

Qazalin:
This diff allows for running type_verify without having to first linearize/toposort UOps.
I don't think a gather all `list(UOp.sink(*outs).toposort)` can scale to large graphs. Currently this is causing the bert OOM.

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -586,7 +586,7 @@ remove_movement_ops = PatternMatcher([
 
 @track_rewrites(named=True)
 def create_schedule_with_vars(outs:list[UOp], skip_check:bool=not __debug__) -> tuple[list[ScheduleItem], dict[Variable, int]]:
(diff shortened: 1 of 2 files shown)


Qazalin:
so CONST(vec, 0.0) is definitely supported, it's explicitly created in WMMA:
```
wmma = UOp(Ops.WMMA, dtype=tc.dtype_out.vec(wmma_sz[2]), src=(
  UOp(Ops.CONTRACT, dtype=srcs[0].dtype.vec(wmma_sz[0]), src=(srcs[0],), arg=upcast_axes[0]),
  UOp(Ops.CONTRACT, dtype=srcs[1].dtype.vec(wmma_sz[1]), src=(srcs[1],), arg=upcast_axes[1]),
  UOp.const(tc.dtype_out.vec(wmma_sz[2]), 0.0)), arg=wmma_arg)
```
that type spec just isn't complete.

PR #8404, opened 2024-12-25
Move define_acc down an unrolled add chain

S-Lykles:
Currently cstyle has a quirk where the uopgraph that represents
`acc=acc+(val0*val1+val2*val3)` gets rendered as `acc=acc+val0*val1+val2*val3`. For floats this does <ins>not</ins> represent the same computation. Crucially, this allows for more efficient accumulation using mulacc which has higher precision than a seperate mul and an add. If the uopgraph is rendered without changing the graph a random metal test fails because of this slight loss of precision. #8156
This also makes assembly style backends more efficient because they can now use an extra mulacc as well. Have to think about how to test this


S-Lykles:
This graph, which gets rendered as `acc0 = (acc0+(val0.x*val1.x)+(val0.y*val1.y)+(val0.z*val1.z)+(val0.w*val1.w));`
image

Now becomes:

image

 Which actually represents what it gets rendered as

S-Lykles:
On PTX process replay:
```
  	selp.b32	%cast_f32_0, 0f3F800000, 0f00000000, %alu_pred_10;
 	and.pred	%alu_pred_11, %alu_pred_9, %alu_pred_5;
 	selp.b32	%cast_f32_1, 0f3F800000, 0f00000000, %alu_pred_11;
-	mul.f32		%alu_f32_0, %val_f32_1, %cast_f32_1;
 	and.pred	%alu_pred_12, %alu_pred_9, %alu_pred_6;
 	selp.b32	%cast_f32_2, 0f3F800000, 0f00000000, %alu_pred_12;
 	and.pred	%alu_pred_13, %alu_pred_9, %alu_pred_7;
 	selp.b32	%cast_f32_3, 0f3F800000, 0f00000000, %alu_pred_13;
-	fma.rn.f32	%alu_f32_1, %val_f32_0, %cast_f32_0, %alu_f32_0;
+	fma.rn.f32	%alu_f32_0, %val_f32_0, %cast_f32_0, %acc_f32_0;
+	fma.rn.f32	%alu_f32_1, %val_f32_1, %cast_f32_1, %alu_f32_0;
 	fma.rn.f32	%alu_f32_2, %val_f32_2, %cast_f32_2, %alu_f32_1;
 	fma.rn.f32	%alu_f32_3, %val_f32_3, %cast_f32_3, %alu_f32_2;
-	add.f32		%alu_f32_4, %acc_f32_0, %alu_f32_3;
-	mov.b32		%acc_f32_0, %alu_f32_4;
+	mov.b32		%acc_f32_0, %alu_f32_3;
 	add.s32		%ridx_s32_0, %ridx_s32_0, 1;
 	setp.lt.s32	%pred_pred_0, %ridx_s32_0, %const_s32_1
```

S-Lykles:
Ultimately this is needed for the reordering+flattening bounty because for that I need to change the order of commutative ops which shouldn't change anything since floats are commuative but it breaks this optimization that cstyle does

tinygrad/codegen/uopgraph.py
```diff
@@ -239,6 +239,9 @@ def reduce_collapse(acc:UOp, ret:UOp, alu:UOp):
 arange_augrng = UPat.any(rng_aug, rng_aug+UPat.var("idx2"), rng_aug+UPat.var("idx2")+UPat.var("idx3"), UPat(Ops.VECTORIZE, name="vec", src=rng_aug))
 arange_m = ((arange_augrng<UPat.cvar("compval"))!=UPat(Ops.CONST, name="ne", arg=True)).where(UPat.cvar("multconst"), UPat.const(None, 0))
 
+# this moves the accumulation variable down an unrolled add chain which allows for more efficient accumulation using mulacc
+mulacc_unrolled = PatternMatcher([(UPat.var("x")+UPat.var("y")+acc_pat, lambda x,y,acc: (acc+x)+y if y.op is not Ops.DEFINE_ACC else None)])
```
chenyuxyz, line 243: does it need its own pass?

test/test_uops.py
```diff
@@ -365,6 +366,17 @@ def test_bitshift_right(self):
     self.assertIn(Ops.SHR, ops)
     self.assertIn(Ops.IDIV, ops)
 
+  def test_mulacc_unrolled(self):
+    # test that     acc = acc + a0*b0 a1*b1 + a2*b2 + a3*b3
```
chenyuxyz, line 370: missed a `+`

tinygrad/codegen/uopgraph.py
S-Lykles: I can't add it to sym since it will break loop/index/reduce_collapse, I tried adding it to `late_reduce` but since that will rewrite from the sources some the mul/adds will have been changed to mulacc and the `DEFINE_ACC` can't be moved down the chain anymore.

S-Lykles: I can add it to the devectorize + load_store_indexing pass but that might be less efficient

chenyuxyz:
i think this is good (mostly speed no-op, 1~2% faster on red sometimes), one question https://github.com/tinygrad/tinygrad/actions/runs/12497139860/job/34869504372?pr=8404#step:18:752 will be fixed later with add chain sorting right?

image

also can you add an unit test that directly tests mulacc_unrolled on several different ordered add chain of DEFINE_ACC and other ALU / CONST, so it's straightforward to know what the sorted result is

S-Lykles:
Mmh do you mean wether `acc6 = (val26+acc6+val24)` will become  `acc6 = (acc6+val24+val26)` again?
The answer is no because this is pattern is happenign after all of symbolic, but `acc6 = (val26+acc6+val24)`  is `acc6 = ((val26+acc6)+val24)`  which is the same as `acc6 = ((acc6+val26)+val24)` since floats and ints are commutative

chenyuxyz:
I see, that's fine. I would still like to see the tests that reflect the behavior. I think this part is going to change a lot in the future and having more examples to show desired behavior makes it easier to maintain.

@geohot any feedback on this one?

chenyuxyz:
discussed offline - okay to merge this with more tests for the orders

PR #8409, opened 2024-12-25
New tuplize and commutative flipping

S-Lykles:
Sorts binary-op-by-a-const, on non-constant part first.

`b+a*3` now becomes `a*3+b` and `c+(a+b)` now gets rendered as `(a+b)+c`

This will fail one metal test without #8404

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -269,7 +269,9 @@ class UOp(MathTrait, metaclass=UOpMetaClass):
     return _toposort(self, cache={})
 
   @functools.cached_property
-  def tuplize(self:UOp) -> tuple[int, Any, Optional[DType], tuple]: return (self.op.value, self.arg, self.dtype, tuple(x.tuplize for x in self.src))
+  def tuplize(self:UOp) -> tuple:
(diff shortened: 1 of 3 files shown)


tinygrad/ops.py
```diff
@@ -269,7 +269,9 @@ def _toposort(u:UOp, cache:dict[UOp, dict[UOp, None]]):
     return _toposort(self, cache={})
 
   @functools.cached_property
-  def tuplize(self:UOp) -> tuple[int, Any, Optional[DType], tuple]: return (self.op.value, self.arg, self.dtype, tuple(x.tuplize for x in self.src))
+  def tuplize(self:UOp) -> tuple:
+    if self.op in GroupOp.Binary and self.src[1].op in (Ops.CONST, Ops.VCONST): return self.src[0].tuplize + (self.op.value, self.src[1].arg)
```
geohot, line 273: Why do we need a special case here?

S-Lykles: This is part of #8123, by itself it doesn't achieve that much but one you start ordering a chain you want x and x*const to be sorted next to each other so that it can get folded to x*(const+1)

S-Lykles: Also should sort x//c*c and x%c very close to each other

S-Lykles:
will be part of #8123 not a seperate pr anymore

PR #8412, opened 2024-12-25
merge all one op views [pr]

geohot:


Qazalin:
some permute here is making the last AST wrong:
Screenshot 2024-12-26 at 19 03 19

```
IMAGE=2 GPU=1 python3 test/test_ops.py TestOps.test_small_gemm_padded
```

geohot:
@Qazalin do we want this?

Qazalin:
I want base on buffer to just work but that's gated on mutability to go, we can't have two instances of a BUFFER.

Issue #7889, opened 2024-11-25
Make TYPED=1 work


geohot:
TYPED=1 will do runtime type checking. Make tinygrad run normally with that option

AndreCorreaSantos:
Sorry for opening this PR too early—it should’ve been a draft since the tests are still failing.

Should we import numpy globally in tensor.py to resolve the type annotation issues Typeguard is causing?

Typeguard seems to be having trouble with the walrus operator. Are these necessary, or can we consider refactoring them?

kayo09:
I couldn't find any documentation for TYPED variable (https://docs.tinygrad.org/env_vars/). Would you mind giving some pointers as to what it does

kayo09:
IDK what is wanted from developers, can some insight be provided before closing a PR request just so there is some direction to work in... Thank you. @chenyuxyz

AndreCorreaSantos:
> IDK what is wanted from developers, can some insight be provided before closing a PR just so there is some direction to work towards, I have STFW and RTFM... Thank you. @chenyuxyz

If I understood it correctly, the goal is to make typeguard run by default and have tinygrad work with typeguard enabled. I tried it but I am yet to make typeguard work with := operators and some numpy imports on tensor.py, if you want check the pr I opened previously.

kayo09:
> > IDK what is wanted from developers, can some insight be provided before closing a PR just so there is some direction to work towards, I have STFW and RTFM... Thank you. @chenyuxyz
> 
> 
> 
> If I understood it correctly, the goal is to make typeguard run by default and have tinygrad work with typeguard enabled. I tried it but I am yet to make typeguard work with := operators and some numpy imports on tensor.py, if you want check the pr I opened previously.
> 
> 
> 
> please correct me if I misunderstood the issue.

Typeguard is imported when program is run TYPED=1, I had no issues with walrus and for numpy I simply had to 
'''If TYPE_CHECKING: 
          import numpy as np''' in tensor.py. Along with a few other minor changes, I had typed=1 do as expected in my previous pr that I referenced (now deleted). I also emailed @chenyuxyz but no avail. Ig I have to leave working on this repo and move on something that supports work from new developers.

sipmine:
I understand correctly that the point of this Issue is that if a tensor / matrix is specified by a certain type, then strict typing must be performed during the implementation process?

avinabneogy23:
Can you please assign this issue to me

rt-op:
[...]

PR #8417, opened 2024-12-26
Fix TYPED=1 by moving ConstLike to appropriate place

powerLEO101:
Hi, 

by moving ConstLike to appropriate place (right after UOp), TYPED=1 is able to run 

Best

Leo


powerLEO101:
#7889

chenyuxyz:
that's not what was asked

powerLEO101:
I am  sorry I really want to do a good first issue. Would you provide an example of what is expected? I thought by having install import hook in __init__ the entire import is type guarded. Thank you very much.

chenyuxyz:
that issue was asking to make tinygrad run with run time type checking

PR #8419, opened 2024-12-26
add symbolic_simple to the scheduler [pr]

Qazalin:
symbolic_simple replaces all the existing elementwise ops folding rules in the scheduler.

It also allows for collapsing `float32 -> CAST(imagef) -> STORE` to the same Buffer.
The new behavior requires `.contiguous()` if the user actually wants to create the image Buffer right after cast:

`float32 -> CAST(imagef) -> STORE` does not allocate memory.
`float32 -> CAST(imagef) -> CONTIGUOUS -> STORE`, allocates a new Buffer.

The master implementation for `image_conv2d` [does this internally](https://github.com/tinygrad/tinygrad/blob/6bb54eb53238b70d0db219fdb4e94af35f7e87b6/tinygrad/tensor.py#L3818-L3820).

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -956,11 +956,11 @@ spec = PatternMatcher([
 
   # most ALUs have all matching dtypes, except CMPLT, CMPNE, and WHERE
   (UPat(Ops.WHERE, name="w", src=(UPat(dtype=dtypes.bool), UPat(name="x"), UPat(name="y"))), lambda w,x,y: w.dtype == x.dtype == y.dtype),
-  (UPat((Ops.CMPLT, Ops.CMPNE), dtype=dtypes.bool, src=(UPat(name="x"), UPat(name="y"))), lambda x,y: x.dtype == y.dtype),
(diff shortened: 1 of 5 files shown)


geohot:
Did you see what the new `29 - 13` gated read image are?

tinygrad/ops.py
```diff
@@ -956,11 +956,11 @@ def graph_rewrite(sink:UOp, pm:PatternMatcher, ctx=None, bottom_up=False) -> UOp
 
   # most ALUs have all matching dtypes, except CMPLT, CMPNE, and WHERE
   (UPat(Ops.WHERE, name="w", src=(UPat(dtype=dtypes.bool), UPat(name="x"), UPat(name="y"))), lambda w,x,y: w.dtype == x.dtype == y.dtype),
-  (UPat((Ops.CMPLT, Ops.CMPNE), dtype=dtypes.bool, src=(UPat(name="x"), UPat(name="y"))), lambda x,y: x.dtype == y.dtype),
+  (UPat((Ops.CMPLT, Ops.CMPNE), dtype=dtypes.bool, src=(UPat(name="x"), UPat(name="y"))), lambda x,y: x.dtype.base == y.dtype.base),
```
geohot, line 959: When is the dtype of the base different from the non base?

Qazalin: Did it because of this #8420 - it only happens on images if the buffer is realized. I can make a helper that restricts this.

PR #8420, opened 2024-12-26
image buffer realization spec [pr]

Qazalin:
As I understand it, "make things that can't be images not images", originally added in #4467, is supposed to override a Buffer's dtype if the ShapeTracker strides do not allow ImageDType ptr arithmetic.

This means calling `.realize` on a Tensor can change its dtype. However, if that Tensor has children, they [stay Image](https://github.com/tinygrad/tinygrad/compare/image_spec?expand=1#diff-01b00890075e5d455c9477e0722d6fc43a51b557f3007b9c57660fd4656a1465R151).

This blocked scheduler symbolic folding because tensor_uops_spec expected the source dtypes to match. I have changed the spec to [check the base dtypes](https://github.com/tinygrad/tinygrad/pull/8419/files#diff-5736814df0b2c07a42d5c1bfef6b5aa5635ce18043f4aa030e21768694259ca3R959).

but this hack can't stay, it also blocks CAST_BEFORE_VIEW=1 moving to the scheduler, is there a better way to deal with this on UOp / Tensor init?


geohot:
I agree this hack is in the wrong place. This should either be a rewrite rule or handled in the device. We should start by documenting the current behavior in relevant cases, than thinking how else we can implement it (if we even need it)

geohot:
This still something we want?

Qazalin:
focusing on tensor_map. I'll get that merged and work on this spec.
I think I just need to add the test case when something goes from image -> can't be image -> back to image.

Also, we discovered there's a swizzle issue with ImageDType STORE yesterday, https://discord.com/channels/1068976834382925865/1230434680000741377/1328808098504114380

This needs to be tested, it was never revealed before we removed CAST_BEFORE_VIEW=1

Issue #3118, opened 2024-01-14
Ubuntu 22.04 cannot load model for conversation.py - _pickle.UnpicklingError: invalid load key, '<'.


helperfunc:
```
$ python3 examples/conversation.py --llama_gen 2 --llama_model weights/LLaMA-2/7B
[nltk_data] Downloading package punkt to /home/hui/nltk_data...
[nltk_data]   Package punkt is already up-to-date!
ram used:  0.11 GB, decoder.mask                                      : 100%|████████████████████████████████████████████████████████████████████████████████████████████| 168/168 [00:00<00:00, 453.90it/s]
loaded weights in 371.60 ms, 0.11 GB loaded at 0.29 GB/s
Traceback (most recent call last):
  File "/home/hui/tinygrad/examples/conversation.py", line 262, in <module>
    synth, emotion_embedding, text_mapper, hps, model_has_multiple_speakers = init_vits(args.vits_model_to_use, args.vits_emotion_path, args.vits_speaker_id, args.vits_seed)
  File "/home/hui/tinygrad/examples/conversation.py", line 167, in init_vits
    net_g = load_model(text_mapper.symbols, hps, model_config)
  File "/home/hui/tinygrad/examples/vits.py", line 535, in load_model
    _ = load_checkpoint(fetch(model[1]), net_g, None)
  File "/home/hui/tinygrad/examples/vits.py", line 540, in load_checkpoint
    checkpoint_dict = torch_load(checkpoint_path)
  File "/home/hui/tinygrad/tinygrad/nn/state.py", line 143, in torch_load
    _, _, _, rwd, _, ids, base_offset = pkl.load(), pkl.load(), pkl.load(), f.tell(), pkl.load(), pkl.load(), f.tell()
_pickle.UnpicklingError: invalid load key, '<'.
```

hristog:
Hi @helperfunc,

I'm not part of the maintainers team, but it seems to me that this issue could be supplemented with a bit more information, in order to be reproducible and debuggable.

Could you, please, perhaps provide what the output of `pip freeze` (or `python3 -m pip freeze`) and `python3 -VV` is, in order to give a better idea about your environment?

Based on a quick Google search, it seems to me that the corresponding model hasn't been downloaded in full. Perhaps you could try to verify the integrity of the downloaded model as well.

Additionally, you might find the suggestion for installing `git-lfs` from [this](https://stackoverflow.com/a/67141340) StackOverflow answer useful.

Good luck!

[...]

PR #8425, opened 2024-12-26
Fix examples/conversation.py 

mulac:
Fixes 2 issues when trying to run conversation example for the first time:
1. `NotImplementedError: mode=1 is not supported` from a signature change to Tensor.pad
2. `AttributeError: 'Tensor' object has no attribute '_slice'` which looks to be removed in preference of Tensor.pad and shrink

### DIFF
--- a/examples/vits.py
+++ b/examples/vits.py
@@ -114,7 +114,7 @@ class StochasticDurationPredictor:
     flows = flows[:-2] + [flows[-1]] # remove a useless vflow
     z = Tensor.randn(x.shape[0], 2, x.shape[2], dtype=x.dtype).to(device=x.device) * noise_scale
     for flow in flows: z = flow.forward(z, x_mask, g=x, reverse=reverse)
-    z0, z1 = split(z, [1, 1], 1)
+    z0, z1 = z.split([1, 1], 1)
     return z0.realize()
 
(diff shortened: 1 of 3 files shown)


examples/vits.py
```diff
@@ -484,10 +484,11 @@ def split(tensor, split_sizes, dim=0):  # if split_sizes is an integer, convert
     dim], "Sum of split_sizes must equal the dimension size of tensor along the given dimension."
   start, slices = 0, []
   for size in split_sizes:
-    slice_range = [(start, start + size) if j == dim else None for j in range(len(tensor.shape))]
```
chenyuxyz, line 487: just use Tensor.split should work

mulac: nice yeah, just used Tensor.split

chenyuxyz:
cool thanks, after install tons of dependencies and fix [this](https://github.com/tinygrad/tinygrad/issues/3118#issuecomment-1949830109)

`PHONEMIZER_ESPEAK_LIBRARY=/opt/homebrew/Cellar/espeak/1.48.04_1/lib/libespeak.dylib PYTHONPATH=. python examples/conversation.py` works fine for me!

this example is pretty outdated and can use some refactor if you are interested

PR #8426, opened 2024-12-26
Ordered tc kernel opts

ignaciosica:
# Tensor core refactor

**New approach** 
Avoids having to swizzle output as the correct swizzle for the output is constructed through kernel opts.

**Overview**
The refactor focuses primarily on the **implementation** side of tensor cores (`TC`) rather than over their abstraction, as George mentioned. While an abstraction refactor could align with the kernel OptOps refactor into pattern-matching (`PM`) rules, this current work keeps tensor cores clean. It addresses:

1. Complex thread access patterns required for specific backends.
2. Simplifying support for new TC shapes.
3. Constructing output swizzles through kernel opts.
4. Use the swizzle from the docs.

> By `thread access pattern` I mean the elements of the tensor core each collaborating thread is in charge of loading (from `A` and `B`) and storing (from `C`).

**Implementation details**
The `TC` class has been simplified, removing redundancies. Each tensor core now requires: `dims`, `upcast_size`, `dtype_in`, `dtype_out`, `opts`, and optionally, `swizzle`. All properties but the swizzle can be extracted directly through the documentation for each tensor core.

``` python
@dataclass(frozen=True)
class TensorCore: # D = A * B + C, A is (M x K), B is (K x N), C and D are (M x N)
dims: tuple[int,int,int] # N, M, K
threads: int # number of threads that construct the warp
elements_per_thread: tuple[int, int, int] # elements per-thread to load/store from A/B/C
dtype_in: DType # dtype for A and B
dtype_out: DType # dtype for C and D
opts: tuple[str, ...] # ordered tuple of "ux" or "lx" specifing kernel opts to perform. "ux" upcasts dim x and "lx" localizes dim x
swizzle: tuple[Optional[tuple[tuple[int, ...], tuple[int, ...]]], Optional[tuple[tuple[int, ...], tuple[int, ...]]]] = (None, None)
```

1. `threads`
Number of threads that construct the warp
2. `elements_per_thread`
Specifies the elements each thread load/store for A/B/C. Although this could theoretically be inferred from the shapetracker of the resulting source for the WMMA uop, padded tensor cores make this logic overly complex. Explicitly defining `elements_per_thread` simplifies this. This property can be derived from documentation.
3. `opts`
Specifies the kernel opts to perform over tensor cores buffers and the order in which to perform them. All kernel OptOps use a fixed value of 2, enabling a more straightforward permutation of shapes for the swizzle.
5. `swizzle`
The `st1_pattern` and `st2_pattern` properties have been replaced with `swizzle`, a tuple of two optional permutations describing the permutation required in order to fix the thread access patterns for the inputs of the tensor core.

**Advantages of the refactor**
The new implementation reduces the complexity of adding support for new tensor core shapes. Previously, the process required defining `thread_patterns`, `reduce_axes`, `upcast_axes`, `expanded_shapes`, and `local/upcast order`, all being specific to Tinygrad implementation. Now, only three elements are necessary apart from input and output dtypes:

1. tensor core dimensions (N, M, K)
2. threads
3. elements_per_thread
4. kernel opts to perform

These values are readily available in backend documentation and will get the tc running. While swizzling still needs to be implemented manually to get the correct result, this approach mostly eliminates concerns about Tinygrad-specific details.

**Making sense of the swizzle**

In order to explain the swizzle, I'm going to make a walkthrough in adding `tf32` tc support for CUDA. As a sidenote; I'm tempted to add the support for the other TC shapes in order to claim the bounties (Turing tc, int8 tc, ...) but it would be a good sign if somebody else could figure them out somehow simply.

Form the docs, we learn that the tensor core shape for `sm_80` with `tf32` as input dtype is:

In the figure, color scaling represent the threads. Threads are continuous from darker to lighter and then they follow the dotted line. That represent the elements that each thread is responsible for loading into the tensor core. Thread0 (T0) is responsible for loading a0-a3, b0-b1 and stores c0-c3. Thread1 (T1) is responsible for the following, lighter orange, elements.

From a first glance into the docs we can get `dims` and `upcast_size` properties of the tensor core:

1. `dims = (N=8, M=16, K=8)`
2. `elements_per_thread = (4(a0,a1,a2,a3), 2(b0,b1), 4(c0,c1,c2,c3))`

And, from a closer look, we can also derive the opts property as well. When deriving the kernel opts for wmma, each operation (“opt”) can either be an upcast (`u`) or a localization (`l`). Each opt is performed with a size of 2, and we traverse the matrix elements in a contiguous order to decide which dimension we need to act on:
1. **Starting at c0 → c1 (first row).**
    - We move along dimension 0 by performing an upcast: `u0`.
    - We continue along that same row with two localizations: `l0`, `l0`.
    - Thus, for the first row, our sequence is `("u0", "l0", "l0")`.

2. **Covering the next rows (still in the top half).**
    - We move to rows 2, 3–4, and 5–8 by localizing along dimension 1 each time. This yields three additional `l1`s:  
        `("u0", "l0", "l0", "l1", "l1", "l1")`.
    - Notice that at this point, we have localized five times total (`l0` three times plus `l1` twice, or vice versa depending how you count). Each localization has a size of 2, so `2**5=32`. This matches the 32 threads in a warp that collaborate for wmma.

3. **Moving to the second half of the matrix.**
    - Once we reach the lower rows (or the “second half” in memory layout terms), we see that it is reached by upcasting along dimension 1, giving `u1`.
    - Therefore, the final sequence becomes:`("u0", "l0", "l0", "l1", "l1", "l1", "u1")`

This final set of opts captures how the warp threads move through the matrix in a pattern of upcasting (to jump over a large dimension range) and localizing (to move sequentially through contiguous sub-blocks) until the entire block of data is covered.

The tensor core can then be added to the backend as follows:
```python
tensor_cores += [TensorCore(dims=(8,16,8), threads=32, elements_per_thread=(4,2,4), dtype_in=dtypes.float, dtype_out=dtypes.float, opts=("u0","l0","l0","l1","l1","l1","u1"))]
```

To figure out the required swizzle, the simplest approach is to perform a matrix multiplication with the same shape as the tensor core being configured. Running the following command: `DEBUG=3 N=8 M=16 K=8 python extra/gemm/simple_matmul.py` yields:

``` bash
TENSOR CORES [(1, 8, 1)] [(0, 16, 8)] WMMA_8_16_8_float_float
r_2_2_2_2_2_2_2_2_2_2
    UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(128), arg=0, src=()),
      UOp(Ops.WMMA, dtypes.float.vec(4), arg=('WMMA_8_16_8_float_float', (8, 16, 8), dtypes.float, dtypes.float, 'CUDA', 32, (((9, 2), (8, 2)), ((9, 2),), ((9, 2), (8, 2))), (5, 6, 7)), src=(
        UOp(Ops.CONTRACT, dtypes.float.vec(4), arg=((9, 2), (8, 2)), src=(
          UOp(Ops.LOAD, dtypes.float, arg=None, src=(
            UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(128), arg=1, src=()),
            UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(0, 0, 8, 16, 32, 1, 2, 4, 0, 64), offset=0, mask=None, contiguous=False),)), src=()),)),)),
        UOp(Ops.CONTRACT, dtypes.float.vec(2), arg=((9, 2),), src=(
          UOp(Ops.LOAD, dtypes.float, arg=None, src=(
            UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(64), arg=2, src=()),
            UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(2, 4, 0, 0, 0, 8, 16, 32, 1, 0), offset=0, mask=None, contiguous=False),)), src=()),)),)),
        UOp(Ops.CONST, dtypes.float.vec(4), arg=0.0, src=()),)),)),)),))

[...]

AssertionError: 
Not equal to tolerance rtol=0.03, atol=0.0001

Mismatched elements: 116 / 128 (90.6%)
Max absolute difference among violations: 2.3470442
Max relative difference among violations: 1.7517731
 ACTUAL: array([[2.176485, 2.176485, 2.176485, 2.176485, 2.176485, 2.176485,
        2.176485, 2.176485],
       [0.830115, 0.830115, 0.830115, 0.830115, 0.830115, 0.830115,...
 DESIRED: array([[0.959163, 1.724498, 2.56779 , 1.680298, 2.10738 , 2.124444,
        2.029053, 2.191531],
       [0.357985, 0.685142, 1.55628 , 0.793698, 1.318517, 1.090871,...
```

While the matmul runs successfully, the assertion error indicates that swizzling has not yet been performed. 
Note, however, that the output does have the correct swizzle; the strides for the local and upcast axes matches the documentation. Local strides are `(2, 4, 8, 16, 32)` and upcast strides are `(1, 64)`. 

We need to configure the ShapeTracker such that each thread is responsible for loading and storing the elements as specified in the documentation. In order to figure out the required swizzle, we are going to focus on the current ShapeTrackers for the buffers:

```python
A => ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(0, 0, 8, 16, 32, 1, 2, 4, 0, 64), [...])))
B => ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(2, 4, 0, 0, 0, 8, 16, 32, 1, 0), [...])))
```

Since the matrix multiplication matches the shape of the tensor core in use, all dimensions are either localized, reduced, or upcasted by TC OptOps. This means there are no global dimensions or extra reduce dimensions.

With a thread count of 32 and each local dimension fixed to 2, the resulting shape for locals is `(2, 2, 2, 2, 2)`, as `32 = 2 ** 5`. The product of `M * N = 8 * 16 = 4 * 32` leaves a remainder of 4 that cannot be localized, so it is upcasted, resulting in an upcasted shape of `(2, 2)`.

The reduce dimension `K = 8` is split into `(2, 2, 2)`. this results in shapetrackers having 10 dimensions of size 2.

The shapetracker configuration is as follows: `(locals)(reduce)(upcast) = (2, 2, 2, 2, 2) (2, 2, 2) (2, 2)`

From the docs we get the desired strides for the threads and upcasts.

In order to get the swizzle permutations, we just need to start from the real strides of the shapetracker and permute to the desired strides for fixing the buffer.

Now, if we introduce the permutations we get from the figures as the swizzle and run the same test as before, we get the correct result. We can see the shapetrackers have been fixed, getting the desired strides from the figures above.

```python
tensor_cores += [TensorCore(dims=(8,16,8), threads=32, elements_per_thread=(4,2,4), dtype_in=dtypes.float, dtype_out=dtypes.float, opts=("u0","l0","l0","l1","l1","l1","u1"), swizzle=(((5,6,2,3,4),(0,1,8,9,7)), ((5,6,8,0,1),(2,3,4,9,7))))]
```

``` bash
TENSOR CORES [(1, 8, 1)] [(0, 16, 8)] WMMA_8_16_8_float_float
r_2_2_2_2_2_2_2_2_2_2
    UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(128), arg=0, src=()),
      UOp(Ops.WMMA, dtypes.float.vec(4), arg=('WMMA_8_16_8_float_float', (8, 16, 8), dtypes.float, dtypes.float, 'CUDA', 32, (((9, 2), (8, 2)), ((9, 2),), ((9, 2), (8, 2))), (5, 6, 7)), src=(
        UOp(Ops.CONTRACT, dtypes.float.vec(4), arg=((9, 2), (8, 2)), src=(
          UOp(Ops.LOAD, dtypes.float, arg=None, src=(
            UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(128), arg=1, src=()),
            UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(1, 2, 8, 16, 32, 0, 0, 0, 64, 4), offset=0, mask=None, contiguous=False),)), src=()),)),)),
        UOp(Ops.CONTRACT, dtypes.float.vec(2), arg=((9, 2),), src=(
          UOp(Ops.LOAD, dtypes.float, arg=None, src=(
            UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(64), arg=2, src=()),
            UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(8, 16, 1, 2, 4, 0, 0, 0, 0, 32), offset=0, mask=None, contiguous=False),)), src=()),)),)),
        UOp(Ops.CONST, dtypes.float.vec(4), arg=0.0, src=()),)),)),)),))
*** CUDA      14 r_2_2_2_2_2_2_2_2_2_2                     arg  3 mem  0.00 GB tm     13.31us/     2.23ms (     0.15 GFLOPS    0.1|0.1     GB/s) ['matmul']
opened device CLANG from pid:818029
*** CLANG     15 copy      512,   CLANG <- CUDA            arg  2 mem  0.00 GB tm     93.66us/     2.32ms (     0.00 GFLOPS    0.0|0.0     GB/s) 
*** CLANG     16 copy      256,   CLANG <- CUDA            arg  2 mem  0.00 GB tm     63.75us/     2.39ms (     0.00 GFLOPS    0.0|0.0     GB/s) 
*** CLANG     17 copy      512,   CLANG <- CUDA            arg  2 mem  0.00 GB tm     38.80us/     2.43ms (     0.00 GFLOPS    0.0|0.0     GB/s) 
```

**tf32 kernel**
``` c
__device__ float4 __WMMA_8_16_8_float_float(float4 a, float2 b, float4 c){
  asm("mma.sync.aligned.m16n8k8.row.col.f32.tf32.tf32.f32"
      "{%0, %1, %2, %3}, {%4, %5, %6, %7},"
      "{%8, %9}, {%0, %1, %2, %3};"
    : "r"(a_pk[0]), "r"(a_pk[1]), "r"(a_pk[2]), "r"(a_pk[3]), "r"(b_pk[0]), "r"(b_pk[1]));
  return c;
}
extern "C" __global__ void __launch_bounds__(32) r_2_2_2_2_2_2_2_2_2_2(float* data0, float* data1, float* data2) {
  int alu0 = (lidx1<<4);
  int alu1 = (lidx2<<5);
  int alu2 = (lidx0>>2);
  int alu3 = (alu2<<3);
  int alu4 = (lidx0&1);
  int alu5 = ((lidx0>>1)&1);
  int alu6 = ((alu5<<1)+alu4+alu3+alu0+alu1);
  float val0 = *(data1+alu6);
  float val1 = *(data1+(alu6+4));
  float val2 = *(data1+(alu6+64));
  float val3 = *(data1+(alu6+68));
  int alu7 = ((alu4<<3)+(alu5<<4)+alu2+(lidx1<<1)+(lidx2<<2));
  float val4 = *(data2+alu7);
  float val5 = *(data2+(alu7+32));
  float4 wmma0 = __WMMA_8_16_8_float_float(make_float4(val0,val2,val1,val3), make_float2(val4,val5), make_float4(0.0f,0.0f,0.0f,0.0f));
  int alu8 = ((alu4<<1)+(alu5<<2)+alu3+alu0+alu1);
  *((float2*)((data0+alu8))) = make_float2(wmma0.x,wmma0.y);
  *((float2*)((data0+(alu8+64)))) = make_float2(wmma0.z,wmma0.w);
}
```

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -594,7 +594,7 @@ class Kernel:
         if (tc := self.tensor_core) and (self.use_tensor_cores == 1 or self.use_tensor_cores == 3):
           wd, tcd = self.global_dims, self.first_upcast
           def get_upcast_axes(buf):
-            upcast_axes = int(math.log2(tc.upcast_size[buf]))
+            upcast_axes = int(math.log2(tc.thread_dims[buf]))
             return tuple((tcd + len(tc.get_reduce_axes()) + len(tc.get_upcast_axes()) - (i+1), 2) for i in range(upcast_axes))
           def get_tc_swizzle_st(shape, wd_swizzle, tcd_swizzle):
             offset = (tcd - (wd + len(wd_swizzle)))
@@ -619,10 +619,10 @@ class Kernel:
           if self.use_tensor_cores == 1: # real WMMA, use CONTRACT/UNROLL to get the vectorization right
             tc_upcast_axes = (get_upcast_axes(0), get_upcast_axes(1), get_upcast_axes(2))
             wmma_arg = (str(tc), tc.dims, tc.dtype_in, tc.dtype_out, self.opts.device, tc.threads, tc_upcast_axes, tc_reduce_axes)
-            wmma = UOp(Ops.WMMA, dtype=tc.dtype_out.vec(tc.upcast_size[2]), src=(
-              UOp(Ops.CONTRACT, dtype=srcs[0].dtype.vec(tc.upcast_size[0]), src=(srcs[0],), arg=tc_upcast_axes[0]),
-              UOp(Ops.CONTRACT, dtype=srcs[1].dtype.vec(tc.upcast_size[1]), src=(srcs[1],), arg=tc_upcast_axes[1]),
-              UOp.const(tc.dtype_out.vec(tc.upcast_size[2]), 0.0)), arg=wmma_arg)
+            wmma = UOp(Ops.WMMA, dtype=tc.dtype_out.vec(tc.thread_dims[2]), src=(
+              UOp(Ops.CONTRACT, dtype=srcs[0].dtype.vec(tc.thread_dims[0]), src=(srcs[0],), arg=tc_upcast_axes[0]),
+              UOp(Ops.CONTRACT, dtype=srcs[1].dtype.vec(tc.thread_dims[1]), src=(srcs[1],), arg=tc_upcast_axes[1]),
+              UOp.const(tc.dtype_out.vec(tc.thread_dims[2]), 0.0)), arg=wmma_arg)
             tc_uop = UOp(Ops.UNROLL, tc.dtype_out, (wmma,), arg=tc_upcast_axes[2])
 
           else: # for TC=3 MUL/SUM instead of WMMA
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -594,7 +594,7 @@ class Kernel:
         if (tc := self.tensor_core) and (self.use_tensor_cores == 1 or self.use_tensor_cores == 3):
           wd, tcd = self.global_dims, self.first_upcast
           def get_upcast_axes(buf):
-            upcast_axes = int(math.log2(tc.thread_dims[buf]))
+            upcast_axes = int(math.log2(tc.elements_per_thread[buf]))
             return tuple((tcd + len(tc.get_reduce_axes()) + len(tc.get_upcast_axes()) - (i+1), 2) for i in range(upcast_axes))
           def get_tc_swizzle_st(shape, wd_swizzle, tcd_swizzle):
             offset = (tcd - (wd + len(wd_swizzle)))
@@ -619,10 +619,10 @@ class Kernel:
           if self.use_tensor_cores == 1: # real WMMA, use CONTRACT/UNROLL to get the vectorization right
             tc_upcast_axes = (get_upcast_axes(0), get_upcast_axes(1), get_upcast_axes(2))
             wmma_arg = (str(tc), tc.dims, tc.dtype_in, tc.dtype_out, self.opts.device, tc.threads, tc_upcast_axes, tc_reduce_axes)
-            wmma = UOp(Ops.WMMA, dtype=tc.dtype_out.vec(tc.thread_dims[2]), src=(
-              UOp(Ops.CONTRACT, dtype=srcs[0].dtype.vec(tc.thread_dims[0]), src=(srcs[0],), arg=tc_upcast_axes[0]),
-              UOp(Ops.CONTRACT, dtype=srcs[1].dtype.vec(tc.thread_dims[1]), src=(srcs[1],), arg=tc_upcast_axes[1]),
-              UOp.const(tc.dtype_out.vec(tc.thread_dims[2]), 0.0)), arg=wmma_arg)
+            wmma = UOp(Ops.WMMA, dtype=tc.dtype_out.vec(tc.elements_per_thread[2]), src=(
+              UOp(Ops.CONTRACT, dtype=srcs[0].dtype.vec(tc.elements_per_thread[0]), src=(srcs[0],), arg=tc_upcast_axes[0]),
+              UOp(Ops.CONTRACT, dtype=srcs[1].dtype.vec(tc.elements_per_thread[1]), src=(srcs[1],), arg=tc_upcast_axes[1]),
+              UOp.const(tc.dtype_out.vec(tc.elements_per_thread[2]), 0.0)), arg=wmma_arg)
             tc_uop = UOp(Ops.UNROLL, tc.dtype_out, (wmma,), arg=tc_upcast_axes[2])
 
           else: # for TC=3 MUL/SUM instead of WMMA
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -595,11 +595,11 @@ class Kernel:
           wd, tcd = self.global_dims, self.first_upcast
           tcd_reduce_len, tcd_upcast_len = len(tc.get_reduce_axes()), len(tc.get_upcast_axes())
           def get_upcast_axes(buf): return tuple((tcd+tcd_reduce_len+tcd_upcast_len - (i+1), 2) for i in range(int(math.log2(tc.upcast_size[buf]))))
-          def get_tc_swizzle_st(shape, wd_pattern, tcd_pattern):
-            offset = (tcd - (wd + len(wd_pattern)))
+          def get_tc_swizzle_st(shape, wd_swizzle, tcd_swizzle):
+            offset = (tcd - (wd + len(wd_swizzle)))
             permaxis = list(range(wd)) \
-              + [wd + x + (offset if x >= len(wd_pattern) else 0) for x in wd_pattern]  + list(range(wd + len(wd_pattern), tcd)) \
-              + [wd + x + (offset if x >= len(wd_pattern) else 0) for x in tcd_pattern] + list(range(tcd + len(tcd_pattern), len(shape)))
+              + [wd + x + (offset if x >= len(wd_swizzle) else 0) for x in wd_swizzle]  + list(range(wd + len(wd_swizzle), tcd)) \
+              + [wd + x + (offset if x >= len(wd_swizzle) else 0) for x in tcd_swizzle] + list(range(tcd + len(tcd_swizzle), len(shape)))
             return ShapeTracker.from_shape(shape).permute(tuple(permaxis))
 
           srcs = list((ret.src[0] if ret.src[0].op is not Ops.CAST else ret.src[0].src[0]).src)
@@ -616,7 +616,7 @@ class Kernel:
 
           tc_reduce_axes = tuple(tcd + ax for ax, _ in tc.get_reduce_axes())
           if self.use_tensor_cores == 1: # real WMMA, use CONTRACT/UNROLL to get the vectorization right
-            tc_upcast_axes= (get_upcast_axes(0), get_upcast_axes(1), get_upcast_axes(2))
+            tc_upcast_axes = (get_upcast_axes(0), get_upcast_axes(1), get_upcast_axes(2))
             wmma_arg = (str(tc), tc.dims, tc.dtype_in, tc.dtype_out, self.opts.device, (2**len(tc.get_local_axes())), tc_upcast_axes, tc_reduce_axes)
             wmma = UOp(Ops.WMMA, dtype=tc.dtype_out.vec(tc.upcast_size[2]), src=(
               UOp(Ops.CONTRACT, dtype=srcs[0].dtype.vec(tc.upcast_size[0]), src=(srcs[0],), arg=tc_upcast_axes[0]),
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -593,7 +593,7 @@ class Kernel:
 
         if (tc := self.tensor_core) and (self.use_tensor_cores == 1 or self.use_tensor_cores == 3):
           wd, tcd = self.global_dims, self.first_upcast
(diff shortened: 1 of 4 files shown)


geohot:
Nice! I knew this had to be possible

ignaciosica:
Yes! Thank you for your feedback!

geohot:
This ready to merge?

tinygrad/codegen/kernel.py
```diff
@@ -596,48 +592,42 @@ def reduced_axes(start, stop):
...
+            permaxis = list(range(wd)) \
+              + [wd + x + (offset if x >= len(wd_swizzle) else 0) for x in wd_swizzle]  + list(range(wd + len(wd_swizzle), tcd)) \
+              + [wd + x + (offset if x >= len(wd_swizzle) else 0) for x in tcd_swizzle] + list(range(tcd + len(tcd_swizzle), len(shape)))
+            return ShapeTracker.from_shape(shape).permute(tuple(permaxis))
 
           srcs = list((ret.src[0] if ret.src[0].op is not Ops.CAST else ret.src[0].src[0]).src)
-          for i, tc_pattern in enumerate([tc.st1_pattern, tc.st2_pattern]):
-            if tc_pattern: srcs[i] = srcs[i].view(fix_st(srcs[i].st_arg if srcs[i].op is Ops.LOAD else srcs[i].src[0].st_arg, *tc_pattern))
+          for i, (src, swizzle) in enumerate(zip(srcs, tc.swizzle)):
+            if swizzle: srcs[i] = apply_swizzle(src.view(get_tc_swizzle_st((src if src.op is Ops.LOAD else src.src[0]).st_arg.shape, *swizzle)))
```
geohot, line 607: Not a merge blocker, but I think this rewrite can wait until later instead of `apply_swizzle` now.

ignaciosica:
If you like the opts style ("u0", "l1", etc.), then yes, I tried other alternatives with tuples ("u", 0) but are minor changes. I was finalizing the guide with a walkthrough on adding tf32 and was planning to mark it as ready for review.

geohot:
Yea, that's fine. It's at least a lot better than what's there

tinygrad/codegen/kernel.py
ignaciosica: Sure, I'll leave it with `view` and `view_left` as before

tinygrad/renderer/__init__.py
```diff
@@ -9,19 +9,15 @@
 @dataclass(frozen=True)
 class TensorCore: # D = A * B + C, A is (M x K), B is (K x N), C and D are (M x N)
   dims: tuple[int,int,int] # N, M, K
+  upcast_size: tuple[int, int, int] # A, B, C/D
```
geohot, line 12: Can you explain what this is in a comment?

ignaciosica:
I have to go now, in ~4 hours I'm back home and I'll address your comments and upload the guide.

tinygrad/renderer/__init__.py
ignaciosica: Yes

geohot:
A suggestion. Add a `__post_init__` to the TensorCore class that checks the self consistency of the parameters while describing each check.

ignaciosica:
The length of the first tuple in swizzle is defined by local, and the length of the second is defined by unrolled + upcasted

(l0,l1,l2,l3,l4)(ur0,ur1,ur2,up0,up1) for example

geohot:
The checks (with comments) I think are the best way to understand code. The sooner it asserts when people are trying to add a new tensor core the better.

Reading this PR is the closest I have ever come to understanding how these things work!

ignaciosica:
Cool! I think the guide will further that as it help adding support to a new shape directly from the docs

ignaciosica:
> This ready to merge?

yes!

geohot:
Great work!

PR #8429, opened 2024-12-27
[wip] onnx running top 100 huggingface models

geohotstan:
Downloading many GBs over a slow internet, over a vpn, is no fun 🫠 .

`extra/datasets/onnx_model_download.py` this is the file. Can use `MODEL=google-bert/bert-base-uncased` for running individual models
I also have a few other PRs included in this (including #8109 lol I forked it) that I will see to get merged first.

This currently passes with `PYTHONPATH=. python extra/datasets/onnx_model_download.py`
We currently test the `POTENTIAL_MODEL_PATHS`. So we aren't testing quantized or encoder stuff or other things I don't understand :D  (gonna try running them to try to understand them)
We do skip some models for ORT verification (ort failure) and 2 other models that have missing OP implementation.

There are a few problems I ran into while working on this:
- some models have a whole pipeline of different models like https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0. The implementation of the pipeline cannot be generic/shared unfortunately
- `get_input` is not robust at all. It was the naive `randn` with value `1` as variable shape before (where we get gather oob error and all nan output), so I have made it a bit better. But lets say when new month starts and new models start appearing, this get_input is likely to fail.

Yeah....
Not sure how this would be tested in CI....


chenyuxyz:
i think it's fine to keep a file of a fixed 100 models, and make it very easy to update with a script, and we can monitor if anything failed when we try to update the model list.

fixed dlownloaded models go into cache so ci should be good

geohot:
How many GB is it? What's the distribution of sizes?

geohotstan:
> How many GB is it? What's the distribution of sizes?

10-100 MB: 5 files
100 MB - 1 GB: 59 files
1 GB - 2 GB: 18 files
2 GB - 5 GB: 11 files
5 GB: 1 files

80G    /Users/zibo/Library/Caches/tinygrad/downloads

chenyuxyz:
to pace this a bit, i suggest (1) having a standalone script that fetch top N model with some stats for how many we support already, and some op info for the ones that we don't support, and (2) support testing in existing benchmark onnx, maybe refactor the existing onnx benchmark test functions too.

then (1) can call the helper functions in (2) and once tools are set up, work on improve the coverage

generally i suggest thinking about what intermediaries has standalone value and can let other people help you too

PR #8432, opened 2024-12-27
start on test rewrite map [pr]

geohot:
~~Test failure is interesting, I think we don't fully understand the GC patterns.~~ frame issue

TODO tomorrow:
1. Add the graph finder to Tensor.schedule, all Tensors and parents
2. Use this rewrite function on the UOp graph
3. Update all lazydata in the Tensors with the rewritten UOp

This will run before create_schedule. The same logic will also be used to remove UOp mutability.

Ugh, MultiLazyBuffer is annoying.


geohot:
We might not even need graph_rewrite_map. Put all the Tensors in sink, then read the sink srcs in the returned src.

Qazalin:
removing the tests I wrote in test_rewrite_map_schedule, the ideas are outdated. There's better things in #8432.

PR #8434, opened 2024-12-28
refactor lower_schedule_item to pattern matchers [pr]

Qazalin:
Refactoring this to a pattern matcher is a prereq for using existing UOps to compose "meta ops".
Then instead of `UPat(Ops.EMPTY)` we can have `VIEW(BUFFER(DEVICE)): EmptyOp` etc.

btw, I'm not happy that Runners need literal Buffers to construct, but this predates `Ops.DEVICE`, I can clean this up next.

Ideally, we want `si_lowerer` to take an AST (UOp), no context, and return a single Runner instance. The Runner abstraction does not need to know about Device Buffers, ExecItem can handle that.


geohot:
Nice!

PR #8436, opened 2024-12-28
const copy folding spec + multi.py behavior [pr]

Qazalin:
The hack in copy_to_device can't stay. It doesn't belong in ops.
This diff includes the spec for rewriting `COPY(CONST) -> CONST`, as it exists in ops_folding.

The extra check for "is const" in ops only exists for multi. Sadly multi calls `contiguous` on all shards:
```
return MultiLazyBuffer([lb if lb.is_unrealized_unmasked_const() else lb.contiguous(allow_buffer_view=False) for lb in sharded_lbs], axis)
```
The scheduler respects contiguous and will dutifully store the thing you call contiguous on.

See how if we delete "early const folding" in copy_to_device, this contiguous prevents further folding:
image

vs if it was just `return MultiLazyBuffer(sharded_lbs, axis)` all MUL 0s would fold until we end up with a NOOP:
image

### DIFF
--- a/test/test_const_folding.py
+++ b/test/test_const_folding.py
@@ -240,6 +240,8 @@ class TestMultiConstFolding(unittest.TestCase):
     _check_ast_count(0, t ** 1)
     _check_ast_count(0, 1 ** t)
 
+  # failing because multi calls .contiguous() on every single sharded uop
+  @unittest.expectedFailure
   def test_multi_const_folding_tensor(self):
     ds = tuple(f"{Device.DEFAULT}:{i}" for i in range(4))
     t = Tensor.arange(16).float().realize().to(ds)
--- a/test/test_multitensor.py
+++ b/test/test_multitensor.py
@@ -5,7 +5,7 @@ from tinygrad.ops import Ops
 from tinygrad.helpers import CI, getenv, prod, Context
 from tinygrad.nn.state import get_parameters, get_state_dict
 from tinygrad.engine.schedule import create_schedule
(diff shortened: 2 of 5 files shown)


Qazalin:
Is the extra is_const check worth keeping around?
image

geohot:
Why does multi have to call contiguous?

Qazalin:
agree that contiguous has to go, it's way too aggressive.


benchmarks fails when I delete it:

https://github.com/tinygrad/tinygrad/actions/runs/12536417701/job/34959405143#step:20:11

test_multitensor doesn't give a signal for why. The only test that failed was test_shard_memory, expecting a VIEW of a shard (on the same device!) to be realized.
```
  File "/Users/qazal/code/tinygrad/test/test_multitensor.py", line 698, in test_shard_memory
    assert lb is lb.base, f"{lb} {t.device}"
           ^^^^^^^^^^^^^
AssertionError: UOp(Ops.SHRINK, dtypes.float, arg=((0, 4), (0, 16)), src=(
  UOp(Ops.VIEW, dtypes.float, arg=ShapeTracker(views=(View(shape=(16, 16), strides=(16, 1), offset=0, mask=None, contiguous=True),)), src=(
    UOp(Ops.BUFFER, dtypes.float, arg=(6, 256), src=(
      UOp(Ops.DEVICE, dtypes.void, arg='LLVM', src=()),)),)),)) ('LLVM', 'LLVM:1', 'LLVM:2', 'LLVM:3')
```
I don't think some of multi's tests are well thought through, neither do they give a comprehensible feedback loop for why benchmarks failed.

PR #8439, opened 2024-12-29
Make tinygrad whisper work on recording

mesozoic-egg:
One early issue is repeated words. This is common with smaller model and was also discused in the [original paper](https://cdn.openai.com/papers/whisper.pdf) (section 4.5). The reason is that tokens are sampled one after another (seq-to-seq) and degradation can easily accumulate.

One solution is to add randomness in sampling and do it multiple times. Openai's implementation does it by  [checking for repetition](https://github.com/openai/whisper/blob/90db0de1896c23cbfaf0c58bc2d30665f709f170/whisper/transcribe.py#L206) and [repeatedly sampling](https://github.com/openai/whisper/blob/90db0de1896c23cbfaf0c58bc2d30665f709f170/whisper/decoding.py#L283), I adopted the same method.

This is the output showing the model can now recover from such repetition `python examples/whisper.py ~/Downloads/2024-12-23_compressed.mp3` (model is `tiny.en`):
Screenshot 2024-12-29 at 18 40 58


mesozoic-egg:
This is the result for 2024-12-23's transcript, running with BEAM=2 for the "tiny" size, at about 80 toks per second on a 3060: `BEAM=2 python examples/whisper.py ~/2024-12-23.mp3", result saved to a file called "transcribed.txt" in pwd.

image

I disabled timestamp prediction for two reasons. On tiny model it requires checking the probability sum of all timestamps and masking text tokens if the value is greater than that for text, this requires boolean indexing (not supported), and doing it in numpy makes things slower too. Not sure how important it is. Second, processing the timestamp need extra empirical code adding to LOC.  To me having a 30seconds segment seems okay. 

I don't think whisper has diariazation, so still need to figure that out. Will also try larger models and add some tinygrad special terms to the prompt.

@geohotstan, would like to get some feedback from you, although still in draft. What's your workflow for the transcript? 

[transcribed.txt](https://github.com/user-attachments/files/18308477/transcribed.txt)

geohotstan:
I tried running [97626d1] (https://github.com/tinygrad/tinygrad/pull/8439/commits/97626d16b696259ea790a25663541092b44b9825) last week
and was still getting occasional repeated texts. (it tried to increase temperature but it had repeated text for every increased temperature.)

I'll try testing it again.

for timestamp prediction, do you mean word level time stamps? We don't currently need word level timestamps. As for the diarization stuff I have been using pyannote. I think I can figure that other stuff out. We just need the text to not repeat first 😅

geohotstan:
ohhh I see what you mean by timestamp prediction. right now the timestamp is just coming out as 30sec chunks. Yeah, I think we need that. The diarization depends on correct timestamp segments.

mesozoic-egg:
thanks for testing! I think accurate timestamp is still required in general for long form audio transcription or any downstream processing. Openai's implementation checks for 1. gzip text compression ratio (✅) 2. avg log probs, 3. no speech probs to decide if resampling is needed. And the resampling is done with 1. multinomial (✅) for low temperature and 2. beam search for higher temperature. Finishing the missing pieces should make this a more robust component. Right now quality is not good enough if timestamp is being generated.

mesozoic-egg:
Turns out suppressing non-speech token greatly improved the result! maybe not all the decoder trickery is needed after all...
image

mesozoic-egg:
see attached transcript for this week's meeting with commit e549d3e with `SMALL=1 python examples/whisper.py {filename}`. It uses argmax sampling, with more logits masking heuristics, and fallback to multinomial depending on repetitiveness. Much improved quality! 
[2025-01-13.txt](https://github.com/user-attachments/files/18411232/2025-01-13.txt)


I still want to add beam search decoding that could make sampling more reliable (specifically, less likely to require a fallback). The challenge is doing kv cache rearrangement on JIT'ted buffer.

geohotstan:
Ooooooooo that's pretty darn good!
The phrase segmentation is nice. 😄 

There are still some differences with whisperx's since I think they also merge some segments, and there are still some transcribe quality stuff like word repetition (which may just be fixed with a larger model instead of post-processing?)
I'm gonna play with this tmrw and try with larger models.

also 
image
ah yes, swizzling cancer course. 🤖

mesozoic-egg:
So I added beam search, but the result seem worse than argmax. I pushed a commit to default to argmax if you are testing it out.

While testing on this week's recording, I noticed that the timestamp quality degrades if any segment falls into the repetitiveness issue. Might just need more time to work on this 😅

I read somewhere that the repetitiveness thing is inherent to the whisper model, so post processing is just a remedy. Also, post processing can only help balance the overall quality (things like timestamp, repetitiveness, start of text randomnness etc.). I'm testing mostly on smaller model because it's easier to reproduce these issues. But the text content themselves can only be solved with bigger ones.

WhisperX has a lot more pre and post processing, I don't think this PR will get to the same quality. But making it useable/reliable for recording is still a reasonable goal.

This seek's transcript `SMALL=1 python examples/whisper.py ~/Downloads/2025-01-20.mp3` (still argmax)
[2025-01-20.txt](https://github.com/user-attachments/files/18482087/2025-01-20.txt)

I'm thinking maybe it's fine to just use argmax with bigger model...

geohotstan:
sry I've been afk recently.

I tried with https://github.com/tinygrad/tinygrad/commit/e549d3e81195ae9f8afc0dcce0fad63f1a7e4f25 and it wasn't working 🤔. I'll try it again with the most recent commit.

Though the txt file you produced did have segmentation that looks pretty close to whisperx's. (some segmentation breakpoints are different and some of their segments are merged better)

I think if we can get tinygrad whisper producing the kind of segmentation you have in your shared files, it's good enough.
I'm just going to replace their part 1 of their pipeline with what we have and see if it works

image

the format of the expected `result` is just 
`{'segments': [{'text': " Are you going to start? Any company update? So you post our revenue line on Twitter? Yeah. Pretty good for a first year of revenue. We're selling hardware. Yeah. Yeah. No, I don't know. I'm hoping for rapid growth next year.", 'start': 0.284, 'end': 29.461}], 'language': 'en'}`

so it would be nice if the output of our thing was json friendly :D 

as for the rest of the pipeline, I've found an onnx model for the diarize stuff and I have no idea about the align loop... It's gonna take some time...

All in all, the output is already pretty good. I'll test replacing their part 1 with our's and if that kinda works, it's good!

mesozoic-egg:
So I cleaned up a few things and fixed beam decoding, also added json output! I tested it on tiny18 (and my two other machines) so it should be pretty reproducible (a268a1a).
[20250120-1.txt](https://github.com/user-attachments/files/18551538/20250120-1.txt)
[20250120-1.json](https://github.com/user-attachments/files/18551539/20250120-1.json)

Usage is `python examples/whisper.py --model="medium.en" --audio="$HOME/20250120.mp3"` (or use`--help`)

I noticed that the large model does not work as well, even in the original whisper implementation by openai. It falls into repetition more easily and time stamp segmentation is less accurate. Specifically, it has a low chance of hitting the "end of text" token that terminate a 30seconds segment. I'm not sure why though.

Speed is painfully slow though. Mostly because updating kv cache during beam search decoding split up the JIT, and the update itself isn't jittable (I haven't found a way to jit it yet...)

Overall I think the functionality is complete. Will clean up the code more, and return from the function a dictionary so you can pass it to other models! and maybe work on speed a bit.

mesozoic-egg:
Yesterday's meeting
[2025-01-28.txt](https://github.com/user-attachments/files/18566062/2025-01-28.txt)
[2025-01-28.json](https://github.com/user-attachments/files/18566063/2025-01-28.json)
I noticed that the generated timestamp is like 30s ahead of some actual audio (for example at 46minutes 0 seconds). There's some long "no speech" segment in this recording and I don't think that's handled well yet. I should add the "no speech probability" handling in the decoder, and debug if there's a bug in the time accumulator code.

PR #8444, opened 2024-12-29
scheduler refactors and renames from the buffer_shape branch [pr]

Qazalin:
Ideally all scheduler side-effects should be done as close to when we return the `list[ScheduleItem]`.
also, EMPTY needs a ShapeTracker in sources at the very least.

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -39,7 +39,7 @@ tensor_uop_spec = PatternMatcher([
   # Tensor variable bindings
   (UPat(Ops.BIND, dtypes.int, (UPat(Ops.DEFINE_VAR), UPat.cvar(dtype=dtypes.int)), arg=None), lambda: True),
 
-  # Tensor const has a ShapeTracker of shape=() and a device
+  # Tensor const has an unmasked ShapeTracker and a device
   (UPat(Ops.CONST, src=(UPat(Ops.VIEW, name="st", src=(UPat(Ops.DEVICE),)),)), lambda st: all(v.mask is None for v in st.st.views)),
 
   # DETACH and CONTIGUOUS change how we interpret the source UOp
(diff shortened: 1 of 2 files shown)


tinygrad/engine/schedule.py
```diff
@@ -32,15 +32,15 @@
...
+   # "make things that can't be images not images" can change the buffer dtype
+   # this is fine as long as it's a realized buffer and base dtypes match.
    ((isinstance(mv.dtype, ImageDType) or isinstance(x.dtype, ImageDType)) and x.dtype.base == mv.dtype.base and x.is_realized)),
 
   # Tensor variable bindings
   (UPat(Ops.BIND, dtypes.int, (UPat(Ops.DEFINE_VAR), UPat.cvar(dtype=dtypes.int)), arg=None), lambda: True),
 
   # Tensor const has a ShapeTracker of shape=() and a device
-  (UPat(Ops.CONST, src=(UPat(Ops.VIEW, src=(UPat(Ops.DEVICE),)),)), lambda: True),
+  (UPat(Ops.CONST, src=(UPat(Ops.VIEW, name="st", src=(UPat(Ops.DEVICE),)),)), lambda st: all(v.mask is None for v in st.st.views)),
```
chenyuxyz, line 43: this reads weird to me as if you are checking already, it should only has one view and that view has all 0 stride.

checking all mask is None in (multiple) views has very little semantic meaning

chenyuxyz (commented):
i think it's good to separate cosmetic changes from code changes in prs, would be easier for readers (and people who digs old prs)

tinygrad/engine/schedule.py
Qazalin: is there a better way to check if the st is unmasked?

chenyuxyz: if you want to a check is a st is unmasked, maybe no None in real_strides, but i don't think that's what you want. also not clear why checking mask is None for all views is right. won't it contain false negatives?

const shapetracker should be single view, no mask, and all 0 stride, or there's something wrong with view math

Qazalin:
chenyu, I removed the two code changes, will move the const spec restriction to a new diff. I do wanna restrict the spec as much as possible.

PR #8445, opened 2024-12-30
add mod to Tensor

geohotstan:
tests passed by splitting the kernel with contiguous 🤔 
I guess some rewrite is rewriting this part `r + b * (((r < 0) & (b > 0)) | ((r > 0) & (b < 0)))` away
gonna stare at this a bit more and maybe practice my VIZ skillz...

  
> possibily can be cleaner with a floor div

something like this without the new `class Mod(Function):`?
```python
def mod(self, x:Union[Tensor, ConstType], reverse=False) -> Tensor:
  a, b = self._broadcasted(x, reverse)
  return a - b * (a/b).floor().cast(self.dtype)
```

or by fixing `def __floordiv__(self, x): return self.idiv(x)  # TODO: idiv is trunc div, not floordiv`?

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -3140,7 +3140,7 @@ class Tensor(SimpleMathTrait):
     ```
     """
     a, b = self._broadcasted(x, reverse)
-    return (r := F.Mod.apply(a, b)) + b * (((r < 0) & (b > 0)) | ((r > 0) & (b < 0)))
+    return a-b*(a/b).floor().cast(dtypes.int32)
 
   def xor(self, x:Union[Tensor, ConstType], reverse=False) -> Tensor:
     """
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -3140,8 +3140,7 @@ class Tensor(SimpleMathTrait):
     ```
     """
     a, b = self._broadcasted(x, reverse)
-    r = F.Mod.apply(a, b).contiguous()
-    return r + b * (((r < 0) & (b > 0)) | ((r > 0) & (b < 0)))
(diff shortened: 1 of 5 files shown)


geohotstan:
disabling this rule makes it pass without contiguous

`(UPat(GroupOp.ALU, name="x"), lambda x: x.const_like(x.vmin) if x.vmin == x.vmax else None),` [here](https://github.com/tinygrad/tinygrad/blob/master/tinygrad/ops.py#L1232)

tinygrad/tensor.py
```diff
@@ -3129,6 +3129,20 @@ def div(self, x:Union[Tensor, ConstType], reverse=False) -> Tensor:
...
+    Mod `self` by `x`.
+    Equivalent to `self % x`.
+    Supports broadcasting to a common shape, type promotion, and integer inputs.
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    print(Tensor([-4, 7, 5, 4, -7, 8]).mod(Tensor([2, -3, 8, -2, 3, 5])).numpy())
+    ```
+    """
+    a, b = self._broadcasted(x, reverse)
+    r = F.Mod.apply(a, b).contiguous()
```
geohot, line 3143: Why is this contiguous needed?

geohotstan: https://github.com/chenyuxyz/tinygrad/pull/636
without it tests fail.

`(UPat(GroupOp.ALU, name="x"), lambda x: x.const_like(x.vmin) if x.vmin == x.vmax else None)`
This is the rule that's causing problems.

chenyuxyz:
see #8458, should unblock you and thanks for root causing the issue!

PR #8447, opened 2024-12-30
restrict tensor const ShapeTracker in spec [pr]

Qazalin:
based on chenyu's suggested way to validate const ShapeTracker.

Also, the extra_spec has to run before the main spec. We should unify CONST's definition, the spec in ops.py [just checks dtypes](https://github.com/tinygrad/tinygrad/blob/96f2a83a62a2bc9814146d3614c1f241f95bb7ae/tinygrad/ops.py#L929).


geohot:
In order to make both rules hit, in the ops pattern matcher, False means no and quit, None means try the next matching rule, and True means yes and quit.

tinygrad/engine/schedule.py
```diff
@@ -39,8 +39,9 @@
   # Tensor variable bindings
   (UPat(Ops.BIND, dtypes.int, (UPat(Ops.DEFINE_VAR), UPat.cvar(dtype=dtypes.int)), arg=None), lambda: True),
 
-  # Tensor const has a ShapeTracker of shape=() and a device
-  (UPat(Ops.CONST, src=(UPat(Ops.VIEW, src=(UPat(Ops.DEVICE),)),)), lambda: True),
+  # Tensor const has an unmasked ShapeTracker of stride 0 and a device
+  (UPat(Ops.CONST, src=(UPat(Ops.VIEW, name="st", src=(UPat(Ops.DEVICE),)),)),
+   lambda st: all(s == 0 for s in st.st.real_strides()) and len(st.st.views) == 1),
```
chenyuxyz, line 44: you can directly check stride of the view if there's only one view

tinygrad/engine/schedule.py
```diff
@@ -39,8 +39,9 @@
   # Tensor variable bindings
   (UPat(Ops.BIND, dtypes.int, (UPat(Ops.DEFINE_VAR), UPat.cvar(dtype=dtypes.int)), arg=None), lambda: True),
 
-  # Tensor const has a ShapeTracker of shape=() and a device
-  (UPat(Ops.CONST, src=(UPat(Ops.VIEW, src=(UPat(Ops.DEVICE),)),)), lambda: True),
+  # Tensor const has an unmasked ShapeTracker of stride 0 and a device
+  (UPat(Ops.CONST, src=(UPat(Ops.VIEW, name="st", src=(UPat(Ops.DEVICE),)),)),
+   lambda st: len(st.st.views) == 1 and all(s == 0 for s in st.st.views[0].strides)),
```
chenyuxyz, line 44: for one view also check `and mask is None`

Qazalin:
status: the core diff is ready. One of the tests fails non deterministically on WebGPU, merging after that is fixed.

PR #8453, opened 2024-12-31
fix rearrange docs

benettia:
fixes small typo in rearrange docs https://github.com/tinygrad/tinygrad/blob/master/tinygrad/tensor.py#L1921 + add the method in the documentation

### DIFF
--- a/docs/tensor/movement.md
+++ b/docs/tensor/movement.md
@@ -7,7 +7,6 @@
 ::: tinygrad.Tensor.flip
 ::: tinygrad.Tensor.shrink
 ::: tinygrad.Tensor.pad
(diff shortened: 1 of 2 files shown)


docs/tensor/movement.md
```diff
@@ -7,6 +7,7 @@
 ::: tinygrad.Tensor.flip
 ::: tinygrad.Tensor.shrink
 ::: tinygrad.Tensor.pad
+::: tinygrad.Tensor.rearrange
```
chenyuxyz, line 10: it should not be here

benettia: my bad, where is the best place to put it?

PR #8460, opened 2024-12-31
Fix uop.st for CLANG+AMX

uuuvn:
```python
# CLANG=1 AMX=1 python fun.py will fail with:
# File "/Users/user/src/tinygrad/tinygrad/ops.py", line 291, in st
#   case Ops.REDUCE_AXIS | Ops.WMMA: shape = src_sts[0].reduce(self.axis_arg)
#                                            ~~~~~~~^^^
# IndexError: list index out of range
from tinygrad import *

x = Tensor.rand(1, 32, 64, 128)
w = Tensor.rand(16,32,1,1)
o = x.conv2d(w).elu()
o.realize()
```
This bug was introduced in https://github.com/tinygrad/tinygrad/pull/8428
In process of rewriting `Ops.EXP2` transcendental calls `bitcast` which calls `can_view` which accesses `.st` which fails because it assumes that all `Ops.REDUCE_AXIS | Ops.WMMA` have at least one src with shapetracker

### DIFF
--- a/test/test_uops.py
+++ b/test/test_uops.py
@@ -592,6 +592,13 @@ class TestShapeSpec(unittest.TestCase):
     r = Tensor.empty(4, 4).sum(axis=1)
     self.assertEqual(r.lazydata.st, ShapeTracker.from_shape((4,)))
 
+  def test_st_wmma_none(self):
+    A = UOp(Ops.DEFINE_VAR, dtypes.float.vec(16), arg=('a', UOp.const(dtypes.float, 0), UOp.const(dtypes.float, 1)))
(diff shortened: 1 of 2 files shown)


Qazalin:
hmm, that doesn't look like the right fix though.

Qazalin:
```diff
- case Ops.REDUCE_AXIS | Ops.WMMA
+ case Ops.REDUCE_AXIS: # ...unchanged
+ case Ops.WMMA: # ....early return if it's shapeless otherwise do the reduce.
```

geohot:
@Qazalin ?

Qazalin:
yea it's good to go, sorry for the delay.

Issue #8457, opened 2024-12-31
sdxl.py: maximum recursion depth exceeded (24.04 Py 3.12.3)


FeepingCreature:
edit: This is probably the Python recursion limit bug but I don't understand why it still happens on Ubuntu's 3.12.3.

edit: Is it the recursion limit bug...? `RewriteContext.rewrite` doesn't have a cache annotation. It shouldn't use C stack. But then why does it ignore setrecursionlimit?

```
tinygrad/examples $ python3 sdxl.py --weight ~/sd_xl_base_1.0.safetensors 
ram used:  6.93 GB, sigmas: 100% ...  2515/2515 [00:04<00:00, 510.56it/s]
loaded weights in 4993.60 ms, 6.93 GB loaded at 1.39 GB/s
created batch
```Traceback (most recent call last):
  File "/home/feep/tinygrad/tinygrad/ops.py", line 897, in rewrite
    new_src = tuple(map(self.rewrite, n.src))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/feep/tinygrad/tinygrad/ops.py", line 897, in rewrite
    new_src = tuple(map(self.rewrite, n.src))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/feep/tinygrad/tinygrad/ops.py", line 897, in rewrite
    new_src = tuple(map(self.rewrite, n.src))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 997 more times]
RecursionError: maximum recursion depth exceeded
```

Happens with 8a8eaa1ed9be7d250b33dc628bd7d6e7c5b4f712 (Wed Dec 18 16:29:12) on Linux 6.12.7 with Radeon 7900 XTX, Python 3.12.3.

edit: By crashing earlier, the start of the backtrace seems to be:
```
Traceback (most recent call last):
  File "/home/feep/tinygrad/examples/sdxl.py", line 409, in <module>
    z = sampler(model.denoise, randn, c, uc, args.steps, timing=args.timing)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/feep/tinygrad/examples/sdxl.py", line 350, in __call__
    x, old_denoised = self.sampler_step(
                      ^^^^^^^^^^^^^^^^^^
[...]

PR #8461, opened 2024-12-31
Workaround for #8457: make RewriteContext rewrite() iterative to avoid Python stack limit

FeepingCreature:
On my 24.04 LTS system, Ubuntu installs Python 3.12.3-0ubuntu2. This seems to run into an unclear bug (#8457) where the SDXL demo busts the Python stack limit, even though `setrecursionlimit` is set high. It doesn't seem to be the 3.12 bug/change where C calls have their own limit either, because there's no C calls corecursing in the backtrace. I'm confused; but just in case, here's a rewrite of the relevant class that makes the recursion explicit so Python can't complain. With this, the SDXL demo works again on my system.

Is it a good idea to merge this? Eeeeehhhhhhhhhhhhhhhh. It is very ugly code. At least I should probably check if it's fixed on a newer Python; if it's just a matter of waiting... then again, it's unclear how long Ubuntu 24 LTS is gonna be on this version.

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -892,6 +892,8 @@ class RewriteContext:
     stack: list[tuple[UOp, list[UOp], bool]] = [(n, [], False)] # n, map result, returning from recursive call
     ret: Optional[UOp] = None
     while stack:
+      if len(stack) > 10000:
+        raise RecursionError("rewrite depth exceeded")
       n, results, returning = stack[-1]
       if len(results) < len(n.src):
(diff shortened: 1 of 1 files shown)


geohot:
I support this in theory, but that's a lot of lines. Can you do this without increasing lines?

geohot:
Closing, no reply for 2 weeks

gilion-joel-b:
@geohot @FeepingCreature This is the same issue which crashes `test.test_linearizer_failures.TestLinearizerFailure.test_failure_53` right?

PR #8464, opened 2025-01-01
tinychat in browser

hooved:
This PR adds support for tinychat in the browser on WebGPU. Both Llama-3.2-1B and tiktoken are run in the browser with full integration to the tinychat client, removing dependence on the previous python API server.

I faced the following critical unresolved issues when making llama work in the browser with WebGPU.

**Issue 1:** When attempting to compile to the javascript llama code with tinygrad from python, using `Variable("start_pos", 1, self.max_context).bind(start_pos)` in python produces javascript code that is not usable because it has python UOp code inserted in it. We should make the original python Variable.bind code compile to javascript properly (or implement a similarly efficient solution for getting our desired kernels). To get something working in the browser as a starting point for this bounty, I modified the llama inference to work with padding tokens which are masked during attention.

**Issue 2**: The llama attention layers' kv caches seem not to be updated during inference, in the js code produced by the current compiler. To make it work for now, I'm not using the kv cache, but of course this is not good.

As a result of the above two issues, the current browser WebGPU inference, although correct, is unacceptably slow, taking a few seconds to produce each new token. I'm interested in addressing these issues but would appreciate feedback.

Other outstanding issues:

- Implement CLANG support
- We could use fp16 weights instead of fp32 weights, but are blocked by lack of support by the python wgpu adapter. I think @wpmed92 has been looking into fp16 support in general for WebGPU adapters


wpmed92:
@hooved I see you used the stable diffusion compile script as a reference. That needs to be cleaned up. 
As for fp16, i have python bindings to Dawn (Google's engine) which supports fp16. I'll PR Dawn and fp16 support soon. After that both stable diffusion and llama compile can be implemented a lot cleaner because you don't need all these weight splitting stuff and weight decompression stuff.

geohot:
You have a link to the demo?

hooved:
I don't have a usable link yet, but I could prioritize that. I'd host the split weights and tokenizer, targeting a similar experience to wpmed's public stable diffusion demo.

I included a README for reproducing everything I've done up until now to run tinychat in the browser, albeit hosted locally.

hooved:
Here's a link to the demo: [https://hooved.github.io/tinychat-wgpu](https://hooved.github.io/tinychat-wgpu)

I want to add a progress bar to the UI for tracking progress of download/decompression, but for now progress can be monitored in the JS console.

After seeing `Transformer setup without exceptions` in the console, the app is ready for prompts. The inference is still very slow (about 0.1 tokens per second) due to the issues I mentioned in my first comment, which I'd look forward to fixing.

The page downloads 1GB of quantized weights and decompresses to float32, ultimately consuming about 6.8 GB of GPU memory.

hooved:
tinychat runs fast now at this link: https://hooved.github.io/tinychat-wgpu, around 15-20 tokens per second.

I fixed the kv cache / symbolic issues by manually debugging / editing the output WebGPU javascript code. Next I could work on getting tinychat running on CLANG.

If you clicked the link before, you may need to empty your indexeddb cache for the site because I refreshed the hosted weights.

hooved:
Based on chenyu's feedback, my priority now is to make the user experience good.

Below are issues I'll address:

- Make the page not lock up while weights are being decompressed
- Communicate model loading status clearly to the user, like with a progress bar
- Implement indexeddb cache file versioning, so the user never has cache problems if we change host weights
- Cache decompressed weights after first load, for fast reloading of tinychat (we could make this optional, because it'd consume ~5.5 GB disk with float32 weights)

hooved:
The demo is better now, https://hooved.github.io/tinychat-wgpu.

Next things to look at:
- [ ] handling conversations that go over max context, which then produce gibberish (also the case in the previous tinychat python server app)
- [ ] switching conversations while model is generating tokens breaks the app (also the case in the previous tinychat python server app)
- [ ] getting faster kernels with BEAM=2 for weights decompression / model inference (they get generated, but don't work in the browser yet), and patching compile code accordingly
- [ ] getting kernels that accept larger input sizes, for weights decompression chunk size, and for increasing max context during inference (tinygrad WEBGPU compiler throws exceptions if the input Tensor sizes are too big)
- [ ] patching the compile code to properly render symbolic variables in rendered JS code. I know what the target output should be because I fixed the output to work in the browser.
- [ ] CLANG support - to me this is the real prize of the bounty. I want this to Just Work (ideally even on a phone), most people aren't going to figure out how to enable WebGPU if it isn't by default.
- [ ] [Fix rendering of exported CLANG code](https://github.com/tinygrad/tinygrad/pull/8464/commits/43824c3d4934d20a25c8c80ff0bb6dc19d3bd7c3)
- [ ] [CLANG compile code has a bug](https://github.com/tinygrad/tinygrad/pull/8464/commits/6f30033e1392361df13d5f55964d194c887ac59c) with BEAM=2, need to fix or workaround for fast kernels

geohot:
Good progress, bounty locked!

geohot:
How is this coming? Good to tweet the link?

hooved:
I just got it working in the browser on clang/wasm locally, not hosted yet.

I wouldn't tweet it yet; we'll get the most value if the tweeted link just works when someone taps the link on their phone, without fiddling with settings, etc. Key issues are making it work without webgpu enabled, and making it fit within phone memory - both of which I have traction on now with CLANG and int8 weights. Immediate priorities are getting the clang deployment clean, hosted and testing it on some phones.

hooved:
Here is an updated demo: https://hooved.github.io/tinychat/, which defaults to WebGPU if enabled, otherwise falls back to WASM (CLANG model). You can force usage of the WASM backend with https://hooved.github.io/tinychat/?backend=wasm

I want to communicate that the CLANG llama inference (1B parameters) is slow despite me trying low hanging fruit (BEAM=2 and above, compiler flags, compiler hints). I'm seeking any advice on making the CLANG inference faster. In tinygrad the jitted CLANG inference takes ~0.6 sec/token, and in the browser ~1.2 sec/token; timestamps show practically all such time elapses while WASM is calling the inference kernels (this does not involve shuffling weight buffers (2GB) which sit at fixed addresses in WASM memory from the start). I've read that WASM can be at least 40% slower than native compiled code. Quantizing to int8 doesn't seem to matter much for speed (which is good).

The WASM inference runs in one thread, in a worker. We could make it faster using multiple threads, but not sure how straightforward multithreading is with CLANG kernels from tinygrad. Also we'd need to enable `SharedArrayBuffer` in the browser, which makes hosting more difficult because of security/server header reqs.

My priority now is making tinychat work on my iphone15 on WebGPU, which I have a good chance of doing within a few days. The current blocker is crashing caused by processing 512 MB download chunks, but I've already split the hosted weights into <= 50 MB chunks, and will try to stream load these chunks into GPU buffers.

I'm leaning toward only supporting WebGPU (not WASM) on phones based mainly on the current slowness of CLANG, and also based on the below phone-centric issues that I've discovered through testing on my iphone 15 (iOS 17.6.1):
- I can't allocate a Uint8Array with size >= 640 MB, [test link](https://hooved.github.io/tinychat-test/?test=maxbuf), so the current 1 GB (float32) llama3-1B token embeddings won't fit into a WASM instance's memory, which is a single contiguous ArrayBuffer.
    - This is potentially solvable using float16 or int8 token embeddings, and using multiple WASM modules for inference.
    - 512 MB browser javascript heap allocations still frequently fail, at least on iphone 15
- I can only reliably allocate up to ~1.7 GB of memory (as 8MB Uint8Arrays), [test link](https://hooved.github.io/tinychat-test/?test=browser_memory), reinforcing that we'd need to shrink model size further beyond current 2 GB
- However I can reliably allocate >6 GB of WebGPU buffers on iphone15, [test link](https://hooved.github.io/tinychat-test/?test=gpu_memory), including a 1 GB buffer, which I hope is real. At least on my PC I confirm that the memory is allocated, via nvidia-smi, and OOM errors
- tests were performed in the best case scenario, after closing all apps then opening only a single safari tab

geohot:
Exciting! Looking forward to merging this when you think it's ready

tinygrad/nn/state.py
```diff
@@ -323,4 +323,4 @@ def read_arr():
 
   for name, dims, typ, off in t_infos: state_dict[name] = ggml_data_to_tensor(tensor[data_start + off:], prod(dims), typ).reshape(*reversed(dims))
 
-  return kv_data, state_dict
+  return kv_data, state_dict, t_infos, data_start
```
geohot, line 326: Why are these needed in general?

hooved: `t_infos`/`data_start` tell us the byte layout and dtypes (also shapes and names) of tensors in the gguf file. I used that data to slice out the Q6_K-quantized token embeddings tensor for direct usage in tinychat, without reimplementing that quantization. That said, I think my code in `examples/tinychat/compile.py` could be a lot simpler if all final model weights derive from the tinygrad model's state_dict, which I'll look at before merge - then I wouldn't need `t_infos`/`data_start`.

rcastill:
Hi, I'm trying to replicate. I successfully launch tinychat and it loads and decompresses the model, but then it gets stuck at:

```Error launching model: TypeError: Cannot read properties of undefined (reading 'bytes')```

image

**Comment by hooved** [comment]
tinychat has been loading+running stably on my iphone 15 (iOS 18.3): https://hooved.github.io/tinychat-test2/, 5 tok/sec. The only occasional problem I've seen is that downloading will stall, but refreshing will pick up where it left off due to chunk caching.

Going from iOS17 to iOS 18.3, Apple reduced max WebGPU buffer size from 1+ GB, to 322 MB. This forced me to quantize the tok embeddings from float32 to int8, so they're now 262 MB. This quantization made inference slower, but BEAM=3 is now giving me WebGPU inference code that works in the browser. Altogether, these changes result in an overall 2-3x speedup in tok/sec on my PC, but more modest change on my iphone.

I'm taking another look at CLANG on mobile, because I have to refactor the CLANG deployment anyway (with the new stream loading of weights).

**Comment by hooved** [comment]
WASM is working on iphone now, at 2.5 tok/sec! https://hooved.github.io/tinychat/?backend=wasm
WebGPU: https://hooved.github.io/tinychat

The loading of weights to WASM on iphone is not as stable as WebGPU, and will cause a page crash/reload sometimes, but does eventually finish if I just reload the page. Will see if i can make it better.

**Comment by hooved** [comment]
Remaining:

- [ ] Enable generation to end when swiping back, which currently breaks the app (also the case in the previous tinychat python server app)
- [ ] Handle conversations that reach max context (4096 on WebGPU, 1024 on WASM), which otherwise produce gibberish (also the case in the previous tinychat python server app). Either halt generation, or implement sliding context if simple
- [ ] Show prefill progress bar for slow prefills so user doesn't think it's frozen (i.e. WASM, or long WebGPU prompts)
- [ ] General code clean up, refactor, improvement

File: examples/tinychat/index.js [hunk 2]
```diff
@@ -1,5 +1,484 @@
...
+
+      const response = await fetch(`${window.MODEL_BASE_URL}/net_metadata.json`);
+      // TODO: cache metadata (and everything else, including tokenizer)
+      // TODO: use service worker to reload page when offline
+      const data = await response.json();
+      data.metadata.files = data.metadata.files.map((file, index) => ({...file, index}));
+      const state_dict = data.metadata.state_dict;
+
+      /*
+      - allocating memory to WASM on mobile has longstanding issues: https://github.com/WebAssembly/design/issues/1397
```
hooved, line 403: WASM loading on iPhone 15 frequently crashed Safari (i.e. forced refreshes), before these changes, which seem to have fixed the problem. Both WebGPU and WASM now are stable enough for me, on the devices that I can routinely test (ubuntu PC, iphone 15)

hooved:
To reproduce building and running tinychat in browser locally, see the [README](https://github.com/tinygrad/tinygrad/blob/b4c5c5ecf01a2a6ee312d16fc1a89c9d4a379ffa/examples/tinychat/tinychat-browser/README.md).

hooved:
Closing this PR because the sub-PRs were merged.

PR #8471, opened 2025-01-02
onnx pool ops clean up

geohotstan:
from #8109 

added a 
``` python
# NOTE: just to make sure that the pool pads are symmetric
def assert_symmetric_pads(pads):
  # pads: (padding_left, padding_right, padding_top, padding_bottom, ...)
  if not all(pB == pA for pB, pA in zip(pads[::2], pads[1::2])): raise ValueError("Pads must be symmetric")
  # use symmetric pads
  return pads[::-2]
```

to filter out non-asymmetric pads for now.
easy to remove if we decide to do asymmetric pads.
High confidence of op correctness this time, except this `  # TODO: if there are repeated values, this is wrong`.


chenyuxyz:
we should support asymmetric pad, but this looks fine

PR #8473, opened 2025-01-02
fuse bitcast for non-disk devices

Qazalin:
This diff allows for only supporting subbuffer for DISK tensors and fusing bitcast in normal kernels.
Because breaking a kernel just to do bitcast does not make sense when we can fuse it and load the value directly from registers.


Qazalin:
gated by moving fix_bf16 to rewrite rules.

Qazalin:
maybe we can keep it and just create the subbuffer in the scheduler

PR #8475, opened 2025-01-02
Skip test that counts Ops.LOAD on CLANG+AMX (upcasts up to float16)

uuuvn:
This test assumes that float4 is the max upcast and tests that 8 float loads are upcasted to 2 float4 loads, however on CLANG+AMX upcasts can be up to float16 and in this test we get one float8 load instead.

The @unittest.skipIf line is copied from test_linearizer.py where a bunch of tests make similar assumptions about upcasts.

What this test assumes and tests for:
image

What actually happens on CLANG+AMX:
image

With this (and #8460) all CLANG+AMX tests pass on my laptop, however i expect it to break again quite soon if there isn't a full test suite running in CI. I thought about making a quick MOCKAMX implementation but it requires writeable global state (amx regs) and clang jit doesn't (and probably shouldn't) support it.

### DIFF
--- a/test/test_uop_graph.py
+++ b/test/test_uop_graph.py
@@ -1,7 +1,7 @@
 from typing import List
 import unittest, time
 from tinygrad import dtypes, Device
-from tinygrad.helpers import DEBUG
+from tinygrad.helpers import DEBUG, AMX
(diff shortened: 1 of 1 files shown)


geohot:
Nice PR, well documented and easy to merge.

PR #8476, opened 2025-01-03
fp8 using the e4m3 format for cuda

theblobinthesky:


geohot:
#8490 is a lot further along, and this is failing tests

PR #8478, opened 2025-01-03
Add QLinearConv, QLinearMatMul, QLinearAdd, QLinearGlobalAveragePool to onnx

geohotstan:
required to run `python3 examples/benchmark_onnx.py https://github.com/xamcat/mobcat-samples/raw/refs/heads/master/onnx_runtime/InferencingSample/InferencingSample/mobilenetv2-7-quantized.onnx`
also tested that with the huggingface branch and ORT verify passes
(there is also numerical inaccuracy here too. Gonna try to see if I can dampen it first)

qmatmul and qconv does op in int32
qadd and qavgpool does it in float

https://github.com/apache/tvm/blob/main/python/tvm/relay/frontend/onnx.py#L5373 it seems that a lot of these QLinear ops for ORT are also doing float math.


geohotstan:
FAILED test/test_hcq.py::TestHCQ::test_speed_exec_time - assert 7612.062 <= 7000
This test might be flaky :o

geohot:
Why is this draft? Can I merge it?

PR #8312, opened 2024-12-18
Clang JIT

uuuvn:

PR #8481, opened 2025-01-03
Clang JIT again

uuuvn:
Diff with #8312:

- Mark AMX's wmma function static + comment about why
- R_AARCH64_CALL26 to call wmma function in AMX
- removed unnecessary i2u usage in R_AARCH64_ADR_PREL_PG_HI21 because getbits already sign-extends negative numbers correctly
- ClangCompiler is back because it is used by ops_dsp.py. Not sure whether i should leave it there or just create a separate DSPCompiler.
- using new python match instead of a bunch of ifs

Tested on: macos 15.2 arm64 (+amx), macos 14.7.2 arm64, ubuntu 24.04 arm64, fedora 41 arm64

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,8 @@ jobs:
         python-version: 3.12
     - name: Install docs dependencies (no cache)
       run: pip install -e '.[docs]'
+    - name: Install capstone for CLANG disassembly
+      run: pip install capstone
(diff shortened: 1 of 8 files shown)


geohot:
Pushed to benchmark, will merge if it passes

geohot:
Congrats, bounty is yours! george@tinygrad.org to claim

PR #8485, opened 2025-01-03
Share a _resolve_pool_pads function for pool ops in Tensor

geohotstan:
from #8109 
basically `_padding2d -> _resolve_pool_pads` and use it for all pool functions.
This also acts as an extra layer of protection over pool padding.

There were times where x.avg_pool(..., padding=[0,1,0,1]) was running and accidentally passing tests in onnx when asymmetric pads wasn't even supported. (e.g. [0,1,0,1] was zip-ed as [0,1] in calculations) 

so _resolve_pool_pads must run at the start of a pool op to ensure padding correctness.

`conv_transpose2d` can use this function when we actually support asymmetric padding.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -1852,8 +1852,10 @@ class TestOps(unittest.TestCase):
       helper_test_op([(2,4,9,9), (4,4,3,3)],
         lambda x,w: torch.nn.functional.conv_transpose2d(x,w,padding=padding).relu(),
         lambda x,w: Tensor.conv_transpose2d(x,w,padding=padding).relu(), grad_rtol=1e-5)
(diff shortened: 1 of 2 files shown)


test/test_ops.py
```diff
@@ -1852,6 +1852,8 @@ def test_padded_conv_transpose2d(self):
       helper_test_op([(2,4,9,9), (4,4,3,3)],
         lambda x,w: torch.nn.functional.conv_transpose2d(x,w,padding=padding).relu(),
         lambda x,w: Tensor.conv_transpose2d(x,w,padding=padding).relu(), grad_rtol=1e-5)
+    with self.assertRaises(ValueError): Tensor.conv_transpose2d(Tensor.randn((2,16,2,2)), Tensor.randn((32,16,3,3)), padding=(1,1,1))
```
chenyuxyz, line 1855: why are these not comparing with torch alert? does torch support it?

PR #8486, opened 2025-01-03
LLVM JIT (+autogen llvm instead of llvmlite)

uuuvn:


uuuvn:
(this is been ready for review)

geohot:
Can you do the CPUProgram refactor (and any other prereq refactors) in a separate PR?

tinygrad/runtime/support/llvm.py
```diff
@@ -0,0 +1,23 @@
+import ctypes, ctypes.util, contextlib, os, subprocess
```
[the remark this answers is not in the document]
uuuvn: If there is then i haven't found it. The closest i've found is `llvm-config` which i use on macos, but it isn't available on fedora for some reason:
```
user@localhost:~$ uname -a
Linux localhost.localdomain 6.11.10-300.fc41.aarch64 #1 SMP PREEMPT_DYNAMIC Sat Nov 23 01:15:57 UTC 2024 aarch64 GNU/Linux
user@localhost:~$ llvm-config
-bash: llvm-config: command not found
user@localhost:~$ python3
Python 3.13.0 (main, Oct  8 2024, 00:00:00) [GCC 14.2.1 20240912 (Red Hat 14.2.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ctypes.util
>>> ctypes.util.find_library('LLVM')
'libLLVM.so.19.1'
```

uuuvn:
Seems like there is a non-determinism in clang2py

geohot:
I think this is pretty close to done. I'll test it tomorrow or Sunday on a bunch of machines, if it works I'll merge and the bounty is yours.

tinygrad/runtime/ops_llvm.py
```diff
@@ -1,49 +1,57 @@
...
-    super().__init__(device, MallocAllocator, LLVMRenderer(), LLVMCompiler(self, getenv("LLVMOPT")), functools.partial(LLVMProgram, self))
+    for component in REQUIRED_COMPONENTS:
+      getattr(llvm, f'LLVMInitialize{HOST_ARCH}{component}')()
+
+    triple = f'{platform.machine()}-none-unknown-elf'.encode()
+    target = expect(llvm.LLVMGetTargetFromTriple(triple, ctypes.pointer(tgt:=llvm.LLVMTargetRef()), err:=cerr()), err, tgt)
+    features = b'+reserve-x18' if platform.machine() == 'arm64' else b''
+    target_machine = llvm.LLVMCreateTargetMachine(target, triple, b'', features, llvm.LLVMCodeGenLevelDefault, llvm.LLVMRelocPIC,
+                                                  llvm.LLVMCodeModelDefault)
+    pbo = llvm.LLVMCreatePassBuilderOptions()
```
geohot, line 28: Why is this created in the Device and not the Compiler?

uuuvn: moved to compiler

geohot:
Good on lines now, but looks like the autogen test is failing?

uuuvn:
oh, i see, now i have to make CPUProgram work on windows too...

converting to draft for now

uuuvn:
re autogen fails: i removed one more non-determinism, not sure if that's all of them or if there is more

geohot:
Don't make that stuff two files, it's tons of lines

uuuvn:
Accidentally found a serious problem: LLVM and METAL backends can't be opened in the same process because llvm loaded by MTLCompiler and llvm loaded by my thing collide with each other and mess everything up quite badly (easiest way to repro: `python tinygrad/device.py` on a system with both metal and llvm, will segfault on release builds and assert on debug builds). Not sure why this doesn't happen with llvmlite, they seem to repackage llvm into their own dylib, maybe that somehow makes it go away.

I'll try to find a way to resolve this cleanly but i'll probably have to do the "llvm tries to open metal, if it succeeds then use metal's llvm instead of loading a new one" (only using metal's llvm if it's already loaded will fail if somebody opens metal after llvm)

geohot:
This ready?

uuuvn:
> This ready?

Haven't had time to properly fix the mtlcompiler's llvm and normal llvm in the same process thing (will likely just end up having to package my llvm build with isolated global state as a homebrew tap and raise if normal non-isloated llvm dylib is used in the same process with metal)

If current thing (don't allow llvm and metal backends be opened in the same process at all) is fine then it's ready, otherwise will be ready soon.

uuuvn:
this is ready

geohot:
Congrats, bounty is yours (assuming benchmark tests are fine)! george@tinygrad.org to remind me to pay

PR #8488, opened 2025-01-03
Graph rewrite 2.0

ttomsa:
I wanted to open this pr when it was fully working but I grew impatient, it's 90% there. The key insights are that matching should start at the leaf nodes instead of the root and the use of a [tree automaton](https://en.wikipedia.org/wiki/Tree_automaton).

The old graph rewrite had two key inefficiencies, both `pm.rewrite()` and `ctx.rewrite()` lead the same uop to be traversed multiple times.
Suppose we have the patterns `a(b,d)` `a(b,c)` and want to match target `a(b,c)` (each letter is a uop and tuple is the src). The matching sequence would be `a-b-d-a-b-c`. When d fails the match, the whole sequence starts over.
But `pm.rewrite()` is called by `ctx.rewrite()` which starts rewriting at the leaf nodes so the full rewrite sequence would be `b-c-a-b-d-a-b-c`.

With the new implementation the full sequence would be `b-c-a`. Each uop is traversed exactly once in the whole rewrite routine.

This [video](https://www.youtube.com/watch?v=ANf7FpHoq0w) is a good overview of tree automata. With a leaf to root traversal we don't need to check the number of srcs.
To support repeated srcs and `allow_any_len` the srcs were decoupled so that the first src leads to the second etc. This way we can add transition values for repeats and variable length srcs.

Whether this results in a 2x or 20x speedup remains to be seen, but I think this is the correct way to implement a state machine for this context.

Here's the TODO list:
- [ ] correctly assign names to uops (somewhat tricky cause names aren't tied to states)
- [ ] add support for any src
- [ ] separate `op`, `dtype`, `arg` into their own states, each will be a simple lookup

### DIFF
--- a/test/unit/test_graph_rewrite.py
+++ b/test/unit/test_graph_rewrite.py
@@ -203,7 +203,7 @@ class TestGEPAndVectorizeRewrite(unittest.TestCase):
 
 
 import inspect
-from tinygrad.ops import graph_rewrite, _substitute, track_rewrites, symbolic_simple
+from tinygrad.ops import graph_rewrite, _substitute, symbolic_simple
 
 class TestBottomUpRewrite(unittest.TestCase):
   def test_const_folding(self):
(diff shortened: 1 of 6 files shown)


geohot:
Nice, yea "tree automata" seems like the right field.

geohot:
How is progress on this?

PR #8491, opened 2025-01-04
am: remove alloc frags logic for now

nimlgen:


geohot:
Do we have tests for this stuff in CI?

PR #8492, opened 2025-01-04
Compiles on Windows for python and gpu backends

Pippepot:
Tinygrad would previously complain about missing imports in ops_clang


geohot:
To merge things like this we'd need working Windows CI. It should be possible to add

Pippepot:
I wrote basic CI for testing on windows but some of the tests compile to clang, which does not work. So windows cannot pass CI without putting a lot of work into compiling to clang for windows, which is beyond the scope of my PR. Will therefore close

PR #8495, opened 2025-01-04
schedule Tensors with graph_rewrite_map

Qazalin:
This is a complete rewrite of the scheduler implementing TIP [4](https://github.com/tinygrad/tinygrad/issues/8294) and [5](https://github.com/tinygrad/tinygrad/issues/8294) using graph_rewrite_map.

Moving the scheduler to this graph_rewrite pattern is the prereq infrastructure for adding CAST_BEFORE_VIEW back and improving the backward pass fusion.

Passes test_ops, ASSIGN (without preload or BFS) and const folding tests. The scheduler graph is a lot simpler here, try it!

There are missing optimizations for reduceop+elementwise children fusion. Currently reduce ops are always realized in this diff. This will bring the kernel counts back down.


Qazalin:
why is JIT=2 so bad?
image
is this a graph_rewrite_map issue?

Qazalin:
[reference run](https://github.com/tinygrad/tinygrad/actions/runs/12621506006/job/35168426247) of the scheduler without late fusion optimizations runs without OOM, this is a bug in graph_rewrite_map.
image

Qazalin:
ASSIGN, symbolic var? and image left. Planning to delete Ops.BUFFER_VIEW from the ops layer soon (it should be in realize.py), so no support for it in this diff.

Qazalin:
- [x] symbolic var
- [ ] image
   - [x] "make things that can't be images not images"
   - [ ] correctness issue in test_conv2d
- [ ]  ASSIGN
   - [ ] double diamond
   - [x] single diamond
   - [ ] cycle and permute asserts
- [x] BUFFER_VIEW

Qazalin:
3 categories of failing tests, ordered by priority for tomorrow:


1. When the base parent realizes, all its (unrealized) children views must update too. See `TestRealizeMeansRealize.test_realize_idempotent_prune_views` for details.


2. While simple diamond assigns works, ASSIGN of specific cases (eg. double diamond, crossunder) is not handled properly yet.

3. Realizing all reduceops into kernels is very slow. This should move back to reduceop+elementwise ops children fused kernels.



----

I could add subbuffer without spec changes, to this diff and pass its tests.
However, I don't have a good solution for "make things that can't be images not images". What would it take to delete that hack? It ruins the purity of graph_rewrite_map. A parent tensor's ShapeTracker impacts all its children's dtype and changes the Buffer midway through schedule.

Qazalin:
moving to #8580.

PR #8498, opened 2025-01-04
viz cleanups [pr]

geohot:


Qazalin:
hmm asdict does what...

geohot:
Oh hmm, you are right, it shouldn't serialize. That line was slow though.

PR #8499, opened 2025-01-04
use UOp.st for kernel reduce axes 

Qazalin:
`verify_ast` needs to be rewritten as a pattern matcher. I think test_padto_group is testing a wrong set of Opts.
Expecting [pr] as none of the benchmarks should regress because of this.


Qazalin:
I think this actually improves benchmarks, and it is more correct this way. Merging.

geohot:
Why did test_padto_group break?

PR #8502, opened 2025-01-04
move llvm_bf16_cast to the renderer [pr]

Qazalin:
`def llvm_bf16_cast` can be deleted from tensor.py after this. Is there any reason not to put this in all renderers?
This is currently blocking https://github.com/tinygrad/tinygrad/pull/8473.


geohot:
If the device supports bfloat16 native better to use that, but overall this is super unoptimized and I'm fine with anything that is correct

PR #8501, opened 2025-01-04
HWInterface abstraction in HCQ devices and support for mac os in mockgpu

patrini32:
The vfio code path hasn't been tested.

PR #8505, opened 2025-01-05
MOCKGPU amd test on OSX

patrini32:
Depends on #8501


nimlgen:
I see the tiny one passes! Can you pls try the same set as AMD runs?

nimlgen:
Cool, bounty locked. Let's get it merged cleanly!

Qazalin:
I would also add the "check Device.DEFAULT and print some sources" step to this workflow.
By default tiyngrad will switch devices if the one specified by env vars does not work properly, we want to make sure mac AMD CI always runs on AMD.

.github/workflows/test.yml
```diff
@@ -570,6 +570,101 @@ jobs:
...
 
+  osxtests:
+    strategy:
+      fail-fast: false
+      matrix:
+        backend: [amd]
+
+    name: Tests on MacOS (${{ matrix.backend }})
+    runs-on: macos-15
+    timeout-minutes: 60
```
Qazalin, line 581: 60 seems way too long, do you need all that time to install clang?

Qazalin:
lgtm, thanks!
@nimlgen do we want this?

.github/workflows/test.yml
```diff
@@ -663,7 +663,7 @@ jobs:
         run: pip install --user -e '.[testing]' --extra-index-url https://download.pytorch.org/whl/cpu
       - name: Check Device.DEFAULT and print some source
         run: |
-          PYTHONPATH=${{ github.workspace }} python3 -c "from tinygrad import Device; assert Device.DEFAULT == 'AMD', Device.DEFAULT"
+          PYTHONPATH=${{ github.workspace }} python3 -c "from tinygrad import Device; assert Device.DEFAULT in ['AMD', 'NV'] Device.DEFAULT"
```
nimlgen: Let's keep only AMD for now.

And since we want to run NV on the same runner later, so can you limit pytest to test/test_hcq.py and test/test_tiny.py only.

And it's good to go imho

nimlgen:
Congrats!

fyi @geohot

geohot:
Testing now. Getting `RuntimeError: library amd_comgr not found`

PR #8517, opened 2025-01-06
Get `MOCKGPU=1 AMD=1 python3 test/test_tiny.py` working in CI on OS X…

dyKiU:
… (build comgr + remu)

build and cache llvm from amd-staging branch - see commit hash for current version
build hipcc and amd device libs
build comgr with amd-staging llvm (wouldn't compile or link with vanilla llvm)
build remu (rust)
provide both lib (libamdcomgr.dylib and libremu.dylib) to run 

MOCKGPU=1 AMD=1 python3 test/test_tiny.py

the tests run but fail because there's no /dev/kfd on macOS

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,10 +10,305 @@ on:
   push:
     branches:
       - master
+      - ci-amd
   pull_request:
   workflow_dispatch:
 
+
 jobs:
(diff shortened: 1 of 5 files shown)


nimlgen:
Similar thing is merged in another PR. Thanks

PR #8525, opened 2025-01-07
minor arg spec check on wmma [pr]

ignaciosica:
Thinking in adding `elements_per_threads` as arg to wmma to simplify `ops_python`, George said he was okay in retrieving simple attributes from tc class like `dimensions`, `threads` and `elements_per_threads`

Is there any value in spec checking every wmma arg, or is it enough with checking the length?
```python
# current spec
wmma_arg = (name, dims, dtype_in, dtype_out, device, threads, upcast_axes, reduce_axes)
```


geohot:
Oh I'd totally merge something that documents and checks each one there, ideally first making it simpler.

PR #8527, opened 2025-01-07
Add 'attend' to Embedding class

Devan-Kerman:
Small LLMs like LLAMA 3.2-1/3B sometimes use the same embeddings for input and output token embeddings. This produces the need to be able to project the output of the last model using the embedding weights. I added an `attend` method to `Embedding` for this.

This is relatively trivial to do without a helper method so the utility/LOC is questionable, but I don't know how strict that is and will leave it up to the maintainers.

```python
if Config.shared_embeds:
	self.unembed = lambda x: x @ self.embed.weight.T
else:
	self.unembed = nn.Linear(Config.model_dim, Config.vocab_size)
```

### DIFF
--- a/tinygrad/nn/__init__.py
+++ b/tinygrad/nn/__init__.py
@@ -325,6 +325,12 @@ class Embedding:
     arange, idx, vals = self.arange.expand(big_shp), idx.reshape(idx.shape+(1, 1)).expand(big_shp), self.weight.expand(big_shp)
     return (arange == idx).mul(vals).sum(-2, acc_dtype=vals.dtype)
 
+  def attend(self, x:Tensor) -> Tensor:
+    """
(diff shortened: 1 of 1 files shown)


chenyuxyz:
how's this done in torch? for nn class we usually match torch or other well known api

Devan-Kerman:
Not sure, I'm more familiar with JAX. Though, I'm not sure how else this would be implemented

geohot:
Going to close, I don't see this in JAX either.

https://flax.readthedocs.io/en/v0.5.3/_autosummary/flax.linen.Embed.html

Devan-Kerman:
https://flax.readthedocs.io/en/latest/_modules/flax/nnx/nn/linear.html#Embed.attend

Linen was replaced with NNX

PR #8528, opened 2025-01-07
comma Benchmark: add clang

YassineYousfi:


wozeparrot:
If this is for benchmarking openpilot model on cpu, it probably shouldn't be run on a comma, unless that is what you were going for.

YassineYousfi:
Yes, just on a github CI runner would be fine. 
Moving to test.yml:testopencl

geohot:
Are you trying to run this in "benchmark" (hosted at comma dc) or "CI" (github actions)?

YassineYousfi:
CI (github actions) is good enough

geohot:
It should be in test.yml and not benchmark.yml then. I don't see it running

YassineYousfi:
it ran here: https://github.com/tinygrad/tinygrad/actions/runs/12663519014/job/35290224470

YassineYousfi:
that's because the test only runs on these branches 
```
- master
- update_benchmark
- update_benchmark_staging
```

geohot:
re: the relationship between tiny corp and comma. for now there isn't one, so all comma specific development is postponed until their is.

Issue #8366, opened 2024-12-21
TIP 6 - shape and device of Tensors created from a BIND UOp


Qazalin:
## Abstract

Tensors created from bound Variables need a ShapeTracker and a Device. This document proposes a spec change to accommodate this.

## Motivation

BIND should contain its full meaning in parents. Currently, it relies on a VIEW and BUFFER child to get its ShapeTracker and device.

## Current state

A Tensor created from a bound Variable has a `VIEW(DEVICE, <BIND>)` UOp.

```
vv = UOp.variable("a", 0, 10).bind(1)
a = Tensor(vv)
print(a.lazydata)

UOp(Ops.VIEW, dtypes.int, arg=ShapeTracker(views=(View(shape=(), strides=(), offset=0, mask=None, contiguous=True),)), src=(
  UOp(Ops.DEVICE, dtypes.void, arg=GPU, src=()),
  UOp(Ops.BIND, dtypes.int, arg=None, src=(
    UOp(Ops.DEFINE_VAR, dtypes.int, arg=('a', 0, 10), src=()),
    UOp(Ops.CONST, dtypes.int, arg=1, src=()),)),))
```

## Proposal

A Tensor Variable becomes a register in the final kernel, eg. RDNA3 allocates an SGPR for it:

```
E_(int* data0, const int a) {
  *(data0+0) = (a+3);
}

s_load_b32 s2, s[0:1], 0x8
s_add_i32 s2, s2, 3
```

Thus Ops.DEVICE can be placed as a source to Ops.DEFINE_VAR,
Scalars always have shape () and stride ().

```
var = UOp(Ops.DEFINE_VAR, dtypes.int, arg=('a', 0, 10), src=(UOp(Ops.DEVICE, arg="GPU"),))
bind = var.bind(1)
print(bind)
```

would naturally work because `def bind` uses `variable.const_like` to derive device and shape.

```
UOp(Ops.BIND, dtypes.int, arg=None, src=(
  UOp(Ops.DEFINE_VAR, dtypes.int, arg=('a', 0, 10), src=(
    x1:=UOp(Ops.DEVICE, dtypes.void, arg=GPU, src=()),)),
  UOp(Ops.CONST, dtypes.int, arg=1, src=(
    UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(), strides=(), offset=0, mask=None, contiguous=True),)), src=(
       x1,)),)),))
```

## Alternatives

[...]

PR #8529, opened 2025-01-08
add VIEW(DEVICE) to tensor variable [pr]

Qazalin:
This completes the proposed spec change from TIP 6: #8366.

Symbolic add AST diff: (a+3 from `TestTensorVariable.test_add_tvar`)

image

Now BIND contains its full meaning locally: device, shape, var, val.


Qazalin:
spending one extra line, but I think this is the more natural representation of BIND in tensors.

PR #8532, opened 2025-01-08
am in ci

nimlgen:
Not going to merge this now, just to check


nimlgen:
@chenyuxyz, we fine with merging this (half with AM, half with AMD)? GPU time is better, but AM is slower when copying from drive now (I think due to missing huge pages).

chenyuxyz:
i think it's fine, 2 things to consider (1) overall ci speed, and (2) state management. so consider adding `sudo modprobe amdgpu` at the top as part of setup, or any other way to recover the state in case the script crashed in between

chenyuxyz:
also need to add to the other red ci script, there's no guarantee that the same machine gets the same job

PR #8538, opened 2025-01-08
Small bug in _reshape_mask

eliotgolding:
Masks that are everywhere-invalid can always be unfolded but the current condition misses that


chenyuxyz:
cool!

Issue #7237, opened 2024-10-23
canonicalize shapetracker


chenyuxyz:
can we get this to 0? might be related to `simplify_merge_adjacent`, but conditional on the last view

`SEED=42 CNT=1000 CHECK_NEQ=1 FUZZ=plus PYTHONPATH="." python3 test/external/fuzz_shapetracker_math.py`

`same but unequal 59/1000 = 5.90%`

chenyuxyz:
deprecating shapetracker

PR #8560, opened 2025-01-10
use unravel in views_to_indexed_uops [pr]

eliotgolding:
- Use `unravel` in `views_to_indexed_uops`
- Replace `unravel` implementation with the one from `views_to_indexed_uops`; this keeps the uop graph returned by that function unchanged. Also adds new symbolic rules for when `unravel` is used in `View.__add__`.
- ~Remove `View.minify` from views_to_indexed_uops and call it whenever the number of views in a shapetracker increases. This works towards #7237 (since minify is provably complete for canonicalizing the action of a single view within a shapetracker).~ see below


eliotgolding:
The minify thing actually increased the "same but unequal" from the issue above; I think this might be because it merges dims in the right-hand shapetracker, and the current view add requires that the right view's shape is preserved, when this is not required for adding in the middle of a shapetracker. So less views get added.

This is just the unravel change now.

chenyuxyz:
this looks more complicated, what's the benefit?

eliotgolding:
Saves lines and conceptually clearer in views_to_indexed_uops. (why is the same function written there with a different implementation?) You lose the lines again in ops.py but then that's two new symbolic rules for free.

chenyuxyz:
not convinced this is simpler, new rules with no new tests are hard to maintain. I think it's fine if the two needed rules are part of `div_and_mod_folding` with some tests for the updated behavior

eliotgolding:
Alright that's done

chenyuxyz:
cool this is much better

PR #8562, opened 2025-01-11
add SINK folding from the tensor_map branch [pr]

Qazalin:


Qazalin:
that extra `is VIEW` diff shouldn't exist, #8572 fixes it, this needs a spec.

PR #8569, opened 2025-01-11
add test_onnx_ops.py

geohotstan:
ok I wasted so much time trying to adapt and fix the tests in https://github.com/apache/tvm/blob/main/tests/python/frontend/onnx/test_forward.py

I think it's better we just write tests from the ground up.
For the tests in `test/test_onnx_ops.py`, we just find the op here
```
# https://github.com/onnx/onnx/blob/main/docs/Operators.md
# https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md
```
and use the docs to write the `inputs`, `attributes`, and expected `outputs`  according to the spec described.

We can also use this to write regression tests when we hit op errors when running onnx.
We can also add CalibrationDataReader from onnxruntime.quantize and automatically add quantizelinear, dequantizelinear for ops to test later.
We can also add tests for multiple ops. Just gotta define the model.

Posting this early to see if this is a good structure to go about this.

`test_qlinear_matmul`, `test_qlinear_add`, and `test_qlinear_global_average_pool` was chatgpted 🫣


test/test_onnx_ops.py
```diff
@@ -0,0 +1,130 @@
...
+    helper_test_single_op("Reshape", inputs, attributes, outputs)
+
+class TestOnnxQuantizedOps(unittest.TestCase):
+  @unittest.skip("TODO: Max absolute difference: 128")
+  def test_qlinear_conv(self):
+    # https://github.com/xamcat/mobcat-samples/raw/refs/heads/master/onnx_runtime/InferencingSample/InferencingSample/mobilenetv2-7-quantized.onnx
+    # first qlinear_conv from mobilnet but with x, w, and b randomized
+    inputs = {
+      "x": np.random.randint(0, 256, [1, 3, 224, 224]).astype(np.uint8),
+      "x_scale": np.array(0.01865844801068306, dtype=np.float32),
```
chenyuxyz, line 72: why is it sometimes `astype` and sometimes `dtype` in array arg?

chenyuxyz (commented):
looks fine, one minor comment

test/test_onnx_ops.py
geohotstan: oooo, that doesn't matter.
I was confused why 
`onnx_outputs = [helper.make_tensor_value_info(name, helper.np_dtype_to_tensor_dtype(np.dtype(dtype)), shape) for name, (shape, dtype) in outs.items()]`
from `helper_test_single_op` was raising error before

chatgpt said some explicit dtype define instead of `astype` fixes it. (which didn't fix it)
true fix was I was missing the `np.dtype` wrapper for `np.dtype(dtype)`

extra/onnx_ops.py
```diff
@@ -424,27 +424,25 @@ def DequantizeLinear(x:Tensor, x_scale:Tensor, x_zero_point:Tensor|int=0, axis:i
...
   return ((x.int() - x_zero_point) * x_scale).cast(x_scale.dtype)
 
-def _quantize_linear(y:Tensor, y_scale:Tensor, y_zero_point:Tensor):
-  assert y_scale.dtype is dtypes.float32 and y_zero_point.dtype in {dtypes.uint8, dtypes.int8}, "used only for qlinear ops"
-  y = (y / y_scale + y_zero_point).round()
-  return y.clamp(dtypes.min(y_zero_point.dtype), dtypes.max(y_zero_point.dtype)).cast(y_zero_point.dtype)
+def _quantize_linear(t:Tensor, scale:Tensor, zero_point:Tensor):
+  assert scale.dtype is dtypes.float32 and zero_point.dtype in {dtypes.uint8, dtypes.int8}, "used only for qlinear ops"
+  t = (t / scale + zero_point).round()
+  return t.clamp(dtypes.min(zero_point.dtype), dtypes.max(zero_point.dtype)).cast(zero_point.dtype)
```
geohotstan, line 430: I have a better version of this with proper factoring of the ideas.
I'll post the PR tmrw.

test/test_onnx_ops.py
```diff
@@ -0,0 +1,154 @@
...
+  def test_reshape(self):
+    inputs = {"in": np.arange(6, dtype=np.float32), "shape": np.array([2,3], dtype=np.int64)}
+    attributes = {}
+    outputs = {"out": ([2,3], np.float32)}
+    helper_test_single_op("Reshape", inputs, attributes, outputs)
+
+class TestOnnxQuantizedOps(unittest.TestCase):
+  def test_qlinear_conv(self):
+    for dtype, zero_point in [(np.uint8, 128), (np.int8, 0)]:
+      for b in (np.ones([32], dtype=np.int32), np.zeros([32], dtype=np.int32)):
```
geohotstan, line 56: I made both this and the C in qgemm minimal enough just to test correctness of it's interaction with qlinear stuffs.
When it's np.zeros, it should be testing the `C = None` optional case, but there's some value_info annoyances that I need to figure out. Maybe for now this is ok?

extra/onnx_helpers.py
```diff
@@ -26,8 +25,10 @@ def validate(onnx_file, inputs, rtol=1e-5, atol=1e-5):
   out_values = ort_sess.run(out_names, np_inputs)
   ort_out = dict(zip(out_names, out_values))
 
-  assert len(tinygrad_out) == len(ort_out) and tinygrad_out.keys() == ort_out.keys()
+  tinygrad_out = run_onnx(inputs)
```
geohotstan, line 28: moved this to the bottom to allow ORT to raise error first. If OnnxRunner raises error first, we don't know if that's because it's OnnxRunner's bug or if it's an invalid ONNX model.

PR #8570, opened 2025-01-11
onnx bug fixes from adapting tvm tests

geohotstan:
I still didn't get all the tests in tvm onnx tests to run (tbh tvm didn't either), but they mostly all passed.
These were the extra fixes that made the tests that ran pass.
There are a few more failures for tvm tests that I didn't fix yet, but I'll add them with regression tests in test_onnx_ops.py if that gets merged.


chenyuxyz:
cool i like adding tests from other legit repo and using it to improve ours!

PR #8571, opened 2025-01-11
FSDP in Tinygrad [WIP]

KhanerX:
FSDP semantics can be fully captured by:

- [x] Sharding model and optimizer parameters across GPUS (already supported by multi.py) 

- [x] Sharding gradients along the same axis as parameters
  - This is done by adding a differentiable `All_Gather` function, that `Scatter`s the gradient in backward, and applying it to all sharded model parameters
  
- [x] After each `All_Gather` and forward pass, free the gathered parameters, `All_gather` again before backward.
  - This is the current challenge, and it can be done using `Activation Checkpointing`, which I will try to implement as a prerequisite here.

See paper [[Simple FSDP](https://arxiv.org/pdf/2411.00284)]

**Approach:** Write a `class Checkpoint(Function)` that takes a function, calculates it in forward, remove the intermediate values, and re-calculated that function on the same input with the same RNG state (for things like dropout) as before, in the backwards, removing the need to store output values.

There are a lot of further optimization opportunities, mostly trying to overlay comms with computation as much as possible, I will also look into those further.

First commit demonstrates sharding and successfully training MNIST on multi GPUs, but currently the memory isn't freed after each forward.


geohot:
Does there have to be wrapper classes for everything? There must be a better way to do this

KhanerX:
Followup with a better approach here: [#8669](https://github.com/tinygrad/tinygrad/pull/8669)
Closing this one

Issue #8546, opened 2025-01-09
845 DSP MobileNetV2 < 13.4 ms


geohot:
First, it has to work in int8.

https://github.com/MTlab/onnx2caffe/blob/master/model/MobileNetV2.onnx is the model I tested with 845's SNPE.

https://huggingface.co/qualcomm/MobileNet-v2-Quantized/tree/main wouldn't compile.

geohot:
Running one QLinearConv `VIZ=1 ONNXLIMIT=1 DEBUGONNX=1 DEBUG=2 python3 examples/benchmark_onnx.py https://github.com/xamcat/mobcat-samples/raw/refs/heads/master/onnx_runtime/InferencingSample/InferencingSample/mobilenetv2-7-quantized.onnx`

And we hit the a scheduler issue:
```
w = w.int() - w_zero_point
y = Conv(x, w, B, auto_pad, dilations, group, kernel_shape, pads, strides)
```
w is uint8, but it's being expanded in the conv (after the subtraction of w_zero_point). That expand is triggering `w.int() - w_zero_point` to be realized

geohot:
TODO: fix `python3 examples/test_onnx_imagenet.py https://github.com/xamcat/mobcat-samples/raw/refs/heads/master/onnx_runtime/InferencingSample/InferencingSample/mobilenetv2-7-quantized.onnx`

geohotstan:
> TODO: fix `python3 examples/test_onnx_imagenet.py https://github.com/xamcat/mobcat-samples/raw/refs/heads/master/onnx_runtime/InferencingSample/InferencingSample/mobilenetv2-7-quantized.onnx`

the problem was the out type was uint8 and casting from float to uint8 currently wraps around.
clamp before final cast for conv gets it to around 67%

geohotstan:
> TODO: fix `python3 examples/test_onnx_imagenet.py https://github.com/xamcat/mobcat-samples/raw/refs/heads/master/onnx_runtime/InferencingSample/InferencingSample/mobilenetv2-7-quantized.onnx`

The problem was after the `Conv` we applied the scale back in `QLinearConv` which made it a `float`, then we were casting from `float` to `uint8` which we implicitly wrap around.
Clamping before final cast (no wrap around) gets it to 67%~

I'll do more testing around these QLinearOps to make sure it's correct 😅

PR #8574, opened 2025-01-12
fix onnx mobilenetv2-7-quantized.onnx

geohotstan:
#8546
is 67% considered good?

also a few notes:
 - this current implementation matches onnxruntime but fails onnx external tests 🤔 
 - we could probably rewrite this `float.clamp(uint_min, uint_max).cast(uint)` on some GPU backends to just `float.cast(uint)` since from what I remember, GPU backends just clamps the value so we can add that rewrite for backends with confirmed clamping behavior. (I struggled with this but current `Tensor.maximum` behavior for float is... oh my.)


geohot:
67% is good, yea

extra/onnx_ops.py
```diff
@@ -431,16 +431,15 @@ def QLinearConv(x:Tensor, x_scale:Tensor, x_zero_point:Tensor|int, w:Tensor, w_s
   w = w.int() - w_zero_point
   y = Conv(x, w, B, auto_pad, dilations, group, kernel_shape, pads, strides)
   y = ((y * (x_scale * w_scale / y_scale)) + y_zero_point).round()
-  return y.cast(y_zero_point.dtype)
+  return y.clamp(dtypes.min(y_zero_point.dtype), dtypes.max(y_zero_point.dtype)).cast(y_zero_point.dtype)
```
geohot, line 434: This stuff really should be abstracted and not copy pasted

geohotstan:
also added qgemm so that 

`QUANT=1 PYTHONPATH=.  python3 examples/test_onnx_imagenet.py https://github.com/xamcat/mobcat-samples/raw/refs/heads/master/onnx_runtime/InferencingSample/InferencingSample/mobilenetv2-7.onnx`
with 
`QuantFormat.QDQ` 
runs with around 67% ~. (not sure if we need this lol, but added it anyways. At least now gemm runs in int)

tested qgemm with #8569

geohot:
I merged this, but the code could be cleaner. Why is it called y_scale in the function? Why are you assigning to y_scale in the functions instead of calling the scale function with the math?

PR #8580, opened 2025-01-12
simplify tensors before scheduling [pr]

Qazalin:
In master, scheduling goes through the following steps:
1. Recurse `big_sink` and give all UOps a BUFFER UOp.
2. Perform symbolic folding on the graph with all those BUFFERs
3. merge BUFFER UOps that simplified to the same UOp
4. Break the SINK into list[ScheduleItem]

This diff simplifies the flow by:

1. Perform symbolic folding directly on the `big_sink`
2. Give the remaining UOps BUFFERs
3. Break the SINK into list[ScheduleItem]


tinygrad/ops.py
```diff
@@ -469,21 +457,14 @@ def can_view(self):
...
   def base(self) -> UOp:
     if self.op in GroupOp.Movement: return self.src[0].base
-    return self.src[0] if self.op is Ops.VIEW and len(self.src) == 1 and self.src[0].op is not Ops.BUFFER else self
-  def view(self, new_st:ShapeTracker) -> UOp:
-    if self.st is None: return UOp(Ops.VIEW, self.dtype.base if not isinstance(self.dtype, ImageDType) else self.dtype, (self,), new_st)
-    ret = UOp(Ops.VIEW, self.dtype, (self.base,), new_st)
-    # instant folding rules
-    if new_st.contiguous and self.base.shape == new_st.shape: return self.base
-    return ret
+    return self.src[0] if self.op is Ops.VIEW and len(self.src) == 1 else self
```
geohot, line 466: self.src[0].base?

Also, can this go in w/o the rest of the PR?

Qazalin: hmm no, that's a fundamental difference between GroupOp.Movmenet and VIEW.

VIEW always applies to base. So it [asserts it resolves ](https://github.com/tinygrad/tinygrad/pull/8580/commits/faa07356ea377fee8b4837a1bbf39a3076aa9626#diff-5736814df0b2c07a42d5c1bfef6b5aa5635ce18043f4aa030e21768694259ca3R467) at src[0],
GroupOp.Movement can recursively search.

Qazalin:
figuring out threefry's contiguous pattern is the key to getting this merged.

before:
image


after:
image

now, how do we treat these tensors?

Qazalin:
prereqs:

- [ ] Fix COPY(CONST) destination device
- [ ] Upstream CONTIGUOUS folding on VIEW(BUFFER)

Then, figuring out why threefry requires that contiguous... 
https://github.com/tinygrad/tinygrad/pull/6869/files#diff-7703e7b63c9844f019ae3609629fd9c9bab290d11ae6608ed2d09edd49bab2dbR491
> also fixes weird jit behaviour with the assign.

Like that diff does not look like the real fix, because the test breaks when forced_realize is removed.

test/test_image_dtype.py
```diff
@@ -121,7 +121,7 @@ def test_lil_model(self):
       loss = x.image_dot(w1).image_dot(w2).float().max()
       loss.backward()
       sched = unwrap(w1.grad).schedule()
-      self.assertEqual(len(sched), 9)
+      self.assertEqual(len(sched), 10)
```
Qazalin, line 124: ah this was wrong before, of course that CAST should not fold with the image cast
image
One is a float tensor, the other is an ImageDType Buffer.

Qazalin:
my bad, test_complex_backward was failing because the order of graph_rewrite_map is reverse of add_buffers. So it expected the _last_ tensor's metadata.
We should really fix this by making metadata include all the tensors that mapped to that same buffer).

PR #8597, opened 2025-01-13
CUDA Turing TC

ignaciosica:
Matrix multiplication can be understood as a series of dot products between rows of the first matrix and columns of the second matrix. For each element in the resulting matrix, the corresponding row vector from the first matrix and the column vector from the second matrix are paired, and their dot product is computed. This involves multiplying each pair of corresponding elements from the two vectors and summing these products into a single scalar value. This scalar becomes the entry in the output matrix at the position defined by the row and column involved in the dot product. Repeating this process for all row-column combinations produces the full result of the matrix multiplication.

<p align="center">

</p>

The way python tc emulation works is, instead of giving the input matrix A and B for calculating C, the inputs are matrices that represent the threads collaborating to perform the matmul operation and the elements each thread is responsible of loading and storing. Following, a diagram of the input matrices for 32 collaborating threads for turing tensor cores, where each thread is responsible for loading 4 elements from matrix A, 2 elements from matrix B, and storing 4 elements from matrix C. 

<p align="center">

</p>

Now, the tricky part, is construction the rows and columns required for the matmul from this input matrices. As we already know, each tensor core has a specific thread pattern that specifies which element is responsible for loading and storing each thread from the input and output matrices. This thread pattern is encoded in the swizzle of the tensor core class. Our goal here is to "reverse" the swizzle in a way in order to perform the matmul and check if the thread pattern was performed correctly in the first place. An important observation here is that our "reverse" swizzle must come from the tensor cores documentation directly and has nothing to do with how the swizzle is performed in tinygrad implementation. In this way, this python tensor cores emulation works as a test for real tensor cores. From the documentation we get the tensor core shape and thread pattern (left). A more detailed view from the thread pattern is presented also (right).

<p align="center">

</p>

Given thread index and element from matrix C (c_map) we get the row and column from which to perform the dot product in order to calculate that element. That row and col number is what we are going to use to know which elements retrieve from the input matrices. Starting from k and the row number, we have to calculate the thread indices and elements for matrix A. Starting from k and the col number, we have to calculate the thread indices and elements for matrix B.

Keep in mind our objective. We have to construct the row from the input matrix A from the real input we have, which are the threads and elements from each thread. Going back to the image (left) we have to, given k and row number, retrieve `(a0t4,a1t4,a0t5,a1t5,a0t6,a1t6,a0t7,a1t7)` from the input matrix. That is our goal.

The easiest way to calculate this indices is to think them as a set of rules that construct the final output. As an example, I'm going to make a walkthrough in defining this rules for the A matrix. B is left as an exercise for the reader (just kidding, I'm planning to include it for both matrices for the more formal *docs* version of this).

First, calculating thread element number:
- if the row is < 8, then only the first two elements (a0, a1) are retrieved, otherwise, the last two (a2, a3). The rule that can be derived from this that can help as get a baseline for element index is: `2 * (row // 8)`. From this, we will start fetching either elements from index 0 or index 2 depending on row number.
- if k is uneven, we get element 0, otherwise, we get element 1 from the set of available elements. The rule, in this case, is simple: `k % 2`

Adding this both "rules" together we get the formula for the thread element index for matrix A: `a[2 * (row // 8) + k % 2]`
For calculating the thread index, the process is similar.

- Each row takes 4 threads to form, so for each new row, we need to add 4 to the thread index. Rules derived from this: `(row % 8) * 4`. Note a special mod 8 added to the rule. This is because, for this specific thread pattern, each thread is responsible from loading two elements 8 row apart. Thread 0 loads elements from row 0 and 8. This mod 8 added to the rule contemplates this.
- Each thread loads two contiguous elements, so the thread index is updated as follow: `k // 2`
- We need to add `goff`to contemplate indexing when the matrices are larger than the tensor core size itself.

Adding all three rules together we get the thread index given row and k: `[goff + (row % 8) * 4 + k // 2]`

Resulting in the following:

``` python
def a_elem(a, k, row, goff): return a[2 * (row // 8) + k % 2][goff + (row % 8) * 4 + k // 2]
```

PS: I know this isn't the simplest explanation to follow, I'm still thinking how to make the guide simpler and cleaner.

Indexing for C/D and B:

``` python
# (col, row) given (lane, elem) for C & D (4 elements on 32 threads); 
# shared by all tc shapes with M=16 N=8
def c_map(lane, elem): return ((lane % 4) * 2 + elem % 2, lane // 4 + (elem // 2) * 8)
```

``` python
def b_elem(b, col, k, goff): return b[k % 2][goff + col * 4 + k // 2]
```


ignaciosica:
Added the following job to ci in order to test python tc emulation that involves goff:

`DEBUG=2 EMULATE_CUDA_SM75=1 FORWARD_ONLY=1 PYTHON=1 python3 test/test_ops.py TestOps.test_gemm_fp16`

Otherwise, in all tests goff will be 0 as the tensor cores tests perform a matmul of the same size as the tested tensor core.

ignaciosica:
I think it's much more simpler to reason over ops_python emulation using 'k' and 'row/col' rather than 'i' and 'j'. I can update all the other backend as well if preferred, or change this one to use 'i' and 'j'

flammit:
> I think it's much more simple to reason over ops_python emulation using `k` and `row/col` rather than `i` and `j`. To make it consistent, if preferred, I can update all the other backend as well, or change this one to use `i` and `j` otherwise.

Sorry that was my misstep and I agree it's definitely better not to use `i` or `j`.

ignaciosica:
Great! Didn't knew it was you who made the ops_python emulation. What do you think about my walkthrough? Does it make it easier to follow what is going on? I would appreciate feedback as I'm yet far from feeling comfortable from adding this to the official docs.

ignaciosica:
I'll make a standalone pr renaming to `k` and `row/col`

ignaciosica:
**generated ast and kernel**
``` python
TENSOR CORES [(1, 8, 1)] [(0, 16, 8)] WMMA_8_16_8_half_float
r_2_2_2_2_2_2_2_2_2_2
    UOp(Ops.DEFINE_GLOBAL, dtypes.half.ptr(128), arg=0, src=()),
    UOp(Ops.CAST, dtypes.half, arg=None, src=(
        UOp(Ops.WMMA, dtypes.float.vec(4), arg=('WMMA_8_16_8_half_float', (8, 16, 8), dtypes.half, dtypes.float, 'NV', 32, (((9, 2), (8, 2)), ((9, 2),), ((9, 2), (8, 2))), (5, 6, 7)), src=(
          UOp(Ops.CONTRACT, dtypes.half.vec(4), arg=((9, 2), (8, 2)), src=(
            UOp(Ops.LOAD, dtypes.half, arg=None, src=(
              UOp(Ops.DEFINE_GLOBAL, dtypes.half.ptr(128), arg=1, src=()),
              UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(2, 4, 8, 16, 32, 0, 0, 0, 1, 64), offset=0, mask=None, contiguous=False),)), src=()),)),)),
          UOp(Ops.CONTRACT, dtypes.half.vec(2), arg=((9, 2),), src=(
            UOp(Ops.LOAD, dtypes.half, arg=None, src=(
              UOp(Ops.DEFINE_GLOBAL, dtypes.half.ptr(64), arg=2, src=()),
              UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2), strides=(16, 32, 1, 2, 4, 0, 0, 0, 0, 8), offset=0, mask=None, contiguous=False),)), src=()),)),)),
          UOp(Ops.CONST, dtypes.float.vec(4), arg=0.0, src=()),)),)),)),)),))
```
```c
#include <cuda_fp16.h>
struct __align__(8) half4 { half x, y, z, w; }; __device__ half4 make_half4(half x, half y, half z, half w) { half4 r={x, y, z, w}; return r; }
__device__ float4 __WMMA_8_16_8_half_float(half4 a, half2 b, float4 c){
  asm("mma.sync.aligned.m16n8k8.row.col.f32.f16.f16.f32"
      "{%0, %1, %2, %3}, {%4, %5},"
      "{%6}, {%0, %1, %2, %3};"
    : "r"(a_pk[0]), "r"(a_pk[1]), "r"(b_pk[0]));
  return c;
}
extern "C" __global__ void __launch_bounds__(32) r_2_2_2_2_2_2_2_2_2_2(half* data0, half* data1, half* data2) {
  int alu0 = (lidx0>>2);
  int alu1 = (lidx0&1);
  int alu2 = ((lidx0>>1)&1);
  int alu3 = ((alu1<<1)+(alu2<<2)+(alu0<<3)+(lidx1<<4)+(lidx2<<5));
  half2 val0 = *((half2*)((data1+alu3)));
  int alu4 = (alu3+64);
  half2 val1 = *((half2*)((data1+alu4)));
  int alu5 = ((alu1<<4)+(alu2<<5)+alu0+(lidx1<<1)+(lidx2<<2));
  half val2 = *(data2+alu5);
  half val3 = *(data2+(alu5+8));
  float4 wmma0 = __WMMA_8_16_8_half_float(make_half4(val0.x,val0.y,val1.x,val1.y), make_half2(val2,val3), make_float4(0.0f,0.0f,0.0f,0.0f));
  *((half2*)((data0+alu4))) = make_half2(((half)(wmma0.z)),((half)(wmma0.w)));
  *((half2*)((data0+alu3))) = make_half2(((half)(wmma0.x)),((half)(wmma0.y)));
}
```

tinygrad/renderer/ptx.py
```diff
@@ -124,11 +124,12 @@ class PTXRenderer(Renderer):
...
   suffix = "PTX"
   global_max, local_max, shared_max = CUDARenderer.global_max, CUDARenderer.local_max, CUDARenderer.shared_max
-  tensor_cores = [tc for tc in CUDARenderer.tensor_cores if tc.dtype_in == dtypes.half]
+  tc_sm80 = [tc for tc in CUDARenderer.tc_sm80 if tc.dtype_in == dtypes.half]
   code_for_op = asm_for_op
   extra_matcher = ptx_matcher
   def __init__(self, arch:str, device="CUDA"):
-    self.device, self.tensor_cores, self.arch = device, PTXRenderer.tensor_cores if int(arch[3:]) >= 80 else [], arch
+    self.device, self.arch = device, arch
+    self.tensor_cores = PTXRenderer.tc_sm80 if int(arch[3:]) >= 80 else CUDARenderer.tc_sm75 if int(arch[3:]) >= 75 else []
```
geohot, line 132: This should not be copy pasted

geohot:
Congrats, this bounty is yours! $300 paid via PayPal

ignaciosica:
Awesome! Received! Thanks

PR #8601, opened 2025-01-14
remove unused src from metaop [pr]

chenyuxyz:


Qazalin:
the whole metaop syntax needs to be updated. I only kept it because tensor.py uses it.
Same goes for Ops.EMTPY.

PR #8611, opened 2025-01-14
image fixup tensor map

geohot:


Qazalin:
removing PR since this enables better const folding on Image kernels, here is an example process_replay diff:
image

That CAST used to keep existing because image fixup was done after `symbolic_simple`.
This diff is a prereq for #9165.

PR #8615, opened 2025-01-14
delete forced_realize

geohot:


Qazalin:
something weird is happening with THREEFRY I don't yet fully understand.

Otherwise the openpilot kernel count is fixed.

Qazalin:
most benchmarks are unchanged. "Llama 3 4 GPUs" is slower.
image
This is gating simple_tensor_map. It has to get upstreamed first.

test/test_const_folding.py
```diff
@@ -220,7 +220,9 @@ def test_multi_const_folding_literal(self):
     t = Tensor.arange(16).float().realize().to(ds)
 
     # non const folding case creates one ast on each shard
-    _check_ast_count(4, t + 1)
+    # NOTE: there's extra contiguous kernels here since it's realizing both the CONTIGUOUS and its parent COPY
+    # why does multi call contiguous on a COPY?
+    _check_ast_count(7, t + 1)
```
Qazalin, line 225: This is the VIZ
image
This is the same with openpilot, it's calling contiguous on a COPY which should fold to the same BUFFER.

PR #8623, opened 2025-01-15
JITBEAM=2 for LLaMA-3 8B on 4 GPUs [pr]

chenyuxyz:
is it fast?


chenyuxyz:
adds 40 second to benchmark job and 3X faster

Issue #8612, opened 2025-01-14
ShapeTracker.real_size is wrong after pad


eliotgolding:
This was mentioned in the meeting. Example: `ShapeTracker.from_shape((10,)).pad(((2,4),)).real_size() == 14`. It happens because the vmax of the idx function is checked without considering valid.

For one view finding real_size is easy, just shrink off the mask first. We can get a good idea of the real_size for multi views by taking the max value of the first view (`st.views[0]`), because the results of other views are just inputs to the first one. But it's hard to find the exact size in general because not all those inputs to `st.views[0]` are accessible per the strides of the other views. See below for an example where just checking the first view would give real_size == 100.
```python
st = ShapeTracker.from_shape((10,10)).pad(((2,4), (3,1))).reshape((16*14,)).stride((17,))
assert st == ShapeTracker(views=(View(shape=(16, 14), strides=(10, 1), offset=-23, mask=((2, 12), (3, 13)), contiguous=False), View(shape=(14,), strides=(17,), offset=0, mask=None, contiguous=False)))
def valid_tiu(x):
  tiu = st.to_indexed_uops((x,))
  return int(tiu[0]) if bool(tiu[1]) else -1

assert max(valid_tiu(x) for x in range(st.size)) == 77
```

Note that for any sequence of MovementOps on `ShapeTracker.from_shape((N,))`, the multi-view ST is built up from single views, so at one point it was equal to the single-view ST with just `views[0]`. Therefore for any sequence of movementops on `from_shape((N,))`, the result of the approximation on that first view will never be bigger than N, and doing this check will never give a false negative for a shapetracker being able to fit in a buffer.

For this reason I think this method is OK for fixing real_size, I wanted to check before writing it though.

PR #8628, opened 2025-01-15
limit real_size to the size of first View of ShapeTracker

eliotgolding:
#8612

Like I said above, it's still wrong but it will never increase after movementops now.


test/external/fuzz_shapetracker_size.py
```diff
@@ -0,0 +1,13 @@
...
+from test.external.fuzz_shapetracker import shapetracker_ops as st_ops
+from test.unit.test_shapetracker_math import MultiShapeTracker
+from tinygrad.helpers import getenv
+import random
+
+random.seed(getenv("SEED", 42))
+for i in range(getenv("CNT", 2000)):
+  if getenv("DEBUG", 0) >= 1: print()
+  N = random.randint(1, 10000)
+  mst = MultiShapeTracker([ShapeTracker.from_shape((N,))])  # st_ops don't mutate regular shapetrackers for some reason
```
chenyuxyz, line 11: without figuring out the reason, this reads like a hack

tinygrad/shape/shapetracker.py
```diff
@@ -72,12 +72,10 @@ def to_uop(self) -> UOp: return UOp(Ops.VIEW, dtypes.void, (), self)
...
     return views_to_indexed_uops(self.views, tuple(_idxs) if _idxs is not None else None)
 
+  # upper bound on buffer size required to fit this shapetracker
   def real_size(self) -> int:
     if 0 in self.shape: return 0
-    idx, valid = self.to_indexed_uops()
-    if not valid.vmax: return 0
-    assert idx.vmax < 1e12, f"real_size broken for {self}"
-    return int(idx.vmax+1)
+    return int( (v.shrink(v.mask) if (v:=self.views[0]).mask else v).to_indexed_uops()[0].vmax + 1 )
```
chenyuxyz, line 78: strange spacing

chenyuxyz (commented):
seems fine, should this have [pr]?

test/external/fuzz_shapetracker_size.py
eliotgolding: I mean that MultiShapeTracker.reshape mutates self, ShapeTracker.reshape returns a new ShapeTracker. `st_ops` were written to be used with the MultiShapeTracker behaviour. If you want I will redo it, or at least remove the confusing comment

PR #8629, opened 2025-01-15
only use BUFFER_VIEW in disk [pr]

Qazalin:
This is currently blocking deleting forced_realize.


geohot:
Fine with just making it for DISK as long as there's no major regressions

PR #8633, opened 2025-01-15
am: bind for sdma

nimlgen:


geohot:
Hmm, why are indirect queues faster/better?

nimlgen:
CPU enqueue times are better. Copy ops from graph is just one indirect op vs copying all of them onto the ring.

We do this for NV and enqueue times are way better (vs AMD)

PR #8635, opened 2025-01-15
tf32 tc for nv and ptx

ignaciosica:
Walkthrough on tf32 tc (#8426)
Added tf32 flag only to correctness tests in benchmark (tensor core tests), I don't know if you want to add it to some real benchmarks as well

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -1982,7 +1982,7 @@ class TestKernelOpts(unittest.TestCase):
     Tensor.manual_seed(1552)
     for tc in Device[Device.DEFAULT].renderer.tensor_cores:
       # bf16 buffer returns float32 numpy outputs so test would fail. testing opt with half suffices.
-      if tc.dtype_in == dtypes.bfloat16: continue
+      if tc.dtype_in != dtypes.half and tc.dtype_out != dtypes.half: continue
       a, b = Tensor.rand(N, N, dtype=tc.dtype_in), Tensor.rand(N, N, dtype=tc.dtype_in)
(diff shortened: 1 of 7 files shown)


test/test_linearizer.py
```diff
@@ -1982,7 +1982,7 @@ def test_tensor_core_opts(self):
     Tensor.manual_seed(1552)
     for tc in Device[Device.DEFAULT].renderer.tensor_cores:
       # bf16 buffer returns float32 numpy outputs so test would fail. testing opt with half suffices.
-      if tc.dtype_in == dtypes.bfloat16: continue
+      if tc.dtype_in != dtypes.half and tc.dtype_out != dtypes.half: continue
```
ignaciosica, line 1985: changed to check for half based on comment above stating that testing with half suffices

geohot:
Bounty is yours! 4 -> 23 TFLOPS w/o BEAM on my 3090. Same PayPal?

ALLOW_TF32 should probably be a context var. See issue:
```
BEAM=2 ALLOW_TF32=1 DEBUG=2 python3 extra/gemm/simple_matmul.py
BEAM=2 ALLOW_TF32=0 DEBUG=2 python3 extra/gemm/simple_matmul.py
...
tinygrad.codegen.kernel.KernelOptError: no tensor core available
```
but this isn't super well thought through (idk what beam does here)

ignaciosica:
Sure, I'll take a look int that issue

> Same PayPal?
yes

geohot:
Paid

ignaciosica:
Received! Thanks

PR #8636, opened 2025-01-15
[poc] Single file onnx.py

geohotstan:
another day, another massive refactor instead of properly doing small PRs 🫠 

is having an Onnx class ok?
added `make_onnx_ops` to keep ops in a scope so it can be added easily.
also ordered the onnx_ops kinda like our tensor op docs (still need cleanup)

I do think the code in `class OnnxSession:` is cleaner though (I gotta add back the debug code but the debug code before was really ugly)


geohotstan:
gonna split this into multiple prs

PR #8639, opened 2025-01-15
move subbuffer to a rewrite rule in the scheduler 

Qazalin:
This has been a dream for me ever since delete_lazy, so much time wasted trying to work around it.
Now it's once and for all in an isolated rewrite rule you can reason about. Tensor and UOp should not ever care if something became a subbuffer.

I first thought it was realize.py's decision but I was wrong, the scheduler decides how BUFFERs get created, realize just allocates/frees(?) them.

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -51,17 +51,6 @@ tensor_uop_spec = PatternMatcher([
   # ASSIGN changes the value of a realized buffer
   (UPat(Ops.ASSIGN, name="assign", src=(UPat.var("target"), UPat.var("new_val"))),
    lambda assign,target,new_val: (target.op is Ops.BUFFER or target.is_realized) and (assign.dtype == target.dtype == new_val.dtype)),
-
-  # TODO: BUFFER_VIEW is overloaded, it should be removed.
-  # BUFFER_VIEW shares the device buffer with its source, it uses a subbuffer of the underlying source buffer
-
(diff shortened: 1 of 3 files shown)


tinygrad/engine/schedule.py
```diff
@@ -51,17 +51,6 @@
   # ASSIGN changes the value of a realized buffer
   (UPat(Ops.ASSIGN, name="assign", src=(UPat.var("target"), UPat.var("new_val"))),
    lambda assign,target,new_val: (target.op is Ops.BUFFER or target.is_realized) and (assign.dtype == target.dtype == new_val.dtype)),
-
-  # TODO: BUFFER_VIEW is overloaded, it should be removed.
-  # BUFFER_VIEW shares the device buffer with its source, it uses a subbuffer of the underlying source buffer
-
-  (UPat(Ops.BUFFER_VIEW, name="root", src=(UPat.var("x"),)), lambda root,x:
```
Qazalin, line 58: These asserts do not belong to tensor_uop_spec anymore.
Because the same compute is represented by the correct Op (BITCAST/CONTIGUOUS) at the tensor level.

At the realize.py level, ViewOp does an assert, which is enough to validate a subbuffer is correctly constructed.

I am 70% sure we don't even need ViewOp anymore. That will be a diff like "delete Empty Op" https://github.com/tinygrad/tinygrad/pull/8544, where the children just have a subbuffer in the ScheduleItem bufs, and ExecItem does the right thing (ensure_allocated).

PR #8640, opened 2025-01-15
remove the uop buffers dict, just give BUFFER UOps a Buffer arg

Qazalin:
The advantage to this is that pickling and freeing buffers just works without having to keep track of external state.

I don't think there's anything different between a BUFFER UOp and a Buffer that it maps to. That buffer_num counter can go too as the arg is deduped.

New VIZ:
image

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -51,17 +51,6 @@ tensor_uop_spec = PatternMatcher([
   # ASSIGN changes the value of a realized buffer
   (UPat(Ops.ASSIGN, name="assign", src=(UPat.var("target"), UPat.var("new_val"))),
    lambda assign,target,new_val: (target.op is Ops.BUFFER or target.is_realized) and (assign.dtype == target.dtype == new_val.dtype)),
-
-  # TODO: BUFFER_VIEW is overloaded, it should be removed.
-  # BUFFER_VIEW shares the device buffer with its source, it uses a subbuffer of the underlying source buffer
-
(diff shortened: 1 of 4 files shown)


Qazalin:
can we deal with PICKLE_BUFFERS in Buffer `__reduce__`?
We really need it for VIZ and process replay, otherwise it'll pickle all the bytes.

Qazalin:
gonna get the BUFFER_VIEW stuff merged first. The Buffer pickling is open for debate, I'd prefer it's in device.py not ops.py.

tinygrad/engine/schedule.py
```diff
@@ -447,8 +446,9 @@ def sink_outputs(ctx:ScheduleContext, sink:UOp) -> None:
 def create_subbuffer(ctx:ScheduleContext, base:UOp, b:UOp, root:UOp, x:UOp):
   if not root.device.startswith("DISK"): return None
   if x.op is not Ops.VIEW: x = x.src[-1] # TODO: remove this once forced_realize is gone
-  buffers[b] = x.buf_uop.buffer.view(b.size, b.dtype, unwrap(x.st).views[0].offset*x.dtype.itemsize)
-  return base.replace(src=(b, root.replace(op=Ops.BUFFER_VIEW)))
+  vbuf = b.replace(arg=x.buf_uop.buffer.view(b.size, b.dtype, unwrap(x.st).views[0].offset*x.dtype.itemsize))
+  ctx.tensor_uops[vbuf] = ctx.tensor_uops[b]
```
Qazalin, line 450: do we need Ops.SUBBUFFER?
is this SHRINK(BUFFER)?

PR #8647, opened 2025-01-16
make onnx runner a class

geohotstan:
Few things of note
- if we ever want to do our own protobuf parsing, I refactored it so that we just have to replace the parse functions under `# ***** protobuf parsing ******` and the model parsing that uses said functions in `OnnxSession.__init__`
- I also added a thing that toggles training. The training ops in onnx that requires gradient all have "ai.onnx.preview.training" as domain. When I was writing tests, I tried to create the node without the domain and it didn't work, so I think that domain name is a must and the check in `__init__` is fine.
- I also added in code the specs that are unsupported. This way we can do without an external src link since the entire spec is in code. (not sure if good idea)
- debug=1 also looks a little different now. DEBUGONNX=1 only shows ops, DEBUGONNX=2 to see attributes, inputs and outputs.

if I can move onnx_ops into onnx, I can make the debug much nicer. 
like right now I can't inspect the ops in `importlib.import_module('extra.onnx_ops')` since there's a darn circular import.

let's see if this refactor is palatable first 🫠

### DIFF
--- a/examples/compile_tensorflow.py
+++ b/examples/compile_tensorflow.py
@@ -8,7 +8,7 @@ import numpy as np
 import subprocess
 import tensorflow as tf
 import tf2onnx
-from extra.onnx import get_run_onnx
+from extra.onnx import OnnxSession
 from tinygrad.tensor import Tensor
 from extra.export_model import export_model_clang, compile_net, jit_model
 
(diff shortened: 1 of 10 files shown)


examples/compile_tensorflow.py
```diff
@@ -25,10 +25,10 @@ class TinyOnnx:
   def __init__(self, keras_model):
     input_signature = [tf.TensorSpec([1,32], tf.float32, name='x')]
     onnx_model, _ = tf2onnx.convert.from_keras(keras_model, input_signature, opset=13)
-    self.run_onnx = get_run_onnx(onnx_model)
+    self.onnx_sess = OnnxSession(onnx_model)
 
   def forward(self, x):
-    return self.run_onnx({"x": x}, debug=False)['predictions']
+    return self.onnx_sess.run({"x": x}, debug=False)['predictions']
```
geohot, line 31: Why is this not just __call__ on the class?

geohot:
`OnnxSession` -> `OnnxRunner`

examples/compile_tensorflow.py
geohotstan: Initial idea was to match ort's `InferenceSession` with their `InferenceSession.run()`
but I'll change it :D

Issue #8648, opened 2025-01-16
Intel MBP doesn't handle metal graph properly


MashDevel:
Apple/AMD driver bug makes it so JIT is slower than without using JIT. Should default to setting JIT=2 for Intel MBPs

PR #8649, opened 2025-01-16
Disable graphs by default if using an intel macbook

MashDevel:
Apple/AMD driver bug makes it so JIT is slower than without using JIT. Should default to setting JIT=2 for Intel MBPs

Issue: https://github.com/tinygrad/tinygrad/issues/8648

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -101,7 +101,8 @@ class ContextVar:
   def __gt__(self, x): return self.value > x
   def __lt__(self, x): return self.value < x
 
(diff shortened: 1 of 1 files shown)


tinygrad/helpers.py
```diff
@@ -16,6 +16,7 @@ def prod(x:Iterable[T]) -> Union[T,int]: return functools.reduce(operator.mul, x
 # fix colors on Windows, https://stackoverflow.com/questions/12492810/python-how-can-i-make-the-ansi-escape-codes-to-work-also-in-windows
 if sys.platform == "win32": os.system("")
 
+def is_intel_mbp(): return platform.system() == 'Darwin' and ('Intel' in platform.processor() or 'i386' in platform.processor())
```
geohot, line 19: This shouldn't be a function

MashDevel: 👍🏻

PR #8657, opened 2025-01-17
remove contiguous in mlb in shard

chenyuxyz:


chenyuxyz:
fetch data for resnet is slower. this should be done with not allowing subbuffer stuff

PR #8668, opened 2025-01-18
fractional global_sizes tested in CI

ibkvictor2:
Added float value test to `tinygrad/tree/master/extra/qcom_gpu_driver/qcom_opencl_interop.py`.

Addresses #7181 

Regards,
Ezekiel.


geohot:
Why do you think CI is testing this?

PR #8669, opened 2025-01-18
FSDP in Tinygrad [WIP]

KhanerX:
Instead of explicitly applying All_Gather to the weights and Reduce_scatter to the gradients, this logic can be done directly in multi's alu, r and expand, and keeping the necessary metadata about how we want parameters and gradients to be sharded in Functions. 

This first proof of concept:

- Trains GPT2-Large on 2 GPUS with 12GB VRAM with BS 16 that would OOM without FSDP
- Uses a Maximum of ~9GB VRAM while the model uses more than 12GB w/o FSDP
- Implements all_gather and reduce_scatter logic in Expand and SUM, relying on the fact we always Expand before Matmul (since BS > 1)

This should be implemented a little better, generalized further, handling edge cases and have a lot of tests!

Also, with a little more memory tolerance and the use of new free_intermediate() function it should with JIT no problem, I will make it work in the coming days.


KhanerX:
image
Ran this test on 2x RTX 3060 with 12GB memory each.

Issue #8617, opened 2025-01-14
stacking many tensors hits recursion limit


chenyuxyz:
```
from tinygrad import Tensor, Device
Device.DEFAULT="CLANG"
tensors = [Tensor.empty(28,28) for _ in range(6131)]
Tensor.stack(*tensors).realize()
```

failed with `RecursionError: maximum recursion depth exceeded while calling a Python object`

chenyuxyz:
still valid

chenyuxyz:
can workaround with `MAX_KERNEL_BUFFERS=100`

fangpenlin:
Actually, the kernel buffer error was after I overcame stack overflow issue. I wrote this to adjust the recursion size:

```python
NEW_RECURSION_LIMIT = 100_000
sys.setrecursionlimit(NEW_RECURSION_LIMIT)
```

And then encountered the kernel buffer error

PR #8670, opened 2025-01-18
make add_buffer iterative

KareemMusleh:
[#8617](https://github.com/tinygrad/tinygrad/issues/8617)
```python
from tinygrad import Tensor, Device
tensors = [Tensor.empty(28,28) for _ in range(6131)]
Tensor.stack(*tensors).realize()
```

- [x] make `add_buffers` iterative
- [x] make `top_down_rewrite` iterative
- [x] make `UOp.key` and `UOp._device` iterative
- [x] make `Kernel.get_optimized_ast` iterative
- [ ] make UOp.full_shape iterative

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -172,12 +172,13 @@ jobs:
       run: NV=1 python test/external/external_benchmark_multitensor_allreduce.py
     - name: Test tensor cores
       run: |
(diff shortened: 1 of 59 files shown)


KareemMusleh:
the problem is with caching in _device, can't figure out how to do it without it failing the tests. Will try to solve this later.

PR #8676, opened 2025-01-19
Remove contiguous on buffer

Qazalin:
Prereq for forced_realize deletion and tensor_map #8580.


Qazalin:
ImageDType is gating this, I'll merge a diff like #8611 first.

It is because CONTIGUOUS(CAST(BUFFER(float))) should not fold.

right now it first becomes:
CONTIGUOUS(BUFFER)
then this diff folds it to a BUFFER.
but the CONTIGUOUS is image, buffer is float and tests expect the output to be image.

PR #8678, opened 2025-01-19
start moving image things to rewrite rules

Qazalin:
This is gating the CONTIGUOUS folding diff because currently the scheduler changes `ImageDType` tensor uops to the base dtype too early, before symbolic_simple is ran.

We cannot safely do this and also merge the forced realize deletion, because CAST(BUFFER(dtype=float), dtype=imagef) can later become a CONTIGUOUS.
a CONTIGUOUS(CAST(BUFFER(dtype=float)), dtype=imagef) can't simplify to a the BUFFER(dtype=float), because the output Buffer is expected to be an ImageDType, not a float.


Qazalin:
New VIZ:
image


Old VIZ:
image

I think it's more correct this way.

Qazalin:
err, two extra kernels in openpilot
image

PR #8680, opened 2025-01-19
add Half-Precision Accumulation Support for Tensor Cores in NV, CUDA, and PTX

ignaciosica:
~Keeping it a draft as there is one tc test failing for PTX. (test_tensor_cores_multi_reduce)~ 

edit: fixed in #8814

I will appreciate a benchmark run though.

**Overview**
For ptx, added output packing/unpacking for the wmma registers in order for them to work with half
For nv, changed the way the elements are accessed and changed reg type from `+f` to `+r`


Qazalin:
pushed to benchmarks.

chenyuxyz:
tests for this? at least something that runs on real green machine in benchmark

ignaciosica:
they should already be running on benchmark

see

image

chenyuxyz:
oh okay i see, thanks will merge once benchmark job finished

PR #8696, opened 2025-01-21
copy zero fails [pr]

geohot:
Err, this is passing cause unit tests are run on clang in CI, but fails locally


geohot:
So it looks like it's better now, it's still a CONST but at least the copy is gone

PR #8698, opened 2025-01-21
Tensor UOps can become a buffer or const after scheduling

Qazalin:
#8696


Qazalin:
uh const with symbolic shape is stride (1,) yeah.

PR #8705, opened 2025-01-21
search over shape

ignaciosica:
I don't think this is worth the bounty, it only consistently improves speed to 150+ for `N=4096`. 

Also, it required a change in `Opt` api as it was not expressive enough. Before,  `OptOps.TC` had `axis` and `amt`. `axis` specified the axis on which to perform the TC optimization and `amt` controlled which kinds of kernels were eligible for tensor cores application (default 2 during BEAM, 0 otherwise). Now, after the api change, `OptOps.TC` has `axis`, `amt` and `arg`. Axis remains the same as before. `amt` now represent `tc_select`, a var that specifies which tc are available to choose from while performing the optimization. `arg` now controlls which kernel is eligible for tensor cores. I founded more suitable `arg` to be the kind of opt and `amt` to be the selection of the tc.

The only other solution for searching over shapes without changing the api is fixing tc_opt to 2 (always allowing padding), but I'm not sure we want that.

Maybe this isn't even worth the change in the api, I'm not sure. My preliminary tests were promising but testing with more shapes the gains in performance are not there. The good thing is that the beam time doesn't seem to be significantly impacted.

```
tc_select -- specifies which tensor core(s) to use for optimization (default -1)
      -1: iterates through all available tensor cores in order and uses 
          the first one that matches the requirements (dims and dtypes)
      [0-N]: uses only the n'th tensor core available; useful for search
```

image
image
image
image
image


ignaciosica:
Maybe I'll just close; only one benchmark seemed to be affected and for the worse

image

chenyuxyz:
We definitely want to search over tc shapes. There are not many beam search benchmark, and very few tc in llama so i don't expect to see a diff in benchmark ci. Other than the pure gemm kernel, you can train resnet with mlperf script or cifar / stable diffusion that has more gemms. Also vs_theoretical should get faster right?

Specifically for next steps, i understand OptOp amt is abused now (for current TC and SWAP it's not even an amount). You can make amt to TC a tuple instead of single number maybe. If you are interested to change the api, write a short TIP proposal for that and we should be able to resolve it quickly.

ignaciosica:
Thanks for your feedback! Yes, I agree it's a good thing to search over, but got discouraged after the preliminary results and the fact that currently only cuda has different shaped tc given the same types. I didn't had in mind the fact that llama had few tc though.

Sure, I'll run benchmarks on those models as well.

Making TIP today. Yes, I like it being a tuple a bit more I think.

ignaciosica:
closing in favor of #8793

PR #8720, opened 2025-01-22
move llvm_bf16_cast to renderer for CLANG and LLVM [pr]

mesozoic-egg:
Uses the same LLVM bitcasting trick to convert b16 to float32/float16 but directly in kernel.

Running llama3 8B

CLANG F16:
clang_f16

LLVM F32:
llvm_f32

NV BF 16:
nv_bf16

NV F16:
nv_f32


tinygrad/renderer/__init__.py
```diff
@@ -113,6 +113,10 @@ def launch_dims(self, var_vals:dict[Variable, int]):
     local_size = [sym_infer(sz, var_vals) for sz in self.local_size] if self.local_size is not None else None
     return global_size, local_size
 
+def bf16_bitcast_to_float(buf:UOp, idx:UOp, root:UOp):
```
geohot, line 116: Hmm, I don't think this belongs in init. The right place is probably in rewriter.py, see `get_late_rewrite_patterns`

mesozoic-egg:
@geohot  updated, I put the pattern into `get_late_rewrite_ops` and check if it's needed via `is_dtype_supported`

geohot:
Bounty is yours! DM to remind me to pay and I'll do next time I'm doing crypto

PR #8722, opened 2025-01-22
Tensor UOps can become a VIEW(BUFFER) with all movement ops applied

Qazalin:


Qazalin:
another untested thing going on with openpilot. blocking merge until it is fixed.
image

Qazalin:
I think it's something with the process replay capture, disabling process replay fixes that increase in the last run.

Qazalin:
```
CAPTURE_PROCESS_REPLAY=1 PYTHONPATH=. QCOM=1 taskset -c 4-7 python3 test/external/external_benchmark_openpilot.py https://github.com/commaai/openpilot/raw/v0.9.7/selfdrive/modeld/models/supercombo.onnx | tee openpilot_0_9_7.txt
```
last step is:
```
jitted:  73.7642 ms
jitted:  395.7401 ms
```

Qazalin:
hmm, does this work with pm_gradient?

Qazalin:
I think I overlooked what happens to gradient in this case, and i can't really trust this green
https://github.com/tinygrad/tinygrad/actions/runs/12948105101/job/36116059131 (with that VIEW upat in pm_gradient)

Qazalin:
status: pm_gradient doesn't have a rewrite rule for VIEW,
the most principled way to write this would be to somehow backtrack from a VIEW to the movement ops that resulted in that VIEW and apply the movement ops on top of the tensor UOp.

I don't know yet if there is a way to reverse MovementOps given the starting (base) VIEW and the final VIEW.

PR #8723, opened 2025-01-22
Move BF16 casting to renderer rewrite rules (#6909) #8721

1957-001B:
Move BF16->float32 autocasting to renderer rewrite rules as requested by @geohot.

Changes:
Add BF16 pattern matching rule in ClangRenderer's extra_matcher
Use explicit Ops.CAST pattern instead of relying on fix_bf16()

This addresses issue #6909 by moving the BF16 casting logic into the renderer's rewrite rules, matching the pattern approach used by other backends.

Fixes #6909

### DIFF
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -183,8 +183,17 @@ class ClangRenderer(CStyleLanguage):
   code_for_op = {**({k:v for k,v in CStyleLanguage.code_for_op.items() if k not in [Ops.EXP2, Ops.SIN, Ops.LOG2]}),
                  Ops.SQRT: lambda x,dtype: f"__builtin_sqrt({x})" if dtype == dtypes.float64 else f"__builtin_sqrtf({x})"}
   # LLVM legalizes double => half cast on systems that don't support it natively (like x86 cpus without AVX512-FP16) into a compiler-rt libcall.
-  extra_matcher = PatternMatcher([(UPat.var("x", dtypes.float64).cast(dtypes.float16), lambda x: x.cast(dtypes.float32).cast(dtypes.float16))]) + \
-    CStyleLanguage.extra_matcher
+
+  extra_matcher = PatternMatcher([
+    # Pattern for float64->float16 conversion via float32 intermediate
+    (UPat.var("x", dtypes.float64).cast(dtypes.float16),
+      lambda x: x.cast(dtypes.float32).cast(dtypes.float16)),
+    # BF16 uses explicit cast op pattern since float->BF16 requires
+    # special handling in hardware/software emulation
+    (UPat(Ops.CAST, dtypes.bfloat16, UPat.var("x")),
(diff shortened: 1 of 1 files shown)


1957-001B:
Reverted back to simpler fix.

lorenzoliusso:
The second rule causes an infinite recursion, you can test it by running llama and turning off `fix_bf16()`

geohot:
Is this still wanted? I merged a hack #8935 but this sort of thing looks more correct.

1957-001B:
Yeah I can fix that issue just been really busy will get on it soon. @geohot

1957-001B:
If all checks pass I believe that this is now working @geohot and also removed the hack https://github.com/tinygrad/tinygrad/pull/8935

tinygrad/renderer/cstyle.py
```diff
@@ -183,8 +183,16 @@ class ClangRenderer(CStyleLanguage):
   code_for_op = {**({k:v for k,v in CStyleLanguage.code_for_op.items() if k not in [Ops.EXP2, Ops.SIN, Ops.LOG2]}),
                  Ops.SQRT: lambda x,dtype: f"__builtin_sqrt({x})" if dtype == dtypes.float64 else f"__builtin_sqrtf({x})"}
   # LLVM legalizes double => half cast on systems that don't support it natively (like x86 cpus without AVX512-FP16) into a compiler-rt libcall.
-  extra_matcher = PatternMatcher([(UPat.var("x", dtypes.float64).cast(dtypes.float16), lambda x: x.cast(dtypes.float32).cast(dtypes.float16))]) + \
-    CStyleLanguage.extra_matcher
+
+  extra_matcher = PatternMatcher([
```
geohot, line 187: What is this formatting? It should look like the rest of the code

1957-001B: Will fix sorry.

geohot:
There's no test for the renderer change. This should be two PRs.

chenyuxyz:
stale

FlorianHeigl:
@chenyuxyz are you sure this ought to be closed?

- the actual problem still exists
- dependent issues are still open 
- affects people 
- author didn't seem to be reluctant to add the tests asked
- author said he was really busy but like 3 days before you closed it as stale

maybe i'm missing something, but i'd say if the issue is open for months already then the person trying to fix it should also get 3 months.

divinity76:
> @chenyuxyz are you sure this ought to be closed?
IMO this should not be closed, but the actual issue has proven extremely difficult to reproduce,
also afaik it has only been reproduced on RTX3090 and RTX4090, and does not reproduce on any other GPUs (eg does not reproduce on RTX3080)

1957-001B:
I have continued to work on this in another pr that was closed for failing one of the windows tests I will fix it at some point.

1957-001B:
This issue (#6905) is device-agnostic, tied to the ClangRenderer in tinygrad/renderer/cstyle.py, and reproducible with CPU=1 (CLANG backend).

PR #8731, opened 2025-01-23
reorder and categorize onnx_ops

geohotstan:
grouped the ops into similar categories much like the ones we have in `/docs/tensor/` 

I have prepared another PR for moving the Tensor Ops into these categories too, it also moves this into onnx.py. Will PR if this is accepted.


chenyuxyz:
seems fine, i don't have strong preference

how's top 100 onnx model support going?

PR #8749, opened 2025-01-25
fix multi with no real srcs

geohot:
chenyu, i still don't fully understand what real is? i need this to fix gradient

### DIFF
--- a/test/test_multitensor.py
+++ b/test/test_multitensor.py
@@ -344,9 +344,7 @@ class TestMultiTensor(unittest.TestCase):
   # NOTE: this is failing on LLVM CI, no idea why. Works locally.
   @unittest.skipIf(CI and Device.DEFAULT in ("CUDA", "NV", "LLVM"), "slow")
   def test_data_parallel_resnet(self):
-    import sys, pathlib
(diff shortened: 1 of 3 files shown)


tinygrad/ops.py
```diff
@@ -896,15 +896,15 @@ def __init__(self, pm, ctx=None):
     self.replace: dict[UOp, UOp] = {}
   def top_down_rewrite(self, n:UOp) -> UOp:
     if (rn := self.replace.get(n)) is not None: return rn
-    new_src = tuple(map(self.top_down_rewrite, n.src))
+    new_src = tuple([self.top_down_rewrite(x) for x in n.src])
```
geohot, line 899: this was masking a stopiteration error that took me 30 min to track down

test/test_multitensor.py
```diff
@@ -899,6 +895,7 @@ def test_uneven(self):
     np.testing.assert_equal((a+a).numpy(), na+na)
     np.testing.assert_equal((b+b).numpy(), nb+nb)
 
+  @unittest.skip("why didn't this work?")
```
geohot, line 898: i don't understand why this is supposed to fail, shouldn't it just be 0?

PR #8753, opened 2025-01-26
remove Function class [pr]

geohot:


docs/developer/developer.md
```diff
@@ -9,7 +9,7 @@ There is a good [bunch of tutorials](https://mesozoic-egg.github.io/tinygrad-not
 
 ## Frontend
 
-Everything in [Tensor](../tensor/index.md) is syntactic sugar around [function.py](function.md), where the forwards and backwards passes are implemented for the different functions. There's about 25 of them, implemented using about 20 basic ops. Those basic ops go on to construct a graph of [UOps](../developer/uop.md).
```
AlexanderYastrebov, line 12: https://tinygrad.org/#faq points to the https://github.com/tinygrad/tinygrad/blob/master/tinygrad/function.py

> Is tinygrad inference only?
No! It supports full forward and backward passes with autodiff. [This](https://github.com/tinygrad/tinygrad/blob/master/tinygrad/function.py) is implemented at a level of abstraction higher than the accelerator specific code, so a tinygrad port gets you this for free.

PR #8756, opened 2025-01-26
FSDP in (New!) Tinygrad [WIP]

KhanerX:
With new differentiable multi UOp, Model parallel is supported out of the box, by sharding optimizer parameters.

I will look into 3D parallelism, that lets you shard both model (optimizer) and data, it currently OOM I think because it realizes one of expanded intermediaries.

This first commit trains GPT2-large on two 8GB cards, would OOM without model parallelism.


KhanerX:
I've found the Issue with sharding both model and data. it's in Embedding layer, it boils down to:
```python
from tinygrad import Tensor, nn

B, T = 4, 64
vocab_size = 50000
embed_dim = 1024

GPUS = ("CLANG", "CUDA")

x = Tensor.randint((B, T), low=0, high=vocab_size).shard_(GPUS, axis=0).realize()
emb = Tensor.randn((vocab_size, embed_dim)).shard_(GPUS, axis=1).realize()

big_shape = x.shape+emb.shape

emb = emb.expand(big_shape)
x = x.reshape(x.shape+(1, 1)).expand(big_shape)

x.mul(emb).sum(-2).realize()
```

This OOMs by trying to allocate a 26 GB buffer, for the following kernel:

```cuda
//E_16_800000_8_16_4

extern "C" __global__ void __launch_bounds__(128) E_16_800000_8_16_4(float* data0, int* data1) {
  int gidx0 = blockIdx.x; /* 800000 */
  int gidx1 = blockIdx.y; /* 16 */
  int lidx1 = threadIdx.y; /* 16 */
  int val0 = *(data1+(lidx0+(gidx1<<3)));
  float cast0 = ((float)(val0));
  long alu0 = (((long)((lidx1<<2)))+((long)((lidx0*51200000)))+((long)((gidx0<<6)))+(((long)(gidx1))*409600000ll));
  *(data0+alu0) = cast0;
  *(data0+(alu0+1ll)) = cast0;
  *(data0+(alu0+2ll)) = cast0;
  *(data0+(alu0+3ll)) = cast0;
}
```


I have looked at it in VIZ, I think the issue is COPY after EXPAND
image

I couldn't work out how to prevent this, however I think it has something to do with how we choose the shared axis in Multi ALU. particularly this line:

`python  # NOTE: they all have to share an axis, we always choose [-1]
  axis, bounds = axes[-1] if len(axes := dedup([(x.axis, x.bounds) for x in msrcs if x.axis is not None])) else (None, None)`
  
The reason for my guess is, when you change the last line of my script to `emb.mul(x).sum(-2).realize()` (reversing the order of msrcs in alu), it doesn't OOM. the difference can be seen in following get_multi_map_3 rewrites:

**When x.mul(emb) (OOM):**
image
notice expand and cast before copy, and notice the multi before reduce_axis has axis=3 which belongs to emb

**When emb.mul(x) (OK):**
image

I'm kinda stuck here, I appreciate any pointers to what I'm missing or should be looking into. thanks.

geohot:
Hmm yea, the copy shouldn't be after expand. I locked the bounty for you if you are still interested. See #9051 for the direction to fix this.

chenyuxyz:
still active?

KhanerX:
@chenyuxyz yeah I was waiting for cast_before_view, now it's merged and I'm continuing work, will update soon.

Qazalin:
@KhanerX I'd start by writing minimum failing tests for the behavior you want.
It's easy to add reordering rules to the scheduler `sym` now. CAST_BEFORE_VIEW is an example, there's much more to optimize in when we realize expands.

KhanerX:
COPY after expand in Multi-GPU only happens in 1 case:
ALU on two MULTI with different axis, where the source that ends up being copied is a result of an ALU, which copy can't move before it later on:
```python
a = Tensor.empty(1, 16).shard_(GPUS, axis=1).realize()
b = Tensor.empty(16, 16).to_(GPUS).realize()
c = Tensor.empty(16, 16).shard_(GPUS, axis=0).realize()
((a+b)*c).realize()
```
image


There are two ways to fix this. 
1. Add a rewrite rule to re-order copy (move before ADD in this case), after multi_pm
2. Before multi_pm, gather the parameters of the tensor that so that its MLB has axis=None, which doesn't produce the problematic COPYs in the first place.
so this:
image

becomes this:
image

I think 2 is the correct solution, so my plan is:

- [ ] Add an argument to MULTI, that tells whether or not this MULTI should be gathered before any compute (We set it to true for model parameters in FSDP)
- [ ] Write a rewrite rule to do the all_gather for these before multi_pm
- [ ] For each FSDP unit(e.g each layer) , realize the forward and backward pass, and Deallocate the buffers.

Qazalin:
We tend to not prefer changing UOp specs unless there is no other path.
I think 1 is the right solution.

chenyuxyz:
stale, i unlocked the bounty

PR #8765, opened 2025-01-27
Remove `SimpleMathTrait`

eitanturok:
Removed `__eq__`, `__lshift__`, `__rshift__` from `Tensor` because they are now implemented in `MathTrait`. Are there more functions I can remove?

I want to get rid the in-place functions (e.g. `__iadd__`). But it seems like they must be in `Tensor.py` because we need `assign()` to take care of reassigning the `lazydata`, something we can only do in `Tensor`, not `UOp`.

It seems all other functions in `Tensor` that are wrappers for `UOp`s must remain because they take care of broadcasting dimensions and casting `dtype`s.


geohot:
Feel free to reopen when this is ready

PR #8766, opened 2025-01-27
usb gpu

nimlgen:
rebasing on master


geohot:
Does this work on 7600 on tiny10? What's the timing for a cold boot on test_tiny?

nimlgen:
sorry, haven't made it work for 7600 today. tried to get >4g addresses, since we got discovery there to offset regs (it's hardcoded for 7900 now). so resizable bar shows only 256mb supported size (and i don't know if 64-bit tlps are supported at all by the controller). will do tomorrow or at least hardcode to collect boot times.

once we get sdma working we'll be like nvme. since nvme is the initiator of all transfers, the fw just sets up several bytes in the ring for the nvme (like addresses to push/pull data to/from). so going to try to see if the controller can be the initiator of fast (large) transfers, since we need this to load fw to the gpu.

geohot:
ahh cool, yea firmware load might have to be through the controller. what's really important for speed is data/weight transfer once the GPU is booted, makes sense SDMA is the same as NVMe. maybe you can boot that firmware first (after the PSP) to bring up the rest.

nimlgen:
so, will move queues to the controller, otherwise putting into the queues makes it so slow.
maybe timeline signals as well.

0xA000-0xAFFF and 0xB000-0xB1FF should be available (maybe not the whole space, since controller will try to write nvme commands)

tinygrad/runtime/ops_amd.py
```diff
@@ -834,15 +879,16 @@ def __init__(self, device:str=""):
...
-    self.sdma_queue = self.create_queue(kfd.KFD_IOC_QUEUE_TYPE_SDMA, 0x800000)
+    self.compute_queue = self.create_queue(kfd.KFD_IOC_QUEUE_TYPE_COMPUTE, queue_size, ctx_save_restore_size=wg_data_size + ctl_stack_size,
+                                           eop_buffer_size=0x1000, ctl_stack_size=ctl_stack_size, debug_memory_size=debug_memory_size)
+    self.sdma_queue = self.create_queue(kfd.KFD_IOC_QUEUE_TYPE_SDMA, queue_size)
 
     super().__init__(device, AMDAllocator(self), AMDLLVMRenderer(self.arch) if getenv("AMD_LLVM", 0) else AMDRenderer(self.arch),
                      AMDLLVMCompiler(self.arch) if getenv("AMD_LLVM", 0) else HIPCompiler(self.arch), functools.partial(AMDProgram, self),
-                     AMDSignal, functools.partial(AMDComputeQueue, self), functools.partial(AMDCopyQueue, self, max_copy_size=max_copy_size))
+                     AMDSignal, functools.partial(AMDComputeQueue, self), functools.partial(AMDCopyQueue, self, max_copy_size=max_copy_size),
+                     kernargs_size=(8 << 10) if self.dev_iface.__class__ is USBIface else (16 << 20))
```
geohot, line 891: use isinstance

Issue #8716, opened 2025-01-22
TIP 9 - Opt API change


ignaciosica:
**Context**
Multiple tensor core shapes can match a given shape and dtype. Larger tensor cores are not always faster because smaller ones enable more memory access permutations. Therefore, a search process is needed to determine which tensor core to use for best performance. Additionally, kernels must be uniquely determined by both the AST and the set of `opt-ops` applied.

**Current Limitation**
The existing Opt API cannot specify which tensor core is selected. Instead, it only allows controlling whether certain kernels are eligible for tensor core application. Currently, this happens via an `amt` field that is repurposed to configure eligibility rather than an amount, which is not fully expressive.

**Proposed Change**
Replace `amt` with `arg` so that:
`axis` remains the axis for the transformation.
`arg` can store different types of values depending on the operation.

```python
@dataclass(frozen=True, order=True)
class Opt:
  op: OptOps
  axis: Optional[int] = None
  arg: Optional[int | tuple] = None
```

For tensor core optimization (`OptOps.TC`), `arg` can be used to specify two aspects:
- `tc_select` (which tensor core to use): A tuple or integer designating the index of the selected tensor core or -1 to indicate a search through all available options and use the first one that matches.
- `tc_opt` (which kernels qualify for tensor core usage).

The rename from 'amt' to 'arg' is also supported by the fact that currently, for OptOps.SWAP, amt does not represents an amount but rather another axis instead.

geohot:
Yea, I'm good with this change

PR #8767, opened 2025-01-27
[TIP-9] rename Opt's `amt` to `arg`

ignaciosica:
# TIP-9 pt.1

#8716

Can't run process replay as changing Opt's API breaks it (I believe due to cache issues).

[Failed PR run](https://github.com/ignaciosica/tinygrad/actions/runs/12989660738/job/36224173366#step:21:7119)

image


Qazalin:
yea if the dataclass is different we can't unpickle the captured process replay.

ignaciosica:
I think this might be an issue with cache. I'm not able to replicate locally on tiny green.

I think it is loading the val (`beam_search`) with the older dataclass:

``` python
def beam_search(lin:Kernel, rawbufs:list[Buffer], amt:int, allow_test_size=True, disable_cache=getenv("IGNORE_BEAM_CACHE")) -> Kernel:
  global beam_pool
  key = {"ast": lin.ast.key, "amt": amt, "allow_test_size": allow_test_size, "device": lin.opts.device, "suffix": lin.opts.suffix}
  if not disable_cache and CACHELEVEL >= 1 and (val:=diskcache_get("beam_search", key)) is not None:
    ret = lin.copy()
    for o in val[len(lin.applied_opts):]: ret.apply_opt(o)
    return ret
  [...]
```

local `test_tiny` run:

image

I don't know how to work around this. Maybe a cache clean?

PR #8768, opened 2025-01-27
Support CLANG backend on Windows

b1tg:
- support CLANG backend on Windows
- Windows CI with CLANG


geohot:
Hmm, do we really need a COFF loader? Can you just compile the elf on Windows the same as Mac?

@uuuvn This fix CPUProgram for LLVM?

b1tg:
The calling conventions on Windows and macOS are different. The ELF I generated previously seems to be using a non-Windows calling convention.

nimlgen:
Doesn't `-mabi=ms` work with elf?

b1tg:
`-mabi=ms` not work: `clang: error: unsupported option '-mabi=' for target 'x86_64-none-unknown-elf'`

victorqhong:
Does the COFF loader process relocations? Wouldn't that be needed for many programs to work?

uuuvn:
> Hmm, do we really need a COFF loader? Can you just compile the elf on Windows the same as Mac?
> 
> Also, there should only be one Windows CI, just put both backends in the same one.
> 
> @uuuvn This fix CPUProgram for LLVM?

I'm sure we don't really need a COFF loader. The problem with mac abi is that it uses sysv calling convention, just slightly modified, thus requiring the 8-byte stack slot hack.  The win thing should be much easier (like one attribute on a function type of easy) and not even a hack.

#8486 passes on windows ci, clang should be a ~4 line change after it's merged.

Also as @victorqhong pointed out this doesn't even handle relocations

uuuvn:
a quick google search: `__attribute__((ms_abi))` should be it

uuuvn:
https://github.com/uuuvn/tinygrad/pull/3/files <-- +2 lines not counting `platform.machine() if sys.platform != 'win32' else 'x86_64'` thing. Obviously no COFF parsers required
image
image

I'm a bit busy right now, will likely finish #8486 tomorrow and then clang is basically no additional complexity

b1tg:
clang [not support](https://github.com/llvm/llvm-project/blob/main/clang/test/Driver/x86-mabi.c) `-mabi=ms` on elf target,  @uuuvn 's `__attribute__((ms_abi))` method solve the ABI issue and is more simple, i removed the coff loader

tinygrad/device.py
```diff
@@ -222,23 +222,33 @@ def _offset(self, buf, size:int, offset:int): return from_mv(self._as_buffer(buf
...
-    self.fxn = ctypes.CFUNCTYPE(None)(mv_address(self.mem))
+    if sys.platform == "win32":
+      PAGE_EXECUTE_READWRITE = 0x40
+      MEM_COMMIT =  0x1000
+      MEM_RESERVE = 0x2000
+      shellcode = bytearray(lib)
+      ctypes.windll.kernel32.VirtualAlloc.restype = ctypes.c_uint64
+      ptr = ctypes.windll.kernel32.VirtualAlloc(ctypes.c_int(0), ctypes.c_int(len(shellcode)), MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE)
+      buf = (ctypes.c_char * len(shellcode)).from_buffer(shellcode)
+      ctypes.windll.kernel32.RtlMoveMemory(ctypes.c_uint64(ptr), buf, ctypes.c_int(len(shellcode)))
```
geohot, line 233: Do you need a custom Windows function for this? `ctypes.memmove`? Less platform specific code is better

geohot:
I think your python package cache is broken, it's only 3MB and "Install deps" is taking way too long

geohot:
Bounty is yours, now I owe you $400 (george@tinygrad.org to claim with either paypal or USDC). Cache thing should be fixed though, I should have caught on last PR.

b1tg:
windows ci cache fixed in https://github.com/tinygrad/tinygrad/pull/8787

PR #8770, opened 2025-01-27
[TIP-9] rename Opt's amt to arg 2

ignaciosica:
add ignore_beam_cache

~should fix #8767~

moved fix to #8771


Qazalin:
wanna instead bump the DB version?

PR #8777, opened 2025-01-28
start work on kernel op

geohot:


Qazalin:
yea I have to fix base not being base for VIEW(BUFFER)
currently `VIEW(BUFFER)` is always treated as base. Otherwise all views are merged once we reach KERNEL.

PR #8784, opened 2025-01-28
add `Tensor.isclose` and `Tensor.allclose`

cverrier:
As [discussed on Discord](https://discord.com/channels/1068976834382925865/1070745817025106080/1333669036533354497), let's try to add `Tensor.isclose` and `Tensor.allclose` operations. For reference:
- [`torch.isclose`](https://pytorch.org/docs/stable/generated/torch.isclose.html)
- [`torch.allclose`](https://pytorch.org/docs/stable/generated/torch.allclose.html)

I tried to implement some tests with `helper_test_op()`, but I am not sure I did that correctly.

Also, the tests for `Tensor.allclose()` are failing, which I think is due to the fact that PyTorch returns a `bool` and not a `torch.Tensor` object:
```
=========================================================================================================================== short test summary info ===========================================================================================================================
FAILED test/test_ops.py::TestOps::test_allclose - AttributeError: 'bool' object has no attribute 'detach'
FAILED test/test_ops.py::TestOps::test_allclose_edge_cases - AttributeError: 'bool' object has no attribute 'detach'
FAILED test/test_ops.py::TestOps::test_allclose_mixed_cases - AttributeError: 'bool' object has no attribute 'detach'
==================================================================================================== 3 failed, 1436 passed, 82 skipped, 24 xfailed, 23 warnings in 47.03s =====================================================================================================
```

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -1712,6 +1712,44 @@ class Tensor(SimpleMathTrait):
     """
     return self.logical_not().any(axis, keepdim).logical_not()
 
+  def isclose(self, other:Tensor, rtol:float=1e-05, atol:float=1e-08) -> Tensor:
+    """
+    Returns a new tensor with element-wise comparison of closeness to `other` within a tolerance.
+
+    The `rtol` and `atol` keyword arguments control the relative and absolute tolerance of the comparison.
+
+    Two `NaN` values are not close to each other.
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t.numpy())
+    ```
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t.isclose(Tensor([1, 2, 3.1])).numpy())
+    ```
+    """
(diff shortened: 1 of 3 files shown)


tinygrad/tensor.py
```diff
@@ -1712,6 +1712,44 @@ def all(self, axis:Optional[Union[int, Sequence[int]]]=None, keepdim=False):
...
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t.numpy())
+    ```
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t.isclose(Tensor([1, 2, 3.1])).numpy())
+    ```
+    """
+    return ((self - other).abs() <= atol + rtol * other.abs()) * (self.isnan() | other.isnan()).logical_not()
```
cverrier, line 1732: By default, PyTorch does not assume that `NaN` values are equal. Although they have a `equal_nan` option to modify this behavior, I am not sure you want to introduce that based on [a recent poll](https://x.com/__tinygrad__/status/1865279121950970360).

PR #8789, opened 2025-01-28
Update OPS.FLIP lambda call for stride -> flip transition

bautista-garcia:


bautista-garcia:
Accidentally pushed to main instead of fork (CI tests)

PR #8790, opened 2025-01-28
allow VIEW source in DEFINE_VAR spec [pr]

Qazalin:
After the unmasked VALID removal we need to keep this ShapeTracker in the DEFINE_VAR sources. eg, this is the new symbolic mean:
image

This spec already exists in the Tensor UOps spec, this diff extends it to ScheduleItem/Kernel ASTs.


Qazalin:
Also found out we can't just return a constant ShapeTracker for DEFINE_VAR. Because of swizzling. eg. see a var_2d kernel:
image
image


I think the final ShapeTracker needs to exist in the DEFINE_VAR sources.

PR #8793, opened 2025-01-28
search over shape 2

ignaciosica:


test/test_linearizer.py
```diff
@@ -1111,7 +1112,9 @@ def test_tensor_cores_multi_reduce(self):
       # check that get_kernel_actions produces all 9 options
       from tinygrad.engine.search import get_kernel_actions
       tc_actions = [k for i, k in get_kernel_actions(Kernel(realized_ast), False).items() if k.applied_opts[0].op == OptOps.TC]
-      assert len(tc_actions) == 9, f"get_kernel_actions should contain 9 possible TC actions, only got {len(tc_actions)}"
+      assert (
+        len(tc_actions) > 0 and len(tc_actions) % 9 == 0
```
chenyuxyz, line 1116: can this be more specific? something like `9 * (available tensor cores)`

chenyuxyz (commented):
i'd like to separate this into one change that adds all select tc infra but default disabled (or select the existing one so it's no-op if you already added it somewhere), and [pr] can pass

and have another change that enables the search, where we can measure the perf diff

ignaciosica:
Sure, adding a tc_select no-op pr later today!

ignaciosica:
I'll try to run some benchmarks against cifar/resnet and post them here.

Are there some reference value for master I could use?

chenyuxyz:
can you show gemm with different input sizes and show that different TC were picked and speed diff (same style to speed_v_theoretical)? you can also do conv but i think gemm would be a good start

ignaciosica:
I'll resume work on the benchmarks on Monday

ignaciosica:
image

ignaciosica:
TC_SEARCH_OVER_SHAPE
You can see how different tc are selected based on the amount of purple 2 in the kernel shape
image

ignaciosica:
Search with default tc

image

chenyuxyz:
from the screenshot this also qualified the `150+ TFLOPS FP16 or BF16 matmul in tinygrad on 4090 (FP32 acc)` bounty

PR #8795, opened 2025-01-29
do not lockup VIZ when rendering big graphs [pr]

Qazalin:
Currently VIZ locks up if we open a big scheduler graph. This diff adds support for a non-blocking VIZ graph renderer.

Calculating graph layout is the heaviest part of the UI, below is the profile for beautiful_mnist schedule:
image
Previously calling dagre.layout blocked the main thread, this diff offloads it to a [WebWorker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers). Because dagre-d3 didn't allow for this I removed it, now we are calling D3 directly to draw the svg. The UI closely matches old VIZ.


Qazalin:
Status: VIZ is very responsive here! I wanna fix the zooming behavior and make updates smoother.
Currently this is the primary constraint on a fast and responsive VIZ on large graphs.

Qazalin:
looks close enough:
image
image

PR #8796, opened 2025-01-29
Fix test_ops.py on Windows

victorqhong:
Some tests in test_ops.py would fail on Windows with "access violation reading 0xFFFFFFFFFFFFFFFF". See https://github.com/victorqhong/tinygrad/actions/runs/13020694560/job/36320461001?pr=1

The fix is to allocate more memory to prevent reads past the end of the buffers.


nimlgen:
Should be fixed in master

PR #8798, opened 2025-01-29
More complete shapetracker.simplify() (draft)

eliotgolding:
Please let me know if you think this is worth pursuing, I don't think it is, just because the practical benefit is very small outside of the toy examples and fuzzers (I found 2 less views in test_winograd.py, that was it.) But if you want complete shapetracker simplification as per the bounty, I don't think there's a simpler approach than this, because a complete simplifier would be able to solve arbitrary ILP problems.

It converts subsequences of views in a shapetracker into ILP problems and tries to find an equivalent single-view. There are two cases, one where we keep the final shape, and one where the final shape is arbitrary (for the subsequences in the st that don't contain the final view). The second case is sometimes slow in the ILP solver, I think because constraints have a less clear geometric interpretation which makes the solving harder? (I also tried using graph_rewrite to convert to_indexed_uops into ILP problems, but this was also slow, probably for the same reason.)

The only cases this doesn't handle are 3 -> 2 view simplifications where both views are new, example:
```python
ShapeTracker(views=(View(shape=(5, 5, 5), strides=(-10, 45, 1), offset=40, mask=None, contiguous=False), View(shape=(29, 5, 1), strides=(1, 25, 0), offset=-2, mask=((2, 27), (0, 5), (0, 1)), contiguous=False), View(shape=(72, 1), strides=(1, 0), offset=50, mask=None, contiguous=False)))
ShapeTracker(views=(View(shape=(5, 5, 5), strides=(45, 1, -10), offset=40, mask=None, contiguous=False), View(shape=(72, 1), strides=(1, 0), offset=40, mask=None, contiguous=False)))
```
You would also get a shapetracker equality check out of this method as well, idk if that's useful.

Obviously needs cleanup.

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -10,10 +10,6 @@ from tinygrad.helpers import PICKLE_BUFFERS, SPLIT_REDUCEOP, DEBUG
 if TYPE_CHECKING:
   from tinygrad.shape.shapetracker import ShapeTracker
   from tinygrad.device import Buffer
-import numpy as np
-import highspy
-
-highs_inst = highspy.Highs()
 
 # wrapper around IntEnum that preserves Enum.__str__ and makes auto() unique across all FastEnum subclasses
 class FastEnum(IntEnum):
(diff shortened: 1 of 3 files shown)


chenyuxyz:
stale

PR #8801, opened 2025-01-29
`tc_select` noop [pr]

ignaciosica:


chenyuxyz:
hmm, i guess pr won't run. i think it's fine as long as there's no functional change

ignaciosica:
should I remove pr tag and rerun?

chenyuxyz:
yea that works

ignaciosica:
I'm not allowed to rerun

tinygrad/codegen/kernel.py
```diff
@@ -353,9 +359,12 @@ def apply_opt(self, opt:Opt, append_opt:bool=True):
 
     if opt.op is OptOps.TC:
       check(len(self.applied_opts) == 0, "tensor core opts must be first") # TODO: things like PADTO might be fine
-      check(opt.axis is not None and opt.arg is not None, "tensor core opts must have an axis and arg")
       check((use_tensor_cores:=USE_TC.value) == 2 or len(self.opts.tensor_cores) > 0, "must have tensor cores or TC=2")
-      check(self._apply_tc_opt(use_tensor_cores, cast(int, opt.axis), cast(int, opt.arg)), "no tensor core available")
+      check(opt.axis is not None, "tensor core opts must have an axis")
```
chenyuxyz, line 363: does not need to be this pr, but having a function that checks the TC opt.arg can avoid many repetitive casts here

ignaciosica: Yes, you are right. I wasn't really convinced about all that part, going to clean it up.

ignaciosica:
happened to me before, removing pr tag does not always deactivate pr if it was on before

Qazalin:
yea I think github actions env doesn't update unless you push a new commit.

ignaciosica:
Ah, nevermind then, I'll just wait for a commit to master

PR #8806, opened 2025-01-29
Better stable sigmoid

S-Lykles:
Uses `1/(x*x) -> 1/x * 1/x`  together with `x/(1+x) -> 1-1/(1+x)` to rewrite sigmoid instead of `x/((x+1)(x+1)) -> 1/(x+1)*(1-1/(x+1))`

The last rule is what fixes `test_sigmoid_alt_extreme`, would not be needed if float multiplication distributed over addition


S-Lykles:
It also need 1/x*1/x to not be rewritten to 1/(x*x). But this combination of rules is also more stable to overflow and should require the same amount of operations

chenyuxyz:
cool thanks, will push to benchmark after i update the db version and merge if all looks good

PR #813, opened 2023-05-27
add retinanet with resnet backbone

sohaiberrabii:
- adds resnext to support loading retinanet pretrained on openimages
- object detection post processing with numpy. Some parts could easily be written with tinygrad ops but im not sure how useful that would be. imo it should stay decoupled from forward until the whole postprocessing is part of the graph.
- Just like the reference implementation i used **fiftyone** to download the dataset and convert it to coco format (probably overkill). And **pycocotools** for mAP evaluation.

evaluation result:
```
Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.375
Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.524
Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.405
Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.024
Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.124
Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.414
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.420
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.597
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.623
Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.077
Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.326
Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.675
```

PR #8807, opened 2025-01-29
Add multi device support for RetinaNet eval

francislata:
# Overview
This PR adds multi device support when running RetinaNet eval. Here's a screenshot of the command that I ran:
```bash
time BS=42 BASE_DIR=/raid/datasets/openimages/ PYTHONPATH=. MODEL=retinanet GPUS=6 python3 examples/mlperf/model_eval.py
```
Screenshot 2025-01-29 at 16 20 27

### DIFF
--- a/examples/mlperf/model_eval.py
+++ b/examples/mlperf/model_eval.py
@@ -83,40 +83,54 @@ def eval_retinanet():
   # RetinaNet with ResNeXt50_32X4D
   from extra.models.resnet import ResNeXt50_32X4D
   from extra.models.retinanet import RetinaNet
-  mdl = RetinaNet(ResNeXt50_32X4D())
-  mdl.load_from_pretrained()
-
-  input_mean = Tensor([0.485, 0.456, 0.406]).reshape(1, -1, 1, 1)
-  input_std = Tensor([0.229, 0.224, 0.225]).reshape(1, -1, 1, 1)
-  def input_fixup(x):
-    x = x.permute([0,3,1,2]) / 255.0
-    x -= input_mean
-    x /= input_std
-    return x
-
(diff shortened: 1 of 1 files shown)


chenyuxyz:
does it have the same eval metrics? also what's the wall time speed improvement

francislata:
it has the same eval metrics as the PR that introduced #813.

for the multi device changes, i got the following times
real    123m52.186s                                                                                                                                                                                                                                                   
user    50m10.836s                                                                                                                                                                                                                                                    
sys     73m46.258s

for the one in `master`, i'm currently doing a run to get the times. i can follow-up with it shortly once it's available.

francislata:
using `time BASE_DIR=/raid/datasets/openimages/ PYTHONPATH=. MODEL=retinanet python3 examples/mlperf/model_eval.py` with `BS=8`, i get the following:
```
real    126m13.321s
user    50m16.997s
sys     76m1.050s
```

not that much worse than the multi device one. i'll test it with the same batch size now.

chenyuxyz:
i'd expect wall time to reduce a lot (like 5.5X or better) with 6 gpus? or is the bottleneck not on gpu at all

francislata:
it looks like step times between the multi-device vs. `master` is almost similar:
**multi-device**:
Screenshot 2025-01-29 at 22 01 55

**master**:
Screenshot 2025-01-29 at 22 03 34

for the same batch size of 24.

it's odd that multi device isn't that much faster than the single GPU case.

chenyuxyz:
if that's the case, i would recommend figuring out why the time is similar and where the bottleneck is. it's not worth the complexity

francislata:
i'll reopen this once i have a correctness training run for retinanet.

PR #8812, opened 2025-01-30
Fix unclosed mmap in disk device

FICTURE7:
The disk device will now close the mmapped file when being closed. This is needed on Windows where operations on the mmapped file will result in "file in use" complaints.

This fixes some of the tests in `test_disk_tensor`, but modifications to some tests were needed to make them work on Windows. Mainly freeing the mmapped file before operating on it again in some cases.

This gets `test_disk_tensor` to pass on Windows.

Before:
image

After:
image


test/unit/test_disk_tensor.py
```diff
@@ -49,7 +49,7 @@ def test_readinto_read_speed(self):
       with Timing("copy in ", lambda et_ns: f" {test_size/et_ns:.2f} GB/s"):
         f.readinto(tst)
   def test_bitcasts_on_disk(self):
-    _, tmp = tempfile.mkstemp()
```
FICTURE7, line 52: changed to `temp('bitcast')` because `tempfile.mkstemp` returns an extra file handle which needs to be closed

FICTURE7:
This causes a nifty free after use style ignored exception in `examples\beautiful_minst.py`.

```pytb
Exception ignored in: <function Buffer.__del__ at 0x000001F9121E82C0>
  File "C:\Users\User\source\repos\tinygrad\tinygrad\device.py", line 144, in __del__
    def __del__(self): (not self.is_allocated()) or self.deallocate()
                                                    ^^^^^^^^^^^^^^^^^
  File "C:\Users\User\source\repos\tinygrad\tinygrad\device.py", line 131, in deallocate
    self.allocator.free(self._buf, self.nbytes, self.options)
  File "C:\Users\User\source\repos\tinygrad\tinygrad\device.py", line 177, in free
    def free(self, opaque, size:int, options:Optional[BufferSpec]=None): self._free(opaque, options if options is not None else BufferSpec())
                                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\User\source\repos\tinygrad\tinygrad\runtime\ops_disk.py", line 83, in _free
    def _free(self, opaque, options): self.dev._might_close()
                                      ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\User\source\repos\tinygrad\tinygrad\runtime\ops_disk.py", line 41, in _might_close
    if self.mem is not None: self.mem.close()
```

geohot:
Any progress on this? Can you fix the use-after-free?

PR #8819, opened 2025-01-30
feat: added the ability to load video if webcam is not available

rcastill:
This feature is for yolov8 webgpu example


geohot:
Is there a test for this?

rcastill:
Not really. What kind of test do you have in mind?

This behavior takes place mainly in the browser.

geohot:
We don't have a good enough way to test this. I'll close for now, if you want it, include a link to a deployed instance and I'll test it.

PR #8821, opened 2025-01-31
[bounty] Make Tensor inherit from MathTrait, delete SimpleMathTrait

bautista-garcia:
- Merged `MathTrait` and `SimpleMathTrait` into a single `MathTrait`.  
- Removed operator overloads for `<<` (lshift) and `>>` (rshift) from `tensor.py`, now handled by `MathTrait`.  
- Updated `maximum` and `minimum` implementations to use `MathTrait` ops.  
- `lshift` and `rshift` could not be fully refactored due to a `ulong` error in `test/test_randomness.py`.  
  - These, along with `idiv` and `mul`, follow the same broadcasted binary op process, but `OPS.SHL & OPS.SHR` end up casting to `ulong`.  

Let me know if you need me to change/add anything.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -3255,7 +3255,7 @@ class Tensor(MathTrait):
     if self.dtype != dtypes.bool and not dtypes.is_int(self.dtype): raise RuntimeError(f"{self.dtype} is not supported")
     return self.logical_not() if self.dtype == dtypes.bool else self ^ -1
 
-  def lshift(self, x:int, reverse=False) -> Tensor:
+  def lshift(self, x:int) -> Tensor:
     """
     Computes left arithmetic shift of `self` by `x` bits. `self` must have unsigned dtype.
     Equivalent to `self << x`.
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -3255,7 +3255,7 @@ def bitwise_not(self) -> Tensor:
     if self.dtype != dtypes.bool and not dtypes.is_int(self.dtype): raise RuntimeError(f"{self.dtype} is not supported")
     return self.logical_not() if self.dtype == dtypes.bool else self ^ -1
 
-  def lshift(self, x:int):
+  def lshift(self, x:int, reverse=False):
```
geohot, line 3258: No

bautista-garcia: Sorry George, was trying to make lshifht and rshift work with OPs.SHL and OPs.SHR with reverse option and left it there when reverted back to idiv and mul implementation. 

Do you want me to make reverse work with mul and idiv implementation?

geohot: If I have to manage at this level, it's not worth the bounty. This change is obviously wrong.

bautista-garcia: Sorry for that, will take the feedback for next contrbutions

PR #8823, opened 2025-01-31
Reopened "move llvm_bf16_cast to renderer for CLANG and LLVM" [pr]

mesozoic-egg:
Reopening #8720 

Merging that failed with this error https://github.com/tinygrad/tinygrad/actions/runs/13007184087/job/36276554066
```
<null>(4): error: duplicate parameter name
  extern "C" __global__ void __launch_bounds__(32) r2_5_8_2_16_128_4_4(nv_bfloat16* data0, nv_bfloat16* data1, float* data2, unsigned short* data3, nv_bfloat16* data3, float* data4) {
                                                                                                                                                                 ^
```

But I'm actually not able to reproduce it, even on tiny18. I'm actually not sure how it would lead to a duplicatee parameter. It seems more related to some rendering code elsewhere.

Screenshot 2025-01-31 at 11 31 33

Maybe just retry the merge?


mesozoic-egg:
There was this failure in https://github.com/tinygrad/tinygrad/actions/runs/13065912346/job/36458135628
```
FAILED test/test_fusion_op.py::TestFusionOp::test_recursive_pad - AssertionError: 2.5171995420000712 not less than 2.0
```

It might be related to the runner itself? re-running it works

bautista-garcia:
I've been investigating the issue in #8823 (Reopened: "Move llvm_bf16_cast to renderer for CLANG and LLVM"). While I couldn't fully resolve it, I found some insights that might help clarify the problem and potential causes.

### **Problem Overview**

The issue appears to be a failed benchmark run on **tinybox green** when executing **hlb\_cifar10** with **BF16 support** enabled. The kernel generates duplicate parameter names (`nv_bf16* data0` and `ushort* data0`).

### **1. PatternMatcher Order Impact**

The only runtime change is the **order of the PatternMatcher** in graph rewrite. This could affect how the graph is rewritten:

- In **top\_down\_rewrite()**, the **BF16 cast rule** is now checked **before** the **move mask rule**.
- I didn't find direct conflicts between these rules, but you might notice something I missed.

```python
# Move masks:
(UPat((Ops.LOAD, Ops.STORE), src=(UPat.any(masked_index := UPat(Ops.INDEX, 
    src=(UPat(name="buf"), UPat(name="idx"), UPat(name="mask"))),
    masked_index.cast(None).named("cast")),), allow_any_len=True, name="x"), 
    move_mask),

# BF16 cast:
pat += [(UPat(Ops.CAST, name="root", 
    src=(UPat.load(UPat.index(UPat.var("buf"), UPat.var("idx")), dtype=dtypes.bfloat16),)), 
    bf16_bitcast_to_float),]
```

### **2. Kernel Parameter Naming Issue**

Kernel parameters in `tinygrad/renderer/cstyle.py` are derived from `DEFINE_GLOBAL`:

```python
r[u] = f"data{u.arg}" if u.op is Ops.DEFINE_GLOBAL else u.arg[0]
```

Examining `schedule_uop`, `_append_buf` dynamically assigns buffer arguments based on order, making duplicate **DEFINE\_GLOBALs** unlikely.

```python
def _append_buf(ctx: ScheduleItemContext, x: UOp) -> UOp:
    ctx.bufs.append(x)
    return UOp(Ops.DEFINE_GLOBAL, x.dtype.ptr(size=x.size), (), len(ctx.bufs)-1)
```

### **3. Possible Issue with ****`self.replace`**** Cache**

The **self.replace** cache in `top_down_rewrite` might be matching duplicate buffers, returning **duplicated DEFINE\_GLOBALS** UOPs. However, I'm not entirely convinced by this conclusion, as each **DEFINE\_GLOBAL** appears to have different dtypes (by the kernel codegen error message). 

```python
if (rn := self.replace.get(n)) is not None: return rn
```
I noticed that the bounty for this issue has been moved to archived. Should this change still be pursued, or is it no longer relevant?

PR #8824, opened 2025-01-31
TQDM quality of life

torn8to:
Added logic to tqdm to invert the it/s to s/it or to get the behavior in the image below. In cases where loop iterations take longer than 10/s to provide better iteration feedback to the user.

example 
Screenshot from 2025-01-30 22-49-47


geohot:
Test? Does this match stock tqdm?

torn8to:
yes this one does, I just need to add/change tests to cover this case as the original tests never check this region of the time, the output is checked only when it/s > 1 but its more like it/s > 4000. Which is why this condition was never caught, I need to add tests to cover these conditions that preferably don't use time.sleep().

PR #8827, opened 2025-01-31
tensor uop spec should be in spec.py [pr]

geohot:


Qazalin:
very supportive of `if not skip_check` getting out of the scheduler API!

geohot:
Yea then the post Scheduled ops spec can go in spec.py too.

I imagine scheduler actually rewriting the Tensor UOps to include the KERNELs. realize will remove these once it's actually set the buffer

PR #8830, opened 2025-01-31
start simplifying the schedule context [pr]

Qazalin:
This ScheduleContext can't stay as a global dict passed around everywhere. This diff removes its need for the becomes_map update.
There is a path to removing this global context completely with a few more refactors.

### DIFF
--- a/tinygrad/engine/schedule.py
+++ b/tinygrad/engine/schedule.py
@@ -35,7 +35,7 @@ class ScheduleItem:
 
 @dataclass(frozen=True)
 class ScheduleContext:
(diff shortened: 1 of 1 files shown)


Qazalin:
ScheduleContext will just exist in group_realizes as GrouperContext, defined as:
```
@dataclass(frozen=True)
class GrouperContext:
  assigns: frozenset[UOp]
  realizes: frozendict[UOp, UOp]
  allbufs: frozendict[UOp, UOp]
  children: frozendict[UOp, dict[UOp, None]]
```

PR #8833, opened 2025-01-31
Remove size arg from DEFINE_LOCAL

KhanerX:
image


test/test_uop_graph.py
```diff
@@ -397,7 +397,7 @@ def test_fold_gated_load(self):
 
   def test_fold_gated_load_local(self):
     glbl0 = UOp(Ops.DEFINE_GLOBAL, dtypes.int.ptr(), (), 0)
-    smem = UOp(Ops.DEFINE_LOCAL, dtypes.int.ptr(local=True), (), ("temp", 1))
```
davidjanoskyrepo, line 400: Why refactor the arg and now have mixed type and interpretations in the render? You can leave it as a tuple and just not render the size info twice with just this change:

"""
UOp(Ops.DEFINE_LOCAL, tc.dtype_in.ptr(size=st.real_size(), local=True), (), (f"temp{i + 1}", st.real_size()))
              local_buffer = UOp(Ops.DEFINE_LOCAL, tc.dtype_in.ptr(size=st.real_size(), local=True), (), f"temp{i}")
"""

KhanerX: Hi @davidjanoskyrepo, thanks a lot for your review, can you please clarify where we have mixed type and interpretations that could be avoided by keeping arg as tuple with one value instead of string?

davidjanoskyrepo: "(UPat(Ops.DEFINE_LOCAL, name="x"),
  --- lambda ctx, x: [f".shared .align 4 .b8 {x.arg[0]}[{x.arg[1]*x.dtype.itemsize}];", f"mov.u64 {ctx.r[x]}, {x.arg[0]}[0];"]),
 +++  lambda ctx, x: [f".shared .align 4 .b8 {x.arg}[{x.dtype.size*x.dtype.itemsize}];", f"mov.u64 {ctx.r[x]}, {x.arg}[0];"]),
...
  (UPat(Ops.DEFINE_VAR, name="x"), lambda ctx, x: f"ld.param.{ctx.mem_types[x.dtype]} {ctx.r[x]}, [{x.arg[0]}+0];"),
])"

Like here. Before it was indexing the tuple with a certain composition but now since it isn't a tuple this has changed in this instance. But further below it is still following the old behaviors which is inconsistent to me. I'm not saying this sort of refactor is good or bad just that:

1.) it isn't the minimum required delta for the task of reducing duplicated information in the print out and generally combining tasks is a bad thing and

2.) the refactor is incomplete to me where some places there is an arg tuple, others just one item for the string.

KhanerX: Thanks for the explanation, you're right, arg remains tuple.

extra/assembly/assembly.py
```diff
@@ -105,7 +105,7 @@ def uops_to_asmstyle(lang, function_name:str, uops:List[UOp]):
   for u in uops:
     uop,dtype,vin,args,_ = u
     if uop == Ops.DEFINE_LOCAL:
-      lang.ins.append(AssemblyInstruction(Ops.DEFINE_LOCAL, None, [], args))
+      lang.ins.append(AssemblyInstruction(Ops.DEFINE_LOCAL, None, [], args+(dtype.size,)))
```
geohot, line 108: Huh? Is this tested?

test/test_uops.py
```diff
@@ -312,7 +312,7 @@ class TestLocalAccess(unittest.TestCase):
   @unittest.skipUnless(Device[Device.DEFAULT].renderer.has_shared, "test requires shared memory")
   def test_local_basic(self):
     uops = []
-    smem = uop(uops, Ops.DEFINE_LOCAL, dtypes.float32.ptr(local=True), (), ('smem', 16))
+    smem = uop(uops, Ops.DEFINE_LOCAL, dtypes.float32.ptr(size=16, local=True), (), ("smem",))
```
geohot, line 315: Why is this a tuple?

test/test_uop_graph.py
davidjanoskyrepo: Well seems GH would prefer it mixed I suppose. I'm used to production C++ and the mixed typing is a little odd to me but it is his code so best to follow his style guidance.

extra/assembly/assembly.py
KhanerX: it's not tested. changing arg breaks code in `extra/assembly/assembly_ptx.py` that accesses arg[1].
but I'm not sure those files in extra are used anymore or obsolete.
should I just don't touch it?

test/test_uops.py
KhanerX: My reason was to be consistent with ops like `DEFINE_ACC` that's used like this:
`python
acc = UOp(Ops.DEFINE_ACC, x.dtype, (x.const_like(identity_element(alu_op, x.dtype.scalar())),) + tuple(reduce_range), (ctx.acc_num,))
`
I changed it to string, is string ok?

PR #8646, opened 2025-01-16
Autogen webgpu dawn, removing wgpu-py dependency (f16 support part 1)

wpmed92:
This PR changes the WebGPU runtime used from `wgpu` to `Dawn` (used by Chrome):
- `wgpu` is not closely following the WebGPU specs, it doesn't impose a limit on storage buffers per kernel, but Dawn does, so some tests had to be skipped due to more than the supported number of buffers per kernel
- the python bindings used are created by me, it's published as `dawn-python` (supports both Mac (Metal), and Linux (Vulkan)
- WebGPU specs doesn’t allow using the same buffer for input and output, JIT does this, and was fine on wgpu (non spec compliant). I patch it in ops_webgpu

Here is a demo of `f16` stable diffusion deployed from another branch that includes the `f16` support: https://wpmed92.github.io/stable-diffusion-tinygrad-f16/
I see around 30% perf gain from `f16` in SD.
The `f16` `wgsl` change is coming in a follow-up PR, I didn't want to put these two relatively big changes into 1 PR.

PR #8839, opened 2025-01-31
weights_only=False to prepare for pytorch 2.6 in unit tests

wpmed92:
While working on https://github.com/tinygrad/tinygrad/pull/8646 I found this failing test:

> FAILED test/unit/test_disk_tensor.py::TestTorchLoad::test_load_resnet - RuntimeError: Cannot use ``weights_only=True`` with files saved in the legacy .tar format. In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.

This was because in that PR I modified setup.py (I removed the `wgpu-py` dependency). By doing so, the actions cache key for linting packages (see: `linting-packages-${{ hashFiles('**/setup.py') }}-3.10`) in `Linters+fuzz+unit Tests` changed, which resulted in not the cached 2.5.1 pytorch version being used, but the new 2.6 being installed.
`weights_only=False` solves the aformentioned error. I made this change in the webgpu PR, but I discussed with @chenyuxyz that it's worth a separate PR.

### DIFF
--- a/test/unit/test_disk_tensor.py
+++ b/test/unit/test_disk_tensor.py
@@ -10,7 +10,7 @@ def compare_weights_both(url):
   import torch
   fn = fetch(url)
   tg_weights = get_state_dict(torch_load(fn))
-  torch_weights = get_state_dict(torch.load(fn, map_location=torch.device('cpu'), weights_only=True), tensor_type=torch.Tensor)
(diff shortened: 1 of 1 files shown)


francislata:
thank you for fixing this. i'm also encountering this in my recent CI checks for the retinanet PR.

PR #8844, opened 2025-02-01
add `Tensor.isclose()`

cverrier:
As [discussed on Discord](https://discord.com/channels/1068976834382925865/1070745817025106080/1333669036533354497), let's try to add `Tensor.isclose`. For reference:
- [`torch.isclose`](https://pytorch.org/docs/stable/generated/torch.isclose.html)

I tried to implement some tests with `helper_test_op()`, but I am not sure I did that correctly.

Also, I tried to implement `Tensor.allclose()`, but unit tests were failing, which I think is due to the fact that `torch.allclose()` returns a `bool` instead of a `torch.Tensor`... So I left that for later. In case you absolutely want to do an `allclose()`, then just do `Tensor.isclose().all()`.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -1714,6 +1714,25 @@ class Tensor(SimpleMathTrait):
     """
     return self.logical_not().any(axis, keepdim).logical_not()
 
+  def isclose(self, other:Tensor, rtol:float=1e-05, atol:float=1e-08) -> Tensor:
+    """
+    Returns a new tensor with element-wise comparison of closeness to `other` within a tolerance.
+
+    The `rtol` and `atol` keyword arguments control the relative and absolute tolerance of the comparison.
+
+    Two `NaN` values are not close to each other.
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t.numpy())
+    ```
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t.isclose(Tensor([1, 2, 3.1])).numpy())
+    ```
+    """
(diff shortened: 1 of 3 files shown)


tinygrad/tensor.py
```diff
@@ -1714,6 +1714,25 @@ def all(self, axis:Optional[Union[int, Sequence[int]]]=None, keepdim=False):
...
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t.numpy())
+    ```
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor([1, 2, 3])
+    print(t.isclose(Tensor([1, 2, 3.1])).numpy())
+    ```
+    """
+    return ((self - other).abs() <= atol + rtol * other.abs()) * (self.isnan() | other.isnan()).logical_not()
```
cverrier, line 1734: By default, PyTorch does not assume that `NaN` values are equal. Although they have a `equal_nan` option to modify this behavior, I am not sure you want to introduce that based on [a recent poll](https://x.com/__tinygrad__/status/1865279121950970360).

geohot: We are keeping NaN, and should match PyTorch here

cverrier:
It seems that the only reason some GitHub workflows are failing is https://github.com/tinygrad/tinygrad/blob/6bcd2e5fb5577041977a1f3ac608a8320519b8f9/test/test_ops.py#L1264

@geohot, do you know what's happening? Is the way I coded the test not correct? Locally, when I run
```python
from tinygrad import Tensor
import torch
x = torch.Tensor([float("nan"), 1.0])
y = Tensor([float("nan"), 1.0])
print(torch.isclose(x, x, equal_nan=True))  # result: `tensor([True, True])`
print(y.isclose(y, equal_nan=True).numpy())  # result: `[ True  True]`
```
it works as expected.

geohot:
Not sure, but as long as tests are failing this is hard to review

PR #8847, opened 2025-02-01
Tensor uses mathtrait (except mod)

eliotgolding:
Is this correct?
- Added `MathTrait._opfix` which makes operands compatible; for UOp and UPat this is basically just const_like. For Tensor this does broadcasting and dtype casting.
- `Tensor._apply_uop` amended to implement `MathTrait.alu`
- Tensor inherits ops from MathTrait and docs moved over, except for `mod` because Tensor.mod adds a factor to fix the sign. I think this is a separate issue, similar to the idiv/truncdiv issue, and the fix shouldn't be in Tensor anyway?
- Threefry doesn't rewrite longs with new Tensor bit shifts, so use mul/idiv in `Tensor._threefry_random_bits` for now

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -20,11 +20,11 @@ class FastEnum(IntEnum):
 class MathTrait:
   # required to implement
   def alu(self:T, arg:Ops, *src) -> T: raise NotImplementedError
-  def _opfix(self:T, b:ConstLike|T, _float=False, _bool=False, bitwise=False) -> tuple[T, T]: raise NotImplementedError
+  def _opfix(self:T, b:ConstLike|T|None=None, match_dtype=False, _float=False, _bool=False, bitwise=False) -> tuple[T, T]: raise NotImplementedError
 
   # great functions you get!
   def const_like(self:T, b:ConstLike) -> T: return self._opfix(b)[1]
-  def _binop(self, op, x, reverse, **kwargs):
+  def _binop(self, op, x, reverse=False, **kwargs):
     a, b = self._opfix(x, **kwargs)
     return b.alu(op, a) if reverse else a.alu(op, b) 
   def logical_not(self): return self.ne(True)
@@ -82,7 +82,7 @@ class MathTrait:
 
   def maximum(self, x): return self._binop(Ops.MAX, x)
   def minimum(self, x): return -(-self).maximum(-x)
-  def where(self, x, y): return self.alu(Ops.WHERE, x, x.ufix(y))
+  def where(self, x, y): return self._opfix(x)[0]._opfix(y,_bool=True)[0].alu(Ops.WHERE, *self._opfix(x,match_dtype=False)[1]._opfix(y))
   def threefry(self, seed): return self.alu(Ops.THREEFRY, seed)
   def reciprocal(self): return self.alu(Ops.RECIP)
   def sqrt(self): return self.alu(Ops.SQRT)
@@ -355,11 +355,12 @@ class UOp(MathTrait, metaclass=UOpMetaClass):
   def sink(self, *srcs:UOp): return UOp(Ops.SINK, dtypes.void, (self,)+srcs)
   def detach(self): return UOp(Ops.DETACH, self.dtype, (self,))
   def index(self, idx:UOp, valid:UOp|None=None): return UOp(Ops.INDEX, self.dtype, (self,idx,valid) if valid is not None else (self,idx))
-  def const_like(self, b:ConstLike):
+  def _opfix(self, b:ConstLike|UOp, **kwargs):
+    if isinstance(b, UOp): return self, b
     # constants can optionally have a DEVICE source
-    if self._device is None: return UOp.const(self.dtype, b)
-    if isinstance(self.device, tuple): return UOp.multi(*[UOp.metaop(Ops.CONST, self.shape, self.dtype, d, b) for d in self.device], axis=None)
-    return UOp.metaop(Ops.CONST, self.shape, self.dtype, self.device, b)
+    if self._device is None: return self, UOp.const(self.dtype, b)
+    if isinstance(self.device, tuple): return self, UOp.multi(*[UOp.metaop(Ops.CONST, self.shape, self.dtype, d, b) for d in self.device], axis=None)
+    return self, UOp.metaop(Ops.CONST, self.shape, self.dtype, self.device, b)
   def broadcast(self, count:int):
     assert self.dtype.count == 1
     if count == 1: return self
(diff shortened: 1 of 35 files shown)


geohot:
Bounty locked, this is the best one I've seen. Will fully review tomorrow

tinygrad/ops.py
```diff
@@ -17,27 +17,17 @@ def __str__(self): return Enum.__str__(self)
...
-  def add(self, x, reverse=False): return self._binop(Ops.ADD, x, reverse)
-  def mul(self, x, reverse=False): return self._binop(Ops.MUL, x, reverse)
-  def bitwise_and(self, x, reverse=False): return self._binop(Ops.AND, x, reverse)
-  def bitwise_or(self, x, reverse=False): return self._binop(Ops.OR, x, reverse)
-  def xor(self, x, reverse=False): return self._binop(Ops.XOR, x, reverse)
-  def idiv(self, x, reverse=False): return self._binop(Ops.IDIV, x, reverse)
-  def mod(self, x, reverse=False): return self._binop(Ops.MOD, x, reverse)
-  def sub(self, x, reverse=False): return self.ufix(x).alu(Ops.ADD, -self) if reverse else self.alu(Ops.ADD, self.ufix(-x))
-  def div(self, x, reverse=False): return (self.ufix(x)*self.alu(Ops.RECIP)) if reverse else (self*self.ufix(x).alu(Ops.RECIP))
+  def ufix(self, b): return self._opfix(b)[1]
```
geohot, line 26: Why is this not _opfix?

tinygrad/ops.py
```diff
@@ -17,27 +17,17 @@ def __str__(self): return Enum.__str__(self)
   @staticmethod
   def _generate_next_value_(_, __, ___, last_values): return 1 + max([0, *last_values, *[max(c) for c in FastEnum.__subclasses__()]])
 
-class SimpleMathTrait:
+class MathTrait:
   # required to implement
   def alu(self:T, arg:Ops, *src) -> T: raise NotImplementedError
-  def const_like(self:T, b:ConstLike) -> T: raise NotImplementedError
+  def _opfix(self:T, b:ConstLike|T|None=None, match_dtype=True, _float=False, _bool=False, bitwise=False) -> tuple[T, T]: raise NotImplementedError
```
geohot, line 23: This is extremely complex looking, how can it be simpler?

geohot:
Can you start with a single function so the diff is readable? I'll leave this locked to you.

geohot:
I mean start with just the add function using opfix so the diff is readable

tinygrad/ops.py
```diff
@@ -355,11 +355,12 @@ def axis_arg(self) -> tuple[int, ...]:
   def sink(self, *srcs:UOp): return UOp(Ops.SINK, dtypes.void, (self,)+srcs)
   def detach(self): return UOp(Ops.DETACH, self.dtype, (self,))
   def index(self, idx:UOp, valid:UOp|None=None): return UOp(Ops.INDEX, self.dtype, (self,idx,valid) if valid is not None else (self,idx))
-  def const_like(self, b:ConstLike):
+  def _opfix(self, b:ConstLike|UOp, match_dtype=True):
```
geohot, line 358: Why is match_dtype here? It's not used in the function

eliotgolding: It's to make the fxn signature the same as MathTrait, since in where() MathTrait would call _opfix(match_dtype=False). Could do `**kwargs` and `# type: ignore` if you don't like it. UOp/UPat ignores match_dtype, Tensor doesn't.

const_like used to be the basic function for making args compatible for math functions, which was required to implement. But for tensor you need to broadcast args even if they are both tensors already, so const_like can't be the basic function. Since uop doesn't care about broadcasting, the functionality of opfix is basically the same as const_like.

tinygrad/ops.py
```diff
@@ -17,14 +17,16 @@ def __str__(self): return Enum.__str__(self)
   @staticmethod
   def _generate_next_value_(_, __, ___, last_values): return 1 + max([0, *last_values, *[max(c) for c in FastEnum.__subclasses__()]])
 
-class SimpleMathTrait:
+class MathTrait:
   # required to implement
   def alu(self:T, arg:Ops, *src) -> T: raise NotImplementedError
-  def const_like(self:T, b:ConstLike) -> T: raise NotImplementedError
+  def _opfix(self:T, b:ConstLike|T, match_dtype=True) -> tuple[T, T]: raise NotImplementedError
```
geohot, line 23: I don't understand this function

tinygrad/tensor.py
```diff
@@ -3258,7 +3267,7 @@ def bitwise_not(self) -> Tensor:
     if self.dtype != dtypes.bool and not dtypes.is_int(self.dtype): raise RuntimeError(f"{self.dtype} is not supported")
     return self.logical_not() if self.dtype == dtypes.bool else self ^ -1
 
-  def lshift(self, x:int):
+  def lshift(self, x:int, reverse=False):
```
geohot, line 3270: Why is this here if it does nothing?

eliotgolding: Temporarily for mypy, because you told me to only change add(), as I wrote in a comment below... will be gone if tensor fully inherits mathtrait like in the original pr

eliotgolding:
Maybe I'm not explaining the motivation for this clearly enough. Right now in Tensor, there are methods `_apply_uop`, `_broadcasted` and `_apply_broadcasted_uop`. The way these are defined and used parallels MathTrait's `alu`, `ufix` and `_binop` respectively. This pr makes use of this so that Tensor will inherit all the elementwise op methods from MathTrait, saving lines and simplicity.

The current `a.ufix(b)` returns `b` unchanged if it is already a mathtrait instance. For Tensor we need a more general function, because of broadcasting, which means `a` and `b` might change even if they are already tensors. This is the point of `_opfix`, it makes operands compatible like `ufix` and replaces `const_like` as a basic function for MathTrait subclasses. It returns a tuple `(a, b)` because `a`'s shape might have been modified by broadcasting as well.

For UOp/UPat, we don't want to care about the shape or dtype. So `_opfix` just needs to check that the other arg is not a number literal. This is why it replaces const_like. (More accurately, it replaces the functionality of ufix, hence the `isinstance` checks). For Tensor, `_opfix` is a wrapper around `_broadcasted`.

Why is the `match_dtype` parameter there? That was used in `MathTrait.where()` in the original pr, similar to how it is used in Tensor right now. UOp and UPat ignore this parameter because they don't care about the dtype of the args anyway. I think I will replace _opfix signature with `_opfix(self, b, **kwargs)` as it's less confusing.

tinygrad/ops.py
eliotgolding: See my comment below; apologies if I have been unclear so far

geohot:
Err, there's still match_dtype, and it's still unused. I obviously can't merge that. Can you break this into smaller PRs that provide standalone value?

eliotgolding:
Yeah, I got rid of the part where `match_dtype` was needed, in order to make the diff smaller so you could see the changes to `alu` and `_opfix`. This wasn't meant to be merged right now. I thought you wanted me to do that, then add the other stuff back in once you had reviewed this part.

I will close this and do it as a few smaller prs then. Thanks for your patience.

PR #8854, opened 2025-02-01
hcq pci signal fuzzer

nimlgen:
it looks fine now. will double check that queues built the way that each agent sees prev signal value before setting a new one.


nimlgen:
ONLY_COMPUTE seems fine on AMD/AM. Ugh, will look into sdma....

PR #8855, opened 2025-02-01
Fix disk tensor assignment

FICTURE7:
Before this change, `test_disk_assignment` would fail with: `AssertionError: can't copyin to unallocated buffer`. The disk buffer was not allocated because the scheduler would fold the entire `SINK(CONTIGUOUS(VIEW(BUFFER(DEVICE))))` tree to an isolated `SINK` node, meaning the buffer would not end up in schedule item to be allocated eventually.

~~Not entirely sure if this is the right fix, but `Tensor.data()` will do something similar.~~

### DIFF
--- a/test/test_assign.py
+++ b/test/test_assign.py
@@ -3,6 +3,7 @@ import unittest
 import numpy as np
 from tinygrad import dtypes, Tensor, TinyJit, GlobalCounters, Variable
 from tinygrad.device import is_dtype_supported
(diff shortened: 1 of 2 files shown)


Qazalin:
yea if EMPTY is childless nothing gets allocated for it. This is because `ensure_allocated` is called in realize.py.

FICTURE7:
should be addressed now

PR #8858, opened 2025-02-01
Single file Onnx

geohotstan:
tried many many abstractions and thought maybe decorating everything is best (also took this idea from reading pytorch's symbolic onnx stuff)
keeps the op relevant info like python_const, op domain, and op version close to the op implementation.
downside is more lines 

most of the current ops are naively matching the op version still. That's a todo!
But I think with this abstraction it's at least possible to enforce correct behavior for choosing domain and domain version.


chenyuxyz:
merging into one file is fine, as i would expect that to be the case in tinygrad main dir.

can you separate this into 2 prs, one is a simple file merge and another is whatever abstraction you want to change?

geohotstan:
2 PRs

PR #8859, opened 2025-02-01
TIP 12: spec for allowed UOps in a ScheduleItem

Qazalin:
Because this AST is captured by the JIT, we need a spec that defines UOps allowed to go inside the JIT.

This TIP proposes allowing for 3 UOp types:

1. COPY(DEVICE, placeholder)
2. STORE(placeholder, st, val)
3. BUFFER_VIEW(placeholder)

The key similarity is the placeholder UOp. BUFFER is not allowed to exist.

COPY and STORE maintain their full meaning, such that the JIT can create replacement bufs from just the AST.
sadly, BUFFER_VIEW does not maintain enough information to recreate a subbuffer, it needs a new spec. Maybe SHRINK(placeholder)?

### DIFF
--- a/tinygrad/engine/realize.py
+++ b/tinygrad/engine/realize.py
@@ -3,6 +3,7 @@ import time, pprint
 from dataclasses import dataclass, replace
 from tinygrad.helpers import all_same, colored, getenv, DEBUG, GlobalCounters, ansilen, BEAM, NOOPT, all_int, CAPTURING, Metadata, TRACEMETA
 from tinygrad.ops import Ops, PatternMatcher, UOp, UPat, Variable, sym_infer
+from tinygrad.spec import type_verify, si_spec
 from tinygrad.device import Device, Buffer
 from tinygrad.renderer import Renderer, ProgramSpec, Estimates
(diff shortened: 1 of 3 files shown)


Qazalin:
err, DEFINE_GLOBAL has no DEVICE. So COPY still loses its full meaning after getting in a ScheduleItem.

geohot:
I don't understand this, why is there not BUFFER? You are saying that KERNEL is BUFFER?

Qazalin:
The AST inside KERNEL needs to enumerate all the BUFFERs.
This is because of the method_cache. So BUFFER has to become DEFINE_GLOBAL at some point in rewrite stages.

ASSIGN after linearization is not needed. It just becomes a STORE.

Qazalin:
I replaced this with kernel_spec from https://github.com/tinygrad/tinygrad/pull/8893/files#diff-adba9dc217f63e1e32992a929dbcba526e6ebe433c8116de1dadbcb9ea6e682bR126
Now that we're close to the line limit we need to just delete ScheduleItem.

PR #8862, opened 2025-02-03
Ops.POW

chenyuxyz:


tinygrad/tensor.py
```diff
@@ -3309,20 +3309,8 @@ def pow(self, x:Union[Tensor, ConstType], reverse=False) -> Tensor:
     # positive const ** self
     if not isinstance(x, Tensor) and reverse and x > 0: return self.mul(math.log(x)).exp()
 
-    base, exponent = self._broadcasted(x, reverse=reverse)
```
geohot, line 3312: I would start by adding this as a transcendental and not changing any renderers

chenyuxyz: yea i am doing it that way now, and i have started to clean up pow more before moving it to transcendental

PR #8864, opened 2025-02-03
replace bfs with KERNEL linearizer

Qazalin:


Qazalin:
merged p1 here #8893 - now need to work on ASSIGN as sources next.

PR #8871, opened 2025-02-03
Cleanup llvm cleanup (and some clang things too)

uuuvn:


tinygrad/runtime/ops_llvm.py
```diff
@@ -13,12 +13,14 @@ def expect(x, err, ret=None):
 
 class LLVMCompiler(Compiler):
   def __init__(self, host_arch:str, opt:bool):
-    for component in ['Target', 'TargetInfo', 'TargetMC', 'AsmPrinter']: getattr(llvm, f'LLVMInitialize{host_arch}{component}')()
-    triple = ({'AArch64': 'aarch64', 'X86': 'x86_64'}[host_arch]+'-none-unknown-elf').encode()
+    for component in ['Target', 'TargetInfo', 'TargetMC', 'AsmPrinter']:
```
geohot, line 16: Don't change whitespace!

tinygrad/runtime/ops_llvm.py
```diff
@@ -13,12 +13,14 @@ def expect(x, err, ret=None):
...
+      getattr(llvm, f'LLVMInitialize{host_arch}{component}')()
 
+    triple = {'AArch64': b'aarch64', 'X86': b'x86_64'}[host_arch] + b'-none-unknown-elf'
     target = expect(llvm.LLVMGetTargetFromTriple(triple, ctypes.pointer(tgt:=llvm.LLVMTargetRef()), err:=cerr()), err, tgt)
-    target_machine = llvm.LLVMCreateTargetMachine(target, triple, b'', b'+reserve-x18' if platform.machine() == 'arm64' else b'',
-                                                  llvm.LLVMCodeGenLevelDefault, llvm.LLVMRelocPIC, llvm.LLVMCodeModelDefault)
+    # +reserve-x18 here does the same thing as -ffixed-x18 in ops_clang.py, see comments there for why it's needed on arm osx
+    self.target_machine = llvm.LLVMCreateTargetMachine(target, triple, b'', b'+reserve-x18' if OSX and host_arch == 'AArch64' else b'',
```
geohot, line 22: This needs tests. Why is it only on OSX?

uuuvn: You just don't use x18 register, that's all that this flag does. It is totally safe to use on linux, not safe on macos.

It was implicitly OSX only before cleanup because platform.machine() reports aarch64 on linux and arm64 on macos, now host_arch is always AArch64 regardless of os so OSX here is to preserve that behaviour.

I don't think it makes sense to test for usage of the register, in fact it looks like llvm uses it so rarely that no test in CI or benchmark both used it and got unlucky enough to get context switched at the wrong time for it to cause failure.

If you mean test the "x18 not used on osx arm64" instead of "x18 used on non-osx arm64" then i'm currently trying to find a way to do this elegantly (this is also the thing that broke after cleanup pr, this really needs to be tested to avoid breaking silently in the future)

tinygrad/runtime/ops_llvm.py
uuuvn: this actually reverts whitespace change from #8867, but ok if you want it to be one long line instead of (imo much more readable) two lines.

.github/workflows/test.yml
```diff
@@ -605,7 +605,11 @@ jobs:
           path: /Users/runner/Library/Python/3.12/lib/python/site-packages
           key: mockgpu-osx-${{ matrix.backend }}-packages-${{ hashFiles('**/setup.py') }}
       - name: Set env
-        run: printf "${{ matrix.backend == 'amd' && 'MOCKGPU=1\nAMD=1\nFORWARD_ONLY=1' }}" >> $GITHUB_ENV
+        run: printf "${{ matrix.backend == 'llvm' && 'LLVM=1' || matrix.backend == 'clang' && 'CLANG=1' || matrix.backend == 'amd' && 'MOCKGPU=1\nAMD=1\nFORWARD_ONLY=1' }}" >> $GITHUB_ENV
+      - name: Install llvm
+        if: matrix.backend == 'llvm'
+        run: |
+          brew install llvm uuuvn/tinygrad/tinyllvm
```
geohot, line 612: I really don't like depending on this. I'm okay with no METAL + LLVM

.github/workflows/test.yml
```diff
@@ -585,7 +585,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        backend: [amd]
+        backend: [amd, clang, llvm]
```
geohot, line 588: Too many runners

PR #8873, opened 2025-02-03
ptx wmma render refactor

ignaciosica:
PTX wmma rendering refactor to work with half.

**Overview**
Added packing/unpacking for the wmma registers in order for them to work with half

Standalone prerequisite for #8680 only refactoring rendering.


Qazalin:
did you expect [pr] to pass?

ignaciosica:
No, it shouldn't have passed

ignaciosica:
Benchmark jobs with real devices will fail, as now it uses temp registers for input and output for packing/unpacking

PR #8875, opened 2025-02-03
delete fold_img_cast [pr]

Qazalin:
Need to see if this is actually doing anything now after the tensor_map simplifier folds image to image casts.
As long as comma doesn't regress we don't need this extra complexity. Otherwise these kinds of things have to be tested.


Qazalin:
passes comma process replay, out goes 6 lines of unneeded complexity.
image

PR #8881, opened 2025-02-04
Add OnnxOp and _OpRegistry to handle onnx ops

geohotstan:
follows https://github.com/onnx/onnx/blob/main/docs/Versioning.md#operator-versioning

debug now looks like
```
0: op 'ai.onnx.Conv:1' opt {'dilations': (1, 1), 'group': 1, 'kernel_shape': (3, 3), 'pads': (1, 1, 1, 1), 'strides': (2, 2)}                                                        
1: op 'ai.onnx.BatchNormalization:1' opt {'epsilon': 9.999999747378752e-06, 'is_test': 1, 'momentum': 0.8999999761581421}     
...                                                       
```

Errors should also now give a good overview of what we support and what we don't support when running 100 huggingface

We default to a min supported version now `since_version:int=1`. Will have to go through each op later. Big chore.
Gonna have to write more than the current onnx end to end tests for this. Once that's done maybe we can release the restriction of `SupportedDomains` so that we can support custom ops as well.

Do we want this? Seems like much more stuff to maintain


chenyuxyz:
feel free to use discord more to discuss high level stuff like this before working on the pr, i think it can help you to iterate faster.

i don't have strong feeling on this, mostly because it looks like a lot of work, and i don't know how likely we would have version issues. maybe we can include a versioned onnx runtime test in ci at some point.

i would suggest finishing the script to check coverage for top 100 models first, then we decide the next steps. it's fine if the error message is not perfect (just showing what is not supported to start with)

geohotstan:
stale.
Might take 1 or 2 ideas from this and open a new PR.

PR #8885, opened 2025-02-04
Don't rewrite idiv to rshift when numerator is negative

eliotgolding:
This rewrite would be correct if the idiv behaviour was floor-division (like torch and np) and not trunc-division. But right now it leads to inconsistent behaviour.

The test on PTX failed because it checked that floordiv(dereferenced int ptr) was rewritten to shift, but the int might have been negative. I changed it to use uint so that the new condition is satisfied.


test/test_uops.py
```diff
@@ -355,12 +355,12 @@ def test_bitshift_left(self):
     self.assertIn(Ops.MUL, ops)
 
   def test_bitshift_right(self):
-    g1 = UOp(Ops.DEFINE_GLOBAL, dtypes.int32.ptr(), (), 0)
-    c1 = UOp(Ops.CONST, dtypes.int, (), 2)
-    c2 = UOp(Ops.CONST, dtypes.int, (), 3)
-    l1 = UOp(Ops.LOAD, dtypes.int, (g1.index(c1),))
-    a1 = UOp(Ops.IDIV, dtypes.int, (l1, c1))
-    a2 = UOp(Ops.IDIV, dtypes.int, (l1, c2))
+    g1 = UOp(Ops.DEFINE_GLOBAL, dtypes.uint32.ptr(), (), 0)
```
geohot, line 358: Why change this test?

PR #8886, opened 2025-02-04
copy is already contiguous

chenyuxyz:
can use this to test the pr schedule diff thing, also would like to investigate where in multi is calling that contiguous


Qazalin:
yea I'm gonna merge pr schedule soon, need to figure out an instability in the toposort order of the comma benchmark.

chenyuxyz:
shard calls contiguous because this partition that's on the original device might be a view, but it's not needed for other devices

chenyuxyz:
a few % faster for all training on 6 gpu benchmarks

PR #8890, opened 2025-02-04
add onnx_helpers to extra and add ort validate to benchmark_onnx

geohotstan:
Should we try to have 1 place to store helper functions around onnx usage?
kind of a prereq to 100 huggingface, but I can make do without this too. 
`get_example_inputs` will get crazy with 100 huggingface

made the changes minimal for now so I don't accidentally write regressions for geohot's quantize stuff lolol.
onnx_helpers isn't used in many places yet. 
Ideally all the onnx related tests and external_model_benchmark can reuse stuff in here but that's gonna take some work.

### DIFF
--- a/extra/onnx_helpers.py
+++ b/extra/onnx_helpers.py
@@ -1,4 +1,4 @@
-from tinygrad import Tensor, fetch
+from tinygrad import Tensor
 from tinygrad.tensor import _to_np_dtype
 from extra.onnx import OnnxRunner, OnnxValue
 import onnx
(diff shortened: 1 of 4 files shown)


extra/onnx_helpers.py
```diff
@@ -0,0 +1,33 @@
...
+
+  ort_options = ort.SessionOptions()
+  ort_options.log_severity_level = 3
+  ort_sess = ort.InferenceSession(onnx_file, ort_options, ["CPUExecutionProvider"])
+  np_inputs = {k:v.numpy() if isinstance(v, Tensor) else v for k,v in inputs.items()}
+  out_names = list(run_onnx.graph_outputs)
+  out_values = ort_sess.run(out_names, np_inputs)
+  ort_out = dict(zip(out_names, out_values))
+
+  assert len(tinygrad_out) == len(ort_out) and tinygrad_out.keys() == ort_out.keys()
```
chenyuxyz, line 29: small nits. if `tinygrad_out.keys() == ort_out.keys()` they will have same len too, and if `tiny_v is None` then `tiny_v == onnx_v` is just `onnx_v is None`

PR #8761, opened 2025-01-27
Add Windows tests to GitHub CI

b1tg:

PR #8896, opened 2025-02-04
cherry-pick reverted "Use full soname for libgcc_s in CPUProgram (#8642)"

peinguim:
I encountered the same issue as #8642 and discovered that #8761 had reverted it, which seems unintentional.


geohot:
Can you add a test that would have caught this?

PR #8899, opened 2025-02-05
rename to KernelContext and move the linearize_sched comment [pr]

Qazalin:
The scheduler creates KERNEL UOps and update the becomes_map in the same loop.
Then, it'll give the sched_sink to the schedule linearizer.


test/external/process_replay/process_replay.py
```diff
@@ -19,7 +19,7 @@
 os.environ["CAPTURE_PROCESS_REPLAY"] = "0"
 early_stop = multiprocessing.Event()
 logging.basicConfig(level=logging.INFO, format="%(message)s")
-MAX_LINES = 1_000
```
Qazalin, line 22: 1K is still too much and github log breaks

PR #8904, opened 2025-02-05
hotfix bug in `get_kernel_actions` after `TC_SEARCH_OVER_SHAPE` was introduced

ignaciosica:
I think some benchmarks might have worsen based on this bug

### DIFF
--- a/tinygrad/engine/search.py
+++ b/tinygrad/engine/search.py
@@ -102,7 +102,8 @@ def bufs_from_lin(lin:Kernel, allocate:bool=True) -> list[Buffer]:
 
 # get dictionary of all possible actions
 def get_kernel_actions(lin:Kernel, include_0=True) -> dict[int, Kernel]:
-  acted_lins, max_up, max_lcl, kernel_actions = {0:lin} if include_0 else {}, getenv("BEAM_UPCAST_MAX", 256), getenv("BEAM_LOCAL_MAX", 1024), actions
+  acted_lins, max_up, max_lcl = {0:lin} if include_0 else {}, getenv("BEAM_UPCAST_MAX", 256), getenv("BEAM_LOCAL_MAX", 1024)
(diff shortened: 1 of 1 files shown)


tinygrad/engine/search.py
```diff
@@ -112,7 +112,7 @@ def get_kernel_actions(lin:Kernel, include_0=True) -> dict[int, Kernel]:
 
   for i,a in enumerate(kernel_actions):
     if a.axis is not None and a.op is not OptOps.TC:
-      if ((ax:=a.real_axis(lin)) >= lin.shape_len) or (lin.full_shape[ax] == a.arg and Opt(a.op, ax, 0) in actions): continue
+      if ((ax:=a.real_axis(lin)) >= lin.shape_len) or (lin.full_shape[ax] == a.arg and Opt(a.op, ax, 0) in kernel_actions): continue
```
chenyuxyz, line 115: hmm do you need to make a copy of `kernel_actions` form `actions`? otherwise it might mutate the global ish `actions`

chenyuxyz: line 105, somehow i cannot comment on that line

ignaciosica: I did that in order to remove the search over the default tc, otherwise, if I only added the actions with specific shapes, it will also make runs with default shape

chenyuxyz: my point was that the current code is not making copy, and i think you want to make a copy

ignaciosica: Ohh, I completely missed that, I thought it was, I'll fix it

ignaciosica: fixed

chenyuxyz:
can you also add a test to catch this?

ignaciosica:
> can you also add a test to catch this?

added in #8907

PR #8915, opened 2025-02-05
Hi :) tiny windows change to fix null ptr ref

stokrotkagucia:
in abstactions2.py


geohot:
Ideally I'd like to find a clang flag that does this.

PR #8863, opened 2025-02-03
construct the sched sink [PR]

geohot:

PR #8925, opened 2025-02-06
construct the schedule sink 2

Qazalin:
There are two key differences with #8863:

- This diff removes the need for a double BFS.
Instead, it constructs the KERNEL ops directly from the big_sink. The algorithm is similar to `append_to_block`.
- `PRELOAD` is removed.

The decision for which UOps to realize comes from the `realize_map`. This step is separate from append_to_kernel.


test/test_assign.py
```diff
@@ -120,15 +120,15 @@ def test_assign_changes(self):
 
   def test_assign_diamond_cycle(self):
     # NOTE: should *not* raise AssertionError from numpy
-    with self.assertRaisesRegex(RuntimeError, "cycle"):
+    with self.assertRaisesRegex(RecursionError, "maximum recursion depth exceeded"):
```
geohot, line 123: You should explicitly detect this still

Qazalin:
Explicitly checking for cycles makes things slower, to the point that multi GPU resnet doesn't train.
https://github.com/tinygrad/tinygrad/actions/runs/13228489382/job/36922384624?pr=8925#step:17:10
It goes to a deeper problem of why we're calling schedule so many times for multi.

PR #8933, opened 2025-02-06
Flash attention

Tzvetomir-Kaydjiev:
This is the 'FlashAttention-2 forward pass' algorithm from the [https://arxiv.org/abs/2307.08691](https://arxiv.org/abs/2307.08691) paper. I have changed the `./example/gpt2.py` demo to use `Tensor.flash_attention` and it seems to work, also I have made a sort `flash_atten.py` script to compare `scaled_dot_product_attention` and `flash_attention`.

I know that this patch is a sketch/prototype, but I want to ask a few questions:

 - Am I on the right path to adding the implementation of the algo?

 - Is there a way to add the backward pass, without stitching the computation graph with python?
( For example [torch](https://pytorch.org/docs/stable/autograd.html#torch.autograd.Function) has a way to control the backward pass. As far as I understand this is against the idea of tinygrad. )

 - Do you expect this to be implemented on the backend ( `LLVM`, `CLANG`, `CUDA`, ... ) and added as an operation in OPs?
( I guess attention is quite important function, but I doubt. )

```
(p_venv) [tinygrad]$ python flash_atten.py
7681395.5
flash attention first run time : 1.4955716133117676

7681396.5
normal attention first run time : 0.8573811054229736

7680352.5
flash attention second run time : 0.5276808738708496

7680366.5
normal attention second run time : 0.8197903633117676

error of the flash attention 0.00675% (casual)
error of the flash attention 0.00306% (with mask)

Trying normal attention on a big sequence
failure: rm_alloc returned 81: Out of memory
Trying flash attention on a big sequence
success
```

Pls, give feedback.

### DIFF
--- a/examples/gpt2.py
+++ b/examples/gpt2.py
@@ -44,7 +44,14 @@ class Attention:
       values = xv
 
     xq, keys, values = xq.transpose(1, 2), keys.transpose(1, 2), values.transpose(1, 2)
-    return self.c_proj(xq.scaled_dot_product_attention(keys, values, mask).transpose(1, 2).reshape(bsz, seqlen, self.dim))
+    old_res = xq.scaled_dot_product_attention(keys, values, mask)
+
+    padding = Tensor.zeros(xq.shape[0], xq.shape[1], (16 - (xq.shape[2]) % 16) % 16, xq.shape[3])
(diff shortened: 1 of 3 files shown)


geohot:
Err, this should be done at a much lower level. tbh, these bounties aren't that doable today without more infrastructure, they are labelled "hard" for a reason

PR #8937, opened 2025-02-06
[bounty] Fix LLVM=1 NO_DEVECTORIZE=1 python3 test/test_ops.py TestOps.test_strided_conv2d_simple

bhavyagada:
Bugfix bounty 

Fix 'LLVM=1 NO_DEVECTORIZE=1 python3 test/test_ops.py TestOps.test_strided_conv2d_simple'


geohot:
Hmm, I don't understand this, but I'll lock the bounty.

Can you add a test that explicitly tests both this fix (with an explanation) and the NO_VECTORIZE in the bounty.

geohot:
I added the failing test here #8940

Add a second test that explicitly tests this fix for all platforms and we should be good. I still don't understand how the non DEVECTORIZE version works.

geohot:
ChatGPT believes in you. https://chatgpt.com/share/67a56511-bbdc-8000-87ab-df833875f818

geohot:
Bounty is yours! I'm still curious about an explanation of why this worked in DEVECTORIZE mode, but it's yours regardless. george@tinygrad.org to claim, can pay via PayPal or USDC on ETH.

bhavyagada:
I emailed you. Thanks!

PR #8945, opened 2025-02-07
[bounty] Fix test_avg_pool3d_failure

Neulus:
* Fixed test_avg_pool3d_failure via reducing adds on pool(ones_like())
* Removed expectedFailure


geohot:
Updated bounty to "Fix TestOps.test_avg_pool3d_failure by fixing the issue in the linearizer, not with workarounds" and raised to $200

PR #8947, opened 2025-02-07
Add `Tensor.scatter_reduce`

joshiemoore:
I saw your request to clean up warnings in `test/test_ops.py`. This PR fixes the two occurrences of
```
UserWarning: The reduce argument of torch.scatter with Tensor src is deprecated and will be removed in a future PyTorch release. Use torch.scatter_reduce instead for more reduction options.
```
in `test_scatter_add` and `test_scatter_mul`.

In case you wonder why I didn't replace every usage of `torch.scatter` with `torch.scatter_reduce`, the answer is that `scatter` is only deprecated for the usage of the `reduce` arg with a Tensor `src`, as the warning indicates ([docs](https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_.html)). It's not deprecated when you're passing scalar values. `scatter_reduce` also doesn't support scalar values like `scatter` does, so I left these ones alone. 

I plan on cleaning up the `std()`/`var()` warnings too, but those are separate issues that I'll give their own PR.


geohot:
Err, yea but now the API doesn't match and you can't look at the test and know it's correct...

joshiemoore:
> Err, yea but now the API calls don't match and you can't look at the test and know it's correct...

Thanks, I get what you mean. I have to do more digging to verify, but my initial impression is that pytorch's newer `scatter_reduce` is mostly the same as `scatter` except `scatter_reduce` doesn't support scalars and it has more reduction operations. So I could add a `scatter_reduce` to tinygrad's tensor which would mostly be an alias for the existing `scatter` but it would also implement the additional reduction operations.

Does this sound like a good approach to you? Not sure if you want me to go that far just to silence a couple deprecation warnings, but if so, I'm game.

joshiemoore:
In summary, I pulled the preprocessing logic for src and mask out into the `_pre_scatter` helper method. I then implemented `scatter_reduce` like [pytorch has](https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_reduce_.html#torch.Tensor.scatter_reduce_) with new reduction ops and `include_self` logic, and added tests.

It may be controversial the way I removed support for using `reduce` with a Tensor `src` from `scatter`. pytorch just has the deprecation warning, but I figured we can rip the bandaid off now since tinygrad is still alpha. Let me know if you guys would prefer something different.

I also updated `ScatterElements` in the onnx backend to use `scatter_reduce` as appropriate, so `ScatterElements` now also supports min and max reductions as a result of this change. I would have done this in a separate PR, but the breaking change to `scatter` broke the onnx tests so I just did it here.

geohot:
@chenyuxyz can you review this?

chenyuxyz:
i think this can be more unified with other reduce mapping and identity element, but fine to merge

PR #8949, opened 2025-02-07
Try to open llvm first when opening metal

uuuvn:

PR #8950, opened 2025-02-07
Fix incorrect __del__

uuuvn:
CPython doesn't make any guarantees about order in which globals like `msg` or `libobjc` are destroyed when the interpreter shuts down

https://github.com/tinygrad/tinygrad/pull/8949 triggered the unlucky ordering which lead to a bunch of errors at exit

There is also a bunch of other places where similar problems exist

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -9,7 +9,11 @@ class objc_id(ctypes.c_void_p): # This prevents ctypes from converting response 
   def __eq__(self, other): return self.value == other.value
 
 class objc_instance(objc_id): # method with name "new", "alloc" should be freed after use
-  def __del__(self): msg("release")(self)
+  def __del__(self):
+    # CPython doesn't make any guarantees about order in which globals (like `msg` or `libobjc`) are destroyed when the interpreter shuts down
+    # https://github.com/tinygrad/tinygrad/pull/8949 triggered the unlucky ordering which lead to a bunch of errors at exit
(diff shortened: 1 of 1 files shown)


geohot:
Which CI run has the errors?

uuuvn:
> Which CI run has the errors? I don't see it looking though some 8949 runs.

checkout #8949 , run literally anything with METAL, get this:
```
(.venv-3.12) [uuuvn@macbook ~/src/tinygrad]% cat fun1.py
from tinygrad import Tensor

Tensor.rand((4,)).contiguous().realize()
(.venv-3.12) [uuuvn@macbook ~/src/tinygrad]% python fun1.py
Exception ignored in: <function objc_instance.__del__ at 0x103aeb100>
  File "/Users/uuuvn/src/tinygrad/tinygrad/runtime/ops_metal.py", line 19, in __del__
TypeError: 'NoneType' object is not callable
Exception ignored in: <function objc_instance.__del__ at 0x103aeb100>
  File "/Users/uuuvn/src/tinygrad/tinygrad/runtime/ops_metal.py", line 19, in __del__
TypeError: 'NoneType' object is not callable
Exception ignored in: <function objc_instance.__del__ at 0x103aeb100>
  File "/Users/uuuvn/src/tinygrad/tinygrad/runtime/ops_metal.py", line 19, in __del__
TypeError: 'NoneType' object is not callable
(.venv-3.12) [uuuvn@macbook ~/src/tinygrad]%
```

I've submitted this as a separate PR because the current usage of `__del__` is a bug even without #8949 , it just triggers the unlucky ordering

PR #8959, opened 2025-02-07
Make tensor use UOp lshift/rshift; delete SimpleMathTrait

eliotgolding:
Previously, Tensor `lshift`/`rshift` just called `.mul(2**x)` or `.idiv(2**x)`. Now it uses `Ops.SHL` and `Ops.SHR`. This lets you do `Tensor << Tensor` for example, which wasn't possible before. Signed ints are also supported.

Once Tensor bit shifts are implemented properly, you can delete SimpleMathTrait.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -8,7 +8,7 @@ from tinygrad.helpers import argfix, make_tuple, flatten, prod, all_int, round_u
 from tinygrad.helpers import IMAGE, WINO, _METADATA, Metadata, TRACEMETA, ceildiv, fetch, polyN, unwrap
 from tinygrad.engine.multi import get_multi_map
 from tinygrad.gradient import compute_gradient
-from tinygrad.ops import smax, smin, resolve, UOp, Ops, sint, Variable, SimpleMathTrait, identity_element
+from tinygrad.ops import smax, smin, resolve, UOp, Ops, sint, Variable, MathTrait, identity_element
 from tinygrad.spec import tensor_uop_spec, type_verify
 from tinygrad.device import Device, BufferSpec
 from tinygrad.engine.realize import run_schedule
@@ -116,7 +116,7 @@ def _flat_to_grouped(padding:Sequence[sint]) -> tuple[tuple[sint, sint], ...]: r
 
 ReductionStr = Literal["mean", "sum", "none"]
 
-class Tensor(SimpleMathTrait):
+class Tensor(MathTrait):
   """
   A `Tensor` is a multi-dimensional matrix containing elements of a single data type.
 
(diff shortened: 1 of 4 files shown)


tinygrad/tensor.py
```diff
@@ -487,9 +487,9 @@ def manual_seed(seed=0):
 
   @staticmethod
   def _threefry_random_bits(key:Tensor, counts0:Tensor, counts1:Tensor):
-    x = (counts1.cast(dtypes.uint64) << 32) | counts0.cast(dtypes.uint64)
-    x = x._apply_uop(UOp.threefry, (key[1]._broadcast_to(x.shape).cast(dtypes.uint64) << 32) | key[0]._broadcast_to(x.shape).cast(dtypes.uint64))
-    counts0, counts1 = (x & 0xffffffff).cast(dtypes.uint32), ((x >> 32) & 0xffffffff).cast(dtypes.uint32)
+    x = (counts1.cast(dtypes.uint64) * 2**32) | counts0.cast(dtypes.uint64)
+    x = x._apply_uop(UOp.threefry, (key[1]._broadcast_to(x.shape).cast(dtypes.uint64) * 2**32) | key[0]._broadcast_to(x.shape).cast(dtypes.uint64))
+    counts0, counts1 = (x & 0xffffffff).cast(dtypes.uint32), ((x // 2**32) & 0xffffffff).cast(dtypes.uint32)
```
eliotgolding, line 492: Longs don't get removed by symbolic 2.0 when you have `Ops.SHL`/`SHR` in the graph, so this makes sure the old rewrite rules are enough

tinygrad/tensor.py
```diff
@@ -3249,29 +3249,35 @@ def bitwise_not(self) -> Tensor:
...
     print(Tensor([4, 13, 125], dtype=dtypes.uint8).rshift(2).numpy())
     ```
     """
-    assert dtypes.is_unsigned(self.dtype) and isinstance(x, int) and x >= 0, f"not supported {self.dtype=} {x=}"
-    return self.idiv(2 ** x)
+    a, b = self._broadcasted(x, reverse=reverse)
+    if a.dtype != dtypes.bool and not dtypes.is_int(a.dtype): raise RuntimeError(f"{self.dtype} is not supported")
+    # shift by a count greater than bit width isn't supported; this fix matches torch
+    b_invalid = (b < 0) | (b >= 8*a.dtype.itemsize)
+    return a._apply_uop(UOp.rshift, b_invalid.where(8*a.dtype.itemsize-1, b))._apply_broadcasted_uop(UOp.rshift, b_invalid)
```
eliotgolding, line 3280: When b is invalid, do two shifts to shift by `8*a.dtype.itemsize` places (can't do that in 1 shift as it's unsupported on the backends). This behaviour is the same as torch. Also, you can't just return 0 when b is invalid because rshift preserves sign (e.g. `-4 >> 64 == -1`).

geohot:
Do the tests have to change? Please make a minimal PR

eliotgolding:
They do if `reverse=True` is supported for Tensor shifts, because then you need to use `Ops.SHL`/`SHR`. Then the new test cases (`<< 100` and `<< -1`) fail without the right checks on `b` in Tensor.lshift.

Only way to make this more minimal is to delete `reverse` from MathTrait lshift/rshift, is that what you want?

PR #8962, opened 2025-02-08
fix test_ops - metal and pytorch diff

akhilmaddineni:
While running on Apple M2 and Pytorch 2.1.0 few tests are failing in test_ops. This PR fixes those errors, I am not sure if this is an issue on other platforms but at least pytorch issue should be common. 

Fixes : 

`test_gemm_fp16 `- Metal doesnt support fp16 (atleast as per docs) so its throwing this error. 
`RuntimeError: "addmm_impl_cpu_" not implemented for 'Half' so skipped this test case for device type metal. `

`test_any_axis,test_any_zero_axis,test_all_axis,test_all_zero_axis `- `torch.any/all `are expecting different i/p params so aligned tests to be in sync with torch.


chenyuxyz:
don't care if it's old torch issue, these tests passes with later version of torch and certainly are passing on ci now

akhilmaddineni:
Cool , upgraded to torch 2.6 and everything passes now without this changes!

PR #8965, opened 2025-02-08
am training bench

nimlgen:


geohot:
Hmm, I see it took 20 minutes. Is AM slower?

nimlgen:
Yep, investigating why resnet mlperf eval is 
loaded initial data        @  2.56s vs loaded initial data        @  181s

PR #8967, opened 2025-02-08
[bounty] TODO: this cast can be removed by putting the casts around the EXPAND…

GMDSantana:
Upgrade bounty:

> TODO: this cast can be removed by putting the casts around the EXPAND, must test properly


chenyuxyz:
missing the point

PR #8971, opened 2025-02-08
[bounty] TODO: this cast can be removed by putting the casts around the EXPAND

GMDSantana:
Upgrade bounty:

> TODO: this cast can be removed by putting the casts around the EXPAND, must test properly

This way the extra cast was removed. If anything else is needed, feel free to let me know.


GMDSantana:
@chenyuxyz can you give more details about what you are expecting?

PR #8972, opened 2025-02-08
Use getpass.getuser() instead of os.getlogin()

samayala22:
os.getlogin() is broken on WSL.

getpass.getuser() is also recommended in the [python docs](https://docs.python.org/3/library/os.html#os.getlogin) and it handles the env vars internally.

### DIFF
--- a/tinygrad/helpers.py
+++ b/tinygrad/helpers.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-import os, functools, platform, time, re, contextlib, operator, hashlib, pickle, sqlite3, tempfile, pathlib, string, ctypes, sys, gzip
+import os, functools, platform, time, re, contextlib, operator, hashlib, pickle, sqlite3, tempfile, pathlib, string, ctypes, sys, gzip, getpass
 import urllib.request, subprocess, shutil, math, contextvars, types, copyreg, inspect, importlib
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


nimlgen:
does `getpass.getuser()` raise if not is set, but `getlogin` fallbacks to pwd.getpwuid(os.getuid())[0]?

nimlgen:
What does it return on WSL?

samayala22:
> does `getpass.getuser()` raise if not is set, but `getlogin` fallbacks to pwd.getpwuid(os.getuid())[0]?

from [cpython lib](https://github.com/python/cpython/blob/3.13/Lib/getpass.py)

```python

def getuser():
    """Get the username from the environment or password database.

    First try various environment variables, then the password
    database.  This works on Windows as long as USERNAME is set.
    Any failure to find a username raises OSError.

    .. versionchanged:: 3.13
        Previously, various exceptions beyond just :exc:`OSError`
        were raised.
    """

    for name in ('LOGNAME', 'USER', 'LNAME', 'USERNAME'):
        user = os.environ.get(name)
        if user:
            return user

    try:
        import pwd
        return pwd.getpwuid(os.getuid())[0]
    except (ImportError, KeyError) as e:
        raise OSError('No username set in the environment') from e
```

> What does it return on WSL exactly?

```
>>> os.getlogin()
OSError: [Errno 6] No such device or address
```

nimlgen:
The behavior on linux seems to be the same, so let's try this since it works in wsl.

PR #8986, opened 2025-02-09
hevc decoder using nvcuvid

theomonnom:
the decoder is always set to output nv12, simplifying the conversion to rgb.
nvcuvid also manages the video parser, allowing it to work with other codecs—I’ve tested it with h264 & hevc
there is no demuxer; it only decodes raw bitstreams. adding a demuxer would increase complexity and introduce dependencies on external libraries like libavcodec.


autogen_stubs.sh
```diff
@@ -161,6 +161,17 @@ nv_status_codes = {}
   python3 -c "import tinygrad.runtime.autogen.nv_gpu"
 }
 
+generate_nvcuvid() {
+  clang2py \
+    nvcuvid/cuviddec.h \
```
theomonnom, line 166: the headers need to be downloaded from the [Nvidia website](https://developer.nvidia.com/nvidia-video-codec-sdk/download) (Video Codec SDK 13.0 Header Files)

geohot:
On the right track (bounty locked), but the bounty is for adding it to the NV driver, not the CUDA driver, so without the library.

chenyuxyz:
still active?

PR #8997, opened 2025-02-10
apply const folding before movement ops

Qazalin:


Qazalin:
UPat.any exists because of this chain of movement ops
image
so it's pattern matching on MOP(VIEW(CONST)) and MOP(CONST)

Qazalin:
this makes the scheduler graph hard to work with. All those movement ops should simplify to a single VIEW or not exist at all.

PR #9002, opened 2025-02-10
remove const_arg

Qazalin:
This is type checked in spec.py already before reaching the simplifier.
I'm happy tensor doesn't access UOp args anymore, `_to_const_val` wasn't good.


Qazalin:
for some reason if I don't do `-m pytest` CI fails, I tried deleting the cache but that didn't help. Why doesn't `pytest` work?
image
https://github.com/tinygrad/tinygrad/actions/runs/13239719380/job/36952197518#step:8:15

PR #9029, opened 2025-02-11
mulacc_unrolled should happen even with no DEVECTORIZE

S-Lykles:
I can add a regression test later today.

### DIFF
--- a/tinygrad/codegen/rewriter.py
+++ b/tinygrad/codegen/rewriter.py
@@ -523,13 +523,13 @@ def full_graph_rewrite(sink:UOp, opts:Optional[Renderer]=None) -> UOp:
   sink = graph_rewrite(sink, sym+expander)
 
   if DEVECTORIZE:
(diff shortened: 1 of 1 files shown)


tinygrad/codegen/rewriter.py
```diff
@@ -523,13 +523,13 @@ def full_graph_rewrite(sink:UOp, opts:Optional[Renderer]=None) -> UOp:
...
-    sink = graph_rewrite(sink, sym+(devectorize+float4_folding if opts is not None and opts.supports_float4 else devectorize)+load_store_indexing+
-      mulacc_unrolled)
+    # devectorize + load_store_indexing
+    sink = graph_rewrite(sink, sym+(devectorize+float4_folding if opts is not None and opts.supports_float4 else devectorize)+load_store_indexing)
   else:
     # new devectorize only for load/store
     sink = graph_rewrite(sink, sym+devectorize_load_store)
 
   # final rules for the renderer (without sym)
+  sink = graph_rewrite(sink, mulacc_unrolled)
```
geohot, line 533: So it's slower to do multiple rewrites, can you just add it to both?

PR #9031, opened 2025-02-11
add backend check command

hurrrsh:
Added ```python -m tinygrad.device``` to MNIST tutorial doc


geohot:
That is meant for notebooks

PR #9034, opened 2025-02-11
TestOps: silence pytorch std()/var() degrees of freedom warnings

joshiemoore:
tinygrad's `std()` and `var()` are correctly returning tensors full of `NaN` here, as evinced by the test cases passing, so we don't need all the [warning spam](https://media.discordapp.net/attachments/1069001075828469790/1337274136057352312/Screenshot_2025-02-07_at_12.09.44_PM.png?ex=67acc7d7&is=67ab7657&hm=b602955354cc7f89aff96b874f447fc4978dd41842ca8eac7a952c5a0edcafa8&=&format=webp&quality=lossless&width=555&height=605) about it from pytorch.


geohot:
Nice, thank you!

PR #9037, opened 2025-02-11
Graph rewrite 2.0 3.0

ttomsa:
Hopefully this is the last one :), much of the last pr isn't relevant anymore. Building the automaton lazily gives us nearly the same performance as building a deterministic automaton upfront without any of the drawbacks. It also makes everything better. Less code, simpler, don't have to worry about creating the minimal number of states, subsumption is straightfoward, repeats and allow_any_len work as is without any extra steps whereas before we had to consider all possible number of repeats and src length, meaning we either couldn't represent a transition as a tuple of the symbol and src or we had to set a max number of srcs.

### TODO list
- [ ] name assignment
- [ ] support for any src

Name assignment remains tricky. If there were no permutations or any src it would be a straightforward root to leaf traversal once a full pattern is matched, so an option would be to separate each of the possible srcs into its own pattern. Also trying to see if there's a way to do the name assignment inside matches() without making it slow.


geohot:
Oh, it's not enabled yet...I was looking and benchmarks look the same

geohot:
Going to unlock this bounty until there's correctness of the new matcher. It's okay if it isn't fast, but this PR as it doesn't even have live code in it.

ttomsa:
@geohot What I wanted to get at is that I have some doubts whether this is worth doing given that it doesn't replace the current matching because it only supports rewriting leaf to root and might not provide a huge speedup. Also, tree automata are very much not the standard way of solving this problem because of variable binding. A more standard way would be to build a decision tree. I'll open a pr if I have something more concrete.

niteria:
I can't put cycles into this, but maybe this will be of use to someone.

GHC has a pretty neat way of handling rewrites of syntax in the presence of lambda bound names (binders).
Here's the relevant code: https://github.com/ghc/ghc/blob/426ae98852be498fa87d10e4c88ba8d726d6b320/compiler/coreSyn/CoreMap.hs#L53

PR #9038, opened 2025-02-12
CUDA backend on Windows

rmtew:
CUDA now works and is selected before GPU on Windows. Tests pass. Windows-compatible parts of `pre-commit` pass.

Requirements:

- CUDA 12+ in the form of `cuda_12.8.0_windows_network.exe` or similar has been installed.
- The environment contains the `CUDA_PATH` variable from the CUDA install. Mine is `CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.8`. This is used to extract the major version for the DLL name on Windows.
- nVidia claims the versioning in the DLL naming is `bin\nvrtc64_Major Release Version_0.dll`.  Grok claims this is for 12+. That there is a different one for 11.3+, and a different one earlier than that. Users with older versions of CUDA (pre 12) therefore must update CUDA to have it detected.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -598,7 +598,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        backend: [llvm, clang]
+        backend: [llvm, clang, cuda]
 
     name: Windows (${{ matrix.backend }})
     runs-on: windows-latest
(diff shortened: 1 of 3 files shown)


rmtew:
I'm not yet understanding something about autogen, I'll try and work it out tomorrow.

rmtew:
Bad approach, need to deal with clang2py.

PR #9047, opened 2025-02-12
Ensure Windows CI correctly tests the specified backends 

rmtew:
- Ensure that the set backend environment variable is persisted to the next step via $GITHUB_ENV
- It doesn't actually persist for Windows unless shell is explicitly set to bash.
- Add the assertion to ensure the selected backend is actually used.


chenyuxyz:
thanks!

PR #9053, opened 2025-02-13
Start add vectorized support to transcendental ($100 bounty)

geohot:
I lowered the bounty to $100 and started it (with failing test)

Fix this only with changes to transcendental.py


chenyuxyz:
done in https://github.com/tinygrad/tinygrad/pull/9058

PR #9056, opened 2025-02-13
test: no early viz rewrite, only view one

geohot:


geohot:
I want to be able to quickly click the next one.

PR #9062, opened 2025-02-13
fix test_failure_53 

yvonmanzi:
# Fix cyclic substitution in graph rewrite

## Problem
Test `test_failure_53` was failing with a maximum recursion depth exceeded error. Upon investigation, this was caused by cyclic substitutions in the graph rewrite system where a node would be replaced by an expression containing itself, leading to infinite recursion.

## Changes Made
1. Added `uop_contains` helper to detect if a node exists in a UOp tree
2. Created `cyclic_safe_substitution` that checks for cycles before applying substitutions
3. Modified the substitution pattern to use this cycle-safe version
4. Updated test `test_assert_inf_recurse` to `test_skip_cyclic_substitution` to verify that cyclic substitutions are skipped rather than causing infinite recursion

## Notes
While this fix works and passes the tests, there might be more elegant ways to handle cyclic substitutions. Initially, I tried handling this in `bottom_up_rewrite` but moved the logic to the substitution pattern as it seemed more appropriate.

Looking for feedback on:
- Is handling cycles at the substitution level the right approach?
- Should we consider alternative strategies for cycle detection?
- Are there performance implications we should consider?
- Notably, while test_failure_53 no longer hits infinite recursion, it now fails with a different assertion:

`assert all(len(x.src) == 0 and x.op not in {Ops.BLOCK, Ops.BLOCKSTART, Ops.BLOCKEND, Ops.BLOCKFORK} for x in uops)`

This indicates that after linearization, some nodes still have children or contain block operations, when they should have been fully flattened to leaf nodes. Is this the expected "COMPILE_ERROR, val scope issue" that was mentioned in the original test comment? I would appreciate guidance on whether:
1. This is the expected failure mode we should be seeing
2. If not, what failure mode should we be targeting instead

### DIFF
--- a/test/unit/test_graph_rewrite.py
+++ b/test/unit/test_graph_rewrite.py
@@ -250,14 +250,11 @@ class TestSubstitute(unittest.TestCase):
     ret = substitute(ret, {a.sin():b})
     self.assertIs(ret, b.sin())
 
-  # broken due to infinite recursion
-  # NOTE: VIZ hangs and doesn't recover if you click this one
-  def test_assert_inf_recurse(self):
+  def test_skip_cyclic_substitution(self):
     a = UOp.variable('a', 0, 10)
     n1 = a.sin()
-    ret = n1
-    with self.assertRaises(RecursionError):
-      ret = substitute(ret, {n1:n1.sqrt()})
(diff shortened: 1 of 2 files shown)


geohot:
Do not use AI

PR #9064, opened 2025-02-13
Bitcast constant folding

quortus:
I'm currently investigating a failure of TestTranscendentalFunctions.test_payne_hanek_reduction.
Bitcast result looks ok so there is a chance some other folding is done incorrectly.

### DIFF
--- a/test/test_const_folding.py
+++ b/test/test_const_folding.py
@@ -1,4 +1,4 @@
-import unittest, math
+import unittest, itertools, math
 from tinygrad import Tensor, Device, dtypes
 from tinygrad.ops import Ops
 from tinygrad.helpers import CI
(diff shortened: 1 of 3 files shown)


quortus:
Yeah, the problem is not with bitcast folding. 
```python
UOp(Ops.CONST, dtypes.ulong, arg=-1023, src=())`

**Comment by quortus** [comment]
So, the root of the problem is that UOp ignores modular arithmetic rules.
`test_payne_hanek_reduction` creates this UOp [here](https://github.com/tinygrad/tinygrad/blob/e02e3b94c3372e21e7d9ecbc6948839b78685bfa/tinygrad/codegen/transcendental.py#L61):

```python
UOp(Ops.CONST, dtypes.ulong, arg=-1023, src=())
```

We have unsigned const with negative value and this breaks python backend.
I am working on a fix and a simpler test case.
Bitcast folding just allowed to fully fold UOp tree and exposed this problem.

quortus:
Blocked by #9073 (generic const folding bugfix)

PR #9068, opened 2025-02-13
Bitcast constant folding (Bounty)

theblobinthesky:
Seems to work fine except for the test TestTranscendentalFunctions::test_payne_hanek_reduction


theblobinthesky:
Turning off this rule fixes the test:
(UPat(Ops.CAST, name="root", src=UPat.cvar("c")), lambda root, c: root.const_like(c.arg)),

Not sure why the other cast rule interferes with my bitcast rule....

theblobinthesky:
Turns out, the problem was not directly related to `(UPat(Ops.CAST, name="root", src=UPat.cvar("c")), lambda root, c: root.const_like(c.arg)),`. Commenting it out just made it so the uop tree wasn't collapsed completely and the problem didn't surface again.

I changed
`   lambda a: a.const_like(exec_alu(a.op, a.dtype, [x.arg for x in a.src], False)) if a.op is not Ops.THREEFRY else None),
`To
`   lambda a: a.const_like(exec_alu(a.op, a.dtype, [x.arg for x in a.src], True)) if a.op is not Ops.THREEFRY else None),
`
Thereby enabling the masking after overflow. Unsigned ints weren't handled correctly by `exec_alu`.

theblobinthesky:
Can you elaborate a bit, this is my first time doing a pr properly.
I thought the schedule should be empty now since it boils down to a const value.
Or is it the truncate_output=True ?

PR #9058, opened 2025-02-13
[Bounty] Vectorize Transcendental

eitanturok:

PR #9072, opened 2025-02-13
[bounty] add vectorized support to transcendental

AlexCheema:
My understanding of this:

**Before:** Previously `rewriter.py` was responsible for handling a special case for vectorization of transcendental operations (`Ops.EXP2`, `Ops.LOG2`, `Ops.SIN`).

**After:**  We remove the special case from `rewriter.py` and instead handle vectorization directly in each transcendental function in `transcendental.py`.

**Why:** I can read `rewriter.py` and understand how it handles both scalar and vector inputs. Easier to understand since it's self-contained.

**Note:** Added `pillow` to `pydeps` in the `testdsp` job. I think it's fine but let me know if you want that different, e.g. using deps instead.

Based on https://github.com/tinygrad/tinygrad/pull/9053


geohot:
This is not the right approach, also this bounty is already locked. https://github.com/tinygrad/tinygrad/pull/9058

PR #9073, opened 2025-02-13
Force truncate all UOp constants

quortus:
I do not understand why all constant folding computations were done using python int but it does not produce the same result as running compiled code due to lack of wraparound.
If you want to keep using python int, you will need to painstakingly analyze all transformation rules to accurately apply delayed truncate when it is needed. See tests for examples.


geohot:
All tests failing

PR #9074, opened 2025-02-14
improve get_fake_data_bert

chenyuxyz:
it was creating a real buffer on device 0, then shard to 6 gpus. so it used more memory than needed and caused oom.

the real run loads data from disk directly to 6 gpus so it has less memory footprint.

`DEFAULT_FLOAT=HALF BENCHMARK=10 BS=66 EVAL_BS=66 GPUS=6 MODEL=bert python3 examples/mlperf/model_train.py` no longer oom


chenyuxyz:
not quite, still oom

PR #9082, opened 2025-02-14
new becomes_map = tensor_map

Qazalin:
The only difference between tensor_map and becomes_map is that becomes_map must map VIEW to movement ops.


Qazalin:
This is the prereq for reorder_expand https://github.com/tinygrad/tinygrad/pull/9051. Merging this will be my focus for tomorrow.
The challenge is in finding the right becomes_map k:v pairings such that we don't infinitely `subsitute`.
with becomes_map = tensor_map you can easily end up with one k/v pair saying RESHAPE(x) -> VIEW(x) and then one saying VIEW(x) -> RESHAPE(x).

Qazalin:
Prereq for this is merging kernel_map from #9165.

Qazalin:
Completed in https://github.com/tinygrad/tinygrad/pull/9201.

Issue #8043, opened 2024-12-04
get_grouped_dims does not split dim


chenyuxyz:
can be triggered by `WEBGPU=1 PYTHONPATH=. WINO=1 python examples/stable_diffusion.py`

getting error
```
  File "/Users/chenyu/code/tinygrad/tinygrad/codegen/lowerer.py", line 34, in get_grouped_dims
    limited = _limit_dims(dims, max_sizes) if max_sizes is not None else dims
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chenyu/code/tinygrad/tinygrad/codegen/lowerer.py", line 26, in _limit_dims
    if dims[i] * dims[i+1] <= m:
                 ~~~~^^^^^
IndexError: tuple index out of range
```

the input has `dims=(102400,), max_sizes=(65535, 65535, 65535)`, and it needs to reshape `102400` to maybe `(51200, 2)` to fit

wpmed92:
We had a limit_dim last time webgpu was in core that handled this case, but i see that the current limit dim has a different purpose.

chenyuxyz:
yea i think that's a regression from lowerer rewrite that's not matched yet

wpmed92:
Worked on it a bit today, got it working in `kernel.py` with `reshape_and_permute` but probably should be in lowerer instead.

```python
def apply_size_limits(self, gmax:Tuple[int], lmax:Tuple[int]):
    # reshapes dims (if possible) so that each element is less then, or equal to the coresponding element in max
    def to_max(dims:Tuple[int], max:Tuple[int]) -> Tuple[int]:
      if len(dims) == 0 or len(dims) > len(max) or all(d <= m for d,m in zip(dims, max)): return dims
      dims = list(dims) + [1]*(3-len(dims))
      for i in range(len(dims)):
        while dims[i] > max[i]:
        # find a divisor
          div = [x[0] for x in [(c, dims[i] % c) for c in [2, 3, 5, 7, int(math.sqrt(dims[i]))]] if x[1] == 0]
          if len(div) == 0: return tuple(dims)
          else: div = div[0]
          dims[i], dims[i+1] = dims[i]//div, dims[i+1]*div
      return tuple(dims)
    gdim, ldim = self.global_dims, self.first_reduce+self.group_for_reduces
    if gmax: self.reshape_and_permute(lambda x: to_max(x[:gdim], gmax) + x[gdim:], None)
[...]

PR #9085, opened 2025-02-14
Solve get_grouped_dims does not split issue

wpmed92:
This closes https://github.com/tinygrad/tinygrad/issues/8043

Our `_limit_dims` in `lowerer` only handles contraction, i.e. cases such as:
`dim=(2,3,4,5)`  `max=(16,16,16)`, so when` len(dim) > len(max)`

But with WebGPU we hit cases not handled by `_limit_dims`, when some element in `dim` is larger than the corresponding element in `max`, resulting in this error:

Screenshot 2025-02-14 at 14 36 17

This can come in various forms, such as:
- `dim=(64,3,4)`, `max=(16,16,16)`
- `dim=(128,)`, `max=(16,16,256)`

The algo tries to find a divisor for the current `dim` if it's larger than `max`, and divides it until it's `<= max`. While dividing, it multiplies the next dim by the same amount so the product of the shape doesn't change.

- I removed all skips which were skipped because of this issue.
- I also removed ignore of `test_speed_v_torch` previously failing.
- It also solves `fp16` stable diffusion previously failing with the same error message shown above.

### DIFF
--- a/tinygrad/codegen/lowerer.py
+++ b/tinygrad/codegen/lowerer.py
@@ -20,7 +20,7 @@ def get_contraction(old_shape:tuple[sint, ...], new_shape:tuple[sint, ...]) -> l
 def _group_dims(dims:tuple[sint, ...], max_sizes:tuple[int, ...]):
   # TODO: symbolic shape
   if not all_int(dims): return dims
-  while len(dims) > len(max_sizes):
+  while len(dims) > len(max_sizes) or any(d > m for d,m in zip(dims, max_sizes)):
     for i,m in enumerate(max_sizes):
       if dims[i] * dims[i+1] <= m:
         dims = dims[:i] + (dims[i]*dims[i+1],) + dims[i+2:]
@@ -29,6 +29,7 @@ def _group_dims(dims:tuple[sint, ...], max_sizes:tuple[int, ...]):
   return dims
 
 def _split_dims(dims, max_sizes):
+  if all(d <= m for d,m in zip(dims, max_sizes)): return dims
   _dims = list(dims) + [1]*(3-len(dims))
   for i in range(len(_dims)):
     while _dims[i] > max_sizes[i]:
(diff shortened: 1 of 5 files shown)


tinygrad/codegen/lowerer.py
```diff
@@ -27,18 +28,38 @@ def _limit_dims(dims:tuple[sint, ...], max_sizes:tuple[int, ...]):
...
+def _fit_to_max(dims:tuple[sint, ...], max_sizes:tuple[int, ...]):
+  if len(dims) == 0 or len(dims) > len(max_sizes) or all(d <= m for d,m in zip(dims, max_sizes)): return dims
+  _dims = list(dims) + [1]*(3-len(dims))
+  for i in range(len(_dims)):
+    while _dims[i] > max_sizes[i]:
+      div = 1
+      for d in range(2, int(math.sqrt(_dims[i])) + 1):
+        if (_dims[i] % d) == 0:
+          div = d
+          break
```
chenyuxyz, line 40: `div = next(d for d in range(2, math.ceil(math.sqrt(_dims[i]))) if (_dims[i] % d) == 0, 1)`

also can you add a test case for the math.ceil behavior?

wpmed92: Changed it to generator, will add the test.

wpmed92: @chenyuxyz Added the test, has to add `+ 1` in the `range` in the `generator`, otherwise it will skip the `sqrt` divisor.

wpmed92:
green benchmark keeps hanging

wpmed92:
@chenyuxyz Can you retrigger benchmark?

wpmed92:
It needs a little more work so that it works with cases hit by NV BREAM as before, but also works with WebGPU. I have a solution, will add test and push it.

wpmed92:
@chenyuxyz Benchmark finally green, all tests passing (except for line count).

The issue was that the "old" `_limit_dims` (which is now `_group_dims`) not only handles `len(dim) > len(max_sizes)`, but also some cases of `len(dim) == len(max_sizes)`, such as: `dim=(512,4,2)` `max=(8192,2,2`) -> `(2048,2)`. In such a case, when elements can be grouped, this should be the first try, because it's faster. Only after that should we try `_split_dims`.
WebGPU mostly hits `_split_dims` cases, and NV `_group_dims` cases, but the two can now co-exist as shown in the tests.

chenyuxyz:
fine to merge since it solves several webgpu problem. feels like there's room to clean up too

PR #9088, opened 2025-02-14
Prevent const folding in test_payne_hanek_reduction

quortus:


quortus:
@chenyuxyz This is kinda the whole point of this commit.

The real issue (const folding bugs):
1. Is already present and can create funny to debug bugs in random places.
2. Has nothing to do with either bitcast or test_payne_hanek_reduction.
3. Is better tested in #9080

I don't see any good reason to block work on bitcast because of this barely related failure. Fixing const folding can take a long time and not working on bitcast does not save you from numerous already present bugs.

geohot:
I'm fine with this for now

PR #9089, opened 2025-02-14
Bitcast constant folding 2.0

quortus:
Now with vconst support and better tests.
Depends on #9088

### DIFF
--- a/examples/test_onnx_imagenet.py
+++ b/examples/test_onnx_imagenet.py
@@ -73,3 +73,6 @@ if __name__ == "__main__":
     t = p.argmax().item()
     hit += y==t
     print(f"target: {y:3d}  pred: {t:3d}  acc: {hit/(i+1)*100:.2f}%")
+
+  import pickle
+  with open("/tmp/im.pkl", "wb") as f: pickle.dump(run_onnx_jit, f)
--- a/extra/replay_pkl.py
+++ b/extra/replay_pkl.py
@@ -1,0 +1,35 @@
+import pickle, sys
+from dataclasses import replace
+from tinygrad import Device
+from tinygrad.helpers import getenv
+from tinygrad.engine.jit import TinyJit
+from tinygrad.engine.realize import CompiledRunner
+from tinygrad.renderer import ProgramSpec
+from tinygrad.codegen.kernel import Kernel, Opt, OptOps
+
+if __name__ == "__main__":
+  with open(sys.argv[1], "rb") as f:
+    fxn: TinyJit = pickle.load(f)
+    print(f"{f.tell()/1e6:.2f}M loaded")
+  print(type(fxn))
+
+  knum = 1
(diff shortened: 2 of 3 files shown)


quortus:
I have found the source of MacOS test failure.
If you run the following code, the same UOp will be printed (0.0 or -0.0) because 1) UOps are cached 2) In python 0.0 == -0.0. Exact value depends on what code was executed before. But I guess this behavior is ok for tinygrad based on how it treats floating point comparisons.

```python
r1 = UOp.const(dtypes.float, 0.0)
r2 = UOp.const(dtypes.float, -0.0)
print(r1, r2)
```

test/unit/test_transcendental_helpers.py
```diff
@@ -1,21 +1,28 @@
...
 from tinygrad import dtypes
-from tinygrad.ops import UOp
+from tinygrad.ops import UOp, Ops
 from tinygrad.codegen.transcendental import payne_hanek_reduction, cody_waite_reduction, frexp, rintk, pow2if
 from test.helpers import eval_uop
 
 class TestTranscendentalFunctions(unittest.TestCase):
   def test_payne_hanek_reduction(self):
-    r, q = (eval_uop(u) for u in payne_hanek_reduction(UOp.const(dtypes.float64, 12 * math.pi + 0.1)))
+    # TODO: Test constant input when constant folding is fixed (or maybe test both variants)
```
geohot, line 10: Is there something preventing the full constant folding version from working?

test/helpers.py
```diff
@@ -52,10 +52,14 @@ def timeit(fxn:Callable[..., T], *args, **kwargs) -> tuple[T, float]:
   ret = fxn(*args, **kwargs)
   return ret, (time.perf_counter_ns()-st)*1e-6
 
-def eval_uop(uop:UOp):
+def eval_uop(uop:UOp, inputs:list[tuple[DType, list[Any]]]=[]):
```
geohot, line 55: Hmm, are the tests not being linted? This should raise a warning in most linters (cause that list is mutable)

quortus: Not by pylint. I though it was not a problem because inputs is read-only. I replaced it with None.

test/unit/test_transcendental_helpers.py
quortus: Const folding uses arbitrary precision arithmetic instead of modular arithmetic. I did some tests on that in #9080.
This PR opens more folding opportunities and triggers the problem. I think this failure should be addressed separately because it is not directly related to this PR.

quortus: I've added reduced test case from test_payne_hanek_reduction to #9080. It contains no bitcasts.

geohot:
Congrats, the bounty is yours! E-mail george@tinygrad.org to claim, can pay via PayPal or USDC on eth

PR #9091, opened 2025-02-14
Extra stable sigmoid pattern

S-Lykles:


chenyuxyz:
if you want this it needs some tests, i don't see any kernel diff either

PR #9092, opened 2025-02-14
Make simplify use symbolic_flat

S-Lykles:
Is there a specific reason this is not using symbolic_flat?


S-Lykles:
Okay so this doesnt work seperately from #8123

PR #9110, opened 2025-02-15
Fix TestLinearizerFailures.test_failure_53

hmih:
I took a look at the bounty list and this one seems easy enough.

At first I just stared at the test for a while trying to decypher the AST, then I rummaged around in the history of this file. I looked at the few commits working on and around this testcase and this one stood out to me: https://github.com/tinygrad/tinygrad/commit/ba17786068255cde78e8d8947b5c16eb0b3f9edc I think the rafactoring missed the last `WHERE -> VALID -> VIEW` piece which results in an open basic block that is never closed. I believe this is the root cause because I see an unclosed block in some debug prints I played around with. The test fails because the bottom up rewrite throws a RecursionError, and it makes sense for rewrites to fail because the block is never closed. The test case makes sense to me after the edits.

I tested this fix on macbook. I appreciate any feedback.


chenyuxyz:
@Qazalin can you verify that was missed? if i add `type_verify([lin.ast])` in `helper_test_lin` 53 indeed seems to be broken

Qazalin:
Err, this is removing a real mask on a CONST, how is that a root cause fix?
This was failing even before that commit.

chenyu, this ran fine for me on master:
```diff
 def helper_test_lin(lin: Kernel, opts, failed_platforms, rtol=1e-2, atol=1e-2):
   if any(b.dtype.base == dtypes.half for b in lin.membufs) and not is_dtype_supported(dtypes.half): return
   if any(b.dtype.base == dtypes.bfloat16 for b in lin.membufs) and not is_dtype_supported(dtypes.bfloat16): return
+  from tinygrad.spec import type_verify, shape_spec
+  type_verify(list(lin.ast.toposort), shape_spec)
```

Qazalin:
please use `PYTHONPATH=. VIZ=1 python test/test_linearizer_failures.py TestLinearizerFailures.test_failure_53` and think about this graph more deeply.

Your diff is changing the nature of this graph entirely:

image


image

chenyuxyz:
`type_verify([lin.ast])` failed on master though

Qazalin:
hmm, kernel.py type_verifies before/after ast_fixup. Could you push the diff you ran?

chenyuxyz:
```
 def helper_test_lin(lin: Kernel, opts, failed_platforms, rtol=1e-2, atol=1e-2):
   if any(b.dtype.base == dtypes.half for b in lin.membufs) and not is_dtype_supported(dtypes.half): return
   if any(b.dtype.base == dtypes.bfloat16 for b in lin.membufs) and not is_dtype_supported(dtypes.bfloat16): return
+  from tinygrad.spec import type_verify
+  type_verify([lin.ast])
```

i don't know if this should pass or not (it passed for other ast in the test). also your diff failed for me on master with `RecursionError` in bottom_up_rewrite
`python test/test_linearizer_failures.py -k test_failure_53`

Qazalin:
Yea, the RecursionError  is because of a bug in `codegen/linearize.py` this bounty is meant to solve.
If type_verify failed you'd see a `RuntimeError("UOp verification failed")`, this is my stack trace with your diff:
```
  File "/Users/qazal/code/tinygrad/test/external/fuzz_linearizer.py", line 98, in run_linearizer
    prg = CompiledRunner(lin.to_program())
                         ^^^^^^^^^^^^^^^^
  File "/Users/qazal/code/tinygrad/tinygrad/codegen/kernel.py", line 686, in to_program
```

PR #9113, opened 2025-02-15
limit_dims is group_dims [pr]

wpmed92:
As discussed with @chenyuxyz I'm separating out `group_dims` refactor from the [split_dims PR](https://github.com/tinygrad/tinygrad/pull/9085)

The grouping failure check is moved out of `_group_dims` and the `None` return is introduced, so that if `_group_dims` fails to limit dims, `_split_dims` can try. There is only `RuntimeError` when `len(limited) > len(max_sizes)`, because in such a case `_split_dims` won't help.

### DIFF
--- a/test/test_linearizer.py
+++ b/test/test_linearizer.py
@@ -1199,6 +1199,8 @@ class TestLinearizer(unittest.TestCase):
     _assert_grouped_dims("gidx", (2,3,4), (16,16,16), False, [2,3,4])
 
     # test splitting globals
+    # test grouping behavior when len(dims) == len(max_sizes)
+    _assert_grouped_dims("gidx", (512,4,2), (8192,2,2), False, [2048,2])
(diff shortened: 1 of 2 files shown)


chenyuxyz:
test? i think you have something in the split dim pr

wpmed92:
@chenyuxyz Added that one test case currently not tested: where `len(dims) == len(mas_sizes)`. The other tests for this cases are for `split_dim`.

chenyuxyz:
do you know why WEBGPU process replay failed? if it really needs to be one single pr that's fine (but we need to clean up some lines)

wpmed92:
The process replay error is because of the in bounds check in: 
`if i < (len(dims)-1) and dims[i] * dims[i+1] <= m:`
The tuple index error it fails with can't happen now, so the failure is due to something like.

wpmed92:
@chenyuxyz Seeing the PR separated out I think it makes more sense to have both (split/group) in one go.

PR #9125, opened 2025-02-16
Fix issues with Quickstart example (mnist data loading, dataset dimensions and Tensor creation from sample of indices)

mlgarchery:
I played with your lib this afternoon. I liked it but I had issues setting up the quickstart tutorial only to notice a few changes were needed for it to work (the way you load the mnist dataset and create a tensor from a sample of indices probably changed since the last time tutorial was updated).

The proposed PR loads the mnist dataset from `tinygrad.nn.datasets` (in the initial example your load from `extra` ? I don't see it anywhere else in the doc) and squeeze then flatten the vectors so it can be fed appropriately to the model in the example (starting with a Linear layer of in_features = 784)

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -187,7 +187,7 @@ with Tensor.train():
     out = net(batch)
 
     # compute loss
-    loss = sparse_categorical_crossentropy(out, labels)
+    loss = Tensor.sparse_categorical_crossentropy(out, labels)
 
     # zero gradients
     opt.zero_grad()
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -228,8 +228,8 @@ with Timing("Time: "):
     out = net(batch)
 
     # calculate accuracy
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -169,21 +173,21 @@ We use `with Tensor.train()` to set the internal flag `Tensor.training` to `True
...
-X_train, Y_train, X_test, Y_test = fetch_mnist()
-
 with Tensor.train():
   for step in range(1000):
     # random sample a batch
-    samp = np.random.randint(0, X_train.shape[0], size=(64))
-    batch = Tensor(X_train[samp], requires_grad=False)
+    samp = Tensor.randint(64, high=X_train.shape[0])
+    batch = X_train[samp]
+    batch.requires_grad = False
```
mlgarchery, line 181: Not sure about this line, I kept it from the initial example

geohot:
Tests failing. Also, there has to be a cleaner fix than this.

PR #9163, opened 2025-02-18
Only add enable f16 directive if ShaderF16 is supported

wpmed92:
We do a check in `ops_webgpu` to only request `ShaderF16` if it is supported.
However, we always render `enable f16;` directive regardless of the f16 check result in `ops_webgpu`.
This PR adds this check in the renderer.


tinygrad/renderer/wgsl.py
```diff
@@ -85,7 +88,8 @@ def render_kernel(self, function_name:str, kernel:list[str], bufs:list[tuple[str
     bind_it = iter(range(len(bufs)))
     external_local_bufs = [line.lstrip() for line in kernel if "var<workgroup>" in line]
     kernel[:] = [line for line in kernel if "var<workgroup>" not in line]
-    prg = "enable f16;\nfn nan() -> f32 { let bits = 0xffffffffu; return bitcast<f32>(bits); }\n"
+    prg = "enable f16;\n" if self.supports_f16 else ""
```
geohot, line 91: Instead, why don't you check if f16 is in any uops? I think this is what we do for OpenCL.

PR #9171, opened 2025-02-19
fix win32 CPUProgram missing cache flush

peppingdore:
Also fix wrong ctypes, rename ptr to self.mem

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -270,6 +270,9 @@ class CPUProgram:
     # The bug was fixed in https://github.com/llvm/llvm-project/commit/454cc36630296262cdb6360b60f90a64a97f7f1a but was only backported to xcode 16+
     if platform.machine() == "arm64" and OSX: args = args[:8] + [ctypes.c_int64(a) if isinstance(a, int) else a for a in args[8:]]
     return cpu_time_execution(lambda: self.fxn(*args), enable=wait)
+  
+  def __del__(self):
+    if sys.platform == 'win32': ctypes.windll.kernel32.VirtualFree(ctypes.c_void_p(self.mem), ctypes.c_size_t(0), 0x8000) #0x8000 - MEM_RELEASE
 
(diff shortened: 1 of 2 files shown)


geohot:
Can you add a test that this fixes?

tinygrad/device.py
```diff
@@ -236,10 +236,13 @@ def __init__(self, name:str, lib:bytes):
       PAGE_EXECUTE_READWRITE = 0x40
       MEM_COMMIT =  0x1000
       MEM_RESERVE = 0x2000
-      ctypes.windll.kernel32.VirtualAlloc.restype = ctypes.c_uint64
-      ptr = ctypes.windll.kernel32.VirtualAlloc(ctypes.c_int(0), ctypes.c_int(len(lib)), MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE)
-      ctypes.memmove(ptr, lib, len(lib))
-      self.fxn = ctypes.CFUNCTYPE(None)(ptr)
+      ctypes.windll.kernel32.VirtualAlloc.restype = ctypes.c_void_p
+      self.mem = ctypes.windll.kernel32.VirtualAlloc(ctypes.c_void_p(0), ctypes.c_size_t(len(lib)), MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE)
```
geohot, line 240: Hmm, this isn't freed.

peppingdore:
> Can you add a test that this fixes?

I have no idea how to test instruction cache not being properly invalidated.
I just added instruction cache flush to win32 that is already present in non-win32 code:
https://github.com/tinygrad/tinygrad/blob/e4a8bf28eac254c39a94f0f35af7546f3b9a99f5/tinygrad/device.py#L253-L258

tinygrad/device.py
peppingdore: Fixed

geohot:
Fair, I don't know how to test that either.

PR #9175, opened 2025-02-19
debug script for slow bert kernel

chenyuxyz:


chenyuxyz:
beam finds group 76 so it's 4ms not 40ms

PR #9178, opened 2025-02-19
linearizer: avoid recursive cycles when merging BLOCKENDs

jfhs:
Bugfix for bounty `Fix TestLinearizerFailures.test_failure_53`


chenyuxyz:
did not fix test_failure_53, also changes like this itself needs some tests

Issue #9158, opened 2025-02-18
Hi from conda-forge, your friendly new distribution channel 👋


h-vetinari:
Hey there!

Tinygrad is really cool ❤ 

I help maintain the [conda-forge](https://conda-forge.org/) ecosystem, which is essentially a rolling cross-platform distribution, based on an alternative packaging paradigm (as opposed to python wheels) that encompasses not only the python bits, but also (almost) everything else.

In particular, we have working compiler toolchains for C/C++/CUDA that users can just install, making it a good fit for the kinds of things that tinygrad is doing, because - for the most part - users won't have to worry about installing anything else but
```
conda install -c conda-forge tinygrad
```
<details>
<summary>Choice of installation tool</summary>

You cannot install conda packages through pip/uv/poetry, etc., as those tools do not understand packages in the conda format. However there are other tools than just conda ([download](https://github.com/conda-forge/miniforge/releases)) that can be used; in particular, notable alternative implementations are [mamba](https://mamba.readthedocs.io/en/latest/index.html) and the next-gen [pixi](https://pixi.sh/latest/) that I highly recommend.
</details>

## Limitations

There are some limitations of course. In particular, conda-forge cannot ship:
* glibc, as it sits too deep in the system to replace. To be compatible with a wide amount of systems, we currently compile against a 2.17 baseline (yes, that's old...)
* the CUDA drivers themselves, due to EULA reasons (the reason we can ship the toolchain is because nvidia itself provides the respective components through conda-forge)
* xcode / MSVC / ... compiler, due to license reasons

## Available variants

That said, we currently provide a matrix of builds for the following:
* linux-64, linux-aarch64, osx-64, osx-arm64, win-64
* python 3.9-3.13
* CUDA-enabled builds on linux-64, linux-aarch64, win-64

By default the builds come with clang (see more below), and where available with a CUDA toolchain. The versions aren't pinned, so you can use any recent clang version, and any cuda toolchain from 12.0 to 12.8 (simply add `cuda-version=12.<x>` to the installation).

[...]

PR #9186, opened 2025-02-20
use torch in test_gradient

chenyuxyz:
used torch.autograd.grad, but not sure if it can be a template like jax


chenyuxyz:
#9158

PR #9197, opened 2025-02-22
different way to write torch backend

geohot:


test/test_ops.py
```diff
@@ -46,8 +50,8 @@ def compare(s, tinygrad_output, torch_output, atol, rtol):
   if DEBUG >= 6:
     np.set_printoptions(linewidth=200, suppress=True)
     print(ret.numpy())
-    print(out.detach().numpy())
-  compare("forward pass", ret.numpy(), out.detach().numpy(), atol=atol, rtol=rtol)
+    print(out.detach().cpu().numpy())
```
albanD, line 53: FYI you can do `out.numpy(force=True)` here

PR #9209, opened 2025-02-22
[Bounty] PyTorch backend make MNIST run

damoonsh:
My understanding is that functionalities can be mapped using TTensor. I have added a few here and the error message has changed asking for more mapping.


geohot:
Can you use backend.py instead of backend2.py. Added a note. It's mostly the same logic

PR #9211, opened 2025-02-23
remove unused memsize_to_str and minor cleanups [pr]

ShikChen:
Encountered `IndexError: list index out of range` when calling `memsize_to_str()` with inputs <= 1. This PR fixes the issue:

```py
from tinygrad.helpers import memsize_to_str

print(memsize_to_str(0))  # error -> 0.00 B
print(memsize_to_str(1))  # error -> 1.00 B
print(memsize_to_str(1000))  # 1000.00 B -> 1.00 KB
print(memsize_to_str(1234))  # same
```

Since there are no existing tests for `memsize_to_str()`, I haven't added any new ones. Let me know if you'd like tests to be included.

### DIFF
--- a/examples/llama3.py
+++ b/examples/llama3.py
@@ -47,7 +47,7 @@ def concat_weights(models, device=None):
     disk_tensors: List[Tensor] = [model[name] for model in models]
     if len(disk_tensors) == 1 or len(disk_tensors[0].shape) == 1:
       return disk_tensors[0].to(device=device)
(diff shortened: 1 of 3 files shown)


chenyuxyz:
since it's not used, i think it's fine to just remove that line

ShikChen:
Ah, I should have expanded my grep scope to catch that earlier. Removed the line and made a few minor tweaks while reviewing `helpers.py`. No behavioral changes expected.

chenyuxyz:
cool thanks!

PR #9220, opened 2025-02-24
insert contiguous in post permuted assignment

Qazalin:
The scheduler can do this now, sadly dealing with ASSIGN is required to merge #9210 - The contiguous in SGD was never a good idea anyways.

This diff has the general direction, recurse the local graph and insert contiguous if we can't fuse the LOAD:
```
VIZ=1 python test/test_assign.py TestAssign.test_post_permuted_assignment
```
image

This way we don't need to assert. I believe we'll end up doing something similar with cycle asserts.
Keeping in draft as there are two cases it doesn't handle yet:
1. Can swizzles result in a wrong ShapeTracker in the final AST?
2. The local graph toposort should stop when there is a KERNEL parent, currently it's recursing it to root.
Both of these suggest `sym` might be early for this, planning to integrate it with `create_kernels`.

### DIFF
--- a/test/test_assign.py
+++ b/test/test_assign.py
@@ -277,13 +277,12 @@ class TestAssign(unittest.TestCase):
     #GlobalCounters.cache = []
     ba1 = a.lazydata.base.realized # noqa: F841
     bb1 = b.lazydata.base.realized # noqa: F841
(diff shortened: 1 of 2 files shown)


Qazalin:
Gonna focus on the KERNEL graph refactor, bring multi output back, then get back to this.

PR #9226, opened 2025-02-24
Stable gradient

S-Lykles:
Also put ctx outside of the where to propagate the sign

### DIFF
--- a/tinygrad/gradient.py
+++ b/tinygrad/gradient.py
@@ -23,8 +23,8 @@ pm_gradient = PatternMatcher([
   (UPat((Ops.CMPLT, Ops.CMPNE)), lambda: (None, None)),
   (UPat(Ops.ADD), lambda ctx: (ctx, ctx)),
   (UPat(Ops.POW, name="ret"), lambda ctx, ret:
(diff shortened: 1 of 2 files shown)


S-Lykles:
I was tryin to write 
```python
ret.src[0].eq(0).where(ret.src[1].eq(0).where(ret.src[1], ret.src[1]*ret.src[0].pow(ret.src[1]-1)), ret.src[1]*ret.src[0].pow(ret.src[1]-1)
``` 
as
```python
(ret.src[0].eq(0) and ret.src[1].eq(0)).where(ret.src[1], ret.src[1]*ret.src[0].pow(ret.src[1]-1))
``` 
but that didn't work for some reason

S-Lykles:
This would pass if this todo gets fixed. Locally the test pass for me with so not sure how to debug this
`WEBGPU=1 WEBGPU_BACKEND="WGPUBackendType_Vulkan" python3 -m pytest test/test_ops.py::TestOps::test_pow_zero_tensor`

https://github.com/tinygrad/tinygrad/blob/f0b24d230cbe0aad509316e13e84152f66bc6d00/test/test_ops.py#L649-L654

chenyuxyz:
i think it's fine to skip WEBGPU... it has issues with inf and nan.

PR #9227, opened 2025-02-24
[Bounty] making beautiful_mnist_torch.py run

damoonsh:
Aim to make mnist run with TINY_BACKEND=1, mapped some of the base functions. Will iteratively debug and add all the mappings needed.

### DIFF
--- a/extra/torch_backend/backend.py
+++ b/extra/torch_backend/backend.py
@@ -55,18 +55,18 @@ def masked_select(self, mask):
 
 @torch.library.impl("aten::as_strided", "privateuseone")
 def as_strided(tensor, size, stride, storage_offset=None):
-  print(tensor.shape, size, stride)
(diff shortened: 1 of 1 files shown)


damoonsh:
Currently failing at gradient computation. I suspect one of the implemented methods is unnecessarily invoking Ops.View while computing gradients and that is source of the problem. Will try to make use realize and contiguous to see if the current error can be mitigated. 

Also not sure to what extent torch._decomps can be used but in the forward pass it seems that they can be replaced by Tiny Tensor and wrap/unwrap.

The error starts on **_loss.backward()_** and goes to tinygrad/tinygrad/gradient.py", line 66, in compute_gradient
    if lgrads is None: raise RuntimeError(f"failed to compute gradient for {t0.op}\n\nin {str(t0)[0:1000]}...")
RuntimeError: failed to compute gradient for Ops.VIEW

PR #9230, opened 2025-02-24
bring cast before view back

Qazalin:


Qazalin:
speedup in resnet/bert/cifar
slower GPT2 HALF=1
@chenyuxyz any ideas on how to tune this UPat?
image

chenyuxyz:
previously we turn if off for gpt2, i don't know how to tune it

Qazalin:
GPT2 is slow because it's wasting time indexing reshapes, this is the process replay diff:
image

Qazalin:
keeps speedups in resnet/bert/cifar, only triggering on EXPAND fixes gpt2 HALF=1:
image

PR #9231, opened 2025-02-24
fix test linearizer failure

bhagwat-rahul:
Bounty Bug Fix TestLinearizerFailures.test_failure_53

less aggressive optimisation using local instead of grouptop which was causing scope errors


chenyuxyz:
missing the point, the bounty is to fix the scope error

PR #9232, opened 2025-02-25
Acc first

S-Lykles:
Hopefully this can remove the mulacc_unrolled pass


geohot:
Nice, much better to do it this way! Test failure does look real though

S-Lykles:
Mmh I think it was just the macos cache misses and timeout, its passing now

S-Lykles:
Im surprised reduce collapse isnt broken by this, I think its not tested in combination with unroll

PR #9233, opened 2025-02-25
(bounty) Make mnist training run with torch backend

tocubed:
It runs but with some performance and accuracy issues.

image

Few notes:

- Had to workaround thrown errors regarding events and streams in the DeviceGuardImplInterface and made them no-ops. Unsure if this is right.
- Patched pytorch Optimizer class to realize weight and optimizer state tensors on step. Otherwise computation graph blew up.
- Edited test file to delay float conversion as the kernel is gather samples is already really slow and 2x worse with float. Matches behavior of other beautiful_mnist example,

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -171,8 +171,8 @@ jobs:
       run: PYTHONPATH=. DEBUG=2 python3 extra/torch_backend/torch_tests.py TestTinyBackendPRIVATEUSE1.test_unary_log_tiny_float32
     - name: Test Ops with TINY_BACKEND (expect failure)
       run: PYTHONPATH=. LLVM=1 LLVMOPT=0 TINY_BACKEND=1 python3 -m pytest -n auto test/test_ops.py || true
-    - name: Test beautiful_mnist in torch with TINY_BACKEND (expect failure)
-      run: PYTHONPATH=. TINY_BACKEND=1 python3 examples/other_mnist/beautiful_mnist_torch.py || true
+    - name: Test beautiful_mnist in torch with TINY_BACKEND
+      run: PYTHONPATH=. ALLOW_DTYPE_MISMATCH=1 TINY_BACKEND=1 python3 examples/other_mnist/beautiful_mnist_torch.py
     - name: Test some torch tests (expect failure)
       run: PYTHONPATH=. python3 -m pytest extra/torch_backend/torch_tests.py -v --tb=no || true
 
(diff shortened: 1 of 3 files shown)


geohot:
Test failing

tocubed:
One more special case on CPU device - skipped the tiny generated kernel and used whatever torch provides. Just too slow for CI otherwise. Doesn't look related to torch backend as examples/beautiful_mnist.py produces the same slow kernel.

geohot:
There's tons of spam in the test, can you remove the expect failure and it passes?

geohot:
I'll lock this to you, good work on the "meta" decomp and optimizer hook.

geohot:
I copied in the hacked thing too. `ALLOW_DTYPE_MISMATCH=1 TINY_BACKEND=1 python3 examples/other_mnist/beautiful_mnist_torch.py` works but the accuracy is low.

tocubed:
Thanks. 

The meta decomps return zero gradients and were the source of the error. Backwards convolution and maxpool are now based on tiny's autograd. Model accuracy matches pytorch native now. 

Performance dropped enough I had to bump CI timeouts. Hope that's fine, running low on time to investigate.

.github/workflows/test.yml
```diff
@@ -171,8 +171,8 @@ jobs:
       run: PYTHONPATH=. DEBUG=2 python3 extra/torch_backend/torch_tests.py TestTinyBackendPRIVATEUSE1.test_unary_log_tiny_float32
     - name: Test Ops with TINY_BACKEND (expect failure)
       run: PYTHONPATH=. LLVM=1 LLVMOPT=0 TINY_BACKEND=1 python3 -m pytest -n auto test/test_ops.py || true
-    - name: Test beautiful_mnist in torch with TINY_BACKEND (expect failure)
-      run: PYTHONPATH=. TINY_BACKEND=1 python3 examples/other_mnist/beautiful_mnist_torch.py || true
+    - name: Test beautiful_mnist in torch with TINY_BACKEND
+      run: PYTHONPATH=. ALLOW_DTYPE_MISMATCH=1 TINY_BACKEND=1 python3 examples/other_mnist/beautiful_mnist_torch.py
```
geohot, line 175: `ALLOW_DTYPE_MISMATCH` no longer needed. Also, can you add `TARGET_EVAL_ACC_PCT=96.0` to the test and check that? Then I think the bounty is yours!

geohot:
I added the target eval check. If CI passes, I'll merge. george@tinygrad.org to claim bounty, can pay via PayPal or USDC on ETH.

PR #9234, opened 2025-02-25
feat(FLUX1): load T5, CLIP, AE from path

yigithanyigit:
This PR using all related model weights from just single `model_path`. Previously, the implementation only took the main model weights from the specified path and downloaded the others from the internet. 

Since weights like T5, CLIP, and AE models are typically stored together in the same folder structure, we now extract that structure from the URLs and use it to look for local files. We can define a arg for each of the model but I believe this is way more nicer than defining bunch of explicit args.

Thanks


yigithanyigit:
> fetch works with local path too, so you can just update the urls and no need to make load model code more complicated

Thanks, Closing then.

PR #9245, opened 2025-02-25
Make const moving not depend on the order

S-Lykles:
Since floats are not being flipped anymore this should help with const folding for floats


geohot:
Why is this draft? I think it's a good idea

S-Lykles:
Wanted to add some tests

geohot:
Still planning on tests?

geohot:
Feel free to add tests in follow up.

PR #9259, opened 2025-02-26
Update tensor.py

whalmcfly:
Fixing "_one_hot_along_dim input needs to be int #9179" caused incorrect behavior in the "sparse_categorical_crossentropy" function.
- Added a cast to int32.
- Uncertain whether similar changes should be applied elsewhere in tensor.py.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -3091,7 +3091,7 @@ class Tensor(SimpleMathTrait):
     print(Tensor([-3., -2., -1., 0., 1., 2., 3.]).leaky_relu(neg_slope=0.42).numpy())
     ```
     """
-    return self.relu() - (-neg_slope*self).relu()
+    return (self<0).where(neg_slope*self, self)
 
   def mish(self):
     """


chenyuxyz:
what are you trying to fix?

whalmcfly:
I was running:
"def step():
    """Perform a single optimization step."""
    Tensor.training = True  # makes dropout work
    samples = Tensor.randint(BATCH_SIZE, high=X_train.shape[0])
    x_train_batch, y_train_batch = X_train[samples], Y_train[samples]

    optim.zero_grad()
    labels, _ = model(x_train_batch.float())
    labels = labels.float()
    y_train_batch = y_train_batch.float()
    loss = labels.sparse_categorical_crossentropy(y_train_batch)
    loss.backward()

    optim.step()
    return loss

jit_step = TinyJit(step)"

and sparse_categorical_crossentropy gave me an error that it takes only integers. Maybe I'm wrong.

chenyuxyz:
the error is saying don't convert y_train_batch to float before passing into sparse_categorical_crossentropy

PR #9260, opened 2025-02-26
Ints ordering flattening

S-Lykles:


S-Lykles:
I could also merge `symbolic` and `symbolic_flat` since `simplify` was the only function that needed `symbolic`. Previously it would break if  you used `symbolic_flat` but with this it doesn't

S-Lykles:
I will still add some unit tests for the new `order` function

S-Lykles:
@chenyuxyz Could this be benchmarked?

test/unit/test_simplify_valid_idx.py
```diff
@@ -258,9 +264,11 @@ def test_openpilot_conv3(self):
...
 
     self.check(load,
-               "((((idx2*2)+ridx0)<11)&((((idx1*8)+ridx1)<3)!=True))",
-               "(((idx0+((idx1*512)+(ridx1*64)))+832)%1024)",
-               "((((idx2*2)+ridx0)+(((idx1+((ridx1+5)//8))+1)//2))+-4)")
+               ("((((idx2*2)+ridx0)<11)&((((idx1*8)+ridx1)<3)!=True))",
+                "(((((idx1*8)+ridx1)<3)!=True)&(((idx2*2)+ridx0)<11))"),
+               "((((idx0+(idx1*512))+(ridx1*64))+832)%1024)",
+               ("((((idx2*2)+(((idx1+((ridx1+5)//8))+1)//2))+ridx0)+-4)",
+                "(((((idx1+(idx2*2))+(((idx1+((ridx1+5)//8))//2)*-1))+ridx0)+((ridx1+5)//8))+-4)"))
```
chenyuxyz, line 271: for the longer ones can you keep only the correct one? easier to understand the change.

the shorter ones are probably fine to keep any correct expressions

chenyuxyz:
benchmark failure seems real, something got really slow?

S-Lykles:
> benchmark failure seems real, something got really slow?
Mmh "Run 10 CIFAR training steps w BF16" has a really high python time on the first step now

chenyuxyz:
yea i think it's pickling the enum value from this pr and read it from master

S-Lykles:
Also some other benchmarks were also definitely slower, I did notice some loads moved from the start of the kernel to much later to right before the computation where they were needed.

S-Lykles:
#9266 Is what caused the slowdown

I think its happening when cast is moved after the math ops becase it causes some loads to be moved

S-Lykles:
First cifar training step is still a lot slower

S-Lykles:
@chenyuxyz could this be benchmarked again?

Issue #9247, opened 2025-02-25
missed fusion in cat with unsafe pad ops


chenyuxyz:
found in threefry, example
```
from tinygrad import Tensor, dtypes
x = Tensor([1, 2, 3])
t = x.cat(x.idiv(2))
print(t.tolist())
```
has 2 kernels, one for idiv and one for pad and add. This is due to `realize_before_view` inserts a realize because idiv is not safe to pad.

(1) idiv is safe to pad if it's idiv to a non-zero const (maybe zero is fine too?)
(2) maybe can rewrite cat

Qazalin:
As a first step we shouldn't generate division by 0 when it is diving by VALID.

#10369 `test/test_schedule.py TestSchedule.test_rand_handcoded` has a custom schedule to fuse rand, see:

```
unsigned long alu86 = (alu0?0ull:4294967296ull)
...(unsigned long)((alu133+alu88+10u))))/alu86
```

chenyuxyz:
fixed
```
#include <metal_stdlib>
using namespace metal;
kernel void E_3_2(device int* data0_6, device int* data1_3, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int lidx0 = lid.x; /* 3 */
  int val0 = (*(data1_3+lidx0));
  *(data0_6+lidx0) = val0;
  int alu1 = ((val0<0)?1:0);
  *(data0_6+(lidx0+3)) = ((val0+alu1)>>1);
}
```

PR #9265, opened 2025-02-26
reorder masked view for IDIV by const

Qazalin:
https://github.com/tinygrad/tinygrad/issues/9247

There's probably a way to make this more generic though.


Qazalin:
This is part of a larger problem with how we realize expands and pads, worth merging as a more generic diff.

PR #9266, opened 2025-02-26
Move unary ops after math ops

S-Lykles:


S-Lykles:
@chenyuxyz Can this be benchmarked, I want to know if its responsible for the slower openpilot 0.9.4 time of #9260

PR #9272, opened 2025-02-26
simpler as_strided

geohotstan:
Taking a break from agonizing over ORT C++ quantization.

I remember also doing the to_movement_ops way back when!!
That was fun.
I once knew how it worked, but I no longer remember.
all I remember is the shifty expand reshape to get fake 1 stride on multiple dimensions.

I think my quick hacks work. Need sleep. Will clean up when wake up and double check if this is actually correct.
I think it's simpler than to_movement_ops.

### DIFF
--- a/extra/torch_backend/backend.py
+++ b/extra/torch_backend/backend.py
@@ -49,7 +49,7 @@ def as_strided(tensor:torch.Tensor, size, stride, storage_offset=None):
   # make big ret with stride 1
   ret = ret.reshape(numel, *wow)
   # stupid 0 stride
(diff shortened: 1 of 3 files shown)


geohot:
So I like the `to_movement_ops`, it should be cleaned up and work on all ShapeTrackers.

geohotstan:
Got it.

geohotstan:
Why do tests pass not pass for python and webgpu but pass for the other ones 🤔

chenyuxyz:
python seems to timeout because it's too slow, i think the approach to make a big tensor then reduce in getitem might waste too much memory/ops

PR #9273, opened 2025-02-26
tinychat in browser, Part 1: llama

hooved:
This PR is a subset of [tinychat in browser #8464](https://github.com/tinygrad/tinygrad/pull/8464), and contains these changes relevant to llama:

- Support int8 quantization of embeddings. This lets us shrink the llama-3.2-1B tok_embeddings from 1GB to 262 MB.
- Allow use of either float16 or float32 scales in quantization. Float32 support is important for broad compatibility of tinychat with browsers.
- Enable skipping of loading weights from a file when building the model (so we can later load from a state_dict instead). The next PR that covers model compiling/export (part 2) shows why this is needed: we start by quantizing float16 weights, but float16 is not supported in the dawn/Vulkan/NVIDIA stack (see [here](https://issues.chromium.org/issues/42251215)), which I depend on.
- Enable jitting of a single llama inference function with symbolic start_pos >= 0. This lets us export a single jitted function for all inference in the browser.
- Improve int8 quantization by rounding the scales before casting to int8


chenyuxyz:
nf4 failing, overall looks good

hooved:
Concerning the Mac Benchmark tests that failed ([here](https://github.com/tinygrad/tinygrad/actions/runs/13553662608/job/37883866619?pr=9273)): The failing tests run llama-1-7B, and failed due to the new quantization logic, but I fixed the issue and got the failing tests to succeed locally (CPU backend on Ubuntu).

Could I please have a rerun of the desired tests, such as above, that are not automatically triggered by me pushing? Thanks.

PR #9283, opened 2025-02-27
nanoGPT train works with tiny torch backend

b1tg:
- `python train.py config/train_shakespeare_char.py` works
- `python train.py --batch_size=6 --compile=False` cause Out of memory on a single 3090, `MemoryError: Can't allocate 1073741824 bytes` on tiny10, gonna test on tiny18 when it's idle


geohot:
Wait really? It was only that?!?

extra/torch_backend/backend.py
```diff
@@ -299,6 +299,7 @@ def _wrap_out(*args, **kwargs):
   "aten.scatter.value": Tensor.scatter,
   "aten.gather": Tensor.gather,
   "aten.where.self": Tensor.where,
+  "aten.where.self_out": lambda cond, self, other, out=None: out.replace(Tensor.where(cond, self, other)),
```
geohot, line 302: Put this in the out functions, don't use replace

geohot:
Is the RAM issue also true on the cuda backend?

b1tg:
NV=1 and CUDA=1 both cause out of memory

geohot:
No, I mean the torch cuda backend.

b1tg:
torch cuda backend works fine

geohot:
We should see how off we are, is it a bug or just torch being more effiicient

b1tg:
test with the small params model `python train.py config/train_shakespeare_char.py`, tiny backend much slower than torch cuda backend and use more vram
```
# tiny backend
step 0: train loss 4.2874, val loss 4.2823
iter 0: loss 4.2711, time 195856.62ms, mfu -100.00%
iter 10: loss 3.1484, time 2348.33ms, mfu 0.16%
iter 20: loss 2.7454, time 2579.35ms, mfu 0.16%
iter 30: loss 2.6310, time 2224.50ms, mfu 0.16%
iter 40: loss 2.5872, time 2174.07ms, mfu 0.16% (6.2G vram)
```

```
# torch cuda backend 
step 0: train loss 4.2874, val loss 4.2823
iter 0: loss 4.2681, time 6879.40ms, mfu -100.00%
iter 10: loss 3.1448, time 40.77ms, mfu 9.14%
iter 20: loss 2.7369, time 40.69ms, mfu 9.14%
iter 30: loss 2.6234, time 40.48ms, mfu 9.15%
iter 40: loss 2.5680, time 40.60ms, mfu 9.15% (3.2G vram)
```

geohot:
As in there's a bug with dtypes? I think it's in the scope of bounty to fix that

b1tg:
oh, i mean the torch.amp.autocast stuff, tiny backend does not implement it, i thought the nearly double memory cost was caused by the dtype casting.

I later found that there's a memory leak in tiny backend, which can be reproduced using the beautiful_mnist_torch example:

With TINY_BACKEND=1, vram increases to 9.18GB.

```
01-70 1.15GB->9.18GB
```

with TINY_BACKEND=0, vram remains stable
```
01-08 1.57GB
09-70 4.35GB
```

future investigate shows the vram increase occurs during `loss.backward()`, which, to my knowledge, is related to the computation graph, temporally hack unwrap function can reduce the top memory from 9.18GB to 5.52GB, and nanoGPT can run under this situation with ~23GB nvram.


```python
def unwrap(x:torch.Tensor) -> Tensor:
  assert isinstance(x, torch.Tensor), f"x isn't {type(x)}"
  tt = mod.unwrap(x)
  tt.realize()
  return tt
```

I have not yet found a proper way to actually fix this problem. Is this on the right track?

tocubed:
#9395 should help with the vram leak over iterations

b1tg:
Seems the one-line change has already been added to the master. tiny18 is doing some benchmarking, I'll test when it's idle.

b1tg:
https://github.com/tinygrad/tinygrad/pull/9395 not fix nanoGPT memory leak, nanoGPT OOM during backward pass.

I learned from @tocubed‘s pr, realized all the grad of torch module's parameters after backward pass, which works (python train.py --batch_size=6 --compile=False)

the problem is `register_module_full_backward_hook` [does not work with ReLU ](https://github.com/pytorch/pytorch/issues/61519), cause extra/torch_backend/example.py failed, so i use `register_backward_hook`, but it seems deprecated

tocubed:
Is there still a leak? Or just too high peak vram consumption compared to torch.cuda?
Current solution should be breaking the graph at model buffers and optimizer state and I assume previous grad doesn't carry over past those breaks. If it's still leaking and grad is somewhere in that leaking chain I would like to try and understand why.

b1tg:
@tocubed the nanoGPT OOM is caused by high vram spark during the backward pass, which was fixed by last commit. and i do notice that if you run beautiful_mnist_torch.py for a longer iterations, there is small cumulative memory growth

```
# examples/other_mnist/beautiful_mnist_torch.py
# iter vram    accuracy
# 0    4.51GB  
# 100  12.31GB 98.72%
# 500  12.51GB 99.23%
# 1000 12.75GB 99.35%
# 2000 13.24GB 99.42%
# 4000 14.41GB 99.46%
# 4400 14.42GB 99.46% | 4406/7000 [1:42:10<1:00:09,  0.72it/s]
```

b1tg:
train.py diff for test

```diff
diff --git a/train.py b/train.py
index de57850..3c0e2f6 100644
--- a/train.py
+++ b/train.py
@@ -16,6 +16,11 @@ $ torchrun --nproc_per_node=8 --nnodes=2 --node_rank=1 --master_addr=123.456.123
 (If your cluster does not have Infiniband interconnect prepend NCCL_IB_DISABLE=1)
 """
 
+import sys, tinygrad, pathlib, os
+TINY = os.getenv("TINY_BACKEND") == "1"
+if TINY:
+    sys.path.append(pathlib.Path(tinygrad.__file__).parent.parent)
+    import extra.torch_backend.backend
 import os
 import time
 import math
@@ -69,7 +74,7 @@ min_lr = 6e-5 # minimum learning rate, should be ~= learning_rate/10 per Chinchi
 # DDP settings
 backend = 'nccl' # 'nccl', 'gloo', etc.
 # system
-device = 'cuda' # examples: 'cpu', 'cuda', 'cuda:0', 'cuda:1' etc., or try 'mps' on macbooks
+device = 'tiny' if TINY else 'cuda' # examples: 'cpu', 'cuda', 'cuda:0', 'cuda:1' etc., or try 'mps' on macbooks
 dtype = 'bfloat16' if torch.cuda.is_available() and torch.cuda.is_bf16_supported() else 'float16' # 'float32', 'bfloat16', or 'float16', the latter will auto implement a GradScaler
 compile = True # use PyTorch 2.0 to compile the model to be faster
 # -----------------------------------------------------------------------------
@@ -106,10 +111,10 @@ if master_process:
 torch.manual_seed(1337 + seed_offset)
 torch.backends.cuda.matmul.allow_tf32 = True # allow tf32 on matmul
 torch.backends.cudnn.allow_tf32 = True # allow tf32 on cudnn
-device_type = 'cuda' if 'cuda' in device else 'cpu' # for later use in torch.autocast
+device_type = 'cuda' if 'cuda' in device else 'tiny' if TINY else 'cpu' # for later use in torch.autocast
 # note: float16 data type will automatically use a GradScaler
 ptdtype = {'float32': torch.float32, 'bfloat16': torch.bfloat16, 'float16': torch.float16}[dtype]
-ctx = nullcontext() if device_type == 'cpu' else torch.amp.autocast(device_type=device_type, dtype=ptdtype)
+ctx = nullcontext() if device_type == 'cpu' or TINY else torch.amp.autocast(device_type=device_type, dtype=ptdtype)
 
 # poor man's data loader
 data_dir = os.path.join('data', dataset)
@@ -219,6 +224,7 @@ def estimate_loss():
     for split in ['train', 'val']:
         losses = torch.zeros(eval_iters)
         for k in range(eval_iters):
+            print(f"eval_iters {k}/{eval_iters}")
             X, Y = get_batch(split)
             with ctx:
                 logits, loss = model(X, Y)
@@ -290,6 +296,7 @@ while True:
     # forward backward update, with optional gradient accumulation to simulate larger batch size
     # and using the GradScaler if data type is float16
     for micro_step in range(gradient_accumulation_steps):
+        print(f"micro_step {micro_step}/{gradient_accumulation_steps}")
         if ddp:
             # in DDP training we only need to sync gradients at the last micro step.
             # the official way to do this is with model.no_sync() context manager, but

```

```
TINY_BACKEND=1 python train.py --batch_size=6 --compile=False
```

geohot:
Very slow, but seems to work. Bounty is yours, george@tinygrad.org to claim, can pay via PayPal or USDC on ETH

PR #9296, opened 2025-02-28
Make TYPED=1 work

utkarshgill:
add support for runtime type checking in tinygrad using typeguard #7889


utkarshgill:
used the @typechecked decorator approach only, not the import hook

this is because tensor.py uses conditional imports of numpy in several places, which conflicts with the import hook mechanism. installing import hook causes issues with scoped numpy imports in tensor.py

PR #9297, opened 2025-02-28
torch fix casting and add ops for sd vae(s)

tocubed:
Fix copy to cast as that is the expected behavior [here](https://github.com/pytorch/pytorch/blob/af720cd5a7edafbf3ee791ce4aaa52a5da2049f6/torch/csrc/lazy/ts_backend/tensor_aten_ops.cpp#L56).

After this commit, the image AE models from https://github.com/madebyollin/taesd should work. Tested in ComfyUI, quite simple to patch in tinygrad. Tried an actual SDXL VAE and it runs but produces incorrect results.

### DIFF
--- a/extra/torch_backend/backend.py
+++ b/extra/torch_backend/backend.py
@@ -140,16 +140,16 @@ def convolution_backward_overrideable(grad_out, input, weight, stride, padding, 
   return tuple([wrap(grads.pop(0)) if m else None for m in output_mask])
 
 @torch.library.impl("aten::_copy_from", "privateuseone")
-def _copy_from(src, dest, non_blocking=False):
+def _copy_from(src: torch.Tensor, dest, non_blocking=False):
+  cast_dtype = _from_torch_dtype(dest.dtype)
   if str(src.device) == "tiny" and str(dest.device) == "tiny":
-    dest, src = unwrap(dest), unwrap(src)
-    dest.replace(src.cast(dest.dtype), allow_shape_mismatch=True)
+    unwrap(dest).replace(unwrap(src).cast(cast_dtype), allow_shape_mismatch=True)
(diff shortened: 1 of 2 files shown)


extra/torch_backend/backend.py
```diff
@@ -142,7 +142,8 @@ def convolution_backward_overrideable(grad_out, input, weight, stride, padding,
 @torch.library.impl("aten::_copy_from", "privateuseone")
 def _copy_from(src, dest, non_blocking=False):
   if str(src.device) == "tiny" and str(dest.device) == "tiny":
```
chenyuxyz, line 144: the cast should happen regardless of the device (before if blocks) right?

chenyuxyz (commented):
do you know why sdxl output is different?

extra/torch_backend/backend.py
tocubed: torch cast recurses back into this (at least when I tried it crashed without traceback…). need to know it’s a tiny device to do the cast, although now looking again I’m not sure if that 3rd block does the cast implicitly actually 

I’ll add a test

tocubed:
> do you know why sdxl output is incorrect?

narrowed it down to the encoder but no further, it may fix itself with the test_ops fixes but if not I’ll revisit

tocubed:
Might have issues with this _copy_from and non-contiguous tensors, which admittedly seems to be rare scenario so far.
SDXL VAE works after a adding tiny implementation for aten.pad. Whatever torch was doing instead was broken, must be one of the other ops.

image

extra/torch_backend/backend.py
Anish9901: this change _does_ seem to fix ~10 additional tests in #9302 which were failing due to dtype mismatch.

chenyuxyz:
cool thanks!

PR #9299, opened 2025-02-28
mod only on int [pr]

nimlgen:
@Qazalin I get this right? we want to support MOD only for ints?


Qazalin:
hmm we should match torch:
```
python -c "import torch; print((torch.tensor([1.0, 2.0])%2).numpy())"
```

nimlgen:
is this Ops.MOD a spec for hw, right? Do we want to handle that at codegen?
Any arch can do mod on float? 

I do not see any instructions in ptx/sass/rdna
like ptx:
`rem.type  d, a, b;
.type = { .u16, .u32, .u64,
          .s16, .s32, .s64 };`

nimlgen:
We can prob can just do `a-b*floor(a/b)` for float somewhere higher level?

Qazalin:
yea I'd put it in symbolic, scheduler will rewrite the tensors. This `spec` is shared across tensors+ast+hw (rendererd) ops.
A prereq for this is floor_div.

chenyuxyz:
done in #9306

PR #9301, opened 2025-02-28
Test Onnx quantization behavior

geohotstan:
~so it appears qlinearop is round away from zero while quantize linear is round to even~
I think it's all round to even...

### DIFF
--- a/test/external/external_test_onnx_backend.py
+++ b/test/external/external_test_onnx_backend.py
@@ -50,7 +50,7 @@ backend_test.exclude('test_qlinearmatmul_3D_int8_float16_cpu')
 backend_test.exclude('test_qlinearmatmul_2D_int8_float32_cpu')
 backend_test.exclude('test_qlinearmatmul_3D_int8_float32_cpu')
 
-# BUG: we don't match ORT here because const folding is done in float64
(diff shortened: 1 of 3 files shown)


geohotstan:
this passes locally with the same python version, onnx version, and onnxruntime version. What.

test/external/external_test_onnx_backend.py
```diff
@@ -50,6 +50,10 @@ def supports_device(cls, device: str) -> bool:
 backend_test.exclude('test_qlinearmatmul_2D_int8_float32_cpu')
 backend_test.exclude('test_qlinearmatmul_3D_int8_float32_cpu')
 
+# BUG: we don't match ORT here due to some div inaccuracy with floats
+backend_test.exclude('test_dynamicquantizelinear_cpu')
+backend_test.exclude('test_dynamicquantizelinear_expanded_cpu')
```
geohotstan, line 55: this is due to slight numerical inaccuracy's sensitivity to rounding.
```
In [145]: (Tensor(-2.5) / (Tensor(5.0) / 255)).item()
Out[145]: -127.5
In [159]: (np.float32(-2.5) / (np.float32(5.0) / np.float32(255))).item()
Out[159]: -127.49999237060547
```

PR #9305, opened 2025-02-28
Fix TRACEMETA=0 in Context

utkarshgill:
When using with Context(TRACEMETA=0) metadata was still being saved

This doesn't match what was expected in #6918

In ops.py, the metadata was being stored regardless of TRACEMETA value so we check TRACEMETA before storing anything

### DIFF
--- a/test/test_tracemeta.py
+++ b/test/test_tracemeta.py
@@ -7,41 +7,40 @@ from tinygrad.helpers import ContextVar
 
 class TestTraceMeta(unittest.TestCase):
   def test_tracemeta_default(self):
-    # Test default behavior (TRACEMETA=1)
     t = Tensor([1,2,3,4])
     t2 = t + t
-    # There should be metadata for the add operation
+
+    # should be metadata for the add operation
     self.assertTrue(any(all_metadata.values()))
-    
+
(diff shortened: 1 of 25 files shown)


tinygrad/__init__.py
```diff
@@ -7,5 +7,5 @@
 from tinygrad.ops import UOp
 Variable = UOp.variable
 from tinygrad.dtype import dtypes                                     # noqa: F401
-from tinygrad.helpers import GlobalCounters, fetch, Context, getenv   # noqa: F401
+from tinygrad.helpers import GlobalCounters, fetch, Context, getenv, TRACEMETA   # noqa: F401
```
geohot, line 10: No

test/test_tracemeta.py
```diff
@@ -0,0 +1,44 @@
...
+    # back to default value
+    self.assertEqual(TRACEMETA.value, 1)
+
+  def test_tracemeta_global_zero(self):
+    all_metadata.clear()
+
+    original_value = TRACEMETA.value
+    try:
+      # set TRACEMETA to 0 globally
+      ContextVar._cache["TRACEMETA"].value = 0
```
geohot, line 33: wIth Context

Issue #9262, opened 2025-02-26
Changes requested from pytorch backend


geohot:
Methods to add to tensor.py
* as_strided -- https://pytorch.org/docs/stable/generated/torch.as_strided.html
* topk -- https://pytorch.org/docs/main/generated/torch.topk.html
* bitwise_xor -- https://pytorch.org/docs/stable/generated/torch.bitwise_xor.html
* var_mean -- https://pytorch.org/docs/stable/generated/torch.var_mean.html

Methods to add features to
* max_pool2d -- add support for return_indices

Methods to rename
* leakyrelu -> leaky_relu -- https://pytorch.org/docs/stable/generated/torch.nn.functional.leaky_relu.html

The indexing situation in tinygrad needs to be dealt with also.

chenyuxyz:
- [ ] as_strided
- [ ] topk
- [ ] bitwise_xor
- [ ] var_mean
- [ ] max_pool2d
- [ ] leaky_relu

geohot:
We need to support << and >>, all three of:
(Tensor) << (Scalar)
(Scalar) << (Tensor)
(Tensor) << (Tensor)

eichenroth:
- [ ] copysign

PR #9310, opened 2025-03-01
Topk

utkarshgill:
#9262


tinygrad/tensor.py
```diff
@@ -2006,6 +2007,68 @@ def argmin(self, axis=None, keepdim=False):
...
+    # handle negative dim
+    if dim < 0:
+      dim += x.ndim
+    
+    # validate k
+    dim_size = x.shape[dim]
+    assert k <= dim_size, f"k ({k}) is too large for dimension {dim} of size {dim_size}"
+    
+    # get numpy array for sorting
+    xnp = x.numpy()
```
geohot, line 2054: This is not in tinygrad and defeats the whole point

PR #9312, opened 2025-03-01
WIP: linearzer rewrite

Tzvetomir-Kaydjiev:
WIP: Global rewrite of SINK.

This attempts to split the compute graph on multiple scopes before the linearization process starts. I am not sure if this is possible with local changes to the graph, because this scopes can be overlapping for this implementation.

For example the case for TestOps.test_avg_pool3d_failure:

ASSIGN1  -- depends -- > RANGE1, RANGE2
ASSIGN2  -- depends -- > RANGE1
ASSIGN3  -- depends -- > RANGE2

We can try to make with by shoving everything in a nested loop:
```
Loop RANGE1 {
  Loop RANGE2 {
    ASSIGN1
    ASSIGN2
    ASSIGN3
  }
}
```

or we can have multiple loops with duplicate operations for RANGE, LOAD ... :
```
Loop RANGE1 {
  Loop RANGE2 {
    ASSIGN1
  }
}
Loop RANGE1 {
  ASSIGN2
}
Loop RANGE2 {
  ASSIGN3
}
```

The current implementation tries to do the first variant and fails because can't shove a ASSIGN2 and ASSIGN3 because one must be in the nested loop.

This scratch implementation tries to do the second approach.

Opinion?

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -2351,7 +2351,7 @@ class TestOps(unittest.TestCase):
       lambda x: Tensor.avg_pool2d(x, kernel_size=(111,28)), rtol=1e-5)
 
   # TODO: linearizer block error
-  @unittest.expectedFailure
+  # @unittest.expectedFailure
   def test_avg_pool3d_failure(self):
     with Context(NOOPT=0):
       helper_test_op([(1,1,16,16,16)],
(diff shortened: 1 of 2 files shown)


geohot:
Can you use rewrite rules?

Tzvetomir-Kaydjiev:
I added rewrite if VIZ is enabled, also fixed some problems with the new implementation, but I think there are still some problems to fix with the implicit dependencies of the Ops.STORE operations. I can add more more rewrite rules for the final SINK transformation, but the scope generation and scope merger, probably can not be broken down.

geohot:
Should be a graph rewrite. Also, with linearizer misspelled in the PR title I have very little faith in this. Before trying big rewrites, become a known contributor with bugfixes and small changes.

PR #9317, opened 2025-03-02
MultiGPU training tiny torch backend

tocubed:
Messy but working with `GPUS=4 LLVM=1 TINY_BACKEND=1 python3 examples/other_mnist/beautiful_mnist_torch.py`
Quite a few stubs and TODOs to work on.

### DIFF
--- a/examples/other_mnist/beautiful_mnist_torch.py
+++ b/examples/other_mnist/beautiful_mnist_torch.py
@@ -43,6 +43,8 @@ if __name__ == "__main__":
   model = Model().to(device)
   optimizer = optim.Adam(model.parameters(), 1e-3)
 
+  if (GPUS:=getenv("GPUS", 1)) > 1: model = torch.nn.DataParallel(model, range(GPUS))
+
   loss_fn = nn.CrossEntropyLoss()
   #@torch.compile
   def step(samples):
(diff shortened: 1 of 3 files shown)


tocubed:
Investigating the training slowing down to a crawl as it progresses.

geohot:
Nice! Does DistributedDataParallel work also?

tocubed:
> Nice! Does DistributedDataParallel work also?

Not yet, work in progress. A few tricky things and more C++ code than I'd like, but close to something running, just need to get a basic multiprocess gather/scatter/reduce with tiny tensors.

tocubed:
Issue narrowed down to torch backend not supporting views for out or dest tensors. Working on fix.
DDP uses slices of larger tensors to accumulate grads from different layers.

tocubed:
This PR now works in the sense that DDP is training mnist with the correct accuracy, but still very much a WIP. I need to implement some small fixes and it should pass CI assuming we bump the timeout.

I am going to take a few more steps to cleanup and test more. The current implementation of the distributed ops falls back to CPU torch...  I'd like to see what it'd take not to do that. And maybe find out why DDP is currently slower on CI (faster on my machine but then again I have more cores...)

tocubed:
If my bisect is right, some behavior changed in #9845. Loss diverges to nan now, looking into why...

chenyuxyz:
stale

PR #9320, opened 2025-03-02
Failing test for not realizing intermediate expand in multi-GPU

KhanerX:
In multi, when the result of an operation is needed on another device, we copy it on the destination device after the operation. this is fine for most cases, and is handled well in scheduling. 

However, in case of a binary op that expands its inputs, 
There's no rule in scheduler that moves copy before binary op. 

and really it doesn't make sense as a rule in scheduler. I think it should be a rule in get_multi_map. because there we expect this type of copy between devices, and moving copies to the beginning (after buffer) makes sense (it's basically all_gather).

I visualize what I mean,
Executing this code
```python
a = Tensor.empty(16, 1).shard_(devices_2, axis=0)
b = Tensor.empty(16, 16).to_(devices_2)
c = Tensor.empty(16, 16).shard_(devices_2, axis=1)
d = a+b
(d*c).realize()
```
This is the result of current get_multi_map:
image

And after scheduling, copies are still after adds and adds get realized (hence failing test)

===================================================================

What I think should happen (after get_multi_map):
image

Then after scheduling, graph will look like this and we don't realize expand:
image

### DIFF
--- a/test/test_multitensor.py
+++ b/test/test_multitensor.py
@@ -782,6 +782,15 @@ class TestMultiTensor(unittest.TestCase):
     run_schedule(sched)
     self.assertListEqual(b.tolist(), [0, 0, 0])
 
+  @unittest.expectedFailure
+  def test_dont_realize_intermediate_expand(self):
+    a = Tensor.empty(16, 1).shard_(devices_2, axis=0)
+    b = Tensor.empty(16, 16).to_(devices_2)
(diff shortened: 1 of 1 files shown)


Qazalin:
Thanks!
This requires a more generic version of the UPat we have for COPY(DEVICE, VIEW) reordering:

https://github.com/tinygrad/tinygrad/blob/00028e87bb05d610b40883b17c9877e00e35d922/tinygrad/engine/schedule.py#L80-L82

PR #9326, opened 2025-03-02
Add consistent typing for python 3.10 in tensor.py

eichenroth:
- Replace all `Union` with `|`
- Replace all `Optional` with `...|None`


chenyuxyz:
conflict

PR #9335, opened 2025-03-03
Move topk, _index_put_impl_, index_tensor, masked_select, and randperm_generator off of cpu in torch_backend

khushi-411:
Hi,

I found this repository very interesting and came across this open bounty via X yesterday. Here's my small attempt to fix it.

Would love to hear your suggestions and feedback on the PR!

Thank you!

cc @geohot


chenyuxyz:
the goal is to move those to tinygrad

PR #9340, opened 2025-03-03
torch backend cleanup: remove duplicate dict keys

eichenroth:
still 57 failing tests


chenyuxyz:
things like this is hard to review, removing stuff and test remained the same might or might not have hidden impacts

eichenroth:
@chenyuxyz I literally removed duplicate keys that got linted - opening the file in an editor would instantly prove that this was the right choice

image
image

image

PR #9343, opened 2025-03-04
add Topk to tensor

geohotstan:
Need to think about this a little more.
Works for some default use-cases. Fails for edge cases.

There seems to be `return_index` use cases for other ops too (maxpool, cummax).
I hacked together a `return_index` for maxpool before but it was really really ugly, as is this starting implementation.
It would be nice if we can factor out a common way to get the indices.


chenyuxyz:
https://github.com/pytorch/pytorch/issues/88184

geohotstan:
> [pytorch/pytorch#88184](https://github.com/pytorch/pytorch/issues/88184)
> 
> and looks like other frameworks pick lowest indices

thanks for helping me investigate :D 

I've been looking at all the new UOp abstractions.
Maybe direct lowering, meaning without doing arange masking, of Tensor indexing can be done now, maybe...... (only bool tensor indexing has data dependent output shape)
Still haven't gotten it to work yet.

chenyuxyz:
cool!

PR #9344, opened 2025-03-04
VIZ lightmode

DKormann:


DKormann:
Screenshot 2025-03-04 at 14 42 35
automatic system preferred color theme

geohot:
@Qazalin this harmless?

Qazalin:
yea, having two color schemes makes one easily breakable. I think we'll keep the dark only for now.

PR #9347, opened 2025-03-04
Adding fixes and features to the GPT2 training example. Also wrote Tensor.topk()

PKD667:
Added a `--device` argument for the GPT2 training example.
Fixed the generate function with topk_k.

Needed topk for gpt2 generate so I implemented `Tensor.topk()`
The implementation is a bit slow, but faster than argmax is guess.


tinygrad/tensor.py
```diff
@@ -2049,6 +2049,113 @@ def parse_formula(formula:str, *operands:Tensor):
...
+          for item in items:
+              heap_push(heap, item)
+          return heap
+      
+      n = tensor.shape[0]
+      if k <= 0:
+          return Tensor([]), Tensor([])
+
+      initial_k = min(k, n)
+      heap = heapify([(tensor[i].item(), i) for i in range(initial_k)])
```
geohot, line 2123: No!

PR #9357, opened 2025-03-05
use am in training benchmarks

nimlgen:


nimlgen:
after the hdp fix this looks fine

PR #9358, opened 2025-03-05
Fix test_ops for tiny backend part 2

Anish9901:
Related to #9302

101 --> 75 failures

### DIFF
--- a/extra/torch_backend/backend.py
+++ b/extra/torch_backend/backend.py
@@ -328,11 +328,11 @@ tiny_backend = {**{k:wrap_out(v) for k,v in tiny_backend_out.items()}, **{
   "aten.var_mean.correction": Tensor.var_mean,
   # NOTE: axis=[] in torch means all, change tinygrad?
   "aten.sum.IntList_out": lambda self,axis,keepdim=False,dtype=None,out=None:
-    out.replace(Tensor.sum(self, axis if axis is None or len(axis) else None, keepdim,
-                           acc_dtype=_from_torch_dtype(dtype) if dtype is not None else dtypes.default_float), allow_shape_mismatch=True),
+    out.replace(self.sum(axis if axis is None or len(axis) else None, keepdim,
+                         acc_dtype = _from_torch_dtype(dtype) if dtype is not None else None), allow_shape_mismatch=True),
   "aten.scatter.value": Tensor.scatter,
   "aten.scatter.value_reduce": Tensor.scatter,
-  "aten.gather": Tensor.gather,
+  "aten.gather": lambda self, dim, index: self.gather(dim, index.cast(dtypes.int)),
   "aten.where.self": Tensor.where, # NOTE: this is needed as well as the out type
   "aten._softmax": lambda self,dim,half_to_float: self.softmax(dim),
   "aten._log_softmax": lambda self,dim,half_to_float: self.log_softmax(dim),
@@ -345,11 +345,11 @@ tiny_backend = {**{k:wrap_out(v) for k,v in tiny_backend_out.items()}, **{
   # these don't work in out form, they have size 0
   "aten.abs": Tensor.abs,
   "aten.logical_not": Tensor.logical_not,
-  "aten.logical_or_": lambda x, y: x | y,
+  "aten.logical_or_": lambda x, y: x.assign(x | y),
   "aten.multinomial": Tensor.multinomial,
   "aten.pad": Tensor.pad,
   "aten.reflection_pad2d": functools.partial(Tensor.pad, mode="reflect"),
-  "aten.masked_fill_.Scalar": lambda self, mask, value: self.assign(Tensor.masked_fill(self, mask, value)),
+  "aten.masked_fill_.Scalar": lambda self, mask, value: self.assign(self.masked_fill(mask, value)),
   "aten.masked_fill.Scalar": Tensor.masked_fill,
   "aten.masked_fill.Tensor": Tensor.masked_fill,
   "aten.all": Tensor.all,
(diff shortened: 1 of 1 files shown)


extra/torch_backend/backend.py
```diff
@@ -270,8 +271,8 @@ def cat_out(tensors, dim=0, out=None):
   # TODO: this might result in overflow issues
   "aten.round.decimals_out": lambda self,decimals: (self*10**decimals).round()/10**decimals,
   # TODO: support this in tinygrad
-  "aten.bitwise_left_shift.Tensor_out": lambda input,other: Tensor(input << other.numpy()),
-  "aten.bitwise_right_shift.Tensor_out": lambda input,other: Tensor(input >> other.numpy()),
+  "aten.bitwise_left_shift.Tensor_out": lambda x,y: x*(2**y),
```
chenyuxyz, line 274: does this pass more test? i don't think it work in tinygrad

extra/torch_backend/backend.py
```diff
@@ -326,8 +327,9 @@ def _wrap_out(*args, **kwargs):
   "aten.var.correction": Tensor.var,
   "aten.var_mean.correction": Tensor.var_mean,
   # NOTE: axis=[] in torch means all, change tinygrad?
-  "aten.sum.IntList_out": lambda self,axis,keepdim=False,out=None:
-    out.replace(Tensor.sum(self, axis if axis is None or len(axis) else None, keepdim), allow_shape_mismatch=True),
+  "aten.sum.IntList_out": lambda self,axis,keepdim=False,dtype=None,out=None:
```
chenyuxyz, line 330: sum of int should not have default_float for acc dtype

extra/torch_backend/backend.py
```diff
@@ -343,10 +345,11 @@ def _wrap_out(*args, **kwargs):
   # these don't work in out form, they have size 0
   "aten.abs": Tensor.abs,
   "aten.logical_not": Tensor.logical_not,
+  "aten.logical_or_": lambda x, y: x | y,
```
chenyuxyz, line 348: inplace?

extra/torch_backend/backend.py
```diff
@@ -361,13 +364,24 @@ def _wrap_out(*args, **kwargs):
   "aten.atanh": Tensor.atanh,
   "aten.fill_.Tensor": Tensor.full,
   "aten.flip": Tensor.flip,
+  "aten.scatter_reduce.two": Tensor.scatter_reduce,
+  "aten.squeeze_.dim": lambda self, dim: self.replace(Tensor.squeeze(self, dim), allow_shape_mismatch=True),
+  "aten.add.Tensor": lambda input,other,alpha=1: input+alpha*other,
+  "aten.linspace": lambda start, stop, steps, dtype=None, **kwargs:
+    Tensor.linspace(start, stop, steps, dtype=_from_torch_dtype(dtype) if dtype is not None else dtypes.default_float),
+  "aten::view.dtype": lambda self, dtype: Tensor.bitcast(self, _from_torch_dtype(dtype)),
```
chenyuxyz, line 372: just `self.bitcast`, similar for others

extra/torch_backend/backend.py
```diff
@@ -361,13 +364,24 @@ def _wrap_out(*args, **kwargs):
...
+  "aten.constant_pad_nd": lambda self, padding, value=0.0: Tensor.pad(self, padding, mode="constant", value=value),
+  "aten.logsumexp": lambda self, axis, keepdim=False: Tensor.logsumexp(self, *axis, keepdim=keepdim),
```
chenyuxyz, line 374: are you sure about `*axis`?

chenyuxyz (commented):
please check carefully, and double check with real torch code when no sure. it needs to be correct in principle and not just passing test_ops

extra/torch_backend/backend.py
```diff
@@ -343,10 +345,11 @@ def _wrap_out(*args, **kwargs):
   # these don't work in out form, they have size 0
   "aten.abs": Tensor.abs,
   "aten.logical_not": Tensor.logical_not,
+  "aten.logical_or_": lambda x, y: x | y,
   "aten.multinomial": Tensor.multinomial,
   "aten.pad": Tensor.pad,
   "aten.reflection_pad2d": functools.partial(Tensor.pad, mode="reflect"),
-  "aten.masked_fill_.Scalar": lambda self,mask,value: self.assign(mask.where(self, value)),
+  "aten.masked_fill_.Scalar": lambda self, mask, value: self.assign(Tensor.masked_fill(self, mask, value)),
```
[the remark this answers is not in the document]
Anish9901: I thought it would be more descriptive now that tinygrad supports `masked_fill`, I'll revert it.

extra/torch_backend/backend.py
Anish9901: It does pass significantly more tests, there are dtype mismatches withouh it, and the test fail with: `Exception: forward pass failed shape (): dtype mismatch: tinygrad=float32 | torch=int32` or similar. 

A similar version to this already exists on master:
https://github.com/tinygrad/tinygrad/blob/2af129c078dde57a8a96f8ac1677c1632dc0130b/extra/torch_backend/backend.py#L308-L311

extra/torch_backend/backend.py
Anish9901: Ah, you're right, will fix that.

extra/torch_backend/backend.py
```diff
@@ -361,13 +364,24 @@ def _wrap_out(*args, **kwargs):
   "aten.atanh": Tensor.atanh,
   "aten.fill_.Tensor": Tensor.full,
   "aten.flip": Tensor.flip,
+  "aten.scatter_reduce.two": Tensor.scatter_reduce,
+  "aten.squeeze_.dim": lambda self, dim: self.replace(Tensor.squeeze(self, dim), allow_shape_mismatch=True),
+  "aten.add.Tensor": lambda input,other,alpha=1: input+alpha*other,
+  "aten.linspace": lambda start, stop, steps, dtype=None, **kwargs:
+    Tensor.linspace(start, stop, steps, dtype=_from_torch_dtype(dtype) if dtype is not None else dtypes.default_float),
```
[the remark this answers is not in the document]
chenyuxyz: i see, passing None instead of repeating the logic seems better

extra/torch_backend/backend.py
chenyuxyz: eh what if axis is a tuple with length more than 2?

Anish9901: it can only contain 1 int according to torch's function signature:

Screenshot 2025-03-05 at 10 53 32 PM

I can change it to `axis[0]` if you prefer

chenyuxyz: i see, yea axis[0] is more clear

extra/torch_backend/backend.py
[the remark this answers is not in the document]
Anish9901: Oh, I think I remember why I didn't add None here, `to_dtype(None)` is invalid, I would have to change the function to something like this:
`Tensor.linspace(start, stop, steps, **({"dtype": _from_torch_dtype(dtype)} if dtype is not None else {}))`

PR #9276, opened 2025-02-26
tinychat in browser, Part 3: browser app

hooved:
- This PR is a subset of [tinychat in browser #8464](https://github.com/tinygrad/tinygrad/pull/8464)
- This PR depends on:
    - [tinychat in browser, Part 1: llama #9273](https://github.com/tinygrad/tinygrad/pull/9273), which is now contained in this branch, as part 1 was merged into master, which was merged back into this branch.
    - [tinychat in browser, Part 2: model export #9274](https://github.com/tinygrad/tinygrad/pull/9274)

This PR contains the tinychat browser app layer:

- Complete client-side (browser) implementation of llama-3.2-1B inference
- Support for both WebGPU and WASM backends in the browser
- Supports running a 1.2 GB model in smartphone browsers, with optimizations for stability:
    - Minimize memory overhead by downloading weights in small chunks, and loading the weight chunks into model buffers as they arrive (versus previous pattern of downloading all weights before loading to model, which would have doubled memory overhead)
    - Make WASM model-loading stable on mobile browsers, [which is notorious](https://github.com/WebAssembly/design/issues/1397), by allocating and filling WASM memory with a specific pattern
- Includes several improvements over previous tinychat python app:
    - Prevent streaming of gibberish when max_context is reached
    - Enable user to stop model generation (exit a conversation) by navigating back
    - Add prefill progress bar, so user doesn't think app is frozen for long prompts or when using slower WASM
- Includes compilation of tiktoken.js (tokenizer) dependency
- Includes missing font for tinychat, which was probably omitted previously by mistake due to the repo's `.gitignore`, which I updated

To reproduce compiling tiktoken.js, hosting the browser app locally, and running the app, see the [README](https://github.com/tinygrad/tinygrad/blob/b4c5c5ecf01a2a6ee312d16fc1a89c9d4a379ffa/examples/tinychat/tinychat-browser/README.md).

PR #9361, opened 2025-03-05
hotfix: load >2 GiB from disk on macOS

hooved:
This PR fixes a bug where on macOS, with disk buffers > 2 GiB, not all data was copied, when copying from disk device to CPU device.

Reproducer: (second assertion fails on macOS before this diff):

```
import os
from tinygrad import Tensor, dtypes, Device
model_path = str(fetch("https://huggingface.co/bartowski/Llama-3.2-1B-Instruct-GGUF/resolve/main/Llama-3.2-1B-Instruct-f16.gguf", "Llama-3.2-1B-Instruct-f16.gguf", subdir="llama3-1b-instruct"))
Device.DEFAULT="CPU"
start = 2425053024

x = Tensor.empty(os.stat(model_path).st_size, dtype=dtypes.uint8, device=f"disk:{model_path}")
assert not all(i==0 for i in x[start:].tolist())

y = Tensor.empty(os.stat(model_path).st_size, dtype=dtypes.uint8, device=f"disk:{model_path}").to(Device.DEFAULT)
assert not all(i==0 for i in y[start:].tolist())
```

I think this bug is what was causing the issue geohot reported in [tinychat in browser, Part 3: browser app #9276](https://github.com/tinygrad/tinygrad/pull/9276), which I was able to reproduce (with WASM, on an intel macbook), and which is resolved for me with this diff.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -664,6 +664,9 @@ jobs:
         run: python3 -m pytest -n=auto test/ --ignore=test/models --ignore=test/unit --durations=20
       - name: Run process replay tests
         uses: ./.github/actions/process-replay
+      - name: Run macOS-specific unit test
+        if: matrix.backend == 'cpu'
+        run: python3 -m pytest test/unit/test_disk_tensor.py::TestDiskTensor::test_copy_to_cpu_not_truncated
 
 # ****** Windows Tests ******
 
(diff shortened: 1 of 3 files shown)


nimlgen:
Can you add a test? Fine to skip it for CI if it takes too long

hooved:
Here is the new test failing in CI without the fix: [test fail](https://github.com/tinygrad/tinygrad/actions/runs/13703982567/job/38324670096)

This test succeeds after applying the fix to `tinygrad/runtime/ops_disk.py` in this PR.

geohot:
Ahh, I see...sort of a Python bug.

PR #9364, opened 2025-03-05
lower test_gemm_8192 amd_tflops to 65

chenyuxyz:


chenyuxyz:
hmm 8192 is slower than 4096

PR #9371, opened 2025-03-06
torch: support in-place operations on views

tocubed:
Torch in-place operations on views aren't common but they do happen.
This spawned off from #9317 since DistributedDataParallel internally uses views for its gradients and in-place updates.

TODOs:

- [x] Wrap all out functions with this support
- [x] More tests around multiple views
- [x] Sanity check against DDP

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -175,8 +175,8 @@ jobs:
       run: PYTHONPATH=. DEBUG=2 python3 extra/torch_backend/torch_tests.py TestTinyBackendPRIVATEUSE1.test_unary_log_tiny_float32
     - name: Test Ops with TINY_BACKEND (expect failure)
       run: PYTHONPATH=. LLVM=1 LLVMOPT=0 TINY_BACKEND=1 python3 -m pytest -n auto test/test_ops.py --durations=20 || true
-    - name: Test in-place operations on views (expect failure)
-      run: PYTHONPATH=. TORCH_DEBUG=1 python3 extra/torch_backend/test_inplace.py || true
+    - name: Test in-place operations on views
+      run: PYTHONPATH=. TORCH_DEBUG=1 python3 extra/torch_backend/test_inplace.py
 
   torchbackendmore:
     name: Torch Backend Tests More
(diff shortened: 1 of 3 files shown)


geohot:
Looks like CI is timing out with this...expected?

tocubed:
It's definitely a bug. I fixed the slowest tests and cut time in half locally (exceptions led to printing of uint8 tensors, which is extremely slow for some reason, will file a separate issue), yet issues persists on CI. Working on it.

geohot:
Yea printing calls slice and `.item()` a ton of times...don't know if that's how it's supposed to work or if there's a faster path we can enable.

tocubed:
Hi @Qazalin is it intended that BUFFER_VIEW is not in VIEW tensor_uop spec?

Started seeing this after recent commits, on a test I'm adding for torch backend. `extra/torch_backend/test_inplace.py TestTorchBackendInplace.test_noncontig`
```
   0 Ops.DEVICE          : dtypes.void                    []                               LLVM
   1 Ops.UNIQUE          : dtypes.void                    []                               4
   2 Ops.BUFFER          : dtypes.long                    [0, 1]                           16
   3 Ops.VIEW            : dtypes.long                    [2]                              ShapeTracker(views=(View(shape=(4, 4), strides=(4, 1), offset=0, mask=None, contiguous=True),))
   4 Ops.RESHAPE         : dtypes.long                    [3]                              (16,)
   5 Ops.SHRINK          : dtypes.long                    [4]                              ((4, 16),)
   6 Ops.SHRINK          : dtypes.long                    [5]                              ((0, 8),)
   7 Ops.RESHAPE         : dtypes.long                    [6]                              (2, 4)
   8 Ops.UNIQUE          : dtypes.void                    []                               1
   9 Ops.BUFFER          : dtypes.long                    [0, 8]                           16
  10 Ops.BUFFER_VIEW     : dtypes.long                    [9]                              (8, 4)
  11 Ops.VIEW            : dtypes.long                    [10]                             ShapeTracker(views=(View(shape=(2, 4), strides=(4, 1), offset=0, mask=None, contiguous=True),))
  12 Ops.ASSIGN          : dtypes.long                    [7, 11]                          None
  13 Ops.SINK            : dtypes.void                    [12]                             None
```

Adding to the spec and running... tests seem to pass but I'm not sure if the above pattern is right to begin with.

Qazalin:
hmm, I see the problem, #9387 should fix the issue you saw.

tocubed:
Thanks!

PR #9378, opened 2025-03-07
Boolean indexing

PKD667:
Adding boolean indexing to tensors.


PKD667:
these fails are weird, maybe my tests are taking too long ?

PKD667:
I think the linter CI fail is a bug. It doesn't understand the unpacking operator.

tinygrad/tensor.py
```diff
@@ -2049,6 +2054,44 @@ def parse_formula(formula:str, *operands:Tensor):
...
+    flat_t = self.reshape(-1)
+    # Create a tensor of 1’s for nonzero elements and 0’s otherwise.
+    if flat_t.dtype == dtypes.bool:
+      m = flat_t.cast(dtype=dtypes.int32)
+    else:
+      m = (flat_t != 0).cast(dtype=dtypes.int32)
+    # Compute the cumulative sum along the flattened dimension.
+    cumsum = m.cumsum(0)
+    # The last element of cumsum is the count of nonzero items.
+    nonzero_count = cumsum[-1].item()
```
geohotstan, line 2075: `.item()` realizes the Tensor and breaks the compute graph

bool Tensor indexing has been a hard nut to crack for quite sometime now. 
The main problem has been the output shape depends on the values of the index Tensor.

PKD667: in this case, without an Op that can act on tensor shapes based on another tensor, its probably impossible. Maybe I should try a hacky solution that abuses REDUCE_AXIS to do some dynamic reshaping

geohotstan: oh I think it might be fine to accept a graph break? <https://github.com/geohotstan/tinycorp-meetings/blob/master/2025-03-03/README.md#geohot-000532>
Maybe this `.item()` is fine 😄

geohot:
You literally have a print in this code. Please read your diffs 3 times before posting PRs

PR #9365, opened 2025-03-05
gfx10 correct ctl stack size

nimlgen:
closes #9363

PR #9384, opened 2025-03-08
fix: correct gfx10 ctl stack size

wozeparrot:
#9365 broke amd driver on non gfx1030

it should instead be capped to `0x7000` following https://github.com/torvalds/linux/blob/21e4543a2e2f8538373d1d19264c4bae6f13e798/drivers/gpu/drm/amd/amdkfd/kfd_queue.c#L441-L447


deftdawg (approved):
Tested on:
- Steam Deck - Works!
- RX6900XT - Works!

PR #9395, opened 2025-03-10
fix torch backend memory leak

tocubed:
Realize all tensors on optimizer step.

Things like running stats from batch norm aren't captured in optimizer state or parameters. One would need to hook the model (isn't an easy way to get at it from the optimizer) and/or the specific layers. This is more foolproof, maybe sub-optimal.


geohot:
Hmm, I'm not sure you want to do this, this includes any intermediate tensors that still are alive and can make tinygrad very slow...

tocubed:
I changed it to using the named buffers on the modules. This is what batch-norm uses and should be used for any non-trainable tensors belonging to the model.

I don't like the solution fully... it's more of a stop gap for now to address mnist and nanoGPT (and possibly others). Can think of a few scenarios, training without an nn.Module or anything iterative, i.e. diffusion models or hidden state from RNNs is probably going to be problematic. 

Not obvious to me yet how to fix, besides some heuristics on the ops based on depth or buffer sizes, shapes, etc. I need to figure out what torch does when non-eager (i.e. under JIT), but the realization point there seems obvious.

geohot:
Yea, longer term this problem should be solved in tinygrad proper

PR #8846, opened 2025-02-01
remove comgr and use LLVM directly for AMD cards

b1tg:
- did basic test, `HIP=1 python test/test_ops.py` passed on tinybox red
- CI has not been fixed yet, I want to make sure that I am on the right track

PR #9403, opened 2025-03-11
refactor llvm compiler

b1tg:


tinygrad/runtime/ops_llvm.py
```diff
@@ -12,13 +12,15 @@ def expect(x, err, ret=None):
   return ret
 
 class LLVMCompiler(Compiler):
-  def __init__(self, host_arch:str):
-    for component in ['Target', 'TargetInfo', 'TargetMC', 'AsmParser', 'AsmPrinter']: getattr(llvm, f'LLVMInitialize{host_arch}{component}')()
+  def __init__(self, target_arch:str, gpu:str|None=None):
+    for component in ['Target', 'TargetInfo', 'TargetMC', 'AsmParser', 'AsmPrinter']: getattr(llvm, f'LLVMInitialize{target_arch}{component}')()
 
-    triple = {'AArch64': b'aarch64', 'X86': b'x86_64'}[host_arch] + b'-none-unknown-elf'
+    if target_arch == "AMDGPU": triple, cpu, feats = b"amdgcn-amd-amdhsa", (gpu or "gfx1100").encode(), b"+cumode"
```
geohot, line 18: So this isn't how this works. This isn't a standalone improvement, this is just a random piece of the other PR

tinygrad/runtime/ops_llvm.py
```diff
@@ -12,13 +12,15 @@ def expect(x, err, ret=None):
   return ret
 
 class LLVMCompiler(Compiler):
-  def __init__(self, host_arch:str):
-    for component in ['Target', 'TargetInfo', 'TargetMC', 'AsmParser', 'AsmPrinter']: getattr(llvm, f'LLVMInitialize{host_arch}{component}')()
+  def __init__(self, target_arch:str, gpu:str|None=None):
```
geohot, line 15: gpu makes no sense here. Think about what the abstraction should be

b1tg:
refactored LLVMCompiler, and the AMDGPULLVMCompiler will look like below:
https://github.com/tinygrad/tinygrad/blob/dddaf56b5c6cb9f3983692b71dbb650c6cd828de/tinygrad/runtime/ops_llvm.py#L61-L71

tinygrad/runtime/ops_llvm.py
```diff
@@ -12,13 +12,10 @@ def expect(x, err, ret=None):
   return ret
 
 class LLVMCompiler(Compiler):
-  def __init__(self, host_arch:str):
-    for component in ['Target', 'TargetInfo', 'TargetMC', 'AsmParser', 'AsmPrinter']: getattr(llvm, f'LLVMInitialize{host_arch}{component}')()
+  def __init__(self, target_arch:str, triple:bytes, cpu:bytes, feats:bytes, jit:bool=False, cache_key:str="llvm"):
```
geohot, line 15: Why is cache_key an arg?

b1tg: I was trying to make LLVMCompiler as a clean abstraction of the LLVM interface for more extensible(HostLLVMCompiler, AMDGPULLVMCompiler), avoiding check `if target_arch == "AMDGPU"` in LLVMCompiler

cache_key as argument is learned from CUDACompiler for set key different for host and amdgpu

b1tg:
set target_arch and jit as class attributes, leave processor and features as parameters, the optional values for features are closely related to the processor, so i choose to place the appropriate processing logic in the subclass.

use processor term because the LLVMCreateTargetMachine named third argument cpu, in amd case should pass the gpu arch to it, use processor can cover both cases. btw [LLVM AMD document](https://llvm.org/docs/AMDGPUUsage.html) use processor term too

 AMDGPULLVMCompiler will look like below:

```python
class AMDGPULLVMCompiler(LLVMCompiler):
  jit = False
  target_arch = "AMDGPU"
  def __init__(self, arch: str):
    self.arch = arch
    super().__init__(self.arch, "+cumode")
  def __reduce__(self): return (AMDGPULLVMCompiler, (self.arch,))
    try: return super().compile(src)
    except RuntimeError as e:
      if "undefined value '@llvm.amdgcn." in str(e): raise CompileError(str(e) + "AMD with LLVM backend requires LLVM >= 18") from e
      raise CompileError(e) from e
  def disassemble(self, lib:bytes): amdgpu_disassemble(lib)
```

PR #9405, opened 2025-03-11
Boolean Indexing and Tensor.nonzero()

PKD667:
Implementing boolean indexing and adding tests. (cleaner)


chenyuxyz:
we want this feature, but both implementation and tests in this pr are not very clean

PKD667:
at least it has more than two (2) half baked tests

PR #9411, opened 2025-03-11
Preserve metadata across shards

bhagwat-rahul:
when we shard, keep original metadata instead of overwriting 
added test to check for same (for both individual shard and on final op)


bhagwat-rahul:
trash

PR #9414, opened 2025-03-12
torch backend multigpu - add devices and tests

tocubed:
For #9317 add just the device support. Currently pulls available device count from GPUS environment variable.


geohot:
Nice!

PR #9417, opened 2025-03-12
MI300 mfma support

ignaciosica:
image

---

Gets 190TFLOPS with hand-coded opts

image

---

325TFLOPS with BEAM=3 

image


tinygrad/runtime/ops_python.py
```diff
@@ -191,6 +196,7 @@ class PythonRenderer(Renderer):
   def __init__(self):
     if getenv("EMULATE_METAL"): self.device, self.tensor_cores = "METAL", MetalRenderer.tensor_cores
     if getenv("EMULATE_AMD"): self.device, self.tensor_cores = "AMD", AMDRenderer.tensor_cores
+    if getenv("EMULATE_AMD_MFMA"): self.device, self.tensor_cores = "AMD", AMDRenderer.tensor_cores_mfma
```
geohot, line 199: untested

ignaciosica: Sorry, missed it

ignaciosica:
tested another mfma shape for f16/f32, it's slower for BEAM=3, so I don't think it's worth adding right now:

```python
  tensor_cores_mfma = [TensorCore(dims=(32,32,8), threads=64, elements_per_thread=(4,4,16), dtype_in=di, dtype_out=do,
    opts=("l0","l0","l0","l0","l0","u1","u1","l1","u1","u1"), swizzle=(((9,10,5,11,12,8),(4,0,1,2,3,6,7)),((0,1,2,3,4,8),(5,9,10,11,12,6,7))))
    for di,do in [(dtypes.half,dtypes.float)]]
```

PR #9428, opened 2025-03-13
Refactor ops_amd.py (MI300X prereq)

uuuvn:


tinygrad/runtime/support/am/amdev.py
```diff
@@ -1,35 +1,28 @@
...
-      mask &= ~m
-      values |= v << s
-    return mask, values
+  def read(self): return self.adev.rreg(self.addr)
+  def read_bitfields(self) -> dict[str, int]: return self.decode_bitfields(self.read())
 
-  def build(self, **kwargs) -> int: return self._parse_kwargs(**kwargs)[1]
+  # *args instead of something like value:int|None=None to avoid key collision with **kwargs if there is a bitfield named value
+  def write(self, *args, **kwargs):
+    if bool(args) == bool(kwargs) or (bool(args) and len(args) != 1): raise RuntimeError('Either value or **kwargs must be specified, but not both')
```
nimlgen, line 22: want both. sometimes no known offsets for regs, just raw values in kernels

uuuvn: i've converted all raw values from kernel in am to bitfields, eg:
image
This is also the only place i found where parts of raw value are undefined, but i've tested it and the two undefined bits that are set don't seem to do anything

uuuvn: If you think there is a significant chance of those two bits subtly breaking something i think it's much better to patch bitfield definitions to include them as something like "unknown1" and "unknown2" and set them explicitly than just ORing magic values from kernel

nimlgen:
don't need to define all regs and blow up the code base (even autogen).

Also, I am not a big fan of umr files. like there is not smu_v13_0_0.py, how often is this updated + it's just one more stop in a dep chain. But that's discussable... Yes, this allows to define ipregs right there, but what's the real benefit compared to AM style?

uuuvn:
> don't need to define all regs and blow up the code base (even autogen).

I autogened everything just because it's easier while i develop the thing and don't know exactly what i need, i'll make it only autogen things that tinygrad uses.

> Also, I am not a big fan of umr files. like there is not smu_v13_0_0.py, how often is this updated + it's just one more stop in a dep chain. But that's discussable...

????? There is no asic_reg/smu_v13_0_0.py because smu_v13_0_0.py doesn't contain any registers? There are just structs, smu commands and stuff like this, i've moved things like that into asic_defs because 1: it's much easier to think about this stuff as separate things 2: later i'm going to move PM4 and SDMA defs into asic_defs too.

> Yes, this allows to define ipregs right there, but what's the real benefit compared to AM style?

I wrote umr autogen because generating from kernel .h were taking like 15 minutes (isn't that why am autogen isn't in verify autogen ci right now?). This generates everything in half a second.

nimlgen:
umr db is prob fine if you can make it usable in one-click. git clone and regen the needed file (see nv autogen).

Makes sense re smu_v13_0_0

tinygrad/runtime/support/am/ip.py
```diff
@@ -211,17 +211,17 @@ def setup_ring(self, ring_addr:int, ring_size:int, rptr_addr:int, wptr_addr:int,
...
       cp_hqd_pipe_priority=0x2, cp_hqd_queue_priority=0xf, cp_hqd_quantum=0x111,
       cp_hqd_pq_base_lo=lo32(ring_addr>>8), cp_hqd_pq_base_hi=hi32(ring_addr>>8),
       cp_hqd_pq_rptr_report_addr_lo=lo32(rptr_addr), cp_hqd_pq_rptr_report_addr_hi=hi32(rptr_addr),
       cp_hqd_pq_wptr_poll_addr_lo=lo32(wptr_addr), cp_hqd_pq_wptr_poll_addr_hi=hi32(wptr_addr),
-      cp_hqd_pq_doorbell_control=self.adev.regCP_HQD_PQ_DOORBELL_CONTROL.build(doorbell_offset=doorbell*2, doorbell_en=1),
-      cp_hqd_pq_control=self.adev.regCP_HQD_PQ_CONTROL.build(rptr_block_size=5, unord_dispatch=0, queue_size=(ring_size//4).bit_length()-2),
-      cp_hqd_ib_control=self.adev.regCP_HQD_IB_CONTROL.build(min_ib_avail_size=0x3), cp_hqd_hq_status0=0x20004000,
-      cp_mqd_control=self.adev.regCP_MQD_CONTROL.build(priv_state=1), cp_hqd_vmid=0,
+      cp_hqd_pq_doorbell_control=self.adev.regCP_HQD_PQ_DOORBELL_CONTROL.encode_bitfields(doorbell_offset=doorbell*2, doorbell_en=1),
+      cp_hqd_pq_control=self.adev.regCP_HQD_PQ_CONTROL.encode_bitfields(rptr_block_size=5, unord_dispatch=0, queue_size=(ring_size//4).bit_length()-2), # noqa: E501
```
nimlgen, line 220: e501 + `encode_bitfields` is a long name. I am fine with `build` ro just `encode`

tinygrad/runtime/support/amd_common.py
```diff
@@ -0,0 +1,32 @@
+import functools
+from collections import defaultdict
+from dataclasses import dataclass
+from math import log2
+from tinygrad.helpers import getbits
+
+@dataclass(frozen=True)
+class IPReg:
```
nimlgen, line 8: amd prefix, `AMDReg`?

nimlgen:
split into several parts. I see at least 2: ipreg/am + ops_amd

autogen_stubs.sh
```diff
@@ -326,11 +326,46 @@ generate_am() {
...
+    extra/hip_gpu_driver/navi10_enum.h \
+    extra/hip_gpu_driver/nvd.h \
+    extra/amdpci/headers/gc_10_3_0_offset.h \
+    extra/amdpci/headers/gc_10_3_0_sh_mask.h \
+    -o $BASE/am/gc_10_3_0.py
+  fixup $BASE/am/gc_10_3_0.py
+
+  clang2py -k cdefstum \
+    extra/hip_gpu_driver/kfd_pm4_headers_ai.h \
+    extra/hip_gpu_driver/soc21_enum.h \
```
nimlgen, line 339: in some recent commits I have moved out soc to a separate autogen file. Do we want to double soc21 defs here?

autogen_stubs.sh
```diff
@@ -326,11 +326,46 @@ generate_am() {
...
+    extra/hip_gpu_driver/nvd.h \
     extra/amdpci/headers/gc_11_0_0_offset.h \
     extra/amdpci/headers/gc_11_0_0_sh_mask.h \
     -o $BASE/am/gc_11_0_0.py
   fixup $BASE/am/gc_11_0_0.py
 
+  clang2py -k cdefstum \
+    extra/hip_gpu_driver/kfd_pm4_headers_ai.h \
```
nimlgen, line 347: I would be fine with having sdma and pm4 as separate packages, leaving gc_x_x_x.py as ~regs only. Do you see any advantages to moving everything into one autogen?

autogen_stubs.sh
uuuvn: Nope, haven't seen that commit, will move to using soc files instead of doing that

autogen_stubs.sh
uuuvn: Only that it's simpler from python (no need to separately select and import pm4 and soc) at a cost of more cluttered autogen

tinygrad/runtime/ops_amd.py
```diff
@@ -415,6 +411,29 @@ def signal_doorbell(self, dev):
...
+  def addr(self): return self.ip.bases[self.segment] + self.offset
+
+@dataclass(frozen=True)
+class AMDIP:
+  name: str
+  version: tuple[int, ...]
+  bases: tuple[int, ...]
+  @functools.cached_property
+  def module(self):
+    for ver in [self.version, self.version[:2]+(0,), self.version[:1]+(0, 0)]:
```
nimlgen, line 427: we should have the same in AM. maybe move to common?

deftdawg:
Commits beginning with this one broke RDNA2 / gfx1030 / RX6000 (#9665) due to errors loading modules `SDMA 5.2.0` and then later with `NBIO 3.3.0`...  Any thoughts to how we might fix?

Issue #9108, opened 2025-02-15
Rand and JIT unpredictable behavior


tobias17:
The following was testing with both the `NV=1` and `CUDA=1` backends with no observable differences.

## Repro

The following code will repeat the values starting with iteration 2 and onward, even though the values should be random.
```python
from tinygrad import Tensor, TinyJit

@TinyJit
def f(x:Tensor) -> Tensor:
   return x.softmax().multinomial().realize()

inputs = [Tensor.rand(100) for _ in range(2)]

for i in range(4):
   print(f"Iteration {i+1}")
   for inp in inputs:
      print(f(inp.realize()).item())
   print()
```

## A hacky workaround

For some reason, changing the inputs line to the following "fixes" it
```python
inputs = [Tensor.rand(100).realize() for _ in range(2)]
```

## Issues with BEAM

Running either pieces of code with `BEAM=1` causes the program to print exceptions in an infinite loop.

## More weirdness

This issue was origienally discovered when my LLM was not producing different results when calling it multiple times with the same input (even though my temperature was 0.95). Poking around there, I was able to have another JIT that just returned a `Tensor.rand(1,1,1).realize()` produce the same outputs, but adding `Tensor.rand(1,1,1).realize()` calls outside of the JIT in specific places would cause it or not cause it to return the same value.

chenyuxyz:
this is similar to `test_jit_random_after_unrealized_random` failure. The seed increment is not captured by jit @Qazalin fyi

chenyuxyz:
seems to be fixed on master, all outputs are different for me

PR #9434, opened 2025-03-13
ensure randn seed is realized + remove contiguous from threefry [pr]

Qazalin:
This diff brings back the behavior https://github.com/tinygrad/tinygrad/pull/6869 introduced to threefry.

Threefry uses `seed.assign(seed+1)` to ensure every Tensor.rand generates a new random.

The root cause of https://github.com/tinygrad/tinygrad/issues/9108 is that if `Tensor.rand` is called once outside the JIT, the seed Tensor UOp becomes a lazy `ADD`. When this ADD enters the JIT, `seed +=1` of each JIT Item exec never gets written to memory, thus all JIT items read the same seed.

I believe the fix #6869 was not correct though - we do not want to break the graph on seed increment, rather, we want to ensure `seed.assign(seed+1)` isn't just a UOp replace, but a write to memory.

### DIFF
--- a/test/test_jit.py
+++ b/test/test_jit.py
@@ -318,7 +318,6 @@ class TestJit(unittest.TestCase):
     assert len(res3) == 10, "All values should be different, rand works in jit."
     assert res3 != res2, "Jit rand is diff with diff seeds"
 
-  @unittest.expectedFailure # requires contiguous folding
   def test_jit_random_after_unrealized_random(self):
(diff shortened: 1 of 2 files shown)


Qazalin:
err, it's still a replace in 9108, JIT exec 2 is an ASSIGN, the captured graph ends up being (ASSIGN+1) for every seed.

PR #9435, opened 2025-03-13
don't apply GROUP to const reduce

chenyuxyz:
for arange the second reduce cannot know the first reduce is a const that has a close form


chenyuxyz:
hmm what if it's fused arange and the whole thing got GROUP

PR #9436, opened 2025-03-14
symbolic fold double where

chenyuxyz:
a.where(b.where(c, d), d) -> (a & b).where(c, d). a pattern in optimizer

process replay diff looks surprisingly smart... trying to figure out why


chenyuxyz:
weird smart diffs are gone. process replay only has the intended change
image

PR #9438, opened 2025-03-14
`simple_matmul.py` uses np to generate random

ignaciosica:
As requested by George on discord

Using generator for ints but not for floats due to the following benchmarks
```
np.random.randint   : 2.6765651249997973
rng.integers        : 1.0878237079996325
np.random.rand      : 1.4534849159999794
rng.random          : 1.8805371250000462
```

### DIFF
--- a/extra/gemm/simple_matmul.py
+++ b/extra/gemm/simple_matmul.py
@@ -19,9 +19,10 @@ INT_HIGH = getenv("INT_HIGH", 10)
 
 if __name__ == "__main__":
   def init_matrix(rows, cols):
+    rng = np.random.default_rng()
     if dtype_in in dtypes.ints:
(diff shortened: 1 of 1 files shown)


geohot:
You want to pass the dtype into rng.random to make it faster, this is currently generating doubles.

ignaciosica:
updated benchmark for 1000x1000 tensor
```
np.random.randint   : 0.26341387499996927
rng.integers        : 0.10546883299957699
np.random.rand      : 0.14306495800065022
rng.random          : 0.09861004199956369
```

set float32 as default for generator:
> dtype : dtype, optional
> Desired dtype of the result, only float64 and float32 are supported. Byteorder must be native. The default value is np.float64.

Issue #3392, opened 2024-02-14
Rethink concatenated constants


geohot:
Winograd uses concatenated constants to build a Tensor that has no backing buffer. Is there a cleaner way to express this?

chenyuxyz:
interesting example
```
from tinygrad import Tensor
N = 3
c = Tensor.stack([Tensor(i+1) for i in range(N)])
x = Tensor.rand(N)
(c * x).realize()
```

if N is upcasted, it generates the clean code
```
kernel void E_3(device float* data0, const device float* data1, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  float val0 = *(data1+0);
  float val1 = *(data1+1);
  float val2 = *(data1+2);
  *(data0+0) = val0;
  *(data0+1) = (2.0f*val1);
  *(data0+2) = (3.0f*val2);
}
```

PR #9439, opened 2025-03-14
Optimize Winograd convolution with symbolic constant tensors [pr]

c5huracan:
This PR optimizes the Winograd convolution implementation by using symbolic constant tensors instead of materialized ones -- Also see _Rethink concatenated constants_ #3392

## Changes
- Modified Winograd matrices to be created as device-agnostic constant tensors
- Updated _get_winograd_matcols to use symbolic operations
- Updated _apply_winograd_matrix to handle broadcasting without forcing materialization
- Updated the conv2d implementation to use these optimized functions with the symbolic constants

## Benefits
- Reduces memory usage by not allocating memory for constant Winograd matrices
- Enables compiler optimizations by fusing constants directly into computation kernels
- Improves code maintainability through a more elegant approach to constant handling

## Performance Notes
Performance testing shows comparable execution times to the original implementation, with some cases showing slightly better consistency (lower standard deviation). While not primarily a performance optimization, the approach doesn't introduce any significant overhead.

## Notes
- Process replay tests show differences in operation scheduling as expected
- These differences represent a change in how operations are scheduled but maintain mathematical equivalence
- The full test suite shows very-similar failures on this branch as on master


geohot:
Tests failing, no benchmarks showing it's better and it's +17 lines

c5huracan:
@geohot That one might have been a bit too ambitious. Thank you for the opportunities!

PR #8385, opened 2024-12-23
RetinaNet MLPerf

francislata:
# Overview
This PR adds the MLPerf training script for the RetinaNet model.

# How to run the training script
Run the following command to install dependencies:
`pip install tqdm numpy wandb pycocotools boto3 pandas torch torchvision`

To download the `openimages` dataset, run the following:
`BASE_DIR=/raid/datasets/openimages/ python3 extra/datasets/openimages.py`
 
Once the dataset has been downloaded, use the following train command:
`WANDB=1 DEFAULT_FLOAT=HALF BEAM_PADTO=0 BEAM_UOPS_MAX=1500 BEAM_UPCAST_MAX=64 BEAM_MIN_PROGRESS=5 TRAIN_BEAM=4 IGNORE_BEAM_CACHE=1 IGNORE_JIT_FIRST_BEAM=1 MODEL=retinanet BASE_DIR=/raid/datasets/openimages GPUS=6 python3 examples/mlperf/model_train.py`

This will use a batch size of 96. Here is a [complete run](https://wandb.ai/flata/MLPerf-RetinaNet/runs/6je5509t/overview) that finished at around 18h2m with a metric of `0.3454`.

# Tasks
Since this is still in draft mode, here are some `TODO`s that I want to get done before it is ready for review:
- [x] Add validation test run in training script
- [x] Add W&B integration
- [x] Perform a full training run of model
- [x] Update newly-added typing annotations

# PR breakdown
Once all `TODO`s are done and correctness is verified, here's the breakdown of independent PRs that I can put up:
- [x] #9442
- [x] #9536
- [x] #9773
- [x] #9747 
- [x] #9822

PR #9442, opened 2025-03-14
RetinaNet dataloader

francislata:
# Overview
As part of #8385, this PR adds the implementation for the RetinaNet dataloader that will be responsible for loading both training and validation datasets.

Note that future PRs will use this as part of `model_eval.py` as well as the main training loop.

# Tasks
- [ ] Add dataloader tests
- [ ] Ensure dataloader CI test works

### DIFF
--- a/examples/mlperf/dataloader.py
+++ b/examples/mlperf/dataloader.py
@@ -348,6 +348,167 @@ def batch_load_unet3d(preprocessed_dataset_dir:Path, batch_size:int=6, val:bool=
       # happens with BENCHMARK set
       pass
 
+### RetinaNet
+
+def load_retinanet_data(base_dir:Path, val:bool, queue_in:Queue, queue_out:Queue,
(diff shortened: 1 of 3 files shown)


test/external/mlperf_retinanet/coco_utils.py
```diff
@@ -0,0 +1,89 @@
+# https://github.com/mlcommons/training/blob/637c82f9e699cd6caf108f92efb2c1d446b630e0/single_stage_detector/ssd/coco_utils.py
```
chenyuxyz, line 1: add `Copied from ...` and below

chenyuxyz (commented):
seems fine, you can update the comment in the next pr

test/external/mlperf_retinanet/coco_utils.py
francislata: addressed this in #9534.

PR #9445, opened 2025-03-14
increase `Linux(DSP)` timeout (10 -> 15min)

ignaciosica:
flaky, failed me twice today in ci [(failed task)](https://github.com/tinygrad/tinygrad/actions/runs/13863016046/job/38795785728)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -467,7 +467,7 @@ jobs:
   testdsp:
     name: Linux (DSP)
     runs-on: ubuntu-24.04
-    timeout-minutes: 10
+    timeout-minutes: 15
     steps:
     - name: Checkout Code
       uses: actions/checkout@v4


Qazalin:
it's failing because of cache misses, see "Setup" is taking 5 mins.
Maybe we should fix the cache instead.

ignaciosica:
image
Don't know if this is the right approach, setup is taking 5 min, maybe there is some problem with cache

ignaciosica:
Thanks Qazalin, I just saw it was cache misses, strange that only that task failed me, if cache is overused other tasks should failed as well.

ignaciosica:
I don't know why its generating so many caches with the same key

image

ignaciosica:
Oh, this is why [reference](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

image

Qazalin:
i think github has some docs about auto deleting caches on closed PRs.

ignaciosica:
> Oh, this is why [reference](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

Based on this they should be able to get them from master right?

ignaciosica:
> i think github has some docs about auto deleting caches on closed PRs.

I'll take a look

ignaciosica:
I'm not sure restore-keys is always safe, the only thing it changes hitting the cache even if setup.py changed, and that may cause problems (if setup doesn't change the key will be the same). 

I'm not sure why sometimes is not hitting the one in master and making one specifically for the pr.

ignaciosica:
Maybe we would need that extra workflow

ignaciosica:
cache cleanup workflow reference:
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries

PR #9458, opened 2025-03-15
Fix 'TC=3 DEBUG=2 python3 extra/gemm/simple_matmul.py' on METAL

vovw:
wip, fixing Incorrect results due to multiple upcasts


vovw:
- Threadgroup memory allocation exceeded 32kb 
- the tiling and memory access patterns are messed up giving incorrect output working on fixing this

PR #9459, opened 2025-03-15
Add olmoe model option

Muennighoff:
The team released a [newer and slightly better OLMoE version](https://allenai.org/blog/olmoe-app) in case you want that as an option; feel free to close if not!


geohot:
I think it should be fast and actually used first

PR #9460, opened 2025-03-15
generalize where alu fold

chenyuxyz:
previously we fold alu of where with same condition and branches like ((x, const), (y, const)) or ((const, x), (const, y)).

we can also fold ((x, const), (const, y)) or ((const, x), (y, const)) because alu with const does not increase total alu


chenyuxyz:
hmm (a, 1) + (1, b) -> (a+1, b+1) does increase alu though

chenyuxyz:
meh does not hit anything

PR #9461, opened 2025-03-15
move copy before expand in multi

KhanerX:
I will try to minimize unwanted changes, this should just solve for #9320, therefore allow for #8756.
I should add tests that show and test behavior better, and do something about its time complexity.


KhanerX:
Hi @Qazalin,
about #9320 —

This rewrite recursively finds any EXPAND ops before a COPY in the graph and moves the COPY earlier. To avoid too much recursion, I do two things:

1. Memoize matches by UOp key

2. Only recurse on base UOps and loop through movement ops separately

Is this on the right track?

Qazalin:
hmm I don't see why you need a global memo.
Rewrite rules should be fully local.
The right way to write this would be an incremental rewrite that pushes expands down one UOp at a time through the graph.

Ops.FUSE was recently merged to enable generic stuff like this, see if that is useful for you? https://github.com/tinygrad/tinygrad/pull/9808

KhanerX:
@Qazalin Thanks for your review.
I made the function simpler at expense of more rewrites.
I removed the global variable, although memoization can still be used to cache result of _expanded(UOp) -> bool.

I tried using Ops.FUSE to fuse the expand instead, couldn't get it working since COPYs in the middle break the schedule, perhaps I'm missing something.

Qazalin:
Cool, yea I can see why. On my list to review tomorrow morning.
Have you tested this on [FSDP](https://github.com/tinygrad/tinygrad/pull/8756) yet?

KhanerX:
The OOM issue when sharding both data and parameters is fixed.
But the first iteration takes too long when training GPT2-large, I will look at why it gets stuck. I switched to the small GPT2, the first iteration took 7 minutes but did eventually resume and next iterations were normal speed.

for GPT2-large, it scheduled 13k kernels, ran some compute on GPUS (for 8 seconds) then hanged and CPU went to 100%. I will debug further.

Qazalin:
I don't think we can upstream this diff, why does the `if REDUCE_AXIS` part exist?
This kind of reordering doesn't belong to multi_pm, Multi is not supposed to alter the Tensor graph's orderings.

We do some of these reorderings in grouper.py, I see CAST_BEFORE_VIEW already handles the EXPAND in the embedding,
but this CAST still exists:
image

PR #9462, opened 2025-03-15
Make OpenCL on Intel compile without opening device, working for A770 and Max 1100

amarmemic:
WIP:

- using intel opencl offline compiler (see. https://github.com/intel/compute-runtime/tree/master/shared/offline_compiler)

- contribution: 
  1. created python bindings for libocloc.so 
  2. implemented  IntelOfflineCompiler class which uses ocloc in order to compile without opening device via opencl api 
  3. TODO: add tests if needed ??
  4. TODO: add documentation and example usage 

IntelOfflineCompiler is currently implemented inside ops_gpu.py; is it the correct location ?


geohot:
I'll leave this open, but it's clear to me this PR wasn't carefully read 3 times

amarmemic:
> I'll leave this open, but it's clear to me this PR wasn't carefully read 3 times

What do you exactly mean ? I am trying to contribute for the first time, so it is a bit tricky to understand all requirements  and the context of the task. 

All I know is that the compilation  with ocloc works. 

My problem is to answer the open question. I would appreciate it I you could give some requirements and hints on that bounty. 


As you did not close the PR, it should not be totally the false way.

geohot:
Bounty locked, but this code needs to sparkle before we can merge it.

deftdawg:
@amarmemic, some silly questions (fellow A770 owner):
1. What is the approximate performance difference from using ocloc vs the current opencl/gpu method?
2. Does one need to do anything special to activate `ocloc`?  Do we still need to provide `INTEL=1` as required by this line (see below), or will `ocloc` work a without?

https://github.com/tinygrad/tinygrad/blob/5662944ff929c53b4f02c9223d12905705d0cbeb/tinygrad/runtime/ops_gpu.py#L112

amarmemic:
> @amarmemic, some silly questions (from a fellow A770 owner):
> 
> 1. What is the approximate performance difference from using ocloc vs the current opencl/gpu method?
> 2. Does one need to do anything special to activate `ocloc`?  Do we still need to provide `INTEL=1` as required by this line (see below), or will `ocloc` work without it?
> 

@deftdawg 

I did not any benchmarks but I think there should not be a difference in performance because both approaches use the same icd background (see second link above) 

No, you don't need to set anything, just using IntelOfflineCompiler class. But anyway, this PR will not be merged according to George from last meeting.

deftdawg:
> I did not any benchmarks but I think there should not be a difference in performance because both approaches use the same igc background (see second link above)

Ok, thanks...  I thought I had it configured, however I hadn't noticed any performance difference, so I was wondering about that.    I fell down a bit of a rabbit hole recently trying to figure out why inference perf seems slow using tinygrad via exo.

geohot:
I'm open to merging this with the new autogen stuff if you are still interested.

amarmemic:
> I'm open to merging this with the new autogen stuff if you are still interested.

Sure, but is it still relevant for tinygrad ?  I think there was a discussion  for a few months about AMD/Nvidia GPUs being more relevant. 

But, lets do it!

geohot:
It seems abstracted enough and testable in CI such that it's fine. We should refactor the interfaces this uses to be cleaner.

Can this work with the CPU Intel OpenCL implementation CI uses? If so, even better.

chenyuxyz:
stale

PR #9468, opened 2025-03-17
add masked_select to tensor.py

tjonari:


PKD667:
this is already implemented in #9405

chenyuxyz:
merging this as it's overall better than #9405

PR #9469, opened 2025-03-17
failed test case for threefry

chenyuxyz:
not sure if it's always like this, but increment before _threefry_random_bits is incorrect. the counts should start with random numbers generated so far.

use jax to generate 20 + 20 + 10 random numbers, the first 20 + 20 matches and the last 10 are different. just moving increment after _threefry_random_bits matches the number but jit test fails


geohot:
Can you leave the increment before for the JIT and subtract the value again when you compute it?

chenyuxyz:
something was still wrong. merging the test case first

PR #9483, opened 2025-03-17
Tiny backend test_ops fix part 3

Anish9901:
Related to #9302

72 --> 24 failures


extra/torch_backend/backend.py
```diff
@@ -441,6 +502,7 @@ def _wrap_out(*args, **kwargs):
   "aten.repeat": Tensor.repeat,
   "aten.lerp.Tensor": Tensor.lerp,
   "aten.expand": Tensor.expand,
+  "aten.ones_like": lambda self, **kwargs: Tensor.ones_like(self),
```
chenyuxyz, line 505: what are the `kwargs` here?

Anish9901: dtype, layout, device, pin_memory & memory_format

chenyuxyz: should pass at least dtype and device

PR #9496, opened 2025-03-18
improve reproducibility of WebGPU CI puppeteer test

hooved:
Here I attempt to show why `node test/web/test_webgpu.js` as part of the CI job `MacOS (WebGPU) / Run WEBGPU Efficientnet` can fail:

The hypothesis is that there is a race condition where sometimes model import is not complete before model is referenced, which I reproduced locally and am testing in CI.


hooved:
Here the test fails when I force delay of the model import in the WebGPU test: [test fail](https://github.com/tinygrad/tinygrad/actions/runs/13923769860/job/38962927412)

Here I make the test work even with the above artificial delay: [test success](https://github.com/tinygrad/tinygrad/actions/runs/13923940726/job/38963492983)

I removed the import delay which forced the test to fail. Merging this diff should improve reproducibility of the test.

PR #9497, opened 2025-03-18
AMDCompiler refactor (no_comgr prereq)

b1tg:

AMDGPULLVMCompiler will use this function from tinygrad/runtime/ops_llvm.py


tinygrad/helpers.py
```diff
@@ -283,6 +283,10 @@ def capstone_flatdump(lib: bytes):
     print(f"{instr.address:#08x}: {instr.mnemonic}\t{instr.op_str}")
   sys.stdout.flush()
 
+def amdgpu_disassemble(lib:bytes):
+  asm = subprocess.check_output(["/opt/rocm/llvm/bin/llvm-objdump", '-d', '-'], input=lib)
```
nimlgen, line 287: no, that's (helpers) the wrong place for that functionality

b1tg: so amdgpu_disassemble should stay in compiler_hip.py?

PR #9498, opened 2025-03-18
fix prg size calculation when there are adjacent mapped ranges(no_comgr prereq)

b1tg:

the follow mapped ranges cause `assert prg_sz > 0` fails
```txt
st             st+sz           prg_addr
0x7370607f7000 0x7370607f8000 0x7370607f7000
0x7370607f6000 0x7370607f7000 0x7370607f7000
```


b1tg:
CI should pass after re-run

PR #9500, opened 2025-03-18
Add return_indices to max_pool2d

wizardling1:
This implementation creates a tensor containing the index of each pooled element (like return_indices for _pool), and indexes into that with the argmax of the _pool output.

return_indices=True does not seem to incur any speed penalty for large tensors and with TinyJit based on the tests I've run.


wizardling1:
this is my first contribution to tinygrad, so any feedback would be appreciated :)

geohotstan:
Does this work for repeated elements?

wizardling1:
yes, it has the same behavior as pytorch for repeated elements

wizardling1:
it seems that geohotstan made an implementation #9506 which has all the same functionality but is half the length of code

PR #9503, opened 2025-03-19
add validation of the bounds of Ops.INDEX

geohot:


tinygrad/spec.py
```diff
@@ -45,6 +45,18 @@
 
 # ***** uop type spec *****
 
+def validate_index(idx:UOp, mask:UOp|None=None):
```
geohot, line 48: @chenyuxyz do we have the stuff to add support for masked index here?

PR #9506, opened 2025-03-19
Add return_indices to max_pool

geohotstan:
Here's another way of doing maxpool indices. #9500
Heard about the single kernel greatness of `Tensor.argmax` in the meeting, so I looked for myself.
That reversed arange trick is so neat! (maybe I can try to do that for sort indices too!!! Man I should've carefully read the argmax impl wayyy earlier)

This is doing maxpool return_indices by adapting the trick from argmax.


chenyuxyz:
cool thanks!

PR #9511, opened 2025-03-20
remove move_mask from the devectorizer

Qazalin:
To check if INDEX is out of bounds, we need three things:
1. buffer size
2. min/max of alu
3. mask

The devectorizer in master removes this mask before renderer. This diff refactors the renderers to keep the mask in INDEX.

PTX does not directly render INDEX, so we still move the mask to LOAD/STORE before doing the pointer arithmetic. This move is safe because in the final linearization of PTX UOps, INDEX doesn't exist.

### DIFF
--- a/tinygrad/ops.py
+++ b/tinygrad/ops.py
@@ -735,7 +735,7 @@ class UPat(MathTrait):
 
   @staticmethod
   def any(*src): return UPatAny(src=src)
-  def or_casted(self, name:str|None=None): return UPat.any(self.named(name or self.name), UPat(Ops.CAST, name=name, src=(self,)))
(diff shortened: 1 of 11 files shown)


tinygrad/ops.py
```diff
@@ -735,6 +735,7 @@ def named(self, name:str): return UPat(self.op, self.dtype, self._in_src, self.a
 
   @staticmethod
   def any(*src): return UPatAny(src=src)
+  def or_casted(self, name:str|None=None): return UPat.any(self if name is None else self.named(name), UPat(Ops.CAST, name=name, src=(self,)))
```
geohot, line 738: Nice way to do it

PR #9518, opened 2025-03-20
add Tensor.max_unpool2d

geohotstan:
also can be used in torch backend `max_pool2d_with_indices_backward`
I gotta make sure this is at least faster than the out.gradient method first 😃


geohotstan:
Darn, I think this implementation of `max_pool2d_with_indices_backward` is _slower_!!!! 😱 
uhhhh, hmmm.

PR #9522, opened 2025-03-21
torch backend: explicit view tracking for inplace ops

tocubed:
Track views as metadata on the C++ TensorImpl so as to not lose it on detach().
Failing test/test_ops.py from 24 => 13.
Part of #9317


tocubed:
No merge yet, going to explore an alternative with storing the metadata on tinygrad Tensor itself (as XLA does with it's own tensor class) avoiding the TensorImpl code.

tocubed:
Close in favor of cleaner #9557

PR #9524, opened 2025-03-21
ops_gpu.py: match getenv("INTEL") or "intel" in device_name

deftdawg:
Without this change an Intel Arc 770 will happily burn GPU cycles without output model tokens in Exo when only `GPU=1` is passed


geohot:
Tests failing. Do you have a way to CI test this? Any reason we still need the env var?

deftdawg:
The env was introduced as part of @CaltropHungerton's Intel XMX [PR](https://github.com/deftdawg/tinygrad/commit/38fb1e14a21dbf5c27bdecd4d87a46764e46a77e) 8 months ago...

There are also mentions of `EMULATE_INTEL` in the same PR.    So maybe it was a thing to make it possible to test via CI faking the hardware?  In any case, it breaks when trying to run on real hardware unless the `INTEL=1` env is set, which seems silly when the hardware can be auto-detected.

PR #9526, opened 2025-03-21
memplan into one buffer

nimlgen:


nimlgen:
@Qazalin, can you take a look please at `DEBUG=1 python3 test/test_schedule.py TestSchedule.test_empty_is_not_realized`

```
test_empty_is_not_realized (__main__.TestSchedule) ... <buf 133416946481520 real:False device:AMD size:10 dtype:dtypes.float offset:0>
buf is <buf 133416946600688 real:False device:AMD size:10 dtype:dtypes.float offset:0> <buf 133416946481520 real:False device:AMD size:10 dtype:dtypes.float offset:0> 0
 memory reduced from 0.00 MB -> 0.00 MB, 1 -> 0 bufs
<buf 133416946481520 real:False device:AMD size:10 dtype:dtypes.float offset:0>
```

a kernel with buffers: buf 133416946600688 and buf 133416946481520 (buf 133416946481520 is `a` in the test). it fails cuz it's optimized out now (since lb_refcount=0). but I don't get why `buf 133416946481520` has `lb_refcount=0`, while it's held?

PR #9537, opened 2025-03-21
raise RuntimeError in bufs_from_lin with 0 size

chenyuxyz:
not happening now so no need that hack


chenyuxyz:
hmm better clean up Kernel before things like this

PR #9541, opened 2025-03-21
Use real shift op in transcendentals

quortus:


quortus:
It works but tests are flaky (due to timeout).

chenyuxyz:
still want this? the WEBGPU dtype error seems real

quortus:
Yeah. I will take a look next week.

chenyuxyz:
stale

PR #9542, opened 2025-03-21
Fix DSP transcendentals

quortus:
Fix for bounty: "Add sqrt support to transcendentals, fixing DSP=1 python test/test_tiny.py TestTiny.test_mnist_model with test"

Fixes log2, exp2, pow and sqrt.
sin is not working right now (will be address by #9541).


geohot:
Bounty is yours, george@tinygrad.org to claim. Can pay via PayPal or USDC on ETH

PR #9543, opened 2025-03-22
amd llvm render (no_comgr prereq)

b1tg:


tinygrad/renderer/llvmir.py
```diff
@@ -48,7 +48,7 @@ def AMX(op, gpr): return f'call void asm sideeffect ".word (0x201000+($0<<5)+0$1
 unsigned_lop = { Ops.ADD: "add", Ops.MUL: "mul", Ops.IDIV: "udiv", Ops.MOD: "urem",
                  Ops.CMPLT: "icmp ult", Ops.CMPNE: "icmp ne", Ops.OR: "or", Ops.AND: "and", Ops.XOR: "xor", }
 signed_lop = {**unsigned_lop, Ops.CMPLT: "icmp slt", Ops.IDIV: "sdiv", Ops.MOD: "srem"}
-flags = " nsz arcp contract afn"
```
b1tg, line 51: remove afn to fix test/test_ops.py::TestOps::test_div_rounding_mode, which made 6/3=1 in trunc/floor mode, and it's only failed on real hw, mockgpu was correct

geohot: Any downside to this change?

b1tg: remove afn would make fdiv use Newton-Raphson method rather than simply use v_rcp_f32, slower but more precious, other ops currently using flags generate the same assembly code with or without afn base on my test.

and afn do not effect fdiv code-gen for cpu target (only tested on x86)

geohot: I think it should use rcp. I'm fine with skipping that test on AMD LLVM

PR #9545, opened 2025-03-22
[no merge] add huggingface onnx modesl to CI

geohotstan:
Test to see the speed of mac runner on main repo


geohotstan:
oh lord top 30 repo takes 10 minutes just to run. I think maybe I go and hand pick some good ones (diverse ones) to run rather than selecting via top downloaded.

PR #9549, opened 2025-03-23
ONNX OnnxRunner file path as input

geohotstan:


chenyuxyz:
what's the benefit of using path as Runner input? I would expect Runner to take a model/program as input and does not introduce other side effects

geohotstan:
> what's the benefit of using path as Runner input? I would expect Runner to take a model/program as input and does not introduce other side effects

george mentioned this in discord
image
so this PR's idea was to only allow loading of python stuff and not onnx stuff, so that maybe later we remove dependency on onnx, we can just change all the onnx stuff in here instead of dealing with changing the input types.

I also now remember us saying it's fine to have onnx as import. Maybe this is a little premature :o

chenyuxyz:
oh okay, i am not familiar with the grand plan here, but also not likely to be the most important stuff for now. we can discuss this later

PR #9550, opened 2025-03-23
HCQ: Increment timeline signal before submitting

uuuvn:
`AMDComputeQueue.__del__` frees `hw_page` which is safe because `AMDAllocator._free` does `self.dev.synchronize()` which is supposed to wait for execution of IB to finish, however that doesn't happen if AMDComputeQueue is dropped right after submit before timeline signal is incremented, which it is in most places leading to a race if .bind() is also used (required for multi-xcc because bug in mec fw treats all PACKET3_PRED_EXECs outside IBs as if they had EXEC_COUNT of zero).


nimlgen:
I am supportive of this hcq function anyway. But bind for every submit sounds crazy (for mi300). Can we just preallocate the ib?

PR #9463, opened 2025-03-15
Initial work on bounty Move _index_put_impl_, index_tensor, masked_select, and randperm_generator off of cpu in torch_backend

yvonmanzi:
-  [x] masked_select
-  [ ] randperm_generator
-  [ ] index_tensor
-  [ ] _index_put_impl
-  [x] cumprod

PR #9554, opened 2025-03-23
Torch randpermgen

yvonmanzi:
Continue moving functions off of CPU in torch backend ... #9463

### DIFF
--- a/extra/torch_backend/backend.py
+++ b/extra/torch_backend/backend.py
@@ -80,9 +80,6 @@ def index_tensor(x, y):
 def index_put(self, indices, values, accumulate=False):
   return aten.index_put(self.cpu(), [z.cpu() if isinstance(z, torch.Tensor) else None for z in indices], values.cpu(), accumulate).tiny()
 
-@torch.library.impl("aten::randperm.generator_out", "privateuseone")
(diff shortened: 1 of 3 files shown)


yvonmanzi:
While reviewing how mapping is handled in the torch backend, I noticed that functions like torch.randperm(n, generator=g) expect a seeding generator. Currently, tinygrad silently ignores the generator.
Ideally, we should:

Raise a NotImplementedError when a generator is passed (to match PyTorch’s explicit handling).
Provide guidance that users can use Tensor.manual_seed(seed) if they want seeded random generation.
One possible approach would be to modify the mapping to sth like this:


```
"aten.random_": inplace_fn("self")(lambda self, generator=None:
    (_ for _ in ()).throw(NotImplementedError("randint with a generator is not supported in tinygrad"))
    if generator is not None else
    self.assign(Tensor.randint(*self.shape, low=dtypes.min(self.dtype), high=dtypes.max(self.dtype), device=self.device, dtype=self.dtype))
)
```
In this PR, I had randperm_generator raise the error when the generator(seed) is provided and wanted to bring up this inconsistency.

We might need to update the mapping for randint() and rand() to fail explicitly when a generator is passed.

PR #9564, opened 2025-03-24
move Buffer refcount increment outside of schedule.py

Qazalin:
`lb_refcount` needs to count alive BUFFER UOps referencing the device.py Buffer.
But schedule only increments references for the new intermediate buffers it creates, or buffers it assigns to.

The new memory planner fails for Tensor.empty https://github.com/tinygrad/tinygrad/pull/9526#issuecomment-2748480115 since there's no ScheduleItem for it, might be more correct to put this when we create a Buffer.

### DIFF
--- a/test/test_gc.py
+++ b/test/test_gc.py
@@ -86,12 +86,10 @@ class TestGC(unittest.TestCase):
     a.realize()
     real_buf = a.lazydata.buffer
     # after the Tensor UOp is deleted there shouldn't be any references on the Buffer
-    with self.assertRaises(AssertionError):
-      self.assertEqual(real_buf.lb_refcount, 1)
+    self.assertEqual(real_buf.lb_refcount, 1)
     self.assertEqual(bufs_allocated()-init, 1)
     del a.lazydata
-    with self.assertRaises(AssertionError):
-      self.assertEqual(real_buf.lb_refcount, 0)
+    self.assertEqual(real_buf.lb_refcount, 0)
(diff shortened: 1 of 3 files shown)


Qazalin:
eventually we need to refactor the memory planner and JIT to not use lb_refcount,
it looks pretty old and untested.

@nimlgen if you can upstream other failing tests for this let me know.
Tomorrow morning I'll see how this works with ASSIGN.

Qazalin:
The buffers dict in ops.py that maps `{UOp(Ops.BUFFER): Buffer}`, can it be a reverse map of that?

Unless you also want ASSIGN to be counted as an extra reference,
if I ASSIGN to a buffer it has refcount of 2?

Qazalin:
TestGC. test_buffer_refcount: buffer.lb_refcount will keep track of in scope BUFFER UOps, regardless of if there's a ScheduleItem, or if the Buffer is allocated.

TestGC.test_assign_refcount: If there's an ASSIGN, we do not increment the refcount. I think this is fine for lb_refcount since it'll stay > 0 until all ASSIGN and BUFFER UOps are out of scope.

nimlgen:
> The buffers dict in ops.py maps {UOp(Ops.BUFFER): Buffer}, can it be a reverse map of that?

I need to check how you use `buffers`. `lb_refcount` should be 0 for all intermediate buffers that the user cannot access or hold a pointer to, so that the memory planner, jit, and others can optimize them out.

The only thing I care about is whether lb_refcount is 0 or not, so is your assign case more of an implementation detail, no?

PR #9584, opened 2025-03-26
LDS

ignaciosica:
This is a rough draft.

This works focuses on correctness (adding smem caching).
In order to be able to swizzle into local memory, we need the dimensions of the reduced axes to be the same as the dimensions of the local axes. This is necessary in order to be able to permute them freely. As a simplification, right now I'm only enabling LDS on powers of 2. But in the future, if you have a reduce dim of size 5 and a locals of size 5, you should be able to swizzle them.

Thinks I'm avoiding
- TC
- Padded
- OptOps with amt 0
- GROUPTOP
- Speed
- UROLL/LOCALS with size that are not power of 2

TODO
- support most of the things we are currently avoiding right now
- swizzle local memory layout

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -655,6 +655,25 @@ class Kernel:
 
     return graph_rewrite(fixup_ast(self.ast), view_left)
 
+  def lds(self, ast) -> UOp:
+    from tinygrad.ops import UPat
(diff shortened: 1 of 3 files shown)


geohot:
I don't understand why locals and reduce need to be the same size.

It should load all global data accessed by the global dispatch into local memory. Then it should read from locals. This should always work, can you show an example when it doesn't?

ignaciosica:
wrong approach

PR #9591, opened 2025-03-27
OptOps.LDS

ignaciosica:
**We want to search for:**
- Tile shape
- Tile layout
- Number of threads & what each thread moves

**OptOps.LDS specification**
- amt (buffer_id)
- arg (tile_shape (4,4), threads (min=1, max=tile size), layout ((0,1) for contiguous or (1,0) for transposed))


ignaciosica:
image

ignaciosica:
wrong approach

PR #9596, opened 2025-03-27
OptOps.LDS

ignaciosica:


ignaciosica:
actually I think it is not right

tinygrad/codegen/kernel.py
```diff
@@ -663,16 +697,17 @@ def linearize(self, name_override:Optional[str]=None, ast_transform:Optional[Cal
     if getenv("VIZ"): graph_rewrite(self.ast, PatternMatcher([]), name="View Base AST")
 
     modified_ast = self.get_optimized_ast(name_override)
+    modified_ast = self.apply_lds(modified_ast)
     if ast_transform is not None: modified_ast = ast_transform(self, modified_ast)
 
     if DEBUG >= 3:
       print(self.name)
-      if DEBUG >= 5: print(self.ast)
+      if DEBUG >= 4: print(self.ast)
```
geohot, line 705: I don't want this in 4, it can be really large. Maybe we move UOps to 6?

PR #9606, opened 2025-03-28
fix helper_tc_allclose

ignaciosica:
Padded TC where not tested because of this

``` python
def helper_tc_allclose(n:int, m:int, k:int, dtype_in:DType, dtype_out:DType, axis:int=0, tc_select:int=-1, tc_opt:int=0):
  a, b = Tensor.rand(m, k, dtype=dtype_in), Tensor.rand(k, n, dtype=dtype_in)
  np_a, np_b = a.numpy(), b.numpy()
  r = a.matmul(b, dtype=dtype_out)
  sched = r.schedule()
  realized_ast = sched[-1].ast
  run_schedule(sched)
  out = r.numpy() # output buffer is computed here
                  # opts applied to kernel after this are noop
  k = Kernel(realized_ast)
  k.apply_tensor_cores(1, axis=axis, tc_select=tc_select, tc_opt=tc_opt)
  k.linearize()
  assert len([uop for uop in k.uops if uop.op is Ops.WMMA]) > 0, "tensor core not triggered"
  assert len([x for x in k.applied_opts if x.op is OptOps.TC]) == 1, "tensor core opt not included"
  np_c = np_a @ np_b
  if dtype_in == dtypes.half: tc_atol, tc_rtol = 1e-2, 1e-3
  elif dtype_in == dtypes.bfloat16: tc_atol, tc_rtol = 1e-2, 1e-2
  else: tc_atol, tc_rtol = 5e-3, 1e-4
  np.testing.assert_allclose(np_c, out, atol=tc_atol, rtol=tc_rtol)
```


ignaciosica:
CI failing for emulated tc, will come back to this later

geohot:
Do not mix syntax changes (the caps) with anything else.

ignaciosica:
will move python fix to independent pr once I figure it out

ignaciosica:
There is a bug that I can only reproduce with AMD TC padding by 1 the N dimension. 

image

The bug appears for both real TC and emulated tc with python.
I can't trigger this bug for any other TC or padding. The strange thing that it only triggers when the pad is +1 for N dimension. If pad is anything else it works

chenyuxyz:
it might be a compiler bug https://github.com/tinygrad/tinygrad/blob/49dafe6d43ed176b2cee2f2c4a923fbe8c095ec8/.github/workflows/benchmark.yml#L379

see if it passed if you disable cache and remove -O3

ignaciosica:
Thanks, it might be that yes. I disabled compiler cache and removed -O3 and it is still present, but double checked and actually for python backend is fine, it was just the assert that was breaking, but the output looks fine. So it must be AMD related

PR #9609, opened 2025-03-28
Improve ONNX ops test coverage

geohotstan:
Adds tests for some long standing untested stuff in onnx ops.
Also some small clean ups of the ops here and there.

For the onnx tests that both we and ORT fail at, I added the ORT runnable ones in `external_test_onnx_ops` to match our behavior with ORT.


geohotstan:
Gonna break this up.

PR #9612, opened 2025-03-28
replay_pkl: add func to move buffers to dsp

nimlgen:


geohot:
How does this work, I trash the bufs on line 46

PR #9619, opened 2025-03-29
TC pattern_matcher

ignaciosica:
I'll use this as a base pr, as there are many interconnected things to address. Once I'm happy with the state of the whole refactor, I will start breaking it up into smaller standalone prs.

I started working on this as naive LDS alone was never going to make BERT fast. We need to properly search over the different access patterns for the threads into locals so they collaborate. We need to be able to properly search even if tensor cores are applied. In the way tensor cores used to apply, they hided the kernel applied opts.

`apply_tc_optimizations` are just handcoded way of adding TC. TC pattern matching works only checks for opts performed and dtype. If the opts are correct for any available TC, it will rewrite.

test_ops run on METAL
~Need to investigate why they are failing, I believe it might have something to do with `simplify_ones`. Before there was a `fix_axes` function for tensor cores~. edit: this is failing because it was choosing the wrong axis for the optimizations for convs.
image

Many things yet to address:

- [ ] Kernel API change: many fails tests due to this
- [ ] TC opts and swizzle for other backends: Only fixed for metal now
- [ ] Simplify get_optimized_ast: I think this could be refactored into much simpler pm rules
- [ ] Code style and cleanup
- [ ] Refactor TensorCore class: I think this can be greatly simplified as all you really require are the opts, swizzle and dtypes (though you may still want some arguments as thread count or dims for making easier the process of adding new ones as it may guide you, maybe not)
- [X] TC handcoded opts
- [ ] Search
- [ ] TC=2, I don't think we will need this anymore
- [ ] TC=3, I think we would want to address this after LDS, using that infrastructure to localize the TC. I think TC=3 is basically LDS with the specific thread access pattern
- [ ] TC_SELECT? I think this is no longer needed. It is basically decided by the opts like what TC shape can you plug in
- [ ] Move wmma rewrite and swizzle into different rules?
- [ ] Failing test: ~20~ ~11~ 9

---

image
image

---

edit: Actually, I don't want this rule to depend at all on the applied opts (that is kind of cheating), I only want to check if it is pluggable. I mean, if the shape was correct in the first place for the tensor core but wasn't produced by the opts, tc will not trigger this was, that's the main reason for this second approach. TC should be a completely independent rule


ignaciosica:
closing in favor of #9645

PR #9642, opened 2025-03-31
torch backend: partial fix for strided related test fails

tocubed:
Report correct strides and offset to torch. The offset is necessary as part of some of the backwards as_strided calculations that torch does (like in [CopySlices](https://github.com/pytorch/pytorch/blob/c97632154108f27a84ae843844b3f7fd15994c65/torch/csrc/autograd/functions/tensor.cpp#L199)).

Fix to repeat not being a view.
Fix to behavior of as_strided. The operation should perform on the base "storage" and doesn't stack. Added test.
Some TODOs added where we need to be respecting requested strides.

Mostly as a result of investigation into `TINY_BACKEND=1 python test/test_ops.py TestOps.test_pad_circular_mode` failure.
The test now passes, but others are still failing. Looking to see results on #9302

This also fixed test_cat but broke two einsum tests, taking a look at those next.


chenyuxyz:
is repeat not being a view an intended behavior? ideally we don't want to call contiguous if we don't need to

tocubed:
Docs mention repeat should return a copy [here.](https://pytorch.org/docs/stable/generated/torch.Tensor.repeat.html#torch.Tensor.repeat)
The contiguous may be avoidable but the seat strides issue in _copy_from needs to be fixed first.

PR #9645, opened 2025-03-31
TC `pattern_matcher` pt.1 [pr]

ignaciosica:
This pr only focuses on the rewrite into Tensor Core. In my last pr I was tackling two different problems at the same time: 1) the heuristics of modifying the kernel to use tensor cores, 2) applying tensor cores through a rewrite rule.

This new pr makes the application of tensor core only dependent on the ast and kernel_info. It no longer checks if the applied opts are the same, it only checks if the reduce and the sources are the correct shape for the tensor core to be applied.

Many of the points from #9619 are still valid and will be addressed later on. This right now need more lines but will make the heuristics for adding TC in the future much simpler. In fact, search should able to find, the problem is that it needs a very specific combination to work, we should guide it. That is where heuristics come.

Many simplifications will come after this. I think it will be easy to reuse LDS infrastructure to do TC=3 instead of it being a specific case.


ignaciosica:
Will close until I'm happier with the refactor. Maybe its not even worth this refactor to check at a shapetracker level, and it should be done at the other level George suggested (treat locals as upcast and make the gep push the access pattern (don't really understand what he meant yet, will go over the last meeting again in the future))

PR #9652, opened 2025-03-31
as_strided

utkarshgill:
~~We build a view by summing broadcasted arange vectors multiplied by the corresponding strides to get a direct index mapping~~

EDIT:

0.001 ms — torch.as_strided
0.1 ms — [STRIDE as a mop, using View.create](https://github.com/tinygrad/tinygrad/pull/9652#issuecomment-2777954959):  fastest, but gradient is [tricky](https://github.com/pytorch/pytorch/blob/ffc7158bf2f97916305217e4203ef846c00161ce/tools/autograd/templates/Functions.cpp#L937-L1488)
0.4 ms — [Direct ShapeTracker + View manipulation](https://github.com/tinygrad/tinygrad/pull/9652#issuecomment-2772944200): `.realize()` breaks graph, not ideal
5.0 ms — [gather-based](https://github.com/tinygrad/tinygrad/pull/9652#issuecomment-2797211746) : slowest, uses existing mops


utkarshgill:
We can consider this alternative too. It doesn't use realize(). It is slower than `View.create()`. Here we use tensor operations rather than direct ShapeTracker manipulation. I'm not certain this is better overall, but it might be worth considering.

```python
  def as_strided(self, size, stride, storage_offset=0) -> Tensor:
    if stride and min(stride) < 0: raise RuntimeError(f"as_strided: negative strides not supported, got {stride}")
    total = prod(size)
    if total == 0: return Tensor.zeros(size)
    flat = self.contiguous().flatten()
    if total == 1: return flat[storage_offset].reshape(size)
    max_idx = storage_offset + sum((s - 1) * st for s, st in zip(size, stride))
    if max_idx >= flat.shape[0]: return flat[storage_offset:].reshape(size)
    if len(size) == 1: return flat[Tensor.arange(size[0], dtype=dtypes.int32) * stride[0] + storage_offset].reshape(size)
    
    idx_strides = [1] * len(size)
    for i in range(len(size) - 2, -1, -1): idx_strides[i] = idx_strides[i + 1] * size[i + 1]
    
    indices = storage_offset
    for i in range(len(size)): indices += (Tensor.arange(total, dtype=dtypes.int32) // idx_strides[i] % size[i]) * stride[i]
    return flat[indices].reshape(size)
```
```bash
Size: (50, 50), Stride: (101, 1), Offset: 1
Tinygrad: 1.222 ms
PyTorch:  0.001 ms
Ratio:    1063.47x

Size: (90, 90), Stride: (101, 1), Offset: 1
Tinygrad: 1.226 ms
PyTorch:  0.001 ms
Ratio:    827.93x

Size: (10, 10, 10), Stride: (100, 10, 1), Offset: 1
Tinygrad: 1.663 ms
PyTorch:  0.002 ms
Ratio:    1088.32x

Size: (50,), Stride: (2,), Offset: 1
Tinygrad: 0.587 ms
PyTorch:  0.001 ms
Ratio:    598.55x
```

utkarshgill:
This is the fastest and seems most correct way. Define AS_STRIDED as a movement op. Still learning about codebase, so I could be wrong about abstraction levels. Open to feedback

```python
# tensor.py
def as_strided(self, size, stride, storage_offset=0) -> Tensor:
  """Creates a view with custom stride/offset"""
  if any(s < 0 for s in stride): raise RuntimeError(f"as_strided: negative strides not supported, got {stride}")
  return self._apply_uop(UOp.as_strided, arg=(size, stride, storage_offset))

```
```python

# shapetracker.py
def as_strided(self, arg: tuple[tuple[sint, ...], tuple[sint, ...], sint]) -> ShapeTracker:
  size, stride, storage_offset = arg
  return ShapeTracker(self.views[0:-1] + (self.views[-1].as_strided(size, stride, storage_offset), ))
``` 
```python
# view.py
def as_strided(self, size: tuple[sint, ...], stride: tuple[sint, ...], storage_offset: sint = 0) -> View:
  if len(size) != len(stride): raise ValueError(f"size {size} and stride {stride} must have the same length")
  if not all(s >= 0 for s in size): raise ValueError(f"size can't contain negative numbers {size}")
  if all_int(stride) and any(s < 0 for s in stride): raise RuntimeError(f"as_strided: negative strides not supported, got {stride}")
  if 0 in size: return View.create(size, (0,) * len(size), 0, None)
  stride = canonicalize_strides(size, stride)
  return View.create(size, stride, self.offset + storage_offset, None)
```
TODO: Add gradient func

utkarshgill:
**`gather`-based approach:** 

-    Handles backward pass implicitly 
-    **Downside:** Very slow (~5.0ms vs ~0.001ms Torch) due to explicit index generation and `gather`.
-    **Issue:** Fails PyTorch gradient check (`TypeError: arange_start() got an unexpected keyword 'layout'`)
```
test     shape    size     stride       offset torch  tiny   ratio
(64, 64) (64, 64) (64, 1)      0      0.0010ms 6.2722ms 5979.0x ✓
(64, 64) (64, 64) (1, 64)      0      0.0010ms 6.2000ms 6389.3x ✓
(64, 64) (64,)    (65,)        0      0.0009ms 2.4910ms 2645.1x ✓
(64, 64) (4096,)  (1,)         0      0.0009ms 6.2263ms 6765.5x ✓
(32, 32) (30, 30, 3, 3) (32, 1, 32, 1) 0      0.0010ms 10.5239ms 10313.2x ✓
```

geohot:
We definitely don't want to add a new ShapeTracker method! The gather approach is more interesting

PR #9654, opened 2025-03-31
hotfix: do not fetch cifar with NULL=1

Qazalin:
needs a generic lazy HTTPTensor really,
is the shape and dtype specified from the kernel graph?


geohot:
I don't think we need this. CIFAR is fast to download, and for imagenet there's MNISTMOCK

PR #9655, opened 2025-03-31
add: single line function to copy rendered code in viz

hurrrsh:
Clicking on rendered code in VIZ copies it to clipboard

https://github.com/user-attachments/assets/e5f63448-01bc-418b-b4b9-d4ae2884e2ea


geohot:
Hidden feature. Use mouse to select code, press ctrl c.

PR #9659, opened 2025-04-01
move uops print to debug >= 6

ignaciosica:
I think it is still convenient to have ast printed independently. `VIZ` is great for many things, but for very simple stuff I think the overhead is not worth launching it, that is why I think `DEBUG=5` printing only the `ast` / `modified_ast` is useful.


Qazalin:
wanna check if print assembly on the Compiler needs bumping too?

PR #9662, opened 2025-04-01
bump assembly debug >= 7

ignaciosica:
DEBUG >= 6 will output linearized uops


Qazalin:
I think it might be better to put this in the same diff that bumps the UOps one.
Merging prereqs is fine but just changing numbers makes diffs hard to follow.

ignaciosica:
For the future cases right? Or you meant revert?

Qazalin:
We only revert if the diff is rejected.
For the future try merging diffs that add value as a standalone piece.

PR #9464, opened 2025-03-16
Bugfix for `TC=3`

ignaciosica:
```c
*** METAL      1 copy   67.11M,   METAL <- NPY             arg  2 mem  0.13 GB tm   4663.58us/     4.66ms (     0.00 GFLOPS   14.4|14.4    GB/s) 
*** METAL      2 copy   67.11M,   METAL <- NPY             arg  2 mem  0.20 GB tm   7503.00us/    12.17ms (     0.00 GFLOPS    8.9|8.9     GB/s) 
TENSOR CORES [(1, 4096, 1)] [(0, 4096, 4096)] WMMA_8_8_8_float_float
r_128_32_2_2_2_2_2_4_512_2_2_2_2_4_4
 0: (128, 32, 2, 2, 2, 2, 2, 4, 1, 1, 1, 1, 2, 4, 4) float.ptr(16777216)  (131072, 128, 2, 4096, 8192, 4, 16384, 32, 0, 0, 0, 0, 1, 32768, 8)
 1: (128, 32, 2, 2, 2, 2, 2, 4, 512, 2, 2, 2, 2, 4, 4) float.ptr(16777216)  (0, 128, 2, 0, 0, 4, 0, 32, 32768, 4096, 8192, 16384, 1, 0, 8)
 2: (128, 32, 2, 2, 2, 2, 2, 4, 512, 2, 2, 2, 2, 4, 4) float.ptr(16777216)  (131072, 0, 0, 4096, 8192, 0, 16384, 0, 8, 1, 2, 4, 0, 32768, 0)
[Opt(op=OptOps.TC, axis=0, arg=(-1, 0)), Opt(op=OptOps.UPCAST, axis=0, arg=4), Opt(op=OptOps.UPCAST, axis=1, arg=4), Opt(op=OptOps.LOCAL, axis=1, arg=4)]
#include <metal_stdlib>
using namespace metal;
kernel void r_128_32_2_2_2_2_2_4_512_2_2_2_2_4_4(device float* data0, device float* data1, device float* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  threadgroup float temp0[256];
  threadgroup float temp1[1024];
  int gidx0 = gid.x; /* 32 */
  int gidx1 = gid.y; /* 128 */
  int lidx0 = lid.x; /* 16 */
  int lidx1 = lid.y; /* 2 */
  int lidx2 = lid.z; /* 4 */
[...]

PR #9667, opened 2025-04-01
Fix 'TC=3 DEBUG=2 python3 extra/gemm/simple_matmul.py' on METAL

7vansh7:
I fixed a simple import issue but the script is still not running, working on it

### DIFF
--- a/extra/gemm/simple_matmul.py
+++ b/extra/gemm/simple_matmul.py
@@ -1,6 +1,6 @@
 import numpy as np
 from tinygrad.helpers import getenv
-from tinygrad.dtype import _to_np_dtype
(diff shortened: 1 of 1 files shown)


7vansh7:
The file runs with no output and no errors when running with python3.12 but encounters many errors when run with using python3.10

ignaciosica:
already fixed on master, see #9464

7vansh7:
I think you should update it here https://docs.google.com/spreadsheets/d/1WKHbT-7KOgjEawq5h5Ic1qUWzpfAzuD_J06N1JwOCGs/edit?gid=0#gid=0

PR #9669, opened 2025-04-01
LDS noop and spec

ignaciosica:


geohot:
So I'm generally opposed to merging things like this, it should wait in a PR until it add some functionality. The issue with spec PRs is it makes it harder to change the spec, adds lines, and doesn't add any functionality.

ignaciosica:
Okay, sorry about that

ignaciosica:
Actually, these specs are wrong, working on correcting them

PR #9674, opened 2025-04-01
don't div_and_mod_folding for negative numerator with remainder

chenyuxyz:
can be wrong in C div since it truncates towards zero


chenyuxyz:
@geohot fyi, with this we can remove the `x.vmin < -10000000: return None` hack in dsp_search and got > 50% acc

PR #9675, opened 2025-04-01
fix: FUSE_ARANGE shrink failed case

### LINKED ISSUE #9509: fuse arange shrink failed case
```
from tinygrad import Tensor
t = Tensor.arange(10)
print((t[:5] + t[5:]).tolist())
```

with `FUSE_ARANGE=1` failed with reshape

```
  File "/Users/chenyu/code/tinygrad/tinygrad/shape/view.py", line 280, in reshape
    if resolve(prod(self.shape) != prod(new_shape), False): raise ValueError(f"size mismatched, can't reshape {self.shape=} -> {new_shape=}")
                                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: size mismatched, can't reshape self.shape=(10, 1) -> new_shape=(5,)
```

CodeMan62:
## Description
This PR fixes #9509 

### Changes
- Added check to detect operations involving sliced tensors
- Prevents FUSE_ARANGE optimization when slices are detected
- Maintains optimization benefits for non-sliced cases

### Test Case
The fix is verified by `test_arange_slices_with_fuse` in test/test_arange.py which tests:
- Creating an arange tensor
- Slicing it into two parts
- Adding the sliced parts
- Verifying correct shape and values

### Implementation Details
- Added check in tensor.__getitem__ to detect when FUSE_ARANGE is enabled and slicing is involved
- Temporarily disables FUSE_ARANGE for slicing operations using context manager
- Allows operation to proceed through general path that correctly handles shapes

### DIFF
--- a/test/test_arange.py
+++ b/test/test_arange.py
@@ -182,5 +182,13 @@ class TestIndexing(unittest.TestCase):
   # at least the arange is being fused
   def test_llama_embedding_opt(self): self.test_llama_embedding(0, 1_736_704_000 if CI else 5_898_240_000)
 
+  def test_arange_slices_with_fuse(self):
+    # Test the specific case from issue #9509
+    with Context(FUSE_ARANGE=1):
+      t = Tensor.arange(10)
+      slices_sum = t[:5] + t[5:]
+      assert slices_sum.shape == (5,)
+      np.testing.assert_allclose(slices_sum.numpy(), np.array([5, 7, 9, 11, 13]))
+
 if __name__ == "__main__":
   unittest.main()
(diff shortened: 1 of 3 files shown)


Qazalin:
Looks complex and it's mutating UOps.
UOps are immutable. This should be fixed with rewrite rules instead.

CodeMan62:
@Qazalin  can you guide me what should i do ??

PR #9678, opened 2025-04-02
improve vmin/vmax for IDIV

chenyuxyz:


chenyuxyz:
https://github.com/tinygrad/tinygrad/actions/runs/14210002882/job/39816800207?pr=9678#step:29:791

removed some div/mod with this
image

PR #9679, opened 2025-04-02
add RECORD_TRACEBACKS=1 option to process replay

Qazalin:
Capturing with RECORD_TRACEBACKS=1 will also include the location of the kernel if process replay fails to generate the diff:

image
`traceback.extract_stack` is very slow though. If we wanna make this the default process_replay path maybe it should manually recurse sys._getframe.


chenyuxyz:
yea looks fine. tested in https://github.com/tinygrad/tinygrad/actions/runs/14210369386/job/39816329352?pr=9678#step:29:2276

PR #9688, opened 2025-04-02
LDS spec refactor

ignaciosica:
No functionality change, only refactor for cleaner code.
Will fix the spec on an independent pr next


geohot:
Can you revert the spec if it adds no functionality?

PR #9698, opened 2025-04-02
gep should look at count, not vcount

geohot:


VickySekhon:
This seems intuitive @geohot. Great work!

Issue #9665, opened 2025-04-01
Failed to load autogen module for NBIO 3.3.0 (breaks RDNA2)


deftdawg:
RDNA2 support was working from ~ March 5th to March 25th; commit ccbcdca473df140cf63aadaf9aadd098d5450d7d is good - after this commit it re-broke at some point with the following error:
```
  File "/nix/store/vcb6iqhcb4sqd2q7x5drs8f64yykbrmp-python3.12-tinygrad-0.10.2+git-2025-04-01/lib/python3.12/site-packages/tinygrad/runtime/ops_amd.py", line 505, in module
    def module(self): return import_module(self.name, self.version)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/vcb6iqhcb4sqd2q7x5drs8f64yykbrmp-python3.12-tinygrad-0.10.2+git-2025-04-01/lib/python3.12/site-packages/tinygrad/runtime/support/amd.py", line 32, in import_module
    raise ImportError(f"Failed to load autogen module for {name.upper()} {'.'.join(map(str, version))}")
ImportError: Failed to load autogen module for NBIO 3.3.0
```

If the source of this error isn't obvious to a dev, I can spend some to bi-sect it the specific commit it was introduced in.

It would be cool if the next release of tinygrad can include working RDNA2 support.

deftdawg:
Last good commit was 3e1168ff5e33fbcbdd5e0dd14dba4ddd3d04a52e (`am: module import in common (#9615)`), commits after that fail with either ` Failed to load autogen module for SDMA 5.2.0` or `Failed to load autogen module for NBIO 3.3.0`.

uuuvn:
> Last good commit was [3e1168f](https://github.com/tinygrad/tinygrad/commit/3e1168ff5e33fbcbdd5e0dd14dba4ddd3d04a52e) (`am: module import in common (#9615)`), commits after that fail with either ` Failed to load autogen module for SDMA 5.2.0` or `Failed to load autogen module for NBIO 3.3.0`.

Can you check my branch? Should be fine there
```bash
git remote add uuuvn https://github.com/uuuvn/tinygrad.git
git fetch --all
git switch rdna2
```
Will mark #9700 ready for review after confirmed fixed

deftdawg:
@uuuvn thanks, I grabbed your branch, there's one unit test that fails due to a missing `z3` import:

```
[...]

PR #9700, opened 2025-04-02
Fix RDNA2 support

uuuvn:
linux amdgpu_discovery.c:amdgpu_discovery_set_ip_blocks is a ton of switch cases with sometimes weird choices like replacing nbio 3.X with 2.3 while nbio 2.5 is somehow nbio 7.0. `import_module` currently just tries to replace revision and minor with zeroes if there is no exact match, but that's not enough to cover all that weirdness

### DIFF
--- a/tinygrad/runtime/ops_amd.py
+++ b/tinygrad/runtime/ops_amd.py
@@ -121,7 +121,7 @@ class AMDComputeQueue(HWQueue):
     return self
 
   def memory_barrier(self):
-    pf = 0 if self.nbio.version[:2] != (7, 11) else 1
+    pf = '' if self.nbio.version[0] == 2 else '0' if self.nbio.version[:2] != (7, 11) else '1'
     self.wait_reg_mem(reg_req=getattr(self.nbio, f'regBIF_BX_PF{pf}_GPU_HDP_FLUSH_REQ').addr,
                       reg_done=getattr(self.nbio, f'regBIF_BX_PF{pf}_GPU_HDP_FLUSH_DONE').addr, value=0xffffffff)
     self.acquire_mem()
@@ -252,7 +252,7 @@ class AMDComputeQueue(HWQueue):
     self.wreg(self.gc.regCOMPUTE_RESTART_X, 0, 0, 0)
     self.wreg(self.gc.regCOMPUTE_STATIC_THREAD_MGMT_SE0, 0xFFFFFFFF, 0xFFFFFFFF)
     self.wreg(self.gc.regCOMPUTE_STATIC_THREAD_MGMT_SE2, 0xFFFFFFFF, 0xFFFFFFFF)
-    if prg.dev.target >= (10,0,0):
+    if prg.dev.target >= (11,0,0):
       self.wreg(self.gc.regCOMPUTE_STATIC_THREAD_MGMT_SE4, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF)
     self.wreg(self.gc.regCOMPUTE_USER_DATA_0, *user_regs)
 
@@ -503,6 +503,8 @@ class AMDIP:
   def regs(self): return collect_registers(self.module, cls=functools.partial(AMDReg, ip=self))
   def __getattr__(self, name:str):
     if name in self.regs: return self.regs[name]
+    # NOTE: gfx10 gc registers always start with mm, no reg prefix
+    if (mmname:=name.replace('reg', 'mm')) in self.regs: return self.regs[mmname]
     return getattr(self.module, name)
 
 class KFDIface:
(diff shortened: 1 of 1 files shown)


uuuvn:
Should close #9665, waiting to confirm that it indeed fixes it and will mark ready for review

deftdawg:
I disabled test `test_valid_becomes_const1_z3` which was failing;  at runtime I now the following error when attempting to inference with exo:
`module 'tinygrad.runtime.autogen.am.nbio_2_3_0' has no attribute 'regBIF_BX_PF0_GPU_HDP_FLUSH_REQ'`

uuuvn:
> I disabled test `test_valid_becomes_const1_z3` which was failing; at runtime I now the following error when attempting to inference with exo: `module 'tinygrad.runtime.autogen.am.nbio_2_3_0' has no attribute 'regBIF_BX_PF0_GPU_HDP_FLUSH_REQ'`

Can you retest? Also in the issue you said one test was failing after the initial fix, it shouldn't have worked at all with that....

deftdawg:
> Can you retest? Also in the issue you said one test was failing after the initial fix, it shouldn't have worked at all with that.... Were you only running unit tests?

I'm using a modified version of the tinygrad derivation to build which [runs/excludes some of the unit tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/tinygrad/default.nix#L167), the runtime error I mentioned in the last post was from actually running exo (which uses tinygrad under the hood for inference on Linux).

I've just rerun with your latest (9bdb92eba9525f561c38ef35c0efe1d040d890ae), it now fails with this error `module 'tinygrad.runtime.autogen.am.gc_10_3_0' has no attribute 'regCOMPUTE_PGM_LO'`

image

uuuvn:
> > Can you retest? Also in the issue you said one test was failing after the initial fix, it shouldn't have worked at all with that.... Were you only running unit tests?
> 
> I'm using a modified version of the tinygrad derivation to build which [runs/excludes some of the unit tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/tinygrad/default.nix#L167), the runtime error I mentioned in the last post was from actually running exo (which uses tinygrad under the hood for inference on Linux).
> 
> I've just rerun with your latest ([9bdb92e](https://github.com/tinygrad/tinygrad/commit/9bdb92eba9525f561c38ef35c0efe1d040d890ae)), it now fails with this error `module 'tinygrad.runtime.autogen.am.gc_10_3_0' has no attribute 'regCOMPUTE_PGM_LO'`
> 
> image

Retest again? Looks like gfx10 has all the registers starting with 'mm', not 'reg'

deftdawg:
4b77923e40c43c60306fbf535199d7bb04914f4f -> `Failed to fetch completions: Error processing prompt (see logs with DEBUG>=2): module 'tinygrad.runtime.autogen.am.gc_10_3_0' has no attribute 'regCOMPUTE_STATIC_THREAD_MGMT_SE4'`

470c1c2c8a3dedb7ae8f22a2d0f471b400f34fc1 ... Testing now will report back in a couple minutes

uuuvn:
> [4b77923](https://github.com/tinygrad/tinygrad/commit/4b77923e40c43c60306fbf535199d7bb04914f4f) -> `Failed to fetch completions: Error processing prompt (see logs with DEBUG>=2): module 'tinygrad.runtime.autogen.am.gc_10_3_0' has no attribute 'regCOMPUTE_STATIC_THREAD_MGMT_SE4'`
> 
> [470c1c2](https://github.com/tinygrad/tinygrad/commit/470c1c2c8a3dedb7ae8f22a2d0f471b400f34fc1) ... Testing now will report back in a couple minutes

https://github.com/tinygrad/tinygrad/commit/470c1c2c8a3dedb7ae8f22a2d0f471b400f34fc1 won't fix it, https://github.com/tinygrad/tinygrad/commit/198c7eaf33b85e8e2da11902245eba2daef36abc hopefully will though

deftdawg:
Fixed!  Ship it! :smile: :beers: 

image

tinygrad/runtime/ops_amd.py
```diff
@@ -121,7 +121,7 @@ def xcc_barrier(self):
     return self
 
   def memory_barrier(self):
-    pf = 0 if self.nbio.version[:2] != (7, 11) else 1
+    pf = '' if self.nbio.version[0] == 2 else 0 if self.nbio.version[:2] != (7, 11) else 1
```
geohot, line 124: what's the type of this?

uuuvn: `str|int`, changed to be just `str`

tinygrad/runtime/ops_amd.py
```diff
@@ -807,8 +809,13 @@ def __init__(self, device:str=""):
     self.pm4 = importlib.import_module(f"tinygrad.runtime.autogen.am.pm4_{'nv' if self.gfxver >= 10 else 'soc15'}")
     self.sdma = import_module('sdma', min(self.dev_iface.ip_versions[am.SDMA0_HWIP], (6, 0, 0)))
     self.gc = AMDIP('gc', self.dev_iface.ip_versions[am.GC_HWIP], self.dev_iface.ip_offsets[am.GC_HWIP])
+    # Looks like on GFX9 nbio ip bases are off by one?
     pad = (0,) if self.gfxver == 9 else () # ?!?!?!?!??!?!?!
-    self.nbio = AMDIP('nbio' if self.gfxver < 12 else 'nbif', self.dev_iface.ip_versions[am.NBIF_HWIP], pad+self.dev_iface.ip_offsets[am.NBIF_HWIP])
+    # linux amdgpu_discovery.c:amdgpu_discovery_set_ip_blocks is a ton of switch cases with sometimes weird choices like replacing nbio 3.X with 2.3
+    # while nbio 2.5 is somehow nbio 7.0. import_module currently just tries to replace revision and minor with zeroes if there is no exact match, but
+    # that's not enough to cover all that weirdness
+    real_nbio_ver = self.dev_iface.ip_versions[am.NBIF_HWIP] if self.dev_iface.ip_versions[am.NBIF_HWIP][0] != 3 else (2, 3, 0)
```
nimlgen, line 817: defaulting to 2.3.0 is strange. if there is something missing I want a runtime error, not silently writing to the wrong reg.
map is fine to override what we want. runtime should support gfx>=10

uuuvn: Hm? I only set to 2.3.0 if major is 3, not defaulting to it if the correct definition can't be found, the kernel uses 2.3.0 for all nbios with major 3: https://github.com/torvalds/linux/blob/80a0e828293389358f7db56adcdcb22b28df5e11/drivers/gpu/drm/amd/amdgpu/amdgpu_discovery.c#L2874

nimlgen: I have no idea how they enumerate them and if there is any guarantee that the next version with major 3 will point to 2.3.0 as well?

Not sure if we want to autogen... Or just use a map...

Also comment above is too verbose.

tinygrad/runtime/ops_amd.py
```diff
@@ -252,7 +252,7 @@ def exec(self, prg:AMDProgram, args_state:CLikeArgsState, global_size:tuple[sint
     self.wreg(self.gc.regCOMPUTE_RESTART_X, 0, 0, 0)
     self.wreg(self.gc.regCOMPUTE_STATIC_THREAD_MGMT_SE0, 0xFFFFFFFF, 0xFFFFFFFF)
     self.wreg(self.gc.regCOMPUTE_STATIC_THREAD_MGMT_SE2, 0xFFFFFFFF, 0xFFFFFFFF)
-    if prg.dev.target >= 100000:
+    if prg.dev.target >= 110000:
```
nimlgen, line 255: why is this changing?

uuuvn: Because there is no regCOMPUTE_STATIC_THREAD_MGMT_SE4 in gfx10

uuuvn: image

PR #9703, opened 2025-04-02
fix gated store with index in python backend 

ignaciosica:
There was a bug for python backend and gated store, see [CI run](https://github.com/tinygrad/tinygrad/actions/runs/14218735982/job/39841347570#step:5:89) for reference. 

The problem was that the default gate was added where it was not expected. The index zipped the gate with the memory and offset, but the STORE was expecting for the gate to be in the input after the val. I think it's related to the move from the old style load/store to the new one.

### DIFF
--- a/test/test_renderer_failures.py
+++ b/test/test_renderer_failures.py
@@ -9,6 +9,7 @@ from tinygrad.engine.realize import CompiledRunner
 from tinygrad.helpers import dedup, flatten, prod
 from tinygrad.renderer.cstyle import CStyleLanguage
 from tinygrad.renderer.ptx import PTXRenderer
+from tinygrad.runtime.ops_python import PythonRenderer
 from tinygrad.ops import UOp, Ops
 from tinygrad.renderer import ProgramSpec
 from tinygrad.tensor import Tensor, _to_np_dtype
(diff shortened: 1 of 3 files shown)


geohot:
Are there no tests for this?

ignaciosica:
I was surprised it didn't pop up before. I discovered it through fixing the tc that actually triggered the padded tensor core. I'll add some specific tests for this tomorrow.

ignaciosica:
image

This renderer ptx test fails on master but it was skipped because it skips unless the renderer is ptx

```python
class TestPTXFailures(unittest.TestCase):
  def test_gated_store_with_alu(self):
    a = UOp(Ops.DEFINE_GLOBAL, dtypes.int.ptr(), (), 0)
    gate_alu = (lidx0:=UOp(Ops.SPECIAL, dtypes.int, (), ('lidx0', 4))).ne(0)
    gated_alu_store = UOp(Ops.STORE, dtypes.void, (a.index(lidx0, gate_alu), UOp.const(dtypes.int, 1)))
    sink = UOp(Ops.SINK, dtypes.void, (gated_alu_store,))
    uops = linearize_uop(full_graph_rewrite(sink, Device[Device.DEFAULT].renderer))
    ret = _test_uop_result([], uops, local_size=[4, 1, 1])[0]
    np.testing.assert_equal(ret, [0, 1, 1, 1])
```

test/test_renderer_failures.py
```diff
@@ -27,6 +28,17 @@ def _recursive_add(uop:UOp) -> List[UOp]: return flatten([_recursive_add(x) for
   ei.exec(outbufs+inbufs)
   return [np.frombuffer(x.as_buffer(), _to_np_dtype(x.dtype)) for x in outbufs]
 
+class TestRendererFailures(unittest.TestCase):
```
geohot, line 31: Is this a renderer failure?

ignaciosica: It works for METAL=1

ignaciosica:
Working on adding gate for if for python, right now in master if is a noop for python

image

## test

```
  def test_gated_store_with_if(self):
    a = UOp(Ops.DEFINE_GLOBAL, dtypes.int.ptr(), (), 0)
    gate_alu = (lidx0:=UOp(Ops.SPECIAL, dtypes.int, (), ('lidx0', 4))).ne(0)
    val = UOp.const(dtypes.int, 1)
    if_uop = UOp(Ops.IF, dtypes.void, (gate_alu,))
    gated_alu_store = UOp(Ops.STORE, dtypes.void, (a.index(lidx0), val, if_uop))
    sink = UOp(Ops.SINK, dtypes.void, (gated_alu_store,))
    uops = linearize_uop(full_graph_rewrite(sink, Device[Device.DEFAULT].renderer))
    ret = _test_uop_result([], uops, local_size=[4, 1, 1])[0]
    np.testing.assert_equal(ret, [0, 1, 1, 1])
```

Qazalin:
Can this be tested with something that uses Tensor methods instead of constructing UOps directly?

PR #9585, opened 2025-03-26
MI300X support

uuuvn:

PR #9705, opened 2025-04-02
Fix double SDMA_OP_FENCE

uuuvn:
Introduced in #9585, probably when i incorrectly resolved merge conflict while rebasing an old, mi300x-only branch. Seems to be the source of multi gpu beam llama hangs


uuuvn:
I'm not sure how writing the same value twice vs one time leads to hangs but so far i was able to reproduce the hangs quite quickly without that and not at all with this fix

geohot:
Merged, as it's clearly a bug. Can you think about what sort of tests would have caught this and add them?

nimlgen:
> I'm not sure how writing the same value twice vs one time leads to hangs

```
dev1                 dev2
write 1 to x   
                      wait for 1 in x  
                      write 2 to x  
write 1 to x
```

@uuuvn, I strongly encourage smaller PRs. The MI300 PR introduced three regressions: dropped RDNA4 support, this bug, and CWSR is now allocated for AM (it's not really needed there and it's not used)

PR #9708, opened 2025-04-03
Fix TestLinearizerFailures.test_failure_53 (bounty)

bokonon1:
```
$ PYTHONPATH=. python3 test/test_linearizer_failures.py TestLinearizerFailures.test_failure_53
.
Ran 1 test in 0.079s

OK
```


bokonon1:
@geohot it seems like the original linearizer bug that was causing infinite recursion is fixed, but there is an additional/separate issue that happens later when the GPU (OpenCL) runtime is used

not sure if that's intended to be in scope, but I can dig more if so

when you say unit test, do you mean a test that more specifically/narrowly triggers the original linearizer bug? or one that would catch the remaining rendering issue in the GPU case?

austin362667 (commented):
This fix looks correct! Thank you~

geohot:
Can you rebase this on the new linearizer?

chenyuxyz:
stale

PR #9709, opened 2025-04-03
Revert "disable CI red llama 3 4 gpu beam"

chenyuxyz:
Reverts tinygrad/tinygrad#9690


chenyuxyz:
ran 4 times all seemed fine

PR #9471, opened 2025-03-17
ONNX true float16

geohotstan:

PR #9711, opened 2025-04-03
fix ONNX IsInf unintended dtype promotion

geohotstan:
Attributes `detect_negative` and `detect_positive` are 1 and 0 only.
Separated from #9471

### DIFF
--- a/test/external/external_test_onnx_ops.py
+++ b/test/external/external_test_onnx_ops.py
@@ -67,7 +67,7 @@ class TestMainOnnxOps(TestOnnxOps):
   def test_isinf(self):
     # https://github.com/onnx/onnx/blob/main/docs/Operators.md#isinf
     # attributes are int but output expects bool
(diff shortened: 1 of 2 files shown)


test/external/external_test_onnx_ops.py
```diff
@@ -58,6 +64,17 @@ def test_maxunpool(self):
...
+  def test_isinf(self):
+    # https://github.com/onnx/onnx/blob/main/docs/Operators.md#isinf
+    # attributes are int but output expects bool
+    x = np.array([-1.2, np.nan, np.inf, 2.8, -np.inf, np.inf], dtype=np.float32)
+    inputs = {"x": x}
+    attributes = {"detect_negative":1, "detect_positive":1}
+    outputs = ["y"]
+    model = self.helper_build_model("IsInf", inputs, attributes, outputs)
+    outputs = OnnxRunner(model)(inputs)
+    assert outputs["y"].dtype is dtypes.bool
```
chenyuxyz, line 76: fine for this one. can think about how to make `helper_test_single_op` always check dtype

PR #9716, opened 2025-04-03
update mi300x bert haparams

chenyuxyz:
borrowed from previous submission that also did BS=1024


chenyuxyz:
consistent < 2hr on mi300x machine https://wandb.ai/chenyuxyz/MLPerf-BERT/runs/bkx21h4c?nw=nwuserchenyuxyz

PR #9720, opened 2025-04-03
ONNX add output shape validation

geohotstan:
Also removed sequence type support for now because it was barely tested and there was no `sequence_type` usage in top 100 huggingface models. Can add back later with good testing coverage.

Currently we don't have output dtype check yet, but for `test/external/external_test_onnx_ops.py`, ORT does runtime checks on shape and dtype for inputs and outputs.

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -1,7 +1,7 @@
-from typing import Any, Sequence, cast, Literal, Callable
+from typing import Any, cast, Literal, Callable
 import dataclasses, functools, io, math, types
 from tinygrad.tensor import Tensor, _broadcast_shape, ReductionStr
-from tinygrad.helpers import getenv, DEBUG, all_same, prod, flatten, make_tuple, argsort
+from tinygrad.helpers import getenv, DEBUG, prod, flatten, make_tuple, argsort
 from tinygrad.dtype import DType, ConstType, dtypes, ImageDType
 from tinygrad.device import is_dtype_supported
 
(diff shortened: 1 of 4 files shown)


test/external/external_test_onnx_backend.py
```diff
@@ -175,6 +175,12 @@ def supports_device(cls, device: str) -> bool:
 backend_test.exclude('test_ai_onnx_ml_label_encoder_tensor_value_only_mapping_cpu') # bad data type string
 backend_test.exclude('test_ai_onnx_ml_label_encoder_tensor_mapping_cpu') # bad data type string
 
+# no support for sequence
```
chenyuxyz, line 178: these passed before? or was it wrong?

geohotstan: They passed before but the sequence test is very simplistic. Like it passed on the condition that I parsed the sequence input into a list type and the return of the sequence remained a list (if it was a tuple it would've been wrong).

chenyuxyz: oh okay fine i think

extra/onnx.py
```diff
@@ -117,28 +115,34 @@ def __init__(self, model: ModelProto):
...
-    # TODO: need true float16 for dtype checking
-    if spec.is_sequence:
-      if not isinstance(value, Sequence): raise RuntimeError(f"{name} received {value}, expected a sequence type")
-      sequence = [Tensor(v, dtype=spec.dtype, requires_grad=self.is_training) if not isinstance(v, Tensor) else v for v in value]
-      if not all_same(tuple(t.shape for t in sequence)): raise RuntimeError(f"Shapes for {name} sequence must be homogeneous")
-      return sequence
-    tensor = Tensor(value, dtype=spec.dtype, requires_grad=self.is_training) if not isinstance(value, Tensor) else value
-    for dim, (onnx_dim, user_dim_input) in enumerate(zip(spec.shape, tensor.shape, strict=True)):
+  def _valid_shape(self, value: Tensor, spec: OnnxValue):
+    if len(spec.shape) != len(value.shape): return False
```
chenyuxyz, line 127: i would reconstruct this into
```
def _valid_shape(self, actual, spec):
  desired_shape = spec.shape. # have a helper to convert this to a shape
  assert value.shape == desired_shape, "error message"
```

comparing two shape tuples is easier to read then 3 separate return statements, also make it clear which is the actual value and which one is desired value.

raise error in validation function to avoid copy pasting handling the output. the difference in `input/output` is reflected in trace anyway

geohotstan: I ended up pushing the `name` through to the validate function. :s I think having the `name` is quite useful to track the error down.

PR #9722, opened 2025-04-03
Refactor renderer test to use frontend

ignaciosica:
This test failing for `PYTHON` backend on the commit before the bug fix (#9703)
image

thanks @Qazalin for the idea of moving the test to frontend, much simpler now


ignaciosica:
And now that it wrote it this way, I don't think it belongs to renderer failures either, will move it to a more sensible place

chenyuxyz:
still want this?

PR #9724, opened 2025-04-03
`test_tensor_cores_padded` refactor

ignaciosica:
See https://github.com/tinygrad/tinygrad/pull/9606#issuecomment-2774610002 for reference

test tensor cores padded expected uops and correctness separately

`test_tensor_cores_padded` for correctness
`test_tensor_cores_padded_uops` for uops

added new `test_tensor_cores_padded_amd` to check correctness of padded tc with pad != 1


test/test_linearizer.py
```diff
@@ -1108,6 +1108,7 @@ def test_tensor_cores_padded(self):
       if not AMX: # AMX tc.dims[2] == 1
         helper_tc_ensure_uops_and_opts_count(tc.dims[0], tc.dims[1], tc.dims[2]//4, tc.dtype_in, tc.dtype_out, tc_opt=2, ensure_triggered=False)
 
+      if Device.DEFAULT == "AMD": pad = 3 # pad == 1 might be broken for AMD tensor cores due to a compiler bug, see discussion in #9606
```
chenyuxyz, line 1111: ideally we don't change what's being tested based on device, it's harder to maintain. it's fine to have a different test that's only enabled for this device, and skip this test for the same device if it really has a bug.

PR #9729, opened 2025-04-03
Raise Error When Setting requires_grad=True on Non-Float Tensors

### LINKED ISSUE #9723: raise error when setting requires_grad=True on non-float tensors
```
from tinygrad import Tensor
```
would happily run, while
```
import torch
```

aturker1:
Closes #9723.

This PR ensures that tensors with non-floating point data types cannot have requires_grad=True.

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -128,7 +128,7 @@ class OnnxRunner:
   def _parse_input(self, name: str, value: Any, spec: OnnxValue):
     if spec.is_optional and value is None: return None
     # TODO: need true float16 for dtype checking
-    requires_grad = self.is_training if spec.dtype in dtypes.floats else False  
+    requires_grad = self.is_training if spec.dtype in dtypes.floats else False
     if spec.is_sequence:
       if not isinstance(value, Sequence): raise RuntimeError(f"{name} received {value}, expected a sequence type")
       sequence = [Tensor(v, dtype=spec.dtype, requires_grad=requires_grad) if not isinstance(v, Tensor) else v for v in value]
--- a/test/unit/test_gradient.py
+++ b/test/unit/test_gradient.py
@@ -134,7 +134,7 @@ class TestTensorRequiresGrad(unittest.TestCase):
       Tensor([1, 2, 3], requires_grad=True)
 
   def test_requires_grad_method(self):
-    x = Tensor([1, 2, 3]) 
+    x = Tensor([1, 2, 3])
     with self.assertRaises(TypeError):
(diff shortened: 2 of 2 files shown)


chenyuxyz:
test failing

aturker1:
@chenyuxyz done, I guess

tinygrad/tensor.py
```diff
@@ -160,6 +160,9 @@ def __init__(self, data:ConstType|bytes|list|tuple|UOp|'np.ndarray'|pathlib.Path
       dtype = dtype or dtypes.uint8
       data = UOp.new_buffer(f"DISK:{data.resolve()}", data.stat().st_size // dtype.itemsize, dtype)
 
+    if dtype is not None and dtype not in dtypes.floats and requires_grad:
```
chenyuxyz, line 163: use data.dtype here since it's been normalized here

chenyuxyz (commented):
can you also add a check in `requires_grad_`?

huy07111:
**_[]()_**

aturker1:
Hey @chenyuxyz, I think there's an issue. When I used data.dtype to check the Tensor input's dtype, some tests started failing. After investigating, I realized that in Tinygrad, ops like Relu use operations like `>` which produce boolean tensors that still have `requires_grad=True`. As a result, the tests fail.

The issue seems to occur in `_apply_uop`. I think we might need to check the dtype as we did previously instead of using data.dtype, but I'm not completely sure.

aturker1:
Just checking in case you missed my message, @chenyuxyz. I reverted the usage of `data.dtype` due to the explanation above.

tinygrad/tensor.py
```diff
@@ -185,9 +186,14 @@ def _apply_broadcasted_uop(self, fxn:Callable, x:Tensor|ConstType, reverse=False
     return lhs._apply_uop(fxn, rhs)
 
   def requires_grad_(self, requires_grad=True) -> Tensor:
+    self._check_grad(self.dtype, requires_grad)
     self.requires_grad = requires_grad
     return self
 
+  def _check_grad(self, dtype:DType|None, requires_grad:bool|None) -> None:
```
geohot, line 193: Why is this not just in requires_grad function?

geohot:
There's a much cleaner way to write this

tinygrad/tensor.py
aturker1: I wrote a function to avoid duplicate code; the _check_grad is used in the Tensor.__init__ and requires_grad methods. I can place the same code in both methods

PR #9731, opened 2025-04-04
record unittest name in process replay

Qazalin:
The type change is fine because process replay just prints the extra args.


Qazalin:
f_code.name is already accessible, don't need to create strong references to f_locals.

PR #9740, opened 2025-04-04
callback functionality

kreefduik:
## What this is
Makes it so that TRACEMETA=0 context var is respected in a Context() environment.
We need to make it such that the wrapper is add/removed when TRACEMETA changes. To do this I added a callback functionality to the ContextVars, such that a function can run on change.

## Why it's good
fixes https://github.com/tinygrad/tinygrad/issues/6918

## Testing
Added unit test


TheTacoMerchant:
Seeing some really weird CI behavior:
image

These test_transcendental cases need to return early for some values of x due to some instability for certain functions, but this isn't happening. I was able to fix that be changing the way data is sampled for that test case, which fixed almost all issues but this one for WEBGPU:
image

Not really sure why this is only happening on my CI jobs.

geohot:
I wouldn't do it like this, just disable it, don't remove it.

TheTacoMerchant:
Can definitely shave off a few lines if we don't want to remove the wrapper in a Context(TRACEMETA=0) block, but some lines are unavoidable if we want to be able to add the wrappers arbitrarily with Context(TRACEMETA=1) when it was set to 0 initially.

The alternative is to always add the wrappers and just eat the function overhead even when TRACEMETA=0 initially (very few lines to add), but I understand that was a design choice not to do it that way.

PR #9746, opened 2025-04-05
Remote multihost p2p with infiniband verbs

uuuvn:


tinygrad/runtime/ops_amd.py
```diff
@@ -597,6 +599,10 @@ def free(self, mem):
     if mem.va_addr: FileIOInterface.munmap(mem.va_addr, mem.size)
     kfd.AMDKFD_IOC_FREE_MEMORY_OF_GPU(self.kfd, handle=mem.meta.handle)
 
+  def as_dmaref(self, mem:HCQBuffer) -> DMAFdRef:
+    base = mem._base if mem._base is not None else mem
+    return DMAFdRef(kfd.AMDKFD_IOC_EXPORT_DMABUF(KFDIface.kfd, handle=base.meta.handle, flags=0).dmabuf_fd, mem.va_addr-base.va_addr, mem.size)
```
wozeparrot: can split this out as another pr. also want to be able to export dmabufs with AM, can probably use vfio for that.

want to test devmem tcp and io_uring zerocopy rx as well

uuuvn: > can split this out as another pr. also want to be able to export dmabufs with AM, can probably use vfio for that.
> 
> want to test devmem tcp and io_uring zerocopy rx as well

Can't use vfio for that, there were some patches in mailing lists but they weren't merged. I wrote a small kernel module some time ago but it was really janky, can write a good one if that's needed

(this pr isn't ready for review btw)

wozeparrot: yea just commenting since this has some dmabuf stuff, cuz i was going to look into it. You know if udmabuf is usable for exporting a device dmabuf at least for AM? 

We do want AM support since all the cloud machines only use AM

uuuvn: afaik no, udmabuf is just for normal cpu pages, not arbitrary things like pci bars

the jank of my kmod came mostly from it being a kmod and not wanting to claim the pci-e device (eg for vfio to work, linux only allows one driver to claim a device), i think it would be much less jank if i could patch the kernel instead (to eg add a ioctl to resource file or another file like there is `resourceX` `resourceX_wc` already, add a `resource0_dma`)

wozeparrot:
this supports roce? we would rather use roce than rdma over infiniband.

how hard would it be to do a userspace driver for the cards? that should also remove the need to export dmabuf for AM

uuuvn:
> this supports roce? we would rather use roce than rdma over infiniband.

This is RoCEv2 only for now.

RoCE = infiniband transport carried over ethernet link. Supporting other links like RoCEv1 or the infiniband link is probably a couple lines change, likely just lids (which are ignored currently, not needed for RoCEv2 which just has ip encoded in gid)

I'm probably not explaining this well, here is nvidia doc - https://docs.nvidia.com/networking/display/mlnxofedv497100lts/rdma+over+converged+ethernet+(roce)#src-129571092_safe-id-UkRNQW92ZXJDb252ZXJnZWRFdGhlcm5ldChSb0NFKS1Sb0NFTW9kZXM

> how hard would it be to do a userspace driver for the cards? that should also remove the need to export dmabuf for AM

This is the thing I planned to propose after this is merged! Should be very fun and unlock having the gpu queue transfers without the cpu. No idea how hard, haven't looked at any kernel code for mellanox yet

wozeparrot:
conflict

tinygrad/runtime/support/ib.py
```diff
@@ -0,0 +1,167 @@
...
+from tinygrad.device import Buffer, DMACPURef, DMAFdRef
+from tinygrad.helpers import getenv, round_up, DEBUG
+
+DEFAULT_PORT, DEFAULT_GID = getenv("DEFAULT_PORT", 1), getenv("DEFAULT_GID", 3) # DEFAULT_GID=0 for RXE
+IOVA_ALIGN = resource.getpagesize()
+
+# Default soft fd limit is 1024, which is not enough, set soft to hard (maximum allowed by the os)
+soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+resource.setrlimit(resource.RLIMIT_NOFILE, (hard, hard))
+if DEBUG>=2: print(f"IB: Increased fd limit from {soft} to {hard}")
```
wozeparrot: Can this not be in the toplevel?

I ran this and it unexpectedly changed the limit on my machine.

uuuvn: Not sure what you mean, setrlimit call here is per process, not systemwide.

I moved it to be `IBCtx.rlimit_fix`, though i don't see any reason for this

uuuvn: rlimit has two values: soft and hard. soft is what the process is bound by, hard is max that a process can set it's soft to. at no point is this code doing anything that affects other processes (except future children of this process)

wozeparrot: ah i see, maybe its fine toplevel? just didn't like that it was doing stuff on import

PR #9748, opened 2025-04-05
cat loop split

pkotzbach:
went from this:
obraz

to this:
obraz


geohot:
That's the right change, but that pattern is extremely specific. It needs to work generically.

PR #7961, opened 2024-11-29
Fast idiv and mod

S-Lykles:
Needs a lot more tests but it should work.

PR #9755, opened 2025-04-06
fast idiv

S-Lykles:
Fast idiv with test and fuzzer
This should also allow for fast mod but that needs some modification to `validate_index` first (using z3?). 

There are a few cases where the idiv could be optimized that are not by the current version. That only happens when x.vmax is close to the max of the dtype. There are ways to do that but it needs some extra steps to prevent intermediate overflow.

I checked compiler output and the assembly this generates is equal or better than what compilers generate themselves. In general tinygrad knows more about the bounds of the variables than nvcc so it can generate better expressions. (Clang I have found to be pretty good though)

### DIFF
--- a/tinygrad/codegen/devectorizer.py
+++ b/tinygrad/codegen/devectorizer.py
@@ -2,7 +2,8 @@ from typing import Optional, Any, Callable, cast
 import functools, operator, itertools
 from collections import defaultdict
 from dataclasses import dataclass
-from tinygrad.dtype import dtypes, ImageDType, PtrDType
+from tinygrad.device import is_dtype_supported
+from tinygrad.dtype import dtypes, ImageDType, PtrDType, promo_lattice
 from tinygrad.ops import UOp, Ops, UPat, PatternMatcher, resolve, graph_rewrite, GroupOp, identity_element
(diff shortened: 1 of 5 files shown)


S-Lykles:
With the fuzzer I feel confident this is actually correct

tinygrad/codegen/devectorizer.py
```diff
@@ -137,6 +159,9 @@ def get_late_rewrite_patterns(ops, force_transcendental=False):
     # no reason to check x>=0 for uints
     pat += [(UPat.var("x", dtypes.uints)//UPat.cvar("c"), lambda x,c: x >> v if (v:=powers_of_two.get(c.arg, 0)) else None)]
     pat += [(UPat.var("x", dtypes.sints)//UPat.cvar("c"), lambda x,c: x >> v if (v:=powers_of_two.get(c.arg, 0)) and resolve(x>=0,False) else None)]
+    pat += [(UPat.var("x", dtypes.ints)//UPat.cvar("d"), lambda x, d: fast_idiv(x, d.arg))]
```
chenyuxyz, line 162: add the flag here? so it does not spend time matching it if not enabled

S-Lykles:
So previously in #7961 this was noticably faster in CIFAR with wino, @chenyuxyz could you benchmark  #9767 to see if the speed was because of fast mod?

PR #9767, opened 2025-04-07
Fast idiv, also fast mod

S-Lykles:
Basically #9755 with the mod pattern enabled and `validate_indexing` commented out.
Just for benchmarking


S-Lykles:
Continued in #10055

PR #9771, opened 2025-04-07
OptOps.LDS

ignaciosica:
Use of shared memory is broken for PTX. (LDS test fails with `RecursionError: maximum recursion depth exceeded`, same as for TC=3 emulation, probably there is a bug on how `Ops.DEFINE_LOCAL` is handled). Will try to address this before merging this pr. 

Can't run with `process replay` as the addition of a new item in `OptOps` enum breaks pickle.

I don't know why NV won't finish the job, it passes all the tests fairly quickly on the real device (tiny19).
image

If added to BEAM, it makes matmul for 4096x4096 more fast on my computer. Right now it only makes it fast for large matmuls. For matmuls of smaller sizes, as the hit rate for the cache is good already good, so it does not actually make it faster.
image
BEAM=4 is faster on this branch vs BEAM=16 on master.
image

This is not addressing thread access pattern to local (LDS_SWAP) and local layout(LDS_LAYOUT) yet. Those are going to be independent kernel opts. 

I've already been trying the former, named. I actually could make the search space quite good, quickly rejecting invalid permutations, but the way BEAM works right now makes it hard for it to end up using it, as it usually needs 3/4 consecutive opts to actually make it good (local and unroll of the same size, LDS on the buffer(that is usually discarded for input buffers as without correcting the thread pattern it makes it slower) and then the LDS_SWAP opt that corrects the thread pattern (only one permutation at a time))

### DIFF
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -357,7 +357,7 @@ class Kernel:
 
     axis = self.real_axis(opt)
     if opt.op is not OptOps.LDS:
-      if opt.op != OptOps.SWAP: check(not any(self.lds), f"cant reshape after LDS {self.applied_opts=} {opt=}")
+      if opt.op != OptOps.SWAP: check(not any(self.lds), f"can't reshape after LDS {self.applied_opts=} {opt=}")
       check(axis < len(self.full_shape), "invalid axis")
 
     if opt.op is OptOps.SWAP: amt = cast(int, opt.arg)  # arg is an axis in the SWAPs
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -356,7 +356,7 @@ class Kernel:
       return
 
     axis = self.real_axis(opt)
-    if opt.op != OptOps.LDS:
+    if opt.op is not OptOps.LDS:
       check(not any(self.lds), f"cant reshape after LDS {self.applied_opts=} {opt=}")
       check(axis < len(self.full_shape), "invalid axis")
 
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -673,7 +673,6 @@ class Kernel:
     del fixup_ast
     return graph_rewrite(fixed_ast, view_left)
 
-
   def apply_lds(self, ast) -> UOp:
     def transform(ctx:tuple[Kernel, set[UOp]], global_access:UOp):
       if (buf:=global_access.src[0]).arg in ctx[1] or global_access.src[0].op is not Ops.DEFINE_GLOBAL: return None
@@ -698,7 +697,6 @@ class Kernel:
       return None
 
     return graph_rewrite(ast, PatternMatcher([(UPat((Ops.LOAD, Ops.STORE), name="global_access"), transform)]), ctx=(self, set()))
-
 
   # **** this is the lowerer ****
 
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -704,9 +704,7 @@ class Kernel:
     if getenv("VIZ"): graph_rewrite(self.ast, PatternMatcher([]), name="View Base AST")
 
     modified_ast = self.get_optimized_ast(name_override)
-    assert modified_ast.arg.dont_use_locals == self.dont_use_locals
     modified_ast = self.apply_lds(modified_ast)
-    assert modified_ast.arg.dont_use_locals == self.dont_use_locals
     if ast_transform is not None: modified_ast = ast_transform(self, modified_ast)
 
     if DEBUG >= 3:
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -675,7 +675,7 @@ class Kernel:
 
   def apply_lds(self, ast) -> UOp:
     def transform(ctx:tuple[Kernel, set[UOp]], global_access:UOp):
-      if (buf:=global_access.src[0]).op is not Ops.DEFINE_GLOBAL or buf.arg in ctx[1] or not (k := ctx[0]).lds[buf.arg]: return None
+      if (buf := global_access.src[0]).op is not Ops.DEFINE_GLOBAL or buf.arg in ctx[1] or not (k := ctx[0]).lds[buf.arg]: return None
       ctx[1].add(buf.arg)
       global_st: ShapeTracker = global_access.src[1].arg
       gd, fr, fu, shape = k.global_dims, k.first_reduce, k.first_upcast, []
--- a/tinygrad/codegen/kernel.py
+++ b/tinygrad/codegen/kernel.py
@@ -5,7 +5,7 @@ from collections import defaultdict
 from typing import Optional, cast, Final, Callable, Sequence
 
 from tinygrad.ops import GroupOp, KernelInfo, UOp, Ops, can_pad, resolve, Variable, sint, graph_rewrite, track_rewrites, view_left, print_uops
-from tinygrad.ops import PatternMatcher
+from tinygrad.ops import PatternMatcher, UPat
 from tinygrad.spec import type_verify, shape_spec
 from tinygrad.device import Device
 from tinygrad.renderer import Renderer, TensorCore, ProgramSpec, Opt, OptOps
@@ -78,6 +78,8 @@ class Kernel:
     self.tensor_core_opts: Optional[TensorCoreOptions] = None
     self.use_tensor_cores: int = 0
     self.dont_use_locals: bool = False
+    self.lds: list[bool] = [False] * len(self.bufs)
+    self.smem_usage:int = 0
 
     # group simplifies
     self.simplify_ones()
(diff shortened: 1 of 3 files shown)


tinygrad/codegen/kernel.py
```diff
@@ -656,6 +671,29 @@ def get_tc_swizzle_st(shape, local_perm, upcast_perm):
     del fixup_ast
     return graph_rewrite(fixed_ast, view_left)
 
+  def apply_lds(self, ast) -> UOp:
+    def transform(ctx:tuple[Kernel, set[UOp]], global_access:UOp):
+      if (buf := global_access.src[0]).op is not Ops.DEFINE_GLOBAL or buf.arg in ctx[1] or not (k := ctx[0]).lds[buf.arg]: return None
+      ctx[1].add(buf.arg)
+      global_st: ShapeTracker = global_access.src[1].arg
+      gd, fr, fu, shape = k.global_dims, k.first_reduce, k.first_upcast, []
+      for i, st in enumerate(global_st.real_strides(True)): shape.append(global_st.shape[i] if i >= gd and st != 0 and (i < fr or i >= fu) else 1)
```
geohot, line 680: what if st is None?

ignaciosica: That's why it's skipped right now for padded kernels. sometimes it breaks, the same happens with TC=3

geohot:
Do you plan to finish this?

ignaciosica:
Yes, yesterday I resumed work but I've been stuck with the ptx bug, it hits maximum recursion depth in the linearizer

geohot:
PTX still failing?

ignaciosica:
yes, the fix is already upstreamed in #9967, I haven't merged it yet because I'm refactoring tc tests to add the tc=3 regression test for the fix

ignaciosica:
move old description to comment for better doc: Use of shared memory is broken for PTX. (LDS test fails with `RecursionError: maximum recursion depth exceeded`, same as for TC=3 emulation, probably there is a bug on how `Ops.DEFINE_LOCAL` is handled). Will try to address this before merging this pr. 

Can't run with `process replay` as the addition of a new item in `OptOps` enum breaks pickle.

I don't know why NV won't finish the job, it passes all the tests fairly quickly on the real device (tiny19).
image

If added to BEAM, it makes matmul for 4096x4096 more fast on my computer. Right now it only makes it fast for large matmuls. For matmuls of smaller sizes, as the hit rate for the cache is good already good, so it does not actually make it faster.
image
BEAM=4 is faster on this branch vs BEAM=16 on master.
image

This is not addressing thread access pattern to local (LDS_SWAP) and local layout(LDS_LAYOUT) yet. Those are going to be independent kernel opts. 

I've already been trying the former, named. I actually could make the search space quite good, quickly rejecting invalid permutations, but the way BEAM works right now makes it hard for it to end up using it, as it usually needs 3/4 consecutive opts to actually make it good (local and unroll of the same size, LDS on the buffer(that is usually discarded for input buffers as without correcting the thread pattern it makes it slower) and then the LDS_SWAP opt that corrects the thread pattern (only one permutation at a time))

TODO:
- [x] gate on masked st
- [x] add test for masked st

ignaciosica:
Both lds and tc=3 tests work on tiny10 for AMD_LLVM but fail in ci benchmark's red box and also in `tiny11`. I think it is related to llvm version. Taking a deeper look into it

ignaciosica:
[Unit Tests / Linux (amdllvm)](https://github.com/tinygrad/tinygrad/actions/runs/14888180735/job/41813326928?pr=9771): failing due to remu with AMD_LLVM=1 bug, same test pass on real device; need to investigate further.

[Benchmarks / tinybox red Training](https://github.com/tinygrad/tinygrad/actions/runs/14885504435/job/41803905207): failing due to outdated llvm version in CI, works on both tiny11 and tiny10 machines.

[Benchmarks / Mac Benchmark](https://github.com/tinygrad/tinygrad/actions/runs/14885504435/job/41803903462): test_lds_tc tests failing with assert error. Worked on two different m series devices. Could not replicate error anywhere else; need to investigate further.

ignaciosica:
adding `-e` to defaults fixed silent fail in Mac Benchmark. Moving afterwards to independent pr

ignaciosica:
thanks for pointing out the issue, I addressed it in #10335

geohot:
Is this close? Looks like one test failure, is it the remu stuff?

ignaciosica:
It is close, the only issue blocking is metal precision that is suspect might be fixed by updating os in the ci machine. I couldn't manage to run the tests yet on the new machine as connection keeps lagging, but one run in that machine should me enough to see if its fixed with the new os build.

geohot:
We can skip the METAL test, I don't think it's your bug. Let's get this merged!

tinygrad/codegen/kernel.py
```diff
@@ -423,6 +428,16 @@ def apply_opt(self, opt:Opt, append_opt:bool=True):
           self.sts[i] = st.pad(((0,0),) * axis + ((0,ru),) + ((0,0),) * (len(st.shape)-axis-1))
           padded = True
       check(padded, "nothing was padded")
+    elif opt.op is OptOps.LDS:
+      check(0 <= axis < len(self.bufs) and not self.lds[axis], f"invalid lds {axis=}")
+      check(self.group_for_reduces == 0, "can't apply lds with group/grouptop")
+      # TODO: remove buf_st hack
+      buf_st = self.sts[axis if axis == 0 else (1 if axis == 2 else 2)] if len(self.bufs) == 3 else self.sts[axis]
```
geohot, line 435: What is this line?

tinygrad/codegen/kernel.py
```diff
@@ -423,6 +428,16 @@ def apply_opt(self, opt:Opt, append_opt:bool=True):
...
+      check(all(not buf_st.axis_is_masked(i) for i in range(len(buf_st.shape))), "can't apply lds with masked axis")
+      self.smem_usage += prod(sz for i,(sz,st) in enumerate(zip(buf_st.shape, buf_st.real_strides()))
```
geohot, line 437: This isn't right. What if dimensions are aliased, like in a conv?

tinygrad/codegen/kernel.py
ignaciosica: this is to correctly keep track of `smem_usage`; if this hack isn't applied, for example, if lds is applied on buf 2 it will calculate the local buf size based on buf 1 st.

this is what makes it required

https://github.com/tinygrad/tinygrad/blob/1e0a59aca48f6855aec00e20c3339beee1197d3e/tinygrad/codegen/kernel.py#L50-L51

tinygrad/codegen/kernel.py
```diff
@@ -529,6 +544,27 @@ def get_tc_swizzle_st(shape, local_perm, upcast_perm):
     del fixup_ast
     return graph_rewrite(fixed_ast, view_left, name="fixup optimized AST")
 
+  def apply_lds(self, ast) -> UOp:
+    def transform(ctx:tuple[Kernel, set[UOp]], global_access:UOp):
+      if (buf := global_access.src[0]).op is not Ops.DEFINE_GLOBAL or buf.arg in ctx[1] or not (k := ctx[0]).lds[buf.arg]: return None
+      ctx[1].add(buf.arg)
+      global_st: ShapeTracker = global_access.src[1].arg
+      gd, fr, fu, shape = k.global_dims, k.first_reduce, k.first_upcast, []
```
geohot, line 552: Why!?! Just write them out each time so it's readable. Don't prematurely optimize.

tinygrad/codegen/kernel.py
geohot: Yea, you have to fix the bug. This is not a solution and can't be merged. It's not correct, it just happens to pass the tests.

tinygrad/codegen/kernel.py
```diff
@@ -529,6 +544,27 @@ def get_tc_swizzle_st(shape, local_perm, upcast_perm):
     del fixup_ast
     return graph_rewrite(fixed_ast, view_left, name="fixup optimized AST")
 
+  def apply_lds(self, ast) -> UOp:
+    def transform(ctx:tuple[Kernel, set[UOp]], global_access:UOp):
+      if (buf := global_access.src[0]).op is not Ops.DEFINE_GLOBAL or buf.arg in ctx[1] or not (k := ctx[0]).lds[buf.arg]: return None
```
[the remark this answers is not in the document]
ignaciosica: `ctx` is the context of the graph rewrite

ignaciosica:
sure, I'll address the nits shortly and take a look into the bufs order bug later

ignaciosica:
new blockers for lds:
- [ ] `bufs` ordering in kernel init bug
- [ ] pad is broken but there is no fundamental reason for it to not work with lds

tinygrad/codegen/kernel.py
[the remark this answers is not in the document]
geohot: no, i know that, but what's in it? maybe unpack to readable names on the first line

ignaciosica:
rebased to a cleaner take after the kernel/realize/get_optimized_ast refactors

PR #9774, opened 2025-04-07
[bounty] z3 index validation [pr]

S-Lykles:


S-Lykles:
Definitely a performance hit, will investigate more

Good enough so far to fix 
```
FAILED test/test_linearizer_failures.py::TestLinearizerFailures::test_failure_58 - AssertionError: unexpected success on GPU
assert 'GPU' not in ['METAL', 'GPU', 'AMD', 'NV']
```

S-Lykles:
Currently in things like gpt nothing is being validated because of 

```python
if (sz := cast(PtrDType, idx.src[0].dtype).size) == -1: return True
```

tinygrad/spec.py
```diff
@@ -46,16 +55,34 @@
...
-      print(f"OUT OF BOUNDS ACCESS in INDEX {vmin} - {vmax} not in 0 - {sz}. {idx.src[1].render()=}")
-      return False
+  if not CHECK_OOB or isinstance(idx.dtype, ImageDType) or (sz := cast(PtrDType, idx.src[0].dtype).size) == -1: return True
+
+  all_uops = list((idx.toposort | mask.toposort).keys() if mask is not None else idx.toposort.keys())
+  all_uops += flatten(map(lambda x: x.arg[1].toposort.keys(), filter(lambda x: x.op is Ops.SPECIAL and isinstance(x.arg[1], UOp), all_uops)))
+  # WEBGPU has a BITCAST in the index. TODO: fix
+  if any(x.op is Ops.BITCAST for x in all_uops): return True
+
+  # We can use UOp min/max to do a faster check, but it can give false positive since its not an exact bound and doesn't consider the mask
```
S-Lykles, line 65: With this its actually not much perf loss

tinygrad/spec.py
```diff
@@ -46,16 +55,34 @@
...
+  if 0<=idx.src[1].vmin and idx.src[1].vmax<sz: return True
```
S-Lykles, line 66: With this its actually not much perf loss

tinygrad/spec.py
```diff
@@ -1,7 +1,18 @@
...
-from tinygrad.helpers import all_same, dedup, prod, getenv, DEBUG
+from tinygrad.helpers import all_same, dedup, flatten, prod, getenv, DEBUG, CHECK_OOB
+if CHECK_OOB:
+  import z3
+
+  # IDIV is truncated division but z3 does floored division and mod by power of two sometimes uses Ops.AND
+  def z3_cdiv(a,b): return z3.If(a<0, (a+(b-1))/b, a/b)
+  z3_alu: dict[Ops, Callable] = python_alu | {Ops.MOD: lambda a,b: a-z3_cdiv(a,b)*b, Ops.IDIV: z3_cdiv, Ops.CAST: lambda x:x,
+    Ops.SHR: lambda a,b: a/(2**b), Ops.SHL: lambda a,b: a*(2**b), Ops.AND: lambda a,b: a%(b+1) if isinstance(b, (int, z3.ArithRef)) else a&b}
+  def z3_eval(u: UOp|int, ctx: dict[UOp, 'z3.ArithRef']) -> 'int|z3.ArithRef':
```
geohot, line 12: Can this be done with a graph_rewrite? Like look at how we do rendering.

The problem with this is that it isn't cached and you can get 2^n blowup.

PR #9779, opened 2025-04-08
Fix test_failure_53: Disable GROUPTOP for uchar reductions

snagasuri:
Fix test_failure_53 by disabling GROUPTOP for uchar reductions

This PR fixes test_failure_53 by preventing GROUPTOP optimization when the reduction involves uchar types, where type compatibility issues can arise. The fix takes a defensive approach by disabling the optimization in cases where it could lead to type safety problems. Submitting for the test_failure_53 bounty.


snagasuri:
Hi @geohot, saw this was closed. Was the fix incorrect, or are you planning to handle test_failure_53 another way? Happy to adjust the approach if needed!

PR #9784, opened 2025-04-08
view/movement ops tests [pr]

Qazalin:
VIEW can apply on top of another VIEW and change the ShapeTracker in an additive way.
MovementOps can only apply on the base and create a new ShapeTracker.


Qazalin:
Merging the tests, fine with the behavior that exists for now.

PR #9792, opened 2025-04-08
Off by one error in start_pos

S-Lykles:
Variable upper bound is inclusive


geohot:
Any reason this is draft? It looks correct to me

S-Lykles:
Yeah forgot to mark it ready

PR #9795, opened 2025-04-08
[Bounty] CPUGraph on ClangDevice

larrythwu:
Work in progress for "CPUGraph working with CLANG+LLVM" bounty.

In CPUGraph, the batched function is rendered at the very bottom of the C program. However, when we compile and load it, the memory pointer points to the very beginning of the shared library (ie. the first function).

from device.py:
` self.mem = mmap(-1, len(lib), MAP_ANON | MAP_PRIVATE | (MAP_JIT if OSX else 0), PROT_READ | PROT_WRITE | PROT_EXEC)
`

then we create a function pointer using this memory address, which is then not pointing to the batched function:

from device.py:
`self.fxn = ctypes.CFUNCTYPE(None)(mv_address(self.mem))
`

Therefore, when we run the JIT, it is not batched function that gets run.

My change labels the batched function with an attribute and when assembling the elf we manually relocate the function that's labeled to the top.

### DIFF
--- a/test/test_graph.py
+++ b/test/test_graph.py
@@ -108,7 +108,8 @@ class TestGraph(unittest.TestCase):
     helper_test_graphs(Device[d0].graph, graphs)
 
   def test_order_copy_writed(self):
-    if not issubclass(Device[Device.DEFAULT].graph, MultiGraphRunner): self.skipTest("graph does not supported (not MultiGraphRunner)")
+    if not issubclass(getattr(Device[Device.DEFAULT].graph, 'func', Device[Device.DEFAULT].graph), MultiGraphRunner): 
+      self.skipTest("graph does not supported (not MultiGraphRunner)")
 
     d0 = Device.DEFAULT
     b0 = [helper_alloc_rawbuffer(d0, fill=True) for _ in range(4)]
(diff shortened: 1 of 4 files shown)


test/test_graph.py
```diff
@@ -108,7 +108,8 @@ def test_order_write_read_same_buf(self):
     helper_test_graphs(Device[d0].graph, graphs)
 
   def test_order_copy_writed(self):
-    if not issubclass(Device[Device.DEFAULT].graph, MultiGraphRunner): self.skipTest("graph does not supported (not MultiGraphRunner)")
+    if not issubclass(getattr(Device[Device.DEFAULT].graph, 'func', Device[Device.DEFAULT].graph), MultiGraphRunner):
```
geohot, line 111: Don't copy paste this looks awful. Refactor

tinygrad/runtime/ops_cpu.py
```diff
@@ -3,6 +3,8 @@
 from tinygrad.device import Compiled, Compiler, MallocAllocator, CPUProgram
 from tinygrad.runtime.support.elf import jit_loader
 from tinygrad.renderer.cstyle import ClangRenderer
+from tinygrad.runtime.graph.cpu import CPUGraph
+import functools
```
geohot, line 7: Please read diffs and always be asking if you can make this nicer

uuuvn:
`static void` on everything except `void batched` instead of `__attribute__((section(".text.last")))` on `void batched`? It shouldn't require any changes to elf stuff

larrythwu:
> `static void` on everything except `void batched` instead of `__attribute__((section(".text.last")))` on `void batched`? It shouldn't require any changes to elf stuff

Good point!

geohot:
Tons of copy paste in diff

larrythwu:
> Tons of copy paste in diff

Sorry could you please clarify, is there anything I am missing?

PR #9807, opened 2025-04-09
Speedup rewrite by checking less often for op match

Yexo:
match() is called from rewrite(), which already checks for the op match, and is recursively called by match itself. By checking for the op match before the recursive call we can remove it as the very first check that match does. Given how often match is called, this gives a 10%-20% speedup on my machine for: `time NULL=1 python3 examples/beautiful_mnist.py`


Yexo:
This got taken over by https://github.com/tinygrad/tinygrad/pull/9737, can't measure improvements after that.

PR #9810, opened 2025-04-09
BF16 autocasts to float32

1957-001B:
fixes #6909

PR #9812, opened 2025-04-09
Removed all references to fix_bf16 hack 

1957-001B:
removed fix_bf16 from extra/models and changed all examples that reference fix_bf16 as #9810 autocasts bf16 to float.


geohot:
Tests failing

PR #9813, opened 2025-04-09
[Bounty] Fix metadata for sharded tensors

kreefduik:
Should fix:
Fix metadata in 'DEBUG=2 LLVM=1 python3 test/test_multitensor.py TestMultiTensor.test_data_parallel_resnet_train_step' + add test


tinygrad/ops.py
```diff
@@ -960,9 +965,9 @@ def graph_rewrite(sink:UOp, pm:PatternMatcher, ctx=None, bottom_up=False, name=N
   return rewrite_ctx.bottom_up_rewrite(sink) if bottom_up else rewrite_ctx.top_down_rewrite(sink)
 
 @track_matches
-def graph_rewrite_map(sink:UOp, pm:PatternMatcher, ctx=None, bottom_up=False, name=None, track_children=False, input_map=None) -> dict[UOp, UOp]:
+def graph_rewrite_map(sink:UOp, pm:PatternMatcher, ctx=None, bottom_up=False, name=None, track_children=False, input_map=None, keep_metadata=False) -> dict[UOp, UOp]:# noqa: E501
   rewrite_ctx = RewriteContext(pm, ctx, children=sink.get_children_map() if track_children else None)
-  new_map = {k:(rewrite_ctx.bottom_up_rewrite(k) if bottom_up else rewrite_ctx.top_down_rewrite(k)) for k in list(sink.toposort)[::-1]}
+  new_map = {k:(rewrite_ctx.bottom_up_rewrite(k) if bottom_up else rewrite_ctx.top_down_rewrite(k, keep_metadata=keep_metadata)) for k in list(sink.toposort)[::-1]} # noqa: E501
```
geohot, line 970: No E501. No new flags to this function

tinygrad/ops.py
```diff
@@ -940,11 +940,16 @@ def update_children(self):
         while (nv:=self.replace.get(tv, None)) is not None and nv is not tv: tv = nv
         new_child[tv] = None
       self.children[k] = new_child
-  def top_down_rewrite(self, n:UOp) -> UOp:
+  def top_down_rewrite(self, n:UOp, keep_metadata=False) -> UOp:
```
geohot, line 943: This will make everything slow

kreefduik: Point taken. Will try a different approach.

kreefduik:
Okay new approach. Now metadata is only preserved when using the multi_pm pattern matcher.

geohot:
Don't modify any core classes like UOp!

kreefduik:
New approach is to extend PatternMatcher in un-compiled mode to accept an on-match wrapper, which can wrap the function that gets called on pattern match. This should create the least overhead for the majority of rewrites at the expense of not getting the multi_pm pattern matcher to be compiled. 

I tested with TRACK_MATCH_STATS=1 for the Resnet train step and saw ~0.5% slowdown, albeit with a lot of variance due to my local hardware.

kreefduik:
@geohot take a look when you have a minute.

geohot:
I'm now repeating the same point three times

Issue #9814, opened 2025-04-09
device="CLOUD" (without env HOST)  fails when the cloud back-end re-uses the Device handle from the parent process.


xl0:
With a clean env:

```py
import tinygrad as tg
from tinygrad import Tensor

print(tg.Device.DEFAULT)

print(Tensor(1).numpy())

print(Tensor(1).to("CLOUD").numpy())
```

Exception
```
NV
1
start cloud server on 6667 with device NV
connection established with ('127.0.0.1', 32882), socket: 85
127.0.0.1 - - [09/Apr/2025 12:34:29] "GET [/renderer](https://file+.vscode-resource.vscode-cdn.net/renderer) HTTP/1.1" 200 -
----------------------------------------
[Errno 111] Connection refused

Exception occurred during processing of request from ('127.0.0.1', 32882)
Traceback (most recent call last):
  File "/home/xl0/work/projects/grads/tinygrad/tinygrad/device.py", line 210, in alloc
    try: return super().alloc(size, options)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xl0/work/projects/grads/tinygrad/tinygrad/device.py", line 190, in alloc
    return self._alloc(size, options if options is not None else BufferSpec())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xl0/work/projects/grads/tinygrad/tinygrad/runtime/ops_nv.py", line 273, in _alloc
    return self.dev._gpu_alloc(size, cpu_access=options.cpu_access, tag=f"user memory ({options})")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xl0/work/projects/grads/tinygrad/tinygrad/runtime/ops_nv.py", line 354, in _gpu_alloc
    if cpu_access: va_addr = self._gpu_map_to_cpu(mem_handle, size, target=va_addr, flags=map_flags, system=uncached)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #7038, opened 2024-10-14
Adding CUDA_HOME for conda env ?


keithyau:
`conda env config vars set CUDA_HOME="XXX"`

at runtime/support/compiler_cuda.py line 50

```
class CUDACompiler(Compiler):
  def __init__(self, arch:str, cache_key:str="cuda"):
    import os
    cuda_home = os.environ.get('CUDA_HOME')
    self.arch, self.compile_options = arch, [f'--gpu-architecture={arch}', "-I/usr/local/cuda/include", "-I/usr/include", "-I/opt/cuda/include/", f"-I{cuda_home}"]
```

 NV=1 CACHELEVEL=0 python3 examples/stable_diffusion.py

geohot:
I don't think that's the right fix

keithyau:
yeah, worked on my conda env

remorses:
Docker images also use `/usr/local/cuda/lib64`, it would be cool to add that too in the default list

xl0:
@geohot Do you have a better idea? I ran into this, and can't find a better solution.

- Try to find the include dir based on the location of libnvrtc.so:
    - fing_library() does not always give you a full path, often it's just the filename
    - /include/ path might be different relatively to the lib on different systems
    -  Not impossible, but many lines.
- Don't use CUDA headers. I have a quick look at PyTorch, and when they use nvrtc, they don't pass any header paths at all.
    - Not sure if feasible.
- Triton uses some ctazy contraption with a custom llvm? No.
- Call `nvcc` instead of using nvrtc. Messy, but solves the header issue, as long as the right nvcc is in the path.
 
- This solution above, but maybe
```py
    self.arch, self.compile_options = arch, [f'--gpu-architecture={arch}'] + [f"-D{CUDA_HOME}/include/"] if CUDA_HOME else ["-I/usr/local/cuda/include", "-I/usr/include", "-I/opt/cuda/include/"]
```
with CUDA_HOME being a ContextVar in helpers.py

I also don't love that we pass multiple possible directories when CUDA_HOME is not set, because we will end up with headers that don't match the cuda version the user expects. It's better to have a clear failure, than a silent version mismatch.

WDYT?

PR #9817, opened 2025-04-09
Fix my pre commit tests

xl0:
I had a few failures when runnign the pre-commit tests:

- test_allreduce fails when less than 6 GPUS are available, at least with NV/CUDA
    - Make N dynamic (up to 8), test is skipped if less than 2.

- compiler_cuda will look for CUDA headers in pre-determined locations, fails with mamba/conda if the system CUDA version is different/missing, see #7038
   - Prepend CUDA_HOME to the -Is if set in environment.

- external_test_example.py fails at least on NV/CUDA for the CLOUD device, see #9814
   -  Exclude CLOUD from the test

Please let me know if you want any of them done differently/dropped.

### DIFF
--- a/docs/env_vars.md
+++ b/docs/env_vars.md
@@ -50,6 +50,7 @@ JIT                 | [0-2]      | 0=disabled, 1=[jit enabled](quickstart.md#jit
 VIZ                 | [1]        | 0=disabled, 1=[viz enabled](https://github.com/tinygrad/tinygrad/tree/master/tinygrad/viz)
 ALLOW_TF32          | [1]        | enable TensorFloat-32 tensor cores on Ampere or newer GPUs.
 WEBGPU_BACKEND      | [WGPUBackendType_Metal, ...]          | Force select a backend for WebGPU (Metal, DirectX, OpenGL, Vulkan...)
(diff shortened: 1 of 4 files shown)


xl0:
TC test fails. I think it's just flaky, I did not touch anything near it.

geohot:
Would merge the no CLOUD in external test example right away. I'd merge the CUDA_HOME but that code should be cleaned up.

xl0:
Ok, I did not like it, so
- Renamed to CUDA_INC
- If set, ONLY look in CUDA_INC
- Added doc

geohot:
What's CUDA_INC? I'd only merge the CLOUD one of these.

xl0:
@geohot CUDA_INC is the path to the cuda headers.

Can you elaborate what's not good with the code?
```py
    self.arch, self.compile_options = arch, [f'--gpu-architecture={arch}']
    self.compile_options += [f"-I{CUDA_INC}"] if CUDA_INC else ["-I/usr/local/cuda/include", "-I/usr/include", "-I/opt/cuda/include"]
```
How could you write it better?
```py
    self.arch, self.compile_options = arch, ([f'--gpu-architecture={arch}']+[f"-I{CUDA_INC}"] if CUDA_INC else
                                             ["-I/usr/local/cuda/include", "-I/usr/include", "-I/opt/cuda/include"])
```
?

And what about reduceall. It fails on hardware with 1-5 local GPUs.

geohot:
"Fix my pre commit tests" is not a good name for a PR

PR #9822, opened 2025-04-10
RetinaNet model type annotations and loss functions

francislata:
# Overview
This PR brings in all of the type annotations and loss functions needed for training RetinaNet (#8385). Running `BASE_DIR=/raid/datasets/openimages/ PYTHONPATH=. MODEL=retinanet python3 examples/mlperf/model_eval.py` works as expected.

### DIFF
--- a/extra/models/retinanet.py
+++ b/extra/models/retinanet.py
@@ -1,8 +1,13 @@
 import math
+from tinygrad import Tensor, dtypes
 from tinygrad.helpers import flatten, get_child
-import tinygrad.nn as nn
+from examples.mlperf.helpers import generate_anchors, BoxCoder
+from examples.mlperf.losses import sigmoid_focal_loss, l1_loss
 from extra.models.resnet import ResNet
+import tinygrad.nn as nn
 import numpy as np
+
+ConvFPN = ConvHead = ConvClassificationHeadLogits = nn.Conv2d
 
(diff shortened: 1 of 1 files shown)


extra/models/retinanet.py
```diff
@@ -145,33 +132,73 @@ def postprocess_detections(self, predictions, input_size=(800, 800), image_sizes
...
+    if Tensor.training:
+      assert labels is not None and matches is not None, "labels and matches should be passed in when training"
+      return self._compute_loss(out.cast(dtypes.float32), labels, matches)
+
+    return out.sigmoid()
+
+  def _compute_loss(self, x:Tensor, labels:Tensor, matches:Tensor) -> Tensor:
+    labels = ((labels + 1) * (fg_idxs := matches >= 0) - 1).one_hot(num_classes=x.shape[-1])
+    valid_idxs = (matches != -2).reshape(matches.shape[0], -1, 1)
+    loss = valid_idxs.where(sigmoid_focal_loss(x, labels), 0).sum(-1).sum(-1)
```
chenyuxyz, line 153: any reason that it's `sum(-1).sum(-1)` and not `sum((-1, -2))`? the later has one less kernel

francislata: yes, this can definitely be converted into the latter option.

PR #9841, opened 2025-04-11
test for broken method_cache with symbolic shape [pr]

Qazalin:
```
python test/test_symbolic_ops.py TestSymbolicOps.test_method_cache
```
re-linearizes the same Kernel, this used to codegen only once. This behavior changed after: https://github.com/tinygrad/tinygrad/commit/6608ba316d04ec0bb5c332dab59ee90355bea7c7.

I assume part of the reason we wanted Variable shape is to avoid repeating codegen?

### DIFF
--- a/test/test_symbolic_ops.py
+++ b/test/test_symbolic_ops.py
@@ -1,10 +1,24 @@
 import unittest
 from tinygrad import Variable
 from tinygrad.tensor import Tensor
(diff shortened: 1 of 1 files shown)


Qazalin:
This is expected because the size of the buffer is different in every allocation.
In real cases like the LLMs we allocate to the max size so it uses the method_cache correctly.

PR #9843, opened 2025-04-11
use CPU for test_schedule_ring

chenyuxyz:
there's a world that sharding contiguous empty does not create copy kernels, but it does now


Qazalin:
The tricky thing with the rewrite rule is that empty creates an unallocated buffer,
we can't really rewrite a COPY(BUFFER) with another BUFFER as the buffer may have state.

PR #9850, opened 2025-04-11
raise error for non floats

utkarshgill:
#9723


utkarshgill:
Isn't this what was expected? All tests passed too?
```
(venv) engelbart@MacBook-Pro tinygrad % python -c "from tinygrad import Tensor; print(Tensor([1, 2, 3], requires_grad=True).tolist())"
  File "<string>", line 1, in <module>
  File "/Users/engelbart/Desktop/stuff/tinygrad/tinygrad/tensor.py", line 168, in __init__
    raise RuntimeError(f"Only Tensors of floating point dtype can require gradients, got {data.dtype}")
RuntimeError: Only Tensors of floating point dtype can require gradients, got dtypes.int
(venv) engelbart@Utkarshs-MacBook-Pro tinygrad % 
```

PR #9855, opened 2025-04-11
fix diamond assigns before mapping tensors to assigns

Qazalin:
It used to need a map from the tensor sink to the fixed up sched_sink with the extra edges for diamond assign.
It should instead be local to the Tensor UOp. ASSIGN(3) it always before ASSIGN(0), figuring out where to put the edge isn't local - we have to check the full graph - but expressing the edge is local to the ASSIGN UOp.
image


Qazalin:
Process replay diffs as expected:
image
Now becomes_map is created after diamond fixup.

PR #9856, opened 2025-04-11
prevent VIZ from blocking GC of buffers

hooved:
This PR prevents `VIZ=1` from interfering with GC of `Ops.BUFFER` UOps (and their corresponding `Buffer` objects).

With `VIZ=1`, `TrackedGraphRewrite` objects are held in `tinygrad.ops.tracked_ctxs` until being pickled at the end of the program. `TrackedGraphRewrite` holds references to UOps, including `Ops.BUFFER` UOps, therefore preventing their GC which could otherwise occur. 

Because some `Ops.BUFFER` UOps are kept alive only by `TrackedGraphRewrite`, it is therefore the case that corresponding `Buffer` objects referenced by `tinygrad.ops.buffers` will not be GC'ed and will maintain `lb_refcount=1`, preventing their recycling in memory planning.

One solution is: when `TrackedGraphRewrite` wants to capture a UOp, first `graph_rewrite` that UOp to change `Ops.BUFFER` UOps to something else (in a way that can easily be reversed). Then at the end of the program, revert all such rewrites, right before we pickle the captured UOps. This fixes the memory planning issue, at the cost of increased runtime due to more rewrites. 

Concerning usage of `CPU` without VIZ (relevant to the bounty) -- it seems that GC and memory planning are working fine in python. However, memory planning on `CPU` uses sub-buffers, and the model export logic exports the sub-buffers not knowing that they are dependent on a master buffer, such that the exported clang code does not benefit from the memory planning. This problem could be fixed by another PR.

Next steps are to write tests and improve solution implementation.


hooved:
The new tests check that tracking rewrites of a UOp (whose op is `Ops.Buffer`) does not prevent GC of a buffer instantiated by that UOp, nor does it prevent that buffer's `lb_refcount` from resetting to 0 (as a proxy for GC of the UOp).

Here are new tests failing in CI without fixes from this PR: [test failures](https://github.com/tinygrad/tinygrad/actions/runs/14420980343/job/40443500474)

The above tests pass with fixes from this PR: [test success](https://github.com/tinygrad/tinygrad/actions/runs/14451679294/job/40525659376)

geohot:
Similar to your other PR, this adds complexity. Trading complexity for speed is almost never worth it.

Why does this happen? Is there like a fix to fix this instead of a workaround?

Qazalin:
Could TRACK_MATCH_STATS=2 pickle in-place? Instead of appending objects, which creates a strong reference, append a serialized bytes stream, and de-serialize it later in PRINT_MATCH_STATS and VIZ server.

hooved:
That would work but I didn't favor that approach because it's solving the problem in a similar way (deleting UOps that block GC then reinstantiating them later), but seems to have more overhead because pickling one UOp at a time will duplicate processing/data of the many UOps that are shared between graphs, and is overkill because many/most UOps aren't blocking Buffer GC.

I don't see a significantly simpler fix, maybe it's possible by refactoring how UOps and Buffers are linked so that Buffer GC isn't gated by UOp GC, but that is also complex.

PR #9858, opened 2025-04-11
Use $CUDA_PATH/include for CUDA headers

### LINKED ISSUE #7038: Adding CUDA_HOME for conda env ?

xl0:
Closes #7038

### DIFF
--- a/tinygrad/runtime/support/compiler_cuda.py
+++ b/tinygrad/runtime/support/compiler_cuda.py
@@ -4,7 +4,7 @@ from tinygrad.helpers import to_char_p_p, colored, init_c_var, getenv
 import tinygrad.runtime.autogen.nvrtc as nvrtc
 from tinygrad.device import Compiler, CompileError
 
-PTX = getenv("PTX")  # this shouldn't be here, in fact, it shouldn't exist
+PTX, CUDA_PATH = getenv("PTX"), getenv("CUDA_PATH", "")  # PTX shouldn't be here, in fact, it shouldn't exist
 
 def _get_bytes(arg, get_str, get_sz, check) -> bytes:
   sz = init_c_var(ctypes.c_size_t(), lambda x: check(get_sz(arg, ctypes.byref(x))))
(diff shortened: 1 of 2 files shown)


geohot:
Is this the CUDA standard? Or do you expect users to know to set this?

xl0:
I added it to the doc.

It seems that CUDA_HOME might be more standard. It's mentioned a few times on the internet, but not in the cuda documentation, and it's not in nvcc strings. Do you want me to change it to CUDA_HOME/include?

tinygrad/runtime/support/compiler_cuda.py
```diff
@@ -40,7 +40,8 @@ def cuda_disassemble(lib, arch):
 
 class CUDACompiler(Compiler):
   def __init__(self, arch:str, cache_key:str="cuda"):
-    self.arch, self.compile_options = arch, [f'--gpu-architecture={arch}', "-I/usr/local/cuda/include", "-I/usr/include", "-I/opt/cuda/include/"]
+    self.arch, self.compile_options = arch, [f'--gpu-architecture={arch}']
+    self.compile_options += [f"-I{CUDA_INC}"] if CUDA_INC else ["-I/usr/local/cuda/include", "-I/usr/include", "-I/opt/cuda/include"]
```
nimlgen, line 44: `CUDA_HOME`, otherwise not sure this pr actually fixes the issue.

how does the error message look now? better to add a note about the env var if includes aren't found.

is there any way to use something from `PATH` to get include paths?

xl0: @nimlgen It looks like CUDA_PATH is actually somewhat more standard, but only on Windows, because that's what the sample Nvidia build scripts used, and people stuck with it. On Linux, it seems that CUDA_HOME is more standard, used by PyTorch and maybe TensorFlow.

From `PyTorch/setup.c`:
```
# Environment variables we respect (these environment variables are
# conventional and are often understood/set by other software.)
#
#   CUDA_HOME (Linux/OS X)
#   CUDA_PATH (Windows)
#     specify where CUDA is installed; usually /usr/local/cuda or
#     /usr/local/cuda-x.y
```

I have documented the current `CUDA_INC` in the `docs/env_vars.md`, and I'll change it to whatever @geohot says because there is no right answer and I dont care what it's called as long as it;s documented. I don't think we should differentiate based on platform, but he's the boss.

Regarding the error message, nothing changes. Nvrtc will return an error saying it can't find the headers:
```
E       tinygrad.device.CompileError: Nvrtc Error 6, NVRTC_ERROR_COMPILATION
E       <null>(3): catastrophic error: cannot open source file "cuda_fp16.h"
```
and we get a backtrace.

While we could have checked for the headers in the paths passed to nvrtc, this is not consistent with how TinyGrad handles errors in general, so I'll stick wth its style unless @geohot says otherwise.

> is there any way to use something from PATH to get include paths?

The best I found is to run `nvcc -v` and grep its default include path from there. Which is not the way, because TinyGrad does not depend on nvcc, and it's too messy anyway.

xl0: nvrtc documentation uses CUDA_PATH in its examples, but it's in no way interpreted by nvrtc, just the example build scrips pass it to `-I` and `-L`:

https://docs.nvidia.com/cuda/nvrtc/#saxpy-build-instructions

xl0:
I've changed it to CUDA_PATH/include. While not "standard", it seems to be the most popular way to set the cuda location in other software and some Nvidia example code.

PR #8631, opened 2025-01-15
FP8 support on NVIDIA

pkotzbach:
without tensorcores, for simple_matmul.py i get
BFLOAT16:
`*** CUDA      21 r_128_64_8_16_1024_4_4_4                  arg  3 mem  0.13 GB tm     19.01ms/   117.17ms (  7229.39 GFLOPS    5.3|3616.5  GB/s) ['matmul']`

FP8E4M3:
`*** CUDA      24 r_128_64_8_16_1024_4_4_4                  arg  3 mem  0.07 GB tm     18.98ms/    98.23ms (  7241.72 GFLOPS    2.7|1811.3  GB/s) ['matmul']`
but without saturation (values are cut to highest possible value, no infinity)

with tencorcores:
BFLOAT16:
`*** CUDA      14 r_42_63_2_2_2_2_2_126_2_2_2_2_2_2_3_4     arg  3 mem  0.03 GB tm    364.22us/     3.15ms ( 44991.77 GFLOPS   67.0|2365.6  GB/s) ['matmul']`

FP8E4M3:
`*** CUDA      23 r_42_63_2_2_2_2_2_63_2_2_2_2_2_2_2_3_4    arg  3 mem  0.02 GB tm    306.53us/     4.71ms ( 53460.30 GFLOPS   39.8|1405.5  GB/s) ['matmul']`

continuation to #8490 (as i used wrong branch there)

PR #9861, opened 2025-04-12
FP8 support on NVIDIA

pkotzbach:
fixes to: https://github.com/tinygrad/tinygrad/pull/8631
sorry for breaking mypy

fixed mypy and tests and added hotfixes
gemm is now tested on CI https://github.com/tinygrad/tinygrad/actions/runs/14419075402/job/40439375479?pr=9861
one thing that is hard to get past is removal of numpy dependency in python device, as native python doesnt support fp8s


chenyuxyz:
let's split this into smaller prs, easier to review and make sure everything is correct and tested

we can add f8s to dtype and making sure the type promotion stuff are correct. then adding support alu device. finally adding support for TC

for python backend you should be able to implement truncate in python and having the numbers in fp8 format in python number (similar to truncate_bf16)

PR #9874, opened 2025-04-13
improve memory planning in JIT

hooved:
Unallocated buffers with `lb_refcount` > 0 are not cloned during JIT capture for good reasons. However, some buffers that have `lb_refcount` > 0 during jit capture ultimately have `lb_refcount` = 0 after jit capture is complete, yet have become allocated, and therefore are excluded from memory planning in the jit.

This PR aims to enable additional memory planning with such buffers. I don't yet understand the cases where we may not want to deallocate buffers with `lb_refcount` = 0 after jit capture.

To illustrate: on commit `513071bd923d143180bb988153a3f0371b0ac7c8`, which uses master branch `jit.py` code, here are two demos:

[demo.py](https://github.com/tinygrad/tinygrad/blob/513071bd923d143180bb988153a3f0371b0ac7c8/demo.py): JIT memory consumption is reduced by 97%
`python demo.py`
```
BEFORE new memory planning in JIT
  -- 100 buffers are used by jit
  -- 2000 bytes are used by jit buffers

AFTER new memory planning in JIT
  -- 3 buffers are used by jit
  -- 60 bytes are used by jit buffers
```

[examples/tinychat/tinychat-browser/compile.py](https://github.com/tinygrad/tinygrad/blob/513071bd923d143180bb988153a3f0371b0ac7c8/examples/tinychat/tinychat-browser/compile.py): a more realistic demo, which shows slight memory savings for the tinychat llama
`PYTHONPATH=. python examples/tinychat/tinychat-browser/compile.py`

CPU model:
```
BEFORE new memory planning in JIT
  -- 348 buffers are used by jit
  -- 1309129748 bytes are used by jit buffers

AFTER new memory planning in JIT
  -- 283 buffers are used by jit
  -- 1308796948 bytes are used by jit buffers
```

WEBGPU model:
```
BEFORE new memory planning in JIT
  -- 365 buffers are used by jit
  -- 1310504440 bytes are used by jit buffers

AFTER new memory planning in JIT
  -- 303 buffers are used by jit
  -- 1309676024 bytes are used by jit buffers
```


hooved:
The new variable `real_at_first_capture_bufs: set[Buffer]` tracks which buffers are allocated when first seen during jit capture.

Using `real_at_first_capture_bufs`, we can recycle more buffers in memory planning than we could before, and the memory planner has simpler logic.

Also, `real_at_first_capture_bufs` is a more comprehensive representation of "model state" than our current definition used in model export (buffers that are written to before being read from), therefore is useful in [#9424: refactor WebGPU export](https://github.com/tinygrad/tinygrad/pull/9424). For example, llama kv caches are written to before they are read from, but they represent state of the model whose export could be useful - and they are included in `real_at_first_capture_bufs`.

A test was added for whether redundant allocated buffers are recycled after jit capture, which fails on master: [failing test](https://github.com/tinygrad/tinygrad/actions/runs/14536730712/job/40786493872), but passes with this diff: [passing_test](https://github.com/tinygrad/tinygrad/actions/runs/14536792172/job/40786669063).

tinygrad/engine/jit.py
```diff
@@ -227,6 +228,8 @@ def __init__(self, fxn:Optional[Callable[..., ReturnType]], captured:Optional[Ca
     self.optimize = optimize
 
   def add_buffer(self, b:Buffer) -> Buffer:
+    if b not in self._seen_bufs and b.is_allocated(): self._real_at_first_capture_bufs.add(b)
```
geohot, line 231: Is this always meaningful, or are you just getting lucky with how it happens to load the model?

hooved: This is based on the assumption that the buffers containing data that must persist between separate calls to the jit are all allocated by the end of the first call to the jit, with no new such allocations occurring during the second call.

hooved:
The memory savings so far is not large in real models (although have only looked at llama), so value is not proven there.

The stateful buffer definition may work in the current version of tinygrad, but may become incorrect if memory patterns change. Will close this PR, to try instead determining stateful buffers based on those having bytes read before written to, which seems like a more pure approach but hasn't been done previously (in master we look only with buffer resolution, not with byte resolution, at what's read before written to).

PR #9875, opened 2025-04-13
Use Buffer in cloud server instead of opaques

uuuvn:
Not-quite-required but makes cloud graph a *lot* cleaner because unlike raw compiled programs `GraphRunner` takes `Buffer`s like other runners.

Otherwise either of: adding a new option to not free on `__del__`, (ab)using `external_ptr` to prevent free, or making something like a `FakeBuffer` is required.

### DIFF
--- a/tinygrad/runtime/ops_cloud.py
+++ b/tinygrad/runtime/ops_cloud.py
@@ -13,7 +13,7 @@ from http.server import HTTPServer, BaseHTTPRequestHandler
 from tinygrad.renderer import Renderer
 from tinygrad.dtype import dtypes
 from tinygrad.helpers import getenv, DEBUG, fromimport, unwrap, Timing
-from tinygrad.device import Compiled, Allocator, Compiler, Device, BufferSpec
+from tinygrad.device import Compiled, Buffer, Allocator, Compiler, Device, BufferSpec
 
 # ***** API *****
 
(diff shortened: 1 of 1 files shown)


geohot:
Hmm, why wasn't it done like this in the first place? Is there any downside to this?

tinygrad/runtime/ops_cloud.py
```diff
@@ -99,21 +98,16 @@ def _do(self, method):
...
-            session.buffers[c.buffer_num] = (Device[CloudHandler.device].allocator.alloc(c.size, c.options), c.size, c.options)
-          case BufferFree():
-            buf,sz,buffer_options = session.buffers[c.buffer_num]
-            Device[CloudHandler.device].allocator.free(buf,sz,buffer_options)
-            del session.buffers[c.buffer_num]
-          case CopyIn(): Device[CloudHandler.device].allocator._copyin(session.buffers[c.buffer_num][0], memoryview(bytearray(req._h[c.datahash])))
-          case CopyOut():
-            buf,sz,_ = session.buffers[c.buffer_num]
-            Device[CloudHandler.device].allocator._copyout(memoryview(ret:=bytearray(sz)), buf)
+            session.buffers[c.buffer_num] = Buffer(CloudHandler.device, c.size, dtypes.uint8, options=c.options, preallocate=True)
```
geohot, line 101: Does this work with images always having dtype uint8?

uuuvn:
> Hmm, why wasn't it done like this in the first place? (I actually don't know, I agree this looks better) Is there any downside to this?

None that i thought of

tinygrad/runtime/ops_cloud.py
uuuvn: Doesn't seem to break anything that isn't already broken, see #9903

geohot: Why not pass the dtype in?

uuuvn: If it's not image i don't know the dtype (`BufferAlloc` is sent from the `def _alloc(self, size:int, options:BufferSpec)` and `BufferSpec` only has a `ImageDType` if it's an image), and i don't see how having a special case for image for no reason is a good idea

geohot:
Btw, I now understand why it wasn't done this way to begin with. It's the dtype thing.

PR #9876, opened 2025-04-13
Cloud graph

uuuvn:
bert step:
cloud master - 1880.25 ms
cloud this - 1232.38 ms
master - 1094.61 ms

llama:
cloud master - 4.55 tok/s
cloud this - 9.20 tok/s
master - 17.63 tok/s

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -66,3 +66,6 @@ def eval_uop(uop:UOp, inputs:list[tuple[DType, list[Any]]]|None=None):
 def not_support_multi_device():
   # REMOTE doesn't support multi device anywhere, GPU, CUDA and METAL don't support multi device if in CI
   return Device.DEFAULT == "REMOTE" or (CI and Device.DEFAULT in ("GPU", "CUDA", "METAL"))
+
+# NOTE: This will open REMOTE if it's the default device
+REAL_DEV = (Device.DEFAULT if Device.DEFAULT != "REMOTE" else Device['REMOTE'].properties['remotedev'])
--- a/test/test_image_dtype.py
+++ b/test/test_image_dtype.py
@@ -5,9 +5,9 @@ from tinygrad.device import LRUAllocator, is_dtype_supported
 from tinygrad.dtype import ImageDType
 from tinygrad.engine.realize import lower_schedule
 from tinygrad.helpers import prod, unwrap
+from test.helpers import REAL_DEV
 
 IMAGE_SUPPORTED_DEVICES = ("QCOM", "GPU")
-REAL_DEV = (Device.DEFAULT if Device.DEFAULT != "REMOTE" else Device['REMOTE'].properties['remotedev'])
 
 @unittest.skipUnless(REAL_DEV in IMAGE_SUPPORTED_DEVICES, "Images not supported")
 class TestImageCopy(unittest.TestCase):
--- a/test/test_jit.py
+++ b/test/test_jit.py
@@ -3,7 +3,7 @@ import unittest, functools
 import numpy as np
 
 from hypothesis import given, settings, strategies as strat
-from test.helpers import assert_jit_cache_len, not_support_multi_device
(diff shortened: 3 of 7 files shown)


tinygrad/runtime/ops_cloud.py
```diff
@@ -42,11 +45,44 @@ class ProgramExec(CloudRequest):
...
+@dataclass(frozen=True)
+class GraphExec(CloudRequest):
+  graph_num: int
+  bufs: tuple[int, ...]
+  var_vals: dict[Variable, int]
+  wait: bool
+
 # for safe deserialization
-whitelist = {x.__name__:x for x in [BufferAlloc, BufferFree, CopyIn, CopyOut, ProgramAlloc, ProgramFree, ProgramExec, BufferSpec]}
+eval_globals = {x.__name__:x for x in [BufferAlloc, BufferFree, CopyIn, CopyOut, ProgramAlloc, ProgramFree, ProgramExec, GraphComputeItem,
```
geohot, line 76: Why is dtypes here?

tinygrad/runtime/ops_cloud.py
```diff
@@ -111,9 +148,24 @@ def _do(self, method):
...
             if r is not None: ret = str(r).encode()
+          case GraphAlloc():
+            graph_cls: type[GraphRunner] = unwrap(Device[CloudHandler.device].graph)
+            def _parse_ji(gi: GraphComputeItem):
+              prg = session.programs[(gi.name, gi.datahash)]
+              ps = ProgramSpec(gi.name, '', CloudHandler.device, UOp(Ops.NOOP), vars=list(gi.vars),
+                               global_size=list(cast(tuple[int], gi.global_size)) if gi.global_size is not None else None,
+                               local_size=list(cast(tuple[int], gi.local_size)) if gi.local_size is not None else None)
+              return ExecItem(CompiledRunner(ps, precompiled=b'', prg=prg), [session.buffers[buf] for buf in gi.bufs])
+            session.graphs[c.graph_num] = graph_cls(list(map(_parse_ji, c.jit_cache)), [session.buffers[buf] for buf in c.bufs], c.var_vals)
```
geohot, line 159: What if this number is already taken?

tinygrad/runtime/ops_cloud.py
```diff
@@ -111,9 +148,24 @@ def _do(self, method):
...
+          case GraphFree(): del session.graphs[c.graph_num]
+          case GraphExec():
+            r = session.graphs[c.graph_num]([session.buffers[buf] for buf in c.bufs], c.var_vals, wait=c.wait)
+            if r is not None: ret = str(r).encode()
     elif self.path == "/renderer" and method == "GET":
       cls, args = Device[CloudHandler.device].renderer.__reduce__()
       ret = json.dumps((cls.__module__, cls.__name__, args)).encode()
+    elif self.path == "/capabilities" and method == "GET":
```
geohot, line 167: This should probably be merged with renderer

geohot:
Why is the graph speed still so much slower than master?

tinygrad/runtime/ops_cloud.py
uuuvn: The old pattern was just hardcoded any get attribute with imageh or imagef returns dtype.imageh / dtypes.imagef. Eg BufferAlloc.imageh will parse as dtype.imageh too, this is very hacky and isn't even the simplest way to do this, i split whitelists into global whitelist and attribute whitelist that says what attributes of those globals can be accessed. The UOp and Ops are required for var_vals and symbolic global_size/local_size, the dtype is required because i fixed the hack where it didn't mattered what object is attribute being accessed on

tinygrad/runtime/ops_cloud.py
uuuvn: Then it will be just free the old graph and make the new graph be at it's number, the exact same behavior as `ProgramAlloc`

PR #9878, opened 2025-04-14
Add support for OpenAI Whisper v3 models

fujimotos:
OpenAI has released `large-v3` model in Nov 2023, which shows
10%-20% reduction of errors compared to `large-v2`.

https://github.com/openai/whisper/discussions/1762

They also released a `large-v3-turbo` model, which is claimed to
be as fast as `base`, while being as accurate as `large-v2`.

https://github.com/openai/whisper/discussions/2363

This patch adds support for them to tinygrad.


fujimotos:
_Note:_ I can confirm that this patch allows `whisper.py` to transcribe audio
files successfully using `v3` models:

```console
$ # Testing with large-v3
$ python3 examples/whisper.py jfk-5s.wav
ram used:  3.09 GB, decoder.mask                                      : 100%|1260/1260 [00:09<00:00, 131.02it/s]
loaded weights in 9620.15 ms, 3.09 GB loaded at 0.32 GB/s
And so my fellow Americans, ask not 

$ # Testing with large-v3-turbo
$ python3 examples/whisper.py jfk-5s.wav
ram used:  1.62 GB, decoder.mask                                      : 100%|588/588 [00:06<00:00, 97.35it/s]
loaded weights in 6041.25 ms, 1.62 GB loaded at 0.27 GB/s
And so, my fellow Americans, ask not to
```

(Tested with Python 3.12 on Ubuntu 24.04)

geohot:
Those tests don't repro using the models, can you add a simple way for me to test this?

fujimotos:
> Can you add a simple way for me to test this?

@geohot I just pushed de158677d that adds a feature to switch
between dfferent Whisper models.

Here is the minimal steps to try out this patch:

```console
$ # Get a short example audio file
$ wget https://upload.wikimedia.org/wikipedia/commons/b/bb/Neil_Armstrong_small_step.wav

$ # Transcribing with 'large-v3'
$ export WHISPER_MODEL=large-v3
$ python3 examples/whisper.py Neil_Armstrong_small_step.wav

$ # Transcribing with 'large-v3-turbo'
$ export WHISPER_MODEL=large-v3-turbo
$ python3 examples/whisper.py Neil_Armstrong_small_step.wav
```

You should obtain transcriptions like this:

```
I'm gonna step off the limb now. That's one small step for man, one giant leap for mankind.
```

Let me know if you find anything unclear.

examples/whisper.py
```diff
@@ -297,7 +306,10 @@ def listener(q):
   print("done listening")
 
 if __name__ == "__main__":
-  model, enc = init_whisper("small.en" if getenv("SMALL") else "tiny.en", batch_size=1)
+  # Set WHISPER_MODEL to change the inference model.
+  # SMALL is for compat with tinygrad<=0.10.2.
+  model_name = getenv("WHISPER_MODEL", default="small.en" if getenv("SMALL") else "tiny.en")
```
geohot, line 311: Why is this an env var?

fujimotos: The existing code has been using an environment variable (`SMALL`)
to switch between models. (It only supported `small.en` and `tiny.en`
for unknown reasons, though).

My intention here is to follow that convention.

fujimotos: _Addendum:_ I can make it a command-line argument if you prefer.
In particular, I'm thinking something like this:

* https://github.com/tinygrad/tinygrad/blob/master/examples/gpt2.py#L204-L214
* https://github.com/tinygrad/tinygrad/blob/master/examples/llama.py#L333-L344

PR #9885, opened 2025-04-14
rewrite the linearizer

geohot:


Qazalin:
`AMD_LLVM=1 AMD=1 python test/test_ops.py TestOps.test_padded_conv3d` looks like a remu bug, can't repro it on tiny13.

PR #9895, opened 2025-04-15
Allow TYPED=1

niniack:
This is a work in progress that addresses #7889! I will be updating my notes below, but I wanted to document things live in case someone has thoughts. Some things I noted and would like to get feedback on:

Main: 
-  numpy types are currently handled with annotations, which I assume is done to avoid importing numpy. Runtime type checking doesn't play well with this. Potential solution: we use annotations when not type checking, otherwise use [numpy's typing module](https://numpy.org/devdocs/reference/typing.html). I think I have a gap in my understanding and may be doing something silly here.

- Some tests cannot pass with runtime type checking because they are _designed_ to have an error. (I think!) this is is a thorn for runtime type checking because we should be passing in `TypeCheckError` for the expected error, but we need to import the typeguard library for that. This was alluded to in #6520.
e.g.
```
self.helper_test_exception(None, lambda x,y: x.div(y, rounding_mode="typo"), lambda x,y: x.div(y, rounding_mode="typo"), forward_only=True, vals=[[5], [0]], expected=RuntimeError)
```
Could a potential solution be to somehow [xfail](https://docs.pytest.org/en/7.1.x/how-to/skipping.html) them when TYPED=1?

- The tests did not seem to like `ctypes._CData`. I cannot seem to import that class either, although the documentation shows that it definitely exists. I started replacing them with `_SimpleCData` and `_Pointer` and potentially others, but I'm not sold on this approach yet. I have yet to do more digging around here.

Minor: 
- Looks like the codebase is split between using `Union` and `|`; `| None` and `Optional[]`. I'm assuming there is a preference towards the modern version of doing things and maybe moving away from `typing.List` to `list`, etc, since the project expects >3.10? If there is a consensus, I can zap those as I come across them! As of now, I follow what neighboring functions are doing.

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 from dataclasses import dataclass, replace
 from collections import defaultdict
-from typing import Optional, Any, Iterator, Generator
+from typing import Optional, Any, Iterator, Generator, Optional
 import multiprocessing, importlib, inspect, functools, pathlib, os, ctypes, ctypes.util, platform, contextlib, sys, re, atexit, pickle, decimal, time
 from tinygrad.helpers import CI, OSX, LRU, getenv, diskcache_get, diskcache_put, DEBUG, GlobalCounters, flat_mv, from_mv, PROFILE, temp, mv_address, \
                              cpu_time_execution, colored, Context, round_up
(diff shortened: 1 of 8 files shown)


kayo09:
Exciting progressed and I love the way you're describing the problem at hand! Makes my understanding of tinygrad a lot clearer as well

chenyuxyz:
stale

PR #9785, opened 2025-04-08
Add a HuggingFace ONNX models to CI

geohotstan:
Changed the structure a bit to make it easier maintenance around CI
dry_run is `collect_metadata.py`
`download_models.py` downloads the models and loads a lot of the information to determine what to add to CI.
we then just run `run.py` with selected models.

Also added a little `--diversity` flag 🏳️‍🌈 to `download_models.py` This selects models with most uncovered ops in the current set of ops in `diverse_models`. Pretty much picking the top few maximizes op coverage.

I only added a few models (hand picked from `diverse_models`) into CI for now.
`FacebookAI/xlm-roberta-large/onnx/model.onnx` `pszemraj/long-t5-tglobal-base-sci-simplify/model.onnx` `patrickjohncyh/fashion-clip/onnx/model.onnx` `mixedbread-ai/mxbai-rerank-base-v1/onnx/model.onnx`

They are ran by `run.py` completely independently, meaning no need to pre-download. `run.py` is the only runner now. `collect_metadata.py` and `download_models.py` just helps inform what to run with `run.py`

Can add more based on usage popularity or manually change to a new set after awhile.

PR #9898, opened 2025-04-15
Revert 9720 and remove symbolic dim check for ONNX

geohotstan:
I misunderstood "A dimension variable “N” represents the same value across the entire graph in a model." to mean outputs too.

model from `HuggingFaceTB/SmolLM-135M/onnx/model.onnx` from #9785
has symbolic dim for output but has a different value than the symbolic value specified at input. (result is validated to ORT to be correct)

It appears that ORT does not check symbolic dims for both input and output.
ORT checks at runtime:
 - explicit static input shapes dims
 - input dtype
 - output dtype

after this revert and removing the symbolic dim check, we are currently only checking `static input shapes`

ORT has options ilke [add_free_dimension_override_by_denotation](https://onnxruntime.ai/docs/api/python/api_summary.html#onnxruntime.SessionOptions.add_free_dimension_override_by_denotation) and [tools](https://onnxruntime.ai/docs/tutorials/mobile/helpers/make-dynamic-shape-fixed.html) to make symbolic shapes static.

I think they just leave validation out in case people want to tamper with the symbolic shapes. I think it is fine if we don't check this? And trust the user to identify the shapes and input the intended dim size for symbolic dimensions.

There are also some other peculiarities I found from ORT while running random tests for this. 
Was pretty puzzled on how to resolve this. Decided to revert.
I think moving on we try to do little hand written validation. Just let the model run with flexibility. It's a little hard to know which validation is absolutely expected and which isn't due to the lack of documentation and the ORT weirdness.

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -1,7 +1,7 @@
-from typing import Any, cast, Literal, Callable
+from typing import Any, Sequence, cast, Literal, Callable
 import dataclasses, functools, io, math, types
 from tinygrad.tensor import Tensor, _broadcast_shape, ReductionStr
-from tinygrad.helpers import getenv, DEBUG, prod, flatten, make_tuple, argsort
+from tinygrad.helpers import getenv, DEBUG, all_same, prod, flatten, make_tuple, argsort
 from tinygrad.dtype import DType, ConstType, dtypes, ImageDType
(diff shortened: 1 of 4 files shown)


geohotstan:
wait maybe I should have a PR only reverting...

PR #9899, opened 2025-04-15
CPUGraph working on CPU

larrythwu:
Label all other functions in CPUGraph as static except batched(), so that when we create a function pointer to the starting memory offset of the compiled object, it correctly points to batched().

### DIFF
--- a/tinygrad/runtime/graph/cpu.py
+++ b/tinygrad/runtime/graph/cpu.py
@@ -31,7 +31,8 @@ class CPUGraph(GraphRunner):
     batched.append("}")
 
     prep = [device.renderer._render(cast(CompiledRunner, ji.prg).p.uops) for i,ji in enumerate(jit_cache)]
-    funcs = dedup(device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops, ["static "]) for i, ji in enumerate(jit_cache))
+    funcs = dedup(device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops, ["static "])
+                  for i, ji in enumerate(jit_cache))
 
     defines = dedup(itertools.chain.from_iterable(device.renderer._render_defines(cast(CompiledRunner, ji.prg).p.uops) for ji in jit_cache))
     entry = device.renderer._render_entry("batched", targs)
--- a/tinygrad/runtime/graph/cpu.py
+++ b/tinygrad/runtime/graph/cpu.py
@@ -31,7 +31,10 @@ class CPUGraph(GraphRunner):
     batched.append("}")
 
     prep = [device.renderer._render(cast(CompiledRunner, ji.prg).p.uops) for i,ji in enumerate(jit_cache)]
-    funcs = dedup(device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops) for i,ji in enumerate(jit_cache))
+    funcs = dedup(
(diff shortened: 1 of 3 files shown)


tinygrad/runtime/graph/cpu.py
```diff
@@ -31,7 +31,10 @@ def render_arg(buf):
     batched.append("}")
 
     prep = [device.renderer._render(cast(CompiledRunner, ji.prg).p.uops) for i,ji in enumerate(jit_cache)]
-    funcs = dedup(device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops) for i,ji in enumerate(jit_cache))
+    funcs = dedup(
+        device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops).replace('void', 'static void')
```
geohot, line 35: No

larrythwu: Using _render_body's option pref=["static "] now.

geohot:
Hmm, I missed this one. Who deserves credit? See #10014

larrythwu:
> Hmm, I missed this one. Who deserves credit? See #10014
> 
> @quortus can you two figure this out?

I think I do, as far as this particular commit for credit, I am willing to share the bounty if @quortus continues to work on LLVMGraph. My reason is: first of all, this is an earlier PR and #10014 is the exact same solution. Second, I have been working on this for past few weeks which eventually led me to this clean solution as evidence in my previous closed PR https://github.com/tinygrad/tinygrad/pull/9795 where @uuuvn commented on and inspired the current solution.

quortus:
@geohot  I have no idea how this kind of situation should be resolved.

1) I knew about this solution back from PRs for "Replace CLANG dlopen with mmap + remove linker step" bounty. It was discussed many times.
2) The only non-trivial part of this PR is "static" usage. Graph init is copypasted from other places, tests are straightforward reactionary changes.
3) That non-trivial part was suggested by @uuuvn so if we are following this line of reasoning it should be his who claim bounty (at least partially).
4) I have evidence that this change is not enough from my work on LLVM graph (#10029). See #10037 for details.

tinygrad/runtime/graph/cpu.py
```diff
@@ -31,7 +31,8 @@ def render_arg(buf):
     batched.append("}")
 
     prep = [device.renderer._render(cast(CompiledRunner, ji.prg).p.uops) for i,ji in enumerate(jit_cache)]
-    funcs = dedup(device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops) for i,ji in enumerate(jit_cache))
+    funcs = dedup(device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops, ["static "])
```
quortus, line 34: Space before last quote is not necessary.

geohot:
Sounds like @uuuvn solved this...

uuuvn:
> 3. That non-trivial part was suggested by @uuuvn so if we are following this line of reasoning it should be his who claim bounty (at least partially).

Which if i recall correctly was suggested to me by @nimlgen back in the first version of clang jit where i was doing the stupid "`pickle(tuple[shellcode, dict[symbol name, offset]])` because compiler doesn't know what symbol is entry" thing

larrythwu:
@geohot  Just my two cents — I believe @uuuvn and I should at least be mentioned in the commit message (https://github.com/tinygrad/tinygrad/commit/9e49721c47801c3fc07f36c008a72b40ab46fc5e), since we independently arrived at the same solution. That said, full credit to @quortus for taking on the heavier lifting at this stage — the bounty rightly goes to him.

PR #9900, opened 2025-04-15
FP8s part 2

pkotzbach:
CUDA device and ALUs

### DIFF
--- a/extra/gemm/simple_matmul.py
+++ b/extra/gemm/simple_matmul.py
@@ -5,7 +5,8 @@ from tinygrad.dtype import _to_np_dtype
 from tinygrad.codegen.kernel import OptOps
 from tinygrad.engine.realize import lower_schedule
 
-dtype_in = dtypes.half if getenv("HALF") else dtypes.bfloat16 if getenv("BFLOAT16") else dtypes.float
+dtype_in = (dtypes.half if getenv("HALF") else dtypes.bfloat16 if getenv("BFLOAT16") else
+            dtypes.fp8e4m3 if getenv("FP8E4M3") else dtypes.fp8e5m2 if getenv("FP8E5M2") else dtypes.float)
 acc_dtype = dtypes.half if getenv("ACC_HALF") else dtypes.bfloat16 if getenv("ACC_BFLOAT16") else None
 if getenv("INT"):  dtype_in, acc_dtype = dtypes.int8, dtypes.int32
 if getenv("UINT"): dtype_in, acc_dtype = dtypes.uint8, dtypes.int32
@@ -14,7 +15,7 @@ N = getenv("N", 4096)
 M = getenv("M", N)
 K = getenv("K", N)
 CNT = getenv("CNT", 10)
-ATOL = getenv("ATOL", 1e-4)
+ATOL = getenv("ATOL", 1e-4 if dtype_in not in dtypes.fp8s else 1)
 RTOL = getenv("RTOL", 3e-2)
 INT_LOW = getenv("INT_LOW", 0)
 INT_HIGH = getenv("INT_HIGH", 10)
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -145,6 +145,7 @@ def _test_ops(a_dtype:DType, b_dtype:DType, target_dtype=None):
   _assert_eq(Tensor([1,1,1,1], dtype=a_dtype)+Tensor.ones((4,4), dtype=b_dtype), target_dtype, 2*Tensor.ones(4,4).numpy())
 
 class TestFp8s(unittest.TestCase):
+  @unittest.skipUnless(is_dtype_supported(dtypes.fp8e4m3), "fp8e4m3 not supported")
   def test_fp8e4m3_creation_numpy(self):
     data = [-1, 1, 2]
     t = Tensor(data, dtype=dtypes.fp8e4m3)
@@ -153,6 +154,7 @@ class TestFp8s(unittest.TestCase):
     assert tnp.dtype == np.float32
     np.testing.assert_allclose(tnp, np.array(data))
 
+  @unittest.skipUnless(is_dtype_supported(dtypes.fp8e5m2), "fp8e5m2 not supported")
   def test_fp8e5m2_creation_numpy(self):
     data = [-1, 1, 2]
     t = Tensor(data, dtype=dtypes.fp8e5m2)
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -39,7 +39,8 @@ def _assert_eq(tensor:Tensor, target_dtype:DType, target, tol_target_dtype:float
   if DEBUG >= 2: print(tensor.numpy())
   try:
     assert tensor.dtype == target_dtype
-    np.testing.assert_allclose(tensor.numpy(), target, rtol={dtypes.float16:1e-3, dtypes.bfloat16:1e-2}.get(target_dtype, tol_target_dtype))
+    rtols = {dtypes.float16:1e-3, dtypes.bfloat16:1e-2, dtypes.fp8e5m2: 1, dtypes.fp8e4m3: 1e-1}
+    np.testing.assert_allclose(tensor.numpy(), target, rtol=rtols.get(target_dtype, tol_target_dtype))
   except AssertionError as e:
     raise AssertionError(f"\ntensor {tensor.numpy()} dtype {tensor.dtype} does not match target {target} with dtype {target_dtype}") from e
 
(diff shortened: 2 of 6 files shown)


extra/gemm/simple_matmul.py
```diff
@@ -5,7 +5,8 @@
 from tinygrad.codegen.kernel import OptOps
 from tinygrad.engine.realize import lower_schedule
 
-dtype_in = dtypes.half if getenv("HALF") else dtypes.bfloat16 if getenv("BFLOAT16") else dtypes.float
+dtype_in = (dtypes.half if getenv("HALF") else dtypes.bfloat16 if getenv("BFLOAT16") else
```
chenyuxyz, line 8: add this to the tensor core change

test/test_dtype.py
```diff
@@ -143,8 +145,49 @@ def _test_ops(a_dtype:DType, b_dtype:DType, target_dtype=None):
   _assert_eq(Tensor([1,1,1,1], dtype=a_dtype)+Tensor.ones((4,4), dtype=b_dtype), target_dtype, 2*Tensor.ones(4,4).numpy())
 
 class TestFp8s(unittest.TestCase):
-  def test_fp8e4m3_creation(self): assert Tensor([-1, 1, 2], dtype=dtypes.fp8e4m3).dtype == dtypes.fp8e4m3
-  def test_fp8e5m2_creation(self): assert Tensor([-1, 1, 2], dtype=dtypes.fp8e5m2).dtype == dtypes.fp8e5m2
+  @unittest.skipUnless(is_dtype_supported(dtypes.fp8e4m3), "fp8e4m3 not supported")
+  def test_fp8e4m3_creation_numpy(self):
```
chenyuxyz, line 149: this and the one below are almost identical, can use for loop instead of copy pasting

test/test_dtype.py
```diff
@@ -703,6 +747,8 @@ def test_sum(self):
     assert (Tensor([0, 1], dtype=dtypes.uint64)).sum().dtype == dtypes.uint64
     assert (Tensor([0, 1], dtype=dtypes.float16)).sum().dtype == dtypes.float16
     #assert (Tensor([0, 1], dtype=dtypes.bfloat16)).sum().dtype == dtypes.bfloat16
+    if (is_dtype_supported(dtypes.fp8e4m3)): assert (Tensor([0, 1], dtype=dtypes.fp8e4m3)).sum().dtype == dtypes.fp8e4m3
```
chenyuxyz, line 750: dtype is known without compuatation, so don't need is_dtype_supported

tinygrad/device.py
```diff
@@ -332,9 +332,7 @@ def is_dtype_supported(dtype:DType, device:Optional[str]=None) -> bool:
   if dtype == dtypes.bfloat16:
     # NOTE: this requires bf16 buffer support
     return device in {"AMD"} or (device in {"CUDA", "NV"} and not CI and not getenv("PTX"))
-  if dtype in dtypes.fp8s:
-    # not supported yet - in progress
-    return False
+  if dtype in dtypes.fp8s: return (device in {"CUDA", "NV"} and not CI and not getenv("PTX"))
```
chenyuxyz, line 335: can you implement PYTHON backend? otherwise non of the alu tests run in CI

pkotzbach:
made python backend
@chenyuxyz

tinygrad/dtype.py
```diff
@@ -126,7 +126,7 @@ def fields() -> dict[str, DType]: return DTYPES_DICT
   uint32: Final[DType] = DType.new(6, 4, "unsigned int", 'I')
   int64: Final[DType] = DType.new(7, 8, "long", 'q')
   uint64: Final[DType] = DType.new(8, 8, "unsigned long", 'Q')
-  fp8e4m3: Final[DType] = DType.new(9, 1, "float8_e4m3", None)
+  fp8e4m3: Final[DType] = DType.new(9, 1, "float8_e4m3fn", None)
```
chenyuxyz, line 129: why this change and how about float8_e5m2?

test/test_dtype.py
```diff
@@ -709,7 +748,7 @@ def test_least_upper_float_input_is_float(self, input_dtype, default_float):
   @given(strat.sampled_from(dtype_ints), strat.sampled_from(dtype_floats))
   def test_least_upper_float_input_is_int(self, input_dtype, default_float):
     dtypes.default_float = default_float
-    self.assertEqual(least_upper_float(input_dtype), default_float)
+    self.assertEqual(least_upper_float(input_dtype), default_float if default_float not in dtypes.fp8s else dtypes.half)
```
chenyuxyz, line 751: why is it different?

tinygrad/runtime/ops_python.py
```diff
@@ -4,25 +4,25 @@
...
-from tinygrad.dtype import DType, dtypes, ImageDType, PtrDType, truncate
+from tinygrad.dtype import DType, dtypes, ImageDType, PtrDType, truncate, fp8_to_float, float_to_fp8
 from tinygrad.helpers import all_same, getenv, flatten, get_single_element
 from tinygrad.device import Compiled, Compiler, Allocator
 from tinygrad.ops import exec_alu, Ops, UOp, GroupOp
 from tinygrad.renderer import Renderer
 from tinygrad.renderer.cstyle import CUDARenderer, MetalRenderer, AMDRenderer, IntelRenderer, ClangRenderer
 
-def _load(m, i):
+def _load(m, i, dtype=None):
```
chenyuxyz, line 14: why do you need to change load and store? changing load implies that the stored value was wrong. what if it's a copy from PYTHON backend to CUDA?

chenyuxyz (commented):
don't think in terms of `if fp8 then else`, think in terms of least_dtype and the relation between fp8 and other dtypes. the goal is have the logic that works for all the dtypes

tinygrad/dtype.py
pkotzbach: because equivalents of these are in CUDA
https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/struct____nv__fp8__e4m3.html - doesnt support infinity (thats why fn)
https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/struct____nv__fp8__e5m2.html - does support infinity and fp8e5m2 with fn doesnt exists

test/test_dtype.py
pkotzbach: because least_upper_float(int64) is float16 and not float8
https://github.com/tinygrad/tinygrad/pull/9887#discussion_r2042626548

tinygrad/runtime/ops_python.py
pkotzbach: memoryview doesnt have default way to store and read fp8s
i store fp8s as "B" format but with fp8 binary (float_to_fp8 changes to fp8 binary and fp8_to_float restores from fp8 binary), so i can use these values as floats in native python for operations and store them as proper fp8s in memory

if i store fp8s as proper fp8s in memory i think copy from PYTHON backend to CUDA should be ok
i tried to create that and it looks fine for me, but maybe i am missing something
obraz

test/test_dtype.py
chenyuxyz: oh i see, yea that's fine

tinygrad/dtype.py
chenyuxyz: i see, does both support tensor cores? i think it's fine for now, not sure how or which one we really want

can you add a comment for this?

pkotzbach: yes these two are supported tensor cores

tinygrad/runtime/ops_python.py
chenyuxyz: @geohot thoughts on this?

pkotzbach: i've changed it so it is stored as int that is casted to fp8 when needed on operations

tinygrad/runtime/ops_python.py
```diff
@@ -24,6 +24,19 @@ def _store(m, i, v):
   if i < 0 or i >= len(m): raise IndexError(f"store out of bounds, size is {len(m)}, access is {i}, value is {v}")
   m[i] = v
 
+def with_fp8_handling(values, dtypes_from: list[DType]|DType|None, dtype_to: DType, operation_fn):
```
chenyuxyz, line 27: we discussed this in the meeting, you can check the recording when it's ready. but generally we don't want things like this. it's not scalable long term with more dtypes.

i don't have a solution now, will think about this later this week

chenyuxyz:
can you also open a draft pr that includes everything including a working tensor core?

i am testing support bfloat16 in PYTHON and see what we are missing now.

PR #9903, opened 2025-04-16
Image test fixes

uuuvn:
test_imageh_copyout_1x1 fails with `ValueError: memoryview: destination format must be a native single character format prefixed with an optional '@'`

test_lru_alloc fails on cloud only with `AssertionError: assert 49 == 48`

No infra for testing cloud in ci, `CLOUDDEV=GPU CLOUD=1 python -m pytest test/test_image_dtype.py` on localhost for now


test/test_image_dtype.py
```diff
@@ -1,17 +1,22 @@
...
+import sys, numpy as np
 from tinygrad import Device, dtypes, Tensor, Context
 from tinygrad.dtype import ImageDType
 from tinygrad.engine.realize import lower_schedule
-from tinygrad.helpers import prod, unwrap
+from tinygrad.helpers import getenv, prod, unwrap
 
-@unittest.skipIf(Device.DEFAULT not in ("QCOM", "GPU"), "only images on GPU")
+IMAGE_SUPPORTED_DEVICES = ("QCOM", "GPU")
+DEFAULT_SUPPORTS_IMAGE = (Device.DEFAULT if Device.DEFAULT != "CLOUD" else getenv("CLOUDDEV", "METAL")) in IMAGE_SUPPORTED_DEVICES
```
geohot, line 9: This getenv CLOUDDEV shouldn't be copied here, what if the default changes?

uuuvn: fixed

test/test_image_dtype.py
```diff
@@ -1,17 +1,22 @@
...
-import numpy as np
+import sys, numpy as np
 from tinygrad import Device, dtypes, Tensor, Context
 from tinygrad.dtype import ImageDType
 from tinygrad.engine.realize import lower_schedule
 from tinygrad.helpers import prod, unwrap
 
-@unittest.skipIf(Device.DEFAULT not in ("QCOM", "GPU"), "only images on GPU")
+IMAGE_SUPPORTED_DEVICES = ("QCOM", "GPU")
+REAL_DEV = (Device.DEFAULT if Device.DEFAULT != "CLOUD" else Device[Device.DEFAULT].properties['clouddev'])
```
geohot, line 9: Hmm, now this opens the default device. Is this okay? Do we have other things that do that?

If this is an issue, a way to refactor is to put CLOUDDEV in helpers

uuuvn: Not sure if that's okay, but i don't think that CLOUDDEV should even have a hardcoded default, it should just be whatever would've been opened if CLOUD=1 wasn't specified if it was specified and Device.DEFAULT otherwise. I also placed CLOUDDEV into ops_cloud.py instead of helpers because helpers don't (and probably shouldn't) import device.py

uuuvn: #9935

uuuvn: Looks like i have to open the CLOUD in tests after all, see https://github.com/tinygrad/tinygrad/pull/9935#discussion_r2051653999

tinygrad/runtime/ops_cloud.py
```diff
@@ -166,9 +165,7 @@ class CloudDevice(Compiled):
   def __init__(self, device:str):
     if (host:=getenv("HOST", "")) != "": self.host = host
     else:
-      p = multiprocessing.Process(target=cloud_server, args=(6667,))
-      p.daemon = True
-      p.start()
+      multiprocessing.Process(target=cloud_server, args=(6667,), name='MainProcess', daemon=True).start()
```
geohot, line 168: Why is a change like this (whitespace) mixed in here?

geohot:
This PR is doing two things, a refactor of CLOUDDEV and a new type of test. This should be two PRs.

uuuvn:
> This PR is doing two things, a refactor of CLOUDDEV and a new type of test. This should be two PRs.

This was a separate pr

uuuvn:
The /renderer => /properties thing isn't standalone, so not submitting it separately

PR #9912, opened 2025-04-16
Don't assume all devices with the same runtime have compatible binaries

Yexo:
`python examples\beautiful_mnist_multigpu.py` fails on my machine. After adding some logging in opg_gpu.py I got this output: `Error: The program ISA amdgcn-amd-amdhsa--gfx1103 is not compatible with the device ISA amdgcn-amd-amdhsa--gfx1102`

I'm running this under windows, my laptop has two GPUs: an integrated AMD Radeon(TM) 780M and also an AMD Radeon(TM) RX 7700S. Looks like the ISAs for both aren't compatible, so an OpenCL program compiled for one will not run on the other.

### DIFF
--- a/tinygrad/engine/realize.py
+++ b/tinygrad/engine/realize.py
@@ -105,8 +105,8 @@ def get_runner(device:str, ast:UOp) -> CompiledRunner:
   context = (BEAM.value, NOOPT.value, DEVECTORIZE.value)
   ckey = (device, ast.key, context, False)
   if cret:=method_cache.get(ckey): return cret
-  bkey = (device.split(":")[0], ast.key, context, True)
-  if bret:=method_cache.get(bkey):
+  bkey = ("%s_%s" % (device.split(":")[0], Device[device].compiler.cachekey), ast.key, context, True)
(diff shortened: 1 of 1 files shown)


geohot:
Needs tests

tinygrad/engine/realize.py
```diff
@@ -105,8 +105,8 @@ def get_runner(device:str, ast:UOp) -> CompiledRunner:
   context = (BEAM.value, NOOPT.value, DEVECTORIZE.value)
   ckey = (device, ast.key, context, False)
   if cret:=method_cache.get(ckey): return cret
-  bkey = (device.split(":")[0], ast.key, context, True)
-  if bret:=method_cache.get(bkey):
+  bkey = ("%s_%s" % (device.split(":")[0], Device[device].compiler.cachekey), ast.key, context, True)
+  if not getenv("DISABLE_COMPILER_CACHE") and (bret:=method_cache.get(bkey)):
```
geohot, line 109: Why does this have to be checked here? I think something probably has to be refactored

chenyuxyz:
stale

PR #9916, opened 2025-04-17
[Bounty] moved index_tensor off cpu in torch_backend

nishant42491:
moving some cpu bottlenecked functions in torch_backend off the cpu as part of bounty. I aim to achieve this by framing the operations to be performed natively as tinygrad tensors.


nishant42491:
If this pr is acceptable should i work on moving the other cpu bottlenecked function in torch_backend as well?

extra/torch_backend/backend.py
```diff
@@ -131,6 +127,10 @@ def upsample_backward(grad_out, output_size, input_size, *args, f=None): return
 
 # *** end bad functions on CPU ***
 
+@torch.library.impl("aten::index.Tensor", "privateuseone")
+def index_tensor(x, y):
+  return wrap(unwrap(x)[[unwrap(_y.to(x.device)) if _y is not None else slice(None) for _y in y]])
```
geohot, line 132: Are they ever on different devices? Also, if it's just this, can you use the autowrappers?

nishant42491: The beautiful_mnist_torch in torch_backend_extra tests fail precisely because the index tensors are on a different device than the input (x) tensors hence placing them on the same device became necessary.

nishant42491: Also what exactly do you mean by autowrapping in case the two tensors are on a different device?

PR #9918, opened 2025-04-17
Move randperm_generator off of cpu in torch_backend

currybab:
- add `randperm` function in tensor.py
- create tests for `randperm` for pytorch backend and tinygrad frontend


tinygrad/tensor.py
```diff
@@ -851,6 +851,23 @@ def kaiming_normal(*shape, a:float = 0.01, **kwargs) -> Tensor:
     std = math.sqrt(2.0 / (1 + a ** 2)) / math.sqrt(prod(argfix(*shape)[1:]))
     return Tensor.normal(*shape, mean=0.0, std=std, **kwargs)
 
+  @staticmethod
+  def shuffle(t: Tensor) -> Tensor:
+    n = int(t.shape[0])
+    t = t.contiguous()
+    for i in reversed(range(1, n)):
+      j = Tensor.randint(1, 1, low=0, high=i, device=t.device)[0,0]
+      t[i], t[j] = t[j].clone().realize(), t[i].clone().realize()
```
geohot, line 860: This is absurdly slow. You can't use realize or an O(n) loop here

currybab: Thanks for feedback. I'll find for alternative solution not using realize and O(n) loop.

currybab: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/Randperm.cu
I read this and checked tinygrad's sort uses bitonic sort, which is not a stable sort.

currybab:
To enable the random generator to work on the tiny backend, it seems necessary to register the generator for the new backend.
https://pytorch.org/tutorials/advanced/privateuseone.html#register-generator-for-the-new-backend

currybab:
To follow the behavior of pytorch's generator in tinygrad, it seems to separate random logic into another class.

PR #9919, opened 2025-04-17
Support unfold op

L1aoXingyu:


geohot:
1. Needs tests
2. Is there a simpler/shorter way to write this in terms of other functions?

L1aoXingyu:
> 1. Needs tests
> 2. Is there a simpler/shorter way to write this in terms of other functions?

i am trying to simplify the unfold op.

PR #9926, opened 2025-04-18
fix `ptx` linearizer bug [pr]

ignaciosica:
This is the smallest diff that fixes the issue with shared memory for ptx.
Not using `idx_context` and only relying on `store_context` works and passes pr too, but decided, as there is ongoing work into refactoring the linearizer, to stick to the smallest diff possible to fix the issue.

Temporarily added the fix to the LDS to showcase the fix ([Linux (ptx) job](https://github.com/tinygrad/tinygrad/actions/runs/14535535917/job/40783069161))


ignaciosica:
I saw this diff is still relevant after the linearizer refactor, I initially thought it was no longer going to be needed. @geohot does it broke anything for the ongoing effort of the refactor? If so I'll revert

PR #9929, opened 2025-04-18
[bounty] [pr] z3 index validation with support for symbolic

S-Lykles:
This is  #9774 but with support for variables in the kernel. This adds a rule to `grouper.py` to change the dtype size of the buffers being loaded/stored to be the maximum of the symbolic shape of the load/store.


tinygrad/engine/grouper.py
```diff
@@ -375,6 +375,9 @@ def check_load_st(glbl:UOp, view:UOp):
   (UPat(GroupOp.All-{Ops.DEFINE_GLOBAL}, name="x"), lambda x: x.replace(dtype=x.dtype.base) if isinstance(x.dtype, ImageDType) else None),
   # if this kernel also assigns to the loaded buffer, ensure we can index it correctly
   (UPat(Ops.LOAD, src=(UPat.var("glbl"), UPat.var("view"))), check_load_st),
+  # set the dtype size of the buffer to the maximum symbolic size
```
chenyuxyz, line 378: can you make a separate pr for this with a test?

chenyuxyz (commented):
overall looks good. did you find any other OOB error? also how fast is this?

tinygrad/engine/grouper.py
Qazalin: agreed, I don't see why this is needed.
DEFINE_GLOBAL's size should always match the size of the BUFFER,
If there's code creating VIEW that goes out of the BUFFER size, it's a bug.

S-Lykles: So the reason this is needed (when we do a index bound check) is that this:
```python
i = Variable('i', 1, 10)
ret = Tensor.ones(2).contiguous().reshape(i.bind(2)) + 1
ret.reshape(2).tolist()
```

Will create a graph like this 

image


This kernel is symbolic so it works with input buffers with size [1, 10], but the buffer dtype is is `dtypes.float.ptr(2)`, which corresponds to the bound shape of the tensor that was used to create the kernel.
This is then a problem when we do the oob check because that uses the size of the buffer dtype.



> If there's code creating VIEW that goes out of the BUFFER size, it's a bug.

Maybe the right abstraction is that a buffer can have a symbolic size? This was the smallest change I could make that would get the index validation working on. I think it is also consistent with how the size for output buffer is calculated in:

```python
  (UPat(Ops.CONTIGUOUS, name="x"), lambda ctx,x: create_kernel(ctx, x, UOp.new_buffer(x.device, x.size, x.dtype))),
  # create a buffer for COPY on the new device
  (UPat(Ops.COPY, src=(UPat(Ops.DEVICE, name="d"), UPat()), name="x"), lambda ctx,d,x: create_kernel(ctx, x, UOp.new_buffer(d.arg, x.size, x.dtype))),
  # otherwise check the context if we're realizing this UOp
  (UPat(GroupOp.All-DONT_PLACE_IN_KERNEL, name="x"),
   lambda ctx,x: create_kernel(ctx, x, UOp.new_buffer(x.device, x.size, x.dtype)) if x in ctx.realizes else None),
```

Since `x.size` uses the maximum size of the (unbound) shapetracker, not the bound size

Qazalin: The real Buffer in memory is a size 2. The code you're showing is accessing index=[0, 10) of a size 2 buffer. That's OOB.

S-Lykles: Mmh, well the kernel is accessing index=[0, i), and the actual value of i is determined from the size of the buffer being passed in.

Qazalin: Yea, but it's wrong.
How often are we using this pattern in tests? You can mark them as expected failure.
It shouldn't happen in the LLM implementations for symbolic shape.

Qazalin:
woah, 39 bad patterns in symbolic ops tests...

Qazalin:
pip install z3-solver fails on the comma device.
```
Failed to build z3-solver
ERROR: Failed to build installable wheels for some pyproject.toml based projects (z3-solver)
```
```
comma@tiny23:~$ uname -a
Linux tiny23 4.9.103+ #2 SMP PREEMPT Fri May 3 16:56:59 PDT 2024 aarch64 aarch64 aarch64 GNU/Linux
```
couldn't ssh into tinymac, the rest of the boxes should have z3-solver installed.

S-Lykles:
`test/models/test_whisper.py` does fail

Qazalin:
hmm, llama failed too.

Qazalin:
hmm, CHECK_OOB=1 won't work in process replay though.
You see why right? This CHECK_OOB=1 isn't upstreamed to master, process_replay compiles AST+OptOps on the **master** version of `tinygrad/*`. It only captures the program output string from your branch.

S-Lykles:
Right. Does BEAM search discard kernels that fail uop verification? The new index validation doesnt have the false positives (positive=OOB) that the old validate_index had. Im struggling to undestand why this is changing asts

S-Lykles:
Also yeah I was under the impression that process replay worked the other way around.

PR #9935, opened 2025-04-19
Don't hardcode default CLOUDDEV

uuuvn:


tinygrad/runtime/ops_cloud.py
```diff
@@ -15,6 +15,8 @@
 from tinygrad.helpers import getenv, DEBUG, fromimport, unwrap, Timing
 from tinygrad.device import Compiled, Buffer, Allocator, Compiler, Device, BufferSpec
 
+CLOUDDEV = getenv("CLOUDDEV", next(Device.get_available_devices()) if Device.DEFAULT == "CLOUD" else Device.DEFAULT)
```
nimlgen, line 18: this opens devices on the client??

uuuvn: Same with just getting Device\.DEFAULT when there is no env set, i don't think there is a way around it that would both not hardcode some preferred device like METAL and allow getting CLOUDDEV on client without opening a connection (eg for #9903)

nimlgen: why do you want to keep `CLOUDDEV` global? moving it back to its original place should solve that, no?

uuuvn: image
(from #9903)

nimlgen: I think this won't work for AM

uuuvn: yeah, missed that, moved into `cloud_server`, separate function won't help because in case of am it will just fail and return wrong device, i think i just have to open the cloud device in tests...

geohot:
Can we add tests that devices aren't opened on the client?

geohot:
Will merge when tests pass

PR #9937, opened 2025-04-19
FP8s truncate

pkotzbach:
in order to make python backend (without numpy) here https://github.com/tinygrad/tinygrad/pull/9900 i need to implement first FP8s-float conversions and truncate, i make new PR for that as it is pretty big
i put these conversions to file in extra folder, as it takes IMO too much lines to put it in dtype.py - although then I import with try..catch statement and i dont know if this is good approach

here are proofs that values in tests are correct:
obraz

obraz


chenyuxyz:
it should be in `tinygrad`, not `extra`. also i think comparing `ml_dtypes` output values in the tests is better to maintain

pkotzbach:
i put truncate functions in dtype.py and added ml_dtypes to tests

tinygrad/dtype.py
```diff
@@ -198,8 +198,73 @@ def truncate_bf16(x):
   bf = struct.unpack('f', struct.pack('I', f32_int & 0xFFFF0000))[0]
   return bf
 
+# fp8-float conversions based on https://gitlab.com/nvidia/headers/cuda-individual/cudart/-/blob/main/cuda_fp8.hpp
+def float_to_fp8(x: float, fp8_interpretation: str) -> int:
+  assert fp8_interpretation in ["E4M3", "E5M2"], "Invalid fp8 interpretation"
```
chenyuxyz, line 203: use `dtypes.fp8e4m3` instead of strs for keys

tinygrad/dtype.py
```diff
@@ -198,8 +198,73 @@ def truncate_bf16(x):
...
+  assert x not in [math.nan, -math.nan], "NaN is not supported"
```
chenyuxyz, line 204: fp8 supports nan right?

chenyuxyz:
overall i think the code can be more concise, but this is fine and should unblock you to implement PYTHON backend alus for fp8s

PR #9939, opened 2025-04-20
add double gemm fuse beam failure

geohotstan:
hits this assert in `kernel.py line:278`
```python
        # can only fuse reduces with the same tc options
        assert all_same(tensor_core_opts)
```
I changed that assert to a `if not all_same(tensor_core_opts): continue` just for the double gemm fuse in olmoe to run and it became real sloooow. 
So maybe it's my fault.
Dunno if it's necessarily a linearizer failure.

### DIFF
--- a/test/test_linearizer_failures.py
+++ b/test/test_linearizer_failures.py
@@ -1430,5 +1430,33 @@ class TestLinearizerFailures(unittest.TestCase):
     opts = [Opt(op=OptOps.GROUPTOP, axis=0, arg=29), Opt(op=OptOps.PADTO, axis=0, arg=32)]
     helper_test_lin(Kernel(ast, opts=Device[Device.DEFAULT].renderer), opts=opts, failed_platforms=["METAL", "GPU", "AMD", "NV"])
 
+  @unittest.skipUnless(Device[Device.DEFAULT].renderer.tensor_cores, "test needs TC")
+  def test_failure_59(self):
+    ast = UOp(Ops.SINK, dtypes.void, arg=None, src=(
+      UOp(Ops.STORE, dtypes.void, arg=None, src=(
+        UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(1024), arg=0, src=()),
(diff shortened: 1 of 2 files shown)


Qazalin:
looks like it only fails on METAL TC?

test/test_softmax_fusion.py
```diff
@@ -67,6 +67,12 @@ def test_double_gemm(self):
       c = (Tensor.rand(N,N)-0.5).realize()
     self._test_fuse(lambda a,b,c: a@b@c, a, b, c, atol=1e-5)
 
+  @unittest.skipUnless(Device.DEFAULT == "METAL", "fails only with METAL TC")
+  @unittest.expectedFailure
+  def test_double_gemm_beam(self):
+    with Context(BEAM=2):
```
chenyuxyz, line 73: can you use BEAM_DEBUG to find the ast and opts that failed (at that point it's better to put in test_linearizer_failures)? now it's non-deterministic because of BEAM

test/test_linearizer_failures.py
```diff
@@ -1430,5 +1430,34 @@ def test_failure_58(self):
...
+              UOp(Ops.DEFINE_GLOBAL, dtypes.float.ptr(1024), arg=3, src=()),
+              UOp(Ops.VIEW, dtypes.void, arg=ShapeTracker(views=(View(shape=(32, 32, 32, 1), strides=(0, 1, 32, 0), offset=0, mask=None, contiguous=False),)), src=()),)),)),)),)),))
+    opts = [Opt(op=OptOps.TC, axis=0, arg=(0, 2))]
+    lin = Kernel(ast, opts=Device[Device.DEFAULT].renderer)
+    try:
+      lin.apply_opts(opts)
+    except AssertionError:
+      # AssertionError is from `assert all_same(tensor_core_opts)` in kernel.py
+      # only fails with metal TC
+      assert Device.DEFAULT == "METAL"
```
chenyuxyz, line 1460: `helper_test_lin(Kernel(ast, opts=Device[Device.DEFAULT].renderer), opts=opts, failed_platforms=["METAL"])`

geohotstan:
Passes now on master 😃

PR #9943, opened 2025-04-20
CPUGraph for LLVM+CLANG

leopf:
Reuses the compiled CPUProgram for both LLVM and CLANG.

I have decided to pass in the functions as void pointers, to make sure `device.renderer._render_entry("batched", targs)` works correctly.

Running `test_order_2_writes_to_same_buf` produces:

```python
void batched(signed char* cbuf0,signed char* cbuf1,signed char* cbuf2,signed char* cbuf3,signed char* cbuf4,void* fxnp0) {
  void (*E_131072_4n2)(int*,int*,int*) = (void (*)(int*,int*,int*))fxnp0;
  E_131072_4n2((int*)(cbuf0 + 0),(int*)(cbuf1 + 0),(int*)(cbuf2 + 0));
  E_131072_4n2((int*)(cbuf0 + 0),(int*)(cbuf3 + 0),(int*)(cbuf4 + 0));
}
```


geohot:
Compare this to #10014 which claimed the bounty. Look how much simpler that one is

PR #9944, opened 2025-04-20
center graph on rewrite step selection

hurrrsh:
Ensures the graph auto-fits on each rewrite selection, preventing overly zoomed views and improving step navigation.

before:

https://github.com/user-attachments/assets/9771cb56-810e-4fca-b378-f3c0b55aa7f8

after:

https://github.com/user-attachments/assets/a8a49980-650d-46f8-9623-687db9598b61


Qazalin:
It's intentionally not this way.

PR #9945, opened 2025-04-20
WebGPU Graph

hooved:
This PR adds `WebGPUGraph`, part of the [WebGPU export refactor](https://github.com/tinygrad/tinygrad/pull/9424). In that branch, 40% more tok/sec was observed for tinychat in browser, as a result of mapping each tinygrad `WebGPUGraph` to a unique browser [`GPUCommandEncoder`](https://developer.mozilla.org/en-US/docs/Web/API/GPUCommandEncoder).

The code needs to be cleaned up to pass tests, and work with master branch WebGPU export logic.


hooved:
This all works but adds 31 lines of complexity for ~40% more speed. Based on discussion this may not be worth it, will close this PR for now.

PR #9950, opened 2025-04-20
RetinaNet INITMLPERF support

francislata:
# Overview
This PR adds support for `INITMLPERF` that can be used to run beam searches by using an empty dataset.

Note that the validation loop is updated so that it does not trigger the `pycocotools` related operations as it is not important for applying beam search on it.

Here's a screenshot of how it looks like when running the following command:
```bash
INITMLPERF=1 BENCHMARK=10 MODEL=retinanet BASE_DIR=/raid/datasets/openimages/ BS=48 GPUS=6 python3 examples/mlperf/model_train.py
```
Screenshot 2025-04-20 at 14 03 21

### DIFF
--- a/examples/mlperf/model_train.py
+++ b/examples/mlperf/model_train.py
@@ -385,7 +385,7 @@ def train_retinanet():
     x = Tensor.zeros(bs, 800, 800, 3, dtype=dtypes.uint8).contiguous()
     if val:
       img_ids, img_sizes = [0] * bs, [(800, 800)] * bs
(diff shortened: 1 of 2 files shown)


chenyuxyz:
estimated training time 0h5m seems wrong

chenyuxyz:
merging now so i can test locally. estimation can be fixed later

francislata:
> estimated training time 0h5m seems wrong

that's true that it is wrong. i'll create a follow-up PR later on to fix this issue.

PR #9960, opened 2025-04-21
Less verbose cloud multiprocessing start

uuuvn:
The set name before starting part used to be required for #9935 when CLOUDDEV was a global variable, now just readability improvement

### DIFF
--- a/tinygrad/runtime/ops_cloud.py
+++ b/tinygrad/runtime/ops_cloud.py
@@ -124,7 +124,6 @@ class CloudHandler(BaseHTTPRequestHandler):
   def do_POST(self): return self._do("POST")
 
 def cloud_server(port:int):
-  multiprocessing.current_process().name = "MainProcess"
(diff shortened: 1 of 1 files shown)


geohot:
Are these two changes related?

uuuvn:
> Are these two changes related?

Two changes in this pr or this pr and #9935?

Two changes in this pr is just setting everything in one place before starting the process instead of part in constructor, then modifying, then starting, then modifying again from the newly started process.

Setting process name before starting the process was required for earlier version of #9935 that had CLOUDDEV as a global variable because of assert that devices are only opened from main process in device.py, now completely unrelated

geohot:
This PR isn't mergable. There's still a whitespace change in it (never mix with changes that affect anything!), and there's two changes (an add of MainProcess and a removal of MainProcess) that I don't understand if you can do separately.

geohot:
Is a PR with only removal of line 127 valid, called something like "Remove MainProcess set in CLOUD device" with a description that explains why that's okay?

geohot:
Oh, I understand.

PR #9966, opened 2025-04-21
retinanet eval combine output on GPUS[0]

chenyuxyz:
eval 35 sec -> 20 sec. it was spending 13 seconds assembling output tensor on CPU backend. GPUS[0] seems to have enough memory, otherwise we can lower EVAL_BS


chenyuxyz:
more like 30.7 -> 25.7 for real run on green
image

PR #9967, opened 2025-04-21
fix `ptx` linearizer bug 2 [pr]

ignaciosica:
The root cause for this bug is that a special workaround within the linearizer for handling non-reduce locals doesn't function correctly following the pointer arithmetic rewrite in the PTX backend.

This was because it was relying on the store first source to check for local buffer, but with pointer arithmetic, its not longer the case, the first source usually is an Ops.ADD or Ops.CAST, thus not falling into special workaround check and not propagating correctly the context.

Now, instead of only checking the first source for local buffer, it checks for a local buffer in the toposort of the first source.

I benchmarked the change with `test/external/external_benchmark_schedule.py`
ptx_bug_2: `Mean: 550.43 ms  |  Std: 9.35 ms  |  Min: 542.32 ms  |  Max: 585.28 ms  (N=20)`
master (b35f94b6): `Mean: 543.89 ms  |  Std: 8.31 ms  |  Min: 536.69 ms  |  Max: 578.06 ms  (N=20)`

benchmark performed in tiny19; command to reproduce:
```
python - <<'PY'
import os, re, subprocess, statistics
N = 20
pat = re.compile(r'model linearize in\s+([0-9.]+)\s+ms')
base_env = os.environ.copy()
base_env["NULL"] = "1"

times = [
    float(pat.search(
        subprocess.check_output(
            ['python', 'test/external/external_benchmark_schedule.py'],
            text=True, env=base_env
        )
    ).group(1))
    for _ in range(N)
]

print(f"Mean: {statistics.mean(times):.2f} ms  |  "
      f"Std: {statistics.stdev(times):.2f} ms  |  "
      f"Min: {min(times):.2f} ms  |  "
      f"Max: {max(times):.2f} ms  (N={N})")
print("Values:", ", ".join(f"{t:.2f}" for t in times))
PY
```

This fixes also TC=3 emulation for ptx that was broken


ignaciosica:
Keeping as a draft until I add regression test. While adding TC=3 job in benchmark I realized the use_tensor_core attribute from apply_tensor_cores was not propagating correctly. Working on fixing.

PR #9974, opened 2025-04-22
replace cached property with ours [pr]

geohot:


geohot:
This isn't faster, closing

PR #9981, opened 2025-04-22
[bounty] [pr] index validation with z3

S-Lykles:
Some notes

- Process replay was failing before because of a kernel created by `extra/optimizations/test_beam_search.py`. The new index validation doesnt really change any kernels, but because the bounds check is now more accurate there are some beamed kernels that used to give an out of bounds error and were discarded that are now valid kernels. This means there are some different kernels in process replay which would then give a uop verification failure on master.

- There were a lot of failing symbolic tests so I have changed them so the tests run with IGNORE_OOB=1 Context. Tests like this example from `test_tiny.py`
```python
i = Variable('i', 1, 10)
ret = Tensor.ones(2).contiguous().reshape(i.bind(2)) + 1
ret.reshape(2).tolist()
```

- `test/test_linearizer_failures.py::TestLinearizerFailures::test_failure_58` now also fails with OOB on CUDA

- Whisper, Llama and gpt2 all fail bounds check. I duplicated one of the tests in `tests/models/test_whisper.py`. One runs with IGNORE_OOB=1 and the other with =0 with a  `@unittest.expectedFailure`.


chenyuxyz:
@geohot can you also take a look at this one? high level logic and usage of z3 look good to me. symbolic seems to have issue when reshape into a symbolic shape but that can be dealt separately

geohot:
Interesting, I see a bunch of OOB in `IGNORE_OOB=0 IGNORE_BEAM_CACHE=1 BEAM=2 DEBUG=2 python3 examples/beautiful_mnist.py` and I think they are real...

Looking into how much this slows CI down, but overall looks good.

geohot:
I'll merge and refactor like that since I took so long to review.

Great work, bounty is yours! george@tinygrad.org to claim, can pay via PayPal or USDC on ETH

S-Lykles:
Sent an email from @lykles.com

PR #9985, opened 2025-04-22
add FLAT formats to remu

Qazalin:
rebase from #9982. The FLAT format is very complex actually, it "may" fall into global memory space or local memory space, depending on a scalar register that is different per wave. This resolves all FLAT instructions to the GDS. But the addresses can fall into SDS as well, I'll need to hand write some tests that use the scratch address space next and resolve those correctly too.
https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna3-shader-instruction-set-architecture-feb-2023_0.pdf#page=120

### DIFF
--- a/extra/remu/src/thread.rs
+++ b/extra/remu/src/thread.rs
@@ -1505,8 +1505,12 @@ impl<'a> Thread<'a> {
             if !self.exec.read() {
                 return Ok(());
             }
-            let offset = sign_ext(instr & 0x1fff, 13);
             let seg = (instr >> 16) & 0x3;
+            let offset_bits = match seg {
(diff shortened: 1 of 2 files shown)


geohot:
oh, we probably don't want that reorder change, but we should merge the remu change

Qazalin:
Focusing on finishing multi output - I'll bring this back with a way to properly test remu comparing it to the real hardware.

We don't really have a way to render RDNA3 code in tinygrad. Maybe to start, I'll handcode the assembly in string literal. But we need a sensible way to assemble the machine code.

PR #9996, opened 2025-04-23
retinanet green scripts

chenyuxyz:
also removed realize in data_get and used empty for fake data. slightly bigger lr. https://wandb.ai/chenyuxyz/MLPerf-RetinaNet/runs/8skid0e8?nw=nwuserchenyuxyz


chenyuxyz:
still training but first 3 epochs look good (green 45 one)
image

PR #10001, opened 2025-04-23
Skip test that requires python 3.12 on older versions

uuuvn:
`out.cast(it.dtype.fmt).tolist()` fails with `ValueError: memoryview: destination format must be a native single character format prefixed with an optional '@'`


uuuvn:
image

test/test_image_dtype.py
```diff
@@ -12,6 +12,8 @@ def test_image_copyout_1x1(self, img_type=dtypes.imagef):
     buf = it.lazydata.buffer
     out = buf.as_buffer()
     np.testing.assert_equal(out.cast(it.dtype.fmt).tolist(), np.arange(4))
+
+  @unittest.skipIf(sys.version_info < (3, 12), 'Half not supported')
```
chenyuxyz, line 16: `@unittest.skipUnless(is_dtype_supported(dtypes.half), "need half")`

uuuvn: changed to that (but with `device="PYTHON"` instead of default device) because cast happens on python memoryview

PR #10003, opened 2025-04-23
Skip test that requires lru if device's allocator isn't lru

uuuvn:


geohot:
Was anything failing without this? This probably shouldn't be a standalone PR.

uuuvn:
> Was anything failing without this? This probably shouldn't be a standalone PR.

This test fails on cloud. The image tests are currently not being ran on cloud but if you run them manually it will fail

geohot:
Can you add them to run on CLOUD in CI with this PR?

uuuvn:
> Can you add them to run on CLOUD in CI with this PR?

Then this pr is going to be the same as #9903

geohot:
I'll merge this, but in the future I want PRs like this to have a way for me to test them, even locally is okay, but CI is better.

PR #10013, opened 2025-04-23
RetinaNet MLPerf logging

francislata:
# Overview
This PR adds `MLLOGGER` to the RetinaNet training script

# Tasks
- [ ] Add all necessary RetinaNet MLPerf logs
- [ ] Do mock run of RetinaNet with MLPerf logger enabled and verify its compliance
- [ ] Perform a real run with MLPerf logger enabled and verify its compliance


chenyuxyz:
you can also cross checked with previous submissions (e.g. https://github.com/mlcommons/training_results_v4.1/blob/main/Dell/results/XE8640x4H100-SXM-80GB/ssd/result_0.txt) to confirm the logged entries. i think the only difference is that it's called retinanet instead of ssd starting from 5.0

francislata:
a mock run is compliant with `compliance_checker`:
Screenshot 2025-04-25 at 11 32 40

francislata:
closing this in favor of #10078 since this PR's squash got ruined.

PR #10020, opened 2025-04-24
fix BlockingIOError in test_shm_tensor.py

leopf:
I was getting `tinygrad/runtime/ops_disk.py:28: BlockingIOError` when running `test/unit/test_shm_tensor.py TestRawShmBuffer.test_e2e_big` on my machine (Linux).

This is because the memlock limit is too low. Which is also why the `e2e_big` test was disabled for CI.
The fix is to check if RLIMIT_MEMLOCK is sufficient (skip if not) when running the test.

As a result, the tests still pass/are skipped in CI even with the `@skipIf(CI, ...)` removed and it doesnt crash on machines with low limits anymore.


geohot:
This doesn't improve anything, just hides a bug. Can the bug be fixed?

leopf:
Ok, but annoying to have a low signal failing test.

Will look into a proper fix. Closing for now.

PR #10023, opened 2025-04-24
Print real device behind cloud instead of an array of renderer properties

uuuvn:
(the first commit of this pr is #10021)

This is kind of a bug because currently with DEBUG>=1 it will say that
remote has device and then an array of renderer props instead of a real
device name which doesn't make sense:

```
127.0.0.1 - - [24/Apr/2025 16:50:44] "GET /properties HTTP/1.1" 200 -
remote has device ['tinygrad.renderer.cstyle', 'MetalRenderer', []]
opened device CLOUD from pid:20210
```

Now it will actually print the name of device behind cloud:

```
127.0.0.1 - - [24/Apr/2025 16:56:29] "GET /properties HTTP/1.1" 200 -
remote has device METAL
opened device CLOUD from pid:20315
```


uuuvn:
force-pushed myself to rerun ci, is there maybe a setting to allow anybody to do ci reruns the correct way?

PR #10029, opened 2025-04-24
Graph support for LLVM

quortus:
Second part of bounty (#10014)


tinygrad/runtime/graph/cpu.py
```diff
@@ -1,45 +1,67 @@
...
 from tinygrad.renderer.cstyle import ClangRenderer
+from tinygrad.renderer.llvmir import LLVMRenderer, ldt
 
-class CPUGraph(GraphRunner):
-  def __init__(self, device, jit_cache: list[ExecItem], input_rawbuffers: list[Buffer], var_vals: dict[Variable, int]):
-    if not issubclass(type(device.renderer), ClangRenderer) and not isinstance(device.renderer, ClangRenderer): raise GraphException
-    super().__init__(jit_cache, input_rawbuffers, var_vals)
+T = TypeVar('T')
+class BatchedGraph(Generic[T], GraphRunner):
+  def __init__(self, device, jit_cache: list[ExecItem], input_rawbufs: list[Buffer], var_vals: dict[Variable, int]):
```
geohot, line 14: Don't change names and functionality, it makes the PR harder to read. Style should always be separate from function

quortus: I did that to satisfy line width limitation, not just because I didn't like the old name.
But yeah, point taken. I rolled back this change and found another way.

geohot:
Looks good, bounty is yours! george@tinygrad.org to claim, can pay via PayPal or USDC on ETH

PR #10030, opened 2025-04-24
Refactor rand functions: extract Generator class for RNG management

currybab:
- split Generator class from Tensor class to support torch-like generator behavior
- create test for Generator class

### DIFF
--- a/test/test_tensor.py
+++ b/test/test_tensor.py
@@ -2,7 +2,7 @@ import subprocess
 import numpy as np
 import torch
 import unittest, copy, mmap, random, math, array
-from tinygrad import Tensor, Device, dtypes
+from tinygrad import Tensor, Device, dtypes, Generator
 from tinygrad.tensor import _METADATA
 from tinygrad.helpers import getenv, temp, mv_address
 from extra.gradcheck import numerical_jacobian, jacobian, gradcheck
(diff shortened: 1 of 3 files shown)


geohot:
Hmm, this is going to need a lot more tests. Does this match torch behavior exactly?

examples/flux1.py
```diff
@@ -442,7 +442,7 @@ def unpack(x:Tensor, height:int, width:int) -> Tensor:
   height = 16 * (args.height // 16)
   width = 16 * (args.width // 16)
 
-  if args.seed is None: args.seed = Tensor._seed
+  if args.seed is None: args.seed = Tensor._generator._seed
```
geohot, line 445: Things like this shouldn't just be changed, what is this API supposed to be?

currybab: The reason for switching from `Tensor._seed` to `Tensor._generator._seed` is that, due to changes in the internal structure of the Tensor library, seed management is now handled via the `_generator` object.

currybab: I understand that your team knows this library much better than I do, and that major changes like this require consensus and approval.
If you'd prefer maintaining access to the seed as before (via `_seed`), I’m happy to revert to the original approach so that seed management through the Tensor class remains unchanged.
Alternatively, if the main concern is about accessibility, perhaps we could manage the underlying seed with the generator, but still provide convenient access through a static method, such as `Tensor.seed()`.
Please let me know what you think.

PR #10031, opened 2025-04-24
Fix and enable jit tests on CLOUD

uuuvn:


test/test_jit.py
```diff
@@ -456,7 +456,7 @@ def f(a, b):
       np.testing.assert_allclose(a.numpy(), xc.numpy(), atol=1e-4, rtol=1e-5)
       np.testing.assert_allclose(b.numpy(), yc.numpy(), atol=1e-4, rtol=1e-5)
 
-  @unittest.skipIf(CI and Device.DEFAULT in {"GPU", "CUDA", "METAL"}, "no GPU/CUDA/METAL in CI, fine to run on AMD/NV")
```
uuuvn, line 459: Old `not_support_multi_device()` was equivalent to `CI and Device.DEFAULT in {"GPU", "CUDA", "METAL"}` and doing this very small noop change in a separate pr seems like going a bit too far, so submitting it together with the rest of the changes

PR #10032, opened 2025-04-24
bert on tiny13

nimlgen:
@chenyuxyz this is what I tested on tiny13. (it's with 6.2.4 fw)


chenyuxyz:
cool so `AM_RESET=1 AMD_SDMA_BIND=1` and fixed firmware version trains fine with AM?

if i patch with this and run on different red machines do you expect that to work?

PR #10034, opened 2025-04-24
improve `bf16` case for `is_dtype_supported` [pr]

ignaciosica:
There is still room for improvement, this checking is still very high level, but is a good tradeoff with that was there before, and more consistent with the actual landscape. I believe the way to go forward in order to be very reliable is to move the checking into each device and handle each case with more granular control, either by checking flags for support or gating on specific versions.

Some references on `bf16` support:

- [clang reference](https://clang.llvm.org/docs/LanguageExtensions.html#half-precision-floating-point)
- [metal reference](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf)


Qazalin:
err, why is it always skipping AMD?
Now the RDNA3 bf16 tests don't run in CI.

ignaciosica:
addressing in #10039

PR #10035, opened 2025-04-24
use `is_dtype_supported` to check dtype support in tc tests

ignaciosica:
`bf16` is not supported for `PYTHON` so no need to check for each `EMULATED_x` individually


chenyuxyz:
PYTHON should support bf16. we have truncate_bf16 now

ignaciosica:
image

ignaciosica:
> we should not make changes based on the assumption that it will not be supported in the future
Yes, I agree, but this change actually improves that by relying on the logic of `is_dtype_supported` to check if its supported. In the future, is easier to have all that checks depending on that function that each test independently checking for support.

chenyuxyz:
cool sounds good

PR #10037, opened 2025-04-24
Make sure clang always inline batched functions

quortus:
Makes #10014 more robust.
While working on LLVM graph (#10029) I found two tests (one was test_jit_multiple_random_regen in test_jit.py) where LLVM did not inlined functions with just linkage modifier (most likely because of the size).
I think there is a good chance that this can happen with clang backend too so it's better to force function inlining.

### DIFF
--- a/tinygrad/runtime/graph/cpu.py
+++ b/tinygrad/runtime/graph/cpu.py
@@ -31,8 +31,8 @@ class CPUGraph(GraphRunner):
     batched.append("}")
 
     prep = [device.renderer._render(cast(CompiledRunner, ji.prg).p.uops) for i,ji in enumerate(jit_cache)]
-    funcs = dedup(device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops, ["static"])
-                  for i,ji in enumerate(jit_cache))
+    funcs = dedup(device.renderer._render_body(prep[i][0], *prep[i][1:], cast(CompiledRunner, ji.prg).p.uops,
+                                               ["static", "__attribute__((always_inline))"]) for i,ji in enumerate(jit_cache))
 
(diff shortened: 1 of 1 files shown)


geohot:
Why does the function need to be inlined?

quortus:
It needs to be inlined to make sure it is always the first function in the generated code.
Function order in object file depends on some LLVM heuristics. `static` enables some inlining and after that LLVM seems to prefers to place it in the beginning. Except that I have two tests for LLVM graph where `static` (aka internal linkage) is not enough to guarantee the right placement:
https://github.com/tinygrad/tinygrad/actions/runs/14660219829/job/41142824874?pr=10029
https://github.com/tinygrad/tinygrad/actions/runs/14660219829/job/41142823292?pr=10029

I don't have a test for clang but it will probably be heavily dependent on compiler version and optimization flags, and thus, will have a high maintenance cost.
I think LLVM failures are enough.

nimlgen:
> I think there is a good chance that this can happen with clang backend too so it's better to force function inlining.

any clang failing tests you can come up with?
also blindly using `always_inline` might hurt performance significantly in some cases

quortus:
> also blindly using `always_inline` might hurt performance significantly in some cases

In generic code - yes. I don't see how this can hurt us here. Control flow is trivial, instruction cache misses are negligible. Or maybe you have specific example from tinygrad?


> any clang failing tests you can come up with?

Any reason why two LLVM failures are not enough? It uses the same linkage modifier.

With `static` right now we are risking getting random crashes (if we are lucky) and silent data corruptions (like in LLVM tests).

nimlgen:
No, the issue is real. And yeah always_inline should solve it for sure

quortus:
One other option is to prepend jump to the first non-static function to the binary image in jit_loader. Should I do that instead?

geohot:
Discussed in meeting, merged. Will review LLVMGraph today/tomorrow.

PR #10047, opened 2025-04-25
Change yolo webgpu res from 416x416 to 640x640

roryclear:
640 is somehow slightly faster than 416 on my iPhone and similar on my computer
https://roryclear.github.io/yolo-webgpu/index_640.html
https://roryclear.github.io/yolo-webgpu/index.html


geohot:
What's the model trained in? Accuracy is more important than speed

roryclear:
> What's the model trained in? Accuracy is more important than speed

The weights are from Ultralytics and they don't say, but all their examples use 640 https://docs.ultralytics.com/models/yolov8/#performance-metrics

PR #10048, opened 2025-04-25
fix `use_tensor_cores` propagation

ignaciosica:
`use_tensor_cores` from `apply_tensor_cores` was not propagated correctly. The value used for this purpose was fetched from `USE_TC` context var again in `apply_opt`.

edit: removed pr as it breaks due to the new arg


ignaciosica:
remu failed in CI but tested locally and the test pass. Maybe it was fixed by the last remu refactors, rerunning job

Qazalin:
remu needs `DS_STORE_B16_D16_HI`, I'll add it now it's easy.

ignaciosica:
thanks!

ignaciosica:
TC=3 fails for LLVM backend, adding llvm local buffer support

geohot:
This broke all beam caches.

ignaciosica:
Thanks for pointing that out, I completely overlooked it. Next time is going to be gated through a TIP.

Is ts still worth incrementing the db number?

This is the number that should be incremented right?
https://github.com/tinygrad/tinygrad/blob/c3ff308abbb1d19437b2ae615b598890a99359a9/tinygrad/helpers.py#L177

PR #10053, opened 2025-04-26
Fix TestLinearizerFailures.test_failure_53

gilion-joel-b:
Maybe a stupid question but is this supposed to be grouped by axis 1?


yvonmanzi:
I think it was noted a long time ago that this bounty is about fixing a bug in linearizer, not modifying the test.

gilion-joel-b:
> I think it was noted a long time ago that this bounty is about fixing a bug in linearizer, not modifying the test.

Yeah makes sense, just wanted to ask, thanks 👍

PR #10068, opened 2025-04-27
Passing test_complexity_w_unroll2

S-Lykles:
The non folded divs stuff wasnt necessary, thats only if you have variables or ranges that dont start at 0. There are two things this is doing, it deals with a multiplication around the div and constants in the div bigger than the divisor.


geohot:
Any reason this is draft? Fix looks good to me, other failing tests seem unrelated

S-Lykles:
Wanted to clean up the function a bit and add a test. Is it blocking something?

geohot:
Yea, I want to get the `bug_in_div_range` branch merged to move toward single src ranges.

PR #10079, opened 2025-04-28
torch: add tiny process group and tests

tocubed:
Add the process group and functions needed for torch DDP. On CPU for now.


chenyuxyz:
stale

PR #10084, opened 2025-04-28
remu: only write v_cmp result if exec is set

Qazalin:
[page 24](https://m7ymdt1k0usrgkaw.public.blob.vercel-storage.com/ref-jRgBvTGLG866V3AJ3U0BmVkIefoNuU.pdf#page=24) states that VALUs are "Not skipped if the instruction writes SGPRs/VCC". V_CMP ops do write to SGPR/EXEC or VCC, so remu didn't skip them regardless of the EXEC mask value.

Though when I tested against real hardware, V_CMP ops in VOP3 do not alter the exec mask on inactive threads.

This makes the [uninitialized VGPR usage](https://discord.com/channels/1068976834382925865/1134866270320328824/1366049909773701150) issue irrelevant, because the output of the comparison does not impact EXEC.

### DIFF
--- a/extra/remu/src/thread.rs
+++ b/extra/remu/src/thread.rs
@@ -810,10 +810,12 @@ impl<'a> Thread<'a> {
                 _ => todo_instr!(instruction)?,
             };
 
-            match op >= 128 {
-                true => self.exec.set_lane(ret),
-                false => self.vcc.set_lane(ret),
-            };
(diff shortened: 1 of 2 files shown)


Qazalin:
would be nice to have llvm-mc in CI to run hwtest.py. I think it needs all of llvm-tools though.

PR #10085, opened 2025-04-28
fix full shape in kernel.py

geohot:


Qazalin:
fyi, PR is not asserting if you change title. But process replay has generated diffs.

To rerun process replay, you have to also push a new commit.

geohot:
I'm aware, pushing new commit in a minute.

PR #10097, opened 2025-04-29
display viz rewrites with tabbing if they are subrewrites

Qazalin:
space can be 0-n depending on the depth:

image


Qazalin:
I should be merging the TypedDict in viz/serve with the ops one, the only reason they're different is because of UOp serialization.

PR #10124, opened 2025-05-01
amd: finalize device in case of failures

nimlgen:
this allows to move AM into sleep mode in case of some device failures (like pf).


Qazalin:
is the MOCKGPU=1 sync hanging forever also easy to fix?

I put a simple repro here: https://github.com/tinygrad/tinygrad/compare/master...Qazalin:tinygrad:sync_hang_forever. If run_asm fails it hangs in the while loop.

nimlgen:
`HCQDEV_WAIT_TIMEOUT_MS=1000` (or smaller) to not wait for a signal which will never come

PR #10125, opened 2025-05-01
[BOUNTY] Fix TestLinearizerFailures.test_failure_53

tusharparimi:
The problem seems to be that, 
when Ops.IF is already present in the BLOCKEND lst, the remove_blockend() func does not work and merge blockends rewrite ends there raising this,

if sink.op is not Ops.BLOCK or not all(x.op in DONT_PLACE_IN_BLOCK for x in sink.src): raise RuntimeError("linearize failure")

So, added an if statement in the remove_blockend() func to look for such cases and make that BLOCKEND a BLOCK, then linearizer continues from there.

Test passes on CUDA


geohot:
Tests failing. I'm deleting this bounty, all I'm getting is hacks and not principled fixes.

tusharparimi:
@geohot i know its not a bounty anymore but can I still open a PR if I make a principled fix?

PR #10129, opened 2025-05-01
kernelize setitem tests

Qazalin:
When we kernelize a full ASSIGN, all future children of the Tensor use the new ASSIGN(buf, kernel).

However, when we kernelize a partial ASSIGN, only the children of the VIEW(BUFFER) use the ASSIGN, in the current spec for BUFFER, there is no way to link a base BUFFER UOp to all the partial ASSIGN edges that it can have.

Thus, when we ASSIGN to a part of the kv cache and later use another VIEW of the kv cache Buffer, there is no link from the first ASSIGN to the base k/v cache BUFFER. Unless the partial ASSIGN gets realized in-place, it will not be reflected in the final output.


Qazalin:
setitem also has two different behaviors, unless realized in-place:

```
at = a[:2].assign(Tensor.full((2,), 1.0))
child = a[:2] + 2
print(child.lazydata) # BUFFER + CONST

a_slice = a[:2]
at = a_slice.assign(Tensor.full((2,), 1.0))
child = a_slice + 2
print(child.lazydata) # ASSIGN + CONST
```

PR #10146, opened 2025-05-03
[bounty] TODO: this cast can be removed by putting the casts around the EXPAND, must test properly

457crash:
The only use of AI was to understand the codebase and not write code. This is my first time contributing to tinygrad. Hope you find this helpful.

Command to run unit tests:

python test_gradient.py TestCastsAroundExpand

### DIFF
--- a/test/unit/test_gradient.py
+++ b/test/unit/test_gradient.py
@@ -131,21 +131,21 @@ class TestCastsAroundExpand(unittest.TestCase):
     dt, dw = t.conv2d(w).sum().gradient(t,w)
     self.assertListEqual(dt.tolist(), [[[[1, 2, 1], [2, 4, 2], [1, 2, 1]]]])
     self.assertListEqual(dw.tolist(), [[[[8, 12], [20, 24]]]])
-    
+
   def test_signed_int(self):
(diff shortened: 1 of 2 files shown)


geohot:
Problem not understood

457crash:
I have tinkered a lot in trying to understand this problem. Can you point me towards what the problem is about based on my several understandings below. Or is it entirely different from what I'm thinking?:

1. We need to remove the cast functions on both sides of the reduce axis function? Which is what I am trying to do in this PR
2. I tried to do it by updating the gradient.py file only. Although you could also do this by making changes in the core Tensor module
3. Are you looking for another formula for calculating the gradient of Ops.EXPAND instead?

457crash:
> Problem not understood

I have tinkered a lot in trying to understand this problem. Can you point me towards what the problem is about based on my several understandings below. Or is it entirely different from what I'm thinking?:

1. We need to remove the cast functions on both sides of the reduce axis function? Which is what I am trying to do in this PR
2. I tried to do it by updating the gradient.py file only. Although you could also do this by making changes in the core Tensor module
3. Are you looking for another formula for calculating the gradient of Ops.EXPAND instead?

PR #10151, opened 2025-05-04
[Bounty] Make sure buffers are GCed on CPU and with VIZ, with good tests

sashimicodechef:
- Use weak references in visualization code
- Add test to verify fix


sashimicodechef:
fixed! :)

geohot:
Way too long. Tests are poorly written

PR #10164, opened 2025-05-05
Take neg out of idiv

S-Lykles:
A nice property about c div is you can take out the sign

### DIFF
--- a/test/unit/test_uop_symbolic.py
+++ b/test/unit/test_uop_symbolic.py
@@ -154,8 +154,8 @@ class TestSymbolic(unittest.TestCase):
     self.helper_test_variable(Variable("a", 0, 6) // 2, 0, 3, "(a//2)")
 
   def test_div_neg_min_max(self):
-    self.helper_test_variable(Variable("a", 1, 7) // -2, -3, 0, "(a//-2)")
-    self.helper_test_variable(Variable("a", 0, 6) // -2, -3, 0, "(a//-2)")
+    self.helper_test_variable(Variable("a", 1, 7) // -2, -3, 0, "((a//2)*-1)")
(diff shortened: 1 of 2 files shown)


S-Lykles:
Will probably fail the fuzzer because that uses python mod/div and this rule is only valid for cdiv/cmod

test/unit/test_uop_symbolic.py
```diff
@@ -154,8 +154,8 @@ def test_div_min_max(self):
     self.helper_test_variable(Variable("a", 0, 6) // 2, 0, 3, "(a//2)")
 
   def test_div_neg_min_max(self):
-    self.helper_test_variable(Variable("a", 1, 7) // -2, -3, 0, "(a//-2)")
-    self.helper_test_variable(Variable("a", 0, 6) // -2, -3, 0, "(a//-2)")
+    self.helper_test_variable(Variable("a", 1, 7) // -2, -3, 0, "((a//2)*-1)")
+    self.helper_test_variable(Variable("a", 0, 6) // -2, -3, 0, "((a//2)*-1)")
```
geohot, line 158: Will this be slower?

S-Lykles: On cstyle definitely not, possibly on some of the asm backends in some cases. Probably needs a few more rules like #10189
Negative divisors are very rare though and its nice to have a canonical form because otherwise the div/mod is not simplified

S-Lykles: Actually if its div by a const then it will get rendered the same by fast idiv, that rule takes the sign out anyways

```
  sign = 1 if d > 0 else -1
  m,s = magicgu(vmax := min(x.vmax, dtypes.max(x.dtype)), abs(d))
  if m * vmax <= dtypes.max(x.dtype): return sign * ((x*m) >> s)
```

PR #10168, opened 2025-05-06
even spacing in viz nodes

kevbuh:
The spacing inside of the VIZ=1 nodes seem to be a few pixels too much on the top and left side. I changed x to 0 and set the first line's dy to 12, and the rest to 14. The change is especially noticeable in the SINK node shown below, it looks centered now.

Before:
Screenshot 2025-05-05 at 10 27 54 PM

After:
Screenshot 2025-05-05 at 10 26 30 PM

### DIFF
--- a/tinygrad/viz/js/index.js
+++ b/tinygrad/viz/js/index.js
@@ -43,8 +43,8 @@ async function renderDag(graph, additions, recenter=false) {
       const x = (d.width-d.padding*2)/2;
       const y = (d.height-d.padding*2)/2;
       return `translate(-${x}, -${y})`;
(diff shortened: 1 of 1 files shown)


Qazalin:
doesn't look like a root cause fix, where is the 12 coming from?

Qazalin:
Still a hack, this needs to be solved in the g.label translate calculation.

kevbuh:
Alright i got rid of that and instead added [text-after-edge](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/dominant-baseline#text-after-edge) and an offset for the stroke width.

Qazalin:
looks good, can you make STROKE_WIDTH a const in index.js?

Also, you know why safari still isn't properly centering it?

PR #10174, opened 2025-05-06
Div and mod folding uses cdiv

S-Lykles:


S-Lykles:
This has become a lot of seperate prs

PR #10178, opened 2025-05-06
Cleanup in `div_and_mod_folding` [pr]

S-Lykles:
I waay overthought this when I wrote it. Im just doing some refactoring because making changes to div_and_mod_folding before  #10174


S-Lykles:
Should pass process replay after #10196

S-Lykles:
Was just some randomness test that was failing. I've seen that fail before

PR #10179, opened 2025-05-06
simplify WebGPU runtime involving infinity

hooved:
By treating INFINITY the same way as we currently treat nan in WGSL rendering, we can simplify the runtime by not requiring an infinity uniform buffer argument for all the kernels, which also simplifies computing bind groups and layouts.


hooved:
Found the below previous context, currently evaluating:

https://github.com/tinygrad/tinygrad/pull/7063
https://github.com/pygfx/wgpu-py/issues/620

The main motivation is to reduce complexity/lines that will result from this refactor: https://github.com/tinygrad/tinygrad/pull/9424

hooved:
I reverted the diff back to master except for a new test that passes on master, but defines the issue, and fails when I try to simplify the webgpu runtime to not require a uniform buffer to which we write `Infinity` from CPU: [failing test](https://github.com/tinygrad/tinygrad/actions/runs/14884625613/job/41800818264)

The test will save time and ensure correctness if anyone tries to modify how `Infinity` is handled in our WebGPU rendering/runtime in the future.

On Apple devices (Metal backend), either with our `dawn` runtime (tested in CI) or with Safari's `webkit` (tested on my iPhone 15), I couldn't find a non-hacky way to avoid using an `Infinity` uniform buffer. Here are things I tried: [live browser tests](https://hooved.github.io/testwgsl/), [code](https://github.com/hooved/testwgsl/blob/master/test_wgsl.js). Of note, the simple fix of using constant `Infinity` instead of a uniform buffer was tested and works on dawn/Vulkan and dawn/D3D12.

I tested the current solution of using an `Infinity` uniform buffer, and confirmed that it works, on the below environments.

- dawn
    - Metal, tested in CI on MacOS
    - Vulkan, tested in CI on Ubuntu, and on Ubuntu (NVIDIA 3080) [with this link](https://hooved.github.io/testwgsl/)
    - D3D12, tested on Windows 10 (NVIDIA 3080) [with this link](https://hooved.github.io/testwgsl/)
- Webkit
    - Metal, tested on iPhone 15 iOS 18.4.1 [with this link](https://hooved.github.io/testwgsl/)

hooved:
Closing because this PR is hard to understand, will reopen with a simpler explanation

PR #10180, opened 2025-05-06
add support for `ds_load_u8` in remu

ignaciosica:
Fixes _RuntimeError: remu does not support the new instruction introduced in this kernel_ for https://github.com/tinygrad/tinygrad/pull/9771

Walkthrough docs are pending

Whenever you have time, feedback is much appreciated @Qazalin


ignaciosica:
[in progress]

Walkthrough in fixing simple case for:
`FAILED [...] RuntimeError: remu does not support the new instruction introduced in this kernel`

Whenever a test in CI fails with the runtime error from above, it means a new instruction is introduced in the kernel that is not yet supported by the remu emulator.

Qazalin:
looks good, thanks!

PR #10182, opened 2025-05-06
Less messy broken graph on paravirtualized metal workaround

uuuvn:
GitHub CI macOS runners use paravirtualized metal which is broken with graph (some comments say that ICB in particular is broken but in my testing it was fine sometimes, but other times hitting an assert inside metal's code related to resouces, so not sure).

> Assertion failed: (resource != nil), function -[IOGPUMetalResource initWithResource:], file IOGPUMetalResource.m, line 458.

This can be reproduced locally with any virtualization software (like utm) that can create macOS VMs with apple's own virtualization framework.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -706,11 +706,11 @@ jobs:
         cuda: 'true'
         llvm: 'true'
     - name: Run real world test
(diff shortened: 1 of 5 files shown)


uuuvn:
Ideally i should be checking for metal being paravirtualized but i couldn't find an easy way to do that

PR #10185, opened 2025-05-07
move benchmark stat tracking to influxdb

wozeparrot:


examples/llama3.py
```diff
@@ -166,7 +167,9 @@ def build_transformer(model_path: Path, model_size="8B", quantize=None, scale_dt
   model = Transformer(**MODEL_PARAMS[model_size]["args"], linear=linear, embedding=embedding, max_context=max_context, jit=True)
 
   if not load_weights: return model
+
   # load weights
+  log_event_start(BenchEvent.LOAD_WEIGHTS)
```
geohot, line 172: with Event?

geohot:
with WallTimeEvent <-- log difference in time.perf_counter
with KernelTimeEvent <-- log difference in GlobalCounters.kernel_time_s

PR #10193, opened 2025-05-07
cat loop split

pkotzbach:
went from this
obraz

to this
obraz


pkotzbach:
is this bounty still possible with RANGE having only 1 source now?
https://github.com/tinygrad/tinygrad/pull/10100

geohot:
This pattern looks extremely fragile

PR #10196, opened 2025-05-07
Fold constant variable

S-Lykles:
#10178 fails on process replay I think because there is a 'fake' variable with vmin==vmax=1 which happens to only get folded if its in an alu with a const. 
Just seeing what happens


S-Lykles:
So kernels that changed in process replay change are the same as in #10178

geohot:
We have `(UPat(GroupOp.ALU, name="x"), lambda x: x.const_like(x.vmin) if x.vmin == x.vmax else None),` want to just add it to this rule?

Issue #10006, opened 2025-04-23
UOp lifetime requirements conflict between systems


hooved:
Keeping UOps alive for the entire program simplifies VIZ implementation, but keeping UOps alive for as short as possible optimizes memory usage.

VIZ captures references to UOps in `TrackedGraphRewrite`, and these UOps live until program exit when they are pickled.

But some UOps are linked to buffers. Such a buffer cannot be GC'ed (or have its memory optimized by the memory planner or JIT capture) until the buffer's linked UOp is GC'ed. Therefore VIZ currently blocks GC of buffers if their linked UOps are referenced by `TrackedGraphRewrite` objects.

Tests demonstrating this behavior, which fail on master, are given here: https://github.com/tinygrad/tinygrad/pull/9856. Workaround solutions are also provided in #9856, which prevent `TrackedGraphRewrite` from referencing UOps that are linked to buffers, but a simpler solution is desired.

------

UOps interact with buffers through this system:
- A `UOp` object whose op is `Ops.BUFFER` can call `buffer()` to instantiate a `Buffer` object and increment the `Buffer` object's `lb_refcount` to 1
- Calling `buffer()` retrieves the `Buffer` object from cache instead of instantiating it:
    - if `buffer()` was previously called,
    - or if the `UOp` object was unpickled after being pickled with buffer data
- The buffer cache is `tinygrad.ops.buffers: WeakKeyDictionary[UOp, Buffer]`, which also enforces a 1:1 mapping between `UOp` and `Buffer` objects
- When a `UOp` object is GC'ed:
    - The `UOp` object's `__del__` method decrements `lb_refcount` to 0 on the `Buffer` object linked through `tinygrad.ops.buffers`
    - A strong reference to the `Buffer` object in `tinygrad.ops.buffers` is also removed, allowing GC of the `Buffer` object
- The `Tensor.lazydata: UOp` -> `UOp.buffer: Buffer` -> `Buffer` reference chain lets us know where data is for realized Tensors
- A `Buffer` object spawned by a `UOp` object that is still alive will have `lb_refcount`=1, causing the memory planner, and JIT capture, to preserve data in that `Buffer` object, which matters in these contexts:
    - For example, when `tinygrad.tensor._apply_map_to_tensors` substitutes part of a Tensor's `lazydata` graph with a UOp linked to a buffer, which prevents redundant computation across separate realizes
[...]

PR #10198, opened 2025-05-07
GC buffers on CPU and with VIZ

lewis-carson:
Fairly straightforward + simple change to ops - added weakrefs to TrackedGraphRewrite so UOps aren't held onto forever. Added tests.

### DIFF
--- a/test/test_viz_gc.py
+++ b/test/test_viz_gc.py
@@ -1,4 +1,6 @@
-import gc, unittest, os, weakref
+import os
+os.environ["VIZ"] = "1"
+import gc, unittest, weakref
 from tinygrad.tensor import Tensor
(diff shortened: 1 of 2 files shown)


hooved:
Won't you lose rewrites that you care about seeing in VIZ if the UOps are GC'ed this way?

Not sure you saw these but it might be helpful: https://github.com/tinygrad/tinygrad/issues/10006, https://github.com/tinygrad/tinygrad/pull/9856

geohot:
Whitespace changes

lewis-carson:
@hooved I think I've fixed it for now? I'm essentially keeping a list of tracked refs and then clearing it when VIZ is done. geohot has closed this though so I'm not sure if it's worth continued work.

PR #10200, opened 2025-05-07
Refactor test: Enable generality in testing UOp alu expressions

hooved:
This makes it easy to generally test execution of alu expressions in kernels, for when we want to write such expressions as UOps. Useful for https://github.com/tinygrad/tinygrad/pull/10201


geohot:
In general, whenever you add a function it should be used in at least two places.

hooved:
I use the new function twice here: https://github.com/tinygrad/tinygrad/pull/10201, this PR is just to show the refactor in isolation.

geohot:
It should really be one PR, but w/e

PR #10219, opened 2025-05-08
more `-a<-b` -> `b<a`

S-Lykles:


S-Lykles:
Actually needs more thinking because `uop_given_valid` expects `x<c`

PR #10229, opened 2025-05-09
fix test_div_numerator_negative

chenyuxyz:
the simplification was wrong with negative const_factor


S-Lykles:
Maybe you want to do
```python3
if x.vmin>=0:
  rems = [min(r, r-c, key=abs) for r in remainders]
  if (rem:=sum(r*v for r,v in zip(rems,svars))+const%c).vmin//c==rem.vmax//c:
    if which is Ops.MOD: return rem - rem.vmin//c*c
    return sum((f-r)//c * v for f,r,v in zip(factors,rems,svars)) + (const-const%c+rem.vmin//c*c)//c
```
You don't need the factors to be positive, its valid as long as its floored div

PR #10238, opened 2025-05-09
take gcd out of trunc div

S-Lykles:
This used to work but 
```
if (x.vmin < 0 or rem.vmin < 0)
```
disabled it.
If we include the const factor in the gcd then we can still remove it since trunc div == floor div  if the numerator is a multiple of the divisor

### DIFF
--- a/tinygrad/codegen/symbolic.py
+++ b/tinygrad/codegen/symbolic.py
@@ -156,6 +156,11 @@ def div_and_mod_folding(x: UOp, y: UOp, which: Literal[Ops.MOD, Ops.IDIV], split
     if which is Ops.MOD: return rem - rem.vmin//c*c
     return sum((f-r)//c * v for f,r,v in zip(factors,rems,svars)) + (const-const%c+rem.vmin//c*c)//c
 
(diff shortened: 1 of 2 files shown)


tinygrad/codegen/symbolic.py
```diff
@@ -156,6 +156,11 @@ def div_and_mod_folding(x: UOp, y: UOp, which: Literal[Ops.MOD, Ops.IDIV], split
     if which is Ops.MOD: return rem - rem.vmin//c*c
     return sum((f-r)//c * v for f,r,v in zip(factors,rems,svars)) + (const-const%c+rem.vmin//c*c)//c
 
+  if math.gcd(gcd, const)!=1:
```
chenyuxyz, line 159: use `:=`

S-Lykles: You don't want to change gcd unless the condition is true though. 
Like you can take out the gcd even if the constant is not exactly divisible

chenyuxyz: yea but can rename this to `g` or something else. the point is to avoid immediately recomputing the same thing

PR #10239, opened 2025-05-09
Correct congruence folding

S-Lykles:
I think this is what will break gated image loads in openpilot


S-Lykles:
So this is what breaks the gated image simplification

S-Lykles:
from #10239 
```
int alu0 = (idx1*24);
...
int alu2 = (ridx5*3);
...
int alu4 = ((idx2*1536)+(ridx4*768)+ridx3+alu0+alu2+-771);
...
float4 val7 = (((((idx2+ridx4)<1)!=1)&(((idx1+ridx5)<1)!=1))?read_imagef(data1, smp, (int2)((alu4%768),(alu4/768))):(float4)(0.0f,0.0f,0.0f,0.0f));
```
The gate condition implies that alu4.vmin >=0 so therefore it can be simplified according to floordiv/mod rules...

PR #10246, opened 2025-05-10
add support for amdgpu-flat-work-group-size to AMD LLVM IR

geohot:


geohot:
@Qazalin 

`FAILED test/test_linearizer.py::TestLinearizer::test_tensor_cores_emulation - RuntimeError: remu does not support the new instruction introduced in this kernel`

PR #10251, opened 2025-05-11
Use substitute in sym infer

S-Lykles:


geohot:
The problem with this is that it's slow

PR #10264, opened 2025-05-12
Move session to RemoteRequest

uuuvn:
This is a prereq refactor for cloud multi which will make it possible to use multiple devices from cloud host instead of just one.

I will do that via changing a session to be a `tuple[token, dev_idx]`

Previously the session was in cookies, this is a problem because a single http request can contain many RemoteRequests with potentially different devices.

The alternatives are either:
\- sending commands for different devices in separate http requests (slow)
\- only adding an idx in RemoteRequest in basically the same way i added session here, keeping session a cookie and concat in server. This is how i've done it previously and it looks just strictly worse than having it all be in the same place.

### DIFF
--- a/tinygrad/runtime/ops_remote.py
+++ b/tinygrad/runtime/ops_remote.py
@@ -7,7 +7,7 @@
 from __future__ import annotations
 from typing import Callable, Optional, Any, cast
 from collections import defaultdict
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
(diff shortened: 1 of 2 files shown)


tinygrad/runtime/ops_remote.py
```diff
@@ -20,7 +20,8 @@
 
 # ***** API *****
 
-class RemoteRequest: pass
+@dataclass(frozen=True)
```
uuuvn, line 23: @dataclass here is needed for it to properly merge the fields with `kw_only=True`

PR #10265, opened 2025-05-12
equal to simplex `x==0` -> `x<1`

S-Lykles:


S-Lykles:
Is failing index validation because the gated store condition isnt being used in the validation

PR #10268, opened 2025-05-12
Failing tests for kernelize assign

hooved:
In the failing tests, nothing is being written to the `target` Tensor with `kernelize()` if we slice the `target` before `kernelize()`

This matches the behavior I see when kernelizing the cache_kv write in llama:

`self.cache_kv[:, :, start_pos:start_pos+seqlen, :, :].assign(Tensor.stack(xk, xv)).kernelize()`

Where at the end of inference, I checked and the realized cache_kv values were zero for all layers - I didn't write a test for this because it involves a lot of data, and more complexity.

This diff should be pruned to only have `test_kernelize_partial_assign` and/or parameterize the `test_assign_kv_cache` to use `kernelize`, the rest of the tests in the first commit are for illustration by comparison.

### DIFF
--- a/test/test_assign.py
+++ b/test/test_assign.py
@@ -240,6 +240,48 @@ class TestAssign(unittest.TestCase):
     out = attn.cache_k.flatten().numpy()
     np.testing.assert_allclose(out, [1.,1.,1.,1.,1.,1.,0.,0.,1.,1.,1.,1.,1.,1.,0.,0.])
 
+  # TODO: in addition, or just instead, repeat test_assign_kv_cache with kernelize instead of realize?
+  def test_kernelize_partial_assign(self):
+    results = []
+    for mode in ("kernelize", "realize"):
+      target = Tensor.zeros(2, 2, dtype=dtypes.int32).contiguous().realize()
+      src = Tensor.full((1, 2), 7, dtype=dtypes.int32)
+      if mode == "kernelize":
+        target[1:2, :].assign(src).kernelize()
+      elif mode == "realize":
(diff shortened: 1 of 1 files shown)


geohot:
Way too much copy/paste in these tests

hooved:
I eliminated the copy/paste, and refactored `test_assign_kv_cache` to allow testing partial assign with either `kernelize()` or `realize()`.

`test_kernelize_partial_assign_kv_cache` fails as expected.

`test_realize_partial_assign_kv_cache` is the closest test to [current code in extra/models/llama.py](https://github.com/tinygrad/tinygrad/blob/master/extra/models/llama.py#L81), but strangely, this test works in some environments, like [WebGPU CI](https://github.com/tinygrad/tinygrad/actions/runs/14999165514/job/42141199579?pr=10268#step:5:156) and locally on my CPU, but not in [CPU CI](https://github.com/tinygrad/tinygrad/actions/runs/14999165514/job/42141198947?pr=10268#step:6:189), [Metal CI](https://github.com/tinygrad/tinygrad/actions/runs/14999165514/job/42141200680?pr=10268#step:6:188), and others.

geohot:
@Qazalin can you look into this?

Qazalin:
@hooved tests are clear, thanks. Lazily building up a chain of `__setitem__` requires a fundamental change to specs.
The current implementation limits setitem to an in-place operation.

To do this correctly, we need to connect partial ASSIGN(s) to the base BUFFER, it's an interesting cyclic dependency challenge in the kernel graph. I'll update with a PR and a TIP to enable this.

PR #10269, opened 2025-05-13
Remote multi (basic)

uuuvn:
Simplest thing to be able to use remote with multiple gpus, very slow because no transfers (copyin copyout for cross-device copies)

### DIFF
--- a/tinygrad/runtime/ops_remote.py
+++ b/tinygrad/runtime/ops_remote.py
@@ -21,7 +21,7 @@ from tinygrad.runtime.graph.cpu import CPUGraph
 # ***** API *****
 
 @dataclass(frozen=True)
-class RemoteRequest: session: str|None = field(default=None, kw_only=True)
+class RemoteRequest: session: tuple[str, int]|None = field(default=None, kw_only=True)
 
 @dataclass(frozen=True)
 class BufferAlloc(RemoteRequest): buffer_num: int; size: int; options: BufferSpec # noqa: E702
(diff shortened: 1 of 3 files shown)


tinygrad/runtime/ops_remote.py
```diff
@@ -219,14 +219,13 @@ def __call__(self, *bufs, global_size=None, local_size=None, vals:tuple[int, ...
...
-      self.host = "127.0.0.1:6667"
+    self.host = getenv("HOST", "") or RemoteDevice.local_server()
 
     # state for the connection
-    self.session = binascii.hexlify(os.urandom(0x10)).decode()
-    self.buffer_num, self.graph_num = 0, 0
+    self.session_token = binascii.hexlify(os.urandom(0x10)).decode()
+    self.device_idx = int(device.split(":")[1]) if ":" in device else 0
+    self.buffer_num = 0
+    self.graph_num = 0
```
uuuvn, line 228: `cannot determine type of "buffer_num"` if i leave it one line, looks like mypy bug

tinygrad/runtime/ops_remote.py
```diff
@@ -254,6 +252,12 @@ def __del__(self):
 
   def q(self, x:RemoteRequest): self.req.q(replace(x, session=self.session))
 
+  @functools.cache
+  @staticmethod
+  def local_server():
```
geohot, line 257: Is this change related?

geohot:
Can you add a test for this to the cloud test?

uuuvn:
> Can you add a test for this to the cloud test?

Added, should be fine to merge after 10274

PR #10270, opened 2025-05-13
ONNX add tests against ORT

geohotstan:
split from #9609 
 - added missing test to cover Squeeze op when axes is None
 - resolved PRelu to match ORT
 - added test coverage for tests in `test/external/external_test_onnx_ops.py` where both we and ORT fails at but we match ORT behavior

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -290,13 +290,8 @@ def get_onnx_ops():
   def HardSigmoid(x:Tensor, alpha:float=0.2, beta:float=0.5): return (alpha*x + beta).clip(0, 1)
   def Gelu(x:Tensor, approximate:str|None=None): return x.gelu() if approximate == "tanh" else 0.5 * x * (1 + (x/math.sqrt(2)).erf())
   def BiasGelu(x: Tensor, bias: Tensor, approximate: str | None = None) -> Tensor: return Gelu(x + bias, approximate)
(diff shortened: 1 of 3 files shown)


extra/onnx.py
```diff
@@ -290,13 +290,8 @@ def Softmax_13(x:Tensor, axis:int=-1): return x.softmax(axis)
...
   def Gelu(x:Tensor, approximate:str|None=None): return x.gelu() if approximate == "tanh" else 0.5 * x * (1 + (x/math.sqrt(2)).erf())
   def BiasGelu(x: Tensor, bias: Tensor, approximate: str | None = None) -> Tensor: return Gelu(x + bias, approximate)
-  def FastGelu(x:Tensor, bias:Tensor|None=None):
-    # this is tanh approximated
-    return (x + bias).gelu() if bias is not None else x.gelu()
-  # TODO: fix this
-  def PRelu(X:Tensor, slope:Tensor):
-    slope = slope[0] if slope.shape[-1] != X.shape[-1] else slope
-    return (X > 0).where(X, X * slope)
+  def FastGelu(x:Tensor, bias:Tensor|None=None): return (x + bias).gelu() if bias is not None else x.gelu() # this is tanh approximated
```
chenyuxyz, line 293: `((x + bias) if bias is not None else x).gelu()`, but w/e

PR #10272, opened 2025-05-13
Tensor.unfold

eitanturok:


test/test_ops.py
```diff
@@ -250,6 +250,17 @@ def test_meshgrid(self):
 
     self.helper_test_exception([], lambda: torch.meshgrid(x, indexing="bad"), lambda: xt.meshgrid(indexing="bad"), expected=RuntimeError)
 
+  def test_unfold(self):
+    helper_test_op([(8,)], lambda x: x.unfold(0, 2, 1), lambda x: x.unfold(0, 2, 1))
```
chenyuxyz, line 254: one lambda is enough

tinygrad/tensor.py
```diff
@@ -1411,6 +1411,30 @@ def meshgrid(self:Tensor, *args:Tensor, indexing:Literal["ij", "xy"]="ij") -> tu
...
+    """
+    dim = self._resolve_dim(dim)
+    if size < 0: raise RuntimeError(f'size must be >= 0 but got {size=}')
+    if step <= 0: raise RuntimeError(f'step must be >0 but got {step=}')
+    if size > self.shape[dim]: raise RuntimeError(f'maximum size for tensor at dimension {dim} is {self.shape[dim]} but size is {size}')
+
+    dim_size = self.shape[dim]
+    num_slices = (dim_size - size) // step + 1
+    slices = [self[(slice(None),)*dim + (slice(i*step, i*step+size),)] for i in range(num_slices)]
+    return Tensor.stack(*slices, dim=dim)
```
chenyuxyz, line 1436: using stack means it is not just a view of original tensor, can it be done with movements only?

eitanturok: hmmm, I'll take a look.

eitanturok:
`PYTHONPATH=. LLVM=1 LLVMOPT=0 TINY_BACKEND=1 python test/test_ops.py TestOps.test_unfold` currently fails because of some pytorch backend stuff. Investigating...

tinygrad/tensor.py
eitanturok: With Claude's [help](https://claude.ai/share/81917d55-5f8f-485f-a6cb-7caa5b056e84), I implemented this used Tensor.gather instead of `Tensor.stack`.

I use `torch.arange` to create indices for `Tensor.gather` and `Tensor.gather` itself calls `Tensor._one_hot_along_dim` which uses `torch.arange`. So there is some external memory usage. But I believe we are **NOT** creating a new view to call `Tensor.unfold` anymore. I'm double checking this....

eitanturok: Because unfold creates **overlapping** windows, multiple elements in the output tensor may refer to the same element in the input tensor. Does this prevent us from having a contiguous ternsor?

Example:
```py
x = torch.tensor([[1,2,3,4,5],
                  [6,7,8,9,10]])
>>> x.unfold(1,2,1)
tensor([[[ 1,  2], [ 2,  3], [ 3,  4], [ 4,  5]],
        [[ 6,  7], [ 7,  8], [ 8,  9], [ 9, 10]]])
```
This diagram is helpful:
image

eitanturok: With Claude's [help](https://claude.ai/share/81917d55-5f8f-485f-a6cb-7caa5b056e84), I implemented `Tensor.unfold` using `Tensor.gather` instead of `Tensor.stack`. However, I don't think this is an improvement.

`Tensor.unfold` with `Tensor.stack` has two kernels:
Screenshot 2025-05-14 at 11 41 34 AM

`Tensor.unfold` with `Tensor.gather` has 6 kernels:
Screenshot 2025-05-14 at 11 42 45 AM

I think the extra kernels when using `Tensor.gather comes from creating indices for `Tensor.gather` which results in two extra calls to `Tensor.arange`. I think we get some extra kernels from this.

But I believe we are NOT creating a new view to call Tensor.unfold anymore.

eitanturok:
In [PyTorch](https://github.com/pytorch/pytorch/blob/4015166e5d51bc39d5a81aa59ad49720ec2a23fe/torch/_decomp/decompositions.py#L1106), the backwards pass of `unfold` in `pytorch` is often done in two kernels
```
    # nb. At the moment this generates two kernels in triton
    # It could potentially be fused into one call to scatter_reduce,
    # in the case step <= size provided scatter_reduce generates 1 kernel
```

eitanturok:
@chenyuxyz I'm worried it may not be possible to compute `Tensor.unfold` in a single kernel / with just movement ops.

`Tensor.unfold` creates overlapping windows, meaning multiple elements in the output tensor may refer to the same element in the input tensor. I think this may prevents us from computing `Tensor.unfold` in a single kernel. PyTorch runs into this issue and computes the backward pass of unfold using two different kernels for this reason. And we run into this same issue ourselves when running `Tensor.unfold` with `TINY_BACKEND=1`.

I also tried implemented `Tensor.unfold` without changing the original tensor by using `Tensor.gather` instead of `Tensor.stack`. But this just increased the number of kernels so I'm not sure if this is what we want.

Can you clarify:
* Do we want to compute `Tensor.unfold` using `Tensor.gather`? Not sure if this is better.
* Do we want to compute `Tensor.unfold` in a single kernel, just with movement ops, or just with preserving the original tensor? Can the views change? How exactly do we want to measure the efficiently of computing `Tensor.unfold`? I want to know what I should be testing and aiming for.

More info can be found in my comments above.

PR #10282, opened 2025-05-13
Factor out remote connection and cache it

uuuvn:
(stacked on top of #10269)

Should be a small speed improvement but the main reason this is needed is to have a defined ordering of RemoteRequests within one host so that transfers won't required doing something like:
```python
src_dev.batch_submit()
dest_dev.q(Transfer(dest, src_dev.session, src))
dest_dev.batch_submit()
```
for correctness.


uuuvn:
Rebased on top of #10283 assuming it's going to be accepted, will revert if that pr is going to be closed without merge

PR #10286, opened 2025-05-13
Make `dev` a property of `Allocator`

uuuvn:
(this is a prereq refactor for #10285)

At least `BufferXfer.copy` accesses it assuming it's always present, currently most devices just add this property on their own repeating the same code over and over again.

This is also a bit footguny, see `RemoteAllocator` that named this property `device` instead of `dev`, i could obviously just change that in one place but doing it globally seems like a better solution (and it reduces code duplication too).

`MallocAllocator` is a bit special, but passing `None` works just fine.

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 from dataclasses import dataclass, replace
 from collections import defaultdict
-from typing import cast, Optional, Any, Generic, TypeVar, Iterator, Generator
+from typing import Optional, Any, Generic, TypeVar, Iterator, Generator
 import multiprocessing, importlib, inspect, functools, pathlib, os, ctypes, ctypes.util, platform, contextlib, sys, re, atexit, pickle, decimal, time
 from tinygrad.helpers import CI, OSX, LRU, getenv, diskcache_get, diskcache_put, DEBUG, GlobalCounters, flat_mv, from_mv, PROFILE, temp, mv_address, \
                              cpu_time_execution, colored, Context, round_up, DISABLE_COMPILER_CACHE
@@ -241,7 +241,7 @@ class _MallocAllocator(LRUAllocator['Compiled']):
   def _copyout(self, dest:memoryview, src): ctypes.memmove(from_mv(dest), src, len(dest))
   def _offset(self, buf, size:int, offset:int): return from_mv(self._as_buffer(buf)[offset:offset+size])
 
-MallocAllocator = _MallocAllocator(cast('Compiled', None))
+MallocAllocator = _MallocAllocator(None) # type: ignore
 
 # NOTE: MAP_JIT is added to mmap module in python 3.13
 MAP_JIT = 0x0800
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 from dataclasses import dataclass, replace
 from collections import defaultdict
-from typing import Optional, Any, Iterator, Generator
+from typing import cast, Optional, Any, Generic, TypeVar, Iterator, Generator
 import multiprocessing, importlib, inspect, functools, pathlib, os, ctypes, ctypes.util, platform, contextlib, sys, re, atexit, pickle, decimal, time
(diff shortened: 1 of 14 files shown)


tinygrad/runtime/support/hcq.py
```diff
@@ -437,10 +437,9 @@ class HCQAllocatorBase(LRUAllocator, Generic[DeviceType]):
   """
 
   def __init__(self, dev:DeviceType, batch_size:int=(2 << 20), batch_cnt:int=32, copy_bufs=None, max_copyout_size:int|None=None):
-    self.dev:DeviceType = dev
+    super().__init__(dev)
```
uuuvn, line 440: implementations of `_alloc` (AMDAllocator, NVAllocator, etc) that is called in the line below access `.dev` so i had to move `super().__init__(dev)` from the end of the `__init__` to the start

tinygrad/device.py
```diff
@@ -186,6 +186,7 @@ def view(self, size:int, dtype:DType, offset:int) -> Buffer:
 
 # TODO: size, dest, src are the same type. can we enforce this?
 class Allocator:
+  def __init__(self, dev): self.dev = dev
```
geohot, line 189: Type?

uuuvn: if i add a a `Compiled` type hint mypy throws like 70 errors like `tinygrad/runtime/ops_dsp.py:100: error: Item "Compiled" of "Compiled | None" has no attribute "ion_fd"  [union-attr]` because backends access properties without casting

geohot: Then it needs to be specialized. Otherwise this PR is removing a lot of types

uuuvn: Added typing, seems to work just fine, that was easier than i expected

uuuvn: image
image

tinygrad/device.py
```diff
@@ -236,7 +241,7 @@ def _copyin(self, dest, src:memoryview): ctypes.memmove(dest, from_mv(src), len(
   def _copyout(self, dest:memoryview, src): ctypes.memmove(from_mv(dest), src, len(dest))
   def _offset(self, buf, size:int, offset:int): return from_mv(self._as_buffer(buf)[offset:offset+size])
 
-MallocAllocator = _MallocAllocator()
+MallocAllocator = _MallocAllocator(cast('Compiled', None))
```
uuuvn, line 244: This is the only part of this pr i don't like, setting bound to `Optional[Compiled]` breaks mypy and it just evaluates everything as `Any`

tinygrad/runtime/ops_npy.py
```diff
@@ -2,9 +2,9 @@
 from tinygrad.helpers import flat_mv
 from tinygrad.device import Compiled, Allocator
 
-class NpyAllocator(Allocator):
+class NpyAllocator(Allocator['NpyDevice']):
```
geohot, line 5: Why is this in quotes? Does from future import annotations fix?

tinygrad/device.py
geohot: Can you just ignore the type of this line?

tinygrad/runtime/ops_npy.py
uuuvn: Future annotations doesn't fix that for some reason, it's in quotes where the device is defined after the allocator, normal otherwise

tinygrad/device.py
uuuvn: don't see how that's better than explicitly casting the thing but sure, changed

uuuvn:
> The real solution to that type ignore is to make a CPUDevice that both CLANG and LLVM use I think

How would that help? The allocator would still need to have an instance of device that it is attached to and unless it's a single instance of `class CPUDevice` then it would be wrong on other instances.

Isn't the only way to solve this (without major refactoring enabling having one allocator instance for multiple devices, might actually be needed if some weirdo device has shared memory for multiple logical tinygrad devices, maybe like an old mi200 or when exposing mi300 as underlying xcds or something like that?) is renaming `class _MallocAllocator` to just `class MallocAllocator` and having each device instance create a new allocator instance with `self` as `dev` instead of one shared `MallocAllocator` instance that would otherwise need to somehow point to the correct device?

PR #10299, opened 2025-05-14
Add floor_divide for UOps

python273:


tinygrad/ops.py
```diff
@@ -45,6 +45,15 @@ def __sub__(self, x): return self.sub(x)
   def __mul__(self, x): return self.mul(x)
   def __truediv__(self, x): return self.div(x)
   def __floordiv__(self, x): return self.idiv(x)  # TODO: idiv is trunc div, not floordiv
+
+  def floor_divide(self, x, reverse=False):
```
chenyuxyz, line 49: where is this used?

python273: In `ceildiv`

https://discord.com/channels/1068976834382925865/1069001075828469790/1371186135446851706
https://discord.com/channels/1068976834382925865/1069001075828469790/1371577821180198942

tinygrad/helpers.py
```diff
@@ -41,7 +41,10 @@ def fully_flatten(l):
   return [l]
 def fromimport(mod, frm): return getattr(__import__(mod, fromlist=[frm]), frm)
 def strip_parens(fst:str): return fst[1:-1] if fst[0] == '(' and fst[-1] == ')' and fst[1:-1].find('(') <= fst[1:-1].find(')') else fst
-def ceildiv(num, amt): return int(ret) if isinstance((ret:=-(num//-amt)), float) else ret
+def ceildiv(num, amt):
+  if hasattr(num, 'op'): return -(num.floor_divide(-amt))  # XXX: isinstance(num, UOp) is circular import
+  if hasattr(amt, 'op'): return -(-amt.floor_divide(num, reverse=True))
+  return int(ret) if isinstance((ret:=-(num//-amt)), float) else ret
```
python273, line 47: Not sure if there's a less hacky way to check `UOp`, maybe only moving to a separate file

tinygrad/ops.py
chenyuxyz: can it be triggered from the tensor level?

if it's just an internal corner case, might not be worth it to patch like this, but it's more serious if it can be triggered from tensor

python273: Yes, the first link's example is broken `conv2d` with symbolic shape. Second mentions the path `Tensor.conv2d` -> `Tensor._pool` -> `ceildiv`.

```python
from tinygrad import Tensor, Variable
from tinygrad.ops import UOp


def eval_uops(a: UOp):
    return a.sym_infer(dict(v.unbind() for v in a.vars()))

v = Variable('qwe', 1, 100000).bind(330) * 120 + 1
assert eval_uops(v) == 39601

x = Tensor.randn(1, 22, 39601)
x = x.reshape(1, 22, v)  # BUG: depending on this, the shape is different
weight = Tensor.randn(256, 22, 12)
bias = Tensor.randn(256)

result = x.conv2d(weight=weight, bias=bias, groups=1, stride=6, dilation=1, padding=(3, 3))
print(result.shape)
ishape = [eval_uops(i) if isinstance(i, UOp) else i for i in result.shape]
print(ishape)
assert ishape == [1, 256, 6600]  # BUG: [1, 256, 6599]
```

python273: hmm, though dunno if it's works even with the fix, seems to generate separate kernels (trying to run an ONNX model with tons of dynamic shapes)

chenyuxyz: we technically does not support reshape symbolic like this, all examples/tests are directly reshape into one bound variable to register symbolic shape. can it be triggered that way?

chenyuxyz: cool, can you add this as a test to test_symbolic_ops? maybe with `expectedFailure` first as a separate pr first

PR #10304, opened 2025-05-14
Benchmark remote hlb cifar

uuuvn:
hlb cifar is fast so added it, can add bert too if you think it's ok

6 real gpus to test multigraph and transfers + accuracy validation

should probably be added to tinystats too, i don't know how though


uuuvn:
> Traceback (most recent call last):
>  File "/Users/tiny/actions-runner/_work/tinygrad/tinygrad/process_replay.py", line 9, in <module>
>    from tinygrad.ops import UOp, Ops
> ModuleNotFoundError: No module named 'tinygrad.ops'

updated branch, should be fine

chenyuxyz:
total benchmark time creeps up to 20 minutes, but this seems fine

PR #10305, opened 2025-05-14
Fix: raise error on requires_grad=True for non-float tensors

### LINKED ISSUE #9723: raise error when setting requires_grad=True on non-float tensors

innofarah:
Fixes #9723.


chenyuxyz:
stale

PR #10324, opened 2025-05-15
remove FUSE_ARANGE_UINT [pr]

Qazalin:
New FUSE_ARANGE does not fuse E_ kernels with arange if there is an expand. This is because rangeless arange folding doesn't work in a lot of cases and ends up generating large for loops in the kernel.

Confirmed with `VIZ=1 GPU=1 FUSE_ARANGE=1 PYTHONPATH=. GPUS=1 BERT_SIZE="tiny" DEFAULT_FLOAT=HALF BENCHMARK=5 MODEL="bert" python examples/mlperf/model_train.py` arange and rand stay split:
image


Qazalin:
Not changing mlperf scripts since this was required in the older commits. Moving forward FUSE_ARANGE_UINT=0 is a no-op.

PR #10326, opened 2025-05-15
usbgpu: scsi read

nimlgen:


geohot:
Any hope on this? Faster read would improve the experience a lot

Issue #10321, opened 2025-05-15
bug in Tensor integer div and mod


chenyuxyz:
```
from tinygrad import Tensor
a = Tensor([110])
b = Tensor([55])
print(f"{a.div(b, rounding_mode='floor').item()=}")
print(f"{(a%b).item()=}")
```
outputs
```
a.div(b, rounding_mode='floor').item()=1
(a%b).item()=55
```

geohotstan:
This is probably the same problem as the `div` and `round` bug in 
 tinygrad/test/external/external_test_onnx_backend.py

```python
# BUG: we don't match ORT here due to some div inaccuracy with floats
backend_test.exclude('test_dynamicquantizelinear_cpu')
backend_test.exclude('test_dynamicquantizelinear_expanded_cpu')
```

I tried at this before but didn't manage to find the proper solution.

piyushK52:
I think the issue is with Ops.RECIP, it is not precise and that error propagates throughout the rest of the calculation. This is the final kernel for the above example.
```
#include <metal_stdlib>
using namespace metal;
kernel void E_(device int* data0, device int* data1, device int* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  int val0 = *(data1+0);
  int val1 = *(data2+0);
  float alu0 = (((float)(val0))*(1/((float)(val1))));   // 1.9999998808 instead of 2
  float cast0 = ((float)(((int)(alu0))));               // 1.0  ---> wrong output
  int alu1 = ((alu0<cast0)?((int)((cast0+-1.0f))):((int)(cast0)));
  *(data0+0) = alu1;
}
```
let me know if I am understanding the problem correctly, I can look into it to figure this out further.

chenyuxyz:
fixed https://github.com/tinygrad/tinygrad/pull/10327

PR #10327, opened 2025-05-15
fix tensor int floor div

chenyuxyz:
#10321 

passed fuzzer for integer mod.

some regression in const folding can wait for a different pr as this is more correct.

need to fix quantize DSP though

### DIFF
--- a/test/test_const_folding.py
+++ b/test/test_const_folding.py
@@ -81,8 +81,10 @@ class TestBinaryOpsConstFolding(unittest.TestCase):
   def test_div_tensor_one(self):
     _check_ast_count(0, Tensor([1.0, 2, 3, 4]) / Tensor.ones(4))
 
+  @unittest.expectedFailure  # TODO: fix
   def test_idiv_literal_one(self):
     _check_ast_count(0, Tensor([1, 2, 3, 4]) // 1)
+  @unittest.expectedFailure  # TODO: fix
   def test_idiv_tensor_one(self):
     _check_ast_count(0, Tensor([1, 2, 3, 4]) // Tensor.ones(4, dtype=dtypes.int32))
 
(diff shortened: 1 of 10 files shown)


geohotstan:
image

```python
  def test_prequant_gemm_intacc(self, xi=np.uint8, wi=np.uint8, replace_src=None, N=512, clip=True, opts=None):
    X = Tensor(m1:=(np.random.uniform(0, 255, size=(N,N)).astype(xi))).realize()
    W = Tensor(m2:=(np.random.uniform(0, 255, size=(N,N)).astype(wi))).realize()
    # ugh, it's so broken with those casts. need DONT_REALIZE_EXPAND=1 python3 test/test_quantize_onnx.py TestQuantizeOnnx.test_prequant
    tg_dtype = dtypes.int8 if xi == np.int8 else dtypes.uint8
    with Context(DONT_REALIZE_EXPAND=1):
      out = (X.int().matmul(W.int())//1000)
      if clip: out = out.clip(dtypes.min(tg_dtype),dtypes.max(tg_dtype))
      out = out.cast(tg_dtype)
      opts = [Opt(op=OptOps.UPCAST, axis=1, arg=128), Opt(op=OptOps.UNROLL, axis=0, arg=4)] if opts is None else opts
      sexec(out, opts, replace_src, run_count=1)
    tout = out.numpy()
    mout = ((m1.astype(np.int32) @ m2.astype(np.int32)) / 1000)
    if clip: mout = mout.clip(dtypes.min(tg_dtype),dtypes.max(tg_dtype))
    mout = mout.astype(xi)
    print(tout)
    print(mout)
    np.testing.assert_equal(tout, mout)
```

`out = (X.int().matmul(W.int())//1000)` the floor div here is causing the issue.
when you match numpy by also floor dividing the numpy implementation or removing the floor div in the tinygrad implementation, test will pass.

I'm not too sure why there's a tinygrad/numpy mismatch in the first place.

PR #10329, opened 2025-05-15
`fold_unrolled_divs` with float cast 

S-Lykles:
Ill make the symbolic flat rule a seperate pr as well

PR #10331, opened 2025-05-15
const float distributes over int cast

S-Lykles:


tinygrad/codegen/symbolic.py
```diff
@@ -299,6 +299,8 @@ def gep_through_wmma(gep:UOp, wmma:UOp):
   (-1 * (UPat.var("x") + UPat.var("y")), lambda x,y: (-x)+(-y)),  # -(x+y) -> -x + -y
   # (x+y)*c -> x*c+y*c. only for int, float has inf*0=nan issue
   ((UPat.var("x", dtypes.ints) + UPat.var("y")) * UPat.cvar("c"), lambda x,y,c: x*c+y*c),
+  ((UPat.var("x", dtypes.ints) + UPat.cvar("y")).cast(dtypes.floats) * UPat.cvar("c"),
```
chenyuxyz, line 302: are there big wins with this? want to see if it's worth it

this is wrong if (x + y) is unsigned int that wraps around

chenyuxyz: oh it's used in #10329

S-Lykles: Ill change it to sints only

S-Lykles:
Still thinking about this a little bit more, #10329 works without this rule now if you have some more variations of fold_unrolled_divs pattern, to do it more generically when `c=1.0` you would also need the pattern:
```
((UPat.var("x", dtypes.sints) + UPat.cvar("y")).cast(dtypes.floats), lambda x,y: x.cast(c.dtype)+y.cast(c.dtype),
```
But that not really something you want to do (unless it means you can fold the divs)

PR #10335, opened 2025-05-15
multi sync support in remu

ignaciosica:
This pull request adds support for `DS_LOAD_B96` and `DS_STORE_B96`, which are required by `test_mid_dim_multireduce`. 

### multi sync

The first approach I took extended the existing implementation. Before executing wave passes, I counted the number of barriers present in the kernel and executed exactly that many passes. However, this approach doesn't handle cases where barriers are placed inside loops, as these barriers may be encountered multiple times.

The second approach addresses this limitation by continuing to execute wave passes until all waves complete. I modified the `exec_wave` API to return `true` if it reaches the end of the program, and `false` if it encounters a barrier. Consequently, `exec_waves` now repeatedly runs wave passes until every wave has finished executing.

Additionally, I simplified barrier synchronization to support only the `s_barrier` instruction, as I couldn't make it reliably work for the multi synchronization instruction matching.

I added basic tests directly to remu. More extensive testing will be incorporated from the frontend using OptOps.LDS.

### DIFF
--- a/extra/remu/src/thread.rs
+++ b/extra/remu/src/thread.rs
@@ -1396,9 +1396,10 @@ impl<'a> Thread<'a> {
 
             match op {
                 // load
-                54 | 118 | 255 => {
+                54 | 118 | 254 | 255 => {
                     let dwords = match op {
                         255 => 4,
+                        254 => 3,
                         118 => 2,
                         _ => 1,
                     };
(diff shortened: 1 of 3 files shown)


Qazalin:
Your s_barrier change is definitely correct, I'll merge it if you wanna put it in a separate diff.
The compiler optimizes away s_barrier (even when we insert it in our renderer) when the L size is <32. That's guaranteed to always be in sync!

ignaciosica:
sure, I'll break it into 3 separate prs
- `s_barrier`
- `DS_LOAD_B96` and `DS_STORE_B96`
- multi sync

ignaciosica:
currently trying out different mechanisms for safer barrier sync emulation

chenyuxyz:
stale

PR #10344, opened 2025-05-15
More efficient shard

uuuvn:
resnet step native:
master: 2364.34 ms
this: 2127.24 ms

resnet step remote:
master: 6754.24 ms
this: 3384.33

For some reason those shrinks are kernels instead of just view, but it's still faster than copying everything


geohot:
I think this should wait for #10183

PR #10352, opened 2025-05-16
cat loop splitting

leopf:
bounty: Fix cat at high level by implementing loop splitting

approach:
- find pivot by searching for cmplt
- duplicate store and substitute new ranges
- apply sym and constant fold the cmplt

---
```python
from tinygrad import Tensor

a = Tensor.rand(64).realize()
b = Tensor.rand(128).realize()
c = (a.pad((0, 64)) + b).realize()
```
becomes
```c
void E_32_4n1(float* restrict data0, float* restrict data1, float* restrict data2) {
  for (int ridx16384 = 0; ridx16384 < 16; ridx16384++) {
    int alu0 = ((ridx16384<<2)+64);
    float4 val0 = *((float4*)((data2+alu0)));
    *((float4*)((data0+alu0))) = val0;
  }
  for (int ridx32768 = 0; ridx32768 < 16; ridx32768++) {
    int alu3 = (ridx32768<<2);
    float4 val1 = *((float4*)((data1+alu3)));
    float4 val2 = *((float4*)((data2+alu3)));
    *((float4*)((data0+alu3))) = (float4){(val1[0]+val2[0]),(val1[1]+val2[1]),(val1[2]+val2[2]),(val1[3]+val2[3])};
  }
}
```

### DIFF
--- a/tinygrad/codegen/expander.py
+++ b/tinygrad/codegen/expander.py
@@ -1,6 +1,6 @@
 # this converts a lowerer program into a vectorized program
 
-import functools, itertools, operator
+import functools, itertools, operator, typing
 from tinygrad.helpers import AMX, dedup, flatten, all_same, prod
 from tinygrad.uop.ops import UOp, Ops, UPat, PatternMatcher, GroupOp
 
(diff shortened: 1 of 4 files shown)


tinygrad/codegen/expander.py
```diff
@@ -115,6 +115,17 @@ def _gate_srcs(u:UOp, gate:UOp) -> UOp:
...
 
+def split_loop(store: UOp, pivot: UOp, prange: UOp):
+  pivot_point = prange.src[0].minimum(pivot)
+  nrange_lower = UOp.range(prange.dtype, pivot_point, prange.arg + 0x8000)
+  # no reliable split (avoid inf. recursion)
+  if (nrange_lower < pivot_point).simplify().op == Ops.CMPLT: return None
+  nrange_upper = UOp.range(prange.dtype, prange.src[0] - pivot_point, prange.arg + 0x4000) + pivot_point
+  return UOp.sink(store.substitute({ prange: nrange_lower }), store.substitute({ prange: nrange_upper }))
+
+pm_split_loop = PatternMatcher([ (UPat(Ops.STORE, src=(UPat(), UPat(Ops.INDEX, src=(UPat(), UPat(), UPat(Ops.CMPLT,
```
geohot, line 126: Why does this have to do with store?

geohot:
No tests

leopf:
Thanks for the review. I reconsidered the approach.

New approach is more general but still WIP. Splits still happen around pivot points that can be found using pattern matchers. Like a upat that finds `[range] < [pivot]`. A split at `[pivot]` would result in the `<` simplifying out (with exceptions...).

The actual splitting happens with a SPLIT Op that is propagated forward in the graph until either a store or assign is reached. ~~There also is a SPLITSELECT (basically a GEP for SPLIT) which is allows a cleaner implementation of the assign split logic.~~

Added tests, but a lot more to be done here.

PR #10357, opened 2025-05-16
Remote beam

uuuvn:
Just overriding device in renderer seems to work fine, ~~currently running bert to check, will mark ready for review if works fine.~~ Bert ran successfully, but ci failed because cstylerenderer checks .device for renderer-specific things instead of having separate properties for them, #10361 should fix it


geohot:
Yea a good change anyway, we shouldn't be checking device name

PR #10359, opened 2025-05-16
Device: Allow DEVICE=X from instead of X=1

haraschax:
This PR isn't very clean. I tried to keep behavior the same, but not clear to me what the differences are between the following groups and whether they are handled well:

- ALL_DEVICES
- Device._devices
- Device._devices - ['DISK', 'NPY']
- Device.get_available_devices()

Should `Device._devices - ['DISK', 'NPY']` be named? Also seems confusing that `NPY=1` will silently switch to another device?

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -37,14 +37,16 @@ class _Device:
       with contextlib.suppress(Exception): yield self[device].device
   @functools.cached_property
   def DEFAULT(self) -> str:
-    from_env = [d for d in self._devices if d not in ["DISK", "NPY"] and getenv(d) == 1]
+    from_env = [d for d in ALL_DEVICES if getenv(d) == 1]
+    if (env_device:=getenv("DEVICE", '')) in ALL_DEVICES: from_env.append(env_device)
     assert len(from_env) < 2, f"multiple devices set in env: {from_env}"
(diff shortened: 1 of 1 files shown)


geohot:
This is going to need tests

haraschax:
I'll write tests.

wozeparrot:
maybe we want the concept of real and fake devices? like disk and npy would be fake devices

geohot:
So they aren't fake, they are just not capable of doing compute.

haraschax:
So NPY/DISK are storage only as I understand. NULL can fake everything but doesn't compute either. Any reason PYTHON, HIP, REMOTE, and NULL? aren't in ALL_DEVICES?

Should ALL_DEVICES just be `COMPUTE_DEVICES=self._devices - [NPY, DISK]`

chenyuxyz:
I think `ALL_DEVICES` shows useful devices. NULL is not included because it does not compute, PYTHON is for emulation only, HIP is for debugging and we don't recommend using it, REMOTE needs a real device etc. I am also not sure what's best here, open to suggestions.

Adding `DEVICE=X` is fine with some tests to validate behavior, and make sure `DEVICE=X Y=1` would throw

PR #10361, opened 2025-05-16
Use renderer properties instead of `.device` [pr]

uuuvn:
This seems to be the only thing that prevents remote backend from overriding device property on renderer for beam instead of changing `class Kernel`/etc to have a separate "real/execution/search device" in addition to `self.opts: Renderer`


geohot:
This should have [pr]

uuuvn:
> This should have [pr]

added, ci failure unrelated ("urllib.error.HTTPError: HTTP Error 429: Too Many Requests") https://github.com/tinygrad/tinygrad/actions/runs/15078203479/job/42390132364?pr=10361#step:10:41

geohot:
No changes left

PR #10362, opened 2025-05-16
don't use UOp.multi in Tensor.rand

geohot:


geohot:
If you want things like this to stay, they have to be tested

PR #10363, opened 2025-05-17
Fix fp16 precision with IMAGE=2 mode

ved1beta:
-Preserved float16 precision when IMAGE >= 2 and FLOAT16=1 is set
-Modified the skip condition for test_gemm_fp16 to run when both IMAGE=2 and FLOAT16=1 instead of skipping when any IMAGE value is set
-Increased tolerance values from 5e-3 to 5e-2
#6378

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -2370,9 +2370,9 @@ class Tensor(SimpleMathTrait):
     if any(s>1 for s in stride):
       # handle strides: (k) -> reshape -> (k,1) -> pad -> (k,s) -> reshape -> (k*s) -> shrink (k-(s-1))
       x = x.reshape(None, None, *flatten((k,1) for k in x.shape[2:]))
-      x = x.pad((None, None, *flatten((None,(0,s-1)) for s in stride)))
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -2370,9 +2370,9 @@ def conv_transpose2d(self, weight:Tensor, bias:Tensor|None=None, groups=1, strid
     if any(s>1 for s in stride):
       # handle strides: (k) -> reshape -> (k,1) -> pad -> (k,s) -> reshape -> (k*s) -> shrink (k-(s-1))
       x = x.reshape(None, None, *flatten((k,1) for k in x.shape[2:]))
-      x = x.pad((None, None, *flatten((None,(0,s-1)) for s in stride)))
+      x = x.pad(tuple((None, None, *flatten((None,(0,s-1)) for s in stride))))
```
geohot, line 2373: Are these changes related?

geohot:
Oh we explicitly don't want to do this. Compute should be in fp32, it would have to be a separate thing to change compute.

PR #10371, opened 2025-05-17
Remote fixedvars

uuuvn:
(first commit of this pr is #10385)

Somehow none of existing tests failed when fixedvars were added, added amd remote runner and extra tests for macos metal runner


geohot:
In general, benchmark shouldn't test correctness, it should only test speed (hence the name). That's why that wasn't merged, I much prefer this approach.

PR #10381, opened 2025-05-17
Fix getbits mask calculation

geohot:
## Summary
- update `getbits` mask formula
- add unit tests for various start/end ranges
- keep `getbits` on one line

## Testing
- `ruff check tinygrad/helpers.py test/unit/test_getbits.py`
- `PYTHONPATH=. python3 test/unit/test_getbits.py -v`


geohot:
hmm is this real?

PR #10383, opened 2025-05-18
Updated test_linearizer.py (test_mean_std_multireduce_multiout)

illesteggplant:
Fixed the "TODO: Similar error to test_multiout_intermediate_multireduce (implicit expand vs shape mismatch)" in test_linearizer.py.  

Ran the following command to verify:   **python test/test_linearizer.py**


geohot:
Do not use AI

PR #10385, opened 2025-05-18
amd mockgpu multigraph support

uuuvn:
For testing remote graph stuff (prompted by #10371) in ci

### DIFF
--- a/test/mockgpu/amd/amddriver.py
+++ b/test/mockgpu/amd/amddriver.py
@@ -39,8 +39,9 @@ class AMDDriver(VirtDriver):
   def __init__(self, gpus=6):
     super().__init__()
 
+    # NOTE: gpu ids start from one (id 0 is skipped in KFDIface._is_usable_gpu)
     self.tracked_files += [VirtFile('/dev/kfd', functools.partial(KFDFileDesc, driver=self))] + \
-      [VirtFile('/sys/devices/virtual/kfd/kfd/topology/nodes', functools.partial(DirFileDesc, child_names=[str(i) for i in range(gpus)]))]
+      [VirtFile('/sys/devices/virtual/kfd/kfd/topology/nodes', functools.partial(DirFileDesc, child_names=[str(i+1) for i in range(gpus)]))]
 
     self.gpus = {}
     self.next_fd = (1 << 30)
@@ -52,7 +53,7 @@ class AMDDriver(VirtDriver):
     self.next_doorbell = collections.defaultdict(int)
     self.mmu_event_ids = []
 
-    for i in range(gpus): self._prepare_gpu(i)
+    for i in range(gpus): self._prepare_gpu(i+1)
 
   def _alloc_fd(self):
     my_fd = self.next_fd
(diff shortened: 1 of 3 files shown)


test/mockgpu/amd/amdgpu.py
```diff
@@ -171,8 +183,7 @@ def _exec_indirect_buffer(self, n):
     wptr = memoryview(bytearray(8)).cast('Q')
     rptr[0] = 0
     wptr[0] = buf_sz
-    PM4Executor(self.gpu, (addr_hi << 32) | addr_lo, buf_sz * 4, mv_address(rptr), mv_address(wptr)).execute()
-    assert rptr[0] == wptr[0], "not everything executed in amdgpu"
```
uuuvn, line 175: The main change is removal of this assert that will fail when anything inside of an ib has to wait which is basically everything with graph and multi (single device graph seems fine but there are probably some cases where it has to wait for sdma or something like that, not sure)

geohot:
Does the mock NV backend work with multiple GPUs?

uuuvn:
> Does the mock NV backend work with multiple GPUs? I'm fine with adding support to AMD also, but if the NV one just works to get started maybe test on that.

It does but tests have tons of annoying skip hacks for nv on ci that make it easier to just fix amd than change all of them to properly detect device capabilities

geohot:
@nimlgen can you review?

test/mockgpu/amd/amdgpu.py
```diff
@@ -59,22 +59,34 @@ def create_sdma_packets():
...
+  @property
+  def executing(self): return self.rptr[0] < self.wptr[0] or self.ib_executor is not None
+
   def execute(self):
-    while self.rptr[0] < self.wptr[0]:
-      cont = True
+    prev_rptr, cont = self.rptr[0], True
+    while self.executing and cont:
+      if self.ib_executor is not None:
+        prev_rptr -= self.ib_executor.execute()
```
nimlgen, line 86: lgtm. just add a var here (like `executed_in_indirect`) and return `self.rptr[0] - prev_rptr + executed_in_indirect`. this should make it a bit easier to see the idea behind this line.

test/mockgpu/amd/amddriver.py
```diff
@@ -39,8 +39,9 @@ class AMDDriver(VirtDriver):
   def __init__(self, gpus=6):
     super().__init__()
 
+    # NOTE: gpu ids start from one (id 0 is skipped in KFDIface._is_usable_gpu)
     self.tracked_files += [VirtFile('/dev/kfd', functools.partial(KFDFileDesc, driver=self))] + \
-      [VirtFile('/sys/devices/virtual/kfd/kfd/topology/nodes', functools.partial(DirFileDesc, child_names=[str(i) for i in range(gpus)]))]
+      [VirtFile('/sys/devices/virtual/kfd/kfd/topology/nodes', functools.partial(DirFileDesc, child_names=[str(i+1) for i in range(gpus)]))]
```
nimlgen, line 44: this is mostly based on topology. but I am fine with matching tinyboxes here

test/mockgpu/amd/amdgpu.py
uuuvn: changed

PR #10395, opened 2025-05-18
Raise error when setting requires_grad=True on non-float tensors

### LINKED ISSUE #9723: raise error when setting requires_grad=True on non-float tensors

VedantRGosavi:
## Description
This PR addresses issue #9723 by adding checks that raise appropriate error messages when attempting to set `requires_grad=True` on non-float tensors.

## Changes
1. Added a check in `Tensor.__init__()` to prevent creating non-float tensors with `requires_grad=True`
2. Added a check in `Tensor.requires_grad_()` to prevent modifying existing non-float tensors to have `requires_grad=True` 
3. Added a check in `Optimizer.__init__()` to prevent optimizers from enabling `requires_grad` on non-float tensors
4. Added appropriate tests in `test_tensor.py` and `test_optim.py`

This behavior now matches PyTorch, which also raises an error when trying to enable gradients on non-floating point tensors.

## Testing
Added tests that verify:
- Creating a non-float tensor with `requires_grad=True` raises an error
- Setting `requires_grad_(True)` on a non-float tensor raises an error
- Passing a non-float tensor to an optimizer raises an error when trying to set `requires_grad=True`
- Float tensors still work properly with `requires_grad=True`

Fixes #9723

### DIFF
--- a/test/test_optim.py
+++ b/test/test_optim.py
@@ -132,5 +132,15 @@ class TestOptim(unittest.TestCase):
     optimizer.step()
     Tensor.training = old_state
 
+  def test_requires_grad_non_float(self):
+    # Test that optimizer raises an error when a non-float tensor with requires_grad=None is passed
+    x = Tensor([1, 2, 3], dtype=dtypes.int32)  # Non-float tensor with default requires_grad=None
+    with self.assertRaises(RuntimeError):
+      Adam([x])
+    with self.assertRaises(RuntimeError):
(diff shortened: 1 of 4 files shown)


geohot:
Don't use AI

VedantRGosavi:
i apologize this was made by mistake.

PR #10403, opened 2025-05-18
count viz javascript in lines

Qazalin:
Sadly there's no tokenizer builtin for javascript. Ignoring comments and empty lines only.

### DIFF
--- a/sz.py
+++ b/sz.py
@@ -10,7 +10,7 @@ TOKEN_WHITELIST = [token.OP, token.NAME, token.NUMBER, token.STRING]
 def is_docstring(t):
   return t.type == token.STRING and t.string.startswith('"""') and t.line.strip().startswith('"""')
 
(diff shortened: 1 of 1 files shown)


sz.py
```diff
@@ -10,18 +10,24 @@
...
   return t.type == token.STRING and t.string.startswith('"""') and t.line.strip().startswith('"""')
 
+def is_js_token(s): return len(s) and not s.startswith('//')
+
 def gen_stats(base_path="."):
   table = []
   for path, _, files in os.walk(os.path.join(base_path, "tinygrad")):
     for name in files:
-      if not name.endswith(".py"): continue
+      if not (name.endswith(".py") or (name.endswith(".js") and "assets" not in path)): continue
```
geohot, line 19: Err, what if a file is named assets?

PR #10408, opened 2025-05-19
no amdgpu kernel driver

geohot:


test/external/speed_v_theoretical.py
```diff
@@ -90,7 +90,7 @@ def f(conv, x) -> Tensor: return conv(x).realize()
   def test_conv_3x3_256_32_32_256_256(self): self._test_conv_3x3(256, 32, 32, 256, 256, nv_tflops=27, amd_tflops=20)
 
   # theoretical is nv_tflops=165, amd_tflops=123
-  def test_gemm_4096(self): self._test_matmul(4096, nv_tflops=115, amd_tflops=74)
+  def test_gemm_4096(self): self._test_matmul(4096, nv_tflops=115, amd_tflops=65)
```
geohot, line 93: Maybe without AM we weren't clearing L2?

PR #10412, opened 2025-05-19
[Bounty] Move topk, _index_put_impl_, index_tensor, masked_select, and randperm_generator off of cpu in torch_backend

mrsuperwealthy:


geohot:
Don't use AI

PR #10420, opened 2025-05-19
Use REAL_DEV for test skips

uuuvn:
This should fix remote cpu tests flakiness (segfaults were in `test_data_parallel_resnet_train_step` which is skipped on cpu but wasn't skipped on remote cpu)


.github/workflows/test.yml
```diff
@@ -523,6 +523,7 @@ jobs:
     timeout-minutes: 15
     env:
       IGNORE_OOB: 0
+      PYTHONPATH: ${{ github.workspace }}
```
uuuvn, line 526: This is needed because i added `from tests...` import in test_tiny.py, otherwise https://github.com/tinygrad/tinygrad/actions/runs/15125158926/job/42515763726#step:6:19

PR #10424, opened 2025-05-20
Only float tensors can require gradients

tusharparimi:
Fixes https://github.com/tinygrad/tinygrad/issues/9723

- Used property decorator for additional validation (only float can require gradients) while setting requires_grad value for a tensor.
- is_float() check added in _apply_uop() as Ops.CMPNE & Ops.CMPLT create tensor of dtype bool.
- is_float() check also added in onnx runner _parse_input() as some onnx backend test for optimizers like Adam that have timestep input which is int64 dtype.
- updated old tests and added new tests for requires_grad


tusharparimi:
working to update with new issue https://github.com/tinygrad/tinygrad/issues/10460

chenyuxyz:
fixed by #10475

PR #10430, opened 2025-05-20
verify rewrites in test_uop_symbolic

S-Lykles:


chenyuxyz:
can you resolve the conflict?

PR #10431, opened 2025-05-20
Add Ops.EQ

S-Lykles:
I understand why this op doesn't exist yet, you want to have a very restricted set of ops during symbolic simplification. But I think its nice to have as a late rewrite, there's a lot of ugly `(a!=b)!=True` in kernels right now


Qazalin:
Yeah if you change enums it can't unpickle in master. Because all Ops now have a different number.

PR #10432, opened 2025-05-20
More div conditions

S-Lykles:


test/unit/test_uop_symbolic.py
```diff
@@ -838,5 +838,16 @@ def test_unrolled_arange(self):
     assert ((alu0+2559)//-4).vmin == -639 and ((alu0+2559)//-4).vmax == 0
     assert (((alu0+2559)//-4)*(-1)).vmin == 0 and (((alu0+2559)//-4)*(-1)).vmax == 639
 
+class TestFuzzFailure(unittest.TestCase):
+  def test_fuzz_failure1(self):
+    v1=UOp(Ops.DEFINE_VAR, dtypes.int, arg=('v1', 0, 8), src=())
```
chenyuxyz, line 843: use `Variable` for construction should be more robust than call UOp directly, also more consistent with other tests

PR #10433, opened 2025-05-20
Correct divmod folding behind flag

S-Lykles:


chenyuxyz:
why does it need a flag?

S-Lykles:
Because it regresses openpilot gated load quite a lot. Was George's suggestion in the meeting to add a flag so I can add the fuzzer

PR #10434, opened 2025-05-20
intiger div mod fix

### LINKED ISSUE #6378: image_dot of 2 half inputs returns in half instead of float

ved1beta:
fix: preserve float16 precision in image_dot with IMAGE=2 mode

- Modify image_dot to return float16 when both inputs are float16 with IMAGE=2 and FLOAT16=1
- Adjust test tolerances to account for typical float16 precision differences
- Fixes #6378

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -4142,7 +4142,7 @@ class Tensor(SimpleMathTrait):
     # groups*cout x cin x H, W
     cw = w.transpose(w.ndim-1, w.ndim-2).reshape((groups*cout, cin, 1, 1))
     if dtype is None and IMAGE >= 2 and getenv("FLOAT16", 0) and self.dtype == dtypes.float16 and w.dtype == dtypes.float16:
-        dtype = dtypes.float16
+      dtype = dtypes.float16
     return cx.image_conv2d(cw, groups=groups, dtype=dtype).reshape(out_shape_t).transpose(self.ndim-1, self.ndim-2)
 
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -4141,7 +4141,11 @@ def image_dot(self, w:Tensor, dtype:DTypeLike|None=None) -> Tensor:
     cx = self.transpose(self.ndim-1, self.ndim-2).reshape((bs//groups, groups*cin, -1, 1))
     # groups*cout x cin x H, W
     cw = w.transpose(w.ndim-1, w.ndim-2).reshape((groups*cout, cin, 1, 1))
-    return cx.image_conv2d(cw, groups=groups, dtype=dtype).reshape(out_shape_t).transpose(self.ndim-1, self.ndim-2)
+    if dtype is None and IMAGE >= 2 and getenv("FLOAT16", 0) and self.dtype == dtypes.float16 and w.dtype == dtypes.float16:
```
chenyuxyz, line 4144: this is too specific and complicated

ved1beta: changing

PR #10435, opened 2025-05-20
onnx parser

b1tg:


python273:
It would need to output offsets into tensor data, so that it's possible to create DISK tensors form .onnx file directly (Seems to use `list` + `struct.unpack` currently?)

geohot:
We want this to be similar to `gguf_load` and `torch_load` where it takes in a Tensor. Can be on disk or can be on device.

extra/onnx_parser.py
```diff
@@ -0,0 +1,266 @@
...
+  else: obj[key_name] = val
+
+def dict_to_namespace(d):
+  if isinstance(d, dict): return SimpleNamespace(**{k: dict_to_namespace(v) for k, v in d.items()})
+  elif isinstance(d, list): return [dict_to_namespace(i) for i in d]
+  else: return d
+
+@accept_filename
+def onnx_load(tensor: Tensor):
+  reader = BufferedReader(TensorIO(tensor), 1_000_000)
```
geohot, line 57: Err, we have to do this for zipfile cause we are using the builtin, but there's no reason we should need BufferedReader here

b1tg: gguf/safesensors puts metadata(offset,type..) of all tensors at the beginning of the file, which is suitable for lazy loading, onnx is protobuf-encoded, which heavily uses varints(variable-width integers) to represent integers, if we want to load it in a GPU context, parsing the variant-length data will cause too many GPU->CPU which are really slow

geohotstan:
Ran it on my makeshift script https://gist.github.com/geohotstan/17d76b74d6efd4c0150ad726d016b97a.
`METAL=1 DISABLE_COMPILER_CACHE=1 DEBUG=2`
Output: https://gist.github.com/geohotstan/9d4cff5146dd63769b6ea551b2fe7234

Can confirm Tensors (weights and biases) are not copied to CPU

chenyuxyz:
`test/external/external_model_benchmark.py` failed in benchmark ci

b1tg:
external_model_benchmark.py should be fixed now. Couldn't reproduce benchmark error `openpilot dmonitoring compile3 0.9.7` because it need Qualcomm GPU, any suggestion to reproduce/debug that (QCOM=1)?

btw the test did pass with LLVM/GPU backend 

```
PYTHONPATH="." LLVM=1 LLVMOPT=1 JIT=2 BEAM=0 IMAGE=0 python examples/openpilot/compile3.py https://github.com/commaai/openpilot/raw/v0.9.7/selfdrive/modeld/models/dmonitoring_model.onnx
PYTHONPATH="." GPU=1  python examples/openpilot/compile3.py https://github.com/commaai/openpilot/raw/v0.9.7/selfdrive/modeld/models/dmonitoring_model.onnx
```

geohotstan:
Ran this on top huggingface models.
The only failures were from models that have [external_data](https://onnx.ai/onnx/repo-docs/ExternalData.html), meaning along with `model.onnx` there is also `model.onnx_data`.

Example repos that have this `onnx_data` [gpt2-large](https://huggingface.co/openai-community/gpt2-large/tree/main/onnx), [roberta-large](https://huggingface.co/FacebookAI/xlm-roberta-large/tree/main/onnx)

Steps to reproduce:
1. change function `run_huggingface_validate` in `extra/huggingface_onnx/run_models.py` to use new onnx_load
2. run `PYTHONPATH=. python extra/huggingface_onnx/run_models.py test --debug FacebookAI/xlm-roberta-large`

It's not very common to see `onnx_data` though.

---

also some larger models with many nodes still take a long time to load... But maybe this is fixable when onnx.py is combined with parser?

chenyuxyz:
`PYTHONPATH=. python extra/huggingface_onnx/run_models.py test --debug openai-community/gpt2-large` segfault on my mac. it validates fine on master (see picture, though there seems to be other unrelated error @geohotstan have you seen those?)

image

geohotstan:
> `PYTHONPATH=. python extra/huggingface_onnx/run_models.py test --debug openai-community/gpt2-large` segfault on my mac. it validates fine on master (see picture, though there seems to be other unrelated error @geohotstan have you seen those?)
> 
> image

ahhhhh that's embarrasing. It's because I wrote a bug D:
`.onnx_data` was not downloading.

geohotstan:
Oh gpt2-large is HUUUUGE (9222 op nodes)
most models are not this huge. xlm-roberta-large is (2480 op nodes), which already was pretty big.

For each node, there are things that have to go to CPU, so currently parsing speed does not scale well with number of nodes.
Instead, parsing speed does scale well with initializer (W & B tensors) size.

chenyuxyz:
`PYTHONPATH=. python extra/huggingface_onnx/run_models.py test --debug FacebookAI/xlm-roberta-large` also segfault and works fine on master

image

b1tg:
works on tinyr5 and my mac

chenyuxyz:
hmm ci works fine. looks like my mac issue. @geohotstan does this look good to you?

geohotstan:
> hmm ci works fine. looks like my mac issue. @geohotstan does this look good to you?

Yeah it runs fine for me :o

chenyuxyz:
running benchmark for one more time, will merge once it passed and bounty is yours!

PR #10436, opened 2025-05-20
raise error non float tensors & updated test to verify it.

dragonlord29:
Fixes https://github.com/tinygrad/tinygrad/issues/9723
- Updated the Tensor __init__ function with the required checks, and raise RuntimeError when it fails.
- Updated test/test_tensor.py to check for the above changes. 
- Test cases 
   - when dtype is inferred 
   - when dtype is explicitly mentioned


dragonlord29:
@chenyuxyz please review

geohot:
See #10460

PR #10437, opened 2025-05-20
Flash attention

Yexo:
This is very much an early draft, I'm sharing this to maybe avoid a bit of double work, and to have a good place to outline the areas of this work I'm struggling with to hopefully get some pointers in the right direction.

What is implemented so far: basic outline of flash attention. When running with FLASH_ATTENTION=1 (default right now for testing only) all calls to scaled_dot_product_attention will be handled via the flash attention implementation.

Next steps for this to actually be an improvement:
- [ ] Fuse the intermediate steps into one kernel, right now there is an explosion in number of kernels
- [ ] Flash attention paper slices the output buffer, and reads+writes to the same memory locations of it multiple times. I've tried to emulate that by assigning a UOp buffer to itself (see commented out code), but that leads to compile failures.
- [ ] The additions in symbolic should be removed again, they're there to eliminate part of the first kernel layers (first computation of m_new and first update of O)
- [x] is_causal is not yet supported
- [ ] Test/fix symbolic shapes for key/value (and attn_mask?)
- [ ] Size of cache memory is hardcoded, should be made device-specific or maybe best sizes discovered via beam search?
- [ ] The backwards pass can be made more efficient by using `l` to recompute softmax more efficiently. I think this needs a custom backwards pass, not autodiscovered via compute_gradient(). Not sure how to include that.

### DIFF
--- a/test/models/test_real_world.py
+++ b/test/models/test_real_world.py
@@ -165,7 +165,7 @@ class TestRealWorld(unittest.TestCase):
       for v in data.values(): v.to_(Device.DEFAULT)
 
       helper_test("train_bert", lambda: (data["input_ids"], data["segment_ids"], data["input_mask"], data["masked_lm_positions"], \
-          data["masked_lm_ids"], data["masked_lm_weights"], data["next_sentence_labels"]), train, 0.25, 346)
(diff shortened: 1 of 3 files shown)


chenyuxyz:
stale, but wrong level of abstraction (should be on fusion)

PR #10442, opened 2025-05-21
Fixing metal compile errors in extra/gemm (similar to #3365)

babusid:
`extra/gemm/metal_matmul.py` and `extra/gemm/metal_matvec.py` both failed to run on M2 Macbook target. 
This slight tweak resolves the issue.


babusid:
Similar to #3365

PR #10447, opened 2025-05-21
tensor: add index_put (also to torch_backend)

bojle:
adds index_put. does it by creating a mask of the size of the tensor.


tinygrad/tensor.py
```diff
@@ -289,6 +289,29 @@ def assign(self, x) -> Tensor:
     self.lazydata = self.lazydata.assign(x.lazydata)
     return self
 
+  def index_put(self, indices: Tensor, values: Tensor, accumulate: bool = False) -> Tensor:
+    if indices.ndim == 1: indices = indices.reshape((-1, 1))
+    assert indices.ndim == 2, "indices should be a single dimensional tensor of tensor-coordinates"
+    assert indices.shape[1] == self.ndim
+    assert values.numel() == 1 or values.shape == (indices.shape[0],), "values tensor should have as many elements as indices"
+    out = self
+    ind_np = indices.numpy()
```
Yexo, line 298: Shouldn't have a hard dependency on numpy(). Additionally, all Tensor functions should be lazy.

bojle: i've modified the code now. remove hard numpy dependence. 
regarding the latter part. i tried making everything lazy. the solutions that i could think of were quite messy and seemingly very memory intensive (creating lots of copies of the tensors). I settled with the current impl. It uses `.item()` twice, once for bounds check on indices and another for getting the values from the `val` tensor so the masks can be created. 

If you have an idea for a fully lazy index_put implementation, please show me the way.

geohot:
Feel free to reopen if still relevant

PR #10449, opened 2025-05-21
refactor `CUDA`, `NV` & `PTX` disassemble [pr]

ignaciosica:
unify CUDA/NV/PTX disassemble pipeline

before, it was quite a mess how it worked; the `disassemble` function from `CUDACompiler` was not working for itself as `nvdisam` expects a `cubin` lib, but the `compile` outputs `ptx`. That `disassemble` function did work for `NVCompiler` that inherited it as the compile function from it does generate `cubin` lib.

`CUDA=1 PTX=0` and `CUDA=1 PTX=1` were disassembled by `cuda_disassemble` helper fun that introduced a step using ptxas for ptx into sass compilation.

Added disassemble to NVPTX that before was a noop.

Removed extra disassembling happening in `cuda_ops` as it was already triggered with `DEBUG >= 7` in realize.

https://github.com/tinygrad/tinygrad/blob/475a7583b38458e5addf74e15026f0888c1feaf9/tinygrad/engine/realize.py#L45

Now disassemble works for all cuda backend/renderer combinations
- `CUDA=1 PTX=1`
- `CUDA=1 PTX=0`
- `NV=1 PTX=1`
- `NV=1 PTX=0`

### DIFF
--- a/tinygrad/runtime/ops_cuda.py
+++ b/tinygrad/runtime/ops_cuda.py
@@ -5,7 +5,7 @@ from tinygrad.device import Compiled, BufferSpec, LRUAllocator
 from tinygrad.renderer.cstyle import CUDARenderer
 from tinygrad.renderer.ptx import PTXRenderer
 from tinygrad.runtime.autogen import cuda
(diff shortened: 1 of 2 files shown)


geohot:
Merged, but could use tests

PR #10451, opened 2025-05-21
Fix METAL virtual device sync issue 

MashDevel:
## Issue
   - Metal virtual devices need to be properly synchronized when tensors are transferred between devices (e.g. METAL:0 → METAL:1), the destination can read data before the source finishes writing 

## Fix
   - Add explicit cross-device synchronization using MTLSharedEvent timeline events
   - Use encodeSignalEvent at end of command buffer on source device and on destination device create wait buffer with encodeWaitForEvent:value: before any dependent work. 
   - Track and increment timeline_value per device to ensure ordered signalling so work on the source device completes before the destination proceeds.
   
   More Details on [Synchronizing Events Across Multiple Devices](https://developer.apple.com/documentation/metal/synchronizing-events-across-multiple-devices-or-processes)


chenyuxyz:
stale draft

PR #10454, opened 2025-05-22
fix: apply ip version fixes during AMDIP creation

wozeparrot:
additionally nbio 2.3.0 is nbio_2_3_offset.h in the kernel

### DIFF
--- a/tinygrad/runtime/support/amd.py
+++ b/tinygrad/runtime/support/amd.py
@@ -15,9 +15,10 @@ class AMDReg:
   @property
   def addr(self): return self.bases[self.segment] + self.offset
 
-@dataclass(frozen=True)
+@dataclass
 class AMDIP:
   name:str; version:tuple[int, ...]; bases:tuple[int, ...] # noqa: E702
+  def __post_init__(self): self.version = fixup_ip_version(self.name, self.version)[0]
 
   @functools.cached_property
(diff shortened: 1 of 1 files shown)


tinygrad/runtime/support/amd.py
```diff
@@ -15,9 +15,10 @@ def decode(self, val: int) -> dict: return {name:getbits(val, start, end) for na
   @property
   def addr(self): return self.bases[self.segment] + self.offset
 
-@dataclass(frozen=True)
+@dataclass
 class AMDIP:
   name:str; version:tuple[int, ...]; bases:tuple[int, ...] # noqa: E702
+  def __post_init__(self): self.version = fixup_ip_version(self.name, self.version)[0]
```
nimlgen, line 21: want to keep the original ip version here. fixup_ip_version is applied to get the correct modules

wozeparrot: problem is here: https://github.com/tinygrad/tinygrad/blob/5e6b96a1be3dc0d8918b4741c10083f2bd0f258d/tinygrad/runtime/ops_amd.py#L121

looks at version and on gfx 10.3.2 its nbio 3.3.2 which should map to 2.3.0, but this check doesn't match with the "correct" version

nimlgen: oh, right. that's correct, versions of ip should be overwritten

PR #10463, opened 2025-05-22
Merge SimpleMathTrait and MathTrait

MashDevel:


tinygrad/tensor.py
```diff
@@ -3559,7 +3559,7 @@ def bitwise_not(self) -> Tensor:
     if self.dtype != dtypes.bool and not dtypes.is_int(self.dtype): raise RuntimeError(f"{self.dtype} is not supported")
     return self.logical_not() if self.dtype == dtypes.bool else self ^ -1
 
-  def lshift(self, x:int) -> Tensor:
+  def lshift(self, x:int, reverse=False) -> Tensor:
```
geohot, line 3562: Why are these functions still in Tensor?

geohot:
meh, I'll merge it, we can delete the functions from Tensor later.

geohot:
That reverse is wrong btw. x<<2**4 reversed is not 4<<2**x

PR #10473, opened 2025-05-22
feat: support request obj in fetch

wozeparrot:
some sites block the default urllib user agent, and the only way to set it is thru the request object.

not sure if its better to allow a custom request object to be passed or just to set our own default useragent


geohot:
I think we should just set a browser-like user agent

PR #10474, opened 2025-05-22
hashing

wozeparrot:
currently implements:

- shake128 (keccak)

scheduling takes a very long time


geohot:
Why`state: list[Tensor]`? That shrink of statet is slow

PR #10499, opened 2025-05-24
functionality to invert views to mops

geohot:


geohot:
I still like this view_cache...will check if just that is faster

PR #10511, opened 2025-05-25
Fix SQTT

pkourouklidis:
The recent AMDIP refactor removed field masks from the regs dict of AMDIP objects. This led to SQTT which relied on this breaking.

This pull request adds a method to AMDReg to be able to get the mask of a field and fixes SQTT.


geohot:
Any way to test this with the mock GPU in github actions?

pkourouklidis:
I wanted to add a unit test for this to prevent regression because I really want to avoid Windows dual-boot to use RGP. I believe there is no way to this with MOCKGPU and remu, at least I couldn't figure it out.

PR #10512, opened 2025-05-25
Very basic mock sqtt

uuuvn:
Would've caught #10511

This mockgpu sqtt emulation will just ignore basically everything and end up with a 0x1000 size trace full of zeroes, but just testing for things like register rename is better than nothing i guess


geohot:
Thanks!

uuuvn:
> This is draft or good to merge?

Was busy, yeah should be good (can't set ready to review from github mobile app)

PR #10517, opened 2025-05-26
simpler variable binding for gpt2 inference

eitanturok:
To perform inference on GPT2 on master we need to do variable binding:
```py
model = GPT2()
model(tokens, Variable("start_pos", 0, 10).bind(0))
```

In this PR, I moved the variable binding to `model.__call__`, like we do in `extra/models/llama.py`.

Now the user can just do:
```py
model = GPT2()
model(tokens, 0)
```

Like in `extra/models/llama.py`, the tokens are also treated as a `Tensor`, not a `UOp`. This allows us to simplify things further.


chenyuxyz:
stale, probably not useful

PR #10518, opened 2025-05-26
Add Tensor.unfold

geohotstan:
for #10272's completion sake
the trick was `_pool()` for movement ops only unfold


eitanturok:
Wow, great job using `_pool` for movement ops only unfold! Very clean!

I was having a bit of trouble finding a way to do `unfold` with only movement ops so I closed my PR but this is great.

geohotstan:
thanks! :D 
I only saw this unfold thing when ya'll closed your PRs. I feel bad for pretty much stealing the PR...
Maybe more docs or tutorials for the tricks Tinygrad uses would be good 🤔 (like _pool)

PR #10521, opened 2025-05-26
nvpci

nimlgen:


nimlgen:
fyi this takes a bit longer than expected.
the missing part is gr (graphics) context, so compute throws errors. copies and signals work...

PR #10530, opened 2025-05-27
[webgpu] Fix atomic shared mem load inside loop 

wpmed92:
`test_arange::TestIndexing::test_index_mnist_opt` with `GROUPTOP` causes a shared mem load inside a loop, which didn't get rewritten to packed load because of a missing `allow_any_len=True` in the wgsl pattern matcher.

The load inside loop part of the problematic kernel before the change:

```wgsl
if (((bool(lidx0))!=true)) {
    var acc0 = 0u;
    for (var ridx1001 = 0; ridx1001 < 16; ridx1001++) {
      var val4 = atomicLoad(&temp0[ridx1001]);
      acc0 = (acc0+val4);
    }
}
```

after the change with proper packed load.

```wgsl
if (((bool(lidx0))!=true)) {
    var acc0 = 0u;
    for (var ridx1001 = 0; ridx1001 < 16; ridx1001++) {
      var val4 = atomicLoad(&temp0[(ridx1001>>2)]);
      acc0 = (acc0+(u32(((val4>>(((u32(ridx1001))&3u)<<3u))&255u))));
    }
}
```

Although the load was atomic, we can see that the uchar is not shifted out in the first version.


chenyuxyz:
do you know why atomic is wrong for webgpu shared memory?

PR #10532, opened 2025-05-27
allow stacking of VIEW UOps [pr]

Qazalin:
UOp.view calling base is a legacy behavior from LazyBuffer's MovementOps.
MovementOps after VIEW replace the ShapeTracker inplace,
VIEW after a VIEW does merge_views, the grouper tolerates additional complexity to add the ShapeTrackers manually, but this should be done properly deeper in ops.


Qazalin:
diff in process_replay `NOOPT=1 PYTHONPATH=. python test/test_multitensor.py TestMultiTensor.test_data_parallel_resnet`
image

image

Edit: fixed, it was re-applying a swizzle.

PR #10541, opened 2025-05-27
refactor LOAD(DEFINE_GLOBAL, VIEW) in kernels to LOAD(VIEW(DEFINE_GLOBAL))

Qazalin:
Currently there's a separate merge_views rule for LOAD/STORE and other UOps. This spec change enables using a generic merge_views rule, that is universally applicable to all elementwise ops and memory ops. This also closely matches the structure in INDEX:
image
DEFINE_GLOBAL describes the pointer, VIEW describes INDEX+optional gate and LOAD/STORE describe the type of operation.


Qazalin:
regenerating dataset last, there is likely some rewrite rule using UOp.const instead of const_like to create constants.
image

PR #10543, opened 2025-05-27
workflow to run sdxl with search

chenyuxyz:


chenyuxyz:
max step time would roughly be the search time, and the median would be the searched step

PR #10559, opened 2025-05-29
Add Test for Setitem

eitanturok:
Indexing with a tensor of size >= 256 fails on my mac. I added a minimal reproducible test for this.

For example, the code
```py
tensor_size = 256
t = Tensor.arange(0, tensor_size).contiguous()
idx_big = Tensor.arange(0, 256, dtype=dtypes.int)
t[idx_big] = 5
```
results in the error
```sh
  File "/Users/eitanturok/tinygrad/test.py", line 5, in <module>
    t[idx_big] = 5
    ~^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/tensor.py", line 4250, in _wrapper
    ret = fn(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/tensor.py", line 1251, in __setitem__
    self.assign(res).realize()
  File "/Users/eitanturok/tinygrad/tinygrad/tensor.py", line 4225, in _wrapper
    if _METADATA.get() is not None: return fn(*args, **kwargs)
                                           ^^^^^^^^^^^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/tensor.py", line 270, in realize
    run_schedule(*self.schedule_with_vars(*lst), do_update_stats=do_update_stats)
  File "/Users/eitanturok/tinygrad/tinygrad/engine/realize.py", line 170, in run_schedule
    for si, ei in lower_schedule(schedule):
  File "/Users/eitanturok/tinygrad/tinygrad/engine/realize.py", line 163, in lower_schedule
    raise e
  File "/Users/eitanturok/tinygrad/tinygrad/engine/realize.py", line 157, in lower_schedule
    try: yield (si, lower_schedule_item(si))
                    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/engine/realize.py", line 152, in lower_schedule_item
    return ExecItem(*cast(tuple[Runner,list], si_lowerer.rewrite(si.ast, si.bufs)), si.metadata, si.fixedvars)
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/uop/ops.py", line 882, in rewrite
    if (ret:=match(uop, ctx)) is not None: return ret
             ^^^^^^^^^^^^^^^
  File "<string>", line 3, in compiled_match
  File "/Users/eitanturok/tinygrad/tinygrad/engine/realize.py", line 145, in <lambda>
    (UPat(Ops.SINK, name="sink"), lambda ctx,sink: (runner:=get_runner(ctx[0].device, sink), [ctx[x] for x in runner.p.globals])),
                                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/engine/realize.py", line 113, in get_runner
    method_cache[ckey] = method_cache[bkey] = ret = CompiledRunner(replace(prg, device=device))
                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/engine/realize.py", line 44, in __init__
    self.lib:bytes = precompiled if precompiled is not None else Device[p.device].compiler.compile_cached(p.src)
                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/device.py", line 321, in compile_cached
    lib = self.compile(src)
          ^^^^^^^^^^^^^^^^^
  File "/Users/eitanturok/tinygrad/tinygrad/runtime/ops_metal.py", line 138, in compile
    if isinstance(ret, Exception): raise ret
                                   ^^^^^^^^^
tinygrad.device.CompileError: program_source:265:271: fatal error: bracket nesting level exceeded maximum of 256
program_source:265:271: note: use -fbracket-depth=N to increase maximum nesting level
```

This seems to be a mac-specific error. So if relevant, here is info on my mac
```
Hardware Overview:

  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro18,1
  Model Number:	MK193LL/A
  Chip:	Apple M1 Pro
  Total Number of Cores:	10 (8 performance and 2 efficiency)
  Memory:	16 GB
  System Firmware Version:	11881.81.4
  OS Loader Version:	11881.81.4
```


chenyuxyz:
i think this is fixed, but fine to still add a test. i did not add this because it's too high level to the root cause. you can clean the test a bit to only keep the previously failing one

test/test_setitem.py
```diff
@@ -158,6 +158,12 @@ def test_setitem_overlapping_inplace2(self):
     t[:-1] = t[1:]
     self.assertEqual(t.tolist(), [[2.0], [1.0], [1.0]])
 
+  def test_setitem_big(self):
+    idx_size, val = 256, 4
+    t = Tensor.arange(0, idx_size+1).contiguous()
+    idx = Tensor.arange(0, idx_size, dtype=dtypes.int)
```
chenyuxyz: why specifying the dtype if it's already int?

test/test_setitem.py
```diff
@@ -158,6 +158,12 @@ def test_setitem_overlapping_inplace2(self):
...
+    t[idx] = val
```
chenyuxyz, line 165: fine to still adding this test, but i think it should at least test the setitem works correctly

also keep the test as clean as possible

test/test_setitem.py
```diff
@@ -158,6 +158,12 @@ def test_setitem_overlapping_inplace2(self):
     t[:-1] = t[1:]
     self.assertEqual(t.tolist(), [[2.0], [1.0], [1.0]])
 
+  def test_setitem_big(self):
+    idx_size, val = 256, 4
+    t = Tensor.arange(0, idx_size+1).contiguous()
```
chenyuxyz: why contiguous?

test/test_setitem.py
eitanturok: it was unnecessary and I removed it

test/test_setitem.py
eitanturok: i added a check that the setitem works correctly

test/test_setitem.py
eitanturok: I needed it from before, but removed it now

PR #10565, opened 2025-05-29
Hlb cifar10 random crop and cutmix to tinygrad

albertaillet:
- I use indexing for `random_crop` instead of the mask because `masked_select` is very slow.
- I have also written a version that skips `pad_reflect` entirely, which is faster but less readable.
- Tested that `random_crop` and `cutmix`  matche the prev version available ([random_crop](https://github.com/albertaillet/tinygrad/blob/albert-debug-branch/test_random_crop.py), [cutmix](https://github.com/albertaillet/tinygrad/blob/albert-debug-branch/test_random_crop_cutmix.py)).
- I would prefer to not have to flatten the indices, but only found this solution for using gather.

I benchmarked multiple solutions for `rand_mask` ([code](https://github.com/albertaillet/tinygrad/blob/albert-debug-branch/test_random_crop.py))
```
METAL=1 JITBEAM=4 python benchmark_and_test_random_crop.py
Dataset shape:   (100, 3, 32, 32), 186.97 ms numpy,  36.78 ms indexing,  576.47 ms masked_select
Dataset shape:   (500, 3, 32, 32), 307.07 ms numpy, 105.58 ms indexing, 9153.08 ms masked_select
Dataset shape: (50000, 3, 32, 32), 670.13 ms numpy, 483.35 ms indexing,   too slow masked_select
```


chenyuxyz:
can yo comment on the performance before/after?

albertaillet:
Another PR was merged that jitted the augmentations, closing this one.

PR #10566, opened 2025-05-29
cifar10 - use tinygrad for crop and cutmix 

geonot:


chenyuxyz:
can you comment on the performance before/after?

geonot:
I'll reopen with a better solution, there are some hangs with the RNG and other issues.

PR #10575, opened 2025-05-30
whisper: pad ultrashort waveforms (<N_FFT) to avoid STFT errors

Ojaswy:
- Modified load_file_waveform(...) in whisper.py so that any loaded WAV shorter than N_FFT (400 samples) gets zero-padded to length 400 before computing the mel-spectrogram.
- This prevents `librosa.stft(...)` from crashing with "signal too short for STFT" when given a very small file.


examples/whisper.py
```diff
@@ -231,6 +231,10 @@ def init_whisper(model_name="tiny.en", batch_size=1):
 
 def load_file_waveform(filename):
```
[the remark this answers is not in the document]
Ojaswy: Updated both

test/models/test_whisper.py
```diff
@@ -76,6 +77,21 @@ def test_transcribe_long_no_batch(self):
     self.assertEqual(2, len(trancriptions))
     self.assertEqual(TRANSCRIPTION_3, trancriptions[0])
     self.assertEqual(TRANSCRIPTION_1, trancriptions[1])
+  
+  def test_transcribe_file_tiny_wav(self):
+    sr = 16000   # matches RATE in whisper.py
```
chenyuxyz: just import `RATE`

test/models/test_whisper.py
```diff
@@ -76,6 +77,21 @@ def test_transcribe_long_no_batch(self):
...
+    tiny_data = np.zeros(100, dtype=np.float32)
+    tmp_f = tempfile.NamedTemporaryFile(suffix=".wav", delete=False)
+    sf.write(tmp_f.name, tiny_data, sr)
+    tmp_f.close()
+
+    try:
```
chenyuxyz, line 88: i don't understand this. what's the expected behavior?

examples/whisper.py
```diff
@@ -231,10 +231,16 @@ def init_whisper(model_name="tiny.en", batch_size=1):
...
   waveform, _ = librosa.load(filename, sr=RATE)
+# Padding added in transcribe_file
```
chenyuxyz: please read the diff carefully and remove unnecessarily stuff

test/models/test_whisper.py
Ojaswy: It creates a temp 100-sample silent WAV (far shorter than the 400-sample FFT window) for testing the padding logic in transcribe_file

chenyuxyz: my question was on the try except part, what does it do?

Ojaswy: Its to verify that, padding was added correctly and a transcription '<|nocaptions|>' is still returned

examples/whisper.py
Ojaswy: Anything else left to do before merging?

test/models/test_whisper.py
Ojaswy: Anything else to do?

chenyuxyz:
probably a desired fix, but the pr is just not clean

PR #10579, opened 2025-05-30
allow x18 register on linux

ttomsa:


ttomsa:
just realized it's implied windows or macos cause on linux target is aarch64 instead of arm64

PR #10584, opened 2025-05-31
fix: resolve pre-comit errors

ddalton15:
# Overview:

Noob first PR for TinyGrad. I noticed some issues while running some testing locally. Feel free to shut this down if these fixes are not needed. I figured I would push up fixes if they were needed.


chenyuxyz:
we'd like to have broken stuff fixed, but pr that fixed stuff needs to have tests in CI that can catch it. also don't reduce test coverage for no reason

PR #10585, opened 2025-05-31
[webgpu] Proper shared mem size for packed types

wpmed92:
When working on https://github.com/tinygrad/tinygrad/pull/10530 I realised that the shared mem size is 4 times the size it should be.
This is because in webgpu for packed types, the size has to be `size//4//dtype.itemsize`
As an example, for `char`, `uchar` we divide by 4//1 = 4, for `short`, `ushort` we divide by 4//2 = 2 to get how many dwords we need.

The kernel from the linked PR before and after the change.

Before:

```wgsl
fn nan() -> f32 { let bits = 0xffffffffu; return bitcast<f32>(bits); }
@group(0) @binding(0)
var<uniform> INFINITY : f32;

// This is a uchar array with 16 elements, so this should be 4 u32
var<workgroup> temp0: array<atomic<u32>, 16>;

@group(0) @binding(1)var<storage,read_write>data0:array<atomic<u32>>;
@group(0) @binding(2)var<storage,read_write>data1:array<atomic<u32>>;
@group(0) @binding(3)var<storage,read_write>data2:array<i32>;
@compute @workgroup_size(16) fn r_512_16_3750_60000(@builtin(workgroup_id) gindex: vec3<u32>,@builtin(local_invocation_id) lindex: vec3<u32>) {
...
}
```

After:

```wgsl
fn nan() -> f32 { let bits = 0xffffffffu; return bitcast<f32>(bits); }
@group(0) @binding(0)
var<uniform> INFINITY : f32;

// 16 u8 = 4 u32
var<workgroup> temp0: array<atomic<u32>, 4>;

@group(0) @binding(1)var<storage,read_write>data0:array<atomic<u32>>;
@group(0) @binding(2)var<storage,read_write>data1:array<atomic<u32>>;
@group(0) @binding(3)var<storage,read_write>data2:array<i32>;
@compute @workgroup_size(16) fn r_512_16_3750_60000(@builtin(workgroup_id) gindex: vec3<u32>,@builtin(local_invocation_id) lindex: vec3<u32>) {
 ...
}
```

Metal for reference:

```msl
#include <metal_stdlib>
using namespace metal;
kernel void r_512_16_3750_60000(device unsigned char* data0, device unsigned char* data1, device int* data2, uint3 gid [[threadgroup_position_in_grid]], uint3 lid [[thread_position_in_threadgroup]]) {
  // unsigned char array of 16
  threadgroup __attribute__((aligned(16))) unsigned char temp0[16];
  ...
}
```


chenyuxyz:
hmm.. can you add a test that would have caught this?

wpmed92:
@chenyuxyz Added a test that checks all packed types, and a non-packed type, for shared memory, and checks if sizes are correct.

test/test_uops.py
```diff
@@ -320,6 +320,19 @@ def test_local_packed(self):
...
+  @unittest.skipUnless(Device.DEFAULT == "WEBGPU", "Test local memory size for packed data types")
+  def test_packed_smem_size(self):
+    # list of (dtype, size in dwords)
+    # half is not packed, so size is 16 as in the normal case of 32-bit types
+    types_w_sizes = [(dtypes.char, 4), (dtypes.uchar, 4), (dtypes.short, 8), (dtypes.ushort, 8), (dtypes.half, 16), (dtypes.float32, 16)]
+    # Define a local array of 16
+    temps = [UOp(Ops.DEFINE_LOCAL, dtype.ptr(size=16, local=True), (), 'smem') for (dtype,_) in types_w_sizes]
+    uops = to_uops_list(temps, opts=Device[Device.DEFAULT].renderer)
+    out = Device[Device.DEFAULT].renderer.render(uops)
+    for i, (dtype, size) in enumerate(types_w_sizes):
```
chenyuxyz: i'd write this similar to
```
for dtype in dtypes:
  temp = UOp(Ops.DEFINE_LOCAL, dtype.ptr(size=16, local=True), (), 'smem')
  uops = to_uops_list([temp], opts=Device[Device.DEFAULT].renderer)
  out = Device[Device.DEFAULT].renderer.render(uops)
  # some comment about half
  size = (packed) if dtype!=dtypes.half else (unpacked)
  self.assertIn(f"temp0: array<{Device[Device.DEFAULT].renderer.buf_map(dtype)},{size}>;", out)
```
to decouple cases between dtypes

wpmed92:
@chenyuxyz Just pushed the refactored test as you suggested.
As for if it fails at higher level: it didn't fail, but we allocated more shared memory than needed. 4x the memory needed for a uchar/char shared mem, and 2x for short/ushort.
This change ensures we allocate the right size for packed shared mem, and the test makes sure it doesn't break later.

chenyuxyz:
got it, thanks!

PR #10589, opened 2025-06-01
enable process replay diff assert in benchmarks [pr]

Qazalin:


Qazalin:
ASSERT_DIFF is 0 despite `[pr]` because update-benchmark push event doesn't have access to pull request title.

PR #10590, opened 2025-06-01
jit capture without realize (refactor WebGPU export: part 1)

hooved:
This is part 1 of https://github.com/tinygrad/tinygrad/pull/9424 (refactor WebGPU export).

This PR implements a new method of graph capture not dependent on `Tensor.realize`, in contrast to the current JIT.

A `GraphRenderer` instance captures all Tensor kernel graphs before and after calling the function that affects Tensors (i.e. the "traced function"). Calling `Tensor.realize` is forbidden within the traced function. The difference in graphs, before versus after, is sufficient to determine what compute occurred.

Implicit inputs are originally-existing BUFFER or ASSIGN root nodes that are contained in Tensor.lazydata that are created or modified by the traced function, after excluding explicit inputs. Knowing the original graph state also allows realization of the implicit input Tensors as they were at the start of the traced function. Such Tensors are stored in the `GraphRenderer` instance's `self.state_dict` for model export.

For testing and comparison to the current JIT, a minimal `jit_v2` is implemented, which works for capturing and running functions with combinations of explicit/implicit I/O.

Within https://github.com/tinygrad/tinygrad/pull/9424, this graph capture method works for compiling and exporting the WebGPU demos (efficientnet, yolov8, stable diffusion, tinychat llama). Separate PRs for WebGPU export refactor, WebGPU demo refactors, and WebGPU app validation/testing can follow in sequence after the current PR's functionality is merged.


hooved:
I'm closing this because more research is needed to specify what the new jit capture should do, and so that two different jit implementations do not need to be supported simultaneously.

PR #10591, opened 2025-06-02
Implementing BitNet

viraatdas:
Implementation of BitNet in Tinygrad

Still ongoing because the weight unpacking doesn't seem right.


chenyuxyz:
you can keep the example in your own repo. core tinygrad changes need to be isolated to smaller prs

PR #10595, opened 2025-06-02
Bounty: Fix METAL virtual device sync issue and reenable "Run LLaMA 7B on 4 virtual GPUs"

sgtBurns7:
Key Changes:

- added is_virtual variable to the function and additional synchronization primitives for virtual devices
- using a smaller command buffer pool for virtual devices to reduce memory pressure
- added aggressive synchronisation for virtual devices

### DIFF
--- a/test/test_metal.py
+++ b/test/test_metal.py
@@ -1,7 +1,12 @@
 import unittest
+import os
+import time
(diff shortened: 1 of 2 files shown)


geohot:
Do not use AI

PR #10601, opened 2025-06-02
Remote multihost p2p

uuuvn:


tinygrad/runtime/ops_remote.py
```diff
@@ -255,11 +282,14 @@ def _copyout(self, dest:memoryview, src:int):
...
-      src_dev.allocator._copyout(tmp:=memoryview(bytearray(sz)), src)
-      dest_dev.allocator._copyin(dest, tmp)
+    if dest_dev.conn != src_dev.conn:
+      dest_dev.q(Event(src_dev.session, start_event:=next(src_dev.event_num)))
+      src_dev.q(Wait(start_event))
+    src_dev.q(Transfer(src, dest_dev.session, dest))
+    if dest_dev.conn != src_dev.conn:
+      src_dev.q(Event(dest_dev.session, end_event:=next(dest_dev.event_num)))
+      dest_dev.q(Wait(end_event))
+    if DEBUG >= 2: dest_dev.conn.batch_submit()
```
uuuvn, line 290: Batch submit if DEBUG is here for accurate timings

tinygrad/runtime/ops_remote.py
```diff
@@ -155,30 +164,39 @@ async def __call__(self, reader:asyncio.StreamReader, writer:asyncio.StreamWrite
...
             rp = RemoteProperties(
-              real_device=dev.device, renderer=(cls.__module__, cls.__name__, args),
+              real_device=dev.device, renderer=(cls.__module__, cls.__name__, args), offset_supported=hasattr(dev.allocator, '_offset'),
               graph_supported=graph_cls is not None, graph_supports_multi=graph_cls is not None and issubclass(graph_cls, MultiGraphRunner),
-              transfer_supported=hasattr(dev.allocator, '_transfer'), offset_supported=hasattr(dev.allocator, '_offset'),
             )
             ret = repr(rp).encode()
+          case Event():
+            if c.session == c.event_session:
+              session.events[c.event].set()
```
uuuvn, line 190: `dev.synchronize()` here is not needed because everything is added to the same queue in the underlying device

tinygrad/runtime/ops_remote.py
```diff
@@ -54,7 +62,7 @@ class CopyIn(RemoteRequest): buffer_num: int; datahash: str # noqa: E702
 class CopyOut(RemoteRequest): buffer_num: int
 
 @dataclass(frozen=True)
-class Transfer(RemoteRequest): buffer_num: int; ssession: tuple[str, int]; sbuffer_num: int # noqa: E702
+class Transfer(RemoteRequest): buffer_num: int; dsession: SessionKey; dbuffer_num: int # noqa: E702
```
uuuvn: Transfer is now queued on the source because rdma reads are significantly slower than rdma writes - https://forums.developer.nvidia.com/t/why-do-larger-bytes-lead-to-smaller-bandwidth-in-ib-read-bw-testing/300274

(this pr is meant to be basically everything needed for ib but using normal tcp instead)

uuuvn: Transfers are also queued on source device in hcq, likely for similar reasons - https://github.com/tinygrad/tinygrad/blob/883bb4541c7173b18894f4ae2bf22bd0a02e1acd/tinygrad/runtime/support/hcq.py#L499-L512

tinygrad/runtime/ops_remote.py
```diff
@@ -21,7 +21,10 @@
 # ***** API *****
 
 @dataclass(frozen=True)
-class RemoteRequest: session: tuple[str, int]|None = field(default=None, kw_only=True)
```
uuuvn: Was on the edge about whether or not to submit this mini refactor separately, decided not to (but it's a separate commit for ease of review). Can submit separately if needed

The change itself is here because otherwise it's adding a third field to a tuple that should really just be a dataclass

Following the style of surrounding code with one line and noqa. I really don't like it and if you're going to accept a separate pr that removes all those noqas i'm going to do it

tinygrad/runtime/ops_remote.py
wozeparrot: yea this is fine

wozeparrot:
can you split the transfer queue change into its own pr?

tinygrad/runtime/ops_remote.py
wozeparrot: yea, I would split the sessionkey change into its own pr as well

uuuvn:
@wozeparrot prereqs have been merged, anything left to merge this?

PR #10603, opened 2025-06-02
removed 'cast' because type of 'unwrap(self._device)' is already 'str | tuple[str, ...]'

iharabukhouski:


chenyuxyz:
mypy failed

PR #10607, opened 2025-06-03
Adds BitNet linear layer implementation

viraatdas:
- `LinearBitNet`: Drop-in replacement for standard Linear layers
- Packed weights: `(out_features//4, in_features)` as `uint8`
- Unpacked weights: Ternary values as `float32`
- Weight scaling: Single `float32` scalar for proper scaling
- Bias: Standard `float32` (unquantized) for precision preservation

This is based on the Huggingface Transformer implementation for Bitnet's linear layer: https://github.com/huggingface/transformers/blob/096f25ae1f501a084d8ff2dcaf25fbc2bd60eba4/src/transformers/integrations/bitnet.py#L126

Huggingface model card for reference: https://huggingface.co/microsoft/bitnet-b1.58-2B-4T


viraatdas:
Hi - sorry about opening this PR - didn't mean to create this!

PR #10610, opened 2025-06-03
Fromfunction arange

S-Lykles:
@geohot so this is kind of the idea that I was talking to you about before
Basis idea is to have something like [`numpy.fromfunction`](https://numpy.org/doc/stable/reference/generated/numpy.fromfunction.html) that allows you to define a tensor as a `(function, shape)`. Where the function maps each index coordinate to some value. Then on top of this Tensor you can have any view you want (starting with the input shape) so that you can do all the movement ops as normal. Finally during the lowering you then use `unravel` and `x.st_arg.to_indexed_uops(ctx.idxs)` to get the indices into your input function.
This then allows you to define arange as `Tensor.fromfunction(lambda idx: idx.cast(dtype)*step+start, (output_len,))` and it is guaranteed to be fused. Something like `Tensor.eye` is also very simple to define using this. 

PR is a proof of concept at the moment but if this is something you like/want I can finish it.


S-Lykles:
Something like padding an arange is 2 kernels on master even with `FUSE_ARANGE=1` but 1 kernel with fromfunction

`Tensor.arange(10).pad((0, 6)).tolist()`

PR #10612, opened 2025-06-03
Implement linalg_eigh function for tensor eigenvalue decomposition in torch backend

L1aoXingyu:


chenyuxyz:
fine for now to make it run, but i don't quite see how tinygrad implements this yet

PR #10613, opened 2025-06-03
return inf on remote exec exceptions

roryclear:
BEAM with remote currently stops (from the remote device) if the host device throws an exception. This catches exceptions on program exec and returns inf to the host, which will then ignore these kernels

### DIFF
--- a/tinygrad/runtime/ops_remote.py
+++ b/tinygrad/runtime/ops_remote.py
@@ -204,7 +204,7 @@ class RemoteHandler:
             try:
               r = session.programs[(c.name, c.datahash)](*bufs, vals=c.vals, wait=c.wait, **extra_args)
               if r is not None: ret = str(r).encode()
-            except Exception as e:
+            except RuntimeError as e:
               if c.wait: ret = 'inf'.encode()
               else: raise e
           case GraphAlloc():


tinygrad/runtime/ops_remote.py
```diff
@@ -201,8 +201,12 @@ def handle(self, method:str, path:str, body:bytes) -> tuple[http.HTTPStatus, byt
           case ProgramExec():
             bufs = [session.buffers[x]._buf for x in c.bufs]
             extra_args = {k:v for k,v in [("global_size", c.global_size), ("local_size", c.local_size)] if v is not None}
-            r = session.programs[(c.name, c.datahash)](*bufs, vals=c.vals, wait=c.wait, **extra_args)
-            if r is not None: ret = str(r).encode()
+            try:
+              r = session.programs[(c.name, c.datahash)](*bufs, vals=c.vals, wait=c.wait, **extra_args)
+              if r is not None: ret = str(r).encode()
+            except Exception as e:
```
wozeparrot: think we only need to catch RuntimeErrors here?

roryclear: you're right, I've switched it

roryclear:
closing stale PR

PR #10616, opened 2025-06-03
specify metal threadgroup size at compile time

uuuvn:
(first commit of this pr is #10617)

previously the metal compiler would use some amount of registers, then when the program is created the api would return the maximum threadgroup size allowed given that amount of registers is used by a shader and throw if the local_size is more than that

this makes it do the limiting the other way around, tinygrad would add the attribute to the kernel source that it must be launchable with some amount of threads in a threadgroup and the metal compiler would limit itself to using the amount of registers available with that in mind

### DIFF
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -281,7 +281,7 @@ class MetalRenderer(CStyleLanguage):
   def __init__(self): self.tensor_cores = tc.metal if hasattr(os, 'uname') and os.uname().machine == "arm64" else []
 
   # language options
-  kernel_typedef = "kernel void"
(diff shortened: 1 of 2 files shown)


uuuvn:
Getting random `RuntimeError: Internal Error (0000000e:Internal Error)` with this...

geohot:
This good to merge? Looks like a great change, maybe it'll even speed things up!

Is there a failing test for this?

uuuvn:
`METAL=1 python examples/beautiful_mnist.py` will either fail with internal error (rarely some other error), or lock up the process (unkillable) or crash the entire ui. Never saw it go past the step 16.

PR #10618, opened 2025-06-03
Add BitNet linear layer implementation

viraatdas:
BitNet is a QAT model that uses ternary weights {-1, 0, 1}. 

This implements BitNet's linear layer as a drop-in replacement for standard Linear layers. The implementation packs 4 ternary weights into each uint8 byte and applies int8 activation quantization with per-token scaling for efficient inference.

## References

- **HuggingFace Implementation of BitLinear**: https://github.com/huggingface/transformers/blob/096f25ae1f501a084d8ff2dcaf25fbc2bd60eba4/src/transformers/integrations/bitnet.py#L126
- **Model Card**: https://huggingface.co/microsoft/bitnet-b1.58-2B-4T
- **Paper**: https://arxiv.org/pdf/2504.12285

### DIFF
--- a/.github/workflows/benchmark_search.yml
+++ b/.github/workflows/benchmark_search.yml
@@ -18,7 +18,7 @@ jobs:
       run: sudo rmmod amdgpu || true
     - name: Run SDXL with new search
       run: |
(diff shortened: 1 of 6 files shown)


wozeparrot:
I don't see why we would want this, its not used in any models in the repo, it should just be in its own examples repo

PR #10625, opened 2025-06-04
AMD_LLVM: support mfma for mi300x

b1tg:
model_train.py should work after this

### DIFF
--- a/tinygrad/renderer/llvmir.py
+++ b/tinygrad/renderer/llvmir.py
@@ -46,6 +46,9 @@ def render_wmma_amx(ctx, wmma: UOp) -> str:
 
 def render_wmma_amd(ctx, wmma: UOp) -> str:
   dt_map = {dtypes.half: "f16", dtypes.float: "f32", dtypes.bfloat16: "bf16", dtypes.ushort: "bf16"}
+  # https://github.com/llvm/llvm-project/blob/main/clang/test/CodeGenOpenCL/builtins-amdgcn-mfma.cl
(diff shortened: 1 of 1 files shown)


tinygrad/renderer/llvmir.py
```diff
@@ -46,6 +46,9 @@ def AMX(op, gpr): return f'call void asm sideeffect ".word (0x201000+($0<<5)+0$1
 
 def render_wmma_amd(ctx, wmma: UOp) -> str:
   dt_map = {dtypes.half: "f16", dtypes.float: "f32", dtypes.bfloat16: "bf16", dtypes.ushort: "bf16"}
+  # https://github.com/llvm/llvm-project/blob/main/clang/test/CodeGenOpenCL/builtins-amdgcn-mfma.cl
+  if ctx.arch.split(":")[0] == "gfx942": return f"  {ctx[wmma]} = call {ldt(wmma.dtype)} @llvm.amdgcn.mfma.{dt_map[wmma.src[-1].dtype.scalar()]}" + \
```
chenyuxyz: i think passing self into ctx and manually managing `self.r` defeats the purpose of making rewrites local

can it be done either through a macro function that's always valid here, or other more local approach

tinygrad/renderer/llvmir.py
```diff
@@ -186,10 +176,10 @@ def _render_kernel(self, uops: list[UOp], prefix:list[str]|None=None) -> tuple[t
           vc += 1
           r[u] = f"%v{vc}"
 
+        if u.op is Ops.WMMA: kernel.append(self._render_wmma(r, u))
```
chenyuxyz: not passing self is better, but now here's a op specific if that's bad.

what other approaches have you considered?

b1tg: passing arch through lambda argument

chenyuxyz:
verified that the beam setup works for bert on MI300X. full training breaks but it's not related to this pr

PR #10629, opened 2025-06-04
linearizer failure from FUSE_ARANGE default diff

Qazalin:
The linearizer seems to be fine without the CAST.
The real avg_pool2d kernel does not lower properly:
image
without CAST, this loop structure is fine, suggests there's a missing or_casted in the linearizer matchers?
image


Qazalin:
The same OptOps pass on WEBGPU and fail on other backends:
image

edit: because of split_load_store having different behavior if device supports_float4.

PR #10637, opened 2025-06-05
tests to motivate decoupling JIT graph capture from execution

hooved:
The current JIT capture executes the function to-be-captured twice, which constrains what graphs can be captured.

### DIFF
--- a/test/test_jit_failures.py
+++ b/test/test_jit_failures.py
@@ -1,0 +1,38 @@
+import unittest
+from tinygrad import TinyJit, Tensor, Device, dtypes
+
+def f16_is_unsupported():
+  assert (Tensor([1], dtype=dtypes.float32) + 1).tolist() == [2], "float32 kernel failed, there is an unexpected problem"
+  try: (Tensor([1], dtype=dtypes.float16) + 1).realize()
+  except: return True
+  return False
+
+class TestJitFailures(unittest.TestCase):
+
+  def test_capture_mutates_implicit_input(self):
+    w = Tensor([-2.0])
+
+    # Current JIT capture requires realizing the function's computation twice, which can "ruin" our state
+    # You can't just JIT this function from "two steps back" because tinygrad doesn't have complex numbers
+    # When exporting this function, if we want `w` to be [-2.0] like when we defined the function, we'll have to do extra manual work to fix `w`
+    @TinyJit
+    def f():
+      w.assign(w ** 2).realize()
+
+    for _ in range(2): f()
+    self.assertEqual(w.tolist(), [16.0])
+
+  # Currently you can't export f16 WebGPU models on dawn+Vulkan+NVIDIA, because f16 isn't supported in that stack
+  # See: https://issues.chromium.org/issues/42251215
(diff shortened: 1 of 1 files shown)


test/test_jit_failures.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @TinyJit
+    def f():
+      return x
+
+    with self.assertRaisesRegex(AssertionError, "didn't JIT anything!"):
+      for _ in range(2): f()
+
+  @unittest.skipUnless(Device.DEFAULT=="WEBGPU" and f16_is_unsupported(), "If realize is necessary, you can't capture kernels that won't run locally")
+  def test_capture_unsupported_dtype_kernel(self):
```
geohot: I really don't see this as an issue. That Tensor you created in the function is on WebGPU because that's the default. It should raise.

test/test_jit_failures.py
```diff
@@ -0,0 +1,57 @@
...
+    w = Tensor([-2.0])
+
+    @TinyJit
+    def f():
+      w.assign(w ** 2).realize()
+
+    for _ in range(2): f()
+    self.assertEqual(w.tolist(), [16.0])
+
+  def test_cannot_capture_closure(self):
```
geohot: This behavior also looks correct to me? Why should it rerun the same compute?

test/test_jit_failures.py
```diff
@@ -0,0 +1,57 @@
...
+    When exporting this function, if we want `w` to be [-2.0] like when we defined the function, we'll have to do extra manual work to fix `w`.
+    """
+    w = Tensor([-2.0])
+
+    @TinyJit
+    def f():
+      w.assign(w ** 2).realize()
+
+    for _ in range(2): f()
+    self.assertEqual(w.tolist(), [16.0])
```
geohot: What do you think the answer here should be? You ran the compute twice

hooved: The comment covers this. If we could capture and export the graph without executing the compute, then it'd be possible for the value of `w` to be the intial value of [-2.0]. But if compute has to be executed, then the value of `w` cannot be [-2.0] unless I write additional steps post-JIT-capture to copyin that data to the graph.

hooved:
"test_jit_failures" is a bad name for this module, I'll rename it.

The point is not that this is unintended behavior of the current JIT. Instead, the point is to document that you can't do as much as you otherwise could do, because of JIT capture depending on realize.

test/test_jit_failures.py
hooved: The behavior is correct. The point is to document how the JIT works. The reason this test is relevant is because it documents non-obvious behavior for a new tinygrad developer. A developer may be confused why the compute in the Tensor is not being captured, because they might not understand how the JIT realizes stuff once before capturing the graph.

If JIT capture didn't involve realize, the behavior would be different.

test/test_jit_failures.py
hooved: The exception is expected behavior based on the current JIT capture implementation, which requires the kernels to run locally. But as the comment describes in the context of exporting models, if I didn't have to run the kernels locally, I could still capture the graph and export the model. This would let me export f16 WebGPU models for use on mac/windows devices, even though I can't run them locally.

PR #10647, opened 2025-06-05
unique const

Qazalin:


Qazalin:
This also fixes https://github.com/tinygrad/tinygrad/issues/10937 - Need to find out why bert has more kernels.
Other kernel count tests get fixed when I dedup the CONST in the big graph.

Qazalin:
We probably still don't want this behavior...
Constants are shared, if you want to ASSIGN to a const Tensor, it should be first realized to a buffer.
Though it's pretty confusing when it happens.

PR #10666, opened 2025-06-06
Fix gpt2 OOB

S-Lykles:
Im assuming that llama and whisper have a similar problem, I'll try to get around to them tomorrow and then `IGNORE_OOB=0` can be enabled in CI for models


chenyuxyz:
for documentation, can you add a command that this pr would fix?

S-Lykles:
sure, `IGNORE_OOB=0 PYTHONPATH=. python3 examples/gpt2.py`

PR #10677, opened 2025-06-07
remove unnecessary 'argfix' because 'view' is an alias to 'reshape'. all functionality must be inside 'reshape'

iharabukhouski:


chenyuxyz:
can you add a few tests here https://github.com/tinygrad/tinygrad/blob/e3805171e238d8d328cb1a872091cd68b6e6794a/test/test_ops.py#L1860
that tinygrad call does not use a parentheses

iharabukhouski:
> can you add a few tests here
> 
> https://github.com/tinygrad/tinygrad/blob/e3805171e238d8d328cb1a872091cd68b6e6794a/test/test_ops.py#L1860
> 
> 
> that tinygrad call does not use a parentheses

in the second commit added for 'view' the same set of tests as for 'reshape' (since 'view' is just an alias for 'reshape')

chenyuxyz:
don't add the same set of tests, it's wasteful for trivial stuff. just add one or two simple case that this pr triggers

iharabukhouski:
@chenyuxyz would you please take another look?
i have added 5 completely new tests for the 'view' op
(hopefully i understood you correctly)

PR #10683, opened 2025-06-07
viz: clicking a KERNEL node can open codegen rewrite

Qazalin:
https://github.com/user-attachments/assets/99fc98d8-84a7-41c9-9e30-a95c503a9de1


Qazalin:
Something like this,
image
The colors are kind of distracting though

edit: much better without the colors
image

PR #10685, opened 2025-06-07
Fix whisper OOB

S-Lykles:
Im not sure exactly what the `+5` in 
```python
self.max_self_attn_cache_len = self.max_tokens_to_sample * 2 + 5  # roughly prompt + start toks + max_tokens_to_sample
``` 
Is supposed to be, but  I think it was maybe a workaround for Variable being off by one


chenyuxyz:
cool thanks!

Issue #10340, opened 2025-05-15
masked_select hangs indefinitely with large tensor inputs


L1aoXingyu:
When using the tiny backend for HLB-CIFAR10 torch training, I encountered a hang issue upon launching the training script. By isolating each component and testing individually, I discovered the issue originates from the masked_select function.

Further examination revealed that torch.masked_select internally calls tinygrad's masked_select. Testing this independently confirmed the problem occurs when processing large tensor sizes, causing the operation to hang indefinitely.

Below is a simplified reproduction test case:

```python
from tinygrad import Tensor, dtypes

def make_random_square_masks(inputs: Tensor, mask_size: int):
    if mask_size == 0:
        return None
    is_even = int(mask_size % 2 == 0)
    in_shape = inputs.shape

    mask_center_y = Tensor.randint(in_shape[0], low=mask_size//2-is_even, high=in_shape[-2]-mask_size//2-is_even + 1, device=inputs.device, dtype=dtypes.long)
    mask_center_x = Tensor.randint(in_shape[0], low=mask_size//2-is_even, high=in_shape[-1]-mask_size//2-is_even + 1, device=inputs.device, dtype=dtypes.long)

    # measure distance, using the center as a reference point
    to_mask_y_dists = Tensor.arange(in_shape[-2], device=inputs.device, dtype=dtypes.long).reshape(1, 1, in_shape[-2], 1) - mask_center_y.reshape(-1, 1, 1, 1)
    to_mask_x_dists = Tensor.arange(in_shape[-1], device=inputs.device, dtype=dtypes.long).reshape(1, 1, 1, in_shape[-1]) - mask_center_x.reshape(-1, 1, 1, 1)

    to_mask_y = (to_mask_y_dists >= (-(mask_size // 2) + is_even)) * (to_mask_y_dists <= mask_size // 2)
    to_mask_x = (to_mask_x_dists >= (-(mask_size // 2) + is_even)) * (to_mask_x_dists <= mask_size // 2)
    final_mask = to_mask_y * to_mask_x ## Turn (y by 1) and (x by 1) boolean masks into (y by x) masks through multiplication. Their intersection is square, hurray! :D

    return final_mask


B, C, H_img, W_img = 10000, 3, 36, 36 # Batch, Channels, Padded Height, Padded Width
crop_size = 32
input_tensor = Tensor.randn(B, C, H_img, W_img).half()

mask = make_random_square_masks(input_tensor, crop_size)

[...]

PR #10701, opened 2025-06-08
Fix large tensor stalling when using masked_select

L1aoXingyu:
Since masked_select is only tested with forward_only=True, it’s safe to use NumPy and break graph preservation. The current implementation consumes excessive memory with large tensors, causing the program to stall.

Ref to https://github.com/tinygrad/tinygrad/issues/10340

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -1624,10 +1624,16 @@ class Tensor(MathTrait):
     """
     if not dtypes.is_bool(mask.dtype): raise RuntimeError(f"masked_select expects bool mask tensor, got {mask.dtype}")
     x, mask = self.flatten(), mask._broadcast_to(self.shape).flatten()
-    mask_cumsum = mask.cumsum()
-    counts = Tensor.zeros(mask_cumsum[-1].item(), dtype=dtypes.int32)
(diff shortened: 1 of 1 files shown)


chenyuxyz:
tinygrad works without numpy and should stay that way. issues like this needs to be understood and fixed in tinygrad. you can add failing tests with expectedFailure and TODO notes to start with

L1aoXingyu:
In the hlb-cifar10 repo, `masked_select` is used to preprocess large tensors like this:
```python
def batch_crop(inputs, crop_size):
    with torch.no_grad():
        crop_mask_batch = make_random_square_masks(inputs, crop_size)
        crop_mask_batch = crop_mask_batch.expand((-1, 3, -1, -1)) 
        cropped_batch = torch.masked_select(inputs, crop_mask_batch).view(inputs.shape[0], inputs.shape[1], crop_size, crop_size)
        return cropped_batch
```
Here, inputs has shape (50000, 3, 36, 36) and crop_mask_batch has shape (50000, 3, 32, 32). This causes masked_select to stall due to high memory usage.
Is it acceptable to use NumPy in torch backend as a workaround to avoid this issue? @chenyuxyz

chenyuxyz:
this is not avoiding the issue, it just hides the issue

PR #10703, opened 2025-06-08
Chunked Implementation of masked_select for large tensor

L1aoXingyu:


geohot:
Why is this faster? This stuff probably shouldn't be in tensor.py, there should be a way to do this at a lower level

L1aoXingyu:
> Why is this faster? This stuff probably shouldn't be in tensor.py, there should be a way to do this at a lower level

The chunked version fixes the stall issue with masked_select on large tensors. Without chunking, it stalls; with chunking, it runs—albeit very very slowly.
Also, could you clarify what you mean by “shouldn't be in tensor.py”? I'd appreciate an example of a lower-level implementation I can learn from.

PR #10716, opened 2025-06-09
visualize tile script

ignaciosica:
This still needs some cleanup/docs. I'm also planning on writing this week some docs on tiled-algos/kernel-opts/tc using this tool.

Also, for small enough shapetrackers, the viz might also be useful, I'll play with it and if I find it useful I'll upstream it too.

One way of using this script is by plugging it within the linearize pipeline in the following way:
``` diff
  @track_rewrites()
  def linearize(self, name_override:Optional[str]=None, ast_transform:Optional[Callable]=None) -> Kernel:
    [...]
    modified_ast = self.get_optimized_ast(name_override)
    if ast_transform is not None: modified_ast = ast_transform(self, modified_ast)
+   from extra import viz_tile              # 1 import helper
+   viz_tile.viz_tile(self, modified_ast)   # 2 show tiles
    [...]
```

### Examples
`TC=1`
image
`TC=2 highlighting thread 7`
image
`TC=3`
image

### DIFF
--- a/extra/viz_tile.py
+++ b/extra/viz_tile.py
@@ -6,54 +6,71 @@ from tinygrad.shape.shapetracker import ShapeTracker
 from tinygrad.shape.view import View, unravel, canonicalize_strides
 from tinygrad.codegen.kernel import Kernel
 
+# TODO: address masking/padding
+# TODO: generic shapetracker layout viz
+# TODO: remove tabulate dependency
+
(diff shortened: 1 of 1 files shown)


geohot:
I like this idea! How can we integrate into VIZ?

chenyuxyz:
stale

Issue #10723, opened 2025-06-09
infinite looping in top_down_rewrite


wozeparrot:
```py
  def test_expand_padded(self):
    for i in range(1, 5):
      vi = Variable("i", 1, 10).bind(i)
      a = Tensor(1).unsqueeze(0).pad((0, 24)).unsqueeze(0)
      symbolic = a.expand(vi, 25).reshape(i, 25).numpy()
      expected = a.expand(i, 25).numpy()
      np.testing.assert_allclose(symbolic, expected, atol=1e-6, rtol=1e-6)
```

traceback:
```
Traceback (most recent call last):
  File "/home/woze/tinygrad/test/test_symbolic_ops.py", line 265, in <module>
    unittest.main()
  File "/usr/lib/python3.12/unittest/main.py", line 105, in __init__
    self.runTests()
  File "/usr/lib/python3.12/unittest/main.py", line 281, in runTests
    self.result = testRunner.run(self.test)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/unittest/runner.py", line 240, in run
    test(result)
  File "/usr/lib/python3.12/unittest/suite.py", line 84, in __call__
    return self.run(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/unittest/suite.py", line 122, in run
    test(result)
  File "/usr/lib/python3.12/unittest/suite.py", line 84, in __call__
    return self.run(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/unittest/suite.py", line 122, in run
    test(result)
  File "/usr/lib/python3.12/unittest/case.py", line 690, in __call__
    return self.run(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/unittest/case.py", line 634, in run
    self._callTestMethod(testMethod)
  File "/usr/lib/python3.12/unittest/case.py", line 589, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/home/woze/tinygrad/test/test_symbolic_ops.py", line 173, in test_expand_padded
    symbolic = a.expand(vi, 25).reshape(i, 25).numpy()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10726, opened 2025-06-09
don't merge shape symbolic reshape symbolic

chenyuxyz:


chenyuxyz:
https://github.com/tinygrad/tinygrad/issues/10723 does not hang, but the output is still off by 1 because the mask is not expanded

chenyuxyz:
no diff for existing kernels

PR #10736, opened 2025-06-09
Update autogen ci runner to ubuntu 24.04

uuuvn:
For `kfd.AMDKFD_IOC_EXPORT_DMABUF`


.github/actions/setup-tinygrad/action.yml
```diff
@@ -157,7 +157,7 @@ runs:
         fi
         # **** WebGPU (dependencies for software-based vulkan) ****
         if [[ "${{ inputs.webgpu }}" == "true" ]]; then
-          pkgs+=" libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers"
+          pkgs+=" libgl1 libglx-mesa0 libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers"
```
uuuvn, line 160: `libegl1-mesa` is a "transitional dummy package" that is removed in 23.10 and up. It's equivalent to `libgl1 libglx-mesa0`

https://launchpad.net/ubuntu/jammy/amd64/libgl1-mesa-glx/22.0.1-1ubuntu2

autogen_stubs.sh
```diff
@@ -83,6 +83,7 @@ generate_kfd() {
   sed -i "/import functools/a from tinygrad.runtime.support.hcq import FileIOInterface" $BASE/kfd.py
   sed -i "s/def _do_ioctl(__idir, __base, __nr, __user_struct, __fd, \*\*kwargs):/def _do_ioctl(__idir, __base, __nr, __user_struct, __fd:FileIOInterface, \*\*kwargs):/g" $BASE/kfd.py
   sed -i "s/fcntl.ioctl(__fd, (__idir<<30)/__fd.ioctl((__idir<<30)/g" $BASE/kfd.py
+  sed -i "s/!!/not /g" $BASE/kfd.py
```
uuuvn: https://github.com/tinygrad/tinygrad/actions/runs/15547245566/job/43770999652?pr=10736#step:6:83

uuuvn:
image

.github/actions/setup-tinygrad/action.yml
```diff
@@ -118,8 +118,8 @@ runs:
       shell: bash
       run: |
         wget https://repo.radeon.com/rocm/rocm.gpg.key -O - | gpg --dearmor | sudo tee /etc/apt/keyrings/rocm.gpg > /dev/null
-        sudo tee /etc/apt/sources.list.d/rocm.list <<'EOF'
-        deb [arch=amd64 signed-by=/etc/apt/keyrings/rocm.gpg] https://repo.radeon.com/rocm/apt/6.1.2 jammy main
+        sudo tee /etc/apt/sources.list.d/rocm.list <<EOF
+        deb [arch=amd64 signed-by=/etc/apt/keyrings/rocm.gpg] https://repo.radeon.com/rocm/apt/6.2 $(lsb_release -cs) main
```
uuuvn, line 122: 6.2 is the earliest version that has support for 24.04

.github/workflows/test.yml
```diff
@@ -96,19 +96,21 @@ jobs:
         cuda: 'true'
         webgpu: 'true'
         llvm: 'true'
+    - name: Install libclang-14 for clang2py
+      run: sudo apt-get install -y --no-install-recommends libclang-14-dev
```
uuuvn: ubuntu 22.04 github ci images have clang 13-15 preinstalled: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#language-and-runtime

ubuntu 24.04 github ci images have clang 16-18 installed: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime

which does this:
image

not adding this to setup env action as it's only for autogen

.github/workflows/test.yml
```diff
@@ -96,19 +96,21 @@ jobs:
         cuda: 'true'
         webgpu: 'true'
         llvm: 'true'
+    - name: Install autogen support packages
+      run: sudo apt-get install -y --no-install-recommends llvm-14-dev libclang-14-dev
```
uuuvn, line 100: I also added llvm-14-dev to this (correct version was preinstalled on 22.04 github images similar to clang thing) which is needed to autogen llvm headers, but not for runtime which can use any llvm ver (so not adding to setup action either)

autogen_stubs.sh
nimlgen: !! is used to cast to bool prob

autogen_stubs.sh
```diff
@@ -83,11 +83,12 @@ generate_kfd() {
...
   python3 -c "import tinygrad.runtime.autogen.kfd"
 }
 
 generate_cuda() {
-  clang2py /usr/include/cuda.h -o $BASE/cuda.py -l /usr/lib/x86_64-linux-gnu/libcuda.so
+  clang2py /usr/include/cuda.h --clang-args="-D__CUDA_API_VERSION_INTERNAL" -o $BASE/cuda.py -l /usr/lib/x86_64-linux-gnu/libcuda.so
```
uuuvn, line 91: CUDA 12 makes some weird symbol versioning changes that break everything unless i do this

https://github.com/tinygrad/tinygrad/actions/runs/15559505609/job/43808064398#step:7:19

autogen_stubs.sh
uuuvn: changed to `not not `

uuuvn:
Should be fine to merge as long as real cuda is tested (the `__CUDA_API_VERSION_INTERNAL` thing is a bit sketchy)

uuuvn:
> Should be fine to merge as long as real cuda is tested (the `__CUDA_API_VERSION_INTERNAL` thing is a bit sketchy)

I'm a bit upset that when i ask to test on real hardware before merge because i suspect that something like #10776 might happen i just get ignored

I don't have any nvidia hw to test, i went manually through the functions to verify that `__CUDA_API_VERSION_INTERNAL` fixes them all assuming that it + mypy (that initially caught the disappearance of v1 functions) + my request to test on real hw before merge would be enough to not break things but apparently not

Going forward will probably just not set ready to review until someone responded that they actually tested the thing when i don't have the hardware instead of relying on them noticing the message

PR #10739, opened 2025-06-10
linearizer failure from wino fuse arange cifar

chenyuxyz:
somehow only failed for AM and NV
`RuntimeError: UOp verification failed at 4 on Ops.LOAD dtypes.half 2 [<Ops.DEFINE_GLOBAL: 24>, <Ops.VIEW: 23>] None`


chenyuxyz:
oh because constraint on local
`tinygrad.codegen.kernel.KernelOptError: exceeds maximum shared memory size: needs 36864, max 32768`

PR #10761, opened 2025-06-10
early stop _time_program on tightly coupled tms

ignaciosica:
draft until `benchmark_search` validates this change makes no regression

### DIFF
--- a/tinygrad/engine/search.py
+++ b/tinygrad/engine/search.py
@@ -52,6 +52,7 @@ def _time_program(p:ProgramSpec, lib:bytes, var_vals:dict[Variable, int], rawbuf
         with Context(DEBUG=0, BEAM=0, CAPTURING=0, TRACK_MATCH_STATS=0): Tensor.ones(1024,1024).contiguous().realize(do_update_stats=False)
     tms.append(cast(float, car(input_bufs, var_vals, wait=True))*factor)
     if early_stop is not None and early_stop < min(tms): break
+    if len(tms) == 2 and abs(tms[0] - tms[1]) <= 0.05 * (tms[0] + tms[1]): break
   return tms
 
 class TimeoutException(Exception): pass


tinygrad/engine/search.py
```diff
@@ -52,6 +52,7 @@ def _time_program(p:ProgramSpec, lib:bytes, var_vals:dict[Variable, int], rawbuf
         with Context(DEBUG=0, BEAM=0, CAPTURING=0, TRACK_MATCH_STATS=0): Tensor.ones(1024,1024).contiguous().realize(do_update_stats=False)
     tms.append(cast(float, car(input_bufs, var_vals, wait=True))*factor)
     if early_stop is not None and early_stop < min(tms): break
+    if len(tms) == 2 and abs(tms[0] - tms[1]) <= 0.05 * (tms[0] + tms[1]): break
```
chenyuxyz, line 55: not really readable. what's the problem that this is trying to solve?

ignaciosica: sorry about that; the idea is that for kernels that the first two runs are already tightly couple, the difference between them is less than 10% of the mean, you can skip your third run; I saves on the first iterations of beam where some kernels are really expensive to run.

0.5 val is pretty arbitrary, it can be an env var

chenyuxyz: how much does it save? would be easier to justify this complexity if there are real wins

measuring time is hard, and i think we want to first make sure we can measure correctly, then think about how to make the measurement fast (for example understand how useful `clear_l2` and `allow_test_size` are)

ignaciosica: Yes I agree.

I had this change in mind for a while and only now I opened it upstream to test with the new benchmark_search ci job. I did see some marginal wins but they are anecdotal, nothing tested or rigorously benchmarked.

So yeah, maybe not worth it

ignaciosica:
I don't think the results are entirely related to the change in the pr, however, based on them and the discussion above, I decide to close 
image

PR #10762, opened 2025-06-10
fix onnx ci

b1tg:


chenyuxyz:
what does this fix?

b1tg:
onnx test is flaky and be reported in discord few times today , can't reproduce locally, but from the [backtrace](https://github.com/tinygrad/tinygrad/actions/runs/15560238034/job/43810450234), it's looks like a io sync issue, trig ci in my branch few times, it not happen again after this

chenyuxyz:
oh cool, thanks for looking into this.

PR #10770, opened 2025-06-11
remove Tensor.test() in retinanet

chenyuxyz:
test was removed


chenyuxyz:
OOM at the start of eval anyway...

PR #10775, opened 2025-06-11
Support symbolic slice with no start [pr]

S-Lykles:
This now allows symbolic slicing with no start and also a step of 1
`a[3:5, :vi:1]` and `a[3:5, :vi]`


tinygrad/tensor.py
```diff
@@ -1143,19 +1143,19 @@ def _getitem(self, indices, v: Tensor|None = None) -> Tensor:
...
+          boundary, stride = [start, stop], step
+          if all(isinstance(s, int) for s in (start,stop,step)):
             # handle int slicing
             *boundary, stride = index.indices(cast(SupportsIndex, size))
             if stride * (boundary[1] - boundary[0]) < 0: boundary = [0, 0]
             elif stride < 0: boundary = [boundary[1] + 1, boundary[0] + 1]
             # update size for slice
             size = ceildiv((boundary[1] - boundary[0]), abs(stride))
-          elif index.step is None and all(isinstance(s,(int,UOp))for s in (index.start,index.stop)) and resolve((index.stop-index.start) > 0, False):
+          elif (step == 1) and isinstance(step, int) and all(isinstance(s,(int,UOp)) for s in (start, stop)) and resolve((stop-start) > 0, False):
```
chenyuxyz, line 1156: `isinstance(step, int)` must be true if `step==1`? (technically maybe a const Variable but that should be correct too)

S-Lykles: The ==1 would compare equal to 1.0 as well, explicitly forbidding floats even if they are whole numbers is more consistent with python was my thinking

chenyuxyz: oh i see... yea i agree.

i think it's better to add a input validation for slice (int or UOp that has integer dtypes) earlier, and test that slice whole float should fail

Issue #10774, opened 2025-06-11
CUDA JIT broken by Update autogen ci runner to ubuntu 24.04 #10736


xl0:
Ubuntu 24.04.2 LTS
CUDA 12.8 (from conda)

The commit 1d15374c7aec1f9c962ac2156f318e196f99f5de broke this.

```
$ CUDA=1 PYTHONPATH=. CACHELEVEL=0 python examples/hlb_cifar10.py 
shuffling training dataset in 4538.70 ms (epoch=0)
  0 18939.09 ms run, 18937.70 ms python,    1.39 ms CUDA, 1185.38 loss, 0.000015 LR, 0.89 GB used,     34.76 GFLOPS,    658.26 GOPS
  1 4959.52 ms run, 4958.22 ms python,    1.30 ms CUDA, 1185.25 loss, 0.000030 LR, 2.09 GB used,    132.04 GFLOPS,    654.88 GOPS
Traceback (most recent call last):
  File "/home/xl0/work/projects/grads/tinygrad/examples/hlb_cifar10.py", line 429, in <module>
    train_cifar()
  File "/home/xl0/work/projects/grads/tinygrad/examples/hlb_cifar10.py", line 404, in train_cifar
    loss = train_step_jitted(model, optim.OptimizerGroup(opt_bias, opt_non_bias), [lr_sched_bias, lr_sched_non_bias], X, Y)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xl0/work/projects/grads/tinygrad/tinygrad/engine/jit.py", line 329, in __call__
    ret = self.captured(input_buffers, var_vals)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xl0/work/projects/grads/tinygrad/tinygrad/engine/jit.py", line 201, in __call__
    self._jit_cache = apply_graph_to_jit(self.jit_cache, input_buffers, var_vals, max_batch_size=getenv("JIT_BATCH_SIZE", 32))
[...]

PR #10776, opened 2025-06-11
Explicitly use CUDA_KERNEL_NODE_PARAMS v1

uuuvn:
CUDA 12 changed `CUDA_KERNEL_NODE_PARAMS` to always be an alias to `CUDA_KERNEL_NODE_PARAMS_v2` instead of `CUDA_KERNEL_NODE_PARAMS_v1`

(looks like `__CUDA_API_VERSION_INTERNAL` hack only fixes functions)

Also it looks like there is no cuda tests that trigger graph even in benchmark???


uuuvn:
closes #10774 assuming there is no other similar things that i missed

nimlgen:
tested, looks fixed

nimlgen:
> Also it looks like there is no cuda tests that trigger graph even in benchmark???

benchmarks is not for tests. we need to extend MOCKCUDA with graph emulation

PR #10728, opened 2025-06-09
[bounty] examples/hlb_cifar10.py dataset augments without numpy (random_crop + cutmix)

IntendedConsequence:
With CUDA=1 and DEFAULT_FLOAT=HALF, on an RTX 3080, using FUSE_ARANGE=1 and LATEWINO=1, running the entire training, and measuring the runtime of the entire command:

Before: 116.59s total
After: ~108s total

FUSE_ARANGE=1 speeds up indexing dataset by a factor of 10, without it each epoch starts with a 5 second step. I tried spreading the shuffling and augmentation over the steps but that just added a flat ~30ms overhead to each step.

LATEWINO=1 introduces a 40 second initial delay, but massively speeds up the entire training, without it it takes 172.42s, adding 64 seconds. With LATEWINO=1 shuffling the dataset at the start of each epoch is near instant, adding ~48 GOPS to the first step.

(Running with LATEWINO=1 and without FUSE_ARANGE makes the first step of each epoch take 3 seconds, adding ~30s to overall training time, 143s total).

~~I also had to hack in a fix for CUDA on windows, these files seem to be autogenerated by clang2py tool and should probably be dealt with there, or moved into a patch. Since this is my working branch and it's a draft PR I'll remove it if it ever comes to a review.~~ Removed since they were failing CI.

First 20 steps:
```
shuffling training dataset in 2703.00 ms (epoch=0)
  0 38344.00 ms run, 38344.00 ms python,    0.00 ms CUDA, 1188.00 loss, 0.000015 LR, 0.74 GB used,      6.73 GFLOPS,    258.21 GOPS
  1 4187.00 ms run, 4172.00 ms python,   15.00 ms CUDA, 1181.00 loss, 0.000030 LR, 1.52 GB used,     49.50 GFLOPS,    207.27 GOPS
  2   79.00 ms run,   32.00 ms python,   47.00 ms CUDA, 1182.00 loss, 0.000045 LR, 1.52 GB used,   2623.73 GFLOPS,    207.27 GOPS
  3   62.00 ms run,   15.00 ms python,   47.00 ms CUDA, 1173.00 loss, 0.000060 LR, 1.52 GB used,   3343.14 GFLOPS,    207.27 GOPS
  4   47.00 ms run,    0.00 ms python,   47.00 ms CUDA, 1175.00 loss, 0.000075 LR, 1.52 GB used,   4410.10 GFLOPS,    207.27 GOPS
[...]

PR #10777, opened 2025-06-11
[bounty] Don't use numpy inside hlb_cifar10 training loop

xl0:
A deep dive: https://xl0.github.io/tinygrad-notes/bounty1.html

So the main issue here was the fact that `Tensor.masked_select()` is so slow that it's unusable. It's a natural choice for implementing `random_crop`, but not the only option.

Since `random_crop` picks a 32x32 rectangle from a 36x36 source image, there are only 16 possible crops. So I shuffle the dataset, split it 16-way, apply all possible crops using `.pad()` with negative padding, put it togeher and shuffle it again.

The result is not as straight-forward as `X.masked_select(mask)`, but it is actually fast - faster than the numpy implementation at least. Maybe we can use `.masked_select()` once it get good.

Another issue - indexing the dataset with a large 1-dim tensor to shuffle it is super slow, unless done with FUSE_ARANGE. Then it's actually quite fast.

@geohot do you want me to add FUSE_ARANGE to the CI flows for this example? Without it it gets slower.

`cutmix` did not pose difficulty, and I also removed any numpy from the trainig loop. There is still numpy in `whitening()`, and I can remove it too. @geohot do you want all trace of numpy gone, or should I do it in another PR?

In total, on my machine, I get 110s -> 102s, so ~7% performance improvement. No difference in training accuracy.

I also noticed that @IntendedConsequence made a good effort in #10728. I'm not sure how to best handle the situation, because I saw his half-way through. I like the training loop cleanup he did, but feel it's a separate PR.

### DIFF
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -244,10 +244,10 @@ def train_cifar():
         if getenv("CUTMIX", 1) and step >= hyp['net']['cutmix_steps']: X, Y = X_cm, Y_cm
       et = time.monotonic()
       print(f"shuffling {'training' if is_train else 'test'} dataset in {(et-st)*1e3:.2f} ms ({epoch=})")
-      vi = Variable("i", 0, X.shape[0]-1)
-      for i in range(0, X.shape[0], BS):
+
+      vi = Variable("i", 0, (full_batches := (X.shape[0] // BS) * BS) - BS)
+      for i in range(0, full_batches, BS):
         step += 1
-        if i+BS > X.shape[0]: break
         vib = vi.bind(i)
         yield X[vib:vib+BS], Y[vib:vib+BS]
       epoch += 1
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -245,15 +245,14 @@ def train_cifar():
       if is_train:
         X, Y = jittable_transforms(X, Y)
         if getenv("CUTMIX", 1) and step >= hyp['net']['cutmix_steps']:
-            X, Y = cutmix(X, Y, mask_size=hyp['net']['cutmix_size'])
+          X, Y = cutmix(X, Y, mask_size=hyp['net']['cutmix_size'])
       et = time.monotonic()
       print(f"shuffling {'training' if is_train else 'test'} dataset in {(et-st)*1e3:.2f} ms ({epoch=})")
       for i in range(0, X.shape[0], BS):
-        # pad the last batch  # TODO: not correct for test
-        batch_end = min(i+BS, Y.shape[0])
         step += 1
+        if i+BS > X.shape[0]: break
         # This needs to be contiguous, or the jitted consumer will fail.
-        yield X[batch_end-BS:batch_end].contiguous(), Y[batch_end-BS:batch_end].contiguous()
+        yield X[i:i+BS].contiguous(), Y[i:i+BS].contiguous()
       epoch += 1
       if not is_train: break
 
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -215,16 +215,6 @@ def train_cifar():
     Xs, Ys, Xi, Yi = make_random_crop_indices(X.shape, crop_size)
     return X.gather(-1, (Xs + Xi).expand(-1, 3, X.shape[2], -1)).gather(-2, ((Ys+Yi).expand(-1, 3, crop_size, crop_size)))
 
-  @TinyJit
-  def jittable_transforms(X:Tensor, Y:Tensor):
-    perms = Tensor.randperm(X.shape[0], device=X.device)
-    if getenv("RANDOM_CROP", 1):
-      X = random_crop(X, crop_size=32)
-    if getenv("RANDOM_FLIP", 1):
-      X = (Tensor.rand(X.shape[0],1,1,1) < 0.5).where(X.flip(-1), X) # flip LR
-    return X[perms], Y[perms], perms # perms is expensive to calculate, and we can reuse it.
-
-  @TinyJit
   def cutmix(X, Y, order, mask_size=3):
     mask = make_square_mask(X.shape, mask_size)
     X_patch, Y_patch = X[order], Y[order]
@@ -233,6 +223,16 @@ def train_cifar():
     Y_cutmix = mix_portion * Y_patch + (1. - mix_portion) * Y
     return X_cutmix, Y_cutmix
 
+  @TinyJit
+  def augmentations(X:Tensor, Y:Tensor):
+    perms = Tensor.randperm(X.shape[0], device=X.device) # We reuse perms for cutmix, because they are expensivne to generate
+    if getenv("RANDOM_CROP", 1):
+      X = random_crop(X, crop_size=32)
+    if getenv("RANDOM_FLIP", 1):
+      X = (Tensor.rand(X.shape[0],1,1,1) < 0.5).where(X.flip(-1), X) # flip LR
+    X, Y = X[perms], Y[perms]
+    return X, Y, *cutmix(X, Y, perms, mask_size=hyp['net']['cutmix_size'])
+
   # the operations that remain inside batch fetcher is the ones that involves random operations
   def fetch_batches(X_in:Tensor, Y_in:Tensor, BS:int, is_train:bool):
     step, epoch = 0, 0
@@ -240,9 +240,8 @@ def train_cifar():
       st = time.monotonic()
       X, Y = X_in, Y_in
       if is_train:
-        X, Y, perms = jittable_transforms(X, Y)
-        if getenv("CUTMIX", 1) and step >= hyp['net']['cutmix_steps']:
-          X, Y = cutmix(X, Y, perms, mask_size=hyp['net']['cutmix_size'])
+        X, Y, X_cm, Y_cm = augmentations(X, Y)
+        if getenv("CUTMIX", 1) and step >= hyp['net']['cutmix_steps']: X, Y = X_cm, Y_cm
       et = time.monotonic()
       print(f"shuffling {'training' if is_train else 'test'} dataset in {(et-st)*1e3:.2f} ms ({epoch=})")
       for i in range(0, X.shape[0], BS):
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -201,19 +201,25 @@ def train_cifar():
     idx_y = Tensor.arange(H, dtype=dtypes.int32).reshape((1,1,H,1))
     return (idx_x >= low_x) * (idx_x < (low_x + mask_size)) * (idx_y >= low_y) * (idx_y < (low_y + mask_size))
 
-  def random_crop(X:Tensor, crop_size=32):
-    mask = make_square_mask(X.shape, crop_size)
-    mask = mask.expand((-1,3,-1,-1))
-    X_cropped = Tensor(X.numpy()[mask.numpy()])
-    return X_cropped.reshape((-1, 3, crop_size, crop_size))
 
-  def cutmix(X:Tensor, Y:Tensor, mask_size=3):
-    # fill the square with randomly selected images from the same batch
+  def negative_pad_grid(n=4):
+      # Generate all possible padding combinations that sum to biting n elements from h and w.
+      pad_options = [(-i, i-n) for i in range(n+1)]
+
+      # Create all possible combinations of padding for left/right and top/bottom
+      return [(pad_options[x % len(pad_options)], pad_options[y % len(pad_options)])
+              for x in range(n) for y in range(n)]
+
+  def random_crop(X:Tensor, crop_size:int=32) -> Tensor:
+      assert X.shape[-1] >= crop_size
+      Xl = X.split(X.shape[0]//16) # XXX This results in B=3125, which is prob not great for performance, not sure if matters.
+      Xp = [ x.pad(((0,0),(0, 0), *npad)) for x, npad in zip(Xl, negative_pad_grid(X.shape[-1] - crop_size)) ]
+      return Tensor.cat(*Xp)
+
+  def cutmix(X, Y, mask_size=3):
     mask = make_square_mask(X.shape, mask_size)
-    order = list(range(0, X.shape[0]))
-    random.shuffle(order)
-    X_patch = Tensor(X.numpy()[order], device=X.device, dtype=X.dtype)
-    Y_patch = Tensor(Y.numpy()[order], device=Y.device, dtype=Y.dtype)
+    order = Tensor.randperm(X.shape[0], device=X.device)
+    X_patch, Y_patch = X[order], Y[order]
     X_cutmix = mask.where(X_patch, X)
     mix_portion = float(mask_size**2)/(X.shape[-2]*X.shape[-1])
     Y_cutmix = mix_portion * Y_patch + (1. - mix_portion) * Y
@@ -226,28 +232,25 @@ def train_cifar():
       st = time.monotonic()
       X, Y = X_in, Y_in
       if is_train:
+        perms = Tensor.randperm(X.shape[0], device=X.device)
         # TODO: these are not jitted
         if getenv("RANDOM_CROP", 1):
+          X, Y = X[perms], Y[perms]
           X = random_crop(X, crop_size=32)
         if getenv("RANDOM_FLIP", 1):
           X = (Tensor.rand(X.shape[0],1,1,1) < 0.5).where(X.flip(-1), X) # flip LR
         if getenv("CUTMIX", 1):
           if step >= hyp['net']['cutmix_steps']:
             X, Y = cutmix(X, Y, mask_size=hyp['net']['cutmix_size'])
-        order = list(range(0, X.shape[0]))
-        random.shuffle(order)
-        X, Y = X.numpy()[order], Y.numpy()[order]
-      else:
-        X, Y = X.numpy(), Y.numpy()
+        X, Y = X[perms], Y[perms]
       et = time.monotonic()
       print(f"shuffling {'training' if is_train else 'test'} dataset in {(et-st)*1e3:.2f} ms ({epoch=})")
       for i in range(0, X.shape[0], BS):
         # pad the last batch  # TODO: not correct for test
         batch_end = min(i+BS, Y.shape[0])
-        x = Tensor(X[batch_end-BS:batch_end], device=X_in.device, dtype=X_in.dtype)
-        y = Tensor(Y[batch_end-BS:batch_end], device=Y_in.device, dtype=Y_in.dtype)
         step += 1
-        yield x, y
+        # This needs to be contiguous, or the jitted consumer will fail.
+        yield X[batch_end-BS:batch_end].contiguous(), Y[batch_end-BS:batch_end].contiguous()
       epoch += 1
       if not is_train: break
 
--- a/examples/hlb_cifar10.py
+++ b/examples/hlb_cifar10.py
@@ -203,18 +203,18 @@ def train_cifar():
 
 
   def negative_pad_grid(n=4):
-      # Generate all possible padding combinations that sum to biting n elements from h and w.
-      pad_options = [(-i, i-n) for i in range(n+1)]
+    # Generate all possible padding combinations that sum to biting n elements from h and w.
+    pad_options = [(-i, i-n) for i in range(n+1)]
 
-      # Create all possible combinations of padding for left/right and top/bottom
-      return [(pad_options[x % len(pad_options)], pad_options[y % len(pad_options)])
-              for x in range(n) for y in range(n)]
+    # Create all possible combinations of padding for left/right and top/bottom
+    return [(pad_options[x % len(pad_options)], pad_options[y % len(pad_options)])
+            for x in range(n) for y in range(n)]
 
   def random_crop(X:Tensor, crop_size:int=32) -> Tensor:
-      assert X.shape[-1] >= crop_size
-      Xl = X.split(X.shape[0]//16) # XXX This results in B=3125, which is prob not great for performance, not sure if matters.
-      Xp = [ x.pad(((0,0),(0, 0), *npad)) for x, npad in zip(Xl, negative_pad_grid(X.shape[-1] - crop_size)) ]
-      return Tensor.cat(*Xp)
+    assert X.shape[-1] >= crop_size
+    Xl = X.split(X.shape[0]//16) # XXX This results in B=3125, which is prob not great for performance, not sure if matters.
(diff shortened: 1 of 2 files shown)


IntendedConsequence:
@xl0 The main reason for changing the main loop is because I'm still vague on the details of how JIT works, and it was my attempt at keeping all data processing and augmentations under two segregated functions that I can slap the JIT decorator on. I'm not even sure if shuffling is needed for eval, but it was just a few lines and it preserves existing behavior.

And I'm still not quite sure what to do about the last batch. The original repository for hlb_cifar10 just skips the last batch if it's not a full batch, I opted for extending the last batch with legitimate data for experiment's sake. I don't like how verbose it is though, and I'm not sure if undertraining by a rand(BS) is better or worse than going over.

IntendedConsequence:
@xl0 speaking of FUSE_ARANGE, after reading George's commits in #10759 I found that in minimal reproducible tests (like https://github.com/tinygrad/tinygrad/issues/10340#issuecomment-2959950543) you can add a `.fuse()` after `X[order]` and get the same speedup as with FUSE_ARANGE=1, though I didn't get to test if it works in hlb_cifar10 yet. I don't know how relevant it would be after/if #10759 gets merged though.

xl0:
There is no reason to shuffle eval, and I think dropping the last batch is reasonable, as long as the data is shuffled, and 1 batch is a small % of the data. And we train based on steps, not epochs, so adding more data makes no difference and adds complexity.

But I like the way you tidied up the code. And yes, let me wrap stuff in jit too.

xl0:
Oh wow. I did
```python
  @TinyJit
  def jittable_transforms(X:Tensor, Y:Tensor):
    perms = Tensor.randperm(X.shape[0], device=X.device)
    if getenv("RANDOM_CROP", 1):
      X, Y = X[perms], Y[perms] # 
      X = random_crop(X, crop_size=32)
    if getenv("RANDOM_FLIP", 1):
      X = (Tensor.rand(X.shape[0],1,1,1) < 0.5).where(X.flip(-1), X) # flip LR
    return X[perms], Y[perms]
```

to combine as much as possible. cutmix can't be part of it because it's enabled half-way through the training.

And the result it unexpectedly good - 88 seconds vs 102 seconds before!

So 110 -> 102 -> 88 = 20% improvement!

xl0:
Also dropped the last batch stuff.

xl0:
Autogen test fails for unrelated reasons.

IntendedConsequence:
@xl0 I tried running your solution on my setup to see if I can steal your crop/mix approach, but under the same settings the whole command took 105.86s, shaving one second from my measurement. What GPU do you test it on? I'm measuring on RTX 3080, CUDA backend, DEFAULT_FLOAT=HALF FUSE_ARANGE=1 and LATEWINO=1. The last one made a huge difference (60 seconds!) for me.

xl0:
~~@IntendedConsequence can you run this benchmark?~~
~~https://gist.github.com/xl0/d4d3f05cc4609f4a7e7565855a585179~~

Edit: Let me clean it up first

I'm running it on my 3080 laptop, and on an RTX a6000 Ada.

xl0:
@IntendedConsequence Please give this premium ai slop a go: 

https://gist.github.com/xl0/4200edbb3074de7b5404bc742042b50d

I'm running it on both, will share the results when done. Make sure the GPU is not thermally throttling.

xl0:
Here's a script to visualize the results:
https://gist.github.com/xl0/eba3b7a5d29bdc3997803dae4824e6b7

`python visualize_benchmarks.py your_benchmark_file.csv`

a6000:
```
Version,FloatType,RealTime,FirstShuffleTime,LastShuffleTime,EvalAccuracy
baseline,fp32,112.65,2826.00,2835.63,93.57
baseline,fp16,93.65,1859.05,1588.49,93.76
baseline,bf16,120.22,3202.38,2866.78,93.89
nonumpy,fp32,102.52,82.80,151.03,93.71
nonumpy,fp16,92.28,82.35,156.55,93.76
nonumpy,bf16,113.10,86.18,154.90,93.87
jit,fp32,89.02,2073.75,0.14,93.85
jit,fp16,81.26,1898.15,0.15,94.01
jit,bf16,98.03,2164.03,0.15,93.72
```
a6000_visualization

xl0:
3080 laptop. Less clear-cut, but jit still wins easily.
```
Version,FloatType,RealTime,FirstShuffleTime,LastShuffleTime,EvalAccuracy
baseline,fp32,352.05,3805.04,3645.29,93.98
baseline,fp16,267.86,2530.40,2108.18,93.68
baseline,bf16,356.57,3871.39,3684.82,93.53
nonumpy,fp32,353.43,103.56,191.60,93.50
nonumpy,fp16,280.46,103.01,190.92,93.71
nonumpy,bf16,343.93,103.02,191.26,93.86
jit,fp32,321.67,2201.36,0.18,93.61
jit,fp16,251.05,2281.55,0.18,94.04
jit,bf16,301.74,2276.05,0.18,93.79
```

benchmark_summary_visualization

IntendedConsequence:
Sorry for wall of text, it started with just a few questions but got a bit out of control:

You're grepping for dataset shuffling measurement, but with numpy removed, doesn't it just measure tinygrad internal graph/ast building/optimizing?

I stopped looking at it after some of my experiments made it near 0ms, and the first step after that gaining its "weight" instead, and given tinygrad's lazy evaluation model, when dataset processing is done with its tensors, even if you put `.realize()` before getting the timestamp you're still not measuring the exact time without an explicit synchronize call. I just look at ms per step and any outliers, but mainly total seconds and eval score.

I noticed you aren't measuring with LATEWINO=1, is there a reason for that? Check out the hlb_cifar tab in tinygrad bounties google sheet, it hasn't been updated in a while though, but gives some insights. 

There's another environment variable there that's used in almost every entry, LATEBEAM, but on windows I can't really use any beam-related flags because of how slow it is.

Also thanks for clearing up which of your measurements were on 3080 and which on A6000, btw is your 3080 a Ti? Mine is a regular one, desktop, and I power throttle it at 85% for various reasons.

I'm still getting the hang of how tinygrad jit works, but I noticed that in official code all jitted functions seem to return realized tensors, with a few exceptions where it happens the first thing after, like `jitted_forward(X).realize()`, is there a reason you didn't do that in your jitted functions?

xl0:
>You're grepping for dataset shuffling measurement, but with numpy removed, doesn't it just measure tinygrad internal graph/ast building/optimizing?

Yeah, it is more of a sanity check, to confirm the code behaves as I expect (it does) and it shows a difference between baseline and nonumpy. The number I care most about is the total time. I guess we could remove the shuffle timer too.

I'm not sure, I just did not have a reason to. WINO changes the way convolutions are implemented, so it's a bit out of the way, isn't it? I might be missing some meta here. :)

But sure, let me fire a benchmark on a6000 - the 3080 takes too long for all tests. And I can throw in LATEBEAM too, let me give it a try. 

> btw is your 3080 a Ti? 

GPU 0: NVIDIA GeForce RTX 3080 Laptop GPU
16Gb VRAM and 115W power consumption. Not sure how it compares to the desktop 3080 and why Nvidia decided to bless it with 16 Gbs, but can't complain.

Can't comment on the .realize(), I'm not sure 

Here are the results from LATEWINO=1:
```
Version,FloatType,RealTime,FirstShuffleTime,LastShuffleTime,EvalAccuracy
baseline,fp32,100.68,2718.86,2808.08,93.90
baseline,fp16,76.57,1827.51,1596.29,93.66
baseline,bf16,98.89,2744.94,2953.20,93.75
nonumpy,fp32,92.08,81.02,146.79,93.96
nonumpy,fp16,77.83,81.44,149.72,93.79
nonumpy,bf16,100.08,79.75,150.16,93.80
jit,fp32,77.28,1678.62,0.13,93.79
jit,fp16,64.86,1737.58,0.13,93.95
jit,bf16,76.14,1745.86,0.13,93.82
```

benchmark_summary_visualization

The BEAM stuff does not seem to actually work for me at the moment, regardless of my changes. I'll need to dig deeped into how it should be used properly.

xl0:
@geohot ready for review btw

IntendedConsequence:
> I'm not sure, I just did not have a reason to. WINO changes the way convolutions are implemented, so it's a bit out of the way, isn't it? I might be missing some meta here. :)

WINO is used in benchmark entries by the team in the google sheets tab I mentioned, and it cuts down total time almost in half in my case without noticeable accuracy loss, good enough for me.

~~Also your results with LATEWINO on A6000 are puzzling, because at least the one I care about (jit fp16) takes about the same or slower as on my 85% power throttled 3080, on which I'm getting consistent 65-67 seconds for the whole run with FUSE_ARANGE and LATEWINO. Wait, did you enable FUSE_ARANGE as well?~~
**[EDIT]** Disregard that, I mixed up my minutes and seconds (1:05s vs 105s) 🦀 

For measuring step times, I found empirical cumulative distribution function (ECDF) works great with video game frame times compared to histograms/mean/median/std/variance plots, so could work for training step times overview as well. The way you plot it is you just sort the timestamps in ascending order, and plot them on the x-axis, with their sorted indices on the y-axis. Divide indices by total count and you get percentiles, but indices work well enough for lower counts as well. I set the x-scale to logarithmic because you usually want to zoom into the top left corner anyway, and added custom xticks/yticks to make it more readable. The best way to read these is with interactive zoom and pan.

Example (not the latest versions, fp16, jit, all with FUSE_ARANGE and LATEWINO).
image

The basic way to compare graphs is to look which one is closer to the top left corner. For example, 90 percent of both your and mine step times are clearly faster by 1-2.5ms than baseline, but both of our approaches have a chunk of step times that are 5ms slower, about 8% or so, it seems to correlate to slightly slower steps of the very first epoch.

Another way to read it is you pick a percentile on the left (y-axis) and follow it horizontally until you hit the curve, then trace down to the ms on the x-axis,  for example, 0.9 (90%) of step times of the original code take 55ms or less.

The only thing ECDF doesn't help you with is if you have an extreme outlier of a step, like with LATEWINO I have the first step take something like 30-40 seconds on all 3 commits, on the plot it's at the very top and far right, somewhere in the 0.999 and isn't visible at all.

xl0:
Would PDF be a better visualization here? Let me have a look.

xl0:
This is from a6000-wino. Not sure what to make of it.

step_times_by_float_type

examples/hlb_cifar10.py
```diff
@@ -226,28 +243,16 @@ def fetch_batches(X_in:Tensor, Y_in:Tensor, BS:int, is_train:bool):
...
-        if getenv("CUTMIX", 1):
-          if step >= hyp['net']['cutmix_steps']:
-            X, Y = cutmix(X, Y, mask_size=hyp['net']['cutmix_size'])
-        order = list(range(0, X.shape[0]))
-        random.shuffle(order)
-        X, Y = X.numpy()[order], Y.numpy()[order]
-      else:
-        X, Y = X.numpy(), Y.numpy()
+        X, Y = jittable_transforms(X, Y)
+        if getenv("CUTMIX", 1) and step >= hyp['net']['cutmix_steps']:
```
geohot: Why isn't this jittable? The other one seems to JIT it.

examples/hlb_cifar10.py
```diff
@@ -201,19 +201,36 @@ def make_square_mask(shape, mask_size) -> Tensor:
...
+    pad_options = [(-i, i-n) for i in range(n+1)]
+
+    # Create all possible combinations of padding for left/right and top/bottom
+    return [(pad_options[x % len(pad_options)], pad_options[y % len(pad_options)])
+            for x in range(n) for y in range(n)]
+
+  def random_crop(X:Tensor, crop_size:int=32) -> Tensor:
+    assert X.shape[-1] >= crop_size
+    Xl = X.split(X.shape[0]//16) # XXX This results in B=3125, which is prob not great for performance, not sure if matters.
+    Xp = [ x.pad(((0,0),(0, 0), *npad)) for x, npad in zip(Xl, negative_pad_grid(X.shape[-1] - crop_size)) ]
```
geohot: strange spacing

examples/hlb_cifar10.py
```diff
@@ -201,19 +201,36 @@ def make_square_mask(shape, mask_size) -> Tensor:
...
-  def random_crop(X:Tensor, crop_size=32):
-    mask = make_square_mask(X.shape, crop_size)
-    mask = mask.expand((-1,3,-1,-1))
-    X_cropped = Tensor(X.numpy()[mask.numpy()])
-    return X_cropped.reshape((-1, 3, crop_size, crop_size))
-
-  def cutmix(X:Tensor, Y:Tensor, mask_size=3):
-    # fill the square with randomly selected images from the same batch
+
+  def negative_pad_grid(n=4):
```
geohot: I don't understand this, why doesn't the other one have it?

geohot:
Okay, here's my thoughts. If you two work together I'll up the bounty to a shared $300 (so $150 each)

examples/hlb_cifar10.py
```diff
@@ -226,28 +243,16 @@ def fetch_batches(X_in:Tensor, Y_in:Tensor, BS:int, is_train:bool):
...
       print(f"shuffling {'training' if is_train else 'test'} dataset in {(et-st)*1e3:.2f} ms ({epoch=})")
       for i in range(0, X.shape[0], BS):
-        # pad the last batch  # TODO: not correct for test
-        batch_end = min(i+BS, Y.shape[0])
-        x = Tensor(X[batch_end-BS:batch_end], device=X_in.device, dtype=X_in.dtype)
-        y = Tensor(Y[batch_end-BS:batch_end], device=Y_in.device, dtype=Y_in.dtype)
         step += 1
-        yield x, y
+        if i+BS > X.shape[0]: break
+        # This needs to be contiguous, or the jitted consumer will fail.
```
geohot: They shouldn't need to be contiguous I don't think

examples/hlb_cifar10.py
xl0: Since cutmix does not kick in until half-way, if we want to jit it with other transforms, we have to play tricks to make it an effective nop. But this gets tricly and with weird performance dynamics (non-obvious .realize() and .contiguous() requred to keep it on par with the version with 2 separate functions). While this jits cutmix separately, it's simpler and performs as well as the other version with non-obvious tricks.

examples/hlb_cifar10.py
xl0: It does, the train step fails with
```
AssertionError: args mismatch in JIT:
self.captured.expected_st_vars_dtype_device=[
  (ShapeTracker(views=(View(shape=(512, 3, 32, 32), strides=(3072, 1024, 32, 1), offset=1572864, mask=None, contiguous=False),)), (), dtypes.half, 'CUDA'),
  (ShapeTracker(views=(View(shape=(512, 10), strides=(10, 1), offset=5120, mask=None, contiguous=False),)), (), dtypes.half, 'CUDA')
] != st_vars_dtype_device=[
  (ShapeTracker(views=(View(shape=(512, 3, 32, 32), strides=(3072, 1024, 32, 1), offset=3145728, mask=None, contiguous=False),)), (), dtypes.half, 'CUDA'),
  (ShapeTracker(views=(View(shape=(512, 10), strides=(10, 1), offset=10240, mask=None, contiguous=False),)), (), dtypes.half, 'CUDA')]
```

because the tensor is indexed at a differen offset. It looks like other examples apply transforms at each step, rather than transforming the whole dataset once per epoch.

examples/hlb_cifar10.py
xl0: I used different approache to work around the unusability of mark indexing. In the end we chose to go with the approach from #10728

examples/hlb_cifar10.py
IntendedConsequence: Yeah, to put both into the same functions we have to rewrite it so it can work branchless inside the jit (passing source indices for the patches from outside, always doing the cutmix, just with unshuffled source indices before cutmix_step), but then randperm has to go outside and is either unjitted and slow, or jitted one-liner but we're back at 2 jitted functions.

xl0:
```
Version,FloatType,RealTime,FirstShuffleTime,LastShuffleTime,EvalAccuracy
baseline,fp32,100.52,2814.42,2793.53,93.90
baseline,fp16,76.16,1898.62,1607.91,93.66
nonumpy,fp32,96.51,80.95,148.33,93.96
nonumpy,fp16,74.91,82.94,146.58,93.79
jit,fp32,76.56,1642.60,0.12,93.79
jit,fp16,64.67,1708.96,0.12,93.95
gather,fp32,75.98,1794.64,0.13,93.65
gather,fp16,63.17,1578.48,0.12,93.78
```
a6000 ada FUSE_ARANGE=1 LATEWINO=1
image

.github/workflows/benchmark.yml
```diff
@@ -109,13 +109,13 @@ jobs:
     - name: Train MNIST
       run: time PYTHONPATH=. TARGET_EVAL_ACC_PCT=96.0 python3.11 examples/beautiful_mnist.py | tee beautiful_mnist.txt
     - name: Run 10 CIFAR training steps
-      run: BENCHMARK_LOG=cifar_10steps JIT=1 STEPS=10 python3.11 examples/hlb_cifar10.py | tee train_cifar.txt
+      run: BENCHMARK_LOG=cifar_10steps FUSE_ARANGE=1 JIT=1 STEPS=10 python3.11 examples/hlb_cifar10.py | tee train_cifar.txt
```
geohot: Don't do this. I told you the right way to deal with this in discord

examples/hlb_cifar10.py
geohot: This can be fixed without contiguous

examples/hlb_cifar10.py
geohot: This needs to be in the jit

examples/hlb_cifar10.py
```diff
@@ -360,7 +361,7 @@ def eval_step(model, X, Y):
   eval_acc_pct = 0.0
   batcher = fetch_batches(X_train, Y_train, BS=BS, is_train=True)
   with Tensor.train():
-    st = time.monotonic()
+    st = time.perf_counter()
```
geohot: This monotonic -> perf_counter change should be a separate PR

examples/hlb_cifar10.py
IntendedConsequence: You mean with Variable?

.github/workflows/benchmark.yml
IntendedConsequence: `Context(FUSE_ARANGE=1).__enter__()` at the top is failing CI.

https://github.com/tinygrad/tinygrad/actions/runs/15719232713/job/44296517880?pr=10728

examples/hlb_cifar10.py
xl0: OK

.github/workflows/benchmark.yml
xl0: If it was not clear, a bunch of tests do
```py
    from examples.hlb_cifar10 import UnsyncedBatchNorm
```

which means that `UnsyncedBatchNorm` should be moved somewhere else, but also means that we can't just blindly enter the context at the top of the file.

Instead, let's do
```py
@Context(FUSE_ARANGE=getenv("FUSE_ARANGE", 1))
def train_cifar():
   ...
```

examples/hlb_cifar10.py
xl0: Ok, we could not figure a fast way to conditionally make cutmix a nop, but we don't have to. We can just always return both X, Y, and X_cutmix, Y_cutmix, and then choose to use either one.

```py
  @TinyJit
  def jittable_transforms(X:Tensor, Y:Tensor):
    perms = Tensor.randperm(X.shape[0], device=X.device) # We reuse perms for cutmix, because they are expensivne to generate
    if getenv("RANDOM_CROP", 1):
      X = random_crop(X, crop_size=32)
    if getenv("RANDOM_FLIP", 1):
      X = (Tensor.rand(X.shape[0],1,1,1) < 0.5).where(X.flip(-1), X) # flip LR
    X, Y = X[perms], Y[perms]
    return X, Y, *cutmix(X, Y, perms, mask_size=hyp['net']['cutmix_size'])

...

        X, Y, X_cm, Y_cm = jittable_transforms(X, Y)
        if getenv("CUTMIX", 1) and step >= hyp['net']['cutmix_steps']: X, Y = X_cm, Y_cm

 ```

@geohot is this what you had in mind?

Comment (xl0) [reply examples/hlb_cifar10.py]: @geohot Me and @IntendedConsequence are trying to guess your preferences here. It would be more productive if you could tell us if you like this version, or if you have something else in mind.

**Comment by xl0** [comment]
| Version | FloatType | RealTime | FirstShuffleTime | LastShuffleTime | EvalAccuracy |
|---|---|---|---|---|---|
| baseline e94ac6e | fp32 | 99.78 | 2676.74 | 2806.28 | 93.90 |
| baseline e94ac6e | fp16 | 76.63 | 1835.99 | 1617.83 | 93.66 |
| nonumpy cd7b2c6 | fp32 | 90.70 | 83.28 | 148.60 | 93.96 |
| nonumpy cd7b2c6 | fp16 | 75.27 | 79.52 | 151.59 | 93.79 |
| jit f8b4279 | fp32 | 76.38 | 1639.97 | 0.12 | 93.79 |
| jit f8b4279 | fp16 | 64.19 | 1677.91 | 0.13 | 93.95 |
| gather 0d5a81f | fp32 | 76.13 | 1539.72 | 0.11 | 93.65 |
| gather 0d5a81f | fp16 | 63.19 | 1568.92 | 0.10 | 93.78 |
| jit_cutmix 6d85582 | fp32 | 73.90 | 1656.97 | 0.07 | 93.52 |
| jit_cutmix 6d85582 | fp16 | 61.59 | 1695.85 | 0.06 | 93.75 |

image

File: examples/hlb_cifar10.py [hunk 7]
```diff
@@ -201,53 +202,54 @@ def make_square_mask(shape, mask_size) -> Tensor:
...
-        if getenv("CUTMIX", 1):
-          if step >= hyp['net']['cutmix_steps']:
-            X, Y = cutmix(X, Y, mask_size=hyp['net']['cutmix_size'])
-        order = list(range(0, X.shape[0]))
-        random.shuffle(order)
-        X, Y = X.numpy()[order], Y.numpy()[order]
-      else:
-        X, Y = X.numpy(), Y.numpy()
+        X, Y, X_cm, Y_cm = augmentations(X, Y)
+        if getenv("CUTMIX", 1) and step >= hyp['net']['cutmix_steps']: X, Y = X_cm, Y_cm
```
geohot, line 244: Why are you computing cutmix when you don't need it?

geohot:
I have no idea what you are doing with cutmix and the perms. It looks like nonsense to me, and there's a typo in the comment which supports that theory.

Ahh, the double index into perms confused me, but it's actually right...

geohot:
Merged, bounty is yours! $150 each, reach out to george@tinygrad.org to claim, can pay via paypal or USDC

PR #10779, opened 2025-06-11
AMD_LLVM=0 for sdxl search

chenyuxyz:
hangs with llvm19 but seems fine with llvm20


chenyuxyz:
merging this for now so the script runs on github action

PR #10782, opened 2025-06-11
hung kernel with AMD_LLVM on llvm19

chenyuxyz:
works with llvm20


chenyuxyz:
also works with `BLOCK_REORDER=0`

chenyuxyz:
i (and LLMs) believe this is register spilling and bug in amdllvm 19, which seems to be solved in llvm 20, so we plan to update ci and tinybox to use llvm 20

PR #10783, opened 2025-06-11
Fix integer overflow in bitwise shift ops by casting base to float

mdaiter:
Replaces `2**y` with `2.0**y` in left and right shift emulations to avoid integer overflow.
Ensures consistent behavior for large y values on non-integer tensor types.

Ran into this issue while testing left and right shift emulations. Happy to provide minimum replication samples.


chenyuxyz:
stale

Issue #10458, opened 2025-05-22
[WebGPU] IndexError: index dtype dtypes.float is not supported


0312birdzhang:
Hi, I'm using rubkipi 3, a Qcom QCS6490 dev board. It has vulkan support, and can get gpu backends via wgpu.
```
>>> import wgpu
>>> print(wgpu.gpu.request_adapter_sync())
<wgpu.backends.wgpu_native.GPUAdapter object at 0x7f8a846090>
```

I have also tried example from https://github.com/pygfx/wgpu-py/blob/main/examples, and works well.

I have also rerun `autogen_stubs.sh`, nothing different. 

Just confused what should i do now. Pls help me, thanks.

The errors are:
```
(openpilot) comma@RUBIKPi:/data/openpilot/tinygrad_repo$ PYTHONPATH="." WEBGPU=1 IMAGE=0 python3 examples/openpilot/compile3.py https://github.com/commaai/openpilot/raw/v0.9.8/selfdrive/modeld/models/driving_policy.onnx
loaded model
Traceback (most recent call last):
  File "/data/openpilot/tinygrad_repo/examples/openpilot/compile3.py", line 136, in <module>
    test_val = compile(onnx_file) if not getenv("RUN") else None
               ^^^^^^^^^^^^^^^^^^
  File "/data/openpilot/tinygrad_repo/examples/openpilot/compile3.py", line 30, in compile
    new_inputs = {k:Tensor.randn(*shp, dtype=_from_np_dtype(input_types[k])).mul(8).realize() for k,shp in sorted(input_shapes.items())}
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/openpilot/tinygrad_repo/tinygrad/tensor.py", line 4246, in _wrapper
    ret = fn(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^
  File "/data/openpilot/tinygrad_repo/tinygrad/tensor.py", line 264, in realize
    run_schedule(*self.schedule_with_vars(*lst), do_update_stats=do_update_stats)
  File "/data/openpilot/tinygrad_repo/tinygrad/engine/realize.py", line 170, in run_schedule
    for si, ei in lower_schedule(schedule):
                  ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10694, opened 2025-06-08
ONNX real float16

geohotstan:
Issue a UserWarning for dtype mismatches instead of explicitly checking.

`PYTHONPATH=. python test/models/test_onnx.py TestOnnxModel.test_benchmark_openpilot_model`

```
/Users/zibo/fun/tiny/tinygrad/extra/onnx.py:140: UserWarning: input_imgs has mismatch on dtype. Expected dtypes.half, received dtypes.float.                                                                                                                                                       
  if tensor.dtype is not spec.dtype: warnings.warn(f"{name} has mismatch on dtype. Expected {spec.dtype}, received {tensor.dtype}.")
/Users/zibo/fun/tiny/tinygrad/extra/onnx.py:140: UserWarning: big_input_imgs has mismatch on dtype. Expected dtypes.half, received dtypes.float.
  if tensor.dtype is not spec.dtype: warnings.warn(f"{name} has mismatch on dtype. Expected {spec.dtype}, received {tensor.dtype}.")
/Users/zibo/fun/tiny/tinygrad/extra/onnx.py:140: UserWarning: desire has mismatch on dtype. Expected dtypes.half, received dtypes.float.
  if tensor.dtype is not spec.dtype: warnings.warn(f"{name} has mismatch on dtype. Expected {spec.dtype}, received {tensor.dtype}.")
/Users/zibo/fun/tiny/tinygrad/extra/onnx.py:140: UserWarning: traffic_convention has mismatch on dtype. Expected dtypes.half, received dtypes.float.
  if tensor.dtype is not spec.dtype: warnings.warn(f"{name} has mismatch on dtype. Expected {spec.dtype}, received {tensor.dtype}.")
/Users/zibo/fun/tiny/tinygrad/extra/onnx.py:140: UserWarning: nav_features has mismatch on dtype. Expected dtypes.half, received dtypes.float.
  if tensor.dtype is not spec.dtype: warnings.warn(f"{name} has mismatch on dtype. Expected {spec.dtype}, received {tensor.dtype}.")
/Users/zibo/fun/tiny/tinygrad/extra/onnx.py:140: UserWarning: features_buffer has mismatch on dtype. Expected dtypes.half, received dtypes.float.
  if tensor.dtype is not spec.dtype: warnings.warn(f"{name} has mismatch on dtype. Expected {spec.dtype}, received {tensor.dtype}.")
ran openpilot model in 2598.41 ms, waited 2466.94 ms for realize, 7.02 ms for GPU queue
```

[...]

PR #10787, opened 2025-06-12
Add ensure_supported_dtype to ONNX

geohotstan:
- Ensures that dtypes of tensors internal to the graph is supported on device for the FP and issues warnings when dtype is defaulting to a default dtype.
- Removes the numpy in buffer load and instead use:
`DISK (bitcast, reshape) -> DEVICE (optionally -> CPU cast to device supported dtype -> back to DEVICE, if dtype not support on device)` if on DISK, or `DEVICE (cast, reshape) and -> CPU cast -> back to DEVICE if dtype not supported` if already on DEVICE

split from #10694

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -1,13 +1,13 @@
 from types import SimpleNamespace
 from typing import Any, Sequence, cast, Literal, Callable
-import dataclasses, functools, io, math, types
+import dataclasses, functools, io, math, types, warnings
 from tinygrad.tensor import Tensor, _broadcast_shape, ReductionStr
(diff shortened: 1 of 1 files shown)


chenyuxyz:
this also fixes https://github.com/tinygrad/tinygrad/issues/10458?

geohotstan:
> this also fixes #10458?

yep fixes the error in #10458 title
"IndexError: index dtype dtypes.float is not supported"

chenyuxyz:
can you split this one more time, so one will have no-op refactor only. then we can fix default behavior with small change

geohotstan:
> can you split this one more time, so one will have no-op refactor only. then we can fix default behavior with small change

yeah sorry, I just now understood what was meant by no-op refactor 😓

PR #10788, opened 2025-06-12
Add fallback dtype to ONNX

geohotstan:
introduces a noop change to add function `ensure_supported_dtype` to notify user when a fallback dtype happens

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -1,13 +1,13 @@
 from types import SimpleNamespace
 from typing import Any, Sequence, cast, Literal, Callable
-import dataclasses, functools, io, math, types
+import dataclasses, functools, io, math, types, warnings
 from tinygrad.tensor import Tensor, _broadcast_shape, ReductionStr
 from tinygrad.helpers import getenv, DEBUG, all_same, prod, flatten, make_tuple, argsort
 from tinygrad.dtype import DType, ConstType, dtypes, ImageDType
 from tinygrad.device import is_dtype_supported, Device
 
 # ***** protobuf parsing ******
-from onnx import AttributeProto, ModelProto, TensorProto, TypeProto, helper
+from onnx import AttributeProto, ModelProto, TensorProto, TypeProto
 import numpy as np
 
 def has_field(onnx_type: TypeProto|SimpleNamespace, field):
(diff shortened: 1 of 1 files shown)


extra/onnx.py
```diff
@@ -26,7 +26,14 @@ def dtype_parse(onnx_dtype: int) -> DType:
     TensorProto.FLOAT8E5M2, TensorProto.FLOAT8E5M2FNUZ, TensorProto.UINT4, TensorProto.INT4
   }
   if onnx_dtype in unsupported: raise NotImplementedError(f"onnx dtype {TensorProto.DataType.Name(onnx_dtype)} is not supported")
-  return supported[onnx_dtype] if is_dtype_supported(supported[onnx_dtype]) else dtypes.float
```
chenyuxyz, line 29: why not add warning and dtype fallback here? it seems like ensure_supported_dtype is needed every time dtype_parse is called

geohotstan: problem is `dtype = ensure_supported_dtype(dtype_parse(elem_type.tensor_type.elem_type), "input type spec parse")` will be removed later. I kept it in just to make sure this PR is a noop change.

There are OnnxRunner methods like 
```
  def get_empty_input_data(self, device:str|None=None) -> dict[str, Tensor]:
    return {name:Tensor.empty(*spec.shape, device=device, dtype=spec.dtype) for name, spec in self.graph_inputs.items()}
```

That currently use spec.dtype as source of truth. (spec.dtype is resulting from "input type spec parse")

----

so there are two use cases for `dtype_parse`
1. parsing spec (model source of truth) to validate user input (external Tensor dtype)
2. parsing value tensor dtypes or op dtypes inside graph (internal Tensor dtype)

"input type spec parse" is for comparing external dtype. This should not have any fallback since this dtype should be used to warn the user if the dtype of their input does not match model spec.

The reason why I separated out ensure_dtype_supported as it's own function since some `dtype_parse` needs true dtype.

chenyuxyz: How about an arg `allow_fallback`?

My concern is that current approach has too many copy paste that easily get misused

chenyuxyz:
eventually the dtype fallback stuff needs to be integrate into tinygrad dtype (we want to distinguish dtype of the tensor, and how we run alu of that dtype). fine for now though. and context debug into should combine with metadata

PR #10789, opened 2025-06-12
OnnxRunner file as input

geohotstan:
minimal change to enable file as input OnnxRunner

The idea of OnnxRunner is
`OnnxRunner.__init__` -> parse model and load model into easy to use/access spec attributes of OnnxRunner
`OnnxRunner.__call__` -> FP of model

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -185,7 +185,7 @@ class OnnxRunner:
   def get_empty_input_data(self, device:str|None=None, dtype:DType|None=None) -> dict[str, Tensor]:
     return {name:Tensor.empty(*spec.shape, device=device, dtype=dtype or spec.dtype) for name, spec in self.graph_inputs.items()}
 
-  def to(self, device:str):
+  def to(self, device:str|None):
     self.graph_values = {k:v.to(device) if isinstance(v, Tensor) else v for k,v in self.graph_values.items()}
     self.graph_nodes = tuple(OnnxNode(n.num, n.op, tuple(n.inputs), tuple(n.outputs),
                                       {k:v.to(device) if isinstance(v, Tensor) else v for k,v in n.opts.items()}) for n in self.graph_nodes)


examples/openpilot/compile3.py
```diff
@@ -12,14 +12,14 @@
...
-from tinygrad.frontend.onnx import OnnxRunner, onnx_load
+from tinygrad.frontend.onnx import OnnxRunner
 
 OPENPILOT_MODEL = sys.argv[1] if len(sys.argv) > 1 else "https://github.com/commaai/openpilot/raw/v0.9.7/selfdrive/modeld/models/supercombo.onnx"
 OUTPUT = sys.argv[2] if len(sys.argv) > 2 else "/tmp/openpilot.pkl"
 
 def compile(onnx_file):
-  onnx_model = onnx_load(onnx_file)
-  run_onnx = OnnxRunner(onnx_model)
+  onnx_model = onnx.load(onnx_file)
```
geohotstan: There are still some `onnx.load` usage here and `external_model_benchmark`
This is because the spec parsing is not parsing the model truthfully. Need true spec parsing before this can be cleaned up.

geohotstan:
`examples/openpilot/compile4.py` is currently broken maybe due to recent refactors.

from `compile4.py`
`schedule, var_vals, becomes_map = create_schedule_with_vars(independent)`

`create_schedule_with_vars` returns `schedule, var_vals`

examples/openpilot/compile3.py
chenyuxyz: what's true spec parsing and what's missing?

preferably we don't have code that reverts `onnx_load` back to `onnx.load`

geohotstan: oh I just saw this now!!
I nested `onnx_load` inside `OnnxRunner.__init__`
currently `onnx_load` and `onnx.load` result in faithful parsing of model, but when loading the model into `OnnxRunner` we get unfaithful stuff like untrue float16.

geohotstan: left it as onnx_load. 👍

geohotstan:
whaaaat, ok investigating...

chenyuxyz:
comma benchmark becomes quite slower. is that expected?

geohotstan:
> comma benchmark becomes quite slower. is that expected?

yeah unfortunately it's doing onnx_load on the model twice now.
from `external_benchmark_openpilot.py`
```
  onnx_model = onnx_load(onnx_path := fetch(OPENPILOT_MODEL))
  run_onnx = OnnxRunner(onnx_path)
```

There's an `onnx_load` inside `OnnxRunner` too.
later we don't need to have a separate onnx_model and run_onnx once OnnxRunner can have faithful parsing of model.

if it's a bit too much, I could try to inch towards faithful parsing first? :s

chenyuxyz:
i can remove some tests so that this is not too slow. can you add a TODO not in external_benchmark_openpilot to explain why it's parsed twice?

chenyuxyz:
i think we want to get faithful parsing before merging this. is it only blocked on true float16?

geohotstan:
> i think we want to get faithful parsing before merging this. is it only blocked on true float16?

Sure thing.
True float16 and removing the fallback_dtype for input spec parsing.

chenyuxyz:
dtype fallback and true fp16 merged, how's this one looking?

geohotstan:
The flakiness persists.
Manual removal (closing of file) of the tempfile holding the onnx model still does not fix the flakiness.

geohotstan:
I think this is ready for a benchmark run.

test/external/external_benchmark_openpilot.py
```diff
@@ -1,24 +1,21 @@
...
 
   Tensor.manual_seed(100)
-  input_shapes = {inp.name:tuple(x.dim_value for x in inp.type.tensor_type.shape.dim) for inp in onnx_model.graph.input}
-  input_types = {inp.name: tensor_dtype_to_np_dtype(inp.type.tensor_type.elem_type) for inp in onnx_model.graph.input}
-  new_inputs = {k:Tensor.randn(*shp, dtype=_from_np_dtype(input_types[k])).mul(8).realize() for k,shp in input_shapes.items()}
-  new_inputs_junk = {k:Tensor.randn(*shp, dtype=_from_np_dtype(input_types[k])).mul(8).realize() for k,shp in input_shapes.items()}
+  input_shapes = {name: spec.shape for name, spec in run_onnx.graph_inputs.items()}
+  input_types = {name: spec.dtype for name, spec in run_onnx.graph_inputs.items()}
+  new_inputs = {k:Tensor.randn(*shp, dtype=input_types[k]).mul(8).realize() for k,shp in input_shapes.items()}
+  new_inputs_junk = {k:Tensor.randn(*shp, dtype=input_types[k]).mul(8).realize() for k,shp in input_shapes.items()}
```
geohotstan, line 18: previous failure was because it was
```
  new_inputs = {k:Tensor.randn(*shp, dtype=input_types[k]).mul(8).realize() for k,shp in sorted(input_shapes.items())}
  new_inputs_junk = {k:Tensor.randn(*shp, dtype=input_types[k]).mul(8).realize() for k,shp in sorted(input_shapes.items())}
```

This gave different values because `Tensor.randn().realize()` had it's order changed so it gave different values.
The order of input (dict entries) does not affect forward pass and thus shouldn't change result.

Not sure how Cursor tab complete read `compile3.py` because there is a sorted in there... It decided to apply it here too when I multi-line tabbed... smh

chenyuxyz:
mac inference benchmark got 2 minutes slower, that's more than 100%. can you repro this locally? i think 20% is fine but 120% is too much

geohotstan:
yeah that's weird, and unexpected. I'll take a close look.
Oh onnx model parse was running independently for each device. I need to think about this a little. Maybe an abstraction problem.

extra/onnx.py
```diff
@@ -178,6 +185,12 @@ def _dispatch_op(self, op, inps, opts):
   def get_empty_input_data(self, device:str|None=None, dtype:DType|None=None) -> dict[str, Tensor]:
     return {name:Tensor.empty(*spec.shape, device=device, dtype=dtype or spec.dtype) for name, spec in self.graph_inputs.items()}
 
+  def to(self, device:str):
+    self.graph_values = {k:v.to(device) if isinstance(v, Tensor) else v for k,v in self.graph_values.items()}
+    self.graph_nodes = tuple(OnnxNode(n.num, n.op, tuple(n.inputs), tuple(n.outputs),
+                                      {k:v.to(device) if isinstance(v, Tensor) else v for k,v in n.opts.items()}) for n in self.graph_nodes)
+    return self
```
geohotstan, line 192: this is a little janky now, but since onnx is `SSA`, the initializer tensors in `graph_values` remains unchanged, so it works. Could be made more explicit in the future though.

PR #10800, opened 2025-06-13
ONNX improve dtype fallback 

geohotstan:
minimal fix for https://github.com/tinygrad/tinygrad/issues/10458


geohotstan:
If we are to expect ONNX to be used with WEBGPU, this probably has to work
`WEBGPU=1 PYTHONPATH=. pytest test/external/external_test_onnx_backend.py`

Currently, there are a lot of int64 usage in the op definitions like
```
  def ArgMax(x:Tensor, axis:int=0, keepdims:int=1, select_last_index:int=0):
    if select_last_index: return ((x.shape[axis]-1) - x.flip(axis).argmax(axis, keepdim=keepdims)).cast(dtypes.int64)
    return x.argmax(axis, keepdim=keepdims).cast(dtypes.int64)
```
This is because tests expect int64 as output (ONNX expects int64 as default)...
This obviously doesn't run on WEBGPU, but hmmmm.

There are still a lot of failures in this test suite with WEBGPU. (This PR fixes the gather ones :))

can confirm the test failures only consists of 
1. int64 stuff
2. 4 tests because buffer_parse is using dtype_parse fallback and bitcasting to it (this is to be fixed next with float16 flip)

chenyuxyz:
webgpu not supporting int64 is fine, webgpu needs a way to deal with int64

can we add some simple tests for this behavior?

geohotstan:
> webgpu not supporting int64 is fine, webgpu needs a way to deal with int64
> 
> can we add some simple tests for this behavior?

already on it.
Gonna add an `external_test_onnx_runner.py` to test stuff related to `OnnxRunner.__init__` (had this idea for the longest time xD)
Later this can be used to investigate speed stuff too.

extra/onnx.py
```diff
@@ -65,7 +61,7 @@ def buffer_parse(onnx_tensor: TensorProto) -> Tensor:
   if has_field(onnx_tensor, "raw_data"):
     raw_data = onnx_tensor.raw_data
     if not isinstance(raw_data, Tensor): raw_data = Tensor(raw_data)
-    if onnx_tensor.data_type == TensorProto.FLOAT16:
+    if onnx_tensor.data_type == TensorProto.FLOAT16 or not is_dtype_supported(data_types[onnx_tensor.data_type]):
```
geohotstan, line 64: This fixed a bug before. (found from onnx_runner test)
It looks reallly really messy now but will be totally cleaned up later.

geohotstan:
https://github.com/tinygrad/tinygrad/actions/runs/15797111154/job/44530569815?pr=10800
another random flaky error. This was after `external_test_onnx_backend.py`. Maybe that test corrupts the runner........
what is this disk voodoo D:

test/external/external_test_onnx_runner.py
```diff
@@ -0,0 +1,77 @@
...
+    input_tensor = onnx.helper.make_tensor_value_info('input', onnx_data_type, ())
+    output_tensor = onnx.helper.make_tensor_value_info('output', onnx_data_type, ())
+    node = onnx.helper.make_node('Identity', inputs=['input'], outputs=['output'])
+    graph = onnx.helper.make_graph([node], 'identity_test', [input_tensor], [output_tensor])
+    model = onnx.helper.make_model(graph)
+    tmp = tempfile.NamedTemporaryFile(suffix='.onnx')
+    onnx.save(model, tmp.name)
+    tmp.flush()
+    model = onnx_load(tmp.name)
+    runner = OnnxRunner(model)
```
chenyuxyz, line 24: too much copy-pasting for this block

geohotstan: There're slight variations to how the model is defined. 
But for the tempfile stuff, I already wrote a helper function in #10789 for it :D
Will clean it up.

PR #10801, opened 2025-06-13
[Bounty] Singular Value Decomposition

NoahKusaba:
# SVD Implementation Notes

## Overview
Implemented SVD using Householder reflection QR decomposition.

## Supported Parameters
The implementation supports parameters from numpy including:
- **`full_matrices`**: `bool` → Adding missing columns for U and V
- **`compute_uv`**: `bool`

## Extra added methods to Tensor
.eig()
.diag()
.norm()
.qr_decompose()

## Accuracy Results
I've achieved accuracy compared to numpy within `rtol = 1e-4`, with more variability for the extra columns during the `full_matrices` case.

> **Note**: Tests comparing to numpy are using absolute values, due to sign changes between implementations. If you actually calculate the recreated A matrix of `A = U @ S @ Vt`, it computes to a valid SVD().

## Accuracy Limitations
To achieve higher accuracy, you would need to copy the implementation that numpy uses from the LAPACK linalg package in Fortran from: https://github.com/Reference-LAPACK/lapack/blob/master/SRC/dgesvd.f

### LAPACK Implementation Details
The LAPACK implementation uses:
- **Golub-Kahan bidiagonalization** algorithm with Householder reflectors
- Additional accuracy-boosting techniques:
  - Wilkinson shifts
  - Givens rotations
  - Other optimization tricks

### Complexity Considerations
The multiple Fortran scripts for SVD total **several thousand lines of code**. The complexity required to execute it to the same precision falls way outside the scope of this bounty, which I believe is simply to "get something working for SVD()".

## Feedback Request
Hoping to get some feedback on the current implementation so far.


geohot:
Are tests passing for this? Will lock bounty on passing tests.

NoahKusaba:
@geohot I'll finish it tonight.

That being said, I'm happy to split the bounty with @F

NoahKusaba:
This PR nailed it https://github.com/tinygrad/tinygrad/pull/10875

PR #10803, opened 2025-06-13
Adapt `>>` for division by power of two to all ints

S-Lykles:
Arithmatic shift is always floored divison, the `+c-1` turns floored division into ceiled division when x is negative, making this truncating divison

### DIFF
--- a/test/test_uops.py
+++ b/test/test_uops.py
@@ -361,15 +361,16 @@ class TestAssembly(unittest.TestCase):
     self.assertIn(Ops.MUL, ops)
 
   def test_division_power_of_two(self):
-    g = UOp(Ops.DEFINE_GLOBAL, dtypes.int32.ptr(), (), 0)
-    c = UOp(Ops.CONST, dtypes.int, (), 2)
-    l = UOp(Ops.LOAD, dtypes.int, (g.index(c),))
-    a = UOp(Ops.IDIV, dtypes.int, (l, c))
-    uops = to_uops_list([a], opts=Device[Device.DEFAULT].renderer)
-    Device[Device.DEFAULT].renderer.render(uops)
-    ops = [x.op for x in uops]
-    self.assertIn(Ops.SHR, ops)
-    self.assertNotIn(Ops.IDIV, ops)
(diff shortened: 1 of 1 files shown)


S-Lykles:
Using .simplify() on the condition gave incorrect results on some cpu backends which is a bit weird

tinygrad/uop/ops.py
```diff
@@ -169,9 +169,10 @@ def size(self) -> int: return self.arg[0] if self.op is Ops.BUFFER_VIEW else sel
 
   # *** uop evaluation ***
 
-  def simplify(self):
+  def simplify(self, no_rewrite=False):
```
chenyuxyz: this arg is confusing. it still is a rewrite but with only one rule. why do you need this?

test/test_uops.py
```diff
@@ -361,10 +361,10 @@ def test_bitshift_left(self):
     self.assertIn(Ops.MUL, ops)
 
   def test_division_power_of_two(self):
-    g = UOp(Ops.DEFINE_GLOBAL, dtypes.uint32.ptr(), (), 0)
-    c = UOp(Ops.CONST, dtypes.uint, (), 2)
-    l = UOp(Ops.LOAD, dtypes.uint, (g.index(c),))
-    a = UOp(Ops.IDIV, dtypes.uint, (l, c))
+    g = UOp(Ops.DEFINE_GLOBAL, dtypes.int32.ptr(), (), 0)
```
chenyuxyz: if uint32 does not work anymore there should be a test to reflect that

S-Lykles: uint32 does still work, previously int32 didn't work. Going from uint32 to int32 just makes the test more generic, Im happy to test both as well

tinygrad/uop/ops.py
S-Lykles: So the problem is that simplify uses full symbolic to simplify, whereas these late rewrite patterns are in a pattern matcher together with symbolic_simple. So if I do .simplify() part of the graph will be simplified using full symbolic which I don't want. I just need to simplify with this one particular rule for the condition to fold into True or False. But yeah the arg is a bit poorly named

PR #10804, opened 2025-06-13
fix buffer GC with VIZ=1

Ziad0dev:
Fixes buffer garbage collection when VIZ=1 is enabled. Uses weak references in TrackedGraphRewrite to break circular references that prevented buffer cleanup.


geohot:
Does this need 24 lines? And I'm not talking about code golf, why does it have to be so complex?

Qazalin:
Fix is upstreamed.

PR #10805, opened 2025-06-13
corealize forward and gradients in test_ops

chenyuxyz:
for speed. a few nan value tests changed behavior, added forward_only for now


chenyuxyz:
not that faster...

PR #10806, opened 2025-06-13
xor self folding

leopf:
relevant for keccak. Reduces scheduling time a bit (2900ms -> 2200ms on my machine).

example:
```python

a = Tensor([ 1, 2 ], dtype=dtypes.int)
b = Tensor([ 1337 ], dtype=dtypes.int).pad((0, 1))

(a ^ b).realize()
```

before:
```c
void E_2(int* restrict data0, int* restrict data1, int* restrict data2) {
  int val0 = *(data1+0);
  int val1 = *(data2+0);
  int val2 = *(data1+1);
  *(data0+0) = (val0^val1);
  *(data0+1) = (val2^0);
}
```
after:
```c
void E_2(int* restrict data0, int* restrict data1, int* restrict data2) {
  int val0 = *(data1+0);
  int val1 = *(data2+0);
  int val2 = *(data1+1);
  *(data0+1) = val2;
  *(data0+0) = (val0^val1);
}
```

### DIFF
--- a/test/unit/test_uop_symbolic.py
+++ b/test/unit/test_uop_symbolic.py
@@ -110,6 +110,9 @@ class TestSymbolic(unittest.TestCase):
   def test_neg(self):
     self.helper_test_variable(-Variable("a", 0, 8), -8, 0, "(a*-1)")
 
+  def test_xor_0(self):
+    self.helper_test_variable(Variable("a", 0, 8) ^ 0, 0, 8, "a")
+
   def test_add_1(self):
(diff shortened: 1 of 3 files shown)


chenyuxyz:
can you add test in https://github.com/tinygrad/tinygrad/blob/master/test/unit/test_uop_symbolic.py?

leopf:
Had to add z3 support for bitwise ops (bit vectors...). Can be extended for OR and AND too, as needed.

PR #10813, opened 2025-06-14
feat: add initial svd code

fiskrt:
Does not work for larger than n>25 matrices
Would be easy if we had a tinygrad qr impl

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -4275,6 +4275,74 @@ class Tensor(MathTrait):
     ret = ret.reshape(bs, oy, ox, cout).permute(0,3,1,2)
     return ret if bias is None else ret.add(bias.reshape(1, -1, 1, 1))
 
+
+  def svd(self:Tensor, k:int=None, max_iter:int=15) -> tuple[Tensor, Tensor, Tensor]:
+    """
+    Args:
+        A: Input matrix (..., m, n)
+        k: Number of singular values to compute (default: min(m,n))
+        max_iter: Power iterations
+    
+    Returns:
+        U, S, V: SVD where A ≈ U @ diag(S) @ V.T
+
+    ```python exec="true" source="above" session="tensor" result="python"
+    t = Tensor.randn(5, 10)
+    U, S, V = t.svd()
+    print(U.numpy(), S.numpy(), V.numpy())
+    ```
+    """
+    batch_dims = self.shape[:-2]
+    A = self.reshape(-1, self.shape[-2], self.shape[-1])
+    batch_size, m, n = A.shape
+    if k is None: k = min(m, n)
+
+    def _power_iteration(M:Tensor, k:int, max_iter:int) -> Tensor:
+      batch_size, n, _ = M.shape
+      Q = Tensor.randn(batch_size, n, k)
+      Q = _gram_schmidt(Q)
+      for i in range(max_iter):
+        Q = M @ Q
+        Q = _gram_schmidt(Q)
+      return Q
+    if m >= n:
+      # Compute V from A^T A, then U = A @ V @ S^-1
+      AtA = A.transpose(-2, -1) @ A  # (batch, n, n)
+      V = _power_iteration(AtA, k, max_iter)
+      AV = A @ V  # (batch, m, k)
+      S = (AV * AV).sum(axis=-2).sqrt()  # (batch, k)
+      S_safe = S.clamp(min_=1e-12)
+      U = AV / S_safe.unsqueeze(-2)
+    else:
+      # Compute U from A A^T, then V = A^T @ U @ S^-1  
+      AAt = A @ A.transpose(-2, -1)  # (batch, m, m)
+      U = _power_iteration(AAt, k, max_iter)
+      AtU = A.transpose(-2, -1) @ U  # (batch, n, k)
+      S = (AtU * AtU).sum(axis=-2).sqrt()  # (batch, k)
+      S_safe = S.clamp(min_=1e-12)
+      V = AtU / S_safe.unsqueeze(-2)
+    
+    S_sorted, idx = S.sort(dim=-1, descending=True)
+    U_sorted = U.gather(-1, idx.unsqueeze(-2).expand(-1, m, -1))
+    V_sorted = V.gather(-1, idx.unsqueeze(-2).expand(-1, n, -1))
+    # TODO: transpose V here to be consistent with torch api?
+    return U_sorted.reshape(*batch_dims, m, k), S_sorted.reshape(*batch_dims, k), V_sorted.reshape(*batch_dims, n, k)
+
+  def _gram_schmidt(self: Tensor) -> Tensor:
+    Q = self
+    batch_size, n, k = Q.shape
+    R_list = []
+    for i in range(k):
+      r = Q[:, :, i:i+1]  # shape: [B, N, 1]
+      for j in range(i):
+        # Project r onto each previous basis vector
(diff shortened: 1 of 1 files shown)


NoahKusaba:
I'm also deep in the weeds with my own PR for svd.

I think it's necessary to at least support some of the parameters from numpy's svd like, full_matrices.

I think Gram-schmidt is absolute dog-water for calculating the orthogonal columns you need to add in.

NoahKusaba:
Took the chance to run your code. --> also in the weeds doing this bounty.

I think you should make your SVD() have as much parity as reasonable to numpys. You should definitely have a full_matrices parameter.

I think you should copy my implementation for tests, where you reconstruct the input matrix from the U,S,VT. 

The accuracy/precision issue with the basic algorithms you can find online feels complicated, as they aren't up to snuff with what numpy provides.

George hasn't really defined expectations with accuracy of the svd, and getting this extremely accurate is a big lift for a 200$ bounty competition. I honestly wish there wasn't money on the line, as it would be less weird to work collaboratively. 

Just my two cents.

fiskrt:
Awesome that you're also working on this @NoahKusaba. Yea I agree `full_matrices=True` should be supported.

The Gram-Schmidt is definitely the bottle-neck here and I wonder if there's a better way.

tinygrad/tensor.py
```diff
@@ -4275,6 +4275,59 @@ def image_conv2d(self, weight:Tensor, bias:Tensor|None=None, groups=1, stride=1,
...
+    def power_iteration(M:Tensor, k:int, max_iter:int=20) -> Tensor:
+      B, n, _ = M.shape
+      Q, _ = Tensor.randn(B, n, k).qr()
+      for _ in range(max_iter): Q, _ = (M @ Q).qr()
+      return Q
+    # Assume A is tall
+    if m < n: A = A.transpose(-2, -1)
+    V = power_iteration(A.transpose(-2, -1) @ A, k)
+    AV = A @ V
+    S = (AV * AV).sum(axis=-2).sqrt().clamp(min_=1e-12)
```
TiborGY, line 4301: How was this hardcoded 1e-12 value determined? This feels like something that should be documented, if its a guesstimate.

tinygrad/tensor.py
```diff
@@ -4275,6 +4275,59 @@ def image_conv2d(self, weight:Tensor, bias:Tensor|None=None, groups=1, stride=1,
...
+    ```python exec="true" source="above" session="tensor" result="python"
+    A = Tensor.randn(5, 10)
+    U, S, Vt = A.svd()
+    print(U.numpy(), S.numpy(), Vt.numpy())
+    ```
+    """
+    *b, m, n = self.shape
+    A = self.reshape(-1, m, n)
+    if k is None: k = min(m, n)
+    def power_iteration(M:Tensor, k:int, max_iter:int=20) -> Tensor:
```
TiborGY, line 4292: What happens when things don't converge in 20 iterations? How was 20 determined to be a good number of iterations?

TiborGY:
> SVD for smaller matrices `n<64`, efficient for very rectangular matrices. 

Wait, a 64x64 matrix is like...microscopic. What happens when you try to compute the SVD of something like a 3000 x 10 000 matrix?

tinygrad/tensor.py
NoahKusaba: It's kind of arbitrary.
Typical solutions would set a tolerance to check against and "break", to ensure some consistency for accuracy;
However the power-iterations [@fiskrt](https://github.com/fiskrt) and I are using, don't really converge well on large matrices I think he left it out (also there is a limit to the accuracy of QR algorithm with householder reflections --> see my explanation in my pr on it).
I think 20 + is a good number, as from my experience too, it is fairly accurate there.

TiborGY: Well, if the goal is only to make something that works for tiny matrices, then fair enough, I guess? But unfortunately this means the range of applications for it will be limited. Good SVD would enable things like solving large rank-deficient linear systems.

NoahKusaba: Making an accurate svd() for large matrices is a gigantic undertaking.
If that is the expectation, I don't think anyone is completing this bounty for a long time.

Everyone just uses the LAPACK package (numpy, pytorch, other languages), because nobody wants to re-implement it from scratch.

TiborGY: I mean yeah, true, but that's how you end up with a mountain of dependencies, potentially including a Fortran77 compiler. And LAPACK is CPU only. Anyhow, if this is going to be tinygrad's SVD implementation for now, then all I ask is that its limitations should be documented.

fiskrt:
`torch.linalg.svd` uses many different highly optimized fortran/cuda algs  (`gesvdj`, `gesvda`, `gesvd`, `gesdd`, ...) for different device/matrix shape/condition and runs the following on M4 cpu with `gesdd` (divide-conquer optimized for big matrices) in 6 seconds:
```
a = torch.randn(10_000, 3000)
start = time.monotonic()
torch.linalg.svd(a)
print(f"{time.monotonic() - start:.2f} s")
```

I think we can expect at-least 100x slower for a pure python implementation which makes the SVD for big matrices hard.
Tinygrad can of-course speed it up but since the algorithms for SVD aren't just matmuls, it's not so easy to benefit. I went for the power iteration approach because I thought it was quite matmul heavy versus the others, also more friendly for autograd.
@TiborGY

tinygrad/tensor.py
fiskrt: It's there to safe-guard against zero divisions. But for sure it should be analysed more to see what matrix conditions lead to it being needed.

TiborGY: > It's there to safe-guard against zero divisions. But for sure it should be analysed more to see what matrix conditions lead to it being needed.

My first instinct is that, this threshold should depend on dtype. 1e-12 is probably not an equally good choice for FP16/FP32/FP64 matrices.

NoahKusaba:
> `torch.linalg.svd` uses many different highly optimized fortran/cuda algs (`gesvdj`, `gesvda`, `gesvd`, `gesdd`, ...) for different device/matrix shape/condition and runs the following on M4 cpu with `gesdd` (divide-conquer optimized for big matrices) in 6 seconds:
> 
> ```
> a = torch.randn(10_000, 3000)
> start = time.monotonic()
> torch.linalg.svd(a)
> print(f"{time.monotonic() - start:.2f} s")
> ```
> 
> I think we can expect at-least 100x slower for a pure python implementation which makes the SVD for big matrices hard. Tinygrad can of-course speed it up but since the algorithms for SVD aren't just matmuls, it's not so easy to benefit. I went for the power iteration approach because I thought it was quite matmul heavy versus the others, also more friendly for autograd. @TiborGY

I feel like if @geohot really does want a tinygrad svd implementation with comparable accuracy/precision to numpy/torch, he should split those sub-functions/algorithms into smaller bounties with a holistic bounty to stitch it them together. That way we can split the work, as its a monumental task to do alone (unless there is some easier way to do this).

Also @fiskrt I think George mentioned a potential rewrite of the backend into rust at some point, once all of the algorithms are determined and hardened, which should solve the speed issue (although Jitting probably solves most of it).

ccoder1233:
> I feel like if @geohot really does want a tinygrad svd implementation with comparable accuracy/precision to numpy/torch, he should split those sub-functions/algorithms into smaller bounties with a holistic bounty to stitch it them together. That way we can split the work, as its a monumental task to do alone (unless there is some easier way to do this).

Maybe worth considering whether Tinygrad wants to implement a subset of Torch's linalg module?

geohot:
Are tests passing for this? Will lock bounty on passing tests, the current check is fake.

PR #10820, opened 2025-06-15
STFT for whisper example in tinygrad without librosa

IntendedConsequence:
WIP. Tested whisper with new stft on some clip, transcription seems to match. I don't know where best to put the functions, temporarily put them in Tensor for easier access from a test.

### DIFF
--- a/examples/whisper.py
+++ b/examples/whisper.py
@@ -157,8 +157,12 @@ def prep_audio(waveforms: List[np.ndarray], batch_size: int, truncate=False) -> 
     # we could have a symbolic batch_size dim instead of manually padding here if conv/layernorm supported symbolic shapes
     waveforms = np.pad(waveforms, pad_width=((0, batch_size - waveforms.shape[0]), (0, 0)))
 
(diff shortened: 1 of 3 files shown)


geohot:
I doubt that stuff belongs in tensor.py

IntendedConsequence:
First working WEBGPU export of whisper's encoder and decoder. Very rough, float32, doesn't include audio preprocessing yet, uses mel_spec.bin dump processed in advance for `test/models/whisper/test.wav`.
Demo at [https://intendedconsequence.github.io/whisper-webgpu/](https://intendedconsequence.github.io/whisper-webgpu/)
Should load 140MB(!) weights and print out streaming transcription result.

IntendedConsequence:
Added stft+mel audio preprocessing to whisper compile, and export to f32 should work even if WEBGPU backend doesn't support it.

IntendedConsequence:
Mostly a sign of life push, still experimenting with whisperx approach of using a helper VAD, no context and no timestamps to reduce hallucinations and loops. This should allow a significant speed boost since without context and serially-dependent seeking from whisper-emitted timestamps we'll be able to batch the whole thing which will make larger models more viable.

PR #10825, opened 2025-06-15
feat/amd-mi300x-runtime-refactor

ascender1729:
### Description

This PR refactors the AMD MI300X runtime by introducing a flexible way to override the number of XCCs (Shader Core Complexes) through an environment variable. This change improves configurability when working with different MI300X hardware configurations.

### Changes Made

* Added a new helper function `parse_xccs` to safely parse the `XCCS` environment variable with boundary checks and default handling.
* Updated the `AMDDevice` initialization to use `parse_xccs` for determining the number of XCCs.
* Introduced a new unit test file `test_parse_xccs.py` to validate override behavior, including:

  * Default fallback
  * Proper override within limits
  * Graceful fallback on invalid or excessive values

### DIFF
--- a/test/test_parse_xccs.py
+++ b/test/test_parse_xccs.py
@@ -1,0 +1,31 @@
+import os, unittest
+from tinygrad.runtime.ops_amd import parse_xccs
+from tinygrad.helpers import getenv
+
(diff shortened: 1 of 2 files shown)


wozeparrot:
do not use AI

PR #10832, opened 2025-06-16
Remove incorrect rule for `x%-d -> (x%d)*-1`

S-Lykles:
For cdiv you can take out the sign for both the numerator and divisor but for cmod only the numerator


test/unit/test_uop_symbolic.py
```diff
@@ -271,6 +271,16 @@ def test_div_const_div_wrong_sign(self):
     a = Variable("a", 0, 124)
     self.helper_test_variable(((a-10)//2+10)//2, 2, 33, "((((a+-10)//2)+10)//2)")
 
+  def test_neg_div(self):
```
chenyuxyz: can you combine these tests with existing `test_div_neg`?

PR #10833, opened 2025-06-16
Change z3 cdiv to euclidian division

S-Lykles:
So the z3 divison operator is actually euclidian division, which is floored division if d is positive. I added the rule:
```
(UPat.var("x") // UPat.var("d"), lambda x,d: -(x//(-d)) if d.vmax <=0 else None),
```
So that d was always positive and I could treat z3 division as floored division for the index validation.
But now for the symbolic fuzzing bounty I need to deal with negative divisors as well.

### DIFF
--- a/tinygrad/uop/spec.py
+++ b/tinygrad/uop/spec.py
@@ -6,7 +6,7 @@ try:
   import z3
 
   # IDIV is truncated division but z3 does euclidian division (floor if b>0 ceil otherwise); mod by power of two sometimes uses Ops.AND
-  def z3_cdiv(a, b):return z3.If(0<b, z3.If((a<0), (a+(b-1))/b, a/b), z3.If((a<0),(a-(b+1))/b, a/b))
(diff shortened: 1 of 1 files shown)


chenyuxyz:
interesting. do you know why z3 picked this div?

S-Lykles:
z3 uses the [SMT-lib definition](https://smt-lib.org/theories-Ints.shtml) (SMT-lib is not really a lib, its more a smt-language spec)
```
- div and mod according to Boute's Euclidean definition [1], that is,
    so as to satify the formula

    (forall ((m Int) (n Int))
      (=> (distinct n 0)
          (let ((q (div m n)) (r (mod m n)))
            (and (= m (+ (* n q) r))
                 (<= 0 r (- (abs n) 1))))))
```

S-Lykles:
Floored division does not always have a positive remainder if d is negative. Whereas euclidian division does always have a positive remainder

PR #10848, opened 2025-06-17
Change rate to int in test_tqdm

simone-pietro:
I’m working on issue #7889.  I’ve verified locally that all unit tests pass under `TYPED=1`. Since these changes touch multiple files, I’m splitting them into several pull requests.

In `test_tqdm`, `typeguard` currently raises an error because `tinytqdm` requires the `rate` parameter to be an integer. To fix this, I replaced the float literal `1e9` with the integer expression `10**9`.


simone-pietro:
Not sure why the pipeline is red. Can we rerun it?

PR #10849, opened 2025-06-17
Perform index validation on load/store, not on the index

S-Lykles:
@geohot this is what was causing OOB failures in beautiful_mnist. Problem was that the gates of the load/stores were not being considered in index validation.


S-Lykles:
I just want to unit test it a little more before I mark it ready

geohot:
Oh I see, if it's gated by an if. Good fix

PR #10859, opened 2025-06-18
Pass PythonRenderer instance to full_rewrite in eval_uop

simone-pietro:
I’m working on issue #7889 and need to ensure the unit tests run with TYPED=1. To simplify the review process, I’m splitting the necessary changes into several PRs. See also #10848.


simone-pietro:
A test seems unstable. Can we trigger another pipeline run?

PR #10863, opened 2025-06-18
bert grad accumulation

chenyuxyz:
training with half BS and GRADIENT_ACC_STEPS=4 https://wandb.ai/chenyuxyz/MLPerf-BERT/runs/kh4fijkx/overview


chenyuxyz:
need to update learning rate. but training looks fine

PR #10864, opened 2025-06-18
Cast int to str for render_cast

simone-pietro:
I’m working on issue #7889 and need to ensure the unit tests run with TYPED=1. To simplify the review process, I’m splitting the necessary changes into several PRs. See also #10848 and #10859.

In the snippet below, we pass an `int` instead of a `str` to `render_cast`:

``` Python
(UPat(Ops.CONST, (dtypes.int8, dtypes.int16), name="x"), lambda ctx,x: f"({ctx.render_cast(x.dtype, x.arg)})"),
```

This is why I've added `int` as a type hint. Alternatively, we could also use `str(x.arg)` in the lambda.


chenyuxyz:
`str(x.arg)` seems more consistent. passing int works but that's the implementation detail of `render_cast`

simone-pietro:
> `str(x.arg)` seems more consistent. passing int works but that's the implementation detail of `render_cast`

Thanks! Done, do I need to add [pr] for this change?

chenyuxyz:
this is fine. you can add [pr] in the future refactor/cleanups

PR #10869, opened 2025-06-18
support copyin/copyout/is_allocated for subbuffers

b1tg:


tinygrad/device.py
```diff
@@ -110,6 +110,7 @@ def __init__(self, device:str, size:int, dtype:DType, opaque:Any=None, options:O
     if isinstance(dtype, ImageDType): options = BufferSpec(image=dtype) # TODO: image hack shouldn't be here. where should it be?
     else: assert isinstance(dtype, DType) and not isinstance(dtype, PtrDType)
     self.device, self.size, self.dtype, self.options, self.offset, self.allocated_views = device, size, dtype, options, offset, 0
+    self.allocator: Allocator = Device[self.device].allocator
```
geohot: I like doing this late because it won't open the device if the Buffer isn't allocated. The problem with doing it here it that it always opens the device

geohot:
So I'm reviewing this, part of me just doesn't believe it could be this simple...tests look good though. Looks like it was supported at the lowest level, it just wasn't wired up.

geohot:
Congrats, bounty is yours. george@tinygrad.org to claim, can pay via PayPal or USDC on ETH

PR #10866, opened 2025-06-18
Generalize type hint for get_single_element [pr]

simone-pietro:
I’m working on issue #7889 and need to ensure the unit tests run with TYPED=1. To simplify the review process, I’m splitting the necessary changes into several PRs. See also #10848, #10859 and #10864.

`get_single_element` is often called with a `tuple[T]`. I've relaxed the type hint to `Sequence[T]` such that `typeguard` does not throw an error.

PR #10876, opened 2025-06-19
Cast ptr to int in test_from_mv_to_mv

simone-pietro:
I’m working on issue #7889 and need to ensure the unit tests run with TYPED=1. To simplify the review process, I’m splitting the necessary changes into several PRs. See also #10848, #10859, #10864 and #10866.

`to_mv` expects `ct` to be of type `int`. If you don’t cast it, `ct` is a `ctypes.Array` and causes a Typeguard error.

### DIFF
--- a/test/unit/test_helpers.py
+++ b/test/unit/test_helpers.py
@@ -1,4 +1,4 @@
-import gzip, unittest
+import ctypes, gzip, unittest
 from tinygrad import Variable
 from tinygrad.helpers import Context, ContextVar, argfix
 from tinygrad.helpers import merge_dicts, strip_parens, prod, round_up, fetch, fully_flatten, from_mv, to_mv, polyN, time_to_str, cdiv, cmod, getbits
@@ -182,7 +182,7 @@ class TestMemoryview(unittest.TestCase):
   def test_from_mv_to_mv(self):
     base = memoryview(bytearray(b"\x11\x22\x33"*40))
     ct = from_mv(base)
-    mv = to_mv(ct, len(base))
+    mv = to_mv(ctypes.addressof(ct), len(base))
     mv[0] = 2
     assert base[0] == 2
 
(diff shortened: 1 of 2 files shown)


simone-pietro:
ONNX CPU check failed. Can we rerun it?

test/unit/test_helpers.py
```diff
@@ -182,7 +182,7 @@ class TestMemoryview(unittest.TestCase):
   def test_from_mv_to_mv(self):
     base = memoryview(bytearray(b"\x11\x22\x33"*40))
     ct = from_mv(base)
```
chenyuxyz, line 184: also add a type for `from_mv`

simone-pietro: Done ✅

tinygrad/helpers.py
```diff
@@ -291,7 +291,7 @@ def capstone_flatdump(lib: bytes):
 # *** ctypes helpers
 
 # TODO: make this work with read only memoryviews (if possible)
-def from_mv(mv:memoryview, to_type=ctypes.c_char):
+def from_mv(mv:memoryview, to_type:type[ctypes._SimpleCData]=ctypes.c_char) -> ctypes.Array:
```
chenyuxyz, line 294: is the return type really `ctypes.Array`?

also @nimlgen can you help reviewing this one? thanks!

simone-pietro: As far as I can tell
```Python
ctypes.cast(<..>, ctypes.POINTER(to_type * len(mv)))
``` 
gives us a pointer to an array because of the multiplication in `ctypes.POINTER`. Since we dereference the pointer with `.contents`, the return type is an array.

nimlgen: lgtm

maybe we should consider deleting `from_mv` entirely. in all but one case it's used just to get addrs (and there's a lightweight `mv_address` for that). only `_MallocAllocator._offset` uses it otherwise, but we can probably inline it there.

PR #10877, opened 2025-06-19
small fix replacing download_file with fetch

NinoRisteski:
i added os as it was missing and replaced download file with fetch from helpers

### DIFF
--- a/examples/so_vits_svc.py
+++ b/examples/so_vits_svc.py
@@ -1,11 +1,11 @@
 # original implementation: https://github.com/svc-develop-team/so-vits-svc
 from __future__ import annotations
-import sys, logging, time, io, math, argparse, operator, numpy as np
+import sys, os, logging, time, io, math, argparse, operator, numpy as np
 from functools import partial, reduce
 from pathlib import Path
 from typing import Tuple, Optional, Type
 from tinygrad import nn, dtypes, Tensor
-from tinygrad.helpers import getenv
+from tinygrad.helpers import getenv, fetch
 from tinygrad.nn.state import torch_load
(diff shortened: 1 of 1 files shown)


examples/so_vits_svc.py
```diff
@@ -20,7 +20,7 @@
 F0_MEL_MAX = 1127 * np.log(1 + F0_MAX / 700)
 
 def download_if_not_present(file_path: Path, url: str):
```
chenyuxyz: remove this function and just use fetch?

NinoRisteski: yes, made those changes in the latest commit.

examples/so_vits_svc.py
```diff
@@ -97,7 +93,7 @@ def extract_features(self, source: Tensor, spk_emb:Tensor=None, padding_mask=Non
     return res, padding_mask
   @classmethod
   def load_from_pretrained(cls, checkpoint_path:str, checkpoint_url:str) -> ContentVec:
-    download_if_not_present(checkpoint_path, checkpoint_url)
+    if not os.path.isfile(checkpoint_path): fetch(checkpoint_url, checkpoint_path)
```
chenyuxyz: still need `if not os.path.isfile`?

NinoRisteski: you are right. sorry for rushing. removed those

chenyuxyz:
have you tested this from a fresh env so made sure it works?

also please update pr title

NinoRisteski:
> have you tested this from a fresh env so made sure it works?

yes, fetch works for me.
> 
> also please update pr title

updated

PR #10880, opened 2025-06-19
fix test logcumsumexp broken devectorize=0

NinoRisteski:
my attempt to fix the broken devectorize=0 which failed because of nan values. I propose a fix by applying the mask before the exponential rather than after, with mask value 1e4.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -2073,7 +2073,7 @@ class Tensor(MathTrait):
     m = self.max(axis=axis, keepdim=True)
     return (self - m).exp().sum(axis=axis, keepdim=keepdim).log() + m.squeeze(axis)
 
-  def logcumsumexp(self, axis=0, mask_val:float=-1e4) -> Tensor:
+  def logcumsumexp(self, axis=0) -> Tensor:
     """
     Computes the log-cumsum-exp of the tensor along the specified axis or axes.
 
@@ -2105,7 +2105,7 @@ class Tensor(MathTrait):
     x_cummax = x_reshaped.cummax(-1).unsqueeze(-1)
     x_expand = x_reshaped.unsqueeze(1).expand(*x_reshaped.shape, last_dim_size)
     mask = Tensor.ones(last_dim_size, last_dim_size, requires_grad=False, device=self.device).tril().unsqueeze(0)
-    ret = mask.where(x_expand - x_cummax, mask_val).exp().sum(-1).log() + x_cummax.squeeze(-1)
+    ret = mask.where(x_expand - x_cummax, dtypes.min(self.dtype)).exp().sum(-1).log() + x_cummax.squeeze(-1)
     return ret.reshape(*x.shape).transpose(-1, axis)
 
   def argmax(self, axis=None, keepdim=False) -> Tensor:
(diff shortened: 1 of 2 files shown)


tinygrad/tensor.py
```diff
@@ -2105,7 +2105,7 @@ def logcumsumexp(self, axis=0) -> Tensor:
     x_cummax = x_reshaped.cummax(-1).unsqueeze(-1)
     x_expand = x_reshaped.unsqueeze(1).expand(*x_reshaped.shape, last_dim_size)
     mask = Tensor.ones(last_dim_size, last_dim_size, requires_grad=False, device=self.device).tril().unsqueeze(0)
-    ret = ((x_expand - x_cummax).exp() * mask).sum(-1).log() + x_cummax.squeeze(-1)
+    ret = mask.where(x_expand - x_cummax, mask_val).exp().sum(-1).log() + x_cummax.squeeze(-1)
```
chenyuxyz: i think `dtype.min(self.dtype)` is more principled than an arbitrary value

NinoRisteski: yes, changed. Thanks for that!

tinygrad/tensor.py
```diff
@@ -2105,7 +2105,7 @@ def logcumsumexp(self, axis=0) -> Tensor:
     x_cummax = x_reshaped.cummax(-1).unsqueeze(-1)
     x_expand = x_reshaped.unsqueeze(1).expand(*x_reshaped.shape, last_dim_size)
     mask = Tensor.ones(last_dim_size, last_dim_size, requires_grad=False, device=self.device).tril().unsqueeze(0)
-    ret = ((x_expand - x_cummax).exp() * mask).sum(-1).log() + x_cummax.squeeze(-1)
```
chenyuxyz, line 2108: so the issue is `exp` returns `nan` and `nan*0` is bad?

why does it only happen with `DEVECTORIZE`?

NinoRisteski: the issue is: exp returns inf, then inf * 0 returns nan, and this only happens with devectorize=0 

to my understanding,
with devectorize=0:
Vector ops like exp * mask are computed exactly as written
No op reordering or optimization

with devectorize=1:
vector ops are devectorized into scalar operations
the devectorizer can reorder or optimize the computation
handle inf * 0 in optimized ways that produce 0 instead of nan

chenyuxyz: how can exp returns inf if `x_expand - x_cummax` is non-positive though?

NinoRisteski: it can be positive in some positions because it computes (x_expand - x_cummax).exp() * mask where x_expand creates a matrix that compares each element against the cumulative max at every position, not just its own position. 

For input [0.0, 100.0], 100.0 gets compared to the cumulative max at position 0 creating positive diff in a position that will later be masked out. but exp(100.0) = inf is computed before the mask is applied, and when inf is multiplied by the mask value 0, it produces inf * 0 = nan, which happens with `DEVECTORIZE=0`

NinoRisteski: Not really convinced? 
I tested it yesterday and devectorize 0 was good. Now both 0 and 1 are working, and tests passed of course.

PR #10884, opened 2025-06-19
add 405B params to llama3.py

chenyuxyz:
tested with `python examples/llama3.py --model /raid/weights/llama31_405b/ --size 405B --shard 8 --benchmark` on tinyamd2


chenyuxyz:
`python examples/llama3.py --model /raid/weights/llama31_405b/ --size 405B --shard 8 --benchmark`
image

with search
image

PR #10890, opened 2025-06-20
WebGPU on Windows

wpmed92:
Add support for WebGPU on Windows.
Out of the 9 backends, WebGPU will be the second cross-platform (mac, win, linux) GPU-backed backend in tinygrad, after `GPU=1` (OpenCL).
On mac, the lib is installed through homebrew, on linux through curl, and on windows we prompt the user to pip install pydawn, although only the dawn dll is used from the installed package.
We could make the `support/webgpu.py` shorter by always getting the dawn binary from the pip installed package, regardless of platform.


setup.py
```diff
@@ -4,6 +4,7 @@
 from setuptools import setup
 
 directory = Path(__file__).resolve().parent
+
```
geohot: Why is this change here?

.github/actions/setup-tinygrad/action.yml
```diff
@@ -83,6 +83,7 @@ runs:
         python -m venv .venv
         if [[ "$RUNNER_OS" == "Windows" ]]; then
           source .venv/Scripts/activate
+          python -m pip install dawn-python
```
geohot: This should be in deps or pydeps, not here

wpmed92:
Added dawn-python to pydeps and removed the empty line from setup.py. Tests are green.

.github/actions/setup-tinygrad/action.yml
```diff
@@ -16,7 +16,7 @@ inputs:
   pydeps:
     description: 'Extra Python dependency groups (space separated)'
     required: false
-    default: ''
+    default: 'dawn-python'
```
geohot: This should not be the default! Only when you need it

wpmed92:
Fixed that `dawn-python` was in default `pydeps`, now it's only installed for Windows webgpu.

.github/actions/setup-tinygrad/action.yml
wpmed92: Fixed it

wpmed92:
Has an unrelated onnx test failure but it otherwise this is ready.

wpmed92:
@geohot this is ready, tests are green. I fixed the pydeps thing so now dawn-python is only installed on webgpu windows.

geohot:
Nice, merged!

PR #10891, opened 2025-06-20
minor onnx dropout cleanup

chenyuxyz:
we should consider removing numpy random and test it similar to test_randomness, unless how seed works is part of spec?


geohotstan:
I remember the test required the same randomness as np.
I looked into replacing it before but I didn't get too far. (chatgpt said it was not easy)

As for the different version of ops, do we want a full comprehensive op version manager? I wrote something for that before, but at the time we weren't sure it was worth the complexity.

*also my guess for the CI flakiness isn't from the OP..., there has been random segfaults occurring in random places during `external_test_onnx_backend.py`

chenyuxyz:
if it's not commonly used (among huggingface popular models), don't bother with new versions

PR #10900, opened 2025-06-20
opt transforms the ast into an optimized ast

geohot:


geohot:
kernel.py linearize remove in the next PR

geohot:
@Qazalin this breaks the link from the kernel graph to codegen, is there a way to register it or is it diving deep into the `Kernel` data structure? I see the `track_rewrites` on kernel doesn't have a name arg.

I'm going to merge this anyway since no tests broke. Can you fix and add a test for that?

PR #10911, opened 2025-06-21
Make sure buffers are GCed on CPU and with VIZ

madaxer:
Not pretty sure if that was the requirement from Bounties , just to create tests for gc on CPU


Qazalin:
Does not solve the bounty.

madaxer:
> Does not solve the bounty.

ok, thank you, would give a try one more time, if more details would be on that bounty

PR #10914, opened 2025-06-21
remove np from beautiful_cartpole

NinoRisteski:
replacing numpy dep with tinygrad operations in cartpole and it should stay on device with stack.

### DIFF
--- a/examples/beautiful_cartpole.py
+++ b/examples/beautiful_cartpole.py
@@ -3,7 +3,6 @@ import time
 from tinygrad import Tensor, TinyJit, nn
 import gymnasium as gym
 from tinygrad.helpers import trange
-import numpy as np  # TODO: remove numpy import
 
 ENVIRONMENT_NAME = 'CartPole-v1'
 #ENVIRONMENT_NAME = 'LunarLander-v2'
@@ -86,7 +85,7 @@ if __name__ == "__main__":
   for episode_number in (t:=trange(EPISODES)):
     get_action.reset()   # NOTE: if you don't reset the jit here it captures the wrong model on the first run through
 
-    obs:np.ndarray = env.reset()[0]
+    obs = env.reset()[0]
     rews, terminated, truncated = [], False, False
     # NOTE: we don't want to early stop since then the rewards are wrong for the last episode
     while not terminated and not truncated:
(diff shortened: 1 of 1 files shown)


examples/beautiful_cartpole.py
```diff
@@ -86,30 +85,27 @@ def get_action(obs:Tensor) -> Tensor:
...
+    obs = env.reset()[0]
     rews, terminated, truncated = [], False, False
     # NOTE: we don't want to early stop since then the rewards are wrong for the last episode
     while not terminated and not truncated:
       # pick actions
       # TODO: what's the temperature here?
       act = get_action(Tensor(obs)).item()
 
-      # save this state action pair
-      # TODO: don't use np.copy here on the CPU, what's the tinygrad way to do this and keep on device? need __setitem__ assignment
```
chenyuxyz, line 98: stacking is slow and we want setitem to work here

examples/beautiful_cartpole.py
```diff
@@ -86,30 +85,27 @@ def get_action(obs:Tensor) -> Tensor:
...
-    # reward to go
-    # TODO: move this into tinygrad
-    discounts = np.power(DISCOUNT_FACTOR, np.arange(len(rews)))
-    Rn += [np.sum(rews[i:] * discounts[:len(rews)-i]) for i in range(len(rews))]
+    discounts = Tensor([DISCOUNT_FACTOR ** i for i in range(len(rews))])
+    Rn += [(Tensor(rews[i:]) * discounts[:len(rews)-i]).sum().item() for i in range(len(rews))]
 
     Xn, An, Rn = Xn[-REPLAY_BUFFER_SIZE:], An[-REPLAY_BUFFER_SIZE:], Rn[-REPLAY_BUFFER_SIZE:]
-    X, A, R = Tensor(Xn), Tensor(An), Tensor(Rn)
+    X = Xn[0].stack(*Xn[1:]) if len(Xn) > 1 else (Xn[0].unsqueeze(0) if len(Xn) == 1 else Tensor([]))
```
geohot: This is a lot harder to read

examples/beautiful_cartpole.py
NinoRisteski: Initially I see setitem as slower... I guess I am doing something wrong? Any tips come to mind?

examples/beautiful_cartpole.py
NinoRisteski: I think now is better. Still have to replace stacking though.

examples/beautiful_cartpole.py
NinoRisteski: can you check my last commit on this? thx

PR #10915, opened 2025-06-21
ONNX external_test_onnx_backend use PYTHON device for model

geohotstan:
I just want to see how much slower this is.
This seems to be one potential fix for the flaky `test/external/external_test_onnx_backend.py` (fixes the os errors I'm having locally 😃 , but is quite a bit slower locally)

another potential fix I might try is manual file management instead of tmpfile

### DIFF
--- a/test/external/external_test_onnx_backend.py
+++ b/test/external/external_test_onnx_backend.py
@@ -30,7 +30,7 @@ class TinygradBackend(Backend):
     input_initializer = [x.name for x in model.graph.initializer]
     net_feed_input = [x for x in input_all if x not in input_initializer]
     print("prepare", cls, device, net_feed_input)
-    model = Tensor(model.SerializeToString(), device="PYTHON") # test
+    model = Tensor(model.SerializeToString(), device="PYTHON")
     run_onnx = OnnxRunner(onnx_load(model))
     return TinygradModel(run_onnx, net_feed_input)
 


geohotstan:
Test ONNX (GPU): 1m17s -> 7m23s
YIIIKES! 💀

geohotstan:
Currently
Test ONNX (GPU): 1m17s -> 2m13s
And fixes the flakiness issue.

I have not root-caused the previous flakiness.

Can we take doubling the time it takes for that test to get rid (most likely!) of the flakiness issue in `external_test_onnx_backend`?

If not, I'll continue to try to root cause.

chenyuxyz:
few questions
(1) can you reproduce the flakiness consistently? is it only for this test?
(2) why does this change fix the flakiness?
(3) why is this change slower?

geohotstan:
> few questions (1) can you reproduce the flakiness consistently? is it only for this test? (2) why does this change fix the flakiness? (3) why is this change slower?

1) From the cases I saw this flakiness, it was either during the run for `external_test_onnx_backend.py` or some test after running `external_test_onnx_backend.py`. I cannot reproduce the flakiness consistently, but LLM says it is related to `pytest -n=auto` usage and `tempfile` usage. Running this test locally consistently fails for me with a bunch of os errors after 27~% of the tests finishes saying too many files opened, but this current fix fixes that.

2) Directly having the model in-memory without having to go to disk would be a solution here. 

3) I need some time on this :o. The flakiness seems really annoying so I just threw this out asap.

Also I know it's ill-advised to use PYTHON backend for running, but hmmm.

geohotstan:
In keeping the modelproto in memory, we'd have to put the bytes modelproto either on "CPU" or "PYTHON".

CI speeds as reference:

Parse model in-memory on CPU
Test ONNX (GPU): 5m13s
```
model_bytes = model.SerializeToString()
model = Tensor(model_bytes, device="CPU")
run_onnx = OnnxRunner(onnx_load(model))
```

Parse model in-memory on PYTHON
Test ONNX (GPU): 2m13s
```
model_bytes = model.SerializeToString()
model = Tensor(model_bytes, device="PYTHON")
run_onnx = OnnxRunner(onnx_load(model))
```

Save model and load from disk
Test ONNX (GPU): 1m17s
This just uses tempfile.


---

*IF* we put it on PYTHON, the PYTHON backend is currently responsible for `Tensor()[start:end].data()` from `TensorIO.readinto()` and occasionally a bitcast or two.
```
class TensorIO(io.RawIOBase, BinaryIO):
  ...
  def readinto(self, buffer: Any) -> int:
    data = self._tensor[self._position:self._position+len(buffer)].data()
    buffer[:len(data)] = data
    self._position += len(data)
    return len(data)
```

Now the right way is probably CPU, but CPU is much slower.
I took the Profiler for a spin here
For the most common operation in onnx_load: `Tensor()[start:end].data()`
```
data = np.arange(100, dtype=np.int32).tobytes()
st = time.time()
t = Tensor(data, device="PYTHON")  # or CPU
with Profiling():
   t[3:10].data()
print((time.time() - st) * 1000, "ms")
```

the Profiling tot time for CPU is double that of PYTHON. Wall clock time is also longer.
The actual E_ kernel time for CPU is very quick, but I think much of the time is spent on compilation and lowering. The Python backend just does straight up memoryview operations, which turns out to be fast.

The model sizes for external_test_onnx_backend.py are all very small, so many many unnecessary compilation and extra rewrites adds up. (due to TensorIO usage)
This is what I gathered about the slowness.

---
Maybe there can be some future feature for onnx_load to deal with ModelProto users... where it takes in a pure bytes object that is fed into BufferedReader that directly does this in memory, instead of using PYTHON backend.

geohotstan:
I'm gonna use this to trigger CI a few more times just to make sure it's fixed.

chenyuxyz:
adds about 10 seconds to each of the comma benchmark tests, seems fine if it fixed the segfault

PR #10921, opened 2025-06-22
update codegen process replay to get_program [pr]

Qazalin:
Updating process replay to use the new get_program API. Sadly, kernel.py is still here. I'll put up an API change diff that removes usage of the Kernel class.

After commit https://github.com/tinygrad/tinygrad/commit/b09c47366f8578beb319817a235af17c0b376949, process replay did not run on codegen:
Screenshot 2025-06-23 at 5 03 32 PM

This diff re-enables it:
Screenshot 2025-06-23 at 5 06 28 PM


geohot:
name_override is fine as get_program arg if you need it.

Qazalin:
Re: name_override, string replace works fine. I actually prefer it over changing the API.

Failing:

- [x] TC=2 #10922
- [x] Openpilot https://github.com/tinygrad/tinygrad/pull/10921/commits/2c8f74953a428aacfd0cc9ed04824f56449528f8
- [x] BEAM cache [not hitting](https://github.com/tinygrad/tinygrad/actions/runs/15810764076/job/44561747801?pr=10921#step:7:350)?
- [x] BEAM/onnx test https://github.com/tinygrad/tinygrad/actions/runs/15810764076/job/44561747761?pr=10921#step:11:335
- [ ] llama beam benchmark

This non-reproducibility is likely because kernel.py is relying on state outside of input args and ContextVars.

test/external/process_replay/process_replay.py
```diff
@@ -67,8 +63,12 @@ def diff(offset:int) -> None:
       break
     try:
       name, args, kwargs, ctx_vals, loc, ret = pickle.loads(row[0])
-      ctx_vars = {k:v.value for k,v in ctx_vals.items() if k != "DEBUG" and (var:=ContextVar._cache.get(k)) is not None and var.value != v.value}
       if (replayer:=replayers.get(name)) is None: continue
+      ctx_vars:dict[str, int] = {}
+      for k,v in ctx_vals.items():
+        # process replay does not allow beaming or debug prints
+        if k not in {"IGNORE_BEAM_CACHE", "DEBUG"} and (default:=ContextVar._cache.get(k)) is not None and default.value != v.value:
```
Qazalin: BEAM cannot be process replayed since it is non-deterministic. Only test_tiny test_beam seems to set IGNORE_BEAM_CACHE=1. Otherwise, Using process replay with BEAM should be fine if the cache exists.

Qazalin:
BEAM still breaks
image

get_program should expose a lower level API that takes applied_opts, this is non-deterministic.

PR #10928, opened 2025-06-23
[Bounty] Switch the meaning of Reduce keepdim

eitanturok:
Most reduce ops are called with `keepdim=False` (e.g. `Tensor.sum` implicitly has `keepdim=False`).

But internally, `REDUCE_AXIS` and `REDUCE` produce outputs with `keepdim=True`.

So whenever the user sets `keepdim=False` we constantly have to add reshape operations to remove the extra 1 dimensions .

This is annoying + expensive.

Let's redesign `REDUCE_AXIS` and `REDUCE` so that by default they produce outputs with `keepdim=True`, sparring us the need to do extra reshapes.

### DIFF
--- a/tinygrad/shape/shapetracker.py
+++ b/tinygrad/shape/shapetracker.py
@@ -83,7 +83,7 @@ class ShapeTracker:
   @property
   def size(self) -> int: return self.views[-1].size()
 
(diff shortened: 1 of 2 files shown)


geohot:
Tests failing

PR #10939, opened 2025-06-23
remove kernelize from llama

geohot:


extra/models/llama.py
```diff
@@ -166,27 +166,29 @@ def sample(logits: Tensor, temp: float, k: int, p: float, af: float, ap: float):
...
 
-    self.freqs_cis = self.freqs_cis.cast(h.dtype).kernelize()
+    self.freqs_cis = self.freqs_cis.cast(h.dtype).contiguous()
     freqs_cis = self.freqs_cis[:, start_pos:start_pos+seqlen, :, :, :]
 
-    mask = Tensor.full((1, 1, seqlen, start_pos+seqlen), float("-inf"), dtype=h.dtype, device=h.device).triu(start_pos+1).kernelize() if seqlen > 1 else None
+    mask = Tensor.full((1, 1, seqlen, start_pos+seqlen), float("-inf"), dtype=h.dtype, device=h.device).triu(start_pos+1) if seqlen > 1 else None
     for layer in self.layers: h = layer(h, start_pos, freqs_cis, mask)
     logits = self.output(self.norm(h)).float()[:, -1, :]
+    if math.isnan(temperature): return logits
```
chenyuxyz, line 189: can temperature be nan?

geohot: if you pass in nan it'll trigger this. it could be a flag, but idk this is simpler?

PR #10940, opened 2025-06-23
[bounty] whisper in pure tinygrad on webpage with WEBGPU

alec-build:
Write whisper in pure tinygrad (no js imports) and make it run on a website with WEBGPU backend


geohot:
Will lock when there's a working link to a website

PR #10950, opened 2025-06-24
add opts_to_apply option to AST KernelInfo

Qazalin:
Process replay and some other tests override OptOps with a custom sequence.

Currently, this is done through interfacing with the Kernel class. This diff moves this to the SINK UOp's argument.

Updating some of the tests to demonstrate use-case, I think this change enables removing kernel.py from the rest of the tests as well.

The get_optimized_ast stage moves opts_to_apply to applied_opts.

opts_to_apply=None. No optimizations specified by the sink. Use the default path (BEAM, NOOPT, heuristic.py)
opts_to_apply=() -> Equivalent to NOOPT=1.


geohot:
I'm fine with this. The other alternative is to move get_optimized_ast out of get_program, but I'm not sure that improves things.

geohot:
Actually I have a different proposal. Put them on the sink in KernelInfo

Qazalin:
Yeah we could just put it in get_optimized_ast, that would make it two calls

PR #10958, opened 2025-06-24
clean up binary_crossentropy_logits

chenyuxyz:


chenyuxyz:
one more kernel for bert training, but both green and red benchmark are slightly faster so probably not well optimized before anyway

PR #10972, opened 2025-06-25
fix viz vscode link on windows

b1tg:
https://github.com/microsoft/vscode-docs/blob/main/docs/configure/command-line.md#opening-vs-code-with-urls


Qazalin:
oh I see it did vscode://filec:, fine with this. Trailing slashes gets removed elsewhere.

PR #10973, opened 2025-06-25
use modified_ast for ProgramSpec [pr]

ignaciosica:
Is there any particular reason we use the original ast for `ProgramSpec`?

I came across this while refactoring some calls to `linearize` method from `kernel` in the tests. I realized that if we use the original ast for the `ProgramSpec` and apply some opts through the new `opts_to_apply` api, the original ast won't correctly track the `applied_opts`. If we want the state to live in the uops, we need to use `modified_ast` for `ProgramSpec`.

Notice that in order to print the `applied_opts`, `modified_ast` is used in line 39.

https://github.com/ignaciosica/tinygrad/blob/cf60ccac6a44002c493c60739b4c4d44f801c50e/tinygrad/engine/realize.py#L17-L48

I'm asking this question and not considering this bug because the original ast was also used for ProgramSpec before all the refactors that have been taking place lately, so maybe there is a reason behind it.

### DIFF
--- a/tinygrad/engine/realize.py
+++ b/tinygrad/engine/realize.py
@@ -44,7 +44,7 @@ def get_program(ast:UOp, renderer:Renderer) -> ProgramSpec:
   if DEBUG >= 6: print_uops(uops)
   src = renderer.render(uops)
 
-  return ProgramSpec(uops[-1].arg.name, src, renderer.device, ast, uops,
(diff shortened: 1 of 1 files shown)


geohot:
We want to save the original ast. You can construct the modified ast from the opt on sink on the uops and the base ast. You can't do that in reverse.

ignaciosica:
I see, that makes sense, thanks

PR #10975, opened 2025-06-25
bugfix: 10850 - fixed input tensors via _find_tensors_recursive

rishavvajpayee:
## bugfix: 10850 - fixed input tensors via _find_tensors_recursive
1. added _find_tensors_recursive that recursively checks and finds Tensors at any nesting levels


rishavvajpayee:
as this makes the support for arrays possible in JIT the test is failing for test_array_jit.

but replacing the test case with only the else case of `np.testing.assert_allclose(c.numpy(), a.numpy()+b.numpy(), atol=1e-4, rtol=1e-5)` makes it work.

let me know if you want me to push that change too with this PR.

cc: @chenyuxyz

PR #10977, opened 2025-06-25
replace ast arg on get_optimized_ast call [pr]

ignaciosica:
It is either replacing the ast arg or adding back applied_opts to ProgramSpec, otherwise the applied opts are not going to be tracked correctly after get_optimized_ast call.

I opted for this approach to keep the state in the ast.

### DIFF
--- a/tinygrad/engine/realize.py
+++ b/tinygrad/engine/realize.py
@@ -27,14 +27,14 @@ def get_program(ast:UOp, renderer:Renderer) -> ProgramSpec:
   """
 
   if getenv("VIZ"): graph_rewrite(ast, PatternMatcher([]), name="View Base AST")
-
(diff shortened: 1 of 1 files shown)


Qazalin:
the expectation is that this line does it:
https://github.com/tinygrad/tinygrad/blob/98d2cde2932e65e9b329b7cef902ad527a5f30b4/tinygrad/opt/kernel.py#L456-L458
have a failing test?

ignaciosica:
Now that you pointed that out I checked and applied_opts are tracked not on the ast but on the uops. 

I'll upstream a failing test just to show the case, but this is no longer necessary if it is the intended behavior.

ignaciosica:
closing, I think the current behavior is the expected one

PR #10978, opened 2025-06-25
remove some linearize calls from tests [pr]

ignaciosica:
speed_compare_cuda_ptx
test_uop_spec
test_linearizer (some tests still call linearize here)
test_uops
test_winograd

I'll keep the pr's small

### DIFF
--- a/test/external/speed_compare_cuda_ptx.py
+++ b/test/external/speed_compare_cuda_ptx.py
@@ -33,7 +33,6 @@ if __name__ == "__main__":
     dev.compiler = PTXCompiler(dev.arch)
     lin = ast_str_to_lin(ast, opts=ptx)
     lin.apply_opts(hand_coded_optimizations(lin))
-    lin.linearize()
     ptx_prg = CompiledRunner(lin.to_program())
 
     # warmup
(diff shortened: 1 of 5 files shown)


test/external/speed_compare_cuda_ptx.py
```diff
@@ -33,7 +33,6 @@
     dev.compiler = PTXCompiler(dev.arch)
     lin = ast_str_to_lin(ast, opts=ptx)
     lin.apply_opts(hand_coded_optimizations(lin))
-    lin.linearize()
```
ignaciosica, line 36: linearize call is followed by to_program call, so the call to linearize is a noop

test/test_winograd.py
```diff
@@ -44,7 +44,6 @@ def test_speed(self):
       with Timing(f"linearize {i} with {len(ops):4d} ops: "):
         l = Kernel(s.ast)
         l.apply_opts(hand_coded_optimizations(l))
-        l.linearize()
```
ignaciosica, line 47: only sts are accessed from the kernel after the linearize call
sts do not change with the linearize call so the call to linearize is noop

test/unit/test_uop_spec.py
```diff
@@ -2,26 +2,26 @@
...
   sink = UOp(Ops.SINK, dtypes.void, stores)
   if DEBUG >= 3:
     for op in stores: print(op)
   try: type_verify(list(sink.toposort()), ast_spec)
   except RuntimeError as e: raise InvalidASTException(e.args)
-  k = Kernel(sink)
-  k.linearize()
-  if DEBUG >= 6: print_uops(k.uops)
-  if DEBUG >= 4: print(k.to_program().src)
-  return k
```
ignaciosica, line 24: kernel was returned but it wasn't actually being used

PR #10983, opened 2025-06-25
llama3 mlperf train

chenyuxyz:
work in progress. now it can overfit small examples and vram roughly matches


chenyuxyz:
merging this so people can run the example command. next step is to run the full eval on a model

PR #10984, opened 2025-06-26
remove onnx buffer_parse rebuild const

chenyuxyz:
do we need this?


chenyuxyz:
i remember there was a regression that const is not rebuilt and fused, and onnx became a lot slower. let's see if there are tests

PR #10991, opened 2025-06-26
Extra spec for renderers

ttomsa:
Backend specific spec should only run on the specific backend.


geohot:
There's two changes in here

ttomsa:
> There's two changes in here

removed

ttomsa:
@geohot is this good to merge?

ttomsa:
relevant for #11611

PR #11000, opened 2025-06-27
ONNX onnx_parser and buffer_parse clean up

geohotstan:
implement uint64 and double buffer parsing in onnx_parser
remove numpy from buffer_parse

### DIFF
--- a/test/external/external_test_onnx_runner.py
+++ b/test/external/external_test_onnx_runner.py
@@ -4,7 +4,6 @@ from tinygrad.frontend.onnx import OnnxRunner, onnx_load
 from tinygrad.device import is_dtype_supported
 from extra.onnx import data_types
 from hypothesis import given, settings, strategies as st
-import numpy as np
 
 device_supported_dtypes = [odt for odt, dtype in data_types.items() if is_dtype_supported(dtype)]
 device_unsupported_dtypes = [odt for odt, dtype in data_types.items() if not is_dtype_supported(dtype)]
--- a/test/external/external_test_onnx_runner.py
+++ b/test/external/external_test_onnx_runner.py
@@ -6,7 +6,6 @@ from extra.onnx import data_types
 from hypothesis import given, settings, strategies as st
 import numpy as np
 
-data_types.pop(16) # TODO: this is bf16, need to support double parsing first.
 device_supported_dtypes = [odt for odt, dtype in data_types.items() if is_dtype_supported(dtype)]
 device_unsupported_dtypes = [odt for odt, dtype in data_types.items() if not is_dtype_supported(dtype)]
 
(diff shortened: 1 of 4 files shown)


geohotstan:
re [52638f1](https://github.com/tinygrad/tinygrad/pull/11000/commits/52638f11b8706ef021ae23b3811437a752a78992)  commit msg: it was because no WEBGPU tensor was being realized before 😅 
We're also testing WEBGPU=1 in `Unit Tests / ONNX (CPU) Tests` which is kinda weird... 🤔

test/external/external_test_onnx_runner.py
```diff
@@ -1,65 +1,139 @@
...
-  def _test_node_attribute_dtype(self, onnx_data_type, tinygrad_dtype):
-    arr = np.array([0, 1], dtype=onnx.helper.tensor_dtype_to_np_dtype(onnx_data_type))
-    output_tensor = onnx.helper.make_tensor_value_info('output', onnx_data_type, arr.shape)
-    value_tensor = onnx.helper.make_tensor('value', onnx_data_type, arr.shape, arr.tobytes(), raw=True)
-    node = onnx.helper.make_node('Constant', inputs=[], outputs=['output'], value=value_tensor)
-    graph = onnx.helper.make_graph([node], 'attribute_test', [], [output_tensor])
+  def test_const_fold_from_memory(self):
+    self._test_const_fold_unary_op(False)
+    # TODO: understand this and fix this, bitcast related
+    # self._test_const_fold_binary_op(False)
```
geohotstan, line 67: the reason for testing const folding here is because I remember a long time ago I introduced a regression where it wasn't const folding. 
That should've been the test, but I couldn't find the regression before unfortunately.

I just ended up copying `test_const_folding.py` over. It does test const folding though.

geohotstan:
@chenyuxyz 
also 
`PYTHONPATH=. GPU=1 python examples/openpilot/compile3.py ../openpilot/selfdrive/modeld/models/driving_vision.onnx` 

is guaranteed to hit the error on this branch
```
...
  File "/Users/zibo/fun/tiny/tinygrad/tinygrad/uop/ops.py", line 737, in rewrite
    if (ret:=match(uop, ctx)) is not None and ret is not uop: return ret
             ~~~~~^^^^^^^^^^
  File "<string>", line 3, in compiled_match
  File "/Users/zibo/fun/tiny/tinygrad/tinygrad/uop/symbolic.py", line 22, in fold_bitcast
    return root.const_like(convert(c.arg) if root.dtype.count == 1 else tuple(map(convert, c.arg)))
                           ~~~~~~~^^^^^^^
  File "/Users/zibo/fun/tiny/tinygrad/tinygrad/uop/symbolic.py", line 21, in convert
    def convert(v:Any): return struct.unpack(to_fmt, struct.pack(from_fmt, v))[0]
                               ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
struct.error: unpack requires a buffer of 2 bytes
```

Before the code path was doing numpy buffer loading instead of bitcast, which is why it passed.

---
I can give this a look also when I wake up tmrw if that should be fixed first.

geohotstan:
confirms that after merging master with #11236 
`driving_monitor.onnx` now compiles properly with this branch too.

PR #11007, opened 2025-06-27
image_dot of 2 half inputs returns half

### LINKED ISSUE #6378: image_dot of 2 half inputs returns in half instead of float

utkarshgill:
Fix #6378

When we call `.sum(dtype=None)`, the sum operation defaults to fp32 accumulation. Nothing was converting it back to the expected fp16 result

The fix is to add simple dtype check at the end of `image_conv2d`. If no explicit dtype is requested and inputs are fp16/bf16, cast the result back to the expected output type

```python
if dtype is None and (ret_dtype := least_upper_dtype(self.dtype, weight.dtype)) in (dtypes.float16, dtypes.bfloat16):
    ret = ret.cast(ret_dtype)
```
This preserves the existing acc behavior (still uses fp32 for compute) while the final result matches what users expect from input dtypes. 

Also cleaned up the test skip condition for IMAGE


utkarshgill:
`GPU=1 IMAGE=2 python -m pytest test/test_ops.py -q -k test_gemm_fp16` passes now
```bash
engelbart@Utkarshs-MacBook-Pro tinygrad % PYTHONPATH=. GPU=1 IMAGE=2 python -m pytest test/test_ops.py -q -k test_gemm_fp16
/Users/engelbart/.pyenv/versions/3.11.9/lib/python3.11/site-packages/pytest_asyncio/plugin.py:217: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"

  warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))
.                                                                                                   [100%]
1 passed, 391 deselected in 1.06s
engelbart@Utkarshs-MacBook-Pro tinygrad % 
```

test/test_ops.py
```diff
@@ -1282,7 +1282,7 @@ def test_small_gemm_range(self):
                                                                          np.arange(64,128,dtype=np.float32).reshape(8,8)])
   def test_small_gemm_eye(self):
     helper_test_op(None, lambda x,y: x.matmul(y), lambda x,y: x@y, vals=[np.eye(8).astype(np.float32), np.eye(8).astype(np.float32)])
-  @unittest.skipIf(CI and Device.DEFAULT in ["NV", "LLVM", "GPU", "CUDA"] or IMAGE, "not supported on these in CI/IMAGE")
+  @unittest.skipIf(CI and Device.DEFAULT in ["NV", "LLVM", "GPU", "CUDA"], "not supported on these in CI")
```
geohot: Is this actually testing anything in CI?

utkarshgill: The fix in `image_conv2d` is correct; it passes locally with `GPU=1 IMAGE=2`, but we can't test it with real FP16 on CI with GPU=1 because opencl lacks FP16 support. It crashes with `error: casting to type 'half' is not allowed`

---
1. Added an explicit test that takes the image_dot → image_conv2d path without compiling kernels, just checking that the resultant dtype is correct. This gives broad coverage across all backends in lazy mode

    ```python
      def test_gemm_fp16_image_path_dtype(self):
        with Context(IMAGE=2):
          x = Tensor.rand(64, 64)
          y = Tensor.rand(64, 64)
          z = x.half().matmul(y.half()) # don't realize
          assert z.dtype == dtypes.half, f"Expected half, got {z.dtype}"
    ```

2. Added a job that runs in `Tensor Core tests` > `Test IMAGE=2 support`
    ```yaml
    # /tinygrad/.github/workflows/test.yml
    
    - name: Test IMAGE=2 support
      run: |
        IMAGE=2 PYTHON=1 python3 test/test_ops.py TestOps.test_gemm
    +   IMAGE=2 PYTHON=1 python3 test/test_ops.py TestOps.test_gemm_fp16
        IMAGE=2 PYTHON=1 python3 test/test_ops.py TestOps.test_simple_conv2d
    ```
---

**Both approach test correctly. I keep both for completeness.**

Long term solution is to add proper FP16 support to the OpenCL so `cl_khr_fp16` compiles in CI. we will need to fix the image_dot kernel generator so it never emits illegal `(half)(…)` casts inside `write_imagef`, or guard it by checking for `cl_khr_fp16`. That seems orthogonal to this IMAGE=2 bugfix, could be a bigger lift and probably deserves its own PR

utkarshgill:
`test_gemm_fp16` passes in CI under `Tensor Core tests` when IMAGE=2 is set:
```bash
Run IMAGE=2 PYTHON=1 python3 test/test_ops.py TestOps.test_gemm
test_gemm (__main__.TestOps.test_gemm) ... ok

Ran 1 test in 0.868s

OK
test_gemm_fp16 (__main__.TestOps.test_gemm_fp16) ... ok

Ran 1 test in 0.935s

OK
test_simple_conv2d (__main__.TestOps.test_simple_conv2d) ... ok

Ran 1 test in 0.5[26](https://github.com/tinygrad/tinygrad/actions/runs/15968585437/job/45034392055?pr=11007#step:4:27)s

OK
```

test/test_ops.py
```diff
@@ -1282,9 +1282,15 @@ def test_small_gemm_range(self):
...
   def test_gemm_fp16(self):
     helper_test_op([(64,64), (64,64)], lambda x,y: x.half().matmul(y.half()), atol=5e-3, rtol=5e-3)
+  def test_gemm_fp16_image_path_dtype(self):
```
chenyuxyz: this test is only for image and does not belong in test_ops

tinygrad/tensor.py
```diff
@@ -4245,6 +4245,7 @@ def image_conv2d(self, weight:Tensor, bias:Tensor|None=None, groups=1, stride=1,
 
     # NCHW output
     ret = ret.reshape(bs, oy, ox, cout).permute(0,3,1,2)
+    if dtype is None and (ret_dtype := least_upper_dtype(self.dtype, weight.dtype)) in (dtypes.float16, dtypes.bfloat16): ret = ret.cast(ret_dtype)
```
chenyuxyz, line 4314: why is the cast done here and not in the `x*w` mul?

utkarshgill: The multiply and reduce needs to be in fp32

If we down-cast the product to fp16 before the .sum(), the accumulation is also fp16 and we lose precision, and it crashes with numerical mismatch so when dtype isnt supplied

1. x*w is fp32
2. `.sum()` accumulates in fp32
3. after the reduce we cast back to fp16

And final tensor has expected dtype. This is similar to cast patterns used in Tensor.sum and Tensor.dot

test/test_ops.py
utkarshgill: Yes right. I’ll move the test into `test/test_image_dtype.py`

PR #11008, opened 2025-06-27
bert grad clipping start with const 0

chenyuxyz:
saved the init kernels


chenyuxyz:
trained fine https://wandb.ai/chenyuxyz/MLPerf-BERT/runs/yb9o30qc?nw=nwuserchenyuxyz

PR #11011, opened 2025-06-27
[bounty] Refactor VIEW(CONST(DEVICE))

S-Lykles:
So this does most of the changes I think you want in kernel/kernelize/grouper/spec

### Main changes
- added `kernel.views`, this fixup_optimized_ast is now way easier since you just find all the views and replace their shapetracker
- `kernel.bufs` is now a list of `CONST`/`DEFINE_VAR`/`DEFINE_GLOBAL`/`DEFINE_LOCAL` instead of `LOAD`/`STORE`, basically the source of the view. `bufs` are therefore the inputs to the graph
- If you create a tensor out of a bound variable, the view sits on top of the bind, this makes binding/unbinding simpler because you ideally don't want `BIND(VIEW(DEFINE_VAR), VIEW(CONST))`
- `DEFINE_VAR` is now also `VIEW(DEFINE_VAR(DEVICE))`

### Problems
- No more constant folding at the scheduling level. You will only get constant folding after lowering. Another problem of no constant folding is that `0/0` won't become `0*inf -> nan` but instead will still be rewritten by `x/x -> 1`. This is one of the tests in `test_ops.py` thats failing
- A lot of the tests are failing because they create CONST(VIEW(DEVICE)) by hand, there was discussion in the discord about removing them

### TODO
- Fix the tests
- I will try to get some degree of const folding back by just not having a view on the const if its not masked

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -44,8 +44,7 @@ def ast_const(dtype:DType, val:ConstType, shape:tuple[sint, ...]=(), st:Optional
   if st_src is None:
     st_src = (st.to_uop() if st is not None else ShapeTracker.from_shape(()).reshape((1,)*len(shape)).expand(shape).to_uop(),)
   st = unwrap(st_src[0].st)
-  if all(v.mask is None for v in st.views): return UOp.const(dtype, val).replace(src=(st.to_uop(),))
-  return graph_rewrite(UOp.const(dtype, val).view(st).valid(), view_left)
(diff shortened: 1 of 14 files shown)


Qazalin:
This diff is in the right direction,
Agree [const folding in the big Tensor graph isn't possible anymore](https://discord.com/channels/1068976834382925865/1230434680000741377/1385547803786350592), and we probably don't want it back at that level. See how testgrad deals with it.

PR #11024, opened 2025-06-29
clean up unused imports in examples and update CI linting

NinoRisteski:


geohot:
The problem with merging things like this is that we don't have tests or a linter. If you want this merged, you'd have to add that.

NinoRisteski:
> The problem with merging things like this is that we don't have tests or a linter. If you want this merged, you'd have to add that.

`examples` was excluded in ruff.toml. I included it https://github.com/tinygrad/tinygrad/pull/11024/commits/e31d0fb3738a6b0f08c713f2a28b283b80c6431c but only added to check for unused imports.

geohot:
Cool, might be worth PRs removing more of these exceptions

PR #11029, opened 2025-06-30
Make `KernelInfo.name` default `None`

eitanturok:
Make `KernelInfo.name=None` by default.

I wasn't sure exactly how to test this behavior we only initalize `KernelInfo` in two places:
* In `kernel.py`'s `fixup_ast()`, we initialize `KernelInfo` with arguments
* In `lower.py`'s `get_index()` without arguments
I didn't want to go so low level and call `get_index()` directly so I just tested that `fixup_ast` works fine even when we set `KernelInfo.name=None` by default.

Also, I noticed that by default all of the renderers set the name of the kernel to "test". So I included a default kernel name in the `Renderer` class.

### DIFF
--- a/tinygrad/renderer/__init__.py
+++ b/tinygrad/renderer/__init__.py
@@ -142,6 +142,7 @@ class ProgramSpec:
 class Renderer:
   device: str = ""
   suffix: str = ""
+  default_name: str = "test"
   # TODO: make this generic with a list of supported types
   supports_float4: bool = True
   has_local: bool = True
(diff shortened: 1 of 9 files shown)


tinygrad/renderer/__init__.py
```diff
@@ -142,6 +142,7 @@ def launch_dims(self, var_vals:dict[Variable, int]):
 class Renderer:
   device: str = ""
   suffix: str = ""
+  default_name: str = "test"
```
geohot, line 145: ?? is this changed?

Always make PRs minimal

PR #11034, opened 2025-06-30
remove simplify upcast [pr]

ignaciosica:
this is not necessary, upcasted axes cannot be transformed through any opt


geohot:
I would just leave this. It's more correct, even though nothing triggers it now

PR #11041, opened 2025-06-30
viz: test for no lockup on infinite loop

Qazalin:
This is a regression after non recursive graph_rewrite. Currently it locks up in the substitute.


Qazalin:
The right sidebar is a huge pain to re-render 200k rewrite items

```
VIZ=1 python test/unit/test_viz.py TestViz.test_inf_loop
```

no sidebar:
image
sidebar:
image

It can be much faster when re-rendering is removed.

PR #11047, opened 2025-07-01
extra: add tracepoints patching to usb

nimlgen:
a tool with some sample tracepoints to patch the controller


nimlgen:
it has nothing to do here. will put it in the docs repo

PR #11049, opened 2025-07-01
use kernel info

ignaciosica:
For now, kernelinfo is kept as a property of kernel. This was done this way on order to keep the pr scoped. Using the kernelinfo directly from the ast breaks the copy semantics of kernel, as the state of the ast will change over time (search, heuristics, tests...).

In the future, kernelinfo could independently be refactored to dynamically get and update the one form from sink without needing to change again the api, (`kernel.info` will point towards the one in the ast, instead of being itself a property of kernel). This hasn't been done in this pr due to the reason stated above.

### DIFF
--- a/tinygrad/opt/kernel.py
+++ b/tinygrad/opt/kernel.py
@@ -408,7 +408,7 @@ class Kernel:
       self.shift_to(axis, amt, insert_before=None)
       self.upcast()
     elif opt.op is OptOps.NOLOCALS:
-      check(self.opts.has_local and not self.dont_use_locals, "invalid NOLOCALS if target does not support local or already not using locals")
+      check(self.opts.has_local and not self.dont_use_locals, "NOLOCALS is meaningless if target does not support local or already not using locals")
       check(self.local_dims == 0 and self.group_for_reduces == 0, "can't have no locals with locals")
       self.update_info(dont_use_locals=True)
     elif opt.op is OptOps.SWAP:
--- a/tinygrad/opt/kernel.py
+++ b/tinygrad/opt/kernel.py
@@ -133,6 +133,9 @@ class Kernel:
   def upcasted(self) -> int: return self.info.upcasted
 
   @property
+  def dont_use_locals(self) -> bool: return self.info.dont_use_locals
+
+  @property
   def applied_opts(self) -> list[Opt]: return list(self.info.applied_opts)
 
   # there's eight chunks of the shape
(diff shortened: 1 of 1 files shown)


tinygrad/opt/kernel.py
```diff
@@ -128,6 +126,18 @@ def shape_len(self) -> int: return len(self.sts[0].shape)
   @property
   def global_dims(self) -> int: return self.first_reduce-self.local_dims
 
+  @property
+  def local_dims(self) -> int: return self.info.local_dims
```
ignaciosica, line 130: These properties were added in order to keep kernel api stable and the pr minimal. This could later be refactored and removed in a pr with no functionality change

tinygrad/opt/kernel.py
```diff
@@ -455,9 +466,8 @@ def fixup_ast(op:UOp) -> UOp:
         return ret.replace(src=(ret.src[0].replace(arg=st),)+ret.src[1:])
       if op.op is Ops.SINK:
         # NOTE: should group_for_reduces be added to the local_dims?
-        return ret.replace(arg = KernelInfo(ret.arg.name if ret.arg is not None else self.name if name_override is None else name_override,
-                                            self.global_dims if self.opts.has_local else 0, self.local_dims+self.group_for_reduces,
-                                            self.upcasted, self.dont_use_locals, tuple(self.applied_opts)))
+        return ret.replace(arg=replace(self.info, name=self.name if name_override is None else name_override,
```
Qazalin, line 469: err, why is ret.arg.name removed?

ignaciosica: its already in info if it was not None:
in kernel init
`self.info: KernelInfo = self.ast.arg if self.ast.arg is not None else KernelInfo()`

ignaciosica: Ah, I see what you mean, fixing

Qazalin, line 469: that KernelInfo construction was intentional,
now opts_to_apply survives even after applied_opts.

Qazalin: yea, not really your fault though we should write behavior tests before refactoring.
Currently only process_replay relies on this, so without [pr] nothing in core tinygrad would give feedback.

ignaciosica: addressing in #11055

ignaciosica: update: closed as saw your pr

PR #11052, opened 2025-07-02
fix process replay diff in PYTHON device [pr]

Qazalin:
The PYTHON backend pickles and encodes UOps, the encoded binary can't be directly diffed in process replay.
It first has to decode and unpickle the renderer output.

image

Currently output of `base64.b64encode(pickle.dumps(lops)).decode()` is not deterministic.


Qazalin:
was never right in the oldest of times:
image

```
~/code/tinygrad ((HEAD detached at 808029873)) > cat test.sh
#!/bin/bash -e
test/external/process_replay/reset.py
PYTHONPATH=. RUN_PROCESS_REPLAY=1 CAPTURE_PROCES_REPLAY=1 PYTHON=1 DEBUG=4 NOOPT=1 python -c "from tinygrad import Tensor; Tensor.arange(4).realize()"
PYTHONPATH=. test/external/process_replay/process_replay.py
```

PR #11054, opened 2025-07-02
add test_kernel_info

Qazalin:


ignaciosica:
why is ci so broken? Ahh, never mind, I think it was because it wasn't on master

PR #11055, opened 2025-07-02
fix kernelinfo init in fixup_ast

ignaciosica:
keep ast.arg.name if it was not none
clean up opts_to_apply


ignaciosica:
closing, I see its already being addressed in #11054

Qazalin:
can you add a test?
This is a logical issue, KernelInfo cannot define both applied_opts and opts_to_apply in one state.

Qazalin:
merging, it's blocking 11051.

Feel free to follow-up with a test.
also, I'd consider explicitly creating a fresh KernelInfo, makes things easier to reason about.
replace is a good choice for update_info, but the return value of get_optimized_ast creates a fundamentally different state.

PR #11061, opened 2025-07-02
remove const 0 clause in "UOp with size 0 is zero" [pr]

chenyuxyz:
don't see why const 0 needs to be special and no test failure nor kernel diff


Qazalin:
oh, it used to infinite loop. but I think we later changed UPat/match to return None in those cases.

PR #11062, opened 2025-07-02
[bounty] Merge ASSIGN and STORE

quortus:
TODO:
- Finish remained minor ASSIGN usages
- Clean the code


geohot:
Err, so the dtypes.void thing shouldn't be a check everywhere. STORE and ASSIGN actually have to be unified, not just disambiguated with dtypes.void

I'll lock the bounty for you, maybe submit sub-PRs moving some ASSIGNs to STORE. I'd start with the scheduler and Tensor.

geohot:
Replacing ASSIGN with STORE+AFTER in #13200

PR #11063, opened 2025-07-02
generate kernel dataset and upload artifact

chenyuxyz:
it generates a downloadable zip. and there's no way to make it not a zip https://github.com/actions/upload-artifact/issues/426


wozeparrot:
should be possible to use: https://github.com/lhotari/gh-actions-artifact-client to get it not to zip, file extension will still end with .zip tho, but the actual contents won't be

PR #11066, opened 2025-07-02
fix: correct args to MemoryManager

wozeparrot:


nimlgen:
rewritten to be shared with am

PR #11067, opened 2025-07-03
rename info to kernel_info [pr]

ignaciosica:
Renaming `info` to `kernel_info` so that if in the future kernel_info is added to the ProgramSpec api, it can be consistent and more clear.

Moved back to init KernelInfo in `get_optimized_ast` based on qaz feedback

Also added regression test


geohot:
I'm not going to merge more of these refactors without some functionality change or big line reduction, it's not clear this is more readable. We shouldn't be refactoring in expectation of something. If you think the first one made it worse feel free to revert.

PR #11068, opened 2025-07-03
start LLM app, tons of clean up required. target is 200 line ollama

geohot:


geohot:
If someone is interested, `SYM=1 python3 tinygrad/apps/llm.py` causes a graph rewrite loop

geohot:
Pretty happy with it, it's 139 lines

PR #11071, opened 2025-07-03
Custom hostname support for viz

quortus:
Useful if you are running dev env in VM or Docker.


Qazalin:
why not just forward the port?

quortus:
Because it takes more time to configure. Specifying custom ip address is a standard feature.

Qazalin:
Port forwarding is easy and works across cloud, docker and ssh.

PR #11105, opened 2025-07-05
Add `tinygrad.frontend` to pip package

adeebshihadeh:
Moving openpilot to use it as a normal pip package and this this:
```
  File "/home/batman/fivepilot/selfdrive/modeld/compile3.py", line 15, in <module>
    from tinygrad.frontend.onnx import OnnxRunner, onnx_load
    from tinygrad.frontend.onnx import OnnxRunner, onnx_load
ModuleNotFoundError: No module named 'tinygrad.frontend'
ModuleNotFoundError: No module named 'tinygrad.frontend'
```

https://github.com/commaai/openpilot/pull/35638


adeebshihadeh:
Ah I see this is because it imports from `extra`.

chenyuxyz:
yea we need onnx to be in tinygrad proper

PR #11106, opened 2025-07-05
remove invalid int div test

ttomsa:
The test divides by 0, crashes in llvm, cpu. A conditional move doesn't branch so the division by 0 is always performed even if the condition is false.


chenyuxyz:
i kinda want this to work so i added it, but fine to remove it since it does not run in CI anyway

Issue #10509, opened 2025-05-25
torch.linalg.svd not implemented yet


TiborGY:
Trying to do a Pytorch singular value decomposition fails when it needs to use aten::_linalg_svd.U, which is not supported yet.
https://docs.pytorch.org/docs/stable/generated/torch.linalg.svd.html

catswe:
Hi, I would like to take this on. Can this be assigned to me?

TiborGY:
> Hi, I would like to take this on. Can this be assigned to me?

Well, I dont have the power to assign you anything, but I am unlikely to have the time to implement it in the foreseeable future. So as far as I am concerned, you are most welcome to try getting this working.

LucianU:
There are several PRs. The most recent one: https://github.com/tinygrad/tinygrad/pull/10875

TiborGY:
The SVD implementation in #10875 does not seem to be hooked up to the torch backend. `torch.linalg.svd(torch.randn(3,3).tiny())` still complains about 'aten::_linalg_svd.U' not being implemented.

TiborGY:
Speed is not ideal but it runs now. Closing.

LucianU:
I'm curious, are you open to having other implementations besides this one: Randomized SVD, Golub-Reinsch, Divide-and-Conquer?

PR #11109, opened 2025-07-06
pipe linalg svd to torch

chenyuxyz:
#10509 
and found a bug in svd @kevvz fyi


kevvz:
I copy pasted your test and the shapes look good? Do you get (5,3) for U, (3,) for S, (3,3) for V? Oops I forgot to add the flag.

Looks like full_matrices (in aten_linalg) isn't propagating correctly, it is true in all cases even when setting it false
I'll figure out a fix

PR #11113, opened 2025-07-06
clean svd tests, set full_matrices false in torch backend

kevvz:
Torch's aten operators make no sense at all


chenyuxyz:
nice find, thanks!

PR #11116, opened 2025-07-07
Adjust vmin/vmax in case of over/underflow

S-Lykles:


S-Lykles:
Discussed in meeting 78

PR #11119, opened 2025-07-07
[DRAFT] not use disk with usb

nimlgen:


tinygrad/device.py
```diff
@@ -220,6 +220,7 @@ class Allocator(Generic[DeviceType]):
   def __init__(self, dev:DeviceType):
     self.dev: DeviceType = dev
     self.default_buffer_spec: BufferSpec = BufferSpec()
+    self.allow_copy_from_disc: bool = True
```
wozeparrot: `disk` with `k`?

PR #11121, opened 2025-07-07
[pr] Unify CONTIGUOUS and GBARRIER

quortus:
I've decided to try smaller and less ambitious refactor bounty first.

### DIFF
--- a/tinygrad/kernelize/kernelize.py
+++ b/tinygrad/kernelize/kernelize.py
@@ -136,9 +136,10 @@ def append_to_kernel(x:UOp):
   if (new_src:=tuple(dedup(new_srcs))) != x.src: return x.replace(src=new_src, arg=Kernel(x.arg.ast, tuple(dedup(metadata))))
 
 create_kernels = PatternMatcher([
-  # always give assign/gbarrier a kernel
+  # always give assign/contiguous a kernel
   (UPat.assign(UPat.var("b"), UPat(GroupOp.All-{Ops.KERNEL}), name="x"), create_kernel),
-  (UPat(Ops.GBARRIER, src=(UPat.var("x"),)), create_kernel),
+  (UPat(Ops.CONTIGUOUS, src=(UPat({Ops.CONST, Ops.VIEW, Ops.RESHAPE}|DONT_PLACE_IN_KERNEL),), name="x"), create_kernel),
+  (UPat(Ops.CONTIGUOUS, src=(UPat(name="x"),)), create_kernel),
   # walk back the local graph until we reach a realized source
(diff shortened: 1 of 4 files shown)


geohot:
How have you tested that nothing changed? I see a lot of diffs in process replay.

Qazalin:
Process replay will have large diffs if the Ops Enum changes, due to how unpickling works.
You don't need to remove the Op for this diff though.

quortus:
@geohot I thought process replay validates all steps including intermediate so I didn't try to use it. Instead I compared kernel outputs of test_uops.py and test_ops.py. And that was not enough.

Qazalin:
[pr] simply validates a function's return value given the same arguments,
and looks like it's passing here.

quortus:
BTW It would be way more convenient if [pr] did not stop on the first error. I removed it during development because it helps to see all (up to a limit) failed tests.

tinygrad/kernelize/kernelize.py
```diff
@@ -136,9 +136,10 @@ def append_to_kernel(x:UOp):
   if (new_src:=tuple(dedup(new_srcs))) != x.src: return x.replace(src=new_src, arg=Kernel(x.arg.ast, tuple(dedup(metadata))))
 
 create_kernels = PatternMatcher([
-  # always give assign/gbarrier a kernel
+  # always give assign/contiguous a kernel
   (UPat.assign(UPat.var("b"), UPat(GroupOp.All-{Ops.KERNEL}), name="x"), create_kernel),
-  (UPat(Ops.GBARRIER, src=(UPat.var("x"),)), create_kernel),
+  (UPat(Ops.CONTIGUOUS, src=(UPat({Ops.CONST, Ops.VIEW, Ops.RESHAPE}|DONT_PLACE_IN_KERNEL),), name="x"), create_kernel),
+  (UPat(Ops.CONTIGUOUS, src=(UPat(name="x"),)), create_kernel),
```
Qazalin: you don't need these two lines, it's just `(UPat(Ops.CONTIGUOUS, name="x"), create_kernel),`

quortus: They are different. First rule puts CONTIGUOUS in kernel, second - don't.

Qazalin: Just remove contiguous in COPY/BUFFER_VIEW? or better, change the UPat that removes SINK from them.

something like CONTIGUOUS(REDUCE_AXIS) can and probably should keep existing.

quortus: I've changed UPat.

geohot:
Congrats, the first refactor bounty is yours! E-mail george@tinygrad.org to claim

PR #11125, opened 2025-07-08
viz: profile RewriteSteps in TINY device

Qazalin:
Tracing down to graph_rewrite/graph_rewrite_map calls
New beautiful_mnist:
image

- Nested graph_rewrites work naturally:
image

- Linking works when the graph_rewrites are named. Most of them are.
- Colors are divided by the top-level call (schedule is green, get_program is blue)


Qazalin:
TRACK_MATCH_STATS=3 traces down to match calls, it's slower to render though. Profiling it, I'd prefer it's the default.
image

Python overhead is minimal, UI is (very) slow.

geohot:
Woah, down to match calls is crazy, maybe too much, but if you can make it fast why not...be careful with the overhead you are adding here though.

I'd be fine with just graph_rewrites (and sub graph_rewrites)

Qazalin:
Yeah just profiling the graph_rewrites works, it stacks nicely too:
image
The algorithm for stacking just looks at timestamps.

PR #11128, opened 2025-07-08
Remove Ops.GBARRIER

quortus:
Depends on #11121


geohot:
Did this here with the color too. https://github.com/tinygrad/tinygrad/commit/289a411f5f934c61c5fc61aedc2df16b634ab636

PR #11136, opened 2025-07-08
Add int ** int support to pow() with error handling

Mohammed0101-lgtm:
# 🚀 Add Support for Integer Exponentiation (`int ** int`) in `Tensor.pow()`

## 🔧 What this PR does:
- Implements element-wise integer exponentiation (`int ** int`) in the `Tensor.pow()` method.
- Adds safety checks for:
  - Zero raised to a non-positive power (`0 ** 0`, `0 ** -n`) — raises `ZeroDivisionError`.
  - Negative integer exponents with integer bases — raises `ValueError`.
- Preserves float behavior when either base or exponent is a float.
- Keeps rounding logic consistent with existing behavior for float → int casting when required.

## 🧪 Why it's useful:
- Allows `Tensor` objects with integer data types to correctly perform power operations with integer exponents.
- Prevents unexpected or undefined behavior by explicitly catching invalid operations.
- Enhances `tinygrad`'s compatibility with typical NumPy-like tensor operations.

## ✅ Example:
```python
Tensor([2, 3, 4]).pow(2)  # => [4, 9, 16]

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -3614,29 +3614,38 @@ class Tensor(MathTrait):
     assert dtypes.is_unsigned(self.dtype) and isinstance(x, int) and x >= 0 and not reverse, f"not supported {self.dtype=} {x=}"
     return self.idiv(2 ** x, reverse)
 
-  def pow(self, x:Tensor|ConstType, reverse=False) -> Tensor:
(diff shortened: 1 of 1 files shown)


chenyuxyz:
don't use ai

mohamed-labbit:
> don't use ai

I did for the description, or is that not allowed?

PR #11138, opened 2025-07-08
Update tensor.py

Mohammed0101-lgtm:
added int pow that was missing


geohot:
No new tests, failing tests, looks like it's from AI or GItHub edit

PR #11144, opened 2025-07-09
viz: stack an event for the entire batch

Qazalin:


Qazalin:
I think we should have a separate "NV Graph" stream that handles the MultiGraphRunner cases though

PR #11145, opened 2025-07-09
record GraphEvents in metal graph

Qazalin:
New Metal beautiful_mnist:
image

vs master:
image

Currently in master, the MetalGraph command buffer is recorded like normal Programs, that approach loses the kernels inside a batch.

This diff records trace events directly inside MetalGraph, which gives a complete picture of the batch.
Tracing flow is similar to HCQGraph.

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -199,7 +199,8 @@ class MetalAllocator(LRUAllocator[MetalDevice]):
     ret = msg("newBufferWithLength:options:", objc_id)(self.dev.sysdevice, ctypes.c_ulong(size), MTLResourceOptions.MTLResourceStorageModeShared)
     if ret.value is None: raise MemoryError(f"Metal OOM while allocating {size=}")
     return MetalBuffer(ret, size)
-  def _free(self, opaque:MetalBuffer, options): msg("release")(opaque.buf)
+  def _free(self, opaque:MetalBuffer, options):
+    if msg is not None and libobjc is not None: msg("release")(opaque.buf)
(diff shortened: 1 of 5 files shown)


geohot:
Why are you getting them from the command buffer? You constructed that command buffer, you know what went into it.

tinygrad/runtime/graph/metal.py
```diff
@@ -100,3 +102,13 @@ def __call__(self, input_rawbuffers: list[Buffer], var_vals: dict[Variable, int]
...
+
+  def collect_timestamps(self):
+    # create a graph event and evenly space each program
+    st, en = decimal.Decimal(cmdbuf_st_time(self.command_buffer)) * 1000000, decimal.Decimal(cmdbuf_en_time(self.command_buffer)) * 1000000
+    ents = [ProfileGraphEntry(self.device, cast(CompiledRunner, ji.prg)._prg.name, i, i+1, is_copy=False) for i,ji in enumerate(self.jit_cache)]
+    step = (en-st)/len(ents)
+    self.dev.profile_events += [ProfileGraphEvent(ents, [], [st+step*i for i in range(len(ents)+1)])]
+
+  def __del__(self):
+    if PROFILE: self.collect_timestamps()
```
nimlgen: wait here? I think we can destroy graphs once cmdbuf is submitted, no?

Qazalin: hmm, I don't think I'm touching any lifetimes here? Will test some more.
Unless you know a failing case already

edit: oh and yeah, python gc is non deterministic, so it might even AttributeError

PR #11158, opened 2025-07-10
metal: fix incorrect _free on interpreter exit

pkova:
The same idea as in #8950, this gets rid of the annoying `TypeError: 'NoneType' object is not callable` errors when running anything metal. This happens in CI every time: https://github.com/tinygrad/tinygrad/actions/runs/16180307775/job/45675203905#step:6:63

The original PR asks why `sys.is_finalizing` doesn't work here. Looks like the reason is that the python interpreter shutdown lifecycle is a complete shitshow with cleanup steps happening in all kinds of wonky orders. For example, in our case here if we were to try to call `sys.is_finalizing` to determine whether to release we would discover that `sys` is already `None`! Another way to write this fix is `if sys is not None: msg("release")(opaque.buf)`, but that's even more gross than this is.


nimlgen:
Ugh, maybe we should add `@fail_if_shutting_down` and catch attribute errors if we shutting down? Can we come up with a global solution to be reused across the whole project?

PR #11159, opened 2025-07-10
Replace ASSIGN with STORE in kernelize and tensor

quortus:
Part of #11062.
Store dtype will be unified in a separate MR.


test/external/process_replay/process_replay.py
```diff
@@ -42,9 +43,14 @@ class ProcessReplayWarning(Warning): pass
 
 # *** replay the function and convert return values to string
 
+pm_store_to_assign = PatternMatcher([
```
quortus: This hack fixes process replay execution divergence from what happens on master branch. 
Process replay tries to replay new version of `Tensor.assign` on master branch (with STORE instead of ASSIGN) and this, obviously, fails.
This hack should be reverted by the following PR #11833.

Qazalin: If your diff does not pass process replay you can remove `[pr]`, do not remove the test!

quortus: But why? This will replace limited process replay checks with zero pr check. Now it still checks that other rules behaves the same.

quortus: I did find quite a few bugs with this setup.

Qazalin: We can still see the diff without `[pr]`, and prefer that over removing the test completely.

quortus: Done. But these diffs are quite unreadable.

geohot:
Replacing ASSIGN with STORE+AFTER in #13200

PR #11173, opened 2025-07-11
torch.compile TinyJIT for beautiful_mnist_torch

mpai17:
- Add torch.compile backend="tiny" support with TINY_BACKEND=1
- Register TinyJIT backend in torch frontend for torch.compile
- Use CPU device for data/model when using tiny backend for compute

Tiny backend fails due to OpaqueTensorImpl storage access errors in PyTorch's compilation infrastructure. Basic torch.compile functionality works, but the full MNIST example encounters this limitation.


chenyuxyz:
stale, test failed

PR #11177, opened 2025-07-11
Remote scheduler changes

uuuvn:
For multihost bert step to be 1300 ms instead of 13 s.

Toposort change is to have a bunch of big graphs instead of a couple thousand graphs with 1-2 kernels each
Multi change is to do topology-aware allreduce, otherwise allreduces would be really slow

The group id approach is the simplest one, devices with multiple idxes and consts instead of colon is a big refactor that i don't have neither time nor motivation to do

### DIFF
--- a/test/test_remote.py
+++ b/test/test_remote.py
@@ -32,7 +32,6 @@ class TestRemoteMultiHost(unittest.TestCase):
     # Verify that everything is in one big cross-host graph
     assert len(do.captured._jit_cache) == 1 and isinstance(do.captured._jit_cache[0].prg, RemoteGraph), repr(do.captured)
 
-  @unittest.expectedFailure # multihost-aware schedule is in separate pr
   @Context(JIT_BATCH_SIZE=2**32)
   def test_multihost_aware_schedule(self):
     @TinyJit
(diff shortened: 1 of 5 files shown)


wozeparrot:
tests?

uuuvn:
> tests?

It's a bit hard to have a generic tests for this without #9746

I rebased things around such that this pr is after the ib one and ib one no longer requires this (but is very slow without those changes).

The test is `test_multihost_aware_schedule`, it is added in 9746 but is `@unittest.expectedFailure`, this pr fixes scheduling and just removes expected failure. This looks good?

uuuvn:
Actually, there is a better way. I factored out cross-host graph changes into #11229. Now it's like this:
image

Now the only dependency of this pr is #11229 which adds the expected fail test that this pr removes

geohot:
@wozeparrot what's the delay with reviewing this?

wozeparrot:
the multi change is fine to me. just if you want to look over the scheduler changes as well.

wozeparrot:
merge conflict

uuuvn:
> merge conflict

resolved, was conflicting with `Optional[]` => `|None` refactor someone did in device.py

PR #11194, opened 2025-07-12
fix device arg to Tensor.randn

nimlgen:


nimlgen:
@chenyuxyz maybe you know a better and more general way to test this?

test/test_randomness.py
```diff
@@ -251,6 +251,12 @@ def test_randn(self):
     self.assertTrue(normal_test(Tensor.randn))
     self.assertTrue(equal_distribution(Tensor.randn, torch.randn, lambda x: np.random.randn(*x)))
 
+  def test_randn_device(self):
+    if Device.DEFAULT == "CPU": self.skipTest("no diff dev to test this")
+    try: _, _ = Device[Device.DEFAULT], Device["CPU"]
+    except Exception: self.skipTest("no diff dev to test this")
```
chenyuxyz: when does this trigger?

chenyuxyz:
we can add something in unit, `test_tensor_device` or similar, and put all Tensor staticmethod with device specified there. hard coded between two device should suffice

test/test_randomness.py
```diff
@@ -251,6 +251,9 @@ def test_randn(self):
     self.assertTrue(normal_test(Tensor.randn))
     self.assertTrue(equal_distribution(Tensor.randn, torch.randn, lambda x: np.random.randn(*x)))
 
+  def test_randn_device(self):
+    assert (rdev:=Tensor.randn(3,3,device="CPU").device) == "CPU", f"{rdev} != CPU"
```
chenyuxyz: `self.assertEqual`

PR #11199, opened 2025-07-12
final mergeable onnx

geohotstan:
the diff looks absolute 🍌s now but the main thing is that 1 file [onnx.py](https://github.com/geohotstan/tinygrad/blob/mergable_onnx/extra/onnx.py)
So mainly give those abstractions a look.

If everything in this PR is merged, then Onnx is ready.
Posted this PR as a merge tracker.

this branch contains:
some already open PRs
- #10789 
- #11000 
added tests in `external_test_onnx_runner.py` for external data (tensor) loading.
big merge of onnx_parser and onnx.

Gonna link everything to this.


extra/onnx.py
```diff
@@ -544,6 +550,20 @@ def ___wrapper(R:Tensor, T:int, *inputs:Tensor, **kwargs):
...
   # ***** Property/Graph Ops *****
+  def If(condition:Tensor, else_branch, then_branch, intermediate_tensors:dict[str, Tensor]):
+    else_graph, then_graph = SubGraphOnnxRunner(else_branch), SubGraphOnnxRunner(then_branch)
+    else_graph.graph_values = intermediate_tensors
+    then_graph.graph_values = intermediate_tensors
+    else_out = else_graph({k:intermediate_tensors[k] for k in else_graph.graph_inputs.keys()})
+    then_out = then_graph({k:intermediate_tensors[k] for k in then_graph.graph_inputs.keys()})
+    assert len(else_out) == len(then_out), f"else_out and then_out must have the same number of outputs: {len(else_out)} != {len(then_out)}"
+    # can use where op when output shape is the same
+    if all(t.shape == e.shape for t,e in zip(then_out.values(), else_out.values())):
```
chenyuxyz, line 561: if the shapes are different this can just broadcast?

geohotstan: ORT expects outputs from If Op to be not broadcasted, so unfortunately no. :c

test/external/external_test_onnx_backend.py
```diff
@@ -165,12 +166,16 @@ def supports_device(cls, device: str) -> bool:
...
-backend_test.exclude('test_resize_upsample_sizes_cubic_*') # unsure how to implement cubic
 backend_test.exclude('test_ai_onnx_ml_tree_ensemble_*') # https://github.com/onnx/onnx/blob/main/onnx/reference/ops/aionnxml/op_tree_ensemble.py#L121
 
+# TODO: cubic not fully supported
+backend_test.exclude('test_resize_downsample_scales_cubic_antialias_cpu')
+backend_test.exclude('test_resize_downsample_sizes_cubic_antialias_cpu')
+backend_test.exclude('test_resize_upsample_scales_cubic_A_n0p5_exclude_outside_cpu')
+backend_test.exclude('test_resize_upsample_scales_cubic_asymmetric_cpu')
+backend_test.exclude('test_resize_upsample_scales_cubic_cpu')
+backend_test.exclude('test_resize_upsample_sizes_cubic_cpu')
```
chenyuxyz, line 177: i also got a similar implementation that somehow failed at these two. i understand that antialias and exclude outside are not implemented, but not sure why upsamples are incorrect while downsamples are fine. is it because of out of bound issue?

geohotstan: I'm not sure.
This is the numpy impl I used as context to vibe the above implementation:  https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_resize.py

It also may be that the test is just wrong. I can cross validate with ORT to see.
Gonna add a PR to include this into the current master onnx.

---
update: 
> I can cross validate with ORT to see.

ORT's output and the reference implementation (op_resize.py) output both have the same value as the expected value for the test. So this is not a test suite issue.

---
update 2:
aaaaaa it seems they use a different clipping method
https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_resize.py#L231-L237

chenyuxyz:
closing as onnx is merged, bounty is yours!

PR #11201, opened 2025-07-12
fix silent mypy failure in ci

uuuvn:
Example: https://github.com/tinygrad/tinygrad/actions/runs/16215577171/job/45784110543?pr=11177#step:7:20

Caused by footguny exception in how `set -e` works:

```bash
python -m mypy --strict-equality --lineprecision-report . && cat lineprecision.txt
```

Will fail (and have non-zero exit code if run in interactive mode) but because there is `&&` it won't count as script-terminating failure in a script with `set -e` and instead as a test (similar to how fail of a command in if condition won't count as a script-terminating failure despite having non-zero exit code)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -332,7 +332,8 @@ jobs:
       run: python -m pylint tinygrad/
     - name: Run mypy
       run: |
-        python -m mypy --strict-equality --lineprecision-report . && cat lineprecision.txt
+        python -m mypy --strict-equality --lineprecision-report .
+        cat lineprecision.txt
         python -m mypy --strict-equality extra/onnx_parser.py
 
   unittest:


uuuvn:
This is really unintuitive, like i get why `||` doesn't cause a failure but any reasonable person would expect `false && true` to fail. it even behaves reasonably in interactive mode, but breaks in script:
image

chenyuxyz:
hmm weird. i think there was a config somewhere to detect early failures in pipe. anyway the fix looks simple enough

uuuvn:
> hmm weird. i think there was a config somewhere to detect early failures in pipe. anyway the fix looks simple enough

assuming you mean pipefail it's an entirely different thing:
image
still fails silently

PR #11211, opened 2025-07-13
update float4 condition in hcopt

chenyuxyz:
don't need all upcast candidates to be upcast-able, only check the actual one

looks like it's mostly no-op


chenyuxyz:
no process replay diff for image and openpilot

PR #11216, opened 2025-07-13
add Ops.CMPGT

ttomsa:
this is relevant for #11611. On x86 vector int `Ops.CMPLT` only exists with avx512. On arm64 no vector dtypes have a `Ops.CMPLT`. We use `Ops.CMPGT` instead with the pattern x < y => y > x


ttomsa:
going to keep it as Ops.CMPLT and emit cmpgt by swapping the operands

PR #11219, opened 2025-07-13
simpler grouptop in hcopt

chenyuxyz:
keep the only perf relevant conditions and the rest is handled by try except


chenyuxyz:
seemed to made openpilot slightly faster...

PR #11224, opened 2025-07-14
remove some first_reduce in hcopt [pr]

chenyuxyz:
`:first_reduce` is `k.global_dims+k.local_dims` when referring to upcast-able axes. this should make moving UPCAST to before REDUCE easier


chenyuxyz:
nvm have a much better idea

PR #11229, opened 2025-07-14
Remote cross-host graph

uuuvn:
Factored out some changes from #9746 for ease of review


tinygrad/runtime/ops_remote.py
```diff
@@ -189,9 +194,9 @@ async def handle(self, method:str, path:str, body:bytes) -> tuple[http.HTTPStatu
             if c.session == c.event_session:
               session.events[c.event].set()
             else:
-              dev.synchronize() # wait for device to finish executing previous stuff
+              [Device[d].synchronize() for d in Device._opened_devices] # wait for device*s* to finish executing previous stuff
```
wozeparrot: write out the for loop here?

uuuvn: assuming you mean `for d in Device._opened_devices: Device[d].synchronize()`, fixed

PR #11237, opened 2025-07-14
Update README and other changes

AlexKissiJr:
# Add BEAM Search Tests to Pre-commit Hooks

## Description
This PR adds BEAM search optimization tests to the pre-commit hooks to ensure that changes don't break the BEAM search functionality. BEAM search is a critical optimization technique used throughout the codebase (as seen in CI workflows and MLPerf benchmarks), but wasn't being tested in pre-commit hooks.

## Changes
- Added a new pre-commit hook that runs the BEAM search tests with `IGNORE_BEAM_CACHE=1`
- Updated README to document the new pre-commit hook
- Positioned the hook before the general tests to catch BEAM-specific issues early

## Testing
- Verified that the pre-commit hook runs successfully
- Confirmed that the tests catch issues with BEAM search optimization

## Why this matters
BEAM search optimization is used in multiple benchmarks and is critical for performance. By adding this to pre-commit hooks, we ensure that developers catch any regressions before they're committed, improving code quality and preventing performance degradation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,12 @@ repos:
         language: system
         always_run: true
         pass_filenames: false
+      - id: beamtests
+        name: beam search tests
+        entry: env PYTHONPATH="." IGNORE_BEAM_CACHE=1 python3 -m pytest extra/optimization/test_beam_search.py
+        language: system
+        always_run: true
+        pass_filenames: false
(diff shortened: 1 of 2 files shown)


chenyuxyz:
please read https://github.com/tinygrad/tinygrad/?tab=readme-ov-file#contributing

PR #11240, opened 2025-07-14
Switch the meaning of the REDUCE_AXIS/REDUCE UOp to be keepdims=False instead of True while preserving all functionality

llvm-x86:
Switch the meaning of the REDUCE_AXIS/REDUCE UOp to be keepdims=False instead of True while preserving all functionality`


llvm-x86:
to achieve bounty. Paypal email: ornuq@protonmail.com

Please let me know any thing I missed so I may fix it. Thank you.

chenyuxyz:
not the right fix

llvm-x86:
@chenyuxyz could you point me in the right direction please?

llvm-x86:
@chenyuxyz I'm really lost. Could you give me an entry-point please?

PR #11249, opened 2025-07-15
only construct Ops.VALID in kernelize [pr]

Qazalin:
lowerer handles VIEW(CONST) fine now.


Qazalin:
we can just remove Ops.VALID once kernel.py doesn't rely on the full_shape stuff.

PR #11252, opened 2025-07-15
axistypes for ranges in gpudims

kevvz:


chenyuxyz:
cool thanks, i also added more refactors in #11255

PR #11253, opened 2025-07-15
add METAL tracer

Qazalin:
Running with `VIZ=1 METAL=1` saves trace file in /tmp/metal.trace.
This can be viewed using `open /tmp/metal.trace` (requires XCode.app install, the final version should not)

- [x] Low overhead GPU counter tracing
     - note: less than 1% runtime overhead. However, startup/exit is slow due to subprocess calls, I think there's a path to removing subprocess usage.
- [x] Generic UI to render bar/area charts
- [x] XML parsing O(n)
- [x] Align Metal Counter timestamps with Program timestamps

speedup work:
Goal is to render beautiful_mnist fast, currently stopping xctrace recording (lightest weight metrics possible) takes 5 seconds.

- [x] Stream counters to UI, we already have this infra for graph_rewrites with 100+ steps
- [x] Aggregate metrics instead of sending raw counters to client
- [x] Add UI to display aggregate metrics
- [x] Put behind a reasonable flag.

Apple specific work (note: pausing work on reversing XCode to work on AMD's stuff)
- [ ] Fast profiling startup, trace without XCode
- [ ] Fast profiling exit/parsing (like [InstrumentsParser.kt](https://github.com/benjaminRomano/instruments-to-gecko/blob/main/src/main/java/com/bromano/instrumentsgecko/InstrumentsParser.kt), [google/instrumentsToPprof ](https://github.com/google/instrumentsToPprof)

Default view aggregates counters into ALU, DRAM, SRAM, Bandwidth and Occupancy.
This view is hardware agnostic:
image

With per-metric breakdown:
image

### DIFF
--- a/tinygrad/runtime/ops_metal.py
+++ b/tinygrad/runtime/ops_metal.py
@@ -72,7 +72,7 @@ class MetalDevice(Compiled):
     self.timeline_value = 0
 
     Compiled.profile_events += [ProfileDeviceEvent(device)]
-    if PROFILE.value >= 2 or MetalDevice.xctrace_proc is None: self.start_xctrace()
+    if PROFILE.value >= 2 and MetalDevice.xctrace_proc is None: self.start_xctrace()
 
     from tinygrad.runtime.graph.metal import MetalGraph
     # NOTE: GitHub CI macOS runners use paravirtualized metal which is broken with graph.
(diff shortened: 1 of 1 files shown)


Qazalin:
47 seconds parsing uop_gemm test...

Qazalin:
something finally
image

tinygrad/runtime/ops_metal.py
```diff
@@ -88,6 +89,25 @@ def synchronize(self):
         Compiled.profile_events += [ProfileRangeEvent(self.device, lb, st, en, is_copy=lb.startswith("COPY"))]
     self.mtl_buffers_in_flight.clear()
 
+  xctrace_proc = None
+  @classmethod
+  def start_xctrace(cls):
+    if PROFILE.value < 2 or cls.xctrace_proc is not None: return
```
geohot: Move to line 75

tinygrad/runtime/ops_metal.py
```diff
@@ -88,6 +89,25 @@ def synchronize(self):
...
+    # fetch xctrace config
+    cfg = "https://gist.githubusercontent.com/qazalin/96680d79e12ab18f19403ac696ced8d2/raw/12268d24c738e4d1272b29a2d26ee6e5b831a6bd/GPUCounter.xml"
+    cfg_loc = os.path.expanduser("~/Library/Application Support/Instruments/Templates/GPUCounter.tracetemplate")
+    os.system(f"plutil -convert xml1 -o '{cfg_loc}' {fetch(cfg)}")
+    # start recording hardware counters
+    os.system("rm -rf "+(output:="/tmp/metal.trace"))
```
geohot: Don't do this! Use pathlib

tinygrad/runtime/ops_metal.py
```diff
@@ -88,6 +89,25 @@ def synchronize(self):
         Compiled.profile_events += [ProfileRangeEvent(self.device, lb, st, en, is_copy=lb.startswith("COPY"))]
     self.mtl_buffers_in_flight.clear()
 
+  xctrace_proc = None
+  @classmethod
+  def start_xctrace(cls):
+    if PROFILE.value < 2 or cls.xctrace_proc is not None: return
+    # fetch xctrace config
+    cfg = "https://gist.githubusercontent.com/qazalin/96680d79e12ab18f19403ac696ced8d2/raw/12268d24c738e4d1272b29a2d26ee6e5b831a6bd/GPUCounter.xml"
+    cfg_loc = os.path.expanduser("~/Library/Application Support/Instruments/Templates/GPUCounter.tracetemplate")
```
geohot: Create dir

Qazalin:
The generic UI for showing hw resource usage is merged with LLVM MCA.
This diff has a lot of Apple specific things and requires XCode.app installation (since it's only on App Store, requires Apple ID login!).

If this isn't locked down we should merge something that reads the internal GPU counters.

PR #11266, opened 2025-07-16
Bump onnx to 1.18.0

geohotstan:
Needs serious clean ups still.

Also #11199 needs to support parsing domain for each node. I initially removed that from the parser because parsing unneeded string is really wasteful, but that has to be added back.

the workaround in _dispatch_op is just to get domain management working.

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -739,21 +739,17 @@ def get_onnx_ops():
     if attn_mask is not None:
       mask_to_add = attn_mask.where(0, -float("inf")) if attn_mask.dtype == dtypes.bool else attn_mask
       scores = scores + mask_to_add
-
     if qk_matmul_output_mode == 1: qk_matmul_return_val = scores
 
-    if softcap > 0.0:
-      scores = (scores / softcap).tanh() * softcap
-      if qk_matmul_output_mode == 2: qk_matmul_return_val = scores
+    if softcap > 0.0: scores = (scores / softcap).tanh() * softcap
+    if qk_matmul_output_mode == 2: qk_matmul_return_val = scores
 
     if softmax_precision: scores = scores.cast({1: dtypes.float32, 10: dtypes.float16, 16: dtypes.bfloat16}[softmax_precision])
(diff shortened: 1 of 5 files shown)


extra/onnx.py
```diff
@@ -488,7 +494,9 @@ def CumSum(X:Tensor, axis:int|list, exclusive:int=0, reverse:int=0):
                         .shrink(tuple((0,X.shape[axis]) if i == axis else None for i in range(X.ndim)))
     return X.cumsum(axis).flip(axis) if reverse else X.cumsum(axis)
 
-  def Trilu(x:Tensor, k:int=0, upper:int=1): return x.triu(k) if upper else x.tril(k)
+  def Trilu(x:Tensor, k:int=0, upper:int=1):
+    k = k[0] if isinstance(k, list) else k
```
chenyuxyz: why is it sometimes `(list, tuple)` and sometimes `list` only? consider adding a helper for this

geohotstan: This is a pure hack I have to change.
I just changed this to get test passing. (attention tests has some multi-node test with multiple ops)
I think it might be to_python_const related.

---
edit:
ok, I added helper for this.

It seems that the onnx test suite just likes to differ from their docs
The node protobuf from official onnx.load gives a dim value of [1] while their docs specify that this value is a 0-dim Tensor. 🙃 

multiple tests also have this.....

geohotstan:
I implemented a minimal opset manager.

Some notes:
- These Attention tests take pretty long to run
- the new implemented ops are currently pretty much just a translation from their [reference implementation](https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_attention.py)
- for the ORT workaround in tests, it's ugly but it's easily removable.

2 more TODOs that can be done later? Or I'll keep going and finish these two TODOs.
- clean up newly added ops (pipe it?). 
- add `def get_onnx_ops` docs to detail how to register ops properly

chenyuxyz:
looks good, thanks!

PR #11276, opened 2025-07-18
[pr] Stable tensor order in _find_all_tensors_for_uops

quortus:
Right now result of `_apply_map_to_tensors` depends not only on its inputs but also on global set `all_tensors`. Return value of `_find_all_tensors_for_uops` can change due to creation/destruction of unrelated tensor. Changing set to dict keeps iteration order the same and helps avoid annoying process replay errors.


Qazalin:
Yeah I think non determinism in set order isn't a desirable behavior.

PR #11301, opened 2025-07-21
viz: defer loading traces

Qazalin:
- Profiler already gets cached on the client side
- Startup time is O(1) -> Just cost of spinning up a server, everything else gets offloaded to the client.
- Explicitly raise FileNotFoundError if trace file does not exist


Qazalin:
I don't think we want this... It's still pushing latency to the client.
Will revisit the idea later, likely with a design that allows for streaming traces - It's the same problem as real-time viz.

PR #11310, opened 2025-07-21
Move remote tests out of onnx

uuuvn:
Should fix https://github.com/tinygrad/tinygrad/actions/runs/16423765198/job/46408807596


Qazalin:
I disabled it for remote, is it easy to make it not open devices?
https://github.com/tinygrad/tinygrad/pull/11320

uuuvn:
> I disabled it for remote, is it easy to make it not open devices? #11320

Not sure, leaning towards no. The easiest way to do this might be to replace colon in device name by const in Ops.DEVICE and then have like a multi-dimensional topology with multiple consts that George proposed in one of the meetings

Having `ScheduleContext` with info needed for scheduling like a dict from device string to it's properties will work too but it seems just annoying and hacky enough that it might make sense to do this properly via the first thing

uuuvn:
(this is no longer a process replay workaround but i still think it's worth merging - remote shouldn't be in onnx runner)

PR #11316, opened 2025-07-21
Keepdims false

kevvz:
- I had to reverse the order of unrolls on shape since it looked like it was messing up something in the lowerer for multireduces.
- Reduce push add ones (and contraction w/reduce) is unnecessary now since 1s are treated just like any other number. It is now handled by swizzle
- Simplify ones is still needed for apply_opt


kevvz:
VIZ=1 python3 test/test_ops.py TestOps.test_matvec
Image 7-22-25 at 1 30 PM

chenyuxyz:
for bounty it needs to keep the behavior exactly the same

kevvz:
So no changes to any tests at all? I ask this since I thought the point of this was to remove as many ones as needed. Pushing the ones outside of REDUCE_AXIS isn't the same as removing them altogether

chenyuxyz:
stale

PR #11351, opened 2025-07-23
support `DEV=` to specify device

chenyuxyz:


chenyuxyz:
fix #6689

`DEV=GPU python -c "from tinygrad import Device; print(Device.DEFAULT)"` works now

PR #11355, opened 2025-07-24
fix imports in assembly and update CI lint

NinoRisteski:


Qazalin:
Same comment as: https://github.com/tinygrad/tinygrad/pull/11024#issuecomment-3016818913

NinoRisteski:
> Same comment as: [#11024 (comment)](https://github.com/tinygrad/tinygrad/pull/11024#issuecomment-3016818913)

updated with https://github.com/tinygrad/tinygrad/pull/11355/commits/25de5174d0e8555beb02cb9f4f9f284d4a5e0935

.github/workflows/test.yml
```diff
@@ -322,14 +322,21 @@ jobs:
...
-    - name: Lint tinygrad with pylint
-      run: python -m pylint tinygrad/
+        python3 -m ruff check . --select F822
+        python3 -m ruff check extra/onnx.py extra/onnx_parser.py --select F822
+        python3 -m ruff check examples/mlperf/ --select F822 --ignore E501
+        python3 -m ruff check extra/ --select F822
+    - name: Lint tinygrad with pylint (E0401 only)
+      run: |
+        python -m pylint tinygrad/ > pylint_tinygrad.log || true
+        if grep -q 'E0401' pylint_tinygrad.log; then cat pylint_tinygrad.log; exit 1; fi
```
chenyuxyz, line 335: this is complexity creep

like is assembly in extra work at all? it's not clear by looking at the pr

if it still does not work, there's no point to fix the import only; if it works with this fix, show so in the pr

NinoRisteski: you are right. will try to simplify

NinoRisteski: couldnt find a simple way for this. i think its not worth it. thx for the feedback!

PR #11364, opened 2025-07-25
don't use const in keccak rnd_const_masks

chenyuxyz:
setting it as a const generates 23 more unique kernels for the loop. 30% faster scheduling and somehow slightly faster kernels on mac


chenyuxyz:
inconclusive

PR #11366, opened 2025-07-25
replace early_exec with subprocess in assembly_rdna

NinoRisteski:
removed the unwanted and broken extra helper and replaced with subprocess


geohot:
huh, why? is this code used? tested?

NinoRisteski:
Sry, guess I misunderstood the comment. Just thought about a quick fix.

PR #11367, opened 2025-07-25
mockgpu: threaded mockgpu

nimlgen:
put executor into a separate thread. should work for severall devices now (e.g CPU and NV)


nimlgen:
amd ci is 1.3x slower, will think of optimizations or maybe can put it behind a flag (do not really want this, but...)

PR #11370, opened 2025-07-25
add kernelize to keccak for each data block

chenyuxyz:
hashing 4096 bytes work with this kernelize, otherwise internal stack grows with data length.

compiling with fused ctensor is very slow though.


chenyuxyz:
includes stuff in https://github.com/tinygrad/tinygrad/pull/10764

skipped some tests in CI due to speed, works locally
image

PR #11374, opened 2025-07-25
add HEVC decode support through CUVID to the NV driver

TAbhishekkumar6:


abhishekkchoudharyy:
i am looking into the failed checks so i can implemnt fix

geohot:
Do not use AI

PR #11377, opened 2025-07-26
don't use the wmma args as much

geohot:
Does someone want to do this for all of them and factorize it nicely?


b1tg:
https://github.com/tinygrad/tinygrad/pull/11385

PR #11389, opened 2025-07-27
fix wmma ptx

b1tg:


geohot:
Can you add a test to CI that would have caught this?

PR #11390, opened 2025-07-27
[Bounty] 165+ TFLOP GEMM multistage kernel on 4090 (FP16/BF16)

DaviDX-32:
168 TFLOPS on RTX 4090 (FP16 multistage GEMM), measured by TestSpeed.test_gemm_multistage.


geohot:
Do not use AI

PR #11398, opened 2025-07-27
remove a branch in UOp.r [pr]

chenyuxyz:


Qazalin:
hmm I'd regression test this a bit more than just [pr]. Because high-level Tensor methods aren't process replayed.

Process replay just compares the schedule output of a fixed Tensor graph, not how the Tensor graph gets constructed given higher level operations.

chenyuxyz:
those two branches are equivalent (permaxis is just range(len(self.shape)) and newaxis is just axis)

PR #11414, opened 2025-07-29
[bounty] Muon optim

kevvz:
Remaining things:
- ~~Try to fix the numerical instability in the NS algorithm~~
- ~~Add more tests~~
- Wait for torch muon to come out?
- ~~Fix high LR failures~~

The torch Muon optimizer is exported from another github repo. This implementation apparently has benchmarked very well in the past, so it is a good proxy for muon for testing.
 ~~While a good proxy for torch, it uses bfloat16 casting under the hood, so this can only be tested on select devices. I could use a modified version with the bfloat casting removed and in line with tinygrad's optims until torch releases their version. It will make testing easier without having to wait on an official torch version.~~ I've exported it and tried to match the tgrad muon to be as similar as possible. 

This implementation isn't as clean as I would've liked (in theory 1 line max, since you just need to call NS on the gradient after momentum) , but I think having a correct functionality against a verified source is more important than being concise in this case.


tinygrad/nn/optim.py
```diff
@@ -101,17 +116,23 @@ def _step(self, params:list[Tensor], grads:list[Tensor]) -> tuple[list[Tensor],
...
-        self.b[i].assign(self.momentum * self.b[i].contiguous() + g)  # NOTE: self.b[i] is zero on the first run, no if required
-        g = (g + self.momentum * self.b[i]) if self.nesterov else self.b[i]
+        if not self.ns_params:
+          self.b[i].assign(self.momentum * self.b[i].contiguous() + g)  # NOTE: self.b[i] is zero on the first run, no if required
+          g = (g + self.momentum * self.b[i]) if self.nesterov else self.b[i]
+        else:
+          self.b[i].assign(((1 - self.momentum) * self.b[i]).contiguous() + g * self.momentum)
+          g = (g * (1 - self.momentum) + self.momentum * self.b[i]) if self.nesterov else self.b[i]
+
+      if self.ns_params is not None: g = Tensor.newton_schulz(g.detach().reshape(g.shape[0], -1), params=self.ns_params).reshape(g.shape)
```
geohot: g.detach().reshape(g.shape[0], -1).newton_schulz(self.ns_params).reshape(g.shape)

tinygrad/nn/optim.py
```diff
@@ -101,17 +116,23 @@ def _step(self, params:list[Tensor], grads:list[Tensor]) -> tuple[list[Tensor],
...
-      if self.wd > 0: g = g + self.wd * t.detach()
+      if self.wd > 0 and not self.ns_params: g = g + self.wd * t.detach()
       # classic momentum does post learning rate update
       if self.classic: g = g * r * self.lr
       if self.momentum:
         # TODO: this contiguous is required for correctness because self.b[i] becomes a non contiguous view
         # the scheduler should detect this and just insert contiguous
-        self.b[i].assign(self.momentum * self.b[i].contiguous() + g)  # NOTE: self.b[i] is zero on the first run, no if required
-        g = (g + self.momentum * self.b[i]) if self.nesterov else self.b[i]
+        if not self.ns_params:
```
geohot: not or None, this can be cleaned up

test/test_linalg.py
```diff
@@ -62,5 +62,13 @@ def test_qr_general(self):
       orthogonality_helper(Q)
       reconstruction_helper([Q,R],a)
 
+  def test_newton_schulz(self):
+    coefficients = [(1.5,-0.5), (2,-1.5,0.5)]
+    Tensor.manual_seed(42)#unsure if any random matrix works
```
geohot: Must be fixed before merging.

kevvz: Ran 2000 times with seed removed, no errors

kevvz:
ready

wozeparrot:
want to make beautiful mnist trainable with muon? this way we can check convergence in an actual model.

kevvz:
> want to make beautiful mnist trainable with muon? this way we can check convergence in an actual model.

Adam
python3 examples/beautiful_mnist.py
Image 7-30-25 at 8 33 PM

Muon
python3 MUON=1 examples/beautiful_mnist.py
Image 7-30-25 at 8 30 PM

Muon params
Image 7-30-25 at 8 38 PM

tinygrad/nn/optim.py
```diff
@@ -81,16 +81,31 @@ def SGD(params: list[Tensor], lr=0.001, momentum=0.0, weight_decay=0.0, nesterov
   """
   return LARS(params, lr, momentum, weight_decay, nesterov, classic, tcoef=0.0, fused=fused)
 
+# Muon applies the newton schulz algorithm on gradient. also can include momentum, nesterov, and weight decay
+def Muon(params: list[Tensor], lr=0.02, momentum=0.95, weight_decay=0.0, nesterov=True, \
```
geohot: Where are these defaults from?

examples/beautiful_mnist.py
```diff
@@ -21,7 +21,7 @@ def __call__(self, x:Tensor) -> Tensor: return x.sequential(self.layers)
   X_train, Y_train, X_test, Y_test = mnist(fashion=getenv("FASHION"))
 
   model = Model()
-  opt = nn.optim.Adam(nn.state.get_parameters(model))
+  opt = nn.optim.Adam(nn.state.get_parameters(model)) if not getenv("MUON") else nn.optim.Muon(nn.state.get_parameters(model))
```
geohot: opt = (nn.optim.Adam if not getenv("MUON") else nn.optim.Muon)(nn.state.get_parameters(model))

tinygrad/tensor.py
```diff
@@ -4060,6 +4060,22 @@ def nll_loss(self, Y:Tensor, weight:Tensor|None=None, ignore_index:int|None=None
     nll = -self.gather(1, Y.unsqueeze(1)).squeeze(1) * masked_weight
     return nll.sum() / masked_weight.sum() if reduction == "mean" else nll._do_reduction(reduction)
 
+  def newton_schulz(self, steps:int=5, params=(2,-1.5,0.5)) -> Tensor:
```
geohot: Where are these defaults from? Type for params?

geohot:
This is close, tiny fix in beautiful_mnist and more thought to defaults

tinygrad/tensor.py
kevvz: Also from the blog. But I guess someone wouldn't use the default function if they didn't know what they were doing

tinygrad/nn/optim.py
kevvz: These were ripped from the torch muon reference and they were good parameters for beautiful_mnist, but I guess that's no reason not to conform to tinygrad's defaults. ns_params comes from the blog https://kellerjordan.github.io/posts/muon/ .

geohot: Are the current defaults good for beautiful_mnist?

kevvz: SGD Default
Image 8-1-25 at 3 06 PM

Muon Default
Image 8-1-25 at 3 07 PM

AdamW Default
Image 8-1-25 at 3 08 PM

Muon Torch Spec
472802305-cc3cf074-f94b-4432-b111-6f2e73621653

Muon is just SGD with Newton Schulz, so I would say it is pretty good compared to SGD. AdamW default looks much better, but I can't tell if they are hand tuned or not. With the torch author's params, Muon looks competitive. Muon does feel slower regardless.

These params will be hand tuned anyways so I really don't have a hard preference for the defaults. I would keep it as I have it now since I prefer a direct comparison over SGD, but I could understand keeping the torch spec for a competitive comparison over Adam/AdamW.

geohot: Yea I want the MUON flag to be good by default!

kevvz: Done

geohot:
It's broken with FUSE_OPTIM it seems.

```
light@blak:~/tinygrad$ ADAM=1 python3 examples/beautiful_mnist.py
loss:   0.06 test_accuracy: 97.95%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 11.06it/s]
light@blak:~/tinygrad$ FUSE_OPTIM=1 ADAM=1 python3 examples/beautiful_mnist.py
loss:   0.05 test_accuracy: 98.29%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 10.32it/s]
light@blak:~/tinygrad$ MUON=1 python3 examples/beautiful_mnist.py       
loss:   0.04 test_accuracy: 98.36%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:08<00:00,  8.72it/s]
light@blak:~/tinygrad$ FUSE_OPTIM=1 MUON=1 python3 examples/beautiful_mnist.py
loss:   0.35 test_accuracy: 93.81%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 10.18it/s]
light@blak:~/tinygrad$ 
```

FUSE_OPTIM shouldn't change behavior, though it's possible FUSE_OPTIM Is also broken with SGD?

geohot:
FUSE_OPTIM not working seems MUON specific. With `opt = nn.optim.SGD(nn.state.get_parameters(model), lr=0.1, momentum=0.9)` in beautiful_mnist it's fine
```
light@blak:~/tinygrad$ python3 examples/beautiful_mnist.py             
loss:   0.07 test_accuracy: 97.76%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 10.66it/s]
light@blak:~/tinygrad$ FUSE_OPTIM=1 python3 examples/beautiful_mnist.py
loss:   0.09 test_accuracy: 96.93%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 10.20it/s]
light@blak:~/tinygrad$ FUSE_OPTIM=1 python3 examples/beautiful_mnist.py
loss:   0.09 test_accuracy: 97.59%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 10.63it/s]
```

test/test_optim.py
```diff
@@ -83,6 +87,14 @@ def test_multistep_sgd_nesterov_momentum_wd(self):
   def test_multistep_sgd_high_lr_nesterov_momentum_wd(self):
     self._test_sgd(10, {'lr': 9, 'momentum': 0.9, 'nesterov': True, 'weight_decay': 0.1}, 1e-5, 3e-4)
 
+  def test_muon(self): self._test_muon(1, {'lr': 0.001 }, 1e-6, 0) #TODO:change to torch.muon when it comes out
+  def test_muon_high_lr(self): self._test_muon(1, {'lr': 10}, 1e-6, 3e-4)
+  def test_muon_wd(self): self._test_muon(1, {'lr': 0.001, 'weight_decay': 0.1}, 1e-6, 0)
+  def test_multistep_muon_momentum(self): self._test_muon(10, {'lr': 0.001, 'momentum': 0.5}, 1e-6, 0)
```
geohot: nesterov with muon is untested

kevvz: Its set to True in the reference implementation. Making a test for nesterov=False

kevvz:
> FUSE_OPTIM not working seems MUON specific. With `opt = nn.optim.SGD(nn.state.get_parameters(model), lr=0.1, momentum=0.9)` in beautiful_mnist it's fine
> 
> ```
> light@blak:~/tinygrad$ python3 examples/beautiful_mnist.py             
> loss:   0.07 test_accuracy: 97.76%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 10.66it/s]
> light@blak:~/tinygrad$ FUSE_OPTIM=1 python3 examples/beautiful_mnist.py
> loss:   0.09 test_accuracy: 96.93%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 10.20it/s]
> light@blak:~/tinygrad$ FUSE_OPTIM=1 python3 examples/beautiful_mnist.py
> loss:   0.09 test_accuracy: 97.59%: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 70/70 [00:06<00:00, 10.63it/s]
> ```

I wouldn't call it wrong, Muon needs local shape information for it to be correct. Catting the buffers into one dimension then applying NS is like comparing a matmul of two flattened tensors and a matmul of the original shaped tensors. It is two different calculations altogether.

geohot:
So FUSE_OPTIM is wrong? Why is it not an assert?

geohot:
Are you sure the momentum change is right? https://chatgpt.com/share/688fbfb1-2048-8000-a685-0419d52453e6

kevvz:
> So FUSE_OPTIM is wrong? Why is it not an assert?

It's just in case someone wants to use multiple optimizers. Keller Jordan in this blog https://kellerjordan.github.io/posts/muon/ recommends using AdamW for embedding layers and Muon for everything else.

kevvz:
> Are you sure the momentum change is right? https://chatgpt.com/share/688fbfb1-2048-8000-a685-0419d52453e6

Keller Jordan uses it in the reference implementation https://github.com/KellerJordan/Muon . However I did some digging and found https://github.com/MoonshotAI/Moonlight/blob/master/examples/toy_train.py . They use normal momentum instead of the one found in the first repo. They also seem to be in the running for implementing torch muon https://github.com/pytorch/pytorch/issues/148819#issuecomment-3076551857 .I tested using normal momentum and it seemed to yield similar results, so I just removed the other momentum code. I just wanted to avoid changing the reference implementation as much as possible.

tinygrad/nn/optim.py
```diff
@@ -101,17 +116,19 @@ def _step(self, params:list[Tensor], grads:list[Tensor]) -> tuple[list[Tensor],
         r2 = g.square().sum().sqrt()
         r:Tensor|float = (r1 > 0).where((r2 > 0).where(self.tcoef * r1 / (r2 + self.wd * r1), 1.0), 1.0)
       else: r = 1.0
-      if self.wd > 0: g = g + self.wd * t.detach()
+      if self.wd > 0 and self.ns_params is None: g = g + self.wd * t.detach()
```
geohot: do you need three if ns_params?

kevvz: I listened to the meeting today, and in hindsight ns_params isn't that readable, I should've been more clear. Regardless I have to predict what torch will implement in their version. All signs lead to them doing weight decay after NS + momentum. ns_params checks are gone, but the weight decay has to stay. I could be wrong, but it's most likely.

geohot:
Like I'm reading this, and there's a whole bunch of nitpicks. Read the diff over and over. It can be cleaner and easier to read.

kevvz:
- Moved Muon arguments to have hyper-parameters first and flags second. NS params are at the end. LARS arguments are also moved accordingly. tcoef breaks that pattern for some reason so its treated as a flag
- Removed function arguments that don't make sense in Muon (classic) 
- Updated tests in optim to accept ns_params and ns_steps, since these two are the relevant hyperparams
-  Updated newton_schulz to accept batched dims (although in Muon you have to change shape to 2 dims first). Tests changed accordingly

kevvz:
getting an unrelated test fail 
test/test_jit.py::TestJitGraphSplit::test_jit_multidev_xfer

wozeparrot:
reach out to george@tinygrad.org to claim bounty, can pay via paypal or USDC

PR #11415, opened 2025-07-29
BPE tokenizer

leopf:
replaces the old tokenizer with a byte pair tokenizer.

To make the tokenization fully match with the reference, the words must be split into smaller words with a regex pattern. That is still missing.

Consistency has improved. Testing with `test/external/external_test_simple_tokenizer.py` the error rate has gone from 10/11 to 10/22.

I added tests for some of the BPE edge cases.

### DIFF
--- a/test/external/external_llm_eval.py
+++ b/test/external/external_llm_eval.py
@@ -21,12 +21,12 @@ if __name__ == "__main__":
                '\n'.join([f"{k}) {v}" for k,v in zip(choices['label'], choices['text'])]) +\
                "\n\nReply with the letter of the correct answer only."
     try:
-      ids = [bos_id] + tok.role(b"user") + tok.encode(phrasing.encode()) + [eos_id] + tok.role(b"assistant") + tok.encode(b"Answer: ")
-    except KeyError:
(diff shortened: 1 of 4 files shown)


geohot:
Does this match tiktoken for everything?

leopf:
Not yet, it improves consistency and is on the way to full compatibility but not an exact match yet.

toktoken is basically 3 steps.
1. extract special tokens and split text into segments - **this pr**
2. split segments further with regex - **~~missing~~ this pr**
3. do the bpe tokenization on the segments - **this pr**

regex step is a little annoying because the model specifc regexes are not syntax-compatible with pythons `re` regex engine.

leopf:
Almost at full compatibility now. There still is a bug where some cyrillic texts aren't merged correctly.
Current error rate of `test/external/external_test_simple_tokenizer.py` is 1708/84437.

Its a pretty big diff. Ill try to cleanup more, but there are a lot of different things in BPE tokenizers unfortunately.

leopf:
Fully compatible now.
Tested by:
- `test/external/external_test_simple_tokenizer.py` testing against the OASST1 dataset
- `test/unit/test_llm_tokenizer.py` fast tests of edge cases for CI

PR #11441, opened 2025-07-30
feat: flag for training on val

wozeparrot:


examples/mlperf/model_train.py
```diff
@@ -1344,7 +1345,7 @@ def train_step(model, tokens):
 
   # overfitting this example should give cross_entropy log(BS)
```
chenyuxyz: delete this comment too

PR #11445, opened 2025-07-30
[bounty] tensor: functionalize 1D scalar __setitem__ to fuse TestSetitemLoop.test_arange into one kernel

skydusky:
## Summary
Functionalizes `tensor[int] = value` for 1D contiguous tensors by replacing side-effectful assign+realize with a pure `Tensor.where(...)` expression and `self.replace(...)`. This exposes a single composed graph and lets the scheduler fuse the setitem loop into one compute kernel.

## Performance
- `TestSetitemLoop.test_arange (N=10)`: ~20 kernels → **1 compute kernel** (copies remain)
- Loop realizes as one compute kernel; the test also runs a separate kernel for the reference Tensor.arange(N)

## Details
- Conditions: 1D tensor, contiguous, scalar index; negative indices normalized
- Update: boolean mask from `arange(N, int32) == idx`; `Tensor.where(mask, v_broadcast, self)`; repoint via `self.replace(updated)`
- Fallback path unchanged for slices, multi-dim, non-contiguous

## Behavior
- Last-write-wins preserved
- OOB indexing behavior preserved (no-op)
- Bool dtype safe via `Tensor.where`
- Aliasing: fast path is functional; other patterns keep original in-place semantics via fallback

## Environment
Validated on CPU & METAL; CUDA to be exercised by CI.

## Repro
```bash
DEBUG=2 pytest test/test_setitem.py::TestSetitemLoop::test_arange -q -s
pytest test/test_setitem.py -q
```

## Tests
test/test_setitem.py: 19 passed, 1 skipped

Dtypes (int/float/bool), negatives, self-on-RHS, alias smoke, reshape→contiguous all pass

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -1266,6 +1266,20 @@ class Tensor(MathTrait):
     if not isinstance(v, Tensor): raise TypeError(f"can't set a {type(v).__name__} to a Tensor")
     if self.requires_grad or v.requires_grad: raise NotImplementedError("setitem with requires_grad is not supported")
 
+    # --- fast functional path for 1D scalar indexing (enables fusion) ---
+    if isinstance(indices, (int, get_args(ConstType))) and len(self.shape) == 1 and unwrap(self.uop.st).contiguous:
+      N = self.shape[0]
+      idx = int(indices)
+      if idx < 0: idx += N  # normalize negative indices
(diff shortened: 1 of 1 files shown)


RMartires:
hey @skydusky @chenyuxyz i see this bug bounty was not merged 
is it still open ?

PR #11450, opened 2025-07-31
tdma for am dmabuf export

uuuvn:
slightly more complicated than the previous version to allow exporting buffers that aren't contiguous in bar memory

with this bert can run at BS=84 with ibv cross-host transfers.

`AMD=1 DEBUG=0 python tinygrad/runtime/ops_remote.py`

`REMOTE=1 HOST=192.168.200.6:6667 PYTHONPATH="." MODEL="bert" DEFAULT_FLOAT="HALF" SUM_DTYPE="HALF" GPUS=6 BS=84 EVAL_BS=84 FUSE_ARANGE=1 FUSE_ARANGE_UINT=0 BASEDIR="/raid/datasets/wiki" PARALLEL=0 RUNMLPERF=1 python3 examples/mlperf/model_train.py` -> ~1120ms

`REMOTE=1 HOST=192.168.200.6:6667*3,192.168.200.4:6667*3 PYTHONPATH="." MODEL="bert" DEFAULT_FLOAT="HALF" SUM_DTYPE="HALF" GPUS=6 BS=84 EVAL_BS=84 FUSE_ARANGE=1 FUSE_ARANGE_UINT=0 BASEDIR="/raid/datasets/wiki" PARALLEL=0 RUNMLPERF=1 python3 examples/mlperf/model_train.py` ->  ~1350 ms

### DIFF
--- a/extra/tdma/tdma.c
+++ b/extra/tdma/tdma.c
@@ -1,0 +1,269 @@
+#include <linux/module.h>
+#include <linux/miscdevice.h>
+#include <linux/pci.h>
+#include <linux/pci-p2pdma.h>
+#include <linux/dma-buf.h>
+#include "tdma.h"
+
+// To see pr_debug logs:
+// echo 'module tdma +p' | sudo tee /sys/kernel/debug/dynamic_debug/control
+
+struct tdma_priv {
+  struct pci_dev *pdev;
+  int bar;
+  resource_size_t bar_start;
+  resource_size_t bar_size;
+  int usgl_size;
+  struct tdma_ioctl_usge* usgl;
+};
+
+static int tdma_attach(struct dma_buf *dmabuf, struct dma_buf_attachment *attach) {
+  struct tdma_priv *priv = dmabuf->priv;
+  pr_debug("tdma: %s is attaching to %s", dev_name(attach->dev), dev_name(&priv->pdev->dev));
+
+  // Negative distance = can't reach
+  if (pci_p2pdma_distance(priv->pdev, attach->dev, true) < 0) {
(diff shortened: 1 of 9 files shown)


extra/tdma/tdma.c
```diff
@@ -0,0 +1,269 @@
...
+    kfree(arg_usgl);
+    return PTR_ERR(dmabuf);
+  }
+  // DMABuf is created succesfully, the ownership of priv, usgl and pdev is now transfered to it and will be released when dmabuf is released
+
+  int dmabuf_fd = dma_buf_fd(dmabuf, O_CLOEXEC); // TODO: flags should be from userspace. for now cloexec as a sane default
+  if (dmabuf_fd < 0) {
+    pr_err("tdma: failed to create dmabuf fd: %d", dmabuf_fd);
+    dma_buf_put(dmabuf);
+    return dmabuf_fd; // if dmabuf_fd < 0 it's the error code instead of an fd
```
nimlgen, line 217: leakage of `arg_usgl` and others here? just use several gotos for deallocation

uuuvn: > leakage of `arg_usgl` and others here? just use several gotos for deallocation

from comment a couple lines earlier:

> DMABuf is created succesfully, the ownership of priv, usgl and pdev is now transfered to it and will be released when dmabuf is released

after dmabuf is created successfully (the code in question creates dmabuf-**fd** from dmabuf)  all the resources, including usgl are freed in `tdma_release`

wozeparrot:
This has the same problem as last time, which is that I'm not sure that we want to maintain a kernel module.

uuuvn:
> This has the same problem as last time, which is that I'm not sure that we want to maintain a kernel module.

image

this version is slightly more complicated than the one this was about, but still really simple

PR #11453, opened 2025-07-31
Print gb/s instead of gbs

eitanturok:
Print gb/s instead of gbs in speed_v_theoretical test.

This was confusing because gbs seems like gigabytes whereas gb/s is gigabytes per second


chenyuxyz:
flops and gbs are fine i think

PR #11457, opened 2025-07-31
viz: profile the compiler in TINY device

Qazalin:
The cyan ones right after get_program are the compiler:
image
Same click logic should work with the "compile E_xx" nodes.


geohot:
Good to merge?

PR #11468, opened 2025-08-01
Add Onnx Huggingface to test/models/test_onnx.py

geohotstan:
I changed the directory of the files saved to `_ensure_downloads_dir()` so that no extra code is needed to ensure models are cached.

typical use is 
1. `python extra/huggingface_onnx/huggingface_manager.py --limit 50` to dry run
2. see output yaml size to know it's ok to download
3. `python extra/huggingface_onnx/huggingface_manager.py --limit 50 --download`
4. `python extra/huggingface_onnx/run_models.py --validate huggingface_repos.yaml` or handpick models from the yaml to add to `test/models/test_onnx.py`

TODO:
run top 100 and cleanup.
I accidentally deleted all my models and have to redownload it again to make sure everything's good.
Will exit draft when ready.

### DIFF
--- a/tinygrad/device.py
+++ b/tinygrad/device.py
@@ -3,8 +3,8 @@ from dataclasses import dataclass, replace
 from collections import defaultdict
 from typing import Any, Generic, TypeVar, Iterator
 import importlib, inspect, functools, pathlib, os, platform, contextlib, sys, re, atexit, pickle, decimal, time
-from tinygrad.helpers import CI, OSX, LRU, getenv, diskcache_get, diskcache_put, DEBUG, GlobalCounters, flat_mv, PROFILE, temp, MAX_BUFFER_SIZE, \
-                             colored, Context, DISABLE_COMPILER_CACHE, ALLOW_DEVICE_USAGE, cpu_events, ProfileEvent, ProfilePointEvent, dedup
+from tinygrad.helpers import CI, OSX, LRU, getenv, diskcache_get, diskcache_put, DEBUG, GlobalCounters, flat_mv, PROFILE, temp, colored, \
+                             Context, DISABLE_COMPILER_CACHE, ALLOW_DEVICE_USAGE, MAX_BUFFER_SIZE, cpu_events, ProfileEvent, ProfilePointEvent, dedup
 from tinygrad.dtype import DType, ImageDType, PtrDType, dtypes, _to_np_dtype
 from tinygrad.renderer import Renderer
 


.github/workflows/benchmark.yml
```diff
@@ -62,8 +62,6 @@ jobs:
       run: BENCHMARK_LOG=stable_diffusion_xl CAPTURE_PROCESS_REPLAY=0 JIT=1 python3.11 examples/sdxl.py --seed 0 --noshow --timing | tee sdxl.txt
     - name: Run model inference benchmark
       run: METAL=1 python3.11 test/external/external_model_benchmark.py
-    - name: Run huggingface_onnx test
-      run: METAL=1 python3.11 extra/huggingface_onnx/run_models.py test --debug FacebookAI/xlm-roberta-large
```
geohotstan, line 66: moved this to `test_onnx.py` because previous run_models.py was doing validation with ORT anyways, which means the total run time includes ORT.
Makes more sense to run it as test instead of benchmark.

test/models/test_onnx.py
```diff
@@ -137,5 +147,39 @@ def run(img):
...
+    print(f"Validating model: {repo_id}/{model_file} ({file_size / (1024**2):.2f} MB)")
+    validate(onnx_model_path, custom_inputs, rtol=rtol, atol=atol)
+
+  def test_xlm_roberta_large(self):
+    repo_id = "FacebookAI/xlm-roberta-large"
+    model_file = "onnx/model.onnx"
+    custom_inputs = {
+      "input_ids": np.array([[0, 31414, 232, 328, 740, 1140, 12695, 69, 46078, 1588, 2]], dtype=np.int64),
+      "attention_mask": np.ones((1, 11), dtype=np.int64),
+    }
```
geohotstan, line 176: should I use realistic inputs for this like the efficientnet tests above 🤔 
edit: nvm just randint with vocab_size as max.

geohotstan:
few errors for current top 100 models.
attribute 5 is graph attribute.
gather bug and segmentation are likely wrong input values.
will go through each model and debug later.
```
  1 Qdrant/bge-small-en-v1.5-onnx-Q/model_optimized.onnx #inaccuracy maybe input problem
  2 HuggingFaceTB/SmolLM2-135M-Instruct/onnx/model.onnx #Segmentation fault 11
  3 cross-encoder/nli-deberta-v3-large/onnx/model.onnx #attribute 5
  4 hooman650/bge-m3-onnx-o4/model_optimized.onnx #external data bug
  5 cross-encoder/nli-deberta-v3-base/onnx/model.onnx #attribute 5
  6 Felix92/onnxtr-parseq-multilingual-v1/model.onnx #gather bug
```

---

also only testing `xlm_roberta_large` for now.
It's pretty easy to add new models now.
Should I add an extra runner to run more models?

.github/workflows/test.yml
```diff
@@ -793,7 +793,7 @@ jobs:
     - name: Run real world test
       run: METAL=1 python -m pytest -n=auto test/models/test_real_world.py --durations=20
     - name: Test models (Metal)
-      run: METAL=1 python -m pytest -n=auto test/models -v --durations=20
+      run: MAX_BUFFER_SIZE=0 METAL=1 python -m pytest -n=auto test/models -v --durations=20
```
chenyuxyz: why change this? if it really is needed for the new tests, it should not affect the old tests

geohotstan: This a hack to get around this
`if (mbs:=getenv("MAX_BUFFER_SIZE", 0)) > 0 and self.size > mbs: raise RuntimeError(f"buffer of size {self.size/1e6:.2f}M is too large")`

MAX_BUFFER_SIZE gets set to a default 200MB for all unit tests from this commit https://github.com/tinygrad/tinygrad/pull/10708/commits/9dc59a512371389873091ce84c27fb0b9d953ebb

and these models are 700MB~ large.
The proper way is probably to `setUp` and `tearDown` for the `TestHuggingFaceOnnxModels` class to override MAX_BUFFER_SIZE to 0, but for w/e reason it just doesn't get set! 

I got frustrated and just passed `MAX_BUFFER_SIZE=0` in as envvar since I saw other places in test.yml use this.

---
I'll take another look at using `setUp` and `tearDown`

test/models/test_onnx.py
```diff
@@ -137,5 +147,41 @@ def run(img):
...
     assert "car" in _LABELS[cls] or _LABELS[cls] == "convertible"
 
+@unittest.skipUnless(HUGGINGFACE_AVAILABLE and Device.DEFAULT == "METAL", "only run on METAL")
+class TestHuggingFaceOnnxModels(unittest.TestCase):
+  @classmethod
+  def setUpClass(cls):
+    cls._old_max_buffer_size = os.environ.get("MAX_BUFFER_SIZE")
+    os.environ["MAX_BUFFER_SIZE"] = "0"
+    # clear tinygrad.helpers.getenv cache so runtime picks up new env
+    getenv.cache_clear()
```
geohotstan: getenv was being cached before 😅 
so setting envvar after `getenv` has ran on that envvar wasn't changing the value on future `getenv`s.

chenyuxyz: make it a ContextVar? messing around with os and cache is bad

chenyuxyz:
cool saw it tests in CI, consider adding a link to the ci test next time which would help reviewer a lot
image

can you confirm the model is cached in ci and won't redownload between runs?

geohotstan:
Sorry. Yeah this PR was not ready when I marked it for review.

The models are all downloaded in the cached directory defined in `tinygrad/helpers.py`
so models will occupy `_ensure_downloads_dir() / "models"`, which is indeed loaded in `setup-tinygrad/actions.yaml`

If I understand this correctly, the models will get cached when `DOWNLOAD_CACHE_VERSION` is incremented? So without incrementing it, it cannot cache despite it already being downloaded to the official cache dir.

I went ahead and incremented `DOWNLOAD_CACHE_VERSION` in [a95016b](https://github.com/tinygrad/tinygrad/pull/11468/commits/a95016bc8534d2de504eb2d8ab984e0b02b480d3) here and ran it with [HF_HUB_OFFLINE](https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#hfhuboffline) in the next commit [b982015](https://github.com/tinygrad/tinygrad/pull/11468/commits/b982015b43a84f83021f0b9bf225382c09ce43b4) and it passed fine. I also confirmed the files are in the cache in the commit after that: [here](https://github.com/tinygrad/tinygrad/actions/runs/16932453777/job/47981037900?pr=11468#step:5:182) 

Gonna revert the DOWNLOAD_CACHE_VERSION to 10 for now and whenever you guys decide to increment, it should be cached then?

---
update also gonna wrap a retry because download is sometimes flaky.

PR #11470, opened 2025-08-01
offload optim to cpu train llama

chenyuxyz:
somehow this ends up using more memory on device...?


geohot:
Feel free to reopen if relevant

PR #11481, opened 2025-08-02
generic double cast folding

chenyuxyz:
b.cast(a).cast(b) -> b if a preserves all values in b


chenyuxyz:
process replay looks good
image
image

PR #11483, opened 2025-08-03
move cast around expand backward to tensor.py

chenyuxyz:


chenyuxyz:
also tested with training bert. used the same memory and similar perf

PR #11485, opened 2025-08-03
combine onnx parser and onnx

geohotstan:
Ran this on top 100 huggingface models and made sure there are no regressions for the new refactored onnx parser.


test/external/external_test_onnx_ops.py
```diff
@@ -251,7 +251,6 @@ def test_momentum_t_greater_than_zero(self):
         outputs = ["X_out", "V_out"]
         self._validate_training("Momentum", onnx_fxn, inputs, attributes, outputs)
 
-  @unittest.expectedFailure  # TODO: regression from removing StrEnum in Domain
```
chenyuxyz: can you fix this in a separate pr (before or after one file is fine)?

one pr is fine only if it's cleanly split by different commits, separating refactors from behavioral changes are important for maintainability

geohotstan:
This PR consistently teases out the segfault for some reason...
Is that a blocker for merging this?

chenyuxyz:
as long as it's not caused by this pr i think it's fine

chenyuxyz:
yea it feels like something in this pr makes more segfault

want to split this into smaller changes? like one that literally just merge files to start with

geohotstan:
cool, segfault's fixed!
Whatever magic george did with contextvar worked for the issue that was plaguing this PR.

PR #11488, opened 2025-08-03
extract conv layer test logic

TheVanadium:
makes adding Conv3d test only a few lines

dim parameter is awkward but I can't really think of a cleaner way to do it while keeping a structure that supports future Conv3d testing; tinygrad's Conv2d layer needs an indicator to determine 3d-ness

### DIFF
--- a/test/test_nn.py
+++ b/test/test_nn.py
@@ -108,7 +108,7 @@ class TestNN(unittest.TestCase):
     _test_linear(Tensor.randn(BS, in_dim), in_dim, out_dim)
     _test_linear(Tensor.randn(BS, T, in_dim), in_dim, out_dim) # test with more dims
 
-  def _test_conv(self, tiny_conv, torch_conv, BS, C1, DIMS, C2, K, S, P, D=1, winograd=False):
+  def _test_conv(self, tiny_conv, torch_conv, BS, C1, DIMS, C2, K, S, P, D=1):
     # create in tinygrad
     layer = tiny_conv(C1, C2, kernel_size=K, stride=S, padding=P, dilation=D)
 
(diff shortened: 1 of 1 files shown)


test/test_nn.py
```diff
@@ -108,45 +108,29 @@ def _test_linear(x, in_dim, out_dim):
...
 
-    # test
-    x = Tensor.uniform(BS, C1, H, W)
-    z = layer(x)
-    torch_x = torch.tensor(x.numpy())
-    torch_z = torch_layer(torch_x)
-    np.testing.assert_allclose(z.numpy(), torch_z.detach().numpy(), atol=5e-4, rtol=1e-5)
+  def test_conv2d(self): self._test_conv(Conv2d, torch.nn.Conv2d, 2)
 
   def test_conv1d_same_padding(self):
```
chenyuxyz, line 138: i like the concept but i think it's all or none. it's good if it covers all of the conv cases, otherwise it's annoying that some tests are in one style and some are in another

TheVanadium: All conv cases are integrated (except winograd, but winograd's instantiation looks different enough to imo add too much complexity to `_test_conv` for a single exception case).

test/test_nn.py
```diff
@@ -245,44 +200,14 @@ def test_conv2d_winograd(self):
...
-      torch_layer.weight[:] = torch.tensor(layer.weight.numpy(), dtype=torch.float32)
-      torch_layer.bias[:] = torch.tensor(layer.bias.numpy(), dtype=torch.float32)
-
-    # test
-    x = Tensor.uniform(BS, C1, W)
-    z = layer(x)
-    torch_x = torch.tensor(x.numpy())
-    torch_z = torch_layer(torch_x)
-    np.testing.assert_allclose(z.numpy(), torch_z.detach().numpy(), atol=5e-4, rtol=1e-5)
+    self._test_conv(ConvTranspose1d, torch.nn.ConvTranspose1d, BS, C1, DIMS, C2, K, S, P)
```
chenyuxyz, line 205: these two are wrong, i am fixing it separately

PR #11493, opened 2025-08-04
fix onnx StrEnum regression

geohotstan:
preparation for #11485

### DIFF
--- a/extra/onnx.py
+++ b/extra/onnx.py
@@ -144,7 +144,7 @@ class OnnxRunner:
   """
   def __init__(self, model_path: Tensor | str | pathlib.Path):
     model = onnx_load(model_path)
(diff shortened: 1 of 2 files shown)


extra/onnx.py
```diff
@@ -144,7 +144,7 @@ class OnnxRunner:
   """
   def __init__(self, model_path: Tensor | str | pathlib.Path):
     model = onnx_load(model_path)
-    self.is_training = any(n.domain in {Domain.AI_ONNX_TRAINING, Domain.AI_ONNX_PREVIEW_TRAINING} for n in model.graph.node)
+    self.is_training = any(n.domain in {Domain.AI_ONNX_TRAINING.value, Domain.AI_ONNX_PREVIEW_TRAINING.value} for n in model.graph.node)
```
chenyuxyz, line 147: hmm domain here is not Enum? this is confusing

geohotstan:
hmmm yeah that is confusing.
Maybe apply this after #11485

PR #11505, opened 2025-08-05
Fix incorrect jit current batch devs reset

uuuvn:
`current_batch_devs = []` (in `flush_batch()`) happens between `new_batched_devs = ...` and `current_batch_devs = new_batched_devs` => doesn't actually reset anything leading to things not jitting properly

which 2xs remote bert step time (should have similar effects on any non-hcq backend)


uuuvn:
hcq backends aren't affected because this basically only happens with cpu which is an hcq and can be in the same graph with real hcq stuff => so the extra cpu device stuck in `current_batch_devs` doesn't actually prevent the jitting

uuuvn:
hard to test for this kind of stuff generically, added a test that should be good enough hopefully

PR #11529, opened 2025-08-06
lift range out of add_chain

S-Lykles:


S-Lykles:
No process replay change, this is similar to some of the stuff I worked on a while ago with the sorting of commutative chains like #9260. Something to keep in mind

PR #11532, opened 2025-08-06
viz: make SINK node grey

ben-waldron-1:
Kernel LOOP axes are coloured white, which is invisible on a white SINK node.
Before:
Screenshot From 2025-08-06 13-46-09
After:
Screenshot From 2025-08-06 13-46-54
After (real kernel):
Screenshot From 2025-08-06 13-49-38
I'm aware that this is a nitpick, but it just looked a bit unpolished imo


Qazalin:
Gray background makes the black text unreadable.
If you want to only give the kernel name text a slightly darker background that's more okay.

PR #11552, opened 2025-08-07
Remove correct_divmod_folding flag

S-Lykles:


S-Lykles:
So this is harder than I thought, some of the gates look like:
```
int alu17 = (idx2+ridx1005);
(((alu17<2)!=1)&(alu17<10));
```

The `<2` instead of `<1` makes this not a normal simplex. I have also seen them with `<3`
This means in uop_given_valid you have to try
idx2.vmin=1, ridx1005.vmin=1
idx2.vmin=0, ridx1005.vmin=2
idx2.vmin=2, ridx1005.vmin=0
etc

PR #11562, opened 2025-08-07
MXFP4 Support

tclesius:
Support for the new quantization format used by openai/gpt-oss. (relevant for `apps/llm.py` bounty).

References:
- [openai/gpt-oss](https://github.com/openai/gpt-oss/blob/d0a300a40d6502a1bdd73d18464f3d69440656e0/gpt_oss/torch/weights.py#L119)
- [llama.cpp](https://github.com/ggml-org/llama.cpp/blob/7ad67ba9fe2b909e271dd31b99c5fce3aba35899/ggml/src/ggml-quants.c#L260)


geohot:
Needs tests

chenyuxyz:
stale

Issue #11548, opened 2025-08-07
z3 hangs while beaming embedding


chenyuxyz:
```
# test.py
from tinygrad import Tensor, dtypes
from tinygrad.nn import Embedding

if __name__ == "__main__":
  embedding = Embedding(30522, 1024)
  input_ids = Tensor.empty(96, 512)
  embedding(input_ids).realize()
```
ran with `IGNORE_BEAM_CACHE=1 BEAM=2 FUSE_ARANGE=1 DEBUG=2 python test.py` hangs if `z3-solver` is installed
<img width="1226" height="114" alt="Image" src="https://github.com/user-attachments/assets/edbc527e-41fe-4a5d-a25b-e0c1baf28c40" />

with `IGNORE_OOB=1` or uninstalling `z3-solver` it runs fine

PR #11563, opened 2025-08-07
add z3 timeout and warning

S-Lykles:
addded a timeout to the solver of 5 seconds. Should also work if you ctrl-c during the solving. In those cases the solver returns `z3.unknown`


S-Lykles:
5 seconds seems to be too slow for some CI...
This expression failed
```
idx=((((((((ridx0*205)>>11)-(7*((((ridx0*205)>>11)*37)>>8)))*11)+((ridx0*937)>>16))+(((ridx0>>1)-(5*(((ridx0>>1)*205)>>10)))*77))+(((((ridx0*3)>>1)+1)-(3*(((((ridx0*3)>>1)+1)*683)>>11)))*385))+((ridx0&1)*1155))
```

S-Lykles:
I coudnt repro #11548 but this should fix it.

chenyuxyz:
what are the expression that takes a long time and should it? #11548 should repro on a tinybox

PR #11575, opened 2025-08-08
benchmark openpilot 0.9.9

chenyuxyz:
not sure what to do with the 0.9.7 ones with IMAGE=2 and validate


chenyuxyz:
@wozeparrot can you clean up rest of the 0.9.7 and decide what to keep / add to dashboard? also a good idea to change default in compile3 too.

i want to update the benchmark to prepare the upcoming kernel / symbolic changes in case of regression

PR #11590, opened 2025-08-09
Fix z3 float cast in indexing

S-Lykles:
Adds support for using a casted float in an index. Previously z3 would raise an exception here


S-Lykles:
This fixes part of #11548, it still hangs a few lines later though

PR #11591, opened 2025-08-09
Late rewrite rules for CMPLT

S-Lykles:


S-Lykles:
PTX gemm seems to be a few percent improvement

PR #11595, opened 2025-08-09
test: conv3d layer

TheVanadium:
If this gets pulled we should add the dimensions note from functional Conv2d. I only added Conv3d tests for cases where we have an equivalent Conv1d test.


chenyuxyz:
conv3d functionality has been tested in test_ops, i don't see the benefit to repeat the tests in nn layer

TheVanadium:
> conv3d functionality has been tested in test_ops, i don't see the benefit to repeat the tests in nn layer

Same applies to conv1d op, but we still have repeated tests for the Conv1d nn layer. For consistency you could argue the Conv1d be removed from test_nn, or if it's not a big enough deal the PR can be closed.

Since there's a Conv3d in PyTorch, it made sense to me to have a comparison test for it. If nothing else, it adds clarification that the "works for any dimension" property of conv2d extends to the nn layer

TheVanadium:
Converting to draft because a lot of tests are failing, I suspect it's due one of the tests being too gpu intensive?

chenyuxyz:
yea not a big deal, Conv1d is tested because there's Conv1d in tinygrad.nn. maybe it's fine to remove some repeated Conv1d tests

the value of tests is to catch regression and unknown errors, don't just add tests for the sake of adding tests

PR #11600, opened 2025-08-10
fix integer floor division by 1 optimization

ntedvs:
Fixes `test_idiv_literal_one` and `test_idiv_tensor_one` by adding early optimization for `x // 1 = x`.

The existing symbolic pattern `(UPat.var("x") // 1, lambda x: x)` doesn't match because integer floor division creates complex `WHERE/AND/CMPLT` UOp structures for Python vs C division semantics.

This change detects division by constant 1 before the complex structure is created and returns the numerator directly, eliminating unnecessary computation.

Fixes both:
- `Tensor([1,2,3,4]) // 1` (literal case)
- `Tensor([1,2,3,4]) // Tensor.ones(...)` (tensor case)

### DIFF
--- a/test/test_const_folding.py
+++ b/test/test_const_folding.py
@@ -81,10 +81,8 @@ class TestBinaryOpsConstFolding(unittest.TestCase):
   def test_div_tensor_one(self):
     _check_ast_count(0, Tensor([1.0, 2, 3, 4]) / Tensor.ones(4))
 
-  @unittest.expectedFailure  # TODO: fix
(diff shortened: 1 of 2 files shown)


chenyuxyz:
that missed the point, tensor.py should not have path that look into its uop

ntedvs:
Are you saying the fix should be in the symbolic optimization layer (symbolic.py) rather than having tensor.py inspect UOp internals? Sorry, I'm new to the project.

PR #11604, opened 2025-08-10
Add onnx If OP support

geohotstan:


extra/onnx.py
```diff
@@ -425,6 +430,9 @@ def __call__(self, inputs:dict[str, Any], debug=debug):
     Tensor.training = self.old_training
     return {name:self.graph_values[name] for name in self.graph_outputs}
 
+class SubGraphOnnxRunner(OnnxRunner):
+  def __init__(self, graph: dict, opset_imports: dict[Domain, int]): self._load_from_graph(graph, opset_imports)
```
geohotstan, line 434: having opset_imports here is due to abstraction limitation of current parser.
New parser abstraction only needs graph.

geohotstan:
Actually maybe add If op support after new abstractions, instead of merging hacks xD
Since there's already a branch with working `sam2_hiear2_tiny`

PR #11609, opened 2025-08-10
list indexing can normalize in python

chenyuxyz:
list index does not need to be normalized in tensor


chenyuxyz:
@geohotstan fyi it's pretty cool

PR #11611, opened 2025-08-11
new x86 backend

ttomsa:
So I quite like this design, it's very information dense and clear (besides the dictionary mess) and keeps the complexity in the pm. There's a clear separation of abstractions, MUOps only know about operands and encoding fields and the pm is the bridge. Hacks to deal with noops or coalescing are gone. A big thing about x86 is that almost every instruction can take a memory operand. Before, adding that functionality would require a bunch of if statements to check for specific ops and dtypes but now we get it for free by just checking if the operand is in the right field. The allocator is still a bit of a mess, I just mostly copied the old implementation. A better way is a priority queue and spill/fill instructions are added to it like any other instruction instead of having the spill code serve as a bridge between register and memory. That's now much easier because we're not dealing with a dag anymore. Still, a decent amount of stuff is left but hopefully this pr will actually get merged.
Also, `DEBUG >=8` prints the assembly with virtual registers.

### DIFF
--- a/tinygrad/codegen/__init__.py
+++ b/tinygrad/codegen/__init__.py
@@ -85,7 +85,7 @@ def full_rewrite_to_sink(sink:UOp, ren:Renderer|None=None, optimize:bool=True) -
   sink = graph_rewrite(sink, symbolic, name="post index symbolic")
 
   # optional pre matcher
-  if ren.pre_matcher is not None: sink = graph_rewrite(sink, ren.pre_matcher, name="pre_matcher")
+  if ren.pre_matcher is not None: sink = graph_rewrite(sink, ren.pre_matcher, ctx=ren, name="pre_matcher")
 
   # decompositions
   supported_ops = tuple(ren.code_for_op.keys())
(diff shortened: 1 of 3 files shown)


ttomsa:
example:
```
from tinygrad import Tensor
a = Tensor([1,2,3])
b = Tensor([1,2,3])
c = a + b
print(c.numpy())
```

before register allocation:
```
mov v0, v0
mov v1, v1
mov v2, v2
mov v3, 0
.LOOP_0:
movsxd v4, v3
mov v5, v4
shl v5, 2
mov v6, v1
add v6, v5
mov v8, [v6]
mov v9, v2
add v9, v5
mov v11, [v9]
mov v12, v8
add v12, v11
mov v13, v0
add v13, v5
mov [v13], v12
add v3, 1
cmp v3, 3
jl .LOOP_0:
```

after register allocation:
```
mov rdi, rdi
mov rsi, rsi
mov rdx, rdx
mov rax, 0
.LOOP_0:
movsxd rcx, rax
mov rcx, rcx
shl rcx, 2
mov rbx, rsi
add rbx, rcx
mov r8, [rbx]
mov rbx, rdx
add rbx, rcx
mov r9, [rbx]
mov r8, r8
add r8, r9
mov r9, rdi
add r9, rcx
mov [r9], r8
add rax, 1
cmp rax, 3
jl .LOOP_0:
ret
```

ttomsa:
this is almost ready I'm going to start opening sub prs

ttomsa:
@geohot this is ready

geohot:
We discussed this on Discord, I don't think assembly backends should have MUOp. Most of the lowering to assembly should be prelinearization.

PR #11612, opened 2025-08-11
Support cubic mode for ONNX Resize OP 

geohotstan:


extra/onnx.py
```diff
@@ -762,7 +762,42 @@ def _apply_transformation(input_sz, output_sz, scale_dim, mode):
...
+        reshape[i] = expand[i] = sizes[i]
+
+        p = index.floor().int()
+        ratio = index - p
+
+        # Calculate indices
+        idx0, idx1, idx2, idx3 = [p + i for i in [-1, 0, 1, 2]]
+
+        # Calculate coefficients
+        c0 = ((A * (ratio + 1) - 5 * A) * (ratio + 1) + 8 * A) * (ratio + 1) - 4 * A
```
chenyuxyz: can you add a comment / reference to these? it's not clear to me why is it `(A * (ratio + 1) - 5 * A` and not `(A * (ratio - 4))` for example

chenyuxyz: having these as polynomials of ration and (1-ratio) seems reasonable, but (ratio+1) is weird to me

extra/onnx.py
```diff
@@ -755,14 +755,60 @@ def _apply_transformation(input_sz, output_sz, scale_dim, mode):
...
-    if mode == "cubic": raise NotImplementedError("cubic interpolation is not implemented")
+
+    if mode == "cubic":
+      A = cubic_coeff_a
+      def keys_kernel(x:Tensor):
+        # https://wikimedia.org/api/rest_v1/media/math/render/svg/7835668875c7c0284063c9f2eafc3b1022692e03
+        ax = x.abs()
+        ax2 = ax * ax
+        ax3 = ax2 * ax
+        w0_1 = (A + 2) * ax3 - (A + 3) * ax2 + 1
```
chenyuxyz: use `polyN`?

geohotstan:
This current vers is a bit slower than the `sam` onnx model branch version, but I think it's more correct and more readable. (after stressing over this a bit, I think I see some other simplifications...)

I think in the onnx reference impl (which is where I got `ratio+1` from), they do some weird simplification with the knowledge that ratio is (0, 1] to start: https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_resize.py#L231-L237

They end up simplifying away the `.abs()` in the piecewise function.

geohotstan:
ok gemini can indeed vibe faster cubic interp impls xD, but they're all pretty crazy and harder to understand.

chenyuxyz:
cool yea this looks fine. perf improvements are welcome but we also need better profiling script to justify the tradeoff

PR #11613, opened 2025-08-11
rewrite sort idx

chenyuxyz:
`((idx.unsqueeze(dim) <= idx.unsqueeze(dim+1))` is just a lower triangular. not clear why rewrite did not figure it out, but this made `JITBEAM=2 PYTHONPATH=. python3 examples/olmoe.py` 5% faster.

more can be done in topk by only compute the idx for the topk and not the full sorted stuff

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -2871,11 +2871,11 @@ class Tensor(MathTrait):
         x = blue_box.cat(flipped_green_box.flip(flip_dims), dim=crossover_dim)
     x = x.flatten(dim, dim+n_stages-1).shrink(tuple((0, s) for s in self.shape))
     # compute indices for sorted values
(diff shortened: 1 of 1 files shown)


geohotstan:
> more can be done in topk by only compute the idx for the topk and not the full sorted stuff

pytorch's topk doesn't have a `stable` parameter so it expects the full sorted idx (if I understood you correctly)
I have tried adding our own `topk(..., stable=False)` and the speed up wasn't much :s

chenyuxyz:
oh it's not related to stable. more like in the last stage of comparison, we don't need to cat top and bottom if k is smaller than half of the dim. can go back more if we are picking top 8 of 64

chenyuxyz:
i am not sure how much can be gained from `sel` on olmoe
`sel = ((probs.unsqueeze(-1)==g)*Tensor.arange(self.num_experts)).sum(1)` does not really help, and i don't think it can be simpler than this

anyway merging this one as i think it's slightly more readable

PR #11615, opened 2025-08-11
viz: timeline perf try 2

Qazalin:


Qazalin:
Two issues blocking this patch:
1. Proxy rects get smaller when zoomed in
2. Lack of tooltip info in merged rects, eg. it should say how many events are merged

It also doesn't address the memory graph, moving to https://github.com/tinygrad/tinygrad/pull/11898

PR #11641, opened 2025-08-12
simplify onnx cubic

chenyuxyz:
we can drop the double where and abs since we know which ranges the inputs map into


chenyuxyz:
@geohotstan fyi

PR #11646, opened 2025-08-13
add launch bounds to ptx

S-Lykles:


S-Lykles:
Tensor core GEMM is ~10% faster for the bigger matrices

PR #11559, opened 2025-08-07
support sam2_hiear2_tiny onnx

geohotstan:
wrote hacks to pass cubic interp test.
There are a lot of problems with that resize op. 
I have some of the problems resolved in https://github.com/tinygrad/tinygrad/pull/9609/files, so I might go from there to once and for all write a correct resize op.

Regarding perf on m1 pro:
```
(tinygrad) zibo@cloud:~/fun/tiny/tinygrad$ MODEL=sam2_hiear2_tiny PYTHONPATH=. python test/external/external_model_benchmark.py 
sam2_hiear2_tiny tinygrad_metal_jitless    1860.92 ms
sam2_hiear2_tiny tinygrad_metal_jit          34.16 ms
sam2_hiear2_tiny tinygrad_cpu_jitless      2655.32 ms
sam2_hiear2_tiny tinygrad_cpu_jit           679.48 ms
sam2_hiear2_tinyonnx2torch       NotImplementedError
sam2_hiear2_tiny onnxruntime_cpu             66.50 ms
sam2_hiear2_tiny onnxruntime_coreml          68.97 ms
sam2_hiear2_tinyoutputs validated on device='METAL' with rtol=2.0e-03, atol=2.0e-03
sam2_hiear2_tinyoutputs validated on device='CPU' with rtol=2.0e-03, atol=2.0e-03
```

Here're the ops used:
```
Counter({'Constant': 923, 'Unsqueeze': 234, 'Cast': 151, 'Shape': 122, 'Gather': 122, 'Add': 121, 'Mul': 109, 'Concat': 98, 'Transpose': 89, 'Reshape': 84, 'MatMul': 75, 'Div': 67, 'Mod': 48, 'Squeeze': 36, 'Sub': 35, 'Slice': 27, 'LayerNormalization': 25, 'Split': 12, 'Softmax': 12, 'Tanh': 12, 'ConstantOfShape': 9, 'Pad': 9, 'Gemm': 7, 'MaxPool': 6, 'Relu': 5, 'If': 2, 'Conv': 1, 'Resize': 1, 'Size': 1, 'Greater': 1, 'Less': 1, 'Tile': 1, 'ReduceMean': 1, 'Flatten': 1, 'ReduceSum': 1, 'Sqrt': 1, 'Clip': 1})
```

PR #11648, opened 2025-08-13
Support onnx If OP

geohotstan:
Thinking of adding both sam_hiear_tiny and a model from huggingface that uses `If` into tests.
Should I do it this PR or another PR? I guess adding both of those does validate If op... 🤔 
also if add test, then I guess #11559 can be closed and this PR can fix that issue.

Also I verified that `SubGraphOnnxRunner` is enough to run `Loop` op too. I haven't tried Scan yet.


chenyuxyz:
let's add `sam_hiear_tiny` only after comma launches that model, my understanding is that they are still updating / exploring the model.

I think unit tests in this pr is good enough to merge, is this faster than the old pr? if so, we can close the old pr and I will ask them for more feedback.

geohotstan:
Figured out the reason for slowness from [ef0df0d](https://github.com/tinygrad/tinygrad/pull/11648/commits/ef0df0d69982e9feed41b224b43471eb6b2a305f)

reason was `SubGraphOnnxRunner` was holding references to `Buffer`s (Tensor) preventing it from `deallocate` by python gc.
speed is the same as old branch when I hand dereference the Tensors (3 less kernels compared to old branch).
Maybe there's a better way than subclassing `OnnxRunner`...

chenyuxyz:
i heard that the if in that model is just some consts, and it's fine if they manually remove that layer in the model. Maybe consider supporting a more basic case here as const is free?

chenyuxyz:
i merged the test and proposed fix

geohotstan:
> i heard that the if in that model is just some consts, and it's fine if they manually remove that layer in the model. Maybe consider supporting a more basic case here as const is free?

Yeah there are a lot of weird parts to that model

Like all the Gather Ops are doing indexing on Shape Op output (Shape Op returns the shape of a Tensor as a Tensor)
Out of all the Gather Ops, there are also several that's doing: 
```
91: op 'Gather' opt {'axis': 0}
	inputs:
		/vision_model/vision/_en/blocks.0/Shape_2_output_0 - <Tensor <UOp METAL (1,) long (<Ops.COPY: 9>, None)> on METAL with grad None>
		/vision_model/vision/_en/blocks.0/Constant_16_output_0 - <Tensor <UOp METAL (1,) long (<Ops.COPY: 9>, None)> on METAL with grad None>
	outputs:
		/vision_model/vision/_en/blocks.0/Gather_2_output_0 - <Tensor <UOp METAL (1,) long (<Ops.COPY: 9>, None)> on METAL with grad None>
```
which... lol

I'll see what I can do.

geohotstan:
[WOOOOOOOO SHAPE CACHE 50% SPEED INCREASE](https://github.com/tinygrad/tinygrad/pull/11648/commits/437ece9076375be8090481c551b88c55a6745625)
sry 100% speed increase since jitted execution time is halved...

geohotstan:
If op can only use `where` when output shapes are equal. `sam2` model hits movement ops shape valueerrors if the output of if op is broadcasted.

If this If op implementation is fine, there are a few follow up PRs to this:
1. shape tensor caching. I'll show the before and after speed of a few models that does shape ops to justify randomly caching Tensors in model fp.
2. `OnnxNode.replace`, like `UOp.replace`, since the OnnxNode recreation stuff isn't exactly pretty.
3. pre-realize non input-dependent values

tinygrad/frontend/onnx.py
```diff
@@ -445,9 +455,10 @@ def get_empty_input_data(self, device:str|None=None, dtype:DType|None=None) -> d
     return {name:Tensor.empty(*spec.shape, device=device, dtype=dtype or spec.dtype) for name, spec in self.graph_inputs.items()}
 
   def to(self, device:str|None):
-    self.graph_values = {k:v.to(device) if isinstance(v, Tensor) else v for k,v in self.graph_values.items()}
+    self.graph_values = {k: (v.to(device) if isinstance(v, Tensor) else v) for k,v in self.graph_values.items()}
     self.graph_nodes = tuple(OnnxNode(n.op, n.opset_id, tuple(n.inputs), tuple(n.outputs),
-                                      {k:v.to(device) if isinstance(v, Tensor) else v for k,v in n.opts.items()}) for n in self.graph_nodes)
+                                      {k: (v.to(device) if isinstance(v, Tensor) or isinstance(v, OnnxRunner) else v) for k,v in n.opts.items()})
```
chenyuxyz: `isinstance(v, (Tensor, OnnxRunner))`

tinygrad/frontend/onnx.py
```diff
@@ -543,6 +554,23 @@ def ___wrapper(R:Tensor, T:int, *inputs:Tensor, **kwargs):
...
+      # dereference intermediate tensors so Buffer can be deallocated
+      for k in intermediate_tensors: del branch.graph_values[k]
+      return out
+    # both branch must be ran before the condition can be evaluated
+    else_out, then_out = run_branch(else_branch), run_branch(then_branch)
+    assert len(else_out) == len(then_out), f"else_out and then_out must have the same number of outputs: {len(else_out)} != {len(then_out)}"
+    # can use where op when output shape is the same
+    if all(t.shape == e.shape for t,e in zip(then_out.values(), else_out.values())):
+      return tuple(condition.where(t,e) for t,e in zip(then_out.values(), else_out.values()))
+    # otherwise, use condition to select the output in python
```
chenyuxyz, line 570: `where` supports broadcasting too right? eval python is wrong in JIT (unless condition is a const and it degenerates)

geohotstan: the output of If OP is not expected to be broadcasted.
When it is broadcasted, the model hits Exception `ValueError: cannot broadcast (1, 1, 1, 1, 1, 1, 1, 96, 1, 8, 1, 8) to new_shape=(1, 1, 1, 1, 1, 1, 1, 1, 4, 1, 8, 96)` during in the next few ops (Tile).

command:
`ORT=1 PYTHONPATH=. python examples/benchmark_onnx.py https://github.com/haraschax/filedump/raw/refs/heads/master/sam2_hiear2_tiny.onnx`

geohotstan:
I just realized `external_model_benchmark` does not use `optimize=True` for its JIT. (unlike `benchmark_onnx`)
So the Shape Tensor cache doesn't seem to have any speed up after `optimize=True`
...

I don't understand JIT 😅 
Why don't we use `optimize=True` everywhere?

tinygrad/frontend/onnx.py
chenyuxyz: printing some shapes, i am seeing
```
condition.shape=(1,), list(then_out.values())[0].shape=(4,), list(else_out.values())[0].shape=(2,)
condition.shape=(1,), list(then_out.values())[0].shape=(1, 1, 1, 96, 8, 8), list(else_out.values())[0].shape=(1, 96, 8, 8)
```

i don't quite understand how this works if subgraph can have non-broadcastable shapes... do both path work with downstream stuff? If the design has a true branching then it's data dependent and we cannot quite support that

geohotstan: so this model is weird
image

both does work with downstream stuff. If one IF op is True, the other one must be False.
so it's either (4,) and (1, 96, 8, 8), or (2,) and (1, 1, 1, 96, 8, 8)

the (4,) and (2,) are args for the `Tile` op (repeat)

geohotstan: what's even more weird is this is downstream from model initializer values (non model input). 
model inputs are `['img', 'big_img']`

chenyuxyz: weird okay, so in this case the whole right sub tree is just some const and we should precompute them all?

geohotstan: yeah it can be precomputed.

update: I tried implementing a thing that runs nodes that are non-input-dependent during `OnnxRunner.__init__` and also pruning the graph of those nodes. It seems to decrease jit memory by a lot for this model. Not much real speed up though (I'm not sure why).

from
```
JIT memory reduced from 0.92 MB -> 0.79 MB, 70 -> 1 bufs
memory reduced from 198.62 MB -> 198.97 MB, 321 -> 1 bufs
```

to 
```
JIT memory reduced from 0.18 MB -> 0.02 MB, 75 -> 1 bufs
memory reduced from 196.22 MB -> 196.57 MB, 312 -> 1 bufs
```

PR #11653, opened 2025-08-13
remove need for cast_vec

geohot:


Rat5ak:
Fucking Oath George, fucking oath.

PR #11657, opened 2025-08-13
amd: enable sdma bind for sdma

nimlgen:
this should be default for am, will run more trainings to see it's safe


geohot:
Still relevant?

PR #11663, opened 2025-08-14
Remove realize from __setitem__ and get TestSetitemLoop.test_arange to be one kernel

oxrinz:
setitem now computes lazily without realize using ops.where


oxrinz:
can't bother, pretty mechanical bounty, if someone wants to take over feel free, this has the most progress out of previous attempts. prob an hour or two of work left

PR #11667, opened 2025-08-14
add Ops.CDIV and Ops.CMOD

S-Lykles:
adding the ops so I can see process replay if I change IDIV to be floor div


Qazalin:
hmm, considering a feature to change the reference commit of process replay.

S-Lykles:
> hmm, considering a feature to change the reference commit of process replay.

That would also be nice

Qazalin:
https://github.com/qazalin/tinygrad/commit/fb586be5b717c002cdb1e7a54bfd3e6106ab7bd0
you can change `master` to any commit hash you want.

For now by manually editing the workflow file, if this becomes common we can auto parse it from the description, eg. `replay-base: 276cfd58c74d7f8432b5f1ff39d33ff85df28133`

geohot:
Like the idea, reopen if relevant

PR #11677, opened 2025-08-15
fix: gracefully handle NVIDIA runtime unavailability in speed tests

ManiBAJPAI22:


nimlgen:
Needs a clear pr which is split in logical parts. I don't understand what this pr should solve

ManiBAJPAI22:
@nimlgen 

Thank you for the feedback! 

## Questions for You
1. Is this a valid issue that should be fixed?
2. Should I create a new, more focused PR?
3. What would be the best approach to solve this?

I want to make sure I'm solving the right problem in the right way. Thanks for your guidance!

PR #11691, opened 2025-08-16
[bounty] Fix METAL virtual device synchronization issue + Re-enable "Run LLaMA 7B on 4 (virtual) GPUs"

JoeruCodes:
Implemented shared state management for virtual METAL devices:
  - Virtual devices now share the same Metal command queue, timeline signal, and buffer tracking from the base device
  - Fixed transfer synchronization to avoid using timeline signals between virtual devices
  - Re-enabled the previously commented test in .github/workflows/benchmark.yml

# Changes

  - tinygrad/runtime/ops_metal.py:
    - Added _shared_devices class dictionary to share Metal resources
    - Modified `__init__` to detect virtual devices and reuse resources
    - Fixed `_transfer` to handle virtual device synchronization correctly
    - Made timeline_value a shared reference via property
  - .github/workflows/benchmark.yml:
    - Re-enabled "Run LLaMA 7B on 4 (virtual) GPUs" test (lines 97-98)

# Testing
- LLaMA 7B now runs successfully with --shard 4 on virtual devices
- unit tests in `tests/test_metal.py`


geohot:
Do not use AI

PR #11692, opened 2025-08-16
viz: add a copy button to the codeblock

b1tg:
it was not easy to copy when you want to do manual diff

image


Qazalin:
For diffing, can you just use PORT= and compare two VIZ graphs?

b1tg:
I usually do both the graph comparison and the uops code diff, if you don't want this, I will close this

Qazalin:
Hmm, I don't think diffing UOp strings is a common use-case of VIZ, you're welcome to keep this in your fork though.

PR #11696, opened 2025-08-16
MathTrait replaces shl, shr in decompositions

leopf:


leopf:
duplicate, oops. llvm+ptx shift ops missing

PR #11706, opened 2025-08-18
llama3 eval train

wozeparrot:


examples/mlperf/model_train.py
```diff
@@ -1384,16 +1387,40 @@ def train_step(model, tokens:Tensor, grad_acc:int):
...
-    iter = batch_load_llama3(GBS, SAMPLES, SEQLEN, Path(getenv("BASEDIR", "/raid/datasets/c4/")), seed=SEED, val=bool(TRAIN_ON_VAL))
+  @TinyJit
+  @Tensor.train(False)
+  def eval_step(model, tokens:Tensor):
+    if (DP := getenv("DP", 1)) > 1:
+      device = tuple(f"{Device.DEFAULT}:{i}" for i in range(DP))
+      tokens = tokens.shard(device, 0)
+    if (MP := getenv("MP", 1)) > 1:
+      device = tuple(f"{Device.DEFAULT}:{i}" for i in range(MP))
+      tokens = tokens.shard(device)
```
wozeparrot, line 1398: @chenyuxyz does sharding here not increase mem usage on first device, since it does a copy to first before copying out to other devices?

chenyuxyz: for DP yea slightly, can realize tokens before calling into loss to mitigate

PR #11713, opened 2025-08-18
fix amd llvm bf16 tc

b1tg:


tinygrad/renderer/llvmir.py
```diff
@@ -222,6 +223,12 @@ def __init__(self, arch:str):
     self.arch = arch
     self.tensor_cores = AMDRenderer.get_tensor_cores(arch)
     self.string_rewrite += PatternMatcher([(UPat(Ops.WMMA, name="wmma"), lambda ctx, wmma, arch=arch: render_wmma_amd(ctx, wmma, arch))])
+    if self.arch.split(":")[0] in {"gfx942", "gfx950"}:
```
chenyuxyz: this should be a helper and not copied

b1tg:
on mi300x:
PYTHONPATH=. AMD_LLVM=1 SHOULD_USE_TC=1 BFLOAT16=1 DEBUG=2 python3 extra/gemm/simple_matmul.py fails on master, works after this change

chenyuxyz:
thanks!

b1tg:
this should be fixed too now, the llvm error message is inaccurate

https://discord.com/channels/1068976834382925865/1382831475199705179/1404503981622821067

PR #11722, opened 2025-08-19
run process replay on branch

Qazalin:
There are two differences with normal process replay:

- Compares the capture on the same branch that created it
- Always asserts any diffs/failure to replay, regardless of [pr]

This can help spot diffs that make codegen non reproducible.


Qazalin:
Makes CI slow...

PR #11729, opened 2025-08-19
Multihost exception handling

uuuvn:
Two issues with beam exceptions:

1 - when one host errors `RemoteConnection.batch_submit` throws the error immediately leaving connections to subsequent hosts in a bad state

2 - `RemoteProgram._fini` doesn't know if the program was created successfully. If it isn't `ProgramFree` request will fail and take down the next batch of requests with it. Can't be fixed properly without rewriting the entire protocol, out of all the possible workarounds modifying `ProgramFree` behavior to delete only if exists seems to be the least bad one


uuuvn:
@wozeparrot ?

wozeparrot:
mid training run with this currently

PR #11743, opened 2025-08-20
a*(1/b) -> a/b on LLVM, CPU

ttomsa:
the LLVM and CPU backends emit fdiv instead of recip. In that case you would want to perform the above rewrite but that doesn't happen (to preserve accuracy). However, the rewrite doesn't seem to break the tests so the extra mul is removed.

### DIFF
--- a/tinygrad/renderer/llvmir.py
+++ b/tinygrad/renderer/llvmir.py
@@ -123,6 +123,7 @@ class LLVMRenderer(Renderer):
   has_local = False
   global_max: tuple[int, ...] | None = None
   string_rewrite = base_rewrite + PatternMatcher([(UPat(Ops.WMMA, name="wmma"), render_wmma_amx)])
+  code_for_op = {Ops.FDIV: lambda: None}
   if AMX: tensor_cores = tc.amx
 
   extra_matcher = PatternMatcher([
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -199,12 +199,14 @@ class ClangRenderer(CStyleLanguage):
   # language options
   buffer_suffix = " restrict"
   type_map = {dtypes.bool:"_Bool", dtypes.half:"__fp16"}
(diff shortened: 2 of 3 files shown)


ttomsa:
before:
Screenshot 2025-08-20 011514

after:
Screenshot 2025-08-20 012515

tinygrad/renderer/llvmir.py
```diff
@@ -123,6 +123,7 @@ class LLVMRenderer(Renderer):
   has_local = False
   global_max: tuple[int, ...] | None = None
   string_rewrite = base_rewrite + PatternMatcher([(UPat(Ops.WMMA, name="wmma"), render_wmma_amx)])
+  code_for_op = {Ops.FDIV: lambda: None}
   if AMX: tensor_cores = tc.amx
 
   extra_matcher = PatternMatcher([
```
chenyuxyz, line 129: cleanup line 131

tinygrad/renderer/llvmir.py
```diff
@@ -123,6 +123,7 @@ class LLVMRenderer(Renderer):
   has_local = False
   global_max: tuple[int, ...] | None = None
   string_rewrite = base_rewrite + PatternMatcher([(UPat(Ops.WMMA, name="wmma"), render_wmma_amx)])
+  code_for_op = {Ops.FDIV: lambda: None}
```
chenyuxyz, line 126: slightly annoying that this is used differently, but i think `code_for_op = float_lop` communicates the idea better

tinygrad/renderer/llvmir.py
ttomsa: you mean remove it or use syntactic sugar? That pattern is still necessary

chenyuxyz: oh i meant move that to decomp too unless there's a reason it's bad for CLANG? basically put all FDIV in one place

ttomsa: oh ok got it

ttomsa:
@chenyuxyz it seems adding "float_lops" breaks amdllvm. It's also a bit beyond the scope, you would also want to add unsigned_lop and signed_lop

PR #11754, opened 2025-08-20
non-supported dtype in transcendental

chenyuxyz:
`CPU=1 python3 test/test_dtype_alu.py TestDTypeALU.test_bfloat16_unary` runs now, but diff is very big for large exp


chenyuxyz:
fixed `CPU=1 python3 test/test_dtype_alu.py TestDTypeALU.test_bfloat16_unary` on mac

when we decomp transcendental, dtypes that does not have transcendental is casted into float32, alu, then cast back

PR #11760, opened 2025-08-20
fix gather with inf values

chenyuxyz:
(mask * x) is wrong because 0*inf is nan. i feel we have a lot of those still...


chenyuxyz:
@geohotstan fyi, we should use where instead of mul 0 if value can be inf

geohotstan:
oh yeah there are a few other places that uses mul. I'll go through and fix. 

I remember mul was faster than where. (this was a long time ago)
There was also discussion back then for not supporting inf or nan.

PR #11770, opened 2025-08-21
[TYPED=1] cvar should allow dtype as a tuple

jordan-chalupka:
With this change you can now import tinygrad with TYPED=1 and not get a type error.

Related: #7889


geohot:
Can you add a test to CI with TYPED=1 that this fixes?

.github/workflows/test.yml
```diff
@@ -63,6 +63,15 @@ jobs:
...
         pip install mypy
         mypy -c "from tinygrad.tensor import Tensor; print(Tensor([1,2,3,4,5]))"
+    - name: Use as an external package (with TYPED=1)
+      run: |
+        mkdir $GITHUB_WORKSPACE/test_dir
+        cd $GITHUB_WORKSPACE/test_dir
+        python -m venv venv_typed
+        source venv_typed/bin/activate
+        pip install $GITHUB_WORKSPACE[typed]
+        TYPED=1 python -c "from tinygrad.tensor import Tensor; print(Tensor([1,2,3,4,5]))"
```
jordan-chalupka, line 73: this failed previously

.github/workflows/test.yml
```diff
@@ -63,6 +63,15 @@ jobs:
...
         mypy -c "from tinygrad.tensor import Tensor; print(Tensor([1,2,3,4,5]))"
+    - name: Use as an external package (with TYPED=1)
+      run: |
+        mkdir $GITHUB_WORKSPACE/test_dir
+        cd $GITHUB_WORKSPACE/test_dir
+        python -m venv venv_typed
+        source venv_typed/bin/activate
+        pip install $GITHUB_WORKSPACE[typed]
+        TYPED=1 python -c "import tinygrad"
+        TYPED=1 mypy -c "import tinygrad"
```
jordan-chalupka: next step will be to have types working for everything in docs/ramp.py

.github/workflows/test.yml
```diff
@@ -63,6 +63,16 @@ jobs:
         python -c "from tinygrad.tensor import Tensor; print(Tensor([1,2,3,4,5]))"
         pip install mypy
         mypy -c "from tinygrad.tensor import Tensor; print(Tensor([1,2,3,4,5]))"
+    - name: Use as an external package (with TYPED=1)
```
geohot: Why is this tested as an external package? Why not just normal TYPED=1?

setup.py
```diff
@@ -104,5 +104,8 @@
             "black",
             "numpy",
         ],
+        'typed': [
```
geohot: This shouldn't be a new category

iashyam:
I just found out that you can comment on any issue. Pretty cool!

.github/workflows/test.yml
jordan-chalupka: right, it's overkill.  Added a simple test to the linter job.

Issue #11763, opened 2025-08-21
Remove using reshape to register symbolic shape


chenyuxyz:
using reshape to register a symbolic shape fails because the input can have a smaller buffer compared to all possible symbolic shapes, which causes z3 index check to fail due to potential OOB access.

update this logic and update the symbolic tests
https://github.com/tinygrad/tinygrad/blob/9f94c25a254efce7161a56254d7f82cd7b05b764/tinygrad/shape/view.py#L321-L331

NOTE: `create_kernel` also has reshaping to symbolic shape

PR #11771, opened 2025-08-21
[bounty] Remove using reshape to register symbolic shape

ben-waldron-1:
Fixes https://github.com/tinygrad/tinygrad/issues/11763

Reshaping all_int -> symbolic now asserts. For the couple places where a buffer is created and reshaped to an existing uop's shape (kernelize, Tensor.empty, rangeify), I added a shrink.

~Still working my way through fixing tests but thought I'd open before fully green just to track progress.~
~Now working on cleanups.~
Ready to review.

After this change, to introduce symbolic shape, you can only shrink, pad, or expand (In tests I use `[:vi]` to replace `.reshape(vi)`)


test/unit/test_shapetracker.py
```diff
@@ -841,10 +841,17 @@ def test_render(self):
...
+  def test_double_reshape(self):
+    st = ShapeTracker.from_shape((10,))
+    i = Variable("i", 1, 10)
+    st = st.shrink(((0, i),))
+    st = st.reshape((i,)).reshape((1, i))
+    assert len(st.views) == 1
+
   def test_reshape_stride_0(self):
     st = ShapeTracker.from_shape((3,), (0,))
     st = st.reshape((Variable("i", 1, 10).bind(3),))
```
ben-waldron-1: `test_reshape_stride_0` currently fails but the behaviour isn't supported with non-symbolic, and it should probably be an expand rather than a reshape. Delete test? Symbolic expand is tested in test_symbolic_ops

chenyuxyz: Do whatever that makes the most sense to you. Will lock and review after all tests are green

ben-waldron-1:
More tests green, but still some left:
- Some unit tests
- `extra/optimization/test_beam_search.py`
- `test_symbolic_jit`
- make sure rangeify works

So far functionality seems solid, and the tests just need outdated behaviour replaced -- will get to them early next week.

tinygrad/shape/view.py
```diff
@@ -312,24 +312,19 @@ def reshape(self, new_shape: tuple[sint, ...]) -> View|None:
     if not all(x >= 0 for x in new_shape): raise ValueError(f"shape can't contain negative numbers {new_shape}")
     # check for the same size
     if (self_all_int := all_int(self.shape)):
-      assert all(isinstance(s, (int, UOp)) for s in new_shape), f"{self.shape=} -> {new_shape=} contains non (int, Variable) dim"
-      if resolve(prod(self.shape) != prod(new_shape), False): raise ValueError(f"size mismatched, can't reshape {self.shape=} -> {new_shape=}")
+      # reshapes cannot introduce symbolic shape
+      assert all(isinstance(s, int) for s in new_shape), f"{self.shape=} -> {new_shape=} contains non int dims"
```
chenyuxyz: all_int

chenyuxyz (commented):
looks decent. locked the bounty to you and will review once all tests pass

ben-waldron-1:
- Added a shrink to Tensor.empty if given symbolic shape (and added a test for Tensor.empty in `test_tensor`)
- Skipped a lot of the tests for reshaping into symbolic in `test_symbolic_shapetracker.py`
- Removed two tests for ShapeTrackerAddVariable because adding relied on reshaping into symbolic being contiguous, whereas shrinking isn't (see temp comment in diff)

Still todo:
- Rangeify likely needs a change similar to kernelize (I'll have to learn how it works first though lol)
- Separate out the major test changes into a different PR
- `test_symbolic_jit`
- ~I forgot to commit my fix for `extra/optimization/test_beam_search.py`~
- Double check error messages and comments, as per feedback from meeting

test/test_symbolic_ops.py
```diff
@@ -1,62 +1,55 @@
...
-
   def test_plus1(self):
     def f(a): return (a+1).realize()
+    a = Tensor.rand(3, 10)
     for i in range(1, 5):
       vi = Variable("i", 1, 10).bind(i)
-      a = Tensor.rand(3, i)
-      symbolic = f(a.reshape(3, vi)).reshape(3, i).numpy()
-      expected = f(a).numpy()
+      # Pad to max size then shrink down to remove symbolic shape
```
chenyuxyz: just reshape from symbolic shape to int shape?

ben-waldron-1: I added an extra check to reshape for potentially reshaping to a shape larger than the current symbolic one could be (that's the new ValueError)
Currently, if there actually is an OOB access, it gets caught by z3, but the check was neat enough I thought to include it (`resolve(prod(self.shape) < prod(new_shape), True)` is True if there's a chance the new shape could lead to OOB for certain values of the variable)

Doing this pad then shrink will always be correct, for all values of the variable, while that isn't guaranteed for certain reshapes to int.

However, I'm happy to be pragmatic, remove that new error and .pad.shrink syntax, and let z3 catch everything, if that's what you want

chenyuxyz: it might be a good idea, but it's better to have a separate pr just for that. changing many things in one pr is hard to review and hard to revert in cast anything goes wrong

test/unit/test_shapetracker.py
```diff
@@ -840,16 +840,27 @@ def test_render(self):
     self.assertEqual(valid.render(), "(ridx0<2)")
 
 class TestVariableReshape(unittest.TestCase):
+  @unittest.skip("reshaping into symbolic no longer supported")
```
chenyuxyz: instead of skipping, update to use shrink. if shrink does not make sense, delete the test

ben-waldron-1:
Tests passing (CI red only because MacOS env setup failed), so starting cleanup and will undraft when ready.

test/unit/test_symbolic_shapetracker.py
```diff
@@ -220,9 +169,10 @@ def test_expand_into_symbols(self):
     assert a.shape == (3, vi, vj)
 
   def test_plus_expands_constant(self):
+    a = Tensor.rand(3, 5)
     for i in range(1, 6):
       vi = Variable("i", 1, 5).bind(i)
-      a = Tensor.rand(3, i).reshape(3, vi)
+      a = a[:, :vi]
```
chenyuxyz: this reassigns `a`

test/unit/test_shapetracker_math.py
```diff
@@ -87,20 +87,6 @@ def test_off_by_one(self):
...
-    out = x + x
-    assert out == x
-
-  def test_self_add_reshape(self):
-    j = Variable("j", 0, 20).bind(10)
-    a = ShapeTracker.from_shape((10,10))
-    x = a.reshape((10, j))
-    out = x.reshape((5, 2, j)) + x
-    assert out == x
-
```
ben-waldron-1, line 103: These tests were deleted because we're lacking the necessary logic in `View.__add__` to support shrinking into symbolic rather than reshaping in.
Reshaping creates contiguous views, so `__add__` just directly replaces with the new shape. However, shrinking creates non-contiguous views, and there's currently very little support for merging non-contiguous symbolic views: 
https://github.com/tinygrad/tinygrad/blob/60dd9a162c92ce1e2e8db34c2367b2f7bc3eebde/tinygrad/shape/view.py#L188-L192

test/unit/test_symbolic_shapetracker.py
```diff
@@ -242,10 +197,10 @@ def test_shrink_symbols(self):
...
   def test_pad(self):
     v = Variable("v", 1, 100).bind(5)
-    t = Tensor.ones(5).reshape(v).pad(((4, 0),)).reshape(9)
-    assert t.shape == (9,)
-    st = t.uop.st
-    print(st)
+    t = Tensor.ones(100)[:v].pad(((4, 0),))
+    t = t.reshape(9)
+    assert t.tolist() == [0,0,0,0,1,1,1,1,1]
+
```
ben-waldron-1, line 203: Test changed significantly because I don't think the current one actually tests for anything?

chenyuxyz (commented):
mostly looks good! please carefully read the diff before marking it as ready

test/test_symbolic_ops.py
```diff
@@ -201,75 +208,75 @@ def test_slice_var_shape(self):
...
+      expected = a[:i].bitcast(dtypes.uint64).numpy()
+      symbolic = a[:vi].bitcast(dtypes.uint64).reshape(expected.shape).numpy()
       np.testing.assert_allclose(symbolic, expected, atol=1e-6, rtol=0)
 
   @unittest.expectedFailure
   def test_conv2d_ceildiv_edge_case(self):
     v = Variable('v', 11, 50_000)
     val = 39601
-    x = Tensor.randn(1, 22, 39601).reshape(1, 22, v.bind(val))
+    x = Tensor.randn(1, 22, 50_000)[:, :, :v.bind(val)]
```
ben-waldron-1, line 279: I checked and this still fails the same way as it does on master

test/unit/test_symbolic_shapetracker.py
```diff
@@ -109,60 +109,44 @@ def test_view_unbind(self):
...
 
   def test_two_symbol_reshape(self):
+    t = Tensor.rand(5, 5)
     for i in range(1, 6):
       for j in range(1, 6):
         vi = Variable("i", 1, 5).bind(i)
         vj = Variable("j", 1, 5).bind(j)
-        t = Tensor.rand(i, j).reshape(vi, vj)
-        assert t.shape == (vi, vj)
-        # NOTE: this is currently not allowed
```
ben-waldron-1, line 161: This works now so I uncommented

test/unit/test_symbolic_shapetracker.py
ben-waldron-1: Whoops I may have done that elsewhere

EDIT: definitely have

ben-waldron-1:
Should be ready for review now.

Regarding the change to rangeify:
- I added the same shrink to `map_contiguous` as I did to `create_kernel` in kernelize, which was enough to pass CI
- Given other symbolic tests currently fail on master with rangeify (`RANGEIFY=1 python test/test_symbolic_ops.py`), I decided not to find the other places where shrinks should be added, since it's evidently not finalised yet
- However, it's likely that other shrinks should be added (I'm guessing `bufferize_to_store` and maaaaybe `map_partial_contiguous`, or maybe a shrink only in `bufferize_to_store`?).

I also don't yet fully understand rangeify, so this was partly guesswork

chenyuxyz:
congrats! bounty is yours. reach out to [george@tinygrad.org](mailto:george@tinygrad.org) to claim bounty, can pay via paypal or USDC

PR #11773, opened 2025-08-21
exclude .git from ruff

jordan-chalupka:
ruff shouldn't look at .git
image

repro by having any file in .git with an unfortunately named extension e.g. `.git/logs/refs/remotes/origin/TYPED-equals-1-working-in-tensor.py` and ruff tries to lint it.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -35,6 +35,7 @@ lint.select = [
 line-length = 150
 
 exclude = [
+  ".git/",
   "docs/",
   "extra/",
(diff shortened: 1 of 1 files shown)


geohot:
Can't repro issue
```
light@blak:~/tinygrad$ python3 -m ruff check .
All checks passed!
```

jordan-chalupka:
@geohot sorry, added instructions to repro. Bit of a silly one but I don't see value in having ruff look in the .git dir

Issue #11626, opened 2025-08-12
x.abs()**2 is x**2


chenyuxyz:
```
from tinygrad import Tensor
print((Tensor([11]).abs()**2).tolist())
```
run with `NOOPT=1 DEBUG=4 CPU=1 python test.py`
```
void E_(int* restrict data0, int* restrict data1) {
  int val0 = (*(data1+0));
  int alu0 = ((val0<0)?-1:1);
  int alu1 = (((_Bool)(val0))?alu0:0);
  int alu2 = (val0*alu1);
  *(data0+0) = (alu2*alu2);
}
```
can simplify to `*(data0+0) = (val0*val0);`

ayushm2003:
can i take it up?

PR #11779, opened 2025-08-22
abs(x) ** 2 is x**2

JoeruCodes:
closes https://github.com/tinygrad/tinygrad/issues/11626

code (run w `NOOPT=1 DEBUG=4 CPU=1`):
```
from tinygrad import Tensor
(Tensor([11]).abs()**2).realize()
```

before:
```
void E_(int* restrict data0, int* restrict data1) {
  int val0 = (*(data1+0));
  int alu0 = ((val0<0)?-1:1);
  int alu1 = (((_Bool)(val0))?alu0:0);
  int alu2 = (val0*alu1);
  *(data0+0) = (alu2*alu2);
}
```

after:
```
void E_(int* restrict data0_1, int* restrict data1_1) {
  int val0 = (*(data1_1+0));
  *(data0_1+0) = (val0*val0);
}
```

### DIFF
--- a/tinygrad/uop/symbolic.py
+++ b/tinygrad/uop/symbolic.py
@@ -352,6 +352,8 @@ symbolic_flat = symbolic+PatternMatcher([
   (-1 * (UPat.var("x") + UPat.var("y")), lambda x,y: (-x)+(-y)),  # -(x+y) -> -x + -y
   # (x+y)*c -> x*c+y*c. only for int, float has inf*0=nan issue
   ((UPat.var("x", dtypes.ints) + UPat.var("y")) * UPat.cvar("c"), lambda x,y,c: x*c+y*c),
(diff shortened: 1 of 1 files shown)


chenyuxyz:
it's wrong if y is 0, also what do you do if it's x**20

uwuxyz:
> it's wrong if y is 0, also what do you do if it's x**20

Hmm yeah, also it would require another pattern for (x*y) * (x * y) too

welp wheres it is lmao

uwuxyz:
Handling the 0 case with `where` brings us back to square one... I think instead of calculating `abs` with `sign` i think if we rewrite abs as `abs(x) = max(x, -x)` we could simplify it better maybe? @chenyuxyz

chenyuxyz:
please think before ask

PR #11792, opened 2025-08-23
viz bytepack format

Qazalin:
Training a 1B llama yields ~20M profiler events:

With JSON serialization, the browser has to load 6GB to memory. This OOMs since each tab is limited to <3-4GB memory usage. Using a packed format, we only need ~600MB.

**Design decisions:**

- Timestamps are in microseconds relative to start time. They're stored in u32, which can express up to ~1 hr of trace events.
- Strings (kernel names, metadata, etc) are deduped.
- Buffer sizes are in u64

More optimization possible:

- The string lookup is a JSON dumped array, we can compress this.
- Can store less for memory by moving the layout to client. (update: done in #11830)

**Results**

|  | Events | JSON | bytepack |
|----------------|---------|-------------|-------------|
| 8 GPU llama 1B train (command:`[1]`) | 24M | 5.8GB | 640MB |
| examples/beautiful_mnist.py | 16.6K | 3.7MB | 770KB |
| examples/gpt2.py | 55K | 12.54MB | 1.40MB |

`[1]`: `VIZ=1 FAKEDATA=1 OFFLOAD_OPTIM=1 DP=8 BS=8 GRADIENT_ACC_STEPS=2 BLOCK_REORDER=0 LR=3e-4 TRAIN_ON_VAL=1 DEFAULT_FLOAT=bfloat16 OPTIM_DTYPE=bfloat16 LLAMA3_SIZE=1B WARMUP_STEPS=36 DECAY_STEPS=360 SEQLEN=8192 PYTHONPATH=. AMD=1 AMD_LLVM=0 MODEL=llama3 python3 examples/mlperf/model_train.py`

### DIFF
--- a/test/unit/test_viz.py
+++ b/test/unit/test_viz.py
@@ -354,7 +354,16 @@ class TestVizProfiler(unittest.TestCase):
     n_events = 1_000
     prof = [ProfileRangeEvent("CPU", name="k_test", st=decimal.Decimal(ts:=i*step), en=decimal.Decimal(ts)+step) for i in range(n_events)]
     sz = len(get_profile(prof))
-    self.assertLessEqual(sz/n_events, 100)
+    self.assertLessEqual(sz/n_events, 27)
+
+  # can pack up to 1hr 11 min of trace events
+  def test_trace_duration(self):
+    dur_mins = 72
+    n_events = 1_000
+    step = decimal.Decimal(dur_mins*60*1e6//n_events)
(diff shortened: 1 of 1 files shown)


geohot:
Looks good, seems like it's around a 10x reduction (~27 bytes per event), and ArrayBuffer should be faster too.

We should have a stress test in CI which tests the maximum number of events we expect to reasonably work.

PR #11843, opened 2025-08-26
fix nan in float_to_bf16

b1tg:


test/unit/test_dtype_spec.py
```diff
@@ -110,15 +110,15 @@ def test_truncate_fp16(self):
 
   def test_truncate_bf16(self):
```
chenyuxyz: oh cool, i also had something similar and was adding tests, can merge your version first.

planned to add more tests here, especially round to even and max finite bf16 + 1/2 ulp etc

chenyuxyz:
i merged part of this, you can update to include the nan fix

b1tg:
I have two more tests. I guess I'll use a new pr later

PR #11847, opened 2025-08-26
viz: add UOp ports to edges

Qazalin:
This looks like a successor to the edge count tags in master.
image


Qazalin:
image
image
I want one descriptive tag for each edge. Ports.length > 1 don't really fit in the UI and it's unclear if enumerating all ports is useful. The primary use-case for this feature is enumerating ALU operands.

I think this is the path forward:
image

PR #11849, opened 2025-08-26
test float_to_bf16 round-to-even behavior

b1tg:


chenyuxyz:
btw if you are interested, $150 bounty to support bfloat16 in PYTHON, once you fixed the rounding, the missing part is storage and bitcast. Storing in float fmt is fine, and the code should work with float8 support in mind

PR #11864, opened 2025-08-26
add background overlay for colored labels

Qazalin:
This allows for better contrast for white axis colors on light backgrounds:
image


Qazalin:
we made the ansi colors a little darker instead.

PR #11866, opened 2025-08-26
viz: place edge tags on the parent node

Qazalin:
image


geohot:
Put on both?

Qazalin:
I don't think we want both, it complicates the UI.
Really prefer this start version over the end one in master. It localizes source+index in one place.

PR #11870, opened 2025-08-27
removed realize() from __setitem__ and fused TestSetitemLop.test_arange into one kernel

kaushal07wick:


chenyuxyz:
test failing, and it needs a lot more than just removing the realize call

PR #11886, opened 2025-08-28
viz: clicking on a selected UOp de selects it

Qazalin:


Qazalin:
This requires tracking more state, outside click still works for deselct.

PR #11889, opened 2025-08-28
Add test_variable_empty

ben-waldron-1:
While this behaviour is tested currently on master, it's only in test_tensor_uop_representation, which I've pasted below:
```python
  def test_empty_buf(self):
    a = Tensor.empty(3, 3)
    is_pattern(a, UPat(Ops.RESHAPE, src=(UPat(Ops.BUFFER),)))
    vi = UOp.variable("i", 1, 3).bind(1)
    a = Tensor.empty(3, vi)
    is_pattern(a, UPat(Ops.RESHAPE, src=(UPat(Ops.BUFFER),)))
    self.assertEqual(a.uop.base.buffer.size, 9)
```

Might be better to have a more atomic test alongside others in `test_tensor`, including testing bound vs unbound variables.

While working on the remove symbolic reshape bounty, I broke this and didn't realise until I checked test_tensor_uop_representation.

### DIFF
--- a/test/external/external_test_onnx_backend.py
+++ b/test/external/external_test_onnx_backend.py
@@ -134,7 +134,6 @@ backend_test.exclude('test_simple_rnn_*')
 
 # no control flow
 # control flow uses AttributeProto.GRAPH
-backend_test.exclude('test_if_*')
 backend_test.exclude('test_loop*')
 backend_test.exclude('test_range_float_type_positive_delta_expanded_cpu') # requires loop
 backend_test.exclude('test_affine_grid_2d_align_corners_expanded_cpu')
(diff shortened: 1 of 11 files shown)


test/test_tensor.py
```diff
@@ -555,6 +555,16 @@ def test_shrink(self):
     t = Tensor.arange(32).contiguous().realize()
     self.assertListEqual(t[16:20].tolist(), [16,17,18,19])
 
+  def test_variable_empty(self):
+    v = Variable("i", 1, 10)
+    t = Tensor.empty(3, v)
```
chenyuxyz: this should assert, so at least add a TODO to show that it's not a desired behavior

chenyuxyz (commented):
we don't have a unit test file for this, but putting in test/test_tensor_variable.py suits better

ben-waldron-1:
MacOS (unit) failing in setup env, but apart from that all good.

PR #11893, opened 2025-08-28
Remove Ops.ASSIGN

quortus:
Depends on #11159


geohot:
See replace ASSIGN with STORE+AFTER

PR #11894, opened 2025-08-28
update gpuocelot brew config to not attempt reinstall

chenyuxyz:
fix
```
but you are trying to install it from the homebrew/core tap.
Formulae with the same name from different taps cannot be installed at the same time.
```


chenyuxyz:
hmm not quite right

PR #11898, opened 2025-08-28
viz: sum all buffers in zoomed out memory graph

Qazalin:
The memory graph is the main bottleneck in a lot of the large VIZ profiler traces because it's trying to draw every single allocation, and whenever there is a dellocation, every alive Buffer's graph has to update its path geometry.
This diff switches the zoomed out memory graph to a single path for total memory usage, which is what we want most of the time. It is still possible to view the per-buffer detailed view and it's one click away.

Default:
image
Per buffer details toggles by clicking on "METAL Memory"
image

This is faster because fundamentally we're just drawing one `sum(nbytes_allocated)` path, instead of N paths * N alloc/dealloc event path changes.

### DIFF
--- a/tinygrad/viz/js/index.js
+++ b/tinygrad/viz/js/index.js
@@ -137,7 +137,7 @@ const formatUnit = (d, unit="") => d3.format(".3~s")(d)+unit;
 
 const colorScheme = {TINY:["#1b5745", "#354f52", "#354f52", "#1d2e62", "#63b0cd"],
   DEFAULT:["#2b2e39", "#2c2f3a", "#31343f", "#323544", "#2d303a", "#2e313c", "#343746", "#353847", "#3c4050", "#404459", "#444862", "#4a4e65"],
-  BUFFER:["#3A57B7","#5066C1","#6277CD","#7488D8","#8A9BE3","#A3B4F2"],
+  BUFFER:["#342483", "#3E2E94", "#4938A4", "#5442B4", "#5E4CC2", "#674FCA"],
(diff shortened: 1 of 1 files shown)


Qazalin:
Nsight systems creates a spectrogram graph at the high level, when you zoom in it becomes a solid, perfetto-like block.

https://github.com/user-attachments/assets/cee95be0-dd18-4de5-be9e-0f734de2b6c0

They document this:
https://docs.nvidia.com/nsight-systems/UserGuide/index.html 
> Nsight Systems calculates the average occupancy for the period of time represented by particular pixel width of screen. It then uses that average to set the top of the colored section. So, for instance, if 25% of that timeslice the kernel is active, the bar goes 25% of the distance to the top of the row.


Perfetto just merges the rects, which doesn't generalize to the memory graph. In nvidia's tools, the nature of the visualization changes as you zoom in.


This is a 15M event visualization of gpt-oss inference
image
zoomed in, it's like current viz:
image

Qazalin:
I'm seriously considering the grid layout now - It'll be useful for visualizing resource usage.

Qazalin:
color by size:
image
color by lifetime:
Screenshot 2025-09-13 at 4 27 35 PM

Qazalin:
min width!
image
vs master
image
in a 10 second beautiful mnist this is all the rects in metal that are 0.5> width when zoomed out
image

currently it's an O(n) pass, it can be a gridmap:

Screenshot 2025-09-14 at 6 49 53 PM
image

PR #11904, opened 2025-08-28
add Ops.INVALID

S-Lykles:


geohot:
Could it just be CONST with arg None on dtypes.index?

Issue #11929, opened 2025-08-31
bf16 PYTHON alu discrepancy


chenyuxyz:
updating test_dtype_alu.py to fuzz bfloat16 input with a correct bfloat16 space (it's sampling from float16 now), found this diff

```
from tinygrad import Tensor, dtypes
from tinygrad.runtime.ops_python import from_storage_scalar
a = from_storage_scalar(20510, dtypes.bfloat16)
print(f"{a=}")
t = Tensor([a], dtype=dtypes.bfloat16)
print(t.tolist())
print(t.sqrt().float().tolist())
```

with `METAL=1`
```
a=10603200512.0
[10603200512.0]
[102912.0]
```

with `PYTHON=1`
```
a=10603200512.0
[10603200512.0]
[100864.0]
```

torch
```
import torch
print(torch.tensor([10603200512.0], dtype=torch.bfloat16).sqrt())
```
```
tensor([102912.], dtype=torch.bfloat16)
```

@b1tg fyi

PR #11932, opened 2025-08-31
fix transcendentals in python renderer

b1tg:


b1tg:
should fix https://github.com/tinygrad/tinygrad/issues/11929

test/test_dtype.py
```diff
@@ -428,5 +428,12 @@ def test_cast(self):
     data = [60000.0, 70000.0, 80000.0]
     np.testing.assert_allclose(Tensor(data).cast("bfloat16").numpy(), torch.tensor(data).type(torch.bfloat16).float().numpy())
 
+  # TODO: AMD_LLVM failed on this
+  @unittest.skipUnless(Device.DEFAULT == "PYTHON", "only test on PYTHON now")
```
Qazalin, line 432: Can you just disable it for AMD_LLVM?
Confirmed it fails on the real device and the emulator.

b1tg: This test currently only runs on PYTHON

chenyuxyz: i think she's trying to say with AMD_LLVM=0 the test also passes, so can disable just for AMD_LLVM

chenyuxyz: anyway it can be added later and does not need to block this

PR #11937, opened 2025-08-31
pre matcher after decomp

ttomsa:
The pre matcher legalizes vector dtypes in the x86/arm64 backends. Transcendental can create vectors that are too large like `dtypes.ulong.vec(4)`. The pre matcher needs to run after decompositions but before the final rewrite.

relevant for #11611


geohot:
This misses the point of the *pre* matcher

PR #11938, opened 2025-08-31
add mask dtype

ttomsa:
In the x86/arm64 backends certain comparisons result in masks instead of booleans. These instructions are necessary for full SIMD support. Masks don't really exist outside this context and are handled by a pattern matcher.

relevant for #11611

### DIFF
--- a/test/test_dtype.py
+++ b/test/test_dtype.py
@@ -111,7 +111,7 @@ class TestDType(unittest.TestCase):
     fields = dtypes.fields()
     self.assertIn("float", fields)
     self.assertIn("float32", fields)
-    self.assertEqual(len(fields), 26)
+    self.assertEqual(len(fields), 30)
     self.assertTrue(all(isinstance(value, DType) for value in fields.values()))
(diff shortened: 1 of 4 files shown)


geohot:
Isn't this a vectorized boolean?

ttomsa:
@geohot you can't treat masks as vectorized booleans. First, they have different sizes so if the vector splitter sees a `dtypes.bool.vec(16)` it will think it's within the max vector length but if the boolean comes from i.e. a vector float32 comparison, the size of each element is actually 4 bytes so the vector exceeds the max length. Masks also have different values, a True mask is all 1s not just a single 1 so if you want to cast from boolean to float you actually need to negate the true mask so that -1 (all 1s == -1 for sints) becomes 1 and then you can cast. If masks are booleans how do you know whether you need to negate or not when casting to another dtype? You would need to walk the srcs until you find a comparison that would produce a mask.
Also scalar float comparisons generate masks not just vector dtypes.

PR #11944, opened 2025-09-01
fix vec dtype _min_max

S-Lykles:
use scalar itemsize, this was giving very high min/max


S-Lykles:
image

This is what was causing this

PR #11960, opened 2025-09-02
python float8 support

b1tg:


tinygrad/dtype.py
```diff
@@ -230,6 +233,8 @@ def float_to_bf16(x):
 # fp8-float conversions based on https://gitlab.com/nvidia/headers/cuda-individual/cudart/-/blob/main/cuda_fp8.hpp
 def float_to_fp8(x: float, dtype: DType) -> int:
   assert dtype in dtypes.fp8s, "Only for fp8s"
+  if dtype == dtypes.fp8e4m3 and not math.isfinite(x): return 0x7f if math.copysign(1, x) > 0 else 0xff
```
chenyuxyz: why `math.copysign(1, x) > 0` and not `x > 0`? if it's needed consider a comment

test/test_dtype.py
```diff
@@ -344,9 +352,11 @@ def test_uint64_load(self):
 
 class TestBoolDType(TestDType): DTYPE = dtypes.bool
 
-@unittest.skipUnless(is_dtype_supported(dtypes.bfloat16), f"no bfloat16 on {Device.DEFAULT}")
```
chenyuxyz: why remove, does it work now? this could be its own pr

tinygrad/dtype.py
```diff
@@ -187,6 +187,9 @@ def _get_recursive_parents(dtype:DType) -> set[DType]:
   return set.union(*[_get_recursive_parents(d) for d in promo_lattice[dtype]], {dtype}) if dtype != dtypes.float64 else {dtypes.float64}
 @functools.cache
 def least_upper_dtype(*ds:DType) -> DType:
+  from tinygrad.device import is_dtype_supported
+  if is_dtype_supported(dtypes.fp8e4m3) or is_dtype_supported(dtypes.fp8e5m2):
```
chenyuxyz: does this work if a device only support one of these?

test/unit/test_dtype_spec.py
```diff
@@ -439,6 +445,8 @@ def test_sum(self):
     assert (Tensor([0, 1], dtype=dtypes.uint16)).sum().dtype == dtypes.uint32
     assert (Tensor([0, 1], dtype=dtypes.uint32)).sum().dtype == dtypes.uint32
     assert (Tensor([0, 1], dtype=dtypes.uint64)).sum().dtype == dtypes.uint64
+    assert (Tensor([0, 1], dtype=dtypes.fp8e4m3)).sum().dtype == dtypes.fp8e4m3
```
chenyuxyz: if these already works, add tests as a separate pr

if it needs some change to work, i feel the dtype change can be its own pr

tinygrad/dtype.py
b1tg: +/- math.nan > 0 both give False, I will add comments

PR #11961, opened 2025-09-02
more llvm intrinsics

b1tg:


test/test_dtype.py
```diff
@@ -433,5 +434,11 @@ def test_no_approximation(self):
     expected = torch.tensor(data, dtype=torch.bfloat16).sqrt().float().numpy()
     np.testing.assert_allclose(Tensor(data, dtype=dtypes.bfloat16).sqrt().numpy(), expected)
 
+  def test_log_nan(self):
+    a = 32769
+    data = [from_storage_scalar(a, dtypes.bfloat16)]
+    expected = torch.tensor(data, dtype=torch.bfloat16).log().float().numpy() # nan
+    np.testing.assert_allclose(Tensor(data, dtype=dtypes.bfloat16).log().numpy(), expected)
```
chenyuxyz: instead of a comment that says it's nan, just also test that output is nan

b1tg:
just found metal failing on this test

log2(x) is undefined on METAL when x<0

https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf p311

test/test_dtype.py
```diff
@@ -433,5 +434,13 @@ def test_no_approximation(self):
     expected = torch.tensor(data, dtype=torch.bfloat16).sqrt().float().numpy()
     np.testing.assert_allclose(Tensor(data, dtype=dtypes.bfloat16).sqrt().numpy(), expected)
 
+  @unittest.skipIf(Device.DEFAULT == "METAL", "log2(x) is undefined on METAL when x<0")
```
chenyuxyz: is this a defined behavior?

if you add `.to("mps")` in torch, it also generates `-inf`, so i think we match torch, and it's not clear what this test is testing

having test_dtype_alu passed is good enough for correctness, but is there more benefit to use llvm intrinsic? like if it's more accurate or faster

b1tg: metal document say it's undefined: https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf p311

the reason I doing this is that test_bfloat16_unary failed in another pr: https://github.com/tinygrad/tinygrad/actions/runs/17403776365/job/49402809541

using llvm intrinsics is more accurate and simple (less llvm ir code and assembly code), not sure it's faster or not

chenyuxyz: i think a better way to handle this would be rejecting the fuzz input for log if it's < 0. unless there an universal definition that says this is nan, we should not have a test that enforce the output

chenyuxyz:
conflict

PR #11963, opened 2025-09-02
assert embedding input is int

S-Lykles:


tinygrad/nn/__init__.py
```diff
@@ -320,6 +320,7 @@ def __init__(self, vocab_size:int, embed_size:int):
 
   def __call__(self, idx:Tensor) -> Tensor:
     if not hasattr(self, 'arange'): self.arange = Tensor.arange(self.vocab_sz, requires_grad=False, device=self.weight.device).unsqueeze(-1)
+    if not dtypes.is_int(idx.dtype): idx = idx.cast(dtypes.default_int)
```
chenyuxyz: this probably should raise an error? i don't think user would want to pass in a float and rely on the cast

PR #12017, opened 2025-09-05
Divmod combine - associative variations

S-Lykles:


S-Lykles:
few of these
```
 -  int alu5 = ((gidx0<<4)+alu1+(alu0<<1)+(lidx1<<2)+(lidx2<<3));
+  int alu5 = (lidx0+(gidx0<<4)+(lidx1<<2)+(lidx2<<3));
```

PR #12042, opened 2025-09-05
assert benchmark times

S-Lykles:


test/external/external_benchmark_openpilot.py
```diff
@@ -35,7 +36,12 @@
       inputs = {**{k:v for k,v in new_inputs_junk.items() if 'img' in k},
                 **{k:Tensor(v) for k,v in new_inputs_junk_numpy.items() if 'img' not in k}}
       ret = next(iter(run_onnx_jit(**inputs).values())).cast(dtypes.float32).numpy()
-    print(f"jitted:  {(time.perf_counter_ns() - st)*1e-6:7.4f} ms")
+      jitted_times.append(t:=(time.perf_counter_ns() - st)*1e-6)
+    print(f"jitted:  {t:7.4f} ms")
+
+  if (assert_time:=getenv("ASSERT_JITTED_TIME")):
+    avg = sum(jitted_times[-10:])/10
```
geohot: I would do the min here. Random bad ones you don't case about

PR #12049, opened 2025-09-06
only run autogen tests on change

jordan-chalupka:
gh allows you to specify when to trigger a workflow using the `on` syntax, I've set it so we only run the autogen tests when those files change.

image


geohot:
Nice, thanks!

PR #12051, opened 2025-09-06
always ignore valid in real_strides

chenyuxyz:


chenyuxyz:
probably not worth it

PR #12052, opened 2025-09-06
Fix gpt2 test

jordan-chalupka:


geohot:
We should probably just delete that test

PR #12053, opened 2025-09-06
only run hcopts once

geohot:
hmm, this should be a NOOP, but it isn't


dataguide-con:
could it be that before hcopts was mutating k and then postrange reapplied the same opts, so they ran twice? wondering if this just removes that replay

PR #12061, opened 2025-09-06
Implement Top-K sampling in Mamba example

Steven-Yiran:
This change adds Top-K sampling, resolving the # TODO: topk comment.

A new --top_k argument can be used with --sample flag, it constrains the token selection to the k most probable candidates.

The new behaviors with `k=5` and `k=50`:
Screenshot 2025-09-06 at 3 50 32 PM
Screenshot 2025-09-06 at 3 50 43 PM

First-time contributor here. Thanks for the guidance in advance!


chenyuxyz:
we don't really maintain that example, but this seems fine

PR #12068, opened 2025-09-08
inline conditional in broadcast and cast

ridoy:
inlining the two conditions here saves 2 lines and still readable


chenyuxyz:
No code golf!

PR #12076, opened 2025-09-08
Move TC tests into `unit/test_emulated_tc.py`

eitanturok:


geohot:
This is not at all what I meant. Remove all the crap from the yml file!

PR #12080, opened 2025-09-09
fix vec dtype in promo lattice

ttomsa:
fixes `PYTHONPATH=. CPU=1 DEVECTORIZE=0 python3 test/test_ops.py TestOps.test_max_unpool2d` resulting in `KeyError: dtypes.int.vec(25)`


chenyuxyz:
can you add a test? also this pr is not "fix vec dtype in promo lattice"

ttomsa:
@chenyuxyz what is the correct name for this pr?

chenyuxyz:
i would put something like `fix vec dtype in fast_idiv`, it's a bug in fast_idiv, not promo lattice

author of promo lattice (me in this case) would think there's an issue in that, while the author of fast_idiv might not be aware of the actual issue

chenyuxyz:
@S-Lykles can you review this one?

S-Lykles:
Looks good, can you add a vec=False to UPat.cvar("d") as well

PR #12087, opened 2025-09-09
remove redundant check

b1tg:


b1tg:
already have the check in TestDType.setUpClass

PR #12088, opened 2025-09-09
change dtype promo lattice when fp8s is supported

b1tg:


tinygrad/dtype.py
```diff
@@ -187,6 +187,9 @@ def _get_recursive_parents(dtype:DType) -> set[DType]:
   return set.union(*[_get_recursive_parents(d) for d in promo_lattice[dtype]], {dtype}) if dtype != dtypes.float64 else {dtypes.float64}
 @functools.cache
 def least_upper_dtype(*ds:DType) -> DType:
+  from tinygrad.device import is_dtype_supported
+  if is_dtype_supported(dtypes.fp8e4m3) and is_dtype_supported(dtypes.fp8e5m2):
```
chenyuxyz: so i don't really like this, because it's checking device.DEFAULT, not the device of the tensor

does it work if we just add an edge from fp8 to fp16/bf16? so that fp8+fp16->fp16 and fp8+bf16->bf16.

what does jax do when adding fp8e4m3 to fp8e5m2 and some different dtype combination? we'd like to match and test that

b1tg: this edge already exists in tinygrad

> does it work if we just add an edge from fp8 to fp16/bf16? so that fp8+fp16->fp16 and fp8+bf16->bf16.

below is the comparation of tinygrad(master)/jax dtype promotion

the differences are:
1. jax don't support promote fp8s with other float type
2. fp8 + int64 or uint64 => half / fp8
3. int64 + uint64 =>  half / float32
4. jax don't promote to 64bit (by default)



```
Comparing dtype promotion (JAX vs tinygrad):

A              B              tinygrad                 jax                     
--------------------------------------------------------------------------------
fp8e4m3        fp8e4m3        fp8e4m3                  float8_e4m3fn           
fp8e4m3        fp8e5m2        half                     <error: TypePromotionError>
fp8e4m3        float16        half                     <error: TypePromotionError>
fp8e4m3        bfloat16       bfloat16                 <error: TypePromotionError>
fp8e4m3        float32        float                    <error: TypePromotionError>
fp8e4m3        float64        double                   <error: TypePromotionError>
fp8e4m3        int64          half                     float8_e4m3fn           
fp8e4m3        uint64         half                     float8_e4m3fn           
fp8e5m2        fp8e5m2        fp8e5m2                  float8_e5m2             
fp8e5m2        float16        half                     <error: TypePromotionError>
fp8e5m2        bfloat16       bfloat16                 <error: TypePromotionError>
fp8e5m2        float32        float                    <error: TypePromotionError>
fp8e5m2        float64        double                   <error: TypePromotionError>
fp8e5m2        int64          half                     float8_e5m2             
fp8e5m2        uint64         half                     float8_e5m2             
float16        float16        half                     float16                 
float16        bfloat16       float                    float32                 
float16        float32        float                    float32                 
float16        float64        double                   float32                 
float16        int64          half                     float16                 
float16        uint64         half                     float16                 
bfloat16       bfloat16       bfloat16                 bfloat16                
bfloat16       float32        float                    float32                 
bfloat16       float64        double                   float32                 
bfloat16       int64          bfloat16                 bfloat16                
bfloat16       uint64         bfloat16                 bfloat16                
float32        float32        float                    float32                 
float32        float64        double                   float32                 
float32        int64          float                    float32                 
float32        uint64         float                    float32                 
float64        float64        double                   float32                 
float64        int64          double                   float32                 
float64        uint64         double                   float32                 
int64          int64          long                     int32                   
int64          uint64         half                     float32                 
uint64         uint64         ulong                    uint32   
```

https://gist.github.com/b1tg/417c333e740bdfd42abd0d2ac275d10f

b1tg: i guess 1. 3. 4. are what we intend to be different with jax and 2. need change to match jax behavior, just leave half for 3. ?

chenyuxyz: my preference would be if the promoted dtype does not loss precision than we can support it. int64+uint64 is always annoying because we don't have float weak dtype, but i am fine with half for now, instead of the lowest float. you can imagine that float4 is definitely not what we want later.

be careful for `in fps`, as we might want to support `e8m0` later and always promo to half will not be a good choice

tinygrad/dtype.py
```diff
@@ -181,13 +181,16 @@ def to_dtype(dtype:DTypeLike) -> DType: return dtype if isinstance(dtype, DType)
   dtypes.uint32: [dtypes.int64, dtypes.uint64], dtypes.uint64: [dtypes.float16, dtypes.bfloat16],
   dtypes.fp8e5m2: [dtypes.float16, dtypes.bfloat16], dtypes.fp8e4m3: [dtypes.float16, dtypes.bfloat16],
   dtypes.float16: [dtypes.float32], dtypes.bfloat16: [dtypes.float32], dtypes.float32: [dtypes.float64], }
+promo_lattice_fp8 = { **promo_lattice, dtypes.int64: [dtypes.fp8e4m3, dtypes.fp8e5m2], dtypes.uint64: [dtypes.fp8e4m3, dtypes.fp8e5m2] }
 
 @functools.cache
-def _get_recursive_parents(dtype:DType) -> set[DType]:
-  return set.union(*[_get_recursive_parents(d) for d in promo_lattice[dtype]], {dtype}) if dtype != dtypes.float64 else {dtypes.float64}
+def _get_recursive_parents(dtype:DType, fp8:bool) -> set[DType]:
```
chenyuxyz: let's not add a arg here, it's annoying to maintain and to add new ones

if it's simpler, i think fine to say `uint64+int64->fp8` for now and if user does not like this then just manually cast dtype

PR #12089, opened 2025-09-09
WIP: Mesa NIR backend (NAK)

sirhcm:
WIP: Everything except local memory seems to work (NAK isn't using the offset stored in `c[0][0x28]` when register spills happen).

Major caveat: right now requires recompiling mesa after running the following:
1. `sed -i '/gnu_symbol_visibility/d' src/compiler/meson.build src/compiler/nir/meson.build src/nouveau/compiler/meson.build`
2. `patchelf --add-rpath $MESA_PATH/lib/x86_64-linux-gnu $MESA_PATH/lib/x86_64-linux-gnu/libvulkan_nouveau.so`

This is because the symbols we need from `libvulkan-nouveau.so` are hidden otherwise. I'm not sure how best to circumvent this: one method would be to store a the offsets of the functions we need for whichever versions of ubuntu we want to support?

Other notes:
* viz is unhappy because ctypes objects can't be pickled (the pattern matcher I use for codegen operates over `nir_def` structs). Similarly `DEBUG=4` can't print the the nir shader in the standard way (although `nir_print_shader` allows us to show it for debugging if needed).
* for testing in CI this needs to run on a real gpu, ocelot wont work because it's not PTX.


geohot:
Cool, is it fast? Like how does the generated code compare in speed?

It's particularly exciting that is NVIDIA, mesa outputs the SASS, not PTX, right? That would make this a fully open source stack.

For CI, is there a different backend we can target? I'm sure mesa has some CPU test thing.

sirhcm:
Performance looks decent, here's `external_test_speed_v_torch.py`. 
image

> It's particularly exciting that is NVIDIA, mesa outputs the SASS, not PTX, right?

Yup. Mesa actually just outputs the raw compiled SASS.

> That would make this a fully open source stack.

The only part that isn't open source is the disassembler!

> I'm sure mesa has some CPU test thing.

Looks like mesa has some ["software drivers"](https://docs.mesa3d.org/systems.html) which should work. I'll take a look at how to interface with these. Interestingly it says that LLVMpipe is multithreaded?

gonsolo:
Running ```./autogen.sh nak``` I get:

>   File "/home/gonsolo/work/tinygrad/.venv/lib/python3.12/site-packages/ctypeslib/codegen/clangparser.py", line 213, in register
    raise DuplicateDefinitionException(
ctypeslib.codegen.handler.DuplicateDefinitionException: "register: nir_variable_mode which has a previous incompatible definition: nir_variable_mode\ndefined here: None\npreviously defined here: ('/tmp/mesa-9e0991eff5aea2e064fc16d5c7fa0ee6cd52d894/src/compiler/nir/nir_defines.h', 61)"

Is this error known?

sirhcm:
> Running `./autogen.sh nak` I get: 
> ...

Hmm, I'm not sure, this doesn't happen on my machine (Ubuntu 24.04). If you use the the version of mesa I pinned in autogen_stubs then you can just download the autogen'd files from this PR.

sirhcm:
~I'm not sure what's going wrong with the openpilot test, since afaik I didn't touch any of these systems and it fails on master on my machine?~ Nevermind, changing `support/mesa.py` fixed it?

chenyuxyz:
@nimlgen can you take a look?

tinygrad/runtime/ops_nv.py
```diff
@@ -185,68 +187,73 @@ def _submit(self, dev:NVDevice): self._submit_to_gpfifo(dev, dev.dma_gpfifo)
...
-      elif m:=re.match(r'\.nv\.constant(\d+)', sh.name): self.constbufs[int(m.group(1))] = (self.lib_gpu.va_addr+sh.header.sh_addr, sh.header.sh_size)
-      elif sh.name.startswith(".nv.info"):
-        for typ, param, data in self._parse_elf_info(sh):
-          if sh.name == f".nv.info.{name}" and param == 0xa: cbuf0_size = struct.unpack_from("IH", data)[1] # EIATTR_PARAM_CBANK
-          elif sh.name == ".nv.info" and param == 0x12: self.lcmem_usage = struct.unpack_from("II", data)[1] + 0x240 # EIATTR_MIN_STACK_SIZE
-          elif sh.name == ".nv.info" and param == 0x2f: self.regs_usage = struct.unpack_from("II", data)[1] # EIATTR_REGCOUNT
+
+    if (NAK:=isinstance(dev.compiler, NAKCompiler)):
+      image, self.cbuf_0, self.regs_usage = memoryview(lib[ctypes.sizeof(info:=mesa.struct_nak_shader_info.from_buffer(lib)):]), [], info.num_gprs
+      self.lib_gpu = self.dev.allocator.alloc(round_up((prog_sz:=image.nbytes), 0x1000) + 0x1000, buf_spec:=BufferSpec(cpu_access=True))
```
nimlgen: can this line be shared between both?

tinygrad/runtime/ops_cpu.py
```diff
@@ -92,7 +106,7 @@ def __init__(self, dev, name:str, lib:bytes):
 
       self.fxn = ctypes.CFUNCTYPE(None)(mv_address(self.mem))
 
-    super().__init__(HCQArgsState, dev, name, kernargs_alloc_size=0)
+    super().__init__(LVPArgsState if not WIN and LVP else HCQArgsState, dev, name, kernargs_alloc_size=12+256 if not WIN and LVP else 0)
```
nimlgen: if an arch/os is not supported, the compiler should assert earlier in init.

tinygrad/runtime/support/elf.py
```diff
@@ -19,6 +19,7 @@ def _to_carray(sh, ctype): return (ctype * (sh.header.sh_size // sh.header.sh_en
   rela = [(sh, sh.name[5:], _to_carray(sh, libc.Elf64_Rela)) for sh in sections if sh.header.sh_type == libc.SHT_RELA]
   symtab = [_to_carray(sh, libc.Elf64_Sym) for sh in sections if sh.header.sh_type == libc.SHT_SYMTAB][0]
   progbits = [sh for sh in sections if sh.header.sh_type == libc.SHT_PROGBITS]
+  progbits.sort(key=lambda sh: sh.name != ".ltext") # LLVM 20 JIT puts .ltext after .rodata
```
nimlgen: why is this needed?

sirhcm: LLVMpipe outputs an object with program text in .ltext instead of .text, and this section appears after .rodata. This wouldn't matter if .rodata was empty, but LLVMpipe puts constants into .rodata, so as a result the start of the buffer we give to CPUProgram is the contents of rodata and not program text. This could be fixed by having CPUProgram be aware of the offset to the start of program text, but it seemed simpler to just force ltext to appear at the beginning.

FWIW, it would be better to load every section into a different buffer and map them all separately, because rodata doesn't need to be RWX...

nimlgen:
so, it's a lot slower.

```
gpt2:
NAK: ran model in  20.32 ms
PTX: ran model in   3.65 ms
NVRTC: ran model in   3.43 ms
```

sirhcm:
Hmm, I see that now. After a brief glance at the disassembly, for some reason it's outputting every load/store with strong semantics (gpu scope): `LDG.E.STRONG.GPU` whereas NVRTC outputs `LDG.E`. Switching the loads to `LDG.E.CONSTANT` gets me from 27ms -> 22ms in gpt2. The stores appear to be [hardcoded](https://elixir.bootlin.com/mesa/mesa-25.2.3/source/src/nouveau/compiler/nak/from_nir.rs#L3568) as `STRONG.GPU` however.

Most of the performance difference seems to be due to nir not being capable of partially unrolling loops: NVRTC can obviously do this, but surprisingly NVPTX does it too. However nir only considers fully unrolling the loop.

tinygrad/runtime/support/mesa.py
```diff
@@ -0,0 +1,5 @@
+import os, ctypes, ctypes.util
+from tinygrad.helpers import getenv
+
+found = (os.path.exists(path:=(PATH:=getenv('MESA_PATH', '/usr/lib'))+"/libtinymesa_cpu.so") or (path:=ctypes.util.find_library('tinymesa_cpu') or '')
```
geohot: If this is only imported in one place it shouldn't be it's own file

tinygrad/runtime/support/elf.py
```diff
@@ -59,5 +60,6 @@ def jit_loader(obj: bytes) -> bytes:
   image, _, relocs = elf_loader(obj)
   # This is needed because we have an object file, not a .so that has all internal references (like loads of constants from .rodata) resolved.
   for ploc,tgt,r_type,r_addend in relocs:
+    if isinstance(tgt, str): raise RuntimeError(f'Attempting to relocate against an undefined symbol {tgt}')
```
geohot: Is there a test for this?

sirhcm: The codepath is hit by `test/unit/test_elf.py::TestElfLoader::test_clang_jit_compiler_external_raise`, but that doesn't explicitly call jit_loader.

tinygrad/runtime/support/compiler_mesa.py
```diff
@@ -0,0 +1,81 @@
...
+
+  def __reduce__(self): return NAKCompiler, (self.arch, self.warps_per_sm)
+
+  def compile(self, src) -> bytes:
+    shader = deserialize(src, self.nir_options)
+    mesa.nak_preprocess_nir(shader, self.cc)
+    ret = bytearray(bytes((out:=mesa.nak_compile_shader(shader, False, self.cc, 0, None).contents).info) + ctypes.string_at(out.code, out.code_size))
+    mesa.nak_shader_bin_destroy(out)
+    mesa.ralloc_free(shader)
+    return ret
```
nimlgen, line 79: beam doesn't work

1) bytearray is unhashable
2) works with PARALLEL=0, but not the default one (PARALLEL=1)

tinygrad/runtime/support/elf.py
nimlgen: this looks like a hack. what would it take for CPUProgram to take entry_offset? let's fix it in a sep prereq pr

tinygrad/runtime/support/compiler_mesa.py
sirhcm: `NV_NAK=1 NV=1 python3 -m pytest -s test/test_tiny.py::TestTiny::test_beam` works on my machine? what command are you running?

nimlgen: it can't find anything

`NV=1 NV_NAK=1 IGNORE_BEAM_CACHE=1 DEBUG=2 BEAM=4 PYTHONPATH=. python3 examples/gpt2.py --seed 0 --timing`

Screenshot 2025-10-12 at 09 46 42

sirhcm: Oh, got it, didn't realize it would silently fail.

tinygrad/runtime/support/elf.py
sirhcm: ~Hmm, what about sorting sections which have `SHF_EXECINSTR` to the beginning? Passing additional parameters to CPUProgram dependent on the output of the Compiler stage seems like a nontrivial architecture change.~ Nevermind, I figured out a way to not need this change at all.

sirhcm:
BEAM works now! In `gpt2.py` with `BEAM=4`, NAK gets ~5ms, NVRTC gets ~3ms, and NVPTX gets ~3ms (on a 3090).

geohot:
Congrats! Sorry it took so long to review, this is one of the best done bounties in a while. george@tinygrad.org to claim, both the base and the Mac bounties.

You're about to see a big payoff btw. Ready for the first Mac with an NVIDIA GPU in 10 years?

sirhcm:
No worries, excited to see what can be done with this backend!

qwertychouskie:
Is there anything preventing this from using ACO for AMD cards?  Would be cool to see another option/codepath for these cards.

sirhcm:
I haven't looked too closely at the ACO backend yet, prioritizing freedreno support right now, but in theory it should just work.

PR #12107, opened 2025-09-10
viz: light up children

Qazalin:
image


Qazalin:
The tag is still incoming parent's index:
image
Do we want the child index?
image
I don't find the child number meaningful, it's possible to implement though:
https://github.com/qazalin/tinygrad/commit/b80a6a186af0a68847978e4d3bda2ad6a9549c58

PR #12109, opened 2025-09-10
Onnx Layer/Group/RMS/Batch-Norm ReduceL2 fp32 intermediates for fp16 

S-Lykles:


tinygrad/frontend/onnx.py
```diff
@@ -897,7 +898,7 @@ def TopK(X:Tensor, K:int|list[int], axis:int=-1, largest:int=1, sorted:int=1):
   def BatchNormalization(X:Tensor, scale:Tensor, B:Tensor, input_mean:Tensor, input_var:Tensor, epsilon:float=1e-05, momentum:float=0.9,
                         training_mode:int=0, spatial=1, is_test=0):
     if training_mode:
-      x_detached = X.detach()
+      x_detached = X.detach().cast(dtypes.float) if X.dtype in (dtypes.float16, dtypes.bfloat16) else X.detach()
```
chenyuxyz: i think what you want is cast to least_upper_dtype with float32

`(dtypes.float16, dtypes.bfloat16)` is not future proof for fp8 or other small dtypes, and cast to float is bad for float64

S-Lykles: right, I was taking the spec quite literally, Ill change it

S-Lykles:
Fixes #12711

S-Lykles:
All this fp16 stuff is hard to test in CI because its not really supported on the CI machines

PR #12113, opened 2025-09-11
viz: spread edges

Qazalin:
aiming to create space for the labels:
image
vs master:
image


Qazalin:
this is a good test case, it's not showing at all that there's two incoming ranges.
image

With offsets:
image

need to fix sink:
image

PR #12123, opened 2025-09-11
fix test_dequantization_mxfp4

chenyuxyz:


chenyuxyz:
@jordan-chalupka fyi the test was wrong. i fixed the sign bit but the output is slightly different on ci

PR #12128, opened 2025-09-12
Fix advanced tensor indexing setitem

usatie:
While trying to complete the bounty issue "Remove realize from __setitem__ and get TestSetitemLoop.test_arange to be one kernel", I found a bug in `__setitem__` with advanced indexing (specifically with multiple tensor indexing that requires permutation). So I added the failing test case and fixed it.

Ready for review.


chenyuxyz:
thanks! also i'd like to see a change that removes `special permute case`, that seems hard to maintain

usatie:
Thanks for merging!
What do you mean by `removes special permute case`?
Is it ok to break compatibility with numpy/pytorch indexing?

PR #12132, opened 2025-09-12
[bounty] Remove realize from __setitem__ and get TestSetitemLoop.test_arange to be one kernel

subeep:
-removed the realize from test_setitem.py and fused the TestSetitemLoop.test_arrange in a single kernel which was taking two kernels before. 
-mainly done by using vectorized assign arange instead of scalar indexing in a loop. kernels are only computed at numpy() instead of realize().


geohot:
Many tests failing

subeep:
Yeah working on it, mainly happens due to removing realize from __setitem__ in tensor.py, the single kernel solution for test_setitem in the first commit works correctly with all tests.

chenyuxyz:
failing

PR #12133, opened 2025-09-12
[WIP] Trying to remove realize from setitem

Bankminer78:
This is a _super rough draft_  (still needs cleanup and prettifying and combing through logic, but the main approach seems right) of removing realize (and subsequently assign) from `__setitem__`. The biggest add is logic to compute padding specs for indices then using this to generate a boolean mask and padded rhs to use in a where op instead of assign. 

Removing realize also means removing assigns from inplace dunder methods like __imul__ because assign does not natively support fusing or stacking assigns on views. open question: is this the right direction, since I’m not sure what the long-term plan for assign is?

TestSetitemLoop now is one kernel (except for the kernel used for testing/comparing purposes like arrange and tolist)!

All tests except TestSetitemLoop.test_simple_jit_setitem pass. Figuring out how this Ops.where paradigm works with TinyJIT is the next big to do along with deciding the formulation for inplace dunders.

### DIFF
--- a/tinygrad/tensor.py
+++ b/tinygrad/tensor.py
@@ -1282,20 +1282,6 @@ class Tensor(MathTrait):
     """
     return self._getitem(indices)
 
-  def _inflate_axis(t: Tensor, axis: int, n_sel: sint, step_abs: int, span_len: sint) -> Tensor:
-    if step_abs == 1:
-      return t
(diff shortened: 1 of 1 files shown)


Bankminer78:
Closed because i dont think this is the right direction

PR #12157, opened 2025-09-13
use CUDA primary context for device interoperability

L1aoXingyu:
When running hlb-cifar10 with the tiny backend, I encountered the following error:

```
E       RuntimeError: CUDA error: invalid resource handle
E       CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
E       For debugging consider passing CUDA_LAUNCH_BLOCKING=1
E       Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

After debugging (with ChatGPT’s help), I found the root cause: when the device "tiny" (tinygrad frontend) is enabled, the hlb-cifar10 training script continues to use torch.cuda.Event for timing. However, tinygrad’s NV/CUDA backend directly uses the GPU driver, which invalidates PyTorch’s CUDA events. As soon as tinygrad touches the GPU, PyTorch CUDA events become unusable, leading to the error.

To reproduce the issue, I wrote a simple unit test. I then modified tinygrad to use the default CUDA context, which resolves the conflict.

### DIFF
--- a/test/mockgpu/cuda/cuda.py
+++ b/test/mockgpu/cuda/cuda.py
@@ -51,7 +51,8 @@ def cuCtxCreate_v2(pctx, flags: int, dev: int) -> int:
   pctx._obj.value = ctx_id
   return orig_cuda.CUDA_SUCCESS
 
-def cuDevicePrimaryCtxRetain(pctx, dev: int) -> int:
+def cuDevicePrimaryCtxRetain(pctx, dev: int | ctypes.c_int) -> int:
+  dev = dev.value if isinstance(dev, ctypes.c_int) else dev
   if dev not in cuda_state.devices or "primary_ctx" not in cuda_state.devices[dev]:
     ctx_id = cuda_state.next_context_id
     cuda_state.next_context_id += 1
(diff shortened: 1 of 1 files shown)


tinygrad/runtime/ops_cuda.py
```diff
@@ -98,7 +98,14 @@ def __init__(self, device:str):
     device_id = int(device.split(":")[1]) if ":" in device else 0
     check(cuda.cuInit(0))
     self.cu_device = init_c_var(cuda.CUdevice(), lambda x: check(cuda.cuDeviceGet(ctypes.byref(x), device_id)))
-    self.context = init_c_var(cuda.CUcontext(), lambda x: check(cuda.cuCtxCreate_v2(ctypes.byref(x), 0, self.cu_device)))
+    # Use the device primary context so events/streams interop with torch.cuda
+    if MOCKGPU:
+      # mock backend may not implement primary ctx APIs, keep previous behavior
```
nimlgen: you can add a support for this if you need

L1aoXingyu: you mean put the primary ctx into MOCKGPU?

nimlgen: extend the cuda mockgpu and implement what cuDevicePrimaryCtxRetain does

L1aoXingyu: review it again?

tinygrad/runtime/ops_cuda.py
```diff
@@ -128,3 +129,9 @@ def synchronize(self):
   @staticmethod
   def synchronize_system():
     for d in CUDADevice.devices: d.synchronize()
+
+  @suppress_finalizing
+  def __del__(self):
+    # drop our retain on the device primary context
+    try: check(cuda.cuDevicePrimaryCtxRelease_v2(self.cu_device))
```
nimlgen, line 136: why this "try"?

L1aoXingyu: In case a runtime error occurs during context release

nimlgen: why do we expect to get a runtime error?

L1aoXingyu: Good point — I don’t have evidence that this throws here; it was added defensively. What’s your suggestion?

L1aoXingyu: hello, any suggestions? @nimlgen

L1aoXingyu: hi, any suggestions? @nimlgen

L1aoXingyu:
hi，can you review it again? @nimlgen

geohot:
I generally like this idea, @nimlgen when you get a chance

geohot:
Actually @L1aoXingyu you got clear feedback, feel free to reopen if fixed and still relevant

PR #12165, opened 2025-09-14
amd: use aql queue v2

nimlgen:
backward compat with v1, but new fields are not correctly filled in this pr (this is only for new FWs)


nimlgen:
this is required only with async reclaim flag. won't update for now

PR #11304, opened 2025-07-21
Stable Diffusion mlperf training

hooved:
This PR implements training and eval of the Stable Diffusion v2 UNet, with every component tested for conformity to the mlperf reference implementation.

Although the code is close to being ready for training runs, I just noticed the mlperf training benchmarks have been updated for v5.1, with Stable Diffusion v2 being replaced by FLUX.1. Therefore it seems that this work (and the Stable Diffusion mlperf training bounty) are probably not relevant going forward, and I will close this PR assuming that's the case. Maybe this PR would still have some value to someone as a reference.

Below would have been the next steps from the current state of the PR:

- Modify training loop to use fp16 math (currently uses fp32 math) and loss scaling
- Implement and use gelu without tanh approximation to match mlperf reference
- Modify training/eval to use multigpu (currently uses 1 gpu)
- Attempt convergent training runs

PR #12170, opened 2025-09-14
Dataset and model download for SD-mlperf training and eval

hooved:
Part of #11304.

This PR includes a script for downloading datasets and models needed for training and eval of mlperf Stable Diffusion v2.


chenyuxyz:
i am testing this on tinyamd2, will merge once it finishes fine

PR #12192, opened 2025-09-15
rangeify: simplify reshaped buffers

Qazalin:
The tags get merged:
```
VIZ=1 RANGEIFY=1 python test/test_schedule.py TestSchedule.test_contiguous_while_contiguous
```
image

image


Qazalin:
pre tagged ops are annoying:
image
> "if it's not tagged by here, it's out"

Qazalin:
not merging that tag_uop change. There should be a way to identify we're bufferizing a Tensor to an existing buffer.

PR #12193, opened 2025-09-15
Add test for 2D tensor indexing in setitem

usatie:
While trying to complete the bounty issue "Remove realize from setitem and get TestSetitemLoop.test_arange to be one kernel", I found another bug in `__setitem__` with 2d-tensor indexing. So I added the failing test case and fixed it.


chenyuxyz:
thanks!

PR #12201, opened 2025-09-16
LR scheduler for Stable Diffusion mlperf training

hooved:
This PR adds the LR scheduler used in mlperf-v5.0 Stable Diffusion-v2 training, which linearly warms up the learning rate over a set number of steps to the final learning rate.

This is a subset of #11304.

Edit: I've updated the PR with a test which can be run as follows:
1. install numpy (testing dependency)
2. run `PYTHONPATH=. python test/external/mlperf_stable_diffusion/external_test_train.py`


chenyuxyz:
for this and things like #12202, either you add a test or have an instruction for me test locally, or it should not be a separate pr. add the code only if the code is used somewhere!

hooved:
I added a test in `test/external/mlperf_stable_diffusion/external_test_train.py`, which can be run with `PYTHONPATH=. python test/external/mlperf_stable_diffusion/external_test_train.py`

chenyuxyz:
can you add the test to https://github.com/tinygrad/tinygrad/blob/ed90de6583fd3ce3e11f0fc0e1a12f90526955cf/.github/workflows/test.yml#L445 so it runs in CI? either combined with existing ones or standalone is fine

hooved:
I combined the test into an existing file, and it ran successfully in CI.

examples/mlperf/lr_schedulers.py
```diff
@@ -36,4 +37,25 @@ def __init__(self, optimizer:Optimizer, base_lr, end_lr, warmup_steps:int, decay
...
+# based on torch.optim.lr_scheduler.LambdaLR
+class LambdaLR(LR_Scheduler):
+  def __init__(self, optimizer:Optimizer, base_lr:Tensor, lr_lambda:Callable):
+    super().__init__(optimizer)
+    self.base_lr, self.lr_lambda = base_lr, lr_lambda
+    self.step()
+
+  def get_lr(self):
+    ret = self.base_lr * self.lr_lambda(self.epoch_counter - 1)
+    return ret.to(self.optimizer.device)
```
chenyuxyz: why do you need this copy? this lr can be computed on each device without a copy assuming the tensors are already properly sharded

hooved: this copy isn't needed, thanks for pointing this out.

PR #12232, opened 2025-09-18
rangeify: fix copy size mismatch errs

Qazalin:
a lot of these are failing
image

Because pm_cleanups removes the bufferizes of COPY's parent.

In the current design, REALIZE can be removed. But realize before a COPY (even if it's a system op), needs to always happen.

We need a way to express this. maybe even insert **contiguous** between COPY and its parents.

update: it's fine if you push the compute to the other device. But currently rangeify is either silently wrong or runtime asserts.


Qazalin:
Same problem with multi, COPY should not do any of this indexing.
image
This is a good copy following the specs:
Screenshot 2025-09-18 at 2 17 27 PM

PR #12235, opened 2025-09-18
RANGEIFY test_tensor

chenyuxyz:


Qazalin:
it'd really help with dev workflow if we use @skipIf(RANGEIFY) instead of -k. I don't usually copy pytest commands from test.yml.

chenyuxyz:
i don't mind either way, maybe less prefer skip because not many people read and clean up old skips.
not pasting the pytest command and you will see the tests failing, isn't that working as intended? feel free to change as you see fit

PR #12245, opened 2025-09-19
enable rangeify test_symbolic_jit

S-Lykles:


S-Lykles:
@Qazalin can you look at this last failing test?

Qazalin:
https://github.com/tinygrad/tinygrad/pull/12247 should fix the tensor map err. I get:
`PYTHONPATH=. RANGEIFY=1 python test/test_symbolic_jit.py TestSymbolicJit.test_ones_sum`
```
AssertionError:
Items are not equal:
 ACTUAL: 2.0
 DESIRED: 1.0
 ```

**Comment by S-Lykles** [comment]
uhm, this is weird, but the jit is affecting the non symbolic `expected` output

```python3
def f(a): return a.sum().realize()
t = Tensor.ones(10)
for i in range(1,4):
  print(f(t[:i]).item())
```
outputs:
```
1.0
2.0
3.0
```

with the jitted function running
```python3
def f(a): return a.sum().realize()
jf = TinyJit(f)
t = Tensor.ones(10)
for i in range(1,4):
  vi = UOp.variable("i", 1, 10).bind(i)
  jf(t[:vi])
  print(f(t[:i]).item())
```
outputs
```
1.0
1.0
1.0
```

And moving the expected value before the jit gives the same error as before 
```python3
def f(a): return a.sum().realize()
jf = TinyJit(f)
t = Tensor.ones(10)
for i in range(1,4):
  print(f(t[:i]).item())
  vi = UOp.variable("i", 1, 10).bind(i)
  jf(t[:vi])
```
`IndexError: list index out of range`

Qazalin:
I can't repro that second [1.0, 1.0, 1.0]. The JIT seems fine with symbolic and RANGEIFY=1 now.

PR #12246, opened 2025-09-19
[Bounty] Remove realize from setitem and get TestSetitemLoop.test_arange to be one kernel

usatie:
Implemented `__setitem__` without using `realize`, and removed setitem-related code (tensor indexing) from `_getitem`.

I fixed JIT failure and inplace ops failure by modifying tests, but not sure if it was right.
Other issues remain, and I’m not sure if modifying __setitem__ alone will be enough.

The code is a bit long for now to help with debugging. Once the approach is confirmed, I'll clean it up.

Issues:
- [x] JIT failure
- [x] Inplace operation failure
- [ ] svd failure
```
test/unit/test_linalg.py::TestLinAlg::test_svd_general - TypeError: <lambda>() missing 1 required positional argument: 'x'
test/unit/test_linalg.py::TestLinAlg::test_svd_nonfull - TypeError: <lambda>() missing 1 required positional argument: 'x'
```
- [ ] ONNX backend failure
```
FAILED test/external/external_test_onnx_backend.py::OnnxBackendNodeModelTest::test_scatternd_cpu - AssertionError: 
```

### DIFF
--- a/test/test_setitem.py
+++ b/test/test_setitem.py
@@ -123,6 +123,7 @@ class TestSetitem(unittest.TestCase):
     @TinyJit
     def f(t:Tensor, a:Tensor):
       t[2:4, 3:5] = a
+      t.realize() # In order to force the setitem to happen here
 
     for i in range(1, 6):
       t = Tensor.zeros(6, 6).contiguous().realize()
--- a/test/test_setitem.py
+++ b/test/test_setitem.py
@@ -34,9 +34,11 @@ class TestSetitem(unittest.TestCase):
     self.assertListEqual(t.tolist(), [0, 1, 11, 3, 11, 5, 6, 7, 8, 9])
 
   def test_setitem_inplace_mul(self):
-    t = Tensor.arange(10).realize()
-    t[:3] *= 10
-    self.assertListEqual(t.tolist(), [0, 10, 20, 3, 4, 5, 6, 7, 8, 9])
+    # This test fails without RANGEIFY
+    with Context(RANGEIFY=1):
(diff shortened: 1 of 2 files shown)


usatie:
The simple JIT test also failed since it’s not JIT-able without realize in __setitem__, so I added a realize call in the test.
```
FAILED test/test_setitem.py::TestSetitem::test_simple_jit_setitem - AssertionError: didn't JIT anything!
```

usatie:
`__setitem__` itself works, but the `__imul__` with indexed test had issues. It passes with RANGEIFY=1, so I updated the test to include that, but I'm not sure if that's the right fix.
```
test/test_setitem.py::TestSetitem::test_setitem_inplace_mul - AssertionError: must be BUFFER Ops.BUFFER_VIEW
```

usatie:
@chenyuxyz Can you review this PR to help confirm if changing only `__setitem__` is enough, or if I also need to modify the kernelizer/linearizer?

chenyuxyz:
i think the idea of the bounty is to remove the realize in setitem and that's the only thing to change in tensor.py, the assign fusion should be handled by tinygrad schedule

usatie:
@chenyuxyz So, as I understand it, this PR is done, right?
And I need to wait for the other PR that fixes the assign fusion issue before this can be locked/merged?

chenyuxyz:
not following, why do we need this pr? the bounty is to fix assign fusion, not messing with setitem implementation

usatie:
Oh, okay. I didn't get the correct meaning of this issue.
Thanks for the clarification!

PR #12254, opened 2025-09-20
RANGEIFY=1 test_jit

Qazalin:
section 0 rewrites must be deleted.
`VIZ=1 PYTHONPATH=. RANGEIFY=1 python test/test_jit.py TestJit.test_jitted_transfers` creates with tag=None and the COPY never realizes.
image


Qazalin:
```
FAILED test/test_schedule.py::TestSchedule::test_mnist_val - AssertionError: buffer copy mismatch, 60000 != 60008, dtypes.uchar != dtypes.uchar
```
disk would be nice, beautiful_mnist is copying the dataset every time.

PR #12257, opened 2025-09-20
RANGEIFY=1 looping in map_pad

Qazalin:
@S-Lykles to repro, you can run `VIZ=1 PYTHONPATH=. RANGEIFY=1 python test/test_multitensor.py TestMultiTensor.test_data_parallel_resnet` (wait for ~5 seconds and ctrl+c).

It seems like map_pad never resolves the UOp.invalid call. I could trace it back to `nest_div_by_smallest_factor`
image

### DIFF
--- a/test/test_multitensor.py
+++ b/test/test_multitensor.py
@@ -372,7 +372,7 @@ class TestMultiTensor(unittest.TestCase):
 
   # NOTE: this is failing on LLVM CI, no idea why. Works locally.
   @unittest.skipIf(CI and REAL_DEV in ("CUDA", "NV", "CPU", "AMD"), "slow, and flaky on CPU")
-  @unittest.skipIf(RANGEIFY, "TODO: pm_rangeify hangs")
+  #@unittest.skipIf(RANGEIFY, "TODO: pm_rangeify hangs")
   def test_data_parallel_resnet(self):
     from extra.models.resnet import ResNet18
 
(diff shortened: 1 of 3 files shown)


S-Lykles:
So its not actually looping, it does finish scheduling after a couple of minutes (on a tiny green). The problem is that the calls to `simplify` are not cached and it keeps having to resimplify the same things. I was working on a version that didnt use simplify a while back #11812 but I have to work out some regressions.

Qazalin:
This is also related:
```
PYTHONPATH=. VIZ=1 RANGEIFY=1 python test/models/test_efficientnet.py TestEfficientNet.test_chicken_car
```
image
we can't currently cache symbolic globally, it blocks the Buffers GC. But I'm open to fixing GC if you believe that's the last option.

S-Lykles:
So even without `nest_div_by_smallest_factor` it still takes quite long, now all the time is being spent in `uop_given_valid` which also does repeated calls to simplify. I dont think we need a global cache but maybe a cache per call to uop_given_valid. If thats still too slow I can make the cache reused between different calls to `uop_given_valid` within a `map_pad` or `get_program`.

Qazalin:
loop issue is fixed, it's just taking 11 seconds to devectorize now
https://discord.com/channels/1068976834382925865/1230434680000741377/1420254852700835892.

PR #12269, opened 2025-09-22
rangeify: fix test_setitem

nimlgen:


Qazalin:
hmm why isn't this rule in normal symbolic?
It's in sym, not symbolic_simple...
image

nimlgen:
`# NOTE: sym (vs symbolic_simple) breaks things here because ranges with len 1 aren't handled right` looks like a hack.

Qazalin:
I'd move the WHERE simplifier pattern out of sym, it's trivial for symbolic_simple to do too.
a whole new graph_rewrite on bigwhere is expensive.

PR #12273, opened 2025-09-23
add split_reduceop to rangeify

S-Lykles:
Brings beautiful_mnist step time down from like 428 -> 265 ms


Qazalin:
This is a pure speed change right? I think we should focus on the correctness issues first.

S-Lykles:
> This is a pure speed change right? I think we should focus on the correctness issues first.

Yeah, its just that big sums are like 80x slower because of this, my thinking was that if we want to make rangeify the default it should not have those types of regressions

Qazalin:
Yea perf is always easier to work on when we're confident rangeify is correct. We shouldn't prematurely optimize things.

PR #12280, opened 2025-09-23
Control precision in attention softmax to enable copying torch AMP behavior

hooved:
In `scaled_dot_product_attention`, when `self` is a bf16 Tensor, you can't get softmax to run in float32. But torch upcasts bf16 to f32 for softmax on CUDA (see test in diff, also https://docs.pytorch.org/docs/stable/amp#cuda-op-specific-behavior), under typical mixed precision settings such as those used in the mlperf reference implementation of Stable Diffusion training.

This diff lets the programmer match torch behavior by giving control over the dtype of the Tensor that is fed into softmax. Using float32 for softmax gives substantially different output than bf16 (see the allclose tolerances in the test).


hooved:
part of #11304

tinygrad/tensor.py
```diff
@@ -3961,7 +3961,8 @@ def scaled_dot_product_attention(self, key:Tensor, value:Tensor, attn_mask:Tenso
     if attn_mask is not None:
       if attn_mask.dtype == dtypes.bool: attn_mask = attn_mask.where(0, -float("inf"))
       qk = qk + attn_mask
-    attn = qk.cast(self.dtype).softmax(-1).dropout(dropout_p) @ value
+    qk_softmax = qk.cast(self.dtype).softmax(-1) if softmax_dtype is None else qk.cast(softmax_dtype).softmax(-1).cast(self.dtype)
```
chenyuxyz: `softmax` can take a dtype, see if that can clean up more. i don't really like a dtype arg can seemingly change the whole expression

hooved:
If we passed the dtype to softmax without changing the expression:
`attn = qk.cast(self.dtype).softmax(-1, dtype=softmax_dtype).dropout(dropout_p) @ value`

The problem with the above is that before the softmax, there is a downcast from qk (float32) to self.dtype (bf16), which would destroy information.

Also, the tinygrad softmax dtype arg only applies to the exponentiation:

```python
  def _softmax(self, axis, dtype:DTypeLike|None=None) -> tuple[Tensor, Tensor, Tensor]:
    m = self - self.max(axis=axis, keepdim=True).detach()
    if dtype is not None: m = m.cast(dtype)
    e = m.exp()
    return m, e, e.sum(axis=axis, keepdim=True)
```

In constrast, torch upcasts to float32 before the max/difference steps of softmax. (Note this is what I have read, but the softmax autocast implementation is buried in compiled C code and I haven't debugged that far down in torch to be 100% sure).

chenyuxyz:
i want the softmax dtype behavior to match torch, the downcast of qk is probably a hack i put in for speed before

not sure if the dtype of max part matter, casting before / after should give the exact same output, so i would expect casting later is faster given softmax uses dtype >= its input

hooved:
The new test results in CI, from the most recent commit `08a2faad3ab917d69b736f663c15bb695f6301ff`, show that the dtype of max and subtract matter for matching torch softmax behavior, when you use bf16 input with autocast enabled on CUDA. To match this torch behavior, all you need to do is just cast bf16 to f32 before softmax (no extra dtype arg required). You could alternatively move the cast in the _softmax function to the beginning, and call softmax with `dtype=dtypes.float32`.

I reverted the change to `Tensor.scaled_dot_product_attention` because I realized the mlperf reference I'm copying is using an attention implementation separate from `torch.nn.functional.scaled_dot_product_attention`. While my initial diff to the tinygrad attention matched the behavior I was targeting, I realized that copying that exact function in torch is a separate rabbit hole that's out of scope.

What's left in this diff are just tests that define torch behavior without modifying tinygrad, so I'll close the PR.

PR #12294, opened 2025-09-24
perf: accelerate cpu sum reductions on mac

ry2009:


nimlgen:
no way this can be merged. too many changes in one pr and no benchmark results at all

PR #12295, opened 2025-09-25
reenable test_backward_sum_acc_dtype

chenyuxyz:
this fails with `RANGEIFY=1`


chenyuxyz:
hmm this test worked before due to const folding, adding contiguous to expand then it fails

PR #12299, opened 2025-09-25
rangeify permuted assign tests

Qazalin:
Loading target buf in permuted assign is wrong.
`CPU=1 RANGEIFY=1 python test/test_assign.py TestAssign.test_post_permuted_assignment` fails in master and passes in this branch.

Master deals with these cases by realizing src[1], this may get optimized away and still get the wrong answer.
This branch loads a clone of the Buffer.

---
notes:
- CPU=1 for deterministic wrong output. GPU may get a lucky right answer. 
- this can remove all the cases we're realizing src[1], CI failing due to diamond issues.

### DIFF
--- a/tinygrad/schedule/rangeify.py
+++ b/tinygrad/schedule/rangeify.py
@@ -86,6 +86,7 @@ do_realize = PatternMatcher([
   (UPat(Ops.ASSIGN, name="a"), realize_assign),
 ])
 
+
 class WrappedContig:
   def __init__(self, x): self.x = x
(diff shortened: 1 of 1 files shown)


Qazalin:
Status: It's correct but inserts extra kernels. This is why `TestJitFree.test_updated_not_freed` fails too.
need to differentiate between these two in the guard function `TestSchedule.test_setitem_sched` vs `TestSchedule.test_setitem_permuted_sched`
image
image

PR #12310, opened 2025-09-27
skip test_softmax_fusion tests if RANGEIFY==1

S-Lykles:


S-Lykles:
These tests only fuse with RANGEIFY>1

PR #12313, opened 2025-09-27
Clip model updates for Stable Diffusion mlperf training

hooved:
This PR makes the Clip tokenizer and embedding model compatible with mlperf-v5.0 Stable Diffusion-v2 multi-GPU training.

To test the new functionality:
1. install these dependencies, a subset of mlperf training dependencies: `pip install numpy ftfy regex pillow`
2. `PYTHONPATH=. python test/external/mlperf_stable_diffusion/external_test_models.py`

This is a subset of #11304

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -259,7 +259,7 @@ jobs:
       uses: ./.github/actions/setup-tinygrad
       with:
         key: unittest-12
-        pydeps: "pillow"
+        pydeps: "pillow numpy ftfy regex"
         deps: testing_unit
     - name: Run unit tests
       run: python -m pytest -n=auto test/unit/ --durations=20
@@ -268,7 +268,7 @@ jobs:
     - name: Run SDXL on NULL backend
       run: MAX_BUFFER_SIZE=0 NULL=1 DEBUG=1 python3 examples/sdxl.py --seed 0 --noshow --timing --fakeweights
     - name: Run Clip tests for SD MLPerf on NULL backend
-      run: NULL=1 python -m pytest -n=auto test/external/mlperf_stable_diffusion/external_test_models.py::TestOpenClip --durations=20
+      run: MAX_BUFFER_SIZE=0 NULL=1 python -m pytest -n=auto test/external/mlperf_stable_diffusion/external_test_models.py::TestOpenClip --durations=20
     # TODO: support fake weights
     #- name: Run LLaMA 7B on 4 fake devices
     #  run: NULL=1 python3 examples/llama.py --gen 1 --size 7B --shard 4 --prompt "Hello." --count 3 --temperature 0 --timing


extra/models/clip.py
```diff
@@ -449,7 +472,7 @@ def encode_tokens(self, tokens:Tensor) -> Tensor:
     x = x + self.positional_embedding
     x = self.transformer(x, attn_mask=self.attn_mask)
     x = self.ln_final(x)
-    x = x[:, tokens.argmax(axis=-1)]
+    x = x[Tensor.arange(x.shape[0], device=x.device), tokens.argmax(axis=-1), :]
```
chenyuxyz: why is this arange needed?

extra/models/clip.py
```diff
@@ -9,6 +9,19 @@
...
+
+# to match behavior of mlperf v5.0 Stable Diffusion training clip tokenizer
+try:
+  import ftfy, html, regex
+  # from open_clip.tokenizer:
+  def basic_clean(text):
+    text = ftfy.fix_text(text)
+    text = html.unescape(html.unescape(text))
+    return text.strip()
+except ImportError: pass
```
chenyuxyz: i think import in the if block directly is easier to know when are these needed

extra/models/clip.py
hooved: `x` has the shape `(BS, 77, 1024)`, and we want the result to have shape `(BS, 1024)`.

The original code `x[:, tokens.argmax(axis=-1)]` gives shape `(BS, BS, 1024)` because the `tokens.argmax(axis=-1)` index is broadcast across the first dim, if the first dim is sliced with `:`. This matches torch behavior.

But when you use the arange like this: `x[Tensor.arange(x.shape[0], device=x.device), tokens.argmax(axis=-1), :]`, the first and second indices are mapped elementwise, giving a result with the desired shape `(BS, 1024)`. This matches torch behavior.

The test in this diff, `test_multigpu_clip_score`, fails with the previous indexing  `x[:, tokens.argmax(axis=-1)]`.

extra/models/clip.py
hooved: added those imports to the if blocks

hooved:
I added the tests to CI.

test/external/mlperf_stable_diffusion/external_test_models.py
```diff
@@ -0,0 +1,53 @@
+import unittest
+from tinygrad import Tensor, dtypes, Device
+from tinygrad.nn.state import get_parameters
+from extra.models import clip
+from examples.mlperf.initializers import gelu_erf
+Device.DEFAULT="NULL"
```
chenyuxyz, line 6: this should pass with all device right? testing with NULL in CI is fine, but don't need to hard code this there

PR #12314, opened 2025-09-28
Stable Diffusion model init for mlperf

hooved:
This PR depends on https://github.com/tinygrad/tinygrad/pull/12313 (that should be merged first), and is a subset of https://github.com/tinygrad/tinygrad/pull/11304.

This PR contains everything needed to initialize the following Stable Diffusion models used in training and eval:
- Clip encoder
- UNet
- VAE

Tests are included to cover the new behavior, including differences from previous SD v1.5 version, differences between train/eval inits, and mixed precision handling.

To run the tests:
1. install dependencies: `pip install numpy ftfy regex pillow`
2. ensure mlperf checkpoint is present at `/raid/weights/stable_diffusion/sd/512-base-ema.ckpt`. Alternatively you can `export CKPTDIR=...` and ensure the mlperf checkpoint is at `$CKPTDIR/sd/512-base-ema.ckpt`. See https://github.com/tinygrad/tinygrad/pull/12170 for how to download the checkpoint.
3. Ensure 6 GB of host RAM is available to load checkpoints (the tests use CPU device)
4. run `PYTHONPATH=. python test/external/mlperf_stable_diffusion/external_test_models.py`

### DIFF
--- a/examples/stable_diffusion.py
+++ b/examples/stable_diffusion.py
@@ -9,11 +9,13 @@ from typing import Dict, Any
 from PIL import Image
 import numpy as np
 from tinygrad import Device, GlobalCounters, dtypes, Tensor, TinyJit
-from tinygrad.helpers import Timing, Context, getenv, fetch, colored, tqdm
+from tinygrad.helpers import Timing, Context, getenv, fetch, colored, tqdm, flatten
 from tinygrad.nn import Conv2d, GroupNorm
 from tinygrad.nn.state import torch_load, load_state_dict, get_state_dict
-from extra.models.clip import Closed, Tokenizer
+from extra.models.clip import Closed, Tokenizer, FrozenOpenClipEmbedder
+from extra.models import unet, clip
 from extra.models.unet import UNetModel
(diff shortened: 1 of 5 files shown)


examples/stable_diffusion.py
```diff
@@ -87,7 +89,12 @@ def __call__(self, x):
         bs,c,py,px = x.shape
         x = x.reshape(bs, c, py, 1, px, 1).expand(bs, c, py, 2, px, 2).reshape(bs, c, py*2, px*2)
         x = l['upsample']['conv'](x)
-      x.realize()
+
+      # for BEAM on mi300x, with decoder BS=384, (and other similar BS), BEAM kernels will hang:
```
chenyuxyz: let's separate the workaround from the absolutely parts. this can go with the main training loop pr and i'd like to train without the workaround to understand its intent better

hooved: I moved this out of `stable_diffusion.py`, into the eval code. Note this issue is only encountered in eval beam, not training. See https://github.com/tinygrad/tinygrad/pull/12316 for instructions to reproduce BEAM hang issue.

Issue #12238, opened 2025-09-18
arange folding leaves uncancelled idx


chenyuxyz:
```
from tinygrad import Tensor
t = Tensor.arange(5)
print((t+t[::-1]).tolist())
```
with `CPU=1 RANGEIFY=1` gives
```
void E_5(int* restrict data0_5, int core_id) {
  for (int ridx0 = 0; ridx0 < 5; ridx0++) {
    *(data0_5+ridx0) = ((ridx0-ridx0)+4);
  }
}
```
with `CL=1 RANGEIFY=1` gives
```
__kernel void E_5(__global int* data0_5) {
  int gidx0 = get_group_id(0); /* 5 */
  *(data0_5+gidx0) = ((gidx0-gidx0)+4);
}
```

wp4032:
Hi @chenyuxyz, can you please check PR #12319 for the fix above?

PR #12319, opened 2025-09-28
Fix arange folding with uncancelled idx

wp4032:
Fixed issue #12238

Fix:
- Added zero folding for Ops.SUB when the src values are the same

Results:
```
CPU=1 RANGEIFY=1 DEBUG=4 python3 arange.py
void E_5(int* restrict data0_5, int core_id) {
  for (int ridx0 = 0; ridx0 < 5; ridx0++) {
    *(data0_5+ridx0) = 4;
  }
}

*** CPU        1 E_5                                          arg  1 mem  0.00 GB tm     10.88us/     0.01ms (     0.00 GFLOPS    0.0|0.0     GB/s) ['__add__', 'arange']
[4, 4, 4, 4, 4]
``` 
```
CL=1 RANGEIFY=1 DEBUG=4 python3 arange.py
__kernel void E_5(__global int* data0_5) {
  int gidx0 = get_group_id(0); /* 5 */
  *(data0_5+gidx0) = 4;
}
*** CL         1 E_5                                          arg  1 mem  0.00 GB tm      8.46us/     0.01ms (     0.00 GFLOPS    0.0|0.0     GB/s) ['__add__', 'arange']
[4, 4, 4, 4, 4]
```

### DIFF
--- a/tinygrad/uop/symbolic.py
+++ b/tinygrad/uop/symbolic.py
@@ -73,7 +73,6 @@ symbolic_simple = propagate_invalid + PatternMatcher([
   (UPat.var("x", dtype=dtypes.ints+(dtypes.bool, dtypes.index)).trunc(), lambda x: x),
   # ** zero folding **
   (UPat.var("x") < UPat.var("x"), lambda x: x.const_like(False).cast(dtypes.bool.vec(x.dtype.count))), # x < x -> False
-  (UPat(Ops.SUB, src=(UPat.var("x"), UPat.var("x"))), lambda x: x.const_like(0)), # x-x -> 0
   (UPat.var("x") % UPat.var("x"), lambda x: x.const_like(0)), # x%x -> 0
   (UPat.var("x", dtype=dtypes.ints+(dtypes.bool, dtypes.index)) != UPat.var("x"),
    lambda x: x.const_like(False).cast(dtypes.bool.vec(x.dtype.count))), # x != x -> False (only ints)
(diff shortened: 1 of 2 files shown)


tinygrad/uop/symbolic.py
```diff
@@ -73,6 +73,7 @@ def convert(v:ConstType): return struct.unpack(to_fmt, struct.pack(from_fmt, v))
   (UPat.var("x", dtype=dtypes.ints+(dtypes.bool, dtypes.index)).trunc(), lambda x: x),
   # ** zero folding **
   (UPat.var("x") < UPat.var("x"), lambda x: x.const_like(False).cast(dtypes.bool.vec(x.dtype.count))), # x < x -> False
+  (UPat(Ops.SUB, src=(UPat.var("x"), UPat.var("x"))), lambda x: x.const_like(0)), # x-x -> 0
```
chenyuxyz: SUB is only introduced after decomposition, so we don't want to match it directly in symbolic

wp4032: Moved it into `decomposition.py`

chenyuxyz: no that's not what i meant, don't match on SUB

chenyuxyz:
i recommend studying the repo more, this and #12321 are bad and we won't merge poor workarounds

wp4032:
Will do, thank you

PR #12320, opened 2025-09-28
Unified rewrite uses deque

S-Lykles:


geohot:
Why is this faster? Shouldn't they be quickly ignored if they were on the stack?

If you want to merge stuff like this, I want a test showing a complexity reduction.

S-Lykles:
So there is something about children and RewriteNotReady thats causing bad behavior. To schedule openpilot you have to increase ctx.progress limit and the stack limit more than 10x. RewriteNotReady will cause parts of the graph to be put on the stack multiple times.

geohot:
I think this needs to be root caused and not hacked around

geohot:
Hmm, if this isn't slower maybe it's fine. I played with a few changes and nothing was better than this. Though I still think this should be root caused and tested and the bad behavior understood.

Going to merge, agreed that this isn't easy and annoying to debug.

PR #12321, opened 2025-09-29
Fixed max backward underflow

### LINKED ISSUE #12296: max backward can underflow
```
from tinygrad import Tensor
t = Tensor.ones(70000, dtype="half", requires_grad=True).contiguous()
t.max().backward()
print(t.grad.tolist()[:10])
```
gives `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]`

https://github.com/tinygrad/tinygrad/blob/476a2a0a962fa8a145c612f63b72eeb4dc3a84f3/tinygrad/gradient.py#L10 uses the gradient dtype for counting

wp4032:
Fixed #12296

Fix:
- Set cast to `dtypes.float32` and then recast output back to `ctx.dtype`
- Added test case to make sure max backward gradient does not underflow for float16, float32, and float64
- Fixed bug for `is_dtype_supported` when device=REMOTE and platform is OSX

Results:
```
from tinygrad import Tensor
t = Tensor.ones(70000, dtype="half", requires_grad=True).contiguous()
t.max().backward()
print(t.grad.tolist()[:10])

python backward.py
[1.430511474609375e-05, 1.430511474609375e-05, 1.430511474609375e-05, 1.430511474609375e-05, 1.430511474609375e-05, 1.430511474609375e-05, 1.430511474609375e-05, 1.430511474609375e-05, 1.430511474609375e-05, 1.430511474609375e-05]
```

### DIFF
--- a/test/unit/test_gradient.py
+++ b/test/unit/test_gradient.py
@@ -2,10 +2,11 @@ from typing import Callable
 import unittest, math
 import torch
 import numpy as np
-from tinygrad import Tensor
+from tinygrad import Device, Tensor
 from tinygrad.dtype import dtypes
 from tinygrad.uop.ops import UOp
 from tinygrad.gradient import compute_gradient
+from tinygrad.helpers import CI
 
 class TestGradient(unittest.TestCase):
   def _cmp_nan_okay(self, x, y):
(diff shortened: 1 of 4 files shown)


tinygrad/gradient.py
```diff
@@ -2,14 +2,15 @@
...
 from tinygrad.uop.ops import UOp, PatternMatcher, UPat, Ops, all_metadata
 from tinygrad.helpers import argsort
+from tinygrad.dtype import dtypes
 
 def reduce_gradient(ctx:UOp, ret:UOp):
   def to_inp_shape(x): return x.reshape(x.shape+(1,)*(len(ret.src[0].shape)-len(x.shape))).expand(ret.src[0].shape)
   if ret.arg[0] == Ops.ADD: return (to_inp_shape(ctx),)
   if ret.arg[0] == Ops.MAX:
-    max_is_1s = ret.src[0].eq(to_inp_shape(ret)).cast(ctx.dtype)
+    max_is_1s = ret.src[0].eq(to_inp_shape(ret)).cast(dtypes.float32)
```
chenyuxyz: does this work with float64?

test/unit/test_gradient.py
```diff
@@ -110,6 +110,12 @@ def test_non_float_tensor_raise(self):
     with self.assertRaises(RuntimeError): x.sum().gradient(x)
     with self.assertRaises(RuntimeError): x.float().sum().gradient(x)
 
+  def test_max_gradient_half_precision(self):
+    t = Tensor.ones(70000, dtype="half", requires_grad=True).contiguous()
+    t.max().backward()
+    expected_grad = [1.430511474609375e-05] * 10
```
chenyuxyz: hard coding the expected value makes it hard to update the tests in the future, like how quickly will `assertListEqual` fail due to numericals

tinygrad/gradient.py
wp4032: Does not work for METAL, so made it so METAL defaults to float32
Else everything else uses float16

test/unit/test_gradient.py
wp4032: Fixed and added more test cases for completeness

tinygrad/gradient.py
chenyuxyz: don't add one-off ifs, study how it's already done in the repo

wp4032: I saw `onnx.py` uses `dtype_fallback`, so I moved that function into `device.py` so that could be used here also

PR #12339, opened 2025-09-30
rangeify: only test correctness in multi

Qazalin:
basic const folding works, this mostly requires `COPY(CONST) -> CONST` simplifications.
few questions left to figure out:
1. Symbolic can make things that weren't const, const. If the COPY(CONST) simplifier rule runs before symbolic it misses those
2. Symbolic relies on a flat CONST structure. Old scheduler pushed views before CONST. Rangeify indexes it so mops disappear. Currently i'm left with graphs like this because symbolic + copy(const) ran and created `op.const_like()` on a shaped op, but we're past pm_rangeify and this doesn't simplify further
image


Qazalin:
failing after merge because cleanup_dead_axes hits before MSTACK
image

PR #12355, opened 2025-09-30
fix cifar training in RANGEIFY

geohot:


geohot:
@Qazalin can you look at this multi failure?

Qazalin:
okay_reduce is removing COPY's BUFFERIZE.

PR #12358, opened 2025-09-30
rangeify: fix test_grouped_store_values

b1tg:


chenyuxyz:
i think it's fine to skip that test in RANGEIFY if it depends on heuristic

PR #12371, opened 2025-09-30
rangeify: ban view pushing

Qazalin:
why doesn't this work?
edit: two bugs
1. It was shaping INDEX(MSELECT) - MSELECT is MSTACK is BUFFER is BUFFERIZE in this context. None of them have shape.
2. Codegen cache wasn't accounting for RANGEIFY=1 context var (it doesn't need to in master)


geohot:
Good idea. I would expect this to work

PR #12379, opened 2025-10-01
rangeify: test_where_fold kernel count

Qazalin:
This diff fuses ASSIGN in non-permute cases. #12384 folded where where where should be folded:
`VIZ=1 RANGEIFY=1 python test/test_linearizer.py TestLinearizer.test_where_fold`
image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -376,9 +376,10 @@ jobs:
           llvm: 'true'
       - name: Test openpilot model kernel count and gate usage
         run: |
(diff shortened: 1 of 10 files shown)


S-Lykles:
> now it's finding where where where should be folded: `VIZ=1 RANGEIFY=1 python test/test_linearizer.py TestLinearizer.test_where_fold` image

Oh I see, the pattern that removes this is not hitting because of the cast after the load, should be a simple fix, Ill add

Qazalin:
cool, I'll work on cleaning up the scheduling part.

PR #12394, opened 2025-10-01
work on double gemm tc

geohot:


Qazalin:
I think if INDEX is on BUFFERIZE as_buf should return the BUFFERIZE:

```
diff --git a/tinygrad/uop/ops.py b/tinygrad/uop/ops.py
index 7d3f807ef..fd01ad6b2 100644
--- a/tinygrad/uop/ops.py
+++ b/tinygrad/uop/ops.py
@@ -476,7 +476,7 @@ class UOp(MathTrait, metaclass=UOpMetaClass):
     if self.op is Ops.MSTACK: return UOp(Ops.MSTACK, self.dtype, src=tuple(x.as_buf() for x in self.src))
     # TODO: this should be the only one of these. this is the one RANGEIFY uses
     s = self
-    while len(s.src) and s.op not in {Ops.BUFFER, Ops.MSTACK}: s = s.src[0]
+    while len(s.src) and s.op not in {Ops.BUFFER, Ops.BUFFERIZE, Ops.MSTACK}: s = s.src[0]
     return s
```
image


It's saying that the buffer of INDEX is the UNIQUE 5 buffer on the far left

PR #12401, opened 2025-10-01
don't use UOp.size in Tensor._frompy

chenyuxyz:
.size needs .st


Qazalin:
I think .size reads better, we can just update .size to not use st once ShapeTracker is gone.
I think the ops.py size method will stay.

chenyuxyz:
yea that's fine, i don't quite see what replace the .st in .size though, is it some kind of shape propagation?

Qazalin:
yea we still need .shape, it'll propagate from (N,) for Buffer-like ops, () for CONST through MovementOps

PR #12402, opened 2025-10-01
reduce simplify early fixes const folding

geohot:


geohot:
oh #12397 looks better

PR #12416, opened 2025-10-02
rangeify: fix test_where_fold (llvm)

b1tg:


tinygrad/renderer/llvmir.py
```diff
@@ -92,6 +92,10 @@ def render_wmma_amd(ctx, wmma: UOp, cdna=False) -> str:
                                                             f", {ldt(u.dtype)} {ctx[u]}, i32 {i}" for i,u in enumerate(x.src)])),
   # unary/binary/ternary ops
   (UPat(Ops.BITCAST, name="x"), lambda ctx,x: f"  {ctx[x]} = bitcast {ldt(x.src[0].dtype)} {ctx[x.src[0]]} to {ldt(x.dtype)}"),
+  # rewrite cast to bool to CMPNE 0
+  # NOTE: this should be a rewrite rule in extra_matcher but it fails test_where_fold with RANGEIFY
```
chenyuxyz: if it fails test_where_fold with RANGEIFY, why should this be a rewrite rule in extra_matcher?

does not sound like a well understood decision to move this

b1tg: the old rewrite rule in extra_matcher remove Ops.CAST so Ops.WHERE optimization in pm_render can't hit, I move the logic to string_rewrite to fix it so I don't have to mess with the rewrite order which may effects other backends

chenyuxyz: i think it's better to justify directly why this should be here, or don't comment at all

the test and RANGEIFY will eventually be gone

PR #12418, opened 2025-10-02
test case for a long rand chain

chenyuxyz:
currently failing with RANGEIFY because device propogates too deep


chenyuxyz:
included in #12419

PR #12419, opened 2025-10-02
ops: n^2 .device property fix

Qazalin:


chenyuxyz:
hmm python does not support tail call optimization...? cool thanks

PR #12430, opened 2025-10-03
skip copies of reshaped buffers

geohot:


geohot:
Issue with adding contiguous here @Qazalin

PR #12436, opened 2025-10-03
undo loaded comparison swap

S-Lykles:


S-Lykles:
Fixes webgpu long problem in #12413

PR #12442, opened 2025-10-03
fix bmnist torch with RANGEIFY=1

Qazalin:


geohot:
Can you include a comment and test with this?

Qazalin:
test_assign_changes_alt isolated the wrong behavior, ASSIGN to unrealized Tensor should always behave like replace. This was a bug before rangeify too, because of something wrong in the old becomes_map, it was never revealed.

Post tensor map for rangeify=0 vs with rangeify (same input graph). The ASSIGN Tensors correctly update in RANGEIFY and reveal the bug.
image
image

PR #12446, opened 2025-10-03
post index symbolic

S-Lykles:


S-Lykles:
Fixes #12238, also needed for #12413

PR #12465, opened 2025-10-06
use recursive_property to prevent RecursionError

geohot:


YugandharrPatil:
come stream brooooooo. and this time do something simple and interesting that our noobie brains could comprehend :P

PR #12468, opened 2025-10-06
rangeify: bufferize reduces

nimlgen:
cifar a bit better, but still slower.

rangeify can chain any number of reduces leading to slow kernels


geohot:
why are the kernels slow if the reduces are chained? this is a feature for the most part

nimlgen:
I mean nested
```
x = Tensor.zeros(512, 32, 32, 32).contiguous().realize()
x = x.sum(3).max(2)
```

should be something like split reduce, but these are 2 reduces.

geohot:
```
x = Tensor.zeros(512, 32, 32, 32).contiguous().realize()
x = x.sum((2, 3))
```
should not be split, so why should it matter if one is a sum and one is a max?

PR #12486, opened 2025-10-07
vars in gated_read_image_count

nimlgen:


chenyuxyz:
oh is this why openpilot is slower? it's not gated though? can i be simplified with oob->0 rules?

nimlgen:
yeah, the ones I've seen are from `map_pad`

PR #12542, opened 2025-10-09
all_same fast path

S-Lykles:


tinygrad/helpers.py
```diff
@@ -24,7 +24,7 @@ def argfix(*x):
     return tuple(x[0])
   return x
 def argsort(x): return type(x)(sorted(range(len(x)), key=x.__getitem__)) # https://stackoverflow.com/questions/3382352/equivalent-of-numpy-argsort-in-basic-python
-def all_same(items:tuple[T, ...]|list[T]): return all(x == items[0] for x in items)
+def all_same(items:tuple[T, ...]|list[T]): return all(x is items[0] for x in items) or all(x == items[0] for x in items)
```
chenyuxyz: `all(x is items[0] or x == items[0])`? otherwise every time it's not all_same it loops twice

PR #12549, opened 2025-10-09
remove on_stack from graph_rewrite

geohot:
If it's faster, why?


geohot:
Why does this break things?

PR #12550, opened 2025-10-09
remove trivial use of RANGEIFY flag

chenyuxyz:
some tests need update still


chenyuxyz:
@Qazalin can you update test_schedule to be independent of RANGEIFY? also update expected failure if it's not expected to fail

Issue #11183, opened 2025-07-11
Whisper Example Seems to breaking.


shashwatmi:
I just ran whisper example from repo. It crashed with an error. 

```
ram used:  0.08 GB, decoder.mask                                      : 100%|██████████████████████████████████████████████████████████████████████████████| 168/168 [00:01<00:00, 128.33it/s]
loaded weights in 1309.49 ms, 0.08 GB loaded at 0.06 GB/s
1 error generated.
Traceback (most recent call last):
  File "/home/dashgrim/Hobby/tinygrad/examples/whisper.py", line 303, in <module>
    print(transcribe_file(model, enc, sys.argv[1]))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dashgrim/Hobby/tinygrad/examples/whisper.py", line 237, in transcribe_file
    return transcribe_waveform(model, enc, [load_file_waveform(filename)])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dashgrim/Hobby/tinygrad/examples/whisper.py", line 275, in transcribe_waveform
    if all(len(c) == len(ctx[0]) for c in ctx): ctx = inferloop(np.array(ctx), encoded_audio)
                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dashgrim/Hobby/tinygrad/examples/whisper.py", line 251, in inferloop
    next_tokens = model.decoder(Tensor(next_tokens), pos, encoded_audio)[:, -1].argmax(axis=-1).numpy().astype(np.int32).reshape(-1, 1)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dashgrim/Hobby/tinygrad/examples/whisper.py", line 108, in __call__
    return self.getjitted[x.shape](x, pos, encoded_audio)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dashgrim/Hobby/tinygrad/.venv/lib/python3.11/site-packages/tinygrad/engine/jit.py", line 268, in __call__
    ret = self.fxn(*args, **kwargs)
[...]

PR #12574, opened 2025-10-09
uop: rewrite sqrt(f16|bf16) via f32; drop OpenCL override

### LINKED ISSUE #12237: `error: call to 'sqrt' is ambiguous` while running `whisper.py` on OpenCL
When running `CL=1 TINY=1 DEBUG=6 python3 examples/whisper.py test/models/whisper/test.wav`, I encounter the following OpenCL compilation error:

```sh
tinygrad.device.CompileError: OpenCL Compile Error

<kernel>:23:27: error: call to 'sqrt' is ambiguous
    *(data0_2+gidx0) = (1/sqrt((((half)(((*(acc1+0))*0.0026041666666666665f)))+((half)(1e-05f)))));
                          ^~~~
cl_kernel.h:1763:25: note: candidate function
double __OVERLOADABLE__ sqrt(double);
                        ^
cl_kernel.h:6288:24: note: candidate function
float __OVERLOADABLE__ sqrt(float in);
                       ^
cl_kernel.h:1764:25: note: candidate function
float2 __OVERLOADABLE__ sqrt(float2);
                        ^
[... additional candidate functions for float3, float4, float8, float16, double2, double3, double4, double8, double16 ...]
```

The error indicates an ambiguous call to sqrt due to multiple overloaded versions in OpenCL. To resolve this, I applied the following fix, which works but may not be optimal:

```diff
diff --git a/tinygrad/renderer/cstyle.py b/tinygrad/renderer/cstyle.py
index 09d25c082..2256942cd 100644
--- a/tinygrad/renderer/cstyle.py
+++ b/tinygrad/renderer/cstyle.py
@@ -251,6 +251,8 @@ class OpenCLRenderer(CStyleLanguage):
   smem_prefix = "__local "
   barrier = "barrier(CLK_LOCAL_MEM_FENCE);"
   float4 = "(float4)"
+  code_for_op = { **CStyleLanguage.code_for_op,
+    Ops.SQRT: lambda x,dtype: f"sqrt((float){x})" if dtype in (dtypes.half, dtypes.bfloat16) else f"sqrt({x})" }
   code_for_workitem = {"g": lambda x: f"get_group_id({x})", "l": lambda x: f"get_local_id({x})", "i": lambda x: f"get_global_id({x})"}
   type_map = { dtypes.int8: "char", dtypes.uint8: "uchar", dtypes.uint32: "uint", dtypes.uint16: "ushort", dtypes.uint64: "ulong",
               dtypes.bfloat16: "ushort" }
```

This fix explicitly casts the input to sqrt as float for half and bfloat16 data types to disambiguate the call. However, I'm unsure if this is the best approach. This issue seems related to #11183. Can you provide feedback on the fix and suggest any improvements?

shua-ie:
Moves sqrt promotion for half/bfloat16 into UOps decompositions.

For f16/bf16: rewrites to cast→sqrt(f32)→cast
For f32: unchanged

Removes OpenCL code_for_op override so all renderers inherit the IR behavior.
Adds test_sqrt_cast_f16_bf16_to_f32 asserting the UOp pattern.

Fixes ambiguous sqrt calls on OpenCL devices without cl_khr_fp16.

Fixes #12237, #11183.

### DIFF
--- a/test/test_uop_graph.py
+++ b/test/test_uop_graph.py
@@ -21,20 +21,25 @@ def to_uops_list(u:List[UOp]) -> List[UOp]:
   assert ret[-1].op is Ops.SINK
   return ret[:-1]
 
-def test_sqrt_cast_f16_bf16_to_f32():
-  from tinygrad.uop.decompositions import get_late_rewrite_patterns
-  pm = get_late_rewrite_patterns((Ops.SQRT,))
-  def rew(dt):
-    x = UOp(Ops.DEFINE_VAR, dt)
-    return graph_rewrite(UOp(Ops.SQRT, dt, (x,)), pm)
-  for dt in (dtypes.float16, dtypes.bfloat16):
-    r = rew(dt)
-    assert r.op is Ops.CAST and r.dtype == dt
-    s = r.src[0]
-    assert s.op is Ops.SQRT and s.dtype is dtypes.float32
-    assert s.src[0].op is Ops.CAST and s.src[0].dtype is dtypes.float32
-  r = rew(dtypes.float32)
-  assert r.op is Ops.SQRT and r.dtype is dtypes.float32
+class TestDecompositions(unittest.TestCase):
+  def test_sqrt_cast_f16_bf16_to_f32(self):
+    from tinygrad.uop.decompositions import get_late_rewrite_patterns
+    pm = get_late_rewrite_patterns((Ops.SQRT,))
+    def rew(dt):
+      x = UOp(Ops.DEFINE_VAR, dt)
+      return graph_rewrite(UOp(Ops.SQRT, dt, (x,)), pm)
+    for dt in (dtypes.float16, dtypes.bfloat16):
+      r = rew(dt)
+      self.assertIs(r.op, Ops.CAST)
+      self.assertIs(r.dtype, dt)
+      s = r.src[0]
+      self.assertIs(s.op, Ops.SQRT)
+      self.assertIs(s.dtype, dtypes.float32)
+      self.assertIs(s.src[0].op, Ops.CAST)
+      self.assertIs(s.src[0].dtype, dtypes.float32)
+    r = rew(dtypes.float32)
+    self.assertIs(r.op, Ops.SQRT)
+    self.assertIs(r.dtype, dtypes.float32)
 
 class TestGraphRewriteConst(unittest.TestCase):
   def test_gep_const(self):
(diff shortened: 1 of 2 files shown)


geohot:
Was there a failing test this fixes?

shua-ie:
> Was there a failing test this fixes?

@geohot 
Not fixing a CI failure—this is portability hardening. On OpenCL 1.2 without `cl_khr_fp16`, `sqrt(half)` is rejected/ambiguous, and bf16 has no overload. Upcasting half/bf16 to float in codegen follows tinygrad’s “sub-32-bit transcendentals in f32” convention (same as Metal), leaves f32 unchanged, and avoids compile errors on those drivers.

geohot:
What's the output from this? I think if you want this cast, it should be in the UOps, not hacked in as a string

shua-ie:
> What's the output from this? I think if you want this cast, it should be in the UOps, not hacked in as a string

Done—moved the cast into UOps. For f16/bf16 the IR rewrites `SQRT(dt, x)` → `CAST(dt, SQRT(f32, CAST(f32, x)))`; f32 stays `SQRT(f32, x)`. Implemented in `get_late_rewrite_patterns`; removed the OpenCL override. Added `test_sqrt_cast_f16_bf16_to_f32`.

test/test_uop_graph.py
```diff
@@ -21,6 +21,21 @@ def to_uops_list(u:List[UOp]) -> List[UOp]:
   assert ret[-1].op is Ops.SINK
   return ret[:-1]
 
+def test_sqrt_cast_f16_bf16_to_f32():
```
wozeparrot: how is this tested? its not under a testcase and doesn't seem to be called anywhere

tinygrad/renderer/cstyle.py
```diff
@@ -255,6 +255,7 @@ class OpenCLRenderer(CStyleLanguage):
   type_map = { dtypes.int8: "char", dtypes.uint8: "uchar", dtypes.uint32: "uint", dtypes.uint16: "ushort", dtypes.uint64: "ulong",
               dtypes.bfloat16: "ushort" }
 
+  code_for_op = {**CStyleLanguage.code_for_op}
```
wozeparrot: why is this needed?

test/test_uop_graph.py
shua-ie: Put the check under a test class `TestDecompositions` in this file and switched to `assertIs` for ops/dtypes. Pytest discovers and runs it alongside the rest.

tinygrad/renderer/cstyle.py
shua-ie: No longer needed—`SQRT` promotion is handled in UOps. Removed `OpenCLRenderer.code_for_op` so it inherits `CStyleLanguage`.

wozeparrot:
this also applies outside of cl? I'm not sure we want that

geohot:
This is not just on OpenCL, it's everywhere. Feel free to reopen with narrower scope and test that tests functionality instead of exact uops

PR #12575, opened 2025-10-09
update tiny torch backend hook

0xbeedee:
In working towards the bounty for making hlb cifar10 work with the tiny torch backend (#12576) , I noticed that the current backward hook raised a `FutureWarning: Using a non-full backward hook when the forward contains multiple autograd [...]`.

Using the full backward hook raised `UserWarning: Full backward hook is firing when gradients are computed with respect to module outputs since no inputs require gradients`, so I opted for using a param hook.


geohot:
When does this fire? I don't see anything related to backward

0xbeedee:
The `FutureWarning` fires with both `extra/torch_backend/example.py`and `examples/other_mnist/beautiful_mnist_torch.py`, when they're run with `TINY_BACKEND=1`.

The `UserWarning` (full backward hook) fires when running `examples/other_mnist/beautiful_mnist_torch.py`, and when running the torchified hlb cifar10 script in #12576 (both with `TINY_BACKEND=1`). It does not fire when running `extra/torch_backend/example.py`.

chenyuxyz:
i see this e.g. https://github.com/tinygrad/tinygrad/actions/runs/18532931438/job/52820308626#step:5:23

and this seems to fix it

PR #12576, opened 2025-10-09
[bounty] hlb CIFAR10 works with tiny torch backend

0xbeedee:
I've used the same structure as in the MNIST examples (a tinygrad-specific python script, and a tiny torch-specific python script, in a separate directory), hence the apparently large changes. 

The code is relatively straightforward, as it's mostly a "torchification" of `hlb_cifar10.py`.

### DIFF
--- a/examples/other_cifar/hlb_cifar10_torch.py
+++ b/examples/other_cifar/hlb_cifar10_torch.py
@@ -170,7 +170,6 @@ hyp = {
 }
 
 
-@Context(FUSE_ARANGE=getenv("FUSE_ARANGE", 1))
 def train_cifar():
   def set_seed(seed):
     torch.manual_seed(seed)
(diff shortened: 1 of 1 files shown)


geohot:
The idea of this bounty is that it has to work with the original file, just with the device changed.

0xbeedee:
I see, thanks for the heads-up!

I thought that the point was for the tiny torch backend to act as a drop-in replacement for torch itself, i.e., that you'd need to write torch-compatible code and make sure it runs with `TINY_BACKEND=1`. In which case testing the original file is guaranteed to work because it's pure tinygrad?

Sorry if I'm not making any sense... I'll get to work on the changes as soon as able.

I noticed a massive slowdown in running both my script and the original one after merging the new updates from master, by the way: is this a known issue?

0xbeedee:
Nevermind, I think I know what you mean now! Closing this PR, will add a new one with the desired changes when completed.

PR #12581, opened 2025-10-10
cleanup UOp.st [pr]

Qazalin:
it should be UOp.shape mostly and a tiny bit of ShapeTracker for wherever we still want .st


Qazalin:
First step here is to get the new shape from args and src[0] for every MovementOp.
https://github.com/Qazalin/tinygrad/blob/c026a41f8f442f0afef5e052c7924302e176cf15/tinygrad/uop/ops.py#L194-L196
Then removing .st is easy, since the rest of the ops just use ShapeTracker.from_shape()

Qazalin:
The shape property change is doing most of this already.

PR #12584, opened 2025-10-10
remove some View add logic [pr]

chenyuxyz:
no longer simplify the case of v0+v1 where v0 has a mask


chenyuxyz:
this might result in more views stack in a shapetracker. i'd expect generated kernels to be the same given that it's only used in is_expanded in heuristic, and rewrite should still simplifies

no visible regression in benchmark running time and step times

PR #12586, opened 2025-10-10
drop invalid clauses not used in index

S-Lykles:


S-Lykles:
Bert gpu step time went from 326 ms to 319ms
Mac Stable Diffusion v2: 4000ms -> 3800ms

S-Lykles:
Merged in #12635

PR #12588, opened 2025-10-10
ttomsa's "new linearizer" that replaces the linearizer with toposort

geohot:
work on understanding #12459

### DIFF
--- a/tinygrad/codegen/__init__.py
+++ b/tinygrad/codegen/__init__.py
@@ -14,7 +14,7 @@ from tinygrad.uop.decompositions import get_late_rewrite_patterns
 from tinygrad.codegen.late.expander import migrate_indexing, expander, pm_pre_expander, pm_group_for_reduce
 from tinygrad.codegen.late.devectorizer import load_store_folding, load_store_indexing, devectorize, pm_reduce, \
   ReduceContext, correct_load_store, pm_render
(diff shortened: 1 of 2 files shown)


geohot:
@ttomsa made a few minor changes here for viz and ran benchmark CI.

I like the fundamental idea of replacing linearizer with toposort.

looks like speed regressions in the BEAMed kernels:
https://github.com/tinygrad/tinygrad/actions/runs/18399305845/job/52424766461
https://github.com/tinygrad/tinygrad/actions/runs/18399305845/job/52424766483

I would rather ENDRANGE be src=(producer, range to end, other...)

The groupbys are hard for me to read.

Have you tested this with BEAM? I know there's some BEAMed kernels the old linearizer would fail on

test/unit/test_block_reorder.py
```diff
@@ -54,16 +54,16 @@ def test_loads(self):
...
-    golden = block_reorder(list(sink.toposort()))
+    golden = schedule(list(sink.toposort()))
 
     # render for test
     print(self._test_render(golden))
     #print_uops(golden)
 
     # assert the loads are in this order
     self.assertListEqual([g.src[0].src[1].render() for g in golden if g.op is Ops.LOAD],
-                         ['(gidx1*4)', '((gidx1*4)+1)', '((gidx1*4)+2)', '((gidx1*4)+3)',
```
geohot, line 65: Why did the order change?

PR #12592, opened 2025-10-10
uop_given_valid cleanup

S-Lykles:


tinygrad/uop/symbolic.py
```diff
@@ -397,8 +397,8 @@ def parse_valid(valid:UOp) -> tuple[UOp, bool, int]:
   if valid.op is Ops.CMPLT and dtypes.is_int(valid.src[0].dtype): return valid.src[0], True, int((valid.src[1]).vmax)-1
   raise ValueError(f"not able to parse {valid=}")
 
-def uop_given_valid(valid:UOp, uop:UOp) -> UOp|None:
-  # return None if valid is always False, otherwise the simplified uop (might be the same as input)
```
chenyuxyz, line 401: if valid is always False is not a concern anymore, i think it's better to make this function return None if nothing is simplified like other rewrite functions

PR #12602, opened 2025-10-10
minimal new linearizer

geohot:
based off #12459

some changes I made:
* switch `schedule` to `linearize` which calls the toposort in the function
* rename the patternmatchers to `pm_add_endrange_endif` and `pm_control_flow`
* fix the `ranges` method on UOp to respect ENDRANGE

overall feedback:
* to merge a PR, the smaller it is the easier it is to merge. i'd split this into three PRs. 
  - PR to fix any tests where the test itself is broken
  - PR to add new functionality. ideally all tests should be fixed with the first PR
  - PR to clean up and remove old functionality
* the groupbys are hard to read. maybe include a comment explaining them or refactor to be easier.
* if you absolutely need a type: ignore, at minimum it must be justified with a comment

### DIFF
--- a/tinygrad/codegen/late/control_flow.py
+++ b/tinygrad/codegen/late/control_flow.py
@@ -1,0 +1,103 @@
+from tinygrad.uop.ops import UOp, Ops, PatternMatcher, UPat
+from tinygrad.helpers import dedup
+from collections import defaultdict
+from itertools import groupby
+from functools import reduce
+import heapq
+
+def linearize(u:UOp) -> list[UOp]:
+  lst = list(u.toposort())
+  in_this_block = set(lst)
+  local_children: defaultdict[UOp, list[UOp]] = defaultdict(list)
+  in_degree:dict[UOp, int] = {}
+  priorities:dict[UOp, int] = {}
+
+  # get local children and assign priorities
+  # NOTE: this requires the lst be locally toposorted
+  for u in reversed(lst):
+    in_degree[u] = 0
+    for s in u.src:
+      if s in in_this_block:
+        local_children[s].append(u)
+        in_degree[u] += 1
+    # put loads in the beginning of the block and prevent priority inversion. hack for BARRIER grouping too
+    priority = [0] + [priorities[x] for x in local_children[u]]
+    if u.op is Ops.LOAD: priority.append(-1000)
+    if u.op is Ops.BARRIER: priority.append(-1500)
+    # ranges are scheduled as late as possible so anything that can be outside is
+    #if u.op is Ops.RANGE: priority = [2000]
+    # move defines and consts to the top
+    if u.op in {Ops.DEFINE_GLOBAL, Ops.DEFINE_LOCAL, Ops.DEFINE_REG, Ops.DEFINE_VAR, Ops.SPECIAL, Ops.CONST}: priority.append(-2000)
+    priorities[u] = min(priority)
+
+  # number the uops in "ideal" order
+  nkey = {u:i for i,u in enumerate(sorted(lst, key=lambda x: (priorities[x],)+x.tuplize))}
+
+  # then force then to be toposorted in as close to the ideal order as possible
+  heapq.heapify(heap:=[(nkey[u],u) for u in lst if in_degree[u] == 0])
+  newlst = []
+  while heap:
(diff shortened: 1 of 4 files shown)


geohot:
local test failures on my machine

```
FAILED test/test_linearizer.py::TestLinearizer::test_simple_unroll_no_between_phi_dependencies - assert 115 < 115
FAILED test/test_linearizer.py::TestLinearizer::test_grouped_store_locals_and_globals - assert 0 == 1
FAILED test/test_uop_graph.py::TestIFUOps::test_expand_ifs_one_gate - AssertionError: 5 != 2
FAILED test/test_uop_graph.py::TestUOpGraph::test_switched_range_order - IndexError: list index out of range
FAILED test/test_uops_stats.py::TestStatsOptimized::test_gemm_group - AssertionError: 65536 != 589824
```

tinygrad/codegen/late/control_flow.py
```diff
@@ -0,0 +1,103 @@
...
+  assert len(newlst) == len(lst), f"len mismatch {len(newlst)} != {len(lst)}"
+  return newlst
+
+def add_endrange(x:UOp):
+  if not ((x.op is Ops.LOAD and x.src[-1].op is Ops.STORE) or all(s.op is Ops.STORE and any(n.op is Ops.RANGE for n in s.src) for s in x.src)):
+    return None
+  src: list[UOp] = []
+  for k,g in groupby(x.src, key=lambda k: tuple(dedup(s for s in k.src if s.op is Ops.RANGE))):
+    if not k: src.extend(g)
+    else: src.extend(reduce(lambda acc,rng: (UOp(Ops.ENDRANGE, src=(rng,) + acc),), reversed(k), tuple(g))) # type: ignore
```
geohot, line 54: this type: ignore needs to be removed. why is the type checker failing here?

tinygrad/uop/spec.py
```diff
@@ -190,17 +191,16 @@ def validate_store(idx:UOp, val:UOp, gate:UOp|None=None):
...
   # WMMA has a <a, b, acc>
   (UPat(Ops.WMMA, src=(UPat(), UPat(), UPat()), name="x"), lambda x: isinstance(x.arg, tuple) and len(x.arg) == 8),
   (UPat(Ops.CONTRACT, name="x"), lambda x: x.dtype.count == prod(y[1] for y in x.arg)),
   (UPat(Ops.UNROLL, name="x"), lambda x: x.src[0].dtype.count == prod(y[1] for y in x.arg)),
 
   # if has a <gate, barrier?>
-  (UPat(Ops.IF, dtype=dtypes.void, src=(UPat(),)), lambda: True),
-  (UPat(Ops.IF, dtype=dtypes.void, src=(UPat(), UPat(Ops.BARRIER))), lambda: True),
-  (UPat(Ops.ENDIF, dtype=dtypes.void, src=(UPat(Ops.IF),)), lambda: True),
+  (UPat(Ops.IF, dtype=dtypes.void, src=(UPat(),), allow_any_len=True), lambda: True),
```
geohot, line 202: Can this spec have the second arg of IF always be BARRIER? The way it was in the source PR the second rule was never triggered.

geohot:
These are the most concerning test failures. Only in benchmark, a BEAMed GEMM speed regression.

On NVIDIA:
https://github.com/tinygrad/tinygrad/actions/runs/18402349283/job/52434347790?pr=12602
FAILED test/external/speed_v_theoretical.py::TestKernelSpeed::test_gemm_8192 - AssertionError: 22.95480445711407 not greater than 125

On AMD:
https://github.com/tinygrad/tinygrad/actions/runs/18402349283/job/52434347846?pr=12602
FAILED test/external/speed_v_theoretical.py::TestKernelSpeed::test_gemm_8192 - AssertionError: 10.536346876597923 not greater than 60

It seems like there's a major speed regression in the BEAMed kernels. Note that if the linearizer fails, the BEAM will silently skip that kernel. Can you find what's causing the regression, fix it, and add a test for it to CI?

tinygrad/codegen/late/control_flow.py
ttomsa: I think it's just misbehaving cause of the lambda. It seems to think acc can be type `Iterable[UOp]` but it is explicitly made a tuple in the initial case and in the lambda

PR #12612, opened 2025-10-11
uop_given_valid uses less simplify

S-Lykles:


S-Lykles:
`devectorize` in `PYTHONPATH=. python3 test/test_multitensor.py  TestMultiTensor.test_data_parallel_resnet`
went from 23s to 9s on my laptop

all of the `Run pytest` are 10-20s faster

PR #12616, opened 2025-10-11
hcq: use exception group

nimlgen:


nimlgen:
oh, I thought we have bumped to 3.11. closing till 3.11

PR #12635, opened 2025-10-13
drop invalid conditions if the ranges are not used inside the index

S-Lykles:


S-Lykles:
From #12586:
Bert gpu step time went from 326 ms to 319ms
Mac Stable Diffusion v2: 4000ms -> 3800ms

PR #12630, opened 2025-10-13
pylint error

b1tg:
pylint just update to [4.0.0](https://github.com/pylint-dev/pylint/releases/tag/v4.0.0), new version found indent error in ops_cpu. I guess pip packages in master ci be cached so not trig the error
```
$ python -m pylint --disable=all -e W0311 -e C0303 --jobs=0 --indent-string='  ' --recursive=y .        
************* Module /home/b1tg/tinygrad/.pylintrc                                                                                 
.pylintrc:1:0: E0015: Unrecognized option found: suggestion-mode (unrecognized-option)                                             
************* Module tinygrad.runtime.ops_cpu                                                                                      
tinygrad/runtime/ops_cpu.py:108:0: W0311: Bad indentation. Found 3 spaces, expected 4 (bad-indentation)                            
tinygrad/runtime/ops_cpu.py:109:0: W0311: Bad indentation. Found 3 spaces, expected 4 (bad-indentation)    
```

PR #12642, opened 2025-10-13
cpu: fix spacing

nimlgen:


wozeparrot:
oop, i merged #12630 which fixes the spacing, don't think we need to pin pylint?

nimlgen:
I think I’ve seen way more linter errors in 4.0

PR #12656, opened 2025-10-14
adds loop splitting for cat

ExpressGradient:
- Enables optional loop splitting for high-level `Tensor.cat` so wide concatenations no longer pad the entire axis at once. 
- Added a `CAT_LOOP_SPLIT` knob, factored the chunked concat helpers, and covered the split path with a regression test (test_cat_loop_split_threshold).
- Verified ruff, mypy, and the usual cat scenarios still pass.


geohot:
This shouldn't be in Tensor. I removed this bounty, it wasn't clear what "high level" meant.

PR #12659, opened 2025-10-14
remove assign contiguous hack

geohot:


geohot:
@Qazalin don't do things like this. this copies nonsense behavior (assign to contiguous is that buffer) to make an incorrect torch backend work. even disabling the torch backend is better than adding hacks.

if the spec is wrong, spending time to match a broken spec is a bad idea. the correct thing to do here is to fix the torch backend

Issue #11495, opened 2025-08-04
torch.linalg.matrix_rank returns incorrect result for identity matrices


TiborGY:
Running `torch.linalg.matrix_rank` sometimes yields incorrect results with the tinygrad backend.
Eg. 2x2 identity matrix, should return 2, tinygrad returns 0 instead.

```
>>> A = torch.eye(2)
Dispatch Log: aten.eye.default
>>> torch.linalg.matrix_rank(A)
Dispatch Log: aten.empty.memory_format
Dispatch Log: aten.zeros.default
Dispatch Log: aten.full.default
Dispatch Log: aten._linalg_svd.default
Dispatch Log: aten.slice.Tensor
Dispatch Log: aten.unsqueeze.default
Dispatch Log: aten.mul.Tensor
Dispatch Log: aten.unsqueeze.default
Dispatch Log: aten.maximum.default
Dispatch Log: aten.gt.Tensor
Dispatch Log: aten.sum.dim_IntList
tensor(2)
>>> torch.linalg.matrix_rank(A.tiny())
Dispatch Log: aten._to_copy.default
empty_strided size=[2, 2] stride=[2, 1] dtype=torch.float32 layout=torch.strided device=device(type='tiny', index=0) pin_memory=False
Dispatch Log: aten.empty.memory_format
empty.memory_format size=[] dtype=torch.int64 layout=torch.strided device=device(type='tiny', index=0) pin_memory=False memory_format=None
Dispatch Log: aten.zeros.default
empty.memory_format size=[] dtype=torch.float64 layout=torch.strided device=device(type='tiny', index=0) pin_memory=False memory_format=None
zero_ torch.Size([])
Dispatch Log: aten.full.default
empty.memory_format size=[] dtype=torch.float64 layout=torch.strided device=device(type='tiny', index=0) pin_memory=False memory_format=None
fill_.Scalar torch.Size([]) 2.384185791015625e-07
Dispatch Log: aten._linalg_svd.default
*** CPU     2948 copy       16,     CPU <- NPY                arg  2 mem  0.00 GB tm     43.19us/    26.21ms (     0.00 GFLOPS    0.0|0.0     GB/s)
*** CPU     2949 E_4n24                                       arg  2 mem  0.00 GB tm     13.09us/    26.22ms (     0.00 GFLOPS    0.0|0.0     GB/s) ['empty', 'contiguous']
[...]

PR #12667, opened 2025-10-14
Added Matrix_rank to tensor, issue 11495

SamE-345:
#11495
Implements a matrix rank function in the tensor class, by using SVD to decompose the matrix into ranks, then summing the number of singular values greater than the tolerance. 
Tests added include on Identity matrices, where the original issue was, and also on rectangular matrices.


chenyuxyz:
not through numpy

PR #12671, opened 2025-10-14
CPU: honor LLVM=1 when selecting backend

ExpressGradient:
### Change
  - Make `LLVM=1` behave like `CPU_LLVM=1` so the CPU LLVM backend is selected automatically.
  - When no explicit device override is set, prefer the CPU device if LLVM is present and ensure child processes inherit the same choice.

### Context
  - Fixes macOS runs of `BEAM=3 LLVM=1 LLVMOPT=1 pytest test/speed/external_test_speed_v_torch.py::TestSpeed::test_sum`, which previously fell back to METAL and failed.


geohot:
Do not use AI

PR #12694, opened 2025-10-15
late merging of where and load

S-Lykles:
Needed to get range/reduce splitting to work


S-Lykles:
Quantized llama is twice as fast! It had some aranges that werent folding

PR #12718, opened 2025-10-16
feat: add thunderkittens 4090 fa2

wozeparrot:


wozeparrot:
seems like nvrtc can't actually compile this since it doesn't have some c++ headers...

PR #12729, opened 2025-10-16
no broadcasting in reduce collapse

S-Lykles:


S-Lykles:
Since reduce collapse is before the expander there wont ever be any gep or vectors

PR #12734, opened 2025-10-16
Only import LLVM with RTLD_GLOBAL when using mesa

sirhcm:


nimlgen:
the same crash for me

sirhcm:
I can't reproduce the crash on my machine (macos 14.7.8)

nimlgen:
yeah I can't repro this now (in normal usage), but opening CPU with LVP and then METAL will crash, no?

sirhcm:
Yeah, closing in favor of #12739.

PR #12738, opened 2025-10-16
Revert "use RTLD_GLOBAL on macos"

sirhcm:
Breaks python3 -m tinygrad.device, but (in theory) fixes metal segfaults?

Will try to fix both today, but tinygrad.device being broken is a lesser evil than metal segfaulting.


chenyuxyz:
seems to help, thanks!

PR #12740, opened 2025-10-16
onnx helper intermediate node output validation

### LINKED ISSUE #12572: onnx helper to compare intermediate values
attempt in https://github.com/tinygrad/tinygrad/pull/11707 was inadequate. 
It'd be best to just have something that outputs a full list of node outputs of both ORT and tinygrad, and do the comparison from there.
Editing the ModelProto to return all the node outputs using an [actual library](https://docs.nvidia.com/deeplearning/tensorrt/latest/_static/onnx-graphsurgeon/index.html) would probably be the way.

related to #11705

geohotstan:
closes #12572
pretty usable imo, though I'm not sure if it's good enough for others.
They are useful for debugging if nodes are giving clearly wrong outputs or detecting numerical imprecision spikes. 

One worry is the `Tensor.realize(*tinygrad_out.values())` in `_get_tinygrad_and_ort_np_outputs` giving different scheduling behavior, which should only cause minor numerical differences maybe. 
Currently the different scheduling occasionally result in wildly different output.

Without realizing them together, validating a model's intermediate outputs takes forever.

### DIFF
--- a/extra/onnx_helpers.py
+++ b/extra/onnx_helpers.py
@@ -10,6 +10,9 @@ def get_example_inputs(graph_inputs:dict[str, OnnxValue], config={}):
   """
   Generate example input tensors based on the provided ONNX graph input specifications.
 
+  NOTE: This is not guaranteed to be reliable. It's a best-effort helper
+  that uses heuristics to guess input shapes and values.
+
   Example:
     from tinygrad.nn.onnx import OnnxRunner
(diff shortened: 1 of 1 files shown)


chenyuxyz:
given it's a helper, i will merge it as along as you think it's useful

do you have more examples that scheduling changes output drastically?

PR #12746, opened 2025-10-16
Mesa: freedreno

sirhcm:


geohot:
Looks active and we still want this, will keep open

PR #12747, opened 2025-10-17
[Bounty] Remove Tensor._pool alternative implementation and verify kernels remain the same

usatie:
Two issues found simply removing alternative implementation. 

The first issue was that `test/test_ops.py::TestOps::test_unfold` failed because `_pool` was not properly handling the edge cases when `kernel = 0` or `stride > input shape`. Fixed this by handling these. 

The second issue is that the backward pass for the pooling kernel taking too long with python backend. The forward pass kernel looks identical from the alternative implementation, but the backward pass kernel doesn't. Especially slow for these tests:
- `PYTHON=1 python -m pytest test/test_ops.py::TestOps::test_global_avg_pool2d` (1.28s vs [almost never ends])
- `PYTHON=1 python -m pytest test/test_ops.py::TestOps::test_avg_pool2d_asymmetric_padding` (2.43s vs 149.98s)

`DEBUG=4 python -m pytest test/test_ops.py::TestOps::test_global_avg_pool2d -rP` on master vs this branch:
Screenshot 2025-10-17 at 6 01 42 PM

```
$ git checkout master
$ DEBUG=5 python -m pytest test/test_ops.py::TestOps::test_global_avg_pool2d -rP > cmp
$ git checkout usatie:remove-tensor-pool-alternative
$ DEBUG=5 python -m pytest test/test_ops.py::TestOps::test_global_avg_pool2d -rP > out
$ diff cmp out
```
Screenshot 2025-10-20 at 2 51 36 PM

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -2556,7 +2556,7 @@ class TestOps(unittest.TestCase):
       with self.subTest(padding=p):
         helper_test_op([shape],
           lambda x: torch.nn.functional.avg_pool2d(x, kernel_size=(5,5), padding=1),
-          lambda x: Tensor.avg_pool2d(x, kernel_size=(5,5), padding=1), rtol=1e-5, forward_only=True)
+          lambda x: Tensor.avg_pool2d(x, kernel_size=(5,5), padding=1), rtol=1e-5)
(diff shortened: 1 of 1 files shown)


usatie:
@chenyuxyz Hi, the bounty issue is to `verify kernels remain the same`, but do I need to fix this (backward pass kernel is different and slow)?

PR #12770, opened 2025-10-17
remove a contiguous in LARS

chenyuxyz:


chenyuxyz:
trains resnet similarly

PR #12773, opened 2025-10-17
use torch 2.9 and its Muon in test

chenyuxyz:


chenyuxyz:
@kevvz fyi in case you know what's up

PR #12775, opened 2025-10-17
openpilot compile tests: assert pre-rangify speeds

haraschax:
This asserts the speeds are similar to before Rangify was made the default. Once speedups are merged to make this pass, openpilot can bump tinygrad again.


geohot:
Are we good on this @chenyuxyz ?

PR #12781, opened 2025-10-18
In-tree autogen

sirhcm:
Notes:

 * `nvjitlink`: versioned functions (eg. `nvJitLinkCreate`) are marked as static in the header, but clang2py parsed them as extern, and they just so happened to be exported. A hard coded header is at `extra/nvJitLink.h` now.
 * metal is 99% autogen: objc runtime (`objc_sendMsg`, `objc_getClass`, and `sel_registerName`) and libdispatch (`dispatch_data_create`) remain, but more significantly objc string manipulation. objc strings require support for [categories](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocCategories.html).
 * `if TYPE_CHECKING` blocks because mypy can't understand metaclasses.
 * All autogen is tested in CI now
 * Should `autogen/__init__.py` and `autogen/autogen.py` count towards the line limit?
 * Eventually, support for autogen that depends on other autogen would be nice (eg. mesa depends on llvm, which could remove all the casts in `compiler_mesa.py`)

### DIFF
--- a/tinygrad/runtime/support/c.py
+++ b/tinygrad/runtime/support/c.py
@@ -1,10 +1,11 @@
 import ctypes, functools, sys
 from typing import TYPE_CHECKING
 
-def _do_ioctl(__idir, __base, __nr, __struct, __fd, **kwargs):
+def _do_ioctl(__idir, __base, __nr, __struct, __fd, __payload=None, **kwargs):
   import tinygrad.runtime.support.hcq as hcq, fcntl
   ioctl = __fd.ioctl if isinstance(__fd, hcq.FileIOInterface) else functools.partial(fcntl.ioctl, __fd)
-  if (rc:=ioctl((__idir<<30)|(ctypes.sizeof(out:=__struct(**kwargs))<<16)|(__base<<8)|__nr, out)): raise RuntimeError(f"ioctl returned {rc}")
+  if (rc:=ioctl((__idir<<30)|(ctypes.sizeof(out:=(__payload or __struct(**kwargs)))<<16)|(__base<<8)|__nr, out)):
+    raise RuntimeError(f"ioctl returned {rc}")
   return out
 
 def _IO(base, nr): return functools.partial(_do_ioctl, 0, ord(base) if isinstance(base, str) else base, nr, None)


tinygrad/runtime/ops_cl.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, dev:CLDevice, compile_key:str):
     self.dev = dev
     super().__init__(f"compile_cl_{compile_key}")
   def compile(self, src:str) -> bytes:
-    program = checked(cl.clCreateProgramWithSource(self.dev.context, 1, to_char_p_p([src.encode()]), None, status := ctypes.c_int32()), status)
+    program = checked(cl.clCreateProgramWithSource(self.dev.context, 1, ctypes.cast(to_char_p_p([src.encode()]), ctypes.POINTER(ctypes.c_char_p)), None, status := ctypes.c_int32()), status)
```
geohot: Why does this have to change?

sirhcm: The argtype changed from `ctypes.POINTER(ctypes.POINTER(ctypes.c_char))` to `ctypes.POINTER(ctypes.c_char_p)`. Will try to find a more elegant solution.

tinygrad/runtime/autogen/__init__.py
```diff
@@ -0,0 +1,118 @@
+import glob, importlib, pathlib, subprocess, tarfile
+from tinygrad.helpers import fetch, flatten
+
+def sys(cmd): return subprocess.check_output(cmd.split()).decode().strip()
```
geohot, line 4: is something like this used elsewhere? maybe this should be in helpers

tinygrad/runtime/autogen/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+
+root = (here:=pathlib.Path(__file__).parent).parent.parent.parent
```
geohot, line 6: parents[3]

tinygrad/runtime/autogen/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+  match nm:
+    case "libc": return load("libc", ["find_library('c')"], lambda: (
+      [i for i in sys("dpkg -L libc6-dev").split() if 'sys/mman.h' in i or 'sys/syscall.h' in i] +
+      ["/usr/include/string.h", "/usr/include/elf.h", "/usr/include/unistd.h", "/usr/include/asm-generic/mman-common.h"]), use_errno=True)
+    case "opencl": return load("opencl", ["find_library('OpenCL')"], ["/usr/include/CL/cl.h"])
+    case "cuda": return load("cuda", ["find_library('cuda')"], ["/usr/include/cuda.h"], args=["-D__CUDA_API_VERSION_INTERNAL"], parse_macros=False)
+    case "nvrtc": return load("nvrtc", ["find_library('nvrtc')"], ["/usr/include/nvrtc.h"])
+    case "nvjitlink": load("nvjitlink", ["find_library('nvJitLink')"], [root/"extra/nvJitLink.h"])
+    case "kfd": return load("kfd", [], ["/usr/include/linux/kfd_ioctl.h"])
+    case "nv_gpu":
```
geohot, line 36: Feel free to use more lines to make this readable

sirhcm:
Closing in favor of #13217, #13220, #13223.

PR #12792, opened 2025-10-19
qcom: add support for 7xx

nimlgen:
It started using “bindless” in some kernels buffers are replaced with textures (`TEX_BUFFER`), so the kernel needs buffer type info. Plan to extract buffer types from the binary and bind texture descriptors directly to kernels (not buffers).


geohot:
Keeping this open

